PR #3541, opened 2026-02-20
[rocprofiler-systems] Add `rocprof-sys-python` binary

kcossett-amd:
## Motivation

Add `rocprof-sys-python` binary.

AIPROFSYST-122

## Technical Details

`rocprof-sys-python` is now in `dist/rocm/bin`.

The corresponding Python package (including the native binding) is located under: 
 - Single Python version: `dist/rocm/lib/python3.<version>/site-packages/rocprofsys`
 - Multiple Python versions: `dist/rocm/lib/python/site-packages/rocprofsys`

Updated artifacts so that it respects https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md#component-types

## Test Plan

Workflows + Manual

## Test Result

Build locally on current branch. Building with a single version produces the correct output, same with multiple versions.

Running `output-linux-portable/build/dist/rocm/bin/rocprof-sys-python --help` produces the expected output.

## Submission Checklist


profiler/CMakeLists.txt
```diff
@@ -230,6 +230,34 @@ endif(THEROCK_BUILD_TESTING)
...
+    # ROOT_DIR tracking so rocprofiler-systems can auto-discover
+    # each Python version from ROCPROFSYS_PYTHON_ROOT_DIRS.
+    set(_rocprofsys_python_config
+      "${CMAKE_CURRENT_BINARY_DIR}/rocprofiler-systems/rocprofsys_python_config.cmake")
+    file(WRITE "${_rocprofsys_python_config}"
+      "# Generated by TheRock - clears Python3 cache state so rocprofiler-systems\n"
+      "# can discover each Python version independently for pybind11 modules.\n"
+      "unset(Python3_EXECUTABLE CACHE)\n"
+      "unset(_Python3_ROOT_DIR_LAST CACHE)\n"
+    )
```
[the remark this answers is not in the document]
kcossett-amd: Incorrect. Per CMake documentation for 3.25 (https://cmake.org/cmake/help/v3.25/command/file.html), it creates the file if it does not exist, including any missing directories in the path.

profiler/CMakeLists.txt
```diff
@@ -230,6 +230,34 @@ endif(THEROCK_BUILD_TESTING)
...
+    set(_rocprofsys_python_root_dirs)
+    foreach(_exe IN LISTS _detected_python_executables)
+      cmake_path(GET _exe PARENT_PATH _bin_dir)
+      cmake_path(GET _bin_dir PARENT_PATH _root_dir)
+      string(REPLACE "/opt/python/" "/opt/python-shared/" _shared_root "${_root_dir}")
+      if(EXISTS "${_shared_root}/lib" AND NOT "${_shared_root}" STREQUAL "${_root_dir}")
+        list(APPEND _rocprofsys_python_root_dirs "${_shared_root}")
+      elseif(EXISTS "${_root_dir}/lib")
+        list(APPEND _rocprofsys_python_root_dirs "${_root_dir}")
+      endif()
```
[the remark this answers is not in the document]
kcossett-amd: hmm... let me take a look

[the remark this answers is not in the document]
kcossett-amd: ̶F̶a̶i̶r̶ ̶p̶o̶i̶n̶t̶,̶ ̶l̶e̶t̶ ̶m̶e̶ ̶u̶s̶e̶ ̶`̶_̶t̶h̶e̶r̶o̶c̶k̶_̶c̶h̶e̶c̶k̶_̶p̶y̶t̶h̶o̶n̶_̶s̶h̶a̶r̶e̶d̶_̶l̶i̶b̶r̶a̶r̶y̶(̶)̶`̶ ̶t̶o̶ ̶e̶n̶s̶u̶r̶e̶ ̶t̶h̶e̶ ̶l̶i̶b̶ ̶a̶c̶t̶u̶a̶l̶l̶y̶ ̶e̶x̶i̶s̶t̶s̶ ̶i̶n̶s̶t̶e̶a̶d̶ ̶o̶f̶ ̶j̶u̶s̶t̶ ̶c̶h̶e̶c̶k̶i̶n̶g̶ ̶t̶h̶a̶t̶ ̶t̶h̶e̶ ̶l̶i̶b̶ ̶f̶o̶l̶d̶e̶r̶ ̶e̶x̶i̶s̶t̶e̶d̶.̶ ̶
 
 We can't do this. If we pre-filter and pass in an empty `ROCPROFSYS_PYTHON_ROOT_DIRS`, then the detection logic in rocprof-sys falls back to `rocprofiler_systems_find_python(_PY REQUIRED)`, which will cause build failure on static-only python. 
 When we pass candidates (what I do originally), rocprof-sys will skip if it lacks libpython. 
 
 Essentially, we let `rocprofiler-systems` handle finding python.

marbre:
@lumachad as you're in a similar boat with debuggers, could you review the approach followed in this patch?

dgaliffiAMD:
I have reviewed the artifacts and confirmed the pybind libraries are built and installed.

```
- lib/python/sites-packages/rocprof-sys/
  - libpyrocprofsys.cpython-310-x86_64-linux-gnu.so
  - libpyrocprofsys.cpython-311-x86_64-linux-gnu.so
  - libpyrocprofsys.cpython-312-x86_64-linux-gnu.so
  - libpyrocprofsys.cpython-313-x86_64-linux-gnu.so
```

My question is: do the libraries have to be installed to `lib/python3.X` rather that a monolithic `lib/python` to be compatible with the "pip install".

kcossett-amd:
Pinging @lumachad for review

dgaliffiAMD:
Hi @marbre & @lumachad, do you have any addition feedback for Kian's change? Thank you.

lumachad:
Sorry, I completely missed these pings. Let me try to understand what this is doing.

profiler/CMakeLists.txt
```diff
@@ -240,6 +240,33 @@ if(THEROCK_ENABLE_ROCPROFV3)
       string(REPLACE ";" "\\;" _ROCPROFSYS_GFX_TARGETS_ESCAPED "${THEROCK_DIST_AMDGPU_TARGETS}")
     endif()
 
+    # rocprofiler-systems requires Development.Embed (libpython) for pybind11.
+    # On manylinux, /opt/python/ lacks libpython but /opt/python-shared/ has it.
+    # For each detected Python, try its shared counterpart first, then the original.
```
lumachad: I think the assumption that /opt/python and /opt/python-shared contain a matching number of versions is wrong. We control what gets put into /opt/python-shared, but we don't control what is in /opt/python, I don't think. They have differing python versions.

Safest here would be to use the discovery code to find out what's available in /opt/python. This gets passed to the configuration via -DTHEROCK_DIST_PYTHON_EXECUTABLES and we have detection code this cmake file already calls and stores the result in _detected_python_executables.

Also, since you seem to require a lookup in /opt/python-shared, which is provided via -DTHEROCK_SHARED_PYTHON_EXECUTABLES, it might make sense to also use the python-shared detection/discovery code like in rocgdb's cmake file. See here: https://github.com/ROCm/TheRock/blob/222992d7fd5dac40fd210df3bd5164d91b276871/debug-tools/CMakeLists.txt#L186

That way you have a couple lists of what's definitely available and detected, and you can make the right choice.

What I don't fully understand here is why we're trying both /opt/python and /opt/python-shared. Is this specific to this particular subproject? It seems to me it would be best to just use /opt/python-shared if that's what gives you the required dependencies/features.

You should also keep in mind we may be building things outside of the manylinux container, and in that case we have no idea where the python executables are. Either the developer will pass that to configure via -DTHEROCK_DIST_PYTHON_EXECUTABLES and -DTHEROCK_SHARED_PYTHON_EXECUTABLES or we will have to detect it some other way.

In any case, I think we should add some status messages here stating what has been found, versions etc... so developers can quickly stare at the build and figure out what was picked and why.

profiler/CMakeLists.txt
```diff
@@ -240,6 +240,33 @@ if(THEROCK_ENABLE_ROCPROFV3)
...
+    set(_rocprofsys_python_root_dirs)
+    foreach(_exe IN LISTS _detected_python_executables)
+      cmake_path(GET _exe PARENT_PATH _bin_dir)
+      cmake_path(GET _bin_dir PARENT_PATH _root_dir)
+      string(REPLACE "/opt/python/" "/opt/python-shared/" _shared_root "${_root_dir}")
```
lumachad: So based on my big comment above, I don't think we should assume /opt/python-shared is available. We should use the detection logic for both dist python (/opt/python) and shared python (/opt/python-shared).

profiler/CMakeLists.txt
```diff
@@ -240,6 +240,33 @@ if(THEROCK_ENABLE_ROCPROFV3)
       string(REPLACE ";" "\\;" _ROCPROFSYS_GFX_TARGETS_ESCAPED "${THEROCK_DIST_AMDGPU_TARGETS}")
     endif()
 
+    # rocprofiler-systems requires Development.Embed (libpython) for pybind11.
+    # On manylinux, /opt/python/ lacks libpython but /opt/python-shared/ has it.
```
lumachad: Just a general comment that mentioning libpython here was slightly confusing, as I did not find any executable or library linked against libpython.

profiler/CMakeLists.txt
```diff
@@ -240,6 +240,33 @@ if(THEROCK_ENABLE_ROCPROFV3)
...
+    # Clear the injected executable and CMake's internal
+    # ROOT_DIR tracking so rocprofiler-systems can auto-discover
+    # each Python version from ROCPROFSYS_PYTHON_ROOT_DIRS.
+    set(_rocprofsys_python_config
+      "${CMAKE_CURRENT_BINARY_DIR}/rocprofiler-systems/rocprofsys_python_config.cmake")
+    file(WRITE "${_rocprofsys_python_config}"
+      "# Generated by TheRock - clears Python3 cache state so rocprofiler-systems\n"
+      "# can discover each Python version independently for pybind11 modules.\n"
+      "unset(Python3_EXECUTABLE CACHE)\n"
+      "unset(_Python3_ROOT_DIR_LAST CACHE)\n"
```
lumachad: I don't fully understand this section. Is there a reason why we need to clear the cache manually in here?

profiler/CMakeLists.txt
```diff
@@ -260,13 +287,16 @@ if(THEROCK_ENABLE_ROCPROFV3)
         -DROCPROFSYS_BUILD_FOR_THEROCK=ON
         -DROCPROFSYS_GFX_TARGETS="${_ROCPROFSYS_GFX_TARGETS_ESCAPED}"
         -DROCPROFSYS_BUILD_SPDLOG=OFF
+        -DROCPROFSYS_USE_PYTHON=ON
```
lumachad: Can -DROCPROFSYS_USE_PYTHON ever be OFF? Maybe if we find no usable Python's? If so, I'd throw a fatal error stating we're missing something fundamental.

profiler/CMakeLists.txt
kcossett-amd: TheRock injects the cached variables into our project when it builds it. It acts as a user override that even forces `find_package(Python3 ... EXACT)` to use the cached variable instead of searching for the proper version. 

We can move this logic to our side of the code. Though I suppose it would be nice if there was an option to prevent TheRock from injecting this (if there is one, please let me know :D)

profiler/CMakeLists.txt
kcossett-amd: Did some digging, we do not require this, but our code was looking for Development as it was written before cmake `3.18`. I have a PR that fixes this and everything worked out great. I'll be removing this part.

profiler/CMakeLists.txt
kcossett-amd: I'll be switching to using `therock_detect_python_versions` since we do not need the share lib.

profiler/CMakeLists.txt
kcossett-amd: Technically yes, but the `rocprofiler-sdk` code throws a `FATAL_ERROR` if no python versions are found. I'll still add a block for it, but unless they change their code, it'll never be reached. 

`rocprofiler-sdk` snippet:
```
  therock_detect_python_versions(_detected_python_executables _detected_python_versions)
  if(NOT _detected_python_versions)
    message(FATAL_ERROR "No Python versions detected for rocprofiler-sdk")
    set(_detected_python_versions_str "")
  else()
```

profiler/CMakeLists.txt
kcossett-amd: Originically, I was using `throck_detect_shared_python_executables`, but I had improperly handled it thinking I had a problem, so I fell back to this. 

Anyhow, I'll be switching to using `therock_detect_python_versions`, which will handle `-D THEROCK_DIST_PYTHON_EXCUTABLES` as we no longer require a shared lib.

kcossett-amd:
Requires https://github.com/ROCm/rocm-systems/pull/3859

lumachad:
I suppose we're waiting for a rocm-systems bump to pick https://github.com/ROCm/rocm-systems/pull/3859 up before we move forward with this one?

kcossett-amd:
> I suppose we're waiting for a rocm-systems bump to pick [ROCm/rocm-systems#3859](https://github.com/ROCm/rocm-systems/pull/3859) up before we move forward with this one?

Yes. Unfortunately, #3904 has failing tests so who knows when the next bump will be.

lumachad:
> > I suppose we're waiting for a rocm-systems bump to pick [ROCm/rocm-systems#3859](https://github.com/ROCm/rocm-systems/pull/3859) up before we move forward with this one?
> 
> Yes. Unfortunately, #3904 has failing tests so who knows when the next bump will be.

Understood. Please ping me (directly or indirectly) when this gets moving again. FTR, the latest changes look reasonable to me.

marbre (commented):
> > > I suppose we're waiting for a rocm-systems bump to pick [ROCm/rocm-systems#3859](https://github.com/ROCm/rocm-systems/pull/3859) up before we move forward with this one?
> > 
> > 
> > Yes. Unfortunately, #3904 has failing tests so who knows when the next bump will be.
> 
> Understood. Please ping me (directly or indirectly) when this gets moving again. FTR, the latest changes look reasonable to me.

Please request a re-review once the build is green. I'd like to take a look at the artifacts produced as I am not able to fully parse what

> If a single version of python is used to build:
>    Respective python lib in dist/rocm/lib/python3.<version>/site-packages/rocprofsys
>
> If multiple versions are used:
>    Python lib under dist/rocm/lib/python/site-packages/rocprofsys

in the PR description means. Normally `python/site-packages` does not exist as far as I am aware and we only use a _single_ Python version to build the Python agnostic wheels. However, the above sounds like this might not be version agnostic any longer and it might need more changes (without me looking at what profiler does in its build).

kcossett-amd:
I've taken a look at the artifacts and all seems in place. In our `lib` artifact, the required python bindings are under:
`profiler/rocprofiler-systems/stage/lib/python/site-packages/rocprofsys`

lumachad:
I'm happy with this version, but CI seems to have skipped testing again.

kcossett-amd:
> I'm happy with this version, but CI seems to have skipped testing again.

Testing is specific to rocprofiler-systems via the `test:rocprofiler-systems` labels.

kcossett-amd:
Pinging @marbre for re-review

kcossett-amd:
> > I've taken a look at the artifacts and all seems in place. In our `lib` artifact, the required python bindings are under: `profiler/rocprofiler-systems/stage/lib/python/site-packages/rocprofsys`
> 
> I didn't noticed this earlier, but why is a _binary_ in the `lib` artifact? It should rather be in run, shouldn't it? I further checked the Python packages and it is not included and I still think the layering might not work in this case as expected. Thus cannot test how this behaves if they go into the wheels.

For your point on the artifacts, yes, some are not where they should be. I'll be addressing that shortly.

For the python wheels, it seems we are not included in the wheels yet until #3720 (and its Phase 3) are merged.

kcossett-amd:
Awaiting `rocm-systems` bump #4193. It fixes the sanity checks.

kcossett-amd:
> could you please clarify a bit your pr description?
> 
> i dont fully get this statement:
> 
> > Respective rocprof-sys-python- scripts are never built as the logic for that lies in rocprof-sys-python-tests.cmake, which is never reached when ROCPROFSYS_BUILD_FOR_THEROCK=ON.
> 
> while at the same time later it is said "pybind libraries are built and installed"? it would be good explain this a bit more for non-rocprofsys experts.

For the `rocprof-sys-python-<version>` scripts: these are test helper scripts generated by `rocprof-sys-python-tests.cmake`. They're not part of a normal install; they only exist for running the Python test suite. I just wanted to call that out in case anyone throught that they were missing... which I did a poor job of.

For the "pybind libraries are built and installed", what I meant was the `rocprofsys` python package (including the pybind11 binding, being the `libpyrocprofsys...`) is build and installed (see updated PR desc).

Issue #3288, opened 2026-02-06
[ci] Re-enable `gfx950-dcgpu` test machines for rocm-libraries and TheRock when enough capacity is back


geomin12:
Currently, we only have 8 `gfx950-dcgpu` test runners, causing a major queue times. For TheRock, `gfx950-dcgpu` test runners are necessary for multi-arch and release workflows.

We will still get signal from post submit and CI nightly gfx950-dcgpu runners, but to remediate queue times, we will turn off gfx950-dcgpu for rocm-libraries

This issue is to re-enable for rocm-libraries once we get enough capacity

geomin12:
Seeing times up to ~ 30h. We are disabling for multi-arch CI

archana-ramalingam:
The tensilelite test component (PR #5529) has 200 gfx950 GPU subtile tests skipped in PR CI due to the gfx950-dcgpu runner exclusion.
These subtile tests are part of customer delivery and must be tracked directly under `rocm-libraries`. 
Prioritizing gfx950-dcgpu runner re-enablement would unblock this coverage. 
Is there a timeline for this? We need to know when gfx950-dcgpu runners will be more widely available.

PR #3542, opened 2026-02-20
[ci] Disabling gfx950-dcgpu for multi-arch tests

geomin12:
Due to large queue times for gfx950-dcgpu, we are excluding gfx950 for multi arch tests. Tagged issue #3288

Here's a cancelled run of gfx950 and gfx94X running properly during a normal CI run: https://github.com/ROCm/TheRock/actions/runs/22237123949 (check CI step summary)

Here's a cancelled run of gfx950 and gfx94X running and gfx950 being ignored properly during multi-arch CI run:
https://github.com/ROCm/TheRock/actions/runs/22237181472 (check CI step summary)

Adding skip-ci label as build is unrelated and no need to waste resources

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -78,7 +78,7 @@ jobs:
         # TODO(TheRock#3288): Re-enable gfx950-dcgpu runners once we get more capacity
         exclude:
           - family_info:
-            amdgpu_family: gfx950-dcgpu
+              amdgpu_family: gfx950-dcgpu
 
     uses: ./.github/workflows/test_artifacts.yml
     with:


.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -75,6 +75,11 @@ jobs:
       fail-fast: false
       matrix:
         family_info: ${{ fromJSON(inputs.matrix_per_family_json) }}
+        # TODO(TheRock#3288): Re-enable gfx950-dcgpu runners once we get more capacity
+        exclude:
+          - family_info:
+              amdgpu_family: gfx950-dcgpu
```
marbre:
```suggestion
          - family_info:
            amdgpu_family: gfx950-dcgpu
```
?

geomin12: oops you are right:
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#excluding-matrix-configurations

geomin12: ah as this is nested field under family_info, this is needed

Tried here: https://github.com/ROCm/TheRock/actions/runs/22237423099 with commit: https://github.com/ROCm/TheRock/commit/73ffa0ce752d188ebc1d74c24ecbcaabdb5b043c and got error `Error when evaluating 'strategy' for job 'test_artifacts_per_family'. ROCm/TheRock/.github/workflows/multi_arch_ci_linux.yml@73ffa0ce752d188ebc1d74c24ecbcaabdb5b043c (Line: 81, Col: 13): Matrix exclude key 'amdgpu_family' does not match any key within the matrix`

However here: https://github.com/ROCm/TheRock/actions/runs/22237582449 with commit: https://github.com/ROCm/TheRock/commit/636fcce084431b90e3fa350c25135d0d6e37417d, worked fine! 

I'll keep as is as it works in tests

amd-justchen: Thanks copilot!

---
It **does** need to be indented (more than `family_info:`), because `amdgpu_family` is a field **inside** the `family_info` mapping.

This is correct:

```yaml
exclude:
  - family_info:
      amdgpu_family: gfx950-dcgpu
```

If you wrote it like this (same indent as `family_info:`), it would be wrong because it makes `amdgpu_family` a *sibling* key of `family_info`, not nested within it:

```yaml
exclude:
  - family_info:
    amdgpu_family: gfx950-dcgpu   # wrong
```

Rule of thumb: anything “contained under” `family_info:` must be indented further than `family_info:`.

PR #3543, opened 2026-02-20
Make sure patchelf/install_name_tool is available

marbre:
patchelf or rather install_name_tool is not only required for `THEROCK_BUNDLE_SYSDEPS=ON` but also for building rocGDB with sysdeps disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -342,7 +342,6 @@ if(THEROCK_BUNDLE_SYSDEPS OR THEROCK_ENABLE_ROCGDB)
     if(NOT PATCHELF)
       message(FATAL_ERROR "Building with THEROCK_BUNDLE_SYSDEPS=ON or THEROCK_ENABLE_ROCGDB=ON on Linux requires `patchelf`")
     endif()
-    message(FATAL_ERROR "!/!")
   elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     # macOS uses install_name_tool (from Xcode) instead of patchelf
     find_program(INSTALL_NAME_TOOL install_name_tool)


CMakeLists.txt
```diff
@@ -330,6 +330,28 @@ if(CMAKE_CXX_VISIBILITY_PRESET)
...
+# install_name_tool on macOS to set RPATH.
+################################################################################
+
+if(THEROCK_BUNDLE_SYSDEPS OR THEROCK_ENABLE_ROCGDB)
+  if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    find_program(PATCHELF patchelf)
+    if(NOT PATCHELF)
+      message(FATAL_ERROR "Building with THEROCK_BUNDLE_SYSDEPS=ON or THEROCK_ENABLE_ROCGDB=ON on Linux requires `patchelf`")
+    endif()
+    message(FATAL_ERROR "!/!")
```
ScottTodd: Stray debug log?
https://github.com/ROCm/TheRock/actions/runs/22241260424/job/64345172580?pr=3543#step:9:98

```
-- Disabled features:
--   * IREE_COMPILER (-DTHEROCK_ENABLE_IREE_COMPILER=OFF)
--   * FUSILLI_PLUGIN (-DTHEROCK_ENABLE_FUSILLI_PLUGIN=OFF)
CMake Error at CMakeLists.txt:345 (message):
  !/!


-- Configured AMDGPU Targets:
--   * gfx942 : MI300A/MI300X CDNA
--   * Dist bundle: gfx94X-dcgpu
-- Configuring incomplete, errors occurred!
```

CMakeLists.txt
```diff
@@ -330,6 +330,28 @@ if(CMAKE_CXX_VISIBILITY_PRESET)
...
+  elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    # macOS uses install_name_tool (from Xcode) instead of patchelf
+    find_program(INSTALL_NAME_TOOL install_name_tool)
+    if(NOT INSTALL_NAME_TOOL)
+      message(FATAL_ERROR "Building with THEROCK_BUNDLE_SYSDEPS=ON or THEROCK_ENABLE_ROCGDB=ON on macOS requires `install_name_tool` (install Xcode Command Line Tools)")
```
lumachad, line 349: No ROCgdb on macOS, so I think we can remove that.

CMakeLists.txt
```diff
@@ -330,6 +330,28 @@ if(CMAKE_CXX_VISIBILITY_PRESET)
   list(APPEND DEFAULT_CMAKE_ARGS ${CMAKE_CXX_VISIBILITY_PRESET})
 endif()
 
+################################################################################
+# patchelf / install_name_tool
+# Building with bundled sysdep or rocGDB requires patchelf on Linux and
+# install_name_tool on macOS to set RPATH.
+################################################################################
+
+if(THEROCK_BUNDLE_SYSDEPS OR THEROCK_ENABLE_ROCGDB)
```
lumachad, line 339: Grepping our submodules, I did spot rocm-systems/projects/rocprofiler-compute/CMakeLists.txt seemingly trying to install patchelf through pip:

rocm-systems/projects/rocprofiler-compute/CMakeLists.txt:            execute_process(COMMAND \${VENV_PYTHON} -m pip install patchelf==0.17.2.4 RESULT_VARIABLE PATCHELF_INSTALL_RESULT)

Maybe it should also use PATCHELF? And it would be an argument in favor of unconditionally having PATCHELF, and not restricting it to BUNDLE_SYSDEPS and ROCGDB.

CMakeLists.txt
marbre: Indeed. That shouldn't go in here. Thanks for the catch.

CMakeLists.txt
marbre: The patch just shuffles around the code that went in with commit dc7a8ec. Even though there might no macOS support yet there is general ennoblement work for macOS support so I would like to keep it here.

CMakeLists.txt
marbre: rocprofiler and its deps are a different thing... I think this needs separate issues. For now we only use the system provided patchelf in for rocgdb and bundled sysdeps. I agree that we may want to further rework this but this will need more work in rocprofiler first or rather I need to take a closer look to rocprofiler. How about to do so in a follow up?

CMakeLists.txt
lumachad: That's fine by me.

CMakeLists.txt
lumachad: Agreed with your assessment. This can be done in a separate PR with the involvement of the profiler folks.

lumachad (approved):
LGTM.

PR #3440, opened 2026-02-17
Reapply "Move non-sysdep third-party projects after base and compiler"

ScottTodd:
## Motivation

This reverts https://github.com/ROCm/TheRock/pull/3439 to reapply https://github.com/ROCm/TheRock/pull/2045 and fix https://github.com/ROCm/TheRock/issues/2003.

## Technical Details

The original PR was missing changes in `BUILD_TOPOLOGY.toml`. If we switch more third-party subprojects to use the amd-llvm toolchain they will need similar changes.

I'm also testing a follow-up change to migrate (nearly) all other third-party subprojects to use amd-llvm.

We could re-evaluate the build stages as "Foundation" gets smaller and "Compiler Runtime" gets larger (could merge them or split out a new stage for third party deps that use the compiler 🤔).

## Test Plan

Run multi-arch CI on this branch: https://github.com/ROCm/TheRock/actions/runs/22112365715
- [x] check logs for "Foundation" stage (should NOT build spdlog)
- [x] check logs for "Compiler Runtime" stage (should build spdlog and upload the artifact)
- [x] check "fetch inbound artifacts" logs for "Profiler Apps" stage (should fetch spdlog artifact)

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3548, opened 2026-02-21
Clarify the difference between sysdeps and other third-party deps

ScottTodd:
## Motivation

Following up on discussion at https://github.com/ROCm/TheRock/pull/3440#discussion_r2824891404.

The descriptions in `BUILD_TOPOLOGY.toml` for `third-party-sysdeps` and `third-party-libs` were misleading. This adds more specific background details explaining what qualifies as a sysdep.

## Submission Checklist

### DIFF
--- a/third-party/sysdeps/README.md
+++ b/third-party/sysdeps/README.md
@@ -6,6 +6,7 @@ symbol versioning, and installation into `lib/rocm_sysdeps/` — so that ROCm ca
 ship private copies without conflicting with system-installed versions.
 
 This is distinct from the other third-party libraries under `third-party/` (fmt,
-spdlog, flatbuffers, etc.), which are build dependencies built as normal CMake
-targets. See [Dependencies](/docs/development/dependencies.md) for
-the full explanation of the two categories.
+spdlog, flatbuffers, etc.), which are build dependencies not given special
+treatment for portable distribution. See
+[Dependencies](/docs/development/dependencies.md) for the full explanation of
+the two categories.


ScottTodd:
> Thanks for picking this up. The _built as normal CMake targets_ for other third-party libraries confuses me a bit. At least from a CMake perspective I don't think one is more special than the other.

"Regular" third-party subprojects seem pretty simple, like https://github.com/ROCm/TheRock/blob/998bf613eba72f39cd32e04fd2b142c5b1a28086/third-party/eigen/CMakeLists.txt#L1-L19

The sysdeps subprojects all do this dance:
* https://github.com/ROCm/TheRock/blob/998bf613eba72f39cd32e04fd2b142c5b1a28086/third-party/sysdeps/linux/elfutils/CMakeLists.txt#L1-L6
* https://github.com/ROCm/TheRock/blob/998bf613eba72f39cd32e04fd2b142c5b1a28086/third-party/sysdeps/linux/elfutils/CMakeLists.txt#L54-L60

Is that just because they use `make`, `meson`, etc.? Here's one that builds with CMake but follows the same pattern: https://github.com/ROCm/TheRock/blob/main/third-party/sysdeps/common/bzip2/CMakeLists.txt.

Should we document that pattern too? Do you want that in this PR or follow-up work?

marbre:
I think you have hit a rather complicated sysdep with `elfutils`.

* host-blas has a custom name as well even though the build is CMake based:
https://github.com/ROCm/TheRock/blob/e40a6da680f0a83fd931a5285ee1dc7cde5fd8c8/third-party/host-blas/CMakeLists.txt#L109-L130
* Another example is boost which has a custom CMake config with https://github.com/ROCm/TheRock/blob/e40a6da680f0a83fd931a5285ee1dc7cde5fd8c8/third-party/boost/cmake_project/CMakeLists.txt in addition to TheRock config (https://github.com/ROCm/TheRock/blob/e40a6da680f0a83fd931a5285ee1dc7cde5fd8c8/third-party/boost/CMakeLists.txt.
* Some of the deps are also patched, see https://github.com/ROCm/TheRock/blob/e40a6da680f0a83fd931a5285ee1dc7cde5fd8c8/third-party/grpc/patch_install_linux.sh (side note, we should avoid patching with sed ...).


What makes the sysdeps really special are the custom so names and version map with private symbol versions.

ScottTodd:
How's this now? I adjusted the language around "normal cmake builds".

marbre:
> How's this now? I adjusted the language around "normal cmake builds".

Thanks for iterating :)

Issue #3399, opened 2026-02-12
[Multi-arch] Add stage-aware "use_prebuilt_artifacts" configuration to CI workflows


ScottTodd:
I'd like to optimize which jobs run for a few specific CI usage scenarios with some generic tooling:

1. PRs in https://github.com/ROCm/rocm-libraries should be able to use prebuilt artifacts for the "foundation" and "compiler-runtime" stages
2. PRs in TheRock that only update [`build_tools/packaging/`](https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/) should be able to use prebuilt artifacts for _all_ stages
3. PRs in TheRock that only update [`external-builds/`](https://github.com/ROCm/TheRock/tree/main/external-builds) should be able to use prebuilt rocm python packages

More generally for the CMake build system, the build topology tracks "source sets", which we can use to compute the list of affected and unaffected artifact groups for a given set of modified paths: https://github.com/ROCm/TheRock/blob/fb66bc4e17443e0b3fd28d357fa6773eaa0f247b/BUILD_TOPOLOGY.toml#L74-L124 https://github.com/ROCm/TheRock/blob/fb66bc4e17443e0b3fd28d357fa6773eaa0f247b/BUILD_TOPOLOGY.toml#L248-L259

Building stage-aware prebuilt artifacts CI configuration would involve changes across these files:

* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_ci.py
  * Compute which stages to use prebuilt artifacts for
* https://github.com/ROCm/TheRock/blob/main/build_tools/find_artifacts_for_commit.py and https://github.com/ROCm/TheRock/blob/main/build_tools/find_latest_artifacts.py
  * Find a workflow run to pull baseline artifacts from
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/setup.yml
  * Propagate the selection of which stages should use prebuilt artifacts (and from which workflow run id)
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml
  * Propagate the selection of which stages should use prebuilt artifacts (and from which workflow run id)
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux.yml
  * Skip or no-op workflow jobs if using prebuilt artifacts for that stage

ScottTodd:
I'm making some good progress on this with some prototypes:

[...]

Issue #3334, opened 2026-02-09
[Multi-arch] Add multi-arch release pipelines


ScottTodd:
Add equivalents for these workflows that build artifacts in multiple stages and do any necessary packaging / postprocessing for multi-arch packages:

* https://github.com/ROCm/TheRock/blob/main/.github/workflows/release_portable_linux_packages.yml
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/release_windows_packages.yml

Notes: 
* Model the "build artifacts" jobs after https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux.yml
* First we'll need the ability to build "dev" release here in TheRock, then later we'll also want nightly and prerelease releases buildable over in https://github.com/ROCm/rockrel.
* Add a Python package release pipeline. Can be based on the old Python packaging until kpack is enabled.
* Avoid name collisions between "single stage" releases and "multi-arch" releases
  * Tarball release pipelines optional?

ScottTodd:
I'm starting work on this, notes: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/multi-arch-releases.md

Design goals:
* Release workflows run the same jobs as CI workflows, uploading intermediate files to `therock-*-artifacts` buckets. Steps that are unique to release jobs then _copy_ files from those artifact buckets to the package buckets like `therock-*-tarball`. See the docs: https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#release-buckets
* Release workflows run in https://github.com/ROCm/rockrel, not this repository

Issues to resolve:
* IAM roles and bucket selection: CI workflows currently assume CI buckets. https://github.com/ROCm/TheRock/pull/3561 from @raramakr and @nunnikri just added a "Determine IAM role" step that bridges CI and release buckets / IAM roles to [`.github/workflows/multi_arch_build_native_linux_packages.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_native_linux_packages.yml) which we can generalize. I think we can compute the bucket and IAM role upfront in the setup job and pass it explicitly to _all_ workflows, rather than recompute in each job.
* Index generation still needs to move server-side for some file types (https://github.com/ROCm/TheRock/issues/3344)
* TBD how existing releases and multi-arch releases will overlap (v3 subfolder? multi-arch subfolder? replace workflows in one step with no conflicts?)

[...]

PR #3561, opened 2026-02-23
Add multi-architecture packaging support

raramakr:
Enable building arch-specific package variants by iterating over
  multiple gfx architectures per base package. Non-versioned packages
  are now only created for generic architecture in multi-arch mode.

  Workflow changes:
  - Add multi_arch_build_native_linux_packages.yml reusable workflow
  - Fetches artifacts for all GPU families (gfx94X-dcgpu, gfx120X-all, etc.)
  - Builds unified DEB/RPM packages containing all architectures
  - Implements comprehensive S3 bucket selection logic with decision tree
  - Supports multiple bucket types: CI artifacts, release packages, internal
  - Determines appropriate IAM roles based on build context
  
  build_package.py enhancements:
  - Add --enable-multi-arch and multi-value --target args to build_package.py
  - Add resolve_versioned_dependencies() and get_dependency_list_for_multiarch() to packaging_utils.py to handle dependency resolution for generic, arch-specific, and single-arch metapackages
  - Add expand_metapackage_to_all_archs() for generic metapackage dependencies
  - Update is_gfxarch_package() and filter_components_fromartifactory() to support multi-arch mode
  - Move clean_package_build_dir() from build_package.py to packaging_utils.py
  - Fix build summary counts by tracking failed packages explicitly instead of inferring from a hardcoded 2x multiplier, which caused negative failure counts
  - Add normalize_target_list() function for flexible input parsing
  - Fix typo in artifact directory error message

upload_package_repo.py improvements:
- Add --s3-prefix parameter for explicit S3 prefix override
- Add "ci" job type to release_type(dev/nightly/prerelease/ci)

Test Output:
https://github.com/ROCm/TheRock/actions/runs/22646265113/job/65634920221
https://github.com/ROCm/TheRock/actions/runs/22646265113/job/65634920219

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -83,8 +83,8 @@ jobs:
             --source-run-id=${{ inputs.baseline_run_id }} \
             --stage="${{ inputs.prebuilt_stages }}" \
             --amdgpu-families="${{ inputs.dist_amdgpu_families }}"
-  
-   build_multi_arch_stages:
+
+  build_multi_arch_stages:
     name: Build Multi-Arch Stages
     needs: copy_prebuilt_stages
     if: ${{ !cancelled() && !failure() }}
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -845,9 +845,9 @@ def run(args: argparse.Namespace):
 
             pkg_info = get_package_info(pkg_name)
             # Check the package is marked as gfxarch package OR meta package
-            if is_gfxarch_package(
-                pkg_info, config.enable_kpack
-            ) or is_meta_package(pkg_info):
+            if is_gfxarch_package(pkg_info, config.enable_kpack) or is_meta_package(
+                pkg_info
+            ):
                 # Use all gfxarch values
                 loop_list = gfxarch_list + [default_gfx_arch]
             else:
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -651,10 +651,7 @@ def filter_components_fromartifactory(
     if enable_kpack:
         dir_suffix = (
             gfx_arch
-            if (
-                is_gfxarch_package(pkg_info, enable_kpack)
-                and gfx_arch != GFX_GENERIC
-            )
+            if (is_gfxarch_package(pkg_info, enable_kpack) and gfx_arch != GFX_GENERIC)
             else "generic"
         )
     else:
@@ -880,9 +877,7 @@ def get_dependency_list_for_multiarch(pkg_info, dep_key, config: PackageConfig):
         return [
             dep
             for dep in dep_list
-            if not is_gfxarch_package(
-                get_package_info(dep) or {}, config.enable_kpack
-            )
+            if not is_gfxarch_package(get_package_info(dep) or {}, config.enable_kpack)
         ]
     elif config.enable_kpack and config.gfx_arch != GFX_GENERIC:
         # Gfx-specific package in multi-arch mode:


build_tools/packaging/linux/package.json
```diff
@@ -1720,6 +1720,12 @@
             "Components": [
               "test"
             ]
+          },
+          {
+            "Name": "rccl-tests",
+            "Components": [
+              "test"
+            ]
```
marbre, line 1754: I don't think we should add rccl tests. These are right now build without Open MPI support enabled.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,271 @@
...
+            echo "✓ Using default CI bucket: ${S3_BUCKET}"
+          fi
+
+          echo "S3 bucket: ${S3_BUCKET}"
+          echo "S3 prefix: ${S3_PREFIX}"
+          echo "Job type: ${JOB_TYPE}"
+
+          echo "s3_bucket=${S3_BUCKET}" >> $GITHUB_OUTPUT
+          echo "s3_prefix=${S3_PREFIX}" >> $GITHUB_OUTPUT
+          echo "job_type=${JOB_TYPE}" >> $GITHUB_OUTPUT
```
ScottTodd: Read the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash. This needs to be in a script (with unit tests).

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,271 @@
...
+            echo "Fetching artifacts for family: ${family}"
+            python ./build_tools/fetch_artifacts.py \
+              --run-id=${{ env.ARTIFACT_RUN_ID }} \
+              --run-github-repo=${{ github.repository }} \
+              --artifact-group="${family}" \
+              --output-dir="${{ env.ARTIFACTS_DIR }}"
+          done
+
+          echo "Fetched artifacts:"
+          ls -lh "${{ env.ARTIFACTS_DIR }}" || echo "No artifacts found"
```
ScottTodd: No need for a loop here, or any of this inlined bash. The `fetch_artifacts.py` script already accepts a list: https://github.com/ROCm/TheRock/blob/222992d7fd5dac40fd210df3bd5164d91b276871/build_tools/fetch_artifacts.py#L311-L316

I have some draft code on local branches that either accepts semicolon-separated lists too, or converts arguments to just use semicolon-separated lists instead of comma-separated. Could do the same here as needed.

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,271 @@
...
+            --dest-dir ${{ env.PACKAGE_DIST_DIR }} \
+            --rocm-version  ${{ inputs.rocm_version }} \
+            --target "${{ env.DIST_AMDGPU_FAMILIES }}" \
+            --artifacts-dir ${{ env.ARTIFACTS_DIR }} \
+            --pkg-type ${{ inputs.native_package_type }} \
+            --version-suffix ${{ env.ARTIFACT_RUN_ID }} \
+            --enable-multi-arch
+
+      - name: Install AWS CLI
+        run: bash ./dockerfiles/install_awscli.sh
```
ScottTodd: We're moving to use the boto3 python module exclusively, no awscli. This is fine for now to keep the scope of this PR small though.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,271 @@
...
+      ref:
+        description: "Branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow"
+        type: string
+        required: false
+
+permissions:
+  id-token: write
+  contents: read
+
+run-name: Build multi-arch native Linux packages (${{ inputs.dist_amdgpu_families }}, ${{ inputs.rocm_version }}, ${{ inputs.native_package_type }}, ${{ inputs.release_type }})
```
ScottTodd: This workflow only has a `workflow_call` trigger, so this `run-name` is not used anywhere.

If keeping it for `workflow_dispatch`, the run name looks very long. Does it fit in the github UI?

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,6 +37,29 @@ permissions:
   contents: read
 
 jobs:
+  compute_native_package_versions:
+    name: Compute Native Package Versions
+    runs-on: ubuntu-latest
```
ScottTodd: Read the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#pin-action-runs-on-labels-to-specific-versions

```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,6 +37,29 @@ permissions:
...
+        run: |
+          # Compute both versions in same step to ensure consistent timestamp
+          deb_version=$(python ./build_tools/compute_rocm_package_version.py --package-type deb --release-type dev)
+          rpm_version=$(python ./build_tools/compute_rocm_package_version.py --package-type rpm --release-type dev)
+
+          echo "deb_version=${deb_version}" >> $GITHUB_OUTPUT
+          echo "rpm_version=${rpm_version}" >> $GITHUB_OUTPUT
+
+          echo "DEB package version: ${deb_version}"
+          echo "RPM package version: ${rpm_version}"
```
ScottTodd: Let's change the script to output both versions and write to github outputs itself. Then this can just be a single line

should end up looking something like

```suggestion
      - name: Compute package versions for both deb and rpm
        id: compute_versions
        run: |
          python ./build_tools/compute_rocm_package_version.py \
              --package-type deb,rpm \
              --release-type dev
```

Then, my second comment is more clear - why `--release-type dev`, when the workflow also says "Empty string for CI builds."?
```yml
      release_type:
        description: "The type of release to build ('dev', 'nightly', or 'release'). Empty string for CI builds."
        required: false
        type: string
        default: ""
```

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -94,3 +117,35 @@ jobs:
...
+    needs: [build_multi_arch_stages, compute_native_package_versions]
+    # Build native packages in parallel with tests after multi-arch build completes
+    # Skip if we're using prebuilt artifacts or expecting build failure
+    if: >-
+      ${{
+        !failure() &&
+        !cancelled() &&
+        inputs.use_prebuilt_artifacts == 'false' &&
+        inputs.expect_failure == false
+      }}
```
ScottTodd: I don't follow this comment. I'm working on https://github.com/ROCm/TheRock/issues/3399, and there won't be a single "use prebuilt artifacts", there will be prebuilt _stages_. Regardless, if artifacts are prebuilt, we still want to build packages (unless we choose to skip them).

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -94,3 +117,35 @@ jobs:
...
+    with:
+      dist_amdgpu_families: ${{ inputs.dist_amdgpu_families }}
+      artifact_run_id: ${{ inputs.artifact_run_id }}
+      rocm_version: >-
+        ${{
+          matrix.package_type == 'deb'
+          && needs.compute_native_package_versions.outputs.deb_version
+          || needs.compute_native_package_versions.outputs.rpm_version
+        }}
+      native_package_type: ${{ matrix.package_type }}
```
ScottTodd: Overloading `rocm_version` here is confusing. I think that might also be possible to do via the matrix itself.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,271 @@
...
+
+jobs:
+  build_native_packages:
+    name: Build ${{ inputs.native_package_type }} packages
+    strategy:
+      fail-fast: false
```
ScottTodd: strategy does nothing without a `matrix`

.github/workflows/multi_arch_ci_linux.yml
ScottTodd, line 1: We could leave the `multi_arch_ci_linux.yml` changes out of this PR to keep the scope smaller - add `.github/workflows/multi_arch_build_native_linux_packages.yml` with a `workflow_dispatch` trigger and iterate from there? We don't yet have multi-arch _releases_, so I see why you'd reach for _ci_, but this PR is quite broad in scope and difficult to review as-is.

build_tools/packaging/linux/build_package.py
```diff
@@ -132,13 +135,21 @@ def create_versioned_deb_package(pkg_name, config: PackageConfig):
...
-        sys.exit(f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting")
+        if config.enable_multi_arch:
+            print(
+                f"ERROR: {pkg_name}: Empty sourcedir_list and not a meta package, skipping"
+            )
+            return []
+        else:
+            sys.exit(
+                f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting"
+            )
```
ScottTodd, line 152: Seems like these errors should raise exceptions, not print/return or sys.exit?

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#fail-fast-behavior

build_tools/packaging/linux/build_package.py
```diff
@@ -563,22 +581,36 @@ def generate_spec_file(pkg_name, specfile, config: PackageConfig):
...
+        # Warn if we have no artifacts for non-meta packages
+        if not sourcedir_list and not is_meta:
+            if config.enable_multi_arch:
+                print(
+                    f"WARNING: {pkg_name}: Empty sourcedir_list and not a meta package, creating empty RPM"
+                )
+            else:
+                sys.exit(
+                    f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting"
+                )
```
ScottTodd, line 595: Same comment about error handling (raise exceptions?)

build_tools/packaging/linux/build_package.py
```diff
@@ -809,18 +857,48 @@ def run(args: argparse.Namespace):
...
+                    # Add failed architecture variant to failed list
+                    variant_name = (
+                        f"{pkg_name}-{gfxarch}"
+                        if gfxarch != default_gfx_arch
+                        else pkg_name
+                    )
+                    failed_pkglist.append(variant_name)
+
+            if not pkg_built:
+                failed_pkglist.append(pkg_name)
```
ScottTodd: Is there a chance for packages to be added to `failed_pkglist` twice here?

build_tools/packaging/linux/build_package.py
```diff
@@ -263,6 +274,11 @@ def generate_rules_file(pkg_info, deb_dir, config: PackageConfig):
     # Get package name for changelog installation
     pkg_name = update_package_name(pkg_info.get("Package"), config)
 
+    # Disable debian dh_strip for multi-arch buillds
```
ScottTodd: nit: typo
```suggestion
    # Disable debian dh_strip for multi-arch builds
```

build_tools/packaging/linux/build_package.py
```diff
@@ -263,6 +274,11 @@ def generate_rules_file(pkg_info, deb_dir, config: PackageConfig):
...
+    # TODO: Fix required for dh_strip error in multi-arch builds
+    if config.enable_multi_arch:
+        disable_dh_strip = "True"
```
ScottTodd: Use an actual boolean instead of `"True"`?

build_tools/packaging/linux/build_package.py
raramakr: Done

build_tools/packaging/linux/build_package.py
raramakr: Done

build_tools/packaging/linux/build_package.py
raramakr: Removed the redundant code , which may add duplicate
if not pkg_built:
                failed_pkglist.append(pkg_name)

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Excellent catch Scott. Will keep it for adding workflow_dispatch later. As it has been useful in rebuilding native packages in case of any failures. For now made the name shorter

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Yes good catch. Removed it

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Done

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: We used the loop as --artifact-group is a mandatory parameter. 
As a workaround passed both --artifact-group and --amdgpu-targets for time being. Which is not ideal. 
Will fix the fetch_artifact.py script and update it soon.
A normalization of gfx-target input was done in this PR itself in build_package.py. May be we can do similar approach in other scripts.

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Sure will do that as a follow up change

.github/workflows/multi_arch_ci_linux.yml
nunnikri: Yes have reverted the changes in this yaml for timebeing. Will bring it later with all the concerns below addressed.

build_tools/packaging/linux/build_package.py
raramakr: Once all the artifacts are available , will enable exceptions.
We do see some artifacts are not having the architecture specific artifacts( ex:rccl_lib_gfx1151 is missing)

build_tools/packaging/linux/build_package.py
raramakr: Once all the artifacts are available , will enable exceptions.
We do see some artifacts are not having the architecture specific artifacts( ex:rccl_lib_gfx1151 is missing)

build_tools/packaging/linux/package.json
raramakr: Disabled rccl-test packaging

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Done.

raramakr:
> Please cross check if the failing CI checks are known issues.

The failing CIs (2) are unnrelated to this PR

.github/workflows/multi_arch_build_native_linux_packages.yml
ScottTodd: We should fix `fetch_artifacts.py` to support downloading artifacts for all GPU families without needing to pass an explicit list like this. We really need to follow the style guide here and not workaround issues in scripts with inlined bash in workflow files.

It looks like the `test_artifacts_structure.yml` workflow is currently only testing generic artifacts on multi-arch CI (example: https://github.com/ROCm/TheRock/actions/runs/23785173360/job/69342666772, compared to https://github.com/ROCm/TheRock/actions/runs/23785173293/job/69329661134), so updating the script would help multiple workflows.

nunnikri: Yes agree @ScottTodd. Will fix the fetch_artifacts.py. 
Is it better to add support --amdgpu-targets=all or any other suggestions?

ScottTodd: I could also see a case for fetching all artifacts by default. The current default of "only generic artifacts" isn't often what developers actually want.

nunnikri: @raramakr lets take this as priority.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -0,0 +1,187 @@
...
+            # CI builds use the shared CI role (for all artifact buckets)
+            IAM_ROLE="arn:aws:iam::692859939525:role/therock-ci"
+            echo "✓ Using CI role: ${IAM_ROLE}"
+          else
+            # Release builds use release-type-specific roles (for package buckets)
+            IAM_ROLE="arn:aws:iam::692859939525:role/therock-${JOB_TYPE}"
+            echo "✓ Using release-type role: ${IAM_ROLE}"
+          fi
+
+          echo "iam_role=${IAM_ROLE}" >> $GITHUB_OUTPUT
```
ScottTodd, line 131: FYI. I'm expecting to refactor this as part of standing up multi-arch releases (https://github.com/ROCm/TheRock/issues/3334). I think we can move this into the setup job and then plumb it through to this and other jobs via workflow inputs, rather than recompute in each job that needs to know a bucket and IAM role for that bucket

.github/workflows/multi_arch_build_native_linux_packages.yml
ScottTodd: See also this pattern used by python package builds: https://github.com/ROCm/TheRock/blob/e5e7d6f7ac1e382de0ad77720680206f3cf8c06d/.github/workflows/build_portable_linux_python_packages.yml#L84-L105

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: @ScottTodd i have done some more changes in s3 config as part of https://github.com/ROCm/TheRock/pull/4310. Let me know if any comments

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Thanks for the suggestion. Have updated that in https://github.com/ROCm/TheRock/pull/4310

.github/workflows/multi_arch_build_native_linux_packages.yml
ScottTodd: I sent https://github.com/ROCm/TheRock/pull/4386. My focus is on the core rocm release pipelines then I can look more closely at what the native packages are doing.

PR #3568, opened 2026-02-23
  Improve packaging scripts: security, performance, and code quality

raramakr:
Security improvements:
  - Replace shell=True with proper command lists in upload_package_repo.py
  - Use Python file handles instead of shell redirection for dpkg-scanpackages and gzip
  - Update run_command() to accept list[str] to prevent shell injection vulnerabilities
  - Add explicit encoding="utf-8" to all text file operations for consistency

  Performance optimization:
  - Optimize artifact directory lookup in packaging_utils.py using prefix index
  - Replace nested loop with dictionary-based lookup for faster artifact matching

  Code quality improvements:
  - Reorder versioned/non-versioned package creation in build_package.py
  - Add clearer comments explaining RPATH package requirements
  - Fix typo: "doesn not exist" → "does not exist"
  - Rename variable: skipped → skipped_list for consistency

### DIFF
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -55,11 +55,13 @@ def create_deb_package(pkg_name, config: PackageConfig):
     print_function_name()
     print(f"Package Name: {pkg_name}")
 
-    # Non-versioned packages are not required for RPATH packages
+    # Build the versioned Debian package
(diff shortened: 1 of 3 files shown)


raramakr:
The changes are incorporated as part of https://github.com/ROCm/TheRock/pull/6789

Issue #2029, opened 2025-11-05
Update PAL in rocm-systems for ubertrace profiles


jayhawk-commits:
The patch [`0002-Revert-SWDEV-543498-Some-compute-Ubertrace-profiles-.patch`](https://github.com/ROCm/TheRock/blob/595de0c74bbede891c6a5e9ad0844e5d3e5ed2c8/patches/amd-mainline/rocm-systems/0002-Revert-SWDEV-543498-Some-compute-Ubertrace-profiles-.patch) is present because it points to code that does not exist in the current version of PAL in `ROCm/rocm-systems`.

Two pull requests from rocm-systems are making changes to the files that were patched.
ROCm/rocm-systems#1587
ROCm/rocm-systems#1680

The sequence should be:
1. update PAL in rocm-systems
2. bump submodule in TheRock and delete the patch
3. unblocks the above PRs.

jayhawk-commits:
CC @pcritchl-amd @lancesix @idass1990 @litvaOo

jayhawk-commits:
PAL has been updated and the ubertrace patch has been removed. A change will be made on rocm-systems and the pending PRs can proceed.

jayhawk-commits:
Change has been made on rocm-systems.

jayhawk-commits:
PAL update was reverted in #2154 and the ubertrace code was patched out again. Re-opening, albeit late.

jayhawk-commits:
ROCm/rocm-systems#2672 updated PAL, which should include the ubertrace profile functions

jayhawk-commits:
The patch was removed on TheRock repo.

PR #3570, opened 2026-02-23
[patches] drop ubertrace patch due to amdgpu-windows-interop update

jayhawk-commits:
- The patched out functions are available with the updated PAL.


jayhawk-commits:
> Do we have CI coverage for what this patch was addressing? Is there a tracking issue?
> 
> LGTM if CI passes, and those questions have good answers ;)

The previous removal of the patch was tracked in #2029, but it looks like I forgot to re-open the issue when the PAL update was reverted around the time I went on vacation. With the older amdgpu-windows-interop and without the patches, TheRock compilation on Windows would fail because those ubertrace functions did not exist on the PAL interfaces. The CI coverage for this PR would be the hip runtimes compiling OK on Windows.

PR #3584, opened 2026-02-24
Users/ibrahimw1/pr test origami

ibrahimw1:
## Submission Checklist


marbre:
No need to open a PR for pure testing. You can trigger the CI to run on a specific branch and limit for which architecture you want to build and test. This saves compute and creates less noise.

astrelsky:
Origami appears to be using nanobind to expose it to python. This doesn't support using the python stable abi until python3.12. It doesn't look like there is a lot exposed by the bindings, so it might not even be difficult to do it manually, or figure out what everything else in therock is using.

PR #3588, opened 2026-02-24
Add amd llvm C++ headers to build artifacts

JonathanLichtnerAMD:
C++ headers are needed for runtime kernel compilation. Without these, kernel runtime compilation can fail due to missing includes for <type_traits> and other standard library headers.

## Motivation

Certain MIOpen unit tests fail during runtime compilation because the type_traits include could not be found:
```
/tmp/comgr-209-20-187dbd/include/miopen_type_traits.hpp:151:10: fatal error: 'type_traits' file not found
  151 | #include <type_traits>
      |          ^~~~~~~~~~~~~
```

However, this same test passes if the rock is built locally.

## Technical Details

The rock build artifacts are missing the C++ headers, so this change adds these headers.

## Test Plan

Download the artifacts built as part of this PR's CI and then verify that the affected tests pass.

## Test Result

- [x] Verify that " ./miopen_gtest --gtest_filter=Full/GPU_LstmMSRnn_FP32.TestFloat/1451*" passes from this PR's CI artifacts, as well as other MIOpen unit tests that were failing due to this

## Submission Checklist


JonathanLichtnerAMD:
I've verified that this change fixes the hip runtime issue.

PR #2001, opened 2025-11-03
Enable hip-tests on TheRock

cjatin:
## Motivation

Enable hip-tests execution on TheRock linux and windows. 
Aligns installation path of hip-tests on linux and Windows
enables sharding of hip-tests 

## Details

Enable hip-tests execution on TheRock linux and windows. 

**This PR fixes -** 
https://github.com/ROCm/TheRock/issues/1354

**Hip-tests source that gets built/executed as part of this PR**
https://github.com/ROCm/rocm-systems/tree/develop/projects/hip-tests/catch

**Related PRs to enable build/execution of tests -**
https://github.com/ROCm/TheRock/pull/2373 - enables building of tests
https://github.com/ROCm/rocm-systems/pull/1624 - prepares rocm-systems to be built with catchlib from upstream , compiler/hip detection, packaging change
https://github.com/ROCm/rocm-systems/pull/2187 - aligns installation path for linux and windows

**few other script changes to have compatible builds on older build frameworks.**
https://github.com/AMD-ROCm-Internal/build-infra/pull/316
https://github.com/AMD-Radeon-Driver/compute-utils/pull/169
https://github.com/AMD-Radeon-Driver/compute-utils/pull/171
hip-tests executed and verified successfully

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3590, opened 2026-02-24
Enable ocltst execution on Rock

agunashe:
## Motivation
Enable ocltst execution on Rock builds on both linux and windows. 

## Technical Details
enable ocltst execution on linux and windows. 

Setup to pick the correct runtime libs
export OCL_ICD_VENDORS=/opt/rocm/etc/OpenCL/vendors/
Linux -
setting LD_LIBRARY_PATH to find the libs
Windows - 
choosing to copy the runtime/icd dlls to local folder instead of regsitry key update. 
Registry key update requires admin permissions so using copy mechanism to pick the right dlls
else the system32 dlls will be loaded, leading to incorrect dlls loaded at runtime

## Test Plan
All opencl tests executed and working

## Test Result
All opencl tests executed and working

## JIRA
AIRUNTIME-267

## Submission Checklist


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -45,6 +45,17 @@ def _get_script_path(script_name: str) -> str:
             "windows": 4,
         },
     },
+    # ocltest
+    "ocltst": {
+        "job_name": "ocltst",
+        "fetch_artifact_args": "--tests",
+        "timeout_minutes": 120,
+        "test_script": f"python {_get_script_path('test_ocltst.py')}",
+        "platform": ["linux", "windows"],
```
marbre, line 80: Do you test on Windows?

agunashe: Yes I need to test on windows as well. It needs to work test_ocltst.py handles both linux and windows

build_tools/github_actions/test_executable_scripts/test_ocltst.py
marbre, line 1: @geomin12 can you review the test script?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,17 @@ def _get_script_path(script_name: str) -> str:
...
+        "total_shards": 4,
```
geomin12: please update to this format for shards:

https://github.com/ROCm/TheRock/blob/4a9e6b47df531beab14070b5d647d9e45372e94f/build_tools/github_actions/fetch_test_configurations.py#L58-L61

rocm-systems
geomin12, line 1: i think this is for testing, but let's avoid doing bumps in PRs with other work

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,17 @@ def _get_script_path(script_name: str) -> str:
             "windows": 4,
         },
     },
+    # ocltest
+    "ocltst": {
+        "job_name": "ocltst",
+        "fetch_artifact_args": "--tests",
```
geomin12: can we use `--ocltst` flag too? otherwise, this test install step will take a while and also install unnecessary artifacts

example here: https://github.com/ROCm/TheRock/commit/5f60db78b53ca25f8bd6299d31f29a8b13791ed9#diff-f827cc49787b6ff82e268d612f178774273ae8c2955e4e75364c654ee6c19301

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,99 @@
...
+        if LD_LIBRARY_PATH is not None:
+            LD_LIBRARY_PATH = Path(LD_LIBRARY_PATH)
+        env["LD_LIBRARY_PATH"] = (
+            f"{ROCK_LIB_PATH}:{OCL_LIB}:{LLVM_LIB}:{ROCM_SYSDEPS_LIB}:{LD_LIBRARY_PATH}"
+        )
+        env["OCL_ICD_VENDORS"] = f"{OCL_ICD_VENDORS}/"
+
+
+def execute_tests(env):
+    if platform.system() == "Linux":
```
geomin12: nit: you can just add a `is_windows = platform.system() == "Windows"` at the top of file and reuse this, since i see it used many places

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,99 @@
...
+        OCLTST_PATH = str(Path(THEROCK_BIN_DIR).parent / "share" / "opencl" / "ocltst")
+        cmd = [
+            "./ocltst",
+            "-J",
```
geomin12: can we use the full flag name? for anyone observing these tests, it makes life a lot easier to see what flags mean

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,99 @@
...
+    # Set ROCM Path, to find rocm_agent_enum etc
+    ROCM_PATH = Path(THEROCK_BIN_DIR).resolve().parent
+    env["ROCM_PATH"] = str(ROCM_PATH)
+    if platform.system() == "Linux":
+        ROCK_LIB_PATH = Path(THEROCK_BIN_DIR).parent / "lib"
+        OCL_LIB = Path(ROCK_LIB_PATH) / "opencl"
+        LLVM_LIB = Path(ROCK_LIB_PATH) / "llvm" / "lib"
+        ROCM_SYSDEPS_LIB = Path(ROCK_LIB_PATH) / "rocm_sysdeps" / "lib"
+        OCL_ICD_VENDORS = Path(THEROCK_BIN_DIR).parent / "etc" / "OpenCL" / "vendors"
+        # logging.info(f"++ contents of ROCK_LIB_PATH={os.listdir(ROCK_LIB_PATH)}")
```
geomin12: remove if not used

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,99 @@
...
+        LD_LIBRARY_PATH = os.getenv("LD_LIBRARY_PATH")
+        if LD_LIBRARY_PATH is not None:
+            LD_LIBRARY_PATH = Path(LD_LIBRARY_PATH)
+        env["LD_LIBRARY_PATH"] = (
+            f"{ROCK_LIB_PATH}:{OCL_LIB}:{LLVM_LIB}:{ROCM_SYSDEPS_LIB}:{LD_LIBRARY_PATH}"
+        )
+        env["OCL_ICD_VENDORS"] = f"{OCL_ICD_VENDORS}/"
+
+
+def execute_tests(env):
```
geomin12, line 71: do these tests have a `smoke` test filter? something like this: https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,17 @@ def _get_script_path(script_name: str) -> str:
...
+        "timeout_minutes": 120,
```
geomin12: from this, it looks like tests takes a total of 8 hours.... that's a pretty insane amount of time, especially for a developer waiting on PR checks.

are there any filters we can use? we are getting teams to start using this format: https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md so we get fast turn around times with the machines we have

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: removed from the script

build_tools/github_actions/fetch_test_configurations.py
agunashe: added the shard_dict to the script

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: done added the windows check at the top of the file

build_tools/github_actions/fetch_test_configurations.py
agunashe: done added the --ocltst flag along with --tests

agunashe: adding --ocltst causes failure -
https://github.com/ROCm/TheRock/actions/runs/23072096718/job/67058673308?pr=3590

removing the --ocltst option

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: removing this flag, as its not required to execute the tests

build_tools/github_actions/fetch_test_configurations.py
agunashe: reduced it to 20 mins. that should be enough to verify

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: as of now, there is no smoke tests for this test.
this is already a shortened test. Maybe we can look at adding smoke tests in the future, needs more work identifying and trimming the tests

build_tools/github_actions/fetch_test_configurations.py
agunashe: actually needed 30 mins of timeout

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        env["OCL_ICD_VENDORS"] = f"{OCL_ICD_VENDORS}/"
+    else:
+
+        OCLTST_PATH = Path(THEROCK_DIR) / "tests" / "ocltst"
+        copy_dlls_exe_path(OCLTST_PATH)
+        OCL_DLL_FILE = Path(OCLTST_PATH) / "amdocl64.dll"
+        OCL_ICD_DLL = Path(THEROCK_BIN_DIR) / "OpenCL.dll"
+        env["OCL_ICD_FILENAMES"] = str(OCL_DLL_FILE)
+    return OCLTST_PATH
```
ScottTodd, line 68: I believe we can (and should) remove nearly all of this setup code from this test script if we fix https://github.com/ROCm/rocm-systems/blob/be8076461de7f44e21a0f273d94f5872c89c1e17/projects/clr/opencl/tests/ocltst/CMakeLists.txt#L7-L11 to install ocltst into a standard install location like `bin/` instead of a new location like `tests/`.

See for example what hipblaslt does:
* https://github.com/ROCm/rocm-libraries/blob/670f8557bbf5062bedaea552827733f245fdbccc/projects/hipblaslt/CMakeLists.txt#L504-L518
* https://github.com/ROCm/rocm-cmake/blob/develop/share/rocmcmakebuildtools/cmake/ROCMInstallTargets.cmake

Or a more recent / simpler case like dnn-providers/hipblaslt-provider: https://github.com/ROCm/rocm-libraries/blob/94453c6cb85db167e96e3f463a12d808734674b5/dnn-providers/hipblaslt-provider/cmake/Tests.cmake#L180-L192

The guiding principles are this:
* Tests should be runnable from a build directory (e.g. via `ctest`)
* Component development should use the same testing workflow as CI systems
* Scripts for CI like these in `build_tools/github_actions/test_executable_scripts/` should be thin wrappers around test runners / executables that possibly add filtering and sharding options applicable to CI. They should NOT be structurally modifying the artifact directory in any way. We granted a narrow exception for that to hip-tests in https://github.com/ROCm/TheRock/pull/2001 but if this is becoming a pattern we need to fix it and not dig the whole deeper

See more detailed analysis of the issue and recommendations to fixing it here: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_3590.md#1-install-to-bin-to-eliminate-dll-copy-and-ld_library_path-workarounds

I ran local experiments with the artifacts produced by this PR here: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_3590.md#local-experiment-install-to-bin-validation
* Moving files to `bin/`, tests passed if I ran _without_ setting `OCL_ICD_FILENAMES` (so using my system OpenCL ICD version 3652.0)
* Moving files to `bin/`, tests failed _with_ `OCL_ICD_FILENAMES` (using OpenCL ICD version 3628.0...?) I'm not sure why, given that tests passed on CI

agunashe: Thanks for your comment.  I agree we should not have different installation paths for Linux vs windows.
However, changing the installation paths in clr could impact compute build. 
So changes will be required in rocm-systems, compute could take a couple of weeks to get this merged again. 
this coverage is important piece to validate Opencl and move away from legacy build systems. 
Also earlier the linux paths were a requirement from packaging team to install the artifacts in a certain folder path. 
We need similar requirements now with Rock which covers both linux and windows
I want to come back and clean up the paths/artifacts through a ticket/issue instead of blocking this PR.
Delaying this further risks test coverage gap in 7.13 too.

Looking at your experiments-
exp 1 - invalid, uses system32 opencl and comgr dlls for loading. Registry key takes precedence. Which is incorrect, need the dlls from the rock build to be loaded
exp 2 - partially valid, setting OCL_ICD_FILENAMES=$(pwd)/amdocl64.dll picks one dll from local folder but does not find all the dlls. So it fails
exp 3 - invalid again. same as exp1. use dlls from system32 instead of local build

To pick the right dlls to be loaded at runtime . Use Listdlls app to see the dlls loaded at runtime.
opt1 - We need to update registry key to point the amdocl64 
or 
opt2 - copy the required dlls to local folder of exe.  

opt1 did not work, because we do not have any facility in Rock to update the registry key. Updating registry key requires admin permissions so could not set it through the test_ocltst.py.  
choosing opt2 - With the interest of time, chose 2. this way local folder always gets precedence over system32 dlls

agunashe:
> Detailed review notes: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_3590.md (this was generated via some back and forth review of the PR, looking at logs, etc.). I expect this level of analysis from PR authors too, so a PR description of just "Technical Details: enable ocltst execution on linux and windows" is insufficient.
> 
> I've added one inline comment for the major architectural issue. I'll wait to comment on the smaller details until that is resolved.

Added more information to the technical details.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
ScottTodd: > However, changing the installation paths in clr could impact compute build.

Impacts on internal builds are not generally a blocking concern for open source code, builds, or releases. The costs for living downstream are paid by downstream projects, not upstream.

I see the points about timelines, but so far I'm seeing more tech debt being accumulated than paid down (I tried to run hip-tests the other day and was completely unable to do so, facing similar issues as on here... these "scripts that only work on CI" need to be fixed)

> Also earlier the linux paths were a requirement from packaging team to install the artifacts in a certain folder path.
> We need similar requirements now with Rock which covers both linux and window

As far as I can tell, only rocgdb (also added recently) writes to a top level `tests/` folder. All other ROCm subprojects use other paths. So this is already in "needs a requirement from packaging" territory. Using the already established layout seems like it would be the path of least resistance here to me, especially given that the `tests/` path does not work (we're not going to ask users who download our packages to copy DLLs to run tests... they should be able to run the tests as easily as developers working on the projects and CI/CD systems can).

> exp 2 - partially valid, setting OCL_ICD_FILENAMES=$(pwd)/amdocl64.dll picks one dll from local folder but does not find all the dlls. So it fails

Which am I missing? I have everything that this PR tries to copy and I don't see any missing DLLs when I run the various exes and dlls through https://github.com/lucasg/dependencies
```
λ ls D:\scratch\claude\artifacts\3590-win\bin
AMD.ROCM.Comgr.MANIFEST  clinfo.exe*   hipcc.exe*      oclperf.dll*     oclruntime.exclude
amd_comgr0702.dll*       cltrace.dll*  hipconfig.bat   oclperf.exclude  ocltst.exe*
amdocl64.dll*
```

agunashe: actually looking at your exp setup 
exp 1 - since you copy the test items to bin. runtime dlls picked are valid but system32 Opencl.dll is loaded 
exp 2 - assuming items are copied from bin to test . setting OCL_ICD_FILENAMES forces to pick local OpenCL.dll. not sure what is causing this failure. Cant repro it locally.
exp 3 - running from tests folder without copying dlls to local folder picks the system32 dlls. So this is invalid.

agunashe: > 
> Impacts on internal builds are not generally a blocking concern for open source code, builds, or releases. The costs for living downstream are paid by downstream projects, not upstream.

I agree with this if its a simple bug fix, upstream gets priority.  But currently we are still holding the legacy/downstream systems because of the test coverage gap in Rock.  Once we have enabled the tests on Rock, we can look at fixing the issues, atleast then we have a gate check that verifies our change. Right now the opencl changes are not tested in Rock, this causes issues to be caught later in the QA cycle.

> I see the points about timelines, but so far I'm seeing more tech debt being accumulated than paid down (I tried to run hip-tests the other day and was completely unable to do so, facing similar issues as on here... these "scripts that only work on CI" need to be fixed)

Please file an issue assign to our team, we will work on it soon. It should work outside CI as well.

> As far as I can tell, only rocgdb (also added recently) writes to a top level `tests/` folder. All other ROCm subprojects use other paths. So this is already in "needs a requirement from packaging" territory. Using the already established layout seems like it would be the path of least resistance here to me, especially given that the `tests/` path does not work (we're not going to ask users who download our packages to copy DLLs to run tests... they should be able to run the tests as easily as developers working on the projects and CI/CD systems can).

ocltst is a test binary. I am not sure if its right thing to copy this to bin folder.  sorry need a proper requirement and design.

also exp2 should work, I am not able to repro it locally.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def execute_tests(env):
+    OCLTST_PATH = setup_env(env)
+    if not is_windows:
+        cmd = [
+            "./ocltst",
+            "-m",
```
geomin12: can we add full flag names? it is unclear to me what `-m` is

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+            "liboclruntime.so",
+            "-A",
```
geomin12: same here for full flag name

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+            "oclruntime.exclude",
+        ]
+        shell_var = False
+        logging.info(f"++ Setting LD_LIBRARY_PATH={env['LD_LIBRARY_PATH']}")
+        logging.info(f"++ Setting OCL_ICD_VENDORS={env['OCL_ICD_VENDORS']}")
+    else:
+        cmd = [
+            "ocltst.exe",
+            "-m",
```
geomin12: full flag name pls

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+            "oclruntime.dll",
+            "-A",
+            "oclruntime.exclude",
```
geomin12: any options for parallelization? can we add a comment on what ./ocltst does? it seems to be a custom script so it's unclear to us what is under the hood. is it ctest/gtest/pytest?

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,96 @@
...
+        copy_dlls_exe_path(OCLTST_PATH)
+        OCL_DLL_FILE = Path(OCLTST_PATH) / "amdocl64.dll"
+        OCL_ICD_DLL = Path(THEROCK_BIN_DIR) / "OpenCL.dll"
+        env["OCL_ICD_FILENAMES"] = str(OCL_DLL_FILE)
+    return OCLTST_PATH
+
+
+def execute_tests(env):
+    OCLTST_PATH = setup_env(env)
+    if not is_windows:
```
geomin12: since both executables are almost identical, can we just do:
```
extension = "" if not is_windows else ".exe"
f"./ocltst{extension}"
```
?

then, this will be less dup code

build_tools/github_actions/test_executable_scripts/test_ocltst.py
geomin12, line 1: 30 mins is long. any options for parallelization? having a GPU machine run 30 mins for quick/sanity checks is really long and costly

build_tools/github_actions/fetch_test_configurations.py
geomin12: okay let's fix it and add it back, we don't need to install everything (costs storage / time / memory to install unneeded artifacts)

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: There is no full flag name just the short names exist. added a comment next to it

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: There is no full flag name just the short names exist. added a comment next to it

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: There is no full flag name just the short names exist. added a comment next to it

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: its a custom test with c/c++ calls no test framework used.  Added 16 threads , think they run parallely. Need to check

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: added common cmd.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: i have added 16 threads for execution.  Lets see if the execution time comes down

build_tools/github_actions/fetch_test_configurations.py
agunashe: okay lets see adding it back. didnt realize had to specify in other files too

agunashe:
> I believe we are adding tests for new components as part of this PR? Before we do that, do we have a report on the below:
> 
> 1. are are introducing any flaky tests / timeouts on tests as part of this that will impact our ci production pipelines/nightlies ?
> 2. how much resources these new tests need to successfully execute all steps ?
> 3. are we adding any extra time as part of this ? if yes, how much ?

This is a coverage gap in the Rock. We are porting it from existing build framework. this test has been from years, mostly stable. From the tests we observed around 30 mins to complete the tests. I have added parallel threads for execution. Hopefully it should take less time

build_tools/github_actions/fetch_test_configurations.py
agunashe: added --ocltst and seems to be working now. Please review

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: Added comments to the flag options

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: Added comments to the flag options

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: Added comments to the flag options

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: seems to have reduced the execution time to around 17 mins.

build_tools/github_actions/fetch_test_configurations.py
agunashe: tested on both linux and windows.

rocm-systems
agunashe: this had to be done to verify the Rock PR since it depends on rocm-systems PRs.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
geomin12: okay, 16 mins is better, but let's open an ocl test ticket to introduce filtering

geomin12: 16 mins is okay now, but let's introduce an internal ticket for ocl test filtering

build_tools/github_actions/test_executable_scripts/test_ocltst.py
geomin12: yes let's open tix for filtering, as we probably want to do expanded testing later

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,90 @@
...
+THEROCK_BIN_DIR = Path(THEROCK_BIN_DIR_STR).resolve()
+THEROCK_DIR = Path(THEROCK_BIN_DIR).parent
+env = os.environ.copy()
+is_windows = platform.system() == "Windows"
+
+
+# copies the dlls to local ocltst path.
+# to overwrite the registry entries
+def copy_dlls_exe_path(ocltst_path):
+    if platform.system() == "Windows":
```
geomin12, line 26: nit: use `is_windows`

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,90 @@
...
+        OCL_DLL_FILE = Path(OCLTST_PATH) / "amdocl64.dll"
+        OCL_ICD_DLL = Path(THEROCK_BIN_DIR) / "OpenCL.dll"
+        env["OCL_ICD_FILENAMES"] = str(OCL_DLL_FILE)
+        logging.info(f"++ Setting OCL_ICD_FILENAMES={env['OCL_ICD_FILENAMES']}")
+    return OCLTST_PATH
+
+
+def execute_tests(env):
+    OCLTST_PATH = setup_env(env)
+    OCLTST = Path(OCLTST_PATH) / "ocltst"
```
geomin12, line 73: doesn't ocltst need to be `ocltst.exe` for windows?

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: https://github.com/ROCm/TheRock/issues/4192 - added to this issue

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: https://github.com/ROCm/TheRock/issues/4192 - can create a internal ticket for this as well

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: no it finds the executable with just the name. you can check the execution of ocltst in the windows build

build_tools/github_actions/test_executable_scripts/test_ocltst.py
agunashe: would prefer to fix it in the clean up. We need this PR merged ASAP for ocl coverage.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
geomin12: can we add that GH issue to the code here, so we don't forget

agunashe:
> > I believe we are adding tests for new components as part of this PR? Before we do that, do we have a report on the below:
> > 
> > 1. are are introducing any flaky tests / timeouts on tests as part of this that will impact our ci production pipelines/nightlies ?
> > 2. how much resources these new tests need to successfully execute all steps ?
> > 3. are we adding any extra time as part of this ? if yes, how much ?
> 
> This is a coverage gap in the Rock. We are porting it from existing build framework. this test has been from years, mostly stable. From the tests we observed around 30 mins to complete the tests. I have added parallel threads for execution. Hopefully it should take less time

@amd-shiraz - if you are satisfied with this PR and the explanation. Can you please review/approve this so we can merge and close the gap please.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
ScottTodd: It's a 1 line change, in the code this PR is directly introducing, immediately prior in the code. Fix it please.

@geomin12 @amd-shiraz please don't approve PRs that so obviously are ignoring reviewer feedback.

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,90 @@
...
+                shutil.copy(dll, ocltst_path)
+                logging.info(f"++ Copied: {dll} to {ocltst_path}")
+            except Exception as e:
+                logging.info(f"++ Error copying {dll}: {e}")
+
+
+# returns ocltst path
+def setup_env(env):
+    ROCM_PATH = Path(THEROCK_DIR)
+    env["ROCM_PATH"] = str(ROCM_PATH)
```
ScottTodd, line 43: This intermediate `ROCM_PATH` variable is doing nothing. `THEROCK_DIR` is already of type `Path`

build_tools/github_actions/test_executable_scripts/test_ocltst.py
```diff
@@ -0,0 +1,90 @@
...
+    if not is_windows:
+        ROCK_LIB_PATH = Path(THEROCK_DIR) / "lib"
+        OCL_LIB = Path(ROCK_LIB_PATH) / "opencl"
+        LLVM_LIB = Path(ROCK_LIB_PATH) / "llvm" / "lib"
+        ROCM_SYSDEPS_LIB = Path(ROCK_LIB_PATH) / "rocm_sysdeps" / "lib"
+        OCL_ICD_VENDORS = Path(THEROCK_DIR) / "etc" / "OpenCL" / "vendors"
+        OCLTST_PATH = Path(THEROCK_DIR) / "share" / "opencl" / "ocltst"
```
ScottTodd, line 50: These variables are all already of type `Path`, no need to wrap in `Path()`

build_tools/install_rocm_from_artifacts.py
```diff
@@ -443,6 +445,10 @@ def retrieve_artifacts_by_run_id(args):
             # rocrtst depends on sysdeps-hwloc (which depends on sysdeps-libpciaccess)
             extra_artifacts.append("sysdeps-hwloc")
             extra_artifacts.append("sysdeps-libpciaccess")
+        if args.ocltst:
+            # ocltst binaries ship in the core-ocl test artifact (base already includes _lib/_dev).
+            if args.tests:
+                argv.append("core-ocl_test")
```
ScottTodd, line 451: This doesn't quite follow existing patterns here.

There is already a code block below that handles `_test` artifacts:

```python
        # Fetch _lib (always) and _test (when --tests) for each artifact.
        # Some projects have self-contained _test archives (just test
        # binaries), while others may also need executables or data from
        # _run. Add those explicitly above via argv.append("<name>_run").
        extra_artifact_patterns = [f"{a}_lib" for a in extra_artifacts]
        if args.tests:
            extra_artifact_patterns.extend([f"{a}_test" for a in extra_artifacts])
```

This can probably just do:

```suggestion
        if args.ocltst:
            # ocltst binaries ship in the core-ocl test artifact (base already includes _lib/_dev).
            extra_artifacts.append("core-ocl")
```

build_tools/github_actions/test_executable_scripts/test_ocltst.py
ScottTodd: We can provide another LIMITED, TEMPORARY exception for these sorts of DLL copying workarounds in tests as part of adding ocltst coverage to TheRock but this needs to be fixed urgently as it impacts core developer workflows and shipping package structures.

We're still recovering from issues where framework (JAX) automated tests, QA processes, and packages were completely misaligned to the point where we were silently falling back to testing on CPU on CI, QA had their own undocumented test procedures, and the end result were user-facing packages that were fundamentally broken.

This has all the same smells to it. We ship these test files as part of releases. Users/developers/etc. can and will try to run the test files in those release archives. They will not work as-is. Users/developers/etc. will not think to find a script like this or copy DLLs / set environment variables themselves.

Tracking follow-ups in https://github.com/ROCm/TheRock/issues/4192.

Issue #3587, opened 2026-02-24
Run hip-tests and opencl on Windows through PAL and ROCR


jayhawk-commits:
To help with the transition of running HIP on Windows via ROCR, there will be a period where hip-tests for Windows will run once through PAL and once through ROCR. This is controlled through the environment variable `GPU_ENABLE_PAL` being set to 1 or 0.

First step: Have hip-tests on Windows run with the flag set to 1, and then run again with the flag set to 0. This can be separated as two jobs. The PAL result can be used for pass/fail, while the ROCR result can be informational.

Once parity is achieved, the version used for the pass/fail signal can be reversed.

AmosLewis:
[Windows::release / Test gfx110X-all / Test hip-tests (PAL) / Test hip-tests (PAL) (shard 1/4) (gfx110X-all)](https://github.com/ROCm/TheRock/actions/runs/24722329856/job/72386788882?pr=4700#logs) fail on today 0421 bump. It supposed to pass https://github.com/ROCm/TheRock/pull/4700

New issue filed here https://github.com/ROCm/rocm-systems/issues/5260

AmosLewis:
still block https://github.com/ROCm/TheRock/pull/4760 and https://github.com/ROCm/TheRock/pull/4761

ScottTodd:
> still block [#4760](https://github.com/ROCm/TheRock/pull/4760) and [#4761](https://github.com/ROCm/TheRock/pull/4761)

Huh? This issue tracks adding new testing. If there are issues with the tests, that should be tracked separately - feature requests do not block regular development work, bugs do.

araravik-psd:
seeing bump failure timeouts for hip-tests https://github.com/ROCm/TheRock/actions/runs/25194856323/job/73885804350

Do we know this timeout failure is also related to the above Windows issue?

@jayhawk-commits

jayhawk-commits:
> seeing bump failure timeouts for hip-tests https://github.com/ROCm/TheRock/actions/runs/25194856323/job/73885804350
> 
> Do we know this timeout failure is also related to the above Windows issue?
> 
> [@jayhawk-commits](https://github.com/jayhawk-commits)

The run you linked is on Linux. The change doesn't apply to Linux. Default behaviour on TheRock has also been updated, so that the Windows hip-tests ROCR paths is opt-in and off by default.

agunashe:
ROCR backend testing needs new 26.20 internal mainline drivers to test the AQL path.
[...]

PR #3593, opened 2026-02-24
Windows hip-tests: Run twice - PAL (pass/fail) and ROCR (informational)

jayhawk-commits:
**Summary**  
hip-tests on Windows run twice for CI: once with **PAL** (pass/fail) and once with **ROCR** (informational), controlled by `GPU_ENABLE_PAL=1` and `0` as described in #3587.

**Technical Details**
- **Single job, matrix-driven:** When `test_artifacts` is run; the test matrix is extended in `fetch_test_configurations.py` so that on Windows the hip-tests component becomes two entries: **hip-tests (PAL)** and **hip-tests (ROCR)**.
- **PAL:** `gpu_enable_pal: "1"`, no `expect_failure`, job result is used for pass/fail.
- **ROCR:** `gpu_enable_pal: "0"`, `expect_failure: true`, same job type but `continue-on-error`, so failures are informational.
- `GPU_ENABLE_PAL` is set in the Test step from the component (matrix). Only the two hip-tests entries set it; other components leave it empty, as it does not affect their execution. In the future, OpenCL tests will also read this.

**Future**  
When ROCR is ready for pass/fail, swap which entry has `expect_failure` (and optionally job names) in `fetch_test_configurations.py`.


jayhawk-commits:
The splitting of hip-tests into two jobs, just on Windows, is working. The `env` is set correctly and toggled between the two cases. However, I see a lot of test failures on the Windows gfx1151 runs.

jayhawk-commits:
I will wait for rocm-systems submodule update to see if it resolves the issue with the hip-tests on ROCR job.

agunashe:
> I will wait for rocm-systems submodule update to see if it resolves the issue with the hip-tests on ROCR job.

https://github.com/ROCm/TheRock/actions/runs/22368278729/job/64870824855?pr=3593 - ROCr backend didnt execute anything. Something else is missing, will try out locally first and keep you updated

jayhawk-commits:
> > I will wait for rocm-systems submodule update to see if it resolves the issue with the hip-tests on ROCR job.
> 
> https://github.com/ROCm/TheRock/actions/runs/22368278729/job/64870824855?pr=3593 - ROCr backend didnt execute anything. Something else is missing, will try out locally first and keep you updated

Yes, my suspicion is that rocm-systems in TheRock is missing updates for hip on ROCR Windows support.

geomin12 (commented):
i am sure you are aware, but tests fail :( 

but overall, lgtm

jayhawk-commits:
It appears further changes are required on rocm-systems to support the Windows ROCR path, but the workflow changes in this PR behave as expected and will be ready when rocm-systems is set.

idass1990:
updating branch as it looks like actions scripts are outdated..

agunashe:
Pulling in all changes from Rock - users/agunashe/pal_static_rocr_both branch (https://github.com/ROCm/TheRock/pull/4044)
tot - 46d27745e1c0fc314e99cf00ff1b9c3074322935

4044 verified the default PAL backend works. This current PR 3593 will test both PAL and ROCr backends independently.

jayhawk-commits:
Behaviour seems fine now. Both ROCR and PAL cases run on Windows. ROCR has some failures, which is expected at this phase and is only informational.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -577,6 +577,45 @@ def run():
...
+            # Hip-tests on Windows run twice: PAL (pass/fail) and ROCR (informational)
+            # for parity tracking until ROCR is the pass/fail path. See:
+            # https://github.com/ROCm/TheRock/issues/3587
+            if key == "hip-tests" and platform == "windows":
+                base = selected_matrix[key]
+                total_shards = base.get("total_shards_dict", {}).get(platform, 1)
+                if test_type == "quick":
+                    total_shards = 1
+                shard_arr = list(range(1, total_shards + 1))
+
```
idass1990, line 590: a note from my agent review:

_pal_entry/rocr_entry duplicate base fields manually. This is a deliberate choice that avoids accidentally carrying future fields. The trade-off is that any new field added to the hip-tests matrix config must be 
  manually added to both entries. Worth a comment or a note in the migration TODO._

idass1990: many fields are common, could have a shared entry and append rocr/pal specific fields to it... no fuctional impact, could be a future clean-up task.

jayhawk-commits: Yes. The duplication was intentional.

Issue #3331, opened 2026-02-09
Refactor `post_build_upload.py` to partly move server side


marbre:
The script `build_tools/github_actions/post_build_upload.py` must be refactored to move the generation of index file for artifacts, logs, etc. to server side.

As a prerequisite the structure where individual artifacts, logs, etc. get pushed need to be defined prior to the refactoring. This is related to what @ScottTodd is partly working on.

ScottTodd:
See https://github.com/ROCm/TheRock/pull/3000 for some of what I'm thinking/planning around a central location for workflow outputs -> S3 storage/upload code.

For this script in particular, we could also stream logs and artifacts to S3, either continuously (during component builds) or incrementally (after each component build), instead of waiting until the overall build completes before uploading anything.

ScottTodd:
Raw design notes: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/run-outputs-layout.md#design-server-side-index-generation-3331

I think we should set up something like https://github.com/razorjack/s3-directory-listing (or use/fork that ourselves).

The [`build_tools/github_actions/post_build_upload.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/post_build_upload.py) script will:
1. Generate archives as needed (ninja logs, etc.)
2. Choose which files to upload and what paths to upload them to
3. Upload those files

On the server-side, we should have some _generic_ infrastructure that:
1. Listens for S3 `PutObject` events
2. Determines the "directory" prefix from an uploaded object's key
3. Lists objects at that prefix (e.g. via `ListObjectsV2`)
4. Generates an `index.html` page with the file listing (name, size, last modified, etc.)
5. Writes that file to the same prefix

Claude Code also suggests setting `TTL=0` on `*/index.html` paths so they are always fresh and not cached?

Note that we should aim for the code running on the server side to NOT know anything about what files will be uploaded or how they relate to one another, at least at the start. That way, we can add more file types in the future without needing to update the server-side code. For some of the more complicated index pages like https://therock-nightly-tarball.s3.amazonaws.com/index.html, it makes more sense to have the server-side code know more details about the files that are present so the generated page includes sorting/filtering/etc. specific to the files. We'll probably get to that point eventually for our CI build artifacts/logs - especially if we add a central summary page for each workflow run that aggregates/summarizes the CI results (build status, artifact sizes, test result summaries, etc.)

ScottTodd:
More design thinking:

[...]

PR #3000, opened 2026-01-20
Add central `RunOutputRoot` class for deriving CI run output paths on S3

ScottTodd:
This PR introduces a new `RunOutputRoot` class that centralizes path construction, clearly documents which files are uploaded where, and adds many new unit tests.

## Motivation

We had multiple files constructing paths on S3 for CI outputs such as artifacts, logs, and reports. Each of these files needed to know what schema to follow and there was risk of the files not agreeing on conventions. Several files even used the `retrieve_bucket_info()` function which just returned a tuple of `[EXTERNAL_REPO, BUCKET]`.

This is prerequisite work for:

* https://github.com/ROCm/TheRock/issues/1559
  * I want to upload ROCm Python packages (with "dev" versioning) to S3 and this builds a foundation for new files
* https://github.com/ROCm/TheRock/issues/2608
  * I want to list and download ROCm artifacts from workflow runs without copying the URL structure again

## Technical Details

Key changes:

1. Added new `build_tools/_therock_utils/run_outputs.py` file with a new `RunOutputRoot` class
2. Migrated existing code to use the new class, like so:

    ```diff
    - analysis_url = f"{bucket_url}/logs/{artifact_group}/build_time_analysis.html"
    + analysis_url = run_root.build_time_analysis_url(artifact_group)
    ```

    ```diff
    - s3_prefix = f"{external_repo}{run_id}-{platform}"
    + s3_prefix = run_root.prefix
    ```

3. Moved `retrieve_bucket_info()` from `github_actions_utils.py` to the new file as an implementation detail
4. Added new documentation for the directory structure of workflow run outputs in `docs/development/run_outputs_layout.md`
5. Added new unit tests

See also:

* Planning and notes: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/run-outputs-layout.md
* Self-review at https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_005_run-outputs.md

## Test Plan

* Added new unit tests and updated existing unit tests
* Watch CI for unexpected file paths, broken logs, errors, etc.

> [!NOTE]
> I also want better ways to test these sorts of changes, so I added some follow-up plans in my plan/notes doc at [Follow-up Work: Local Testing & Backend Abstraction](https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/run-outputs-layout.md#follow-up-work-local-testing--backend-abstraction). The main ideas are to
[...]

PR #3596, opened 2026-02-25
[CI] Centralize code for uploading artifacts/logs/packages in workflows

ScottTodd:
> [!NOTE]
> This is a rewrite of https://github.com/ROCm/TheRock/pull/3000. The code was written primarily via Claude Code and has gone through several design iterations over the last ~month, with a few smaller pieces of work already split out and landed.
> 
> See also:
> * Task file with research, prototypes, notes, etc.: [`tasks/active/run-outputs-layout.md`](https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/run-outputs-layout.md)
> * Self-review 1: [`reviews/local_014_run-outputs-locations-2.md`](https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_014_run-outputs-locations-2.md)
> * Self-review 2: [`reviews/local_015_run-outputs-locations-2.md`](https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_015_run-outputs-locations-2.md)

## Motivation

We have multiple scripts that directly construct paths on S3 and use the `aws` CLI tool like so:

```python
def run_aws_cp(source_path: Path, s3_destination: str, content_type: str = None):
    if source_path.is_dir():
        cmd = ["aws", "s3", "cp", str(source_path), s3_destination, "--recursive"]
    else:
        cmd = ["aws", "s3", "cp", str(source_path), s3_destination]
...

def upload_manifest_to_s3(artifact_group: str, build_dir: Path, bucket_uri: str):
    ...
    dest = f"{bucket_uri}/manifests/{artifact_group}/therock_manifest.json"
    log(f"[INFO] Uploading manifest {manifest_path} -> {dest}")
    run_aws_cp(manifest_path, dest, content_type="application/json")
...

def run(args):
    ...
    external_repo_path, bucket = retrieve_bucket_info()
    run_id = args.run_id
    bucket_uri = f"s3://{bucket}/{external_repo_path}{run_id}-{PLATFORM}"
    bucket_url = (
        f"https://{bucket}.s3.amazonaws.com/{external_repo_path}{run_id}-{PLATFORM}"
    )
```

this moves the path computation and uploading code into shared classes:

```python
def upload_manifest(
    artifact_group: str, build_dir: Path, run_root: WorkflowOutputRoot, backend: StorageBackend,
):
    ...
    backend.upload_file(manifest_path, output_root.manifest(artifact_group))

def run(args):
    ...
    # This handles "retrieve bucket info" as an internal implementation detail
    output_root = WorkflowOutputRoot.from_workflow_run(run_id=args.run_id, platform=PLATFORM)
    # This chooses between LocalStorageBackend and S3StorageBackend depending on if staging_dir is set
    backend = create_upload_backend(staging_dir=args.output_dir, dry_run=args.dry_run)
```

The net result is:

* One file (`workflow_outputs.py`) that documents the directory structure and expected file types that will be uploaded, so upload and download scripts can share paths
* One file (`storage_backend.py`) that handles uploading to S3 via `boto3.client` or "uploading" to a local directory, dropping some uses of the `aws` CLI
* Simpler code and more shared utilities for scripts that want to upload or download files

## Technical Details

I tried to keep this PR as small as possible, but several of the design changes required sweeping updates that would be difficult to split further.

### In scope for this PR

* Defining new interfaces for path and I/O abstractions
* Migrating existing upload scripts for CI logs, artifacts, manifests, python packages to those abstractions
* Migrating existing download scripts for the same
* Adding tests and docs for the above

### Out of scope for this PR

* Migrating `build_tools/github_actions/upload_test_report_script.py`
* Migrating code from release github actions workflow files into scripts. This focuses on just CI workflows, not e.g. https://github.com/ROCm/TheRock/blob/0de4d6b3a84139cf5dd88eca941f3b7651c63a21/.github/workflows/release_portable_linux_packages.yml#L286-L293
  * I did start designing `storage_backend.py` so it might be used here (see `copy_file()` function and some TODOs)
* Moving index HTML generation out of post_build_upload.py (https://github.com/ROCm/TheRock/issues/3331)
* Project-wide refactoring/cleanup discovered as part of this work (there was quite a bit that I had to rein in)

### Feature/architecture highlights

(See also the new `docs/development/run_outputs_layout.md` file)

* `WorkflowOutputRoot`: source of truth for CI output paths
  * `from_workflow_run()` pulls bucket information from the current CI context
    * `retrieve_bucket_info()` is now an implementation detail of that class, not something usable from anywhere in the project via `github_actions_utils.py`
  * `from_workflow_run(lookup_workflow_run=True)` allows fetching from another run's artifacts
  * `for_local()` creates a simulated output path on the local filesystem
  * Methods like `log_dir()`, `python_packages()` return `OutputLocation`
* `StorageLocation`: Backend-agnostic location (relative path + bucket metadata)
  * `.s3_uri`, `.https_url`, `.local_path(staging_dir)`
* `StorageBackend`: base class for uploading
  * `upload_file()`, `upload_directory()` functions that support filtering and specifying content types
  * `S3StorageBackend` uses boto3
  * `LocalStorageBackend` uses a local filesystem (for unit tests and local developer use)

All together, these features enable more natural unit testing and local workflows built on the upload scripts that we already have.

## Test Plan

* New and existing unit tests (+1600 LOC of test code, +800 LOC of non-test code)
* Watch CI workflows to ensure that logs/artifacts/packages/etc. are uploaded as expected
* Watch post-submit multi-arch CI and release workflows (or trigger test runs)

## Submission Checklist

### DIFF
--- a/build_tools/find_artifacts_for_commit.py
+++ b/build_tools/find_artifacts_for_commit.py
@@ -165,10 +165,14 @@ def find_artifacts_for_commit(
     found: dict[str, ArtifactRunInfo] = {}
     for workflow_run in workflow_runs:
         # Bucket info depends only on the workflow run, not the artifact group.
-        external_repo, bucket = retrieve_bucket_info(
+        output_root = WorkflowOutputRoot.from_workflow_run(
+            run_id=str(workflow_run["id"]),
+            platform=platform,
             github_repository=github_repository_name,
             workflow_run=workflow_run,
         )
+        external_repo = output_root.external_repo
+        bucket = output_root.bucket
         for group in artifact_groups:
             if group in found:
                 continue
--- a/build_tools/tests/find_artifacts_for_commit_test.py
+++ b/build_tools/tests/find_artifacts_for_commit_test.py
@@ -13,6 +13,7 @@ from find_artifacts_for_commit import (
     ArtifactRunInfo,
     find_artifacts_for_commit,
 )
+from _therock_utils.workflow_outputs import WorkflowOutputRoot
 from github_actions.github_actions_utils import (
     GitHubAPIError,
     is_authenticated_github_api_available,
@@ -271,14 +272,19 @@ class FindArtifactsCrossRunTest(unittest.TestCase):
     }
 
     @mock.patch("find_artifacts_for_commit.check_if_artifacts_exist")
-    @mock.patch("find_artifacts_for_commit.retrieve_bucket_info")
+    @mock.patch("find_artifacts_for_commit.WorkflowOutputRoot.from_workflow_run")
     @mock.patch("find_artifacts_for_commit.gha_query_workflow_runs_for_commit")
     def test_accumulates_groups_across_runs(
-        self, mock_query_runs, mock_bucket_info, mock_check
+        self, mock_query_runs, mock_from_wfr, mock_check
     ):
         """Groups found across different retriggered runs are accumulated."""
         mock_query_runs.return_value = [self.FAKE_RUN_NEWER, self.FAKE_RUN_OLDER]
-        mock_bucket_info.return_value = ("", "therock-ci-artifacts")
+        mock_from_wfr.return_value = WorkflowOutputRoot(
+            bucket="therock-ci-artifacts",
+            external_repo="",
+            run_id="0",
+            platform="linux",
+        )
 
         # Newer run only built gfx120X; older run built gfx110X.
         def check_by_run_and_group(info):
@@ -305,14 +311,17 @@ class FindArtifactsCrossRunTest(unittest.TestCase):
         self.assertEqual(results[1].workflow_run_id, str(self.FAKE_RUN_NEWER["id"]))
 
     @mock.patch("find_artifacts_for_commit.check_if_artifacts_exist")
-    @mock.patch("find_artifacts_for_commit.retrieve_bucket_info")
+    @mock.patch("find_artifacts_for_commit.WorkflowOutputRoot.from_workflow_run")
     @mock.patch("find_artifacts_for_commit.gha_query_workflow_runs_for_commit")
-    def test_newer_run_takes_priority(
-        self, mock_query_runs, mock_bucket_info, mock_check
-    ):
+    def test_newer_run_takes_priority(self, mock_query_runs, mock_from_wfr, mock_check):
         """When multiple retriggered runs have the same group, the newer wins."""
         mock_query_runs.return_value = [self.FAKE_RUN_NEWER, self.FAKE_RUN_OLDER]
-        mock_bucket_info.return_value = ("", "therock-ci-artifacts")
+        mock_from_wfr.return_value = WorkflowOutputRoot(
+            bucket="therock-ci-artifacts",
+            external_repo="",
+            run_id="0",
+            platform="linux",
+        )
         mock_check.return_value = True  # both runs have all groups
 
         results = find_artifacts_for_commit(


docs/development/run_outputs_layout.md
```diff
@@ -0,0 +1,197 @@
...
+All outputs for a given run live under a common prefix:
+
+```
+s3://{bucket}/{external_repo}{run_id}-{platform}/
+```
+
+| Component       | Example                   | Description                                       |
+| --------------- | ------------------------- | ------------------------------------------------- |
+| `bucket`        | `therock-ci-artifacts`    | Selected based on repo, fork status, release type |
+| `external_repo` | `""` or `"Fork-TheRock/"` | Non-empty for forks and non-TheRock repos         |
```
marbre: Is "Fork-TheRock/" a valid example we see in the wild?

docs/development/workflow_outputs.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Directory structure
+
+```
+{prefix}/
+    {artifact_name}_{component}_{target_family}.tar.xz
+    {artifact_name}_{component}_{target_family}.tar.xz.sha256sum
+    index-{artifact_group}.html
+
+    logs/{artifact_group}/
```
marbre, line 58: In the artifacts `target_family` is used but the subdirectory will be `artifact_group`. From the code this looks like the same eventually? The document could state the difference if there is one here as this is at least right now confusing to me.

Furthermore, we need to check how this aligns with multi-arch.

docs/development/workflow_outputs.md
```diff
@@ -0,0 +1,197 @@
...
+```
+RELEASE_TYPE set?
+├─ Yes ──> therock-{RELEASE_TYPE}-artifacts
+└─ No
+   └─ ROCm/TheRock and not fork?
+      ├─ Yes
+      │  └─ Pre-cutover? ─Yes─> therock-artifacts
+      │                  ─No──> therock-ci-artifacts
+      └─ No
+         └─ ROCm/therock-releases-internal and not fork?
```
marbre, line 95: Not sure if `ROCm/therock-releases-internal` is used anymore.

build_tools/_therock_utils/run_outputs.py
```diff
@@ -0,0 +1,394 @@
...
+
+    # -- Python packages --------------------------------------------------------
+
+    def python_packages(self, artifact_group: str) -> OutputLocation:
+        """Location for the Python packages directory.
+
+        Args:
+            artifact_group: Build variant (e.g., 'gfx110X-all')
+        """
+        return OutputLocation(self.bucket, f"{self.prefix}/python/{artifact_group}")
```
marbre: Should this be fixed to `{self.prefix}/python/{artifact_group}` or do we have a way to make this more flexible, e.g. `whl` vs / and `whl-next` for example with regards to multi-arch.

build_tools/_therock_utils/workflow_outputs.py
```diff
@@ -0,0 +1,394 @@
...
+        bucket = f"therock-{release_type}-artifacts"
+    else:
+        if external_repo == "":
+            bucket = "therock-ci-artifacts"
+            if curr_commit_dt and curr_commit_dt <= _BUCKET_CUTOVER_DATE:
+                bucket = "therock-artifacts"
+        elif (
+            repo_name == "therock-releases-internal"
+            and owner == "ROCm"
+            and not is_pr_from_fork
```
marbre, line 340: This might no longer be actively used and we can eventually drop it.

build_tools/_therock_utils/storage_backend.py
```diff
@@ -0,0 +1,222 @@
...
+# ---------------------------------------------------------------------------
+
+# Explicit content-type overrides for extensions we know about.
+_CONTENT_TYPE_OVERRIDES: dict[str, str] = {
+    ".gz": "application/gzip",
+    ".log": "text/plain",
+    ".md": "text/plain",
+    ".xz": "application/x-xz",
+    ".zst": "application/zstd",
+}
```
marbre, line 49: What about `.whl` and `.html`?

build_tools/_therock_utils/artifact_backend.py
```diff
@@ -83,23 +85,30 @@ def base_uri(self) -> str:
...
-        {staging_dir}/run-{run_id}-{platform}/
+    Directory structure mirrors S3 layout via RunOutputRoot::
+
+        {staging_dir}/{run_root.prefix}/
             {artifact_name}_{component}_{target_family}.tar.zst
-            {artifact_name}_{component}_{target_family}.tar.zst.sha256sum
     """
 
-    def __init__(self, staging_dir: Path, run_id: str, platform: str = "linux"):
+    def __init__(self, staging_dir: Path, run_root: RunOutputRoot):
```
marbre: I kind of associate `run_root` with run-as-root but naming might be fine here though.

marbre (commented):
From a high-level this looks good to me and I see no major issues.

docs/development/workflow_outputs.md
ScottTodd: I think it's safe to remove, but keeping it for now so this change is as close to existing behavior as possible.

build_tools/_therock_utils/run_outputs.py
ScottTodd: Keeping as-is for now to avoid behavior changes in this PR but yeah, we may need to adjust for multi-arch.

build_tools/_therock_utils/artifact_backend.py
ScottTodd: Good point. I considered a few other names for this "root of all outputs for one CI workflow run, stored in a cloud bucket". Which do you like?

Class name | Variable name 
-- | --
RunOutputRoot | run_root
RunOutputStore | run_store
OutputRoot | output_root
WorkflowOutputRoot | workflow_root
WorkflowRunOutputs | run_outputs
RunOutputs | run_outputs

docs/development/workflow_outputs.md
ScottTodd: It's a bit complicated... I added a comment with some more details.

build_tools/_therock_utils/artifact_backend.py
ScottTodd: Pushed some changes renaming the other modules/classes to better support using these new classes to replace upload code in workflows like
```yml
      - name: Upload Releases to S3
        if: ${{ github.repository_owner == 'ROCm' }}
        run: |
          aws s3 cp ${{ env.DIST_ARCHIVE }} s3://${{ env.S3_BUCKET_TAR }}/${{ env.S3_SUBDIR_TAR }}
          aws s3 cp ${{ env.OUTPUT_DIR }}/packages/dist/ s3://${{ env.S3_BUCKET_PY }}/${{ env.S3_SUBDIR }}/${{ matrix.target_bundle.amdgpu_family }}/ \
          --recursive --no-follow-symlinks \
          --exclude "*" \
          --include "*.whl" \
          --include "*.tar.gz"
```

build_tools/_therock_utils/storage_location.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @property
+    def s3_uri(self) -> str:
+        """S3 URI (e.g., ``s3://bucket/path/file``)."""
+        return f"s3://{self.bucket}/{self.relative_path}"
+
+    @property
+    def https_url(self) -> str:
+        """Public HTTPS URL for browser access."""
+        return f"https://{self.bucket}.s3.amazonaws.com/{self.relative_path}"
```
ScottTodd, line 50: I had some ideas here after reviewing https://github.com/ROCm/TheRock/pull/3663 from @jayhawk-commits :

We might want to move this to an explicit `s3_https_url` and add a new `https_url` implementation that can return a custom URL like our current CloudFront pages. That would let us include the CDN-enabled URLs in our logs instead of just the raw S3 URLs (which may or may not be accessible, depending on the bucket).

We could pass through the URL mappings explicitly, inject them via a configuration file (loaded from an env var?), or do something else. For example:

```jsonc
// bucket_url_mappings.json
{
  "therock-dev-tarball": "https://rocm.devreleases.amd.com/tarball/",
  "therock-nightly-tarball": "https://rocm.nightlies.amd.com/tarball/",
}
```

```python
# storage_location.py or some other file
url_mappings = {}
with open(default_bucket_url_mappings_file) as f:
    url_mappings = json.load(f)

extra_bucket_url_mappings_file = os.environ.get("THEROCK_EXTRA_BUCKET_MAPPINGS_FILE", "")
if extra_bucket_url_mappings_file:
  with open(extra_bucket_url_mappings_file) as f:
    url_mappings = url_mappings | json.load(f)
```

build_tools/_therock_utils/storage_backend.py
```diff
@@ -0,0 +1,276 @@
...
+    return guessed or _DEFAULT_CONTENT_TYPE
+
+
+# ---------------------------------------------------------------------------
+# StorageBackend ABC
+# ---------------------------------------------------------------------------
+
+
+class StorageBackend(ABC):
+    """Abstract base class for storage operations."""
```
ScottTodd, line 73: I'm liking the name for this class more and more, as I consider what else in the project we could use this abstraction for. Having a common class (with unit tests and docs) will make it significantly easier to match local/dry-run behavior to behavior in github actions workflows.

Places to use this:
* Artifact/log/package upload for CI workflows/scripts
* Artifact download for CI workflows/scripts
* Release package copying (CI bucket -> release bucket v2-staging -> release bucket v2)

We could add a config struct with:
* Retry logic (max retries, back-off method)
* Upload logic (max parallel uploads, chunking)
* Download logic (max parallel downloads, chunking, checksum validation, caching)
* Logging settings (output to file, verbosity)

We can add new methods:
* `download_file`
* `download_directory` (and/or `_files`)

build_tools/_therock_utils/artifact_backend.py
marbre: > Good point. I considered a few other names for this "root of all outputs for one CI workflow run, stored in a cloud bucket". Which do you like?
> Class name 	Variable name
> RunOutputRoot 	run_root
> RunOutputStore 	run_store
> OutputRoot 	output_root
> WorkflowOutputRoot 	workflow_root
> WorkflowRunOutputs 	run_outputs
> RunOutputs 	run_outputs

`OutputRoot` 	`output_root` might be a good choice?

ScottTodd: Hmmm.... just `OutputRoot` feels too generic to me.

docs/development/workflow_outputs.md
marbre: Sure, can clean up in a follow up.

build_tools/_therock_utils/storage_location.py
marbre: That's actually a nice idea. Do you tend to land it as part of this change or do you plan for a follow up?

ScottTodd: Follow-up. This PR is large enough already. Could at least do the simple rename from `https_url` to `s3_https_url` though.

marbre (approved):
From a high-level this looks good and is a huge improvement. Due to the size of this PR there are clearly things we _could_ improve (see the Claude reviews and suggestions) but I think it is well designed and tested enough to land as is and iterate further once merged to also not block others on their work.

build_tools/_therock_utils/artifact_backend.py
marbre: `WorkflowOutputRoot` works for me as well (could still use `output_root` as variable :laughing: ). No too strong feelings.

marbre: Actually, looking at the naming, as log as the variable does not start with `run_` I am fine :) Can either be `RunOutputRoot`, `output_root` or `RunOutputStore` `output_store` or `WorkflowOutputRoot`, `output_root` (in that nomenclature always skipping the first part of the classname).

ScottTodd: I'll go with `WorkflowOutputRoot` and either `output_root` or `workflow_output_root` then.

ScottTodd: Pushed a renaming. Will wait until tomorrow to merge.

build_tools/_therock_utils/storage_backend.py
```diff
@@ -0,0 +1,276 @@
...
+            dest: Destination location (the directory root in the backend).
+            include: Optional glob patterns to filter files (e.g.
+                ``["*.tar.xz*"]``). If ``None``, all files are uploaded.
+
+        Returns:
+            Number of files uploaded.
+
+        Symlinks are skipped. Subdirectory structure is preserved.
+
+        TODO(scotttodd): Upload files in parallel (perhaps in the derived classes)
```
ScottTodd: Whoops, going to need parallel upload or some tuning on the upload settings. The uploads are taking ~58 minutes: https://github.com/ROCm/TheRock/actions/runs/22602649569/job/65487839754?pr=3596#step:15:168, up from 6 minutes prior to these changes: https://github.com/ROCm/TheRock/actions/runs/22608341595/job/65505405482

ScottTodd:
A few changes (including merge conflict resolution) since you last approved. Want to review again @marbre ?

marbre (approved):
Looks good!

HereThereBeDragons:
This PR failed building pytorch in this nightly as boto3 is missing. The pytorch workflows are not installing any `requirements.txt`.

```
Upload PyTorch manifest to S3 artifacts       0s
Run python build_tools/github_actions/upload_pytorch_manifest.py \
  python build_tools/github_actions/upload_pytorch_manifest.py \
    --dist-dir "/home/runner/_work/TheRock/TheRock/output/packages/dist" \
    --run-id "22708867520" \
    --amdgpu-family "gfx1151" \
    --python-version "3.10" \
    --pytorch-git-ref "release/2.8"
```

```
  File "/__w/TheRock/TheRock/build_tools/github_actions/upload_pytorch_manifest.py", line 145, in <module>
    main(sys.argv[1:])
  File "/__w/TheRock/TheRock/build_tools/github_actions/upload_pytorch_manifest.py", line 141, in main
    backend.upload_file(manifest_path, dest)
  File "/__w/TheRock/TheRock/build_tools/_therock_utils/storage_backend.py", line 268, in upload_file
    self.s3_client.upload_file,
  File "/__w/TheRock/TheRock/build_tools/_therock_utils/storage_backend.py", line 215, in s3_client
    import boto3
ModuleNotFoundError: No module named 'boto3'
Error: Process completed with exit code 1.
```

ScottTodd:
Argh, thanks for noticing and sending the fix-forward.

Brainstorming some ideas for how to catch issues with python requirements in workflows for the future:
1. More careful review
2. Add static analysis / unit tests checking for pip installs in workflows that run scripts
3. Have scripts install what they need on demand: https://docs.astral.sh/uv/guides/scripts/#running-a-script-with-dependencies
4. Install pip requirements into build dockerfiles (basically replacing the system-level awscli with boto3 in this case)
    * Then cache dockerfiles / layers on the build VMs
5. Run full release workflows when changing them
    * I'm deliberately trying to steer us away from this - while that's the most reliable way to know that workflows are still functional, the time to signal is far too long for productive development.
    * Running partial release workflows (for a light/cheap target or a no-op somehow) could work too

I'm thinking we should focus on (4), but (1) can be a crutch too.

HereThereBeDragons:
More careful reviews are always better :)

Dockerfile will only help us so and so much - and we would need to actively monitor if a newer version is available and build a new image. I think the main argument would be if we could cache it locally to not have so much network traffic and delays.

Overall I am a fan to also have some manual workflow runs for such large changes. Especially if the script changes are just touching pytorch:  just use the latest nightly and run the pytorch release workflow. Takes around 30-40min - which is totally within a reasonable time frame. 

Maybe we could have a smarter logic based on the files being changed and in which workflows they are called to automatically build jax/torch/native linux based on the current nightly release and not the full build to get earlier signaling on those exterior layers of rocm? (in addition to the normal "build presubmit rocm" workflows)

As we also have green CI requirements before merging, so we have to wait at least for one ROCm build.

PR #3597, opened 2026-02-25
Updating Releases.md to reflect installation instruction for native packages

ammallya:
## Motivation

To add install instructions for packages generated by Nightly CI runs.

## Technical Details

Updating Releases.md to reflect installation instruction for native packages

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -683,7 +683,8 @@ sudo apt install -y ca-certificates
 echo "deb [trusted=yes] https://rocm.nightlies.amd.com/deb/${RELEASE_ID} stable main" \
   | sudo tee /etc/apt/sources.list.d/rocm-nightly.list
 sudo apt update
-sudo apt install amdrocm-${GFX_ARCH} amdrocm-core-sdk-${GFX_ARCH}
+sudo apt install amdrocm-core-sdk-${GFX_ARCH}
+# If only runtime is needed, install amdrocm-${GFX_ARCH} instead
 ```
 
 ### Installing on RPM-based systems (RHEL, SLES, AlmaLinux etc.)
@@ -711,7 +712,8 @@ enabled=1
 gpgcheck=0
 priority=50
 EOF
-sudo dnf install amdrocm-${GFX_ARCH} amdrocm-core-sdk-${GFX_ARCH}
+sudo dnf install amdrocm-core-sdk-${GFX_ARCH}
+# If only runtime is needed, install amdrocm-${GFX_ARCH} instead
 ```
 
 ## Verifying your installation


RELEASES.md
```diff
@@ -491,9 +490,72 @@ ls install
...
+
+```bash
+sudo apt update
+sudo apt install -y ca-certificates
+```
+
+2. Add the nightly repository
+
+```bash
+echo "deb [trusted=yes] https://rocm.nightlies.amd.com/deb/<YYYYMMDD>-<action-run-id> stable main" \
```
marbre: @nunnikri Looking at the above instructions, would it make sense to drop the `<YYYYMMDD>-<action-run-id>` prefix and have nightlies structured like prereleases?

nunnikri: @marbre  if we set folder structure like prerelease. Setting repo will be easy. But if QA wants to test previous day package, then install command have to be like
`dnf install amdrocm-gfx94x-7.12.0~<YYYYMMDD>-<action-run-id>`

Just giving dnf install amdrocm-gfx94x will always install the latest build. After discussing with QA we decided current folder structure is more convenient for their daily test. But I am open for any changes needed. We have time till 8.0 to do any changes needed. 
@amd-aakash FYI

RELEASES.md
```diff
@@ -417,7 +417,6 @@ such as setting environment variables.
 > For most users, we recommend installing via package managers:
 >
 > - [Installing releases using pip](#installing-releases-using-pip)
-> - (TODO) Installing native Linux deb/RPM packages
```
ScottTodd, line 543: Keep the line but remove `TODO`?

Should be something like 
```markdown
> For most users, we recommend installing via package managers:
>
> - [Installing releases using pip](#installing-releases-using-pip)
> - [Installing native Linux deb/RPM packages](#installing-from-native-packages)
```

RELEASES.md
```diff
@@ -491,9 +490,72 @@ ls install
 > #     ...
 > ```
 
+## Installing from native packages
```
ScottTodd, line 621: Add to the table of contents at the top too: https://github.com/ROCm/TheRock/blob/a5f10cc96ad56f46de60216c7abd68b40a5213c2/RELEASES.md?plain=1#L22-L35

I think we should also sort this above tarballs:
* Python packages from the pip/uv package managers (Windows + Linux)
* Native packages from the apt/dnf package managers (currently just Linux)
* Underlying/raw tarballs and artifacts (Windows + Linux)

RELEASES.md
```diff
@@ -491,9 +490,72 @@ ls install
...
+
+In addition to Python wheels and tarballs, ROCm native Linux packages are
+published for Debian-based and RPM-based distributions.
+
+> [!WARNING]
+> These builds are primarily intended for development and testing and are currently **unsigned**.
```
ScottTodd, line 627: What are the actual implications of these being unsigned? Is that significantly different from our python packages not being signed either? I'm wondering if we should move this warning to the top of the file to apply to _all_ release types, instead of call it out specifically for native linux packages.

RELEASES.md
amd-aakash: @marbre : I agree with @nunnikri ’s assessment. Unlike pre-releases we will have use case for QA to be able test different nightly builds hence we will keep the current convention. We can revisit as we go forward and if any other alternative suits the need better.

RELEASES.md
nunnikri: Done

RELEASES.md
nunnikri: It not really a security concern. But more for telling its a legitimate package created by AMD itself. So that customers feel more comfortable in using the package. We are planning to enable signing for native packages in nightly build soon. We are singing manually for release/prerelease area as of now

RELEASES.md
```diff
@@ -491,9 +495,415 @@ ls install
...
+
+### Native packages release status
+
+| Platform |                                                                                                                                                                                                                     Native Linux packages |
+| -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
+| Linux    | [![Build Native Linux Packages](https://github.com/ROCm/TheRock/actions/workflows/build_native_linux_packages.yml/badge.svg?branch=main)](https://github.com/ROCm/TheRock/actions/workflows/build_native_linux_packages.yml?query=branch%3Amain) |
+| Windows  |                                                                                                                                                                                                                      (Coming soon)           |
```
ScottTodd: update header to "Native packages" instead of "Native Linux packages" if there's a row for Windows. (then format with mdformat / pre-commit)
```suggestion
| Platform |                                                                                                                                                                                                                     Native packages |
| -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
| Linux    | [![Build Native Linux Packages](https://github.com/ROCm/TheRock/actions/workflows/build_native_linux_packages.yml/badge.svg?branch=main)](https://github.com/ROCm/TheRock/actions/workflows/build_native_linux_packages.yml?query=branch%3Amain) |
| Windows  |                                                                                                                                                                                                                      (Coming soon)           |
```

RELEASES.md
ScottTodd: I think we shouldn't list `deb for gfx####` exhaustively for each GPU [family] if the instructions are not copy/paste friendly. At the point where someone already needs to type in `<YYYYMMDD>-<action-run-id>`, they might as well also type in `gfx####`.

We could structure as
```bash
export RELEASE_DATE=YYYYMMDD
export ACTION_RUN_ID=12345
export GFX_ARCH=gfx####

sudo apt update
sudo apt install -y ca-certificates
echo "deb [trusted=yes] https://rocm.nightlies.amd.com/deb/${RELEASE_DATE}-${ACTION_RUN_ID} stable main" \
  | sudo tee /etc/apt/sources.list.d/rocm-nightly.list
sudo apt update
sudo apt install amdrocm-${GFX_ARCH}
```

nunnikri: Thats a great suggestion. Let me incorporate that.

nunnikri: Done

RELEASES.md
```diff
@@ -614,9 +618,103 @@ ls install
...
+### Installing on Debian-based systems (Ubuntu, Debian, etc.)
+
+```bash
+# Step 1: Find the latest release from https://rocm.nightlies.amd.com/deb/
+#         Look for directories like "20260310-12345678"
+# Step 2: Find your package name from the table above
+# Step 3: Set the variables below
+
+export RELEASE_ID=20260310-12345678  # Replace with actual date-runid
+export GFX_ARCH=gfx110x              # Replace with your arch (e.g., gfx94x, gfx110x, gfx1151)
```
dbieleck: Using "arch" doesn't clearly point out to a user that they need to actually look at the GFX Family in the table above.  Maybe point that out.  You could just change "GPU family and package mapping" on line 636 and say "GPU family and gpu architecture package mapping".  I would also say architecture instead of "arch" in the comment.

RELEASES.md
```diff
@@ -614,9 +618,103 @@ ls install
...
+
+# Step 4: Add repository and install
+sudo dnf install -y ca-certificates
+sudo tee /etc/yum.repos.d/rocm-nightly.repo <<EOF
+[rocm-nightly]
+name=ROCm Nightly Repository
+baseurl=https://rocm.nightlies.amd.com/rpm/${RELEASE_ID}/x86_64
+enabled=1
+gpgcheck=0
+repo_gpgcheck=0
```
dbieleck: I don't think you need this.  In the main docs we set priority=50 instead.  You can add that to match.

nunnikri: Done

RELEASES.md
nunnikri: Updated. please take a look

RELEASES.md
```diff
@@ -614,9 +618,105 @@ ls install
...
+export RELEASE_ID=20260310-12345678  # Replace with actual date-runid
+export GFX_ARCH=gfx110x              # Replace with GFX Family from the mapping table
+
+# Step 4: Add repository and install
+sudo apt update
+sudo apt install -y ca-certificates
+echo "deb [trusted=yes] https://rocm.nightlies.amd.com/deb/${RELEASE_ID} stable main" \
+  | sudo tee /etc/apt/sources.list.d/rocm-nightly.list
+sudo apt update
+sudo apt install amdrocm-${GFX_ARCH} amdrocm-core-sdk-${GFX_ARCH}
```
dbieleck: If you install amdrocm-core-sdk-gfxXYZ, you don't need to install amdrocm-gfxXYZ, since core-sdk includes the packages in amdrocm.

RELEASES.md
```diff
@@ -614,9 +618,105 @@ ls install
...
+sudo dnf install -y ca-certificates
+sudo tee /etc/yum.repos.d/rocm-nightly.repo <<EOF
+[rocm-nightly]
+name=ROCm Nightly Repository
+baseurl=https://rocm.nightlies.amd.com/rpm/${RELEASE_ID}/x86_64
+enabled=1
+gpgcheck=0
+priority=50
+EOF
+sudo dnf install amdrocm-${GFX_ARCH} amdrocm-core-sdk-${GFX_ARCH}
```
dbieleck: You don't need amdrocm if you install amdrocm-core-sdk.

nunnikri: done

PR #3606, opened 2026-02-25
[debug-tools/rocgdb] Switch rocgdb submodule to amd-staging-rocgdb-16 branch

lumachad:
Switch from using amd-staging to using amd-staging-rocgdb-16 branch, as it is more stable than amd-staging.


lumachad:
@lancesix Updated now to pick up 0f4c21549e8883e3151a06db067a56901bb26375.

Issue #3337, opened 2026-02-09
[Multi-arch] Enable multi-arch CI on pre-submit


marbre:

ScottTodd:
While working on https://github.com/ROCm/TheRock/issues/3399, I've been building a new configure_multi_arch_ci.py and setup_multi_arch.yml. I plan on putting some policies in there like:

* If a `run-multi-arch-ci` label is added to a PR, run multi-arch CI
* If multi-arch-relevant files like `BUILD_TOPOLOGY.toml` are modified, enable multi-arch CI
* Initially only build Linux and do not run tests on multi-arch CI on PRs
  * This will decrease load on Linux build/test runners from overlapping coverage
  * Windows builds are still slow since they have no build cache (see https://github.com/ROCm/TheRock/issues/3622 and https://github.com/ROCm/TheRock/issues/902)

As the feature gap closes between "CI" and "Multi-arch CI", we can adjust those policies and make the switch

ScottTodd:
We're hoping to switch over from "CI" to "Multi-arch CI" as the default workflow that runs on pull requests as soon as this week.

Rather than switch gradually with an "implicit opt-in to running multi-arch CI on pull_request based on files modified", I now think we should instead:
1. Complete MVP tasks to get multi-arch CI workable:
    * https://github.com/ROCm/TheRock/pull/4161
    * https://github.com/ROCm/TheRock/pull/4168
    * https://github.com/ROCm/TheRock/pull/4169
    * https://github.com/ROCm/TheRock/issues/4184
2. Check that multi-arch and ci have the expected GPU family coverage.
3. Run "multi-arch CI" on all pull requests, drop "CI" down to opt-in based on a `ci:run-old-ci` label or equivalent (to let us debug issues during the migration). Continue running both workflows on `push` triggers. Continue running the existing release workflows

> [!IMPORTANT]
> Any features that are limited to "CI" instead of "multi-arch CI" will no longer be tested on pull requests and will need to migrate or risk being deleted as part of https://github.com/ROCm/TheRock/issues/3340.

Risks:
* Release workflows are still single-stage, not based on multi-arch workflows.
  * This (https://github.com/ROCm/TheRock/issues/3334) needs to be worked on soon, but we can't do it easily without continuous CI coverage keeping the builds working, it's too much of a moving target
* rocm-libraries and rocm-systems are still using forked single-stage workflows
[...]

PR #1714, opened 2025-10-07
[hipDNN] Third party deps install for hipDNN

SamuelReeder:
## Motivation

hipDNN needs spdlog, flatbuffers, nlohmann json, and fmt as runtime deps for its installed header-only libs to be easily consumed. This PR enable this by installing each of them with hipDNN automatically. 

## Technical Details

Calls the `therock_provide_artifact` on hipDNN's third party runtime deps when hipDNN is enabled.

#1689 (already merged) is a prerequisite for this PR.

## Test Plan

I've conducted manual testing so far by consuming hipDNN from TheRock install. However, any further recommendations are welcome.

## Test Result

Runtime deps are installed if hipDNN is enabled. Consumers can compile with the installed hipDNN SDK and FE headers.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3609, opened 2026-02-25
[hipDNN] Removing unnecessary dependencies on spdlog / fmt for dnn-providers

mousdahl-amd:
## Motivation
We are in the process of cleaning up the required dependencies for our header-only hipdnn_frontend library.  Since the requirement for spdlog / fmt has been removed from the frontend, we're able to remove those dependencies from TheRock CI

## Technical Details
- Moved sdplog / fmt from a runtime dep to a build dep for:
    - hipDNN (required for backend lib still)
    - rocprofiler-systems
- Removed spdlog / fmt as a build dep for:
   - MIOpen provider
   - hipBLASlt provider
   - Fusilli provider
   - hipDNN samples
 - Removed artifact install for spdlog / fmt 

## Test Plan
- Run TheRock build on CI

## Test Result
- [ ] Build succeeds

## Submission Checklist


SamuelReeder (approved):
It looks good to me. However, I wonder if we should remove the universal installs of these deps (fmt, spdlog) that were introduced in: https://github.com/ROCm/TheRock/pull/1714

We may want to consider removing them if no other projects depend on them, but it's probably fine leaving it unchanged for now.

BrianHarrisonAMD (approved):
LGTM. I agree with Samuel, but we can make that a separate follow-up also.

ScottTodd (approved):
Nice, thanks for the cleanup. I don't see any issues with removing these artifacts.

Let's watch Multi-arch CI (https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci.yml?query=branch%3Amain) after merge to ensure that the topology is correct there too. (we could do that prior to merge too, but I don't mind the extra risk here... still need to flip the default: https://github.com/ROCm/TheRock/issues/3337)

Issue #3451, opened 2026-02-17
Add `iree-libs` stage to multi-arch CI pipeline


stellaraccident:
The `BUILD_TOPOLOGY.toml` already defines `iree-libs` as a generic build stage (added in the initial fusilli integration and refined in #3264). Once #3264 lands, this stage needs a corresponding job in the multi-arch CI workflow so it actually runs.

Everything in the topology and source-set plumbing is already wired up — the only missing piece is the workflow job itself.

## What's already in place

The full chain from stage → source sets → submodules is configured:

```
BUILD_TOPOLOGY.toml:
  [build_stages.iree-libs]           →  artifact_groups = ["iree-libs"]
  [artifact_groups.iree-libs]        →  type = "generic", source_sets = ["iree-libs"]
  [source_sets.iree-libs]            →  submodules = ["iree", "fusilli"]
  [artifacts.iree-compiler]          →  artifact_group = "iree-libs", type = "target-neutral"
  [artifacts.fusilli-plugin]         →  artifact_group = "iree-libs", type = "target-neutral"
```

Source fetching also works: `fetch_sources.py --stage iree-libs` resolves the topology chain and fetches `iree` + `fusilli` submodules, including IREE's nested submodules (`third_party/llvm-project`, `third_party/stablehlo`, etc.) via the `--nested-submodules` defaults (updated in #3264).

## What needs to change

**File:** `.github/workflows/multi_arch_build_portable_linux.yml`

Add an `iree-libs` job following the same pattern as `dctools-core` or `profiler-apps` (both are generic stages that depend on `compiler-runtime` and run in parallel with `math-libs`):

```yaml
  # ==========================================================================
  # STAGE: iree-libs (generic, parallel to math-libs)
  # ==========================================================================
  iree-libs:
    needs: compiler-runtime
    uses: ./.github/workflows/multi_arch_build_portable_linux_artifacts.yml
    secrets: inherit
    with:
      stage_name: iree-libs
      stage_display_name: "Stage - IREE Libs"
      timeout_minutes: 180  # 3 hours - IREE compiler build is substantial
      dist_amdgpu_families: ${{ inputs.dist_amdgpu_families }}
      rocm_package_version: ${{ inputs.rocm_package_version }}
    permissions:
      contents: read
      id-token: write
```

Key points:
- **`needs: compiler-runtime`** — iree-libs depends on `hip-runtime` artifacts which are produced by the `compiler-runtime` stage
[...]

PR #3611, opened 2026-02-25
Split `iree-libs` into `iree-compiler` + `fusilli-libs` stages and add to Multi-Arch CI pipeline

AaronStGeorge:
## Motivation

Closes: https://github.com/ROCm/TheRock/issues/3451

Adds IREE/fusilli related stages to the multi-arch CI pipeline and splits `iree-compiler` and `fusilli-libs` into separate build stages. `iree-compiler` requires an entire LLVM build, so there's a significant savings by overlapping with `math-libs` etc. stages.

## Technical Details

Changes:
  - Split `[build_stages.iree-libs]` and `[artifact_groups.iree-libs]` into separate `iree-compiler` and `fusilli-libs`  stages in `BUILD_TOPOLOGY.toml`
  - Added corresponding workflow jobs in `multi_arch_build_portable_linux.yml`

## Test Plan

  - CI pipeline runs the new iree-compiler and fusilli-libs stages end-to-end
  - fusilliprovider test job fetches artifacts and runs test_fusilliprovider.py

## Test Result

The fusilli build is currently OFF in `TheRock`, but by temporarily turning it ON we can see a successful build/test [here](https://github.com/ROCm/TheRock/actions/runs/22513063330/job/65231593574?pr=3611). 

## Submission Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,7 +241,8 @@ option(THEROCK_ENABLE_DC_TOOLS "Enable building of data center tools" "${THEROCK
 option(THEROCK_ENABLE_MEDIA_LIBS "Enable building of Media libraries" "${THEROCK_ENABLE_ALL}")
 option(THEROCK_ENABLE_HOST_MATH "Build all bundled host math libraries by default" OFF)
 option(THEROCK_RESET_FEATURES "One-shot flag which forces all feature flags to their default state for this configuration run" OFF)
-option(THEROCK_ENABLE_IREE_LIBS "Enable building of IREE libraries" "${THEROCK_ENABLE_ALL}")
+# TODO(iree-org/fusilli/issues/57): Enable fusilli build once multi-arch support lands.
+option(THEROCK_ENABLE_IREE_LIBS "Enable building of IREE libraries" OFF)
 
 ################################################################################
 # Artifact-based features are auto-generated from BUILD_TOPOLOGY.toml
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -352,14 +352,14 @@ test_matrix = {
     },
     # TODO(iree-org/fusilli/issues/57): Enable fusilli tests once build is
     # enabled by default.
-    "fusilliprovider": {
-        "job_name": "fusilliprovider",
-        "fetch_artifact_args": "--hipdnn --fusilliprovider --iree-compiler --tests",
-        "timeout_minutes": 15,
-        "test_script": f"python {_get_script_path('test_fusilliprovider.py')}",
-        "platform": ["linux"],
-        "total_shards": 1,
-    },
+    # "fusilliprovider": {
+    #     "job_name": "fusilliprovider",
+    #     "fetch_artifact_args": "--hipdnn --fusilliprovider --iree-compiler --tests",
+    #     "timeout_minutes": 15,
+    #     "test_script": f"python {_get_script_path('test_fusilliprovider.py')}",
+    #     "platform": ["linux"],
+    #     "total_shards": 1,
+    # },
     # hipBLASLt provider tests
     "hipblasltprovider": {
         "job_name": "hipblasltprovider",


CMakeLists.txt
```diff
@@ -241,8 +241,7 @@ option(THEROCK_ENABLE_DC_TOOLS "Enable building of data center tools" "${THEROCK
 option(THEROCK_ENABLE_MEDIA_LIBS "Enable building of Media libraries" "${THEROCK_ENABLE_ALL}")
 option(THEROCK_ENABLE_HOST_MATH "Build all bundled host math libraries by default" OFF)
 option(THEROCK_RESET_FEATURES "One-shot flag which forces all feature flags to their default state for this configuration run" OFF)
-# TODO(iree-org/fusilli/issues/57): Enable fusilli build once multi-arch support lands.
-option(THEROCK_ENABLE_IREE_LIBS "Enable building of IREE libraries" OFF)
+option(THEROCK_ENABLE_IREE_LIBS "Enable building of IREE libraries" "${THEROCK_ENABLE_ALL}")
```
AaronStGeorge: Temporary, uncommented just to show things [work](https://github.com/ROCm/TheRock/actions/runs/22513063330/job/65231593574?pr=3611). This will be enabled for real when the multi-arch pipeline becomes the default.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -352,14 +352,14 @@ def _get_script_path(script_name: str) -> str:
...
-    #     "total_shards": 1,
-    # },
+    "fusilliprovider": {
+        "job_name": "fusilliprovider",
+        "fetch_artifact_args": "--hipdnn --fusilliprovider --iree-compiler --tests",
+        "timeout_minutes": 15,
+        "test_script": f"python {_get_script_path('test_fusilliprovider.py')}",
+        "platform": ["linux"],
+        "total_shards": 1,
+    },
```
AaronStGeorge: Temporary, uncommented just to show things [work](https://github.com/ROCm/TheRock/actions/runs/22513063330/job/65231593574?pr=3611). This will be enabled for real when the multi-arch pipeline becomes the default.

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -175,6 +177,40 @@ jobs:
...
+    secrets: inherit
+    with:
+      stage_name: fusilli-libs
+      stage_display_name: "Stage - Fusilli Libs"
+      timeout_minutes: 60  # 1 hour
+      dist_amdgpu_families: ${{ inputs.dist_amdgpu_families }}
+      rocm_package_version: ${{ inputs.rocm_package_version }}
+    permissions:
+      contents: read
+      id-token: write
```
ScottTodd, line 212: Does the fusilli-libs stage build on Windows yet? We'll also need edits here when it does: https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows.yml

build_tools/github_actions/fetch_test_configurations.py
ScottTodd: Please try to keep temporary changes out of non-draft PRs. Including such changes adds extra review complexity (need to judge "how much do I trust that these changes will actually be removed prior to merge" and "how much do I trust that the CI results won't materially change after removing them", etc.)

It's a bit of extra effort, but you could push to both
1. A test branch
2. A review branch

iree-libs/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 ##############################################################################
 if(THEROCK_ENABLE_IREE_COMPILER)
   therock_cmake_subproject_declare(iree-compiler
+    USE_DIST_AMDGPU_TARGETS
     NO_MERGE_COMPILE_COMMANDS
     BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/iree-compiler"
     EXTERNAL_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/iree"
```
ScottTodd, line 13: The compiler shouldn't need a list of AMDGPU targets from `USE_DIST_AMDGPU_TARGETS`... right? What would that be used for?

(I could see including tuning specs for a filtered list of targets, but I don't believe any code like that exists in IREE itself yet)

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -175,6 +177,40 @@ jobs:
...
+  # STAGE: iree-compiler (generic, parallel to math-libs)
+  # ==========================================================================
+  iree-compiler:
+    needs: compiler-runtime
+    uses: ./.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+    secrets: inherit
+    with:
+      stage_name: iree-compiler
+      stage_display_name: "Stage - IREE Compiler"
+      timeout_minutes: 180  # 3 hours - includes captive LLVM build
```
ScottTodd: 3 hour timeout seems excessive for this. Drop to 60 or 90 minutes?

Logs at https://github.com/ROCm/TheRock/actions/runs/22513062870/job/65476655032?pr=3611 have this stage taking 4 minutes with 96% cache hits. I would expect on these build machines to not get above 45 minutes with a cold cache.

If keeping this, I don't understand the "includes _captive_ LLVM build" comment. Reword?

build_tools/github_actions/fetch_test_configurations.py
AaronStGeorge: 👍 I'll remove the temporary changes before the next round of review.

.github/workflows/multi_arch_build_portable_linux.yml
marbre: Agreed, please reword.

build_tools/github_actions/fetch_test_configurations.py
marbre: Why should this only be enabled _after_ multi-arch CI becomes the default? multi-arch is already running on post-submit.

iree-libs/CMakeLists.txt
marbre: This is passed to the other compiler builds but we should indeed clarify why this is needed for IREE as well if it is.

CMakeLists.txt
marbre: Why do we want to limit this to multi-arch becoming the default?

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -175,6 +177,40 @@ jobs:
...
+      rocm_package_version: ${{ inputs.rocm_package_version }}
+    permissions:
+      contents: read
+      id-token: write
+
+  # ==========================================================================
+  # STAGE: fusilli-libs (generic, after math-libs + iree-compiler)
+  # ==========================================================================
+  fusilli-libs:
+    needs: [compiler-runtime, math-libs, iree-compiler]
```
marbre, line 201: I assume the split is due to the rather long compile time for iree-compiler?

.github/workflows/multi_arch_build_portable_linux.yml
AaronStGeorge: It does not, 👍 will update when windows build lands.

iree-libs/CMakeLists.txt
AaronStGeorge: > The compiler shouldn't need a list of AMDGPU targets from USE_DIST_AMDGPU_TARGETS... right? What would that be used for?

I don't think IREE needs it at all, but without `USE_DIST_AMDGPU_TARGETS` or `DISABLE_AMDGPU_TARGETS` the configure step for `iree-compiler` stage will fail this check ([example](https://github.com/ROCm/TheRock/actions/runs/22640992547/job/65620896000)): https://github.com/ROCm/TheRock/blob/b1ebfea3f4304b90827b4e70cdc1eb42a7c69a0c/cmake/therock_subproject.cmake#L1390-L1416

I tried `DISABLE_AMDGPU_TARGETS` and can confirm the configure step (and all others) works ([link](https://github.com/ROCm/TheRock/actions/runs/22608134632/job/65506813697)), but the doc comment for `DISABLE_AMDGPU_TARGETS` makes it sound like something one should prefer not to use:
https://github.com/ROCm/TheRock/blob/b1ebfea3f4304b90827b4e70cdc1eb42a7c69a0c/cmake/therock_subproject.cmake#L212-L219

CMakeLists.txt
AaronStGeorge: Can we turn this on only for the multi-arch pipeline? I wasn't aware there was a way to do that?

build_tools/github_actions/fetch_test_configurations.py
AaronStGeorge: IREE/fusilli are not building on the main pipeline, and we were not planning on enabling it. The added compile time for IREE is just too much.

Can we selectively add `fusilliprovider` based on the pipeline here? If we can build fusilli only on the multi-arch pipeline then enabling this test only on the multi-arch pipeline would be great!

If memory serves this controls the test jobs that are enabled for both pipelines, we definitely don't want to enable `fusilliprovider` tests on the main pipeline as the artifacts won't be available and the test will fail.

.github/workflows/multi_arch_build_portable_linux.yml
AaronStGeorge: Yes, that's the idea.

iree-libs/CMakeLists.txt
ScottTodd: Ok, fine to use `USE_DIST_AMDGPU_TARGETS` given that other target-neutral targets (LLVM, rocm-core, rocm-cmake, etc.) use it too. I think it could be worth a comment in this case, given how IREE is designed as a compiler with general support for many GPUs/backends.

.github/workflows/multi_arch_build_portable_linux.yml
AaronStGeorge: Done.

build_tools/github_actions/fetch_test_configurations.py
AaronStGeorge: > 👍 I'll remove the temporary changes before the next round of review

Leaving now, as some of the in-progress conversation won't make sense without the changes.

iree-libs/CMakeLists.txt
AaronStGeorge: Done.

CMakeLists.txt
marbre: I think we eventually can similar to how we exclude running multi-arch for a specific gfx-architecure (see commit 6aa8bc45b027d7bea86cb7a3490ca59cc46150d9). @geomin12 any thoughts, guidance here as this would be somewhere in the test scripts / CI.

geomin12: i think adding that logic here would be good: https://github.com/ROCm/TheRock/blob/a45247d447fa7b54759a729663ed29b46a51e643/build_tools/github_actions/configure_ci.py#L455

although for TheRock, we build everything. it sounds like if it is multi-arch, we want to enable THEROCK_ENABLE_IREE_LIBS otherwise, no?

marbre: Correct, the ask here is to _only_ build in case of multi-arch but not else.

AaronStGeorge: It looks like we're going to be moving the plugin into `rocm-libraries` ([link to AMD internal teams chat](https://teams.microsoft.com/l/message/19:Og5wsI_YU1T9w0xPy4QxPibM14kOXMgFYfZFlMivNMs1@thread.tacv2/1772575191431?tenantId=3dd8961f-e488-4e60-8e11-a82d994e183d&groupId=69e7275a-4d49-495d-8bd5-04fb519c8e9c&parentMessageId=1772575191431&teamName=AIG%20ROCm&channelName=Project%20-%20hipdnn%20enablement&createdTime=1772575191431&ngc=true&allowXTenantAccess=true)). Given that, we definitely don't want to enable the build until that move is complete.

AaronStGeorge: I think it still makes sense to merge this even with the possible change in stage topology. What's here works today, and should be easy to adapt to the final layout once other pieces have moved around. The current changes give us a base to iterate on.

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -11,7 +8,9 @@
 # 5. debug-tools (generic) - amd-dbgapi, rocr-debug-agent, rocgdb (parallel to math-libs)
 # 6. dctools-core (generic) - RDC (parallel to math-libs)
 # 7. profiler-apps (generic) - rocprofiler-systems (parallel to math-libs)
-# 8. media-libs (generic) - sysdeps-amd-mesa, rocdecode, rocjpeg
+# 8. iree-compiler (generic) - IREE compiler (parallel to math-libs)
+# 9. fusilli-libs (generic) - Fusilli hipdnn provider (after math-libs + iree-compiler)
```
AaronStGeorge, line 12: Image

Adding comment thread from https://github.com/ROCm/TheRock/pull/3791

sjain-stanford: My comment isn't blocking - just wanted to get an idea of how things come together eventually. We can take incremental steps to get there.

BUILD_TOPOLOGY.toml
```diff
@@ -298,11 +302,19 @@ type = "generic"
...
 
-[artifact_groups.iree-libs]
-description = "IREE and Fusilli integration"
+[artifact_groups.iree-compiler]
+description = "IREE compiler"
 type = "generic"
 artifact_group_deps = ["hip-runtime"]
-source_sets = ["iree-libs"]
+# TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
+source_sets = ["iree-libs", "rocm-systems"]
```
marbre, line 310: Does IREE compiler really need `projects/hip/VERSION`? Furthermore, with the renaming `iree-libs` now only is a source set. Is there a better name for this?

BUILD_TOPOLOGY.toml
```diff
@@ -298,11 +302,19 @@ type = "generic"
...
+
+[artifact_groups.fusilli-libs]
+description = "Fusilli hipdnn provider"
+type = "generic"
+artifact_group_deps = ["hip-runtime", "iree-compiler"]
+# TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
+source_sets = ["iree-libs", "rocm-systems"]
```
marbre, line 317: As above, does this really need `rocm-systems`?

BUILD_TOPOLOGY.toml
AaronStGeorge: It doesn't, this is a workaround for a pre-existing issue.  Currently, without `"rocm-systems"` configure crashes during an unconditional read of `projects/hip/VERSION` in `TheRock` 's main `CMakeLists.txt` (before we would even get to IREE):
https://github.com/ROCm/TheRock/blob/c6bc4782f8c3fdfb960c681ce63daecb8f27b6ce/CMakeLists.txt#L179-L182

The same pattern is included in every other base artifact group at the moment:

https://github.com/ROCm/TheRock/blob/c6bc4782f8c3fdfb960c681ce63daecb8f27b6ce/BUILD_TOPOLOGY.toml#L260-L261

https://github.com/ROCm/TheRock/blob/c6bc4782f8c3fdfb960c681ce63daecb8f27b6ce/BUILD_TOPOLOGY.toml#L273-L274

https://github.com/ROCm/TheRock/blob/c6bc4782f8c3fdfb960c681ce63daecb8f27b6ce/BUILD_TOPOLOGY.toml#L286-L287

It seems to me that the underlying issue + all the other stages should definitely be fixed, but it's out of scope for this PR.

BUILD_TOPOLOGY.toml
AaronStGeorge: Same

BUILD_TOPOLOGY.toml
marbre: Ohh... missed that.

PR #3579, opened 2026-02-24
Media Libraries  - Python package

kiritigowda:
## Motivation

Python packaging for media libraries (rocdecode & rocjpeg)

## Technical Details

Libs are being included in rocm-sdk-core (via core_artifact_filter + _dist_info.py LibraryEntry)

## Test Plan

Verify if media packages are generated

## Test Result

No current tests should fail

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3614, opened 2026-02-25
Media - linux package

kiritigowda:
## Motivation

Split PR #3579 into linux and python -- This PR contains linux packaging implementation

## Technical Details

Media linux packages

## Test Plan

Current tests

## Test Result

No current tests should fail

## Submission Checklist

- [x ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/.github/workflows/build_linux_jax_wheels.yml
+++ b/.github/workflows/build_linux_jax_wheels.yml
@@ -59,7 +59,9 @@ on:
           - gfx1152
           - gfx1153
           - gfx120X-all
-          - gfx90X-dcgpu
+          - gfx906
+          - gfx908
+          - gfx90a
           - gfx94X-dcgpu
           - gfx950-dcgpu
         default: gfx94X-dcgpu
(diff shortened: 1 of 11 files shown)


build_tools/packaging/linux/package.json
```diff
@@ -1451,6 +1451,18 @@
           }
         ]
       },
+      {
+        "Artifact": "sysdeps-amd-mesa",
+        "Artifact_Subdir": [
+          {
+            "Name": "sysdeps-amd-mesa",
```
raramakr: Subdir name should be "amd-mesa". 
cat sysdeps-amd-mesa_lib_generic/artifact_manifest.txt 
third-party/sysdeps/linux/**amd-mesa/**build/stage

Subdir is used to identify the specific artifacts : For ex if we have 3 directories in artifact_manifest.txt 
third-party/sysdeps/linux/aaaaa/build/stage
third-party/sysdeps/linux/bbbbb/build/stage
third-party/sysdeps/linux/cccccbuild/stage
If we need only one of the artifacts , then select the unique identifier  (ex: aaaaa)
if wee need all , then we can  either choose "sysdeps" or "list all three in Artifact_Subdir

kiritigowda: @raramakr thanks for the correction. I made the changes, please review the latest

raramakr (approved):

PR #3617, opened 2026-02-25
Changing tagging bot to systems assistant

ammallya:
## Motivation

Current bot does not have access and permissions to tag rocm-systems repo, switching it out to use systems-assistant bot

## Technical Details

Changing APP-ID and KEY


jayhawk-commits (approved):
lgtm

PR #3632, opened 2026-02-26
Move rocm-kpack from base submodule to core (via rocm-systems)

stellaraccident:
rocm-kpack has been subtree-merged into the rocm-systems monorepo at shared/kpack/. This commit updates TheRock to consume it from there instead of as a standalone submodule.

Changes:
- Remove base/rocm-kpack submodule and .gitmodules entry
- Remove THEROCK_KPACK_DIR indirection; reference rocm-systems path directly in cmake/therock_artifacts.cmake
- Move rocm-kpack subproject declaration from base/CMakeLists.txt to core/CMakeLists.txt, gated by THEROCK_ENABLE_KPACK
- Add core-kpack artifact with its own artifact-core-kpack.toml
- Declare core-kpack in BUILD_TOPOLOGY.toml (hip-runtime group) and add it as a dependency of core-hip
- Unconditionally enable ROCM_KPACK_ENABLED in hip-clr (was conditional on THEROCK_KPACK_SPLIT_ARTIFACTS)
- Add core-kpack to python package build list

Requires a rocm-systems dep roll to pick up shared/kpack/.

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_hiptests.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hiptests.py
@@ -78,7 +78,12 @@ def get_asan_lib_path():
 def copy_dlls_exe_path():
     if platform.system() == "Windows":
         # hip and comgr dlls need to be copied to the same folder as exectuable
-        dlls_pattern = ["amdhip64*.dll", "amd_comgr*.dll", "hiprtc*.dll"]
+        dlls_pattern = [
+            "amdhip64*.dll",
+            "amd_comgr*.dll",
+            "hiprtc*.dll",
+            "rocm_kpack*.dll",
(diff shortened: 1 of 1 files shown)


stellaraccident:
Rebased and included a temporary rocm-systems bump to test. Can land once rocm-systems moves past this point on main.

stellaraccident:
I'm landing this with these failing test shards as it is impossible that this change is causing that.

stellaraccident:
This has had a lot of soak time, but it does introduce a new dependency to libamdhip64 which is not something we do every day. If it causes problems, a partial revert is possible by turning this flag off: https://github.com/ROCm/TheRock/blob/main/core/CMakeLists.txt#L216

PR #3640, opened 2026-02-26
[llvm-tools] enable build/install of llvm-dis llvm-as

ronlieb:
There are internal/external customers that compile with -save-temps and then convert the bc to ll using llvm-dis so they can edit/protype changes, then they convert the ll back to bc and feed it to compiler.


ronlieb:
@stellaraccident   ok to land ?

Issue #3109, opened 2026-01-27
[Issue]: Torch 2.9.1 from therock whls throws warnings to stdout on a machine without GPUs


gshtras:
### Problem Description

Per the discussion in https://teams.microsoft.com/l/message/19:HKqqVUgpG31aOnEzxVscKSKYEk_nbAh8cfxYFShj4GE1@thread.tacv2/19:HKqqVUgpG31aOnEzxVscKSKYEk_nbAh8cfxYFShj4GE1@thread.tacv2::1769103640520?parentMessageId=19:HKqqVUgpG31aOnEzxVscKSKYEk_nbAh8cfxYFShj4GE1@thread.tacv2::1769103640520

vLLM build process depends on the output of running scripts with torch import. Polluting the output with messages breaks the build process.
Repro instructions:

```
docker run -it --rm ubuntu:24.04 # Without GPUs
1  apt-get update -y     && apt-get install -y software-properties-common build-essential git curl sudo vim less libopenmpi-dev libpci-dev python3-venv python3-dev
2  python3 -m venv /opt/python --system-site-packages
3  export PATH=/opt/python/bin:$PATH
4  source /opt/python/bin/activate
5  export PIP_EXTRA_INDEX_URL=https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu/
6  pip install --index-url ${PIP_EXTRA_INDEX_URL} --pre "rocm[libraries,devel]>=7.11"
7  rocm-sdk init
8  pip install --index-url ${PIP_EXTRA_INDEX_URL} 'torch==2.9.1' torchaudio torchvision triton
9  python -c 'import torch' 2> /dev/null
```
Observe the warning message
```
[WARNING] offload-arch failed with return code 1
[stderr]
```
printed into the standard output

### Operating System

Ubuntu 24.04 docker

### CPU

Intel(R) Xeon(R) Platinum 8480C

### GPU

N/A

### ROCm Version

7.11

### ROCm Component

rocm-core

ScottTodd:
That logging is coming from here: https://github.com/ROCm/TheRock/blob/6539814ad31a93c29ee4231cd159b1b6cfff9b5d/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py#L138-L140

We discussed logging here: https://github.com/ROCm/TheRock/pull/1269#discussion_r2285943846
[...]

PR #1269, opened 2025-08-18
Added the implementation of discover_current_target_family

umarinkovic:
## Motivation

Solved a TODO in the _dist_info.py template

## Technical Details

Implemented dynamic discovery of gpu target using amdgpu-arch binary.

PR #3644, opened 2026-02-26
Fix stdout pollution from offload-arch warnings (#3109)

WBobby:
## Motivation

Suppress warning messages printed to stdout when offload-arch fails in GPU-less environments. This fixes build failures in vLLM and other tools that depend on clean stdout from `import torch`.

Fixes: https://github.com/ROCm/TheRock/issues/3109

## Technical Details

- Capture subprocess stdout/stderr to prevent offload-arch output leaking
- Gate warning messages behind ROCM_SDK_VERBOSE environment variable
- Output warnings to stderr instead of stdout when verbose mode enabled
- Use os.environ.copy() to avoid modifying global environment

Set ROCM_SDK_VERBOSE=1 to enable diagnostic messages.

## Test Plan

1. Build ROCm in personal branch with the fix
2. Test in local machine with ROCm build in step 1.
3. Run "python -c 'import torch' 2> /dev/null" to check output.

## Test Result

After fix:
root@b8e75cfcdb2b:/app# python -c 'import torch'  2>/dev/null
(no output)

## Submission Checklist

### DIFF
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
@@ -144,7 +144,11 @@ def discover_current_target_family() -> str | None:
                     return arch
     except subprocess.CalledProcessError as e:
         if _VERBOSE:
-            print(f"[rocm_sdk] offload-arch failed: {e.returncode}", file=sys.stderr)
+            print(
+                f"[rocm_sdk] offload-arch failed with return code {e.returncode}",
+                file=sys.stderr,
+            )
+            print(f"[rocm_sdk] output: {e.output}", file=sys.stderr)
     except FileNotFoundError:
         if _VERBOSE:
             print("[rocm_sdk] offload-arch not found", file=sys.stderr)


WBobby:
@ScottTodd Requesting your review - this implements the environment variable solution you suggested in PR #1269. Fix
  verified and tested. Thanks!

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -136,12 +140,14 @@ def discover_current_target_family() -> str | None:
                 if arch in AVAILABLE_TARGET_FAMILIES:
                     return arch
     except subprocess.CalledProcessError as e:
-        print(f"[WARNING] offload-arch failed with return code {e.returncode}")
-        print(f"[stderr] {e.output}")
+        if _VERBOSE:
+            print(f"[rocm_sdk] offload-arch failed: {e.returncode}", file=sys.stderr)
```
ScottTodd: Should we keep `e.output` here (or just `e` and let string conversion handle it... it might)? Just the return code may not be useful enough.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -7,8 +7,10 @@
 import importlib.util
 import os
 import subprocess
+import sys
 from pathlib import Path
 
+_VERBOSE = os.getenv("ROCM_SDK_VERBOSE", "0") == "1"
```
ScottTodd, line 16: Could add a comment here

```suggestion
# Some downstream users of these APIs do not expect or handle
# stdout/stderr output from this code, so logging is disabled by default.
# See https://github.com/ROCm/TheRock/issues/3109.
_VERBOSE = os.getenv("ROCM_SDK_VERBOSE", "0") == "1"
```

PR #3647, opened 2026-02-26
[rocprofiler-systems] Add `test` artifact

kcossett-amd:
## Motivation

I noticed in #3602 that our GPU tests would be skipped.
For now, adding a test artifact may help as previously, both `lib` and `run` would include the example binaries and that may have caused problems.

AIPROFSYST-222

## Technical Details

Add `test` artifact to `rocprofiler-systems` and exclude `share/rocprofiler-systems/examples` and `share/rocprofiler-systems/tests` from being part of both the `run` and `lib` artifacts. 

## Submission Checklist

### DIFF
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -278,6 +278,7 @@ endif(THEROCK_BUILD_TESTING)
         doc
         lib
         run
+        test
       SUBPROJECT_DEPS
         amdsmi
         rocprofiler-sdk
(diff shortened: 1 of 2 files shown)


kcossett-amd:
I do realize that we should be building tests based on `THEROCK_BUILD_TESTING`. However, to do so, I need to change some code involving `ROCPROFSYS_BUILD_FOR_THEROCK` on our side. I'll fix it in a future PR once those changes are in

kcossett-amd:
Closing. This won't fix our problem. We need just the `examples` to be target-specific, which currently can't be done without changes on our side.

Issue #3634, opened 2026-02-26
[Issue] Nightly Linux package workflow fails to trigger JAX run due to invalid `amdgpu_family` input value


chiranjeevipattigidi:
**Effected Archs:** `gfx906, gfx908, gfx90a`
### Jobs:
- https://github.com/ROCm/TheRock/actions/runs/22428091619/job/64940890782
- https://github.com/ROCm/TheRock/actions/runs/22428091619/job/64940890785
- https://github.com/ROCm/TheRock/actions/runs/22428091619/job/64940890756

```
🏃 Workflow Dispatch Action v1.2.4
🔎 Found workflow, id: 233271992, name: Release portable Linux JAX Wheels, path: .github/workflows/release_portable_linux_jax_wheels.yml
🚀 Calling GitHub API to dispatch workflow...
Error: Provided value 'gfx906' for input 'amdgpu_family' not in the list of allowed values - https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event

```

amd-chiranjeevi:
`release_portable_linux_jax_wheels.yml` was not updated in commit https://github.com/ROCm/TheRock/commit/8f17ad65be8002c5ba3d58097f9d44bd5d9131c0, which is causing the current issue

opened a PR:  https://github.com/ROCm/TheRock/pull/3633

ScottTodd:
Let's add test coverage for this to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/workflow_dispatch_inputs_test.py

We also may want to move away from using `type: choice`, given how error-prone it is and how much the list of targets is duplicated across workflows right now.

ScottTodd:
This was fixed by https://github.com/ROCm/TheRock/pull/3633. I'm sending a PR with unit tests that would have caught this earlier.

PR #2869, opened 2026-01-12
Split gfx906/gfx908/gfx90a into separate families

stellaraccident:
Each gfx9 generation has different instruction support (e.g., fp8 variants, WMMA capabilities), so CK and MIOpen need to build/test them individually rather than as a combined gfx90X family.

This change:
- Removes the shared gfx90X-all family grouping
- Gives each target its own family (gfx906, gfx908, gfx90a)
- Updates nightly CI matrix to build each separately
- Updates workflow_dispatch options in wheel build workflows

This is consistent with how gfx942 and gfx950 are already handled as separate families.

PR #3651, opened 2026-02-26
Add unit tests for amdgpu_family workflow dispatch choices

ScottTodd:
## Motivation

https://github.com/ROCm/TheRock/pull/2869 split a GPU family but did not update all blocks of:
```yml
  workflow_dispatch:
    inputs:
      amdgpu_family:
        type: choice
```
causing https://github.com/ROCm/TheRock/issues/3634 , where one of the workflows that was not updated failed to be dispatched:
```
🏃 Workflow Dispatch Action v1.2.4
🔎 Found workflow, id: 233271992, name: Release portable Linux JAX Wheels, path: .github/workflows/release_portable_linux_jax_wheels.yml
🚀 Calling GitHub API to dispatch workflow...
Error: Provided value 'gfx906' for input 'amdgpu_family' not in the list of allowed values - https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event
```

This adds unit test coverage that would have caught that issue.

## Technical Details

Two new tests are included:

1. A generic test in `build_tools/github_actions/tests/workflow_dispatch_inputs_test.py` for cases where a literal string is passed to a `choice` input type
2. A module-specific test in `build_tools/github_actions/tests/fetch_package_targets_test.py` for cases where the _dynamic_ result of `fetch_package_targets.determine_package_targets()` produces a value not covered exhaustively by a `amdgpu_family` option.
    * If a workflow uses a different name the test will NOT catch that, but it will at least assert if _all_ options are renamed without also updating the test

We could switch these workflows to use `type: string` instead of `type: choice` to avoid this particular class of issue. These tests are useful until then (and for other future workflow changes).

## Test Plan

* Run the unit tests (local and CI), check that they pass/fail as expected

## Test Result

1. Invalid static value (should fail):

    ```diff
    (.venv) λ git diff
    diff --git a/.github/workflows/release_portable_linux_packages.yml b/.github/workflows/release_portable_linux_packages.yml
    index fc384166..58a7ae2a 100644
    --- a/.github/workflows/release_portable_linux_packages.yml
    +++ b/.github/workflows/release_portable_linux_packages.yml
    @@ -331,7 +331,7 @@ jobs:
               workflow: release_portable_linux_pytorch_wheels.yml
               inputs: |
    -            { "amdgpu_family": "${{ matrix.target_bundle.amdgpu_family }}",
    +            { "amdgpu_family": "gfxSomeUnknownValue",
                   "release_type": "${{ env.RELEASE_TYPE }}",
    ```
    
    ```python
    (.venv) λ pytest build_tools/github_actions/tests/workflow_dispatch_inputs_test.py
    
    build_tools\github_actions\tests\workflow_dispatch_inputs_test.py F.....                                         WorkflowDispatchInputsTest.test_literal_values_valid_for_choices__release_portable_linux_packages
    
            if errors:
    >           self.fail("\n".join(errors))
    E           AssertionError: step 'Trigger building PyTorch wheels' passes literal 'gfxSomeUnknownValue' for 'amdgpu_family' to 'release_portable_linux_pytorch_wheels.yml', but allowed options are: ['gfx101X-dgpu', 'gfx103X-dgpu', 'gfx110X-all', 'gfx1150', 'gfx1151', 'gfx1152', 'gfx1153', 'gfx120X-all', 'gfx906', 'gfx908', 'gfx90a', 'gfx94X-dcgpu', 'gfx950-dcgpu']
    ```
2. Missing dynamic value (should fail):

    ```diff
    (.venv) λ git diff
    diff --git a/.github/workflows/release_portable_linux_jax_wheels.yml b/.github/workflows/release_portable_linux_jax_wheels.yml
    index eae45d80..544006b2 100644
    --- a/.github/workflows/release_portable_linux_jax_wheels.yml
    +++ b/.github/workflows/release_portable_linux_jax_wheels.yml
    @@ -48,9 +48,7 @@ on:
               - gfx1152
               - gfx1153
               - gfx120X-all
    -          - gfx906
    -          - gfx908
    -          - gfx90a
    +          - gfx90X-dcgpu
               - gfx94X-dcgpu
               - gfx950-dcgpu
             default: gfx94X-dcgpu
    ```
    
    ```python
    (.venv) λ pytest build_tools/github_actions/tests/fetch_package_targets_test.py
    FetchPackageTargetsTest.test_amdgpu_families_accepted_by_choice_workflows ______________________
    
            if errors:
    >           self.fail("\n".join(errors))
    E           AssertionError: release_portable_linux_jax_wheels.yml is missing amdgpu_family options that fetch_package_targets can produce: ['gfx906', 'gfx908', 'gfx90a']
    ```

3. Switch to `type: string` (should pass):

    ```diff
    (.venv) λ git diff
    diff --git a/.github/workflows/release_portable_linux_jax_wheels.yml b/.github/workflows/release_portable_linux_jax_wheels.yml
    index eae45d80..1a98aeec 100644
    --- a/.github/workflows/release_portable_linux_jax_wheels.yml
    +++ b/.github/workflows/release_portable_linux_jax_wheels.yml
    @@ -38,21 +38,7 @@ on:
       workflow_dispatch:
         inputs:
           amdgpu_family:
    -        type: choice
    -        options:
    -          - gfx101X-dgpu
    -          - gfx103X-dgpu
    -          - gfx110X-all
    -          - gfx1150
    -          - gfx1151
    -          - gfx1152
    -          - gfx1153
    -          - gfx120X-all
    -          - gfx906
    -          - gfx908
    -          - gfx90a
    -          - gfx94X-dcgpu
    -          - gfx950-dcgpu
    +        type: string
             default: gfx94X-dcgpu
           release_type:
             description: The type of release to build ("dev", "nightly", or "prerelease"). All developer-triggered jobs should use "dev"!
    ```
    
    ```python
    (.venv) λ pytest build_tools/github_actions/tests/fetch_package_targets_test.py
    build_tools\github_actions\tests\fetch_package_targets_test.py ......                                            [100%]
    ================================================== 6 passed in 0.21s ==================================================
    ```

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/tests/fetch_package_targets_test.py
+++ b/build_tools/github_actions/tests/fetch_package_targets_test.py
@@ -11,12 +11,20 @@ from workflow_utils import WORKFLOWS_DIR, get_choice_options, load_workflow
 
 class FetchPackageTargetsTest(unittest.TestCase):
     def test_amdgpu_families_accepted_by_choice_workflows(self):
-        """Every amdgpu_family from the matrix is in every workflow's choice list.
+        """Checks workflow choice lists against fetch_package_targets outputs.
 
         fetch_package_targets.py produces amdgpu_family values that get
-        dispatched to workflows with type: choice inputs. If a produced value
-        isn't in the target's choice list, GitHub rejects the dispatch.
-        See https://github.com/ROCm/TheRock/issues/3634.
+        dispatched to workflows. Those workflow inputs can either use
+        "type: choice" or "type: string". If they use "type: choice", this tests
+        that the choices are valid:
+
+        1. Check that every amdgpu_family from the matrix is in every workflow's
+          choice list. If a produced value isn't in the target's choice list,
+          GitHub rejects the dispatch.
+          See https://github.com/ROCm/TheRock/issues/3634.
+        2. Check that the workflow choice list does not contain any extra
+          families. While not an error, extra families may indicate an
+          incomplete migration.
         """
         # Collect all amdgpu_family values the matrix can produce.
         produced_families = set()
@@ -45,12 +53,19 @@ class FetchPackageTargetsTest(unittest.TestCase):
         )
 
         errors = []
-        for name, options in choice_workflows.items():
-            missing = produced_families - set(options)
-            if missing:
+        for workflow_name, options in choice_workflows.items():
+            # Test for missing families
+            missing_families = produced_families - set(options)
+            if missing_families:
                 errors.append(
-                    f"{name} is missing amdgpu_family options "
-                    f"that fetch_package_targets can produce: {sorted(missing)}"
+                    f"{workflow_name} is missing amdgpu_family options "
+                    f"that fetch_package_targets can produce: {sorted(missing_families)}"
+                )
+            # Test for extra families
+            extra_families = set(options) - produced_families
+            if extra_families:
+                errors.append(
+                    f"{workflow_name} has extra amdgpu_family options not listed in fetch_package_targets: {sorted(extra_families)} "
                 )
 
         if errors:


build_tools/github_actions/tests/fetch_package_targets_test.py
ScottTodd, line 1: RE: the "skip-ci" label I added... I think we could also exclude the `build_tools/*/tests/` directories from running CI with some changes to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_ci_path_filters.py. I don't _think_ any part of the CI workflow now depends on the unit tests...

build_tools/github_actions/tests/workflow_utils.py
```diff
@@ -0,0 +1,110 @@
...
+            inputs:
+              amdgpu_family:
+                required: true
+              release_type:
+                required: true
+                default: dev
+
+    Returns: {"amdgpu_family"}  (release_type has a default)
+    """
+    dispatch = _get_workflow_dispatch_block(workflow)
```
HereThereBeDragons: using `get_workflow_dispatch_inputs` could save a couple of lines?

build_tools/github_actions/tests/workflow_utils.py
```diff
@@ -0,0 +1,110 @@
...
+                options:
+                  - gfx94X-dcgpu
+                  - gfx110X-all
+
+    get_choice_options(workflow, "amdgpu_family") returns:
+        ["gfx94X-dcgpu", "gfx110X-all"]
+
+    Returns None if the input doesn't exist or isn't type: choice.
+    """
+    dispatch = _get_workflow_dispatch_block(workflow)
```
HereThereBeDragons: same here: using `get_workflow_dispatch_inputs` could save a couple of lines?

build_tools/github_actions/tests/fetch_package_targets_test.py
```diff
@@ -6,8 +6,56 @@
...
+        self.assertGreater(
+            len(choice_workflows),
+            0,
+            "No workflows found with amdgpu_family as type: choice - "
+            "was the input renamed?",
+        )
+
+        errors = []
+        for name, options in choice_workflows.items():
+            missing = produced_families - set(options)
```
HereThereBeDragons: lets also do the reverse: 
```
set(options) - produced_families
```
so we can catch when there are choices that we dont have any more in amdgpu family matrix.

build_tools/github_actions/tests/workflow_dispatch_inputs_test.py
HereThereBeDragons, line 208: took me quite a bit of time to understand all this logic here. 
can we somehow make this more obvious that this is the main loop logic?

build_tools/github_actions/tests/fetch_package_targets_test.py
ScottTodd: Oh good call. Sure, adding that too. It'll look like this:
```diff
git diff
diff --git a/.github/workflows/release_portable_linux_jax_wheels.yml b/.github/workflows/release_portable_linux_jax_wheels.yml
index eae45d80..c25ed402 100644
--- a/.github/workflows/release_portable_linux_jax_wheels.yml
+++ b/.github/workflows/release_portable_linux_jax_wheels.yml
@@ -51,6 +51,7 @@ on:
           - gfx906
           - gfx908
           - gfx90a
+          - gfx99999
           - gfx94X-dcgpu
           - gfx950-dcgpu
         default: gfx94X-dcgpu
```

```python
        if errors:
>           self.fail("\n".join(errors))
E           AssertionError: release_portable_linux_jax_wheels.yml has extra amdgpu_family options not listed in fetch_package_targets: ['gfx99999']

build_tools\github_actions\tests\fetch_package_targets_test.py:62: AssertionError
```

build_tools/github_actions/tests/workflow_dispatch_inputs_test.py
ScottTodd: I'm not sure how much restructuring or documentation could help here. There are a few different patterns for dynamically generating test cases with unittest/pytest, this is just following one of them.

build_tools/github_actions/tests/workflow_utils.py
ScottTodd: Extracted a helper function. Couldn't just use the existing function directly due to it losing some information (returning only a `set` of names, without the other fields)

PR #3658, opened 2026-02-26
create runtime docker images for ocltst

agunashe:
## Motivation
Creating new docker files for runtime ocltst execution. 
ocltst requires ICD Opencl.

## Technical Details
Creating new docker files for runtime ocltst execution. 
ocltst requires ICD Opencl.

## Test Plan
NA

## Test Result
NA

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -40,7 +40,10 @@ test_matrix = {
         "timeout_minutes": 120,
         "test_script": f"python {_get_script_path('test_hiptests.py')}",
         "platform": ["linux", "windows"],
-        "total_shards": 4,
+        "total_shards_dict": {
+            "linux": 4,
+            "windows": 4,
+        },
     },
     # BLAS tests
     "rocblas": {
@@ -49,7 +52,10 @@ test_matrix = {
         "timeout_minutes": 15,
         "test_script": f"python {_get_script_path('test_rocblas.py')}",
         "platform": ["linux", "windows"],
-        "total_shards": 1,
+        "total_shards_dict": {
+            "linux": 1,
+            "windows": 1,
+        },
     },
     "rocroller": {
         "job_name": "rocroller",
@@ -57,7 +63,10 @@ test_matrix = {
         "timeout_minutes": 60,
         "test_script": f"python {_get_script_path('test_rocroller.py')}",
         "platform": ["linux"],
-        "total_shards": 5,
+        "total_shards_dict": {
+            "linux": 5,
+            "windows": 5,
+        },
         "exclude_family": {
             # rocroller does not plan to support Linux and Windows gfx115X architectures
             "linux": [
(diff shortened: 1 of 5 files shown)


dockerfiles/README.md
marbre, line 1: Looking at prior document structure, it seems could just append:

----

| Source .Dockerfile                                                                           | Published package                                                               |
| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| [`no_rocm_image_ubuntu24_04_rocgdb.Dockerfile`](no_rocm_image_ubuntu24_04_rocgdb.Dockerfile) | https://github.com/ROCm/TheRock/pkgs/container/no_rocm_image_ubuntu24_04_rocgdb |

Extended version of no_rocm_image_ubuntu24_04.Dockerfile, containing additional packages and tools required for validation of rocgdb.
This includes dejagnu, make, gcc, g++ and gfortran.

| Source .Dockerfile                                                                           | Published package                                                               |
| -------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| [`no_rocm_image_ubuntu24_04_runtime.Dockerfile`](no_rocm_image_ubuntu24_04_runtime.Dockerfile) | https://github.com/ROCm/TheRock/pkgs/container/no_rocm_image_ubuntu24_04_runtime 

Used in ocltst execution. It installs OCL ICD package required by ocltst.

----

Eventually we can also come up with a better name than `runtime`? Eventually we could append `_opencl_runtime` or `_ocl`?

agunashe: yeah i wanted to just append to the table, instead of creating a new one.  
Did not want to restrict the docker file to ocl, since we have Hip as well. We could potentially reuse the docker incase we need it for hip.  So thats why name it as just runtime.

dockerfiles/no_rocm_image_ubuntu24_04_runtime.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest
+
+# packages for runtime hip/ocl validation
+RUN sudo apt-get install -y --no-install-recommends \
+    ocl-icd-libopencl1 \
+    ocl-icd-opencl-dev
```
marbre: Missing newline.

dockerfiles/README.md
marbre: Adding a new table is fine, we already have one at the top of the section https://github.com/ROCm/TheRock/blob/968b7f82e7955fba6e04ab774e8efed3394e5315/dockerfiles/README.md#build_manylinux_dockerfile. The list above the table doesn't renders where will and isn't a list:

Image

 So one table per docker file is what we had before and it looks cleaner to just follow this approach. 

Regarding naming, thee docker file has no rocm preinstalled and thus no hip preinstalled. That is the whole idea behind the docker file. So what you do is you install ocl in a _no rocm_ dockerfile (see the name). ROCm and thus HIP get downloaded via the test job but are never part of the docker image. If it goes beyond ocl we can rename in the future.

agunashe: renamed the docker file to _ocl_rt and appended to the table.

dockerfiles/no_rocm_image_ubuntu24_04_runtime.Dockerfile
agunashe: Added a new line at the end of the file

PR #3659, opened 2026-02-26
Python Package - devel sdk - Updates for C++ consumption

kiritigowda:
## Motivation

Wheels don't support post-install hooks, so we provide an explicit command to extract the devel tarball into site-packages. Users must run `rocm-sdk-devel-init` once after install to make headers, CMake files, and dev libraries available for C++ compilation.

## Technical Details

Add rocm-sdk-devel-init function

## Test Plan

Init should untar the package once and existing flow should work as is

## Test Result

All current tests should pass

## Submission Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,33 +161,37 @@ Individual features can be controlled separately (typically in combination with
 `-DTHEROCK_ENABLE_ALL=OFF` or `-DTHEROCK_RESET_FEATURES=ON` to force a
 minimal build):
 
-| Component flag                         | Description                                    |
-| -------------------------------------- | ---------------------------------------------- |
-| `-DTHEROCK_ENABLE_AMD_DBGAPI=ON`       | Enables the ROCm debug API library             |
-| `-DTHEROCK_ENABLE_COMPILER=ON`         | Enables the GPU+host compiler toolchain        |
-| `-DTHEROCK_ENABLE_HIPIFY=ON`           | Enables the hipify tool                        |
-| `-DTHEROCK_ENABLE_CORE_RUNTIME=ON`     | Enables the core runtime components and tools  |
-| `-DTHEROCK_ENABLE_HIP_RUNTIME=ON`      | Enables the HIP runtime components             |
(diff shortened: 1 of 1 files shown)


build_tools/packaging/python/templates/rocm-sdk-devel/setup.py
```diff
@@ -40,14 +40,26 @@ def import_dist_info():
...
+    # Wheels don't support post-install hooks, so we provide an explicit
+    # command to extract the devel tarball into site-packages. Users must
+    # run `rocm-sdk-devel-init` once after install to make headers, CMake
+    # files, and dev libraries available for C++ compilation.
     entry_points={
-        "console_scripts": [],
+        "console_scripts": [
+            "rocm-sdk-devel-init=rocm_sdk_devel._cli:init",
+        ],
     },
```
ScottTodd, line 64: We already have `rocm-sdk init` for this: https://github.com/ROCm/TheRock/blob/c655fc589049bd98a32f550e45fc172786d39cc0/build_tools/packaging/python/templates/rocm/src/rocm_sdk/__main__.py#L143-L147

That is documented here: https://github.com/ROCm/TheRock/blob/main/RELEASES.md#using-rocm-python-packages

No need to add something unique to the `rocm-sdk-devel` package.

ScottTodd (changes requested):

build_tools/packaging/python/templates/rocm-sdk-devel/setup.py
```diff
@@ -40,14 +40,26 @@ def import_dist_info():
...
     zip_safe=False,
     include_package_data=True,
+    # include_package_data=True only picks up version-controlled files.
+    # The devel tarball is generated at build time, so it must be
+    # explicitly listed here to be included in the wheel.
+    package_data={
+        "rocm_sdk_devel": ["*.tar", "*.tar.xz"],
```
kiritigowda, line 47: @ScottTodd do you want this change? if not this PR can be closed

ScottTodd: No, that doesn't look right either. Tarballs are a separate concept not connected to the python packages.

PR #3661, opened 2026-02-26
[ci] Adding additional flag details in `test_environment_reproduction.md`

geomin12:
Adding additional flag details in `test_environment_reproduction.md` as a developer did not get this information from the docs

### DIFF
--- a/docs/development/test_environment_reproduction.md
+++ b/docs/development/test_environment_reproduction.md
@@ -28,7 +28,7 @@ $ python build_tools/github_actions/test_executable_scripts/test_rocblas.py
 - CI_RUN_ID is sourced from the CI run (ex: https://github.com/ROCm/TheRock/actions/runs/16948046392 -> CI_RUN_ID = 16948046392)
 - GPU_FAMILY is the LLVM target name (ex: gfx94X-dcgpu, gfx1151, gfx110X-all)
 - GITHUB_REPO is the GitHub repository that this CI run was executed. (ex: ROCm/rocm-libraries, ROCm/rccl)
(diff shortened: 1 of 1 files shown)


docs/development/test_environment_reproduction.md
```diff
@@ -28,5 +28,7 @@ $ python build_tools/github_actions/test_executable_scripts/test_rocblas.py
 - CI_RUN_ID is sourced from the CI run (ex: https://github.com/ROCm/TheRock/actions/runs/16948046392 -> CI_RUN_ID = 16948046392)
 - GPU_FAMILY is the LLVM target name (ex: gfx94X-dcgpu, gfx1151, gfx110X-all)
 - GITHUB_REPO is the GitHub repository that this CI run was executed. (ex: ROCm/rocm-libraries, ROCm/rccl)
+- ADDITIONAL FLAGS (optional) are found in [installing_artifacts.md](https://github.com/ROCm/TheRock/blob/085571a87f8050f94c77eb90454995b6d1b75658/docs/development/installing_artifacts.md#component-selection)
```
jayhawk-commits: Why the link to this specific blob SHA, instead of the version of the doc on main?

geomin12: ah that is in case if the file ever got deleted / moved, it would be a 404. with the SHA, it'll always exist

although, i guess in our case we won't delete it and most likely won't get moved, so we can add main branch in that case

geomin12: i updated it to main, to be most up to date

jharryma (approved):

Issue #3340, opened 2026-02-09
[Multi-arch] Remove single stage (non-multi-arch) CI


marbre:
> [!IMPORTANT]
> **2026-03-25**
> Any features that are limited to "CI" instead of "multi-arch CI" will soon longer be tested on pull requests and shortly afterwards will be deleted.

lumachad:
I'm wondering how this will affect rocm-systems' ability to invoke TheRock CI and if we have a transition plan to move it to multi-arch CI.

ScottTodd:
> I'm wonder how this will affect rocm-systems' ability to invoke TheRock CI and if we have a transition plan to move it to multi-arch CI.

That's https://github.com/ROCm/TheRock/issues/3343. Yes there is a plan (which still needs to be written down).

lumachad:
Thanks. I've posted comments there.

ScottTodd:
I'm starting the process of deleting old workflows and supporting code:
* https://github.com/ROCm/TheRock/pull/5790
* https://github.com/ROCm/TheRock/pull/5794

SamuelReeder:
#6823 will update `install_rocm_from_artifacts.py` to rely on multi-arch artifacts.

PR #3663, opened 2026-02-27
Optional S3 artifacts features 

jayhawk-commits:
## Summary

Adds optional upload behavior to `post_build_upload.py` to support private (GitHub EMU) and other non-standard workflows. All new flags are optional; existing default behavior is unchanged.

These features integrate with the `WorkflowOutputRoot` and `StorageLocation` abstractions introduced in PR #3596 rather than hardcoding bucket-specific paths.

### New CLI flags in `post_build_upload.py`

- **`--artifacts-bucket`**: Override the default artifacts bucket (e.g. for private workflows using a different bucket). When set, bypasses the automatic bucket selection in `WorkflowOutputRoot.from_workflow_run()`.
- **`--skip-manifest`**: Skip `therock_manifest.json` upload and omit the manifest link from the job summary. Useful for builds that do not produce a manifest.
- **`--path-prefix`**: Storage path prefix prepended to all output paths for this run (e.g. `v3/artifacts`). When set, all outputs are placed under `{prefix}/{run_id}-{platform}/` instead of `{run_id}-{platform}/`.
- **`--summary-base-url`** / **`--summary-path`**: Use a custom base URL (e.g. CloudFront CDN) for job summary links instead of the default public HTTPS URL. `--summary-path` sets an optional path prefix on the CDN URL. Avoids exposing raw bucket URLs in job summaries.

### `StorageLocation.cdn_url()` (new method in `storage_location.py`)

Backend-agnostic URL construction for CDN or custom base URLs. Strips a storage path prefix (to map from storage-relative paths to CDN-relative paths) and prepends an optional CDN path prefix.

### `WorkflowOutputRoot` updates (`workflow_outputs.py`)

- New `path_prefix` field: when set, all output paths for the run include the prefix.
- `from_workflow_run()` gains a `bucket_override` parameter to support `--artifacts-bucket`.

## Tests

- `build_tools/tests/workflow_outputs_test.py`: 3 new test classes covering `cdn_url()`, `path_prefix` field behavior, and `bucket_override` / `path_prefix` in `from_workflow_run()`.
- `build_tools/github_actions/tests/post_build_upload_test.py`: 2 new test classes covering `--skip-manifest` behavior and CDN URL substitution in `write_gha_build_summary()`.


jayhawk-commits:
I'll add pytest unit tests for these new features tomorrow. Wanted to get the work saved on the server that is ready for review.

marbre (commented):
This might interfere with @ScottTodd‘s current work in #3596.

jayhawk-commits:
> This might interfere with @ScottTodd‘s current work in #3596.

Thanks for the heads up. I can wait for that to land first. Either way, the workflows on EMU will need changes as the patches it presently have would no longer apply cleanly.

build_tools/github_actions/post_build_upload.py
```diff
@@ -7,6 +7,20 @@
...
+  Must not end with a forward slash. The script joins segments with "/".
+  --s3-subdir-artifacts is the path prefix in the S3 bucket; --s3-summary-path is
+  the corresponding path as it appears in the CloudFront (summary) URL. This
+  optional path aligns with the prerelease CloudFront layout used for tarballs
+  and wheels.
+
+  Example (CloudFront base URL for job summary links):
+    --s3-subdir-artifacts v3/artifacts
+    --s3-summary-base-url https://rocm.foobar.amd.com
+    --s3-summary-path artifacts
```
ScottTodd: I don't think we should be mentioning CloudFront at all in most of our code. It's a CDN, an implementation detail. What matters is the URL, which could be backed by any CDN or other webserver and this code would not change.

build_tools/github_actions/github_actions_utils.py
```diff
@@ -538,6 +538,14 @@ def retrieve_bucket_info(
     If neither is provided, |is_pr_from_fork| is populated from the
     IS_PR_FROM_FORK environment variable.
 
+    Environment (optional override):
+      S3_BUCKET_ARTIFACTS: When set, use this bucket for the artifacts bucket
+        instead of the default chosen from repository/fork/release_type.
```
ScottTodd: An override seems useful, but we need to be careful here. Environment variables are global, affecting all processes. Some of our workflows need the ability to read artifacts from one S3 bucket and write to another S3 bucket. We can do things like only set the environment variable for certain workflow steps, but at that point I think directly passing some setting through to the top level scripts and then into this function (or where I'm moving it to in https://github.com/ROCm/TheRock/pull/3596) would be safer.

build_tools/github_actions/post_build_upload.py
```diff
@@ -363,6 +391,30 @@ def run(args):
...
+        type=str,
+        default="",
+        help="Base URL for job summary links (e.g. CloudFront). Must not end with /. If set, summary uses this instead of raw S3 URL.",
+    )
+    parser.add_argument(
+        "--s3-summary-path",
+        type=str,
+        default="",
+        help="Path prefix for summary URL when --s3-summary-base-url is set (e.g. artifacts). Must not end with /.",
+    )
```
ScottTodd, line 363: Either strip trailing `/` characters or assert with an argument parsing error. Comments are not error handling.

ScottTodd (commented):
Yeah I think we can more cleanly support this once https://github.com/ROCm/TheRock/pull/3596 goes in.

amd-shiraz:
@jayhawk-commits  started looking at this. some very initial comments off the bat:
1) --skip-manifest should also suppress the job summary manifest link
2) in the log message, consider including external repo/fork run info if applicable 
Overall, Concept/design is good and useful for private workflows. on the backend side, not sure  what bucket we are using for this usecase but make sure the s3 bucket is private with versioning and encryption on, cloudfront is the only reader of the bucket, internal users access via vpn, external users via signed urls and so on.

jayhawk-commits:
Updated to work with #3596 and apply some of the feedback provided. I wanted to get that part out of the way, but I will continue after I am back from travel.

jayhawk-commits:
Adding skip-ci labels temporarily to not consume CI sources until I return from travel

build_tools/github_actions/post_build_upload.py
jayhawk-commits: Removed mentions of CloudFront to support any CDN.

build_tools/github_actions/post_build_upload.py
jayhawk-commits: Trailing `/` characters are now `rstrip`'d.

jayhawk-commits:
> @jayhawk-commits started looking at this. some very initial comments off the bat:
> 
>     1. --skip-manifest should also suppress the job summary manifest link
> 
>     2. in the log message, consider including external repo/fork run info if applicable
>        Overall, Concept/design is good and useful for private workflows. on the backend side, not sure  what bucket we are using for this usecase but make sure the s3 bucket is private with versioning and encryption on, cloudfront is the only reader of the bucket, internal users access via vpn, external users via signed urls and so on.

1. `write_gha_build_summary()` now takes a `skip_manifest` parameter and skips both the upload and the manifest link in the job summary when set.
2. The existing `_retrieve_bucket_info()` already logs `external_repo` and `is_pr_from_fork` as part of its output, which appears before our `bucket_override` log line. So the fork/repo context is always visible in the log when the override is active

Security concerns are accounted for when using the bucket overrides. One of the use cases involve buckets that we only want accessible through restricted CloudFront URLs.

build_tools/github_actions/github_actions_utils.py
jayhawk-commits: Replaced the usage of an env with an explicit `--artifacts-bucket` arg that gets passed directly into `WorkflowOutputRoot.from_workflow_run()` as a `bucket_override` parameter.

build_tools/github_actions/post_build_upload.py
ScottTodd, line 1: This script is only used by "CI", not "Multi-arch CI", and thus is scheduled for deletion (https://github.com/ROCm/TheRock/issues/3340). New feature work should be in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/post_stage_upload.py

build_tools/_therock_utils/storage_location.py
```diff
@@ -59,3 +59,38 @@ def local_path(self, staging_dir: Path) -> Path:
...
+    def cdn_url(
+        self,
+        base_url: str,
+        strip_prefix: str = "",
+        cdn_prefix: str = "",
+    ) -> str:
+        """CDN URL, optionally remapping the storage path prefix to a CDN prefix.
+
+        Strips ``strip_prefix`` from the start of ``relative_path`` (if present),
+        then prepends ``cdn_prefix``, and finally prepends ``base_url``.
```
ScottTodd, line 72: These changes overlap with more recent proposals:
* https://github.com/ROCm/TheRock/pull/4088
* https://github.com/ROCm/TheRock/pull/4167

ScottTodd (commented):
This PR has been open for a while I'm not sure about the approach here given how workflows have been changing as a result of switching to multi-arch CI. This also overlaps with other ongoing work.

build_tools/github_actions/post_build_upload.py
jayhawk-commits: Unless NPI pipelines are changed to work with multi-arch CI, such scripts that the older CI used cannot be deleted. A discussion on how multi-arch CI can intersect NPI pipelines is overdue.

jayhawk-commits:
Outdated and no longer compatible with current upload scripts.

PR #3664, opened 2026-02-27
Forward CC/CXX and launcher to Meson-based sysdep builds

### LINKED ISSUE #3522: manylinux build docker image fails to build with resource unavailable errors
Repro:

```
(host)$ ./build_tools/linux_portable_build.sh --docker=podman --interactive
$ pip install -r /therock/src/requirements.txt
$ cmake -S /therock/src -B /therock/output/build -GNinja . -DCMAKE_C_COMPILER_LAUNCHER=ccache   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache   -DTHEROCK_KPACK_SPLIT_ARTIFACTS=ON   -DTHEROCK_AMDGPU_DIST_BUNDLE_NAME=gfxlocal   '-DTHEROCK_AMDGPU_TARGETS=gfx1151'   '-DTHEROCK_DIST_AMDGPU_FAMILIES=gfx1151' -DTHEROCK_ENABLE_ALL=OFF -DTHEROCK_ENABLE_RAND=ON
$ cd /therock/output/build
$ ninja
```

Somewhere early (libdrm?) resource errors start appearing like:
```
[therock-libdrm] FAILED: amdgpu/librocm_sysdeps_drm_amdgpu.so.1.127.0.p/amdgpu_vm.c.o
[therock-libdrm] sccache cc -Iamdgpu/librocm_sysdeps_drm_amdgpu.so.1.127.0.p -Iamdgpu -I../patch_source/amdgpu -I. -I../patch_source -I../patch_source/include/drm -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c11 -O2 -g -inc
lude /therock/output/build/third-party/sysdeps/linux/libdrm/build/build/config.h -fPIC -pthread -Wsign-compare -Werror=undef -Werror=implicit-function-declaration -Wpointer-arith -Wwrite-strings -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declara
tions -Wnested-externs -Wpacked -Wswitch-enum -Wmissing-format-attribute -Wstrict-aliasing=2 -Winit-self -Winline -Wshadow -Wdeclaration-after-statement -Wold-style-definition -Wno-unused-parameter -Wno-attributes -Wno-long-long -Wno-missing-field-initi
alizers '-DAMDGPU_ASIC_ID_TABLE="/share/libdrm/amdgpu.ids"' -MD -MQ amdgpu/librocm_sysdeps_drm_amdgpu.so.1.127.0.p/amdgpu_vm.c.o -MF amdgpu/librocm_sysdeps_drm_amdgpu.so.1.127.0.p/amdgpu_vm.c.o.d -o amdgpu/librocm_sysdeps_drm_amdgpu.so.1.127.0.p/amdgpu_
vm.c.o -c ../patch_source/amdgpu/amdgpu_vm.c
[therock-libdrm] sccache: error: Resource temporarily unavailable (os error 11)
```
The issue seems to be that the drm build is implicitly picking up and trying to use sccache. If I `rm /usr/local/bin/sccache` then reconfigure/rebuild, it seems to work.

Not exactly sure what is going on here. Some foul play between the outer build and the meson internal build for drm?

marbre:
Meson does not inherit CMAKE_C_COMPILER_LAUNCHER, so any launcher configured at the CMake level (e.g. ccache via -DCMAKE_C_COMPILER_LAUNCHER) was silently dropped for the libdrm, amd-mesa, and libpciaccess builds. Instead Mesa auto-detects and picks a launcher, see https://mesonbuild.com/Feature-autodetection.html#ccache.

This computes THEROCK_MESON_CC and THEROCK_MESON_CXX in _therock_cmake_subproject_setup_toolchain and bakes them into the generated toolchain file. These combine the launcher (if any) and the compiler into a space-separated string as expected by meson. The correct compiler is selected by mirroring the existing toolchain logic: amd-llvm/amd-hip subprojects use the built compiler, all others use the CMake-specified one.

Fixes #3522.

### DIFF
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -1550,6 +1550,40 @@ function(_therock_cmake_subproject_setup_toolchain
   endif()
 
   string(APPEND _toolchain_contents "${_sanitizer_stanza}")
+
+  # Compute Meson-compatible CC/CXX env strings.
+  # To allow Meson-based subprojects to use the same compiler and launcher
+  # as CMake does, CC and CXX are passed. Otherweise, Meson auto-detects
+  # launchers and will use it automatically, resulting in not using the
+  # one selected in CMake-land.
+  # We mirror the toolchain selection logic above: amd-llvm/amd-hip
+  # subprojects use the built compiler, all others use the CMake-specified
+  # compiler.
+  if(compiler_toolchain STREQUAL "amd-llvm" OR compiler_toolchain STREQUAL "amd-hip")
+    set(_meson_actual_cc "${AMD_LLVM_C_COMPILER}")
+    set(_meson_actual_cxx "${AMD_LLVM_CXX_COMPILER}")
+  else()
+    set(_meson_actual_cc "${CMAKE_C_COMPILER}")
+    set(_meson_actual_cxx "${CMAKE_CXX_COMPILER}")
+  endif()
+  if(CMAKE_C_COMPILER_LAUNCHER)
+    # CMAKE_C_COMPILER_LAUNCHER is a CMake list (semicolon-separated) to support
+    # chained launchers (e.g. "ccache;distcc"). Meson expects a space-separated
+    # string, so convert before concatenating with the compiler path.
+    string(REPLACE ";" " " _meson_c_launcher "${CMAKE_C_COMPILER_LAUNCHER}")
+    set(_meson_cc "${_meson_c_launcher} ${_meson_actual_cc}")
+  else()
+    set(_meson_cc "${_meson_actual_cc}")
+  endif()
+  if(CMAKE_CXX_COMPILER_LAUNCHER)
+    string(REPLACE ";" " " _meson_cxx_launcher "${CMAKE_CXX_COMPILER_LAUNCHER}")
+    set(_meson_cxx "${_meson_cxx_launcher} ${_meson_actual_cxx}")
+  else()
+    set(_meson_cxx "${_meson_actual_cxx}")
+  endif()
+  string(APPEND _toolchain_contents "set(THEROCK_MESON_CC \"@_meson_cc@\")\n")
+  string(APPEND _toolchain_contents "set(THEROCK_MESON_CXX \"@_meson_cxx@\")\n")
+
   set(_compiler_toolchain_addl_depends "${_compiler_toolchain_addl_depends}" PARENT_SCOPE)
   set(_compiler_toolchain_init_contents "${_compiler_toolchain_init_contents}" PARENT_SCOPE)
   set(_build_env_pairs "${_build_env_pairs}" PARENT_SCOPE)
--- a/third-party/sysdeps/linux/amd-mesa/CMakeLists.txt
+++ b/third-party/sysdeps/linux/amd-mesa/CMakeLists.txt
@@ -60,6 +60,8 @@ add_custom_target(
     # This is required before patching because patch_source.sh modifies files in the libva subproject
     "${CMAKE_COMMAND}" -E env
       "LDFLAGS=-Wl,-rpath='$$ORIGIN' -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds"
+      "CC=${THEROCK_MESON_CC}"
+      "CXX=${THEROCK_MESON_CXX}"
       --
       "${MESON_BUILD}" setup
         --reconfigure
@@ -82,6 +84,8 @@ add_custom_target(
     # This ensures the build system uses the modified source files
     "${CMAKE_COMMAND}" -E env
       "LDFLAGS=-Wl,-rpath='$$ORIGIN' -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds"
+      "CC=${THEROCK_MESON_CC}"
+      "CXX=${THEROCK_MESON_CXX}"
       --
       "${MESON_BUILD}" setup
         --reconfigure
--- a/third-party/sysdeps/linux/libdrm/CMakeLists.txt
+++ b/third-party/sysdeps/linux/libdrm/CMakeLists.txt
@@ -77,6 +77,8 @@ add_custom_target(
     "${CMAKE_COMMAND}" -E env
       # Escaping hack: experimentally determined to persist through the layers.
       "LDFLAGS=-Wl,-rpath='$$ORIGIN' -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds"
+      "CC=${THEROCK_MESON_CC}"
+      "CXX=${THEROCK_MESON_CXX}"
       --
       "${MESON_BUILD}" setup "${CMAKE_CURRENT_BINARY_DIR}"
         --reconfigure
--- a/third-party/sysdeps/linux/libpciaccess/CMakeLists.txt
+++ b/third-party/sysdeps/linux/libpciaccess/CMakeLists.txt
@@ -66,6 +66,8 @@ add_custom_target(
     "${CMAKE_COMMAND}" -E env
       # Apply symbol versioning via version.lds. RPATH is set post-install via patch_install.py.
       "LDFLAGS=-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds"
+      "CC=${THEROCK_MESON_CC}"
+      "CXX=${THEROCK_MESON_CXX}"
       --
       "${MESON_BUILD}" setup "${CMAKE_CURRENT_BINARY_DIR}"
         --reconfigure


cmake/therock_subproject.cmake
```diff
@@ -1550,6 +1550,40 @@ function(_therock_cmake_subproject_setup_toolchain
   endif()
 
   string(APPEND _toolchain_contents "${_sanitizer_stanza}")
+
+  # Compute Meson-compatible CC/CXX env strings.
+  # To allow Meson-based subprojects to use the same compiler and launcher
+  # as CMake does, CC and CXX are passed. Otherweise, Meson auto-detects
+  # launchers and will use it automatically, resulting in not using the
+  # one selected in CMake-land.
```
ScottTodd, line 1558: nit: typo
```suggestion
  # Compute Meson-compatible CC/CXX env strings.
  # To allow Meson-based subprojects to use the same compiler and launcher
  # as CMake does, CC and CXX are passed. Otherwise, Meson auto-detects
  # launchers and will use it automatically, resulting in not using the
  # one selected in CMake-land.
```

cmake/therock_subproject.cmake
```diff
@@ -1550,6 +1550,40 @@ function(_therock_cmake_subproject_setup_toolchain
...
+  # We mirror the toolchain selection logic above: amd-llvm/amd-hip
+  # subprojects use the built compiler, all others use the CMake-specified
+  # compiler.
+  if(compiler_toolchain STREQUAL "amd-llvm" OR compiler_toolchain STREQUAL "amd-hip")
+    set(_meson_actual_cc "${AMD_LLVM_C_COMPILER}")
+    set(_meson_actual_cxx "${AMD_LLVM_CXX_COMPILER}")
+  else()
+    set(_meson_actual_cc "${CMAKE_C_COMPILER}")
+    set(_meson_actual_cxx "${CMAKE_CXX_COMPILER}")
+  endif()
```
ScottTodd, line 1568: I'm not thrilled with the idea of putting meson-specific code on this common code path and I think we can simplify a bit to make it more palatable. 

Above, the compiler is always added to the toolchain file once, then it _might_ be added again (overwriting):

```cmake
  string(APPEND _toolchain_contents "set(CMAKE_C_COMPILER \"@CMAKE_C_COMPILER@\")\n")
  string(APPEND _toolchain_contents "set(CMAKE_CXX_COMPILER \"@CMAKE_CXX_COMPILER@\")\n")

  if(NOT compiler_toolchain)
    ...
  elseif(compiler_toolchain STREQUAL "amd-llvm" OR compiler_toolchain STREQUAL "amd-hip")
    ...
    string(APPEND _toolchain_contents "set(CMAKE_C_COMPILER \"@AMD_LLVM_C_COMPILER@\")\n")
    string(APPEND _toolchain_contents "set(CMAKE_CXX_COMPILER \"@AMD_LLVM_CXX_COMPILER@\")\n")
```

that puts code like this in each `_toolchain.cmake` file:
```cmake
set(CMAKE_C_COMPILER "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/Hostx64/x64/cl.exe")
set(CMAKE_CXX_COMPILER "C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/Hostx64/x64/cl.exe")
set(CMAKE_C_COMPILER "D:/projects/TheRock/build/core/clr/dist/lib/llvm/bin/clang.exe")
set(CMAKE_CXX_COMPILER "D:/projects/TheRock/build/core/clr/dist/lib/llvm/bin/clang++.exe")
```

Here's some code golfed alternatives to this code block (from a session with Claude Code). Both avoid extra intermediate variables and branches, evaluating the existing values in the toolchain file:
1. Simplest, using `[=[` syntax I hadn't seen before (https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#bracket-argument):
    ```cmake
      string(APPEND _toolchain_contents [=[
      # Meson-compatible CC/CXX: combine launcher (if any) with the compiler.
      set(THEROCK_MESON_CC "${CMAKE_C_COMPILER}")
      set(THEROCK_MESON_CXX "${CMAKE_CXX_COMPILER}")
      if(CMAKE_C_COMPILER_LAUNCHER)
        string(REPLACE ";" " " _therock_meson_launcher "${CMAKE_C_COMPILER_LAUNCHER}")
        set(THEROCK_MESON_CC "${_therock_meson_launcher} ${CMAKE_C_COMPILER}")
      endif()
      if(CMAKE_CXX_COMPILER_LAUNCHER)
        string(REPLACE ";" " " _therock_meson_launcher "${CMAKE_CXX_COMPILER_LAUNCHER}")
        set(THEROCK_MESON_CXX "${_therock_meson_launcher} ${CMAKE_CXX_COMPILER}")
      endif()
      ]=])
    ```
    
2. Match project style, inline escapes:
    ```cmake
      string(APPEND _toolchain_contents "set(THEROCK_MESON_CC \"\${CMAKE_C_COMPILER}\")\n")
      string(APPEND _toolchain_contents "set(THEROCK_MESON_CXX \"\${CMAKE_CXX_COMPILER}\")\n")
      string(APPEND _toolchain_contents "if(CMAKE_C_COMPILER_LAUNCHER)\n")
      string(APPEND _toolchain_contents "  string(REPLACE \";\" \" \" _therock_meson_launcher
      \"\${CMAKE_C_COMPILER_LAUNCHER}\")\n")
      string(APPEND _toolchain_contents "  set(THEROCK_MESON_CC \"\${_therock_meson_launcher} \${CMAKE_C_COMPILER}\")\n")
      string(APPEND _toolchain_contents "endif()\n")
      string(APPEND _toolchain_contents "if(CMAKE_CXX_COMPILER_LAUNCHER)\n")
      string(APPEND _toolchain_contents "  string(REPLACE \";\" \" \" _therock_meson_launcher
      \"\${CMAKE_CXX_COMPILER_LAUNCHER}\")\n")
      string(APPEND _toolchain_contents "  set(THEROCK_MESON_CXX \"\${_therock_meson_launcher} \${CMAKE_CXX_COMPILER}\")\n")
      string(APPEND _toolchain_contents "endif()\n")
    ```

---

Another idea would be to push the code down into the sysdeps that use meson, either fully inlined to each callsite or using a shared `therock_meson_env.cmake` that they include.

```cmake
  include("${CMAKE_CURRENT_SOURCE_DIR}/../therock_meson_env.cmake")
```

  The helper itself would be something like:

```cmake
  # Assemble CC/CXX env strings for meson from CMake's compiler + launcher
  # variables (which are already set correctly in the sub-project toolchain).
  function(therock_get_meson_compiler_env out_cc out_cxx)
    set(_cc "${CMAKE_C_COMPILER}")
    set(_cxx "${CMAKE_CXX_COMPILER}")
    if(CMAKE_C_COMPILER_LAUNCHER)
      string(REPLACE ";" " " _c_launcher "${CMAKE_C_COMPILER_LAUNCHER}")
      set(_cc "${_c_launcher} ${_cc}")
    endif()
    if(CMAKE_CXX_COMPILER_LAUNCHER)
      string(REPLACE ";" " " _cxx_launcher "${CMAKE_CXX_COMPILER_LAUNCHER}")
      set(_cxx "${_cxx_launcher} ${_cxx}")
    endif()
    set(${out_cc} "${_cc}" PARENT_SCOPE)
    set(${out_cxx} "${_cxx}" PARENT_SCOPE)
  endfunction()
```

  Then in each consumer:
```cmake
  include("${CMAKE_CURRENT_SOURCE_DIR}/../therock_meson_env.cmake")
  therock_get_meson_compiler_env(_meson_cc _meson_cxx)
```

---

WDYT? The current code is pretty easy to read, but it does involve some extra branching that could complicate future maintenance.

marbre: I actually considered all the alternatives and went for extending `cmake/therock_subproject.cmake` as this seems most consistent. It duplicates some code but with that everything is in one place. However, the longer I look at alternatives a function (or a macro) might be suitable as well as this is indeed Meson specific code. It depends on how much we consider Meson to be "standard" / required (`therock_subproject` sets all the defaults for our CMake build system). If Meson is an integral part I think it is fine to keep it where it is, if it is rather an optional thing a function might not be too bad and I can give it a shot.

marbre:
Regarding the prior discussion if modifying the toolchain file or adding a function or macro is the better approach:

---
❯ Review commit a72542a02251a76d086a520e72512204b184b636 and commit 6ece687281b584ce054491306c3d5df2ca8c9e59. Which is the better approach?

● Both commits solve the same problem (forwarding CC/CXX + launcher to Meson-based sysdep builds), but differ significantly in where the logic lives.                                                                                 
                                                                                                                                                                                                                                      
  therock (a72542a) — Helper function approach                                                                                                                                                                                      
                                                                                                                                                                                                                                      
  - Adds cmake/therock_meson_env.cmake with therock_get_meson_compiler_env()                                                                                                                                                          
  - Each sysdep must include() the helper and call it explicitly                                                                                                                                                                      
  - Reads CMAKE_C_COMPILER from the sub-project's toolchain context (which is already set correctly)                                                                                                                                  
  - Handles chained launchers (ccache;distcc → space-separated) correctly                                                                                                                                                             
                                                                                                                                                                                                                                      
  therock (6ece687) — Bake into toolchain approach                                                                                                                                                                                    
                                                                                                                                                                                                                                      
  - Logic is added to _therock_cmake_subproject_setup_toolchain in cmake/therock_subproject.cmake                                                                                                                                     
  - THEROCK_MESON_CC / THEROCK_MESON_CXX are computed once and baked into the generated toolchain file
  - Sub-projects just use ${THEROCK_MESON_CC} — 2 lines vs 3 per sysdep                                                                                                                                                               
  - Explicitly handles the amd-llvm/amd-hip toolchain distinction (uses AMD_LLVM_C_COMPILER for those)                                                                                                                                
                                                                                                                                                                                                                                      
  Verdict: therock (6ece687) is the better approach                                                                                                                                                                                   
                                                                                                                                                                                                                                      
  The centralization in _therock_cmake_subproject_setup_toolchain is the right architectural choice:                                                                                                                                  
  
  1. Right layer: Compiler/launcher forwarding is a subproject infrastructure concern, not individual sysdep logic. therock_subproject.cmake already owns all toolchain selection.                                                    
  2. Lower friction for new sysdeps: A new Meson-based sysdep just uses the variable — no need to know about or include a helper.
  3. Complete: Explicitly handles the amd-llvm/amd-hip case (even if Meson sysdeps don't currently use that toolchain, it's correct by construction rather than relying on the sub-project toolchain already having it right).


❯ Should meson specific setup go to the toolchain file really?                                                                                                                                                                        
                                                               
● That's a fair pushback. The CMake toolchain file is a CMake-specific artifact — its job is to configure CMAKE_C_COMPILER, CMAKE_SYSTEM_NAME, etc. for CMake's own consumption. Stuffing THEROCK_MESON_CC into it is a leaky         
  abstraction: the toolchain file now has to know about Meson's conventions, and anyone debugging Meson compiler selection wouldn't think to look there.                                                                              
                                                                                                                                                                                                                                      
  The therock (a72542a) approach actually has better separation of concerns on this axis:                                                                                                                                                     
  - Meson-specific logic lives in a file named therock_meson_env.cmake — discoverable and self-contained                                                                                                                              
  - The toolchain infrastructure stays focused on CMake concerns                                                                                                                                                                      
  - The helper function is explicit at the call site about what it's doing and why
                                                                                                                                                                                                                                      
  The cost is 3 lines of boilerplate per sysdep (include + call + usage), but that's a reasonable price for not polluting the toolchain layer with build-system-specific knowledge.                                                   
                                                                                                                                                                                        
So I'd revise: therock (a72542a) is the better approach on architectural grounds, even though therock's centralization initially looks appealing. The centralization in therock (6ece687) is in the wrong place.

---

So there is no clear winner but I think I might prefer the function/macro-based approach over what I have sent for review here. Will spin up another PR.

marbre:
Function-based approach is implemented here: https://github.com/ROCm/TheRock/pull/3874

marbre:
Closing in favor of #3874.

PR #2999, opened 2026-01-19
Add new workflow to support jax branches and python support

kiran-thumma:
## Motivation

Refactors the JAX wheels CI to use a matrix (Python version × JAX branch), align with the PyTorch wheels pattern, and fix paths: rocm-jax at `rocm-jax/`, JAX source at `jax/`, wheel output at `rocm-jax/jax_rocm_plugin/wheelhouse`. Updates the default JAX branch from **rocm-jaxlib-v0.8.0** to **rocm-jaxlib-v0.8.2**.

## Changes

- **release_portable_linux_jax_wheels.yml** (new): Matrix orchestrator with `python_version: ["3.12"]` and `jax_ref: ["rocm-jaxlib-v0.8.2"]`. Calls `build_linux_jax_wheels.yml` per combination; no defaults in `with:`.
- **build_linux_jax_wheels.yml**: Added `jax_ref` input (default `rocm-jaxlib-v0.8.2`). Checkouts: rocm-jax and JAX at `path: rocm-jax` and `path: jax`; refs use `inputs.jax_ref` (with fallback). Build uses `--jax-source-dir="${{ github.workspace }}/jax"`. `PACKAGE_DIST_DIR` set to `rocm-jax/jax_rocm_plugin/wheelhouse`.
- **test_linux_jax_wheels.yml**: rocm-jax at `rocm-jax`, tests at `rocm-jax/jax_tests`; refs use `inputs.jax_ref || 'rocm-jaxlib-v0.8.2'`.
- **release_portable_linux_packages.yml**: Triggers `release_portable_linux_jax_wheels.yml` instead of the build workflow.

## Flow

`release_portable_linux_packages.yml` → `release_portable_linux_jax_wheels.yml` (matrix) → `build_linux_jax_wheels.yml` → `test_linux_jax_wheels.yml`.

## Test Plan

Manipulate the input of workflow dispatch to trigger a test job in order to test the yml workflow

## Test Result

passsing test: https://github.com/ROCm/TheRock/actions/runs/22120125752

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3678, opened 2026-02-27
Adding back jax 0.8.0 support.

JeniferC99:
## Motivation

We removed 0.8.0 support in https://github.com/ROCm/TheRock/pull/2999
There was request to bring it back and it is currently working. 
Internal ticket: https://amd-hub.atlassian.net/browse/AIDEVOPS-67

## Technical Details

Adding jax v. 0.8.0 in our current jax workflow with python 3.11-3.14
Nightlies will now build both jax v. 0.8.0 and 0.8.2

## Test Plan

User our workflow and update matrix to use jax 0.8.0 branch

## Test Result

Passing test: https://github.com/ROCm/TheRock/actions/runs/22495116958/job/65167455574
matrix change test: https://github.com/ROCm/TheRock/actions/runs/22500172269 - able to trigger all jax version

## Submission Checklist

### DIFF
--- a/.github/workflows/release_portable_linux_jax_wheels.yml
+++ b/.github/workflows/release_portable_linux_jax_wheels.yml
@@ -82,14 +82,6 @@ on:
         description: "TheRock branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow."
         type: string
         default: ''
-      jax_ref:
-        description: "jax ref to build. 'all' = build both 0.8.0 and 0.8.2."
-        type: choice
-        options:
-          - all
-          - rocm-jaxlib-v0.8.0
-          - rocm-jaxlib-v0.8.2
-        default: all
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


JeniferC99:
Passing test: https://github.com/ROCm/TheRock/actions/runs/22495116958/job/65167455574

.github/workflows/build_linux_jax_wheels.yml
```diff
@@ -167,12 +165,13 @@ jobs:
...
           ls -lah
           pushd rocm-jax
+          [[ "${{ inputs.jax_ref }}" == *"0.8.2"* ]] && SOURCE_ARG="--jax-source-dir=${{ github.workspace }}/jax" || SOURCE_ARG=""
           python3 build/ci_build \
             --compiler=clang \
             --python-versions="${{ inputs.python_version }}" \
             --rocm-version="${ROCM_VERSION}" \
             --therock-path="${{ inputs.tar_url }}" \
-            --jax-source-dir="${{ github.workspace }}/jax" \
+            $SOURCE_ARG \
```
charleshofer, line 174: Why do we ever build with `${{ github.workspace }}/jax` at all? Shouldn't we always use the JAX that Bazel pulls?

JeniferC99: I can add that to the workflow improvement pr I'm working on. This pr is solely just to add jax v. 0.8.0

JeniferC99: That is needed because without specifying, we always hit this issue FileNotFoundError: Did not find jaxlib wheel. but if it can be automatically found then that's better.

charleshofer: Ah, now I understand why you need it. We do need that jax repo for 0.8.2, and we'll probably need it for 0.8.0 as well.

PR #3680, opened 2026-02-27
Improve rocprofiler-compute test scripts

jbonnell-amd:
## Motivation

Making some minor changes to the way rocprofiler-compute tests are implemented

## Technical Details

- Updated `test_rocprofiler_compute.py`
  - Removed unused vars
  - Refactored to include `main` function and organized core logic into separate functions
- Updated `fetch_test_configurations.py`
  - Updated sharding definition for `rocprofiler_compute`
- Updated `install_additional_requirements.py`
  - Changed the way that `THEROCK_OUTPUT_DIR` is built
  - Updated input variable name in `install_requirements()` function
  - Added instructions on usage as a comment block at the top of the script

## Test Plan

Ensure builds and tests pass without any new regressions.

## Test Result

Tests still behave as expected and no regression issues noticed.

## Submission Checklist

### DIFF
--- a/build_tools/install_additional_requirements.py
+++ b/build_tools/install_additional_requirements.py
@@ -6,7 +6,7 @@
 """Script to install additional requirements.txt files for projects that require additional files for testing.
 
 Requires a requirements-file input parameter that is a list of comma separated paths to requirements.txt files.
-This path will always be relative to the absolute path of the OUTPUT_ARTIFACTS_DIR. 
+This path will always be relative to the absolute path of the OUTPUT_ARTIFACTS_DIR.
 
 Usage:
 python install_additional_requirements.py


build_tools/install_additional_requirements.py
```diff
@@ -10,13 +10,15 @@
 
 SCRIPT_DIR = Path(__file__).resolve().parent
 THEROCK_DIR = SCRIPT_DIR.parent
-THEROCK_OUTPUT_DIR = str(THEROCK_DIR / "build")
+THEROCK_OUTPUT_DIR = str(
+    THEROCK_DIR / os.getenv("OUTPUT_ARTIFACTS_DIR").removeprefix("./")
```
jbonnell-amd, line 42: This was added to try and avoid hardcoding a `build` string here. Since `OUTPUT_ARTIFACTS_DIR` seems to always be set to `./build` as a relative path, this will remove the `./` from `OUTPUT_ARTIFACTS_DIR` and build the absolute path from `THEROCK_DIR`.

This should ensure that, if `OUTPUT_ARTIFACTS_DIR` changes to something else (for example, `./build-dir`), that this script won't need to be updated to accommodate this new value.

jbonnell-amd:
Noticed some PyTorch and hip-tests failures, but they appear to be unrelated to these changes

build_tools/install_additional_requirements.py
geomin12: sounds good, but let's add a comment here to reflect this :)

build_tools/github_actions/test_executable_scripts/test_rocprofiler_compute.py
```diff
@@ -63,30 +33,66 @@
...
+
+
+def execute_tests():
+    # Sharding
+    shard_index = int(os.getenv("SHARD_INDEX", "1")) - 1
+    total_shards = int(os.getenv("TOTAL_SHARDS", "1"))
+
+    # Run tests
+    cmd = [
+        "ctest",
```
geomin12, line 69: do we support parallelization? `-j` flag? would help speed up tests if possible

geomin12 (commented):
i see it passing woo! overall looks good, just added a few comments

build_tools/install_additional_requirements.py
jbonnell-amd: I already added some comments regarding this at the top of the file, let me know if that is good enough or if I should explicitly add a comment to this portion of the code here instead (I mentioned that the path will always be relative to the absolute path of `OUTPUT_ARTIFACTS_DIR`)

build_tools/github_actions/test_executable_scripts/test_rocprofiler_compute.py
jbonnell-amd: We tried doing this before, however when running compute tests in parallel we often see intermittent test failures due to conflicts. @vedithal-amd could provide some more insight on this

vedithal-amd: To provide more context here, when multiple profiling related tests are run parallelly, they can put heavy stress on the underlying GPU and cause deadlocks and GPU hangs leading to test case failures. So, please let us not run GPU profiling related tests parallelly in order to prevent unstable CI.

However, it's totally fine to run profiling tests in parallel with analysis and unit tests (which does not use GPU).
I can help you identify which test cases can be run parallelly if required.

jbonnell-amd: Since we're using ctest, is it possible to add a `RUN_SERIAL` option to the heavy GPU tests in compute's `CMakeLists.txt`? That way, we can run one ctest command with parallel option set but it won't run the selected ones in parallel.

https://cmake.org/cmake/help/latest/prop_test/RUN_SERIAL.html

PR #3687, opened 2026-02-28
Fix provider plugin artifact names in Python packaging

SamuelReeder:
## Summary

- The CMake targets for MIOpen and hipBLASLt providers were renamed from `miopen-plugin` / `hipblaslt-plugin` to `miopenprovider` / `hipblasltprovider`
- `build_python_packages.py` still referenced the old hyphenated names, so the `libraries_artifact_filter` silently dropped the provider libraries from the wheels
- Updated the artifact name strings to match the current CMake target names

## Test plan

- [ ] Verify wheels build includes the provider libraries (`miopenprovider`, `hipblasltprovider`)

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -122,8 +122,8 @@ def libraries_artifact_filter(target_family: str, an: ArtifactName) -> bool:
             "fft",
             "hipdnn",
             "miopen",
-            "miopen-plugin",
-            "hipblaslt-plugin",
+            "miopenprovider",
+            "hipblasltprovider",
             "rand",
             "rccl",
         ]


SamuelReeder:
> I'm not a big fan of mocking like this as it makes it very difficult to refactor. But I'm going to approve so you all can make progress. Can you hit me with an issue to clean this up so it can be directly tested?

I was going to remove the tests and try to figure out something better, since it doesn't hit the condition I wanted. Would you rather I do that and then make an issue to test, or leave it for now?

---

I think I'm going to remove it for now and make an issue to follow-up on.

SamuelReeder:
Created an issue to investigate & implement appropriate tests: https://github.com/ROCm/TheRock/issues/3688

SamuelReeder:
@stellaraccident we're going to wait to merge until we inspect and test the wheels. Our green signal on the python packaging had this silent merge conflict included, so the providers aren't guaranteed to be correct until we test.

stellaraccident:
> @stellaraccident we're going to wait to merge until we inspect and test the wheels. Our green signal on the python packaging had this silent merge conflict included, so the providers aren't guaranteed to be correct until we test.

Your call on sequencing. I appreciate the time adding unit testing and an perfectly fine with followups to improve if it helps to not break flow.

PR #3694, opened 2026-03-01
[rocm-libraries] Fix rocBLAS OpenMP CONFIG find escaping build sandbox (GCC 15)

tlee933:
## Summary

- Adds `NO_DEFAULT_PATH` to the `find_package(OpenMP CONFIG ...)` call in `rocblas/clients/CMakeLists.txt`
- Without it, CMake discovers system-installed OpenMP configs via PATH-derived prefixes (e.g. `/opt/rocm/lib/cmake/openmp`), injecting `-isystem` flags for the Clang resource directory
- This breaks HIP device compilation on the `amdgcn` target when using GCC 15 host headers (Fedora 43+): `error: no member named 'int_fast8_t' in the global namespace`
- The existing fallback to `find_package(OpenMP)` in MODULE mode continues to work correctly

## Reproduction

Build rocBLAS clients with:
- Clang (ROCm) as the HIP compiler  
- GCC 15 host headers (Fedora 43, or any distro shipping GCC 15's libstdc++)
- A system ROCm installation present at `/opt/rocm`

The CONFIG-mode `find_package` escapes the build sandbox and picks up the system OpenMP, whose `INTERFACE_INCLUDE_DIRECTORIES` point into the Clang resource dir. When those get added as `-isystem` to device compilation, `stdint.h` on `amdgcn` fails because `__x86_64__` is not defined.

## References

- https://gcc.gnu.org/gcc-15/porting_to.html

## Test plan

- [ ] Build rocBLAS with Clang + GCC 15 host headers on a system with `/opt/rocm` present
- [ ] Verify `find_package(OpenMP CONFIG)` no longer escapes to system paths
- [ ] Verify OpenMP is still found via MODULE mode fallback

### DIFF
--- a/patches/amd-mainline/rocm-libraries/0001-rocblas-Prevent-OpenMP-CONFIG-find-from-escaping-buil.patch
+++ b/patches/amd-mainline/rocm-libraries/0001-rocblas-Prevent-OpenMP-CONFIG-find-from-escaping-buil.patch
@@ -1,0 +1,51 @@
+From c7dcdeb057acd57e725a7b64fecff77d3b91000d Mon Sep 17 00:00:00 2001
+From: therockbot <therockbot@amd.com>
+Date: Sat, 14 Feb 2026 05:57:38 -0700
+Subject: [PATCH] rocblas: Prevent OpenMP CONFIG find from escaping build
+ sandbox
+
+When building rocBLAS clients with Clang and GCC 15 host headers,
+find_package(OpenMP CONFIG) without NO_DEFAULT_PATH allows CMake to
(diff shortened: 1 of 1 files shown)


patches/amd-mainline/rocm-libraries/0001-rocblas-Prevent-OpenMP-CONFIG-find-from-escaping-buil.patch
marbre, line 1: Please see https://github.com/ROCm/TheRock/tree/main/patches#patches-for-therock. We strongly discourse adding new patches and those should rather go upstream.

PR #3696, opened 2026-03-02
[rocprofiler-systems] Build rocprofiler-systems examples as target-specific

kcossett-amd:
## Motivation

**DO NOT REVIEW**: I am just testing some workflows right now :D

## Submission Checklist


kcossett-amd:
With #3698 , this PR is most likely no longer needed. We'll just build `rocprofiler-systems` using all possible GPU targets.

marbre:
> **DO NOT REVIEW**: I am just testing some workflows right now :D

Fyi, in that case you can either send a draft PR but actually you don't need an PR at all. You can trigger a CI run on any branch in TheRock which also saves resources as you can choose which architectures to build for an run on.

Issue #2070, opened 2025-11-10
[PyTorch] TestBinaryUfuncsCUDA::test_batch_vs_slicing regression across all PyTorch versions


HereThereBeDragons:
There are new tests failing in pytest for PyTorch 2.7 and nightly that havent been failing before:

https://github.com/ROCm/TheRock/actions/runs/19179292911/job/54834551403

for pytorch 2.7
```
TestBinaryUfuncsCUDA::test_batch_vs_slicing___rpow___cuda_complex64 
TestBinaryUfuncsCUDA::test_batch_vs_slicing__refs_pow_cuda_complex32 
TestBinaryUfuncsCUDA::test_batch_vs_slicing__refs_pow_cuda_complex64 
TestBinaryUfuncsCUDA::test_batch_vs_slicing_pow_cuda_complex64 
```

for nightly (same tests failing as 2.7 and + 3 more tests failing)

```
TestTorch::test_index_add_correctness  << only nightly
TestFXMemoryProfiler::test_fx_memory_profiler_augmentation  << only nightly
TestBinaryUfuncsCUDA::test_batch_vs_slicing___rpow___cuda_complex64
TestBinaryUfuncsCUDA::test_batch_vs_slicing__refs_pow_cuda_complex32
TestBinaryUfuncsCUDA::test_batch_vs_slicing__refs_pow_cuda_complex64
TestBinaryUfuncsCUDA::test_batch_vs_slicing_pow_cuda_complex64
TestAutogradStreamSynchronizationCUDA::test_warn_on_accumulate_grad_stream_mismatch_flag_cuda  << only nightly
```

short test summary pytorch 2.7
```
=========================== short test summary info ============================
FAILED [0.0033s] external-builds/pytorch/pytorch/test/test_binary_ufuncs.py::TestBinaryUfuncsCUDA::test_batch_vs_slicing___rpow___cuda_complex64 - AssertionError: Tensor-likes are not close!

Mismatched elements: 647 / 16384 (3.9%)
Greatest absolute difference: 11380077887488.0 at index (31, 186) (up to 1e-05 allowed)
Greatest relative difference: 4.084608917764854e-06 at index (23, 352) (up to 1.3e-06 allowed)
FAILED [0.0153s] external-builds/pytorch/pytorch/test/test_binary_ufuncs.py::TestBinaryUfuncsCUDA::test_batch_vs_slicing__refs_pow_cuda_complex32 - AssertionError: Tensor-likes are not close!

[...]

PR #3704, opened 2026-03-02
PyTorch UT: Update skip_tests based on nightly 20260302

HereThereBeDragons:
This PR adjust the skip_test logic based on results from nightly 20260302:

- Enable better arch matching strategy for skip_tests
- Move test_allocator_settings and test_cpp_warnings_have_python_context_cuda to common skipping for 2.10
- Move test_fx_memory_profiler_augmentation to generic.py
- Renable all cuda pow tests (see #2070)

To note:
- disabled only for pytorch 2.8 and gfx120X: test_repeat_graph_capture_cublas_workspace_memory --> this might be true regression

### DIFF
--- a/external-builds/pytorch/skip_tests/README.md
+++ b/external-builds/pytorch/skip_tests/README.md
@@ -41,11 +41,17 @@ skip_tests = {
     "common": {
         <PyTorch test module> : [ <Tests> ],
     },
-    "<amdgpu family>": {,
+    "<amdgpu family short form>": {,
         <PyTorch test module> : [ <Tests> ],
     },
 }
 ```
+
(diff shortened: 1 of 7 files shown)


HereThereBeDragons:
No, I didnt. Never have before for the pytorch adjustments as it was always 1:1 copy from failed tests in the nightlies to the script.
What I did is verified locally that the new selection logic in skip test is working correctly and now really picks up tests to be skipped for gfx120 (which wasnt the case before - before only exact matchs of gpu arch would skip it (e.g. gfx120X-all runs on gfx1200 or gfx1201)

But here is a newly started gfx120X-all linux run: https://github.com/ROCm/TheRock/actions/runs/22614469327 with the nightly mentioned above.

HereThereBeDragons:
test run looks good

Issue #2406, opened 2025-12-03
[Feature] Evaluate using PyTorch DISABLED_TESTS_FILE to skip unit tests in our test workflows


ScottTodd:
PyTorch has a dynamic list of disabled tests that can be updated by filing github issues and _not just by modifying code_. See the dashboard at https://hud.pytorch.org/disabled.

This file: https://github.com/pytorch/pytorch/blob/main/tools/stats/import_test_stats.py loads the underlying list of disabled tests from https://ossci-metrics.s3.amazonaws.com/disabled-tests-condensed.json.

Note that the list is for "latest pytorch" and I don't think historical versions for prior releases are maintained.

We should

- [ ] Diff our https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch/skip_tests against that list
- [ ] Explore how we can combine the skip_tests file with the upstream disabled-tests-condensed.json file to generate a consolidated skip list for the PR run
- [ ] Reformat skip_tests files to match the upstream file?
- [ ] If there are test cases in both then maybe we can use the upstream list [in addition to our own extra filtering]

cc @jeffdaily @HereThereBeDragons

ScottTodd:
I don't see much documentation in the pytorch repository, so here's a summary from Claude:

---

> PyTorch has the ability to filter/skip/disable tests based on github issues. Can you find the code that references
 that external list of disabled tests when running on CI?

● I'll help you find the code that handles test filtering based on GitHub issues in PyTorch's CI system. Let me
  search for this functionality.

● Explore(Find GitHub issue test filtering)
  ⎿  Done (41 tool uses · 82.1k tokens · 3m 23s)

● Perfect! I found the code that handles test filtering based on GitHub issues. Here's the key information:

  Main External List Source

  The disabled tests are fetched from an S3 bucket:

  URL: https://ossci-metrics.s3.amazonaws.com/disabled-tests-condensed.json

  Code location: `tools/stats/import_test_stats.py:241-256`

```python
  def get_disabled_tests(
      dirpath: str, filename: str = DISABLED_TESTS_FILE
  ) -> dict[str, Any] | None:
      try:
          url = "https://ossci-metrics.s3.amazonaws.com/disabled-tests-condensed.json"
          return fetch_and_cache(dirpath, filename, url, process_disabled_test)
      except Exception:
          print("Couldn't download test skip set, leaving all tests enabled...")
          return {}
```

  Test Skipping Logic

[...]

Issue #3075, opened 2026-01-23
[Feature]: Evaluate whether sharding is enabled for PyTorch tests in our test workflows


jithunnair-amd:
### Suggestion Description

The pytorch test infra needs a [test-times.json](https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/test-times.json) file to determine the run times for each test suite. Based on the run times, it will shard each test suite to break it into smaller running chunks/shards. 

This file needs to be downloaded during the test run, so we added code in run_test.py via PR [2921](https://github.com/ROCm/pytorch/pull/2921) in one of the ROCm/pytorch branches. However, that's not necessarily the best approach.

TODO:
- [x] Invoke `export_test_times.py` from wrapper script around run_tests.py: part of changes in [PR 3705](https://github.com/ROCm/TheRock/pull/3705)
- [ ] Download a snapshot version for release/2.11, release/2.10, release/2.9 branch and commit it to branch
- [x] Ensure all test run environment is set up to enable sharding in theRock test workflows
- [ ] Have different test times used for sharding for each architecture?
- [ ] How to handle cases where shardig

Issue #999, opened 2025-07-10
[Windows] PyTorch processes do not terminate on their own


ScottTodd:
Repro steps:

1. Follow https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-pytorch-python-packages to install torch:
    
    ```bash
    python -m pip install --index-url https://d2awnip2yjpvqn.cloudfront.net/v2/gfx110X-dgpu/ torch
    
    (.venv) λ pip freeze
    filelock==3.18.0
    fsspec==2025.5.1
    Jinja2==3.1.6
    MarkupSafe==3.0.2
    mpmath==1.3.0
    networkx==3.5
    rocm==7.0.0.dev0+e96d36b9b628476463ef6cecee752f601052a4d2
    rocm-sdk-core==7.0.0.dev0+e96d36b9b628476463ef6cecee752f601052a4d2
    rocm-sdk-libraries-gfx110X-dgpu==7.0.0.dev0+e96d36b9b628476463ef6cecee752f601052a4d2
    setuptools==80.9.0
    sympy==1.14.0
    torch==2.7.0a0+rocm7.0.0.dev0.e96d36b9b628476463ef6cecee752f601052a4d2
    typing_extensions==4.14.0
    ```

2. Run the minimal repro:

    ```python
    # test.py
    import torch
    tensor_gpu = torch.randn(1, 1, device="cuda")
    # Calling `python test.py` does _not_ terminate
    ```
    
    ```python
    # test_print.py
    import torch
    tensor_gpu = torch.randn(1, 1, device="cuda")
    print(tensor_gpu)
    # Calling `python test_print.py` _does_ terminate
    ```

    ```python
    # test_with_synchronize.py
    import torch
    tensor_gpu = torch.randn(1, 1, device="cuda")
    torch.cuda.synchronize()
    # Calling `python test_with_synchronize..py` _does_ terminate
    ```

---

You can also install pytest and run the tests at https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch/smoke-tests if you remove this line of code: https://github.com/ROCm/TheRock/blob/d7a2241eb8076de014485e534cbf4fc70c8f3f0d/external-builds/pytorch/smoke-tests/pytorch_smoke_test.py#L92-L95

```
(.venv) λ pip install pytest
(.venv) λ pytest -v D:\projects\TheRock\external-builds\pytorch\smoke-tests
[...]

PR #3705, opened 2026-03-02
Run PyTorch unit tests using run_test.py

ethanwee1:
4 shard run: 
with mi325: https://github.com/ROCm/TheRock/actions/runs/22960250009
with mi355: https://github.com/ROCm/TheRock/actions/runs/22934947300
Resolves 
https://github.com/ROCm/TheRock/issues/2406
https://github.com/ROCm/TheRock/issues/3075

Testing: https://github.com/ROCm/TheRock/actions/runs/22864069244/job/66325924683
Ran 4 shards with the limited set of unit test suites (same as in https://github.com/ROCm/TheRock/blob/a926e619ad3fc2235a265ceee09166e977e4ecce/external-builds/pytorch/run_pytorch_tests.py#L255)

3 errors were observed:
`test_nn.py::TestNN::test_pdist_inf_nan_propagation`
`test_nn.py::TestNNDeviceTypeCUDA::test_CTCLoss_cudnn_cuda`
and
```
  File "/__w/TheRock/TheRock/external-builds/pytorch/pytorch/test/test_cuda.py", line 36, in <module>
    from torch.testing._internal.common_cuda import (
ImportError: cannot import name 'blas_library_context' from 'torch.testing._internal.common_cuda' (/__w/TheRock/TheRock/.venv/lib/python3.12/site-packages/torch/testing/_internal/common_cuda.py)
```


external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,301 @@
...
+        os.environ["TEST_CONFIG"] = test_config
+
+    test_dir = str(pytorch_dir / "test")
+    old_pythonpath = os.getenv("PYTHONPATH", "")
+    if old_pythonpath:
+        os.environ["PYTHONPATH"] = f"{test_dir}:{old_pythonpath}"
+    else:
+        os.environ["PYTHONPATH"] = test_dir
+
+    fetch_disabled_tests(test_dir)
```
jithunnair-amd: @ethanwee1 Do we need this step to be explicitly called when `run_test.py` will download the file anyway: https://github.com/pytorch/pytorch/blob/2f0a6bd3f5efddf0eadf6252924ac4bd0276ee71/test/run_test.py#L2143?

ethanwee1: No, I suppose that is not necessary. I just wanted to guarantee that the script would download it

jithunnair-amd: Okay, let's remove it. If you find a snippet like below in the log file, it means it is downloading the file:
```
Downloading https://ossci-metrics.s3.amazonaws.com/disabled-tests-condensed.json to /var/lib/jenkins/pytorch/test/.pytorch-disabled-tests.json
```

ethanwee1:
https://github.com/ROCm/TheRock/actions/runs/22640239769
Current state of testing

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,292 @@
...
+    cmd = [sys.executable, run_test_path]
+
+    if not args.no_exclude_jit_executor:
+        cmd.append("--exclude-jit-executor")
+    if not args.no_exclude_distributed:
+        cmd.append("--exclude-distributed-tests")
+    if not args.no_exclude_quantization:
+        cmd.append("--exclude-quantization-tests")
+
+    cmd.append("--keep-going")
```
ethanwee1, line 302: @ScottTodd We are going to add the --keep-going flag because we want to get the full run of UT's rather than stopping at the first failure since this is a nightly job that will not run on every commit.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,301 @@
...
+
+from pytorch_utils import (
+    detect_pytorch_version,
+    set_gpu_execution_policy,
+)
+
+THIS_SCRIPT_DIR = Path(__file__).resolve().parent
+
+#Use mi300 test times because https://github.com/ROCm/TheRock/actions/workflows/test_pytorch_wheels.yml chooses mi300 as default
+ROCM_BUILD_ENVIRONMENT = "linux-noble-rocm-py3.12-mi300"
```
erman-gurses: `ROCM_BUILD_ENVIRONMENT` is hardcoded in both the workflow and the script. That seems fragile, especially since this workflow already takes `python_version`, `amdgpu_family`, and runner as inputs.

If those change, this value can become stale or misleading relative to the actual test environment. Can we prefer the workflow-provided env var here and only set a default in the script if it is missing?

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,301 @@
...
+]
+
+
+def setup_env(pytorch_dir: Path, test_config: str) -> None:
+    os.environ["CI"] = "1"
+    os.environ["BUILD_ENVIRONMENT"] = ROCM_BUILD_ENVIRONMENT
+    os.environ["PYTORCH_TEST_WITH_ROCM"] = "1"
+    os.environ["PYTORCH_TESTING_DEVICE_ONLY_FOR"] = "cuda"
+    os.environ["PYTORCH_PRINT_REPRO_ON_FAILURE"] = "0"
+    os.environ["MIOPEN_CUSTOM_CACHE_DIR"] = tempfile.mkdtemp()
```
erman-gurses: `setup_env()` is force-setting several environment variables unconditionally, including BUILD_ENVIRONMENT. Since the workflow already sets these, could we use `os.environ.setdefault(...)` for values that are meant to have defaults?

That would make the script easier to reuse outside this exact workflow and avoid unintentionally overriding caller-provided values.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+    parser.add_argument(
+        "--include",
+        nargs="+",
+        default=None,
+        metavar="TEST",
+        help="Only run these test files (passed to run_test.py --include). "
+        "Also settable via TESTS_TO_INCLUDE env var (run_test.py reads it directly). "
+        "Default: run all tests.",
+    )
```
erman-gurses, line 207: The `--include` help says “Default: run all tests”, but in this workflow we always set `TESTS_TO_INCLUDE` to a limited default list of suites.

I think the current help text is a bit misleading in the context this script is actually used. Maybe reword it to clarify that the script itself defaults to “all tests”, but the workflow currently provides a limited default include set via `TESTS_TO_INCLUDE`.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,301 @@
...
+        action="store_true",
+        default=False,
+        help="Pass --dry-run to run_test.py to list tests without running them.",
+    )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        default=True,
+        help="Pass --verbose to run_test.py (default: True).",
+    )
```
erman-gurses: Minor argparse nit: `--verbose` is declared with `store_true` but already defaults to True, so it is effectively always enabled and the flag itself does not change behavior.

If the intent is “always verbose”, maybe just drop the argument and append `--verbose` unconditionally. Otherwise, this probably wants default=False.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,229 @@
...
+run-name: Test PyTorch Full (${{ inputs.amdgpu_family }}, ${{ inputs.torch_version}}, ${{ inputs.test_runs_on }})
+
+jobs:
+  test_wheels:
+    name: Test PyTorch Full | ${{ inputs.amdgpu_family }} (shard ${{ matrix.shard }}/4)
+    runs-on: ${{ inputs.test_runs_on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
```
erman-gurses: The shard count is currently duplicated in a few places (job name, matrix, and NUM_TEST_SHARDS) which makes later tuning a bit error-prone.
Could be worth centralizing this via a workflow env/variable if we expect the shard count to change again.

erman-gurses (commented):
Also your pre-commit is failing.

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: BUILD_ENVIRONMENT is now set via os.environ.setdefault(), so the workflow YAML value takes precedence if set. 
The script now dynamically selects the correct value from GFX_TO_BUILD_ENV based on the detected GPU arch, falling back to ROCM_BUILD_ENVIRONMENT_DEFAULT for something not set. We have to use the "linux-XXX-rocm-py3.XX-XXX" because that is the value that is set it PyTorch's generated https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/test-times.json. But I added a dictionary that maps to the gfx arch's.

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: All env vars in setup_env() now use setdefault.

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: Updated

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: The --verbose argument is removed entirely. --verbose is now unconditionally appended in build_run_test_cmd.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Updated so that shard count is now defined once per matrix entry (num_shards) and referenced by both the job name and NUM_TEST_SHARDS

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,307 @@
...
+        cmd.extend(["--exclude"] + args.exclude)
+
+    if tests_to_skip:
+        cmd.extend(["-k", tests_to_skip])
+
+    cmd.extend(passthrough_args)
+    return cmd
+
+
+def main() -> int:
```
erman-gurses: Small style-guide nit: would it make sense to switch `main()` to `main(argv: list[str]) -> int` and pass `sys.argv[1:]` from the `__main__ `guard? That tends to make CLI code easier to test and reuse.
https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#script-structure-and-organization

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,307 @@
...
+        cmd = build_run_test_cmd(args, tests_to_skip, passthrough_args)
+        print(f"Executing: {' '.join(cmd)}")
+
+        result = subprocess.run(cmd, cwd=str(args.pytorch_dir))
+        print(f"run_test.py finished with return code: {result.returncode}")
+        return result.returncode
+
+    except (ValueError, IndexError) as e:
+        print(f"[ERROR] Exception in PyTorch full test runner: {e}")
+        return 1
```
erman-gurses: It would be good to make it broad in a way that may hide distinct failure modes:

- bad shard/env parsing
- unexpected structure from set_gpu_execution_policy
- bugs in skip-test generation

to distinguish error conditions and fail fast and loud.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,307 @@
...
+            tests_to_skip = args.k
+        else:
+            tests_to_skip = get_tests(
+                amdgpu_family=first_arch,
+                pytorch_version=pytorch_version,
+                platform=platform.system(),
+                create_skip_list=not args.debug,
+            )
+
+        setup_env(args.pytorch_dir, args.test_config, amdgpu_family=args.amdgpu_family)
```
erman-gurses: Since `setup_env()` now takes a few config-like parameters, would it be clearer to call it with fully named arguments?
Example:
```
setup_env(
    pytorch_dir=args.pytorch_dir,
    test_config=args.test_config,
    amdgpu_family=args.amdgpu_family,
)
```

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+def force_exit_with_code(retcode: int) -> None:
+    import signal
+
+    retcode_file = Path("run_pytorch_tests_full_exit_code.txt")
+    retcode_int = int(retcode)
+    print(f"Writing retcode {retcode_int} to '{retcode_file}'")
+    with open(retcode_file, "w") as f:
+        f.write(str(retcode_int))
```
erman-gurses: `force_exit_with_code()` uses plain `open(...)` instead of `Path.write_text()` - suggesting to use `pathlib` for filesystem operations. Since `retcode_file` is already a Path, this could be simpler and more consistent as:
`retcode_file.write_text(str(retcode_int))`

ScottTodd: We'll be able to remove the "force exit" code soon. A fix for https://github.com/ROCm/TheRock/issues/999 is making its way here via rocm-systems.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,233 @@
...
+    defaults:
+      run:
+        shell: bash
+    env:
+      VENV_DIR: ${{ github.workspace }}/.venv
+      AMDGPU_FAMILY: ${{ inputs.amdgpu_family }}
+      SHARD_NUMBER: ${{ matrix.shard }}
+      NUM_TEST_SHARDS: ${{ matrix.num_shards }}
+      CI: "1"
+      BUILD_ENVIRONMENT: "linux-noble-rocm-py3.12-mi300"
```
amd-shiraz: nit: why does MI300 be hardcoded here and not coming from some variable like derive it from AMDGPU_FAMILY ?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,233 @@
...
+      PYTORCH_TEST_WITH_ROCM: "1"
+      PYTORCH_TESTING_DEVICE_ONLY_FOR: "cuda"
+      TEST_CONFIG: "default"
+      TESTS_TO_INCLUDE: ${{ inputs.tests_to_include }}
```
amd-shiraz, line 115: nit: can we check if this is actually converting into a --include argument at the end ? 
recommendation:  pass --include ${{ inputs.tests_to_include }} explicitly from the workflow

.github/workflows/test_pytorch_wheels_full.yml
ScottTodd: Yeah this doesn't seem right. We run these tests on multiple GPU types across instinct (mi300, mi325, etc.) and radeon (w7900, strix halo, etc.), Linux _and_ Windows

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: https://github.com/pytorch/pytorch/blob/a3ca67c68bd98ff74792b105dc8583dc33eef39b/test/run_test.py#L2117
TESTS_TO_INCLUDE is the canonical way to set which test files are included.

ethanwee1:
@amd-shiraz, Requesting rereview after addressing your comments.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+THIS_SCRIPT_DIR = Path(__file__).resolve().parent
+
+# Maps AMDGPU_FAMILY to BUILD_ENVIRONMENT for test timing data lookup.
+# Must stay in sync with pytorch/pytorch PR #176445.
+# mi300 uses noble/py3.12; all others use jammy/py3.10.
+AMDGPU_FAMILY_TO_BUILD_ENV = {
+    "gfx90X-dcgpu": "linux-jammy-rocm-py3.10-mi200",
+    "gfx94X-dcgpu": "linux-noble-rocm-py3.12-mi300",
+    "gfx950-dcgpu": "linux-jammy-rocm-py3.10-mi355",
+    "gfx110X-all": "linux-jammy-rocm-py3.10-navi31",
```
jithunnair-amd, line 74: @ScottTodd This will be a maintenance burden, just FYI, since we will likely change the OS version or python version for these jobs upstream and that will impact this logic. Having said that, it's definitely a good thing to use GPU-specific timings for more accurate sharding. Once https://github.com/pytorch/pytorch/pull/176445 goes in, we will at least have the GPU SKU suffix to always tie into. Maybe searching through the keys of the dict to pick up one that matches the corresponding SKU (e.g. searching for "-mi200" in the keys) might be a more robust method?

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd: Please keep in mind that this `BUILD_ENVIRONMENT` env var is not merely indicating what GPU we are running tests for, but instead to look up a matching key in a dict inside [`test-times.json`](https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/test-times.json). So we can't use any arbitrary values. However, Ethan's changes in https://github.com/ROCm/TheRock/pull/3705/changes/1ca12b35e5303319b9c823cfb394c387126860b2 make this a bit more GPU-specific, while still using key values that are present in the aforementioned dict.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -1,6 +1,3 @@
-# Copyright Advanced Micro Devices, Inc.
-# SPDX-License-Identifier: MIT
-
```
jithunnair-amd, line 3: @ethanwee1 Need the license here as well as in the `.github/workflows/test_pytorch_wheels_full.yml` file.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+        type: string
+        default: "3.12"
+      torch_version:
+        description: torch package version to install. (e.g. "2.7.1+rocm7.10.0a20251120")
+        required: true
+        type: string
+      pytorch_git_ref:
+        description: PyTorch ref to checkout test sources from. (e.g. "nightly", or "release/2.7")
+        type: string
+        default: "release/2.7"
```
HereThereBeDragons: as we dropped 2.7 support, the default should become maybe 2.9?
(same hten also for the other test yml script)

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+      tests_to_include:
+        description: List of test suites to run
+        type: string
+        default: test_nn test_torch test_cuda test_unary_ufuncs test_binary_ufuncs test_autograd
```
HereThereBeDragons:
```suggestion
        default: "test_nn test_torch test_cuda test_unary_ufuncs test_binary_ufuncs test_autograd"
```

everywhere else we use "

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+        type: string
+      python_version:
+        required: true
+        type: string
+      torch_version:
+        required: true
+        type: string
+      pytorch_git_ref:
+        type: string
+        default: "release/2.7"
```
HereThereBeDragons: see above

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+      repository:
+        description: "Repository to checkout. Otherwise, defaults to `github.repository`."
+        type: string
+      ref:
+        description: "Branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow."
+        type: string
+      tests_to_include:
+        description: List of test suites to run
+        type: string
+        default: test_nn test_torch test_cuda test_unary_ufuncs test_binary_ufuncs test_autograd
```
HereThereBeDragons: see above

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+      run:
+        shell: bash
+    env:
+      VENV_DIR: ${{ github.workspace }}/.venv
+      AMDGPU_FAMILY: ${{ inputs.amdgpu_family }}
+      SHARD_NUMBER: ${{ matrix.shard }}
+      NUM_TEST_SHARDS: ${{ matrix.num_shards }}
+      CI: "1"
+      PYTORCH_TEST_WITH_ROCM: "1"
+      PYTORCH_TESTING_DEVICE_ONLY_FOR: "cuda"
```
HereThereBeDragons, line 113: did you double check what the default behavior is of the run scripts? maybe we dont have to set all the vars here.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+            --torch-version=${{ inputs.torch_version }} \
+            --pytorch-git-ref=${{ inputs.pytorch_git_ref }} \
+            --index-url=${{ inputs.package_index_url }} \
+            --index-subdir=${{ inputs.amdgpu_family }}
+
+      - name: Set git options
+        run: |
+          git config --global core.longpaths true
+
+      - name: Checkout PyTorch Source Repos from nightly branch
```
HereThereBeDragons, line 148: we might want to keep a comment about why we dont need hipify here?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -0,0 +1,232 @@
...
+        run: |
+          python -m pip install -r external-builds/pytorch/requirements-test.txt
+          python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
+          pip freeze
+
+      - name: Run rocm-sdk sanity tests
+        run: |
+          rocm-sdk test
+
+      - name: Print environment variables
```
HereThereBeDragons: please remove the above section with 1) extending the git config section
```
      # safe.directory must be set before Runner Health Status
      - name: Adjust git config
        run: |
          git config --global --add safe.directory $PWD
          git config fetch.parallel 10
```

and 2) using this script instead
```
- name: Runner health status
        run: |
          ./build_tools/health_status.py
```

you can then also remove the `pip freeze` above,

the script gives server status, pip packages installed, etc.

see `.github/workflows/build_portable_linux_artifacts.yml`
and for output see e.g.  https://github.com/ROCm/TheRock/actions/runs/21412357929/job/61652257254

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+    if test_config != "distributed":
+        os.environ["PYTORCH_TEST_RUN_EVERYTHING_IN_SERIAL"] = "1"
+
+    test_dir = str(pytorch_dir / "test")
+    old_pythonpath = os.getenv("PYTHONPATH", "")
+    if old_pythonpath:
+        os.environ["PYTHONPATH"] = f"{test_dir}:{old_pythonpath}"
+    else:
+        os.environ["PYTHONPATH"] = test_dir
+
```
HereThereBeDragons, line 125: any reason why this section got deleted?

```
    # Force update the PYTHONPATH to be part of the sys path
    # Otherwise our current python process that will run pytest will NOT
    # find it and pytest will crash!
    if test_dir not in sys.path:
        sys.path.insert(0, test_dir)
```

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+
+def print_env() -> None:
```
HereThereBeDragons, line 133: if you want to put more time into formatting, have a look at `build_tools/hack/env_check/device.py  def section_bar(self, title: str):`

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        "--pytorch-dir",
+        type=Path,
+        default=THIS_SCRIPT_DIR / "pytorch",
+        help="Path to the PyTorch repository root.",
+    )
+    parser.add_argument(
+        "--test-config",
+        type=str,
+        default=os.getenv("TEST_CONFIG", "default"),
+        help='TEST_CONFIG value for run_test.py sharding/config logic (default: "default").',
```
HereThereBeDragons, line 182: please improve the description. 
for all the args that also can be set by a environment var it should be clearly stated.
and what does default mean here?

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        type=int,
+        default=int(os.getenv("NUM_TEST_SHARDS", "0")),
+        help="Total number of shards. Also reads NUM_TEST_SHARDS env var.",
+    )
+    parser.add_argument(
+        "--include",
+        nargs="+",
+        default=None,
+        metavar="TEST",
+        help="Only run these test files (passed to run_test.py --include). "
```
HereThereBeDragons: we need examples for this at the beginning of the file.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        "Also settable via TESTS_TO_INCLUDE env var (run_test.py reads it directly). "
+        "If neither is set, run_test.py runs all tests. The workflow may "
+        "provide a limited default set via TESTS_TO_INCLUDE.",
+    )
+    parser.add_argument(
+        "--exclude",
+        nargs="+",
+        default=None,
+        metavar="TEST",
+        help="Exclude these test files (passed to run_test.py --exclude).",
```
HereThereBeDragons: examples..

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+    )
+    parser.add_argument(
+        "--no-exclude-jit-executor",
+        action="store_true",
```
HereThereBeDragons: please use for all flags `action=argparse.BooleanOptionalAction`

with this comes that args cannot begin with `--no`, see https://github.com/python/cpython/issues/117941

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        default=False,
+        help="Do NOT pass --exclude-jit-executor (excluded by default).",
+    )
+    parser.add_argument(
+        "--no-exclude-distributed",
```
HereThereBeDragons: `action=argparse.BooleanOptionalAction`
and arg not starting with `--no`

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        action="store_true",
+        default=False,
+        help="Do NOT pass --exclude-distributed-tests (excluded by default).",
+    )
+    parser.add_argument(
+        "--no-exclude-quantization",
```
HereThereBeDragons: `action=argparse.BooleanOptionalAction`
and arg not starting with `--no`

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        "-k",
+        default="",
+        help="Override the pytest -k expression (bypasses TheRock skip-test generation).",
+    )
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        default=False,
+        help="Pass --dry-run to run_test.py to list tests without running them.",
+    )
```
HereThereBeDragons, line 261: can we modify the caching behavior of pytest from here?

we had:
```
parser.add_argument(
        "--cache",
        default=True,
        required=False,
        action=argparse.BooleanOptionalAction,
        help="""Enable pytest caching (default). Use --no-cache when only having read-only access to pytorch directory""",
    )
```

which was useful when using docker and mounting TheRock from the host but which still had the host-user ID and not the docker user ID. pytest would try to create a cache in that directory and obviously fail, so --no-cache would allow to not hit that issue

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+
+    cmd = build_run_test_cmd(args, tests_to_skip, passthrough_args)
+    print(f"Executing: {' '.join(cmd)}")
+
+    result = subprocess.run(cmd, cwd=str(args.pytorch_dir))
+    print(f"run_test.py finished with return code: {result.returncode}")
+    return result.returncode
+
+
+def force_exit_with_code(retcode: int) -> None:
```
HereThereBeDragons, line 367: please keep the comments in this function.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+        cmd.extend(["-k", tests_to_skip])
+
+    cmd.extend(passthrough_args)
+    return cmd
+
+
+def main(argv: list[str]) -> int:
+    args, passthrough_args = cmd_arguments(argv)
+
+    ((first_arch, _),) = set_gpu_execution_policy(args.amdgpu_family, policy="single")
```
HereThereBeDragons, line 334: pleas ekeep the comment about why we need to do `set_gpu_execution_policy` here

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+
+    if args.shard > 0 and args.shard > args.num_shards:
+        parser.error(
+            f"--shard ({args.shard}) cannot exceed --num-shards ({args.num_shards})."
+        )
+
+    return args, passthrough_args
+
+
+def build_run_test_cmd(
```
HereThereBeDragons, line 282: haha i know what you want to say with this.. but if you dont know the context that you want to construct the command this is just a very unspecific, multi-functional function name :-D

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -0,0 +1,303 @@
...
+    print("=== TheRock PyTorch Test Environment ===")
+    for var in THEROCK_ENV_VARS:
+        val = os.environ.get(var, "<not set>")
+        print(f"  {var}={val}")
+    print("=========================================")
+    sys.stdout.flush()
+
+
+def cmd_arguments(argv: list[str]) -> tuple[argparse.Namespace, list[str]]:
+    try:
```
HereThereBeDragons, line 148:
```suggestion
# Extract passthrough pytest args after "--"
    try:
```

ethanwee1: added a comment

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: updated

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: updated

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: added

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: Pytorch has three test configs (default, distributed, inductor)

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: we now have the "Runner health status" step using ./build_tools/health_status.py, and pip freeze removed.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: We want to set these because run_test.py recognizes them but does not set them

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Added a comment

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: added --cache/--no-cache

external-builds/pytorch/run_pytorch_tests_full.py
ScottTodd: Interesting... do we need the `linux-[OS]-` in there? For gfx110X, we test on Linux _and_ Windows

Should we also add "build environment"s for other gpu families, like gfx1151?

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -136,10 +173,16 @@ jobs:
             --index-subdir=${{ inputs.amdgpu_family }} \
             --activate-in-future-github-actions-steps
 
+      - name: Export test times for sharding
+        run: |
+          pushd external-builds/pytorch/pytorch
```
HereThereBeDragons: it would be good to add a comment why

`python external-builds/pytorch/pytorch/tools/stats/export_test_times.py`

does not work

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: Yes, the full linux-[OS]-rocm-py3.XX-XXXX string is required because these are lookup keys into PyTorch's [test-times.json ](https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/test-times.json)— they must match entries in that file exactly.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -129,10 +135,20 @@ jobs:
...
 
-      - name: Set git options
+      # safe.directory must be set before Runner Health Status
+      - name: Adjust git config
         run: |
+          git config --global --add safe.directory $PWD
+          git config fetch.parallel 10
           git config --global core.longpaths true
 
+      - name: Runner health status
```
HereThereBeDragons: sorry, i guess my explanation was not arcuate explicit enough:
`adjust git config` is at a good location but `runner health status` needs to be in a later step.

you want to move `runner health status` before `- name: Run rocm-sdk sanity tests`
so we also capture all the other python dependencies we install in `- name: Install test requirements`

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: We don't have gfx1151 testing in pytorch upstream so it will default to the mi300 test times until we do add 1151 runners upstream. Not sure when/if that is happening.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -6,6 +6,33 @@
     python test/run_test.py \\
         --exclude-jit-executor --exclude-distributed-tests \\
         --exclude-quantization-tests --shard N M --verbose
+
+Usage examples:
+
```
HereThereBeDragons, line 11: could we have one example with "run test XY for file test_nn"?

HereThereBeDragons (commented):
thanks! only 3 minor changes left from my side. otherwise lgtm

and then please also a new run of test_pytorch_wheels_full workflow :)

ethanwee1:
> thanks! only 3 minor changes left from my side. otherwise lgtm
> 
> and then please also a new run of test_pytorch_wheels_full workflow :)

Actually I've been updating the linked jobs at the top of the description with the job runs each time.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -83,9 +100,19 @@ jobs:
...
         shell: bash
+    # These env vars are set at the job level so they are available to all steps
+    # (e.g. export_test_times.py, run_pytorch_tests_full.py). The Python script
+    # uses os.environ.setdefault() so values set here take precedence.
     env:
       VENV_DIR: ${{ github.workspace }}/.venv
       AMDGPU_FAMILY: ${{ inputs.amdgpu_family }}
+      SHARD_NUMBER: ${{ matrix.shard }}
+      NUM_TEST_SHARDS: ${{ matrix.num_shards }}
+      CI: "1"
```
ScottTodd: This `CI` environment variable is already set by github actions: https://docs.github.com/en/actions/reference/workflows-and-actions/variables

Probably harmless to make it explicit, but I think the expected usage is that developers running the scripts get some behavior by default and CI gets other behavior by default.

.github/workflows/test_pytorch_wheels_full.yml
ScottTodd, line 1: I thought we decided to just create this new `_full` workflow and keep the existing workflow unchanged for this initial PR. This is duplicating a lot of workflow code. What's the actual diff?

If we want different triggers to run different sets of tests, it looks like we can do that through the `tests_to_include` input, without needing two workflows.

I'd prefer if this PR outlined the expected progression in the description:
1. Add new workflow that uses the new script and runs more tests
2. Run that workflow nightly / weekly, collect data
3. Once stable, migrate the existing workflow callers to the new workflow, delete the old workflow and old script

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -139,10 +169,23 @@ jobs:
...
+        run: |
+          pushd external-builds/pytorch/pytorch
+          PYTHONPATH=. python tools/stats/export_test_times.py
+          popd
+
       - name: Install test requirements
         run: |
           python -m pip install -r external-builds/pytorch/requirements-test.txt
-          pip freeze
+          python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
```
ScottTodd: Let's keep `pip freeze` to show the list of installed packages

external-builds/pytorch/run_pytorch_tests_full.py
ScottTodd: This information is worth adding to the comment block.

* What the "BUILD_ENVIRONMENT" is used for specifically
* Why having a non-exhaustive and platform-specific list is okay (today, it might not be in the future - that's what the comment helps with)

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: We did, but because I can't trigger the _full workflow to test, I updated the test_pytorch_wheels.yml file in the meantime to see the results. Once we are happy with the changes, I will revert the file back and the PR will only have the two new files in.

ScottTodd: Ok, that wasn't clear from reading the PR. In general, please try to keep draft/testing changes out of PRs marked "ready for review". You can create a parallel branch with those changes for testing but the expectation is that code that is reviewed/approved is code that is merged.

.github/workflows/test_pytorch_wheels.yml
ScottTodd: Bump. Keep the license here.

.github/workflows/test_pytorch_wheels.yml
HereThereBeDragons: this is not needed as we have the runner health status script just below that prints the python packages

PR #3707, opened 2026-03-02
Adding owner and repositories to generate PAT for hip-tagging-automation

ammallya:
## Motivation

Bot PAT is defaulting to ROCm/TheRock instead of rocm-systems


marbre:
@ammallya is there a way to avoid a PAT? We normally want to get rid where ever we can.

jayhawk-commits:
> @ammallya is there a way to avoid a PAT? We normally want to get rid where ever we can.

This is a github app token, and not a PAT.

PR #3728, opened 2026-03-03
RFC0011: Quartz: Central CI/CD Data Hub for the ROCm Ecosystem

HereThereBeDragons:
This RFC proposes Quartz, a central CI/CD data hub that collects TheRock build and test results into a database, distributes status notifications to downstream projects, accepts results reported back by downstream projects, and exposes all data via SQL for dashboard analytics.

### DIFF
--- a/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
+++ b/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
@@ -21,7 +21,7 @@ Quartz closes this gap. It is implemented as GitHub Actions workflows in a dedic
    — when new nightlies or prereleases are available, via push (workflow trigger) or pull (status.json polling)
    — when any other branch has finished the CI, via push (workflow trigger)
 1. **Accept results back from downstream projects** — build and test outcomes against TheRock artifacts, stored in the database
-1. **Power analytics dashboards** — Grafana queries database directly; no GitHub API calls at query time
+1. **Power analytics dashboards** — Dashboards query database directly; no GitHub API calls at query time
 1. **Stay within the GitHub ecosystem** — no external infrastructure required other than the database
 
 ## Non-Goals
@@ -57,9 +57,9 @@ Quartz sits between TheRock CI and the rest of the ROCm ecosystem. TheRock jobs 
                INSERT │     (GH App: Quartz Hunt or Quartz Kibble-<project>)
                       │
                       ▼
-            ┌──────────────────────────┐                  ┌──────────────────────────┐
-            │       Database           │◄─────────────────│  Dashboards (Grafana)    │
-            │  • therock_workflow_runs │    SQL queries   └──────────────────────────┘
+            ┌──────────────────────────┐                  ┌──────────────────────────────────────────┐
+            │       Database           │◄─────────────────│  Dashboards (Grafana, TheRock HUD, ...)  │
+            │  • therock_workflow_runs │    SQL queries   └──────────────────────────────────────────┘
             │  • therock_workflow_jobs │
             │  • downstream_*          │
             └──────────────────────────┘
@@ -247,7 +247,7 @@ Across all phases, a dedicated test repository mocks TheRock and downstream proj
 Create the `ROCm/quartz` repository and database, stand up the Quartz Hauly GitHub App, and implement the TheRock data ingest workflow. Scope is limited to nightly and prerelease workflows. Publish `status.json` artifacts so downstream projects can begin polling immediately.
 
 **Phase 2 — Subscription: Quartz → Downstream + Dashboards**
-Implement the Quartz Conveyor app and outbound notification workflow. Onboard the first downstream subscriber. Connect Grafana to ClickHouse Cloud for analytics dashboards.
+Implement the Quartz Conveyor app and outbound notification workflow. Onboard the first downstream subscriber. Connect the dashboards to use ClickHouse Cloud for analytics.
 
 **Phase 3 — Reporting Back: Downstream → Quartz**
 Define the downstream callback schema and implement the Quartz Hunt (Tier 1) and Quartz Kibble (Tier 2) ingest workflows. Provide onboarding templates for downstream projects. Covered in a follow-up — see Scope and Deferred Work.
@@ -317,11 +317,11 @@ When a downstream project PR triggers a TheRock CI run, notify that project auto
 - Validate full push notification flow end-to-end using the test repository mocking TheRock and a downstream subscriber
 - Onboard real subscribers (starting with rocm-examples) only after successful validation
 
-### C. Grafana Dashboards
+### C. Dashboards
 
-- Create read-only ClickHouse service account for Grafana
+- Create read-only ClickHouse service account for dashboards
 - Create ClickHouse materialized views for pre-aggregated dashboard queries (pass rates per architecture, nightly history, build duration trends)
-- Connect Grafana to ClickHouse using the read-only account
+- Connect dashboards to ClickHouse using the read-only account
 - Create new dashboards or update existing ones to use ClickHouse as the data backend, querying the materialized views
 
 ## Alternatives Considered
@@ -373,3 +373,4 @@ This RFC proposes Quartz, a central CI/CD data hub that gives the ROCm ecosystem
 - 2026-03-16: Add Phase 1 and Phase 2 implementation plans; update deduplication strategy to version integer with GitHub API timestamps; resolve dispatch mechanism to `workflow_dispatch`; various consistency fixes (Laura Promberger)
 - 2026-03-18: Add missing scripts to repository structure; add GitHub Issues and GitHub Project tracking note; minor fixes (Laura Promberger)
 - 2026-03-19: Refactor workflow permissions to read-all default with job-level elevation; move Scope and Deferred Work before Implementation Phases; rewrite Summary; swap Phase 3 (downstream reporting) and Phase 4 (all TheRock workflows); add `docs/` directory to repository structure (Laura Promberger)
+- 2026-03-31: Clarify dashboards are not Grafana-specific — any dashboard related the TheRock CI is in scope (Laura Promberger)


HereThereBeDragons:
one part that is missing i will add tomorrow is the handling of projects that dont want to expose their names in the subscriber.yml --> for this we can use github secrets.

docs/rfcs/RFC0010-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+            │       Database           │◄─────────────────│  Dashboards (Grafana)    │
+            │  • therock_workflow_runs │    SQL queries   └──────────────────────────┘
+            │  • therock_workflow_jobs │
+            │  • downstream_*          │
+            └──────────────────────────┘
+```
+
+Data flows:
+
+- **TheRock → Quartz:** Each CI job dispatches job data and parent workflow metadata via `workflow_dispatch` (Hauly app). A final workflow step (`if: always()`) dispatches a completion signal so Quartz can mark unfinished jobs as `timed_out`.
```
ScottTodd: Where do you expect the "final workflow step" will go? We have some pretty complex workflows that in practice can be cancelled, restarted, partially re-run, etc.

I wonder if finer-grained status updates (per-job instead of per-workflow?) might be useful too.

docs/rfcs/RFC0010-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Implementation Phases
+
+Quartz is delivered in four phases, each building on the previous.
+
+**Phase 1 — TheRock → Quartz data collection + status.json**
+Create the `ROCm/quartz` repository and database, stand up the Hauly GitHub App, and implement the TheRock data ingest workflow. Begin writing TheRock CI results to ClickHouse Cloud. Publish `status.json` artifacts per nightly and prerelease so downstream projects can begin polling immediately.
+
+**Phase 2 — Subscription: Quartz → Downstream + Dashboards**
+Implement the Conveyor app and outbound notification workflow. Onboard the first downstream subscriber. Connect Grafana to ClickHouse Cloud for analytics dashboards.
```
ScottTodd: I think for phase 1/2, your focus is just on _release_ builds (scheduled jobs that run once a day, NOT CI builds that run per-commit on `pull_request` and `push` triggers), so you could clarify "TheRock CI results" here.

I would like us to be able to extend this system with per-commit triggering and reporting eventually, so projects have a clear path from being tested only _after_ ROCm releases are built to being tested/test-able on individual commits while they are still in review. Maybe those projects should just not use Quartz?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Goals
+
+1. **Collect TheRock CI results into a structured database** — per job, per architecture, per branch, for any TheRock branch (nightly, prerelease, PR, ...)
+1. **Notify downstream projects automatically**
+   — when new nightlies or prereleases are available, via push (workflow trigger) or pull (status.json polling)
+   — when any other branch has finished the CI, via push (workflow trigger)
+1. **Accept results back from downstream projects** — build and test outcomes against TheRock artifacts, stored in the database
+1. **Power analytics dashboards** — Grafana queries database directly; no GitHub API calls at query time
+1. **Stay within the GitHub ecosystem** — no external infrastructure required other than the database
```
amd-shiraz, line 25: I like the “stay in GitHub” Goal.  Is consistent with the chosen approach (Actions-only processing and no long-running service).

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+- **Status:** Draft
+- **Discussion:** [TBD - Add GitHub issue/PR link]
+
+This RFC proposes Quartz, a central CI/CD data hub that collects TheRock build and test results into a database, distributes status notifications to downstream projects, accepts results reported back by downstream projects, and exposes all data via SQL for dashboard analytics.
+
+## Overview
+
+TheRock CI produces build and test results continuously, but there is no unified place to observe them, no mechanism for downstream projects to be automatically notified when a (nightly) build succeeds, and no way for downstream projects to report their test results back. Each project polls GitHub, scrapes logs, or relies on manual handoff.
+
+Quartz closes this gap. It is implemented as GitHub Actions workflows in a dedicated `ROCm/quartz` repository. All ingestion, validation, and notification logic is written in Python. As a backend it will use a database (choice: ClickHouse Cloud). Four GitHub Apps handle authenticated data transport between repositories.
```
amd-shiraz: choice of clickhouse cloud is also good imo. Good for high-frequency inserts. has out of the box aws module for provisioning using terraform.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+| **Conveyor**         | Quartz → Downstream        | Downstream projects subscribing to push notifications     |
+| **Hunt**             | Internal AMD/ROCm → Quartz | ROCm org projects (shared app, Tier 1)                    |
+| **Kibble-{project}** | External → Quartz          | External community projects (one app per project, Tier 2) |
+
+Tier 1 uses a single shared app across all ROCm-org projects — simpler onboarding, but a compromised credential affects all Tier 1 reporters. Tier 2 uses one app per external project, installed in that project's own org — narrower blast radius at the cost of more setup per project.
+
+**Authentication:** Every incoming `workflow_dispatch` to Quartz must pass two independent checks:
+
+1. **GitHub App token** — GitHub validates the token before accepting the dispatch; cannot be forged.
+1. **Allowlist match** — `github.event.installation.id` must match the expected App ID for that repository in `config/allow-list/<project>.yml`.
```
amd-shiraz: Wil be also good to pin the allowed dispatcher workflows by name and expected inputs
Don’t accept arbitrary inputs or “freeform JSON” without a strict schema boundary per direction). Make it explicit that the workflow should fail fast before any checkout, command executions, github app ephemeral token access etc.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+Quartz commits a `status.json` per nightly and per prerelease to the repository. Downstream projects may poll these files on a schedule:
+
+- `release-nightly/<date>/status.json`
+- `prerelease/<version>/status.json`
+
+`latest.json` and `latest_good.json` point to the most recent nightly and the most recent fully passing nightly respectively. No installation required — any project can poll these files.
+
+## Security Considerations
+
+### GitHub App Permissions and Blast Radius
```
amd-shiraz, line 206: Please explicitly add and make sure to constrain the repository permissions for the workflow token even if the app needs Actions: write, the workflow should run with:
permissions: contents: read (or none)
permissions: actions: read (unless we truly need more)
avoid contents: write except for the json flow

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+Apps installed on Quartz require `Actions: Write`, which unavoidably also permits disabling workflows, cancelling runs, and deleting run logs. There is no "dispatch-only" GitHub permission.
```
amd-shiraz, line 208: Add branch protection / CODEOWNERS and other similar gates we have on therock and therock-infra repo on .github/workflows/** and scripts/** in Quartz repo as part of safe guards.

docs/rfcs/RFC0010-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+`latest.json` and `latest_good.json` point to the most recent nightly and the most recent fully passing nightly respectively. No installation required — any project can poll these files.
+
+## Security Considerations
+
+### GitHub App Permissions and Blast Radius
+
+Apps installed on Quartz require `Actions: Write`, which unavoidably also permits disabling workflows, cancelling runs, and deleting run logs. There is no "dispatch-only" GitHub permission.
+
+Blast radius is limited to `ROCm/quartz` only — apps are installed on Quartz alone. A compromised Hunt (Tier 1) app can inject false data for all ROCm-org projects and disable ingestion workflows, but cannot affect any other repository.
```
amd-shiraz: good to explore as part of design to enforce that the dispatch comes from a specific repo+workflow pair (e.g., include source_repo_id, source_workflow_file, run_id, and verify via GitHub API that such a run exists and belongs to that repo). This uses GitHub as the source of truth and reduces “fake data injection” even with a compromised installation token.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Push (Conveyor)
+
+Downstream projects install the Conveyor GitHub App and declare their subscription in `config/subscriber.yml`. When a relevant TheRock status changes (e.g. a new nightly passes all checks), Quartz triggers a workflow in the downstream project.
+
+*Note: Whether Conveyor uses `workflow_dispatch` (targets a single named workflow) or `repository_dispatch` (triggers all listening workflows) is to be decided during Phase 2 design.*
+
+### Pull (status.json)
+
+Quartz commits a `status.json` per nightly and per prerelease to the repository. Downstream projects may poll these files on a schedule:
```
amd-shiraz, line 197: recommend having versioning for status.json files. e.g., schema_version field

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+## Database Design
+
+### Schema
+
+Inspired by the PyTorch HUD — all workflow runs and jobs are captured as individual rows, with a structured schema plus an `extra_info` JSON field for data that does not fit fixed columns.
+
+- `therock_workflow_runs` — one row per TheRock CI run
+- `therock_workflow_jobs` — one row per job within a run
+
+ClickHouse's ReplacingMergeTree engine is used for both tables: inserts are always appends, and deduplication happens in the background using `updated_at` as the version column (last write wins). Multiple job retries and status updates are safe without application-level locking.
```
amd-shiraz, line 169: pls add in RFC if already thought for but not mentioned here basically around how are we making sure updated_at is monitonic enough ? 
recommendations:
1) Keep updated_at from GitHub (or Quartz receive time). 
2) Call out clock-skew risk ? If runners have bad clocks, “newest wins” can be wrong. Prefer GitHub timestamps where possible.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Phase 3 — Reporting Back: Downstream → Quartz**
+Define the downstream callback schema and implement the Hunt (Tier 1) and Kibble (Tier 2) ingest workflows. Provide onboarding templates for downstream projects. Covered by a follow-up RFC — see Scope and Deferred Work.
+
+**Phase 4 — PR-Subscription**
+When a downstream project PR triggers a TheRock CI run, notify that project automatically on completion. Covered by a follow-up RFC — see Scope and Deferred Work.
+
+## Alternatives Considered
+
+### No central database (status.json only)
```
amd-shiraz, line 329: would recommend going with central database. considering other project needs and skill set within the team to maintain such a database at scale, we can set up once and use for multiple use cases.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,279 @@
...
+- All business logic must be in Python scripts, not in workflow YAML.
+
+### Mitigations
+
+- Pydantic schema validation on all incoming payloads before any database insert.
+- GitHub App rate limit (5,000 API calls/hour per installation) provides natural spam protection; no additional mechanism is available in GitHub Actions.
+- Workflow disable events should trigger an external alert. GitHub Actions cannot self-monitor a disabled workflow — an external webhook or watchdog in a separate repository is required.
+- Anomaly detection should flag impossible values (e.g. test duration of 1 second, non-existent ROCm versions).
+
+## Implementation Phases
```
amd-shiraz, line 240: I know we are not there yet but will be good to add soft milestones dates when ready to reflect project status as we make progress along the way. . Refer to https://amd.atlassian.net/wiki/spaces/SHARK/pages/1317959419/Cross+Cloud+Migration+Azure+to+AWS#Project-Milestones as example.

docs/rfcs/RFC0010-Quartz-CICD-Datahub.md
HereThereBeDragons: Ah I think this is a bit misleading here:
- Yes we will report to Quartz about every job update. This is goal 1)

> 1. **Collect TheRock CI results into a structured database** — per job, per architecture, per branch, for any TheRock branch (nightly, prerelease, PR, ...)

- Here in this context it talks about how in the database jobs that hang/time-out/dont report can be marked like this by having a final step of the entry job to say "all jobs part of me must have finished now, if not mark them as timed out in the database"

docs/rfcs/RFC0010-Quartz-CICD-Datahub.md
HereThereBeDragons: Yeah we can limit it first for releases. I would even start first only with nightly builds.

Phase 4 is about PRs etc. There your 2nd paragraph should be discussed to see what and how we want to support those individual per-commit triggers.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: you mean for the downstream --> quartz communication?
I will need to figure out how much extra info we can get from Github.
e.g. getting workflow ID, API URL, and workflow URL is only possible since 3 weeks or so: https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/

we will for sure only allow workflow_dispatch commands to communicate with quartz to have the param verification, authentication, etc.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: true, thanks. we could probably even do:

for therock -> quartz
```
permissions:
    contents: write   # needed to commit status.json
    actions: none
```

for quartz -> downstream and downstream -> quartz:
```
permissions:
    contents: read
    actions: none
```

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: already part of the draft for it :)

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: yes, this was the earliest iteration where the scope from quartz was smaller. central database is the choice now, also to not be dependent on what downstream does or doesnt do

HereThereBeDragons:
I created a discussion in case there are larger points to discuss:
https://github.com/ROCm/TheRock/discussions/3782

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
ScottTodd, line 1: Someone else grabbed RFC0010 in https://github.com/ROCm/TheRock/commit/0282a67f540e32f4112427f6742736dfa19736bd. Rename to RFC0011?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: i will create issues in the respective repos and accumulate them here: https://github.com/orgs/ROCm/projects/156  based on the implementation plan part of this rfc

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
marbre, line 19: > for any TheRock branch (nightly, prerelease, PR, ...)

Is this an opt-in or is it really intended to collect the status on every branch always?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,370 @@
...
+
+Blast radius is limited to `ROCm/quartz` only — apps are installed on Quartz alone. A compromised Quartz Hunt (Tier 1) app can inject false data for all ROCm-org projects and disable ingestion workflows, but cannot affect any other repository.
+
+### Development Rules
+
+- Internal Quartz processing workflows must not expose a `workflow_dispatch` trigger. Any workflow that does accept `workflow_dispatch` must enforce allowlist verification as its first step — every incoming dispatch must be treated as potentially malicious and carrying crafted payloads.
+- Every incoming `workflow_dispatch` workflow must verify
+  - the App ID against the allowlist as its first step.
+    = if possible: verify that the request comes from a actively running workflow (see newest [meta data available for workflows](https://github.blog/changelog/2026-02-19-workflow-dispatch-api-now-returns-run-ids/) )
+- Quartz workflows must have strict, minimal scope over what they can change in the repository and database. Every workflow must declare explicit `permissions:` to enforce this:
```
marbre: It might be worth to be more explicit here. Top-level should be `read-all` or `contents: read`. We can enforce this at the repo level and set workflow permissions to `Read repository contents and packages permissions` instead of `Read and write permissions`. Any additional, required permission can than be set at the job level.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,370 @@
...
+| `repository_dispatch`               | Rejected | Requires `contents:write` (broader than needed); triggers all listening workflows (event type based) in the repo rather than one specific workflow; no payload schema enforcement by GitHub |
+| `workflow_call` (reusable workflow) | Rejected | Cannot verify App ID, cross-repo call limitations                                                                                                                                           |
+| `workflow_dispatch`                 | Selected | Targets a specific workflow; requires only `actions:write`; inputs are declared in the workflow YAML and validated by GitHub; App ID verifiable via `github.event.installation.id`          |
+
+## Scope and Deferred Work
+
+This RFC covers the core architecture, authentication model, and the first two phases of the data flow: TheRock CI results flowing into Quartz (Phase 1) and Quartz notifying downstream projects (Phase 2).
```
marbre: I was actually wondering why the others phases are not covered here, well actually they are in https://github.com/ROCm/TheRock/blob/060234de037922d76072bf4665fc03c820ad9103/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md#implementation-phases, but why no further details are provided. Does it makes sense to move this up and let the reader know before diving into the implementation phases?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,370 @@
...
+
+The following are out of scope and will be addressed in a follow-up RFC once downstream project requirements are gathered:
```
marbre: Follow up or added later to this RFC (as it is at least partly covered already)?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
```diff
@@ -0,0 +1,370 @@
...
+
+- **Downstream reporting back (Phase 3):** The schema for downstream callback data, mandatory vs. optional fields, and the Quartz Hunt Tier 1 shared credential model (acceptable blast radius vs. per-project apps) will be defined once downstream projects have confirmed their reporting requirements.
+- **PR-subscription flow (Phase 4):** How Quartz identifies and notifies the originating downstream project when a PR-triggered TheRock CI run completes.
+
+## Summary
+
+Quartz provides the ROCm ecosystem with a unified CI/CD data hub: TheRock results flow in, downstream projects are notified, downstream results flow back, and dashboards query the database directly. The design stays within the GitHub ecosystem, uses ClickHouse Cloud for its high-frequency upsert and insert-triggered materialized view support, and secures all data transport with a pseudo-"2FA" GitHub App + allowlist model.
```
marbre: Duplicates the overview?

marbre (approved):
Just some minor NITs.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: i think this is a good point to have another discussion about.
in theory it was requested "everything" but that might not what we want to have, as we have workflows for:
- prereleases
- release nightly
- dev build triggered on schedule (nightly)
- dev build triggered by PR
- dev build triggered manually by user

we will start with prerelease and release nightly. moving then over to the dev nightly.
- for the 2nd last point we will need more decisions closer to phase 4
- for the last point i think is something we need to discuss about, how we could use it and what benefit there is to put it in the database

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: we dont have yet enough information about the downstream users usage pattern - so i would like to defer the details to a later point where we have parts of quartz already working and can get a more realistic idea of what is needed - and possible to provide.

i can move the "scope and deferred work" further up if that helps to reduce confusion.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
HereThereBeDragons: i thought rfcs are not to be changed once they are published? thats why i thought the need to have another rfc. 
if it is ok to extend this rfc, we can also do that.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
marbre: In that case eventually rephrase and replace `any TheRock branch` with `specific TheRock branches`?

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
marbre: It is perfectly fine to not have this written down yet. Moving this up would make it clear straight from the start that there is more to come.

docs/rfcs/RFC0011-Quartz-CICD-Datahub.md
marbre: Fair point. Should rather go to a follow-up RFC indeed but we're seem to be not as strict (so at least minor edits have been accepted so far).

marbre:
@HereThereBeDragons shouldn't this be merged? :)

HereThereBeDragons:
i am waiting for review from @amd-aakash - but if it takes too long maybe we just merge it anyway.

PR #3730, opened 2026-03-03
[ci] Adding gfx110X windows for presubmit and postsubmit

geomin12:
Currently, we have no windows signal during presubmit and postsubmit due to limited gfx1151 capacity.

As we have now 17+ gfx110X windows runners on Azure, we will re-enable this arch for signal!


geomin12:
windows gfx110X tests pass, landing

Issue #318, opened 2025-03-31
Expand support to CMake 4.0.0+


ScottTodd:
CMake released a major version 4.0.0 on March 28, 2025:

* Announcement: https://www.kitware.com/cmake-4-0-0-available-for-download/
* Release notes: https://cmake.org/cmake/help/v4.0/release/4.0.html

We should support this version across the full build of ROCm.

### Windows amd-llvm

Our Windows CI caught issues with amd-llvm using this new version: https://github.com/ROCm/TheRock/actions/runs/14182869714/job/39732671543.

```
[amd-llvm configure] CMake Error at C:/home/runner/_work/TheRock/TheRock/compiler/amd-llvm/amd/device-libs/cmake/OCL.cmake:12 (cmake_policy):
[amd-llvm configure]   Policy CMP0053 may not be set to OLD behavior because this version of CMake
[amd-llvm configure]   no longer supports it.  The policy was introduced in CMake version 3.1.0,
[amd-llvm configure]   and use of NEW behavior is now required.
[amd-llvm configure] 
[amd-llvm configure]   Please either update your CMakeLists.txt files to conform to the new
[amd-llvm configure]   behavior or use an older version of CMake that still supports the old
[amd-llvm configure]   behavior.  Run cmake --help-policy CMP0053 for more information.
```

That stems from https://cmake.org/cmake/help/latest/policy/CMP0053.html

> **Warning:** The `OLD` behavior of this policy was removed in CMake version 4.0. This policy must be set to `NEW` by a call to [`cmake_minimum_required()`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [`cmake_policy()`](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy).

and this code: https://github.com/ROCm/llvm-project/blob/82b836e284b8e24104789bdc1b35a710d65dfaea/amd/device-libs/cmake/OCL.cmake#L12

### rocprofiler-register

https://github.com/ROCm/TheRock/actions/runs/14182869714/job/39732671543#step:11:276

```
[rocprofiler-register configure] CMake Error at external/glog/cmake/GetCacheVariables.cmake:2 (cmake_policy):
[rocprofiler-register configure]   Compatibility with CMake < 3.5 has been removed from CMake.
[rocprofiler-register configure] 
[rocprofiler-register configure]   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
[...]

PR #3735, opened 2026-03-03
Add Windows environment validation script and update setup docs

### LINKED ISSUE #3733: [Documentation]: There seems to be a GUI setting for long paths now
### Description of errors

I was having trouble with the usual registry edit ways of enabling long paths (security policy). Then I noticed that there is now a toggle for it, at least in the version of Windows I have. This may or may not have only become available after I enabled developer mode.

### Attach any links, screenshots, or additional evidence you think will be helpful.

Image

stellaraccident:
- Add build_tools/validate_windows_install.ps1: checklist-style validator covering RAM/disk, long paths (registry + functional test), symlinks, MSVC/ATL, CMake (<4.0.0), Ninja, Git, bash, patch, DVC, Python (venv + pip dry-run for requirements), gfortran/Strawberry Perl, ccache, pkg-config, and git config settings
- Update docs/development/windows_support.md: add validation script reference at the top of the prerequisites section; expand long path instructions to include registry command, GUI toggle (Settings > System
  > For developers > Enable Long Paths, requires Developer Mode + reboot),
  and reliability caveats

Fixes #3733

### DIFF
--- a/build_tools/validate_windows_install.ps1
+++ b/build_tools/validate_windows_install.ps1
@@ -278,6 +278,56 @@ if ($env:VCINSTALLDIR) {
     Write-Info "Skipping ATL check (MSVC environment not activated)"
 }
 
+# rc.exe (Windows Resource Compiler - lives in SDK bin, not MSVC bin)
+$rcCmd = Get-Command rc.exe -ErrorAction SilentlyContinue
+if ($rcCmd) {
+    Write-Pass "rc.exe in PATH: $($rcCmd.Source)"
+} else {
+    Write-Fail "rc.exe not found in PATH (Windows SDK bin directory missing)" `
+        -Detail "The VS dev shell may not have set up the SDK paths. Check that the Windows SDK is installed and its bin\x64 dir is on PATH."
+}
+
+# Test compile: exercises headers from each SDK include directory and links
+# kernel32.lib.  This is the most reliable way to detect a broken VS dev
+# shell where cl.exe is on PATH but SDK include/lib dirs are not configured.
+#   windows.h  → Include\<ver>\um
+#   stdio.h    → Include\<ver>\ucrt
+#   rpcndr.h   → Include\<ver>\shared
+#   winstring.h→ Include\<ver>\winrt
+#   Link kernel32.lib → Lib\<ver>\um\x64 + Lib\<ver>\ucrt\x64
+if ($clCmd) {
+    $testDir = Join-Path $env:TEMP "therock_compile_test_$([System.Guid]::NewGuid().ToString('N'))"
+    $testSrc = Join-Path $testDir "test.c"
+    $testExe = Join-Path $testDir "test.exe"
+    try {
+        New-Item -ItemType Directory -Path $testDir -Force -ErrorAction Stop | Out-Null
+        Set-Content -Path $testSrc -Value @"
+#include <windows.h>   /* um */
+#include <stdio.h>     /* ucrt */
+#include <rpcndr.h>    /* shared */
+#include <winstring.h> /* winrt */
+int main(void) {
+    HANDLE h = GetProcessHeap();
+    printf("ok %p\n", (void*)h);
+    return 0;
+}
+"@ -ErrorAction Stop
+        $compileOut = (& cl.exe /nologo /Fo:"$testDir\" /Fe:"$testExe" "$testSrc" /link kernel32.lib 2>&1) -join "`n"
+        if ((Test-Path $testExe)) {
+            Write-Pass "Test compile + link succeeded (um, ucrt, shared, winrt headers; kernel32.lib)"
+        } else {
+            Write-Fail "Test compile failed - SDK environment is incomplete" `
+                -Detail $compileOut.Substring(0, [Math]::Min(300, $compileOut.Length))
+        }
+    } catch {
+        Write-Fail "Test compile could not run: $($_.Exception.Message)"
+    } finally {
+        Remove-Item $testDir -Recurse -Force -ErrorAction SilentlyContinue
+    }
+} else {
+    Write-Info "Skipping test-compile check (cl.exe not available)"
+}
+
 # ============================================================================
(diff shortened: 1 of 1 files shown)


build_tools/validate_windows_install.ps1
```diff
@@ -0,0 +1,575 @@
+#!/usr/bin/env pwsh
+<#
+.SYNOPSIS
+    Validates the Windows development environment for building TheRock/ROCm from source.
+
+.DESCRIPTION
+    Checks system configuration, required tools, and settings needed for a
+    successful TheRock source build on Windows. Outputs a checklist-style
+    report with PASS/WARN/FAIL for each item.
```
ScottTodd, line 9: We already have https://github.com/ROCm/TheRock/blob/main/build_tools/hack/diagnose.py and https://github.com/ROCm/TheRock/tree/main/build_tools/hack/env_check, and those are already in use on our CI runners. There's significant overlap here, though this file is more self-contained.

stellaraccident: Those don't work without a maze of subprocess errors and python warnings about relative imports. I'd prefer one user level script like I have here.

build_tools/validate_windows_install.ps1
```diff
@@ -0,0 +1,575 @@
...
+
+# CMake - must be < 4.0.0 per issue #318
+$cmakeCmd = Get-Command cmake -ErrorAction SilentlyContinue
+if ($cmakeCmd) {
+    $cmakeVerStr = ((& cmake --version 2>&1 | Select-Object -First 1) -replace "cmake version ", "").Trim()
+    try {
+        $cmakeVer = [Version]$cmakeVerStr
+        if ($cmakeVer -ge [Version]"4.0") {
+            Write-Fail "CMake $cmakeVerStr is too new (< 4.0.0 required, see issue #318)" `
+                -Detail "Downgrade: winget install cmake -v 3.31.0"
```
ScottTodd: That issue (https://github.com/ROCm/TheRock/issues/318) was fixed. We can trim this back down to just check against the minimum supported version instead, and update the docs.

build_tools/validate_windows_install.ps1
ScottTodd: Here's output from my system for both tools: https://gist.github.com/ScottTodd/2dcffd8fdbe9f4be6bb0c94fdf687662

I like the formatting of the new tool, and it looks AI-agent-friendly for making edits (self-contained file, not overly generalized). As a human I tend to lean more towards the other code since it's Python and I can't stand powershell syntax. This code is pretty much write-once though, so I guess this is fine.

I'm fine with keeping both scripts for now, but we could consolidate in the future (especially when we update supported versions of tools and find ourselves making edits in both places).

stellaraccident: My prompt was basically, "read the windows setup doc and write me a script that verifies it" then three or four back and forths as it found my environment was not set up right (and I had to reboot three times). I like straight-line, no abstracted code like this for these things. Also, since a part of what it needs to validate is actually the python install, not coding the script in python has some merit.

stellaraccident:
@ScottTodd I ran across a novel failure in the VS install today where the powershell developer profiler only *half initializes* the environment, leaving off the SDK tools dir (for rc.exe) and the SDK include/lib dirs. This cost me a lot of frustration, so I added a test compilation step to the script that verifies the failure (some of which only surfaced deep in the build).

ScottTodd:
@jeffdaily ran into the same issue with `rc.exe` today x_x. Thanks for improving the docs + env test scripts.

stellaraccident:
> @jeffdaily ran into the same issue with `rc.exe` today x_x. Thanks for improving the docs + env test scripts.

I can never figure out what is going wrong with those awful MSFT scripts... I just had Claude vibe code what was needed. Might vary from system to system, but we're at the level of just self serving vs trying to fix broken old VS stuff being easier. Here was what "fixenv.ps1" script turned out to be:

```
# fixenv.ps1 â€” Fix missing Windows SDK environment for Developer PowerShell.
# Usage: . D:\develop\fixenv.ps1

$sdkRoot = "C:\Program Files (x86)\Windows Kits\10"
$sdkVer  = "10.0.26100.0"

if (-not (Test-Path "$sdkRoot\bin\$sdkVer\x64")) {
    Write-Error "Windows SDK $sdkVer not found under $sdkRoot"
    return
}

function Add-EnvPath([string]$VarName, [string[]]$Dirs) {
    $current = [Environment]::GetEnvironmentVariable($VarName, 'Process')
    $parts = if ($current) { $current -split ';' } else { @() }
    foreach ($d in $Dirs) {
        if ($parts -notcontains $d) {
            $parts = @($d) + $parts
            Write-Host "  $VarName += $d" -ForegroundColor Green
        }
    }
    [Environment]::SetEnvironmentVariable($VarName, ($parts -join ';'), 'Process')
}

Write-Host "Fixing Windows SDK env ($sdkVer)..." -ForegroundColor Cyan

# rc.exe, mt.exe, etc.
Add-EnvPath 'PATH' @("$sdkRoot\bin\$sdkVer\x64")

# kernel32.lib (um), ucrt.lib (ucrt)
Add-EnvPath 'LIB' @(
    "$sdkRoot\Lib\$sdkVer\um\x64"
    "$sdkRoot\Lib\$sdkVer\ucrt\x64"
)

# SDK headers
Add-EnvPath 'INCLUDE' @(
    "$sdkRoot\Include\$sdkVer\um"
    "$sdkRoot\Include\$sdkVer\ucrt"
    "$sdkRoot\Include\$sdkVer\shared"
    "$sdkRoot\Include\$sdkVer\winrt"
    "$sdkRoot\Include\$sdkVer\cppwinrt"
)

Write-Host "Done." -ForegroundColor Cyan
```

(I first ran vcvars32.bat in 1998. There is no way that this junk was meant to last for nearly 30 years, basically unchanged. It is just rotting and needs to be replaced)

PR #3743, opened 2026-03-04
Update codeowners to include debug tools ownership

lumachad:
We have a ROCgdb cmake wrapper in debug-tools/rocgdb, so it would be nice to include the debugger folks to review those changes alongside TheRock's folks.

While at it, debugger folks can also help with the recipes and artifacts files in debug-tools/*, as well as the testing scripts in build_tools/github_actions/test_executable_scripts/.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,3 +11,7 @@
 # Debug tools configs and ROCgdb cmake wrapper.
 /debug-tools            @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
 /debig-tools/rocgdb     @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
+
+# Debug tools testing scripts.
+/build_tools/github_actions/test_executable_scripts/test_rocgdb.py           @ROCm-DevTools-Debugger
+/build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py @ROCm-DevTools-Debugger


lumachad:
Now that I've put this through, I'm wondering if I should also add debugger reviewers to:
* build_tools/github_actions/test_executable_scripts/test_rocgdb.py
* build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py

.github/CODEOWNERS
```diff
@@ -7,3 +7,7 @@
 
 # New patches are strongly discouraged. See patches/README.md.
 /patches/               @ScottTodd @marbre
+
+# Debug tools configs and ROCgdb cmake wrapper.
+/debug-tools            @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
+/debig-tools/rocgdb     @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
```
marbre, line 13: Not sure it must be `ROCm/TheRock-infra-write`. @ScottTodd wdyt?

marbre:
> Now that I've put this through, I'm wondering if I should also add debugger reviewers to:
> 
> * build_tools/github_actions/test_executable_scripts/test_rocgdb.py
> * build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py

Probably? If you and peers what to take ownership here and feel responsible that is highly appreciated :)

lumachad:
> > Now that I've put this through, I'm wondering if I should also add debugger reviewers to:
> > 
> > * build_tools/github_actions/test_executable_scripts/test_rocgdb.py
> > * build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py
> 
> Probably? If you and peers what to take ownership here and feel responsible that is highly appreciated :)

Makes sense. I'll put a commit on top of this one adding those.

.github/CODEOWNERS
ScottTodd: Some background: we're currently using CODEOWNERS to _require approval_ from at least one listed member when a path is modified. 

Image

Another use of CODEOWNERS in some other projects is to be _notified for changes_. There isn't really a good way to mix both uses of CODEOWNERS. If you use it for one or the other, you need a separate system (like a code review bot) that handles the other use.

So historically we've added entries to this file when someone outside of the list (inadvertently) broke something, prompting tighter review policies. I suspect your requirements for the debug tools are a bit looser (or at least proactive instead of reactive).

With that being said, I could go either way on including `ROCm/TheRock-infra-write` here.
* Do you want that team to automatically have review requests for all changes here?
* Note that `ROCm/TheRock-infra-write` still does not include `ROCm/TheRock-infra-admin` as a subteam...
* ...but, those of us in `ROCm/TheRock-infra-admin` can bypass reviews as needed anyways.

lumachad: Thanks for clarifying it. Given these debug-tools/** files live in TheRock, I think it makes sense to also have TheRock members approval here, as things could potentially break in the recipe files, for instance. For the testing scripts, I think we're good with just the debug-tools folks.

lumachad: Thinking further about this, if a group has overriding/bypassing powers to revert broken stuff, we might not need to have @ROCm/TheRock-infra-admin for debug-tools/. That might reduce the load on infra-admin a bit (review-wise) with @ROCm-DevTools-Debugger the sole responsible for these files.

marbre (approved):
LGTM

.github/CODEOWNERS
```diff
@@ -7,3 +7,11 @@
 
 # New patches are strongly discouraged. See patches/README.md.
 /patches/               @ScottTodd @marbre
+
+# Debug tools configs and ROCgdb cmake wrapper.
+/debug-tools            @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
```
ScottTodd, line 12: The https://github.com/ROCm/TheRock/blob/main/.github/CODEOWNERS file has errors now

```
Unknown owner on line 12: make sure @ROCm-DevTools-Debugger exists and has write access to the repository

…    @ROCm/TheRock-infra-write @ROCm-DevTools-Debugger
```

We would need to grant that team write access for it to get review requests... which _might_ be fine today (write access is already the default).

lumachad: Should I get this one reverted while we do that step first?

ScottTodd: No, that's okay. The errors aren't that impactful. Other codeowner assignments will still work as expected, just that group won't get reviews requested until we add it.

geomin12, line 12:
```suggestion
/debug-tools            @ROCm/TheRock-infra-write @ROCm/ROCm-DevTools-Debugger
```

PR #3745, opened 2026-03-04
Enable XNACK for hip-tests when we run with ASAN

cjatin:
## Motivation

Ideally this should be done from `TheRock` itself, when user passes ASAN preset in config, TheRock automatically appends the `xnack+` based on target (if it supports or not). When it does that, it makes sense to TheRock to drive the XNACK env variable as well.
For now, we make sure XNACK is enabled when we run hip-tests.

## Technical Details

If we run hip-tests build with xnack on systems with xnack disable, we will see lots of errors. If we are running ASAN builds, we make sure we enable xnack.

## Test Plan

Running ASAN builds locally to verify.

## Test Result

Script seems to run ok.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_hiptests.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hiptests.py
@@ -110,6 +110,7 @@ def setup_env(env):
         # For ASAN mode, we preload it for test count query and test running
         if is_asan():
             env["LD_PRELOAD"] = get_asan_lib_path()
+            env["HSA_XNACK"] = "1"
             # TODO: enable this when we have symbolizer patch in
             # env["ASAN_SYMBOLIZER_PATH"] = str(Path(THEROCK_BIN_DIR).parent / "lib" / "llvm" / "bin" / "llvm-symbolizer")
     else:


build_tools/github_actions/test_executable_scripts/test_hiptests.py
```diff
@@ -110,6 +110,7 @@ def setup_env(env):
         # For ASAN mode, we preload it for test count query and test running
         if is_asan():
             env["LD_PRELOAD"] = get_asan_lib_path()
+            env["HSA_XNACK"] = "1"
```
geomin12, line 113: please add a comment here explaining what this is

cjatin: Hey Geo,
Do not think it requires a comment. The operation is self explanatory, we set a env variable.
Regarding what the flag is, its explained here in a better way.

https://github.com/ROCm/TheRock/blob/16aa2411c5d1fa3cf756d6670b6935a85ae81129/docs/development/sanitizers.md?plain=1#L7

https://rocm.docs.amd.com/projects/HIP/en/latest/how-to/hip_runtime_api/memory_management/unified_memory.html#memory-allocation-approaches-in-unified-memory

other projects also had similar approach
https://github.com/ROCm/TheRock/blob/16aa2411c5d1fa3cf756d6670b6935a85ae81129/build_tools/github_actions/test_executable_scripts/test_rocblas.py#L27-L29

geomin12: ah okay, that documentation helps! thanks

PR #3752, opened 2026-03-04
Add ASAN_OPTIONS and HSA_XNACK environment variables to test workflow

radhaksri:
## Detail

For ASAN tests, the environment variables HSA_XNACK and ASAN_OPTIONS must be set for all test execution runs. This PR sets these environment variables for all components

## Test Plan

These have been tested thoroughly in the ASAN enablement based on Nightly build 711, for e.g., like here: https://github.com/ROCm/TheRock/actions/runs/22311510308

## Test Result

Test executions have been confirmed to take these environment variables into account. Refer to the results from the test runs in the link above.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -145,7 +145,8 @@ jobs:
           TEST_TYPE: ${{ fromJSON(inputs.component).test_type }}
           ROCM_KPACK_DEBUG: "1"
           ASAN_OPTIONS: "detect_odr_violation=0:quarantine_size_mb=600"
-          HSA_XNACK: 1          
+          HSA_XNACK: 1
+          ASAN_SYMBOLIZER_PATH: ${{ env.VENV_DIR }}/llvm/bin/llvm-symbolizer
         run: |
           ${{ fromJSON(inputs.component).test_script }}
 


.github/workflows/test_component.yml
```diff
@@ -144,6 +144,8 @@ jobs:
           TOTAL_SHARDS: ${{ fromJSON(inputs.component).total_shards }}
           TEST_TYPE: ${{ fromJSON(inputs.component).test_type }}
           ROCM_KPACK_DEBUG: "1"
+          ASAN_OPTIONS: "detect_odr_violation=0:quarantine_size_mb=600"
```
geomin12, line 147: does each component tests automatically pick this up? how is this picked up?

.github/workflows/test_component.yml
```diff
@@ -144,6 +144,8 @@ jobs:
...
+          HSA_XNACK: 1
```
geomin12: this cannot be not enable here, we should follow something like this: https://github.com/ROCm/TheRock/commit/bc0fec77d06ea39f65590d65608af0c6768c722a

this change will affect all tests across ROCm, and we may get unexpected/failing results. let's copy similar as above

.github/workflows/test_component.yml
geomin12, line 147: alternatively, we need to add a conditional here, where if "asan", we add these options, otherwise, do not add

.github/workflows/test_component.yml
geomin12: personal reminder: this only applies to Instinct machines, does not work for consumer

geomin12:
Closing in lieu of #3756

Issue #3755, opened 2026-03-04
[ci] Make sanitizer environment variables cleaner once flags are finalized


geomin12:
As ASAN tests are still being configured and tested, we are playing around with flags. Once flags are finalized and required for ASAN tests, we can add a cleaner environment variable step to enable these flags

Specific flags like HSA_XNACK cannot be enabled for consumer GPUs

PR #3756, opened 2026-03-04
[ci] Adding sanitizer specific test flags

geomin12:
As experimentation continues for ASAN, we are enabling these sanitizer test flags

I have issue #3755 to clean this up once the experimentation / ASAN tests are passing

- Regular CI running and skipping step: https://github.com/ROCm/TheRock/actions/runs/22683289975/job/65760585007
- ASAN CI running and picking up env variables correctly: https://github.com/ROCm/TheRock/actions/runs/22683307018/job/65762049349#step:12:21

As this is test specific and local tests pass, I am adding `skip-ci` label

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -137,6 +137,14 @@ jobs:
           python ./build_tools/install_additional_requirements.py \
             --requirements-files=${{ join(fromJSON(inputs.component).additional_requirements_files, ',') }}
 
+        # TODO(#3755): Clean this up after final sanitizer environment variable flags are determined
+      - name: Enable sanitizer-specific test flags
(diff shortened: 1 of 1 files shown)


idass1990 (approved):
LGTM

.github/workflows/test_component.yml
jayhawk-commits, line 155: Why not put those new entries into this env block?

geomin12: i considered this, but the `test_component.yml` impact ALL of ROCm testing, so we aren't sure of the behavior of tests when we enable those additional options

So for now, we are keeping it specific to ASAN if asan is triggered, otherwise, normal flow

jayhawk-commits: Ah, fair, these env do need an if check for ASAN.

Issue #3701, opened 2026-03-02
[Issue] Linux Pytorch Build failure due to missing argument `--target_gpus` to `generate.py` - gfx906, gfx908


chiranjeevipattigidi:
**Platform:** `Linux`
**Pytorch Versions:** `2.8, 2.9, 2.10, nightly`
**Python Versions:** `3.10, 3.11, 3.12, 3.13`
**Effected Archs:** `gfx906, gfx908`

### Runs:
- https://github.com/ROCm/TheRock/actions/runs/22564835731
- https://github.com/ROCm/TheRock/actions/runs/22570332110

```
2026-03-02T07:03:33.0484685Z -- AOTRITON_OVERRIDE_TARGET_GPUS 
2026-03-02T07:03:33.0484986Z -- EFFECTIVE_TARGET_GPUS 
2026-03-02T07:03:34.1008335Z AOTRITON_COMPILER /__w/TheRock/TheRock/external-builds/pytorch/pytorch/build/aotriton/src/aotriton_runtime/v3python/compile.py
2026-03-02T07:03:34.7809277Z usage: generate.py [-h]
2026-03-02T07:03:34.7810704Z                    [--target_gpus {gfx90a_mod0,gfx942_mod0,gfx950_mod0,gfx1100_mod0,gfx1101_mod0,gfx1102_mod0,gfx1151_mod0,gfx1150_mod0,gfx1201_mod0,gfx1200_mod0,gfx1250_mod0} [{gfx90a_mod0,gfx942_mod0,gfx950_mod0,gfx1100_mod0,gfx1101_mod0,gfx1102_mod0,gfx1151_mod0,gfx1150_mod0,gfx1201_mod0,gfx1200_mod0,gfx1250_mod0} ...]]
2026-03-02T07:03:34.7812393Z                    [--build_dir BUILD_DIR] [--root_dir ROOT_DIR]
2026-03-02T07:03:34.7812836Z                    [--archive_only] [--library_suffix LIBRARY_SUFFIX]
2026-03-02T07:03:34.7813237Z                    [--noimage_mode] [--build_for_tuning]
2026-03-02T07:03:34.7813591Z                    [--build_for_tuning_second_pass]
2026-03-02T07:03:34.7814069Z                    [--build_for_tuning_but_skip_kernel [BUILD_FOR_TUNING_BUT_SKIP_KERNEL ...]]
2026-03-02T07:03:34.7814562Z                    [--verbose] [--lut_sanity_check]
[...]

PR #3759, opened 2026-03-04
Disable AOTriton for gfx906 and gfx908 in PyTorch builds

araravik-psd:
This PR adds gfx906 and gfx908 to AOTRITON_UNSUPPORTED_ARCHS in the PyTorch build script.

AOTriton is not currently supported on these architectures, which can lead to build failures when attempting to enable Flash Attention. Disabling it ensures PyTorch wheels build successfully for these targets.

This aligns the build configuration with the current AOTriton architecture support.

### DIFF
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -265,6 +265,7 @@ def get_rocm_path(path_name: str) -> Path:
         ).strip()
     )
 
+
 def get_rocm_init_contents(args: argparse.Namespace):
     """Gets the contents of the _rocm_init.py file to add to the build."""
     sdk_version = get_rocm_sdk_version()
@@ -281,6 +282,7 @@ def get_rocm_init_contents(args: argparse.Namespace):
                 check_version='{sdk_version}')
         """
     )
+
 
 def remove_dir_if_exists(dir: Path):
     if dir.exists():


araravik-psd:
Test run above changes:

https://github.com/ROCm/TheRock/actions/runs/22685240099/job/65766302927

Apex issues are present in pytorch 2.9, which were discussed earlier today morning

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -784,7 +794,7 @@ def do_build_pytorch(
 
     # aotriton is not supported on certain architectures yet.
     # gfx101X/gfx103X: https://github.com/ROCm/TheRock/issues/1925
```
jayhawk-commits: Please add to this existing ticket or create a separate one for gfx906 and gfx908, and then update this comment.

araravik-psd: Thanks @jayhawk-commits I have updated the case and the comment.

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -272,7 +272,17 @@ def get_rocm_init_contents(args: argparse.Namespace):
...
+
+    # Dynamically filter based on what this target actually provides
+    try:
+        import rocm_sdk._dist_info as _di
+
+        available = [s for s in library_preloads if s in _di.ALL_LIBRARIES]
+    except Exception:
+        available = library_preloads  # fallback
+
+    library_preloads_formatted = ", ".join(f"'{s}'" for s in available)
```
ScottTodd: This is risky and I don't think it's the right approach here.

* What libraries are not being provided for which target families? Where is the code for that?
* Does PyTorch handle libraries being only available for certain GPU targets?

I'd rather we just disable pytorch release builds for these targets instead of get into this support grey area that could break in unexpected/silent ways later.

araravik-psd: Hi @ScottTodd Thanks for calling this out — that’s a completely fair point.

What we’re running into is that some libraries (like hipdnn) aren’t actually included in the rocm[libraries] package for certain targets such as gfx906. PyTorch assumes all preload libraries are present, so when one isn’t, initialization fails immediately.

The filtering was just meant to avoid that hard failure by only preloading what’s actually packaged for the given target. It wasn’t intended to broaden support silently — just to prevent an avoidable crash during init.

That said, I understand the concern about drifting into a grey area of partial support. If gfx906 (or similar targets) aren’t meant to be supported for PyTorch release builds, I’m completely fine with explicitly disabling those builds instead.

Happy to adjust based on what direction we’d prefer here.

ScottTodd: As discussed on https://github.com/ROCm/TheRock/issues/3701 and in meetings this morning, the simple path forward to get our release builds back to green is to mark the targets as failing to build for pytorch via the `expect_pytorch_failure` setting in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py.

gfx906 in particular excludes many projects: https://github.com/ROCm/TheRock/blob/0b76f11707b8262742ce69f0de19cb175e3f3f4c/cmake/therock_amdgpu_targets.cmake#L47-L55

Just skipping the preloads for those projects is unlikely to produce functional pytorch packages. We're beyond the point where just iterating on github actions workflows is fruitful. Need to build locally, test on actual hardware, and see what is and isn't working -- and then add support to those projects where needed. None of that is directly on our team to do, we'll need to involve the math library component teams and possibly the pytorch framework team.

ScottTodd: For some more context on "this is risky", the library preload list _for pytorch_ is the list of libraries that we know/expect pytorch to depend on. It's not an exhaustive list of all libraries included in ROCm and other projects may have different lists. Removing libraries from the preload list because they aren't available for a given configuration has the potential to lead to errors at runtime. It _may_ be okay for a limited set of targets and libraries, if we know that pytorch does not use those libraries for those targets, but we'd need some more detailed code than what was first proposed here for that.

araravik-psd: Thanks @ScottTodd , I updated the PR to verify disable of only the pytorch build using "expect_pytorch_failure" attribute works as suggested and works as expected. Pytorch builds are skipped

https://github.com/ROCm/TheRock/actions/runs/22701252338

araravik-psd: Test runs:

gfx908: [Release portable Linux PyTorch Wheels (gfx908, dev, 7.12.0a20260305) · ROCm/TheRock@ca03e08](https://github.com/ROCm/TheRock/actions/runs/22736967525/job/65944524965)
 
gfx906: [Release portable Linux PyTorch Wheels (gfx906, dev, 7.12.0a20260305) · ROCm/TheRock@ca03e08](https://github.com/ROCm/TheRock/actions/runs/22736944779)

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -783,8 +783,8 @@ def do_build_pytorch(
     is_pytorch_2_11_or_later = pytorch_build_version_parsed.release[:2] >= (2, 11)
 
     # aotriton is not supported on certain architectures yet.
-    # gfx101X/gfx103X: https://github.com/ROCm/TheRock/issues/1925
-    AOTRITON_UNSUPPORTED_ARCHS = ["gfx101", "gfx103"]
+    # gfx101X/gfx103X/gfx906/gfx908: https://github.com/ROCm/TheRock/issues/1925
+    AOTRITON_UNSUPPORTED_ARCHS = ["gfx101", "gfx103", "gfx906", "gfx908"]
```
ScottTodd: nit: could sort 90X before 101X and 103X (yeah that's confusing , since the 101 and 103 are prefixes for 4 digit numbers and the 90X are 3 digit numbers :P)

araravik-psd: sure done, updated the ordering

ScottTodd (approved):

PR #3761, opened 2026-03-04
feat: add a passthrough for PROJECT_ENABLE_COVERAGE flag

### LINKED ISSUE #6572: Enable Code Coverage Flag Passthrough from TheRock to Subprojects
## Summary

TheRock superbuild needs the ability to pass project-specific code coverage flags (e.g., `-DHIPDNN_ENABLE_COVERAGE=ON`) through to individual component subprojects. This enables centralized control of code coverage builds for components like hipDNN, rocBLAS, etc., from the top-level TheRock CMake configuration.

## Background

ROCm component libraries in the [rocm-libraries](https://github.com/ROCm/rocm-libraries) repository are adding code coverage support (see [ROCm/rocm-libraries#8656](https://github.com/ROCm/rocm-libraries/pull/8656)). These libraries use a standardized naming convention for their coverage flags:

```cmake
-D<PROJECT_NAME>_ENABLE_COVERAGE=ON
```

For example:
- `HIPDNN_ENABLE_COVERAGE` for hipDNN
- `ROCBLAS_ENABLE_COVERAGE` for rocBLAS
- `ROCFFT_ENABLE_COVERAGE` for rocFFT

## Problem

TheRock's subproject system (`therock_cmake_subproject_declare` / `therock_cmake_subproject_activate`) did not have a mechanism to pass these project-specific flags through to the underlying component CMake configurations. When a user specified:

```bash
cmake -B build -DHIPDNN_ENABLE_COVERAGE=ON
```

This flag would be set in TheRock's top-level configuration but **not propagated** to the hipDNN subproject's CMake invocation, rendering it ineffective.

## Solution

Add an automatic passthrough of `<PROJECT_NAME>_ENABLE_COVERAGE` flags in `cmake/therock_subproject.cmake`:

1. **Uppercase normalization**: The logical target name (e.g., `hipDNN`) is converted to uppercase (`HIPDNN`) to match CMake's standard project name convention
2. **Conditional passthrough**: If `HIPDNN_ENABLE_COVERAGE` is defined at the TheRock level, it is automatically added as a CMake argument to the hipDNN subproject's configure command
3. **Per-project control**: Each component can independently enable/disable coverage without affecting others

### Example Usage

```bash
# Enable code coverage for hipDNN only
cmake -B build -GNinja \
  -DTHEROCK_AMDGPU_FAMILIES=gfx1100 \
  -DHIPDNN_ENABLE_COVERAGE=ON

# Enable code coverage for multiple components
cmake -B build -GNinja \
  -DTHEROCK_AMDGPU_FAMILIES=gfx1100 \
  -DHIPDNN_ENABLE_COVERAGE=ON \
  -DROCBLAS_ENABLE_COVERAGE=ON
```

### Implementation Details

The passthrough logic in `therock_cmake_subproject_activate()`:

```cmake
# Passthrough -D<PROJECT_NAME>_ENABLE_COVERAGE=ON to the subproject
# Convert logical target name to uppercase (e.g., hipDNN -> HIPDNN_ENABLE_COVERAGE)
set(_coverage_arg)
string(TOUPPER "${_logical_target_name}" _coverage_project_name)
set(_coverage_var_name "${_coverage_project_name}_ENABLE_COVERAGE")

if(DEFINED ${_coverage_var_name})
  set(_coverage_arg "-D${_coverage_var_name}=${${_coverage_var_name}}")
endif()
```

This variable is then appended to the CMake configure command for the subproject.

## Benefits

1. **Unified interface**: Developers can control code coverage from the TheRock build without modifying individual component configurations
2. **CI/CD integration**: Coverage builds can be triggered via single CMake flag changes in CI pipelines
3. **Per-component granularity**: Enable coverage only for components under test, reducing build overhead
4. **Standards compliance**: Aligns with the `<PROJECT>_ENABLE_COVERAGE` convention established in rocm-libraries

## Related Work

- **Prerequisite for**: [ROCm/rocm-libraries#8656](https://github.com/ROCm/rocm-libraries/pull/8656) - Adds actual code coverage implementation to component libraries
- **Tested with**: hipDNN, rocBLAS, rocFFT, and other math/ML libraries in the ROCm ecosystem

## Testing

Verify the passthrough works correctly:

```bash
# Configure with coverage enabled for hipDNN
cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx1100 -DHIPDNN_ENABLE_COVERAGE=ON

# Verify the flag appears in the hipDNN subproject's configure command
ninja -C build -v hipDNN | grep HIPDNN_ENABLE_COVERAGE
# Expected: -DHIPDNN_ENABLE_COVERAGE=ON appears in the cmake invocation
```

reboss:
## Motivation

We want to migrate code coverage jobs to TheRock

## Technical Details

This PR introduces a passthrough for cmake definitions to their respective projects with:
```cmake
set(_coverage_arg "-D${_logical_target_name}_ENABLE_COVERAGE=${${_logical_target_name}_ENABLE_COVERAGE}")
```

## Test Plan

We'll use the following PR to test that a code coverage job can be kicked off and run for any given project:
https://github.com/ROCm/rocm-libraries/pull/4441

## Submission Checklist

Resolves: #6572


cmake/therock_subproject.cmake
```diff
@@ -957,6 +957,16 @@ function(therock_cmake_subproject_activate target_name)
...
 
+    # Passthrough -D<PROJECT_NAME>_ENABLE_COVERAGE=ON to the subproject
+    # Convert logical target name to uppercase (e.g., hipDNN -> HIPDNN_ENABLE_COVERAGE)
+    set(_coverage_arg)
+    string(TOUPPER "${_logical_target_name}" _coverage_project_name)
+    set(_coverage_var_name "${_coverage_project_name}_ENABLE_COVERAGE")
+
+    if(DEFINED ${_coverage_var_name})
+      set(_coverage_arg "-D${_coverage_var_name}=${${_coverage_var_name}}")
+    endif()
```
ScottTodd: Please document `{project}_ENABLE_COVERAGE` support here: https://github.com/ROCm/TheRock/blob/60333132be9a33a6b901d23a171270becb6b1c55/docs/development/development_guide.md?plain=1#L178-L202

ScottTodd: You can also move `_coverage_arg` up to be next to the other per-project CMake flags here: https://github.com/ROCm/TheRock/blob/60333132be9a33a6b901d23a171270becb6b1c55/cmake/therock_subproject.cmake#L872-L881

and I think you can simplify this "check if the variable exists" to just match the style used for `CMAKE_ARGS`:
```cmake
list(APPEND _cmake_args "${${target_name}_CMAKE_ARGS}")
```
if the variable doesn't exist, list append will do nothing I think...

cmake/therock_subproject.cmake
```diff
@@ -957,6 +957,16 @@ function(therock_cmake_subproject_activate target_name)
       "STAGE_DESTINATION_DIR=${_rel_stage_destination_dir}"
     )
 
+    # Passthrough -D<PROJECT_NAME>_ENABLE_COVERAGE=ON to the subproject
+    # Convert logical target name to uppercase (e.g., hipDNN -> HIPDNN_ENABLE_COVERAGE)
+    set(_coverage_arg)
+    string(TOUPPER "${_logical_target_name}" _coverage_project_name)
```
ScottTodd: If we standardize the case we should do it for all per-project variables. Is this actually needed?

geomin12 (commented):
lgtm but as Scott mentioned, docs to show the support!

cmake/therock_subproject.cmake
reboss: Yes it's needed, or rather should be there, as it ensures that we always get `-DHIPDNN_ENABLE_COVERAGE`, for example, instead of `-DhipDNN_ENABLE_COVERAGE`, which doesn't follow CMake's standard for definitions.

As it is defined, it will only check if the former is being passed, and if it is, propagate it to hipDNN.  If someone uses some non-standard case, then it will ignore it.

See docs: https://github.com/ROCm/TheRock/pull/3761/changes/1ced5b9abfb7b1e07974d3f6a474b67dcfc4e647

cmake/therock_subproject.cmake
reboss: I reordered the logic and used `_cmake_args` instead.  I simplified the logic a little bit, but the `TOUPPER` is still required.

https://github.com/ROCm/TheRock/pull/3761/changes/53d303ce2f62b93de0965199b289a38848d16a9f

reboss: Added docs here: https://github.com/ROCm/TheRock/pull/3761/changes/1ced5b9abfb7b1e07974d3f6a474b67dcfc4e647

cmake/therock_subproject.cmake
ScottTodd: All per-project variables should use consistent casing, possibly standardizing with `TOUPPER`.

docs/development/development_guide.md
```diff
@@ -201,6 +201,18 @@ For example, this will add options to set tensilelite build parallel and keep bu
   -DhipBLASLt_CMAKE_ARGS="-DTENSILELITE_BUILD_PARALLEL_LEVEL=32;-DTENSILELITE_KEEP_BUILD_TMP=ON"
 ```
 
+- `{PROJECT}_ENABLE_COVERAGE` (uppercase project name required)
+
+TheRock checks for `-D<PROJECT_NAME>_ENABLE_COVERAGE` flags (with uppercase project names) and passes them through to the corresponding subproject. If defined, the flag and its value are forwarded automatically. Note that the project name must be uppercase (e.g., `HIPDNN_ENABLE_COVERAGE`, not `hipDNN_ENABLE_COVERAGE`).
```
ScottTodd, line 210: If other per-project options allow mixed casing, this should too. Need to decide one way or the other, not make this inconsistent and trust people to read the docs.

reboss: I don't think this PR is the place to make that decision.

To be clear, when passing `-DHIPDNN_ENABLE_COVERAGE` to TheRock superbuild will propagate that flag to hipDNN's CMake infra to enable code coverage.  Passing `-DhipDNN_ENABLE_COVERAGE` will not be propagated nor will it be converted to uppercase.

If anything other than `-DHIPDNN_ENABLE_COVERAGE` is passed to hipDNN's CMake infra, it will do nothing, because that is how code coverage is designed in hipDNN, and all the other rocm-libraries projects.

I put the guard here so that the user knows when they pass a non-uppercase value, that it wasn't used by CMake, and that they must use an uppercase value.

CMake defines are always uppercase, that is CMake's standard.  The user will not have to read the docs to know this.  If we do accept any letter-case, then we would be inconsistent with what is defined in rocm-libraries and in CMake.

ScottTodd: This PR is introducing inconsistency. It is the place to handle it, or another PR should be made prior to this one.

ScottTodd: How about having the option in TheRock take the subproject name's casing and then use TOUPPPER before forwarding? Here it would be `hipDNN`: https://github.com/ROCm/TheRock/blob/cb90066183b93a80ca4d8ee7b88e7e4bdf3c740b/ml-libs/CMakeLists.txt#L93

Or, we could not use a passthrough at all, and instead use the same method as we use for `ENABLE_TESTING`:
* https://github.com/ROCm/TheRock/blob/cb90066183b93a80ca4d8ee7b88e7e4bdf3c740b/ml-libs/CMakeLists.txt#L40
* https://github.com/ROCm/TheRock/blob/cb90066183b93a80ca4d8ee7b88e7e4bdf3c740b/math-libs/CMakeLists.txt#L32
* https://github.com/ROCm/TheRock/blob/cb90066183b93a80ca4d8ee7b88e7e4bdf3c740b/math-libs/CMakeLists.txt#L476
* https://github.com/ROCm/TheRock/blob/cb90066183b93a80ca4d8ee7b88e7e4bdf3c740b/core/CMakeLists.txt#L206

We'd add a `THEROCK_ENABLE_COVERAGE` and then at each location that wants to set the option it can choose the name that is needed, e.g. `BUILD_COVERAGE`, `PROJECTNAME_ENABLE_COVERAGE`, etc.?

ScottTodd: Yeah I'm liking the idea of not using a passthrough that assumes all projects will use the same internal CMake option names. I added a few more examples to my prior comment.

reboss: Can you please explain how this is introducing inconsistencies?

The passthrough is the most elegant way to handle this. If I distribute this to multiple files, then the footprint of what needs to be maintained increases significantly and also increases the likelihood that things _will_ become inconsistent as the codebase is maintained.

I can make it just convert anything to uppercase, but that also leads to inconsistencies between running code coverage in a superbuild and building a codecov build for an individual project.

ScottTodd: The inconsistency is here in the docs:

```
-DhipBLASLt_BUILD_TYPE=...
-DHIPDNN_ENABLE_COVERAGE=...
```
have different casing requirements (one must match the subproject name here in TheRock, the other must be all uppercase)

Looking at the docs, we already have a generic passthrough `{project}_CMAKE_ARGS`. Can't that be used for enabling coverage? E.g. `-DhipDNN_CMAKE_ARGS="-DHIPDNN_ENABLE_COVERAGE=ON"`

I'm not sure how effectively we'll be able to standardize on `ALLCAPSPROJECT_ENABLE_COVERAGE` across all projects (in rocm-libraries, rocm-systems, llvm-project, rocgdb, etc.), so I think either of these would be more appropriate here:
* The generic passthrough
* A top level switch that gets translated to project-specific options
Both of those can use more flexible casing rules without introducing an inconsistency to the documentation like that.

reboss: I see what you mean about the inconsistency now.  I can have it accept any-case and then change it to uppercase before passing it to the individual components in this PR.

I do like this method: `-DhipDNN_CMAKE_ARGS="-DHIPDNN_ENABLE_COVERAGE=ON"`, however that is going to require some more testing and likely trial and error in https://github.com/ROCm/rocm-libraries/pull/8656, which depends on these changes.

If it's ok with you, I'll do the case-insensitive change for this PR, and then I can open a follow-up ticket to try using `<project_name>_CMAKE_ARGS` in the future.  Does that work?

ScottTodd: Scanning through https://github.com/ROCm/rocm-libraries/pull/8656, I'm not sure where to start with feedback. Why is this being built first via github actions? Has someone proven out a functional workflow for collecting code coverage instrumentation via local source builds first? That PR is forking per-family CI workflow code that is deprecated and it's rebuilding multiple subprojects from source for every component that wants to use code coverage.

These sorts of changes to build/test should always start from local development and then graduate to the CI system.

As for this particular plumbing, you should already be able to pass individual flags to subprojects via `{project}_CMAKE_ARGS` as demonstrated, and no, it's not worth adding this additional custom passthrough just to save on testing and trial & error. Iterate locally and _then_ migrate to CI, and your iteration time should be ~minutes. If this isn't easy to work on, make it easy to work on. If a separate flag is needed to affect the options across all projects together (similar to "enable tests"), then we can look at some of the other options.

@stellaraccident @suchitravemuri1 might need some support / realignment here, I see people spinning for months on build system and CI changes without adequate design review or oversight.

yxsamliu:
The documented command enables coverage for the project, but it does not enable the GPU profiling runtime. For device coverage, both are required; otherwise the build may fail with an unresolved `__llvm_profile_instrument_gpu` or produce no device profiles. Could the project coverage option also enable `COMPILER_RT_BUILD_PROFILE_ROCM`, or could the documentation show that both options are required?

yxsamliu:
Enabling device coverage currently removes Flang and several normal AMDGPU runtimes, including OpenMP and libc++. This may break other components or produce an incomplete compiler package. Could coverage support be added without removing the existing compiler projects and runtimes?

yxsamliu:
The current CI does not test the new compiler/runtime coverage path. TheRock already has a sanity test that compiles and runs a small HIP kernel. Could we extend it, or add a similar smoke test, to build an instrumented HIP kernel and confirm that a device `.profraw` file is produced? The library-specific coverage report can remain in rocm-libraries.

geomin12 (commented):
in the thurs meeting, we discussed a plan before impl! let's wait on GH issue and make sure this aligns

PR #3762, opened 2026-03-04
update : bypass_tests_for_releases True 906, 908, 90a

kiran-thumma:
## Motivation

JAX and PyTorch wheels for gfx906 and gfx908 are being built and published to v2-staging but never promoted to v2. These families lack test hardware, so there is no test runner available. The current promotion logic in `configure_target_run.py` only searches the presubmit and postsubmit matrices, making nightly-only families invisible to the promotion step. This results in wheels permanently stuck in v2-staging.

## Technical Details

Two changes are needed to unblock promotion for gfx906 and gfx908:

1. **`build_tools/github_actions/amdgpu_family_matrix.py`**: Added `bypass_tests_for_releases: True` to gfx906 and gfx908 Linux entries in the nightly matrix. Since these families have no test hardware (`test-runs-on: ""`), the bypass flag allows promotion without requiring test results.

2. **`build_tools/github_actions/configure_target_run.py`**: Extended the search scope in both `get_runner_label()` and `get_upload_label()` from `["presubmit", "postsubmit"]` to `["presubmit", "postsubmit", "nightly"]`. Without this change, the promotion script cannot find nightly-only families, so their `bypass_tests_for_releases` flag is never read.

**Root cause**: `fetch_package_targets.py` (which drives builds) already searches all three trigger types, but `configure_target_run.py` (which drives promotion) only searched presubmit and postsubmit — creating a mismatch where wheels are built but can never be promoted.

**Promotion policy flow after this change**:
- gfx906: no runner + `bypass_tests_for_releases: True` → promoted to v2
- gfx908: no runner + `bypass_tests_for_releases: True` → promoted to v2

## Test Plan

- [ ] Verify `configure_target_run.py` correctly resolves `bypass_tests_for_releases: True` for gfx906 and gfx908 targets
- [ ] Trigger a nightly build for gfx906 and gfx908 and confirm wheels are promoted from v2-staging to v2
- [ ] Verify no regression for existing presubmit/postsubmit families (gfx94X-dcgpu, gfx950-dcgpu, gfx110X-all, gfx1151, gfx120X-all)

## Submission Checklist


build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -167,6 +167,7 @@
             "test-runs-on": "",
             "family": "gfx906",
             "fetch-gfx-targets": [],
+            "bypass_tests_for_releases": True,
```
ScottTodd, line 170: What is special about these targets that warrants adding `bypass_tests_for_releases` when other targets that we _also_ don't have test runners for don't have this set?

Have we tested packages for these targets on real hardware at all? There's a reason we have the "v2-staging" concept - it ensures that we only promote packages that have been tested and we have confidence in their correctness. gfx906, 906, and 90a are all still not "sanity tested" per https://github.com/ROCm/TheRock/blob/main/SUPPORTED_GPUS.md, so I don't agree with just flipping the bit and releasing them.

PR #3764, opened 2026-03-04
Update RFC0009-OS-Packaging-Requirements.md

LiamfBerry:
## Motivation

amdrocm-core-devel meta packages don't exist in ROCm 7.11 but they do exist as amdrocm-core-dev. There are too many instances of this package that making engineering change the naming scheme is redundant so instead it makes more sense to update the RFC to reflect the current state.

Additionally, rocSHMEM needs to be included in this RFC as a component-specific package

AIROCSHMEM-328

## Technical Details

N/A

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist

### DIFF
--- a/docs/rfcs/RFC0009-OS-Packaging-Requirements.md
+++ b/docs/rfcs/RFC0009-OS-Packaging-Requirements.md
@@ -156,8 +156,8 @@ Using `yum` ROCm Core SDK runtime components and ROCm Core SDK runtime + develop
 yum install rocm # ROCm 8.0
 yum install rocm-core # ROCm 8.0
 yum install rocm-core<ver>
-yum install rocm-core-devel
-yum install rocm-core-devel<ver>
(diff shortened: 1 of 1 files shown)


docs/rfcs/RFC0009-OS-Packaging-Requirements.md
```diff
@@ -202,6 +204,7 @@ Package granularity will be increased with ROCm 8.0. Development packages contai
 | amdrocm-jpeg                            |                             | rocJPEG                                                                       |                                          |
 | amdrocm-file                            |                             | hipFile, rocFile (future addition)                                            |                                          |
```
gaoikawa: Not part of this change I realize, but in the future there will only be 'hipFile' (and not 'rocFile').

So you could change it to 'hipFile (future addition)'.

ScottTodd (approved):

PR #3767, opened 2026-03-04
Fix IndexError when only generic artifacts exist

marbre:
`Parameters.__init__` raised `IndexError` when only generic artifacts were present (e.g. a core-only or headers-only build).

Make `default_target_family` `Optional[str]`, set to `None` when the family set is empty, and guard the `DEFAULT_TARGET_FAMILY` line in the generated `dist_info` contents accordingly. The template's fallback value `DEFAULT_TARGET_FAMILY = "DEFAULT"` is a sentinel for the no-families case; since `AVAILABLE_TARGET_FAMILIES` is also empty in that case, any call to `determine_target_family()` will raise a `ValueError` regardless.

### DIFF
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -80,7 +80,10 @@ class Parameters:
         self.version_suffix = version_suffix
         self.artifacts = artifacts
         self.all_target_families = artifacts.all_target_families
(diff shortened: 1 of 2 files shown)


ScottTodd (approved):

marbre:
CI failures is unrelated.

PR #3595, opened 2026-02-24
Changes to integrate Test Filter Standardization for miopen in TheRock

dileepr1:
## Motivation

This PR is to integrate the Test Filter Standardization POC done in MIOpen to TheRock

## Technical Details

The POC in MIOpen is done here: https://github.com/ROCm/rocm-libraries/pull/3513.
This PR adds test_filters.py as a generic runner to call the required test categories from MIOpen, and can be expanded to other components.

## Test Result

Once active, the MIOpen CI tests will be triggered through this.

PR #3769, opened 2026-03-05
Add rocblas to test_runner (Test Filter Standardization)

tony-davis:
Adds rocBLAS support to the generic test runner so rocBLAS CI uses the same CTest-based filtering as MIOpen.

**Platform:** rocBLAS tests run on **Linux only**. Windows has known issues that will need to be addressed in a follow-up.

**Sharding:** The rocBLAS job runs on **6 shards** on Linux to parallelize the test suite.

**CI:** Example run where the rocBLAS test job ran: [Test rocblas (shard 1/6) — Actions run](https://github.com/ROCm/TheRock/actions/runs/23266184096/job/67672087900).

**Related work:** This PR works in conjunction with [PR 4659](https://github.com/ROCm/rocm-libraries/pull/4659) in [rocm/rocm-libraries](https://github.com/ROCm/rocm-libraries), which adds the CTest test categories and configuration consumed by the runner here.

---

**Changes:**

- **`build_tools/github_actions/test_executable_scripts/test_runner.py`**
  - Add `rocblas` to `COMPONENT_DIR_MAPPING` so the generic test runner can discover and run rocBLAS CTest tests.

- **`build_tools/github_actions/fetch_test_configurations.py`**
  - Point the rocblas job at `test_runner.py` instead of `test_rocblas.py`, so CI uses the same CTest-based filtering as MIOpen.
  - Set `platform` to `["linux"]` only (Windows support deferred).
  - Set `total_shards_dict` for Linux to **6** shards.
  - Bump `timeout_minutes` to 1728 (24h full category + 20% margin) to cover the longest rocBLAS test category.

- **`math-libs/BLAS/artifact-blas.toml`**
  - Include `bin/rocblas/**` in the artifact so the test runner has the `bin/rocblas/` directory and `CTestTestfile.cmake` needed for CTest-based execution.

Relates to AIROCBLAS-45.

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -75,14 +75,14 @@ test_matrix = {
     "rocblas": {
         "job_name": "rocblas",
         "fetch_artifact_args": "--blas --tests",
-        # GHA step timeout: max category timeout in rocBLAS test_categories.yaml is 24h (full);
-        # 1440 min + 20% margin = 1728 min
-        "timeout_minutes": 1728,
+        # GHA step timeout: max category timeout in rocBLAS should be 24 hours / 6 shards = 4 hours per shard
+        # 240 min + 20% margin = 288 min
+        "timeout_minutes": 288,
         "test_script": f"python {_get_script_path('test_runner.py')}",
         "platform": ["linux"],
         "total_shards_dict": {
             "linux": 6,
-            "windows": 1,
+            "windows": 6,
         },
     },
     "rocroller": {


build_tools/github_actions/test_executable_scripts/README.md
geomin12, line 1: nit: I'm happy to call this just `README.md`!

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -0,0 +1,247 @@
...
+ctest_parallel_count = 8
+if AMDGPU_FAMILIES and "gfx1152" in AMDGPU_FAMILIES:
+    ctest_parallel_count = 4
+elif AMDGPU_FAMILIES and "gfx1153" in AMDGPU_FAMILIES:
+    ctest_parallel_count = 4
+environ_vars = os.environ.copy()
+# Set the GTEST env vars for Gtest based tests
+# Set ROCM_PATH for tests that rely on it
+environ_vars["GTEST_SHARD_INDEX"] = str(int(SHARD_INDEX) - 1)
+environ_vars["GTEST_TOTAL_SHARDS"] = str(TOTAL_SHARDS)
```
geomin12, line 82: i am not sure this is needed, as ctest below shards these using --test-information

i think if ctests are sharded and the underlying gtest is further sharded, this may result in unwanted behavior

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -0,0 +1,247 @@
...
+ROCM_PATH = Path(THEROCK_BIN_DIR).resolve().parent
+environ_vars["ROCM_PATH"] = str(ROCM_PATH)
+
```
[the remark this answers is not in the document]
geomin12: this is no longer needed as we set those fields on a global level

tony-davis:
> also hold on, we have duplicate work here: #3595
> 
> this PR has rocblas + miopen, so happy to use this one

It's not a duplicate. It depends on it. It's from that branch, enabling just rocBLAS. rocm-libraries is in conflict right now because I would like to run this through CI with that particular commit.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -75,11 +75,13 @@ def _get_script_path(script_name: str) -> str:
     "rocblas": {
         "job_name": "rocblas",
         "fetch_artifact_args": "--blas --tests",
-        "timeout_minutes": 15,
-        "test_script": f"python {_get_script_path('test_rocblas.py')}",
-        "platform": ["linux", "windows"],
+        # GHA step timeout: max category timeout in rocBLAS test_categories.yaml is 24h (full);
+        # 1440 min + 20% margin = 1728 min
+        "timeout_minutes": 1728,
```
geomin12: For the 1728 mins, is that with 6 shards? so 10368 mins???

Or should this be 288 mins (since 6 shards)

tony-davis: Yeah, we should reduce that since it's now 6 shards. That was based on the prior settings. I'll adjust.

tony-davis: Done.

PR #3536, opened 2026-02-20
Add UCX functional test to extended_tests framework

lajagapp:
## Summary

This PR adds UCX (Unified Communication X) functional test to the extended_tests framework.

### New UCX Functional Test

- **test_ucx.py**: Functional test for UCX ROCm integration
  - Clones UCX repository from GitHub
  - Builds UCX with ROCm support
  - Runs UCX gtests filtered for ROCm tests
  - Parses gtest native JSON output for results

- **functional_test_matrix.py**: UCX test matrix entry

## Depends on

This PR depends on initial functional framework change: https://github.com/ROCm/TheRock/pull/3648

Will rebase once initial PR is merged.

PR #3784, opened 2026-03-05
Add UCX build support in external-builds/

lajagapp:
## Summary

Adding UCX (Unified Communication X) build support in `external-builds/ucx/`, moving the build step out of the test runner and into the CI build phase.

## Motivation

Review feedback on the UCX test PR #3536, builds should happen on CPU runners during the build step, not on GPU runners during testing. UCX is an external project that builds on top of ROCm (via `--with-rocm`), so follows the `external-builds/` pattern used by UCCL and PyTorch.

### New: `external-builds/ucx/`

- **`ucx_repo.py`** — Checkout script for upstream UCX sources (shallow clone, configurable branch/tag)
- **`build_ucx.py`** — Autotools-based build script (autogen → configure → make → install) with ROCm support
- **`README.md`** — Prerequisites, quickstart, and usage docs

## Usage

1. Checkout UCX
python ucx_repo.py checkout --depth 1
2. Build with ROCm
python build_ucx.py --rocm-path /path/to/rocm --output-dir ./output

### DIFF
--- a/dockerfiles/build_manylinux_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_x86_64.Dockerfile
@@ -79,6 +79,9 @@ RUN yum install -y epel-release && \
       patchelf \
       vim-common \
       git-lfs \
+      autoconf \
+      automake \
+      libtool \
     && yum install -y \
       texinfo \
       flex \


dockerfiles/build_manylinux_x86_64.Dockerfile
geomin12, line 1: this change will be needed in a separate PR, in order for the changes to take effect in here (will require update in build linux artifacts yml)

dockerfiles/build_manylinux_x86_64.Dockerfile
```diff
@@ -79,6 +79,9 @@ RUN yum install -y epel-release && \
       patchelf \
       vim-common \
       git-lfs \
+      autoconf \
```
geomin12, line 82: add a comment to why this is needed

PR #3788, opened 2026-03-05
[rocgdb] Ignore non-FAIL and non-UNRESOLVED timeouts

lumachad:
I spotted the testing script considering a timeout in a KFAIL result as a failure, when it shouldn't be.

The patch adjusts things so we only consider a timeout a failure when it is in a FAIL or UNRESOLVED category.

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_rocgdb.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rocgdb.py
@@ -421,6 +421,13 @@ class TestResults:
                         test_name
                     )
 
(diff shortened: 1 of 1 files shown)


lumachad:
@spatrang Thoughts on this one?

spatrang (approved):
LGTM. The timeout demotion correctly limits failure tracking to cases where the underlying status is already FAIL or UNRESOLVED. Note that after the demotion, or is_timeout is technically redundant with the status check, but it reads well as a statement of intent and is harmless to keep.

PR #3789, opened 2026-03-05
Use `--reference` for nested `llvm-project`(s)

AaronStGeorge:
## Motivation

https://github.com/ROCm/TheRock/pull/3738 disabled `iree-libs` projects during `fetch_sources.py` by default as the cost of fetching IREE's LLVM was too high. The `ROCm` and `iree-org` llvm forks share the vast majority of git history, so borrowing objects locally saves ~3.3GB of downloads and minutes of clone time -- making the fetch much more palatable.

## Technical Details

When fetching nested submodules, use `compiler/amd-llvm` as a `--reference` repo for `llvm-project` forks.  In stage-aware CI where `amd-llvm` may not be fetched, the reference is skipped gracefully.

~With the clone cost addressed, re-enable `--include-iree-libs` in `fetch_sources.py` by default.~ We'll leave the fetch disabled for now as there still is a cost.

## Test Plan

Fetching sources is adequately covered by existing CI.

## Submission Checklist

### DIFF
--- a/build_tools/fetch_sources.py
+++ b/build_tools/fetch_sources.py
@@ -140,17 +140,49 @@ def fetch_nested_submodules(args, projects):
 
         # Fetch the nested submodules
         parent_dir = THEROCK_DIR / get_submodule_path(parent)
-        nested_submodule_paths = [
-            get_submodule_path(nested_submodule, cwd=parent_dir)
-            for nested_submodule in nested_submodules
-        ]
-        run_command(
-            ["git", "submodule", "update", "--init"]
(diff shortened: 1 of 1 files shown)


build_tools/fetch_sources.py
ScottTodd, line 1: Let's check how this should compose with
* https://github.com/ROCm/TheRock/pull/3867

AaronStGeorge:
Fusilli has been de-prioritized, this is no longer relevant.

PR #3791, opened 2026-03-05
Use `fusilli-provider` defined in `rocm-libraries`

AaronStGeorge:
## Motivation

Update TheRock build for fusilli provider to pull from new location in `rocm-libraries` (see: https://github.com/ROCm/rocm-libraries/pull/5149)

## Technical Details

Really just a couple boilerplate additions.

Note: as an optimization, we may want to add a feature to allow a sparse checkout when fetching the `rocm-libraries` source set ([link](https://github.com/ROCm/TheRock/pull/3791/changes#diff-9c53c23f4ebfe2986e8dfe8af30208eb6a1a6294946bb07b8fd69f9e1b84d379R317)). The `fusilli-libs` stage is only using a very small portion of that repo.

## Test Plan

As the IREE / fusilli build are not enabled by default, see this test branch with temporary changes required to enable the IREE / fusilli build: https://github.com/ROCm/TheRock/pull/3912

## Test Result

It works! [successful build + CI](https://github.com/ROCm/TheRock/actions/runs/22972523986/job/66726952506?pr=3912)

## Submission Checklist

### DIFF
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -314,7 +314,7 @@ description = "Fusilli hipdnn provider"
 type = "generic"
 artifact_group_deps = ["hip-runtime", "iree-compiler"]
 # TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
-source_sets = ["iree-libs", "rocm-systems"]
+source_sets = ["iree-libs", "rocm-libraries", "rocm-systems"]
 
 [artifact_groups.media-libs]
 description = "Media Libraries"
--- a/iree-libs/CMakeLists.txt
+++ b/iree-libs/CMakeLists.txt
@@ -88,7 +88,7 @@ if(THEROCK_ENABLE_FUSILLIPROVIDER)
   ##############################################################################
   therock_cmake_subproject_declare(fusilliprovider
     USE_DIST_AMDGPU_TARGETS
-    EXTERNAL_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/fusilli/plugins/hipdnn-plugin"
+    EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_LIBRARIES_SOURCE_DIR}/dnn-providers/fusilli-provider"
     BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/fusilliprovider"
     BACKGROUND_BUILD
     CMAKE_ARGS


.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -11,7 +11,9 @@
 # 5. debug-tools (generic) - amd-dbgapi, rocr-debug-agent, rocgdb (parallel to math-libs)
 # 6. dctools-core (generic) - RDC (parallel to math-libs)
 # 7. profiler-apps (generic) - rocprofiler-systems (parallel to math-libs)
-# 8. media-libs (generic) - sysdeps-amd-mesa, rocdecode, rocjpeg
+# 8. iree-compiler (generic) - IREE compiler (parallel to math-libs)
+# 9. fusilli-libs (generic) - Fusilli hipdnn provider (after math-libs + iree-compiler)
```
sjain-stanford, line 15: I maybe missing the details here but I think Stella wanted `iree-libs` to remain unchanged (i.e. no separate fusilli-libs). Also, fusilli-libs is pretty lightweight and shouldn't add to the build latency when combined with iree-libs, if that's the concern you're working around.

sjain-stanford: Looks like you have them in the same `IREE_LIBS` feature group below. Not sure how those map to "stages" but I think stages are kind of meaty chunks of libraries that are built logically together, so having a separate stage for fusilli seems like an outlier. Just my thought, not a strong opinion.

AaronStGeorge: good question, but you likely want to comment on https://github.com/ROCm/rocm-libraries/pull/5149. This PR is a CI test that I'm going to close after https://github.com/ROCm/rocm-libraries/pull/5149 lands. The changes are roll-up of https://github.com/ROCm/TheRock/pull/3611 so that the multi-arch build works at all.

But to answer your question here (then I'll probably copy paste it over), I think the splitting `iree-compiler` stage makes sense given the stage topology that *currently* exists. https://github.com/ROCm/TheRock/pull/3611 was only intended to enable the multi-arch build for fusilli not to refactor the stage topology.

Personally I'd prefer to land https://github.com/ROCm/TheRock/pull/3611 so that we have a working multi-arch build then have a separate effort to refactor the stage topology. Collapsing these stages likely makes sense then. But we could do it now 🤷‍♂️, it shouldn't be a huge amount of effort or either way - I'd bet Claude can 1 shot it.

AaronStGeorge: Moved to https://github.com/ROCm/TheRock/pull/3611#discussion_r2892184914

BUILD_TOPOLOGY.toml
```diff
@@ -314,7 +314,7 @@ description = "Fusilli hipdnn provider"
 type = "generic"
 artifact_group_deps = ["hip-runtime", "iree-compiler"]
 # TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
-source_sets = ["iree-libs", "rocm-systems"]
+source_sets = ["iree-libs", "rocm-libraries", "rocm-systems"]
```
sjain-stanford, line 317: Curious, where do we specify that this is a sparse checkout of rocm-libs? Or is that future work (after hipdnn moves to the core stage)?

AaronStGeorge: That's future work, it's not a sparse checkout at the moment.
> Note: as an optimization, we may want to add a feature to allow a sparse checkout when fetching the rocm-libraries source set ([link](https://github.com/ROCm/TheRock/pull/3791/changes#diff-9c53c23f4ebfe2986e8dfe8af30208eb6a1a6294946bb07b8fd69f9e1b84d379R317)). The fusilli-libs stage is only using a very small portion of that repo.

PR #3773, opened 2026-03-05
Filter mxDataGenerator from rocRoller artifact entries

ScottTodd:
## Motivation

Tentative fix for https://github.com/ROCm/TheRock/issues/3770.

The `blas_dev` artifact includes
```
math-libs/
  BLAS/
    rocRoller/
      stage/
        include/
          mxDataGenerator/
            bf16.hpp
            ...
```
but these files are also in the `support_dev` artifact:
```
math-libs/
  support/
    mxDataGenerator/
      stage/
        include/
          mxDataGenerator/
            bf16.hpp
            ...
```

When flattened, these files conflict. The artifact extra code attempts an unlink/replace (most recent wins), but race conditions are possible during parallel artifact extraction.

## Technical Details

Together with the `aqlprofile-tests` / `rocprofiler-sdk` artifact fixes in https://github.com/ROCm/TheRock/pull/3698, this gets us back to having zero duplicate files _across_ all artifacts (e.g. `blas` and `support`). We still have duplicates _within_ artifacts (e.g. `miopen_test` and `miopen_run`).

## Test Plan

* Regular CI workflows
  * Show that this doesn't break any existing tests
  * Check that `blas_dev` no longer includes the mxDataGenerator files
* New unit tests to enforce no duplicates going forward (future PR, nontrivial to add)

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3727, opened 2026-03-03
Error when installing test package for dnn and sparse

jonatluu:
# Motivation

Error when installing test package for dnn and blas. All packages are built from artifacts, test package added containing test bins and test files for miopen and sparse.

## Technical Details
bin/miopen_gtest is present in amdrocm-dnn-dev this is causing an error when installing amdrocm-dnn-test. 
hipsparse/test/hipsparse_clientmatrices.cmake is present in amdrocm-sparse-dev which causes an error when installing amdrocm-sparse-test. 

https://github.com/ROCm/TheRock/issues/3544
https://github.com/ROCm/TheRock/issues/3545

Resolves https://github.com/ROCm/TheRock/issues/2784

## Test Plan

build and install all packages including amdrocm-dnn and amdrocm-sparse packages to see if error is gone. 

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3793, opened 2026-03-05
[ci] fixing `hipdnn` and `miopen` to prevent duplicate `tar` file extraction

### LINKED ISSUE #2784: [Issue]: Windows test machines failing due to error `The process cannot access the file because it is being used by another process` for hipdnn
For Windows test machines (particularly strix halos), the processes from previous jobs are not being cleaned up properly, resulting in failures during the setup phase.

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'build\\bin\\hipdnn_test_sdk_tests.exe'
```

Failures from one day, urls here: 
https://github.com/ROCm/TheRock/actions/runs/20736116290/job/59579474623
https://github.com/ROCm/TheRock/actions/runs/20736116290/job/59540308782
https://github.com/ROCm/TheRock/actions/runs/20736116290/job/59540308717
https://github.com/ROCm/TheRock/actions/runs/20721998822/job/59505541828

geomin12:
Resolves #2784

Discovered that MIOpen and hipDNN have identical files, thus causing the failure when trying to extract files in parallel (sometimes)

MIOpen:
Screenshot 2026-03-05 110045

hipDNN:
Screenshot 2026-03-05 105851

### DIFF
--- a/ml-libs/artifact-hipdnn.toml
+++ b/ml-libs/artifact-hipdnn.toml
@@ -6,6 +6,7 @@
 exclude = [
   "lib/libhipdnn_backend_private.a",
   "lib/test_plugins/**",
+  "bin/*hipdnn_*_test*",
 ]
 [components.test."ml-libs/hipDNN/stage"]
 include = [
--- a/ml-libs/artifact-miopen.toml
+++ b/ml-libs/artifact-miopen.toml
@@ -11,3 +11,6 @@ include = [
   "bin/miopen_gtest*",
 ]
 [components.run."ml-libs/MIOpen/stage"]
+exclude = [
+  "bin/miopen_gtest*",
+]


ScottTodd:
:O I'm fixing another duplicate file issue with https://github.com/ROCm/TheRock/pull/3773, and I'm working on some unit tests that check for this. My tests missed hipdnn/miopen though... hmmm...

ml-libs/artifact-miopen.toml
```diff
@@ -11,3 +11,6 @@ include = [
   "bin/miopen_gtest*",
 ]
 [components.run."ml-libs/MIOpen/stage"]
+exclude = [
+  "bin/miopen_gtest*",
+]
```
ScottTodd, line 16: This fix looks correct, for miopen.

The inheritance chain is lib → run → dbg → dev → doc, with each extending the previous (and filtering out files that the previous includes). See for example: https://github.com/ROCm/TheRock/blob/b059e0c28d65b1f804ee1ebb6dc18f3a47aee0e3/build_tools/_therock_utils/artifact_builder.py#L50-L64

Note that "test" is its own thing - it doesn't extend any of the others, so it doesn't know that "run" had already included `miopen_gtest*`.

I _think_ there's a systemic issue here though. Forgive the LLM output... I'm chasing/debugging this live while adding new tests:
```
● Whoa. Here's the full within-artifact component overlap picture for classic Linux gfx94X:

  ┌─────────────────────┬─────────────┬───────────────────┐
  │      Artifact       │ Components  │ Overlapping files │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocprofiler-compute │ lib vs test │ 4,139             │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocprofiler-sdk     │ run vs test │ 452               │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocprofiler-sdk     │ lib vs test │ 10                │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ base                │ run vs test │ 33                │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocprofiler-compute │ run vs test │ 22                │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocrtst             │ run vs test │ 12                │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rdc                 │ lib vs test │ 4                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocdecode           │ dev vs test │ 3                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocgdb              │ dev vs test │ 2                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ core-ocl            │ lib vs test │ 2                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ blas                │ dev vs test │ 1                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ miopen              │ run vs test │ 1                 │
  ├─────────────────────┼─────────────┼───────────────────┤
  │ rocprofiler-compute │ doc vs test │ 1                 │
  └─────────────────────┴─────────────┴───────────────────┘

  The pattern is consistent: every overlap involves test. This confirms the root cause — test has no extends chain, so
  it re-claims files already taken by other components. rocprofiler-compute is the worst offender at 4,139 duplicated
  files between lib and test.

  PR #3793 fixes miopen (1 file) with an exclude on run. But this is a systemic issue affecting 11 artifacts. The real
  fix might be to make test extend doc (top of the chain) by default in the scanner, rather than patching each
  descriptor individually. That's an artifact_builder.py change, not a descriptor change.
```

ml-libs/artifact-hipdnn.toml
```diff
@@ -6,6 +6,7 @@
 exclude = [
   "lib/libhipdnn_backend_private.a",
   "lib/test_plugins/**",
+  "bin/*hipdnn_*_test*",
 ]
```
ScottTodd, line 10: This change may not be necessary. `lib` already only includes certain file extensions: https://github.com/ROCm/TheRock/blob/b059e0c28d65b1f804ee1ebb6dc18f3a47aee0e3/build_tools/_therock_utils/artifact_builder.py#L36-L49

geomin12:
Closing as #3727 fixes this better

PR #3810, opened 2026-03-06
Store fingerprint in stage.prebuilt and use to check if prebuilt artifacts are out-of-date with source and rebuild if necessary

PeterCDMcLean:
## Motivation

In the prebuilt artifact flow, artifacts may be out-of-date with the source tree. In this case, it is not obvious that they are. However, because the fingerprinting information is readily we can use this information to detect the out of date artifacts and force a rebuild.

## Technical Details

During artifact creation, output the fingerprint `.fprint` file into the root of the artifact xz. During artifact unpacking, copy the content of the .fprint file in to the stage.prebuilt file. During CMake configuration of the rock, check the source-tree fingerprint against the prebuilt fingerprint contents of the stage.prebuilt file.

## Test Plan

TBD

## Test Result

TBD

## Submission Checklist

TBD

### DIFF
--- a/cmake/therock_artifacts.cmake
+++ b/cmake/therock_artifacts.cmake
@@ -171,6 +171,10 @@ function(therock_provide_artifact slice_name)
   if(_fprint_is_valid)
     string(SHA256 _fprint "${_fprint_content}")
   endif()
+  if(THEROCK_VERBOSE_FPRINT)
+    message(STATUS "  ARTIFACT FPRINT (${slice_name}) = ${_fprint}")
+    message(STATUS "  ARTIFACT FPRINT CONTENT = ${_fprint_content}")
+  endif()
 
   # Populate commands.
   set(_fileset_tool "${THEROCK_SOURCE_DIR}/build_tools/fileset_tool.py")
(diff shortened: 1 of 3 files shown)


cmake/therock_artifacts.cmake
```diff
@@ -347,11 +347,14 @@ function(therock_provide_artifact slice_name)
       continue()
     endif()
     set(_component_dir "${_artifacts_dir}/${slice_name}_${_component}${_bundle_suffix}")
-    set(_fprint_file "${_component_dir}.fprint")
+    # Write fingerprint file INSIDE the component directory so it gets archived
+    set(_fprint_file "${_component_dir}/.fprint")
```
ScottTodd: We might also want to upload `{component_name}.fprint` files alongside artifact archives so we can check for fprint matches from prior workflow runs for https://github.com/ROCm/TheRock/issues/3399 without needing to download archives and open them up.

Could build a database with fast lookups for "this is the fprint for my current source, find me an already built artifact".

PeterCDMcLean: Now that I've realized that artifacts are 1-to-many mapping for artifacts->subprojects, I agree and am renaming and placing the .fprint files as {slice_name}.fprint and they exist outside the xz tar balls.

The fprint files in the artifacts don't collapse the 'subproject = subproject_fprint' any more those lines are kept verbatim and as builtctl.py unpacks the artifacts it will correlate the fprints with the subproject calculated fprint.

The 1-to-many mapping will complicate any database.

I am still in the process of testing this.

PR #3814, opened 2026-03-06
Install rocm-develop packages during torch testing

### LINKED ISSUE #2152: [issue] release linux pytorch Error building extension 'dummy_allocator'
Run: https://github.com/ROCm/TheRock/actions/runs/19358650787
effected gfx: 
- gfx1151

pytroch version:
- 2.8

python version:
- 3.13

jobs:
https://github.com/ROCm/TheRock/actions/runs/19358650787/job/55387944049

> 2025-11-14T09:42:52.5202505Z     _run_ninja_build(
2025-11-14T09:42:52.5202603Z     ~~~~~~~~~~~~~~~~^
2025-11-14T09:42:52.5202700Z         build_directory,
2025-11-14T09:42:52.5202801Z         ^^^^^^^^^^^^^^^^
2025-11-14T09:42:52.5202899Z         verbose,
2025-11-14T09:42:52.5202988Z         ^^^^^^^^
2025-11-14T09:42:52.5203098Z         error_prefix=f"Error building extension '{name}'")
2025-11-14T09:42:52.5203241Z         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-14T09:42:52.5203486Z   File "/__w/TheRock/TheRock/.venv/lib/python3.13/site-packages/torch/utils/cpp_extension.py", line 2612, in _run_ninja_build
2025-11-14T09:42:52.5203733Z     raise RuntimeError(message) from e
2025-11-14T09:42:52.5203888Z RuntimeError: Error building extension 'dummy_allocator_v1'

darren-amd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/2152

Several PyTorch tests (`test_mempool_with_allocator`, `test_mempool_empty_cache_inactive`, `test_mempool_limited_memory_with_allocator`, etc.) fail because they compile C++ extensions at runtime (e.g. `dummy_allocator`) which require development headers like `hipblas/hipblas.h`. These tests were previously skipped. Installing `rocm[devel]` alongside `torch` provides the missing headers and lets the tests pass.

## Technical Details

- Install rocm[devel] packages based on the pinned version from the torch packages.
- Set `ROCM_PATH` via `hipconfig --rocmpath` so torch's `_find_rocm_home()` resolves correctly.
- Remove skip entries for `dummy_allocator`-related tests across `generic.py`, `pytorch_2.8.py`, `pytorch_2.9.py`, `pytorch_2.10.py`, and `pytorch_2.11.py`.
- Add a note to `README.md` about installing `rocm[devel]` for local test runs.

## Test Plan

- [x] Run the test workflow and verify the previously-skipped `dummy_allocator` tests now pass

## Test Result

The tests now pass: https://github.com/ROCm/TheRock/actions/runs/22738981360

### DIFF
--- a/.github/workflows/build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/build_portable_linux_pytorch_wheels.yml
@@ -343,7 +343,6 @@ jobs:
       python_version: ${{ inputs.python_version }}
       torch_version: ${{ needs.build_pytorch_wheels.outputs.torch_version }}
       pytorch_git_ref: ${{ inputs.pytorch_git_ref }}
-      rocm_devel_version: ${{ inputs.rocm_version }}
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ inputs.ref || '' }}
 
--- a/.github/workflows/build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/build_portable_linux_pytorch_wheels.yml
@@ -343,6 +343,7 @@ jobs:
       python_version: ${{ inputs.python_version }}
       torch_version: ${{ needs.build_pytorch_wheels.outputs.torch_version }}
       pytorch_git_ref: ${{ inputs.pytorch_git_ref }}
+      rocm_devel_version: ${{ inputs.rocm_version }}
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ inputs.ref || '' }}
 
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -141,10 +141,17 @@ jobs:
       - name: Set up virtual environment
         run: |
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages "torch==${{ inputs.torch_version }} rocm[devel]==${{ inputs.rocm_devel_version }}" \
+            --packages "torch==${{ inputs.torch_version }}" \
             --index-url=${{ inputs.package_index_url }} \
             --index-subdir=${{ inputs.amdgpu_family }} \
             --activate-in-future-github-actions-steps
+
+      # Temporary workaround for CI since 0305 packages are missing
+      - name: Install rocm[devel]
+        run: |
+          python -m pip install --force-reinstall --no-deps \
+            --index-url="${{ inputs.package_index_url }}/${{ inputs.amdgpu_family }}" \
+            "rocm[devel]==${{ inputs.rocm_devel_version }}"
 
       # rocm[devel] provides development headers needed by tests that compile
       # C++ extensions (e.g. dummy_allocator). hipconfig --rocmpath points to
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -141,17 +141,10 @@ jobs:
       - name: Set up virtual environment
         run: |
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages "torch==${{ inputs.torch_version }}" \
+            --packages "torch==${{ inputs.torch_version }} rocm[devel]==${{ inputs.rocm_devel_version }}" \
             --index-url=${{ inputs.package_index_url }} \
             --index-subdir=${{ inputs.amdgpu_family }} \
             --activate-in-future-github-actions-steps
-
-      # Temporary workaround for CI since 0305 packages are missing
-      - name: Install rocm[devel]
-        run: |
-          python -m pip install --upgrade \
-            --index-url="${{ inputs.package_index_url }}/${{ inputs.amdgpu_family }}" \
-            "rocm[devel]==${{ inputs.rocm_devel_version }}"
 
       # rocm[devel] provides development headers needed by tests that compile
       # C++ extensions (e.g. dummy_allocator). hipconfig --rocmpath points to
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -33,6 +33,10 @@ on:
         description: PyTorch ref to checkout test sources from. (e.g. "nightly", or "release/2.7")
         type: string
         default: "release/2.7"
+      rocm_devel_version:
+        description: 'rocm[devel] package version to install (e.g. "7.12.0a20260304")'
+        required: true
+        type: string
 
   workflow_call:
     inputs:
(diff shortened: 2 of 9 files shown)


.github/workflows/build_portable_linux_pytorch_wheels.yml
```diff
@@ -343,6 +343,7 @@ jobs:
       python_version: ${{ inputs.python_version }}
       torch_version: ${{ needs.build_pytorch_wheels.outputs.torch_version }}
       pytorch_git_ref: ${{ inputs.pytorch_git_ref }}
+      rocm_devel_version: ${{ inputs.rocm_version }}
```
marbre: torch already has the correct dependency on rocm[libraries] which is hard coded to the correct version. We could extract that via `pip show build| grep Version` instead.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -134,11 +141,19 @@ jobs:
       - name: Set up virtual environment
         run: |
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages torch==${{ inputs.torch_version }} \
+            --packages "torch==${{ inputs.torch_version }} rocm[devel]==${{ inputs.rocm_devel_version }}" \
```
marbre: torch version also encoded the rocm version, could also simply reuse that one.

marbre (changes requested):
I don't think this should be explicitly passed. It is already encoded in the torch's version extra version or we extract the rocm version which torch depends on. This is a bit more scripting but less error prone (as those rocm[devel] and rocm[libraries] always need to match). I don't see any benefit in making this flexible. It also makes workflows a bit easier.

.github/workflows/test_pytorch_wheels.yml
darren-amd: Thanks Marius, that makes sense. I initially did it this way in case we wanted to test different torch/dev packages but this is much simpler.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -133,12 +133,22 @@ jobs:
 
       - name: Set up virtual environment
         run: |
+          TORCH_VERSION="${{ inputs.torch_version }}"
+          ROCM_VERSION="${TORCH_VERSION#*+rocm}"
```
marbre: This will not work with dev builds:
https://github.com/ROCm/TheRock/blob/497d042aa63d782df9b21f13e6e3b89d752b3fa7/build_tools/github_actions/determine_version.py#L58

darren-amd: Thanks, adjusted to just pull the rocm version instead since I didn't want a messy regex

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -134,11 +134,24 @@ jobs:
       - name: Set up virtual environment
         run: |
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages torch==${{ inputs.torch_version }} \
+            --packages "torch==${{ inputs.torch_version }}" \
```
marbre:
```suggestion
            --packages torch==${{ inputs.torch_version }} \
```

Can we leave as is to have this consistent?

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -134,11 +134,24 @@ jobs:
...
             --index-url=${{ inputs.package_index_url }} \
             --index-subdir=${{ inputs.amdgpu_family }} \
             --activate-in-future-github-actions-steps
 
+      - name: Install rocm[devel]
+        run: |
+          ROCM_VERSION=$(pip show rocm | sed -n 's/^Version: //p')
```
marbre: `setup_venv.py` might use `uv`, thus this might not be fail safe. Eventually add a TODO at least?

marbre (commented):
Do we have a test run for this?

.github/workflows/test_pytorch_wheels.yml
darren-amd: Yes, will make this change.

.github/workflows/test_pytorch_wheels.yml
darren-amd: Hmm, I think I can change ROCM_VERSION to use `importlib.metadata.version` instead, but will add a note in case we want to `uv pip install` vs `pip install` the develop packages.

darren-amd:
Thanks for another review Marius, will run a test with the new changes. Ran tests before the most recent changes that passed: https://github.com/ROCm/TheRock/actions/runs/22738981360, but there seem to be some unrelated kernel issues on the runner's right now.

darren-amd:
Latest test: https://github.com/ROCm/TheRock/actions/runs/22928843805/job/66644082366

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -139,6 +139,20 @@ jobs:
             --index-subdir=${{ inputs.amdgpu_family }} \
             --activate-in-future-github-actions-steps
 
+      # TODO: If setup_venv.py switches to uv, update this to use uv pip install.
+      - name: Install rocm[devel]
+        run: |
+          ROCM_VERSION=$(python -c "from importlib.metadata import version; print(version('rocm'))")
+          python -m pip install \
+            --index-url="${{ inputs.package_index_url }}/${{ inputs.amdgpu_family }}" \
+            "rocm[devel]==${ROCM_VERSION}"
```
ScottTodd, line 148: This came up in discussion with @stellaraccident yesterday. We need to test both with and without `rocm[devel]` installed. Can't only test with `rocm[devel]` installed, as that would allow production impacting issues to slip through our testing.

(Another option would be to make the `torch` packages always depend on `rocm[devel]` packages, but we'd really rather not do that and inflict the larger package installs on all users)

stellaraccident:
This is very unwise and will cause regressions in the field with no way to test them. It also diverges from how cuda operates with respect to runtime vs sdk split iiuc.

I'm open to finding practical solutions to this, but I don't want to compromise testing integrity by including packages that will mask issues with the core runtime build.

darren-amd:
Thanks all, I agree that this is not the correct approach. Created a ticket: https://github.com/ROCm/TheRock/issues/3939, which proposes splitting the tests into 2 buckets (with/without `rocm[devel]`), then running the tests that don't require `rocm[devel]` -> install/run tests that require it. Will work on it after torch tests are stabilized.

PR #3817, opened 2026-03-06
fix: replace copy_directory with timestamp-preserving copy for sysdeps

### LINKED ISSUE #2719: [Issue]: cmake -E copy_directory does not preserve timestamps of the source directory
### Problem Description

Projects with build systems based on autotools are susceptible to timestamp changes to the source files. If a timestamp is changed in a particular way, it may trigger calls to regenerate files via tools that won't be available in the host system doing the build. This can be autoconf, automake or makeinfo.

This is the case for the third-party/sysdeps libraries we download/extract/patch, of which gmp is one. For those libraries we usually do a copy of the source directory and build off of that copy.

For example, here's the original file, extracted from the tarball:

```
ls -lah --full-time /therock/output/build/third-party/sysdeps/linux/gmp/source/doc/gmp.texi
-rw-r--r-- 1 root root 446K 2025-12-30 01:22:40.116677079 +0000 /therock/output/build/third-party/sysdeps/linux/gmp/source/doc/gmp.texi
```

And here's the copy of that file:

```
ls -lah --full-time /therock/output/build/third-party/sysdeps/linux/gmp/build/build/s/doc/gmp.texi
-rw-r--r-- 1 root root 446K 2025-12-30 01:22:40.837679081 +0000 /therock/output/build/third-party/sysdeps/linux/gmp/build/build/s/doc/gmp.texi
```

As we can see the timestamps are not the same. Depending on the order of the copy, files might end up with timestamps that will signal a need to call regeneration tools.

For now I've proposed a workaround for GMP (https://github.com/ROCm/TheRock/pull/2718), but it would be nice to have stable timestamps to avoid any issues in the future.

Just an additional piece of information. The downloading/extracting phase *can* have the original timestamp preserved via the "DOWNLOAD_EXTRACT_TIMESTAMP TRUE" option. But those timestamps will still be lost after a directory copy.

### Operating System

OS: NAME="AlmaLinux" VERSION="8.10 (Cerulean Leopard)"

### CPU

CPU: model name      : AMD Ryzen Threadripper PRO 7975WX 32-Cores

### GPU

GPU:   Name:                    AMD Ryzen Threadripper PRO 7975WX 32-Cores   Marketing Name:          AMD Ryzen Threadripper PRO 7975WX 32-Cores   Name:                    gfx1201   Marketing Name:          AMD Radeon RX 9070 XT       Name:                    amdgcn-amd-amdhsa--gfx1201       Name:                    amdgcn-amd-amdhsa--gfx12-generic

### ROCm Version

7.11 (top of tree)

### ROCm Component

_No response_

### Steps to Reproduce

Do a build of therock-gmp component and inspect build/third-party/sysdeps/linux/gmp/build/build/s and build/third-party/sysdeps/linux/gmp/source directories for the timestamps.

_No response_

### Additional Information

_No response_

sshi-amd:
## Motivation
cmake -E copy_directory does not preserve file modification timestamps, causing autotools-based build systems to incorrectly detect that generated files (e.g. .info docs) need to be regenerated via makeinfo/gettext, which are not available in all build environments.

## Technical Details

Add build_tools/copy_dir_with_timestamps.py which uses Python's shutil.copytree with copy_function=shutil.copy2 to preserve nanosecond- precision mtime. Replace cmake -E copy_directory with this script in all affected sysdeps CMakeLists.txt files:

- third-party/sysdeps/common/zlib
- third-party/sysdeps/common/liblzma
- third-party/sysdeps/linux/elfutils (also removes stamp-po touch workaround)
- third-party/sysdeps/linux/libdrm
- third-party/sysdeps/linux/libcap
- third-party/sysdeps/linux/numactl

Both THEROCK_SOURCE_DIR and Python3_EXECUTABLE are already injected into all subprojects via THEROCK_DEFAULT_CMAKE_VARS, so no outer CMakeLists.txt changes are needed.

Fixes #2719

## Test Plan                                                                                
   
  Built all 6 affected sysdeps on Ubuntu 22.04, gfx1100 (AMD RX 7900 XTX):                  
    therock-zlib, therock-liblzma, therock-libcap, therock-libdrm,
    therock-numactl, therock-elfutils

  Ran the TheRock ctest validation suite for all affected components.

## Test Result:

  100% tests passed, 0 tests failed out of 10

    11/10 Passed  therock-validate-shared-lib-libcap.so        0.08s
    12/10 Passed  therock-validate-shared-lib-libdrm.so        0.02s
    13/10 Passed  therock-validate-shared-lib-libdrm_amdgpu.so 0.02s
    14/10 Passed  therock-validate-shared-lib-liblzma.so       0.02s
    37/10 Passed  therock-validate-shared-lib-libnuma.so       0.02s
    39/10 Passed  therock-validate-shared-lib-libz.so          0.02s
    40/10 Passed  therock-validate-shared-lib-libzstd.so       0.02s
    41/10 Passed  therock-validate-shared-lib-libelf.so        0.02s
    42/10 Passed  therock-validate-shared-lib-libdw.so         0.02s
    43/10 Passed  therock-validate-shared-lib-libasm.so        0.09s

  No makeinfo/gettext doc regeneration errors in any build log.
  elfutils built cleanly without the stamp-po touch workaround.

## Submission Checklist

### DIFF
--- a/build_tools/copy_dir_with_timestamps.py
+++ b/build_tools/copy_dir_with_timestamps.py
@@ -35,7 +35,7 @@ def main():
     if dst.exists():
         sys.stderr.write(
             f"ERROR: Destination already exists: {dst}\n"
-            f"       Remove it first with `cmake -E rm -rf {dst}`\n"
+            f" Remove it first with `cmake -E rm -rf {dst}`\n"
         )
         sys.exit(1)
 
--- a/third-party/sysdeps/common/expat/CMakeLists.txt
+++ b/third-party/sysdeps/common/expat/CMakeLists.txt
@@ -76,15 +76,19 @@ project(EXPAT_BUILD VERSION "2.7.3")
 set(patch_source_commands)
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   list(APPEND patch_source_commands   COMMAND
-    bash "${CMAKE_CURRENT_SOURCE_DIR}/patch_source.sh" "${SOURCE_DIR}")
+    bash "${CMAKE_CURRENT_SOURCE_DIR}/patch_source.sh" "${CMAKE_CURRENT_BINARY_DIR}/s")
 endif()
 
 add_custom_target(
   build ALL
   WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
-  #
-  # OPTIONAL: Patching of the sources should happen here.
-  #
+  COMMAND
+    "${CMAKE_COMMAND}" -E rm -rf -- "${CMAKE_INSTALL_PREFIX}" "${CMAKE_CURRENT_BINARY_DIR}/s"
+  COMMAND
+    # Use copy_dir_with_timestamps.py instead of `cmake -E copy_directory` to
+    # preserve file modification timestamps, preventing autotools from trying
+    # to regenerate files (which would require additional tools).
+    "${Python3_EXECUTABLE}" "${THEROCK_SOURCE_DIR}/build_tools/copy_dir_with_timestamps.py" "${SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}/s"
   COMMAND
     ${patch_source_commands}
   COMMAND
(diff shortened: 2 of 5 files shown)


lumachad:
I think this also needs to apply for gmp, mpfr, expat and ncurses, as described in #2719. And we should stop using the extracted files directly and move to using their copies.

Issue #3818, opened 2026-03-06
[Feature] gfx90c build enablement


harkgill-amd:
Enable gfx90c (Ryzen 4000/5000/6000 series iGPUs) nightly builds. At minimum, the following is required,

- [x] Addition of gfx90c to https://github.com/ROCm/TheRock/blob/main/cmake/therock_amdgpu_targets.cmake
- [ ] Successful build of TheRock targeting gfx90c
              - Currently failing during rocBLAS/tensile build. With https://github.com/ROCm/rocm-libraries/pull/5282 in place, build passes.

- [ ] gfx90c support enabled in as many libraries as possible. https://github.com/ROCm/TheRock/issues/2310 to be used as reference.
- [ ] Addition of gfx90c to the nightly build scripts, similar to https://github.com/ROCm/TheRock/pull/3564.

malcolmroberts:
Do we have gating runtime/compiler/driver tests for gfx90c?  If not, I'm worried that we will have more "bump" issues where we get a bunch of failures showing up in library tests (and I'm concerned with rocFFT mostly) when there's a change in the stack below the library level.

harkgill-amd:
Once gfx90c is added as a build target in TheRock, we'll have nightly builds that'll be subject to end to end testing. https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md covers the different levels of testing done. All in all, it should be to the same level as what's being done with gfx900 which already exists in most libraries and has nightlies being published.

malcolmroberts:
An alternative solution is to enable spir-v for this (and analogous) architectures.  We have a PR open at https://github.com/ROCm/rocm-libraries/pull/8139 which we are pushing to get through CI.

If we end up adding explicit gfx90c support, we would probably remove it when spir-v was enabled in any case.

Zakhrov:
Any progress on this? I'd really like to use my vega iGPU for overnight runs where I can leverage a big-ass framebuffer compared to what my gfx1010 dGPU can fit into its measly 6GB frambuffer

harkgill-amd:
> Any progress on this? 

Yes! We're opting to go the route of enabling nightly builds first (https://github.com/ROCm/TheRock/pull/6538) and then work backwards to enable libraries for gfx90c. Will share an update here once the first set of nightly wheels are published at which point we can see what works/doesn't and fill the gaps where possible.

malcolmroberts:
> > Any progress on this?
> 
[...]

PR #3821, opened 2026-03-06
Add gfx90c build support

harkgill-amd:
## Motivation

First step in enabling gfx90c https://github.com/ROCm/TheRock/issues/3818.

## Technical Details

Adding in new gfx90c entry that will be apart of igpu-all and gfx90c-igpu families. We could potentially group gfx906, gfx900 and gfx90c into a gfx90X-dgpu (non-dcgpu) family going forward but this should be fine for now.

Following the same excludes as gfx906.

## Test Plan

Building with `-DTHEROCK_AMDGPU_FAMILIES=gfx90c-igpu` in parallel though that shouldn't be a blocker for this.

## Submission Checklist

### DIFF
--- a/cmake/therock_amdgpu_targets.cmake
+++ b/cmake/therock_amdgpu_targets.cmake
@@ -44,6 +44,15 @@ function(therock_add_amdgpu_target gfx_target product_name)
   endforeach()
 endfunction()
 
+therock_add_amdgpu_target(gfx90c "AMD Renoir/Lucienne/Cezanne iGPU" FAMILY igpu-all gfx90c-igpu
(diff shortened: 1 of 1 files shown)


geomin12 (commented):
I added label so gfx90c builds! want to make sure it works, so we can wait for CI to build

ScottTodd:
Adding to `cmake/therock_amdgpu_targets.cmake` is fine even if the build does not yet succeed, fyi.

harkgill-amd:
@geomin12, looks like the CI failures are unrelated to the changes here.

PR #3825, opened 2026-03-06
Build Correctly In Presence of /opt/rocm

linuxrocks123:
Various projects built by TheRock are fascinated with the `/opt/rocm` directory on host systems and try to use it for things.  Ideally, they would lose that fascination.  This PR instead attempts to help work around their fascination by exporting more symbols to control the locations of ROCm tools.

### DIFF
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -7,6 +7,8 @@
 
 include(ExternalProject)
 
+set(ROCM_PATH ${THEROCK_BINARY_DIR}/dist/rocm)
+
 # Global properties.
 # THEROCK_DEFAULT_CMAKE_VARS:
 # List of CMake variables that will be injected by default into the
@@ -23,6 +25,10 @@ set_property(GLOBAL PROPERTY THEROCK_DEFAULT_CMAKE_VARS
   THEROCK_ENABLE_LLVM_TESTS
   LLVM_LIT_ARGS
   THEROCK_USE_SAFE_DEPENDENCY_PROVIDER
+  HIP_DIR
+  HIP_PATH
+  ROCM_DIR
+  ROCM_PATH
   ROCM_SYMLINK_LIBS
 
   # RPATH handling.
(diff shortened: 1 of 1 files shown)


cmake/therock_subproject.cmake
```diff
@@ -10,6 +7,8 @@
 
 include(ExternalProject)
 
+set(ROCM_PATH ${THEROCK_BINARY_DIR}/dist/rocm)
```
ScottTodd, line 10: This won't generally work. Subproject builds are deliberately self-contained and isolated from one another. See https://github.com/ROCm/TheRock/blob/main/docs/development/build_system.md#build-directory-layout

> - `dist/` : This is populated by the `dist` build phase by hard linking (or
>   copying if hard-links are not possible) the cone of all runtime dep project's
>   `stage/` and this project's `stage/` directory contents. In this way, each
>   individual project's `dist/` directory should consist of a self-contained
>   slice that is relocatable and usable (for testing, etc). Top-level
>   distributions are created in this same way by having runtime deps on all
>   relevant sub-projects. Keeping sub-projects isolated in this way aids in
>   testing, especially to ensure that dependencies are declared and layered
>   properly.

Need to fix the subprojects that escape the sandbox and point at system paths.

lumachad: Don't we empty ROCM_PATH, ROCM_DIR, HIP_PATH and HIP_DIR on purpose so subprojects don't rely on those for their configs/builds?

For instance, one of out projects relied on ROCM_PATH for building tests, and I'd like that to fail rather than silently pass, as that gives us the opportunity to fix their behavior.

marbre: We do, for example at
https://github.com/ROCm/TheRock/blob/65c64d0eb2002315cc765d5bb91033afc4c7fd80/math-libs/CMakeLists.txt#L15-L17

but we might miss some places.

stellaraccident:
We should remove the "fascination" and verify with strace -- even at the expense of breaking legacy CI systems.

PR #3828, opened 2026-03-06
[ci] Collect and report failed tests for `ctest` and `gtest` components

### LINKED ISSUE #3492: [TheRock CI] reports for PR merge results need some processing
Merged PR produces mail with failure message.
All visible tests are pass, need to drill down to find what failed.
Upstream failures on hipblaslt for both gfx:
https://github.com/ROCm/rocm-libraries/actions/runs/18014289860/job/512651538

Needs to report downstream as failures to review separately, there is no need diagnose issues that can't be caused by a downstream merge.
If you must report upstream put them in a different section.

Separate note, test logs for hipblaslt are unmanagably large.   
Please use GTEST_FILTER=NO_PASS_LINE_IN_LOG or ask them to do this by default.
If you have over a million tests we need to cut to the failures by default.

geomin12:
We want to collect and report failed tests among all components. This will be the first PR towards automating the failed test reporting.

Changes: 
- wrapper around each ctest/gtest to report failures to console output
- adding ctest/gtest output reporting

Ideally, we would:

1. collect failed tests (this PR)
2. turn off collection of good tests and only collect failed tests (?)
3. add code generation to ignore tests for each component
4. add failed tests to github issue and report to team

Adding `skip-ci` label as this is purely test-related. Will run local workflow dispatches to test

Closes #3492 

Tests ran here: https://github.com/ROCm/TheRock/actions/runs/22871406535
Works here: https://github.com/ROCm/TheRock/actions/runs/22871406535/job/66357078995


amd-shiraz (commented):
Initial thoughts. (not revieweing code yet) @geomin12  as discussed offline, our goal should be to get as much reporting and cleaning of the mess on component test side outside of therock as possible. tools we created like https://github.com/ROCm/actions-orchestrator will help with that where we can directly work with component teams on test issues separately Vs in flight integrated with therock as that causes production CI disruption.  

This is fine on post integration reporting side for nightlies perhaps. but the reproduction, optimization, fixing etc. should happen away from therock as much as possible. I know the question of "but this works on my machine" will come up, but that is where we will have BYOM (bring your own machine) model for component teams. and what we provide is tooling to discover, profile and optimize on tests. cc @suchitravemuri1 @amd-aakash @scottt  @jayhawk-commits

jayhawk-commits:
Please see #3907 on moving the project/component test scripts out of TheRock into the super-repos.

I see a lot of repeated code changes across the individual test scripts we would want to move out, suggesting there is potential for better design.

One suggestion would be if we limited the enhancement to common functions in TheRock that auto-detects if the test command does support an XML output format, appends the corresponding parameters to the execution, handles the uploading of the test logs to the correct buckets, and an option for developers to download the test logs from a specific run similar to how they have the means to fetch build artifacts from a run.

The individual test scripts could leverage these common functions.

geomin12:
> Initial thoughts. (not revieweing code yet) @geomin12 as discussed offline, our goal should be to get as much reporting and cleaning of the mess on component test side outside of therock as possible. tools we created like https://github.com/ROCm/actions-orchestrator will help with that where we can directly work with component teams on test issues separately Vs in flight integrated with therock as that causes production CI disruption.
> 
> This is fine on post integration reporting side for nightlies perhaps. but the reproduction, optimization, fixing etc. should happen away from therock as much as possible. I know the question of "but this works on my machine" will come up, but that is where we will have BYOM (bring your own machine) model for component teams. and what we provide is tooling to discover, profile and optimize on tests. cc @suchitravemuri1 @amd-aakash @scottt @jayhawk-commits

indeed, this makes sense. For the following action items from these tests, I will execute in CI nightly, however, this will still help developers when trying to identify tests! (resolving issue #3492)

After this PR, this extended actions-orchestrator / work will go with CI nightly

geomin12:
> Please see #3907 on moving the project/component test scripts out of TheRock into the super-repos.
> 
> I see a lot of repeated code changes across the individual test scripts we would want to move out, suggesting there is potential for better design.
> 
> One suggestion would be if we limited the enhancement to common functions in TheRock that auto-detects if the test command does support an XML output format, appends the corresponding parameters to the execution, handles the uploading of the test logs to the correct buckets, and an option for developers to download the test logs from a specific run similar to how they have the means to fetch build artifacts from a run.
> 
> The individual test scripts could leverage these common functions.

#3907 looks great, I 100% agree as ownership would fall towards component teams + easier to manage.

I think this can land for now as this provides relief on #3492 then we can work removing dup-ed code as we work on #3907 and @amd-shiraz 's actions-orchestrator work

geomin12:
Closing this PR in lieu of actions-orchestrator work which handles this collection of data and reports to component teams.

actions-orchestrator will be integrated into CI nightly and eventually handle CI -> green signal motivations

PR #3648, opened 2026-02-26
extended_tests: Add functional test base class and MIOpen convolution test

lajagapp:
## Summary
Part 2 of 3 — Split from #3533
This PR adds the functional test framework base class and MIOpenDriver convolution test to the extended_tests framework.

### New Functional Testing Framework
- **functional_base.py: Base class providing common functionality for functional tests:**
  - JSON configuration loading from configs/ directory
  - Test result parsing and statistics calculation (pass/fail/error counts)
  - GitHub Actions step summary integration
  - ExtendedTestClient integration for system detection and result upload
  - Standardized error handling with TestExecutionError and TestResultError
  - Sharding support for parallelized test execution

- **MIOpenDriver Convolution Test**
  - test_miopendriver_conv.py: Functional test verifying MIOpenDriver convolution operations
  - miopendriver_conv.json: Test configuration with convolution commands and GPU-specific settings
- **Documentation**
  - functional/README.md: Complete documentation for functional testing framework

### Full series
PR 1: #3646 — Rename BenchmarkClient → ExtendedTestClient + add TestResultError
→ PR 2 (this): #3648 — Add functional test framework base class and MIOpen convolution test
PR 3: #3649 — Add functional test CI workflow infrastructure

PR #3829, opened 2026-03-07
Adding code obj version backward comptability test

dkrottap:
adding a code object version backward compatibility test that builds  for default, n-1, and n-2 versions and verifies binaries execute successfully


tests/extended_tests/functional/scripts/functional_base.py
geomin12, line 1: duplicate of #3648, will we land #3648 first then rebase this?

tests/extended_tests/functional/functional_test_matrix.py
geomin12, line 1: duplicate of #3648, same as other comment

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+        if not self.include_dir.exists():
+            raise TestExecutionError(f"Include path not found: {self.include_dir}")
+
+    def _resolve_tool(self, candidates: List[Path], default_tool: str) -> str:
+        for candidate in candidates:
+            if candidate.exists():
+                return str(candidate)
+        return default_tool
+
+    def _compile_sample(self, output_binary: Path, code_obj_version: int = None) -> int:
```
geomin12, line 104: how long does this compilation take?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+
+        def _summarize_output(output: str, max_lines: int = 6) -> str:
+            lines = [line.strip() for line in output.splitlines() if line.strip()]
+            if not lines:
+                return "no output"
+            if len(lines) <= max_lines:
+                return " | ".join(lines)
+            return " | ".join(lines[:max_lines]) + " | ..."
+
+        def abi_to_code_object_version(abi_version: int, source_desc: str) -> int:
```
geomin12: nit:


```suggestion
        def _abi_to_code_object_version(abi_version: int, source_desc: str) -> int:
```

tests/extended_tests/functional/scripts/test_code_obj_version.py
geomin12: we should generally try to avoid building in the test stage (meaning we shouldn't waste GPU resources on building)

can this be added in the build step?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+
+    def _run_and_validate_binary(self, binary_path: Path) -> Tuple[int, str, str]:
+        env = self._get_rocm_env_with_path()
+        rc, output = self._execute_command_with_output(
+            [str(binary_path)], cwd=self.sample_dir, env=env
+        )
+        if rc != 0:
+            return rc, output, f"Binary execution failed with return code {rc}"
+        if "PASSED!" not in output:
```
geomin12, line 238: not a fan of using `PASSED!` as the check. does the executable not return 0 / 1 if pass / fail?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def run_tests(self) -> None:
+        log.info(f"Running {self.display_name} Tests")
+        self._ensure_sources_ready()
+
+        # 1) Build with default compiler behavior (no explicit code object version),
+        #    then detect current code object version n from the produced binary.
+        default_variant = "default_build_detect_n"
+        default_compile_cmd = f"hipcc {self.source_file} -I {self.include_dir}"
+        default_binary = self.sample_dir / f"{self.binary_prefix}_default"
```
geomin12, line 277: this is fine if it take a few seconds, but if longer, we should add this to build step

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+        rc, output = self._execute_command_with_output(
+            [str(binary_path)], cwd=self.sample_dir, env=env
+        )
+        if rc != 0:
+            return rc, output, f"Binary execution failed with return code {rc}"
+        if "PASSED!" not in output:
+            return rc, output, "Binary did not print expected token: PASSED!"
+        return rc, output, ""
+
+    def _record_result(
```
geomin12, line 242: can we just log out the results? no need to report to a JSON in my opinion

it seems as it is just reporting to a JSON file just to re-open and log the contents of the JSON files, we should just log out to console if this is the case

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+                self.create_test_result(
+                    test_name=self.test_name,
+                    subtest_name=result["test_case"],
+                    status=status,
+                    suite=result["test_suite"],
+                    command=result.get("command", ""),
+                )
+            )
+
+        return parsed_results
```
geomin12, line 407: as mentioned earlier, we can just log out the results instead of writing to JSON / reopening file to log out contents.

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: The specific purpose of this test is to test the backward compatibility of code objects which requires building for the supported code objects. Regarding the compilation, hipcc does not use GPU resources and compilation time is also quite less.

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: done

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: this is a tests from hip-samples so checking the o/p from the same.

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: done removed the redundant json file

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: This shouldn't take much time and it does not use any gpu resources

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: removed redundant json file

tests/extended_tests/functional/scripts/functional_base.py
dkrottap: agreed added it as earlier patches has not yet landed. removing both base and test matrix files. Once 3648 lands will rebase and apply this patch.

tests/extended_tests/functional/functional_test_matrix.py
dkrottap: removing it for now. Please find test results
## Cov Backward Comptability - Functional Test Results

```
+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+
|  Total TestSuites |  Total TestCases  |       Passed      |       Failed      |      Errored      |      Skipped      |    Final Result   |
+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+
|         1         |         3         |         3         |         0         |         0         |         0         |        PASS       |
+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+-------------------+

dkrottap:
branch is corrupted will create a new PR

Issue #3796, opened 2026-03-05
Fix overlapping files in artifacts and enforce correctness through automated tests


ScottTodd:
We've spotted a few issues recently where multiple artifacts contain overlapping files:
* https://github.com/ROCm/TheRock/issues/2784
* https://github.com/ROCm/TheRock/issues/3770
* https://github.com/ROCm/TheRock/issues/3758

We're fixing the few that we spot, but we can go a few steps further:
1. Update our docs at https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md (cover some of the trickier details and suggest best practices)
2. Add test coverage ensuring that there are no overlapping files
    * We can also test for some other properties of artifacts

ScottTodd:
Summarized notes from my local debugging/working sessions: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/artifact-overlap.md

PR #3802, opened 2026-03-06
Introduce tests for artifact structure

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/3796. This adds a more comprehensive set of tests using the actual built artifacts, unlike the unit test in https://github.com/ROCm/TheRock/pull/3765 that just does static analysis of the artifact descriptors (toml files).

> [!IMPORTANT]
> The tests are currently FAILING, so I didn't hook them up to any workflows yet.
>
> Proposed plan:
> 1. (This PR) Land the tests, runnable via workflow_dispatch
> 2. Fix the artifact definitions / population code to the tests start passing
> 3. Integrate the tests into existing CI workflows

Running these tests continuously will guard against future regressions that would otherwise go unnoticed or be difficult to debug (race conditions!).

## Technical Details

Files in artifacts should be unique across all artifacts. If they are not, we use extra storage and can run into race conditions during artifact extraction/flattening.

This adds two key tests:
1. Cross-artifact collision testing (e.g. 'blas' vs 'support')
2. Within-artifact collision testing (e.g. 'blas_dev' vs 'blas_test')

## Test Plan

* Fetch various sets of artifacts with `fetch_artifacts.py` locally and run the tests locally
    ```bash
    set THEROCK_ARTIFACTS_DIR=D:/scratch/claude/artifacts/22681848469-linux-gfx94X-dcgpu
    python -m pytest tests/test_artifact_structure.py -v
    ```
* Triggered test runs with workflow_dispatch in my fork

## Test Result

| Run | Format | Fetch | Validate | Result |
|---|---|---|---|---|
| [22743296445](https://github.com/ScottTodd/TheRock/actions/runs/22743296445) | classic `.tar.xz` | 74s | 169s | 2 FAILED |
| [22743345907](https://github.com/ScottTodd/TheRock/actions/runs/22743345907) | kpack `.tar.zst` | 12s | 18s | 2 FAILED | 

> [!NOTE]
> * These are on github-hosted runners which may have slower network access to S3 than our eventual S3 hosted runners
> * The multi-arch tests using .zst are much faster than the classic tests using .xz

Cross-artifact logs snippet:
```
E           Failed: Found 29 cross-artifact collision(s) across 193 archives (see https://github.com/ROCm/TheRock/issues/3796):
E             include/mxDataGenerator/DataGenerator.hpp
E               - blas (blas_dev_gfx94X-dcgpu.tar.xz)
E               - support (support_dev_generic.tar.xz)
[...]

PR #1483, opened 2025-09-13
Refactor artifact builder to make it more strict.

stellaraccident:
We've been getting reports of some files missing from distribution artifacts (notably for rocprofiler, which added some additional python files recently).

Since the tool that does artifact building was the first thing written 9 months ago, and since it was a bit of a pragmatic attempt, I took the chance to rewrite it so that I could enforce stricter invariants and catch more issues at build time.

Functional changes:

* Components now "extend" another component and a file can only match the deepest component in its hierarchy. Only unmatched files are eligible for child components. This replaces the carefully balanced include/exclude defaults which attempted to ensure that each component's patterns represented a mutually exclusive set. Now mutual exclusion is by construction.
* Prior, the build scripts would run the artifact builder once per component, but now all components in a descriptor are processed in one invocation. This allows us to form a global view of what goes where once and then do the populating all in one go.
* The above enables stricter verification options. Particularly, unmatched files which are not included in any component are now tracked. By default, any unmatched component will produce a build error. The error message contains the changes needed to declare them. All files must either be matched by a component or declared in the descriptor options to be ok to ignore.
* The tighter checks have revealed a number of descriptors which were leaving some of their contents out. These have been fixed (and more that the CI surfaces will be fixed prior to landing).

Other changes:

* Complete unit tests have been implemented for the artifact builder. This was being tested only by a couple of very course integration tests.
* The artifact cmake scripts was inadvertently invoking the flatten tool multiple times as directories were accumulated, with each invocation adding one more directory to the list of things to flatten. This was due to improper nesting in a foreach and has been reduced to one invocation.
* Strict verification of all descriptor fields and tables was added.

Fixes https://github.com/ROCm/rocm-libraries/issues/1622

PR #3830, opened 2026-03-07
Fix overlapping artifacts across the project

ScottTodd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/3796. Distinct artifact archives contained files sharing flattened paths, causing race conditions during artifact extraction as part of build and test workflows.

Tentative fix for https://github.com/ROCm/TheRock/issues/2784 too.

## Technical Details

This fixes all instances of overlapping artifacts across the project by
1. Adding `ComponentDefaults("test", extends=["doc"])` to `build_tools/_therock_utils/artifact_builder.py` so the "test" artifact component participates in the artifact extends chain (`lib -> run -> dbg -> dev -> doc -> test`). Now, only files that are not included in other artifact components (or are explicitly re-included) will appear in the "test" artifact component.
2. Fixing `artifact-*.toml` files that were too loose with catch-all "run" or "dbg" groups so they filter out files that are expected to go in the "test" artifact component instead
3. Fixing `build_tools/install_rocm_from_artifacts.py` to fetch additional "run" artifacts with clear rationale where needed
4. Updating the documentation so these aspects of the build system are easier to understand
5. Wiring up the `test_artifacts_structure.yml` workflow into the CI and Multi-Arch CI workflows so we can watch for regressions

Detailed notes on each artifact fix are available at https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/artifact-overlap.md#descriptor-fix-details. Some of these may need some deeper discussion with component teams to be sure that the setup is still correct.

See also:
* https://github.com/ROCm/TheRock/pull/3802
* https://github.com/ROCm/TheRock/pull/3727
* https://github.com/ROCm/TheRock/pull/3793

### Alternatives Considered

We could keep "test" as a separate artifact component, decoupled from "lib", "run", "dbg", "dev", and "doc" and treat test artifacts having duplicate files as acceptable. I think we'd still run into race conditions in that case.

From reading https://github.com/ROCm/TheRock/pull/1483:
> Components now "extend" another component and a file can only match the deepest component in its hierarchy. Only unmatched files are eligible for child components. This replaces the carefully balanced include/exclude defaults which attempted to ensure that each component's patterns represented a mutually exclusive set. Now mutual exclusion is by construction.

I think having "test" be separate was an accidental omission, not a conscious design choice.

## Test Plan

* Baseline showing overlapping files: https://github.com/ROCm/TheRock/actions/runs/22783664921/job/66095389040
    ```
      ┌─────────────────────┬─────────────┬───────────────────┐
      │      Artifact       │ Components  │ Overlapping files │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocprofiler-compute │ lib vs test │ 4,139             │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocprofiler-sdk     │ run vs test │ 452               │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocprofiler-sdk     │ lib vs test │ 10                │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ base                │ run vs test │ 33                │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocprofiler-compute │ run vs test │ 22                │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocrtst             │ run vs test │ 12                │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rdc                 │ lib vs test │ 4                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocdecode           │ dev vs test │ 3                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocgdb              │ dev vs test │ 2                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ core-ocl            │ lib vs test │ 2                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ blas                │ dev vs test │ 1                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ miopen              │ run vs test │ 1                 │
      ├─────────────────────┼─────────────┼───────────────────┤
      │ rocprofiler-compute │ doc vs test │ 1                 │
      └─────────────────────┴─────────────┴───────────────────┘
    ```
* Workflow runs during development showing test failures without artifact tweaks:
  * CI: https://github.com/ROCm/TheRock/actions/runs/22782491222
  * Multi-Arch CI: https://github.com/ROCm/TheRock/actions/runs/22782514106
* CI on this PR
  * Existing ROCm subproject unit tests should remain passing
  * New artifact structure unit tests should start passing

## Test Result

* Existing project CI tests all passed
* New artifact structure tests are newly passing: https://github.com/ROCm/TheRock/actions/runs/22787970447/job/66116109177?pr=3830
    Linux gfx94X:
    ```
    INFO     test_artifact_structure:test_artifact_structure.py:251 Checked 38887 unique paths across 204 archives, no cross-artifact overlaps
    PASSED                                                                   [ 50%]
    tests/test_artifact_structure.py::TestArtifactStructure::test_no_within_artifact_component_overlaps 
    -------------------------------- live log call ---------------------------------
    INFO     test_artifact_structure:test_artifact_structure.py:308 Checked 48 artifacts, no within-artifact component overlaps
    PASSED                                                                   [100%]
    ======================== 2 passed in 168.60s (0:02:48) =========================
    ```
    Windows gfx1151:
    ```
    INFO     test_artifact_structure:test_artifact_structure.py:251 Checked 10150 unique paths across 92 archives, no cross-artifact overlaps
    PASSED                                                                   [ 50%]
    tests/test_artifact_structure.py::TestArtifactStructure::test_no_within_artifact_component_overlaps 
    -------------------------------- live log call ---------------------------------
    INFO     test_artifact_structure:test_artifact_structure.py:308 Checked 24 artifacts, no within-artifact component overlaps
    PASSED                                                                   [100%]
    ======================== 2 passed in 104.73s (0:01:44) =========================
    ```

## Submission Checklist

### DIFF
--- a/debug-tools/artifact-rocgdb.toml
+++ b/debug-tools/artifact-rocgdb.toml
@@ -1,6 +1,9 @@
 # rocgdb
 [components.dbg."debug-tools/rocgdb/stage"]
 [components.dev."debug-tools/rocgdb/stage"]
+exclude = [
+  "tests/**",
+]
 [components.doc."debug-tools/rocgdb/stage"]
 include = [
   "share/doc/**",
--- a/math-libs/BLAS/artifact-blas.toml
+++ b/math-libs/BLAS/artifact-blas.toml
@@ -187,6 +187,9 @@ optional = true
 include = [
   "**/*.cmake"
 ]
+exclude = [
+  "share/hipsparse/test/**",
+]
 [components.doc."math-libs/BLAS/hipSPARSE/stage"]
 optional = true
 [components.lib."math-libs/BLAS/hipSPARSE/stage"]
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -371,15 +371,11 @@ def retrieve_artifacts_by_run_id(args):
         if args.blas:
             extra_artifacts.append("blas")
         if args.debug_tools:
-            # Add extra artifacts so we generate _lib and _test artifact
-            # entries later.
             extra_artifacts.append("amd-dbgapi")
             extra_artifacts.append("rocgdb")
             extra_artifacts.append("rocr-debug-agent")
             extra_artifacts.append("rocr-debug-agent-tests")
-
-            # Add the rest of the artifacts not handled automatically (non-lib
-            # and non-test).
+            # Contains the rocgdb executable.
             argv.append("rocgdb_run")
 
             # Libraries rocgdb depends on.
@@ -396,7 +392,7 @@ def retrieve_artifacts_by_run_id(args):
             extra_artifacts.append("hipdnn-samples")
         if args.miopen:
             extra_artifacts.append("miopen")
-            # We need bin/MIOpenDriver executable for tests.
+            # Contains bin/MIOpenDriver executable for tests.
             argv.append("miopen_run")
             # Also need these for runtime kernel compilation (rocrand includes).
             argv.append("rand_dev")
@@ -435,6 +431,8 @@ def retrieve_artifacts_by_run_id(args):
             argv.append("rocprofiler-sdk_run")
         if args.rocprofiler_compute:
             extra_artifacts.append("rocprofiler-compute")
+            # Contains the rocprof-compute CLI executable.
+            argv.append("rocprofiler-compute_run")
         if args.rocprofiler_systems:
             extra_artifacts.append("rocprofiler-systems")
             # Contains executables (rocprof-sys-run, rocprof-sys-instrument, etc.)
@@ -453,6 +451,10 @@ def retrieve_artifacts_by_run_id(args):
             argv.append("amd-llvm_lib")
             argv.append("base_dev_generic")
 
+        # Fetch _lib (always) and _test (when --tests) for each artifact.
+        # Some projects have self-contained _test archives (just test
+        # binaries), while others may also need executables or data from
+        # _run. Add those explicitly above via argv.append("<name>_run").
         extra_artifact_patterns = [f"{a}_lib" for a in extra_artifacts]
         if args.tests:
             extra_artifact_patterns.extend([f"{a}_test" for a in extra_artifacts])


ml-libs/artifact-miopen.toml
```diff
@@ -11,3 +11,6 @@ include = [
   "bin/miopen_gtest*",
 ]
 [components.run."ml-libs/MIOpen/stage"]
+exclude = [
+  "bin/miopen_gtest*",
+]
```
ScottTodd, line 16: This `miopen_gtest` binary was in both `_run` and `_test`. Now it will just be in `_test`.

profiler/artifact-aqlprofile.toml
```diff
@@ -4,6 +4,9 @@
 [components.doc."profiler/aqlprofile/stage"]
 [components.lib."profiler/aqlprofile/stage"]
 [components.run."profiler/aqlprofile/stage"]
+exclude = [
+  "share/hsa-amd-aqlprofile/**",
+]
```
ScottTodd, line 9: These are test files like `gfx90a_simple_convolution.hsaco`, so I'm excluding them from the 'run' component and letting them just be in the 'test' component only.

math-libs/BLAS/artifact-blas.toml
```diff
@@ -187,6 +187,9 @@ optional = true
 include = [
   "**/*.cmake"
 ]
+exclude = [
+  "share/hipsparse/test/**",
+]
```
ScottTodd, line 192: These files were in both 'dev' and 'test'. Now they'll just be in 'test'.

profiler/artifact-rocprofiler-sdk.toml
```diff
@@ -17,6 +17,9 @@ include = [
   "share/rocprofiler-sdk-roctx/**",
   "lib/python*/**",
 ]
+exclude = [
+  "share/rocprofiler-sdk/tests/**",
+]
```
ScottTodd, line 22: These files were in 'run' and 'test'. Now they'll just be in 'test'

debug-tools/artifact-rocgdb.toml
```diff
@@ -1,6 +1,9 @@
 # rocgdb
 [components.dbg."debug-tools/rocgdb/stage"]
 [components.dev."debug-tools/rocgdb/stage"]
+exclude = [
+  "tests/**",
+]
```
ScottTodd, line 6: Moving 'test' files from 'dev' to 'test'

core/artifact-core-rocrtst.toml
```diff
@@ -3,6 +3,9 @@
 [components.dev."core/rocrtst/stage"]
 [components.doc."core/rocrtst/stage"]
 [components.run."core/rocrtst/stage"]
+exclude = [
+  "bin/**",
+]
```
ScottTodd, line 8: test binaries like `bin/gfx*/rocrtst`, moving from 'run' to 'test'

base/artifact.toml
```diff
@@ -45,6 +45,9 @@ optional = "windows"
 optional = "windows"
 [components.run."base/rocprofiler-register/stage"]
 optional = "windows"
+exclude = [
+  "share/rocprofiler-register/tests/**",
+]
```
ScottTodd, line 50: moving test files from 'run' to 'test'

build_tools/_therock_utils/artifact_builder.py
```diff
@@ -83,6 +89,12 @@ def get(name: str) -> "ComponentDefaults":
 )
 ComponentDefaults("doc", includes=["**/share/doc/**"], extends=["dev"])
 
+# Test components extend the full chain (lib → run → dbg → dev → doc → test)
+# so they automatically skip files claimed by earlier components. Descriptors
+# that want specific files in test (instead of run) must exclude those files
+# from run — see docs/development/artifacts.md for details.
+ComponentDefaults("test", extends=["doc"])
```
ScottTodd, line 96: > ### Alternatives Considered
> 
> We could keep "test" as a separate artifact component, decoupled from "lib", "run", "dbg", "dev", and "doc" and treat test artifacts having duplicate files as acceptable. I think we'd still run into race conditions in that case.
> 
> From reading #1483:
> 
> > Components now "extend" another component and a file can only match the deepest component in its hierarchy. Only unmatched files are eligible for child components. This replaces the carefully balanced include/exclude defaults which attempted to ensure that each component's patterns represented a mutually exclusive set. Now mutual exclusion is by construction.
> 
> I think having "test" be separate was an accidental omission, not a conscious design choice.

@stellaraccident You might want to review this part of the change since you added the original component "extends" code on that linked PR.

stellaraccident:
Thanks for this. Did we add any errors if it regresses? This is like round three of hardening it -- hopefully the last if we make it hard fail if split this way again

ScottTodd:
> Thanks for this. Did we add any errors if it regresses? This is like round three of hardening it -- hopefully the last if we make it hard fail if split this way again

Yep. https://github.com/ROCm/TheRock/pull/3802 added https://github.com/ROCm/TheRock/blob/main/tests/test_artifact_structure.py. The tests there are newly passing with these changes and will run on every PR now. That checks that the _artifacts_ are structurally sound (no duplicates at all). It does _not_ check that project tests behave as expected, for example if a test depends on files coming from the 'run' artifact.

ScottTodd:
Pulling out the relevant parts of the PR description:
* Baseline failing test artifact structure: https://github.com/ROCm/TheRock/actions/runs/22783664921/job/66095389040
* Newly passing test artifact structure: https://github.com/ROCm/TheRock/actions/runs/22787970447/job/66112863462?pr=3830

build_tools/_therock_utils/artifact_builder.py
marbre: For many cases test depend on lib and thus I don't think test should have duplicate files. Furthermore, I don't think having test was an accidental omission (if I understand the phrasing right). Where should test artifacts go otherwise? I think we definitely want to be able to split out test artifacts even though these go in to rocm-sdk-core right now.

core/artifact-core-rocrtst.toml
marbre: Should `rocrtst_run_generic/core/rocrtst/stage/bin/rocrtst64` go to `test` as well?

ml-libs/artifact-miopen.toml
marbre: I wish there would be a way to not need to exclude this if included in an previous defined stage.

marbre (commented):
LGTM but would feel more confident if Stella approves :) Let me know if I should stamp.

ml-libs/artifact-miopen.toml
ScottTodd: The order written in these files is irrelevant. The actual construction order is fixed: `lib -> run -> dbg -> dev -> doc -> test`. I may send a second PR that changes artifacts to follow that.

The issue here is that `run` has no explicit includes, so if it is added, it acts as a catch-all. See the new recommendations in `artifacts.md`.

build_tools/_therock_utils/artifact_builder.py
ScottTodd: > For many cases test depend on lib and thus I don't think test should have duplicate files.

There was nothing enforcing that before, it relied entirely on explicit include/exclude patterns. Now there are no duplicates by default and you have to try hard with explicit includes to get duplicates.

Test artifacts still contain test files, that's not changing.

Running tests can still depend on more than the "test" artifact, as is the case for a few subprojects where the tests need binaries in "run", for example. For some subprojects (e.g. blas), the "test" artifact is actually a self-contained collection of just `foo_test.exe` binaries, no extra subproject deps.

core/artifact-core-rocrtst.toml
ScottTodd: I think so. The whole project is a test project (source is https://github.com/ROCm/rocm-systems/tree/develop/projects/rocr-runtime/rocrtst/suites/test_common). A better fix here may be to remove the "run" component (it acts as a catch-all if included, hence the need for an exclude here).

ml-libs/artifact-miopen.toml
marbre: Yeah, I know :) But if there are explicit includes in other stages, the fetch all could exclude those. Nothing to address as part of this PR.

ScottTodd: Ah that should be possible. Slightly deeper change that what I'd like to tackle here though.

ScottTodd:
Might do another rev after this, adding earlier unit tests for overlapping includes/excludes within a toml file.

PR #3835, opened 2026-03-07
Fix checkout repo in reusable JAX wheel workflow

araravik-psd:
Ensure the reusable JAX wheel build workflow checks out the correct repository when invoked from wrapper repos (e.g. rockrel). Previously, actions/checkout defaulted to github.repository, causing the caller repo to be checked out instead of ROCm/TheRock and resulting in missing build scripts.

The checkout step now explicitly uses the repository and ref inputs, matching the behavior used in the PyTorch wheel workflow.


araravik-psd:
Passing: Rockrel Test workflow run with branch users/arravikum/JAX_build_release_changes pinned

https://github.com/ROCm/rockrel/actions/runs/22805163609/job/66152996282

PR #3838, opened 2026-03-08
Fix OpenMP -isystem clang resource dir breaking cstdint across math libs

powderluv:
## Summary

- When a system ROCm install exists at `/opt/rocm`, the OpenMP config-mode cmake (`openmp-config.cmake`) adds the clang resource dir as `INTERFACE_INCLUDE_DIRECTORIES` on `OpenMP::omp`. This injects `-isystem .../clang/22/include`, causing clang's `stdint.h` to be found before the system one, which breaks GCC's `<cstdint>` with `no member named 'int_fast8_t'` errors.
- Fix by pre-creating `OpenMP::omp`, `OpenMP::OpenMP_CXX`, and `OpenMP::OpenMP_C` imported targets in pre_hooks so `openmpTargets.cmake`'s guard returns early, skipping the problematic include directory.
- Affected subprojects: rocBLAS, rocSPARSE, hipBLAS (all client builds)

## Changes

- **`pre_hook_rocBLAS.cmake`** — Upgraded from cache-var-only workaround to full OpenMP target pre-creation
- **`pre_hook_rocSPARSE.cmake`** — New pre_hook with same OpenMP target pre-creation
- **`pre_hook_hipBLAS.cmake`** — New pre_hook with same OpenMP target pre-creation  
- **`CMakeLists.txt`** — Added `-DOpenMP_DIR` pointing to bundled LLVM's openmp cmake config for rocSPARSE and hipBLAS

## Test plan

- [x] Full TheRock build with `gfx120X-all` on Linux with system ROCm at `/opt/rocm` — all subprojects pass
- [ ] CI validation


stellaraccident (changes requested):
While this gets the job done, I think the root cause here is that the amd-llvm project isn't marked as exporting openmp. This is then causing resolution to fall back to a variety of system search heuristics.

Can you file this as an issue with the expected behavior and listing this as a workaround and I'll get an agent to fix it correctly.

powderluv:
Thank you. I opened https://github.com/ROCm/TheRock/issues/3839

PR #3689, opened 2026-02-28
Add ExtendedTestBase shared base class for benchmark and functional tests

lajagapp:
### Summary
Extracts shared test infrastructure from `BenchmarkBase` into a new `ExtendedTestBase` base class, reducing code duplication between benchmark and functional tests.
This addresses the reviewer feedback on #3648 about sharing common logic between `benchmark_base.py` and `functional_base.py`.

#### New: `utils/extended_test_base.py`
- Created ExtendedTestBase class with following shared infrastructure:
execute_command(), get_gpu_architecture(), detect_gpu_count(), create_test_result(), calculate_statistics(), upload_results()

#### Updated: `benchmark/scripts/benchmark_base.py`
- BenchmarkBase now inherits from ExtendedTestBase
- Removed duplicated methods: execute_command, _detect_gpu_count, calculate_statistics, upload_results

#### Updated: `benchmark/scripts/test_rccl_benchmark.py`
- self._detect_gpu_count() → self.detect_gpu_count() (now inherited from base)

#### Updated: `utils/__init__.py`, `README.md`, `utils/README.md`
- Added ExtendedTestBase to exports and documentation

#### Inheritance Hierarchy
```
ExtendedTestBase (utils/extended_test_base.py)
├── BenchmarkBase (benchmark/scripts/benchmark_base.py)
│   ├── ROCfftBenchmark, RCCLBenchmark, ROCblasBenchmark, ...
└── FunctionalBase (functional/scripts/functional_base.py)  ← will inherit in follow-up PR
    ├── MIOpenDriverConv, RcclTestInfra, ...
```

#### Follow-up
- Update FunctionalBase to inherit from ExtendedTestBase (after this PR merges)

PR #3840, opened 2026-03-09
Add new HeCBench SPIR-V test harness

sriasapu:
**Description**
This PR introduces a new executable test script, test_hecbench_spirv.py, to run HeCBench HIP benchmarks with SPIR-V support in CI. The script automates cloning, building, and running benchmarks, applies Makefile patches for SPIR-V compilation, and produces structured results for CI integration.

**Key Features**
- Benchmark automation: Clones HeCBench and runs HIP benchmarks.
- SPIR-V support: Injects --offload-arch=amdgcnspirv into HIP builds.
- ROCm path flexibility: Replaces hardcoded /opt/rocm with detected ROCM_PATH.
- Timeout handling: Configurable compile/run timeouts via environment variables.
- Error logging: Captures and prints full stdout/stderr for failures.
- Assertion detection: Flags device‑side assertions as explicit failures.
- Results summary: Outputs JSON with totals, passed, failed, timed out, and assertion failures.

**Benefits**
- Provides a new standardized test harness for HeCBench SPIR-V benchmarks.
- Improves diagnostics with clear logging and structured results.
- Ensures compatibility with non‑default ROCm installations.


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -448,6 +448,18 @@ def _get_script_path(script_name: str) -> str:
             "windows": 1,
         },
     },
+    # hecbench SPIR-V tests
+    "hecbench_spirv": {
+        "job_name": "hecbench_spirv",
+        "fetch_artifact_args": "--tests",
```
geomin12: can we use `--hecbench_spirv` flag too? otherwise, this test install step will take a while and also install unnecessary artifacts

example here: https://github.com/ROCm/TheRock/commit/5f60db78b53ca25f8bd6299d31f29a8b13791ed9#diff-f827cc49787b6ff82e268d612f178774273ae8c2955e4e75364c654ee6c19301

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
geomin12, line 1: if this is a benchmark test, this should belong under https://github.com/ROCm/TheRock/tree/main/tests/extended_tests/benchmark

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -448,6 +448,18 @@ def _get_script_path(script_name: str) -> str:
             "windows": 1,
         },
     },
+    # hecbench SPIR-V tests
```
geomin12: this should belong under https://github.com/ROCm/TheRock/blob/main/tests/extended_tests/benchmark/benchmark_test_matrix.py

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,266 @@
...
+
+HECBENCH_BENCHMARKS = os.getenv("HECBENCH_BENCHMARKS", "")
+
+COMPILE_TIMEOUT = int(os.getenv("HECBENCH_COMPILE_TIMEOUT", "300"))
+RUN_TIMEOUT = int(os.getenv("HECBENCH_RUN_TIMEOUT", "300"))
+
+logging.basicConfig(level=logging.INFO)
+
+
+def _run(
```
geomin12: can we re-use #3689 once landed

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,266 @@
...
+
+            _run(["make", "clean"], cwd=bench_dir, env=env)
+
+            try:
+                compile_proc = _run(
+                    ["make", "-j2"],
+                    cwd=bench_dir,
+                    env=env,
+                    timeout=COMPILE_TIMEOUT,
+                )
```
geomin12: all building should belong in build step. tests should only be dedicated to testing (to avoid wasting GPU resources)

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,266 @@
...
+def main() -> None:
+    env = _setup_env()
+    rocm_path = Path(env["ROCM_PATH"])
+
+    with tempfile.TemporaryDirectory(prefix="hecbench_spirv_") as tmp:
+        tmp_path = Path(tmp)
+        clone_cmd = [
+            "git",
+            "clone",
+            "https://github.com/zjin-lcf/HeCBench",
```
geomin12: for TheRock, we avoid cloning external repos in code. what if this repo goes away? what happens?

can we add this as a `third-party` package so it is a part of TheRock and we can use it and guarantee it'll exist?

https://github.com/ROCm/TheRock/tree/main/third-party

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,266 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
```
geomin12: what is hecbench_spirv? can we add a comment in regards to what that is?

geomin12 (commented):
also, it fails: https://github.com/ROCm/TheRock/actions/runs/22850078408/job/66292916629?pr=3840

geomin12 (commented):
also for future reference, ask for "reviewers" instead of "assignees"

build_tools/github_actions/fetch_test_configurations.py
sriasapu: Updated

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
sriasapu: Moved

build_tools/github_actions/fetch_test_configurations.py
sriasapu: Yes, Updated

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
sriasapu: Yes, implemented

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
sriasapu: Moved to the build step

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
sriasapu: used submodule instead of direct clone

build_tools/github_actions/test_executable_scripts/test_hecbench_spirv.py
sriasapu: Added the description

build_tools/fetch_sources.py
```diff
@@ -169,13 +168,25 @@ def run(args):
...
+                cwd=THEROCK_DIR,
+            )
+        if HECBENCH_SUBMODULE_PATH in submodule_paths:
+            run_command(
+                ["git", "submodule", "update", "--init"]
+                + update_args
+                + ["--", HECBENCH_SUBMODULE_PATH],
+                cwd=THEROCK_DIR,
+                env={"GIT_LFS_SKIP_SMUDGE": "1"},
+            )
```
geomin12: ? we should fetch it anyway, no need for this conditional. we should build everything and have everything needed for the user

build_tools/install_rocm_from_artifacts.py
```diff
@@ -23,6 +23,7 @@
     [--blas | --no-blas]
     [--debug-tools | --no-debug-tools]
     [--fft | --no-fft]
+    [--hecbench_spirv | --no-hecbench_spirv]
```
geomin12: hm... these are meant for ROCm specific items, not for third-party libraries

geomin12 (commented):
i added some build folks to this to review. not sure if third-party is the right path for this, but for benchmark code looks fine

sriasapu:
> i added some build folks to this to review. not sure if third-party is the right path for this, but for benchmark code looks fine

As per the last suggestion, I have added under third-party.
Please help me with the right path to keep this.

build_tools/fetch_sources.py
sriasapu: I added GIT_LFS_SKIP_SMUDGE=1 because the repo’s LFS quota is exceeded. Without it, checkout fails and the build can’t proceed. This flag ensures the code builds cleanly.

error: external filter 'git-lfs filter-process' failed
fatal: src/cmp-cuda/data.tar.bz: smudge filter lfs failed
warning: Clone succeeded, but checkout failed.

marbre (changes requested):
I only briefly went over this PR but see a few issues:

* `hecbench-spirv` should not be added to the top-level and should go into a subfolder. I don't have enough context but it feels this should rather go to third-party.
* As a third-party dep we should _not_ add this as a submodule. Instead a release / snapshot should be mirrored to and then pulled from our S3 bucket for third-party dependencies.

third-party/hecbench-spirv/project/build_hecbench_spirv.py
marbre, line 1: Why is this + the custom CMake config needed? Can't https://github.com/ORNL/HeCBench/blob/master/CMAKE_BUILD.md be followed?

build_tools/install_rocm_from_artifacts.py
sriasapu: Removed

third-party/hecbench-spirv/project/build_hecbench_spirv.py
sriasapu: Updated,
Now makefile configurations are based on flags only except compiler linking changes

third-party/hecbench-spirv/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+therock_subproject_fetch(therock-hecbench-sources
+  CMAKE_PROJECT
+  # TODO: Upload to S3 mirror and restore the S3 URL before merging.
+  URL https://github.com/zjin-lcf/HeCBench/archive/a2efabf38919f26c0c607e22e84cd2b7dde911a1.tar.gz
+  URL_HASH SHA256=787732c67ec0adfe702c88da6278b655635f752707cfcbc99097d472eb8b81fa
```
sriasapu, line 8: @marbre As discussed offline, Please help to upload the latest hecbench repo tar file to S3

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,651 @@
...
+from benchmark_base import BenchmarkBase
+from utils.logger import log
+
+THEROCK_DIR = Path(__file__).resolve().parents[4]
+
+TIME_UNIT_PATTERN = re.compile(
+    r"(?i)^(?:s|sec|secs|second|seconds|ms|msec|millisecond|milliseconds|us|usec|microsecond|microseconds|ns|nanosecond|nanoseconds)$"
+)
+
+COMMON_TIME_PATTERNS: List[str] = [
```
geomin12, line 30: wow this is a massive list, does the hecbench executable already have some output provided with JSON? i'm not a fan of parsing logs out and wondering if there is a better way

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,651 @@
...
+
+class HeCBenchSPIRVBenchmark(BenchmarkBase):
+    """HeCBench SPIR-V benchmark test."""
+
+    def __init__(self):
+        super().__init__(
+            benchmark_name="hecbench_spirv", display_name="HeCBench SPIR-V"
+        )
+        self._results: List[Dict[str, Any]] = []
+        self.hecbench_benchmarks = os.getenv("HECBENCH_BENCHMARKS", "")
```
geomin12, line 448: it doesn't seem like this is ever set

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,651 @@
...
+    return any(bench_dir.glob("*.o")) or (bench_dir / "main").is_file()
+
+
+def _select_bench_dirs(
+    src_dir: Path,
+    hecbench_benchmarks: str = "",
+) -> List[Path]:
+    """Return benchmark directories matching the filter (or all *-hip dirs)."""
+    if not hecbench_benchmarks:
+        hecbench_benchmarks = os.getenv("HECBENCH_BENCHMARKS", "")
```
geomin12, line 367: this doesn't ever get set

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
```diff
@@ -0,0 +1,651 @@
...
+        if not src_dir.is_dir():
+            raise RuntimeError(
+                "Prebuilt HeCBench benchmarks not found. "
+                f"Expected directory: {src_dir}. "
+                "Build HeCBench as part of TheRock build/install step before running this test."
+            )
+
+        log.info("Using prebuilt HeCBench SPIR-V benchmark tree at %s", src_dir)
+
+        selected_benchmarks = _select_bench_dirs(src_dir, self.hecbench_benchmarks)
```
geomin12, line 467: since `HECBENCH_BENCHMARKS` isn't ever set, it seems this won't ever be called

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
geomin12, line 1: i'm not a fan of using `re.search` to find results. does the executable provide a way to look at logs? maybe a json output or csv or xml? 

re.search feels very brittle and prone to error

tests/extended_tests/utils/extended_test_base.py
```diff
@@ -111,19 +120,54 @@ def execute_command(
...
+        if capture_output and stderr_data:
+            output_for_log = f"{output_for_log}{stderr_data}"
+
+        for line in output_for_log.splitlines(True):
+            log.info(line.rstrip())
             if log_file_handle:
-                log_file_handle.write(f"{line}")
+                log_file_handle.write(line)
+
+        rc = timeout_exit_code if timed_out else process.returncode
```
geomin12, line 161: do we need this? pretty sure if it times out, it will print out the expected error

geomin12 (commented):
going to let build folks review the build part

I also expect to see a test run of this with CI nightly, to ensure it works

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
sriasapu: This enables the user to run his own set of tests
If not set then all BMs will run

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
sriasapu: No,
Executable gives the results in the simple text format only

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
sriasapu: Unfortunately, HeCBench doesn't provide structured JSON output — it's a collection of ~400 independent benchmarks each with their own ad-hoc text output format. The regex list is a necessary evil to extract metrics from the diverse output formats.

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
sriasapu: This enables the user to run his own set of tests
If not set then all BMs will run

tests/extended_tests/benchmark/scripts/test_hecbench_spirv.py
sriasapu: This enables the user to run his own set of tests
If not set then all BMs will run

tests/extended_tests/utils/extended_test_base.py
sriasapu: Yes, this is needed.
Some benchmarks appear to get stuck even though they are actually running fine and producing correct results. To ensure we can properly validate those results, we are returning both the stdout and stderr data.

geomin12 (commented):
thanks for responses. overall looks good, but can we add a test to validate this works? you can use CI nightly and target this specific test: https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md

also, i am only reviewing test. we need another approval for build

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+# ---------------------------------------------------------------------------
+# Resolve hipcc from the amd-hip toolchain
+# ---------------------------------------------------------------------------
+if(DEFINED THEROCK_TOOLCHAIN_ROOT AND EXISTS "${THEROCK_TOOLCHAIN_ROOT}")
+  find_program(HECBENCH_HIPCC
+    NAMES hipcc hipcc.bin
+    HINTS "${THEROCK_TOOLCHAIN_ROOT}/bin"
+    NO_DEFAULT_PATH
+  )
+endif()
```
marbre: This should be automatically injected via the toolchain file and there should be no need to set it here. We should really try to use the upstream CMake config if there are no super strong reasons not to.

third-party/hecbench-spirv/project/CMakeLists.txt
marbre, line 1: Why do we need our own `CMakeLists.txt`? Can't we just pick up https://github.com/ORNL/HeCBench/blob/master/CMakeLists.txt instead of writing our own? The motivation isn't really clear here.

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Yes, Updated

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: The upstream build uses generic HIP flags. We need custom **--offload-arch=amdgcnspirv flags** makefile fixes for SPIR-V validation.

marbre: This whole CMake overlay config looks sketchy, will add some more inline comments. Instead of providing our own, why don't we contribute to upstream to improve https://github.com/ORNL/HeCBench/blob/master/CMakeLists.txt?

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+      set(HECBENCH_BASE "${_child}")
+      break()
+    endif()
+  endforeach()
+endif()
+if(NOT EXISTS "${HECBENCH_BASE}/src")
+  message(FATAL_ERROR
+    "Could not locate HeCBench src/ under ${HECBENCH_SOURCE_DIR}. "
+    "Expected ${HECBENCH_SOURCE_DIR}/src or ${HECBENCH_SOURCE_DIR}/*/src")
+endif()
```
marbre: This should not be needed as we bundle the source or what is this for?

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+
+# ---------------------------------------------------------------------------
+# Resolve HIP compiler from configured toolchain
+# ---------------------------------------------------------------------------
+set(HECBENCH_HIPCC "${CMAKE_CXX_COMPILER}")
+if(NOT HECBENCH_HIPCC)
+  message(FATAL_ERROR "CMAKE_CXX_COMPILER is not set; expected a configured HIP-capable toolchain")
+endif()
+message(STATUS "HeCBench: using compiler ${HECBENCH_HIPCC}")
```
marbre: Don't set your own `HECBENCH_HIPCC`. Instead rely on the `CMAKE_CXX_COMPILER` and the toolchain file.

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+      "CC=${HECBENCH_HIPCC}"
+      "EXTRA_CFLAGS=${HECBENCH_SPIRV_CFLAGS}"
+      "LDFLAGS=${HECBENCH_SPIRV_LDFLAGS}"
+    COMMAND "${CMAKE_COMMAND}" -E touch "${_stamp}"
+    WORKING_DIRECTORY "${_dir}"
+    DEPENDS "${_fix_stamp}"
+    COMMENT "Building HeCBench SPIR-V: ${_name}"
+    VERBATIM
+  )
+  list(APPEND _stamps "${_stamp}")
```
marbre, line 102: This is mixing up CMake + make why I would have strongly preferred to go with https://github.com/ORNL/HeCBench/blob/master/CMakeLists.txt. Not sure `add_custom_command` is the right way here. Minor not, it also not best practice to hardcore the level of parallelism via `-j2`.

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+set(HECBENCH_SPIRV_LDFLAGS
+  "--hip-link --offload-arch=amdgcnspirv"
+  CACHE STRING "LDFLAGS for HeCBench SPIR-V benchmarks")
+
+# ---------------------------------------------------------------------------
+# Prefer upstream package config discovery for hipcub / rocprim include paths
+# ---------------------------------------------------------------------------
+set(_hecbench_extra_include_dirs)
+
+find_package(hipcub CONFIG QUIET)
```
marbre: This will most likely always fail, at least in CI and you will never notice. While the CMake config has a dependency modeled, there is no artifact dependency defined in the build topology. As this belong to `hip-runtime` (not sure it should at all) it will not pick up math libs and this it will never find hipcub or rocprim.

third-party/hecbench-spirv/project/CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+  set(_hipcub_includes ${hipcub_INCLUDE_DIRS})
+elseif(DEFINED hipcub_INCLUDE_DIR)
+  set(_hipcub_includes ${hipcub_INCLUDE_DIR})
+endif()
+if(_hipcub_includes)
+  list(APPEND _hecbench_extra_include_dirs ${_hipcub_includes})
+  message(STATUS "HeCBench: using hipcub includes from package config")
+endif()
+
+find_package(rocprim CONFIG QUIET)
```
marbre: As above. This will always be bypassed in CI.

BUILD_TOPOLOGY.toml
```diff
@@ -298,6 +302,12 @@ type = "generic"
 artifact_group_deps = ["profiler-core", "compiler", "core-amdsmi"]
 source_sets = ["rocm-systems"]
 
+[artifact_groups.hecbench-spirv]
+description = "HeCBench SPIR-V benchmark suite (prebuilt benchmarks)"
+type = "generic"
+artifact_group_deps = ["hip-runtime"]
```
marbre: To me this seems to be the wrong artifact group if there are deps on math-libs.

BUILD_TOPOLOGY.toml
```diff
@@ -737,3 +747,13 @@ type = "target-neutral"
 artifact_deps = ["amd-llvm", "amd-dbgapi", "sysdeps-gmp", "sysdeps-mpfr", "sysdeps-expat", "sysdeps-ncurses"]
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
 disable_platforms = ["windows"]
+
+# --- Benchmark Artifacts ---
+
+[artifacts.hecbench-spirv]
+artifact_group = "hecbench-spirv"
+type = "target-neutral"
+artifact_deps = ["core-hip", "amd-llvm"]
```
marbre: Missing dep on math-libs.

BUILD_TOPOLOGY.toml
```diff
@@ -737,3 +747,13 @@ type = "target-neutral"
 artifact_deps = ["amd-llvm", "amd-dbgapi", "sysdeps-gmp", "sysdeps-mpfr", "sysdeps-expat", "sysdeps-ncurses"]
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
 disable_platforms = ["windows"]
+
+# --- Benchmark Artifacts ---
+
+[artifacts.hecbench-spirv]
+artifact_group = "hecbench-spirv"
+type = "target-neutral"
```
marbre: This depends on target-specifics libs (`prim`) and thus most likely will need more work to get you what you want. For which target should your target neutral build pull the artifacts or rather should it pull for all? `hip-runtime` is the wrong stage for this.

marbre (changes requested):

BUILD_TOPOLOGY.toml
sriasapu: Updated

BUILD_TOPOLOGY.toml
sriasapu: Updated

BUILD_TOPOLOGY.toml
sriasapu: Updated

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Removed

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Removed

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Addressed. Hardcoded `-j2` is removed

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Addressed. The build now uses `CMAKE_CXX_COMPILER` only

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Yes removed

third-party/hecbench-spirv/CMakeLists.txt
```diff
@@ -24,8 +24,10 @@ therock_cmake_subproject_declare(hecbench-spirv
     therock-hecbench-sources
   CMAKE_ARGS
     "-DHECBENCH_SOURCE_DIR=${CMAKE_CURRENT_BINARY_DIR}/source"
+    "-DHECBENCH_SPIRV_CFLAGS=${_hecbench_cflags}"
   BUILD_DEPS
-    ${_hecbench_optional_build_deps}
+    hipCUB
+    rocPRIM
```
marbre: What you had before here was fine as is allows to build with prim disabled.

sriasapu: reverted

third-party/hecbench-spirv/project/CMakeLists.txt
sriasapu: Agreed with the suggested way.
As discussed offline, We will create the separate PR for these changes once the upstream is updated.

sriasapu:
As per the discussion, we can hold this test addition as its not a priority.

PR #3852, opened 2026-03-09
[CI] Enable testing of PyTorch's different test configs

ethanwee1:
Provide ability to run pytorch UT's using different test configs. Running the full gamut of UT's will follow in a seperate PR. https://github.com/ROCm/TheRock/pull/3929

Latest run: https://github.com/ROCm/TheRock/actions/runs/23267813952

### DIFF
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -90,7 +90,7 @@ run-name: Test PyTorch Full (${{ inputs.amdgpu_family }}, ${{ inputs.torch_versi
 
 jobs:
   prepare_matrix:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     outputs:
       matrix: ${{ steps.generate.outputs.matrix }}
     steps:
@@ -99,7 +99,6 @@ jobs:
         with:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref || '' }}
-          sparse-checkout: external-builds/pytorch/generate_test_sharding_matrix.py
 
       - id: generate
         run: |
--- a/external-builds/pytorch/generate_test_sharding_matrix.py
+++ b/external-builds/pytorch/generate_test_sharding_matrix.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 import argparse
 import json
 import os
-import sys
 
 # Shard counts mirror the parallelism used by upstream PyTorch CI for the
 # corresponding ROCm test configurations.  Chosen to keep each shard under
@@ -104,8 +103,7 @@ def main() -> None:
 
     configs = args.test_configs.split()
     if not configs:
-        print("Error: --test-configs must not be empty", file=sys.stderr)
-        sys.exit(1)
+        parser.error("--test-configs must not be empty")
 
     matrix = build_matrix(configs, args.default_runner, args.multi_gpu_runner)
     matrix_json = json.dumps(matrix, separators=(",", ":"))


.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -16,6 +16,10 @@ on:
         required: true
         type: string
         default: "linux-mi325-1gpu-ossci-rocm"
+      distributed_test_runs_on:
+        description: Runner label for distributed config (multi-GPU)
+        type: string
+        default: "linux-mi325-8gpu-ossci-rocm"
```
jithunnair-amd: @HereThereBeDragons @scottt Instead of making the distributed runner as another workflow input, I'd prefer to just substitute "4gpu" instead of "1gpu" (with The assumption that a consistent naming pattern is/ will be followed for the runner labels which should not be hard and in fact should be insisted upon for consistency sake).
Side note: @ethanwee1 we do not need 8 gpus for distributed test. That's a waste of runners/GPUs.

ethanwee1: We don't have any 4GPU nodes right now @jithunnair-amd per @geomin12's comment.

> linux-mi325-4gpu-ossci-rocm runner was decommissioned a while back

ethanwee1: I did originally use the [4GPU label](https://github.com/ROCm/TheRock/pull/3852/changes/c08682b87680e34d83e6f1fe43caad94dfe5056e), but the nodes were facing connection issues and I then asked in the CICD help chat and found that those nodes were decommissioned.

jithunnair-amd: @ethanwee1 Okay, thanks for the info. I'd expect 4-GPU runners would be a better use of our limited CI capacity than 8-GPU runners here, but that's a problem for theRock's CI infra team (@geomin12?) to address.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +74,58 @@ on:
...
+  prepare_matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - id: generate
+        run: |
+          get_num_shards() {
+            case "$1" in
+              default)     echo 6 ;;
```
jithunnair-amd: @ethanwee1 Please create a follow-up story to see if we can make this number be informed by the estimated duration of the test suites to be run as part of the test_config (based on test-times.json), with a maximum intended duration of each shard being 3h. That would ensure we use the optimal number of runners based on the requirement.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +74,58 @@ on:
...
 jobs:
+  prepare_matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - id: generate
+        run: |
+          get_num_shards() {
+            case "$1" in
```
jithunnair-amd: @ethanwee1 Please add a comment here to mention where these "magic" numbers come from (i.e. upstream CI)

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +74,58 @@ on:
...
+          done
+          matrix="$matrix]}"
+          echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
+
   test_wheels:
-    name: Test PyTorch Full | ${{ inputs.amdgpu_family }} (shard ${{ matrix.shard }}/${{ matrix.num_shards }})
-    runs-on: ${{ inputs.test_runs_on }}
+    needs: prepare_matrix
+    name: Test PyTorch Full | ${{ inputs.amdgpu_family }} / ${{ matrix.test_config }} (shard ${{ matrix.shard }}/${{ matrix.num_shards }})
+    runs-on: ${{ matrix.test_config == 'distributed' && inputs.distributed_test_runs_on || inputs.test_runs_on }}
```
jithunnair-amd: And this would be `matrix.runs_on` instead of the above, so that the matrix generation logic can handle the regex substitution as per https://github.com/ROCm/TheRock/pull/3852/changes#r2941659677

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +74,58 @@ on:
...
 run-name: Test PyTorch Full (${{ inputs.amdgpu_family }}, ${{ inputs.torch_version}}, ${{ inputs.test_runs_on }})
 
 jobs:
+  prepare_matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - id: generate
+        run: |
```
jithunnair-amd, line 104: @ethanwee1 I think it'll be more readable and maintainable to have a separate helper python script such as `generate_matrix.py` in https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch

HereThereBeDragons: could we have a more meaningful file name please? what kind of matrix is generated?

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,102 @@
...
+
+Outputs a ``matrix`` variable via $GITHUB_OUTPUT suitable for consumption by
+``fromJSON()`` in a workflow strategy block.
+
+Usage (in a workflow step)::
+
+    python external-builds/pytorch/generate_matrix.py \
+        --test-configs 'default distributed inductor' \
+        --default-runner 'linux-rocm-docker-mi300-1gpu-ossci'
+"""
```
HereThereBeDragons, line 27: what would be the example output with this command?
important as we want to process the json in the workflow.

external-builds/pytorch/generate_matrix.py
```diff
@@ -0,0 +1,102 @@
...
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import sys
+
+# Shard counts are chosen to keep each shard under ~6 h on current runners.
```
HereThereBeDragons: what are current runners? maybe add the type and date of this comment.

external-builds/pytorch/generate_matrix.py
```diff
@@ -0,0 +1,102 @@
...
+# inductor-rocm-mi300.yml).
+SHARDS_PER_CONFIG: dict[str, int] = {
+    "default": 6,
+    "distributed": 3,
+    "inductor": 2,
+}
+DEFAULT_SHARDS = 4
+
+
+def derive_runner(default_runner: str, config: str) -> str:
```
HereThereBeDragons: this assumption of replacing 1gpu with 8gpu will not work.

we have in `amdgpu_family_matrix.py` a label called: `test-runs-on-multi-gpu`. it would be best to take this as input of the workflow file.

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -329,9 +329,11 @@ def main(argv: list[str]) -> int:
...
     # Determine AMDGPU family and set HIP_VISIBLE_DEVICES BEFORE importing
     # torch or running pytest.  Once torch.cuda is initialized, changing
-    # HIP_VISIBLE_DEVICES has no effect.  For unit tests we run on a single
-    # device (policy="single") to avoid multi-GPU contention.
-    ((first_arch, _),) = set_gpu_execution_policy(args.amdgpu_family, policy="single")
+    # HIP_VISIBLE_DEVICES has no effect.  Distributed tests need all GPUs;
+    # other configs use a single device to avoid multi-GPU contention.
+    gpu_policy = "all" if args.test_config == "distributed" else "single"
+    selected = set_gpu_execution_policy(args.amdgpu_family, policy=gpu_policy)
+    first_arch = selected[0][0]
```
HereThereBeDragons, line 336: this is not what we want:
if we run on all gpus, we also want to list all gpus we selected.

selected can have the value 
```
[('gfx942', 0),
 ('gfx942', 1),
 ('gfx942', 2),
 ('gfx942', 3),
 ('gfx942', 4),
 ('gfx942', 5),
 ('gfx942', 6),
 ('gfx942', 7)]
```

please note if the requested family is e.g. gfx120X-all it can contain values of gfx1200 and gfx1201. as such we should print all values.

see description `- "all": Use all supported devices (every detected, possibly multiple per arch).`

i think there are also pytorch versions that support multiple gpu archs? that is also something to consider that than when "all" is used all gpus matching those archs will be returned.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: https://github.com/orgs/ROCm/projects/18/views/1?filterQuery=sprint%3A%40current+assignee%3A%22ethanwee1%22&pane=issue&itemId=166431378&issue=ROCm%7Cframeworks-internal%7C15890

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,124 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+
+# Shard counts mirror the parallelism used by upstream PyTorch CI for the
+# corresponding ROCm test configurations (rocm-mi300.yml and
+# inductor-rocm-mi300.yml as of March 2026).  Chosen to keep each shard
```
ethanwee1: Leave permalinks to what is currently in upstream to trunk.yml

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,124 @@
...
+
+Outputs a ``matrix`` variable via $GITHUB_OUTPUT suitable for consumption by
+``fromJSON()`` in a workflow strategy block.
+
+Usage (in a workflow step)::
+
+    python external-builds/pytorch/generate_test_sharding_matrix.py \\
+        --test-configs 'default distributed inductor' \\
+        --default-runner 'linux-rocm-docker-mi300-1gpu-ossci' \\
+        --multi-gpu-runner 'linux-rocm-docker-mi300-8gpu-ossci'
```
ethanwee1: Update from 8gpu runner name to 4gpu runner name

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,124 @@
...
+
+import argparse
+import json
+import os
+import sys
+
+# Shard counts mirror the parallelism used by upstream PyTorch CI for the
+# corresponding ROCm test configurations (rocm-mi300.yml and
+# inductor-rocm-mi300.yml as of March 2026).  Chosen to keep each shard
+# under ~3 h on MI300 1-GPU runners (default/inductor) and MI300 8-GPU
```
ethanwee1: Based on mi300 4gpu runners

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +75,46 @@ on:
...
 
-
 permissions:
   contents: read
 
 run-name: Test PyTorch Full (${{ inputs.amdgpu_family }}, ${{ inputs.torch_version}}, ${{ inputs.test_runs_on }})
 
 jobs:
+  prepare_matrix:
+    runs-on: ubuntu-latest
```
ScottTodd: nit: follow the style guide and pin this: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#pin-action-runs-on-labels-to-specific-versions
```suggestion
  prepare_matrix:
    runs-on: ubuntu-24.04
```

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -64,29 +75,46 @@ on:
...
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || '' }}
+          sparse-checkout: external-builds/pytorch/generate_test_sharding_matrix.py
```
ScottTodd: nit: I don't think a sparse checkout here will save much. As long as you don't fetch TheRock's submodules, the checkout step is pretty fast (seconds). Seems like unnecessary complexity and would make it riskier to edit the files in the future (e.g. for generate_test_sharding_matrix.py took a dep on another python file)

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,129 @@
...
+            "'test-runs-on-multi-gpu' in amdgpu_family_matrix.py "
+            "(e.g. 'linux-mi325-4gpu-ossci-rocm')"
+        ),
+    )
+    args = parser.parse_args()
+
+    configs = args.test_configs.split()
+    if not configs:
+        print("Error: --test-configs must not be empty", file=sys.stderr)
+        sys.exit(1)
```
ScottTodd: This can use `parser.error()` instead of `print` + `sys.exit`

* https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#use-argparse-for-cli-flags
* https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error

external-builds/pytorch/generate_test_sharding_matrix.py
```diff
@@ -0,0 +1,129 @@
...
+            f"  {entry['test_config']} shard {entry['shard']}/{entry['num_shards']}"
+            f" -> {entry['runs_on']}"
+        )
+
+    github_output = os.environ.get("GITHUB_OUTPUT")
+    if github_output:
+        with open(github_output, "a") as f:
+            f.write(f"matrix={matrix_json}\n")
+    else:
+        print(f"\nmatrix={matrix_json}")
```
ScottTodd, line 123: FYI we have a helper for this in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py. If we move this script to that folder we could use that util.

Argument for keeping the file in `external-builds/pytorch`:
* Keeps pytorch code near other pytorch code
* Easier for potential upstreaming (or at least referencing upstream, without connecting too much to TheRock's other code)

Argument for putting in `build_tools/github_actions/`:
* Co-located with https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/generate_pytorch_manifest.py, https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/write_torch_versions.py, etc.

ethanwee1: I don't have a strong preference. Just left it here since the script is PyTorch-specific (shard counts, test configs, runner selection all tied to upstream PyTorch CI). But open to moving it to build_tools/github_actions/ works also.

Issue #3850, opened 2026-03-09
extended_tests: Results API unreachable from CI nodes


lajagapp:
### Summary
The results API (used in extended_tests/benchmark) is unreachable from CI runners, causing benchmark tests to fail with network errors during result upload and LKG comparison.

### Root Cause
As per the recent network changes, the Azure-hosted API appears to be blocking CI runner IPs (403 Ip Forbidden) and causing connection resets.

### Current Workaround
LKG comparison and API upload will be temporarily disabled in benchmark_base.py with TODO. Benchmark scores are still printed iin console and saved locally.

### Actions
- Will get the API network/firewall configuration fixed to allow CI runner IPs
- Re-enable compare_with_lkg() in benchmark_base.py (restore original implementation)
- Remove the TODO(lajagapp) workaround warnings

PR #3853, opened 2026-03-09
extended_tests: Disable results API upload and LKG comparison

lajagapp:
### Summary

Temporarily disabling results API upload and LKG comparison in extended_tests/benchmark due to network connectivity issues.

### Problem

Benchmark CI jobs are failing with `403 Client Error: Ip Forbidden` on result upload & LKG fetch
This causes the entire benchmark run to error out even when all benchmarks pass.

### Changes

- Add `ENABLE_RESULTS_API` flag in benchmark_base.py (set to False)
- When disabled:
  - upload_results skips API call, logs warning
  - compare_with_lkg prints raw score tables, skips LKG fetch

### Actions

Set `ENABLE_RESULTS_API = True` in benchmark_base.py once the API network/firewall issue is resolved. Tracked in https://github.com/ROCm/TheRock/issues/3850.

### DIFF
--- a/tests/extended_tests/benchmark/scripts/benchmark_base.py
+++ b/tests/extended_tests/benchmark/scripts/benchmark_base.py
@@ -22,7 +22,7 @@ from utils.extended_test_base import ExtendedTestBase, gha_append_step_summary
 ENABLE_RESULTS_API = False
 
 
-class BenchmarkBase:
+class BenchmarkBase(ExtendedTestBase):
     """Base class providing common benchmark logic.
 
     Inherits shared infrastructure from ExtendedTestBase (execute_command,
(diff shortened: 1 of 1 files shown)


tests/extended_tests/benchmark/scripts/benchmark_base.py
```diff
@@ -24,6 +24,11 @@
 from github_actions_utils import gha_append_step_summary
 
 
+# TODO(lajagapp): Set to True once the results API network/firewall issue is
+# resolved. (https://github.com/ROCm/TheRock/issues/3850)
+ENABLE_RESULTS_API = False
```
geomin12, line 22: i'm also okay with commenting out results API upload code

eventually, we want it to always be uploading

geomin12 (approved):

PR #3854, opened 2026-03-09
Add Python package building and testing to multi-arch CI

marbre:
## Motivation                                                                                                                                                                                                                       
                                                                                                                                                                                                                                    
The multi-arch CI pipeline built ROCm artifacts for all GPU families                                                                                                                                                                
but stopped short of producing distributable Python wheels. This change
extends `multi_arch_ci_linux.yml` to build and test Python packages for
all families in a single job, following the same pattern as `ci_linux.yml`.                                                                                                                                                         
 
## Changes                                                                                                                                                                                                                          
                
### CI Workflow (`multi_arch_ci_linux.yml`)                                                                                                                                                                                         
* Adds a `build_python_packages` job after `build_multi_arch_stages`,
  fetching all family artifacts via `artifact_manager.py --stage=all`                                                                                                                                                               
  and building family-specific wheels.
* Adds a `test_python_packages_per_family` job with per-family matrix                                                                                                                                                               
  testing using per-family `--find-links` index URLs.                                                                                                                                                                               
                                                                                                                                                                                                                                    
### Artifact Fetching (`artifact_manager.py`)                                                                                                                                                                                       
* Adds `--amdgpu-families=all` mode that fetches all available artifacts
  matching the inbound set by prefix, bypassing the component/family                                                                                                                                                                
  enumeration loop.                                                                                                                                                                                                                 
* Fixes the `--output-dir` convention: the tool appends `/artifacts` to                                                                                                                                                             
  the given path in non-flatten mode; the workflow now passes `$workspace`                                                                                                                                                          
  so artifacts land in `$ARTIFACTS_DIR` as `ArtifactCatalog` expects.                                                                                                                                                               
* Drops `--flatten`: it caused a concurrent directory-creation race and                                                                                                                                                             
  was incompatible with `ArtifactCatalog`'s required per-artifact                                                                                                                                                                   
  subdirectory layout.                                                                                                                                                                                                              
                                                                                                                                                                                                                                    
### Per-Family Indexing (`generate_package_index.py`, `upload_python_packages.py`)                                                                                                                                                  
* Adds `generate_package_index.py`, a new module that generates
  per-family `index.html` files with relative paths to both                                                                                                                                                                         
  family-specific wheels (`./`) and generic wheels (`../`), allowing pip                                                                                                                                                            
  to resolve both from a single `--find-links` URL.                                                                                                                                                                                 
* Updates `upload_python_packages.py` to omit the `artifact_group` path                                                                                                                                                             
  component when `--multiarch` is set, uploading directly to `python/`                                                                                                                                                              
  instead of `python/multi-arch-release/` - the run ID already uniquely                                                                                                                                                             
  identifies the build.                                                                                                                                                                                                             
                                                                                                                                                                                                                                    
### Backward Compatibility
Single-family builds via `ci_linux.yml` are unaffected. All new
multi-arch behavior is gated on `amdgpu_families` and `multiarch_index`                                                                                                                                                             
inputs.
                                                                                                                                                                                                                                    
## Test Plan    
* Adds unit tests for `generate_package_index.py`, `artifact_manager.py`
  fetch logic, `workflow_outputs.python_packages()` with empty artifact                                                                                                                                                             
  group, and the multi-arch upload path.
* Multi-arch CI run: https://github.com/ROCm/TheRock/actions/runs/22913716354

### DIFF
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -141,7 +141,7 @@ def parse_target_families(args: argparse.Namespace) -> List[str]:
         log("Using generic (host) artifacts only")
     else:
         if args.amdgpu_families:
-            target_families.extend(args.amdgpu_families.replace(";", ",").split(","))
+            target_families.extend(args.amdgpu_families.split(";"))
         if args.amdgpu_targets:
             target_families.extend(
                 t.strip() for t in args.amdgpu_targets.split(",") if t.strip()
@@ -913,7 +913,7 @@ def do_info(args: argparse.Namespace):
 
     # Show target families if provided
     if args.amdgpu_families:
-        families = args.amdgpu_families.replace(";", ",").split(",")
+        families = args.amdgpu_families.split(";")
         target_families = ["generic"] + families
         log(f"\nTarget families: {', '.join(target_families)}")
 
@@ -954,7 +954,7 @@ def _add_target_args(parser: argparse.ArgumentParser):
     target_group.add_argument(
         "--amdgpu-families",
         type=str,
-        help="Comma or semicolon-separated GPU families (e.g., gfx94X-dcgpu,gfx1100)",
+        help="Semicolon-separated GPU families (e.g., gfx94X-dcgpu;gfx1100)",
     )
     target_group.add_argument(
         "--generic-only",


build_tools/github_actions/upload_python_packages.py
```diff
@@ -120,9 +141,36 @@ def upload_packages(
...
+    """Write GitHub Actions summary with pip install instructions.
+
+    Args:
+        packages_loc: Storage location for packages
+        multiarch: If True, this is a multi-arch build with per-family indexes
+    """
+    if multiarch:
+        # Multi-arch: show base URL with explanation
+        base_url = packages_loc.https_url
+        install_instructions_markdown = f"""[ROCm Python packages]({base_url})
```
ScottTodd: From https://github.com/ROCm/TheRock/actions/runs/22876872994#summary-66385866694, the https://therock-ci-artifacts.s3.amazonaws.com/22876872994-linux/python page does not exist. Should it? I'd say if it isn't a valid link, don't format as a hyperlink.

```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Key>22876872994-linux/python</Key>
<RequestId>2J6NDKAMME2PXP42</RequestId>
<HostId>
eAWme5uBgVgDSqGuBBi6Z/mGKD6xp40zCLDuLb+HJOXkvSFEeW7cpUWMvk6yJ9oi2yoMoqMctDI=
</HostId>
</Error>
```

Compare with a run on CI: https://github.com/ROCm/TheRock/actions/runs/22876190271?pr=3854#summary-66464556136
https://therock-ci-artifacts.s3.amazonaws.com/22876190271-linux/python/gfx94X-dcgpu/index.html

build_tools/github_actions/upload_python_packages.py
```diff
@@ -120,9 +141,36 @@ def upload_packages(
...
+```bash
+pip install rocm[libraries,devel] --pre {LINE_CONTINUATION_CHAR}
+    --find-links={base_url}/gfx94X-dcgpu/index.html
+```
+
+Example for gfx120X-all:
+```bash
+pip install rocm[libraries,devel] --pre {LINE_CONTINUATION_CHAR}
+    --find-links={base_url}/gfx120X-all/index.html
+```
```
ScottTodd: Rather than include examples (which may not exist for the current configuration), how about all the actual links?

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -94,3 +94,62 @@ jobs:
...
+          inputs.use_prebuilt_artifacts == 'false' ||
+          inputs.use_prebuilt_artifacts == 'true'
+        ) &&
+        inputs.expect_failure == false
+      }}
+    uses: ./.github/workflows/build_portable_linux_python_packages.yml
+    with:
+      artifact_run_id: "${{ inputs.artifact_run_id != '' && inputs.artifact_run_id || github.run_id }}"
+      artifact_group: ${{ inputs.artifact_group }}
+      amdgpu_families: "all"
```
ScottTodd: Can this use `${{ inputs.dist_amdgpu_families }}` instead of `"all"`? If we know the actual list, I think we should pass it and not lose that information here.

build_tools/packaging/python/generate_local_index.py
ScottTodd, line 1: I think we should move this to https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python , next to https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/generate_release_index.py (then consider deleting that one? Or name them "local" and "s3"?)

Just "package index" isn't specific enough when we also have native linux packages too.

build_tools/generate_package_index.py
```diff
@@ -0,0 +1,231 @@
...
+
+    print(f"Found {len(generic_packages)} generic packages at top level: {dist_dir}")
+
+    # Process each family subdirectory
+    family_dirs = [d for d in dist_dir.iterdir() if d.is_dir()]
+
+    if not family_dirs:
+        print(f"Warning: No family subdirectories found in {dist_dir}")
+        print("This may not be a multi-arch build.")
+        return
```
ScottTodd: Seems like this should raise an exception instead of just logging and returning. Shouldn't be calling `generate_multiarch_indexes` on a non-multi-arch build.

.github/workflows/multi_arch_ci_linux.yml
marbre: `dist_amdgpu_families` is semicolon-separated, but `amdgpu_families` in the called workflow expects a comma-separated input. So rather we translate from semicolon- to comma-separated (regex for the win) and pass it down or just grab _all_ artifacts available. I was going with the latter but can also try to find a clean way for the former.

ScottTodd: I ran into a similar issue on https://github.com/ROCm/TheRock/pull/3856. I changed `build_tools/artifact_manager.py` to take a semicolon-separated list instead of a comma-separated list there.

build_tools/github_actions/upload_python_packages.py
marbre: Done.

build_tools/github_actions/upload_python_packages.py
marbre: Done.

build_tools/packaging/python/generate_local_index.py
marbre: Moved and renamed those now. I think we need to delete one and harmonize as there are way to far scripts now doing related things :D However, I think it will be smoother to clean up in a separate, follow-up PR.

build_tools/generate_package_index.py
marbre: Done.

marbre:
@ScottTodd all your review comments have been addressed. I will trigger another multi-arch CI run.

marbre:
Latest multi-arch CI run: https://github.com/ROCm/TheRock/actions/runs/22913716354

marbre:
`rocm[libraries]` and `rocm[libraries,devel]` tests pass on Windows and Linux as well as on Linux in multi-arch CI. PyTorch builds are unaffected. CI failures are mostly due to aborted jobs and timeouts but not related to the changes in this PR. Going forward with merging.

PR #3801, opened 2026-03-05
Add artifact_manager.py copy subcommand for S3-to-S3 artifact copying

ScottTodd:
## Motivation

Progress on
* https://github.com/ROCm/TheRock/issues/3399

This will allow us to copy artifacts from prior workflow runs into the storage for the current workflow run to avoid rebuilding from source (on explicit request, or when algorithms determine that it is safe to do so).

## Technical Details

The logic is basically a merge of the "fetch" and "push" commands (but staying within the same backend and not round tripping through the local filesystem), so I also extracted some common code into helper functions.

The copies run in parallel and are _fast_ (mostly S3 metadata edits under the covers I believe, not actually moving gigabytes of data around) - 16 seconds for all gfx94X-dcgpu artifacts.

I added some special case code for .sha256sum files since I didn't have them locally (from fetch_artifacts.py). Without that special case code, if the files don't exist, the copies would be retried until hitting a limit. We can iterate on that behavior as these artifact handling workflows/scripts get more mature.

Self-review: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_018_multi-arch-prebuilt-1.md (see also the earlier iterations in that directory showing reviews prior to extracting code into shared helper functions).

## Test Plan

This PR itself (adding new code but not using it yet, checking that the refactoring did not break anything):

* New and existing unit tests
* CI on this PR
* Multi-arch CI: https://github.com/ROCm/TheRock/actions/runs/22741871900

Integration with future work (using the new code):
* https://github.com/ROCm/TheRock/actions/runs/22686204334/job/65769807284 Copied artifacts for the "foundation" and "compiler-runtime" stages.
* https://github.com/ROCm/TheRock/actions/runs/22686926501/job/65772434648 Copied artifacts for all stages to completely skip builds and go straight to tests.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3856, opened 2026-03-09
Plumb prebuilt_stages input through multi_arch_ci workflows

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/3399. This allows us to reuse artifacts from prior workflow runs to skip any build stages we want.

## Technical Details

This first uses the `artifact_manager.py copy` subcommand added in https://github.com/ROCm/TheRock/pull/3801 to copy artifacts from the `baseline_run_id` workflow run for a `prebuilt_stages` list to the current workflow run's S3 storage. Then, it marks stage build steps as skipped depending on the `prebuilt_stages` values.

I settled on a single `prebuilt_stages` and `baseline_run_id` for Linux and Windows - not allowing different values per platform in a single run.

Planned future work, out of scope for this PR:

* Pass `prebuilt_stages` through `configure_ci.py` for pre-processing
  * Special `all` option expands to all stages
  * Individual options expand to include their dependencies (e.g. `compiler-runtime` --> `foundation,compiler-runtime`)
  * Transform into a JSON object for safer handling (`contains()` on a string can match if any stages share substrings)
  * Report configuration information in job summary
* Compute `prebuilt_stages` and `baseline_run_id` automatically based on a PR's diff and recent workflow runs (e.g. using [`build_tools/find_latest_artifacts.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/find_latest_artifacts.py), or something more specific to rocm-libraries/rocm-systems when workflows run from there)

## Test Plan

* https://github.com/ROCm/TheRock/actions/runs/22875528559
  * Skipped `foundation,compiler-runtime`
  * Built `math-libs`
  * Ran tests and they pass as before
* No unit tests for this workflow code. Unit tests will be added for configure_ci.py

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -60,8 +60,13 @@ jobs:
       - name: Install artifact manager dependencies
         run: pip install boto3
 
+      # Assume the therock-ci OIDC role in ROCm/TheRock. Other repos
+      # fall back to runner base credentials (therock-ci-artifacts-external).
+      #
+      # This is just needed for write access to the current run's bucket,
+      # all possible source run buckets have public read access.
       - name: Configure AWS Credentials
-        if: ${{ !github.event.pull_request.head.repo.fork }}
+        if: ${{ github.repository == 'ROCm/TheRock' && !github.event.pull_request.head.repo.fork }}
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
         with:
           aws-region: us-east-2
--- a/.github/workflows/multi_arch_ci_windows.yml
+++ b/.github/workflows/multi_arch_ci_windows.yml
@@ -65,8 +65,13 @@ jobs:
       - name: Install artifact manager dependencies
         run: pip install boto3
 
+      # Assume the therock-ci OIDC role in ROCm/TheRock. Other repos
+      # fall back to runner base credentials (therock-ci-artifacts-external).
+      #
+      # This is just needed for write access to the current run's bucket,
+      # all possible source run buckets have public read access.
       - name: Configure AWS Credentials
-        if: ${{ !github.event.pull_request.head.repo.fork }}
+        if: ${{ github.repository == 'ROCm/TheRock' && !github.event.pull_request.head.repo.fork }}
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
         with:
           aws-region: us-east-2


docs/development/ci_behavior_manipulation.md
```diff
@@ -18,12 +29,64 @@ However, if additional options are wanted, you can add a label to manipulate the
...
+> See https://github.com/ROCm/TheRock/issues/3399 for details.
+
+The [Multi-Arch CI](https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci.yml)
+workflow supports skipping individual build stages by copying their artifacts
+from a previous workflow run. This will be used in a few scenarios. For example:
+
+- Changes to the rocm-libraries project will use prebuilt artifacts for
+  `foundation,compiler-runtime`
+- Changes to just test scripts or python packages will use prebuilt artifacts for
+  all stages
```
ScottTodd, line 57: @geomin12 I'm planning this as one general solution to https://github.com/ROCm/TheRock/pull/3857#pullrequestreview-3918405038

PRs that only modify `build_tools/github_actions/test_executable_scripts/*` will
1. Set `prebuilt_stages` to `all`
2. Pull artifacts from an automatically determined baseline (e.g. latest successful build on `main` ... perhaps walking back from the PR's head commit to avoid cross-branch issues)
3. Run tests using those artifacts

ScottTodd:
Synced, resolved conflicts, and triggered a new multi-arch CI test run (using prebuilt `foundation,compiler-runtime`): https://github.com/ROCm/TheRock/actions/runs/22925554327

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,9 +41,45 @@ permissions:
...
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.12"
+
+      - name: Install artifact manager dependencies
+        run: pip install boto3
+
+      - name: Configure AWS Credentials
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
geomin12: maybe add TheRock check similar to #3886  ?

.github/workflows/multi_arch_ci_windows.yml
```diff
@@ -37,9 +41,50 @@ permissions:
...
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: "3.12"
+
+      - name: Install artifact manager dependencies
+        run: pip install boto3
+
+      - name: Configure AWS Credentials
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
geomin12: same comment as above for TheRock check

build_tools/github_actions/configure_ci.py
```diff
@@ -637,11 +637,13 @@ def main(base_args, linux_families, windows_families):
...
     test_type_reason = "default (smoke tests)"
 
-    # In the case of a scheduled run, we always want to build and we want to run full tests
+    # Scheduled and manually dispatched runs always build.
     if is_schedule:
         enable_build_jobs = True
         test_type = "full"
         test_type_reason = "scheduled run triggers full tests"
+    elif is_workflow_dispatch:
+        enable_build_jobs = True
```
geomin12, line 647: oh, this may impact `therock-ci.yml`'s workflow dispatch behavior? i guess if we migrate to multi-arch though, this is fine

docs/development/ci_behavior_manipulation.md
geomin12: love it, mutli-arch will be NEAT! saving even more resources

build_tools/github_actions/configure_ci.py
ScottTodd: It will, yes. I slipped that change into this PR but could pull it out to its own PR.

I was testing via workflow_dispatch and the most recent commit on my branch only edited .md files so the builds were skipped. I wanted builds to run for any workflow_dispatch (there's no meaningful "base" commit for workflow_dispatch like there is for "pull_request" or "push")

.github/workflows/multi_arch_ci_linux.yml
ScottTodd: Depends on which PR lands first. I'll have merge conflicts with myself either way 😅

build_tools/github_actions/configure_ci.py
ScottTodd: For example... right now when I wanted to test: https://github.com/ROCm/TheRock/actions/runs/22926894170

```
modified_paths (max 200): ['docs/development/s3_buckets.md']
Checking modified files since this had a workflow_dispatch trigger
is_ci_run_required findings:
  related_to_ci: False
  contains_other_non_skippable_files: False
Only unrelated and/or skippable paths were modified, skipping build jobs
```

ScottTodd: Working on pulling that into its own PR... need to consider a few details more carefully. (I want to fully rewrite that configure_ci script though)

ScottTodd: https://github.com/ROCm/TheRock/pull/3893. I'll remove from this PR.

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -22,12 +22,16 @@ on:
...
         type: string
-      artifact_run_id:
-        type: string
       expect_failure:
         type: boolean
-      use_prebuilt_artifacts:
+      prebuilt_stages:
+        type: string
+        default: ""
+        description: "Comma-separated build stages to skip; artifacts are copied from baseline_run_id instead"
```
marbre, line 30:
```suggestion
        description: "Comma-separated build stages to skip; artifacts are copied from `baseline_run_id` run instead"
```
?

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,9 +41,50 @@ permissions:
   contents: read
 
 jobs:
+  copy_prebuilt_stages:
```
marbre, line 44: Move `copy_prebuilt_stages` to a reusable workflow? Looks identical to what is used in Windows workflow.

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,9 +41,50 @@ permissions:
...
+    name: Copy Prebuilt Stages
+    if: ${{ inputs.prebuilt_stages != '' && inputs.baseline_run_id != '' }}
+    runs-on: azure-linux-scale-rocm
```
marbre, line 47: Can we use a smaller runner type for this?

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -37,9 +41,50 @@ permissions:
...
+      # fall back to runner base credentials (therock-ci-artifacts-external).
+      #
+      # This is just needed for write access to the current run's bucket,
+      # all possible source run buckets have public read access.
+      - name: Configure AWS Credentials
+        if: ${{ github.repository == 'ROCm/TheRock' && !github.event.pull_request.head.repo.fork }}
+        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        with:
+          aws-region: us-east-2
+          role-to-assume: arn:aws:iam::692859939525:role/therock-ci
```
marbre, line 73: Note, we'll need to change this role if we want to reuse the entire workflow for releases.

build_tools/artifact_manager.py
```diff
@@ -1132,7 +1132,7 @@ def main(argv: Optional[List[str]] = None):
     info_parser.add_argument(
         "--amdgpu-families",
         type=str,
-        help="Comma-separated GPU families to show file lists for",
+        help="Semicolon-separated GPU families to show file lists for",
```
marbre, line 1135: This partly got in as part of https://github.com/ROCm/TheRock/commit/11d92a1188b799448fdf9633066166304d278c12#diff-a00290aeb17a08a2675418088a4627d967fa56c7ae10d2783d94519b1fa96395R957 but doesn't seem to need a rebase.

.github/workflows/multi_arch_ci_linux.yml
ScottTodd: We need the auth/credentials on the self-hosted runners so PRs from forks will be able to write to the external CI bucket, but we don't _need_ an expensive build server, a lower spec server would do just fine here I think. The job is pretty fast (20-60s), so I'm not too concerned either way.

See also the test I added recently: https://github.com/ROCm/TheRock/blob/0d5a00251b3eb7f6989a0d2da6b5cf53b8e971e8/.github/workflows/test_artifacts_structure.yml#L40-L42 (there, it will help to be on the same AWS network as our S3 storage to avoid egress charges... though our builders are still on Azure)

@amd-shiraz do we have lower spec CPU runners planned on either cloud?

.github/workflows/multi_arch_ci_linux.yml
ScottTodd: I think this is fine as-is, or we could reword both descriptions a bit differently. These get used in the UI, and markdown isn't supported there: 

Image

(these refer to each other by name, but the names don't appear automatically)

.github/workflows/multi_arch_ci_linux.yml
ScottTodd: 40 LOC, not sure if it's worth moving/sharing. I'll keep an eye on it if it starts to grow more.

ScottTodd:
This failed pre-commit on merge. Reverting.

PR #3860, opened 2026-03-09
Refactor `update_dependencies.py` to support server-side execution

### LINKED ISSUE #3324: Refactor `update_dependencies.py`  to move server side
The `build_tools/third_party/s3_management/update_dependencies.py` script should be refactored to run server side. Currently the script is manually executed. Once a retention policy is in place, dependencies will be tidied up and will need to be re-uploaded. Furthermore, we want to occasional check for updated dependencies instead of shipping old ones via the self-contained repository APIs. The script needs to either run when dependencies get clean up / get deleted or on a fixed interval.

erman-gurses:
Progress on: #3324
Related PR : https://github.com/ROCm/TheRock-Infra/pull/133

## Motivation:
Refactor `build_tools/third_party/s3_management/update_dependencies.py` to support server-side execution.

This change restructures `update_dependencies.py` so it can be reused in both CLI and AWS Lambda environments, following the thin-wrapper pattern.

The core reconciliation logic now lives in this script, while the Lambda entry point is implemented separately via a `lambda_function.py` wrapper on the AWS side.

---

## Architectural Change

Previously:
- `update_dependencies.py` was CLI-oriented
- bucket resolution happened at import time
- execution flow assumed direct CLI invocation
- prefix iteration was hardcoded

Now:
- the script exposes reusable runtime entrypoints
- bucket resolution happens lazily at execution time
- callers provide reconciliation scope dynamically
- the same logic can be reused by CLI, CI, or Lambda wrappers

---

## Key Changes

- Remove import-time S3 bucket initialization and implicit global bucket usage
- Add `get_s3_bucket()` for runtime bucket resolution via `--bucket` or `S3_BUCKET_PY`
- Extract reusable core logic into `run_update_dependencies()`
- Add `get_project_paths()` helper for centralized project detection
- Add `get_dependency_package_names()` helper for project-specific dependency lookup
- Add fail-fast validation for unsupported package names
- Add `detect_prefixes_from_bucket()` to discover prefixes dynamically from S3
- Add `resolve_target_prefixes()` to support:
  - explicit prefix selection via `--prefix`
  - dynamic prefix detection via `--auto-detect-prefixes` with `--base-prefix`
- Enforce paired validation for `--auto-detect-prefixes` /  `--base-prefix`
- Introduce `is_wheel_allowed()` to replace hardcoded filtering logic with explicit platform/python-tag checks
- Expand `PACKAGES_PER_PROJECT` to support multiple project dependency sets:
  - `rocm`
  - `torch`
  - `jax`
- Keep prefix selection flexible to support roots such as:
  - `v2/`
  - `v2-staging/`
  - future roots like `v3/`
- Update CLI to support explicit and dynamic prefix selection
- Remove hardcoded version/subfolder iteration from the execution path

---

## Lambda Integration

This file does not define a `lambda_handler`.

It is designed to be invoked from a thin Lambda wrapper (lambda_function.py) on the AWS side.

The Lambda wrapper is event-driven:
- S3 events provide the source of truth for:
  - bucket name
  - affected reconciliation root (e.g. v2/, v2-staging/)
  - affected prefix (e.g. v2/gfx110X-all)

Lambda wiring and deployment live outside this repo (TheRock-Infra).

Execution behavior is controlled separately:

- AUTO_DETECT_PREFIXES=true:
  derive root from the S3 event (e.g. v2/) and dynamically discover all prefixes under that root

- AUTO_DETECT_PREFIXES=false:
  derive explicit prefix from the S3 event (e.g. v2/gfx110X-all) and reconcile only that prefix

Environment variables are used only for behavior control, not for selecting architecture or prefixes.

---

## Design
This script is the reusable execution core for dependency reconciliation.

It selects target prefixes based on:
- event-derived root/prefix (Lambda)
- CLI inputs

It is responsible for:
- resolving the target bucket at runtime
- selecting supported dependency packages by project
- discovering target prefixes dynamically
- filtering unsupported wheels
- uploading only missing dependency wheels

This keeps behavior consistent across:
- CLI/manual execution
- CI usage
- AWS Lambda wrappers

---

## Behavior

- Reconciliation can be run repeatedly:
  - missing dependency wheels are uploaded
  - existing ones are skipped

---

### Key Tests (AWS end-to-end)

End-to-end testing was performed using:

S3 → EventBridge → SQS → Lambda → update_dependencies.py

---

### 1. Filtering works (non-wheel ignored)

```bash
echo test > /tmp/not-a-wheel.txt

aws s3 cp /tmp/not-a-wheel.txt \
  s3://therock-dev-python/v2/gfx110X-all/not-a-wheel.txt

aws s3 rm \
  s3://therock-dev-python/v2/gfx110X-all/not-a-wheel.txt
  ```
  Observed:

- Ignored with reason: `not a wheel`

---

### 2. Unknown wheel ignored
```bash
touch /tmp/random-package-1.0.0-py3-none-any.whl

aws s3 cp /tmp/random-package-1.0.0-py3-none-any.whl \
  s3://therock-dev-python/v2/gfx110X-all/random-package-1.0.0-py3-none-any.whl

aws s3 rm \
  s3://therock-dev-python/v2/gfx110X-all/random-package-1.0.0-py3-none-any.whl
```
Observed:
- Ignored as non-dependency package

---

### 3. Project detection (rocm / torch / jax)
```bash
touch /tmp/setuptools-1.0.0-py3-none-any.whl   # rocm
touch /tmp/jinja2-1.0.0-py3-none-any.whl       # torch
touch /tmp/ml_dtypes-1.0.0-py3-none-any.whl    # jax

aws s3 cp /tmp/setuptools-1.0.0-py3-none-any.whl s3://therock-dev-python/v2/gfx110X-all/
aws s3 cp /tmp/jinja2-1.0.0-py3-none-any.whl s3://therock-dev-python/v2/gfx110X-all/
aws s3 cp /tmp/ml_dtypes-1.0.0-py3-none-any.whl s3://therock-dev-python/v2/gfx110X-all/

aws s3 rm s3://therock-dev-python/v2/gfx110X-all/setuptools-1.0.0-py3-none-any.whl
aws s3 rm s3://therock-dev-python/v2/gfx110X-all/jinja2-1.0.0-py3-none-any.whl
aws s3 rm s3://therock-dev-python/v2/gfx110X-all/ml_dtypes-1.0.0-py3-none-any.whl
```
Observed:
- setuptools correctly mapped to rocm
- jinja2 correctly mapped to torch
- ml_dtypes correctly mapped to jax
- reconciliation triggered only for the matching project

---

### 4. Multi-root support (v2-staging/)
```bash
touch /tmp/jinja2-1.0.0-py3-none-any.whl

aws s3 cp /tmp/jinja2-1.0.0-py3-none-any.whl \
  s3://therock-dev-python/v2-staging/gfx110X-all/jinja2-1.0.0-py3-none-any.whl

aws s3 rm \
  s3://therock-dev-python/v2-staging/gfx110X-all/jinja2-1.0.0-py3-none-any.whl
```
Observed:
- Correctly routed to v2-staging/
- Reconciliation executed for that root

---

### 5. Dependency recovery
```bash
aws s3 ls s3://therock-dev-python/v2/gfx110X-all/ | grep numpy

aws s3 rm s3://therock-dev-python/v2/gfx110X-all/numpy-*.whl
```
Observed:
- Missing dependency wheels are re-uploaded
- Bucket returns to consistent state

### 6. Unit Tests
Added focused unit coverage in:

`build_tools/third_party/s3_management/tests/update_dependencies_test.py`

Coverage includes:

- `normalize_package_name()`
- `get_project_paths()`
- `get_dependency_package_names()`
- `resolve_target_prefixes()` validation:
  - explicit prefix mode
  - missing `--prefix`
  - missing `--base-prefix`
  - missing `--auto-detect-prefixes`

Executed with:

```bash
python -m pytest \
  build_tools/third_party/s3_management/tests/update_dependencies_test.py
```

---


erman-gurses:
Hi @araravik-psd , can you please give initial review for that and also I believe I need to test it on CI -  any information about it would be great.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -40,6 +36,24 @@
     "setuptools": {"versions": ["latest"], "project": "rocm"},
 }
 
+SUBFOLDERS = [
```
araravik-psd: As we are improving the update_dependencies.py, I would prefer having a way to remove this hardcoding. Have always had issues with missing new archs added and unintentionally creating new arch folders while running this, few ideas that could be iterated on:

1. Have a helper function to search for different subfolders in each directory (v2/v3/v2-staging) and populate a list of available arch sub dirs to update deps
2. As this is lambda function could to be called when an event of retention policy kicking in and removing whls or tarballs the arch can be dynamically found out and script execution could happen there(This avoids updating deps unnecessarily for all archs in the dir).

marbre:
@araravik-psd please go ahead with the first review and ping me next.

marbre (commented):
As @araravik-psd didn't reviewed yet I asked my AI agent to give it a shot. Do not be too concerned about the `Missing lambda_handler function` and simply address this with a more clear comment in the PR description.

----

## Critical Issue

### Missing `lambda_handler` function

The PR description extensively documents adding a `lambda_handler()` entry point for
AWS Lambda — including local test results showing it working — but the diff does not
include this function. The diff ends at `main()` / `if __name__ == "__main__": main()`.

Either:
- The `lambda_handler` code was accidentally left out of this PR, or
- The PR description is describing a future/separate change

This should be clarified before merging. If the Lambda handler is part of the stated
goal ("support server-side execution"), its absence is a significant gap.

---

## Style Violations

### `Any` type on `get_s3_bucket()` and `upload_missing_whls()`

Per the [TheRock Python style guide](https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#add-specific-type-hints-liberally):

> **"Never use `Any` except in rare generic code."**

```python
# Violates style guide:
def get_s3_bucket(bucket_name: str | None = None) -> Any:
    ...

def upload_missing_whls(
    bucket: Any,
    ...
```

The `Any` import was added specifically to support this. Since boto3 is untyped
(`# type: ignore[import-untyped]` is already used elsewhere in the file), a
Protocol capturing the `.name` and `.Object(key=...)` interface would be
appropriate, or at minimum `boto3.resources.base.ServiceResource`.

---

## Minor Issues

- **`get_project_paths()` silently deduplicates** via `set()` — a comment would
  clarify the intent.
- **`CLIENT` removal**: The original `CLIENT = boto3.client("s3")` is removed.
  Confirm it is not referenced elsewhere in the file outside the diff context.

---

## Positives

- Lazy S3 initialization is correct for Lambda/server-side reuse.
- `VERSION` → `version` rename for the loop variable is correct (uppercase for
  constants, not iterators).
- String formatting consistency fix (`f"..."` with inner `'` quotes) is cleaner.
- `SUBFOLDERS` promoted to module scope is the right call for shareability.
- `run_update_dependencies()` extraction with explicit keyword-only args follows
  the style guide's named-arguments guidance.

---

## Test Plan Gap

The PR notes: *"Plan to test on CI — have not done yet."* No CI results are linked.
Since this touches S3 upload logic, a dry-run CI pass should be required before merging.

---

## Verdict

**Block on:**
1. Clarify/add the `lambda_handler` that the PR description claims is there.
2. Fix `Any` type to comply with the [style guide](https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md).
3. Link CI dry-run results before merging.

The core refactor structure is sound.

erman-gurses:
> ress this with a more clear comment in the PR description.

Yes, the PR description is outdated after the last changes -  will update it. I was able to test it on AWS side and it is working. Arvind will check those results. 

Updated!

erman-gurses:
@araravik-psd , can you please give the initial review for this.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -41,7 +36,6 @@
         "cp311",
         "cp312",
         "cp313",
-        "cp314",
```
araravik-psd, line 44: Was this removed on purpose or removed by mistake?

araravik-psd (approved):
LGTM

araravik-psd:
@erman-gurses LGTM, we checked that the tests are passing too for different scenarios on the call. @marbre Please go ahead with your review when you get chance

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -162,82 +243,138 @@ def upload_missing_whls(
...
+    parser.add_argument(
+        "--auto-detect-prefixes",
+        action="store_true",
+        help=(
+            "Automatically detect architecture prefixes under the given base "
+            "path using S3 CommonPrefixes."
+        ),
+    )
+    parser.add_argument(
+        "--starting-from",
```
marbre: Maybe call it `--base-prefix` as in the help text?

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -162,82 +243,138 @@ def upload_missing_whls(
...
-        set(pkg_info["project"] for pkg_info in PACKAGES_PER_PROJECT.values())
-    )
+    parser = ArgumentParser("Upload dependent packages to S3")
+    project_paths = get_project_paths()
     parser.add_argument("--package", choices=project_paths, default="torch")
+    parser.add_argument("--bucket", type=str, help="S3 bucket name")
+    parser.add_argument(
+        "--prefix",
+        type=str,
+        help="Explicit prefix to update",
```
marbre: Please provide an example here.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -67,6 +61,85 @@
...
+    )
+
+
+def get_dependency_package_names(project: str) -> frozenset[str]:
+    """Return dependency package names for the given project."""
+    return frozenset(
+        pkg_name
+        for pkg_name, pkg_info in PACKAGES_PER_PROJECT.items()
+        if pkg_info["project"] == project
+    )
```
marbre, line 112:
```suggestion
```
This seems to be dead code which is never called. If is used by the lambda wrapper, add a comment to make clear it is actually used.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -9,16 +9,10 @@
...
+from botocore.exceptions import ClientError
 
-S3 = boto3.resource("s3")
-CLIENT = boto3.client("s3")
-# We also manage `therock-nightly-python` (not the default to make the script safer to test)
-BUCKET = S3.Bucket(getenv("S3_BUCKET_PY", "therock-dev-python"))
-# Note: v2-staging first, in case issues are observed while the script runs
-# and the developer wants to more safely cancel the script.
-VERSIONS = ["v2-staging", "v2"]
+import re
```
marbre: NIT: Move `import re` to above the boto imports to match PEP 8 grouping (stdlib then third-party).

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -163,82 +248,164 @@ def upload_missing_whls(
...
-        "gfx101X-dgpu",
-        "gfx103X-all",
-        "gfx110X-all",
-        "gfx1150",
-        "gfx1151",
-        "gfx120X-all",
-        "gfx90X-dcgpu",
-        "gfx94X-dcgpu",
-        "gfx950-dcgpu",
-    ]
```
marbre, line 218: Removing the hardcoded SUBFOLDERS list means CLI users must now know which prefixes exist. The --auto-detect-prefixes --starting-from v2/ flow recreates the old behavior. As this is a change of the behavior, let us consider adding an example to the script's --help or a docstring at the top of the file so a future operator doesn't have to read the diff.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -67,6 +61,85 @@
...
+
+    if auto_detect_prefixes:
+        if not starting_from:
+            raise RuntimeError(
+                "--starting-from must be provided when using --auto-detect-prefixes"
+            )
+        return detect_prefixes_from_bucket(bucket, starting_from)
+
+    raise RuntimeError(
+        "Must provide either --prefix or --auto-detect-prefixes with " "--starting-from"
```
marbre:
```suggestion
        "Must provide either --prefix or --auto-detect-prefixes with --starting-from"
```

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -163,82 +248,164 @@ def upload_missing_whls(
...
+        pkg_name: pkg_info
+        for pkg_name, pkg_info in PACKAGES_PER_PROJECT.items()
+        if pkg_info["project"] == package
+        and (
+            normalized_dependency_names is None
+            or normalize_package_name(pkg_name) in normalized_dependency_names
+        )
+    }
+    if not selected_packages:
+        raise ValueError(f"No dependency packages selected for project '{package}'")
```
marbre, line 376: From my review agent:

----

Two distinct failure modes both collapse into one error message:

1. **Genuine empty configuration**: `--dependency-package` wasn't passed, and
   `PACKAGES_PER_PROJECT` has no entries with `project == package`. (Today
   this can't happen for `rocm`/`torch`/`jax` because each project has at
   least one entry, but the code does not assume that.)
2. **User typo / cross-project mismatch**: every requested name in
   `--dependency-package` failed to match. Examples:
   - `--package torch --dependency-package numyp` (typo)
   - `--package torch --dependency-package setuptools` (`setuptools` exists,
     but belongs to `rocm`, not `torch`)
   - `--package jax --dependency-package ml-types` (does not exist anywhere)

In case 2 the user gets back the exact same message they'd get in case 1:

```
No dependency packages selected for project 'torch'
```

That message tells them nothing about *which* of the names they passed didn't
match, or *why* (typo vs. wrong project). For a CLI that an operator may
invoke at 2am to repair a bucket after a missed Lambda event, that's a real
ergonomics problem — they have to re-read the source to figure out the valid
names per project.

See https://gist.github.com/marbre/09f7ee014d013564489472546b3f0383 for a suggestion how to fix.

marbre (commented):
Looks  mostly good but tests have not been touched yet as part of this PR.

`test_update_dependencies.py` only exercises `is_wheel_allowed`, which is unchanged. The PR adds nontrivial logic that's not covered by any unit test:

- `resolve_target_prefixes` validation (the `--starting-from` /  `--auto-detect-prefixes` paired requirement is exactly the kind of CLI contract that's easy to break).
- `normalize_package_name`.
- `get_project_paths` / `get_dependency_package_names` selection logic.

Pure-function tests for at least `resolve_target_prefixes` and `normalize_package_name` would be low cost and high value.

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: It is used at `dependency_event_filter.py`. I added comment to make it clear.

erman-gurses:
> Looks mostly good but tests have not been touched yet as part of this PR.
> 
> `test_update_dependencies.py` only exercises `is_wheel_allowed`, which is unchanged. The PR adds nontrivial logic that's not covered by any unit test:
> 
> * `resolve_target_prefixes` validation (the `--starting-from` /  `--auto-detect-prefixes` paired requirement is exactly the kind of CLI contract that's easy to break).
> * `normalize_package_name`.
> * `get_project_paths` / `get_dependency_package_names` selection logic.
> 
> Pure-function tests for at least `resolve_target_prefixes` and `normalize_package_name` would be low cost and high value.

Added focused unit coverage in update_dependencies_test.py for the new pure-function logic:

- normalize_package_name()
- get_project_paths()
- get_dependency_package_names()
- resolve_target_prefixes() validation for explicit prefix, missing prefix/auto-detect, missing --base-prefix, and missing --auto-detect-prefixes

The tests keep the existing is_wheel_allowed coverage intact.

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: Added

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: Addressed them

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: Fixed

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: Added

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: Moved

build_tools/third_party/s3_management/update_dependencies.py
erman-gurses: changed to `--base-prefix`

marbre (commented):
Thanks for addressing the review feedback! One more thing before this can go in:

The PR doesn't touch `build_tools/third_party/s3_management/README.md`, but its documented invocations no longer work:

  - README:92 update_dependencies.py --package rocm --dry-run
  - README:99 update_dependencies.py --package rocm
  - README:117 update_dependencies.py --package rocm

Update the README with the equivalent invocation that recreates the previous behavior, e.g.

```bash
  export S3_BUCKET_PY=therock-dev-python
  python ./build_tools/third_party/s3_management/update_dependencies.py \
      --package rocm --auto-detect-prefixes --base-prefix v2/ --dry-run
```

marbre (approved):

PR #3866, opened 2026-03-10
Adding code obj version backward comptability test

dkrottap:
Adding backward comptability test for code object versions. 
Removed redundant json file. We need to build for each code obj version which doesn't take much time and will not use any gpu resources. 
Will rebase this PR once 3648 PR is merged for function base and test matrix files.

### DIFF
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -334,11 +334,12 @@ def retrieve_artifacts_by_run_id(args):
         "core-ocl_dev",
         "rocprofiler-sdk_lib",
         "host-suite-sparse_lib",
-        "core-cov-tests_test",
     ]
 
     if args.base_only:
         argv.extend(base_artifact_patterns)
+        if args.tests:
+            argv.append("core-cov-tests_test")
     elif any(
         [
             args.aqlprofile,
@@ -460,6 +461,7 @@ def retrieve_artifacts_by_run_id(args):
         extra_artifact_patterns = [f"{a}_lib" for a in extra_artifacts]
         if args.tests:
             extra_artifact_patterns.extend([f"{a}_test" for a in extra_artifacts])
+            extra_artifact_patterns.append("core-cov-tests_test")
 
         argv.extend(extra_artifact_patterns)
     else:
--- a/tests/extended_tests/functional/functional_test_matrix.py
+++ b/tests/extended_tests/functional/functional_test_matrix.py
@@ -33,7 +33,7 @@ functional_matrix = {
     },
     "test_code_obj_version": {
         "job_name": "test_code_obj_version",
-        "fetch_artifact_args": "--base-only",
+        "fetch_artifact_args": "--base-only --tests",
         "timeout_minutes": 5,
         "test_script": (
             "python " f"{_get_functional_script_path('test_code_obj_version.py')}"


tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+        if not self.include_dir.exists():
+            raise TestExecutionError(f"Include path not found: {self.include_dir}")
+
+    def _resolve_tool(self, candidates: List[Path], default_tool: str) -> str:
+        for candidate in candidates:
+            if candidate.exists():
+                return str(candidate)
+        return default_tool
+
+    def _compile_sample(self, output_binary: Path, code_obj_version: int = None) -> int:
```
geomin12: how long does this take? we generally want to keep builds to TheRock build step, and this should be responsible for testing only

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+                        if len(bundle_errors) > 3:
+                            offloading_reason += "; ..."
+
+        raise TestExecutionError(
+            "Could not determine code object version from binary "
+            f"{binary_path}. Offloading path: {offloading_reason}. "
+            "Host-binary readelf fallback is disabled."
+        )
+
+    def _run_and_validate_binary(self, binary_path: Path) -> Tuple[int, str, str]:
```
geomin12: what is the diff between this and `run_tests`?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+                    "PASS",
+                    0,
+                    detected_version=detected,
+                )
+
+        with open(self.results_json, "w") as f:
+            json.dump(self.test_results, f, indent=2)
+        log.info(f"{self.display_name} results saved to {self.results_json}")
+
+    def parse_results(self) -> List[Dict[str, Any]]:
```
geomin12, line 128: this seems overkill, can we not just report findings as find them? why write to a json file when it just gets written to and immediately gets read and reports results?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+import re
+import shlex
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[2]))  # For utils
+sys.path.insert(0, str(Path(__file__).resolve().parent))  # For functional_base
```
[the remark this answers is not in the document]
dkrottap: Will rebase once the 3648 PR is merged

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+from utils.logger import log
+
+
+class TargetIdBitExtractTest(FunctionalBase):
+    """Validate bit_extract code object backward compatibility."""
+
+    def __init__(self):
+        super().__init__(
+            test_name="cov_backward_comptability",
+            display_name="Cov Backward Comptability",
```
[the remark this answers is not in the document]
dkrottap: done

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
...
+    def _run_and_validate_binary(self, binary_path: Path) -> Tuple[int, str, str]:
+        env = self._get_rocm_env_with_path()
+        rc, output = self._execute_command_with_output(
+            [str(binary_path)], cwd=self.sample_dir, env=env
+        )
+        if rc != 0:
+            return rc, output, f"Binary execution failed with return code {rc}"
+        if "PASSED!" not in output:
+            return rc, output, "Binary did not print expected token: PASSED!"
+        return rc, output, ""
```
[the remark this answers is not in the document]
dkrottap: compile step does not use gpu resources

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,442 @@
+"""
+code object version  functional test.
+
+Build default bit_extract, detect code object version n, rebuild with n-1/n-2,
+then run all binaries and verify they print "PASSED!".
+"""
```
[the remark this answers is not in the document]
dkrottap: will add it during rebase

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: This takes very less time not even seconds and does not utilize any gpu resources.

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: run_and_validate_binary -- is a lower level method and it does the following executes single compiled binary and validates the o/p where as run_tests  compiles with the default binary detects the cov and calls run_and_validate_binary which executes the binary and records the result

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: done removed writing to json file

geomin12 (commented):
lgtm, but can you validate this passes with a CI Nightly workflow dispatch run?

dkrottap:
> lgtm, but can you validate this passes with a CI Nightly workflow dispatch run?

Please find the passing ci nightly for mi300 https://github.com/ROCm/TheRock/actions/runs/23556534570

tests/extended_tests/functional/functional_test_matrix.py
```diff
@@ -31,4 +31,14 @@ def _get_functional_script_path(script_name: str) -> str:
         "platform": ["linux"],
         "total_shards": 1,
     },
+    "test_code_obj_version": {
+        "job_name": "test_code_obj_version",
+        "fetch_artifact_args": "--tests",
```
geomin12: can we import the project directly? just `--tests` imports a **ton** of packages, which is wasted resources

is there a `--code-obj-version` flag? can we add one?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+code object version  functional test.
+
+Build default bit_extract, detect code object version n, rebuild with n-1/n-2,
```
geomin12: please re-write this, hard to read

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+        self.sample_relative_path = "projects/hip-tests/samples/0_Intro/bit_extract"
+        self.include_relative_path = "projects/hip-tests/samples/common"
+        self.source_file = "bit_extract.cpp"
+        self.binary_prefix = "bit_extract"
+
+        self.rocm_systems_dir = self.therock_dir / "rocm-systems"
+        self.sample_dir = self.rocm_systems_dir / self.sample_relative_path
+        self.include_dir = self.rocm_systems_dir / self.include_relative_path
+
+    def _execute_command_with_output(
```
geomin12: i think the functional_base has this, can we re-use?

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+        output_lines = []
+        for line in process.stdout:
+            line_text = line.rstrip()
+            log.info(line_text)
+            output_lines.append(line_text)
+
+        process.wait()
+        return process.returncode, "\n".join(output_lines)
+
+    def _get_rocm_env_with_path(self) -> Dict[str, str]:
```
geomin12: same here, i think this is shared stuff, no de-dup = better code! let's re-use when we can

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+            str(self.rocm_path / "bin"),
+            str(self.rocm_path / "llvm" / "bin"),
+            "/opt/rocm/llvm/bin",
+        ]
+        existing_path = env.get("PATH", "")
+        env["PATH"] = ":".join(d for d in extra_dirs + [existing_path] if d)
+        env["HIP_PLATFORM"] = "amd"
+        return env
+
+    def _ensure_sources_ready(self) -> None:
```
geomin12: please remove. **we should not update rocm-systems here**. TheRock already does rocm-systems submodule bumps.

What happens if something is broken in rocm-systems? this test will fail if we do a careless submodule update

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+        if not self.include_dir.exists():
+            raise TestExecutionError(f"Include path not found: {self.include_dir}")
+
+    def _resolve_tool(self, candidates: List[Path], default_tool: str) -> str:
+        for candidate in candidates:
+            if candidate.exists():
+                return str(candidate)
+        return default_tool
+
+    def _compile_sample(self, output_binary: Path, code_obj_version: int = None) -> int:
```
geomin12: we should handle build in TheRock, this is only for testing - building with CPU here is wasted resources

tests/extended_tests/functional/scripts/test_code_obj_version.py
geomin12: okay, please add a comment in code to reflect this, so we can tell the difference

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+        )
+
+    def _run_and_validate_binary(self, binary_path: Path) -> Tuple[int, str, str]:
+        env = self._get_rocm_env_with_path()
+        rc, output = self._execute_command_with_output(
+            [str(binary_path)], cwd=self.sample_dir, env=env
+        )
+        if rc != 0:
+            return rc, output, f"Binary execution failed with return code {rc}"
+        if "PASSED!" not in output:
```
geomin12: ? this is brittle. what if the code gets updated and `PASSED!` is not the case? can we use exit codes 0 or 1 for checking? any error messages? i see above, we check for 0 or 1. what is this?

if needed, add comment pls

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,440 @@
...
+        log.info(f"Running {self.display_name} Tests")
+        self._ensure_sources_ready()
+
+        # 1) Build with default compiler behavior (no explicit code object version),
+        #    then detect current code object version n from the produced binary.
+        default_variant = "default_build_detect_n"
+        default_compile_cmd = f"hipcc {self.source_file} -I {self.include_dir}"
+        default_binary = self.sample_dir / f"{self.binary_prefix}_default"
+
+        rc = self._compile_sample(default_binary, code_obj_version=None)
```
geomin12: compilation should be done in TheRock. tests shouldn't compile...

What if a developer wants all tests / build artifacts from just a build? they won't be included as this is compiled in the test executable (outside of build)

tests/extended_tests/functional/functional_test_matrix.py
```diff
@@ -31,4 +31,14 @@ def _get_functional_script_path(script_name: str) -> str:
...
+        "timeout_minutes": 30,
```
geomin12: i see it ran for ~ 15 seconds. let's reduce this time to 5 mins

dkrottap:
> build / compilation **must** be done in TheRock's build side, not in test executable scripts.

This test validates supported code object versions through a lightweight build, requiring minimal compilation time and negligible CPU resources.

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: modified accordingly to handle in functional base

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: Done

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: done modified accordingly

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: The purpose of this test is to compile it with default and check the backward compatibility with the supported code object versions .. it does not consume any cpu resources. compilation time is very less

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: modified accordingly

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: This test verifies compatibility across different code object versions. Note that packaging multiple code object versions in build artifacts is an anti-pattern; this multi-version setup is isolated to this specific test.

tests/extended_tests/functional/functional_test_matrix.py
dkrottap: modified to base only to fetch rocm runtime/compiler artifacts which are required for bit extract.

tests/extended_tests/functional/functional_test_matrix.py
dkrottap: done

tests/extended_tests/functional/scripts/test_code_obj_version.py
dkrottap: Removed dependency on rocm-systems submodule

tests/extended_tests/functional/functional_test_matrix.py
geomin12: please modify, it was not modified

dkrottap:
> we need to have build on TheRock's build side.
> 
> What if a developer wants to run this test? if we run install artifacts with `--base-only` and, they would expect this test to be available without any extra effort (no additional compilation)
> 
> Also for release packages (pip install) - this executable will not be available as it is compiled in the test side (during test run), a developer should not have to run:
> 
> ```
> pip install rocm
> python test/extended_tests/functional/scripts/test_code_obj_version.py
> ```
> 
> to get this executable, which is unnecessary
> 
> they should only be able to run `pip install rocm` and have all these tests / packages available

Separated the COV backward compatibility test into a build phase and a test phase — the build phase compiles a self-contained hipRTC executable  during TheRock's build and packages it as the core-cov-tests artifact, while the test phase simply runs the pre-built binary and parses its JSONL output, removing the need for rocm-systems, hipcc, or any LLVM tools at test time. 
Please find CI nightly for the same https://github.com/ROCm/TheRock/actions/runs/23919072345

build_tools/install_rocm_from_artifacts.py
```diff
@@ -334,6 +334,7 @@ def retrieve_artifacts_by_run_id(args):
         "core-ocl_dev",
         "rocprofiler-sdk_lib",
         "host-suite-sparse_lib",
+        "core-cov-tests_test",
```
geomin12: i know that we have a `--tests` flag. in the case that `--tests` is not available, can we ignore this package?

dkrottap:
> nevermind, i see tests run. let's resolve that other comment though
> 
> also noticing this error in logs: `Error: ]: API upload failed for 'cov_backward_compatibility' functional results. Results were saved locally to '/__w/TheRock/TheRock/tests/extended_tests/functional/results'.`, should we be uploading? should we ignore this?

We can ignore this error. The results API endpoint (BENCHMARK_DB_URL) is not configured in this CI environment, so the upload fails gracefully. The test results are still saved locally. This is the same behavior seen by other extended tests when the API environment variables are not set — it's a cosmetic warning, not a test failure.

tests/extended_tests/functional/srcs/cov_backward_compat_hiprtc.cpp
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * AMDGPU Code Object Version Backward Compatibility Test (hipRTC)
+ *
+ * Self-contained executable that uses hipRTC to compile a simple vector-add
+ * kernel with the default code object version and then with the two prior
+ * versions (n-1 and n-2).  For each variant it:
+ *   1. Compiles the kernel source at runtime via hipRTC.
+ *   2. Detects the actual code object version embedded in the compiled blob.
+ *   3. Loads and launches the kernel through the HIP driver API.
+ *   4. Verifies the computation results on the host.
```
ScottTodd, line 13: Drive-by: did you consider putting this test in hip-tests in rocm-systems or another existing subproject?

It feels fairly heavy-handed to create a new artifact in TheRock for a single test like this.

dkrottap: COV backward compatibility is an integration-level concern. It validates that the current runtime can load and execute code objects compiled with older versions. This spans both the compiler which produces COV  and the HIP/ROCr runtime  which loads it, so it doesn't belong solely to hip-tests or any single rocm-systems component. It's added as part of the extended tests framework. The artifact is minimal just the single binary the test harness needs.

build_tools/install_rocm_from_artifacts.py
dkrottap: had this package inside tests. please find ci nightly with the latest changes
https://github.com/ROCm/TheRock/actions/runs/24165298591

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,153 @@
...
+                return candidate
+
+        searched = "\n  ".join(str(c) for c in candidates)
+        raise TestExecutionError(
+            f"Pre-built test binary '{self.BINARY_NAME}' not found.\n"
+            f"Searched:\n  {searched}\n"
+            "Ensure the cov-backward-compat artifact was built and installed.\n"
+            "The binary is built as part of TheRock when THEROCK_BUILD_TESTING "
+            "is enabled and is included in the core-cov-tests artifact."
+        )
```
geomin12, line 65: pretty wordy, i think we can just do a global variable with path (as we know expect it to be there from artifacts - if not there, that's a larger issue at hand)

tests/extended_tests/functional/scripts/test_code_obj_version.py
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    "test_suite": "cov_backward_compat",
+                    "test_case": variant,
+                    "command": str(binary),
+                    "return_code": 0 if status == "PASS" else 1,
+                    "status": status,
+                    "detected_code_object_version": detected_cov,
+                    "info": cov_info,
+                    "error": error,
+                }
```
geomin12, line 112: what is purpose of this? is this a duplicate of `parse_results`? should we just do all the parsing in one method?

tests/extended_tests/functional/srcs/cov_backward_compat_hiprtc.cpp
geomin12: i think this can be moved to rocm-systems, as it is a `system` related change, that way we don't need to include this file in here.

can be done with this PR or in a future cleanup PR

CMakeLists.txt
```diff
@@ -477,6 +477,7 @@ endif()
 # Testing
 ################################################################################
 
+message(STATUS "THEROCK_BUILD_TESTING=${THEROCK_BUILD_TESTING} (before extended_tests)")
```
geomin12, line 480: can delete, seems to be a test message

dkrottap:
will comeup with a new PR after discussing internally whether to have it as part of extended tests or part of rocm-systems.

PR #3875, opened 2026-03-10
Revert "Include rocr-debug-agent-tests in Python package (#3815)"

marbre:
This reverts commit f5a25316bc84ec3945e1f28e059c6bfa8385e048.

Test artifacts are entirely excluded from Python packaging and thus rocr-debug-agent-tests should not be added to the list as it won't have any effect.

build_python_packages.py builds four package types - core, libraries, meta, and devel - using two artifact filters:

* core_artifact_filter: accepts lib and run components only (plus dev for HIP headers)
* libraries_artifact_filter: accepts lib component only

Neither filter includes the test component. There is no test package type defined and dedicated test artifacts are intentionally excluded.

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -155,7 +155,6 @@ def core_artifact_filter(an: ArtifactName) -> bool:
         "rocjpeg",
         "rocprofiler-sdk",
         "rocr-debug-agent",
-        "rocr-debug-agent-tests",
         "sysdeps",
         "sysdeps-amd-mesa",
         "sysdeps-expat",


lumachad:
This looks OK to me. Does it mean testing is not expected to be done with python packages or does it mean rocr-debug-agent needs to adapt its testing artifact to follow some other pattern?

marbre:
> This looks OK to me. Does it mean testing is not expected to be done with python packages or does it mean rocr-debug-agent needs to adapt its testing artifact to follow some other pattern?

I'd say it would need a broader discussion to decide if we want to support testing via Python packages at all or not. If we conclude that this should be supported, we may want to add a `rocm-sdk-test` package eventually. Anyway, so far I'd say this is not a bug and rather a feature that was never requested or implemented.

lumachad:
> > This looks OK to me. Does it mean testing is not expected to be done with python packages or does it mean rocr-debug-agent needs to adapt its testing artifact to follow some other pattern?
> 
> I'd say it would need a broader discussion to decide if we want to support testing via Python packages at all or not. If we conclude that this should be supported, we may want to add a `rocm-sdk-test` package eventually. Anyway, so far I'd say this is not a bug and rather a feature that was never requested or implemented.

Thanks. That clarifies it for me.

marbre:
Test failures are unrelated.

Issue #3876, opened 2026-03-10
[Issue] Jax Build Fails: Missing `libamd_comgr_stub.a` When Linking `jaxlib/rocm/_hybrid.so`


chiranjeevipattigidi:
The build fails during the linking stage of `jaxlib/rocm/_hybrid.so` due to a missing file:
`libamd_comgr_stub.a.`

**Platform:** `Linux`
**Effected Archs:** `All supported archs`
**Jax Versions:** `0.8.2, 0.9.0`

### Run:
- https://github.com/ROCm/TheRock/actions/runs/22899060682/job/66440631650
- https://github.com/ROCm/TheRock/actions/runs/22897835891

```
[13,793 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 20s local ... (64 actions, 62 running)
[13,814 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 21s local ... (65 actions, 64 running)
[13,835 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 22s local ... (65 actions, 64 running)
[13,855 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 23s local ... (65 actions, 64 running)
[13,881 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 24s local ... (65 actions, 63 running)
[13,914 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 25s local ... (63 actions, 61 running)
[13,941 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 26s local ... (63 actions, 61 running)
[13,964 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 27s local ... (59 actions running)
[13,988 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 28s local ... (35 actions running)
[14,009 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 29s local ... (15 actions running)
[14,019 / 14,849] Compiling src/google/protobuf/descriptor.cc [for tool]; 30s local ... (6 actions, 5 running)
[14,021 / 14,849] Compiling jaxlib/gpu/prng_kernels.cu.cc; 16s local ... (6 actions, 4 running)
[14,049 / 14,849] Compiling jaxlib/gpu/prng_kernels.cu.cc; 17s local ... (9 actions, 8 running)
[...]

PR #3872, opened 2026-03-10
Fix: Remove trailing comma in `package.json` to resolve packaging build failure

chiranjeevipattigidi:
## Motivation
To fix packaging build failures(eg: https://github.com/ROCm/TheRock/actions/runs/22899062783)

This update removes the trailing comma that was left out during the changes introduced in the following commit: https://github.com/ROCm/TheRock/commit/864f3d35fe87c520a78ee3e28b0dc09a0c8bd291#diff-070517c2975180a7276c82419c6a28f2ec176b2eb83daf8723844f6856187efbR242
## Test Result

passing - https://github.com/ROCm/TheRock/actions/runs/22899879672/job/66443429314

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3877, opened 2026-03-10
Revert "Remove comgr delay-load stub; statically link LLVM into comgr"

chiranjeevipattigidi:
Reverts ROCm/TheRock#3731
Jax (v0.8.2 , v0.9.0) builds are failing due missing of `libamd_comgr_stub.a` during linking process
more details: https://github.com/ROCm/TheRock/issues/3876

### Test Plan
Testing: https://github.com/ROCm/TheRock/actions/runs/22904894874
### Test Result
TBD


amd-chiranjeevi:
there is another PR https://github.com/ROCm/TheRock/pull/3872   to fix the small bug(missed removing trailing comma in a json file) introduced by original pr.
If we agree to merge this revert , can close the https://github.com/ROCm/TheRock/pull/3872.

stellaraccident:
Neither of these are good candidates for revert. For build issues not covered by CI, please fix forward (and consider how to add CI coverage so it doesn't happen again).

stellaraccident:
> ```
> ERROR: /root/.cache/bazel/_bazel_root/35c03de1b96cb539bbf04926e943e4d4/external/jax/jaxlib/rocm/BUILD:387:19: Linking external/jax/jaxlib/rocm/_hybrid.so failed: missing input file '@@local_config_rocm//rocm:rocm_dist/lib/libamd_comgr_stub.a'
> ```
> 
> seems like this rather needs a fix in JAX or at some other place. The revert is not the right long term fix. Before accepting a temporarily revert we should triage the root cause more carefully.

Jax has a really fragile manual list of libraries that needs update. Until that is fixed in a more robust way, we can't block landing changes like this (there is another one coming today that will also change the list of libraries Jax must bundle).

PR #3879, opened 2026-03-10
updating test runner label to 8gpu ossci node

rahulc-gh:
## Motivation

The OSSCI cluster - linux-mi325-4gpu-ossci-rocm is no longer active and should be changed to linux-mi325-8gpu-ossci-rocm.


geomin12 (changes requested):
wrong branch / commit?

i am seing rocm-systems submodule bump

geomin12 (commented):
also we don't use any instances of 4gpu, it has all been converted to 8gpu (as far as i am aware)

HereThereBeDragons:
can we close this?

HereThereBeDragons:
please reopen it in case it is still relevant

PR #3481, opened 2026-02-18
Run Linux Nightly Package Build Entirely Inside Container

chiranjeevipattigidi:
## Motivation

Removing uncessary calls for helper scripts `build_tools/linux_portable_build.py`, `build_tools/detail/linux_portable_build_in_container.sh` in [release_portable_linux_packages.yml](https://github.com/ROCm/TheRock/blob/main/.github/workflows/release_portable_linux_packages.yml)

## Technical Details
**Removed Invocation of:**
-  [build_tools/linux_portable_build.py](https://github.com/ROCm/TheRock/tree/main/build_tools/linux_portable_build.py),  
- [build_tools/detail/linux_portable_build_in_container.sh](https://github.com/ROCm/TheRock/tree/main/build_tools/detail/linux_portable_build_in_container.sh)  

**Added:**
- native `container:` tag in the workflow 
- Defined container image directly in workflow configuration
- Executed CMake commands directly in workflow steps

## Files Changed
- `.github/workflows/release_portable_linux_packages.yml`: added `container` tag and launch cmake commands in workflow
- `build_tools/github_actions/build_configure.py`: added support to accept `BUILD_DIR` as a command-line option

## Test Plan

Linux Packages - https://github.com/ROCm/TheRock/actions/runs/22212899021

## Test Result

- Build completed without errors
- No regression observed
- Artifacts generated correctly

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3881, opened 2026-03-10
Add mount command to linux_portable_build.py

PeterCDMcLean:
## Motivation

Allow arbitrary mounts to linux_portable_build.py . Utility is for rocm-npi-dev git_reference repo mount so fprints work properly

## Technical Details

Add the option and allow for either src:dst asymmetric mapping or path:path symmetric mapping

## Submission Checklist

- [x] pre-commit run

### DIFF
--- a/build_tools/linux_portable_build.py
+++ b/build_tools/linux_portable_build.py
@@ -66,6 +66,20 @@ def do_build(args: argparse.Namespace, *, rest_args: list[str]):
         ]
     )
 
+    # Add user-specified bind mounts.
+    if args.mounts:
+        for mount_spec in args.mounts:
+            if ":" in mount_spec:
+                src, dst = mount_spec.split(":", 1)
+            else:
+                src = dst = mount_spec
+            cl.extend(
+                [
+                    "--mount",
+                    f"type=bind,src={src},dst={dst}",
+                ]
+            )
+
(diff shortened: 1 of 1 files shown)


build_tools/linux_portable_build.py
ScottTodd, line 1: Is this script still in use? We've been migrating workflows away from it, to just use `container:` for the entire job instead of using a wrapper script like this for specific steps within the job. See https://github.com/ROCm/TheRock/pull/3481 for example.

Could still be useful for developers, but I think we could consider deleting the script.

PeterCDMcLean: `rocm-npi-dev` is still using this script in its github workflows. I'm sure those workflows could eventually do something similar. We can fork that particular script to the rocm-npi-dev to help the transition.

ScottTodd: As a general rule of thumb, open source code should stand on its own. If we have no uses here on github, it should be safe to delete the script. I think rocm-npi-dev should adapt to that PR.

PeterCDMcLean:
This is no longer necessary. First git alternates are no longer expected to be used by rocm-npi-dev, and rocm-npi-dev _may_ fork the scripts entirely to own them.

PR #3882, opened 2026-03-10
[rocshmem] Add missing artifacts

Yiltan:
Some artifacts were missing - this PR adds them.

AIROCSHMEM-327


marbre (commented):
I think this needs more work. `share/rocmshem` seems to be not in lib artifacts I recently checked. Instead this has duplicates in test and dev. Anyway, why is librocshmem a static library and not a dynamic library? Static libraries should only be needed at build time and must not be runtime dependencies.

This will also be affected by https://github.com/ROCm/TheRock/pull/3830.

Yiltan (commented):
@marbre 

I've updated the paths given the feedback


"Why is librocshmem a static library and not a dynamic library? Static libraries should only be needed at build time and must not be runtime dependencies."

rocSHMEM only produces a static library. The library generates device code that users compile into their kernel. To my knowledge there isn't a way to generate a dyanmic librariy. I think other libraries CK do something similar  

Also what specifically will be affected by https://github.com/ROCm/TheRock/pull/3830.

marbre:
> "Why is librocshmem a static library and not a dynamic library? Static libraries should only be needed at build time and must not be runtime dependencies."
> 
> rocSHMEM only produces a static library. The library generates device code that users compile into their kernel. To my knowledge there isn't a way to generate a dyanmic librariy. I think other libraries CK do something similar

I think in that case it is fine if `librocshmem.a` only goes into the `dev` artifact which is intended to provide those pieces a user will pull to develop something based upon the component. `lib` on the other side is more or less something you pull for runtime but this does not hold true for a static library. Wdyt?

Yiltan:
@marbre , that sounds reasonable to me.

I've moved the rocshmem library to dev. 

However I have kept the include in lib and added an additional exclude. In a future rocSHMEM PR we will be creating LLVM bitcode files `*.bc` and we would like them to be a part of the the runtime libs

marbre (commented):
> @marbre , that sounds reasonable to me.
> 
> I've moved the rocshmem library to dev.
> 
> However I have kept the include in lib and added an additional exclude. In a future rocSHMEM PR we will be creating LLVM bitcode files `*.bc` and we would like them to be a part of the the runtime libs


Keeping the include is fine as long as it doesn't create duplicates. There are merge-conflicts however with need to be resolved, most likely due to https://github.com/ROCm/TheRock/commit/90554eea98f7005d790ebba0291ad11abda39ff1 which reordered the artifacts to match the extends chain.

Yiltan:
@marbre , I'm pretty happy with this, it should be good for review

It creates what we want and doesn't create duplicates

comm-libs/artifact-rocshmem.toml
```diff
@@ -1,16 +1,29 @@
...
+  "share/doc/**",
+]
 
 # Tests
 [components.test."comm-libs/rocshmem/stage"]
 optional = true
+include = [
+  "bin/rocshmem_functional_tests",
+  "share/rocshmem/rocshmem_functional_driver.sh",
+]
```
marbre, line 29:
```
$ tar tvf rocshmem_test_gfx94X-dcgpu.tar.xz 
-rw-r--r-- root/root        25 2026-03-18 18:17 artifact_manifest.txt
```
Seems those files are not present?

marbre (commented):
Please request a re-review next time. I am checking the review queue way more frequently that mentions.

PR #3884, opened 2026-03-10
[rocprofiler-systems] Disable transpose runtime-instrument test

kcossett-amd:
## Motivation

I've noticed that since https://github.com/ROCm/TheRock/pull/3731 was merged, our `runtime-instrument` test for `transpose` would timeout.

After my investigation, the best course of action for now is to disable it.

## Technical Details

I noticed that `transpose → libamdhip64.so.7 → libamd_comgr.so → libLLVM.so + libclang-cpp.so` 
The last two shared libraries are huge and Dyninst fails to parse them, causing failure after ~450 seconds.
Even if you exclude both libraries, the test will still take ~370 seconds, which is far too long for now.

A more permanent solution requires some changes to how `rocprof-sys-instrument` works at a core level.

## Test Plan

Disable the failing test.

## Test Result

See workflows.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
@@ -47,9 +47,9 @@ pytest_package_exec = (
 cmd = [
     sys.executable,
     str(pytest_package_exec),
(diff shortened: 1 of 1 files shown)


kcossett-amd:
`rocprofiler-systems` tests pass

Issue #3873, opened 2026-03-10
[Issue] Native packaging build fails while parsing `package.json`


chiranjeevipattigidi:
**Platform:** `Linux`
### Runs:
- https://github.com/ROCm/TheRock/actions/runs/22899062783/job/66440637963

```
Building deb packages for gfx1153 22887330403
Traceback (most recent call last):
  File "/home/runner/_work/TheRock/TheRock/./build_tools/packaging/linux/build_package.py", line 919, in <module>
In function: clean_package_build_dir
    main(sys.argv[1:])
  File "/home/runner/_work/TheRock/TheRock/./build_tools/packaging/linux/build_package.py", line 915, in main
    run(args)
  File "/home/runner/_work/TheRock/TheRock/./build_tools/packaging/linux/build_package.py", line 801, in run
    pkg_list, skipped_list = parse_input_package_list(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/_work/TheRock/TheRock/./build_tools/packaging/linux/build_package.py", line 718, in parse_input_package_list
    pkg_list, skipped_list = get_package_list(artifact_dir)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/_work/TheRock/TheRock/build_tools/packaging/linux/packaging_utils.py", line 252, in get_package_list
    data = read_package_json_file()
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/_work/TheRock/TheRock/build_tools/packaging/linux/packaging_utils.py", line 63, in read_package_json_file
Removed directory: /home/runner/_work/TheRock/TheRock/build_tools/packaging/linux/__pycache__
Directory does not exist: /home/runner/_work/TheRock/TheRock/output/packages/deb
In function: parse_input_package_list
    data = json.load(file)
           ^^^^^^^^^^^^^^^
  File "/home/runner/_work/_tool/Python/3.12.13/x64/lib/python3.12/json/__init__.py", line 293, in load
    return loads(fp.read(),
           ^^^^^^^^^^^^^^^^
  File "/home/runner/_work/_tool/Python/3.12.13/x64/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3885, opened 2026-03-10
Add JSON syntax validation to pre-commit hooks

raramakr:
Extends the pre-commit validation suite with the check-json hook from
  pre-commit/pre-commit-hooks. This addition ensures that all JSON files
  in the repository are syntactically valid before commits are accepted.


raramakr:
> Would this have caught #3873 (with this fix: #3872)?

Yes Scott

PR #3894, opened 2026-03-10
Add per-family PyTorch wheel builds to multi_arch_ci_linux

marbre:
Wire up PyTorch CI builds for the multi-arch pipeline, mirroring what
ci_linux.yml already does for single-arch variants.

Each GPU family gets its own wheel built against that family's ROCm
package index slice. The build_pytorch gate follows the same pattern
as ci_linux.yml: a variant-level boolean input threads through
multi_arch_ci.yml into multi_arch_ci_linux.yml and gates the job via
inputs.build_pytorch. All current multi-arch families support PyTorch,
so no per-family filtering is needed in the workflow.

Multi-arch CI run: https://github.com/ROCm/TheRock/actions/runs/22931640914

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -32,6 +32,9 @@ on:
         type: string
       test_type:
         type: string
+      build_pytorch:
+        type: boolean
+        default: false
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -141,6 +144,43 @@ jobs:
       contents: read
       id-token: write
 
+  # NOTE: Per-family PyTorch builds — each family gets its own wheel built
+  # against that family's ROCm package index slice.
+  # TODO(future): Consider adding a combined multi-arch pytorch build job
+  # that builds once against the full multi-arch index (amdgpu_families=all).
+  build_pytorch_wheels_per_family:
```
ScottTodd, line 151: We should also be able to add ROCM python + pytorch builds to the Windows multi-arch CI. Anything blocking that, or just starting workflow changes here?

marbre: Just starting here :) Windows can be done next.

marbre:
Test failures are unrelated to the changes here and only failed on some platforms. So either flaky tests or infra but unrelated to the patch.

PR #3896, opened 2026-03-10
Add extra_cmake_options to workflow_call inputs for Windows release workflow

araravik-psd:
The wrapper workflow passes extra_cmake_options when calling the reusable Windows release workflow, but this input is only defined under workflow_dispatch and not workflow_call. This causes the workflow validation to fail.

Failed run: [https://github.com/[ROCm/rockrel](https://github.com/ROCm/rockrel/actions/runs/22926931213)/actions/runs/22926931213](https://github.com/%5BROCm/rockrel%5D(https://github.com/ROCm/rockrel/actions/runs/22926931213)/actions/runs/22926931213)

This change adds extra_cmake_options to workflow_call.inputs so the reusable workflow accepts the same parameter when triggered by another workflow.

### DIFF
--- a/.github/workflows/release_windows_packages.yml
+++ b/.github/workflows/release_windows_packages.yml
@@ -27,6 +27,9 @@ on:
       prerelease_version:
         description: "(Optional) Number of the prerelease"
         type: string
+      extra_cmake_options:
+        description: "Extra options to pass to the CMake configure command"
+        type: string
       repository:
         description: "Repository to checkout. Otherwise, defaults to `github.repository`."
         type: string


araravik-psd:
@ScottTodd Please take a quick look when you have a moment, so that we can merge this and get the windows release started with cherry pick. Thanks

ScottTodd:
You don't need my approval for changes like this.

To catch this in the future, I think we can add a unit test checking that it workflow_dispatch and workflow_call are used, they have the same inputs

araravik-psd:
> You don't need my approval for changes like this.
> 
> To catch this in the future, I think we can add a unit test checking that it workflow_dispatch and workflow_call are used, they have the same inputs

Sure thanks @ScottTodd  will work on the unit test check shortly

PR #3897, opened 2026-03-11
Fix rocprofiler dependency on MPI

marbre:
rocprofiler-sdk has an optional dependency on MPI which is guarded via `-DROCPROFSYS_USE_MPI=${THEROCK_ENABLE_MPI}`. rocprofiler-compute unconditionally calls `find_package(MPI)` and will fail if the bundled packages is enabled (`THEROCK_ENABLE_MPI` set to `ON`).

This patch adds the dependency to the build topology and adds the missing runtime dep to rocprofiler-compute.


araravik-psd (approved):
LGTM, thanks for the fix

PR #3899, opened 2026-03-11
Enable JAX wheel release workflow for prerelease builds

araravik-psd:
This change removes the RELEASE_TYPE != 'prerelease' guard from the JAX wheel release steps so they also run for prerelease builds.

Previously, the steps responsible for URL-encoding the tarball URL and triggering the release_portable_linux_jax_wheels.yml workflow were skipped when RELEASE_TYPE=prerelease. With this change, the workflow will now trigger JAX wheel builds for prerelease releases as well.

This enables prerelease builds to produce JAX wheels consistently alongside other release artifacts.


amd-aakash (approved):
Nit you can remove the comment next instance

araravik-psd:
> Nit you can remove the comment next instance

The comment was removed in the second commit

PR #3785, opened 2026-03-05
Add rocm-bandwidth-test as a core build subproject

lajagapp:
## Summary

Adding ROCm Bandwidth Test (RBT) as a standalone submodule under core/, integrated into TheRock's build system using therock_cmake_subproject_declare, following the same pattern as rocgdb.

## Motivation
Review feedback on the RBT test PR #3539, builds should happen on CPU runners during the build step, not on GPU runners during testing.

### New submodule
- **`core/rocm-bandwidth-test`**: Added `ROCm/rocm_bandwidth_test` as a git submodule (branch: ` amd-mainline`)

### Build topology (`BUILD_TOPOLOGY.toml`)
- Added `source_sets.rocm-bandwidth-test` (Linux-only)
- Added `artifact_groups.rocm-bandwidth-test` (generic type, depends on hip-runtime)
- Added `artifacts.rocm-bandwidth-test` (target-neutral, feature group: CORE)
- Added to `build_stages.compiler-runtime`

### CMake integration (`core/CMakeLists.txt`)
- Declared `rocm-bandwidth-test` as a CMake subproject using `therock_cmake_subproject_declare` with:
  - `ROCR-Runtime` as runtime dependency (provides `hsa-runtime64`)
  - `INTERFACE_LINK_DIRS` / `INTERFACE_INSTALL_RPATH_DIRS` for shared libraries (`libamd_tb.so`, `libamd_work_bench.so`)
- Linux-only (`if(NOT WIN32)`)

### Artifact descriptor (`core/artifact-core-rbt.toml`)
- Components: `dbg`, `doc`, `lib`, `run`
- Covers: binaries (`bin/rocm_bandwidth_test`), shared libraries (`lib/libamd_tb.so`, `lib/libamd_work_bench.so`), plugins  (`lib/rocm_bandwidth_test/plugins/`), and documentation

### Testing
- Nightly-CI build logs - https://therock-ci-artifacts.s3.amazonaws.com/22727144584-linux/logs/gfx94X-dcgpu/index.html
- Workflow - https://github.com/ROCm/TheRock/actions/runs/22727144584

PR #3906, opened 2026-03-11
Add libcurl to the manylinux build dockerfile for rocm-bandwidth-test

lajagapp:
## Motivation

We are onboarding rocm-bandwidth-test (RBT) build as part of #3785, it fails during CMake configure in CI with:

```
CMake Error: Could NOT find CURL (missing: CURL_LIBRARY CURL_INCLUDE_DIR)
Call Stack:
cmake/build_utils.cmake:963 (find_package)
CMakeLists.txt:200 (add_bundled_libraries)
```
RBT's work_bench library has a hard dependency on libcurl. Since its not present in base image, cmake fails with above error.

## Technical Details

CURL is required by the following files in the RBT repo:

https://github.com/ROCm/rocm_bandwidth_test/blob/amd-mainline/cmake/build_utils.cmake#L963
-  cmake/build_utils.cmake:963 -- `find_package(CURL REQUIRED)`
- deps/work_bench/CMakeLists.txt:140 -- `target_link_libraries(... CURL::libcurl)`
- deps/work_bench/include/awb/http_ops.hpp:41 -- `#include <curl/curl.h>`
- deps/work_bench/src/http_ops.cpp:39 -- `#include <curl/curl.h>`
- cmake/sdk/CMakeLists.txt:80 -- `find_package(CURL REQUIRED)`

**Changes:**
- Install `libcurl-devel` in manylinux dockerfiles (`build_manylinux_x86_64.Dockerfile`)

## Test Plan

- Ensure Dockerfile is built and contains necessary packages
- Test the docker image in CI workflow to ensure RBT builds as expected

## Test Result

- Full CI validation pending for the docker image rebuild

### DIFF
--- a/dockerfiles/build_manylinux_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_x86_64.Dockerfile
@@ -82,6 +82,7 @@ RUN yum install -y epel-release && \
     && yum install -y \
       texinfo \
       flex \
+      libcurl-devel \
     && yum clean all && \
     rm -rf /var/cache/yum
 


dockerfiles/build_manylinux_x86_64.Dockerfile
```diff
@@ -82,6 +82,7 @@ RUN yum install -y epel-release && \
     && yum install -y \
       texinfo \
       flex \
+      libcurl-devel \
```
marbre, line 85: `devel` packages are normally not to be added to the build image. Any devel package will require approval by @stellaraccident.

lajagapp: Thanks for the details. I reviewed the policy in docs/design/manylinux_builds.md and agree that -devel packages shouldn't be added to build image.
RBT has an unconditional dependency with libcurl headers & libs as mentioned in PR description. So, it can't be skipped without upstream changes.
Would it be okay to vendor libcurl as a bundled sysdep under third-party/sysdeps/linux/ (same approach as zlib, elfutils, etc.)?

marbre: That could be an alternative, depending on what rocm-bandwidth-test actually is (intended to ship? internal tool?). However, on-hold if there hasn't been a decision from product yet as Geo mentioned.

PR #3915, opened 2026-03-11
Update test_filtering.md

dileepr1:
## Motivation

Make updates to the document based on the current implementation of Test Filter Standardization.

## Technical Details

NA

## Test Plan

NA

## Test Result

NA

## Submission Checklist

### DIFF
--- a/docs/development/test_filtering.md
+++ b/docs/development/test_filtering.md
@@ -44,6 +44,7 @@ To do this the implementation uses a test_categories.yaml file which provides th
 TheRock CI uses the environment variables `TEST_TYPE` to specify the test category and `AMDGPU_FAMILIES` for gpu.
 
 A sample ctest command for a `quick` test run on `gfx110X` will look like
+
 ```
 ctest -L quick -L ex_gpu_gfx110X
 ```
@@ -52,7 +53,6 @@ More information on implementation and integration is available in the below lin
 
 https://github.com/ROCm/rocm-libraries/blob/develop/shared/ctest/README.md
 https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/test_executable_scripts/README.md
-
 
 ## Additional information
 


geomin12 (changes requested):
actually, I am fine with keeping smoke / standard / nightly / full

reasons:
- a majority of the component teams use smoke filters already
- smoke filters provide a clearer message of "sanity check"
- nightly filters run once a day at night and `nightly` already provide a clearer idea of when it will be run (while trying to guess with comprehensive)

docs/development/test_filtering.md
```diff
@@ -36,26 +36,24 @@
...
 ```
-SMOKE_TESTS = [
-  "*smoke_tests*",
-  "*basic_tests*"
-]
-environ_vars = os.environ.copy()
-test_type = os.getenv("TEST_TYPE", "full")
-if test_type == "smoke":
-    environ_vars["GTEST_FILTER"] = ":".join(SMOKE_TESTS)
+ctest -L quick -L ex_gpu_gfx110X
```
geomin12, line 49: ? i don't know what this is, can you add full flag names?

remember, these updates need to apply to **all components**. if `ex_gpu_gfx110X` is applied to all components, this is fine to keep but i was looking for a code example here

docs/development/test_filtering.md
```diff
@@ -36,26 +36,24 @@
...
 
-For gtest executables, using `gtest_filter` is sufficient
+Test filter implementation is done with CTest.
+Whatever be the underlying test framework - say gtest, pytest etc - a ctest wrapper will be created over it exposing the capability to run each test category using ctest labels.
 
-```
-./gtest-executable --gtest_filter=*smoke*
-./gtest-executable --gtest_filter=*nightly*
-```
+To do this the implementation uses a test_categories.yaml file which provides the template to add/exclude the tests to be run for each category, which has to be updated by the component teams. We can add/exclude tests based on the gpu model and OS where the tests are run.
```
geomin12, line 42: code example? some components don't know what test_categories.yaml looks like

docs/development/test_filtering.md
dileepr1: If any component have any tests excluded for a gpu, the parser will add a ctest entry avoiding those tests and assign the label ex_gpu_gfx<number>.
The component teams doesn't need to know about this. They just fill in test_categories.yaml and its parser will take care of the rest.

TheRock side will check if there is any ctest labels in the component for the current gpu platform and applies the labels to the test command accordingly automatically.

docs/development/test_filtering.md
dileepr1: TheRock side run will be handled by test_runner.py and the files like test_miopen.py test_hipblast.py etc will no longer be needed. 
Component teams will not need to touch TheRock CI, they just need to add their tests in test_categories.yaml in the component side.

jayhawk-commits:
> actually, I am fine with keeping smoke / standard / nightly / full
> 
> reasons:
> 
> * a majority of the component teams use smoke filters already
> * smoke filters provide a clearer message of "sanity check"
> * nightly filters run once a day at night and `nightly` already provide a clearer idea of when it will be run (while trying to guess with comprehensive)

Remember the focus should not be what terms being used, but how we define their meaning. If teams are already using "smoke", then it MUST align with the definition in this document. I doubt this is the case for most teams.

There is guess work with both nightly and comprehensive. How long will nightly tests take? The terms are not what is important. If a term leads to incorrect assumptions that don't match what we want them to mean, then it is probably a bad candidate term.

PR #3126, opened 2026-01-28
Deb and rpm install sanity check

jonatluu:
## Motivation

This pull request enhances the native Linux package build and test workflows by introducing automated install verification for built packages. It adds new workflow inputs, implements conditional install testing, and creates a dedicated workflow to validate package installation across multiple OS profiles. These changes improve the reliability and traceability of package builds, especially for prerelease and nightly releases.

## Summary

- `test_native_linux_packages_install.yml` (new file):
  - Full repo-based install and verification workflow for native packages
  - Supports DEB (Ubuntu 24.04) and RPM (RHEL 10, SLES 16) via OS profile
  - Configurable test types: `sanity` (repo install + basic verification) and
    `full` (sanity + RDHC verification)
  - Sets up package manager repo from `package_install_url`, installs ROCm
    packages, verifies key components and installed package list
  - Uses a shared Python venv across steps for consistent dependency management
  - Supports optional GPG key verification for prerelease builds
  - Install test invocation using pytest

## Test plan
- [ ] Trigger `build_native_linux_packages.yml` for a dev/nightly build and
      verify `package_install_url` is correctly emitted to job outputs
- [ ] Trigger `test_native_linux_packages_install.yml` with DEB repo URL on
      ubuntu2404 and confirm `amdrocm` and `amdrocm-core-sdk` install cleanly
- [ ] Trigger with RPM repo URL on rhel10 and sles16 and confirm install

## Test Result
No Verify nightly: 
https://github.com/ROCm/TheRock/actions/runs/23602292215
https://github.com/ROCm/TheRock/actions/runs/23610612186

Verify nightly:
https://github.com/ROCm/TheRock/actions/runs/23785348518 (rpm)
https://github.com/ROCm/TheRock/actions/runs/23778928188 (deb)

Full Test workflow:
https://github.com/ROCm/TheRock/actions/runs/24353145372
https://github.com/ROCm/TheRock/actions/runs/24571097705, https://github.com/ROCm/TheRock/actions/runs/24569164778

PR #3922, opened 2026-03-11
Add scripts for get repo, url params

arvindcheru:
## Motivation
This pull request adds a new utility script, `get_url_repo_params.py`, to the `build_tools/packaging/linux` directory. The script provides command-line tools to extract or construct repository-related URLs and parameters, standardizing and simplifying logic that was previously handled inline in workflows.

## Technical Details
**New utility script for repo URL parameter management: **

* Added `get_url_repo_params.py`, a Python script that provides three subcommands to:
  - Extract the base URL from any input URL (`get-base-url`)
  - Derive a `repo_sub_folder` from an S3 prefix (`get-repo-sub-folder`)
  - Construct a full repository URL from multiple components (`get-repo-url`)
  The script outputs results in `KEY=value` format for easy integration with CI/CD workflows.

* Sample Usage references
1. https://github.com/ROCm/TheRock/pull/3126/changes#diff-2cd881a0753ee30a727e1e80e4b6937d06ff76f102131f8a305422f6e400ebd0R111
2. https://github.com/ROCm/TheRock/pull/3126/changes#diff-d9a4fa08d6e7822fa8a91d37178134468a43b165caf949992543877a122a7e9cR223

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/get_url_repo_params.py
+++ b/build_tools/packaging/linux/get_url_repo_params.py
@@ -5,16 +5,18 @@
 """
 Get URL/repo parameters: base URL from any URL, repo_sub_folder from an S3 prefix, or full repo URL from components.
 
+Output is always KEY=value (suitable for GITHUB_OUTPUT).
+
 Subcommands (get operations):
 
-  get-base-url         Get base URL (scheme + netloc) from an input URL. With --output-format kvp prints KEY=value for GITHUB_OUTPUT.
-  get-repo-sub-folder  Get repo_sub_folder from an S3 prefix (last segment if YYYYMMDD-<id>, else empty).
-  get-repo-url        Get full repo URL from release_type, native_package_type, repo_base_url, os_profile, repo_sub_folder.
+  get-base-url         Get base URL (scheme + netloc) from an input URL. Prints repo_base_url=<value>.
+  get-repo-sub-folder  Get repo_sub_folder from an S3 prefix (last segment if YYYYMMDD-<id>, else empty). Prints repo_sub_folder=<value>.
+  get-repo-url         Get full repo URL from components(release_type, native_package_type, repo_base_url, os_profile, repo_sub_folder). Prints repo_url=<value>.
 
-Usage (--output-format must appear before the subcommand; default is value):
-  python build_tools/packaging/linux/get_url_repo_params.py [--output-format value|kvp] get-base-url --from-url <url>
-  python build_tools/packaging/linux/get_url_repo_params.py [--output-format value|kvp] get-repo-sub-folder --from-s3-prefix <prefix>
-  python build_tools/packaging/linux/get_url_repo_params.py [--output-format value|kvp] get-repo-url ...
+Usage:
+  python build_tools/packaging/linux/get_url_repo_params.py get-base-url --from-url <url>
+  python build_tools/packaging/linux/get_url_repo_params.py get-repo-sub-folder --from-s3-prefix <prefix>
+  python build_tools/packaging/linux/get_url_repo_params.py get-repo-url ...
 
 Examples:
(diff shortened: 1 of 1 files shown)


build_tools/packaging/linux/get_url_repo_params.py
```diff
@@ -0,0 +1,125 @@
...
+"""
+Get URL/repo parameters: base URL from a CloudFront URL, or repo_sub_folder from an S3 prefix.
+
+Subcommands (get operations):
+
+  get-base-url         Get base URL (scheme + netloc) from a CloudFront URL. With --format env outputs repo_base_url=<value>.
+  get-repo-sub-folder  Get repo_sub_folder from an S3 prefix (last segment if YYYYMMDD-<id>, else empty).
+
+Usage:
+  python build_tools/packaging/linux/get_url_repo_params.py get-base-url --from-cloudfront-url <url> [--format env|value]
```
nunnikri: why so many parameters for a simple operation? 
the parameter name --from-cloudfront-url, will it work only for cloudfront url? 

Its better to have the usage as this. 
python build_tools/packaging/linux/get_url_repo_params.py --get-base-url
python build_tools/packaging/linux/get_url_repo_params.py --get-runid-from-prefix
python build_tools/packaging/linux/get_url_repo_params.py --get-date-from-prefix

build_tools/packaging/linux/get_url_repo_params.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+# --- main ---
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description="Get URL/repo parameters: base URL (from CloudFront URL) or repo_sub_folder (from S3 prefix).",
+    )
+    parser.add_argument(
+        "--format",
```
nunnikri: format doesnt look apt for this parameter. Need to find a better name here.
Also any get operation users will expect the value. So value should be the default case

build_tools/packaging/linux/get_url_repo_params.py
```diff
@@ -0,0 +1,125 @@
...
+        print(base_url)
+    return 0
+
+
+# --- repo_sub_folder ---
+
+DATE_ARTIFACT_PATTERN = re.compile(r"^\d{8}-\d+$")
+
+
+def get_repo_sub_folder(s3_prefix: str) -> str:
```
nunnikri, line 59: It returns the full date-id. Is that required. If this full string is not needed, as suggested above getting date and run id separately is better.

build_tools/packaging/linux/get_url_repo_params.py
arvindcheru: Thanks updated the name of option, as it works for all url.

build_tools/packaging/linux/get_url_repo_params.py
arvindcheru: Thanks for the comment. --format option renamed to --output, and default case updated to value. Updated in the latest version.

build_tools/packaging/linux/get_url_repo_params.py
arvindcheru: We need to use the (date-id} for our full url operations. So we currently not using these values separately hence continuing to use (date-id) for now.

build_tools/packaging/linux/get_url_repo_params.py
nunnikri: Now i understood the usage of this option. I think better to rename as --output-format and the possible parameters as env/shell and github. And update the comments accordingly/

arvindcheru: Thank you. Update name to --output-format and choices rename (value, kvp)

PR #3928, opened 2026-03-12
Enable flang and openmp use throughout ROCm.

stellaraccident:
## Summary

Splits the ROCm LLVM build into a small bootstrap compiler plus optional Fortran and OpenMP/offload layers, then adjusts TheRock's build topology so downstream projects can consume the right compiler view without forcing every build to pay for Flang/offload.

The main goals are:

- Build a base LLVM compiler that contains Clang/LLD/backends/device-libs, but not Flang, MLIR, OpenMP/offload, or flang-rt.
- Keep Fortran and offload available for full Linux builds, while allowing minimal HIP/runtime builds to disable Fortran/offload completely.
- Make downstream Fortran use explicit via TheRock subproject metadata and the `ROCM_BUILD_FORTRAN_LIBS` flag.
- Avoid advertising LLVM's installed OpenMP CMake config as the global provider for `find_package(OpenMP)`.
- Collapse the old `foundation` stage into `compiler-runtime` so CI does not add a lightweight root stage or require a second LLVM checkout.
- Move `hip-tests` and `rocrtst` to a separate `runtime-tests` stage so runtime tests do not rebuild on downstream stages that only need HIP/core runtime artifacts.

## LLVM Layering

The LLVM-related artifacts are now layered as follows:

- `amd-llvm-base`: bootstrap compiler layer. Builds the LLVM project with `clang`, `lld`, `clang-tools-extra`, X86/AMDGPU targets, compiler-rt/device-libs, and no Flang/MLIR/OpenMP/offload. This is the base compiler used by later layers.
- `amd-llvm`: compiler infrastructure layer. Builds `amd-comgr` and `hipcc` on top of `amd-llvm-base`.
- `amd-llvm-flang`: optional Linux-only Flang frontend layer. Builds Flang/MLIR as implementation details needed by the frontend, but does not install MLIR into the final ROCm dist.
- `amd-llvm-offload`: optional Linux-only runtime layer. Builds OpenMP/offload runtimes and, when Flang is enabled, flang runtime pieces. Depends on `amd-llvm`, `amd-llvm-flang` when enabled, and `core-runtime`.

`hip-clr` uses the `amd-llvm-offload` compiler view when that target exists. That keeps HIP, Flang, OpenMP headers/libraries, and offload runtime visibility together in the `amd-hip` toolchain view without making the base compiler advertise OpenMP globally.

## Build Topology

The old `foundation` stage has been removed. Its former artifact groups are now part of `compiler-runtime`:

- `third-party-sysdeps`
- `base`
- `compiler`
- core/runtime/profiler groups

This avoids a root stage that mostly adds critical path and avoids splitting LLVM-related source checkout/build behavior across stages.

A new `runtime-tests` stage contains:

- `core-hiptests`
- `rocrtst`

`runtime-tests` depends on the runtime/compiler artifacts it needs and runs parallel to math-libs. Linux builds both `hip-tests` and `rocrtst`; Windows has the same stage so `core-hiptests` artifacts are still produced, while `rocrtst` remains disabled by platform metadata.

## Fortran Controls

This adds or wires:

- `THEROCK_ENABLE_FLANG`
- `THEROCK_ENABLE_AMD_LLVM_OFFLOAD`
- `THEROCK_FLAG_BUILD_FORTRAN_LIBS`
- `FORTRAN_OPTIONAL` / `FORTRAN_REQUIRED` subproject declaration keywords
- `ROCM_BUILD_FORTRAN_LIBS=ON/OFF` propagation to downstream subprojects

Known Fortran consumers can now declare whether Fortran is optional or required. When Fortran libs are disabled, optional users can configure without forcing Flang/offload into the build.

## OpenMP Handling

- HIP-facing projects get OpenMP visibility through the `amd-hip` toolchain view when `hip-clr` is built on top of `amd-llvm-offload`.
- `rocprofiler-systems` gets a targeted pre-hook that passes `libomptarget` explicitly and seeds module-mode `FindOpenMP` cache entries where needed.
- GCC-configured subprojects use GCC's own `libgomp` path, avoiding accidental selection of the offload package's `libgomp.so` shim.

## Downstream Wiring

The change updates downstream projects that need Fortran/OpenMP awareness, including RAND/BLAS-family projects and `rocprofiler-systems`.

Notable fixes:

- Math library Fortran wrappers are gated by `ROCM_BUILD_FORTRAN_LIBS`.
- `rocprofiler-systems` receives the offload library directory explicitly for `libomptarget`.
- `rocprofiler-systems` examples are disabled in TheRock integration builds to avoid using deep example builds as the first test of the Fortran/offload toolchain.

## Verification

Local verification was done on a `gfx1100` machine.

Static checks:

```bash
python build_tools/topology_to_cmake.py --validate-only
python build_tools/configure_stage.py --stage compiler-runtime --print
python build_tools/configure_stage.py --stage runtime-tests --print
python build_tools/configure_stage.py --stage runtime-tests --platform windows --print
python -m unittest discover -s build_tools/github_actions/tests -p configure_multi_arch_ci_test.py
python -m unittest build_tools.tests.build_topology_test build_tools.github_actions.tests.post_stage_upload_test
git diff --check
```

No-Fortran HIP/RAND build:

- Configured with `THEROCK_ENABLE_ALL=OFF`, `THEROCK_ENABLE_HIP_RUNTIME=ON`, `THEROCK_ENABLE_RAND=ON`, `THEROCK_ENABLE_FLANG=OFF`, `THEROCK_ENABLE_AMD_LLVM_OFFLOAD=OFF`, and `THEROCK_FLAG_BUILD_FORTRAN_LIBS=OFF`.
- Built `dist-rocm`.
- Verified the dist contains HIP/RAND pieces and does not contain Flang/OpenMP/offload outputs such as `amdflang`, `flang`, `libomp.so`, `libomptarget.so`, `omp_lib.mod`, or `libflang_rt*`.
- Built and ran a tiny HIP smoke test with `hipcc --offload-arch=gfx1100`.

Fortran/offload toolchain build:

- Configured with compiler/core runtime/HIP/RAND/BLAS enabled plus `THEROCK_ENABLE_FLANG=ON`, `THEROCK_ENABLE_AMD_LLVM_OFFLOAD=ON`, and `THEROCK_FLAG_BUILD_FORTRAN_LIBS=ON`.
- Built and verified `amd-llvm-base`, `amd-llvm`, `amd-llvm-flang`, `amd-llvm-offload`, HIP runtime, RAND, and BLAS-family artifacts.
- Verified `dist/rocm` contains `amdflang`, `flang`, `omp_lib.mod`, flang runtime libraries, `libomp.so`, `libomptarget.so`, AMDGCN device runtime pieces, BLAS/RAND libraries, and no installed MLIR CMake package in the final ROCm dist.
- Passed small Fortran, Fortran OpenMP, C OpenMP offload, Fortran OpenMP/offload, and HIP smoke tests.

BLAS/RAND integration:

- Expunged cached downstreams after toolchain changes.
- Built/refreshed `mxDataGenerator`, `hipBLASLt`, `rocBLAS`, `hipBLAS`, `rocRAND`, and `hipRAND`.
- Confirmed Fortran wrappers are enabled in the Fortran build and absent in the no-Fortran build.


marbre:
Builds are failing:
```
[rocprofiler-systems configure] -- Checking out 'https://github.com/OpenMP-Validation-and-Verification/OpenMP_VV.git' @ 'master'...
[rocprofiler-systems configure] Cloning into 'external/ompvv'...
[rocprofiler-systems configure] CMake Error at cmake/MacroUtilities.cmake:20 (message):
[rocprofiler-systems configure]   [rocprofiler-systems] Could not find amdflang.  This is required for the
[rocprofiler-systems configure]   OMPVV tests.
[rocprofiler-systems configure] Call Stack (most recent call first):
[rocprofiler-systems configure]   examples/openmp/external/CMakeLists.txt:184 (rocprofiler_systems_message)
[rocprofiler-systems configure] 
[rocprofiler-systems configure] 
[rocprofiler-systems configure] -- Configuring incomplete, errors occurred!
```

compiler/artifact-core-runtime.toml
```diff
@@ -0,0 +1,11 @@
...
+[components.dbg."compiler/ROCR-Runtime/stage"]
+[components.dev."compiler/ROCR-Runtime/stage"]
+[components.doc."compiler/ROCR-Runtime/stage"]
+[components.lib."compiler/ROCR-Runtime/stage"]
+[components.run."compiler/ROCR-Runtime/stage"]
+
+# rocminfo
+[components.dbg."compiler/rocminfo/stage"]
+[components.doc."compiler/rocminfo/stage"]
+[components.run."compiler/rocminfo/stage"]
```
ScottTodd: Let's reorder these, following https://github.com/ROCm/TheRock/pull/3890 to the `lib -> run -> dbg -> dev -> doc -> test` ordering that artifacts are built in. See also the new docs I added at https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md#component-extends-chain

(or we can revert that PR if you prefer the alphabetical ordering, despite it not matching the usage in the tooling)

```suggestion
# ROCR-Runtime
[components.lib."compiler/ROCR-Runtime/stage"]
[components.run."compiler/ROCR-Runtime/stage"]
[components.dbg."compiler/ROCR-Runtime/stage"]
[components.dev."compiler/ROCR-Runtime/stage"]
[components.doc."compiler/ROCR-Runtime/stage"]

# rocminfo
[components.run."compiler/rocminfo/stage"]
[components.dbg."compiler/rocminfo/stage"]
[components.doc."compiler/rocminfo/stage"]
```

base/artifact-amd-llvm-base.toml
ScottTodd, line 1: Same ordering comment in here

https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md#component-extends-chain

`lib → run → dbg → dev → doc → test`

BUILD_TOPOLOGY.toml
```diff
@@ -451,13 +451,33 @@ feature_group = "CORE"  # Part of core, enabled by default
...
+artifact_deps = ["amd-llvm"]
+feature_name = "FLANG"
+feature_group = "ALL"
+disable_platforms = ["windows"]
+
+[artifacts.amd-llvm-offload]
+artifact_group = "compiler"
+type = "target-neutral"
+artifact_deps = ["amd-llvm", "core-runtime"]
+disable_platforms = ["windows"]
```
ScottTodd: Should this have `feature_*` parameters like the other `amd-llvm-*` artifacts above?

cmake/therock_subproject.cmake
```diff
@@ -484,6 +484,31 @@ function(therock_cmake_subproject_declare target_name)
...
+    elseif(WIN32)
+      # Windows: let CMake find system gfortran (no toolchain dep).
+    else()
+      # Linux, no flang built.
+      if(ARG_FORTRAN_REQUIRED)
+        message(FATAL_ERROR
+          "Sub-project ${target_name} requires Fortran (FORTRAN_REQUIRED) "
+          "but THEROCK_ENABLE_FLANG is OFF and no system Fortran compiler is available on Linux.")
+      endif()
+    endif()
```
ScottTodd: Is this missing a case for "use system Fortran compiler" on Linux?

The "and no system Fortran compiler is available on Linux" message makes it seem like a check ran already and failed (Windows checks later, Linux just hits this error?)

ScottTodd (commented):
Overall looks good once the CI passes

compiler/artifact-core-runtime.toml
stellaraccident: I do have a strong affinity to alphabetic vs bespoke ordering. Going to leave this as-is for now.

base/artifact-amd-llvm-base.toml
stellaraccident: Prefer alpha.

compiler/artifact-core-runtime.toml
stellaraccident: But please don't revert the ordering PR now. I'm growing tired of resolving the conflicts induced by that.

stellaraccident:
Try 2.

marbre:
> Try 2.

```
[rocRoller configure] CMake Error at /__w/TheRock/TheRock/rocm-libraries/shared/mxdatagenerator/CMakeLists.txt:14 (target_link_libraries):
[rocRoller configure]   The link interface of target "mxDataGenerator" contains:
[rocRoller configure] 
[rocRoller configure]     OpenMP::OpenMP_CXX
[rocRoller configure] 
[rocRoller configure]   but the target was not found.  Possible reasons include:
[rocRoller configure] 
[rocRoller configure]     * There is a typo in the target name.
[rocRoller configure]     * A find_package call is missing for an IMPORTED target.
[rocRoller configure]     * An ALIAS target is missing.
[rocRoller configure] 
[rocRoller configure] 
[rocRoller configure] 
[rocRoller configure] CMake Error at test/common/CMakeLists.txt:7 (target_link_libraries):
[rocRoller configure]   Target "common-test-utilities" links to:
[rocRoller configure] 
[rocRoller configure]     OpenMP::OpenMP_CXX
[rocRoller configure] 
[rocRoller configure]   but the target was not found.  Possible reasons include:
[rocRoller configure] 
[rocRoller configure]     * There is a typo in the target name.
[rocRoller configure]     * A find_package call is missing for an IMPORTED target.
[rocRoller configure]     * An ALIAS target is missing.
```

:cry:

ronlieb:
what are the valid combinations of the 3 amd-llvm* 

can amd-llvm + amd-llvm-offload be built ?  

same for amd-llvm + amd-llvm-flang ?

if i specify amd-llvm-flang does it imply amd-llvm ?  or do i need to explicitly as for amd-llvm?

stellaraccident:
> what are the valid combinations of the 3 amd-llvm*
> 
> can amd-llvm + amd-llvm-offload be built ?
> 
> same for amd-llvm + amd-llvm-flang ?
> 
> if i specify amd-llvm-flang does it imply amd-llvm ? or do i need to explicitly as for amd-llvm?

My agent organized my answer a bit:

The three sub-projects form a dependency chain:                                                                                                                                                                    
                                                                                                                                                                                                                   
- amd-llvm-base → always built. Minimal compiler: clang, lld, compiler-rt, device-libs, openmp host runtime.
- amd-llvm → depends on amd-llvm-base. Adds comgr, hipify, etc. This is what most of ROCm needs.                                                                                                                   
- amd-llvm-flang → depends on amd-llvm. Adds the flang compiler frontend (+ MLIR). Linux-only, gated by THEROCK_ENABLE_FLANG.                                                                                      
- amd-llvm-offload → depends on amd-llvm, and optionally on amd-llvm-flang. Adds libomptarget, device RTLs, and (if flang is enabled) flang-rt device runtime. Linux-only.                                         
                                                                                                                                                                                                                   
Valid combinations:                                                                                                                                                                                                
                                                                                                                                                                                                                   
- amd-llvm + amd-llvm-offload (no flang): Yes. You get libomptarget + device RTLs but no flang-rt. C/C++ OpenMP offload works, Fortran doesn't.                                                                    
- amd-llvm + amd-llvm-flang (no offload): Technically valid but not useful — you get the flang compiler binary but no Fortran runtime (flang-rt lives in amd-llvm-offload), so you can't link anything.
- All three: The full stack. flang frontend + flang-rt + offload runtimes.                                                                                                                                         
                                                                                                                                                                                                                   
Dependencies are enforced via artifact_deps in the topology, so enabling amd-llvm-flang automatically pulls in amd-llvm — you don't need to request it explicitly.

marbre:
```
[rocprofiler-systems] error: Semantic errors in /__w/TheRock/TheRock/build/profiler/rocprofiler-systems/build/examples/openmp/external/ompvv/tests/5.0/teams/test_team_default_shared.F90
[rocprofiler-systems] /__w/TheRock/TheRock/build/profiler/rocprofiler-systems/build/examples/openmp/external/ompvv/tests/5.0/teams/test_team_default_shared.F90:28:7: error: Cannot parse module file for module 'iso_fortran_env': Source file 'iso_fortran_env.mod' was not found
[rocprofiler-systems]     USE iso_fortran_env
[rocprofiler-systems]         ^^^^^^^^^^^^^^^
[rocprofiler-systems] /__w/TheRock/TheRock/build/profiler/rocprofiler-
```

In the latest run rocprofiler-systems is unhappy about the Fortran related changes.

stellaraccident:
A month later - needs more cleanup before landing but pushing a WIP branch that passes my local build to see what CI says.

compiler/amd-llvm-flang/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+set(MLIR_INCLUDE_TESTS OFF)
+set(FLANG_INCLUDE_TESTS OFF)
+set(LLVM_ENABLE_PROJECTS "")
+
+# OpenMP modules are produced by the standalone offload runtime layer.
+set(LLVM_ENABLE_RUNTIMES "openmp")
+set(FLANG_RUNTIME_F128_MATH_LIB "libquadmath")
+
+include(${LLVM_MONOREPO_DIR}/flang/cmake/modules/FlangCommon.cmake)
+if(FLANG_INCLUDE_QUADMATH_H)
```
stellaraccident, line 62: Mark this with a TODO as this should be included by handled by default in the LLVM flang config.

compiler/amd-llvm-offload/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+set(LIBOMP_COPY_EXPORTS OFF)
+
+add_subdirectory("${LLVM_MONOREPO_DIR}/runtimes" "runtimes")
+
+find_package(Clang REQUIRED CONFIG)
+find_package(AMDDeviceLibs REQUIRED CONFIG)
+
+set(_openmp_device_build_dir "${CMAKE_CURRENT_BINARY_DIR}/openmp-device")
+set(_openmp_device_install_dir "${CMAKE_INSTALL_PREFIX}")
+set(_openmp_device_cmake_args
```
stellaraccident, line 72: I think we want to pass through the C/CXX compiler launcher? And probably the CMAKE_PREFIX_PATH.

compiler/amd-llvm-offload/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
...
+  "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+  "-DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>"
```
stellaraccident, line 75: What's up with this? `<INSTALL_DIR>` isn't a thing, is it?

core/CMakeLists.txt
```diff
@@ -158,6 +158,67 @@ if(THEROCK_ENABLE_CORE_RUNTIME)
...
+    set(_offload_deps amd-llvm ROCR-Runtime)
+    set(_extra_offload_cmake_args)
+    if(THEROCK_ENABLE_FLANG)
+      if(NOT TARGET amd-llvm-flang)
+        message(FATAL_ERROR "THEROCK_ENABLE_AMD_LLVM_OFFLOAD with THEROCK_ENABLE_FLANG requires amd-llvm-flang")
+      endif()
+      list(APPEND _offload_deps amd-llvm-flang)
+      get_target_property(_flang_dist_dir amd-llvm-flang THEROCK_DIST_DIR)
+      list(APPEND _extra_offload_cmake_args
+        -DCMAKE_Fortran_COMPILER_WORKS=YES
```
stellaraccident, line 184: Add a descriptive note please, referencing LLVM docs for standalone builds if applicable.

core/CMakeLists.txt
```diff
@@ -158,6 +158,67 @@ if(THEROCK_ENABLE_CORE_RUNTIME)
...
+    endif()
+
+    therock_cmake_subproject_declare(amd-llvm-offload
+      USE_DIST_AMDGPU_TARGETS
+      EXTERNAL_SOURCE_DIR "${THEROCK_SOURCE_DIR}/compiler/amd-llvm-offload"
+      BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/amd-llvm-offload"
+      BACKGROUND_BUILD
+      CMAKE_ARGS
+        -DTHEROCK_ENABLE_FLANG=${THEROCK_ENABLE_FLANG}
+        -DOFFLOAD_EXTERNAL_PROJECT_UNIFIED_ROCR=ON
```
stellaraccident, line 198: Descriptive note please on what this does and what it should be doing (i.e. we should refactor this to depend on ROCr, not do this reacharound build of it).

core/CMakeLists.txt
```diff
@@ -222,6 +285,10 @@ if(THEROCK_ENABLE_HIP_RUNTIME)
       rocprofiler-register
       ROCR-Runtime
     )
+    if(TARGET amd-llvm-offload)
```
stellaraccident, line 296: Note please on what the distinction is: If we built for offload, then the resulting CLR dist will include HIP and omp/offload capabilities. Otherwise, it will only include hip.

patches/amd-mainline/llvm-project/0013-OpenMP-Include-generated-libomp-headers-when-split.patch
```diff
@@ -0,0 +1,32 @@
...
+diff --git a/openmp/runtime/src/CMakeLists.txt b/openmp/runtime/src/CMakeLists.txt
+index c224a10eb1ef..c6ad5095fd53 100644
+--- a/openmp/runtime/src/CMakeLists.txt
++++ b/openmp/runtime/src/CMakeLists.txt
+@@ -12,7 +12,7 @@ include(ExtendPath)
+ 
+ # The generated headers will be placed in clang's resource if using it to
+ # compile openmp in a runtimes-bootstrapping build.
+-if(LLVM_TREE_AVAILABLE AND LLVM_RUNTIMES_BUILD)
++if(LLVM_TREE_AVAILABLE AND LLVM_RUNTIMES_BUILD AND HAVE_LLVM_LIT)
```
stellaraccident: Are you just using HAVE_LLVM_LIT here for its byproduct of something? This seems like you are using it to detect a form of out of tree build, but HAVE_LLVM_LIT is a very weak indicator of that.

profiler/CMakeLists.txt
```diff
@@ -307,6 +316,7 @@ if(THEROCK_ENABLE_ROCPROFV3)
         -DROCPROFSYS_BUILD_BOOST=ON
         -DROCPROFSYS_BUILD_DYNINST=ON
         -DROCPROFSYS_BUILD_FOR_THEROCK=ON # TODO: Will be deprecated
+        -DROCPROFSYS_BUILD_EXAMPLES=OFF
```
stellaraccident: Extra debugging? We should not be disabling features that were enabled at TOM.

profiler/pre_hook_rocprofiler-systems.cmake
```diff
@@ -0,0 +1,96 @@
...
+      set(OpenMP_C_LIB_NAMES "omp" CACHE STRING "" FORCE)
+      set(OpenMP_CXX_LIB_NAMES "omp" CACHE STRING "" FORCE)
+      set(OpenMP_omp_LIBRARY "${LIBOMP_LIBRARY}" CACHE FILEPATH "" FORCE)
+      set(OpenMP_C_INCLUDE_DIR "${_rocprofsys_openmp_include_dir}" CACHE PATH "" FORCE)
+      set(OpenMP_CXX_INCLUDE_DIR "${_rocprofsys_openmp_include_dir}" CACHE PATH "" FORCE)
+      set(OpenMP_C_SPEC_DATE "201511" CACHE INTERNAL "")
+      set(OpenMP_CXX_SPEC_DATE "201511" CACHE INTERNAL "")
+    endif()
+  else()
+    execute_process(
```
stellaraccident: When can this branch even happen in our builds? Aren't we always building with clang?

patches/amd-mainline/llvm-project/0013-OpenMP-Include-generated-libomp-headers-when-split.patch
```diff
@@ -0,0 +1,53 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheRock Fortran Agent <fortran-agent@local>
+Date: Fri, 1 May 2026 18:35:00 -0700
+Subject: [PATCH] [OpenMP] Generate libomp headers locally when split
+
+When OpenMP is built standalone against an installed LLVM tree,
+LIBOMP_HEADERS_INTDIR can point at the installed LLVM resource include
+directory instead of the local runtime build directory. That makes GCC
```
ronlieb, line 8: shouldnt this build of openmp files always use clang rather than gcc ?

lumachad:
I wonder if we need to wire something in the cmake/therock_sanitizers.cmake to handle passing sanitizer options to flang.

compiler/amd-llvm-flang/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+if(TARGET FortranSupport AND TARGET MLIRBuiltinTypeInterfacesIncGen)
+  add_dependencies(FortranSupport MLIRBuiltinTypeInterfacesIncGen)
+endif()
+if(TARGET FIRSupport AND TARGET MLIRLinalgOpsIncGen)
+  add_dependencies(FIRSupport MLIRLinalgOpsIncGen)
+endif()
+if(TARGET FortranSupport)
+  target_include_directories(FortranSupport PRIVATE
+    ${CMAKE_BINARY_DIR}/flang/lib/Support)
+endif()
```
Meinersbur, line 93: Followup from today's conversation:

I think duplicating the CMakeLists.txt from upstream is highly undesirable. If changes upstream, it is hard to debug what changes and replicated it in this copy of CMakeLists.txt. `FLANG_INCLUDE_QUADMATH_H` is a great example for this: Build will not fail initially, but `REAL(16)` will be silently unavailable and some poor soul has to find out why. I am planning larger changes to this file upstream (https://github.com/llvm/llvm-project/pull/171515) and would rather not have to do the same a second time for TheRock.

Upstream LLVM already has a means to build components separately in standalone builds. That is: LLVM+Flang -> MLIR -> Flang. You mentioned the disadvantage is that the MLIR build includes a lot of other stuff that is actually not needed for Flang, and a common build dir just handles dependency resolution.

This is an argument for MLIR, Flang which are build through LLVM_ENABLE_PROJECTS, but not for LLVM_ENABLE_RUNTIMES (i.e. amd-llvm-offload). A runtimes build can build multiple runtimes via passing them all to LLVM_ENABLE_RUNTIMES, e.g `-DLLVM_ENABLE_RUNTIMES=openmp;offload`.

For MLIR, it would be less work to only build the components actually needed, either by `ninja MLIRLinalgOpsIncGen ...`, creating a new target in our downstream llvm-project (such as 
```cmake
add_custom_target(mlir-flang-deps)
add_dependencies(mlir-flang-deps MLIRLinalgOpsIncGen  ...)
```
) that depend only on needed targets, or straight up commenting-out upstream components that we do not need. Dependencies are already managed in this CMakeLists copy and a missing dependency due to an upstream change is much more obvious than hunting down changes to configuration parameters.

If we have to manually manage out top-level dependencies anyway, why not keep everything in a single build dir and create targets that only build/install components that are needed (replacing `ninja all` and `ninja install`)? From AI:
```cmake
# Install base_lib into the "base" component
install(TARGETS base_lib 
        DESTINATION lib 
        COMPONENT base)

# Create the custom target for 'ninja install-base'
add_custom_target(install-base
    # 1. The command to run the component-specific install
    COMMAND ${CMAKE_COMMAND} --install ${CMAKE_BINARY_DIR} --component base
    
    # 2. The targets that MUST be built before installing
    DEPENDS base_lib
    
    COMMENT "Building and installing the 'base' component..."
)
```
(upstream LLVM already uses `COMPONENT` in install commands)

or reconfigure the same build dir with `LLVM_ENABLE_PROJECTS=${LLVM_ENABLE_PROJECTS};flang` a second build to incrementally build flang on top of the previous build.

PR #3929, opened 2026-03-12
[CI] Run full PyTorch UT suite 

ethanwee1:
This PR adds a nightly workflow that runs the full PyTorch unit test suite across all three test configs (default, distributed, inductor) using the same nightly staging wheels and PyTorch nightly source as the existing nightly release pipeline. We will merge this PR so that we can begin to test, but as for the UT's that fail we will address this task afterwards as it is outside the scope of this PR.

Latest run: https://github.com/ROCm/TheRock/actions/runs/24155228835/job/70492662532

### Excluded test modules

Modules excluded entirely at the `run_test.py` level (`--exclude`) due to hanging:
    "nn/test_convolution",
    "inductor/test_max_autotune",
    "inductor/test_torchinductor_opinfo_properties",
    "inductor/test_compiled_autograd",
    "distributed/_composable/fsdp/test_fully_shard_autograd",
    "distributed/_composable/test_composability/test_2d_composability",
    "distributed/_composable/test_composability/test_pp_composability",
    "distributed/_composable/test_replicate",
    "distributed/tensor/test_view_ops",
    "dynamo/test_dynamic_shapes",
    "functorch/test_control_flow",

### Test skips (pytorch_2.12.py) — 176 total

> **Note:** All skips are under "common" (applied to every GPU family). CI currently only tests on gfx94X (MI300). Once additional GPU families are enabled, GPU-specific failures should be split into their own family key.

Tracking followup issues in https://github.com/ROCm/TheRock/issues/4082#issuecomment-4203436133
**ci_sanity_check** (1)

| Test File | Test Name |
|-----------|-----------|
| test/ci_sanity_check | `test_env_vars_exist` |

**cuda** (8)

| Test File | Test Name |
|-----------|-----------|
| test/cuda | `test_mempool_limited_memory_with_allocator` |
| test/cuda | `test_tensor_delete_after_allocator_delete` |
| test/cuda | `test_deleted_mempool_not_used_on_oom` |
| test/cuda | `test_mempool_ctx_multithread` |
| test/cuda | `test_mempool_expandable` |
| test/cuda | `test_cuda_graph_tensor_item_not_allowed` |
| test/cuda | `test_allocator_memory_fraction_setting` |
| test/cuda | `test_resnet` |

**dataloader** (6)

| Test File | Test Name |
|-----------|-----------|
| test/dataloader | `test_large_sampler_indices` |
| test/dataloader | `test_multiprocessing_contexts` |
| test/dataloader | `test_nested_tensor_multiprocessing_context_forkserver_cuda` |
| test/dataloader | `test_nested_tensor_multiprocessing_context_spawn_cuda` |
| test/dataloader | `test_sparse_tensor_multiprocessing_context_forkserver_cuda` |
| test/dataloader | `test_sparse_tensor_multiprocessing_context_spawn_cuda` |

**distributed** (50)

| Test File | Test Name |
|-----------|-----------|
| test/distributed | `test_replicate_pp_ScheduleClass3_bfloat16` |
| test/distributed | `test_replicate_pp_ScheduleClass3_float32` |
| test/distributed | `test_replicate_pp_ScheduleClass4_bfloat16` |
| test/distributed | `test_replicate_device_id` |
| test/distributed | `test_depthwise_convolution` |
| test/distributed | `test_fsdp_memory_ckpt_ckpt` |
| test/distributed | `test_resume` |
| test/distributed | `test_get_memory_stats` |
| test/distributed | `test_suspend` |
| test/distributed | `test_monitored_barrier_allreduce_hang_wait_all_ranks` |
| test/distributed | `test_monitored_barrier_allreduce_hang` |
| test/distributed | `test_clip_grad_norm_1d` |
| test/distributed | `test_clip_grad_norm_2d` |
| test/distributed | `test_compiled_autograd_fsdp2_backward` |
| test/distributed | `test_all_gather_extensions_train_parity` |
| test/distributed | `test_gradient_scaler` |
| test/distributed | `test_ddp_A_fsdp_B_ddp_C` |
| test/distributed | `test_compute_dtype` |
| test/distributed | `test_cached_state_dict` |
| test/distributed | `test_explicit_prefetching` |
| test/distributed | `test_mixture_of_experts_offload_true_shard_grad_op_cuda` |
| test/distributed | `test_3d_mlp_with_nd_mesh` |
| test/distributed | `test_linalg_ops` |
| test/distributed | `test_post_localSGD_optimizer_step_reload` |
| test/distributed | `test_fully_shard_communication_count` |
| test/distributed | `test_train_parity_single_group_shard_largest_dim` |
| test/distributed | `test_compiler_collectives_automatic_dynamic_tensor` |
| test/distributed | `test_nontensor_activations` |
| test/distributed | `test_dtensor_train_parity` |
| test/distributed | `test_multi_forward_mixed_requires_grad` |
| test/distributed | `test_fully_shard_training_memory` |
| test/distributed | `test_fully_shard_training_overlap` |
| test/distributed | `test_tp_with_fsdp_offloading` |
| test/distributed | `test_3d_with_tp_dp_pp_ScheduleClass0_bfloat16` |
| test/distributed | `test_replicate_fully_shard_init` |
| test/distributed | `test_grad_acc_with_reduce_dtype` |
| test/distributed | `test_multi_forward_module` |
| test/distributed | `test_replicate_pp_grads_ScheduleClass1` |
| test/distributed | `test_replicate_ignore_module` |
| test/distributed | `test_manual_reshard_with_reshard_after_forward_false` |
| test/distributed | `test_train_parity_with_shared_params` |
| test/distributed | `test_train_parity_2d_mlp` |
| test/distributed | `test_replicate_pp_grads_ScheduleClass2` |
| test/distributed | `test_replicate_pp_grads_ScheduleClass3` |
| test/distributed | `test_replicate_pp_grads_ScheduleClass4` |
| test/distributed | `test_replicate_move_args_kwargs_to_device` |
| test/distributed | `test_replicate_multi_module` |
| test/distributed | `test_replicate_single_module` |
| test/distributed | `test_unused_forward_module` |
| test/distributed | `test_view_ops` |

**distributions** (1)

| Test File | Test Name |
|-----------|-----------|
| test/distributions | `test_entropy_monte_carlo` |

**dynamo** (3)

| Test File | Test Name |
|-----------|-----------|
| test/dynamo | `test_hops_compile_backend_aot_eager_inline_asm_elementwise_simple_cuda_float32` |
| test/dynamo | `test_triggers` |
| test/dynamo | `test_logs_out` |

**export** (10)

| Test File | Test Name |
|-----------|-----------|
| test/export | `test_aot_export_inline_asm_elementwise_simple_cuda_float32` |
| test/export | `test_fake_export___getitem___cuda_float32` |
| test/export | `test_fake_export_nn_functional_batch_norm_cuda_float32` |
| test/export | `test_fake_export_nn_functional_batch_norm_without_cudnn_cuda_float32` |
| test/export | `test_fake_export_nn_functional_conv2d_cuda_float32` |
| test/export | `test_fake_export_nn_functional_instance_norm_cuda_float32` |
| test/export | `test_fake_export_nn_functional_multi_margin_loss_cuda_float32` |
| test/export | `test_fake_export_nn_functional_scaled_dot_product_attention_cuda_float32` |
| test/export | `test_fake_export_nonzero_cuda_float32` |
| test/export | `test_preserve_original_behavior_cuda` |

**functorch** (1)

| Test File | Test Name |
|-----------|-----------|
| test/functorch | `test_grad_nn_functional_conv3d_cuda_float32` |

**inductor** (81)

| Test File | Test Name |
|-----------|-----------|
| test/inductor | `test_equivalent_extern_code` |
| test/inductor | `test_activation_functions` |
| test/inductor | `test_require_stride_order_non_owning` |
| test/inductor | `test_index_add_fallback` |
| test/inductor | `test_compiling_create_block_mask_no_recompile_cuda` |
| test/inductor | `test_linear_and_cel_max_autotune` |
| test/inductor | `test_bias_addmm_lookup_table_entry` |
| test/inductor | `test_autotune_gemm_choice_validation_op_addmm_max_autotune_True` |
| test/inductor | `test_lazy_template_fusion_multiple_candidates_use_async_compile_False` |
| test/inductor | `test_deferred_triton_kernels` |
| test/inductor | `test_addmm_fp16` |
| test/inductor | `test_binary_ufunc_numerical_fmod_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_benchmark_choice_in_subproc` |
| test/inductor | `test_flex_attention_with_dynamic_max_autotune_graph_partition_cuda` |
| test/inductor | `test_combo_kernel_dynamic_shapes_grid_changes` |
| test/inductor | `test_combo_kernel_yz_overflow` |
| test/inductor | `test_mm_plus_mm` |
| test/inductor | `test_external_template_prologue_epilogue_fusion` |
| test/inductor | `test_compile_with_exporter` |
| test/inductor | `test_return_aux_deprecation_warnings_cuda_float16` |
| test/inductor | `test_concat_unbacked_shape_tensor` |
| test/inductor | `test_hints_wrapper` |
| test/inductor | `test_while_loop_with_mixed_device_dynamic_False_cpu_with_stack_allocation` |
| test/inductor | `test_while_loop_with_mixed_device_dynamic_True_cpu_with_stack_allocation` |
| test/inductor | `test_while_loop_with_pytree_inputs_cpu_with_stack_allocation` |
| test/inductor | `test_filled_cache_precompile` |
| test/inductor | `test_eager_equivalence_exp_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_rsqrt_backend_inductor_numerics_cuda_float32` |
| test/inductor | `test_eager_equivalence_log_backend_inductor_default_cuda_float16` |
| test/inductor | `test_remote_cache_load_function_device_cuda_float32_dynamic_False_bundle_triton_False_use_static_triton_launcher_False` |
| test/inductor | `test_determinism_log1p_backend_inductor_default_cuda_float32` |
| test/inductor | `test_combo_kernel_per_config_subkernel_block_size` |
| test/inductor | `test_cond_branches_no_arguments_no_closure` |
| test/inductor | `test_flex_attention_logging_cuda` |
| test/inductor | `test_shared_memory_pruning_mm_float32_mat1_transposed_False_mat2_transposed_True_use_tma_False` |
| test/inductor | `test_shared_memory_pruning_mm_float32_mat1_transposed_True_mat2_transposed_False_use_tma_False` |
| test/inductor | `test_shared_memory_pruning_mm_float32_mat1_transposed_True_mat2_transposed_True_use_tma_False` |
| test/inductor | `test_template_epilogue_fusion_static_analysis_test_case_timing_reject_use_async_compile_False` |
| test/inductor | `test_template_epilogue_fusion_static_analysis_test_case_timing_reject_use_async_compile_True` |
| test/inductor | `test_unary_ufunc_numerical_cos_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_reciprocal_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_log10_backend_inductor_default_cuda_float32` |
| test/inductor | `test_tensor_and_unbacked_symbol_closure` |
| test/inductor | `test_tensor_to_list_closure` |
| test/inductor | `test_eager_equivalence_log1p_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_numerics_cuda_bfloat16` |
| test/inductor | `test_unary_ufunc_numerical_exp2_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_eager_equivalence_log_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_numerics_cuda_float16` |
| test/inductor | `test_unary_ufunc_numerical_expm1_backend_inductor_default_cuda_float32` |
| test/inductor | `test_async_autotuner_cache_same_inputs` |
| test/inductor | `test_compilation_after_inactivity` |
| test/inductor | `test_tensor_with_unbacked_shape_closure` |
| test/inductor | `test_bmm_out_dtype` |
| test/inductor | `test_functional_call` |
| test/inductor | `test_grad_has_aux` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_numerics_cuda_float32` |
| test/inductor | `test_unary_ufunc_numerical_log10_backend_inductor_default_cuda_float32` |
| test/inductor | `test_cat_max_autotune_extern` |
| test/inductor | `test_grad_capture_tensor` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_default_cuda_float16` |
| test/inductor | `test_eager_equivalence_rsqrt_backend_inductor_default_cuda_float32` |
| test/inductor | `test_unary_ufunc_numerical_log_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_unary_ufunc_numerical_exp2_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_remainder_backend_inductor_default_cuda_float32` |
| test/inductor | `test_unary_ufunc_numerical_log_backend_inductor_default_cuda_float16` |
| test/inductor | `test_unary_ufunc_numerical_sqrt_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_unary_ufunc_numerical_exp_backend_inductor_default_cuda_float32` |
| test/inductor | `test_eager_equivalence_tanh_backend_inductor_default_cuda_float32` |
| test/inductor | `test_lazy_template_fusion_multiple_candidates_use_async_compile_True` |
| test/inductor | `test_grad_closure_scalar` |
| test/inductor | `test_grad_freevar_python_scalar` |
| test/inductor | `test_hops_in_bwd_inline_asm_elementwise_simple_cuda_float32` |
| test/inductor | `test_binary_ufunc_numerical_fmod_backend_inductor_default_cuda_float32` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_default_cuda_float16` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_default_cuda_bfloat16` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_default_cuda_float32` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_numerics_cuda_float32` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_numerics_cuda_bfloat16` |
| test/inductor | `test_binary_ufunc_numerical_remainder_backend_inductor_numerics_cuda_float16` |

**linalg** (1)

| Test File | Test Name |
|-----------|-----------|
| test/linalg | `test_addmm_relu_tunableop_rocm_cuda_float32` |

**modules** (2)

| Test File | Test Name |
|-----------|-----------|
| test/modules | `test_cpu_gpu_parity_nn_CTCLoss_cuda_float32` |
| test/modules | `test_forward_nn_CTCLoss_cuda_float32` |

**multiprocessing** (1)

| Test File | Test Name |
|-----------|-----------|
| test/multiprocessing | `test_fs` |

**multiprocessing_spawn** (1)

| Test File | Test Name |
|-----------|-----------|
| test/multiprocessing_spawn | `test_exception_all` |

**nn** (2)

| Test File | Test Name |
|-----------|-----------|
| test/nn | `test_pdist_inf_nan_propagation` |
| test/nn | `test_CTCLoss_cudnn_cuda` |

**profiler** (3)

| Test File | Test Name |
|-----------|-----------|
| test/profiler | `test_activity_filter_backward_compat` |
| test/profiler | `test_activity_filter_dict_syntax` |
| test/profiler | `test_kineto_kernel_metadata_in_trace` |

**scaled_matmul** (1)

| Test File | Test Name |
|-----------|-----------|
| test/scaled_matmul | `test_scaled_mm_deepseek_error_messages_bfloat16_lhs_block_128_rhs_block_1_M_256_N_256_K_256_cuda` |

**serialization** (2)

| Test File | Test Name |
|-----------|-----------|
| test/serialization | `test_load_njt_weights_only_should_import_False` |
| test/serialization | `test_debug_set_in_ci` |

**stateless** (1)

| Test File | Test Name |
|-----------|-----------|
| test/stateless | `test_private_stateless_warns` |

**utils** (1)

| Test File | Test Name |
|-----------|-----------|
| test/utils | `test_load_standalone` |

### DIFF
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -56,7 +56,7 @@ on:
         required: true
         type: string
       test_runs_on:
-        required: true
+        required: false
         type: string
         default: "linux-gfx942-1gpu-ossci-rocm"
       test_runs_on_multi_gpu:
--- a/external-builds/pytorch/run_pytorch_tests_full.py
+++ b/external-builds/pytorch/run_pytorch_tests_full.py
@@ -388,11 +388,7 @@ def build_run_test_cmd(
     if args.include:
         cmd.extend(["--include"] + args.include)
     test_dir = args.pytorch_dir / "test"
-    excludes = [
-        m
-        for m in EXCLUDED_TEST_MODULES
-        if (test_dir / (m + ".py")).exists()
-    ]
+    excludes = [m for m in EXCLUDED_TEST_MODULES if (test_dir / (m + ".py")).exists()]
     if args.exclude:
         excludes.extend(args.exclude)
     if excludes:
@@ -493,7 +489,10 @@ def _run_inductor(
         print(f"{'=' * 60}")
         print(f"Executing: {' '.join(cmd)}")
         result = subprocess.run(cmd, cwd=str(args.pytorch_dir))
-        print(f"run_test.py [{label}] finished with return code: {result.returncode}", flush=True)
+        print(
+            f"run_test.py [{label}] finished with return code: {result.returncode}",
+            flush=True,
+        )
         if result.returncode != 0:
             worst_rc = result.returncode
 


jithunnair-amd:
@ScottTodd Do you know why this PR is running all CI tests even though the `skip-ci` label is added?

ScottTodd:
A new `ci:skip` label was added in https://github.com/ROCm/TheRock/pull/4037. @geomin12 this is part of why I wanted to rename the existing label instead of creating a new one. I'll do that now.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,83 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+name: CI Nightly PyTorch Full Test
+
+on:
+  schedule:
+    - cron: "0 12 * * *" # 12:00 UTC, after nightly builds (~04:00 UTC) have completed
```
ScottTodd, line 25: Let's keep an eye on the trigger and defaults here. Maybe add some comments up here laying out the plans we have.

* This currently only runs scheduled tests on `gfx94X-dcgpu` Linux. In the long run I think we'll also want to test on Radeon, Linux and Windows. Ideally we'd test on all GPUs that we have available. We may want to encode "run full pytorch tests" in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py / https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/new_amdgpu_family_matrix.py so we have more central tracking of what builds/tests we enable on each GPU and platform
* This also assumes that the nightly Linux gfx94X build succeeds and completes before this trigger. Neither is necessarily true by construction - the build could fail or take longer than expected, in which case this would re-run tests using the most recent previous nightly build. We could add a separate workflow_dispatch trigger around these files, or find another approach:
  * https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml (see where `test_pytorch_wheels` runs)
  * https://github.com/ROCm/TheRock/blob/main/.github/workflows/release_portable_linux_packages.yml (see where "Trigger building PyTorch wheels" uses the `benc-uk/workflow-dispatch` action

I'd say the current behavior may be fine to start collecting data, but we'll want to solve for both of those design constraints relatively soon.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,83 @@
...
+            echo "Using provided torch_version: ${{ inputs.torch_version }}"
+            exit 0
+          fi
+
+          pip install packaging
+          TORCH_VERSION=$(pip index versions torch \
+            --index-url "${INDEX_URL}" \
+            --python-version "${PYTHON_VERSION}" 2>/dev/null \
+            | head -1 \
+            | sed 's/torch (\(.*\))/\1/')
```
ScottTodd, line 101: We should design for the ability to run these full tests on Windows too.

This `pip index versions torch` code will only include packages that match the current OS, so this `resolve_torch_version` job would need to run on `windows-*` if running tests on Windows, or we would need to add an explicit `--platform` argument.

```
λ pip index --help

Usage:
  pip index versions <package>


Description:
  Inspect information available from package indexes.

Index Options:
  --platform <platform>       Only use wheels compatible with <platform>. Defaults to the platform of the running
                              system. Use this option multiple times to specify multiple platforms supported by the
                              target interpreter.
```

On Windows:
```
λ pip index versions torch \
  --index-url https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu
ERROR: No matching distribution found for torch

λ pip index versions torch \
  --index-url https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu \
  --platform linux_x86_64
torch (2.10.0+rocm7.13.0a20260320)
Available versions: 2.10.0+rocm7.13.0a20260320, 2.10.0+rocm7.13.0a20260319, ...

λ pip index versions torch \
  --index-url https://rocm.nightlies.amd.com/v2-staging/gfx1151 \
  --platform win_amd64
torch (2.10.0+rocm7.13.0a20260320)
Available versions: 2.10.0+rocm7.13.0a20260320, 2.10.0+rocm7.13.0a20260319, ...

λ pip index versions torch \
  --index-url https://rocm.nightlies.amd.com/v2-staging/gfx1151 \
  --platform linux_x86_64
torch (2.10.0+rocm7.13.0a20260320)
Available versions: 2.10.0+rocm7.13.0a20260320, 2.10.0+rocm7.13.0a20260319, ...
```

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,83 @@
...
+          echo "torch_version=${TORCH_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "Resolved torch_version: ${TORCH_VERSION}"
+
+  run_full_tests:
+    needs: resolve_torch_version
+    uses: ./.github/workflows/test_pytorch_wheels_full.yml
+    with:
+      amdgpu_family: ${{ inputs.amdgpu_family || 'gfx94X-dcgpu' }}
+      test_runs_on: "linux-gfx942-1gpu-ossci-rocm"
+      test_runs_on_multi_gpu: "linux-gfx942-8gpu-ossci-rocm"
```
ScottTodd: The test runner should be included in the job's workflow_dispatch inputs and/or it should be loaded from scripts like
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_target_run.py

The pytorch build workflows use that configure_target_run.py script which pulls from amdgpu_family_matrix.py, here: https://github.com/ROCm/TheRock/blob/1b0705f93ac5fa0d0326b4826d7f3e4ccd502f9f/.github/workflows/build_portable_linux_pytorch_wheels.yml#L311-L331

If we put the job trigger in that workflow (right next to the existing test_pytorch_wheels?), we can use that already, then keep the workflow matching https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml / https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels_full.yml

.github/workflows/ci_nightly_pytorch_full_test.yml
ScottTodd, line 1: Thinking through my other comments, I'm not sure about the "ci_nightly_" file name prefix here. It makes sense given that this has its own schedule, but I think we might rather want to integrate into the existing workflows to get reliable triggering...

Another angle to consider is monitoring - if this does gets its own workflow file then we can give it its own status badge and workflow history page like https://github.com/ROCm/TheRock/actions/workflows/ci_nightly.yml?query=branch%3Amain . We might just end up monitoring job status on our https://therock-hud-dev.amd.com/ dashboard most of the time though.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,83 @@
...
+      package_index_url: "https://rocm.nightlies.amd.com/v2-staging"
```
ScottTodd: Similar comment about workflow inputs. It would be useful to be able to run from any index, not just rocm.nightlies.amd.com. We have these other channels too: https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#distribution-channels-dev-nightly-release

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -225,6 +225,11 @@ jobs:
             --log-cli-level=INFO \
             -v
 
+      - name: Disable pytest reruns
+        run: |
+          sed -i "s/rerun_options = \[\"-x\", \"--reruns=2\"\]/rerun_options = [\"-x\", \"--reruns=0\"]/" \
+            external-builds/pytorch/pytorch/test/run_test.py
```
ScottTodd: This is patching the source of https://github.com/pytorch/pytorch/blob/main/test/run_test.py? Uhhh that seems risky, and annoying for developers to reproduce. Should we contribute an option that would allow overriding the current behavior instead of this?

jithunnair-amd: Yes, that's what I suggested to @ethanwee1 as well, and if upstream approves having an env var to override the rerun count, that'd be perfect.

jithunnair-amd:
@ethanwee1  I'm converting this to Draft since we are still working through the test results to see if the PR is working as intended.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: https://github.com/pytorch/pytorch/pull/178007/changes

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: For now keeping it standalone so we can start collecting data and iterate on shard counts / timeout behavior independently. The dedicated workflow history page is also useful for monitoring while we stabilize. Will plan a follow-up to rewire it once the test suite is in a steady state.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+# TODO: This schedule-based trigger assumes the nightly build has completed.
+#       Consider integrating into the existing build/release workflows
+#       (build_portable_linux_pytorch_wheels.yml / release_portable_linux_packages.yml)
+#       so tests only run after a successful build.
+
+name: CI Nightly PyTorch Full Test
+
+on:
+  schedule:
+    - cron: "0 12 * * *" # 12:00 UTC, after nightly builds (~04:00 UTC) have completed
```
HereThereBeDragons, line 25: do we want to run it after every nightly?

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+        type: string
+        default: ""
+      pytorch_git_ref:
+        description: "PyTorch git ref for test sources"
+        type: string
+        default: "nightly"
+      pytorch_git_repo:
+        description: "PyTorch git repo for test sources (Options: rocm/pytorch, pytorch/pytorch)"
+        type: string
+        default: "rocm/pytorch"
```
HereThereBeDragons: if default pytorch_git_ref is nightly, this should be pytorch/pytorch?

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+        default: "default distributed inductor"
+      python_version:
+        description: "Python version"
+        type: string
+        default: "3.12"
+
+permissions:
+  contents: read
+
+run-name: CI Nightly PyTorch Full Test (${{ inputs.amdgpu_family || 'gfx94X-dcgpu' }})
```
HereThereBeDragons:
```suggestion
run-name: CI Nightly PyTorch Full Test (${{ inputs.amdgpu_family }})
```
might be wrong but inputs.amdgpu_family has always a value: your default value so no need for this extra "if empty"

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+jobs:
+  resolve_torch_version:
+    runs-on: ubuntu-24.04
+    outputs:
+      torch_version: ${{ steps.resolve.outputs.torch_version }}
+    steps:
+      - name: Resolve latest nightly torch version
+        id: resolve
+        run: |
+          AMDGPU_FAMILY="${{ inputs.amdgpu_family || 'gfx94X-dcgpu' }}"
```
HereThereBeDragons: same here

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+          PYTHON_VERSION="${{ inputs.python_version || '3.12' }}"
```
HereThereBeDragons: here

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+          PACKAGE_INDEX_URL="${{ inputs.package_index_url || 'https://rocm.nightlies.amd.com/v2-staging' }}"
```
HereThereBeDragons: here

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+          INDEX_URL="${PACKAGE_INDEX_URL}/${AMDGPU_FAMILY}"
+
+          if [ -n "${{ inputs.torch_version }}" ]; then
```
HereThereBeDragons, line 84: please no bash - windows gets unhappy with it.

have a look at external-builds/pytorch/run_pytorch_tests.py as there is already logic in it for auto detection of  the version (to be more specific: external-builds/pytorch/pytorch_utils.py function `def detect_pytorch_version() -> str:`)

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -116,7 +123,8 @@ jobs:
       matrix: ${{ fromJSON(needs.prepare_matrix.outputs.matrix) }}
     container:
       image: ${{ contains(inputs.test_runs_on, 'linux') && 'ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26' || null }}
-      options: --ipc host
+      options: --shm-size=10g
+        --cap-add=SYS_PTRACE
```
HereThereBeDragons, line 130: are these specific test that need it or is it for the monitoring of the tests?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -138,6 +146,7 @@ jobs:
       CI: "1"
       PYTORCH_TEST_WITH_ROCM: "1"
       PYTORCH_TESTING_DEVICE_ONLY_FOR: "cuda"
+      OPENBLAS_NUM_THREADS: "64"
```
HereThereBeDragons: i am not sure this is a sane assumption that all our machines have at least 64 vcores in the container. why does this needs to be statically set?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -153,14 +162,10 @@ jobs:
...
-      - name: Summarize workflow inputs
+      - name: Install system dependencies
+        if: ${{ runner.os == 'Linux' }}
         run: |
-          python build_tools/github_actions/summarize_test_pytorch_workflow.py \
-            --torch-version=${{ inputs.torch_version }} \
-            --pytorch-git-ref=${{ inputs.pytorch_git_ref }} \
-            --index-url=${{ inputs.package_index_url }} \
-            --index-subdir=${{ inputs.amdgpu_family }}
+          apt-get update && apt-get install -y build-essential
```
HereThereBeDragons: for what do you need it? was it the clang vs gcc story?

external-builds/pytorch/skip_tests/pytorch_2.12.py
```diff
@@ -17,6 +17,17 @@
             # Same hipblas.h compilation error as test_mempool_with_allocator.
             # See https://github.com/pytorch/pytorch/pull/173330
             "test_mempool_expandable",
+            # torch.AcceleratorError: HIP error: operation not permitted when
```
HereThereBeDragons, line 18: did you test on more than gfx94X?
i am wondering if these are tests failing on all archs or just gfx94X specific

external-builds/pytorch/skip_tests/pytorch_2.12.py
```diff
@@ -17,6 +17,17 @@
...
+            # stream is capturing
+            "test_cuda_graph_tensor_item_not_allowed",
+            # AssertionError: CalledProcessError not raised
+            "test_allocator_memory_fraction_setting",
+            # AssertionError: False is not true (memory_plots)
+            "test_memory_plots",
```
HereThereBeDragons: we might want to add it to `generic.py`. i see we are skipping it for all pytorch versions so far other than 2.12
same for test_memory_snapshot

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -93,11 +111,66 @@
 ]
 
 
+PYTEST_TIMEOUT_SECONDS = 900  # 15 minutes per test function
```
HereThereBeDragons, line 106: that seems like a large timeout for the thousand and thousand of tests we are running?

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -328,15 +399,103 @@ def build_run_test_cmd(
...
+    """
+    # Upstream runs verify_dynamo.py first as a quick smoke test.
+    verify_script = args.pytorch_dir / "tools" / "dynamo" / "verify_dynamo.py"
+    if verify_script.exists():
+        print("Running verify_dynamo.py …")
+        vr = subprocess.run(
+            [sys.executable, str(verify_script)], cwd=str(args.pytorch_dir)
+        )
+        if vr.returncode != 0:
+            print(f"verify_dynamo.py failed with return code {vr.returncode}")
```
HereThereBeDragons, line 474: worth capturing the output of stderr?

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -328,15 +399,103 @@ def build_run_test_cmd(
...
+    worst_rc = 0
+    for label, cmd in zip(labels, cmds):
+        print(f"\n{'=' * 60}")
+        print(f"Inductor phase: {label}")
+        print(f"{'=' * 60}")
+        print(f"Executing: {' '.join(cmd)}")
+        result = subprocess.run(cmd, cwd=str(args.pytorch_dir))
+        print(f"run_test.py [{label}] finished with return code: {result.returncode}")
+        if result.returncode != 0:
+            worst_rc = result.returncode
```
HereThereBeDragons, line 497: worth printing a list afterwards of just the failed labels/error codes?

.github/workflows/ci_nightly_pytorch_full_test.yml
ScottTodd: See my earlier review comment here (https://github.com/ROCm/TheRock/pull/3929/changes#r2967948714, currently marked as "resolved" as some comments were added).

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,116 @@
...
+          echo "torch_version=${TORCH_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "Resolved torch_version: ${TORCH_VERSION}"
+
+  run_full_tests:
+    needs: resolve_torch_version
+    uses: ./.github/workflows/test_pytorch_wheels_full.yml
+    with:
+      amdgpu_family: ${{ inputs.amdgpu_family || 'gfx94X-dcgpu' }}
+      test_runs_on: ${{ inputs.test_runs_on || 'linux-gfx942-1gpu-ossci-rocm' }}
+      test_runs_on_multi_gpu: ${{ inputs.test_runs_on_multi_gpu || 'linux-gfx942-8gpu-ossci-rocm' }}
```
ScottTodd: Also here - trust the default value on the workflow_dispatch input. If someone omits an input, I'd rather either:
* The workflow fails
* The workflow derives values on its own (e.g. "no version specified --> use latest version" or "amdgpu family is gfx94X --> run on linux-gfx942 runner")

What I want to avoid is someone specifying say:
```
amdgpu_family=gfx1151
test_runs_on=
```
and that running on `linux-gfx942-1gpu-ossci-rocm` instead of finding the correct runner or failing

.github/workflows/ci_nightly_pytorch_full_test.yml
ScottTodd: This code should technically work on Windows under `shell: bash`, and I'm willing to look the other way on this despite the style guide section (https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash) this time.

Rather than extract this into a script, I think we can structure the workflows such that a `resolve_torch_version` job is not needed at all. Ideas:
* When triggered after building pytorch, we know the version that we just built using https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/write_torch_versions.py, see https://github.com/ROCm/TheRock/blob/e1000f1f08e232a9e12be98b0cfd33887d7f443e/.github/workflows/build_portable_linux_pytorch_wheels.yml#L240-L249 https://github.com/ROCm/TheRock/blob/e1000f1f08e232a9e12be98b0cfd33887d7f443e/.github/workflows/build_portable_linux_pytorch_wheels.yml#L138-L157
* When triggered manually, an explicit version can be specified 
* We could build some sort of trigger using Quartz (https://github.com/ROCm/TheRock/pull/3728) or another system where after nightly ROCm releases we trigger test jobs. That trigger can include in its payload all the package versions that were built and test jobs could choose from among them, without needing to do their own queries on the index
* Maybe the workflows don't need to know all these versions ahead of time and the scripts could just work with what they are given ("install latest", "download from this git tag", etc.)

---

I'd say here we can proceed as-is then refactor later as we integrate this nightly test workflow more closely with other workflows.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -196,6 +192,48 @@ jobs:
...
+          echo "CMAKE_PREFIX_PATH=${ROCM_CMAKE_PREFIX}:${CMAKE_PREFIX_PATH:-}" >> "${GITHUB_ENV}"
+          echo "HIP_DEVICE_LIB_PATH=${ROCM_HOME}/lib/llvm/amdgcn/bitcode" >> "${GITHUB_ENV}"
+          echo "ROCM_DEVICE_LIB_PATH=${ROCM_HOME}/lib/llvm/amdgcn/bitcode" >> "${GITHUB_ENV}"
+          echo "ROCM_SYSDEPS_INCLUDE=${ROCM_SYSDEPS_INCLUDE}" >> "${GITHUB_ENV}"
+          echo "CPLUS_INCLUDE_PATH=${ROCM_SYSDEPS_INCLUDE}:${CPLUS_INCLUDE_PATH:-}" >> "${GITHUB_ENV}"
+          echo "C_INCLUDE_PATH=${ROCM_SYSDEPS_INCLUDE}:${C_INCLUDE_PATH:-}" >> "${GITHUB_ENV}"
+          echo "PKG_CONFIG_PATH=${ROCM_SYSDEPS_PKGCONFIG}:${PKG_CONFIG_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LD_LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LD_LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "MAGMA_HOME=${ROCM_HOME}/magma" >> "${GITHUB_ENV}"
```
ScottTodd: We aren't building magma in TheRock's pytorch builds so I don't think this will do anything. Is this env var required despite that?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -227,11 +265,24 @@ jobs:
...
             --import-mode=importlib
 
+      - name: Upload test reports
+        if: always()
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: test-reports-${{ matrix.test_config }}-${{ matrix.shard }}-${{ matrix.num_shards }}
+          path: external-builds/pytorch/pytorch/test/test-reports/
+          if-no-files-found: ignore
+          retention-days: 1
```
ScottTodd: We can put these on S3 eventually. For now, how about increasing the `retention-days`? If the files are small (under 1GB), for this trigger frequency I'd be fine with 15 days, 30 days, 45 days, etc. (the organization limit for artifact and log retention is 45 days)

Relevant docs:
* https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md#directory-structure
* https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#ci-buckets

external-builds/pytorch/run_pytorch_tests_full.py
ScottTodd, line 1: We could update these docs at some point to reference this script and the different workflows we have: https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#running-full-pytorch-tests

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: Good catch — fixed. Default changed to `pytorch/pytorch` since the `nightly` branch lives on upstream.

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: Fixed — removed all the fallback defaults. Inputs always have default values, so the `|| '...'` patterns were unnecessary.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: We set it because leaving it unset was causing crashes in some UT's test_entropy_monte_carlo for example, and we wanted to get around the crash quicky.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Some of the PyTorch UT's need JIT, so without it some UT's fail

external-builds/pytorch/skip_tests/pytorch_2.12.py
ethanwee1: We only tested on gfx942. Once we open up more testing we can add more specific GFX skips

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: We will remove this in a followup PR, this is for debugging purposes

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Ah, we copied that from our upstream CI script. Removed.

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: Addressed the redundant fallback defaults — the with: block now passes ${{ inputs.* }} directly, trusting the workflow_dispatch default values. The other issue (deriving runner labels from amdgpu_family automatically, or failing on mismatches) is tracked in the TODO at the top of the file

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: This was added specifically because we had shards hanging for 6+ hours on tests like nn/test_convolution. The timeout prevented individual tests from running forever.

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: since this is meant for CI, we can add it in a followup PR that wires in more capability

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: We'll add a summary at the end of both phases listing which ones failed and their return codes. Will do in a follow-up.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Increased to 30 days.

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: Will address in a followup PR.

external-builds/pytorch/skip_tests/pytorch_2.12.py
ethanwee1: added

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd: @ethanwee1 @pragupta Please elaborate how these flags are debugging-only. Please also file an issue on theRock to address all the TODOs in this PR, so we don't lose track of them.

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd: @ethanwee1 Reading more about OPENBLAS_NUM_THREADS, it appears that the monte carlo test is especially sensitive to variations due to different number of threads used for OpenBLAS. Leaving the variable unset uses the number of CPU cores available, but potentially introduces non-determinism during reduction. As a result, setting it to 1 is the way to get a deterministic answer. We should also add a comment above this variable setting to explain that this is needed for deterministic computation.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -0,0 +1,124 @@
...
+        default: "default distributed inductor"
+      python_version:
+        description: "Python version"
+        type: string
+        default: "3.12"
+
+permissions:
+  contents: read
+
+run-name: CI Nightly PyTorch Full Test (${{ inputs.amdgpu_family }})
```
jithunnair-amd: `pytorch_git_ref` would be the other critical input to reflect in the run name to easily distinguish the runs for various branches.

ethanwee1: Added

.github/workflows/ci_nightly_pytorch_full_test.yml
jithunnair-amd: If this workflow were to be triggered by an upstream workflow, I'd suggest we pass the exact version down to avoid any ambiguity or surprises. However, for a manual trigger, I do like the ability to "just pull the latest" from the package index url, and thus figure out the version if it's not already specified, so I'd vote for retaining this.

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd, line 19: @ethanwee1 Why is this one not `required: true`, when the `test_runs_on` is? That could result in the user running single-GPU tests on a different architecture than the multi-GPU ones. Let's force it to be "correct-by-design".

jithunnair-amd: Looking at [amdgpu_family_matrix.py](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py) though, it seems we don't always have multi-GPU runner labels for every case...

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -37,14 +37,18 @@ on:
         description: PyTorch ref to checkout test sources from. (e.g. "nightly", or "release/2.9")
         type: string
         default: "release/2.9"
```
jithunnair-amd: @ScottTodd Should these defaults be `nightly` and `pytorch/pytorch` to:
1. run with latest supported pytorch by default
2. reduce maintenance burden by not needing updates as we move to newer versions

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd, line 60: @ethanwee1 Why no default `linux-gfx942-1gpu-ossci-rocm` here?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -69,6 +73,9 @@ on:
       pytorch_git_ref:
         type: string
         default: "release/2.9"
+      pytorch_git_repo:
+        type: string
+        default: "rocm/pytorch"
```
jithunnair-amd: Same comment as https://github.com/ROCm/TheRock/pull/3929/changes#r3049059699

ethanwee1: Updated

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Added a default

ethanwee1: Yeah, we have it for a few, not all

jithunnair-amd: Also, the multi_gpu runner label is only relevant if `distributed` is in the list of `test_configs`. So maybe this one can't be made `required`, and since it's in the `workflow_dispatch` trigger, we take a "caller's responsibility" approach?

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -227,11 +270,24 @@ jobs:
 
       - name: (Linux) Run PyTorch full test suite
         if: ${{ runner.os == 'Linux' }}
+        env:
+          PYTORCH_NUM_PYTEST_RERUNS: "2"
```
jithunnair-amd: @ethanwee1 Why is this env var here as opposed to where all the other PyTorch testing-related env vars (e.g. `PYTORCH_TESTING_DEVICE_ONLY_FOR`) are located?

external-builds/pytorch/run_pytorch_tests_full.py
jithunnair-amd: @ethanwee1 This last line is not needed anymore since pytorch/pytorch#176445 is merged

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -227,11 +270,24 @@ jobs:
...
         run: |
-          python ./external-builds/pytorch/run_pytorch_tests_full.py -- \
+          python ./external-builds/pytorch/run_pytorch_tests_full.py \
+            --exclude dynamo/test_dynamic_shapes functorch/test_control_flow \
```
jithunnair-amd: Why are these testfiles excluded here instead of as part of [`EXCLUDED_TEST_MODULES`](https://github.com/ROCm/TheRock/pull/3929/changes#diff-ba2472b298c8bf5591dba3fbcd4ea6031bfaa4ccd5e6b7e09cc40dd7f328cad9R120)?

.github/workflows/test_pytorch_wheels_full.yml
jithunnair-amd: @ethanwee1 Is `gcc` (which comes with `build-essential`) a strict requirement, or can we use the `clang` that ships with ROCm?

external-builds/pytorch/skip_tests/pytorch_2.12.py
jithunnair-amd: @ethanwee1 I think you added `test_memory_snapshot` to `generic.py`, but not `test_memory_plots`?

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -328,15 +406,103 @@ def build_run_test_cmd(
...
+    ]
+
+    worst_rc = 0
+    for label, cmd in zip(labels, cmds):
+        print(f"\n{'=' * 60}")
+        print(f"Inductor phase: {label}")
+        print(f"{'=' * 60}")
+        print(f"Executing: {' '.join(cmd)}")
+        result = subprocess.run(cmd, cwd=str(args.pytorch_dir))
+        print(f"run_test.py [{label}] finished with return code: {result.returncode}")
```
jithunnair-amd:
```suggestion
        print(f"run_test.py [{label}] finished with return code: {result.returncode}", flush=True)
```

I do see the `print(f"run_test.py [{label}] finished with return code: {result.returncode}")` output for the first phase, but not the second phase in [this job](https://github.com/ROCm/TheRock/actions/runs/24086897853/job/70262249816). I wonder if it's because the print needs to be flushed.

external-builds/pytorch/run_pytorch_tests_full.py
jithunnair-amd: @ethanwee1 Is this not already done above? Or were you planning to do something else in addition?

.github/workflows/test_pytorch_wheels_full.yml
ScottTodd: We currently build `release/2.10` on every commit to TheRock. That insulates us a bit from upstream, so PRs here shouldn't be blocked by breaks due to changes outside of their control: https://github.com/ROCm/TheRock/blob/58d321a8c88ecad8f52a40edab70d952a4d71234/.github/workflows/ci_linux.yml#L211-L215

In our nightly releases we test across each supported version.

For this "full tests" workflow, I'd lean towards maybe the latest stable release version and the latest nightly.

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Moved

external-builds/pytorch/run_pytorch_tests_full.py
ethanwee1: The for loop should dump the exit codes

Issue #3332, opened 2026-02-09
[Multi-arch] Build multi-arch PyTorch Python packages


ScottTodd:
https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0008-Multi-Arch-Packaging.md#pytorch-packages

PyTorch wheels will be optimized for WheelNext using split packages:

```
# Base wheel (host code)
torch_rocm-2.12.0-cp311-linux_x86_64.whl

# Architecture-specific wheels (device code)
torch_rocm_gfx110X-2.12.0-cp311-linux_x86_64.whl
torch_rocm_gfx94X-2.12.0-cp311-linux_x86_64.whl

# WheelNext will dynamically resolve dependencies based on detected GPU or explicit hints
```

Implementation plan:

1. Build PyTorch packages for individual GPU targets or families (i.e. using existing `PYTORCH_ROCM_ARCH` env var setting), either with the existing set of 2+4+4+1+... list or by fanning out to 1+1+1+1+1+...
2. Run a (not yet implemented) tool to repackage those packages into a base wheel (containing host code) and architecture-specific wheels (containing device code)
3. Construct pip/uv install commands to install the right wheels for a given configuration manually
4. Set up a wheelnext variant provider that dynamically installs the right wheels for a given configuration (with override env vars for usage on non-GPU machines and for testing)

References:

* https://wheelnext.dev/
* https://www.amd.com/en/developer/resources/technical-articles/2025/pytorch-2-9-wheel-variant-support-expands-to-rocm.html
* https://github.com/wheelnext/amd-variant-provider
* https://github.com/wheelnext/variants-index/

marbre:
@ScottTodd I think we should move wheelnext to it's own issue and call this done as soon as we have CI (which we already have) + release workflows in place.

ScottTodd:
Sure, we can split the issue.

ScottTodd:
CI support for building **torch 2.10** for **Python version 3.12** was added in https://github.com/ROCm/TheRock/pull/4783 . We're working next on getting these into the nightly releases across the full "torch version" x "python version" matrix.

Dev torch packages from https://github.com/ROCm/TheRock/actions/runs/24915005026?pr=4783 are available for testing (and future commits will build and upload similar packages)

CI artifact index pages:
* https://therock-ci-artifacts.s3.amazonaws.com/24915005026-windows/python/index.html
[...]

PR #3944, opened 2026-03-12
[pytorch] Add build_dev.py for creating editable PyTorch build

rkayaith:
Adds `build_dev.py` for local development. Does an editable install (`pip install -e .`) so Python changes are immediate and C++ changes only require an incremental rebuild.

This is for the most part just re-using code from `build_prod_wheels.py`. Common code has been extracted into a new `pytorch_build_env.py` module.

### DIFF
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -132,138 +132,35 @@ inline system deps into the audio and vision wheels as needed.
 """
 
 import argparse
-import json
 import os
 from pathlib import Path
-from packaging.version import parse
-import platform
 import shutil
-import shlex
 import subprocess
 import sys
 import tarfile
 import tempfile
-import textwrap
 import urllib.request
+
+from pytorch_build_env import (
+    capture,
+    directory_if_exists,
+    do_install_rocm,
+    find_built_wheel,
+    find_dir_containing,
+    get_installed_package_version,
+    get_rocm_env,
+    get_version_suffix_for_installed_rocm_package,
+    is_windows,
+    prepare_pytorch_build,
+    remove_dir_if_exists,
+    run_command,
+    sanity_check_pytorch,
+)
 
 script_dir = Path(__file__).resolve().parent
 
-is_windows = platform.system() == "Windows"
-
 # LLVM download URL for triton-windows
 LLVM_BASE_URL = "https://oaitriton.blob.core.windows.net/public/llvm-builds"
-
-# List of library preloads for Linux to generate into _rocm_init.py
-LINUX_LIBRARY_PRELOADS = [
-    "amd_comgr",
-    "amdhip64",
-    "rocprofiler-sdk",  # Linux only: needed by torch since kineto uses rocprofiler-sdk.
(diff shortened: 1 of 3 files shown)


external-builds/pytorch/build_prod_wheels.py
```diff
@@ -1272,9 +1315,14 @@ def add_common(p: argparse.ArgumentParser):
...
         type=Path,
-        required=True,
         help="Directory to copy built wheels to",
     )
+    build_p.add_argument(
+        "--editable",
+        action="store_true",
+        default=False,
+        help="Editable install packages (pip install -e) instead of building wheels",
+    )
```
ScottTodd: Drive-by: I think this is stretching the "prod" part of the "build_prod_wheels.py" name. What workflow do you want to support? We could split this file up into several smaller files more fit for interactive development, but the longer term solution is really to upstream ~all of this code and just call upstream build scripts. (We'll still need _some_ code here for managing version pins, source code checkouts, and setting ROCm-specific configuration options that need to live with the ROCm source, but this file is much larger than it needs to be for that right now).

rkayaith: I'm happy to refactor some of this out, but the gist is that I'm doing pytorch + rocm development, but building requires a whole bunch of env vars that aren't really documented as far as I could tell. TheRock has this all figured out and works great locally, with the exception that it builds wheels instead of doing `pip install -e`. I'm mainly just looking for a simple build setup that can be shared with other devs.

ScottTodd: @ethanwee1 @jithunnair-amd do you have thoughts here? Some of the scripts / environment variables / etc. were ported to upstream in https://github.com/pytorch/pytorch/pull/168377 and related PRs. I'd like for the natural place for "pytorch + rocm development" docs and scripts to be upstream, not downstream here in TheRock so I'm hesitant to let this "build_prod_wheels.py" script grow more features like this when it's already nearly 1.4k LOC.

ethanwee1: As far as I understand, this script is meant to be used in CI and is "not optimized for providing a development experience for PyTorch." It sets up the build env, gets the rocm wheels, then builds pytorch wheels. This PR then adds in the ability to get the env var's setup already. I don't think it should be upstreamed/not sure where it would be as these seem theRock wheel specific. If we do want something for dev setup for wheels, could we get a python file similar to https://github.com/ROCm/TheRock/blob/main/dockerfiles/install_rocm_tarball.sh? We could move _setup_common_build_env() and do_build_pytorch() in the same file for dev work. Either way, I think this turns build_prod_wheels.py into almost a swiss-army-knife kind of script

rkayaith: I've refactored this, moving the new functionality into a new `build_dev.py` script, and moving the shared code into a new module `pytorch_build_env.py`.

ScottTodd (commented):
Structurally this looks okay to me. Timing-wise I think we should sequence this *after* we rework the torch packages to support multi-arch rocm packages (https://github.com/ROCm/TheRock/issues/3332)

Can someone else do the first round code review too? I won't have time myself.

PR #3952, opened 2026-03-12
Extract hipdnn into new ml-libs-generic artifact group

marbre:
hipdnn is architecture-independent and has no per-GPU-family code; its dependencies (core-runtime, core-hip, spdlog) are all available after the generic compiler-runtime stage. Grouping it in the per-arch ml-libs stage wastes one build per GPU family and misrepresents the dependency graph.

Introduce a new generic ml-libs-generic artifact group and CI stage that runs in parallel with the per-arch math-libs stage:

- BUILD_TOPOLOGY.toml:
  - Add build_stages.ml-libs-generic and artifact_groups.ml-libs-generic
  - Move artifacts.hipdnn from ml-libs → ml-libs-generic
  - Add ml-libs-generic to artifact_group_deps of ml-libs (so the artifact manager fetches hipdnn before building miopenprovider, hipblasltprovider, and hipdnn-samples)
  - Add ml-libs-generic to artifact_group_deps of fusilli-libs (so hipdnn is available when building fusilliprovider)

- CI workflows (Linux + Windows):
  - Add ml-libs-generic job (needs: compiler-runtime)
  - Update math-libs needs to include ml-libs-generic
  - Update fusilli-libs needs to use ml-libs-generic instead of math-libs (Linux only; Windows has no fusilli-libs stage)

Multi-arch CI run: https://github.com/ROCm/TheRock/actions/runs/23029434143

### DIFF
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -272,7 +272,8 @@ source_sets = ["rocm-libraries", "rocm-systems", "math-libs"]
 description = "Architecture-neutral ML library API (hipDNN)"
 type = "generic"
 artifact_group_deps = ["hip-runtime", "third-party-libs"]
-source_sets = ["rocm-libraries"]
+# TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
+source_sets = ["rocm-libraries", "rocm-systems"]
 
 [artifact_groups.ml-libs]
 description = "Machine learning libraries"


.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -6,14 +6,15 @@
...
-# 9. fusilli-libs (generic) - Fusilli hipdnn provider (after math-libs + iree-compiler)
-# 10. media-libs (generic) - sysdeps-amd-mesa, rocdecode, rocjpeg
+# 3. ml-libs-generic (generic) - hipdnn (parallel to math-libs)
+# 4. math-libs (per-arch) - BLAS, FFT, etc. (needs ml-libs-generic for ml-libs group)
+# 5. comm-libs (per-arch) - RCCL (parallel to math-libs)
+# 6. debug-tools (generic) - amd-dbgapi, rocr-debug-agent, rocgdb (parallel to math-libs)
+# 7. dctools-core (generic) - RDC (parallel to math-libs)
+# 8. profiler-apps (generic) - rocprofiler-systems (parallel to math-libs)
+# 9. iree-compiler (generic) - IREE compiler (parallel to math-libs)
+# 10. fusilli-libs (generic) - Fusilli hipdnn provider (after ml-libs-generic + iree-compiler)
```
AaronStGeorge, line 16: If `hipDNN` moves out of `math-libs` we might want to combine into a single `iree-libs` stage. The split was to do a little latency hiding as the IREE compiler could be built parallel to `math-libs`, but this new `ml-libs-generic` stage is likely pretty fast so the latency hiding would be minimal.

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -85,11 +86,29 @@ jobs:
       contents: read
       id-token: write
 
+  # ==========================================================================
+  # STAGE: ml-libs-generic (generic, parallel to math-libs)
+  # ==========================================================================
+  ml-libs-generic:
```
ScottTodd, line 92: This is not "parallel to math-libs" if math-libs take a dep on it.

BUILD_TOPOLOGY.toml
```diff
@@ -264,10 +268,17 @@ artifact_group_deps = ["hip-runtime", "profiler-core"]
 # TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
 source_sets = ["rocm-libraries", "rocm-systems", "math-libs"]
 
+[artifact_groups.ml-libs-generic]
+description = "Architecture-neutral ML library API (hipDNN)"
+type = "generic"
+artifact_group_deps = ["hip-runtime", "third-party-libs"]
+# TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
+source_sets = ["rocm-libraries", "rocm-systems"]
```
ScottTodd, line 276: It should make sense to add this stage and artifact in the long term, but introducing a new stage adds substantial overhead to the pipelines as they exist today:

Platform | Build Time | Total Job Time (including overhead) | Logs 
-- | -- | -- | --
Linux | 11s | 5m09s | [logs here](https://github.com/ROCm/TheRock/actions/runs/23029434143/job/66888121035)
Windows | 1m48s | 10m34s | [logs here](https://github.com/ROCm/TheRock/actions/runs/23029434143/job/66889723415)

Dropping the dep on `rocm-systems` sources should help a bit there, but given the fast build time I think we may be better off keeping this as part of the existing target-specific math-libs stage :/

.github/workflows/multi_arch_build_portable_linux.yml
ScottTodd: If iree-compiler and fusilli-libs are both generic I agree that they could build together in a single stage, saving job overhead.

When we get further with pre-build artifact caching (https://github.com/ROCm/TheRock/issues/3399), the decision on whether or not to split the stages will get fuzzier. If iree-compiler (including its own LLVM build) is updated infrequently but fusilli-libs (faster to build) is updated frequently, keeping them separated would let us more aggressively cache the whole iree-compiler stage. Finer-grained caching of just the iree-compiler artifacts would still be possible if the stages were merged.

.github/workflows/multi_arch_build_portable_linux.yml
marbre: Good fetch.

PR #3953, opened 2026-03-13
Add missing third_party/printf to iree nested submodules

zjgarvey:
## Summary
- iree-org/iree#23694 (merged 2026-03-09) replaced all libc `snprintf`/`vsnprintf` calls in the IREE runtime with `eyalroz/printf`, making `third_party/printf` an unconditional build dependency (`printf::printf` is required by `iree::base`).
- `third_party/printf` was missing from the default `--nested-submodules` list for `iree` in `fetch_sources.py`, causing it not to be fetched and the IREE build to fail.

### DIFF
--- a/build_tools/fetch_sources.py
+++ b/build_tools/fetch_sources.py
@@ -434,6 +434,7 @@ def main(argv):
                     "third_party/benchmark",
                     "third_party/llvm-project",
                     "third_party/torch-mlir",
+                    "third_party/printf",
                 ],
             )
         ],


build_tools/fetch_sources.py
```diff
@@ -434,6 +434,7 @@ def main(argv):
                     "third_party/benchmark",
                     "third_party/llvm-project",
                     "third_party/torch-mlir",
+                    "third_party/printf",
```
ScottTodd, line 437: This may be going away soon: https://github.com/iree-org/iree/pull/23761

Fine to include for now, but we'll just need to remove it later. WDYT?

AaronStGeorge: I can pull it out on he next iree/fusilli submodule bump. I'd like to bump soon, and this change is needed, so it would be nice to get it merged.

ScottTodd: https://github.com/iree-org/iree/pull/23761 was merged two hours ago. Does that change how you want to proceed here?

AaronStGeorge: Yeah it's fine, bumps are manual until the build is on by default and I create an auto-bump workflow.

Currently the plan of record is to bump IREE + fusilli in TheRock to top of main fusilli and whatever IREE fusilli is using (bumped nightly to the latest IREE release) on a nightly cadence.

ScottTodd: Ok, I'm fine with this either way then.

marbre:
This broke multi-arch CI as printf is not a submodule in the currently pinned IREE submodule. Sent PR #4003 to revert.

PR #3960, opened 2026-03-13
[test-shift-left] increasing rocsolver test coverage

madkasul:
## Motivation

Increasing rocsolver test coverage as part of test shift left activities

## Technical Details

With this change, when called with `test_type = full` the test script will launch rocsolver tests with no gtest-filters. Which inturn runs all tests of the suite.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -150,8 +150,8 @@ test_matrix = {
         "job_name": "rocsolver",
         "fetch_artifact_args": "--blas --tests",
         # 68350(approx) tests needs 48 mins, so 48 mins / 2 shards = 24 mins per shard
-        # 24 mins + 20% margin = 30 mins
-        "timeout_minutes": 30,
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -149,13 +149,15 @@ def _get_script_path(script_name: str) -> str:
     "rocsolver": {
         "job_name": "rocsolver",
         "fetch_artifact_args": "--blas --tests",
+        # 68350(approx) tests needs 48 mins, so 48 mins / 2 shards = 24 mins per shard
+        # 24 mins + 20% margin = 30 mins
         "timeout_minutes": 30,
```
geomin12: 30 mins is pretty tight for 24 mins. we should extend the timeout even more

~ 6 mins could be hit from GPU delays / lag

madkasul:
> eventually, we want to get to these categories for all components: https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md
> 
> in some time, we'll have timeouts per level and this will start failing

Sure, we can consider at that time.

Issue #3950, opened 2026-03-12
[Issue] `llvm-strip` causes `libunwind.so.99` to have offset ELF load command


kcossett-amd:
When I pulled an `amdrocm-profiler7.12_7.12` artifact (for ex: [amdrocm-profiler7.12_7.12.0~20260311-22937000989_amd64.deb](https://rocm.nightlies.amd.com/deb/20260311-22937000989/pool/main/amdrocm-profiler7.12_7.12.0~20260311-22937000989_amd64.deb)) , and I ran the following:

`ldd <path-to-extract>/opt/rocm/core-7.12/bin/rocprof-sys-instrument`, it fails with the following:
```
error while loading shared libraries: libunwind.so.99: ELF load command address/offset not page-aligned
```

Following my investigation from #3932 , the problem lies with `llvm-strip` being used on the library. If in a fresh build, you package the binary without having `llvm-strip` being used, the `ldd` command above does not fail. However, after using `llvm-strip`, it fails with the exact error above. 

**To reproduce the problem locally**:

```
# Build the profiler in the manylinux container
cmake -S /therock/src -B /therock/output/build -GNinja \
  -DTHEROCK_AMDGPU_FAMILIES=gfx94X-all \
  -DTHEROCK_ENABLE_ALL=OFF \
  -DTHEROCK_ENABLE_PROFILER=ON \
  -DTHEROCK_ENABLE_CORE_RUNTIME=ON \
  -DTHEROCK_ENABLE_HIP_RUNTIME=ON \
  -DTHEROCK_ENABLE_COMPILER=ON
cmake --build /therock/output/build -- -l 16
```
Then using the `build_package.py` script, build the deb package:
```
python3 build_tools/packaging/linux/build_package.py \
  --artifacts-dir output-linux-portable/build/artifacts \
  --dest-dir build/packages \
  --target gfx94X-all \
  --pkg-type deb \
  --rocm-version 7.12.0 \
  --pkg-names amdrocm-profiler
```
Extract it:
```
dpkg-deb -x build/packages/amdrocm-profiler7.12_7.12.0_amd64.deb build/packages/extracted
```
Then run the `ldd` command and observe the failure
```
ldd build/packages/extracted/opt/rocm/core-7.12/bin/rocprof-sys-instrument
build/packages/extracted/opt/rocm/core-7.12/bin/rocprof-sys-instrument: error while loading shared libraries: libunwind.so.99: ELF load command address/offset not page-aligned
```

[...]

PR #3977, opened 2026-03-13
Pin LLVM tools to version 20 to fix ELF page alignment issue

raramakr:
Recent versions of llvm-strip have a bug that breaks ELF LOAD segment
  alignment, causing "address/offset not page-aligned" errors at runtime.
  This manifests as libraries like libunwind.so.99 failing to load with
  the error: "ELF load command address/offset not page-aligned".

  The bug occurs when llvm-strip compacts the file after removing debug
  sections but fails to maintain the required alignment constraint:
  (file_offset % pagesize) == (vaddr % pagesize)

  Pin to llvm-20 tools (llvm-strip-20, llvm-objcopy-20) which have the
  correct behavior, ensuring all stripped libraries maintain proper ELF
  segment alignment.

  Changes:
  - Update GitHub workflow to install llvm-20 package
  - Pin debian packaging to use llvm-strip-20 and llvm-objcopy-20
  - Pin RPM packaging to use llvm-strip-20 and llvm-objcopy-20

### DIFF
--- a/build_tools/packaging/linux/README.txt
+++ b/build_tools/packaging/linux/README.txt
@@ -13,7 +13,7 @@ apt update
 apt install -y python3
 apt install -y python3-pip
 apt install -y debhelper
-apt install -y llvm
+apt install -y llvm-20
 pip install -r requirements.txt
 
 #Usage:
--- a/build_tools/packaging/linux/package.json
+++ b/build_tools/packaging/linux/package.json
@@ -1254,8 +1254,7 @@
     ],
 
     "Gfxarch": "False",
-    "Disable_DWZ": "True",
-    "Disable_DEB_STRIP": "True"
+    "Disable_DWZ": "True"
   },
   {
     "Package": "amdrocm-hipify",


astrelsky:
Is there an open issue in the llvm project for this? I'm just curious as I've seen some wacky things with sections, but I've never had a problem with segments not being properly aligned (when doing other things, not rocm).

Edit: I just realized it is being pinned to a newer version, not an older one.

raramakr:
> Is there an open issue in the llvm project for this? I'm just curious as I've seen some wacky things with sections, but I've never had a problem with segments not being properly aligned (when doing other things, not rocm).

https://github.com/ROCm/TheRock/issues/3950

kcossett-amd:
Assuming this fix works, you should be able to revert the workaround change from #3932, being the `"Disable_DEB_STRIP": "True"` (https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/linux/package.json#L1258).

raramakr:
> Assuming this fix works, you should be able to revert the workaround change from #3932, being the `"Disable_DEB_STRIP": "True"` (https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/linux/package.json#L1258).

Updated the same

raramakr:
> Works locally on my system.
> 
> I do wonder why `llvm-strip` was never seemingly called on the `rpm` package, but this should prevent the same failure from occurring.

Debuginfo package generation is disabled for rpm package. Will be working on it soon

PR #3981, opened 2026-03-15
Correct passing of amdgpu_family to get_tests for skipping pytorch tests.

astrelsky:
## Motivation

Filtering of pytorch tests per gpu family was not working correctly due to how the amdgpu_family was being passed.

## Technical Details

It was previously iterating over the string characters to find the filter instead of filtering by the family.

## Test Plan

Testing locally and via ci when pytorch tests are run.

## Test Result

Ran tests locally and the filter I added for my gpu (gfx1150) was applied correctly. This pull request does not contain the added filter. The process finishes with `= 693 failed, 2112 passed, 1101 skipped, 314 deselected, 6 xfailed, 37448 errors in 508.91s (0:08:28) =` vs crashing due to failing to filter problematic tests.

## Submission Checklist

### DIFF
--- a/external-builds/pytorch/run_pytorch_tests.py
+++ b/external-builds/pytorch/run_pytorch_tests.py
@@ -225,6 +225,9 @@ def main() -> int:
         )
         print(f"Using AMDGPU family: {first_arch}")
 
+        # get_tests amdgpu_family requires list[str]
+        first_arch = [first_arch]
+
         # Determine PyTorch version
         pytorch_version = args.pytorch_version
         if not pytorch_version:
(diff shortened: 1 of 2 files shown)


astrelsky:
> thanks for catching this!
> 
> i guess it was half working/hiding the problem and matching all gfx filters ("g" is in "gfx942" of skip_test dict) until #3704 where I corrected and inverted the way how filters are being selected. ("gfx94" of skip_test dict is in "g")
> 
> i was wondering if putting it here would be better:
> ```
>  tests_to_skip = get_tests(
>             amdgpu_family=[first_arch],
> ```
> but maybe also not a bad idea to keep it as verbose as it is 
> 
> let me know in case you cant merge it yourself.

I'm an external contributor so I am not able to merge.

I don't think it was applying the skip_test filters at all for any GPU specific skip tests, that's how I found this. It would try to use each character as the key when looking up the skip tests. So tbl['g'] would match nothing, tbl['f'] would match nothing, etc.

PR #3986, opened 2026-03-16
Consolidate extended test flags into a single run_extended_tests boolean

### LINKED ISSUE #3901: Consolidate extended test flags into a single run_extended_tests object in workflow inputs
## Problem

Extended test gating is currently spread across two separate inputs with different sources and scopes:

- `run_functional_tests` (boolean) — a global flag generated by `configure_ci.py`, threaded through `setup.yml` -> `ci_nightly.yml` -> `ci_linux.yml` / `ci_windows.yml` -> `test_artifacts.yml`
- `benchmark_runs_on` (string) — a per-architecture runner label defined in `amdgpu_family_matrix.py`, embedded in the variant matrix and passed via `ci_nightly.yml` -> `ci_linux.yml` / `ci_windows.yml` -> `test_benchmarks.yml`

There is no single place to globally enable/disable benchmarks — benchmark gating is implicit (runner label is non-empty). Adding `run_functional_tests` as a standalone workflow_call input in every workflow adds more inputs.

## Proposal

Introduce a single `run_extended_tests` JSON object as a global flag in `configure_ci.py`, containing enablement booleans for both extended test types:

```yaml
# Global object from configure_ci.py (passed through the workflow chain)
run_extended_tests: '{"functional": true, "benchmarks": true}'
```

Per-architecture `benchmark-runs-on` runner label stays in the variant matrix (in `amdgpu_family_matrix.py`) since it varies per GPU family. The global `benchmarks` flag controls whether benchmarks run at all.

Benchmark gating in `ci_linux.yml` / `ci_windows.yml` would check both:

```yaml
# Benchmarks run only if globally enabled AND a runner is available for this variant
if: >-
  ${{
    fromJSON(inputs.run_extended_tests).benchmarks == true &&
    matrix.variant.benchmark-runs-on != ''
  }}
```

lajagapp:
Closes #3901

## Summary

Consolidates `run_functional_tests` and `benchmark_runs_on` into a single `run_extended_tests` boolean flag, and merges benchmark test execution into `test_artifacts.yml`, eliminating the separate `test_benchmarks.yml` workflow.

## Problem

Extended test gating was spread across two separate inputs with different sources and scopes:

- **`run_functional_tests`** (boolean) - threaded through `setup.yml -> ci_nightly.yml -> ci_linux/windows.yml -> test_artifacts.yml`
- **`benchmark_runs_on`** (string) - per-architecture runner label passed via `ci_nightly.yml -> ci_linux/windows.yml -> test_benchmarks.yml`

There was no single place to globally enable/disable extended tests, and benchmark gating was implicit (runner label is non-empty and it runs always for workflow_dispatch).

## Solution

- **Single flag:** `run_extended_tests` (boolean string `'true'`/`'false'`) generated by `configure_ci.py`, passed through the workflow chain.
- **Pre-resolved test_modes in Python:** `configure_ci.py` builds a `test_modes` matrix per variant containing pre-resolved runner labels and test types. Benchmark mode is only included when `run_extended_tests` is true and `build_variant` is `release`. Regular PR builds get a single-entry matrix with only "Test Artifacts".
- **Merged workflows:** Benchmarks run inside `test_artifacts.yml` via the `test_modes` strategy matrix alongside component tests.
- **Deleted `test_benchmarks.yml`:** No longer needed as a separate workflow.

**Docs:**
- `tests/extended_tests/README.md` - update architecture diagram and key differences
- `tests/extended_tests/benchmark/README.md` - remove duplicated sections, add prerequisites link


lajagapp:
Started trial run with custom branch to enable all tests in is_workflow_dispatch trigger (only for testing purpose).
https://github.com/ROCm/TheRock/actions/runs/23208282716/job/67449839907

geomin12 (commented):
going to let other folks review this!

.github/workflows/test_artifacts.yml
```diff
@@ -127,15 +126,20 @@ jobs:
       matrix:
         components: ${{ fromJSON(needs.configure_test_matrix.outputs.components) }}
     uses: './.github/workflows/test_component.yml'
+    secrets: inherit
```
geomin12, line 152: can remove, secrets are not applicable as secrets do not work for forked repos

.github/workflows/test_artifacts.yml
```diff
@@ -127,15 +126,20 @@ jobs:
...
       artifact_run_id: ${{ inputs.artifact_run_id }}
       artifact_group: ${{ inputs.artifact_group }}
       amdgpu_families: ${{ inputs.amdgpu_families }}
       amdgpu_targets: ${{ inputs.amdgpu_targets }}
-      # If the test requires a multi GPU, use the multi GPU test runner
-      # Otherwise, use the standard test runner
+      # Runner selection priority:
+      # 1. Benchmark runner (for benchmark tests on dedicated performance runners)
+      # 2. Multi-GPU runner (for multi-GPU tests like RCCL)
+      # 3. Standard test runner (default)
```
geomin12: pretty sure the `configure_ci.py` logic handles this, where benchmark runner becomes `test_runs_on`

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -480,28 +480,47 @@ def run():
...
+                )
+                for key, value in benchmark_matrix.items():
+                    merged = value.copy()
+                    merged["benchmark_runner"] = benchmark_runner
+                    selected_matrix[key] = merged
+            else:
+                logging.info(
+                    f"Skipping benchmarks: no benchmark runner for "
+                    f"{amdgpu_families} on {platform}"
+                )
```
geomin12: this can be handled in `configure_ci.py` (and i believe already is). this file is specifically related to test selection, not runner selection

geomin12 (commented):
thanks for clean up, i have some comments

build_tools/github_actions/configure_ci.py
ScottTodd, line 1: Note that I have a PR up forking configure_ci.py and setup.yml for multi-arch CI (https://github.com/ROCm/TheRock/pull/4123), and we plan on deleting the non-multi-arch CI (ci_linux.yml, ci_windows.yml, etc.) imminently (https://github.com/ROCm/TheRock/issues/3340)

Code will need to migrate to multi-arch CI if you want to keep it.

ScottTodd: We just switched the default pipeline for pull requests from ci.yml to multi_arch_ci.yml today with https://github.com/ROCm/TheRock/pull/4204, fyi.

lajagapp: Thanks for the heads-up on #4204 and #3340.

I see nightly/scheduled pipeline (ci_nightly.yml) still uses the non-multi-arch files (setup.yml, configure_ci.py, ci_linux.yml/ci_windows.yml). Since run_extended_tests is primarily triggered by scheduled builds, non-multi-arch file changes are still needed until nightly is migrated.

For the multi-arch side: configure_multi_arch_ci.py already has full schedule event support, but there's no multi-arch nightly workflow to add run_extended_tests support now.

Can I go ahead and add run_extended_tests support to multi-arch files (configure_multi_arch_ci.py, setup_multi_arch.yml) in this PR, or wait and do that as follow-up PR once nightly workflow migrated to multi-arch?

ScottTodd: I'd suggest sending a separate PR to add "extended test" support to the multi-arch CI workflows/scripts. Perhaps have them conditionally enabled based on a workflow input and allow PRs to opt-in via a label. I wouldn't wait "for nightly workflows to be migrated", as then there would be a period where this coverage would be entirely absent.

.github/workflows/test_artifacts.yml
lajagapp: secrets: inherit is needed here so nightly/dispatch runs on main repo can pass BENCHMARK_DB_URL / BENCHMARK_DB_FALLBACK_URL to test_component.yml for benchmark results submission.
And yes fork PRs won't have these secrets, but the benchmark scripts already handle that gracefully by skipping submission when they're empty.

lajagapp: We have used secrets in [test_benchmark.yml](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_benchmarks.yml#L83) for the same purpose.

.github/workflows/test_artifacts.yml
lajagapp: Thanks for the feedback. Now I’ve moved benchmark runner resolution to configure_ci.py, and it pre-resolves benchmark-runs-on (blanks it when extended tests are disabled).
ci_linux.yml/ci_windows.yml now maps benchmark_runs_on -> test_runs_on directly, no run_extended_tests re-check is needed.

build_tools/github_actions/fetch_test_configurations.py
lajagapp: I've removed all runner lookup logic from fetch_test_configurations.py. It now only does test selection:
test_type=benchmark -> benchmark_matrix
run_extended_tests -> merge functional_matrix

.github/workflows/test_artifacts.yml
lajagapp: Sample run with this change - https://github.com/ROCm/TheRock/actions/runs/23932295257

build_tools/github_actions/configure_ci.py
lajagapp: Sure, I will create separate PR to add "extended test" support to multi-arch CI scripts/workflows and update it here.

.github/workflows/ci_linux.yml
```diff
@@ -106,42 +106,31 @@ jobs:
...
         inputs.expect_failure == false
       }}
+    strategy:
+      fail-fast: false
+      matrix:
+        test_mode:
+          - name: "Test Artifacts"
+            test_type_override: ""
+          - name: "Test Benchmarks"
+            test_type_override: "benchmark"
```
geomin12: this is not right, this will run both test artifacts and test benchmarks

.github/workflows/ci_linux.yml
```diff
@@ -106,42 +106,31 @@ jobs:
...
       artifact_group: ${{ inputs.artifact_group }}
       amdgpu_families: ${{ inputs.amdgpu_families }}
-      test_runs_on: ${{ inputs.test_runs_on }}
+      # For regular tests: use test_runs_on. For benchmarks: use benchmark_runs_on.
+      test_runs_on: >-
+        ${{
+          (matrix.test_mode.test_type_override != 'benchmark' && inputs.test_runs_on) ||
+          (matrix.test_mode.test_type_override == 'benchmark' && inputs.benchmark_runs_on) ||
+          ''
+        }}
```
geomin12: can this come from `configure_ci.py`? we should try to minimize logic as much as possible in workflow files

.github/workflows/ci_windows.yml
```diff
@@ -106,42 +106,31 @@ jobs:
...
         inputs.expect_failure == false
       }}
+    strategy:
+      fail-fast: false
+      matrix:
+        test_mode:
+          - name: "Test Artifacts"
+            test_type_override: ""
+          - name: "Test Benchmarks"
+            test_type_override: "benchmark"
```
geomin12: same here, this is running both workflows

.github/workflows/ci_windows.yml
```diff
@@ -106,42 +106,31 @@ jobs:
...
       amdgpu_families: ${{ inputs.amdgpu_families }}
-      test_runs_on: ${{ inputs.test_runs_on }}
+      # For regular tests: use test_runs_on. For benchmarks: use benchmark_runs_on.
+      test_runs_on: >-
+        ${{
+          (matrix.test_mode.test_type_override != 'benchmark' && inputs.test_runs_on) ||
+          (matrix.test_mode.test_type_override == 'benchmark' && inputs.benchmark_runs_on) ||
+          ''
+        }}
       artifact_run_id: ${{ inputs.artifact_run_id }}
```
geomin12: logic can be done in configure_ci.py

.github/workflows/test_artifacts.yml
```diff
@@ -61,7 +61,7 @@ permissions:
 
 jobs:
   configure_test_matrix:
-    name: "Configure test matrix (${{ inputs.amdgpu_families }})"
+    name: "Configure ${{ inputs.test_type == 'benchmark' && 'benchmark' || 'test' }} matrix (${{ inputs.amdgpu_families }})"
```
geomin12: i think we can keep original

.github/workflows/test_artifacts.yml
```diff
@@ -100,13 +100,14 @@ jobs:
...
-          # TODO(#3901): Consolidate run_functional_tests and benchmark_runs_on into a single run_extended_tests input.
-          RUN_FUNCTIONAL_TESTS: ${{ inputs.run_functional_tests }}
+          RUN_EXTENDED_TESTS: ${{ inputs.run_extended_tests }}
         run: python ./build_tools/github_actions/fetch_test_configurations.py
 
   test_sanity_check:
     name: 'Test Sanity Check'
     needs: configure_test_matrix
+    # Skip sanity when the matrix has no sanity entry (e.g. benchmark-only runs).
+    if: ${{ needs.configure_test_matrix.outputs.sanity_component != 'null' }}
```
geomin12: why? we should still test sanity and make sure machines work fine during benchmark tests

build_tools/github_actions/configure_ci.py
```diff
@@ -599,6 +599,14 @@ def main(base_args, linux_families, windows_families):
...
 
     print(f"test_type decision: '{test_type}' (reason: {test_type_reason})")
+    print(f"run_extended_tests: {run_extended_tests}")
+
+    # Pre-resolve benchmark runner: blank it out when extended tests are
+    # disabled so downstream YAML doesn't need to re-check run_extended_tests.
+    if not run_extended_tests:
+        for row in linux_variants_output + windows_variants_output:
+            if "benchmark-runs-on" in row:
+                row["benchmark-runs-on"] = ""
```
geomin12: eh i don't think this is needed, we pretty much know when benchmarks will run (nightly)

.github/workflows/test_artifacts.yml
lajagapp: Yes, reverted back to original name.

.github/workflows/test_artifacts.yml
lajagapp: It was existing behavior with benchmark job.
Now, I've added Sanity entry into benchmark_matrix and removed sanity skip condition. Sanity test will always run.

.github/workflows/ci_linux.yml
lajagapp: Fixed now. Created test_modes matrix in configure_ci.py which is gated by run_extended_tests.

For PR or other builds, test_modes only contains "Test Artifacts", benchmark job is skipped entirely.
For nightly release builds and workflow_dispatch with test labels, both entries are present so they run in parallel on their respective runners.

.github/workflows/ci_linux.yml
lajagapp: Yes, moved these conditions for runner selection and test_type to `configure_ci.py`.

Each test_modes entry contains final test_runs_on and test_type values. And workflow reads them as `matrix.test_mode.test_runs_on` and `matrix.test_mode.test_type`.

.github/workflows/ci_windows.yml
lajagapp: Same as linux. ci_windows.yml uses the same pre-resolved test_modes from `configure_ci.py`.

.github/workflows/ci_windows.yml
lajagapp: Yes, moved these conditions to configure_ci.py.

build_tools/github_actions/configure_ci.py
lajagapp: Yes got it. Now blanking logic is removed entirely. Benchmark inclusion is now gated by `run_extended_tests`. When this is False, the benchmark entry simply isn't added to test_modes.

.github/workflows/ci_windows.yml
```diff
@@ -87,42 +84,21 @@ jobs:
         ) &&
         inputs.expect_failure == false
       }}
+    strategy:
+      fail-fast: false
+      matrix:
+        test_mode: ${{ fromJSON(inputs.test_modes) }}
```
geomin12: i feel this may not be the best approach. why not just append benchmark jobs in addition to the current set of jobs if "extended_tests" is added? (for nightly)?

geomin12: actually from `fetch_test_configurations.py`, this is done. this means this can be removed (along with the matrix effort)

build_tools/github_actions/configure_ci.py
```diff
@@ -619,6 +619,28 @@ def main(base_args, linux_families, windows_families):
...
+            },
+        ]
+        if include_benchmarks:
+            row["test_modes"].append(
+                {
+                    "name": "Test Benchmarks",
+                    "test_type": "benchmark",
+                    "test_runs_on": row.get("benchmark-runs-on", ""),
+                }
+            )
```
geomin12: this isn't needed. we can just pass `extended_tests` all the way to `test_artifacts.yml`, then we can determine to run benchmarks then

geomin12 (commented):
as mentioned, we can do this logic in `fetch_test_configurations.py` instead having to pass additional matrix data from `configure_ci.py`

.github/workflows/ci_windows.yml
lajagapp: Yes, similar changes are done in first commit of this PR.
https://github.com/ROCm/TheRock/pull/3986/changes/5167ddcd1f3de01658b20b1a4c5afbfcd1b769a7

build_tools/github_actions/configure_ci.py
lajagapp: Yes, updated the PR with requested changes.
Updated fetch_test_configurations.py to define which runner to use, and test_artifacts.yml will run each test based on the selected runner.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -629,7 +634,9 @@ def run():
                 continue
 
             job_config_data = selected_matrix[key]
-            job_config_data["test_type"] = test_type
+            job_config_data["test_type"] = (
+                "full" if job_config_data.get("is_benchmark") else test_type
```
geomin12: does benchmarks have a `test_type` logic? i don't think this is applicable

.github/workflows/test_artifacts.yml
```diff
@@ -128,22 +133,37 @@ jobs:
...
+      #   2. Multi-GPU components use the multi-GPU runner
+      #   3. Everything else uses the standard test runner
       test_runs_on: >-
         ${{
+          matrix.components.is_benchmark == true &&
+            inputs.benchmark_runs_on != '' &&
+            inputs.benchmark_runs_on ||
           matrix.components.multi_gpu_runner != '' &&
             matrix.components.multi_gpu_runner ||
             inputs.test_runs_on
```
geomin12: can we add `()` to help read this?

geomin12 (commented):
two comments, but looking good, thanks for working on comments. can we run a test to make this operates as expected?

.github/workflows/test_artifacts.yml
lajagapp: Done. Added parentheses to group each runner selection clause.

build_tools/github_actions/fetch_test_configurations.py
lajagapp: Yes correct, benchmarks don't have test_type. They always run the same way regardless of test_type. Removed the conditional now.

lajagapp:
> two comments, but looking good, thanks for working on comments. can we run a test to make this operates as expected?

Yes, started nightly-ci run with this PR (_used custom branch with this PR + run_extended_tests enabled for is_workflow_dispatch_) - https://github.com/ROCm/TheRock/actions/runs/24814444762/job/72626935251

geomin12 (approved):
thanks for updates! let's make these changes in future PR for multi_arch_ci as this CI will be decomissioned soon

PR #3917, opened 2026-03-11
[ci] Expanding test filter options via PR label selection

geomin12:
In order to work with test standardization, we are expanding test filters.

Changes:
- PR labels for additional tests can be added (i.e. `test_filter:nightly`)
- `is_schedule` now runs nightly tests

Currently, each component has different filters, so this is a CI update only, and we will update component shards / test filters accordingly once filters are ready per component

Working as test_filter:nightly is properly displayed https://github.com/ROCm/TheRock/actions/runs/22973838135?pr=3917

PR #3992, opened 2026-03-16
[ci] Expanding test filter options via PR label selection

geomin12:
In order to work with test standardization, we are expanding test filters.

Changes:

PR labels for additional tests can be added (i.e. test_filter:nightly)
is_schedule now runs nightly tests
Currently, each component has different filters, so this is a CI update only, and we will update component shards / test filters accordingly once filters are ready per component

Working as test_filter:nightly is properly displayed https://github.com/ROCm/TheRock/actions/runs/22973838135?pr=3917

Fixed post-submit CI issue from #3917 and added unit tests to cover this scenario


geomin12:
Landing as case from #3917 is fixed

PR #3993, opened 2026-03-16
Add native Linux package building to CI without S3 upload

raramakr:
Integrate native package building into the CI workflow by calling
  build_native_linux_packages.yml from ci_linux.yml. Packages are built
  for both deb and rpm formats but not uploaded to S3.

  Changes:
  - Add upload_packages input to build_native_linux_packages.yml (defaults to true for backward compatibility)
  - Make AWS CLI install, credentials config, and S3 upload steps conditional on upload_packages flag
  - Call build_native_linux_packages.yml from ci_linux.yml with upload_packages: false to build packages during CI tests
  - Build both deb and rpm packages via matrix strategy
  - Use rocm_package_version and build_variant_suffix from inputs

  This enables package validation during CI while keeping S3 uploads
  limited to native packaging workflows triggered individually

### DIFF
--- a/.github/workflows/build_native_linux_packages.yml
+++ b/.github/workflows/build_native_linux_packages.yml
@@ -85,7 +85,7 @@ run-name: Build native Linux packages (${{ inputs.artifact_group }}, ${{ inputs.
 
 jobs:
   build_native_packages:
-    name: Build Linux native Packages::${{ inputs.native_package_type }}
+    name: Build Linux native Packages::${{ inputs.native_package_type }}::${{ inputs.artifact_group }}
     strategy:
       fail-fast: false
     runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}


.github/workflows/build_native_linux_packages.yml
```diff
@@ -136,15 +145,18 @@ jobs:
             --version-suffix ${{ env.ARTIFACT_RUN_ID }}
 
       - name: Install AWS CLI
+        if: ${{ !inputs.skip_package_upload }}
```
nunnikri: This step is getting removed. Better to avoid this check

.github/workflows/ci_linux.yml
```diff
@@ -218,3 +218,37 @@ jobs:
...
+    # Run if:
+    # - Build succeeded (or using prebuilt artifacts)
+    # - Not expecting failure
+    if: >-
+      ${{
+        !failure() &&
+        !cancelled() &&
+        (
+          inputs.use_prebuilt_artifacts == 'false' ||
+          inputs.use_prebuilt_artifacts == 'true'
```
nunnikri, line 235: Looks like we are passing if prebuilt is both  true and false. Is this check really needed?

geomin12: this is a hacky github scenario where without this, the "skip build to test" will not work. this is needed

.github/workflows/build_native_linux_packages.yml
geomin12: this can stay! if no package upload, we don't need to waste time installing AWS CLI

.github/workflows/ci_linux.yml
```diff
@@ -218,3 +218,37 @@ jobs:
...
+        ) &&
+        inputs.expect_failure == false
+      }}
+    strategy:
+      fail-fast: false
+      matrix:
+        package_type: [deb, rpm]
+    uses: ./.github/workflows/build_native_linux_packages.yml
+    secrets: inherit
```
geomin12: is this needed? secrets don't work for forks, so we can remove if not needed

.github/workflows/build_native_linux_packages.yml
raramakr: https://github.com/ROCm/TheRock/pull/4018 , with this aws install is removed

.github/workflows/ci_linux.yml
raramakr: Remove secrets:inherit

.github/workflows/ci_linux.yml
```diff
@@ -218,3 +218,36 @@ jobs:
...
+          inputs.use_prebuilt_artifacts == 'false' ||
+          inputs.use_prebuilt_artifacts == 'true'
+        ) &&
+        inputs.expect_failure == false
+      }}
+    strategy:
+      fail-fast: false
+      matrix:
+        package_type: [deb, rpm]
+    uses: ./.github/workflows/build_native_linux_packages.yml
```
geomin12, line 243: for build_native_linux_package, can we update it to include (deb) or (rpm)?

In the build: https://github.com/ROCm/TheRock/actions/runs/23309328171/job/67807519670?pr=3993, there is no difference between the jobs (so it's hard to tell what is what)

raramakr: Done

raramakr:
> oh, can we also add the gfx architecture to the job name? I realize it is building for different archs

Done

.github/workflows/ci_linux.yml
ScottTodd, line 1: This workflow is deprecated. Please no new feature work here. Needs to go in https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci_linux.yml instead/also

.github/workflows/build_native_linux_packages.yml
ScottTodd, line 159: We should upload to S3 on CI too.

We're planning on reworking the release (CD) workflows to be based on the workflows currently used for CI:
1. Always upload to therock-ci-artifacts (or therock-ci-artifacts-external) bucket
2. Release workflows _copy_ those artifacts to a release bucket after testing them

.github/workflows/ci_linux.yml
```diff
@@ -218,3 +218,36 @@ jobs:
     permissions:
       contents: read
       id-token: write
+
+  build_native_linux_packages:
+    needs: [build_portable_linux_artifacts]
```
ScottTodd, line 223: This can be moved up in the workflow. Sort by dependency ordering, ROCm core first, then frameworks:

1. build rocm artifacts
2. test rocm artifacts
3. build rocm native linux packages
4. test rocm native linux packages
5. build rocm python packages
6. build pytorch python packages
7. test pytorch python packages
8. build jax python packages
9. test jax python packages

nunnikri: Thanks @ScottTodd  for the feedback. This is exactly the plan for native packaging. 
The test PR https://github.com/ROCm/TheRock/pull/3126 and multi arch PR https://github.com/ROCm/TheRock/pull/3561 and coming to shape. Once this is fully ready, we will get the same flow as you suggested. This current PR is a stop gap mechanism till these matures and hence the upload was skipped.

.github/workflows/ci_linux.yml
ScottTodd: We just switched the default pipeline from ci.yml to multi_arch_ci.yml today with https://github.com/ROCm/TheRock/pull/4204. This workflow now only runs on postsubmit or opt-in on PRs with the `ci:run-non-multi-arch` label.

Docs: https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

PR #3995, opened 2026-03-16
[ci] disabling gfx90a

geomin12:
Due to lab maintenance, gfx90a machines are down from 3/17 - 3/18. Disabling to prevent delays.

I am working on better way of updating labels in a separate PR

Tests work here and properly ignored: https://github.com/ROCm/TheRock/actions/runs/23162961369

Since this is a test update only, we are adding `skip-ci` label.


marbre:
Can we be more explicit in the title next time? I was just start wondering why we disabled build.

Issue #3291, opened 2026-02-06
[CI] Build and test PyTorch Python packages as part of ci.yml


ScottTodd:
## Background

Similar to https://github.com/ROCm/TheRock/issues/1559, we've had issues like https://github.com/ROCm/TheRock/issues/3042 slip through since we don't build PyTorch as part of our CI, only opt-in via manually triggered dev releases or regularly scheduled nightly releases.

## Goals

- [x] Pull requests in TheRock should be able to opt-in to building PyTorch as part of the regular CI workflow
- [x] Start with latest stable PyTorch version (currently 2.9 or 2.10), for a single Python version (3.12 or 3.13), for Linux
- [x] Add Windows too
- [ ] Add the option to build for more versions (via configure_ci, github labels, etc.)
- [ ] After building packages, we should be able to run a subset of PyTorch tests
- [ ] Add the option to trigger these jobs for jobs in rocm-libraries and rocm-systems (see https://github.com/ROCm/TheRock/issues/3177 and ongoing work to de-duplicate the workflows in use across the different repos)

## Design

My full design plan and notes are at https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/pytorch-ci.md.

### Workflow structure

At a high level, I'm currently planning on forking [`build_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml) to a new `build_portable_linux_pytorch_wheels_ci.yml` that will reuse the same underlying scripts but contain CI-specific logic (uploading to the artifacts bucket, _not_ uploading to a release bucket).

I would ideally want to have each workflow only be responsible for one thing:

* Build workflows build artifacts/packages and upload them to an "artifacts" bucket
* Test workflows download artifacts/packages
* Release workflows copy artifacts/packages into a "release" bucket based on a policy (e.g. "tests passed")

That's currently difficult to implement in the existing release workflows due to how they use `strategy.matrix` and rely on plumbing the `torch_version` output of each build job to a test job. If we move to a commit manifest approach for our workflows (see https://github.com/ROCm/TheRock/issues/1236), we could avoid the need for that plumbing. We could also unroll the job matrix as PyTorch does for workflows like https://github.com/pytorch/pytorch/blob/main/.github/workflows/generated-linux-binary-manywheel-nightly.yml.

### Package uploading

[...]

PR #2795, opened 2026-01-06
[torch] Remove support for .patch files to PyTorch projects

ScottTodd:
## Motivation

Since https://github.com/ROCm/TheRock/pull/2419 (merged 1 month ago), we've had zero patches to PyTorch repositories here in TheRock, so this code has been unused. We now use the release branches in the https://github.com/ROCm/pytorch repository exclusively, and landing cherry-picks for issues like https://github.com/ROCm/pytorch/issues/2889 has been proceeding smoothly.

I think it's time to shut the door on .patch files for PyTorch, then we can eventually do the same thing for ROCm in general once the https://github.com/ROCm/TheRock/tree/main/patches folder is similarly empty.

### Discussion of counter-arguments

While there is the https://github.com/ROCm/pytorch repository, there are not equivalents for other projects and this patch structure allows us to change any submodule of pytorch too. We really need to lean on upstream, or at least _collaborative_ downstream development here though, using regular git features and not patch files managed through bespoke scripts separate from the project code.

## Technical Details

The checkout scripts are still useful since they handle reading from the "related commits" file to select a commit to checkout and they also run HIPIFY. The checkout process is:

1. Determine which commit to checkout (e.g. by reading the related commits file from the torch repo)
2. Checkout that commit and tag it
3. **(⚠️ removed here)** Apply "base" patches
4. Run HIPIFY and commit + create a tag
5. **(⚠️ removed here)** Apply "hipify" patches

## Test Plan

* Sanity checked checkout scripts locally
* Test workflow runs:

    Workflow | Test details | Run logs
    -- | -- | --
    `build_portable_linux_pytorch_wheels.yml` | Expected to checkout `release/2.9` from ROCm/pytorch then build successfully | https://github.com/ROCm/TheRock/actions/runs/20761710618/job/59618093185
    `build_windows_pytorch_wheels.yml` | Expected to checkout `release/2.9` from ROCm/pytorch then build successfully | https://github.com/ROCm/TheRock/actions/runs/20761686755/job/59618014098
    `test_pytorch_wheels.yml` | Showing that the error in the Windows run above is fixed by removing the `--no-patch` argument | https://github.com/ROCm/TheRock/actions/runs/20763733809/job/59624927054
    `release_windows_pytorch_wheels.yml` | Expected to populate a matrix with the expected Python and PyTorch versions (cancelled before actually running those builds) | https://github.com/ROCm/TheRock/actions/runs/20761830824
[...]

PR #4006, opened 2026-03-17
Implement triton-windows integration

m-gallus:
## Motivation
Adds Windows support for building Triton wheels as part of the PyTorch build process, enabling `torch.compile` functionality on Windows.

Currently, PyTorch on Windows lacks Triton integration, which limits access to PyTorch's compilation and optimization features. This PR integrates the community-maintained [triton-windows](https://github.com/triton-lang/triton-windows) fork, which provides native Windows support for Triton. AMD is actively contributing to triton-windows production readiness (see [triton-windows#2](https://github.com/triton-lang/triton-windows/issues/2)) to ensure a first-class PyTorch experience on Windows with ROCm.

  This integration enables TheRock CI to build PyTorch wheels with full Triton support on Windows, bringing feature parity with Linux builds.

## Technical Details
**Changes to `build_prod_wheels.py`:**
  - Added `download_llvm_for_triton_windows()` to download pre-built LLVM binaries from `oaitriton.blob.core.windows.net`
    - LLVM is downloaded to a sibling directory of triton (e.g., `C:\b\llvm-{hash}-windows-x64`)
    - Version is matched against triton-windows' `cmake/llvm-hash.txt`
    - Caching prevents re-downloading on subsequent builds
  - Added Windows-specific build path using `python -m build --wheel` (matches triton-windows build system)
  - Environment configuration sets LLVM paths, disables Proton, and prevents CMake from finding wrong LLVM
  - Wheel name handling for `triton_windows` vs `triton` package names

  **Changes to `pytorch_triton_repo.py`:**
  - Platform detection to use `triton-windows` repository on Windows, `ROCm/triton` on Linux
  - Commit pin management via `ci_commit_pins/triton-windows.txt`
  - Disabled hipify and patching for Windows (not needed for triton-windows)
  - Unified `_do_hipify` callback to write `build_env.json` for both platforms

  **Prerequisites:**
  - Windows builds must run in a Visual Studio Developer Command Prompt (or have vcvars64.bat already executed)

## Test Plan

```
cd C:\path_to\TheRock\external-builds\pytorch
python pytorch_torch_repo.py checkout --checkout-dir C:/b/pytorch
python pytorch_triton_repo.py checkout --checkout-dir C:/b/triton --torch-dir C:/b/pytorch
python build_prod_wheels.py build ^
    --triton-dir C:/b/triton ^
    --output-dir C:/wheels_test ^
    --build-triton ^
    --no-build-pytorch-audio ^
    --no-build-pytorch-vision ^
    --no-build-apex ^
    --version-suffix "+test" ^
    --pytorch-rocm-arch gfx1100
```

## Test Result

```
(...)
Downloading LLVM for triton-windows...
    Hash: 278dba37
    URL: https://oaitriton.blob.core.windows.net/public/llvm-builds/llvm-278dba37-windows-x64.tar.gz
    Downloading (this may take a few minutes, ~500MB)...
    Extracting...
    LLVM downloaded to: C:\b\llvm-278dba37-windows-x64
  +++ Installing build dependencies:
  +++ Building triton:
(...)
  [474/474] Linking CXX shared module libtriton.cp312-win_amd64.pyd
  Successfully built triton_windows-3.6.0+git6fe895f6-cp312-cp312-win_amd64.whl
  Found built wheel: C:\b\triton\dist\triton_windows-3.6.0+git6fe895f6-cp312-cp312-win_amd64.whl
  ++ Copy C:\b\triton\dist\triton_windows-3.6.0+git6fe895f6-cp312-cp312-win_amd64.whl -> C:\wheels_test
  --- Builds all completed
```

## Submission Checklist

The contribution guide suggests implementing unit tests for each new implemented feature. Are Unit Tests for testing the triton build applicable here? I didn't find any triton unit tests for Linux.


external-builds/pytorch/build_prod_wheels.py
```diff
@@ -651,6 +718,62 @@ def do_build(args: argparse.Namespace):
 def do_build_triton(
     args: argparse.Namespace, triton_dir: Path, env: dict[str, str]
 ) -> str:
+    """Build triton wheel. Uses triton-windows on Windows, ROCm/triton on Linux."""
+    triton_wheel_name = env.get("TRITON_WHEEL_NAME", "triton")
```
ScottTodd: We have a support table here: https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#project-and-feature-support-status. I'd like to see triton added there:

* List as "Supported" on Linux
* List as "In progress" on Windows (can link to https://github.com/triton-lang/triton-windows/issues/2 or another issue)
* Add mentions of which branches/commits are used for different pytorch versions to the table below, like is done for ROCm/triton with the ci_commit_pins/triton.txt file

(That can be separate from this PR that does the initial build integration)

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -651,6 +718,62 @@ def do_build(args: argparse.Namespace):
...
+            env=windows_env,
+        )
+
+        # triton-windows produces wheels named "triton_windows" not "triton".
+        windows_wheel_name = "triton_windows"
+        try:
+            built_wheel = find_built_wheel(triton_dir / "dist", windows_wheel_name)
+        except RuntimeError:
+            built_wheel = find_built_wheel(triton_dir / "dist", triton_wheel_name)
+            windows_wheel_name = triton_wheel_name
```
ScottTodd: * Is this package name difference intentional? Should it eventually use the same name on both platforms? (I'm expecting that various build scripts, workflows, and code would be better behaved with a consistent name, but I can see an argument for using a different name if the code/ownership is substantially different)
* Why check both if we know the name should be `triton_windows`? If/when the name changes we can just change the code instead of having a branch here

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -651,6 +718,62 @@ def do_build(args: argparse.Namespace):
 def do_build_triton(
     args: argparse.Namespace, triton_dir: Path, env: dict[str, str]
 ) -> str:
+    """Build triton wheel. Uses triton-windows on Windows, ROCm/triton on Linux."""
+    triton_wheel_name = env.get("TRITON_WHEEL_NAME", "triton")
+
+    if is_windows:
+        print("Building Triton for Windows (using triton-windows repository)")
```
ScottTodd: Rather than:

```python
def do_build_triton:
  if is_windows:
    # build for windows
    return
  # build for linux
```

I think we should _either_

1. Put both Windows and Linux in branches within the existing function:
    ```python
    def do_build_triton:
      # common setup for both (computing versions, etc.)
      if not is_windows:
        # narrow set of Linux-specific build steps
      else:
        # narrow set of Windows-specific build steps
      # common finalize steps
    ```
  
2. Pull both out into functions:
    ```python
    
    def build_triton_windows:
      # build for windows
    
    def build_triton_linux:
      # build for linux
    
    def do_build_triton:
      # common setup for both (computing versions, etc.)
      if is_windows:
        build_triton_windows()
      else:
        build_triton_linux()
    ```

I'm leaning towards 2, given how much code is different. For both cases, I want the branching to be clear and keep both platforms explicitly separated, not having a "default" platform.

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -258,6 +263,68 @@ def get_version_suffix_for_installed_rocm_package() -> str:
...
+    return hash_file.read_text().strip()
+
+
+def download_llvm_for_triton_windows(triton_dir: Path) -> Path:
+    """Download and extract pre-built LLVM binaries for triton-windows.
+
+    triton-windows requires a specific LLVM version that matches the hash
+    in cmake/llvm-hash.txt. Pre-built binaries are hosted at oaitriton.blob.core.windows.net.
+    """
+    full_hash = get_triton_windows_llvm_hash(triton_dir)
```
ScottTodd, line 280: What is this prebuilt LLVM used for specifically? If it's just running `clang`, `clang-cl`, `lld`, etc. then I'd like to chart a path towards using LLVM from a more central upstream source or from our own amd-llvm build that we distribute as part of ROCm. If it's depending on LLVM internals (e.g. as MLIR compilers do), then we may want to build from source instead of using prebuilt here. We can add compiler caching and all the usual tricks.

Note also that our CI systems will not cache this `llvm_dir` so they'll at least spend the time downloading that blob for every build.

external-builds/pytorch/build_prod_wheels.py
ScottTodd, line 801: We should use `args.version_suffix` for triton-windows too. See the versioning docs we have here: https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md, specifically for packages like pytorch where we create a composite version that includes the rocm version: https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#external-python-package-versions. That allows us to upload multiple nightly/prerelease/stable releases for the same base (e.g. pytorch/triton) version and different rocm versions. See https://rocm.nightlies.amd.com/v2/gfx1151/triton/ for example.

(This could be a follow-up change after landing the initial build support, but we'd need it resolved to enable nightly and later stable releases)

external-builds/pytorch/ci_commit_pins/triton-windows.txt
```diff
@@ -0,0 +1 @@
+6fe895f6b494b2ffc688e1e552dc2d4a8b5ae3b3
```
ScottTodd, line 1: This will limit us to a single triton commit pin for all pytorch versions. What pytorch versions is this compatible with? We can certainly start limited to a single version during bring-up work, but long term I'd like to see this commit pin file move into some upstream pinned commit files. See the table at https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#supported-pytorch-versions for what we do on Linux:

* pytorch/pytorch upstream latest nightly: https://github.com/pytorch/pytorch/blob/nightly/.ci/docker/ci_commit_pins/triton.txt
  * Note that there are already related files like https://github.com/pytorch/pytorch/blob/nightly/.ci/docker/ci_commit_pins/triton-cpu.txt
* ROCm/pytorch fork long term support branch for `release/2.10`: https://github.com/ROCm/pytorch/blob/release/2.10/.ci/docker/ci_commit_pins/triton.txt
  * Note cherry-picks like https://github.com/ROCm/pytorch/commit/94de74feccf27f71977a976374af648029bca207 are possible

How about we add `triton-windows.txt` upstream? Maybe once we've proved it out here in TheRock?

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -258,6 +263,68 @@ def get_version_suffix_for_installed_rocm_package() -> str:
...
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to download LLVM from {download_url}: {e}\n"
+                "You may need to download manually and extract to "
+                f"{llvm_dir}"
+            )
+
+        print("  Extracting...")
+        with tarfile.open(download_path, "r:gz") as tar:
+            tar.extractall(triton_dir.parent)
```
ScottTodd: (spotted by my code review agents)

This https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall call is a security risk - tar archives can contain `..` components that this will extract by default. Python 3.12 added a `filter="data"` parameter to make this safer. In Python 3.14 that filter is the default.

external-builds/pytorch/pytorch_triton_repo.py
```diff
@@ -108,9 +160,15 @@ def add_common(command_parser: argparse.ArgumentParser):
...
-        default=True,
+        default=default_hipify,
         help="Run hipify",
     )
+    checkout_p.add_argument(
+        "--patch",
+        action=argparse.BooleanOptionalAction,
+        default=default_patch,
+        help="Apply patches for the repo-hashtag",
+    )
```
ScottTodd: Let's remove this unused argparse option (see https://github.com/ROCm/TheRock/pull/2795 which removed support for `.patch` files and `--patch` args, we'd really like to avoid reintroducing those)

external-builds/pytorch/build_prod_wheels.py
ScottTodd, line 1: > This integration enables TheRock CI to build PyTorch wheels with full Triton support on Windows, bringing feature parity with Linux builds.

Yep, this is a great start to getting triton building. This will need a few extra steps to fully integrate into our nightly release workflows. A search across the repository for "triton" should spot most areas that need updating. These come to mind for me:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_windows_pytorch_wheels.yml add checkout steps, build step, manifest file version stamping, wheel uploading/copying (can see the diff between linux and windows workflows to know what to change)
  * See https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#testing-pytorch-release-workflows for how to test these workflows
  * https://github.com/ROCm/TheRock/blob/3e6ced8508f372f11520ddbdc3e4423a9ff19420/build_tools/github_actions/write_torch_versions.py#L83-L88
  * https://github.com/ROCm/TheRock/blob/3e6ced8508f372f11520ddbdc3e4423a9ff19420/build_tools/github_actions/generate_pytorch_manifest.py#L191-L199
* Any mentions of "triton" at https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-pytorch-python-packages
* "triton_requirement" code in this `build_prod_wheels.py` file: we add triton as an install requirement to our Linux torch packages (upstream packages do NOT do this), so just installing torch automatically pulls triton too. We could do that if we want, or at least document why we do it for Linux but NOT for Windows

Other nice-to-haves:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml currently only tests pytorch (even on Linux). If there are triton-specific tests, either in pytorch's unit tests or elsewhere, we could run those
* The workflow linked above is for nightly releases. We also have a similar workflow that runs on every commit: https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_windows_pytorch_wheels_ci.yml . Linux and Windows both currently only build torch there, not torchaudio/torchvision/triton/apex/etc. but we could expand that (see https://github.com/ROCm/TheRock/issues/3291)

external-builds/pytorch/build_prod_wheels.py
m-gallus: - The package name has been set to be this way by the `triton-windows` fork maintainer. This is also the name under which the package is distributed at [PyPI](https://pypi.org/project/triton-windows/). However when building, it's name can be altered by setting the `TRITON_WHEEL_NAME` environmental variable. We can use it to our advantage to rename it to `triton`.
- I agree we should rather strive for having the same name for both Windows and Linux, exactly for the reasons you've described. I left the fall back in case we wanted to switch to the `triton` name by using the aforementioned env var, but I can remove it.

external-builds/pytorch/build_prod_wheels.py
m-gallus: Great, thank you for taking your time to find those.  Will make note of these for the next PR.

external-builds/pytorch/ci_commit_pins/triton-windows.txt
m-gallus: This commit in `triton-windows.txt` is currently top of develop and is meant to be compatible with the latest PyTorch mainline.

Good thing is, that the `triton-lang/triton-windows` hosts [several branches/tags](https://github.com/triton-lang/triton-windows/branches/all?query=release%2F) corresponding to triton releases which support their respective PyTorch versions:
PyTorch | Triton
-- | --
2.6 | 3.2
2.7 | 3.3
2.8 | 3.4
2.9 | 3.5
2.10 | 3.6

If we wanted to, we could add support for building from those branches/tags, depending on the selected PyTorch version.

As for the upstreaming of the pin - yes, this is the plan, which could make matching the pins much easier. We plan to submit such change in the near future.

However, unfortunately, even if we wanted to, it seems that for the older PyTorch releases, we no longer have such option. If we happen to consider adding triton-windows support for the older versions of PyTorch, we'd need to maintain a local list of pins.

external-builds/pytorch/build_prod_wheels.py
m-gallus: It is used for LLVM internals (MLIR). As for running `clang-cl` and etc. in fact, we already use this path for the runtime part of triton and we grab it directly from the rocm-sdk package. 

We can sure build from source if this is cheaper/faster. Does the caching system only work for binaries built on the CI, or can we trick it to cache something that has been downloaded?

To give more context, this is how often the llvm hash changes (and hence how often it'd need to be rebuilt/downloaded):
https://github.com/triton-lang/triton-windows/commits/main-windows/cmake/llvm-hash.txt

external-builds/pytorch/build_prod_wheels.py
m-gallus: Nice one, I wasn't aware of it. Will update.

external-builds/pytorch/build_prod_wheels.py
m-gallus: Sure thing, will post it in a follow-up. Thank you for pointing me to those.

external-builds/pytorch/build_prod_wheels.py
m-gallus: Totally agreed. As for the approach - yeah, I was trying to find a common path, but the commonalities amounted to maybe 20% of the scattered code. I'm hence leaning towards 2 as well.

external-builds/pytorch/build_prod_wheels.py
m-gallus: Oh, that's perfect. Thanks, I couldn't initially find where this was described. I'll adjust this either here, or later in the follow-up.

external-builds/pytorch/pytorch_triton_repo.py
m-gallus: Alrighty, I wasn't sure if it's still needed for the linux version. Will gladly remove.

external-builds/pytorch/ci_commit_pins/triton-windows.txt
ScottTodd: > However, unfortunately, even if we wanted to, it seems that for the older PyTorch releases, we no longer have such option. If we happen to consider adding triton-windows support for the older versions of PyTorch, we'd need to maintain a local list of pins.

That's why we maintain the https://github.com/ROCm/pytorch fork. We have full control there, and we do regularly update pinned versions for different dependencies on the "extended support" release branches.

m-gallus:
Updated the PR with:
  - Added `filter="data"` to tarfile extraction
  - Removed deprecated `--patch` argument
  - Split `do_build_triton()` into platform-specific functions
  - Both platforms now produce `triton` wheels (via `TRITON_WHEEL_NAME` env var)
  - Added triton to support table and version matrix

external-builds/pytorch/build_prod_wheels.py
ScottTodd: We don't have a general purpose cache that will just work out of the box (oh how I sometimes miss Bazel...).

A few complementary options:
* We could mirror the source or binaries to our `rocm-third-party-deps` S3 bucket, as we've done for other dependencies. That gives us better control over the supply chain, more efficient downloads from build machines in the same network / cloud provider region, etc.: https://github.com/ROCm/TheRock/blob/7ba1cea5dd65492a84ef102b94a050caccf2e3a4/third-party/eigen/CMakeLists.txt#L4-L9
* We could use ccache or sccache as part of building triton and/or LLVM here. See other code in this file for using those tools for the pytorch build

I'm not too concerned about build time at the moment - we have 4-8 hour builds for ROCm so another 30m-1h for a leaf component like this is unfortunate but not a bottleneck. The supply chain security aspect is worth tracking (but not blocking this initial PR) I think.

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -851,6 +847,16 @@ def do_build_triton(
...
 
 
+def do_build_triton(
+    args: argparse.Namespace, triton_dir: Path, env: dict[str, str]
+) -> str:
+    """Build triton wheel. Dispatches to platform-specific build functions."""
+    if is_windows:
+        return build_triton_windows(args, triton_dir)
+    else:
+        return build_triton_linux(args, triton_dir, env)
```
ScottTodd, line 857: Windows isn't passing through the `env`. That seems intentional... building with MSVC instead of `clang-cl`?

This is in `_setup_common_build_env`: https://github.com/ROCm/TheRock/blob/7ba1cea5dd65492a84ef102b94a050caccf2e3a4/external-builds/pytorch/build_prod_wheels.py#L416-L424

Note that the common env setup does other general useful things like configuring ccache or sccache: https://github.com/ROCm/TheRock/blob/7ba1cea5dd65492a84ef102b94a050caccf2e3a4/external-builds/pytorch/build_prod_wheels.py#L569-L581

m-gallus: Good catch, that was semi-intentional. It didn't use `env` because as you have guessed, it uses MSVC to build triton instead of clang-cl, so the env would overwrite the `CC/CXX` env vars. I wanted to keep the environment simple and avoid potential conflicts.
However, ccache support is something that we're certainly losing here with that approach. In the next PR I'll add it's support, and I'm considering two options:
a) Passing `env` in, but removing `CC/CXX` env vars
b) Selectively copying the `ccache` vars

I'm leaning towards a), but if you have any preference, please let me know.

external-builds/pytorch/build_prod_wheels.py
m-gallus: Understood, thanks for sharing more context.
For the mirroring option, I'd need your support to have it hosted (or from whoever that has access to the `rocm-third-party-deps`). This would require minimal code changes, unless you suggest that we download triton-designated llvm using FetchContent, similarly to how Eigen is downloaded. As for the (s)ccache, we can do that too, even for the sole purpose of speeding up triton builds. Whichever of two you see fit and perhaps we could even combine those later if needed ((s)ccache-assisted triton + fetched mirrored llvm).

ScottTodd: Yeah I can help upload to that s3 bucket (or @marbre / other devops team members could).

external-builds/pytorch/build_prod_wheels.py
ScottTodd: We could also split `_setup_common_build_env` into a few smaller functions and choose what we want from it in the build functions, instead of passing a single env through each function 🤔

```python
# triton build
env = _setup_common_build_env()
env = env | _setup_cache_build_env()
env = env | _setup_msvc_build_env()

# pytorch build
env = _setup_common_build_env()
env = env | _setup_cache_build_env()
env = env | _setup_clang_build_env()
```
(there's a balancing act there between having code that is too abstract and repeating ourselves too much though)

PR #4011, opened 2026-03-17
Changes to accomodate ctest based tests in test filters (rocprof-compute)

dileepr1:
## Motivation

Add test filter standardization support for rocprofiler-compute

## Technical Details

With this change rocprofiler-compute calls test_runner.py instead of test_rocprofiler_compute.py to initiate CI tests.

This PR adds support for component specific env vars in test_runner.py
Adds support for additional labels for ctest based tests in test_runner.py.
Adds unit tests for test_runner.py

Rocprofiler-compute works with the companion PR: https://github.com/ROCm/rocm-systems/pull/4155.


dileepr1:
The run at https://github.com/ROCm/TheRock/actions/runs/23558192220/job/68607817748?pr=4011 shows that rocprof-compute tests are getting called correctly from TheRock.
`ctest -L full --output-on-failure --parallel 8 --timeout 7200 --test-dir /__w/TheRock/TheRock/build/libexec/rocprofiler-compute -V --tests-information 1,,2 -LE full_exclude -LE ex_gpu`

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -85,6 +85,22 @@
 ROCM_PATH = Path(THEROCK_BIN_DIR).resolve().parent
 environ_vars["ROCM_PATH"] = str(ROCM_PATH)
 
+
+TEST_DIR = str(Path(THEROCK_BIN_DIR) / TEST_COMPONENT)
+
+# Add component-specific environment variables here
```
geomin12: i think each component may encounter this more. is there a way to add some dictionary where it handles all this? and then per component, we just add the correct vars? easier to update that way and cleaner (less dup)

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -154,15 +174,19 @@ def get_available_gpu_suite_tests():
         )
 
         gpu_archs = set()
-        prefix = "ex_gpu_"
+        exclude_labels = set()
```
geomin12, line 216: could we write unit tests for this?

geomin12 (commented):
two comments

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: I think this is a good suggestion, but the problem faced is that we cannot be sure of the env vars used by each component. TEST_DIR is a common one, but some others depends upon the component libraries. So I thought of adding the component specific vars in the beginning for visibility.

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: Added a set of unit tests for FindMatchingGpuArch and BuildCtestCommand.

build_tools/github_actions/test_executable_scripts/test_runner.py
geomin12: sure that is fine, but I think we can start with an empty dictionary like:

```
{
    "rocprofiler-compute": {...env settings},
}
```

and update as we go. although either or, eventually we will need to convert to dictionary instead of a giant if tree

build_tools/github_actions/tests/unit_test_runner.py
geomin12, line 1: fix naming

`test_runner_test.py`

dileepr1: Pre-commit flags it saying tests should be named as *_test.py (though it is!).

ScottTodd: Yeah it's banning any file from starting with `test_`, even it it also ends with `_test`:
https://github.com/ROCm/TheRock/blob/e65b265142e1bb6b306be13b8170b08f9e3e6880/.pre-commit-config.yaml#L53-L59

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: done!

build_tools/github_actions/tests/unit_test_runner.py
dileepr1: Is it good to go with testrunner_test.py, or are there any name suggestions?

dileepr1: Renamed testrunner_test.py to unit_test_runner_test.py

geomin12 (commented):
tests fail: https://github.com/ROCm/TheRock/actions/runs/24205784004/job/70701092396?pr=4011, expected?

otherwise, lgtm

dileepr1:
> tests fail: https://github.com/ROCm/TheRock/actions/runs/24205784004/job/70701092396?pr=4011, expected?
> 
> otherwise, lgtm

The test failure is due to quick tests going beyond 5 minutes.

dileepr1:
There was a problem with excluded tests being run, fixed it.
Tests are passing now.

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
 ROCM_PATH = Path(THEROCK_BIN_DIR).resolve().parent
 environ_vars["ROCM_PATH"] = str(ROCM_PATH)
 
+
+# Component-specific overrides applied on top of defaults.
+# - test_dir_from_rocm: path parts relative to ROCM_PATH to override the test directory
+# - env_prepend: env var name -> list of path-part lists (each relative to ROCM_PATH)
+#                to prepend to the existing value
```
geomin12: reading these out loud, and it still isn't completely clear (even to me). can we clarify? add examples?

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
...
+COMPONENT_OVERRIDES = {
+    "rocprofiler-compute": {
+        "test_dir_from_rocm": ["libexec", "rocprofiler-compute"],
+        "env_prepend": {
+            "PATH": [["bin"]],
+            "LD_LIBRARY_PATH": [["lib"], ["lib", "rocm_sysdeps", "lib"]],
```
geomin12: might be a better way to clarify path, seeing three libs was initially confusing

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
...
+        },
+    },
+}
+
+
+def _prepend_env_paths(env, base_path, prepend_dict):
+    """Prepend paths (relative to base_path) to environment variables."""
+    for env_key, path_parts_list in prepend_dict.items():
+        new_paths = [str(base_path.joinpath(*parts)) for parts in path_parts_list]
+        existing = env.get(env_key, "")
```
geomin12: better variable name pls

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
...
+        env[env_key] = ":".join(filter(None, new_paths + [existing]))
```
geomin12: is filter needed here?

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
...
 environ_vars["ROCM_PATH"] = str(ROCM_PATH)
 
+
+# Component-specific overrides applied on top of defaults.
+# - test_dir_from_rocm: path parts relative to ROCM_PATH to override the test directory
+# - env_prepend: env var name -> list of path-part lists (each relative to ROCM_PATH)
+#                to prepend to the existing value
+COMPONENT_OVERRIDES = {
+    "rocprofiler-compute": {
+        "test_dir_from_rocm": ["libexec", "rocprofiler-compute"],
```
geomin12: also, could have a better key name

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -86,6 +86,48 @@
...
+        "env_prepend": {
```
geomin12: additional_env_paths might be better?

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: That filter will protect the paths from ending with an ':' in case existing path is empty or not declared.
it will be similar to `env[env_key] = ":".join(new_paths + ([existing] if existing else [])`, but filter will protect both new_paths and existing.

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: Added better comments for clarification.

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: used new var names and made it cleaner

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: used new names

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: renamed!

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -463,7 +463,7 @@ def _get_script_path(script_name: str) -> str:
             "libexec/rocprofiler-compute/requirements.txt",
             "libexec/rocprofiler-compute/requirements-test.txt",
         ],
-        "test_script": f"python {_get_script_path('test_rocprofiler_compute.py')} -v",
+        "test_script": f"python {_get_script_path('test_runner.py')}",
```
vedithal-amd, line 466: Do we need to get rid off test_rocprofiler_compute.py if its not being used?
Please ensure the functionality is preserved when running test_runner.py and no regressions observed

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: done!

build_tools/github_actions/fetch_test_configurations.py
dileepr1: I'll remove it once test_runner.py is fully integrated.

geomin12 (approved):

Issue #3627, opened 2026-02-26
[Issue]: JAX wheels not built with rocm support


GeneDer:
### Problem Description

Installed jax wheels from the rock index https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu

For example
```
pip install jax_rocm7_pjrt jax_rocm7_plugin jaxlib \
    --index-url https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu 
```

Got the following error at runtime
```
  RuntimeError: Unable to initialize backend 'rocm': Backend 'rocm' is not in the list of known backends: ['cpu', 'tpu']. (set JAX_PLATFORMS='' to automatically choose an available backend)
```

### Operating System

Ubuntu24

### CPU

NA

### GPU

Mi325

### ROCm Version

TheRock 7.12

GeneDer:
If you need an example image you can use this one `rocm/mad-private:jax_the_rock_ci_66a406c_20260226`
These were the jax version installed 

<img width="1249" height="199" alt="Image" src="https://github.com/user-attachments/assets/bb36fcb8-e11c-4840-8a4c-b7c0341e08ea" />

GeneDer:
@amd-aakash @kiran-thumma just for your awareness

ScottTodd:
Another report of this here: https://github.com/ROCm/TheRock/issues/3878#issuecomment-4060226258

appears to be happening on our CI/CD test runs too: https://github.com/ROCm/TheRock/actions/runs/23082792559/job/67057733617

```
Test session starting on GPU ?

jax/tests/multi_device_test.py::MultiDeviceTest::test_broadcast 
-------------------------------- live log call ---------------------------------
INFO     jax_plugins.xla_rocm7:__init__.py:96 No ROCm wheel installation found
WARNING  jax_plugins.xla_rocm7:__init__.py:140 rocm_plugin_extension not found
INFO     jax._src.xla_bridge:xla_bridge.py:808 Unable to initialize backend 'tpu': INTERNAL: Failed to open libtpu.so: libtpu.so: cannot open shared object file: No such file or directory
PASSED                                                                   [  5%]
jax/tests/multi_device_test.py::MultiDeviceTest::test_closed_over_values_device_placement PASSED [ 11%]
jax/tests/multi_device_test.py::MultiDeviceTest::test_computation_follows_data PASSED [ 16%]
```

angelaAMD:
[...]

PR #4012, opened 2026-03-17
fix(ci): configure ROCm library paths for JAX wheel tests

WBobby:
## Motivation

JAX wheel tests in CI can currently report passing results without actually exercising the ROCm backend, which makes the real JAX wheel status hard to evaluate.

Two issues are contributing to this:

1. ROCm tarball installs to `/opt/rocm-<version>/` without a `/opt/rocm` symlink, while JAX wheels expect `/opt/rocm/lib` in their RUNPATH.
2. Without `JAX_PLATFORMS=rocm`, JAX may fall back to CPU, so tests can pass even when ROCm backend initialization is broken.

This PR is intended to improve CI signal for JAX ROCm testing by making the test environment match the wheel expectations and by preventing silent CPU fallback.

Related:
- https://github.com/ROCm/TheRock/issues/3627
- https://github.com/ROCm/jax/pull/737
- https://github.com/ROCm/rocm-jax/pull/362
- https://github.com/ROCm/rocm-jax/pull/363
- https://github.com/ROCm/rocm-jax/pull/364

## Changes

- **`ln -sfn "${DEST}" /opt/rocm`** — create a `/opt/rocm` symlink so wheel RUNPATH entries such as `/opt/rocm/lib` resolve correctly in CI.
- **`JAX_PLATFORMS: rocm`** — force JAX to initialize the ROCm backend in the test step, so CI fails loudly instead of silently running on CPU.

## Scope Of This PR

This PR adjusts the CI test environment and improves test correctness; it does **not** by itself fix the underlying JAX wheel packaging problem reported in `#3627`.

The underlying wheel-side fix is being addressed separately in:
- [ROCm/jax#737](https://github.com/ROCm/jax/pull/737)
- [ROCm/rocm-jax#362](https://github.com/ROCm/rocm-jax/pull/362)
- [ROCm/rocm-jax#363](https://github.com/ROCm/rocm-jax/pull/363)
- [ROCm/rocm-jax#364](https://github.com/ROCm/rocm-jax/pull/364)

## Why `LD_LIBRARY_PATH` Is Not Added Here

This PR intentionally avoids adding `LD_LIBRARY_PATH` as a CI-only workaround.

The goal here is to:
- make CI reflect whether the published wheels work with the expected install layout
- avoid masking remaining wheel packaging issues behind extra linker settings

If the companion wheel fixes are present in the tested artifacts, the `/opt/rocm` symlink should be sufficient for the expected RUNPATHs to resolve in CI.

## Dependency

For CI to pass end-to-end without extra linker overrides, the tested JAX wheels must include the corresponding RUNPATH fixes from:
- [ROCm/jax#737](https://github.com/ROCm/jax/pull/737) for `amd-main`
- [ROCm/rocm-jax#362](https://github.com/ROCm/rocm-jax/pull/362)
- [ROCm/rocm-jax#363](https://github.com/ROCm/rocm-jax/pull/363)
- [ROCm/rocm-jax#364](https://github.com/ROCm/rocm-jax/pull/364) for release branches

Until those fixes are available in the wheels used by CI, this PR may cause CI to fail earlier and more explicitly, which is the intended behavior.

## Test Plan

- [ ] Run the workflow on a GPU runner with ROCm tarball install
- [ ] Confirm JAX does not silently fall back to CPU
- [ ] Verify that backend initialization succeeds when fixed wheels are used
- [ ] Verify logs show ROCm backend initialization rather than CPU-only execution

## Test Result

Pending CI validation

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests

### DIFF
--- a/.github/workflows/test_linux_jax_wheels.yml
+++ b/.github/workflows/test_linux_jax_wheels.yml
@@ -214,7 +214,7 @@ jobs:
             --artifact-group "${{ inputs.amdgpu_family }}" \
             --output-dir "${DEST}"
           # Temporary CI workaround.
-          # TODO(#<issue-number>): Remove this /opt/rocm symlink once JAX wheel
+          # TODO(#4201): Remove this /opt/rocm symlink once JAX wheel
           # packaging/RUNPATH fixes are validated and CI no longer depends on
           # /opt/rocm -> /opt/rocm-<version> path translation.
           ln -sfn "${DEST}" /opt/rocm


.github/workflows/test_linux_jax_wheels.yml
```diff
@@ -213,6 +213,12 @@ jobs:
             --release "${{ inputs.rocm_version }}" \
             --artifact-group "${{ inputs.amdgpu_family }}" \
             --output-dir "${DEST}"
+          # Create symlink so that hardcoded RUNPATH /opt/rocm/lib in JAX wheels resolves
+          ln -sfn "${DEST}" /opt/rocm
+          # Set ROCm environment variables for dynamic linker.
+          # Both lib paths are required: main ROCm libs + vendored sysdeps libs
+          echo "ROCM_PATH=${DEST}" >> "$GITHUB_ENV"
+          echo "LD_LIBRARY_PATH=${DEST}/lib:${DEST}/lib/rocm_sysdeps/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" >> "$GITHUB_ENV"
```
marbre: I think adding `/opt/rocm/lib` to `LD_LIBRARY_PATH` is somehow acceptable but I don't think adding `/opt/rocm/lib/rocm_sysdeps/lib` is the right fix and needs another solution. The question is why those sysdeps aren't get picked up correctly and if we add this, the issue will be kind of masked in CI.

WBobby: Good point — adding `rocm_sysdeps/lib` to `LD_LIBRARY_PATH` is intentionally a short-term CI unblock, not the proper fix. Let me explain why it's needed here and what the real fix looks like.

### The proper fix (JAX side)

  I'll submit a fix to JAX's `jaxlib/rocm/rocm_rpath.bzl` to add `rocm_sysdeps/lib` to `_WHEEL_RPATHS`:

  Once that lands and new wheels are published, the LD_LIBRARY_PATH line here can be simplified to just ${DEST}/lib (or removed entirely if the symlink is sufficient).

  Why this CI PR is still needed

  Even after the JAX RPATH fix, this workflow still needs:
  1. ln -sfn — ROCm installs to /opt/rocm-version/ but RUNPATH is hardcoded to /opt/rocm/lib
  2. JAX_PLATFORMS=rocm — without this, tests silently pass on CPU (896 passed, 0 on GPU)

So this PR unblocks CI now; the JAX-side fix eliminates the need for rocm_sysdeps/lib in LD_LIBRARY_PATH going forward.

marbre: `JAX_PLATFORMS=rocm` is fine and doesn't concern me. I am also aware that you need to set some path but pointing to `/opt/rocm/lib` (and symlinking) should be it. So the what we set `LD_LIBRARY_PATH` should be different or rather we only want this as a short term fix.

WBobby: I submitted a JAX PR (https://github.com/ROCm/jax/pull/737). After it is merged and confirmed working, I can remove the LD_LIBRARY_PATH part from this PR.

WBobby:
Submitted a follow-up PR after https://github.com/ROCm/jax/pull/737 was merged.

WBobby:
Updated the PR description.

marbre:
Any update on the test plan?

WBobby:
JAX wheel builds are currently blocked by [#3876](https://github.com/ROCm/TheRock/issues/3876) (missing libamd_comgr_stub.a). This affects all JAX versions (v0.8.2 & v0.9.0) across all Python versions and architectures. The latest nightly run ([#626](https://github.com/ROCm/TheRock/actions/runs/23289094367), today Mar 19) still fails with the same error. This means no JAX wheel tests can run end-to-end in CI right now.

.github/workflows/test_linux_jax_wheels.yml
```diff
@@ -213,6 +213,8 @@ jobs:
             --release "${{ inputs.rocm_version }}" \
             --artifact-group "${{ inputs.amdgpu_family }}" \
             --output-dir "${DEST}"
+          # Create symlink so that hardcoded RUNPATH /opt/rocm/lib in JAX wheels resolves
+          ln -sfn "${DEST}" /opt/rocm
```
ScottTodd: > Fixes #3627

This does NOT fix the issue. It adds a **limited workaround for CI**. I'm seeing a lot of discussion and patching around incomplete software here, but I'm not seeing coordination on the actual development work and fixes.

Here is what I would expect/accept:
1. In any order, get the CI testing under control:
    1. Add this sort of workaround to start getting real CI signal
    2. Fix the JAX tests to fail if no GPU is detected, not fallback to CPU
    3. Add a sanity check test _in TheRock_ that directly tests our expectations for packaging and GPU detection, separate from what upstream unit tests do. For example: https://github.com/ROCm/TheRock/blob/1b0705f93ac5fa0d0326b4826d7f3e4ccd502f9f/external-builds/pytorch/smoke-tests/pytorch_smoke_test.py#L10-L21
2. Update our release instructions for JAX to include the workaround: https://github.com/ROCm/TheRock/blob/main/RELEASES.md#using-jax-python-packages
3. Land the actual fixes to JAX
    1. RUNPATH edits
    2. Integrate with the packaging guidelines we've already been following for PyTorch for the last ~9 months: https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md, stop relying on `/opt/rocm` and dockerfiles

So I think the code here _could_ be fine to land, but it needs to be part of that larger context. Changing how we _test our software_ does not _actually fix the software itself_, and we need to stay honest there.

WBobby: Thanks for the detailed review, @ScottTodd. I'd like to address your first point about getting CI testing under control.

I believe JAX_PLATFORMS=rocm (added in this PR) addresses all three sub-items:

i. Workaround to get real CI signal — The symlink + JAX_PLATFORMS=rocm ensures tests run against the ROCm backend. This is covered by this PR.

ii. Tests should fail if no GPU is detected — With JAX_PLATFORMS=rocm set, JAX will not silently fall back to CPU. If the ROCm backend fails to initialize (due to packaging issues, missing libraries, no GPU, etc.), JAX raises RuntimeError: Unable to initialize backend 'rocm' immediately, causing all tests to fail. This is exactly the behavior we want.

iii. Separate smoke test for packaging and GPU detection — I understand the reasoning here, since PyTorch has pytorch_smoke_test.py for this purpose. However, there's a key architectural difference: PyTorch silently skips GPU tests when torch.cuda.is_available() returns False, so CI can appear green even when GPU support is completely broken. That's why PyTorch needs an explicit assertion. JAX doesn't have this problem — JAX_PLATFORMS=rocm is an upstream-supported mechanism that forces backend initialization, making every test an implicit GPU availability check. Adding a separate TheRock-specific smoke test would be redundant and introduce code that diverges from upstream without providing additional coverage.

I'd prefer to use JAX's native platform selection rather than maintaining a custom smoke test, to keep our CI aligned with upstream patterns.

Happy to discuss if you see a gap that JAX_PLATFORMS=rocm doesn't cover.

ScottTodd: > ii. Tests should fail if no GPU is detected — With JAX_PLATFORMS=rocm set, JAX will not silently fall back to CPU. If the ROCm backend fails to initialize (due to packaging issues, missing libraries, no GPU, etc.), JAX raises RuntimeError: Unable to initialize backend 'rocm' immediately, causing all tests to fail.

Can you prove this with some logs and repro steps? I'd like to take your word for it, but so far that hasn't been sufficient with JAX... we've gone through multiple rounds of documentation reviews where the docs are just clearly wrong.

> However, there's a key architectural difference: PyTorch silently skips GPU tests when torch.cuda.is_available() returns False, so CI can appear green even when GPU support is completely broken.

No, it doesn't. At a minimum we hit this code path: https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/external-builds/pytorch/pytorch_utils.py#L102-L104

```bash
pip install torch  # Note: not using our ROCm release index, so this pull CPU or CUDA wheels
pip install -r D:/b/pytorch/.ci/docker/requirements-ci.txt
python D:/projects/TheRock/external-builds/pytorch/run_pytorch_tests.py --pytorch-dir D:/b/pytorch --amdgpu-family=gfx1100 --allow-version-mismatch

# ... (other logs)
# Getting GPU information from PyTorch...
# [ERROR] Failed to retrieve GPU info: ERROR:ROCm is not available
```

We need defense in depth here. We've gone _months_ without actually testing JAX GPU code. A bit of redundancy (with _current upstream_, not necessarily _future upstream_ which we don't control) is a small cost to pay to ensure that we don't ship clearly broken software.

WBobby: Thanks, @ScottTodd. I checked this a bit more.

From `xla_bridge.py`:
https://github.com/ROCm/jax/blob/amd-main/jax/_src/xla_bridge.py#L789-L833

if `JAX_PLATFORMS=rocm` is explicitly set, JAX raises when the ROCm backend cannot be initialized. If it is unset, JAX may continue with another available backend.

I verified this on the same CPU-only build server:
- `tests/multi_device_test.py` passes when `JAX_PLATFORMS` is unset
- the same test fails immediately with `JAX_PLATFORMS=rocm` set:
  `RuntimeError: Unable to initialize backend 'rocm'`

The passing case also logs:
- `No ROCm wheel installation found`
- `rocm_plugin_extension not found`

So a passing result here does not necessarily mean the test is actually running on ROCm.

I still think the cleanest approach is to merge this PR first to fix the immediate test issue, then follow up with a separate PR for JAX-specific backend enforcement / skip handling, similar to `pytorch_utils.py`.

PR #3947, opened 2026-03-12
add additional targets to CK blacklist

SreecharanGundaboluAMD:
This pull request updates the list of unsupported GPU targets for the composable kernel feature in the `ml-libs/CMakeLists.txt` file. The change expands the set of GPU architectures for which composable kernel is disabled, to address the issue https://github.com/ROCm/TheRock/issues/3931

GPU support configuration:

* Expanded the `_ck_unsupported_gfx_targets` list to include additional AMD GPU architectures (`gfx900`, `gfx1030`, `gfx1031`, `gfx1032`, `gfx1034`), ensuring composable kernel is not enabled on these unsupported targets.
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4019, opened 2026-03-17
[CI] Revert "add additional targets to CK blacklist"

brockhargreaves-amd:
Serious enough build time regression to warrant a revert.

Reverts ROCm/TheRock#3947


brockhargreaves-amd:
I reverted the wrong PR!

PR #4023, opened 2026-03-18
[ci] Adding new mi325 label

geomin12:
Adding new mi325 label provided by OSSCI. We have 13N and max of 17N (as of 3/17/2026 5:37pm PST)

As requested in standup, we are only running this when it is a non-quick test (aka submodule bumps)


geomin12:
Landing as test in PR description is enough proof. `run-full-tests-only` works well with gfx1151 Windows

PR #4038, opened 2026-03-18
Add patch to fix Comgr memory leak in anonymous bitcode naming

lamb-j:
Cherry-picks https://github.com/ROCm/llvm-project/commit/cb5b241d5de5a86471da5599ec06a6d6d15ce443 from [llvm-project](https://github.com/ROCm/llvm-project) amd-staging branch to fix a memory leak where malloc'd buffers were not freed when assigning anonymous bitcode names.

Removal: Week 12 (amd-compiler-2026-03-19) or later submodule bump

### DIFF
--- a/patches/amd-mainline/llvm-project/0011-Comgr-Fix-memory-leak-when-assigning-anon-bit-rm-w12.patch
+++ b/patches/amd-mainline/llvm-project/0011-Comgr-Fix-memory-leak-when-assigning-anon-bit-rm-w12.patch
@@ -1,0 +1,37 @@
+From cb5b241d5de5a86471da5599ec06a6d6d15ce443 Mon Sep 17 00:00:00 2001
+From: Jacob Lambert <jacob.lambert@amd.com>
+Date: Fri, 13 Mar 2026 08:43:55 -0700
+Subject: [PATCH] [Comgr] Fix memory leak when assigning anonymous bitcode
+ names (#1742)
+
+Use setName() instead of manually managing memory with malloc/free. The
+previous code leaked the original empty name buffer when overwriting
+Input->Name directly.
+
+Co-authored-by: Claude <noreply@anthropic.com>
+---
+ amd/comgr/src/comgr-compiler.cpp | 8 +++-----
+ 1 file changed, 3 insertions(+), 5 deletions(-)
+
+diff --git a/amd/comgr/src/comgr-compiler.cpp b/amd/comgr/src/comgr-compiler.cpp
+index 51b03ab36b0c..da1590b9aa29 100644
+--- a/amd/comgr/src/comgr-compiler.cpp
++++ b/amd/comgr/src/comgr-compiler.cpp
+@@ -1644,11 +1644,9 @@ amd_comgr_status_t AMDGPUCompiler::linkBitcodeToBitcode() {
+       // string to assign. This string is used when the DataObject is written
+       // to the file system via SAVE_TEMPS, or if the object is a bundle which
+       // also needs a file system write for unpacking
+-      const size_t BufSize = sizeof(char) * 30;
+-      char *Buf = (char *)malloc(BufSize);
+-      snprintf(Buf, BufSize, "comgr-anon-bitcode-%d.bc", std::rand() % 10000);
+-
+-      Input->Name = Buf;
++      std::string Name =
++          "comgr-anon-bitcode-" + std::to_string(std::rand() % 10000) + ".bc";
++      Input->setName(Name);
+     }
+ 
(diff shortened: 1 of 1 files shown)


patches/amd-mainline/llvm-project/0011-Comgr-Fix-memory-leak-when-assigning-anon-bit-rm-w12.patch
ScottTodd, line 1: We have these docs for introducing new patches: https://github.com/ROCm/TheRock/tree/main/patches#rules-for-creating-patches , FYI. A bit more structure around commit or PR links would help:

> Cherry-picks cb5b241d5de5 from llvm-project amd-staging to fix a memory leak where malloc'd buffers were not freed when assigning anonymous bitcode names.

Here's the link to the commit: https://github.com/ROCm/llvm-project/commit/cb5b241d5de5a86471da5599ec06a6d6d15ce443

> Removal: amd-compiler-ww-2026-12

Can you clarify here - is this an open pull request already, a branch, or something else? I'm not familiar with the naming scheme and "2026-10" reads like "~~November~~ October 2026" to me, which is pretty far away :P

I see these:
* https://github.com/ROCm/TheRock/tree/users/lajagapp/amd-compiler-ww-2026-10
* https://github.com/ROCm/llvm-project/tree/amd-compiler-ww-2026-10

lamb-j: Ok updated the PR description with some more details. I'm leaning on this case from the documentation:

> Temporary workarounds: Short-term fixes for critical issues while waiting for commits to be merged upstream. Any commits carried as patches must be on a path to being reviewed and accepted into their respective repositories.

The patch is already landed in our "develop" (amd-staging). But the submodule bump to bring it in won't happen for a few weeks (we're working on shortening this runway).

I'm mostly experimenting with creating a patch in TheRock vs doing a custom 1-off submodule bump to include that patch (but that breaks consistency with our amd-staging branch).

It seems like the patch system is set up well for this, assuming:
- we only create patches for things already landed in our "develop", and scheduled to be brought in with a future submodule update (in this case week 12 or later update)
- we only need a few patches between submodule updates (so the total number of active patches stays below 5 for example)

But curious to hear your thoughts (maybe this is a gross abuse of the patch system)

marbre: Actually there is no strong need for us to strictly follow `amd-mainline` (we kind of don't do this anyway if we apply additional patches). An alternative could be an `amd-mainline + N-off`-branch where the patches directly get applied to this branch? This would get us the same result, so not sure about pros or cons but it would be kind of more under your control.

lamb-j: Right, that makes sense. And we're already moved away from having an amd-mainline branch (at least we're experimenting with it). Instead, we're taking a snapshot of amd-staging every week and trying to update the submodule pointer with that.

But there's still a 2-3 week delay for testing and fixing issues. In this example, we have a change that will be the week 12 snapshot (amd-compiler-ww-2026-12). Which will hit TheRock naturally in ~3 weeks. But the fix is needed "yesterday".

I could cherry-pick directly this change on top of our current submodule, which points to week 6 (amd-compiler-ww-2026-06). But then I would also need to make sure it gets into every other submodule pointer update between now and the week 12 branch update; for example amd-compiler-ww-2026-10 update, amd-compiler-ww-2026-11 update (we'll probably skip weeks 7-9).

Alternatively we create one TheRock patch here, and have it get cleaned out with the amd-compiler-ww-2026-12 update.

I think both approaches are valid, and honestly others are probably more in favor of maintaining cherry-picks across multiple branches than maintain patches.

I'm mostly just brainstorming through the patch mechanism to see if it would be helpful for us.

lamb-j: The patch approach could be problematic for our local component testing with TheRock. We're using submodule override to switch the branch (to test our "develop"). But the patches may or may not apply cleanly there, which would break the local build.

But really our local build shouldn't need any patches (at least not compiler patches). Per the first bullet above, any patches should have already been merged into "develop". So local build can just omit compiler patches and avoid the issue entirely

Going to land this for now, track it, and see how it goes.

PR #4045, opened 2026-03-18
RFC0010: fetch test scripts from super-repos in CI

jayhawk-commits:
Part of the TheRock test script migration in RFC0010.

This PR updates TheRock CI to resolve test scripts from their owning
super-repos when available:

- `rocm-systems/test/therock/`
- `rocm-libraries/test/therock/`

## What changed

- `fetch_test_configurations.py` prefers canonical script locations in
  `rocm-systems` and `rocm-libraries`, falling back to the existing
  `test_executable_scripts/` copies when those super-repo paths are not present.
- Test workflows now run `fetch_sources.py` on the test runner so the
  `rocm-systems/` and `rocm-libraries/` submodules exist when tests execute.
- Windows test workflows were updated to match existing build-side behavior:
  - enable `core.longpaths` before `fetch_sources.py`
  - set `core.longpaths` globally in `test_component.yml`
  - install DVC before `fetch_sources.py`
  - use `--depth 1` for `fetch_sources.py`

## Behavior

- Scripts that still live in TheRock (`test_sanity.py`, `test_runner.py`) are unchanged.
- Out-of-scope scripts (for example rocgdb, libhipcxx, fusilli/iree-libs providers)
  continue to use `THEROCK_SCRIPT_DIR`.
- The fallback to TheRock-local script copies remains in place for compatibility
  when super-repos are not checked out.

## Dependencies

Depends on:
- ROCm/rocm-systems#4190
- ROCm/rocm-systems#4197
- ROCm/rocm-libraries#5563
- ROCm/rocm-libraries#5572
- corresponding submodule bumps in TheRock

RFC:
- https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0010-Test-Scripts-Migration.md

### DIFF
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -106,7 +106,7 @@ jobs:
       - name: "Fetch sources"
         timeout-minutes: 30
         run: |
-          ./build_tools/fetch_sources.py --jobs 12
+          ./build_tools/fetch_sources.py --jobs 12 --depth 1
 
       - name: "Configuring CI options"
         id: configure
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -132,7 +132,7 @@ jobs:
       - name: "Fetch sources"
         timeout-minutes: 30
         run: |
-          ./build_tools/fetch_sources.py --jobs 12
+          ./build_tools/fetch_sources.py --jobs 12 --depth 1
 
       - name: Runner health status
         run: |


jayhawk-commits:
Will not try to merge until submodules are bumped and have the aforementioned changes.

jayhawk-commits:
Submodules have been bumped.

dileepr1 (approved):
Looks good!

.github/workflows/test_artifacts.yml
```diff
@@ -93,6 +93,21 @@ jobs:
...
+
+      - uses: iterative/setup-dvc@4bdfd2b0f6f1ad7e08afadb03b1a895c352a5239 # v2.0.0
+        if: ${{ runner.os == 'Windows' }}
+        with:
+          version: '3.62.0'
+
+      - name: "Fetch sources"
+        timeout-minutes: 30
+        run: |
+          ./build_tools/fetch_sources.py --jobs 12 --depth 1
```
geomin12, line 109: anyway we can just checkout rocm-libs / systems? particularly the test directories? fetch_sources.py can take a while...

might need a fetch_sources.py adjustment with a new flag

jayhawk-commits: That can be a separate improvement to `fetch_sources.py`, where submodules can be configured with sparse checkouts.

jayhawk-commits:
Sanity tests passed on WIndows gfx110X, so the previous hurdles with checkouts and long paths on the Windows test workflows seem to be resolved.

jayhawk-commits:
The current CI failures are not from the changes in this PR.

.github/workflows/test_artifacts.yml
ScottTodd: Please reconsider here. I don't think a follow-up issue (https://github.com/ROCm/TheRock/issues/4559) is sufficient, this should be blocking.

Fetching all submodules took 8 minutes on https://github.com/ROCm/TheRock/actions/runs/24378505268/job/71391558133, and that is unreasonably high overhead to be adding to every test job, especially for sanity checks and quick/smoketests when our test runners are already oversubscribed.

Tests also **cannot** depend on project source files (they should get their binaries, test files, etc. from artifacts; test harness scripts are fine), and we've had a few PRs that would happily take a dep on these files if they were present: https://github.com/ROCm/TheRock/pull/4370#discussion_r3060983786

jayhawk-commits: That makes sense. I'll revert.

PR #4039, opened 2026-03-18
Allow running multi-arch CI on pull requests with an opt-in label

ScottTodd:
## Motivation

This is a step towards enabling multi-arch CI on presubmit. I'm planning this sequence:
1. (Done) Allow running multi-arch CI manually via `workflow_dispatch`
2. (Done) Run multi-arch CI on `push`
3. (This PR) Allow explicit opt-in to running multi-arch CI on `pull_request`
4. Add implicit opt-in to running multi-arch CI on `pull_request` based on files modified
5. Run multi-arch CI by default for most changes: https://github.com/ROCm/TheRock/issues/3337
6. Disable the older CI: https://github.com/ROCm/TheRock/issues/3340

## Technical Details

* The `ci:run-multi-arch` label naming matches the `ci:skip` and `ci:run-all-archs` names from https://github.com/ROCm/TheRock/pull/4037.
* I'm planning on deeper changes to how "configure CI" works for multi-arch CI, see https://github.com/ROCm/TheRock/issues/3399. I think we've outgrown `build_tools/github_actions/configure_ci.py`, so I have a larger feature branch that completely forks that file for multi-arch CI.

## Test Plan

1. Add the label to this PR
2. Check that multi-arch CI runs with the expected set of targets, compare against runs on `push` (https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci.yml?query=branch%3Amain)

## Test Result

https://github.com/ROCm/TheRock/actions/runs/23267154576?pr=4039
* Uploads to and downloads from `therock-ci-artifacts-external` work from this fork PR
* Configure results:
    > ## Workflow configure results
    > 
    > * `linux_variants`: [['gfx1151', 'gfx94X-dcgpu', 'gfx120X-all', 'gfx110X-all']]
    > * `linux_test_labels`: []
    > * `linux_use_prebuilt_artifacts`: false
    > * `windows_variants`: [['gfx1151', 'gfx120X-all', 'gfx110X-all']]
    > * `windows_test_labels`: []
    > * `windows_use_prebuilt_artifacts`: false
    > * `enable_build_jobs`: true
    > * `test_type`: quick
    > * `run_functional_tests`: false
    
## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4051, opened 2026-03-18
Simplify s3_client auth/credentials setup using boto3

ScottTodd:
## Motivation

The credentials setup in `build_tools/_therock_utils/artifact_backend.py` was not functional on https://github.com/ROCm/TheRock/pull/4039 - the `AWS_SHARED_CREDENTIALS_FILE` credentials we use to upload to therock-ci-artifacts-external was getting bypassed so uploads were failing.

It appears that we can just use the default credential configuration behavior provided by boto3 here too: https://docs.aws.amazon.com/boto3/latest/guide/credentials.html#configuring-credentials.

## Technical Details

I couldn't find any modern documentation stating that `signature_version=UNSIGNED` would be required for reading from public buckets when unauthenticated, and testing on a few machines suggests that it is NOT required. Several coding agents were vehement that we need it though... maybe outdated guidance in the training set 🤷?

## Test Plan and Results

* Tested on https://github.com/ROCm/TheRock/pull/4039
  * https://github.com/ROCm/TheRock/actions/runs/23267154576?pr=4039 was able to upload and download to therock-ci-artifacts-external
* Removed unit tests that were doing nothing (just checking that APIs were called, not actually interacting with the credentials or auth in a meaningful way)
* We'll continue to monitor workflows afterwards to see if anything was missed

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/artifact_backend.py
+++ b/build_tools/_therock_utils/artifact_backend.py
@@ -220,9 +220,14 @@ class S3Backend(ArtifactBackend):
         """Lazy-initialized boto3 S3 client.
 
         Credentials are resolved through boto3's default credential chain
-        (environment variables, shared credentials file, instance profile,
-        etc.). This matches how the ``aws`` CLI and storage_backend.py
-        resolve credentials.
+        (see https://docs.aws.amazon.com/boto3/latest/guide/credentials.html).
+        Relevant locations are checked in order:
+        1. Environment variables:
+           `AWS_ACCESS_KEY_ID`
+           `AWS_SECRET_ACCESS_KEY`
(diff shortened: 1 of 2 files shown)


marbre (approved):
Thanks!

marbre:
@ScottTodd, this broke among other the validate artifact structure test. Reverted with PR #4058.

ScottTodd:
Argh. Thanks. Guess I missed in my testing which code path was being used, so `signature_version=UNSIGNED` _is_ required, and `build_tools/_therock_utils/storage_backend.py` is likely failing without credentials. I'll get them in alignment the other way and add unit tests that exercise that path (maybe try to list files in some of our public buckets)

PR #4056, opened 2026-03-19
Use COMGR_DLL_NAME via patches to subprojects (workaround for unmerged submodule changes)

lamb-j:
## Summary

Make TheRock build with the new `amd_comgr.dll` naming **without waiting for submodule bumps** to pick up the upstream Comgr (llvm-project#1278, merged) and CLR (rocm-systems#4211, pending) changes. Patches the subprojects so they behave as if those upstream PRs were already in the pinned commits.

**End state matches #4730** (the planned cleanup of #4535's bridging code), with two additional patch files that make the patched submodules behave as if upstream PRs landed.

## Context

- #4535 added a versioned `COMGR_DLL_NAME` bridge in TheRock's pre-hook and CLR `CMAKE_ARGS` to keep things working while the upstream submodule changes were in flight.
- llvm-project#1278 is merged but not yet in TheRock's `compiler/amd-llvm` pin.
- rocm-systems#4211 is open, with CI blocked by an unrelated TheRock/rocm-systems Docker image SHA mismatch (TheRock #4596).
- We can't reliably pull either change in via submodule bump in the near term, so this PR applies them as patches.

## Changes

1. **Cleanup (matches #4730)** — remove the bridging code from #4535:
   - `compiler/pre_hook_amd-comgr.cmake`: drop the `COMGR_DLL_NAME` cache var and deferred `set_target_properties` fallback.
   - `core/CMakeLists.txt`: drop the `_comgr_dll_name` computation and `-DCOMGR_DLL_NAME=...` args passed to hip-clr / ocl-clr.
   - `compiler/CMakeLists.txt`: drop `-DTHEROCK_HIP_MAJOR_VERSION` / `-DTHEROCK_HIP_MINOR_VERSION` from amd-comgr `CMAKE_ARGS` (new Comgr no longer reads them).

2. **Patches** — apply the upstream PRs to the pinned submodule sources:
   - `patches/amd-mainline/llvm-project/0012-Comgr-Use-CMake-configurable-COMGR_DLL_NAME-on-Windows.patch` — backports llvm-project#1278.
   - `patches/amd-mainline/rocm-systems/0001-CLR-Use-CMake-configurable-Comgr-DLL-name-instead-of.patch` — backports rocm-systems#4211.

## Verified

Both patches `git apply --check` cleanly against the current submodule pins (`compiler/amd-llvm` and `rocm-systems`).

## Follow-up cleanup

When the relevant upstream changes are picked up via submodule bump, remove the corresponding patch file. Once both patches are gone, the diff against main is exactly #4730.

## Test plan

- [ ] Linux CI passes (no Windows-specific changes affect Linux builds)
- [ ] Windows build produces `amd_comgr.dll`
- [ ] CLR loads Comgr successfully at runtime on Windows
- [ ] HIP runtime compilation (hiprtc) works end-to-end


patches/amd-mainline/rocm-systems/0001-CLR-Use-CMake-configurable-Comgr-DLL-name-instead-of.patch
```diff
@@ -0,0 +1,141 @@
...
+Date: Sun, 30 Mar 2026 00:00:00 -0700
+Subject: [PATCH] [CLR] Use CMake-configurable Comgr DLL name instead of
+ runtime versioning
+
+Replace the runtime is_versioned parameter in Comgr::LoadLib with a
+CMake cache variable COMGR_DLL_NAME. This simplifies the DLL loading
+logic and allows the DLL name to be configured at build time rather
+than computed at runtime based on HIP version numbers.
+
+Upstream PR: https://github.com/ROCm/rocm-systems/pull/4211
```
ScottTodd: Can we land https://github.com/ROCm/rocm-systems/pull/4211 instead of introducing this patch? We're at 0 patches for rocm-systems and I don't want to slip back there.

external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
```diff
@@ -110,8 +110,6 @@ def test_conv_transpose2d(self):
         inputs = torch.randn(1, 4, 5, 5, device="cuda")
         weights = torch.randn(4, 8, 3, 3, device="cuda")
         # Simply running any conv op exercises MIOpen and library loading.
-        # On Windows, this may fail if `amd_comgr_3.dll` (from build output) is
-        # used instead of `amd_comgr0605.dll` that is expected at runtime.
         result = torch.nn.functional.conv_transpose2d(inputs, weights, padding=1)
```
ScottTodd, line 115: This comment was useful at the time. Maybe keep it, but update the DLL names to something more generic?

patches/amd-mainline/rocm-systems/0001-CLR-Use-CMake-configurable-Comgr-DLL-name-instead-of.patch
stellaraccident: If Jacob is going to own the landing and retiring of this patch across a bump, then I'm fine with it. It can be hard to do these config changes automically across branches.

ScottTodd: Fair enough, but please keep the timeline as short as possible.

We do still have the patch handling code in rocm-systems: https://github.com/ROCm/rocm-systems/blob/83e5988323c52d3cd136ce8dd975781ec16bdf51/.github/workflows/therock-ci-linux.yml#L69-L72
```yml
      - name: Patch rocm-systems
        run: |
          # Remove patches here if they cannot be applied cleanly, and they have not been deleted from TheRock repo
          # rm ./TheRock/patches/amd-mainline/rocm-systems/0003-Use-is_versioned-true-consistently-in-both-Comgr-Loa.patch
```
I would hope that at some point we can delete that and simplify.

Carrying _any_ patches is going to block CI de-duplication efforts (cc @jayhawk-commits), complicate manifest file generation, and lead more people to proposing workarounds like https://github.com/ROCm/TheRock/pull/4057

external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
```diff
@@ -110,8 +110,6 @@ def test_conv_transpose2d(self):
         inputs = torch.randn(1, 4, 5, 5, device="cuda")
         weights = torch.randn(4, 8, 3, 3, device="cuda")
         # Simply running any conv op exercises MIOpen and library loading.
```
lamb-j, line 112:
```suggestion
        # Simply running any conv op exercises MIOpen and library loading.
        # May fail on Windows if an incompatible `amd_comgr.dll` is used.
```

external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
lamb-j: Updated. Though hopefully this patch, and the other planned Windows improvements, move us to a place where errors like this don't happen

patches/amd-mainline/rocm-systems/0001-CLR-Use-CMake-configurable-Comgr-DLL-name-instead-of.patch
lamb-j: Good points. Planning to coordinate everything with rocm-systems/projects/clr team to make sure we're not introducing any long-lived patches

lamb-j:
I've re-worked this patch into https://github.com/ROCm/TheRock/pull/4535

The new PR updates TheRock to handle both the old behavior and new behavior for both llvm-project and rocm-systems, so we can land both PRs in their respective repositories, and the submodule bumps won't cause breakage. Which avoids creating new patches.

After that lands, we can do some additional cleanup in TheRock to get to the "final" state represented by this PR.

This PR was/is still useful for testing the final product though

lamb-j:
Closing again in favor of #4730. We can wait until the submodules are updated properly

PR #1790, opened 2025-10-14
Add a tool to generate manifest file

erman-gurses:
Closes  #1748 
The tool generates the manifest file in `json` format, for example:
```
{
  "the_rock_commit": "be5181cebb7c3741806e61f1f5a6aa68bffd2cee",
  "submodules": [
    {
      "submodule_name": "amdsmi",
      "submodule_path": "base/amdsmi",
      "submodule_url": "https://github.com/ROCm/amdsmi.git",
      "pin_sha": "8758b8f75aca44b4aea60a647df0c3868242330d",
      "patches": [
        "patches\\amd-mainline\\amdsmi\\0001-Fix-finding-libamd_smi.so.patch"
      ]
    },
    {
      "submodule_name": "half",
      "submodule_path": "base/half",
      "submodule_url": "https://github.com/ROCm/half.git",
      "pin_sha": "207ee58595a64b5c4a70df221f1e6e704b807811",
      "patches": []
    },
    {
      "submodule_name": "rocm-cmake",
      "submodule_path": "base/rocm-cmake",
      "submodule_url": "https://github.com/ROCm/rocm-cmake.git",
      "pin_sha": "10155d7272ea1bf79f6b5a9dbc339657af1aa372",
      "patches": []
    },
    {
      "submodule_name": "rccl",
      "submodule_path": "comm-libs/rccl",
      "submodule_url": "https://github.com/ROCm/rccl.git",
      "pin_sha": "d23d18f423d661556a5091f0dc9f27dec2adf67f",
      "patches": []
    },
    {
      "submodule_name": "rccl-tests",
      "submodule_path": "comm-libs/rccl-tests",
      "submodule_url": "https://github.com/ROCm/rccl-tests.git",
      "pin_sha": "e1b8a3aefcdf6482f0f5a35224e19eed2acc685d",
      "patches": []
    },
    {
      "submodule_name": "llvm-project",
      "submodule_path": "compiler/amd-llvm",
      "submodule_url": "https://github.com/ROCm/llvm-project.git",
      "pin_sha": "a7d47b26ca0ec0b3e9e4da83825cace5d761f4bc",
      "patches": [
        "patches\\amd-mainline\\llvm-project\\0001-Enable-MSVC-support-in-comgr.patch",
        "patches\\amd-mainline\\llvm-project\\0002-hipcc-fix-default-include-path-on-Windows-and-adapt-.patch",
[...]

PR #4057, opened 2026-03-19
Script To Generate Repro Shell Script

linuxrocks123:
This script generates a list of shell commands that, when run, will bring a TheRock checkout into the same state used by the CI to run a PR.  This can help with reproducing and debugging CI failures.

### DIFF
--- a/build_tools/generate_repro_checkout.py
+++ b/build_tools/generate_repro_checkout.py
@@ -4,13 +4,15 @@ import os
 import subprocess
 
 def get_output_for_cmd(cmd_str):
-    return subprocess.run(cmd_str,shell=True,stdout=subprocess.PIPE).stdout
+    return subprocess.run(cmd_str,shell=True,text=True,stdout=subprocess.PIPE).stdout
 
 def recurse_git_diff(diffcmd):
-    lines = map(get_output_for_cmd(diffcmd).readlines(),str.rstrip)
+    lines = list(map(str.rstrip,get_output_for_cmd(diffcmd).split('\n')))
     i=0
     while i<len(lines):
         diff_line = lines[i]
+        if diff_line=='':
+            break
         i+=1
         index_line = lines[i]
         if index_line.split(' ')[-1]=="160000":
@@ -18,11 +20,11 @@ def recurse_git_diff(diffcmd):
             submodule_line = lines[i]
             submodule = diff_line.split(' ')[-1][2:]
 
-            sha_dirty = submodule.split(' ')[-1]
+            sha_dirty = submodule_line.split(' ')[-1]
             sha_dirty_parts = sha_dirty.split('-')
             print("pushd "+submodule)
             print("git reset --hard "+sha_dirty_parts[0])
-            if len(sha_dirty_parts > 1) and sha_dirty_parts[1]=="dirty":
+            if len(sha_dirty_parts) > 1 and sha_dirty_parts[1]=="dirty":
                 recurse_git_diff(submodule)
             print("popd")
             i+=1
@@ -34,6 +36,6 @@ def recurse_git_diff(diffcmd):
                 i+=1
             print("DIRTY_EOF")
 
-print("git reset --hard "+get_output_for_cmd("git rev-parse HEAD~1").read().rstrip())
+print("git reset --hard "+get_output_for_cmd("git rev-parse HEAD~1").rstrip())
 recurse_git_diff("git diff")
 recurse_git_diff("git diff HEAD~1")


build_tools/generate_repro_checkout.py
```diff
@@ -0,0 +1,41 @@
...
+            i+=1
+            print("patch -p1 <<DIRTY_EOF")
+            while i < len(lines) and lines[i].split(' ')[0]!="diff":
+                print(lines[i])
+                i+=1
+            print("DIRTY_EOF")
+
+print("git reset --hard "+get_output_for_cmd("git rev-parse HEAD~1").rstrip())
+recurse_git_diff("git diff")
+recurse_git_diff("git diff HEAD~1")
```
ScottTodd, line 41: I don't see what problem this is solving. Can you share some output of the script and show why
```
git checkout [commit]
python3 ./build_tools/fetch_sources.py
```
isn't sufficient?

linuxrocks123: @ScottTodd this script is the first of an effort to make reproducing CI failures more convenient and reliable.  It runs after the sources have been fetched and resets the checkouts to the ones used by the CI.

Example output:
```
psimmons@firestorm:~/work/DockerRock/EphemeralRock$ python3 ../../MyRock/build_tools/generate_repro_checkout.py 
git reset --hard 8e7aa12835fc97260bcc1ce3ae911132be8eb275
pushd compiler/amd-llvm
git reset --hard c68a655a56b474e3f62259ff7f2144523ba9dea4
popd
```

This script will run inside the CI, and its output will be tarred into an asset.  Running the script in the asset will reset a checkout to the SHA IDs used by a PR.

The observed behavior of the CI is that the CI will add exactly one commit to TheRock reflecting its PR-specific changes.  If this is not always done, or if more than one commit may be added, the script will need to be elaborated to handle any number of PR-specific commits, instead of assuming just one.

ScottTodd: The ways to reproduce CI checkouts reliably are:
1. Checkout the PR branch itself (easier, typical git workflow)
2. Checkout the workflow run merge commit (matching what github actions does).

then running `fetch_sources.py`.

For an example of (2):
* https://github.com/ROCm/TheRock/pull/4048
* https://github.com/ROCm/TheRock/actions/runs/23291563723/job/67780571317?pr=4048#step:2:73
    ```
    Checking out the ref
      /usr/bin/git checkout --progress --force refs/remotes/pull/4048/merge
      Note: switching to 'refs/remotes/pull/4048/merge'.
      
      HEAD is now at 50acead Merge 0806f31e3177d2be4aced37f8b614a3f632ba1b5 into ba7457c07d18c2d3741a02b20aa19b17e433b0d6
    /usr/bin/git log -1 --format=%H
    50aceadf952a67a182747f1da4e67d3ba493f878
    ```
* PR commit: https://github.com/ROCm/TheRock/commit/0806f31e3177d2be4aced37f8b614a3f632ba1b5
* Base commit: https://github.com/ROCm/TheRock/commit/ba7457c07d18c2d3741a02b20aa19b17e433b0d6
* Merge commit: https://github.com/ROCm/TheRock/commit/50aceadf952a67a182747f1da4e67d3ba493f878

linuxrocks123: Hi @ScottTodd, I don't think this is sufficient for reproducing failures, except possibly for failures related to PRs for TheRock itself.  For a PR for any other repository, this workflow will not synchronize the SHA values of the other submodules between the CI and the checkout.  To reliably reproduce library build failures related to an LLVM PR, for instance, it is necessary to ensure that the exact same version of, for instance, `rocm-libraries` is in the checkout.

The logs at https://github.com/ROCm/llvm-project/actions/runs/23255094555/job/67608698422?pr=1752 show a `git commit` command to TheRock, but no `git push` command, so I don't think it will be possible to directly check out that CI commit from outside the CI.  Future CI changes could also add patches to the source that the CI scripts either do not commit or do not push.  Therefore, I think the best approach to solve this issue is to provide a script asset that, when run as a single command, sets a checkout of TheRock to exactly the state it was when running a particular CI job.

ScottTodd: Ah. Those extra requirements help. For those cases I'd say:

* We could generate branches in TheRock for all CI runs. This is what I was originally proposing before all these workflows got forked into different repositories, as it solves many of these problems naturally, no extra code needed.
* We generate "manifest" files for TheRock, including information about commits and patches used for all submodules. We should generate these consistently when run in other repositories if we aren't already. Then, we could add a script that loads a manifest file and gets the repository in sync with it.

ScottTodd: Here's an example manifest file we already generate: https://therock-ci-artifacts.s3.amazonaws.com/23291563723-linux/manifests/gfx94X-dcgpu/therock_manifest.json

linuxrocks123: @ScottTodd, those approaches could work, but they sound like more total work to me, and also possibly less robust compared with a direct translation of the final state of the source code into a script.  Can we merge this for now and delete it later if someone later implements those alternatives and they prove superior?  This project is my highest priority at the moment.

Looking at this, I probably should add comments.  I will do that later today.

ScottTodd: Any code in TheRock needs to be aligned with the common infra. We're not building multiple unrelated ways to do the same thing.

You can chat with @erman-gurses about the manifest generation or @amd-hsivasun for the manifest diffing (when I originally reviewed that it was also doing its own thing and not using the manifest files, I'm not sure where that landed after reviews).

erman-gurses: > Any code in TheRock needs to be aligned with the common infra. We're not building multiple unrelated ways to do the same thing.
> 
> You can chat with @erman-gurses about the manifest generation or @amd-hsivasun for the manifest diffing (when I originally reviewed that it was also doing its own thing and not using the manifest files, I'm not sure where that landed after reviews).

The initial TheRock manifest generation PR is here: https://github.com/ROCm/TheRock/pull/1790.  `build_tools/generate_therock_manifest.py`‎ is the file that generates the manifest.
@linuxrocks123, please let me know if you have any questions about it.
For manifest diff, @amd-hsivasun can point his PR's location and answer questions if it is needed.

linuxrocks123: @ScottTodd `generate_therock_manifest.py` is larger than mine to solve only part of the same problem, so I am not seeing a reason to favor that approach yet.  It would be a nicer user experience to have this in-tree than to have someone add a patch to each PR that needs it.  Must we really do it that way?

erman-gurses: > sets a checkout of TheRock to exactly the state it was when running a particular CI job.

Can you please explain in the PR description what is lacking and what problem you are solving for the current manifest?

linuxrocks123: @erman-gurses it doesn't appear that the manifest file lists patches that are not in fact applied, and the manifest also appears to list submodules that are not in fact checked out at the time of the build.  However, I'll have to run my script on the CI to make sure.  I'll stuff it in my own repo and do that now.

linuxrocks123: @erman-gurses so, problems with the manifest, here:

https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-llvm-project/23316920001-linux/manifests/gfx94X-dcgpu/therock_manifest.json

versus the script output, here:

https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-llvm-project/23316920001-linux/index-gfx94X-dcgpu.html

The manifest:
1.  Claims that patches are applied which in fact do not appear to be applied, and
2.  Refers to a TheRock SHA that is internal to the CI's local git checkout and is not pushed anywhere

erman-gurses: Thank you -  I will take a look at it in this week and get back to you.

PR #4059, opened 2026-03-19
Extend Python packages and PyTorch wheels to multi-arch Windows CI

marbre:
Mirrors the Linux multi-arch CI patterns to build ROCm Python packages and PyTorch in multi-arch Windows CI. Further fixes an inverted conditional on Linux.

### DIFF
--- a/.github/workflows/multi_arch_ci.yml
+++ b/.github/workflows/multi_arch_ci.yml
@@ -125,6 +125,7 @@ jobs:
       baseline_run_id: ${{ inputs.baseline_run_id }}
       rocm_package_version: ${{ needs.setup.outputs.rocm_package_version }}
       test_type: ${{ needs.setup.outputs.test_type }}
+      build_pytorch: ${{ matrix.variant.build_pytorch == true }}
     permissions:
       contents: read
       id-token: write


.github/workflows/multi_arch_ci_windows.yml
```diff
@@ -131,7 +134,45 @@ jobs:
...
+      id-token: write
+
+  # NOTE: Per-family PyTorch builds — each family gets its own wheel built
+  # against that family's ROCm package index slice.
+  # TODO(future): Consider adding a combined multi-arch pytorch build job
+  # that builds once against the full multi-arch index (amdgpu_families=all).
+  build_pytorch_wheels_per_family:
+    needs: [build_python_packages]
+    name: Build PyTorch | ${{ matrix.family_info.amdgpu_family }}
+    if: ${{ !failure() && !cancelled() && inputs.build_pytorch == false }}
```
ScottTodd: Oh 🤦 I typo'd this when editing Linux from
```yml
    if: >-
      ${{
        !failure() &&
        !cancelled() &&
        (
          inputs.use_prebuilt_artifacts == 'false' ||
          inputs.use_prebuilt_artifacts == 'true'
        ) &&
        inputs.expect_failure == false &&
        inputs.build_pytorch == true
      }}
```

This condition should be inverted. Should be checking `build_pytorch == true` here and in the Linux workflow.

We still need ` if: ${{ !failure() && !cancelled()` in case `build_python_packages` is _skipped_

.github/workflows/multi_arch_ci_windows.yml
```diff
@@ -36,6 +36,9 @@ on:
         type: string
       test_type:
         type: string
+      build_pytorch:
+        type: boolean
+        default: false
```
ScottTodd, line 41: Missing plumbing for this in https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml , as Linux has: https://github.com/ROCm/TheRock/blob/070488fe3dbbf409546712c4c6846d53f983ce79/.github/workflows/multi_arch_ci.yml#L96

marbre: Good catch! It went through with the inverted conditional however :wink:  See https://github.com/ROCm/TheRock/actions/runs/23292750908/job/67782767124

ScottTodd: Yeah both Windows _and_ Linux ran with the inverted condition... scary. My "multi-arch configure" feature branch has different plumbing for this... hopefully safer + easier to validate 🤞

marbre: @ScottTodd do you think we need to trigger another full test run or is the previous run (without plumbing but inverted conditional) enough signal?

ScottTodd: It's worth watching that correcting the condition doesn't cause pytorch builds to be skipped when we want them to run. Other than that (which I'm still sketched out by), fine to merge without waiting for more complete signal.

marbre:
All builds get triggered as expected.

PR #4074, opened 2026-03-19
Enable SPIRV for Comgr builds by default

lamb-j:
Comgr will still disable SPIRV APIs and Actions if the SPIRV-LLVM-Translator headers are not found.

Also setting in two different place before conflicted


mgcarrasco:
LGTM

PR #4085, opened 2026-03-20
Convert rocshmem to target-neutral build

marbre:
rocshmem is a communication library which `rocshmem_info` embeds
device code using `.hip_fatbin` / `.hip_gpubin_*` ELF sections
rather than the `__CudaFatBinaryWrapper`-based format that produces
a `.hipFatBinSegment` section.

When building with `THEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON`, the
artifact splitter finds `rocshmem_info` as a fat binary and attempts
kpack injection, which requires a `.hipFatBinSegment` wrapper table
to redirect device code loading at runtime. Since that section is
absent, injection fails with `RuntimeError: .hipFatBinSegment section
not found`.

One approach to resolve this is to move rocshmem to a target-neutral
build, building for all supported targets via `USE_TEST_AMDGPU_TARGETS`
so that each per-arch CI job uploads an identical artifact (making
upload races harmless).

Changes:
- Add `USE_TEST_AMDGPU_TARGETS` and `TARGET_NEUTRAL` to rocshmem
  subproject
- Move rocshmem artifact to `target-neutral` in comm-libs group

rccl remains target-specific in the existing comm-libs per-arch stage.

## Build: Size & Time Comparison

Configured with `gfx1201`, `THEROCK_ENABLE_CORE/PROFILER/COMM_LIBS/
ROCSHMEM=ON,` ccache disabled for rocshmem step.

|                        | target-specific | target-neutral |
|------------------------|-----------------|----------------|
| GPU targets compiled   | gfx1201 only    | all test archs |
| Build time (no ccache) | ~14s            | ~4m41s         |
| `librocshmem.a` size   | 5.9 MB          | 110 MB         |

### DIFF
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -676,6 +676,9 @@ artifact_deps = ["core-runtime", "core-hip", "hipify", "rocprofiler-sdk", "core-
 disable_platforms = ["windows"]
 
 [artifacts.rocshmem]
+# target-neutral: rocshmem embeds device code via .hip_fatbin ELF sections
+# rather than the __CudaFatBinaryWrapper format (.hipFatBinSegment), so kpack
+# injection is not supported. Built for all targets to make upload races harmless.
 artifact_group = "comm-libs"
 type = "target-neutral"
 artifact_deps = ["core-runtime", "core-hip", "sysdeps"]
--- a/comm-libs/CMakeLists.txt
+++ b/comm-libs/CMakeLists.txt
@@ -118,6 +118,9 @@ if(THEROCK_ENABLE_ROCSHMEM)
   ##############################################################################
   set(_rocshmem_subproject_names)
   therock_cmake_subproject_declare(rocshmem
+    # rocshmem uses .hip_fatbin/.hip_gpubin_* ELF sections for device code
+    # instead of .hipFatBinSegment, making it incompatible with kpack injection.
+    # Build all GPU targets in a single target-neutral artifact as a workaround.
     USE_TEST_AMDGPU_TARGETS
     EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR}/projects/rocshmem"
     BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/rocshmem"


marbre:
> Thanks. I'd like to try a similar thing with RCCL, which is the last component that lacks a convergent build. But in that case, we would leave the artifact as target specific but switch the build to all targets. It might be fine because most of the RCCL compilation time is a single threaded lto step, so as long as the builder has more cores than gfx architectures, it shouldn't cost much more to build them all.

So you're okay with the bloating static library? This is in the dev artifact and users are expected to link this into their applications but it might be fine if the linker is able to drop unused code.

For RCCL, let me check in a follow up. Will follow up regarding 
>we would leave the artifact as target specific but switch the build to all targets

with you.

stellaraccident:
> > Thanks. I'd like to try a similar thing with RCCL, which is the last component that lacks a convergent build. But in that case, we would leave the artifact as target specific but switch the build to all targets. It might be fine because most of the RCCL compilation time is a single threaded lto step, so as long as the builder has more cores than gfx architectures, it shouldn't cost much more to build them all.
> 
> So you're okay with the bloating static library? This is in the dev artifact and users are expected to link this into their applications but it might be fine if the linker is able to drop unused code.
> 
> For RCCL, let me check in a follow up. Will follow up regarding
> 
> > we would leave the artifact as target specific but switch the build to all targets
> 
> with you.

I don't love the bloat, but I'm fine with it to get this unblocked. The bigger issue is that how this is done will taint any consumer of it into being non convergent. I need to reapproach architecturally with the rocshmem team before that happens.

BUILD_TOPOLOGY.toml
```diff
@@ -677,7 +677,7 @@ disable_platforms = ["windows"]
 
 [artifacts.rocshmem]
 artifact_group = "comm-libs"
-type = "target-specific"
+type = "target-neutral"
```
ScottTodd, line 683: Given how this is not the ideal end state, how about putting a comment here to highlight the tradeoffs?

Here's what my review agent suggested:

```
# target-neutral: rocshmem embeds device code via .hip_fatbin ELF sections
# rather than the __CudaFatBinaryWrapper format (.hipFatBinSegment), so kpack
# injection is not supported. Built for all targets to make upload races harmless.
# Ideally this would move to its own generic stage to avoid redundant builds.
```

comm-libs/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ if(THEROCK_ENABLE_ROCSHMEM)
   ##############################################################################
   set(_rocshmem_subproject_names)
   therock_cmake_subproject_declare(rocshmem
+    USE_TEST_AMDGPU_TARGETS
```
ScottTodd, line 124: See my other comment - how about adding an explanation here?

Suggested comment (per my review agent):

```suggestion
therock_cmake_subproject_declare(rocshmem
    # rocshmem uses .hip_fatbin/.hip_gpubin_* ELF sections for device code
    # instead of .hipFatBinSegment, making it incompatible with kpack injection.
    # Build all GPU targets in a single target-neutral artifact as a workaround.
    USE_TEST_AMDGPU_TARGETS
```

BUILD_TOPOLOGY.toml
marbre: I had moved it to a separate build stage but this would not avoid redundant builds in non-multi-arch CI. Furthermore, the cost of downloading the compiler + cloning rocm-systems is not super low which is why we decided to not yet move hipDNN to its own stage. Same argument applies here. Anyway, I will copy the suggested comment, skipping the last line.

comm-libs/CMakeLists.txt
marbre: Done.

PR #4088, opened 2026-03-20
Add HTTPBackend for read-only artifact downloads from workflow summary index.html

PeterCDMcLean:
Implement HTTPBackend to enable downloading prebuilt artifacts
  from workflow summary HTML index files. Supports S3-hosted summaries with
  configurable base URL defaulting to therock-dev-artifacts.s3.amazonaws.com.

  Key features:
  - Read-only access with SHA256 checksum verification
  - HTML index parsing for artifact discovery
  - Environment-based configuration via THEROCK_WORKFLOW_RUN_ID

  Updated create_backend_from_env() to prioritize HTTPBackend
  when THEROCK_WORKFLOW_RUN_ID is set.

## Motivation

Not all tooling has access to S3, but everyone will have access to the hosted HTTP index.html. This provides a read-only backend so artifact_manager.py can be used to fetch artifacts/fprints

## Technical Details

Added HTTPBackend to artifact_backends.py

## Test Plan

Manual + Claude testing. Added new unit tests

## Test Result

It was able to download and parse the index.html from each family. Tested with both public TheRock artifacts base url as well as internal case url

## Submission Checklist

- [x] pre-commit run

### DIFF
--- a/build_tools/_therock_utils/artifact_backend.py
+++ b/build_tools/_therock_utils/artifact_backend.py
@@ -670,7 +670,4 @@ def create_backend_from_env(
     output_root = WorkflowOutputRoot.from_workflow_run(
         run_id=run_id, platform=platform_name
     )
-    return HTTPBackend(
-        output_root=output_root,
-        gfx_families=targets
-    )
+    return HTTPBackend(output_root=output_root, gfx_families=targets)
--- a/build_tools/_therock_utils/artifact_backend.py
+++ b/build_tools/_therock_utils/artifact_backend.py
@@ -580,7 +580,6 @@ class HTTPBackend(ArtifactBackend):
             # Artifacts are allowed to be downloaded without checksums
             pass
 
-
     def upload_artifact(self, source_path: Path, artifact_key: str) -> None:
         """Upload is not supported - HTTPBackend is read-only."""
         raise NotImplementedError("HTTPBackend is read-only")
--- a/build_tools/_therock_utils/artifact_backend.py
+++ b/build_tools/_therock_utils/artifact_backend.py
@@ -13,9 +13,10 @@ modules manage S3 clients and local directory mirroring. ArtifactBackend has
 download/list/exists operations that StorageBackend doesn't have yet.
 
 Environment-based switching:
-- THEROCK_HTTP_RUN_ID set → use HTTPBackend (read-only)
 - THEROCK_LOCAL_STAGING_DIR set → use LocalDirectoryBackend
-- Otherwise → use S3Backend
+- AWS credentials present → use S3Backend
+- THEROCK_AMDGPU_FAMILIES set → use HTTPBackend (read-only via public HTTPS URLs)
+- Otherwise → use S3Backend (unsigned requests for public buckets)
 """
 
 from abc import ABC, abstractmethod
@@ -26,8 +27,10 @@ import hashlib
 import os
 import re
 import shutil
+import urllib.error
 import urllib.request
 
+from .storage_location import StorageLocation
 from .workflow_outputs import WorkflowOutputRoot
 
 
@@ -339,18 +342,20 @@ class S3Backend(ArtifactBackend):
 class HTTPBackend(ArtifactBackend):
     """Read-only backend for HTTP artifact server.
 
-    Accesses artifacts from:
-    {base_url}/{run_id}-{platform}/
+    Uses WorkflowOutputRoot.https_url for all artifact downloads, which automatically
+    constructs public HTTPS URLs for S3 buckets (e.g., https://bucket.s3.amazonaws.com/path).
 
-    Artifacts are organized by gfx target via index-{gfx_family}.html files.
+    Uses WorkflowOutputRoot for path computation, ensuring consistent handling
+    of fork prefixes and bucket selection.
 
     Usage:
         # Create backend for specific run
+        output_root = WorkflowOutputRoot.from_workflow_run(
+            run_id="23309603946", platform="linux"
+        )
         backend = HTTPBackend(
-            run_id="23309603946",
-            platform="linux",
+            output_root=output_root,
             gfx_families=["gfx94X-dcgpu", "gfx1200"],
-            base_url="https://example.com/artifacts"
         )
 
         # List available artifacts (across all specified GFX families)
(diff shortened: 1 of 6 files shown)


build_tools/_therock_utils/artifact_backend.py
```diff
@@ -330,28 +336,288 @@ def artifact_exists(self, artifact_key: str) -> bool:
...
+        self.run_id = run_id
+        self.base_url = base_url
+        self.platform = platform
+        self.gfx_families = gfx_families
+        self._artifact_cache: Optional[List[str]] = None  # Lazy-loaded artifact list
+
+    @property
+    def base_uri(self) -> str:
+        """Return the base URI for this backend."""
+        return f"{self.base_url}/{self.run_id}-{self.platform}"
```
ScottTodd: See the docs at https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md , which we should also update as part of this PR

I'd prefer to treat this URI (specifically how the platform and run id are joined) as more of an implementation detail. See what `S3Backend` does: https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/_therock_utils/artifact_backend.py#L256-L258

And this code:
* https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/_therock_utils/workflow_outputs.py#L88-L94
* https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/_therock_utils/storage_location.py#L42-L45

It matters since the base URL is also dependent on the repository (TheRock, rocm-libraries, a fork, etc.) too.

See concrete suggestions here: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md#3-artifact_backendpy--httpbackend-should-use-workflowoutputroot-for-path-construction

ScottTodd (commented):
I ran out of time to do a detailed review myself today. I flagged one thing that stood out to me, and my code review agent spotted a few other areas to look into as well: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md

build_tools/_therock_utils/artifact_backend.py
PeterCDMcLean: Thank you for the feedback, Scott. I incorporated using WorkflowOutputRoot, but I felt I had to make a bit of an ugly hack within StorageLocation. There's no structured way to make a substitution of the base {bucket}.s3.amazon.com base URL without threading through lots of base URL schemas through to StorageLocation. I'm open to suggestions for alternatives. Internally, I am asking some devops whether we can have a domain that fits this schema rather than needing something special like this.

PeterCDMcLean: Would you prefer that commands like `post_build_upload.py`,  `artifact_manager.py` and others took a `--https_url_schema` argument that looks like "--https_url_schema "https://{bucket}.s3.amazon.com/sub/path", and the https_url_schema was passed through the API to StorageLocation where StorageLocation would do:

```
    https_url_schema : str
    
    @property
    def https_url(self) -> str:
        """Public HTTPS URL for browser access."""
        return 
        return self.https_url_schema.format({"bucket": self.bucket}).rstrip("/") + "/" + self.relative_path
```

PeterCDMcLean: I made a [stacked pr](https://github.com/ROCm/TheRock/pull/4167) onto this one where I implemented the schema. Feel free to merge that PR to this branch if you prefer that approach to env vars

ScottTodd: Revised review after some consideration and comparison between this PR and the stacked PR: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md

See the section at the top in particular: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md#design-direction-base_url-on-httpbackend. More specifically, I think the direction here is cleanest: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md#recommended-approach-base_url-on-httpbackend

Some key points:
* Environment variable overrides are tricky as they operate process-wide. We have cases where a workflow wants to _fetch_ from one bucket/location but _push_ or _copy_ to another
* Some of our S3 buckets have CDNs (currently using CloudFront) whose names and paths do not necessarily map directly to a S3 bucket name or consistent schema. See https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#bucket-inventory. Even the fully public buckets are still sometimes worth accessing via the CDNs instead of via `boto3` or other AWS APIs/tools/interfaces
  * I've been considering adding more configuration into `StorageLocation` so `https_url` isn't always the 1:1 mapping from S3 URI to public S3 HTTPS URL. We could offer a way to inject custom configurations and then override the bucket selection for a general solution. Claude rejected that idea during discussion: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4088.md#alt-1-bucket_override-with-default-mapping, but I think we could still make it work :P

PeterCDMcLean: - Agreed about environment variables. I will say that typically in one workflow or another there is some typical 'global' mapping or schema, but if you want to copy s3 artifacts from one schema set (Say public TheRock) to another (something internal/external) then environment variables are as you say not a solution.

- A couple points about the review .md:

> Can't express CDN URLs: Templates use {bucket} and {path} placeholders, but CDN URLs like https://rocm.devreleases.amd.com/v2/{path} have no relationship to the bucket name.

This is missing the point, because if the schema doesn't want to use the bucket it can just leave it out, 'https://rocm.devreleases.amd.com/v2/{path}' is a valid schema. This is the whole purpose of the schema. This would effectively allow genesis and other CDN base paths to be used in artifact_manager or other scripts.

> The s3_url_schema customization has almost no consumers — only base_uri display and storage_backend.py uploads, neither of which needs customization.

There is a use case internally which is not part of this commit. There is a prefix that gets applied to only s3 buckets because these s3 buckets don't follow the typical naming:
instead of 

`s3://{bucket}/{RepoOwner}-{RepoName}/{RunId}-{Platform}`

it is

`s3://{bucket}/v3/artifacts/{RepoOwner}-{RepoName}/{RunId}-{Platform}`

The 'v3/artifacts' cannot go in the prefix, because the prefix is applied to https URLs too.

The internal use case is another reason I added the bucket schema, because the naming is has a different 'RELEASE_TYPE' that needs a patch. In general, I've been hoping to reduce the patches that the npi meta repo applies to TheRock to match the infrastructure / s3 buckets / url schemes.

> I've been considering adding more configuration into StorageLocation so https_url isn't always the 1:1 mapping from S3 URI to public S3 HTTPS URL. We could offer a way to inject custom configurations and then override the bucket selection for a general solution.

This is more or less what the schema PR is about, but my pr is a little ham-fisted and a lot of function parameters. What did you have in mind? Part of the complexity here is that the URLs are composed in several different areas, which is why there are so many schemas as opposed to one https schema and one s3. What if there was one single data structure that had the schemas for bucket, s3_root, s3 prefix, https_root, https_prefix? That would clean up the many _schemas in function calls, but would mean that supplying a different schema would require setting each (ex: s3,http,bucket) even if only one type is used. Again, I'd like to know what you are thinking.

PeterCDMcLean: I've prototyped a different [stacked pr 'B'](https://github.com/ROCm/TheRock/pull/4199) which has the consolidated 'StorageConfig' which holds the schemas.

ScottTodd (commented):
Marking as "reviewed" to drop this old PR from my review queue. LMK if you want to continue this work in some form (and then please request review from another developer first, I don't have bandwidth to do so many first level reviews)

PR #2783, opened 2026-01-06
Add Python 3.14 support for PyTorch builds

subodh-dubey-amd:
## Summary

Adds Python 3.14 support for PyTorch builds on both Linux and Windows platforms.

> **Depends on #3535 and #XXXX** — both must merge first.

## Python 3.14 Support Matrix

| PyTorch Version | Python 3.14 Support |
|-----------------|---------------------|
| 2.7 | ❌ Not supported (no upstream support) |
| 2.8 | ❌ Not supported (no upstream support) |
| 2.9 | ✅ Preview support |
| nightly | ✅ Full support |

## Changes

- Updated `.github/workflows/release_portable_linux_pytorch_wheels.yml` to include Python 3.14 for PyTorch 2.9 and nightly
- Updated `.github/workflows/release_windows_pytorch_wheels.yml` to include Python 3.14 for PyTorch 2.9 and nightly
- Updated `RELEASES.md` to document Python 3.14 support (PyTorch 2.9+ only)
- Updated `build_tools/third_party/s3_management/update_dependencies.py` to update s3 index for pip

## Motivation

https://github.com/ROCm/TheRock/issues/2640: Add Python 3.14 support for PyTorch

## References

- https://github.com/pytorch/pytorch/issues/156856
- https://dev-discuss.pytorch.org/t/torch-compile-support-for-python-3-14-completed/3276

## Tests:

**Release portable Linux PyTorch Wheels Tests:** 
- https://github.com/ROCm/TheRock/actions/runs/22221004091

**Release Windows PyTorch Wheels:** 
- https://github.com/ROCm/TheRock/actions/runs/22133299032

## Related

- Fixes #2640
- Python 3.14 support for PyTorch: pytorch/pytorch#156856
- https://dev-discuss.pytorch.org/t/torch-compile-support-for-python-3-14-completed/3276

## Submission Checklist

- [ ] Rebased on `main` after #3535 is merged
- [ ] All workflow runs pass (Linux + Windows)
- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

Issue #2640, opened 2025-12-18
Add python 3.14 support for PyTorch


ScottTodd:
Python 3.14 was released on October 7, 2025 but we do not yet publish packages for it and it is already the default Python version on Fedora and other operating systems.

References:
* https://www.python.org/downloads/release/python-3140/
* https://docs.python.org/3/whatsnew/3.14.html

Related issues:
* https://github.com/pytorch/pytorch/issues/156856
  * We generally track what PyTorch publishes
* https://github.com/ROCm/TheRock/issues/2449
  * https://github.com/ROCm/TheRock/pull/2422
  * https://github.com/ROCm/TheRock/pull/2580
  * https://github.com/ROCm/TheRock/issues/2463

Code:
* https://github.com/ROCm/TheRock/blob/41b9e775d1b50fb1accc3cbf4e49dadcde4e13ea/RELEASES.md?plain=1#L41
* https://github.com/ROCm/TheRock/blob/41b9e775d1b50fb1accc3cbf4e49dadcde4e13ea/.github/workflows/release_portable_linux_pytorch_wheels.yml#L90
* https://github.com/ROCm/TheRock/blob/41b9e775d1b50fb1accc3cbf4e49dadcde4e13ea/.github/workflows/release_windows_pytorch_wheels.yml#L90

ScottTodd:
The error messages from Python when a package is not supported on a given Python version can be confusing:
```
D:\scratch\therock
λ py -V:3.14 -m venv 3.14.venv && .\3.14.venv\Scripts\activate.bat

D:\scratch\therock
(3.14.venv) λ pip install --index-url=https://rocm.nightlies.amd.com/v2-staging/gfx110X-all/ torch
Looking in indexes: https://rocm.nightlies.amd.com/v2-staging/gfx110X-all/
ERROR: Could not find a version that satisfies the requirement torch (from versions: none)
ERROR: No matching distribution found for torch
```

Users are confused too: https://old.reddit.com/r/ROCm/comments/1pkxihc/amd_radeon_rx_9070_xt_not_a_supported_wheel_on/. There they were following https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installrad/windows/install-pytorch.html, which uses an explicit wheel version (see `cp312` in the wheel names and the "To install the following wheels, Python 3.12 must be installed." text at the top of the page):
```
[...]

PR #4090, opened 2026-03-20
Switch wheel filtering from blacklist to whitelist

marbre:
The old approach used 13 substring checks to reject unwanted platforms and Python versions. This is fragile: new PyPI variants (e.g. `cp315`, new ARM flavors, iOS) are admitted by default unless explicitly added.

Replace with is_wheel_allowed(), which parses the PEP 427 wheel stem and checks platform and Python tags against explicit allowlists:

- Platform: `linux_x86_64`, `win_amd64`, `any`, or `manylinux*_x86_64`
- Python: `cp310`-`cp313`, or `py3` (pure-Python)

The manylinux check uses startswith/endswith rather than a substring match to correctly reject `musllinux_*_x86_64` and `macosx_*_x86_64`, which the old blacklist handled via separate pattern checks.

Adds test cases covering allowed wheels, rejected platforms, rejected Python tags, and malformed filenames.

### DIFF
--- a/build_tools/third_party/s3_management/test_update_dependencies.py
+++ b/build_tools/third_party/s3_management/test_update_dependencies.py
@@ -25,7 +25,6 @@ from update_dependencies import is_wheel_allowed
         # pure-Python / platform-independent
         "sympy-1.13.0-py3-none-any.whl",
         "filelock-3.15.0-py3-none-any.whl",
-        "typing_extensions-4.12.0-py310-none-any.whl",
         # Windows x64 — was not excluded by the old blacklist
         "torch-2.3.0-cp312-cp312-win_amd64.whl",
         "numpy-2.0.0-cp310-cp310-win_amd64.whl",
@@ -85,6 +84,9 @@ def test_rejected_platform(pkg: str) -> None:
         "numpy-2.0.0-cp313t-cp313t-linux_x86_64.whl",
         "numpy-2.0.0-cp314-cp314-linux_x86_64.whl",
         "numpy-2.0.0-cp314t-cp314t-linux_x86_64.whl",
+        # Python 2 and py2.py3 universal tags
+        "six-1.16.0-py2-none-any.whl",
+        "six-1.16.0-py2.py3-none-any.whl",
     ],
 )
 def test_rejected_python(pkg: str) -> None:
--- a/build_tools/third_party/s3_management/update_dependencies.py
+++ b/build_tools/third_party/s3_management/update_dependencies.py
@@ -107,9 +107,9 @@ def is_wheel_allowed(pkg: str) -> bool:
     1. Platform tag is "linux_x86_64", "win_amd64", "any", or starts with
        "manylinux" and ends with "_x86_64" (e.g., "manylinux_2_17_x86_64").
        This rejects win32, win_arm64, macOS, musllinux, ARM, RISC-V, iOS, etc.
-    2. Python tag is in _ALLOWED_CPYTHON_TAGS, or starts with "py"
-       (pure-Python wheels such as "py3" or "py310").
-       This rejects PyPy (pp*), cp39, cp313t, cp314, cp314t, etc.
+    2. Python tag is in _ALLOWED_CPYTHON_TAGS, or is exactly "py3"
+       (pure-Python wheels). This rejects PyPy (pp*), cp39, cp313t,
+       cp314, cp314t, py2, py2.py3, etc.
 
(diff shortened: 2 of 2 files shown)


build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -76,6 +100,39 @@ def get_wheels_of_version(idx: Dict[str, str], version: str) -> Dict[str, str]:
...
+    if len(parts) < 5:
+        return False  # Malformed — skip rather than guess
+
+    platform_tag = parts[-1]
+    python_tag = parts[-3]
+
+    platform_ok = platform_tag in _ALLOWED_PLATFORM_TAGS or (
+        platform_tag.startswith("manylinux") and platform_tag.endswith("_x86_64")
+    )
+    python_ok = python_tag in _ALLOWED_CPYTHON_TAGS or python_tag.startswith("py")
```
erman-gurses: LGTM overall. One nit regarding the Python tag criteria:

Referring the current rule :

`Python tag is in _ALLOWED_CPYTHON_TAGS, or starts with py (pure-Python wheels such as py3 or py310). This rejects PyPy (pp*), cp39, cp313t, cp314, cp314t, etc.`

With the current implementation `python_tag.startswith("py")`, we will also admit tags like py2 or py2.py3, not just pure-Python py3-style wheels mentioned in the criteria.

Is that intentional here, or would it be safer to narrow this to py3 (and/or specific supported forms) to better align with the intended constraint?

marbre: I don't think it matters to much but addressed this.

build_tools/third_party/s3_management/test_update_dependencies.py
ScottTodd, line 1: As is, this unit test will not be run automatically:
* https://github.com/ROCm/TheRock/blob/1b0705f93ac5fa0d0326b4826d7f3e4ccd502f9f/.github/workflows/unit_tests.yml#L40-L49
* https://github.com/ROCm/TheRock/blob/1b0705f93ac5fa0d0326b4826d7f3e4ccd502f9f/build_tools/pyproject.toml#L1-L7

Can you move this file into https://github.com/ROCm/TheRock/tree/main/build_tools/tests and match the `{module_name}_test.py` naming?

We enforce that naming via pre-commit:
https://github.com/ROCm/TheRock/blob/1b0705f93ac5fa0d0326b4826d7f3e4ccd502f9f/.pre-commit-config.yaml#L53-L59

I can see wanting to keep the test near the code, especially for "third party" code, so another option is to expand the scope of those files to include this directory too.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -20,6 +20,30 @@
...
+# Pure-Python wheels (python tag starting with "py") are also allowed
+# regardless of version — they carry no CPython ABI dependency.
+_ALLOWED_CPYTHON_TAGS: frozenset[str] = frozenset(
+    {
+        "cp310",
+        "cp311",
+        "cp312",
+        "cp313",
+    }
+)
```
ScottTodd, line 45: Should we also allow cp314, or handle that as part of https://github.com/ROCm/TheRock/issues/2640 ? I just noticed was closed prematurely... we really should have that soaking in nightly releases by now, Python 3.14 was released on October 7 and PyTorch upstream has supported it for over 1 month now too.

build_tools/third_party/s3_management/test_update_dependencies.py
marbre: I could but it would mix up the layering as this is currently in a `third_party` directory as you also noticed. I am aware that therefore we don't run unit tests and would suggest to rather expand the scope of the tests to include the directory. I would like to see that this is addressed by someone else than me in a follow up and can file an issue for this.

build_tools/third_party/s3_management/update_dependencies.py
marbre: No super strong feelings here and I had hopped that Python 3.14 would long been resolved but unfortunately it isn't. I'd rather leave this as part of #2640 / #2783 as we otherwise already upload deps for 3.14 otherwise which might raise expectations. Seeing how slowly #2783 moved I am however not in favor to do so.

build_tools/third_party/s3_management/test_update_dependencies.py
marbre: Filed #4105 and assigned it to @cgoea as it is a good first issue :)

PR #4101, opened 2026-03-22
 set submodule pointer for weekly amd-compiler-2026-03-19 try3 

ronlieb:
Set submodule pointer for weekly amd-compiler-2026-03-19 try3

     llvm   3d881804bbd3
     hipify 774da5d09
     spirv  eed2022f42

Remove 0010-Comgr-Add-COMGR_STATIC_LLVM-option-for-static-LLVM-l.patch
Add 0007-CK-lifetime-override-Wno-lifetime-safety-intra-tu-su.patch
Remove 0011-Comgr-Fix-memory-leak-when-assigning-anon-bit-rm-w12.patch


ronlieb:
This PR has a patch to disable by default lifetime warnings so as to get CK buildable , then we can test ww12 to look for blocking issues.

This PR cannot land with the lifetime patch present. Must be removed first.

Issue #4034, opened 2026-03-18
[Issue] rocprofiler-compute test_profile_pc_sampling Fails in CI Due to Unsupported PC Sampling on CI Agents


chiranjeevipattigidi:
### Summary:
Recent CI runs are failing due to an error in the PC sampling test. The test test_profile_pc_sampling (in tests/test_profile_general.py, specifically test_pc_sampling_stochastic) fails with the following error:

rocm-systems Bump PR: https://github.com/ROCm/TheRock/pull/3984

### Tests Failing:
```
test_profile_pc_sampling 
```

### log:
```
'test_pc_sampling_stochastic_output', 'ROCPROF_OUTPUT_FILE_NAME': 'ps_file', 'ROCPROFILER_PC_SAMPLING_BETA_ENABLED': '1', 'ROCPROF_PC_SAMPLING_UNIT': 'cycles', 'ROCPROF_PC_SAMPLING_INTERVAL': '1048576', 'ROCPROF_PC_SAMPLING_METHOD': 'stochastic'}
2026-03-18T05:11:12.7149331Z 11:   [34mDEBUG[0m pc sampling rocprof sdk user provided command: ['./tests/mat_mul_max']
2026-03-18T05:11:12.7726922Z 11:    [32mINFO[0m    |-> [rocprofiler-sdk] W20260318 05:11:12.768270 140560788107328 agent.cpp:470] Attempt to enable hip visiblity for agent-2 which is not visible to HSA (ROCR)
2026-03-18T05:11:12.8402803Z 11:    [32mINFO[0m    |-> [rocprofiler-sdk] E20260318 05:11:12.839683 140560788107328 tool.cpp:2005] Given PC sampling configuration is not supported on any of the agents
2026-03-18T05:11:12.8654751Z 11:   [31mERROR[0m [31mPC sampling failed.[0m
2026-03-18T05:11:12.8979918Z 11: F
2026-03-18T05:11:12.8980085Z 11: 
2026-03-18T05:11:12.8980310Z 11: =================================== FAILURES ===================================
2026-03-18T05:11:12.8980616Z 11: _________________________ test_pc_sampling_stochastic __________________________
2026-03-18T05:11:12.8980822Z 11: 
2026-03-18T05:11:12.8981044Z 11: binary_handler_profile_rocprof_compute = <function binary_handler_profile_rocprof_compute.<locals>._handler at 0x7fd7f7c620c0>
2026-03-18T05:11:12.8981308Z 11: 
[...]

PR #4110, opened 2026-03-23
[rocprofiler-compute] Disable test_profile_pc_sampling

jbonnell-amd:
## Motivation

#4034 
The `test_profile_pc_sampling` test consistently fails due to the CI runners not supporting the pc_sampling feature. This PR will disable this test from running.

## Technical Details

- Add `test_profile_pc_sampling` to `EXCLUDED_TESTS` list in `test_rocprofiler_compute.py`

## Test Plan

- Run a test run using this branch and confirm that the `test_profile_pc_sampling` test does not run

## Test Result

- https://github.com/ROCm/TheRock/actions/runs/23442786556
  - Tests pass, and the `test_profile_pc_sampling` test is not ran

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_rocprofiler_compute.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rocprofiler_compute.py
@@ -15,9 +15,7 @@ THEROCK_LIB_PATH = str(THEROCK_PATH / "lib")
 ROCPROFILER_COMPUTE_DIRECTORY = THEROCK_PATH / "libexec" / "rocprofiler-compute"
 
 # Set up excluded tests
-EXCLUDED_TESTS = [
-    "test_profile_live_attach_detach",
-]
+EXCLUDED_TESTS = ["test_profile_live_attach_detach", "test_profile_pc_sampling"]
 
 # quick Tests
 QUICK_TESTS = [


vedithal-amd:
I think it is only the `test_pc_sampling_stochastic` pytest function which is problematic.

Excluding the entire ctest suite for pc sampling is not ideal and can cause to gaps and less coverage in our testing

Better to create a PR for rocprofiler-compute and mark this stochastic pc sampling test function as skipped

vedithal-amd:
Please close this against https://github.com/ROCm/rocm-systems/pull/4387

build_tools/github_actions/test_executable_scripts/test_rocprofiler_compute.py
```diff
@@ -15,9 +15,7 @@
 ROCPROFILER_COMPUTE_DIRECTORY = THEROCK_PATH / "libexec" / "rocprofiler-compute"
 
 # Set up excluded tests
-EXCLUDED_TESTS = [
-    "test_profile_live_attach_detach",
-]
+EXCLUDED_TESTS = ["test_profile_live_attach_detach", "test_profile_pc_sampling"]
```
geomin12, line 18: can we follow a format like so: 

https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/github_actions/test_executable_scripts/test_hiptests.py#L43-L77

we are doing some code gen to auto-handle this shortly, but also some tests may pass for some archs and not for others too

jbonnell-amd:
Closing this PR since https://github.com/ROCm/rocm-systems/pull/4387 disables the test on unsupported machines/runners

PR #4112, opened 2026-03-23
[debug-tools/rocgdb] Bump test timeout from 30 to 45 minutes

lumachad:
I spotted rocgdb tests timing out after 30 minutes. It looks like we are dangerously close to reaching the 30 minutes timeout if we manage to get allocated a slower machine.

Bump it to 45 minutes for extra safety. We don't want to make it too big since that may take too long to fail in case the tests misbehave.


lumachad:
Using the ci:run-multi-arch label for testing purposes.

PR #4114, opened 2026-03-23
Add automated IREE libs submodule bump workflow

AaronStGeorge:
## Motivation

Adds a daily workflow that bumps `iree-libs/iree` and `iree-libs/fusilli` submodules.
## Technical Details

IREE is pinned to the version tag declared in fusilli's [`version.json`](https://github.com/iree-org/fusilli/blob/main/version.json), and fusilli is bumped to HEAD of main.

## Test Plan

Tested in fork

## Test Result

[link GitHub Action run](https://github.com/AaronStGeorge/TheRock/actions/runs/23440330718) and [resulting PR](https://github.com/AaronStGeorge/TheRock/pull/2).

## Submission Checklist


.github/workflows/bump_iree_libs.yml
```diff
@@ -0,0 +1,93 @@
...
+        run: python3 build_tools/github_actions/bump_iree_libs.py
+
+      # App token so the created PR triggers CI (GITHUB_TOKEN cannot).
+      - name: Generate GitHub App token
+        if: env.VERSIONS_CHANGED == 'true'
+        uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
+        id: generate-token
+        with:
+          app-id: ${{ secrets.IREE_LIBS_BUMP_APP_ID }}
+          private-key: ${{ secrets.IREE_LIBS_BUMP_APP_PRIVATE_KEY }}
```
AaronStGeorge: We'll need a GitHub App token, the default `GITHUB_TOKEN` is not allowed to trigger CI. Taking a look around `TheRock` I don't see an existing app we can use so I think we'll need to create one. Below is my understanding of how that's done (taken form https://github.com/iree-org/fusilli/pull/207#discussion_r2869135540).

--- 

### 1. Create the app under ROCm

Go to: https://github.com/organizations/ROCm/settings/apps/new

Fill in:
- **GitHub App name**: `iree-libs-bump-deps` (or similar)
- **Homepage URL**: `https://github.com/ROCm/TheRock`
- **Webhook**: uncheck **Active** (no webhook needed)
- **Permissions** (Repository permissions only):
  - **Contents**: Read & Write
  - **Pull requests**: Read & Write
- **Where can this GitHub App be installed?**: Only on this account

Click **Create GitHub App**.

### 2. Note the App ID

On the app settings page, copy the numeric **App ID** shown near the top.

### 3. Generate a private key

On the same settings page, scroll to **Private keys** and click **Generate a private key**.
A `.pem` file will download — keep it secure.

### 4. Install the app on TheRock

Go to the app's **Install App** tab, select `ROCm`, and restrict access to only
the `TheRock` repository.

### 5. Add secrets to the TheRock repo

Go to: https://github.com/ROCm/TheRock/settings/secrets/actions

Add two repository secrets:

| Secret name | Value |
|-------------|-------|
| `IREE_LIBS_BUMP_APP_ID` | The numeric App ID from step 2 |
| `IREE_LIBS_BUMP_APP_PRIVATE_KEY` | The full contents of the `.pem` file from step 3 |

.github/workflows/bump_iree_libs.yml
ScottTodd, line 1: I'd rather we reuse the existing https://github.com/ROCm/TheRock/blob/main/.github/workflows/bump_submodules.yml instead of adding a new workflow.

build_tools/github_actions/bump_iree_libs.py
ScottTodd, line 1: Same here - extend https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/bump_automation.py instead?

AaronStGeorge: There's not a lot of common ground in what the two workflows are doing. The submodule script has some logic for opening PRs (stuff we get "for free" from `peter-evans` github action in the new workflow) but other than that the scripts are doing different jobs (IREE and fusilli need to be bumped together, rocm-libraries and systems can be bumped independently; there's no need to do any type of backwards patch into `rocm-libraries` when a bump merges for fusilli / iree).

We can't use `peter-evans` github action in the `rocm-libraries` workflow as it only handles a single PR.

They could be combined but it would have to create a fairly generic framework for handling events in the python script then have two instantiations of that that do reasonably disjoint things.

AaronStGeorge: Done in https://github.com/ROCm/TheRock/pull/4114/changes/5082c7a52c602fbec9738f63065039a6bfc105db

build_tools/github_actions/bump_automation.py
```diff
@@ -234,100 +246,289 @@ def create_therock_bump(submodule, token):
...
     parser.add_argument("--after")
     parser.add_argument("--systems_token", required=True)
     parser.add_argument("--libraries_token", required=True)
-    args = parser.parse_args()
+    args = parser.parse_args(argv)
+
+    tokens = {
+        "systems_token": args.systems_token,
+        "libraries_token": args.libraries_token,
+    }
```
AaronStGeorge, line 523: the new IREE uses systems token, if that's not OK we'll need a new GitHub App token. Below is my understanding of how that's done (taken form https://github.com/iree-org/fusilli/pull/207#discussion_r2869135540).

--- 

### 1. Create the app under ROCm

Go to: https://github.com/organizations/ROCm/settings/apps/new

Fill in:
- **GitHub App name**: `iree-libs-bump-deps` (or similar)
- **Homepage URL**: `https://github.com/ROCm/TheRock`
- **Webhook**: uncheck **Active** (no webhook needed)
- **Permissions** (Repository permissions only):
  - **Contents**: Read & Write
  - **Pull requests**: Read & Write
- **Where can this GitHub App be installed?**: Only on this account

Click **Create GitHub App**.

### 2. Note the App ID

On the app settings page, copy the numeric **App ID** shown near the top.

### 3. Generate a private key

On the same settings page, scroll to **Private keys** and click **Generate a private key**.
A `.pem` file will download — keep it secure.

### 4. Install the app on TheRock

Go to the app's **Install App** tab, select `ROCm`, and restrict access to only
the `TheRock` repository.

### 5. Add secrets to the TheRock repo

Go to: https://github.com/ROCm/TheRock/settings/secrets/actions

Add two repository secrets:

| Secret name | Value |
|-------------|-------|
| `IREE_LIBS_BUMP_APP_ID` | The numeric App ID from step 2 |
| `IREE_LIBS_BUMP_APP_PRIVATE_KEY` | The full contents of the `.pem` file from step 3 |

build_tools/github_actions/bump_iree_libs.py
AaronStGeorge: @ScottTodd I created generic handlers (`Bumpers`) with some re-used infrastructure for creating rolling PRs (moved away from existing pattern of opening multiple PRs then having an automated close of N PRs - update instead). I think this satisfies your original request, feel free to take a look!

AaronStGeorge:
@ammallya @rahulc-gh if you could review that would be much appreciated.

I updated your original script in a few ways:
- Created a `Bumper` class for organization, IREE libs bumps need some shared infrastructure but otherwise are pretty different from ROCm libraries/system bumps, generic class contains the concept of something that handles schedule and push events while allowing different jobs to be done by [`ROCmBumper`](https://github.com/ROCm/TheRock/pull/4114/changes#diff-37be2961e636171092bf47053fd8c78fb5ead81096b5c0f2d98dd12d82f159abR261) and [`IreeLibsBumper`](https://github.com/ROCm/TheRock/pull/4114/changes#diff-37be2961e636171092bf47053fd8c78fb5ead81096b5c0f2d98dd12d82f159abR371).
- Added IREE libs [bump automation](https://github.com/ROCm/TheRock/pull/4114/changes#diff-37be2961e636171092bf47053fd8c78fb5ead81096b5c0f2d98dd12d82f159abR371-R454).
- Moved to rolling PRs, if previous bump PR wasn't merged the script now _updates_ that PR rather than opening a new one. This allows us to remove the PR cleanup logic and IMHO reduces GitHub noise for reviewers. 
- Fixed missing [pip dep](https://github.com/ROCm/TheRock/pull/4114/changes#diff-b47c85e273f035487773b0b04275813a7623817e62c9c56cc4b11aa090f6d599R36-R37) that was [breaking the run](https://github.com/ROCm/TheRock/actions/workflows/bump_submodules.yml).
- Added action level logging through `GITHUB_STEP_SUMMARY` so we get a nice explanation of what the action accomplished [example test run](https://github.com/AaronStGeorge/TheRock/actions/runs/23495086203).

.github/workflows/bump_submodules.yml
```diff
@@ -45,12 +54,8 @@ jobs:
           owner: "ROCm"
           repositories: "rocm-libraries"
 
-      - name: Install prerequisites
-        run: |
-          pip install requests
-
       - name: Run scheduled bump automation
-        if: github.event_name == 'schedule'
+        if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
rahulc-gh, line 64: if we are adding workflow dispatch than we should add the capability for updating it for certain components only instead of defaulting it to all.

AaronStGeorge: We can just remove this too, I only added it for testing.

rahulc-gh:
@ScottTodd  or  @amd-aakash  - Do we want to combine IREE libs submodule update with rocm-sys and rocm-libs automation ?? They can stick to more stable dependabot automation for updates as they don't need to create multiple PRs or datapoints for IREE bump PRs.

AaronStGeorge:
@rahulc-gh with dependabot we could bump IREE and fusilli to current HEAD, but (correct me if I'm wrong) we couldn't accomplish the workflow we have right now. 

Fusilli stores the IREE it's tested and validated against in [version.json](https://github.com/iree-org/fusilli/blob/main/version.json), when bumping fusilli we would like to keep the IREE version in sync and update them atomically. In practice it's pretty common for new versions of IREE to break fusilli in some way, operationally we would have a nicer time if we took advantage of the testing fusilli has and only bumped to validated IREE versions. That might not be a viable path if other things in TheRock start depending on IREE, but for now that workflow saves some headaches.

IIUC dependeabot doesn't support this type of custom atomic bumping that we're looking for.

I'd be happy to split the workflows back up, that's what I had originally: https://github.com/ROCm/TheRock/pull/4114/changes/71fd8effa9373589a222f5c9ecc7717e1d2e0f57

build_tools/github_actions/bump_iree_libs.py
AaronStGeorge: @ScottTodd after discussing with @rahulc-gh and internal discussions I'm going to remove work don in https://github.com/ROCm/TheRock/commit/5082c7a52c602fbec9738f63065039a6bfc105db, and move forward with a separate workflow.

lumachad:
Hypothetically, if rocgdb also wanted to take advantage of such auto bumps, would we be able to reuse some of this code?

AaronStGeorge:
Fusilli has been de-prioritized, this is no longer relevant.

PR #4116, opened 2026-03-23
rocDecode Tests - Add third-party deps to enhance testing

kiritigowda:
## Motivation

rocdecode requires FFmpeg dev libraries (libavcodec-dev, libavformat-dev, libavutil-dev) at test build time. These are not bundled in TheRock artifacts, so a specialized container image is needed.

## Technical Details

**New Dockerfile** (`dockerfiles/no_rocm_image_ubuntu24_04_media.Dockerfile`)
- Extends `no_rocm_image_ubuntu24_04` with FFmpeg dev libraries required by rocdecode tests.

**Publish workflow** (`.github/workflows/publish_no_rocm_image_ubuntu24_04.yml`)
- Added `publish_no_rocm_image_ubuntu24_04_media` as a dependent job, consistent
  with how `rocgdb` and `ocl_rt` are handled. This ensures the media image is
  rebuilt whenever the base image changes.

**Test configuration** (`build_tools/github_actions/fetch_test_configurations.py`)
- `rocdecode`: Left a commented-out template with a TODO to pin
  to a `@sha256:...` digest after the media image is first published (matching
  the `rocgdb` pattern).

> [!NOTE]
> Once the media image is published to GHCR, update `rocdecode` in `fetch_test_configurations.py`:

## Test Plan

rocDecode Tests should add additional tests that uses ffmpeg

## Test Result

All new tests and existing tests should pass

## Submission Checklist


.github/workflows/publish_no_rocm_image_ubuntu24_04.yml
geomin12, line 1: instead of adding here, let's make it a separate docker file like so: https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04_rocgdb.Dockerfile

that way, we don't waste resources deploying both (as some updates to no_rocm_image_ubuntu_24_04 may not apply to here

kiritigowda: I would like to track the base image as changes to base might be important for our flow. This outweighs rebuilt with changes to base.

ScottTodd: The changes in this file look fine to me.

I did notice a potential (preexisting) issue though:
* https://github.com/ROCm/TheRock/blob/3843f9517da514a5f0c92c1bb9ce96e7366a2dc5/.github/workflows/publish_no_rocm_image_ubuntu24_04_rocgdb.yml#L20-L24
* https://github.com/ROCm/TheRock/blob/3843f9517da514a5f0c92c1bb9ce96e7366a2dc5/.github/workflows/publish_no_rocm_image_ubuntu24_04.yml#L28-L33

We have two workflows running on overlapping triggers (paths matching `dockerfiles/no_rocm_image_ubuntu24_04*.Dockerfile`) that both try to publish `no_rocm_image_ubuntu24_04_rocgdb` and `no_rocm_image_ubuntu24_04_ocl_rt`. They are likely to race and both try publishing slightly different images.

This PR only modifying `.github/workflows/publish_no_rocm_image_ubuntu24_04.yml` makes sense to me, though I'm not sure if `needs: [publish_no_rocm_image_ubuntu24_04]` is actually enough of a wait for `FROM ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest` to pull the image that was just built/published. I used some more sophisticated code for multi-stage docker builds on a prior project: https://github.com/iree-org/iree/tree/0f3bea22adf9ee1a7f6627254183d0094c60be61/build_tools/docker

kiritigowda: @geomin12  @ScottTodd - We are following the precedence set by OCL RT and rocGDB. We we merge these changes?

geomin12:
Landing as errors are unrelated and this is only relevant to docker updates

PR #4119, opened 2026-03-23
[hipDNN] Add hipdnn_list_engines utility to hipDNN install artifacts.

CMiservaAMD:
## Motivation

Utility to list which hipDNN plugins are available and which engines/engineIDs are available in those plugins.

## Technical Details

```
$ hipdnn_list_engines --plugin-dir /opt/rocm/lib/hipdnn_plugins/engines/
Loaded engines:
  HIPBLASLT_ENGINE (0x79AEB17432D4A22B)
    Plugin:  hipblaslt_plugin
    Version: 1.0.0
    Type:    HIPDNN_PLUGIN_TYPE_ENGINE
  MIOPEN_ENGINE (0x15B46865C717A122)
    Plugin:  miopen_provider_plugin
    Version: 1.0.0
    Type:    HIPDNN_PLUGIN_TYPE_ENGINE
  MIOPEN_ENGINE_DETERMINISTIC (0xA258541A6DAA1DE3)
    Plugin:  miopen_provider_plugin
    Version: 1.0.0
    Type:    HIPDNN_PLUGIN_TYPE_ENGINE
```

## Submission Checklist

### DIFF
--- a/.github/workflows/build_native_linux_packages.yml
+++ b/.github/workflows/build_native_linux_packages.yml
@@ -30,6 +30,11 @@ on:
         description: The type of release to build ("dev", "nightly", or "prerelease"). All developer-triggered jobs should use "dev"!
         required: false
         type: string
+      skip_package_upload:
+        description: Skip uploading packages to S3. Set to true when called from test workflows.
+        required: false
+        type: boolean
+        default: false
       repository:
           description: "Repository to checkout. Otherwise, defaults to `github.repository`."
           type: string
@@ -64,6 +69,10 @@ on:
         description: The type of release to build ("dev", "nightly", or "prerelease"). All developer-triggered jobs should use "dev"!
         type: string
         default: "dev"
+      skip_package_upload:
+        description: Skip uploading packages to S3. Set to true when called from test workflows.
+        type: boolean
+        default: false
       ref:
           description: "Branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow."
           type: string
@@ -76,7 +85,7 @@ run-name: Build native Linux packages (${{ inputs.artifact_group }}, ${{ inputs.
 
 jobs:
   build_native_packages:
-    name: Build Linux native Packages
+    name: Build Linux native Packages::${{ inputs.native_package_type }}::${{ inputs.artifact_group }}
     strategy:
       fail-fast: false
     runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
@@ -136,12 +145,14 @@ jobs:
             --version-suffix ${{ env.ARTIFACT_RUN_ID }}
 
       - name: Configure AWS Credentials
+        if: ${{ !inputs.skip_package_upload }}
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
         with:
           aws-region: us-east-2
           role-to-assume: arn:aws:iam::692859939525:role/therock-${{ inputs.release_type }}
 
       - name: Upload Package repo to S3
+        if: ${{ !inputs.skip_package_upload }}
         id: upload-packages
         run: |
           echo "Uploading to s3 bucket : ${{ inputs.release_type }}"
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -218,3 +218,36 @@ jobs:
     permissions:
       contents: read
       id-token: write
+
+  build_native_linux_packages:
+    needs: [build_portable_linux_artifacts]
+    name: Build Native Packages (${{ matrix.package_type }})
+    # Build native packages for CI validation without uploading to S3.
+    # Run if:
+    # - Build succeeded (or using prebuilt artifacts)
+    # - Not expecting failure
+    if: >-
+      ${{
+        !failure() &&
+        !cancelled() &&
+        (
+          inputs.use_prebuilt_artifacts == 'false' ||
+          inputs.use_prebuilt_artifacts == 'true'
+        ) &&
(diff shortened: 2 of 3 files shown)


build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
```diff
@@ -102,6 +102,78 @@ def run_tests(build_dir: Path):
...
+    if is_windows:
+        plugin_dir = artifacts_path / "bin" / "hipdnn_plugins" / "engines"
+    else:
+        plugin_dir = artifacts_path / "lib" / "hipdnn_plugins" / "engines"
+
+    # Expected engines that must be present in the output
+    expected_engines = [
+        "MIOPEN_ENGINE (0x15B46865C717A122)",
+        "MIOPEN_ENGINE_DETERMINISTIC (0xA258541A6DAA1DE3)",
+    ]
```
BrianHarrisonAMD: I don't think we will always have this installed if you were to build TheRock with MIOpen provider, or MIOpen off.

We probably need someway to flag what plugins are actually installed / expected to be found.
We would probably have to check what's in the plugin folder, or maybe we can just ignore that for now, and make this purely about validating no errors are encountered trying to load and check plugins.

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
```diff
@@ -102,6 +102,112 @@ def run_tests(build_dir: Path):
...
+    1. With explicit --plugin-dir argument
+    2. Without --plugin-dir (using internal default path)
+
+    If the miopen_plugin is present, both runs must contain the expected
+    engine strings. If the plugin is not present, only the return code is verified.
+    """
+    artifacts_path = Path(OUTPUT_ARTIFACTS_DIR).resolve()
+    is_windows = platform.system() == "Windows"
+
+    exe_suffix = ".exe" if is_windows else ""
```
SamuelReeder, line 122: This is duplicated with the above `run_tests`. It could be consolidated into an environment setup helper or global variables.

SamuelReeder: This test also doesn't setup the LD_LIBRARY_PATH, so I guess `hipdnn_list_engines` doesn't have a hipDNN runtime dep?

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
```diff
@@ -102,6 +102,112 @@ def run_tests(build_dir: Path):
...
+    if not has_miopen_plugin:
+        logging.info(
+            f"{miopen_plugin_name} not found in {plugin_dir}, "
+            "skipping engine output validation (will only verify return code)"
+        )
+
+    # Expected engines that must be present in the output (only checked if plugin exists)
+    expected_engines = [
+        "MIOPEN_ENGINE (0x15B46865C717A122)",
+        "MIOPEN_ENGINE_DETERMINISTIC (0xA258541A6DAA1DE3)",
```
SamuelReeder, line 149: Where does the hexadecimal representation come from? I'm probably just missing this.

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
```diff
@@ -102,6 +102,112 @@ def run_tests(build_dir: Path):
...
+            logging.error(f"stdout: {result.stdout}")
+            logging.error(f"stderr: {result.stderr}")
+            raise RuntimeError(
+                f"hipdnn_list_engines ({description}) exited with code {result.returncode}"
+            )
+
+        logging.info(f"hipdnn_list_engines ({description}) output:\n{result.stdout}")
+
+        # Only validate output content if miopen_plugin is present
+        if not has_miopen_plugin:
```
SamuelReeder, line 170: It would be nice if we could check whether the miopen plugin was supposed to be built, and only skip if not. I honestly doubt there's a way to determine this from the installed artifacts however, so it's probably fine as is.

SamuelReeder (approved):
I think this test looks good! Left a few comments with questions or improvement, but they're all minor.

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
CMiservaAMD: Is LD_LIBRARY_PATH required? I am expecting that TheRock sets the RUNPATH such that hipdnn_list_engines will be able to locate the hipdnn_backend.so relative to its install location.

Did you find this wasn't working when you were developing tests like this?

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
CMiservaAMD: It's just the engine ID -- the hash of the engine name, registered in https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipdnn/data_sdk/include/hipdnn_data_sdk/utilities/EngineNames.hpp#L147.

https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipdnn/tools/ListEngines.cpp#L78-L79

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
CMiservaAMD: Agreed. This will have to be a future improvement at this point though (assuming things don't fall apart with the test in its current form anyways).

build_tools/github_actions/test_executable_scripts/test_hipdnn_install.py
SamuelReeder: It might be fine to omit for the `hipdnn_list_engines` executable since it's built with TheRock. The above test built by the separate cmake configure lacked it, so I needed to manually set.

PR #4008, opened 2026-03-17
Use TEST_TYPE directly as categories for test filtering

dileepr1:
## Motivation
#3992 adds support for the test categories for test filtering.
Use them directly for selecting the test groups.

## Technical Details

Use the new categories - quick, standard, comprehensive and full - from TEST_TYPE variable instead of smoke, nightly etc.

## Test Plan

Make sure the tests run as usual

PR #4120, opened 2026-03-23
MIOpen used categories are quick and standard not full

aalabsi-amd:
## Motivation

With recent change in this PR https://github.com/ROCm/TheRock/pull/4008/changes, the test runner started running full as the category for miopen, while miopen is using standard.
This caused tests that were not previously run to be executed and fail to cause the bumps in this issue to fail.
https://github.com/ROCm/rocm-libraries/issues/5707

## Technical Details

This PR makes the category for miopen standard
## Test Plan

Full tests should not be selected, only standard.
## Test Result

Standard tests are run and passed
## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_runner.py
+++ b/build_tools/github_actions/test_executable_scripts/test_runner.py
@@ -221,7 +221,12 @@ def build_ctest_command(category, gpu_arch, available_gpu_archs):
 
 
 def main():
-    category = TEST_TYPE.lower() if TEST_TYPE else "quick"
+    # Determine test category based on TEST_TYPE and component-specific constraints                                                              
+    # MIOpen currently only maintains 'standard' and 'quick' categories
+    if test_component_job_name == "miopen":
+        category = "standard" if TEST_TYPE else "quick"
+    else:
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -221,7 +221,12 @@ def build_ctest_command(category, gpu_arch, available_gpu_archs):
 
 
 def main():
-    category = TEST_TYPE.lower() if TEST_TYPE else "quick"
+    # Determine test category based on TEST_TYPE and component-specific constraints                                                              
+    # MIOpen currently only maintains 'standard' and 'quick' categories
```
dileepr1, line 225: TheRock now supports calling all of the categories depending on scenarios for CI runs - quick, standard, comprehensive and full.
If running all of the tests under comprehensive and full is not working out for miopen currently, can we update the tests lists in test_categories.yaml - like using the same test sets as standard in other categories as well - rather than making project specific updates in the code.

aalabsi-amd:
Addressed using: https://github.com/ROCm/rocm-libraries/pull/5766

PR #1732, opened 2025-10-09
[CI] Weekly CI and introducing a new amdgpu matrix generator 

HereThereBeDragons:
This introduces a new weekly CI. 

To make this comfortable, larger changes of the amdgpu matrix generator were necessary. The amdgpu matrix generator can now be used to create build, test and/or release target config in a single run for both Windows and Linux. For pull request it keeps the functionality of dynamically deciding based on the modified files if it needs to build the target or not.

In addition, the `new_setup.yml` is used to provide the container image. It can either use the default CI TheRock image or accepts a custom one.

PR #4123, opened 2026-03-23
Fork configuration scripts and workflows for multi-arch CI

ScottTodd:
This adds new `setup_multi_arch.yml` and `configure_multi_arch_ci_summary.py` configuration code for multi-arch CI.

## Motivation

The existing [`build_tools/github_actions/configure_ci.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_ci.py) script and [`.github/workflows/setup.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/setup.yml) workflow were both tightly coupled to the non-multi-arch CI pipelines and multi-arch CI has some unique needs:

* https://github.com/ROCm/TheRock/issues/3399
* Different workflow I/O (single-arch CI has a matrix across each family, multi-arch CI has a single pipeline per platform)
* (Related) Setting alternate schedules, see https://github.com/ROCm/TheRock/pull/1732

I also judged that starting fresh would be easier, given architectural issues with the existing code:
* Mixed responsibilities in a 300 LOC `matrix_generator()` function
* Sequencing of decisions (whether to skip, what to build, what to test, etc.) was scattered and sometimes duplicated

## Technical Details

> [!NOTE]
> See my worklog for this feature branch here: [`tasks/active/multi-arch-configure.md`](https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/multi-arch-configure.md)

The new code is structured as a sequence of stages:

1. Parse inputs from github (triggers, labels) and git (files changed)
2. Check "skip CI" gate to early return
3. Decide which jobs to run and with what options (skip/use prebuilt/build)
4. Decide which GPU targets/families to build (trigger type + labels --> per-platform GPU families)
5. Expand per-platform build configuration data structures
6. Write outputs to github outputs and step summary

It includes several ✨NEW✨ features:
* Rendered markdown summarizing the configure output
    Before | After
    -- | --
    https://github.com/ROCm/TheRock/actions/runs/23509071878?pr=4142 image | https://github.com/ROCm/TheRock/actions/runs/23465566745?pr=4123 image
* A collapsed build graph instead of a matrix with a single build variant
    Before | After
    -- | --
    https://github.com/ROCm/TheRock/actions/runs/23509071878?pr=4142 image | https://github.com/ROCm/TheRock/actions/runs/23465566745?pr=4123 image
* Reading inputs via `GITHUB_EVENT_PATH` instead of explicit/verbose `github.event.inputs` plumbing
* Passing outputs via `build_config` JSON instead of explicit/verbose `artifact_group`, `matrix_per_family_json`, `dist_amdgpu_families`, etc. (to stay under input count limits and make cross-file maintenance hopefully easier)

As well as a few 🪦REMOVED 🪦 features:
* No longer using `determine_long_lived_branch()` to change `push` behavior based on the branch name - if we enable the workflow on a branch we should run the same set of jobs
* No `run_functional_tests` plumbing - this has not been added to multi-arch CI [yet?]

### Comparison Metrics

<details><summary>📊 Feature Comparison</summary>
<p>

| Feature | Old | New | Change |
|---------|-----|-----|--------|
| Pipeline architecture | Monolithic `matrix_generator` (295 lines) | 6-step pipeline of pure functions | Redesigned |
| Typing | Untyped `base_args` dict | 11 frozen dataclasses, `JobAction` enum | New |
| Skip CI gate | Buried in `matrix_generator` + `main()` | `should_skip_ci()` | Simplified |
| Target selection | Interleaved with matrix expansion | `select_targets()` | Separated |
| Test type | Post-hoc mutation in `main()` | `_determine_test_type()` | Separated |
| Matrix output | Array of per-variant rows via `strategy.matrix` | Single `build_config` JSON per platform | Simplified |
| Prebuilt stages | Boolean `use_prebuilt_artifacts` | Per-stage `dict[str, JobAction]` on `BuildRocmDecision` | More granular |
| Workflow contract | Implicit (YAML ↔ Python drift possible) | Contract tests extract `fromJSON` refs from YAML, assert against dataclass fields | Validated |
| Step summary | Limited information | Markdown with skip reasons, per-family test table, non-default callouts | Redesigned |
| `setup.yml` coupling | 8+ env vars piped through YAML → Python | Script reads `GITHUB_EVENT_PATH` directly | Decoupled |

</p>
</details>

<details><summary>🔎 Code Complexity</summary>
<p>

Around the same number of statements but with more comments and structure:

| Metric | Old `configure_ci.py` (main) | New `configure_multi_arch_ci.py` + summary | Remaining `configure_ci.py` |
|--------|-----|-----|-----|
| Lines | 828 | 947 + 204 = 1,151 | 676 |
| Statements | 348 | 335 + 114 = 449 | 302 |
| Functions | 8 | 30 + 6 = 36 | 7 |
| Classes/dataclasses | 0 | 11 | 0 |

</p>
</details>

<details><summary>🧪 Test Coverage</summary>
<p>

Logic and dataclasses are tested step by step, inputs/outputs are pushed to the boundaries for easier testing.

| Metric | Old (main) | New | Delta |
|--------|-----------|-----|-------|
| Test functions | 33 (configure_ci_test) | 47 (multi_arch) | +14 |
| Test lines | 907 | 826 | -81 |
| Remaining configure_ci tests | — | 27 | -6 (multi-arch tests removed) |
| Statement coverage (main script) | ~63% (configure_ci.py) | 84% (configure_multi_arch_ci.py) | +21pp |
| Uncovered code | `main()`, multi-arch paths mixed with single-arch | I/O boundary only (`from_environ`, `write_outputs`, `main`) | cleaner boundary |

</p>
</details> 

## Test Plan

* New unit tests
  * Tests for each stage, using the dataclasses/enums/etc. that are passed between them
  * A few integration tests for the whole `configure()` pipeline
  * Tests for the workflow YAML files and their `fromJSON` usage
* Manual testing
  * on `pull_request`, multi-arch CI should run by default
  * on `push` it should run with the same behavior as before
* Watch multi-arch CI behavior for `push` events after merge

## Test Result

Expected jobs ran on https://github.com/ROCm/TheRock/actions/runs/23773060560?pr=4123, configure output markdown looks as expected.

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_native_linux_packages.yml
+++ b/.github/workflows/multi_arch_build_native_linux_packages.yml
@@ -1,0 +1,187 @@
+name: Build Multi-Arch Native Linux Packages
+
+on:
+  workflow_call:
+    inputs:
+      dist_amdgpu_families:
+        description: "Semicolon-separated list of all GPU families (e.g., 'gfx94X-dcgpu;gfx120X-all')"
+        required: true
+        type: string
+      artifact_group:
+        type: string
+        description: "Artifact group (e.g. multi-arch-release)"
+        required: true
+      artifact_run_id:
+        description: "Workflow run id to download the artifacts from"
+        required: true
+        type: string
+      rocm_version:
+        description: "ROCm version to append to the package (8.0.0, 8.0.1rc1, ...)"
+        required: true
+        type: string
+      native_package_type:
+        description: "Package type (deb or rpm)"
+        required: true
+        type: string
+      package_suffix:
+        description: "The suffix to be added to package name (asan, tsan, static or rpath)"
+        required: false
+        type: string
+        default: ""
+      release_type:
+        description: "The type of release to build ('dev', 'nightly', or 'release'). Empty string for CI builds."
+        required: false
+        type: string
+        default: ""
+      repository:
+        description: "Repository to checkout. Otherwise, defaults to `github.repository`"
+        type: string
+        required: false
+      ref:
+        description: "Branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow"
+        type: string
+        required: false
+
+permissions:
+  id-token: write
+  contents: read
+
+run-name: Build ${{ inputs.native_package_type }} packages (${{ inputs.rocm_version }}${{ inputs.release_type && format(', {0}', inputs.release_type) || '' }})
+
+jobs:
+  build_native_packages:
+    name: Build ${{ inputs.native_package_type }} packages
+    runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
+    env:
+      ARTIFACT_RUN_ID: ${{ inputs.artifact_run_id }}
+      DIST_AMDGPU_FAMILIES: ${{ inputs.dist_amdgpu_families }}
+      PACKAGE_SUFFIX: ${{ inputs.package_suffix }}
+      OUTPUT_DIR: ${{ github.workspace }}/output
+      ARTIFACTS_DIR: ${{ github.workspace }}/output/artifacts
+      PACKAGE_DIST_DIR: ${{ github.workspace }}/output/packages
+      RELEASE_TYPE: ${{ inputs.release_type }}
+    steps:
+      - name: "Checking out repository"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || '' }}
+
+      - name: Set up Python
+        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
+        with:
+          python-version: '3.12'
+
+      - name: Install Python requirements
+        run: |
+          pip install pyelftools boto3 jinja2
+
+      - name: Install System requirements
+        run: |
+          # Install the needed tools for creating rpm / deb packages
+          # Also install tools for creating repo files
+          sudo apt update
+          sudo apt install -y llvm
+          sudo apt install -y rpm debhelper-compat build-essential
+          sudo apt install -y dpkg-dev createrepo-c
+
+      - name: Determine S3 bucket and prefix
+        id: s3_config
+        run: |
+          python ./build_tools/packaging/linux/get_s3_config.py \
+            --release-type "${{ env.RELEASE_TYPE }}" \
+            --repository "${{ github.repository }}" \
+            --is-fork "${{ github.event.pull_request.head.repo.fork || 'false' }}" \
+            --pkg-type "${{ inputs.native_package_type }}" \
+            --artifact-id "${{ env.ARTIFACT_RUN_ID }}" \
+            --rocm-version "${{ inputs.rocm_version }}" \
+            --output-format github >> $GITHUB_OUTPUT
+
+      - name: Determine IAM role
+        id: iam_role
+        run: |
+          # ================================================================
+          # IAM Role Selection Logic
+          # ================================================================
+          # Determines which AWS IAM role to assume based on job_type from s3_config step.
+          #
+          # Role Mapping:
+          #   ├─ IF job_type == "ci"
+          #   │  └─ Use: therock-ci role
+          #   │     (For all CI buckets: therock-ci-artifacts, therock-ci-artifacts-external, therock-artifacts-internal)
+          #   │
+          #   └─ ELSE (job_type == dev/nightly/prerelease/release)
+          #      └─ Use: therock-${job_type} role
+          #         (For package buckets: therock-dev-packages, therock-nightly-packages, etc.)
+          #
+          # ================================================================
+
+          JOB_TYPE="${{ steps.s3_config.outputs.job_type }}"
+
+          if [[ "${JOB_TYPE}" == "ci" ]]; then
+            # CI builds use the shared CI role (for all artifact buckets)
+            IAM_ROLE="arn:aws:iam::692859939525:role/therock-ci"
+            echo "✓ Using CI role: ${IAM_ROLE}"
+          else
+            # Release builds use release-type-specific roles (for package buckets)
+            IAM_ROLE="arn:aws:iam::692859939525:role/therock-${JOB_TYPE}"
+            echo "✓ Using release-type role: ${IAM_ROLE}"
+          fi
+
+          echo "iam_role=${IAM_ROLE}" >> $GITHUB_OUTPUT
+
+      - name: Fetch Artifacts for all GPU families
+        run: |
+          echo "Fetching artifacts for build ${{ env.ARTIFACT_RUN_ID }}"
+
+          # Convert semicolon-separated to comma-separated
+          FAMILIES_CSV="${DIST_AMDGPU_FAMILIES//;/,}"
+          echo "Fetching artifacts for GPU families: ${FAMILIES_CSV}"
+
+          # Create artifacts directory
+          mkdir -p "${{ env.ARTIFACTS_DIR }}"
+
+          python ./build_tools/fetch_artifacts.py \
+            --run-id="${{ env.ARTIFACT_RUN_ID }}" \
+            --run-github-repo="${{ github.repository }}" \
+            --artifact-group="${{ inputs.artifact_group }}" \
+            --platform="linux" \
+            --amdgpu-targets="${FAMILIES_CSV}" \
+            --output-dir="${{ env.ARTIFACTS_DIR }}"
+
+      - name: Build Packages
+        id: build-packages
+        run: |
+          echo "Building ${{ inputs.native_package_type }} packages for all GPU families"
+          echo "Families: ${{ env.DIST_AMDGPU_FAMILIES }}"
+
+          # Pass the target families as-is (semicolon-separated)
+          # build_package.py's normalize_target_list() handles all separators
+          python ./build_tools/packaging/linux/build_package.py \
+            --dest-dir ${{ env.PACKAGE_DIST_DIR }} \
+            --rocm-version  ${{ inputs.rocm_version }} \
+            --target "${{ env.DIST_AMDGPU_FAMILIES }}" \
+            --artifacts-dir ${{ env.ARTIFACTS_DIR }} \
+            --pkg-type ${{ inputs.native_package_type }} \
+            --version-suffix ${{ env.ARTIFACT_RUN_ID }} \
+            --enable_kpack=false
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        with:
+          aws-region: us-east-2
+          role-to-assume: ${{ steps.iam_role.outputs.iam_role }}
+
+      - name: Upload Package repo to S3
+        id: upload-packages
+        run: |
+          echo "Uploading to S3 bucket: ${{ steps.s3_config.outputs.s3_bucket }}"
+          echo "Using prefix: ${{ steps.s3_config.outputs.s3_prefix }}"
+          echo "Job type: ${{ steps.s3_config.outputs.job_type }}"
+
+          python ./build_tools/packaging/linux/upload_package_repo.py \
+            --pkg-type ${{ inputs.native_package_type }} \
+            --s3-bucket ${{ steps.s3_config.outputs.s3_bucket }} \
+            --artifact-id ${{ env.ARTIFACT_RUN_ID }} \
+            --job ${{ steps.s3_config.outputs.job_type }} \
+            --s3-prefix "${{ steps.s3_config.outputs.s3_prefix }}"
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -374,7 +374,7 @@ test_matrix = {
     "miopenprovider": {
         "job_name": "miopenprovider",
         "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
-        "timeout_minutes": 15,
+        "timeout_minutes": 20,
         "test_script": f"python {_get_script_path('test_miopenprovider.py')}",
         "platform": ["linux", "windows"],
         "total_shards_dict": {
--- a/build_tools/github_actions/post_build_upload.py
+++ b/build_tools/github_actions/post_build_upload.py
@@ -48,37 +48,6 @@ def log(*args):
     sys.stdout.flush()
 
 
-# This method will output logs of the Windows Time Service and is meant
-# to help debug spurious AWS auth issues caused by time differences when
-# uploading with the AWS CLI tool. For context, see this issue and PR:
-# https://github.com/ROCm/TheRock/issues/875
-# https://github.com/ROCm/TheRock/pull/1581#issuecomment-3490177590
-def write_time_sync_log():
-    if platform.system().lower() != "windows":
-        log("[*] Current OS not windows, Skipping.")
-        return
-
-    # Logs are from `w32tm` run in Windows HostProcess containers on Azure VMs
-    # with `/query /status` and `/stripchart /computer:time.aws.com /dataonly`
-    # and are mounted via the readonly H: drive for Github Runner Pods to access
-    startfile = Path("H:\\start.log")
-    timefile = Path("H:\\time.log")
-
-    # Only output if these files exist in the H: drive as expected on Build VMs
-    if startfile.is_file() and timefile.is_file():
-        log(f"[*] Checking time sync at: {datetime.now()}")
-
-        log("[*] Start Time Sync Log:")
-        log(startfile.read_text())
-
-        log("[*] Time Sync Log (last ~50 lines):")
-        timef = open(timefile)
-        timelines = timef.readlines()
-        log("".join(timelines[-51:]))
-    else:
-        log("[*] time.log and/or start.log not present in H:")
-
-
 def create_ninja_log_archive(build_dir: Path):
     log_dir = build_dir / "logs"
 
@@ -269,10 +238,6 @@ def run(args):
         run_id=args.run_id, platform=PLATFORM
     )
     backend = create_storage_backend(staging_dir=args.output_dir, dry_run=args.dry_run)
-
-    log("Write Windows time sync log")
-    log("----------------------")
-    write_time_sync_log()
 
     # Upload artifacts only if the job not failed
     if not args.job_status or args.job_status == "success":
--- a/build_tools/github_actions/test_executable_scripts/test_miopenprovider.py
+++ b/build_tools/github_actions/test_executable_scripts/test_miopenprovider.py
@@ -28,6 +28,8 @@ TEST_TO_IGNORE = {
 
 logging.basicConfig(level=logging.INFO)
 
+# If you increase the timeout here you need to also increase the timeout for the job
+# See file build_tools/github_actions/fetch_test_configurations.py and search for miopenprovider
 cmd = [
     "ctest",
     "--test-dir",
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -23,6 +23,7 @@ import os
 import shutil
 import subprocess
 import sys
+import traceback
 
 from datetime import datetime, timezone
 from email.utils import format_datetime
@@ -56,8 +57,10 @@ def create_deb_package(pkg_name, config: PackageConfig):
     print(f"Package Name: {pkg_name}")
 
     # Non-versioned packages are not required for RPATH packages
+    # In multi-arch mode, only create non-versioned packages for generic architecture
     if not config.enable_rpath:
-        create_nonversioned_deb_package(pkg_name, config)
+        if not config.enable_kpack or config.gfx_arch == GFX_GENERIC:
+            create_nonversioned_deb_package(pkg_name, config)
 
     create_versioned_deb_package(pkg_name, config)
     output_list = move_packages_to_destination(pkg_name, config)
@@ -132,13 +135,21 @@ def create_versioned_deb_package(pkg_name, config: PackageConfig):
 
     sourcedir_list = []
     dir_list = filter_components_fromartifactory(
-        pkg_name, config.artifacts_dir, config.gfx_arch
+        pkg_name, config.artifacts_dir, config.gfx_arch, config.enable_kpack
     )
     sourcedir_list.extend(dir_list)
 
     print(f"sourcedir_list:\n  {sourcedir_list}")
     if not sourcedir_list and not is_meta:
-        sys.exit(f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting")
+        if config.enable_kpack:
+            print(
+                f"ERROR: {pkg_name}: Empty sourcedir_list and not a meta package, skipping"
+            )
+            return []
+        else:
+            sys.exit(
+                f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting"
+            )
 
     if not sourcedir_list:
         print(f"{pkg_name} is a Meta package")
@@ -263,6 +274,14 @@ def generate_rules_file(pkg_info, deb_dir, config: PackageConfig):
     # Get package name for changelog installation
     pkg_name = update_package_name(pkg_info.get("Package"), config)
 
+    # Disable debian dh_strip for multi-arch builds
+    # WORKAROUND: dh_strip's debugedit incorrectly truncates ELF files with
+    # unconventional layouts (e.g., program headers at end of file).
+    # This causes "program header goes past the end of the file" errors.
+    # See: https://github.com/ROCm/TheRock/issues/4047
+    if config.enable_kpack:
+        disable_dh_strip = True
+
     env = Environment(loader=FileSystemLoader(str(SCRIPT_DIR)))
     template = env.get_template("template/debian_rules.j2")
     # Prepare  context dictionary
@@ -291,47 +310,32 @@ def generate_control_file(pkg_info, deb_dir, config: PackageConfig):
     """
     print_function_name()
     control_file = Path(deb_dir) / "control"
-
     pkg_name = pkg_info.get("Package")
-    provides = ""
-    replaces = ""
-    conflicts = ""
-    debrecommends = ""
-    debsuggests = ""
+    is_meta = is_meta_package(pkg_info)
+
+    # Initialize optional fields
+    provides = replaces = conflicts = ""
+    debrecommends = debsuggests = ""
 
     if config.versioned_pkg:
-        recommends_list = pkg_info.get("DEBRecommends", [])
-        debrecommends = convert_to_versiondependency(recommends_list, config)
-        suggests_list = pkg_info.get("DEBSuggests", [])
-        debsuggests = convert_to_versiondependency(suggests_list, config)
-
-        depends_list = pkg_info.get("DEBDepends", [])
+        # Get -> Filter -> Transform
+        debrecommends = process_dependency_field(pkg_info, "DEBRecommends", config)
+        debsuggests = process_dependency_field(pkg_info, "DEBSuggests", config)
+        depends = process_dependency_field(
+            pkg_info, "DEBDepends", config, use_multiarch=True
+        )
     else:
-        depends_list = [pkg_name]
-        provides_list = [
-            debian_replace_devel_name(pkg)
-            for pkg in (pkg_info.get("Provides", []) or [])
-        ]
-        provides = ", ".join(provides_list)
-        replaces_list = [
-            debian_replace_devel_name(pkg)
-            for pkg in (pkg_info.get("Replaces", []) or [])
-        ]
-        replaces = ", ".join(replaces_list)
-        conflicts_list = [
-            debian_replace_devel_name(pkg)
-            for pkg in (pkg_info.get("Conflicts", []) or [])
-        ]
-        conflicts = ", ".join(conflicts_list)
-
-    depends = convert_to_versiondependency(depends_list, config)
-    if is_meta_package(pkg_info):
-        depends = append_version_suffix(depends, config)
+        # Get -> Transform -> Join
+        provides = process_name_field(pkg_info, "Provides", debian_replace_devel_name)
+        replaces = process_name_field(pkg_info, "Replaces", debian_replace_devel_name)
+        conflicts = process_name_field(pkg_info, "Conflicts", debian_replace_devel_name)
+        # Non-versioned package depends on versioned package itself
+        depends = resolve_versioned_dependencies([pkg_name], config, is_meta)
 
     pkg_name = update_package_name(pkg_name, config)
+
     env = Environment(loader=FileSystemLoader(str(SCRIPT_DIR)))
     template = env.get_template("template/debian_control.j2")
-    # Prepare your context dictionary
     context = {
         "source": pkg_name,
         "depends": depends,
@@ -478,8 +482,11 @@ def create_rpm_package(pkg_name, config: PackageConfig):
     print_function_name()
     print(f"Package Name: {pkg_name}")
 
+    # Non-versioned packages are not required for RPATH packages
+    # In multi-arch mode, only create non-versioned packages for generic architecture
     if not config.enable_rpath:
-        create_nonversioned_rpm_package(pkg_name, config)
+        if not config.enable_kpack or config.gfx_arch == GFX_GENERIC:
+            create_nonversioned_rpm_package(pkg_name, config)
 
     create_versioned_rpm_package(pkg_name, config)
     output_list = move_packages_to_destination(pkg_name, config)
@@ -547,15 +554,12 @@ def generate_spec_file(pkg_name, specfile, config: PackageConfig):
     os.makedirs(os.path.dirname(specfile), exist_ok=True)
 
     pkg_info = get_package_info(pkg_name)
-    # populate packge version details
     version = f"{config.rocm_version}"
-    # TBD: Whether to use component version details?
-    #    version = pkg_info.get("Version")
-    provides = ""
-    obsoletes = ""
-    conflicts = ""
-    rpmrecommends = ""
-    rpmsuggests = ""
+    is_meta = is_meta_package(pkg_info)
+
+    # Initialize optional fields
+    provides = obsoletes = conflicts = ""
+    rpmrecommends = rpmsuggests = ""
     sourcedir_list = []
     rpm_scripts = []
     # amdrocm-debugger: Exclude libpython requirements
@@ -564,20 +568,31 @@ def generate_spec_file(pkg_name, specfile, config: PackageConfig):
     exclude_libpython_requires = pkg_name == "amdrocm-debugger"
 
     if config.versioned_pkg:
-        recommends_list = pkg_info.get("RPMRecommends", [])
-        rpmrecommends = convert_to_versiondependency(recommends_list, config)
-        suggests_list = pkg_info.get("RPMSuggests", [])
-        rpmsuggests = convert_to_versiondependency(suggests_list, config)
-
-        requires_list = pkg_info.get("RPMRequires", [])
+        # Get -> Filter -> Transform
+        rpmrecommends = process_dependency_field(pkg_info, "RPMRecommends", config)
+        rpmsuggests = process_dependency_field(pkg_info, "RPMSuggests", config)
+        requires = process_dependency_field(
+            pkg_info, "RPMRequires", config, use_multiarch=True
+        )
 
         dir_list = filter_components_fromartifactory(
-            pkg_name, config.artifacts_dir, config.gfx_arch
+            pkg_name, config.artifacts_dir, config.gfx_arch, config.enable_kpack
         )
         sourcedir_list.extend(dir_list)
 
         # Filter out non-existing directories
         sourcedir_list = [path for path in sourcedir_list if os.path.isdir(path)]
+
+        # Warn if we have no artifacts for non-meta packages
+        if not sourcedir_list and not is_meta:
+            if config.enable_kpack:
+                print(
+                    f"WARNING: {pkg_name}: Empty sourcedir_list and not a meta package, creating empty RPM"
+                )
+            else:
+                sys.exit(
+                    f"{pkg_name}: Empty sourcedir_list and not a meta package, exiting"
+                )
 
         if is_postinstallscripts_available(pkg_info):
             rpm_scripts = generate_rpm_postscripts(pkg_info, config)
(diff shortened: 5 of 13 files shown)


ScottTodd:
Alright, I think this is ready for human review now. It's gone through several rounds of agent review already:
* https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_021_multi-arch-configure.md
* https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/local_023_multi-arch-configure.md
* https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4123.md

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+      routine changes that are unlikely to break GPU-specific behavior.
+    - "standard": More thorough than quick, but not full nightly coverage.
+      Only available via explicit test_filter:standard PR label.
+    - "comprehensive": Full nightly test suite. Used for scheduled runs
+      to catch regressions across all components without requiring a
+      submodule change to trigger it.
+    - "full": Everything, including tests for specific components named
+      by test:* labels. Triggered when a submodule changes (the actual
+      GPU libraries changed, so we need thorough validation) or when
+      test labels explicitly request specific component tests.
```
geomin12: perhaps link to the test_filtering.md? and maybe just have a quick blurb for these?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+                f"Unrecognized test_filter value: {filter_type!r}. "
+                f"Valid values: {sorted(_VALID_TEST_FILTER_TYPES)}"
+            )
+        return filter_type, f"test_filter label: {label}"
+
+    # Priority 2: test:* labels request specific component tests (e.g.
+    # test:rocprim). When someone explicitly asks for tests, run the full
+    # suite — they're investigating something specific.
+    if _has_test_labels(ci_inputs):
+        return "full", "test labels specified"
```
geomin12, line 520: there could / will be a scenario where a specific test and filter was requested, but I think this priority would skip one or the other

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    # Priority 4: a submodule change means actual library code changed
+    # (e.g. rocBLAS, MIOpen). These need full testing since the change
+    # could affect any downstream consumer.
+    if (
+        git_context.changed_files is not None
+        and git_context.submodule_paths is not None
+    ):
+        matching = set(git_context.submodule_paths) & set(git_context.changed_files)
+        if matching:
+            return "full", f"submodule(s) changed: {sorted(matching)}"
```
geomin12, line 536: "full" may require a lotttt of tests (i think rocblas is like 24h??? or something gnarly like that). should we just do comprehensive instead? turn around time may be brutal with full

.github/workflows/multi_arch_ci.yml
```diff
@@ -60,72 +60,44 @@ concurrency:
...
-    uses: ./.github/workflows/setup.yml
+    uses: ./.github/workflows/setup_multi_arch.yml
     with:
       build_variant: "release"
-      multi_arch: true
-      prebuilt_stages: ${{ inputs.prebuilt_stages }}
 
   linux_build_and_test:
-    name: Linux::${{ matrix.variant.build_variant_label }}
+    name: Linux
```
HereThereBeDragons: where do we now put the build variant?

.github/workflows/setup_multi_arch.yml
```diff
@@ -0,0 +1,66 @@
...
+      # workflow_dispatch inputs — no env var pass-through needed.
+      - name: Configuring CI options
+        id: configure
+        env:
+          BUILD_VARIANT: ${{ inputs.build_variant }}
+        run: python ./build_tools/github_actions/configure_multi_arch_ci.py
+
+      - name: Compute package version
+        id: rocm_package_version
+        run: python ./build_tools/compute_rocm_package_version.py --release-type=dev
```
HereThereBeDragons, line 66: shouldnt be release-type be adjustable?

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+
+def format_summary(
+    ci_inputs: CIInputs,
+    outputs: CIOutputs,
+) -> str:
+    """Generate the full step summary markdown."""
+    lines = ["## Multi-Arch CI Configuration", ""]
+
+    if not outputs.is_ci_enabled:
+        return _format_skipped(lines, ci_inputs)
```
HereThereBeDragons:
```suggestion
        return _format_skipped_ci(lines, ci_inputs)
```

just to clarify that you format the output for the skipped ci :)

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if not outputs.jobs:
+        return "\n".join(lines)
+
+    # One-liner: trigger, branch, variant
+    lines.append(
+        f"Trigger: `{ci_inputs.event_name}` on `{ci_inputs.branch_name}` branch, "
+        f"`{ci_inputs.build_variant}` variant."
+    )
+    lines.append("")
```
HereThereBeDragons, line 47: i guess thats a new line? might have it as a
```
NEW_LINE="" 
lines.append(NEW_LINE)
```
so its easier to understand?

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+    elif ci_inputs.is_pull_request and "ci:run-multi-arch" not in ci_inputs.pr_labels:
+        reason = "PR does not have `ci:run-multi-arch` label"
+    else:
+        reason = "no CI-relevant files changed"
+
+    lines.append(f"CI was **skipped**: {reason}. See logs for details.")
+    return "\n".join(lines)
+
+
+def _non_default_callouts(ci_inputs: CIInputs, outputs: CIOutputs) -> list[str]:
```
HereThereBeDragons: i never heard callout in that context. maybe a  few comments could help to set the expectation

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+    callouts: list[str] = []
+    jobs = outputs.jobs
+
+    # Explicit family selection (workflow_dispatch)
+    if ci_inputs.is_workflow_dispatch:
+        if ci_inputs.linux_amdgpu_families or ci_inputs.windows_amdgpu_families:
```
HereThereBeDragons: if you want to reduce indentation here you can remove this `if` here and add below the ifs another if `if parts:
  callouts.append...
```

File: build_tools/github_actions/configure_multi_arch_ci_summary.py [hunk 10]
```diff
@@ -0,0 +1,202 @@
...
+            )
+        elif label.startswith("test_filter:"):
+            callouts.append(
+                f"Label `{label}`: overrode test level (default would be `quick`)"
+            )
+        elif label.startswith("test:"):
+            callouts.append(f"Label `{label}`: requested component tests")
+        elif label.startswith("ci:"):
+            callouts.append(f"Label `{label}`")
+
```
HereThereBeDragons, line 122: we also have a `test_runner:` label to run e.g. oem kernel. geo had some pr for it for gfx1151

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+    return callouts
+
+
+def _append_build_rocm(lines: list[str], outputs: CIOutputs) -> None:
+    jobs = outputs.jobs
+
+    # Prebuilt info
+    prebuilt = jobs.build_rocm.prebuilt_stages
```
HereThereBeDragons, line 148: havent looked at the configure_ci yet. we always have prebuilt_stages in build_rocm?

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+            callouts.append(
+                f"Explicit Linux test labels: `{ci_inputs.linux_test_labels}`"
+            )
+        if ci_inputs.windows_test_labels:
+            callouts.append(
+                f"Explicit Windows test labels: `{ci_inputs.windows_test_labels}`"
+            )
+
+    # Prebuilt stages
+    if jobs and jobs.build_rocm.prebuilt_stages:
```
HereThereBeDragons: do we need this here if we are listing it also in `_append_build_rocm`?

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+
+    # Nothing to build (e.g. workflow_dispatch with no families selected)
+    if outputs.builds.linux is None and outputs.builds.windows is None:
+        lines.append("No GPU families selected — nothing to build or test.")
+        return "\n".join(lines)
+
+    # Non-default callout
+    callouts = _non_default_callouts(ci_inputs, outputs)
+    if callouts:
+        lines.append("> [!NOTE]")
```
HereThereBeDragons, line 57: i dont think we need this as a note. just as a normal headline ### Custom Config or something like that. if it is very long i think the "note" status might be not so nice to read

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -0,0 +1,202 @@
...
+            families = ", ".join(
+                f"`{f}`" for f in config.dist_amdgpu_families.split(";")
+            )
+            lines.append(f"| {platform} | {families} | `{config.artifact_group}` |")
+    lines.append("")
+
+
+def _append_test_rocm(lines: list[str], outputs: CIOutputs) -> None:
+    jobs = outputs.jobs
+    test_rocm = jobs.test_rocm
```
HereThereBeDragons, line 204: we always have a non empty test_rocm when we end here?

ScottTodd:
Thanks for the feedback. We've been landing substantial changes to the old/current configure_ci.py and workflows as part of switching our default workflow from ci.yml to multi_arch_ci.yml. I'll sync this, resolve merge conflicts, then start addressing the comments next week.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    GITHUB_REF_NAME     : Branch name
+    GITHUB_OUTPUT       : Path to write workflow output variables
+    GITHUB_STEP_SUMMARY : Path to write workflow summary
+    BUILD_VARIANT       : Build variant (workflow_call input, not in event payload)
+
+Outputs (written to GITHUB_OUTPUT):
+    linux_build_config    : JSON object with build config, or "" if skipped
+    windows_build_config  : JSON object with build config, or "" if skipped
+    enable_build_jobs     : "true" or "false"
+    test_type             : "quick", "standard", "comprehensive", or "full"
```
HereThereBeDragons, line 44: difference between smoke and quick? or is it the same just with new naming?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+@dataclass(frozen=True)
+class CIInputs:
+    """All external inputs to the CI configuration pipeline.
+
+    Constructed once from the GitHub Actions environment. Every downstream
+    function takes this (or a subset) as a plain argument — no environment
+    access needed.
+    """
+
+    event_name: str  # push, pull_request, schedule, workflow_dispatch
```
HereThereBeDragons: enum? literal? frozen set?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    windows_test_labels: str = ""
+
+    # Prebuilt configuration (from workflow_dispatch)
+    prebuilt_stages: str = ""
+    baseline_run_id: str = ""
+
+    def log(self) -> None:
+        """Log parsed inputs for CI diagnostics."""
+        print("CIInputs:")
+        print(f"  event: {self.event_name}")
```
HereThereBeDragons: claude says the following should work
`from dataclasses import dataclass, field, fields`

```
def log(self) -> None:                                                                                                                                                                       
      print("CIInputs:")                                                                                                                                                                       
      for f in fields(self):                                                                                                                                                                   
          print(f"  {f.name}: {getattr(self, f.name)!r}") 
```

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        print(f"  pr_labels: {self.pr_labels}")
+        print(f"  linux_amdgpu_families: {self.linux_amdgpu_families}")
+        print(f"  windows_amdgpu_families: {self.windows_amdgpu_families}")
+        print(f"  linux_test_labels: {self.linux_test_labels!r}")
+        print(f"  windows_test_labels: {self.windows_test_labels!r}")
+        print(f"  prebuilt_stages: {self.prebuilt_stages!r}")
+        print(f"  baseline_run_id: {self.baseline_run_id!r}")
+
+    @property
+    def is_pull_request(self) -> bool:
```
HereThereBeDragons, line 114: if using enums this is not needed

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        return self.event_name == "schedule"
+
+    @property
+    def is_workflow_dispatch(self) -> bool:
+        return self.event_name == "workflow_dispatch"
+
+    @staticmethod
+    def from_environ() -> "CIInputs":
+        """Parse from GitHub Actions environment."""
+        event_name = os.environ.get("GITHUB_EVENT_NAME", "")
```
HereThereBeDragons: `if not event_name:` missing

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        # Read the full event webhook payload.
+        event_path = os.environ.get("GITHUB_EVENT_PATH", "")
+        if event_path and Path(event_path).exists():
+            with open(event_path) as f:
+                event = json.load(f)
+        else:
+            event = {}
+
+        # Extract fields based on event type.
+        inputs = event.get("inputs") or {}
```
HereThereBeDragons, line 144: comment: workflow_dispatch inputs only
workflow_call inputs are not in the event payload — GitHub injects them as environment variables instead (via the env: block in the calling workflow).

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        pr_labels: list[str] = []
+        base_ref = "HEAD^1"
+
+        if event_name == "pull_request":
+            pr_obj = event.get("pull_request", {})
+            pr_labels = [label["name"].lower() for label in pr_obj.get("labels", [])]
```
HereThereBeDragons, line 156: maybe small example how it looks in the end? seems like you are getting a list of strings just containing the labels and not the json part around it anymore labels [{ "name": "ci:skip", "color": "fffff", ...}, {"name": "other_label", "id": 12312, ...}  ] --> ["ci:skip", "other_label"]

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+@dataclass(frozen=True)
+class GitContext:
+    """Git-derived data for the current commit/PR.
+
+    Separated from CIInputs because these require git operations to compute,
+    while CIInputs is parsed from the GitHub Actions environment. Tests
+    construct GitContext directly without touching git.
+    """
+
+    changed_files: list[str] | None = None
```
HereThereBeDragons, line 193: small comment would be good here for both params

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+
+    def log(self) -> None:
+        """Log selected targets for CI diagnostics."""
+        print("TargetSelection:")
+        print(f"  linux: {self.linux_families}")
+        print(f"  windows: {self.windows_families}")
+
+
+# ---------------------------------------------------------------------------
+# Job decisions — the CI pipeline as a DAG of job groups
```
HereThereBeDragons, line 244: maybe some ascii art here :) ?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    """Action for a node in the CI job graph or a build stage."""
+
+    RUN = "run"
+    PREBUILT = "prebuilt"
+    SKIP = "skip"
+
+
+@dataclass(frozen=True)
+class JobGroupDecision:
+    """Decision for one node in the CI job graph.
```
HereThereBeDragons, line 286: yeah i think some explanation would be good here. probably enough if you add it above to the job decisions overview comment just above jobAction

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+            if action == JobAction.RUN
+        ]
+
+
+@dataclass(frozen=True)
+class TestRocmDecision(JobGroupDecision):
+    """Test-rocm job group with test filtering details.
+
+    test_type levels (from least to most testing):
+    - "quick"         — default for PRs and push
```
HereThereBeDragons, line 325: smoke vs quick?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    expect_failure: bool
+    build_pytorch: bool
+    # Prebuilt stage configuration — set by configure() from JobDecisions.
+    prebuilt_stages: list[str] = field(default_factory=list)
+    baseline_run_id: str = ""
+
+    def to_dict(self) -> dict:
+        """Convert to dict for JSON serialization."""
+        return {
+            "per_family_info": self.per_family_info,
```
HereThereBeDragons: `from dataclasses import asdict, dataclass, field`

```                                                                                                                                                                                               
  def to_dict(self) -> dict:                                                                                                                                                                   
      d = asdict(self)      
      d["prebuilt_stages"] = ",".join(self.prebuilt_stages)                                                                                                                                    
      return d   
```

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        print("BuildConfigs:")
+        self._log_platform("linux", self.linux)
+        self._log_platform("windows", self.windows)
+
+
+@dataclass(frozen=True)
+class CIOutputs:
+    """All outputs from the CI configuration pipeline."""
+
+    is_ci_enabled: bool = True
```
HereThereBeDragons, line 419:
```suggestion
    is_ci_needed: bool = True
```
maybe more clear?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    - 'ci:skip' PR label
+    - pull_request without 'ci:run-multi-arch' label (opt-in during transition)
+    - Only skippable files changed (docs, .md, etc.)
+    - No files changed
+    """
+    if "ci:skip" in ci_inputs.pr_labels:
+        print("  Skipping: 'ci:skip' PR label")
+        return True
+
+    # Multi-arch CI on PRs requires explicit opt-in via label to avoid
```
HereThereBeDragons: outdated

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        """Produce empty outputs when CI is skipped."""
+        return CIOutputs(is_ci_enabled=False)
+
+
+# ---------------------------------------------------------------------------
+# Step 2: Check Skip CI
+# ---------------------------------------------------------------------------
+
+
+def should_skip_ci(
```
HereThereBeDragons, line 437: should we keep the naming it consistent with `is_ci_run_required`?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    """Determine test_type and reason based on trigger, labels, and changed files.
+
+    Test types from least to most testing:
+
+    - "quick": Fast sanity checks. Default for PRs and push where only
+      build infra or non-submodule files changed. Keeps CI fast for
+      routine changes that are unlikely to break GPU-specific behavior.
+    - "standard": More thorough than quick, but not full nightly coverage.
+      Only available via explicit test_filter:standard PR label.
+    - "comprehensive": Full nightly test suite. Used for scheduled runs
```
HereThereBeDragons: test:nightly might be easier to remember

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+            raise ValueError(
+                f"Unrecognized test_filter value: {filter_type!r}. "
+                f"Valid values: {sorted(_VALID_TEST_FILTER_TYPES)}"
+            )
+        return filter_type, f"test_filter label: {label}"
+
+    # Priority 2: test:* labels request specific component tests (e.g.
+    # test:rocprim). When someone explicitly asks for tests, run the full
+    # suite — they're investigating something specific.
+    if _has_test_labels(ci_inputs):
```
HereThereBeDragons, line 519: in the future i guess we want to enable only run selected tests and not default to `full`?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        test_type_reason=test_type_reason,
+    )
+
+    # Other jobs run unconditionally with no configuration.
+    # TODO: job pruning: skip pytorch if only JAX has been edited, etc.
+
+    return JobDecisions(
+        build_rocm=build_rocm,
+        test_rocm=test_rocm,
+        build_rocm_python=JobGroupDecision(action=JobAction.RUN),
```
HereThereBeDragons, line 580: why do you have extra of `JobGroupDecision` instead of just JobAction?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+        test_pytorch=JobGroupDecision(action=JobAction.RUN),
+    )
+
+
+# ---------------------------------------------------------------------------
+# Step 4: Select Targets
+# ---------------------------------------------------------------------------
+
+
+def _validate_family_names(
```
HereThereBeDragons, line 591: these are so small. not sure if you need separate functions for it,

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    if ci_inputs.is_pull_request:
+        for label in ci_inputs.pr_labels:
+            if label == "ci:run-all-archs":
+                # Override to all families.
+                linux_names = list(all_families.keys())
+                windows_names = list(all_families.keys())
+                print("  Label 'ci:run-all-archs' -> all families")
+                break
+            if label.startswith("gfx"):
+                target = label.split("-")[0]
```
HereThereBeDragons, line 688: so gfx94X-dcgpu will become gfx94X?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+                linux_names.append(target)
+                windows_names.append(target)
+                print(f"  Label '{label}' -> adding target {target}")
+
+    # De-dup, validate, then filter by platform availability.
+    linux_names = list(dict.fromkeys(linux_names))
+    windows_names = list(dict.fromkeys(windows_names))
+    _validate_family_names(linux_names, all_families)
+    _validate_family_names(windows_names, all_families)
+    # TODO: For workflow_dispatch, a family requested for a specific platform
```
HereThereBeDragons, line 698: https://github.com/ROCm/TheRock/pull/2952 will allow you to do this easily

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+
+def write_outputs(
+    ci_inputs: CIInputs,
+    outputs: CIOutputs,
+) -> None:
+    """Write results to GITHUB_OUTPUT and GITHUB_STEP_SUMMARY.
+
+    This is the only function with side effects (besides from_environ).
+    """
+    test_type = outputs.jobs.test_rocm.test_type if outputs.jobs else "quick"
```
HereThereBeDragons: why should this be empty?

.github/workflows/multi_arch_ci.yml
HereThereBeDragons, line 17: needs to also run for release/therock-*

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+    prebuilt_stages: list[str] | None = None,
+    baseline_run_id: str = "",
+) -> BuildConfig | None:
+    """Build a BuildConfig for one platform, or None if no families match.
+
+    Collects per-family info for all families that support the requested
+    build variant on this platform, then bundles them into a BuildConfig.
+
+    Per-family info fields:
+    - amdgpu_family: family name for THEROCK_AMDGPU_FAMILIES
```
HereThereBeDragons, line 731: i think there is sometimes confusion between THEROCK_AMDGPU_FAMILIES and THEROCK_AMDGPU_TARGET. might be worth to add a bit more context here.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+                ),
+            }
+        )
+
+    if not per_family_info:
+        return None
+
+    family_names = [f["amdgpu_family"] for f in per_family_info]
+    expect_failure = variant_config.get("expect_failure", False)
+    expect_pytorch_failure = variant_config.get("expect_pytorch_failure", False)
```
HereThereBeDragons, line 797: are we still having htis?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+
+    return BuildConfig(
+        per_family_info=per_family_info,
+        dist_amdgpu_families=";".join(family_names),
+        artifact_group=f"multi-arch-{suffix or 'release'}",
+        build_variant_label=variant_config["build_variant_label"],
+        build_variant_suffix=suffix,
+        build_variant_cmake_preset=variant_config["build_variant_cmake_preset"],
+        expect_failure=expect_failure,
+        build_pytorch=not expect_failure and not expect_pytorch_failure,
```
HereThereBeDragons, line 808: as we are skipping the build, i would like that in the new amdgpu family matrix we better call it skip_broken_pytorch or something like this - or build it and makred it as xfail and passing in the workflow

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -0,0 +1,943 @@
...
+
+    linux_config: BuildConfig | None = None
+    windows_config: BuildConfig | None = None
+
+    for platform, families in [
+        ("linux", targets.linux_families),
+        ("windows", targets.windows_families),
+    ]:
+        variant_config = all_build_variants.get(platform, {}).get(build_variant)
+        if not variant_config:
```
HereThereBeDragons, line 838: is this needed?
you already have 
```
if not per_family_info:
        return None
```
in `_expand_build_config_for_platform()` that could be used to check.

.github/workflows/setup_multi_arch.yml
ScottTodd: Matching exiting code from non-multi-arch CI: https://github.com/ROCm/TheRock/blob/bce0158bf62ab210183ace724f30b609cd986a86/.github/workflows/setup.yml#L95-L97

Will need to edit as part of https://github.com/ROCm/TheRock/issues/3334 (while it may not appear that way from the line count, I'm trying to keep this PR as focused as possible)

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: Would "highlights" make more sense than "callouts"? These are any parts of the CI configuration that are different from what runs by "default", such as when a label was added to opt-in to some extra builds or tests. I figured that if we only showed tables with the computed values, it wouldn't be obvious what about those values is worth paying attention to.

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: I'm adding prebuilt stage support in https://github.com/ROCm/TheRock/issues/3399. This PR continues the existing plumbing and prepares for more complicated logic in the configure script itself.

* Current: stages can be manually specified with `workflow_dispatch`, see the docs at https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#prebuilt-stages-multi-arch-ci
* Planned: `pull_request` and `push` events will compute which stages to use from which historical artifacts automatically or via some code

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: https://github.com/ROCm/TheRock/actions/runs/23515635067?pr=4123
image

I think the different coloring adds a lot here. Otherwise, the callouts/highlights would be easier to miss. We could just remove them entirely... idk, I'm experimenting here to see what we find useful.

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: Today yes, as we never skip the test_rocm job. That may change in the future.

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: I'll remove from the callouts/summary and just put in `_append_build_rocm`

.github/workflows/multi_arch_ci.yml
ScottTodd: I dropped it as part of collapsing the matrix that CI used. Will add it back.

Note that the workflow only supports a single variant at a time, with separate entry points (`ci.yml` vs `ci_asan.yml`) for separate variants. We _could_ add back job matrix support for multiple variants in a single workflow file, but the underlying scripts don't currently cleanly route artifacts to variant-specific subfolders for a shared workflow run id.

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: I think appending an empty line is clear enough, no need to indirect through a `NEW_LINE=""` constant.

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: Ack. Will need to port https://github.com/ROCm/TheRock/pull/2902 over to these new files. Keeping this unresolved for now.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Yes, new naming. See https://github.com/ROCm/TheRock/pull/3915

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Literal string. Comes directly from github `GITHUB_EVENT_NAME`. Will update the comments.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: This is optimizing for concise code in callers:

```python
    if ci_inputs.is_workflow_dispatch:
        pass
    elif ci_inputs.is_pull_request:
       pass
    elif ci_inputs.is_push:
        pass
```

compared to:

```python
    # prone to typos
    if ci_inputs.event_name == "workflow_dispatch":
        pass
    # verbose
    elif ci_inputs.event_name == EventName.PULL_REQUEST:
       pass
```

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: I'll switch these to lookups that will naturally fail if the environment variable is unset (it should always be set when running in github actions)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: I think "enabled" makes more sense here. "needed" is a reason to make a decision. Here the decision is already made, and future workflows just need to follow that instruction.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Similar response to `is_ci_enabled` vs `is_ci_needed`. This is what makes the decision, then once the decision is made it gets a different name. At least, that makes sense to me 😅

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Renaming test filters is out of scope for this PR.

.github/workflows/multi_arch_ci.yml
ScottTodd: Yes, but that change belongs in a separate PR.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Changes to test filtering are out of scope for this PR. Yes, we should have a target determinator for the ROCm test job (and the pytorch test job, etc.)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: That's how the inheritance hierarchy is set up.

Each argument here is a `JobGroupDecision`, which contains _at least_ a `JobAction`. Some stages inherit from `JobGroupDecision` to provide additional options like how `TestRocmDecision` has `test_type` and `test_type_reason` too

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Agreed. Keeping them here for now, then we should adjust build_tools/github_actions/amdgpu_family_matrix.py / build_tools/github_actions/new_amdgpu_family_matrix.py to provide the data that we need more directly, with appropriate validation and unit tests.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Copying existing logic: https://github.com/ROCm/TheRock/blob/abc5ff2b7e8dcd7767000e37cf2a9ef4c78de082/build_tools/github_actions/configure_ci.py#L384-L389

(changing that is out of scope for this PR)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Good catch. It should always be set, at least so long as test_rocm is enabled. I'll trim it down for now and then add boundary condition testing when tests can be skipped.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Agreed, but out of scope for this PR. That should go in `[new_]amdgpu_family_matrix.py` and this code should be simple plumbing.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Yes, it's still used to set `build_pytorch` true/false

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Different checks.

This is for the variant (release/asan/tsan), since we currently only have 'release' for Windows: https://github.com/ROCm/TheRock/blob/abc5ff2b7e8dcd7767000e37cf2a9ef4c78de082/build_tools/github_actions/amdgpu_family_matrix.py#L23-L51

Below is for the GPU target / family: https://github.com/ROCm/TheRock/blob/abc5ff2b7e8dcd7767000e37cf2a9ef4c78de082/build_tools/github_actions/amdgpu_family_matrix.py#L103-L125

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: That PR doesn't look relevant? Wrong link?

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: I'm just copying what configure_ci.py did for ci.yml here: https://github.com/ROCm/TheRock/blob/abc5ff2b7e8dcd7767000e37cf2a9ef4c78de082/build_tools/github_actions/configure_ci.py#L698-L704

Changing the filtering is out of scope for this PR, but yes I think we should refine over time. Hopefully the code here makes it easier to make such changes (dedicated `_determine_test_type()` function instead of mixing with all sorts of other code in `main()`).

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: I think this is correct.

* If you specify `test_filter:comprehensive` _and_ `test:rocprim`, you should run just rocprim's tests with the "comprehensive" filter
* If you specify just `test:rocprim`, you get "full" tests for rocprim
* If you specify just `test_filter:comprehensive`, you get "comprehensive" tests for all components

Is there a different case you want supported?

build_tools/github_actions/configure_multi_arch_ci_summary.py
ScottTodd: Added `test_runner` label support

ScottTodd:
> i would recommend splitting this in multiple files. you already have 5 distinct steps + the data classes.

I think splitting files will make more sense once we start filling in new logic for the build and test jobs. As is, this is _just_ porting the existing configure_ci.py logic to a new framework.

ScottTodd:
Addressed all comments. PTAL?

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: oh no that's right!

build_tools/github_actions/configure_multi_arch_ci_summary.py
HereThereBeDragons: ping pong with claude suggests:
_config_overrides
_override_notes
_active_overrides

build_tools/github_actions/configure_multi_arch_ci_summary.py
HereThereBeDragons: if it is max 2-3 values ok, but if we have 20 i think it will be too much

build_tools/github_actions/configure_multi_arch_ci.py
HereThereBeDragons: i remember that logic. maybe worth a small comment?
i remember that was added to make it easier for users to call the workflow manually - but i also know that in the new amdgpu family matrix we wont the full match.

build_tools/github_actions/configure_multi_arch_ci.py
HereThereBeDragons: yes.. not sure how that happened: https://github.com/ROCm/TheRock/pull/3653
the amdgpu refactor

build_tools/github_actions/configure_multi_arch_ci.py
HereThereBeDragons: i think my argument here was you already have that check indirectly in `_expand_build_config_for_platform()`

because there is
```
per_family_info: list[dict] = []
for family_name in families:
        # ...
        if build_variant not in platform_info["build_variants"]:
            print(
                f"  Family {family_name} does not support variant "
                f"{build_variant} on {platform}, skipping"
            )
            continue
```

and later
```
 if not per_family_info:
        return None
```

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Okay yeah. I'd like to move more of the logic and validation into separate modules like that, so thanks for your work there (and patience with all the comments 😅)

ScottTodd:
> ## Test Plan
> * Watch multi-arch CI behavior for `push` events after merge

Looks good: https://github.com/ROCm/TheRock/actions/runs/23810551251.
* Same set of targets, test types, etc. as on the pull request, 
* Trigger detected as "push"

PR #4125, opened 2026-03-24
[RFC] race-emulator: new rocm-libraries/shared component

newling:
Discussion: https://github.com/ROCm/TheRock/discussions/4124

## Motivation

This is an RFC for a new component in rocm-libraries/shared, with a new github workflow.

## Technical Details

The RFC is added as a new file in docs/rfcs. It is chosen as 0012 instead of 0011 as rfc-0011 accepted is expected to land. 

## Test Plan

RFCs don't need testing, testing of proposal project discussed in RFC!

## Submission Checklist

### DIFF
--- a/docs/rfcs/README.md
+++ b/docs/rfcs/README.md
@@ -26,6 +26,7 @@ benefit from further discussion.
 - [RFC0008: Multi-Arch Packaging](./RFC0008-Multi-Arch-Packaging.md)
 - [RFC0009: OS Packaging Requirements](./RFC0009-OS-Packaging-Requirements.md)
 - [RFC0010: Test Scripts Migration to Owning Repositories](./RFC0010-Test-Scripts-Migration.md)
+- [RFC0012: Race Emulator](./RFC0012-Race-Emulator.md)
 
 ## Adding an RFC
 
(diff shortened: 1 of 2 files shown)


stellaraccident (commented):
Can we discuss at office hours. Not opposed but there are several projects like this happening in various quarters and I'd like to better understand where this is going.

newling:
This will be discussed in the ROCm office hours today at 11-12 PST, ping me if you can't make this and would still like to discuss

newling:
To summarize the discussion (and @stellaraccident can correct me if I misinterpreted):

This new component offers value and will improve hipblaslt very soon after being added (my team's claim). As such, even if it is only around for 6 months, we probably want to have it in rocm-libraries. At the same time, duplicating the effort of building an emulator is not good. And so I need to think about strategies for merging this with similar efforts, in particular [rocjitsu](https://github.com/ROCm/rocm-systems/blob/3899ac42d3df81a55f0ccea6c2dc72740329994b/experimental/rocjitsu/README.md). I will meet with @atgutier to make sure the design or rocjitsu is aligned with our needs, and make an effort to keep the 'race-' and 'emulator' of 'race-emulator' as independent as possible so that the 'emulator' part can, in theory, be dropped, and the 'race-' part can be reused.

newling:
The race detection tool implemented here has moved to rocjitsu.

PR #4135, opened 2026-03-24
Revert "Use TEST_TYPE directly as categories for test filtering"

aalabsi-amd:
Reverts ROCm/TheRock#4008

This caused issues in miopen CI running the unintended `full` category instead of the `standard`.
CI is blocking PRs now, need to unblock until we address this change from miopen.

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_runner.py
+++ b/build_tools/github_actions/test_executable_scripts/test_runner.py
@@ -8,8 +8,7 @@ TEST_COMPONENT: Job name of the component to test (e.g., "miopen", "rocrand", "h
     This is automatically set by the GitHub Actions workflow from the job_name field.
     The script maps these job names to actual test directory names (e.g., "miopen" -> "MIOpen")
     Defaults to "miopen" if not set.
-TEST_TYPE: Test category to run - one of "quick", "standard", "comprehensive", or "full".
-    Defaults to "quick". Invalid values fall back to "quick" with an error message.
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -221,15 +219,11 @@ def build_ctest_command(category, gpu_arch, available_gpu_archs):
...
 
 def main():
-    category = TEST_TYPE.lower() if TEST_TYPE else "quick"
-    if category not in VALID_TEST_CATEGORIES:
-        print(
-            f"ERROR: Invalid TEST_TYPE '{TEST_TYPE}'. "
-            f"Must be one of: {', '.join(sorted(VALID_TEST_CATEGORIES))}. "
-            f"Falling back to 'quick'.",
-            file=sys.stderr,
-        )
```
ScottTodd, line 231: I also came across this code yesterday. This should not have been merged as-is. This should have raised an exception instead of silently falling back to "quick" (stderr does not count if no one is checking it)

See the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#fail-fast-behavior

ScottTodd: (then we could also have a unit test checking that only valid test categories are used, so we don't need to wait until late in the CI pipeline to see issues)

PR #4136, opened 2026-03-24
Fix workflow input descriptions for build_*_python_packages.yml

ScottTodd:
## Motivation

Actual inputs did not match the descriptions

Input | Issue
-- | --
`amdgpu_families` | https://github.com/ROCm/TheRock/pull/3854 changed `artifact_manager.py` to accept a semicolon-delimited list instead of a comma-delimited list but these input descriptions still suggested comma
`artifact_group` | multi-arch CI uses "artifact-group" values like `multi-arch-release`: https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/github_actions/configure_ci.py#L240 https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/github_actions/configure_ci.py#L526-L531

## Technical Details

Example workflow run, issues highlighted with `+`:
https://github.com/ROCm/TheRock/actions/runs/23356803838/job/67968231459#step:1:15
```diff
 Inputs
    artifact_github_repo: 
    artifact_run_id: 
+   artifact_group: multi-arch-release
+   amdgpu_families: gfx94X-dcgpu;gfx1151;gfx120X-all
    multiarch_index: true
    package_version: 7.13.0.dev0+8027b7b9eb0d87c9bb0d28ee6a0902a1047feb27
    linux_amdgpu_families: gfx94X, gfx1151, gfx120X
    linux_test_labels: test:aqlprofile
```

## Submission Checklist

### DIFF
--- a/.github/workflows/build_portable_linux_python_packages.yml
+++ b/.github/workflows/build_portable_linux_python_packages.yml
@@ -18,7 +18,7 @@ on:
         description: "artifact_group to build (e.g. 'gfx94X-dcgpu' for single family builds, 'multi-arch-release' for multi-arch builds with 'amdgpu_families' set)"
         type: string
       amdgpu_families:
-        description: "amdgpu_families for multi-arch builds as a semicolon-separated list (e.g. 'gfx94X-dcgpu;gfx120X-all'). Leave empty for single-family builds (those only use artifact_group)."
+        description: "amdgpu_families for multi-arch builds as a semicolon-separated list (e.g. 'gfx94X-dcgpu;gfx120X-all'). Leave empty for single-family builds (those only use artifact_group)"
         type: string
         default: ""
       multiarch_index:
--- a/.github/workflows/build_windows_python_packages.yml
+++ b/.github/workflows/build_windows_python_packages.yml
@@ -18,7 +18,7 @@ on:
         description: "artifact_group to build (e.g. 'gfx94X-dcgpu' for single family builds, 'multi-arch-release' for multi-arch builds with 'amdgpu_families' set)"
         type: string
       amdgpu_families:
-        description: "amdgpu_families for multi-arch builds as a semicolon-separated list (e.g. 'gfx94X-dcgpu;gfx120X-all'). Leave empty for single-family builds (those only use artifact_group)."
+        description: "amdgpu_families for multi-arch builds as a semicolon-separated list (e.g. 'gfx94X-dcgpu;gfx120X-all'). Leave empty for single-family builds (those only use artifact_group)"
         type: string
         default: ""
       multiarch_index:


.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -15,10 +15,10 @@ on:
         type: string
         default: "17865324892" # TODO: default to the most recent successful run (using a script)
       artifact_group:
-        description: "The artifact group to build (ex: gfx94X-dcgpu, gfx101X-dgpu, gfx1151, gfx120X-all)"
+        description: "The artifact group to build (ex: gfx94X-dcgpu, gfx120X-all, multi-arch-release)"
```
marbre: Not sure if `multi-arch-release` works if `amdgpu_families` is falling back to `amdgpu_groups`.

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -37,7 +37,7 @@ on:
       artifact_group:
```
marbre, line 38: Should repro the above description here as well?

.github/workflows/build_windows_python_packages.yml
```diff
@@ -37,7 +37,7 @@ on:
       artifact_group:
```
marbre, line 38: As above.

.github/workflows/build_portable_linux_python_packages.yml
marbre: From my agent's review:

----

## Review comment r2986763464 — `multi-arch-release` in `artifact_group` description

**Comment (marbre):** "Not sure if `multi-arch-release` works if `amdgpu_families` is
falling back to `amdgpu_groups`."

**Assessment: Valid concern.**

### The two code paths

The fetch step in these workflows branches on whether `amdgpu_families` is set:

```bash
if [ -n "${{ inputs.amdgpu_families }}" ]; then
    # Multi-arch path → artifact_manager.py --amdgpu-families (semicolon-sep)
else
    # Single-family path → fetch_artifacts.py --artifact-group="${{ inputs.artifact_group }}"
fi
```

### Why `multi-arch-release` is wrong as a standalone `artifact_group`

`multi-arch-release` is assigned as the `artifact_group` for multi-arch CI runs in
`configure_ci.py:240`:

```python
"artifact_group": f"multi-arch-{info.get('build_variant_suffix') or 'release'}",
```

But this value is only meaningful on the **upload side** (`upload_python_packages.py`
uses it as a path segment in the S3 key). On the **fetch side**, `fetch_artifacts.py`
matches artifacts by `ArtifactName.target_family` parsed from filenames. No artifact
files are tagged `multi-arch-release` — they're tagged with real family names like
`gfx94X-dcgpu` or `gfx120X-all`.

So if a user passes `artifact_group=multi-arch-release` with `amdgpu_families` left
empty, the workflow falls into the single-family path, calls:

```
fetch_artifacts.py --artifact-group=multi-arch-release
```

...and finds nothing. The only multi-arch references in the codebase outside of
`configure_ci.py` are in a test asserting that the upload path does *not* create a
`multi-arch-release` subdirectory.

### What the PR should do instead

Either:
- Remove `multi-arch-release` from the `artifact_group` example — it is not a valid
  standalone value for these fetch workflows.
- Or, keep it but add an explicit note that `artifact_group=multi-arch-release` requires
  `amdgpu_families` to also be non-empty (otherwise the fetch falls into the wrong path).
- Better still: the `artifact_group` description should clarify it is only used in
  **single-family mode** (i.e., when `amdgpu_families` is empty).

ScottTodd: I wanted to document 1:1 the behavior that the CI already has today (which is most often what you should do when triggering a workflow manually), but I see the point about marking which fields are expected together in which combinations.

.github/workflows/build_portable_linux_python_packages.yml
ScottTodd: I'd rather only put descriptions on `workflow_dispatch` instead of duplicating on `workflow_call`

.github/workflows/build_portable_linux_python_packages.yml
ScottTodd: How's this?
image

.github/workflows/build_portable_linux_python_packages.yml
marbre: Removing the others is also a good idea! :)

PR #4140, opened 2026-03-24
Disable ASAN instrumentation for third-party dependencies

sa-faizal:
## Motivation

This PR disables ASAN (AddressSanitizer) instrumentation for third-party components in the CMake build configuration. The goal is to prevent Address Sanitizer from instrumenting external dependencies, which can cause false positives, increased build times, and potential compatibility issues when these libraries are built with different instrumentation settings than the main ROCm codebase.

## Technical Details

Modified `CMakePresets.json` to disable ASAN instrumentation for the following third-party libraries by setting their respective `*_SANITIZER` cache variables to "OFF":

- boost
- Catch2
- googletest
- eigen
- fftw3
- flatbuffers
- fmt
- frugally-deep
- FunctionalPlus
- grpc
- host-blas
- libdivide
- msgpack-cxx
- nlohmann-json
- simde
- SuiteSparse
- yaml-cpp

This ensures that these external dependencies are built without ASAN instrumentation while allowing the main ROCm components to maintain their sanitizer settings independently.

## Test Plan

- Verified CMakePresets.json is valid JSON syntax
- Built TheRock with the new configuration to ensure all third-party components configure and compile successfully
- Confirmed that the sanitizer settings do not affect the main ROCm build process

## Test Result

All builds completed successfully. Third-party components compile without ASAN instrumentation, and the main ROCm codebase can still apply its own sanitizer settings as appropriate.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -43,7 +43,24 @@
         "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O2 -g1",
         "CMAKE_C_FLAGS_RELWITHDEBINFO": "-O2 -g1",
         "CMAKE_CXX_FLAGS": "-g1",
-        "CMAKE_C_FLAGS": "-g1"
+        "CMAKE_C_FLAGS": "-g1",
+        "therock-boost_SANITIZER": "OFF",
+        "therock-catch2_SANITIZER": "OFF",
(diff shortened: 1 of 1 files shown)


sa-faizal:
converted to draft because of revert https://github.com/ROCm/TheRock/pull/4142

PR #4153, opened 2026-03-25
Bump expat to version 2.7.5

marbre:
Fixes multiple security issues, see changelogs:
* https://github.com/libexpat/libexpat/blob/R_2_7_5/expat/Changes
* https://github.com/libexpat/libexpat/blob/R_2_7_4/expat/Changes

### DIFF
--- a/third-party/sysdeps/common/expat/CMakeLists.txt
+++ b/third-party/sysdeps/common/expat/CMakeLists.txt
@@ -70,7 +70,7 @@ endif()
 
 # Otherwise, this is the sub-project build.
 cmake_minimum_required(VERSION 3.25)
-project(EXPAT_BUILD VERSION "2.7.3")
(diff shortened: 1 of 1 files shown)


marbre:
Linux builds passing, Windows is unaffected.

lumachad:
Based on my comment in #4158, I see we also skipped rocgdb tests for this PR, which is suspicious. We want to see rocgdb tests passing to make sure things are sane. Do we know why they were skipped?

marbre:
> Based on my comment in #4158, I see we also skipped rocgdb tests for this PR, which is suspicious. We want to see rocgdb tests passing to make sure things are sane. Do we know why they were skipped?

Missed that, sorry :( I think test filtering might be the issue here. Anyway, there is no way around bumping as there are several CVEs for the old version. Let's monitor post-submit.

PR #3244, opened 2026-02-04
[CI]Enable ccache for nightly portable Linux builds

chiranjeevipattigidi:
## Summary
This PR introduces ccache support for Linux nightly (non-dev) builds to significantly reduce compilation time through cached build artifact reuse. This optimization aims to accelerate CI pipeline execution, reduce computational resource consumption, and provide faster build feedback.
## Motivation
The current Linux nightly builds utilize GitHub's hosted cache system, which presents two key limitations:
- Reliability issues: Inconsistent cache availability affecting build performance
- Storage constraints: Limited cache size restricting optimization potential

This PR addresses these limitations by migrating to a dedicated bazel-remote server hosted within our infrastructure cluster.

## Files Changed
- `.github/workflows/release_portable_linux_packages.yml` - Updated workflow configuration to enable ccache
- `build_tools/setup_ccache.py` - added new bazelremote service instance

## Technical Approach
- Implemented bazel-remote server integration for improved cache reliability and capacity
- Configuration details available at:https://github.com/nod-ai/GitHub-ARC-Setup/commit/fb76c4e20197cd671f605e6f8d4309fb779f94cc

Related Work:
- Windows ccache enablement is currently in progress: [PR #2415](https://github.com/ROCm/TheRock/pull/2415)

## Test Results
Conducted two test runs to validate performance improvements:
- Run 1: [Build #22711400987](https://github.com/ROCm/TheRock/actions/runs/22711400987)
- Run 2: [Build #22723667746](https://github.com/ROCm/TheRock/actions/runs/22723667746)

### Build Time Improvements
|Arch|Baseline|With ccache|Improvement|
|-----|------|-------|-------|
|**gfx120x**|4h31m|1h16m|~71%|
|**gfx103**|3h1m|2h39m|~12%|

✅ Confirmed cache hit rates improve with subsequent builds
✅ Verified ccache functionality across multiple build configurations

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #2415, opened 2025-12-04
[CI] Use windows cluster bazelremote service for builds

amd-justchen:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/902 but is the Windows counterpart to https://github.com/ROCm/TheRock/pull/930
Removes the dependency on Github cache servers and their cache size limits. Standing this up in anticipation for AWS migration.

## Technical Details

Switch Windows builders to use a self hosted cache server local to the CPU builder cluster that is defined in this https://github.com/nod-ai/GitHub-ARC-Setup/pull/22 pr.

## Test Plan

Wait for CI

## Test Result

TBD from CI

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

Issue #1049, opened 2025-07-16
[ci] MI355 runner hanging during tests


geomin12:
From [CI workflow](https://github.com/ROCm/TheRock/actions/runs/16305988414/job/46087953062).

For GitHub runners, we notice these errors:

```
FAILED tests/test_rocm_sanity.py::TestROCmSanity::test_hip_printf - Exception: b'Memory access fault by GPU node-2 (Agent handle: 0x562d2b5c1610) on address (nil). Reason: Write access to a read-only page.\n'
```

For ssh testing, the executable hangs at [`hipLaunchKernelGGL`](https://github.com/ROCm/TheRock/blob/142fcbc288251e34585535eab941191a1837dd73/tests/hipcc_check.cpp#L14) after hipcc compiles that file

To reproduce for hipcc test:

```
$ git clone https://github.com/ROCm/TheRock.git
$ cd TheRock
$ python3 build_tools/install_rocm_from_artifacts.py --run-id 16350224221 --amdgpu-family gfx950-dcgpu --tests
$ export THEROCK_BIN_DIR=./therock-build/bin/
$ ./${THEROCK_BIN_DIR}/hipcc tests/hipcc_check.cpp -Xlinker -rpath=${THEROCK_BIN_DIR}../lib/ -o test_hipcc
$ ./test_hipcc
```

To reproduce for ctest or gtest:
```
$ git clone https://github.com/ROCm/TheRock.git
$ cd TheRock
$ python3 build_tools/install_rocm_from_artifacts.py --run-id 16350224221 --amdgpu-family gfx950-dcgpu --tests
$ sudo apt install cmake
 // for ctest
$ cd therock-build/bin/rocprim
$ ctest
// for gtest
cd therock-build/bin
$ ./rccl-UnitTests
```

artifacts pulled from [this commit](https://github.com/ROCm/TheRock/commit/8189f38d44874f2dacd35fe505685e3b39ee34a7)

MI355 were working on [this commit](https://github.com/ROCm/TheRock/commit/f686688e59531b5ea59c67ffd49b1e2ac7a15d54), but failed on [this commit](https://github.com/ROCm/TheRock/commit/daafcd0b82a1cb2484c6c3c907d179377900c3eb)

geomin12:
After investigation with @saienduri , the ROCm 7.0 bump requires a dkms upgrade past 30.10_alpha2 (which is what was installed on mi355). This is what caused the hang, and after upgrade, tests passed

PR #442, opened 2025-04-18
Adding 1101 target

geomin12:
Adding gfx110X target

Status as of 5/8/2025 - rocPRIM tests take a significantly long time, resulting in time outs. Working with rocPRIM team to resolve and ignore high memory tests

Closes #438

PR #4161, opened 2026-03-25
Enable ccache + bazel-remote caching for Windows builds

subodh-dubey-amd:
## Motivation

Windows multi-arch CI builds have **no caching** -- every stage compiles from scratch. The old single-job Windows workflows use unreliable GitHub Actions cache (10GB limit, unpredictable eviction). Linux multi-arch CI already uses ccache + bazel-remote with 71% improvement (#3244). This PR brings the same infrastructure to all Windows workflows, closing the gap from #2415.

## Technical Details

- **`setup_ccache.py`**: Made cross-platform -- skips POSIX-only compiler fingerprinting (`ldd`/`sha256sum`) on Windows; ccache default `mtime` check works with MSVC.
- **`multi_arch_build_windows_artifacts.yml`**: Added ccache install, `setup_ccache.py` with bazel-remote, and `CMAKE_*_COMPILER_LAUNCHER=ccache`. Previously had zero caching.
- **`build_windows_artifacts.yml`**: Replaced `actions/cache` with `setup_ccache.py` + bazel-remote.
- **`release_windows_packages.yml`**: Same migration, using presubmit/postsubmit presets based on release type.

### 1. Multi-Arch CI: `main` (no caching) vs this PR (ccache + bazel-remote)

Baseline on main: [run #23577483606](https://github.com/ROCm/TheRock/actions/runs/23577483606) (no caching)
This PR: [run #1049](https://github.com/ROCm/TheRock/actions/runs/23558629177) (ccache + bazel-remote)

| Stage | `main` (no caching) | This PR | Remote Cache Hits | Delta |
|---|---|---|---|---|
| foundation | [6m](https://github.com/ROCm/TheRock/actions/runs/23577483606/job/68652966009) | [6m](https://github.com/ROCm/TheRock/actions/runs/23558629177/job/68592512368) | 986 / 1,002 (98.4%) | -- |
| compiler-runtime | [**69m**](https://github.com/ROCm/TheRock/actions/runs/23577483606/job/68653739959) | [**38m**](https://github.com/ROCm/TheRock/actions/runs/23558629177/job/68595011358) | **9,991 / 10,042 (99.5%)** | **-31m (-45%)** |
| math-libs gfx1151 | [142m](https://github.com/ROCm/TheRock/actions/runs/23577483606/job/68659480194) | [119m](https://github.com/ROCm/TheRock/actions/runs/23558629177/job/68602876177) | -- | -23m |

### 2. Release Windows Packages: cold vs warm cache

| Metric | [Run 1 - cold cache](https://github.com/ROCm/TheRock/actions/runs/23537176310/job/68515372666) | [Run 2 - warm cache](https://github.com/ROCm/TheRock/actions/runs/23545004200/job/68542765533) |
|---|---|---|
| **Total duration** | **3h 17m** | **2h 43m (-17%)** |
| Cacheable calls | 17,734 / 27,259 (65%) | 17,734 / 27,259 (65%) |
| Overall cache hits | 8,997 / 17,734 (50.7%) | 9,002 / 17,734 (50.8%) |
| Remote storage hits | 8,923 / 17,660 (50.5%) | 8,929 / 17,661 (50.6%) |
| Remote timeouts | 0 | 1 |

### 3. All Windows build stages passed

- [Multi-arch #821](https://github.com/ROCm/TheRock/actions/runs/23436418752): foundation, compiler-runtime, math-libs -- all passed
- [Multi-arch #1049](https://github.com/ROCm/TheRock/actions/runs/23558629177): all 5 Windows build stages passed (failures are unrelated Linux/Windows test jobs)
- [Release #440](https://github.com/ROCm/TheRock/actions/runs/23537176310) and [#442](https://github.com/ROCm/TheRock/actions/runs/23545004200): both passed

## Submission Checklist

### DIFF
--- a/.github/workflows/build_windows_artifacts.yml
+++ b/.github/workflows/build_windows_artifacts.yml
@@ -193,4 +193,3 @@ jobs:
             --artifact-group ${{ inputs.artifact_group }} \
             --build-dir ${{ env.BUILD_DIR }} \
             --upload
-
--- a/.github/workflows/release_windows_packages.yml
+++ b/.github/workflows/release_windows_packages.yml
@@ -351,4 +351,3 @@ jobs:
               "rocm_version": "${{ needs.setup_metadata.outputs.version }}",
               "ref": "${{ inputs.ref || '' }}"
             }
-


ScottTodd:
If you want to test multi-arch CI on this PR, add the `ci:run-multi-arch` label. See the docs at https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

build_tools/setup_ccache.py
```diff
@@ -153,7 +160,10 @@ def run(args: argparse.Namespace):
                 file=sys.stderr,
             )
     # Output options.
-    print(f"export CCACHE_CONFIGPATH={config_file}")
+    if IS_WINDOWS:
+        print(f"set CCACHE_CONFIGPATH={config_file}")
```
amd-justchen, line 164: Just to note, the workflow right now actually might be running `setup_ccache.py` with bash even in Windows, so probably should check which shell is running and use the appropriate environment variable command. However, our current Linux and Windows CI workflow does not use the result to set the environment variable for the current shell, so it might be okay and this could be addressed in a follow up PR.
```yaml
      - name: Setup ccache
        run: |
          python ./build_tools/setup_ccache.py \
```
```yaml
jobs:
...
    defaults:
      run:
        shell: bash
```

Typical use per header comments of `setup_ccache.py`:
```
Typical usage for the current shell (will set the CCACHE_CONFIGPATH var):
    eval "$(./build_tools/setup_ccache.py)"
```

build_tools/setup_ccache.py
```diff
@@ -87,11 +91,14 @@ def gen_config(dir: Path, compiler_check_file: Path, args: argparse.Namespace):
...
 
-    # Compiler check.
-    lines.append(
-        f"compiler_check = {sys.executable} {compiler_check_file} "
-        f"{dir / 'compiler_check_cache'} %compiler%"
-    )
+    # Compiler check: on POSIX we use a custom script that fingerprints the
+    # compiler binary and its shared libraries via ldd + sha256sum. On Windows
+    # (MSVC) those tools don't exist; ccache's default mtime check works well.
+    if not IS_WINDOWS:
```
amd-justchen, line 97: This might be necessary also on Windows to allow for more ccache hits. Will have to confirm with @stellaraccident (Can be addressed in a follow up PR)

ScottTodd (approved):
Great, thanks!

Follow-up tasks for us:
* When we set up new release pipelines based on the multi-arch workflows, we should be careful about the ccache config preset ("presubmit" vs "postsubmit"). Might want to split further to presubmit (`pull_request` trigger) / postsubmit (`push` trigger) / release (`schedule` trigger`), with `workflow_dispatch` trigger doing... something
* Check https://github.com/ROCm/TheRock/actions/caches over time to see what else is left. We'll hopefully stop rolling over the 10GB limit after this so the jobs that still do use github actions cache should see better performance (pre-commit, maybe python package installation?)

.github/workflows/build_windows_artifacts.yml
```diff
@@ -71,8 +71,7 @@ jobs:
     env:
       BUILD_DIR: B:\build
       CACHE_DIR: "${{github.workspace}}/.cache"
-      CCACHE_DIR: "${{github.workspace}}/.cache/ccache"
-      CCACHE_MAXSIZE: "4000M"
+      CCACHE_CONFIGPATH: "${{github.workspace}}/.ccache/ccache.conf"
       TEATIME_FORCE_INTERACTIVE: 0
```
amd-justchen, line 75: Due to peculiarities with the Windows setup, this may not use the local temporary disk (mounted on drive `B:`) The default workspace may only be on the OS disk which is limited in size (30GB), and may run into disk space issues along with performance issues (IOPS / Storage) if ccache tries to write locally (if for some reason misses on remote).

It could be slated for a follow up PR for now since it does technically work, but ideally should be stored on B: like my original PR (#2415).

subodh-dubey-amd: https://github.com/ROCm/TheRock/issues/4195 - Created a follow up issue to address all these comments.

Issue #4040, opened 2026-03-18
[Issue]: `rocm-sdk` test `testCLIUsesDevelRootPath` fails on nightlies and 7.12.0rc0/1 on RHEL 10.1


pbhandar-amd:
### Problem Description

I am getting two errors when running `rocm-sdk test` after installing 7.12.0 rc0 on RHEL 10.1.

```
======================================================================
ERROR: testConsoleScripts (rocm_sdk.tests.core_test.ROCmCoreTest.testConsoleScripts) [Check console-script amdflang]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/rocm/rocm-install/.venv/lib64/python3.12/site-packages/rocm_sdk/tests/core_test.py", line 141, in testConsoleScripts
    output_text = subprocess.check_output(
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '[PosixPath('/home/rocm/rocm-install/.venv/bin/amdflang'), '--help']' returned non-zero exit status 255.

======================================================================
FAIL: testCLIUsesDevelRootPath (rocm_sdk.tests.devel_test.ROCmDevelTest.testCLIUsesDevelRootPath)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/rocm/rocm-install/.venv/lib64/python3.12/site-packages/rocm_sdk/tests/devel_test.py", line 87, in testCLIUsesDevelRootPath
    self.assertEqual(
AssertionError: Posix[21 chars]-install/.venv/lib64/python3.12/site-packages/_rocm_sdk_devel') != Posix[21 chars]-install/.venv/lib/python3.12/site-packages/_rocm_sdk_devel') : Expected `hipconfig --rocmpath` to return /home/rocm/rocm-install/.venv/lib64/python3.12/site-packages/_rocm_sdk_devel, not /home/rocm/rocm-install/.venv/lib/python3.12/site-packages/_rocm_sdk_devel
```

[...]

PR #4164, opened 2026-03-25
[rocm_sdk/tests] Compare devel root vs. hipconfig with samefile

lucbruni-amd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/4040 (the latter point)

`testCLIUsesDevelRootPath` asserted that `rocm_sdk path --root` and `hipconfig --rocmpath` return identical path strings. On common Linux `venv` layouts (e.g. RHEL-style images where `lib64` is a symlink to `lib`), both commands still refer to the same `_rocm_sdk_devel` tree, but one path may be spelled with `lib64/...` and the other after `realpath`-style resolution as `lib/...`. That produced a false failure though the install was correct.

## Technical Details

- Replace strict `Path` equality with `os.path.samefile(root_path, rocmpath)` so we compare the same directory on disk, not the string form.
- `assertTrue(..., is_dir())` on both paths before `samefile` so failures are clear if either side is missing or not a directory.

## Test Plan

- Run `rocm-sdk test` from a `venv` with `rocm[devel]` installed

## Test Result

- Confirmed `testCLIUsesDevelRootPath` passes when `path --root` and `--rocmpath` differ only by symlink spelling.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/devel_test.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/devel_test.py
@@ -84,7 +84,9 @@ class ROCmDevelTest(unittest.TestCase):
             .strip()
         )
         rocmpath = Path(rocmpath_output)
-        self.assertTrue(root_path.is_dir(), msg=f"Expected root path {root_path} to exist")
+        self.assertTrue(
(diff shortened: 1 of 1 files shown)


marbre:
@lucbruni-amd don't forget to assign an reviewer next time :)

lucbruni-amd:
Merged, thanks @marbre for the review and reminder. Will do!

PR #4167, opened 2026-03-25
STACKED PR. Add s3_url_schema support and remove THEROCK_HTTPS_URL_ env var

PeterCDMcLean:
- Add s3_url_schema parameter to artifact_manager.py and post_build_upload.py
- Pass s3_url_schema through all backend creation calls
- Remove deprecated THEROCK_HTTPS_URL_ environment variable override
- Update tests to remove env var tests and add comprehensive schema tests
- Add tests for custom S3/HTTPS schemas and bucket schema
- Update workflow_outputs.md to document new schema system
- Document command line usage for all three schema parameters

All three URL schemas (s3_url_schema, https_url_schema, bucket_schema) are now configurable via command line arguments and use a single source of truth for defaults.

Use if we prefer threading all these arguments through instead of magic env vars

### DIFF
--- a/build_tools/_therock_utils/artifact_backend.py
+++ b/build_tools/_therock_utils/artifact_backend.py
@@ -30,7 +30,6 @@ import shutil
 import urllib.error
 import urllib.request
 
-from .storage_location import StorageLocation
 from .workflow_outputs import WorkflowOutputRoot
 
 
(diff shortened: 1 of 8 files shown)


ScottTodd (commented):
Marking as "reviewed" to drop this old PR from my review queue. LMK if you want to continue this work in some form (and then please request review from another developer first, I don't have bandwidth to do so many first level reviews)

PR #4138, opened 2026-03-24
Fix URL encoding in multi-arch Python package index and add logging

ScottTodd:
## Motivation

Multi-arch CI jobs like https://github.com/ROCm/TheRock/actions/runs/23356803838/job/67968231459 produce index pages like https://therock-ci-artifacts.s3.amazonaws.com/23356803838-linux/python/gfx1151/index.html with URLs like https://therock-ci-artifacts.s3.amazonaws.com/23356803838-linux/python/rocm_sdk_core-7.13.0.dev0+8027b7b9eb0d87c9bb0d28ee6a0902a1047feb27-py3-none-linux_x86_64.whl which 404 due to the `+` character which should be escaped as `%2B`.

I also took the chance to add more logging of what files were uploaded, so we can verify that 404s aren't upload failures.

## Technical Details

Here is where the package uploading code branches between index scripts:
https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/build_tools/github_actions/upload_python_packages.py#L98-L105

The `indexer.py` script used by the single-arch CI code path already escapes here: https://github.com/ROCm/TheRock/blob/7cb75f20690d0b49b33b616570c6346cc99033a0/third-party/indexer/indexer.py#L346

## Test Plan

* New unit test
* CI on the PR itself (check logs, index page)
* Manual testing ahead of the PR
  1. Rebase this PR on https://github.com/ROCm/TheRock/pull/4122
  2. Trigger a workflow run with inputs matching those on https://github.com/ROCm/TheRock/actions/runs/23356803838/job/67968231459#step:1:15
      ```
       Inputs
          artifact_github_repo: 
          artifact_run_id: 
          artifact_group: multi-arch-release
          amdgpu_families: gfx94X-dcgpu;gfx1151;gfx120X-all
          multiarch_index: true
          package_version: 7.13.0.dev0+8027b7b9eb0d87c9bb0d28ee6a0902a1047feb27
      ```
  3. Check the generated index pages and try downloading files

## Test Result

Manual testing: https://github.com/ROCm/TheRock/actions/runs/23504275622

<details><summary>URL encoding comparison</summary>
<p>

Baseline index: https://therock-ci-artifacts.s3.amazonaws.com/23356803838-linux/python/gfx1151/index.html
[...]

PR #4168, opened 2026-03-25
Split out index file generation from post upload script

marbre:
## Summary

Removes the client-side HTML index generation from `post_build_upload.py` and 
introduces a server-side replacement as a standalone script to be called from an
AWS Lambda handler.

## Technical details

* `post_build_upload.py`: Modifies the script to only uploads artifacts, logs, and
  the manifest; index generation no longer runs on the CI runner.

* `build_tools/generate_s3_index.py`: Adds as new standalone script and library for
  server-side index generation. The core function
  ` generate_index_for_directory(bucket, dir_prefix, backend)` lists the immediate files
  in one S3 (or local) directory and writes an `index.html`. Subdirectories appear as
  navigable entries linking to their own `index.html`; The cli discovers every
  directory containing files under the run prefix and indexes each one, handling both
  single-arch (`logs/{group}/`) and multi-arch (`logs/{stage}/{family}/`) layouts without
  structural assumptions. A `--bucket` flag allows overriding the bucket resolved from 
  GHA environment variables. 

* `_therock_utils/workflow_outputs.py`: Moves the `github_actions_api` import to
  inside `_retrieve_bucket_info()` (used only when `lookup_workflow_run=True`)
  from the module-level; this keeps the Lambda deployment package free of
  `github_actions_api` as a dependency.

## Not part of this PR

* Lambda handler triggered by S3 PutObject events which computes the parent directory
  and calls `generate_index_for_directory` for each uploaded object while skipping
  `index.html` files to prevent loops. The deployment requires `generate_s3_index.py` +
  `_therock_utils/storage_backend.py` + `_therock_utils/storage_location.py`.
* Removal of third-party `indexer.py` script.

## Design notes

* Index generation is now purely reactive: one S3 event, one `index.html` generated
  for the directory that received the upload.
* No artifact-group awareness needed; any new directory type (`python/`, `reports/`,
  etc.) is automatically indexed.

## Testing

```bash
python -m pytest build_tools/tests/generate_s3_index_test.py \
                 build_tools/tests/workflow_outputs_test.py \
                 build_tools/github_actions/tests/post_build_upload_test.py
```

Progress on #3331.

### DIFF
--- a/build_tools/_therock_utils/workflow_outputs.py
+++ b/build_tools/_therock_utils/workflow_outputs.py
@@ -324,6 +324,7 @@ def _retrieve_bucket_info(
     # all environments (e.g. local dev without the GHA support package installed).
     if workflow_run is None and workflow_run_id is not None:
         from github_actions.github_actions_api import gha_query_workflow_run_by_id
+
         workflow_run = gha_query_workflow_run_by_id(github_repository, workflow_run_id)
 
     # Extract metadata from workflow_run if available
--- a/build_tools/github_actions/post_build_upload.py
+++ b/build_tools/github_actions/post_build_upload.py
@@ -48,7 +48,6 @@ def log(*args):
     sys.stdout.flush()
 
 
-
 # This method will output logs of the Windows Time Service and is meant
 # to help debug spurious AWS auth issues caused by time differences when
 # uploading with the AWS CLI tool. For context, see this issue and PR:


build_tools/_therock_utils/workflow_outputs.py
```diff
@@ -297,6 +296,7 @@ def _retrieve_bucket_info(
 
     # Fetch workflow_run from API if not provided but workflow_run_id is set
     if workflow_run is None and workflow_run_id is not None:
+        from github_actions.github_actions_api import gha_query_workflow_run_by_id
         workflow_run = gha_query_workflow_run_by_id(github_repository, workflow_run_id)
```
ScottTodd: Style guide typically requires imports at the top. Add a comment here for why that isn't applicable here, and maybe a warning at the top of the file about adding new imports if this file is going to get copied to AWS somewhere.

I'm expecting that we'll add at least a common logging module that we import across our scripts, so expecting this script to be safe to copy/paste standalone into another project/location is risky.

build_tools/tests/generate_s3_index_test.py
```diff
@@ -0,0 +1,244 @@
...
+
+            entries = generate_s3_index._list_files_local(
+                staging_dir, "12345-linux/logs/gfx94X-dcgpu"
+            )
+            names = [e.name for e in entries]
+            self.assertIn("build.log", names)
+            self.assertIn("ninja_logs.tar.gz", names)
+            # Subdirectory files are NOT included
+            self.assertNotIn("therock-build-prof/comp-summary.html", names)
+            self.assertNotIn("comp-summary.html", names)
```
ScottTodd: I think I do want subdirectory files for multi-arch. We can iterate on that.

build_tools/tests/generate_s3_index_test.py
```diff
@@ -0,0 +1,244 @@
...
+            root.mkdir(parents=True)
+            (root / "index.html").write_text("<html></html>")
+            (root / "build.log").write_text("log")
+
+            entries = generate_s3_index._list_files_local(
+                staging_dir, "12345-linux/logs"
+            )
+            names = [e.name for e in entries]
+            self.assertNotIn("index.html", names)
+            self.assertIn("build.log", names)
```
ScottTodd: nit / personal preference:

I've been merging similar test cases like these where applicable. The boilerplate setup is mostly the same between `test_lists_immediate_files_only` and `test_excludes_index_html`, just the list of asserts at the end differ.

build_tools/generate_s3_index.py
```diff
@@ -0,0 +1,348 @@
...
+    return "\n".join(lines)
+
+
+def _escape_html(text: str) -> str:
+    return (
+        text.replace("&", "&amp;")
+        .replace("<", "&lt;")
+        .replace(">", "&gt;")
+        .replace('"', "&quot;")
+    )
```
ScottTodd: Use `html.escape` instead of rolling our own: https://docs.python.org/3/library/html.html#html.escape. This implementation does not handle `'`

build_tools/generate_s3_index.py
```diff
@@ -0,0 +1,348 @@
...
+        if entry.last_modified:
+            mod_iso = entry.last_modified.isoformat()
+            mod_str = entry.last_modified.strftime("%Y-%m-%d %H:%M UTC")
+            mod_cell = f'<time datetime="{_escape_html(mod_iso)}">{mod_str}</time>'
+        else:
+            mod_cell = "&mdash;"
+        lines.append(
+            f'<tr><td><a href="{_escape_html(entry.href)}">{_escape_html(entry.name)}</a></td>'
+            f"<td>{size_str}</td><td>{mod_cell}</td></tr>"
+        )
```
ScottTodd, line 138: Use `urllib.parse.quote` instead of HTML escape for the `href` URL segment: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote

Need to encode characters like `+`, as on https://github.com/ROCm/TheRock/pull/4138

build_tools/generate_s3_index.py
```diff
@@ -0,0 +1,348 @@
...
+    """Write html to a temp file and upload it to dest."""
+    with tempfile.NamedTemporaryFile(
+        mode="w", suffix=".html", encoding="utf-8", delete=False
+    ) as tmp:
+        tmp.write(html)
+        tmp_path = Path(tmp.name)
+    try:
+        backend.upload_file(tmp_path, dest)
+    finally:
+        tmp_path.unlink(missing_ok=True)
```
ScottTodd: Either use the `dry_run` arg or delete it

(I'd suggest keeping it so this can be tested locally)

build_tools/tests/generate_s3_index_test.py
marbre: I fully addressed this during a refactoring.  I added subdirectory entries (`CommonPrefixes`) to the listings so `gfx94X/`, `gfx1100/`, etc. appear in the parent index as navigable links.

build_tools/tests/generate_s3_index_test.py
marbre: Done.

build_tools/generate_s3_index.py
marbre: Done.

build_tools/generate_s3_index.py
marbre: Done.

build_tools/generate_s3_index.py
marbre: I added the missing part to make dry-run working.

build_tools/tests/generate_s3_index_test.py
```diff
@@ -0,0 +1,259 @@
...
+
+# Add build_tools to path so _therock_utils and generate_s3_index are importable.
+sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
+
+from _therock_utils.storage_backend import LocalStorageBackend
+import generate_s3_index
+
+
+class TestListFilesLocal(unittest.TestCase):
+    """Tests for _list_files_local()."""
```
ScottTodd, line 23: Typically functions with a `_` prefix are private functions / implementation details. While they _can_ be unit tested, they aren't actually part of the public API.

Reading the library and its tests, I'd say we could:

* Make `_generate_index_html` public by dropping the prefix, also make `_FileEntry` public.
* Rename `generate_index_for_directory` to differentiate it and signal that it list files, generates an index, and uploads the index. Ideas:
  * `index_directory`
  * `publish_index_for_directory`
  * `create_directory_index`
  * `sync_directory_index`
  * `update_directory_index`
  * `process_directory`
* Consider dropping unit tests for the remaining private functions, test the behavior we want through the public functions instead

Or, we can keep the private classes/functions as they are, but at least update the similar names

build_tools/_therock_utils/workflow_outputs.py
marbre: I added a comment to elaborate on the reason for this.

build_tools/tests/generate_s3_index_test.py
marbre: Will address this in a follow up.

PR #4170, opened 2026-03-25
Update package.json with dependency and artifact changes

raramakr:
- Add amd-comgr-stub component to amdrocm-runtime package
- Add amdrocm-rand dependency to amdrocm-fft-test package
- Add fftw3 artifact to amdrocm-fft-test package
- Update amdrocm-dnn-test dependencies to include runtime components (amdrocm-runtime, amdrocm-blas, amdrocm-sysdeps)

### DIFF
--- a/build_tools/packaging/linux/package.json
+++ b/build_tools/packaging/linux/package.json
@@ -248,12 +248,6 @@
               "run",
               "doc"
             ]
-          },
-          {
-            "Name": "amd-comgr-stub",
-            "Components": [
-              "lib"
-            ]
           }
         ]
       }


build_tools/packaging/linux/package.json
```diff
@@ -248,6 +248,12 @@
               "run",
               "doc"
             ]
+          },
+          {
+            "Name": "amd-comgr-stub",
+            "Components": [
+              "lib"
+            ]
           }
```
ScottTodd: Drive-by: wasn't this `amd-comgr-stub` removed in https://github.com/ROCm/TheRock/pull/3731 ?

raramakr: Thanks Scott for catching this. Removed the same. 
Also confirmed the recent binaries doesn't have a dependency to amd_comgr_loader.so .
Following returns nothing
ldd hipfft-test | grep libamd_comgr_loader
ldd rocfft-test | grep libamd_comgr_loader

Also the artifacts doesn't have amd-comgr-stub

ScottTodd: You can thank my code review agent 😂. I happened to point it at this PR while I was testing my setup, since this PR was small and recent. I was surprised it found a real issue!

PR #4185, opened 2026-03-26
[ci] Allowing mi325 back to normal for CI

geomin12:
As queues are better after mi325 enablement and more nodes, we will enable mi325 back to normal to get signal


marbre (approved):

geomin12:
https://github.com/ROCm/TheRock/pull/4193 fixes the sanity check errors. The CI is enough proof that mi325 machines are back

PR #4187, opened 2026-03-26
[ci] Adding default for release workflow_dispatch triggers

geomin12:
Somewhat frequently, engineers will trigger a release pipeline with workflow dispatch with no arguments for families. When this occurs, it runs 10 * 16 jobs on both release and pytorch release workflows and cause delays on TheRock CI.

When chatting with the engineer, they usually only need 1-2 archs. 

With this default input, we will no longer run into this issue!

Works here:

Screenshot 2026-03-26 152543
Screenshot 2026-03-26 152557

### DIFF
--- a/.github/workflows/release_windows_packages.yml
+++ b/.github/workflows/release_windows_packages.yml
@@ -56,6 +56,7 @@ on:
       families:
         description: "A comma separated list of AMD GPU families, e.g. `gfx94X,gfx103x`, or empty for the default list"
         type: string
+        default: "gfx110X"
       prerelease_version:
         description: "(Optional) Number of the prerelease"
         type: string
--- a/.github/workflows/release_portable_linux_packages.yml
+++ b/.github/workflows/release_portable_linux_packages.yml
@@ -53,6 +53,7 @@ on:
       families:
         description: "Comma separated list of AMD GPU families, e.g. `gfx94X,gfx103x`"
         type: string
+        default: "gfx94X"
       prerelease_version:
         description: "(Optional) Number of the prerelease"
         type: string


.github/workflows/release_windows_packages.yml
```diff
@@ -56,6 +56,7 @@ on:
       families:
         description: "A comma separated list of AMD GPU families, e.g. `gfx94X,gfx103x`, or empty for the default list"
         type: string
+        default: "gfx110X"
```
ScottTodd, line 59: Could pick the same default between workflows (others use gfx1151):

* https://github.com/ROCm/TheRock/blob/59d986221af9310c8543e7476155c64dbe7971e1/.github/workflows/build_windows_pytorch_wheels.yml#L55-L72
* https://github.com/ROCm/TheRock/blob/59d986221af9310c8543e7476155c64dbe7971e1/.github/workflows/release_windows_pytorch_wheels.yml#L38-L57

amd-justchen: I was thinking normalizing to gfx1101 as well. We do have plentiful amount of azure runners!

amd-justchen (approved):
Making gfx1101 default seems to be good!

.github/workflows/release_windows_packages.yml
geomin12: ah, good option! although I know some folks want 2+ archs during dev releases, so I'll keep it this way to have that option open.

This is primarly so folks don't just trigger it blank and run a lotttt of archs

ScottTodd: My point is that the defaults should be the same between our workflows. The `type:choice` and `options` are tangential.

PR #4189, opened 2026-03-26
Compress ccache log files prior to uploading them in CI workflows

ScottTodd:
## Motivation

Our ccache.log files are pretty large (~800MB, or 25-30% of the entire storage used per workflow run). We don't yet have a retention policy set on logs or artifacts, so this is contributing quite a bit to our monthly storage billing costs.

<details><summary>Metrics: CI vs multi-arch CI size for ccache.log (Linux)</summary>
<p>

| Source | ccache.log size |
|--------|-----------------|
| Single-stage CI (all subprojects, gfx950-dcgpu) | 787 MB |
| Multi-arch: compiler-runtime | 467 MB |
| Multi-arch: math-libs/gfx1151 | 133 MB |
| Multi-arch: math-libs/gfx120X-all | 130 MB |
| Multi-arch: math-libs/gfx110X-all | 109 MB |
| Multi-arch: comm-libs/gfx1151 | 24 MB |
| Multi-arch: comm-libs/gfx110X-all | 23 MB |
| Multi-arch: foundation | 16 MB |
| Multi-arch total | ~900 MB+ |

</p>
</details> 

<details><summary>Metrics: CI artifact vs ccache.log sizes (Linux gfx950-dcgpu)</summary>
<p>

| Category | Size |
|----------|------|
| blas_lib | 396 MB |
| blas_test | 349 MB |
| prim_test | 355 MB |
| composable-kernel_dev | 160 MB |
| amd-llvm_dev | 158 MB |
| amd-llvm_lib | 113 MB |
| miopen_lib | 96 MB |
| rocdecode_test | 88 MB |
| amd-llvm_run | 50 MB |
| rocwmma_test | 51 MB |
| Total artifacts (estimated) | ~2 GB |
| ccache.log | 787 MB |
| ccache.log as % of artifacts | ~40% |

</p>
</details>

> [!NOTE]
> Unlike subproject build logs (e.g. `MIOpen_build.log` our next largest log file at 80MB-200MB), these files are less frequently browsed casually in a web browser, and I don't think we're actually actively _using_ them at all, but they could still be helpful. We can at least compress them prior to upload.

## Technical Details

This moves the files from the root `build/logs/` folder to a `build/logs/ccache/` folder then archives and compresses that folder as `ccache_logs.tar.gz`. Moving the files into a subfolder made filtering simple reliable. My agent originally tried to delete the original files after archival and prior to "upload", but I wanted to be able to run and re-run the upload script locally without behavior changing for testing.

## Test Plan and Results

* Measured compression savings
  * ~11:1 (91% reduction), e.g. 787MB --> 65.7MB
* Added new unit tests
* Ran `post_stage_upload.py` with a local output directory and inspected the output files:
  image
* Watch CI and Multi-arch CI workflow runs to check that files are archived, uploaded, and included on index pages as expected

## Submission Checklist


HereThereBeDragons:
could we use zstd for compression? it is faster for the same compression level as zlib and/or you can compress better for a similar speed as the default zlib

ScottTodd:
> could we use zstd for compression? it is faster for the same compression level as zlib and/or you can compress better for a similar speed as the default zlib

Good point. It took a bit of extra code, but my testing shows that switching from gzip compression (level 9) to zstd compression (default level of 3) cuts the compression time and improves the compression ratio:

For a 787MB input:

method | size | time | ratio
-- | -- | -- | --
gzip (level 9) | 65.8MB | 12.9s | 12.0x
zstd (level 3) | 59.2MB | 1.2s | 13.3x

The time saving alone is worth it for these workflows IMO

build_tools/github_actions/post_build_upload.py
```diff
@@ -214,8 +257,11 @@ def upload_logs(
     #   - index.html (generated by index_log_files)
     #   - therock-build-prof/ subdirectory (resource profiling)
     # Content-type is inferred per file by the backend.
+    # Exclude raw ccache logs — they're uploaded compressed as ccache_logs.tar.gz.
```
amd-shiraz: Nit: should this say `ccache_logs.tar.zst` in the comment as well ?

build_tools/tests/storage_backend_test.py
```diff
@@ -267,6 +267,40 @@ def test_dry_run_does_not_copy(self):
...
+    def test_exclude_filters_matching_files(self):
+
+        with tempfile.TemporaryDirectory() as staging, tempfile.TemporaryDirectory() as src:
+            staging_dir = Path(staging)
+            source_dir = Path(src) / "artifacts"
+            self._make_tree(source_dir)
+
+            dest = StorageLocation("bucket", "run-1")
+            backend = LocalStorageBackend(staging_dir)
+            count = backend.upload_directory(source_dir, dest, exclude=["sub/*"])
```
amd-shiraz, line 284: nit/optional: Can we add a test that excludes a deeper nested path as well? `rglob("sub/*")` will also match recursively under `sub/`, so having a nested case here would make the intended exclude semantics explicit.

build_tools/github_actions/post_stage_upload.py
```diff
@@ -93,12 +137,14 @@ def upload_stage_logs(
         return
 
     dest = output_root.stage_log_dir(stage_name, amdgpu_family)
-    backend.upload_directory(log_dir, dest)
+    # Exclude raw ccache logs — they're uploaded compressed as ccache_logs.tar.gz.
```
amd-shiraz: same comment as above. ccache_logs.tar.zst in comment as well.

amd-shiraz:
conflicting file. pls resolve + some nits and optional reviews. looks good to merge otherwise.

build_tools/github_actions/post_build_upload.py
```diff
@@ -108,6 +108,49 @@ def create_ninja_log_archive(build_dir: Path):
...
+    with pyzstd.ZstdFile(archive_path, mode="wb") as zst:
+        with tarfile.open(mode="w|", fileobj=zst) as tar:
+            for file_path in found_files:
+                tar.add(file_path, arcname=file_path.name)
+                log(f"[+] Archived ccache log: {file_path.name}")
+
+    archive_size = archive_path.stat().st_size
+    log(
+        f"[INFO] Created {archive_path.name} "
+        f"({archive_size // 1024 // 1024}MB from {len(found_files)} files)"
```
marbre, line 150:
```suggestion
        f"({archive_size // 1024 // 1024}MiB from {len(found_files)} files)"
```

NIT: MiB if dividing by 1024^2 😜

ScottTodd: lol I don't care enough to fix the logging :P

PR #4203, opened 2026-03-27
Add SHA256 conflict detection to artifact uploads

### LINKED ISSUE #4202: Generic artifacts can silently overwrite existing artifacts with different SHA in S3 uploads
## Summary

Generic (`_generic`) artifacts uploaded to S3 can silently overwrite existing artifacts without any verification, even when the new artifact has a different SHA256 checksum. This can mask build system problems where a subproject was incorrectly marked as `target-neutral` but actually produces target-specific binaries.

## Problem

The current upload logic in `artifact_manager.py` and `storage_backend.py` performs uploads without:

1. Checking if the S3 object already exists
2. Comparing checksums between existing and new artifacts
3. Warning or failing when different content would overwrite an existing artifact

### Affected Code Paths

**`storage_backend.py:265-279`** - `S3StorageBackend.upload_file()`:
```python
def upload_file(self, source: Path, dest: StorageLocation) -> None:
    # No existence check, no SHA comparison
    _s3_retry(
        "upload",
        dest.s3_uri,
        self.s3_client.upload_file,
        str(source),
        dest.bucket,
        dest.relative_path,
        ExtraArgs={"ContentType": content_type},
    )
```

**`artifact_backend.py:302-305`** - `S3Backend.upload_artifact()`:
```python
def upload_artifact(self, source_path: Path, artifact_key: str) -> None:
    loc = self.output_root.artifact(artifact_key)
    self.s3_client.upload_file(str(source_path), self.bucket, loc.relative_path)
```

### Risk Scenario

1. Stage A (generic) builds `foo_lib_generic.tar.zst` on Job 1 with SHA `abc123`
2. Stage A (generic) runs again on Job 2 in parallel (different runner) with different content (e.g., due to non-deterministic build or incorrectly-marked target-specific artifact), producing SHA `def456`
3. Both jobs call `artifact_manager.py push` which uploads to the same S3 path
4. The second upload silently overwrites the first - **no warning, no error**

This is particularly concerning because:
- If an artifact is incorrectly marked as `target-neutral` in `BUILD_TOPOLOGY.toml` but actually contains target-specific binaries, multiple parallel jobs could produce different content
- The "last writer wins" with no audit trail
- SHA256sum files *are* uploaded alongside artifacts, but there's no pre-upload check against existing checksums

## Proposed Solution

For generic artifacts, add SHA-based conflict detection:

1. Before uploading, check if the artifact already exists in S3
2. If it exists, download or fetch the existing `.sha256sum` file
3. Compare with the local artifact's SHA256
4. If they differ, **fail the upload with an error** - this surfaces incorrectly-marked artifacts
5. If they match, skip the upload (artifact is already there with correct content)

This ensures:
- Genuinely generic/deterministic artifacts can be safely uploaded from any parallel job
- Non-deterministic or incorrectly-marked artifacts are detected and flagged
- No silent data corruption

## Additional Context

The artifact type (`target-neutral` vs `target-specific`) is defined in `BUILD_TOPOLOGY.toml` and propagated to CMake via `topology_to_cmake.py`. The CMake logic in `therock_artifacts.cmake` uses the `TARGET_NEUTRAL` flag to determine whether to use `_generic` suffix.

Target-specific artifacts are split by `split_artifacts.py` into per-architecture artifacts, so they don't have this issue (each arch gets its own S3 key).

PeterCDMcLean:
## Summary
- Add `ArtifactConflictError` exception to detect when uploading an artifact that already exists with different content
- Add `sha256sum_path` parameter to `upload_artifact()` for efficient conflict detection without recomputing hashes
- `upload_artifact()` now uploads both artifact and sha256sum file in one call
- When remote sha256sum is missing, fall back to downloading and computing hash from file contents
- Use existing `hash_util.calculate_hash` instead of custom implementation

This prevents silent overwrites that could mask build system problems where a subproject was incorrectly marked as target-neutral or built inconsistently across different GPU families.

Fixes #4202

## Test plan
- [x] Added comprehensive unit tests for conflict detection in both `LocalDirectoryBackend` and `S3Backend`
- [x] Tests cover: matching SHA skips upload, differing SHA raises error, missing sha256sum falls back to file comparison
- [x] All existing tests pass
- [x] Pre-commit checks pass

### DIFF
--- a/.github/workflows/build_portable_linux_artifacts.yml
+++ b/.github/workflows/build_portable_linux_artifacts.yml
@@ -122,6 +122,7 @@ jobs:
           package_version: ${{ inputs.package_version }}
           extra_cmake_options: ${{ inputs.extra_cmake_options }}
           BUILD_DIR: build
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           python3 build_tools/github_actions/build_configure.py --manylinux
 
--- a/.github/workflows/build_windows_artifacts.yml
+++ b/.github/workflows/build_windows_artifacts.yml
@@ -135,6 +135,7 @@ jobs:
           amdgpu_families: ${{ inputs.amdgpu_families }}
           package_version: ${{ inputs.package_version }}
           extra_cmake_options: ${{ inputs.extra_cmake_options }}
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           python3 build_tools/github_actions/build_configure.py
 
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -118,6 +118,8 @@ jobs:
         run: pip install ${{ steps.stage_config.outputs.pip_install_cmd }}
 
       - name: Configure
+        env:
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           cmake -B "${BUILD_DIR}" -S . -GNinja \
             -DTHEROCK_PACKAGE_VERSION=${{ inputs.rocm_package_version }} \
--- a/.github/workflows/multi_arch_build_windows_artifacts.yml
+++ b/.github/workflows/multi_arch_build_windows_artifacts.yml
@@ -143,6 +143,8 @@ jobs:
         run: pip install ${{ steps.stage_config.outputs.pip_install_cmd }}
 
       - name: Configure
+        env:
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           cmake -B "${BUILD_DIR}" -S . -GNinja \
             --preset=${{ inputs.build_variant_cmake_preset }} \
--- a/.github/workflows/release_portable_linux_packages.yml
+++ b/.github/workflows/release_portable_linux_packages.yml
@@ -235,6 +235,7 @@ jobs:
           package_version: ${{ needs.setup_metadata.outputs.version }}
           extra_cmake_options: ''
           cmake_preset: ''
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           python3 ./build_tools/github_actions/build_configure.py \
             --build-dir ${{ env.BUILD_DIR }} \
--- a/.github/workflows/release_windows_packages.yml
+++ b/.github/workflows/release_windows_packages.yml
@@ -236,6 +236,7 @@ jobs:
           amdgpu_families: ${{ matrix.target_bundle.amdgpu_family }}
           package_version: "ADHOCBUILD"
           extra_cmake_options: ${{ inputs.extra_cmake_options }}
+          ROCM_BUILD_ID: ${{ github.sha }}
         run: |
           python3 build_tools/github_actions/build_configure.py
 


.github/workflows/build_portable_linux_artifacts.yml
```diff
@@ -122,6 +122,7 @@ jobs:
           package_version: ${{ inputs.package_version }}
           extra_cmake_options: ${{ inputs.extra_cmake_options }}
           BUILD_DIR: build
+          ROCM_BUILD_ID: ${{ github.sha }}
```
ScottTodd, line 125: > I would like to be able to drop:
> 
> [948e06e](https://github.com/ROCm/TheRock/pull/4203/commits/948e06e8ef8b84697f814b674676dd529dce7bbf) and [dc33592](https://github.com/ROCm/TheRock/pull/4203/commits/dc335929e6e63f508f524ce75f069b7ecdc8f657)
> 
> from this PR, but I would need to make a simultaneous change to amd-dbgapi to remove the timestamp. Also, I am not familiar with ROC_BUILD_ID it looks like it was used in the past and was deprecated, but some components (amd-dbgapi) still expect it. I would appreciate some guidance how it should be addressed.

Yes, please remove `ROCM_BUILD_ID` changes from this PR. PRs should be as small and focused as possible, and this is a behavior change that I don't believe should be necessary. We can review separately.

* We already have different mechanisms for setting a version or build id, this one seems redundant
* If we set it, a value like a workflow run id may be more appropriate than the git commit sha
* If any projects are depending on this for reproducible/hermetic builds, we may want to change the source for those projects rather do this plumbing

My code review agent agrees that this should be out of this PR's scope and had more suggestions: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4203.md#%EF%B8%8F-important-rocm_build_id-embeds-into-binaries-changing-artifact-content

PeterCDMcLean:
I would like to be able to drop:

[948e06e](https://github.com/ROCm/TheRock/pull/4203/commits/948e06e8ef8b84697f814b674676dd529dce7bbf) 
and
[dc33592](https://github.com/ROCm/TheRock/pull/4203/commits/dc335929e6e63f508f524ce75f069b7ecdc8f657)

from this PR, but I would need to make a simultaneous change to amd-dbgapi to remove the timestamp. Also, I am not familiar with ROC_BUILD_ID it looks like it was used in the past and was deprecated, but some components (amd-dbgapi) still expect it. I would appreciate some guidance how it should be addressed.

build_tools/_therock_utils/storage_backend.py
```diff
@@ -276,6 +278,53 @@ def upload_file(self, source: Path, dest: StorageLocation) -> None:
...
+                tmp_path = Path(tmp.name)
+            try:
+                _s3_retry(
+                    "download",
+                    dest.s3_uri,
+                    self.s3_client.download_file,
+                    dest.bucket,
+                    dest.relative_path,
+                    str(tmp_path),
+                )
```
ScottTodd, line 313: Let's try to keep this PR focused on the artifact uploads, which use artifact_backend.py and not storage_backend.py (today).
* There are clear cases where multiple build stages upload the same artifact files but the scenarios for logs and build reports are different.
* The code path here that downloads just to compare is inefficient.
* I think we could also flip some settings in the upload command to have AWS do this checking for us, so we don't need to do our own comparisons: https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-writes.html

Review agent had some other thoughts: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4203.md#%EF%B8%8F-important-conflict-detection-at-two-independent-layers-is-confusing-and-expensive

build_tools/fileset_tool.py
```diff
@@ -78,6 +78,22 @@ def do_artifact(args):
...
+    File permissions are preserved. When extracted as a non-root user, ownership
+    defaults to the extracting user. When extracted as root with -p, files are
+    owned by root:root.
+    """
+    tarinfo.mtime = 0
+    tarinfo.uid = 0
+    tarinfo.gid = 0
+    tarinfo.uname = "root"
+    tarinfo.gname = "root"
+    return tarinfo
```
ScottTodd, line 94: Can you sequence this PR as:
1. Changes necessary for there to be no conflicts
2. Tests that enforce no conflicts

?

build_tools/tests/artifact_backend_test.py
```diff
@@ -678,5 +690,417 @@ def test_s3_backend_when_no_local_dir(self, mock_retrieve):
...
+        with self.assertRaises(ArtifactConflictError) as ctx:
+            self.backend.upload_artifact(source_file, artifact_key)
+
+        error_msg = str(ctx.exception)
+        self.assertIn("existing_sha_123", error_msg)
+        self.assertIn("different_sha_456", error_msg)
+
+    def test_backwards_compat_alias(self):
+        """Test that ArtifactConflictError is an alias for ArtifactConflictError."""
+        self.assertIs(ArtifactConflictError, ArtifactConflictError)
```
ScottTodd, line 911: This test does nothing :P

(Claude Code really likes "backwards compatible" aliases... within code that has no external uses, need to watch for that)

build_tools/tests/artifact_backend_test.py
```diff
@@ -678,5 +690,417 @@ def test_s3_backend_when_no_local_dir(self, mock_retrieve):
...
+            self.assertEqual(result, "deadbeef12345678")
+
+    def test_returns_none_if_no_sha256sum(self):
+        """Test that None is returned if sha256sum file doesn't exist."""
+        with tempfile.TemporaryDirectory() as temp_dir:
+            artifact_path = Path(temp_dir) / "test.tar.zst"
+            artifact_path.touch()
+
+            result = _read_local_sha256sum(artifact_path)
+            self.assertIsNone(result)
```
ScottTodd, line 749: These tests don't seem useful, or they could be merged with `TestParseSha256sumContent`. `_read_local_sha256sum` is a private function with a simple implementation:
1. read a file
2. call `_parse_sha256sum_content()`

While helper functions are smaller units to test, the public interface is what matters.

build_tools/artifact_manager.py
ScottTodd, line 562: This shouldn't retry on `ArtifactConflictError` (or any error known to be non-transient)

https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4203.md#1-artifact_managerpy--artifactconflicterror-is-retried

build_tools/tests/artifact_backend_test.py
```diff
@@ -678,5 +690,417 @@ def test_s3_backend_when_no_local_dir(self, mock_retrieve):
...
+    def test_artifact_matching_sha_skips_upload(self, mock_client_prop):
+        """Test that artifact with matching SHA skips upload."""
+        mock_client = mock.MagicMock()
+        mock_client_prop.return_value = mock_client
+
+        # Mock artifact exists
+        mock_client.head_object.return_value = {}  # exists
+
+        # Mock sha256sum download
+        import io
```
ScottTodd, line 937: Move imports to the top, unless there is a technical reason why importing inline is necessary

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

PR #4207, opened 2026-03-27
Adding TheROCK repository access to bot token (Submodule Automation)

ammallya:
## Motivation

Adding TheROCK repository access to bot token

### DIFF
--- a/.github/workflows/bump_submodules.yml
+++ b/.github/workflows/bump_submodules.yml
@@ -4,6 +4,7 @@
 name: Bump Submodule automation
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: "0 */12 * * *" # Runs every 12 hrs at 12AM and 12PM
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_submodules.yml
```diff
@@ -34,7 +35,7 @@ jobs:
           app-id: ${{ secrets.ROCM_SYSTEMS_APP_ID }}
           private-key: ${{ secrets.ROCM_SYSTEMS_APP_PRIVATE_KEY }}
           owner: "ROCm"
-          repositories: "rocm-systems"
+          repositories: "rocm-systems,TheRock"
```
ScottTodd, line 38: Why are we using tokens for rocm-systems and rocm-libraries at all? Public access to those should be fine, isn't this workflow only creating PRs in TheRock?

Ah, I see that it creates PRs like https://github.com/ROCm/rocm-libraries/pull/5887 too. Please add comments to this workflow and/or https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/bump_automation.py explaining the intent/usage.

.github/workflows/bump_submodules.yml
ScottTodd, line 1: FYI bump_automation.py could use https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py instead of rolling its own implementations of API calls

PR #4210, opened 2026-03-27
[ci] benchmark testing back due to increased mi325 capacity

geomin12:
As more mi325 capacity is back, we are re-enabling benchmark nightly testing


amd-justchen (approved):
Glad we got the runners back!

Issue #4211, opened 2026-03-27
[sysdeps] Link shared libraries with `-Bsymbolic` on Linux


stellaraccident:
Reported by a user (on a recent Debian):

```
#5  0x00007fffa7ce592d in ZSTD_freeDCtx () from /lib/x86_64-linux-gnu/libzstd.so.1
#6  0x00007fffa60e3e78 in ZSTD_decompress () from /opt/rocm/lib/rocm_sysdeps/lib/librocm_sysdeps_zstd.so.1
#7  0x00007ffd55144103 in llvm::compression::zstd::decompress(llvm::ArrayRef<unsigned char>, unsigned char*, unsigned long&) () from /opt/rocm/lib/llvm/lib/libLLVM.so.22.0git
```

System zstd is typically shipped without symbol versioning. Therefore, if zstd loads in the namespace first, it can interpose on internal symbols within the (versioned) bundled ROCm zstd. This can result in crashes/corruption/etc, depending on whichsystem zstd is installed.

Adding `-Bsymbolic` for shared linking on Linux solves this issue. Verified with a customer stack trace that the issue was due to internal symbol interposition. Symbol versioning between ROCm libraries is taking care of entry-points properly.

We should fix the reported issue and hold this open so that all libraries can be audited and adjusted. We should also extend our sysdep validation step to verify that they were built symbolic.

Note that not all sysdeps are susceptible to this. Many of them always build with symbol versioning and are therefore more resistant to internal interposition issues. However, there is little reason for us to not be conservative and tag all of our bundled libraries for symbolic linking.

sjoerdsimons:
As a related note; I've seen the rocm zstd symbol is using the @@ version, e.g. `ZSTD_decompressDCtx@@AMDROCM_SYSDEPS_1.0`  -- Which if i'm not mistaken means it's a valid target for both lookups with that version and unversioned lookups. As Debian ships an unversioned zstd by default, this may mean this library might get used rather then the system library. Potentially triggering the same symbol confusion (or abi clashes).

Note that is purely theoretical, but maybe worth considering. 

Fwiw a potentially nicer solution might be actually use different symbol names (e.g. add rocm_ prefixes)  in the library (keeping api compatibility via e.g.  header file fun).. That would avoid all potential symbol mixups without relying on versioning etc.

stellaraccident:
[...]

PR #4212, opened 2026-03-27
[sysdeps] Link zstd with -Bsymbolic on Linux.

stellaraccident:
System zstd is typically shipped without symbol versioning. Therefore, if zstd loads in the namespace first, it can interpose on internal symbols within the (versioned) bundled ROCm zstd. This can result in crashes/corruption/etc, depending on which system zstd is installed.

Adding `-Bsymbolic` for shared linking on Linux solves this issue. Verified with a customer stack trace that the issue was due to internal symbol interposition. Symbol versioning between ROCm libraries is taking care of entry-points properly.

This point fix should solve the reported issue, but we should add more testing and do an audit of all of the libraries and adjust as needed.

Progress on #4211


stellaraccident:
Change validated modulo flaky runners. Landing because this fixes an issue in the field and we will post-review on Monday.

third-party/sysdeps/common/zstd/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set(CONFIGURE_ARGS
     -DZSTD_BUILD_STATIC=OFF
     -DZSTD_BUILD_SHARED=ON
-    "-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds"
+    "-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds -Bsymbolic"
```
ScottTodd, line 68: I'm out of my depth here but my code review agent is skeptical that this flag gets plumbed through successfully.

Should this be `-Wl,-Bsymbolic` instead? That's what most other uses on github that I can see do: https://github.com/search?q=CMAKE_SHARED_LINKER_FLAGS+Bsymbolic&type=code

I pulled down `librocm_sysdeps_zstd.so.1.5.7` from https://therock-ci-artifacts.s3.amazonaws.com/23674255041-linux/index.html on this PR and `readelf -d librocm_sysdeps_zstd.so.1.5.7 | grep -i symbolic` found no results.

more notes from code review agent: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4212.md#1--bsymbolic-flag-placement

stellaraccident: Thanks. I'll fix it properly today.

PR #4228, opened 2026-03-28
[AMD-SMI] Use detect_asic_filter.sh for ASIC-specific test exclusions

marifamd:
## Motivation

ASIC-specific `amdsmitst` exclusions were hardcoded in this script as a per-arch `TESTS_TO_IGNORE` dict. That list duplicates the canonical `amdsmitst.exclude` + `detect_asic_filter.sh` that already ship in the amd_smi test package, so the two drift out of sync. Source the package's filter instead and keep the exclusions in one place owned by amd-smi.

## Technical Details

`build_tools/github_actions/test_executable_scripts/test_amdsmi.py`:
- Add `get_asic_exclude_filter()` to source `amdsmitst.exclude` and `detect_asic_filter.sh` from the test dir, returning the `GTEST_EXCLUDE` string (warns and falls back to no filter when either script is missing or detection fails)
- Full mode: replace the hardcoded `TESTS_TO_IGNORE` per-arch dict (and the `is_windows` / `platform_key` / `AMDGPU_FAMILIES` plumbing) with the negative-only filter sourced from the package, matching the upstream amd-smi CI pattern (`./amdsmitst --gtest_filter="-${GTEST_EXCLUDE}"`)
- Quick mode: convert the include filter to a list for extensibility
- Improve variable naming and logging for clarity

JIRA ID : ROCM-9123

## Test Plan

- Confirm CI passes in full mode across different ASICs; quick-mode logic is unchanged

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
+++ b/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
@@ -48,6 +48,10 @@ def get_asic_exclude_filter(test_dir):
         logging.warning(f"detect_asic_filter.sh not found in {test_dir}")
         return ""
 
+    # check=True: a detection failure is a hard error (raises
+    # CalledProcessError and ends the run) rather than silently degrading to an
+    # unfiltered test set. capture_output is kept because we still need the
+    # GTEST_EXCLUDE value echoed on stdout.
     result = subprocess.run(
         [
             "bash",
@@ -57,13 +61,8 @@ def get_asic_exclude_filter(test_dir):
         capture_output=True,
         text=True,
         cwd=str(test_dir),
+        check=True,
     )
-
-    if result.returncode != 0:
-        logging.warning(
-            f"ASIC detection failed (rc={result.returncode}): {result.stderr.strip()}"
-        )
-        return ""
 
     gtest_exclude = result.stdout.strip()
     if gtest_exclude:
@@ -81,6 +80,17 @@ def _cgroup_mentions_container():
     return any(marker in cgroup for marker in ("docker", "kubepods", "containerd"))
 
 
+def _in_container():
+    # /.dockerenv (docker) and /run/.containerenv (podman) are the standard
+    # marker files; the `container` env var is set by systemd-nspawn and several
+    # OCI runtimes. Fall back to cgroup inspection for runtimes that set none.
+    if Path("/.dockerenv").exists() or Path("/run/.containerenv").exists():
+        return True
+    if os.environ.get("container"):
+        return True
+    return _cgroup_mentions_container()
+
+
 def detect_privilege_tier():
     """Classify the runtime as 'baremetal', 'privileged', or 'unprivileged'.
 
@@ -90,8 +100,7 @@ def detect_privilege_tier():
     alone cannot tell the two apart. Inspect the effective capability mask to
     distinguish them so the caller can set AMDSMI_NON_PRIVILEGED accordingly.
     """
-    in_container = Path("/.dockerenv").exists() or _cgroup_mentions_container()
-    if not in_container:
+    if not _in_container():
         return "baremetal"
 
     # CapEff is a 64-bit hex mask in /proc/self/status. A privileged container
--- a/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
+++ b/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
@@ -103,7 +103,8 @@ def detect_privilege_tier():
         for line in Path("/proc/self/status").read_text().splitlines():
             if line.startswith("CapEff:"):
                 cap_eff = int(line.split()[1], 16)
-                return "privileged" if cap_eff & (1 << cap_sys_admin) else "unprivileged"
+                has_admin = cap_eff & (1 << cap_sys_admin)
+                return "privileged" if has_admin else "unprivileged"
     except (OSError, ValueError):
         pass
     # Unknown capability state inside a container: treat as unprivileged so we
--- a/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
+++ b/build_tools/github_actions/test_executable_scripts/test_amdsmi.py
@@ -4,14 +4,16 @@
 
 """
 ===============================================================================
-AMDSMI Test Runner (Manual Execution Only)
+AMDSMI Test Runner
 
-This script is NOT part of automated CI runs.
+Runs the `amdsmitst` GTest binary shipped with the amd-smi component.
 
-`amdsmitst` requires GPU device access (/dev/kfd, /dev/dri), elevated
-permissions, and execution on a ROCm-enabled system. GitHub-hosted CI
-environments do not expose these capabilities, so this script must be run
-manually by developers inside a privileged ROCm environment or container.
+ASIC-specific exclusions are delegated to the amd-smi test tree itself
+(amdsmitst.exclude + detect_asic_filter.sh), so this runner stays agnostic to
+which tests each ASIC skips. Environment tiering (privileged docker /
+unprivileged docker / baremetal) is detected here and surfaced to the binary
+via AMDSMI_NON_PRIVILEGED so state-modifying tests skip themselves when the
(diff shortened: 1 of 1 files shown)


marifamd:
Dependent on landing rocm-systems patch first: https://github.com/ROCm/rocm-systems/pull/4523

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
```diff
@@ -27,16 +27,52 @@
...
 
-AMDSMITST_BIN = (
-    THEROCK_DIR / "build" / "share" / "amd_smi" / "tests" / "amdsmitst"
-).resolve()
+TESTS_DIR = (THEROCK_DIR / "build" / "share" / "amd_smi" / "tests").resolve()
+AMDSMITST_BIN = TESTS_DIR / "amdsmitst"
+
+
+def get_asic_exclude_filter(test_dir):
+    """Source amdsmitst.exclude and detect_asic_filter.sh, return GTEST_EXCLUDE."""
```
geomin12, line 40: better comment pls, to someone outside of amdsmi, what is asic?

tests/test_rocm_sanity.py
geomin12, line 1: i see this is duplicated from test_amdsmi.py.  any re-usability opportunities?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -123,6 +123,16 @@ def _get_script_path(script_name: str) -> str:
             "windows": 1,
         },
     },
+    "amdsmi": {
```
geomin12, line 239: are we planning to run this test in `test_rocm_sanity.py` or in `test_amdsmi.py` in this phase? I know @araravik-psd has some to clean this up

marifamd: test_amdsmi.py only, I'm still using the ci workflow files to debug failures, I'll move the ticket to draft

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
marifamd: gfx/asic same thing from our terminology, we can rename it if you want

tests/test_rocm_sanity.py
marifamd: don't think it should be the same, test_amdsmi should just be for amdsmi specifically.

build_tools/github_actions/fetch_test_configurations.py
marifamd: currently test_amdsmi.py is the one being used for TheRock CI so that's what I'm targeting.

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
```diff
@@ -21,27 +23,94 @@
...
+    detect_script = test_dir / "detect_asic_filter.sh"
+
+    if not exclude_script.exists():
+        logging.warning(f"amdsmitst.exclude not found in {test_dir}")
+        return ""
+    if not detect_script.exists():
+        logging.warning(f"detect_asic_filter.sh not found in {test_dir}")
+        return ""
+
+    result = subprocess.run(
```
geomin12, line 55: instead of run, can we use `check_call`? so if error, we end the script?

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
```diff
@@ -21,27 +23,94 @@
...
+    # (or --cap-add=ALL) exposes the full mask; unprivileged containers get the
+    # trimmed default docker set. CAP_SYS_ADMIN (bit 21) is the practical marker
+    # for "can touch the sysfs/debugfs knobs the write tests need".
+    cap_sys_admin = 21
+    try:
+        for line in Path("/proc/self/status").read_text().splitlines():
+            if line.startswith("CapEff:"):
+                cap_eff = int(line.split()[1], 16)
+                has_admin = cap_eff & (1 << cap_sys_admin)
+                return "privileged" if has_admin else "unprivileged"
```
geomin12: FYI, we run inside a docker container for these tests. will this still work?

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
```diff
@@ -52,77 +121,85 @@ def is_windows():
...
+        "*Unit*",
     ]
 
+    # Manual exclusions — always applied regardless of ASIC. Listed under both
+    # the legacy and renamed suite names so they survive the rename.
     exclude_tests = [
         "amdsmitstReadOnly.TempRead",
         "amdsmitstReadOnly.TestFrequenciesRead",
         "amdsmitstReadWrite.TestPowerReadWrite",
+        "GpuFunctionalReadOnly.TempRead",
```
araravik-psd, line 187: Are these new GpuFunctional tests added newly?

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
```diff
@@ -21,27 +23,94 @@
...
+def detect_privilege_tier():
+    """Classify the runtime as 'baremetal', 'privileged', or 'unprivileged'.
+
+    amdsmitst gates state-modifying tests on is_sudo_user() (uid == 0). Inside an
+    unprivileged docker container the process is uid 0 yet lacks the full
+    capability set a privileged container or baremetal root has, so is_sudo_user()
+    alone cannot tell the two apart. Inspect the effective capability mask to
+    distinguish them so the caller can set AMDSMI_NON_PRIVILEGED accordingly.
+    """
+    in_container = Path("/.dockerenv").exists() or _cgroup_mentions_container()
```
araravik-psd: Add a couple more container signals before falling back to /proc/1/cgroup, e.g. the container env var (set by systemd-nspawn and some runtimes) and /run/.containerenv:
  
Example:

def _in_container():
      if Path("/.dockerenv").exists() or Path("/run/.containerenv").exists():
          return True
      if os.environ.get("container"):
          return True
      return _cgroup_mentions_container()

araravik-psd (commented):
Can we also add test runs for gfx90a, gfx110X-all, gfx103X-all. I see test runs for only gfx94X here:

[Multi-Arch CI · ROCm/TheRock@04e0ffa](https://github.com/ROCm/TheRock/actions/runs/29448358549)
[Multi-Arch CI · ROCm/TheRock@45b7da8](https://github.com/ROCm/TheRock/actions/runs/29448362081)

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
marifamd: They are going to be added in https://github.com/ROCm/rocm-systems/pull/7650 
So the idea is to add them here first and then merge 7650 so there isn't any gap of CI failures

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
marifamd: Done in 25dee50

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
marifamd: Done in 25dee50

build_tools/github_actions/test_executable_scripts/test_amdsmi.py
marifamd: Yes it would still work

1. Container detected 
	- `_in_container()` matches (the test job sets `/.dockerenv` / cgroup markers).

2. Capabilities checked
	- the amdsmi job runs with `--user 0:0` and the default docker cap set (no `--privileged`, no `--cap-add`), so `CapEff` has no `CAP_SYS_ADMIN`.

3. Classified unprivileged
	- because it's a container without `CAP_SYS_ADMIN`, so we set `AMDSMI_NON_PRIVILEGED=1`.

geomin12 (approved):
let's eventually sync with @dileepr1  so this goes in the test standarization effort!

PR #4235, opened 2026-03-30
Add THEROCK_DEV_PROJECTS option for source change detection in large subprojects

aobolensk:
## Motivation

Large subprojects like amd-llvm skip source file globbing because the monorepo is too large for it to be practical by default. This means source changes go undetected by the outer build system, requiring manual rebuilds.

## Technical Details

Add a `THEROCK_DEV_PROJECTS` CMake option that lets developers opt in to source file globbing for specific subprojects. When `amd-llvm` is listed, `therock_cmake_subproject_glob_c_sources` is called conditionally, reusing the existing glob mechanism rather than introducing a separate code path.

## Test Plan

Development-only option - verify on a local build environment.

## Test Result

With `-DTHEROCK_DEV_PROJECTS=amd-llvm`, modifying files inside the LLVM tree triggers a rebuild. Without it, behavior is unchanged.

## Submission Checklist


AlexeySachkov:
It is unfortunate that extra flags are needed to make this work, which makes me wonder: why can't we unconditionally run the compiler sub-build? If there was no changes since last build, then it would be no-op, meaning minimal impact on those who doesn't modify compiler sources. I suppose there are some test/install target which wouldn't behave as nicely and therefore prevent us from running the compiler sub-build unconditionally?

aobolensk:
> It is unfortunate that extra flags are needed to make this work, which makes me wonder: why can't we unconditionally run the compiler sub-build?

They mentioned in the comments that it is an intentional behavior, because subprojects are huge, there are a lot of files and llvm is huge by itself. To be honest, I'd like to re-index it automatically, but it seems it is not the case here.

That is actually not really no-op, it looks up the file updates, but in my understanding it is not a heavy operation. I'd like to hear the opinion from the maintainers

compiler/CMakeLists.txt
```diff
@@ -133,8 +133,8 @@ if(THEROCK_ENABLE_COMPILER)
       "lib/llvm/lib"
   )
   # Note that we do not trigger on LLVM source changes because the monorepo is
-  # too large to glob like that. Consider having a project dev mode option for
-  # enabling better ergonomics here.
+  # too large to glob like that. Use THEROCK_DEV_PROJECTS=amd-llvm to always
+  # invoke the inner build during local development.
```
ScottTodd: I had to research a bit here to see that LLVM is deliberately omitting a call to `therock_cmake_subproject_glob_c_sources`. These comments would be more clear if they included that text.

An alternate approach to the sentinel file could be including `therock_cmake_subproject_glob_c_sources` here but making it conditional, either on a generic dev mode option (`THEROCK_DEV_MODE=ON` that applies to all subprojects or on a more precise option like `THEROCK_DEV_PROJECTS=amd-llvm`).

What is the actual overhead like for the glob (on Linux and Windows)? I'd prefer to reuse the existing glob mechanism but make it opt-in for LLVM instead of introducing a separate code path.

cmake/therock_subproject.cmake
```diff
@@ -1024,19 +1024,42 @@ function(therock_cmake_subproject_activate target_name)
...
+    # THEROCK_DEV_PROJECTS: always invoke the inner build for listed projects.
+    # The build stamp depends on a sentinel file that is touched *after* the
+    # stamp, keeping it perpetually newer and forcing a rebuild every time.
+    set(_is_dev_project FALSE)
+    if(THEROCK_DEV_PROJECTS)
+      list(FIND THEROCK_DEV_PROJECTS "${target_name}" _dev_idx)
+      if(NOT _dev_idx EQUAL -1)
+        set(_is_dev_project TRUE)
+      endif()
+    endif()
```
ScottTodd: nit: CMake has `IN_LIST` for this: https://cmake.org/cmake/help/latest/command/if.html#in-list

```suggestion
    set(_is_dev_project FALSE)
    if(THEROCK_DEV_PROJECTS AND target_name IN_LIST THEROCK_DEV_PROJECTS)
      set(_is_dev_project TRUE)
    endif()
```

aobolensk: Done

compiler/CMakeLists.txt
aobolensk: Added conditional `therock_cmake_subproject_glob_c_sources` call insted

ScottTodd: Thanks, this approach looks easier to maintain (so long as it has the behavior you want).

Issue #3336, opened 2026-02-09
[Multi-arch] Extend and improve multi-arch CI


marbre:
Parent issue tracking shipping ROCm multi-architecture packaging with Kpack based on this RFC: https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0008-Multi-Arch-Packaging.md.

## Overview

To fully support multi-arch the CI needs to partly be refactored and extended. This included to build artifacts, run all tests, uploads logs etc and to enable the CI on pre-submit and to fully transition to the multi-arch CI.

This issue tracks implementing the work for CI improvements required to fully support multi-arch.

Major milestones:

* Enable multi-arch CI on post-submit
* Enable multi-arch CI on pre-submit
* Enable `THEROCK_KPACK_SPLIT_ARTIFACTS` by default for developers
* Remove single stage (non-multi-arch) CI

## Implementation plan and timelines

TBD

stellaraccident:
## Status update — run #22085898944

**Build:** All 14 stages pass (required 2 retries to get past #3447 build race).

**Tests (gfx94X-dcgpu):** 10/29 passed, 19 failed. Same `hipErrorInvalidImage` pattern as before — the fetch fix (PR #3452 for #3444) did not resolve the test failures. See detailed analysis on #3444.

**Separate finding:** Many test suites mask kernel launch failures with misleading numeric errors (using `EXPECT_*` instead of `ASSERT_*`). Filed ROCm/rocm-libraries#4633 to track this testing debt.

ScottTodd:
I think we can close this parent issue now. Future work is tracked in sub-issues.

PR #4242, opened 2026-03-30
Always configure creds prior to log upload for multi-arch CI

ScottTodd:
## Motivation

Related to https://github.com/ROCm/TheRock/issues/3336

On https://github.com/ROCm/TheRock/actions/runs/23663263447/job/68945135064?pr=4209, log upload failed since the configure credentials step did not run:

```
WARNING:_therock_utils.storage_backend:S3 upload attempt 1/3 failed for s3://therock-ci-artifacts/23663263447-windows/logs/math-libs/gfx110X-all/hipBLAS-common_install.log: Failed to upload B:\build\logs\hipBLAS-common_install.log to therock-ci-artifacts/23663263447-windows/logs/math-libs/gfx110X-all/hipBLAS-common_install.log: An error occurred (AccessDenied) when calling the PutObject operation: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: s3:PutObject on resource: "arn:aws:s3:::therock-ci-artifacts/23663263447-windows/logs/math-libs/gfx110X-all/hipBLAS-common_install.log" because no identity-based policy allows the s3:PutObject action. Retrying in 1.0s...

RuntimeError: Failed to upload 88/88 files. First failure: s3://therock-ci-artifacts/23663263447-windows/logs/math-libs/gfx110X-all/hipBLAS-common_install.log: S3 upload failed after 3 attempts: s3://therock-ci-artifacts/23663263447-windows/logs/math-libs/gfx110X-all/hipBLAS-common_install.log
```

This fixes that by copying the pattern from CI to multi-arch CI: https://github.com/ROCm/TheRock/blob/4a115c720d51397c61a0efd7cc370fb78d862e80/.github/workflows/build_portable_linux_artifacts.yml#L170-L185

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -141,12 +141,17 @@ jobs:
       # Assume the therock-ci OIDC role in ROCm/TheRock. Other repos
       # fall back to runner base credentials (therock-ci-artifacts-external).
       - name: Configure AWS Credentials
-        if: ${{ github.repository == 'ROCm/TheRock' && !github.event.pull_request.head.repo.fork }}
+        if: ${{ always() && github.repository == 'ROCm/TheRock' && !github.event.pull_request.head.repo.fork }}
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
         with:
           aws-region: us-east-2
           role-to-assume: arn:aws:iam::692859939525:role/therock-ci
 
+      # TODO(#3336): Attempt upload `always()`?
+      #   * If some part of the build failed, partial artifacts may be useful for debugging.
+      #   * This script fails if `build_dir / "artifacts"` does not exist. That's useful
+      #     if the build succeeded but did not produce artifacts. That's _not_ useful
+      #     if the workflow failed or was cancelled before the build even started though.
       - name: Push stage artifacts
         run: |
           python build_tools/artifact_manager.py push --run-id ${{ github.run_id }} \
@@ -154,7 +159,7 @@ jobs:
             --build-dir="${BUILD_DIR}"
 
       - name: Upload stage logs
-        if: ${{ !cancelled() }}
+        if: ${{ always() }}
         run: |
           python build_tools/github_actions/post_stage_upload.py \
             --run-id=${{ github.run_id }} \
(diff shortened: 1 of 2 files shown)


.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -141,20 +141,25 @@ jobs:
...
         uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
         with:
           aws-region: us-east-2
           role-to-assume: arn:aws:iam::692859939525:role/therock-ci
 
+      # TODO(#3336): Attempt upload `always()`?
+      #   * If some part of the build failed, partial artifacts may be useful for debugging.
+      #   * This script fails if `build_dir / "artifacts"` does not exist. That's useful
+      #     if the build succeeded but did not produce artifacts. That's _not_ useful
+      #     if the workflow failed or was cancelled before the build even started though.
```
HereThereBeDragons, line 154: are we only uploading complete artifacts? or could it be that a broken build will result in broken artifacts/partially created artifacts?

a bit of side tracking: do we have a good policy how to mark artifacts as healthy or unhealthy to know which can be used as pre-built?
possible scenario: RL is in rocm-libraries but depends on rocm-system/RS. RL tests are failing due to RS, but RS is not seeing those failures due to missing test coverage - but because RS didnt show any failures we think that artifact is a healthy one eventhough it isnt.

ScottTodd: > are we only uploading complete artifacts? or could it be that a broken build will result in broken artifacts/partially created artifacts?

Yes, currently we only upload complete artifacts.

Consider a few cases:
1. For a change to the compiler, the `amd-llvm` artifact is built successfully but the `hip-clr` artifact fails to compile _with that newer compiler_. Here the `amd-llvm` artifact could be considered "broken" and we would have partial artifacts from the "compiler-runtime" stage. The partial artifacts may be useful for debugging.
2. For a change to HIP, `hip-clr` fails to compile. Here we have an incomplete set of "valid" (not "broken") artifacts.

> a bit of side tracking: do we have a good policy how to mark artifacts as healthy or unhealthy to know which can be used as pre-built?

Not yet. So far the pre-built artifact code paths require manually specifying a workflow run and list of stages. I have a few ideas:
* (conservative + safe, but harder to implement) Only pull prebuilt artifacts where the source fingerprint (fprint) matches exactly. Healthy/unhealthy doesn't matter here - it's _just_ a build system optimization that doesn't need to care about the behavior of the artifacts.
* (riskier) Pull prebuilt artifacts from a commit on the base branch. Tests should have already passed for the commit to be merged in the first place.
* (simple but needs continual effort) In rocm-libraries, pull prebuilt artifacts for prior stages ("foundation", "compiler-runtime") using the rocm-systems etc. commit pins matching the TheRock version hash. No dynamic lookup code needed, pull from known-working when we sync across the repos.
* (simpler than the previous point) In rocm-systems, pull prebuilt artifacts for LLVM (a subset of "compiler-runtime")  using a fixed pin

Issue #3290, opened 2026-02-06
[Issue]: rocm-sdk test fails when it runs amdflang on RHEL 10.0


pbhandar-amd:
### Problem Description

rocm-sdk fails because amdflang fails when users don't have `libquadmath` installed when installing latest nightly via pip install on RHEL 10.0.

```
OS:
NAME="Red Hat Enterprise Linux"
VERSION="10.0 (Coughlan)"
CPU: 
model name	: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
GPU:
  Name:                    Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
  Marketing Name:          Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
  Name:                    gfx1100                            
  Marketing Name:          AMD Radeon RX 7900 XTX             
      Name:                    amdgcn-amd-amdhsa--gfx1100         
      Name:                    amdgcn-amd-amdhsa--gfx11-generic 
```

### Operating System

RHEL 10.0 (Coughlan)

### CPU

Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz

### GPU

AMD Radeon RX 7900 XTX (Navi31 XTX)

### ROCm Version

7.12.0a20260206

### ROCm Component

flang

### Steps to Reproduce

I installed latest nightly on a RHEL 10.0 baremetal system with a Navi31 XTX GPU

1. Setup
```
sudo dnf install -y python3.12 python3.12-pip
sudo usermod -a -G render,video $LOGNAME 

python3.12 -m venv .venv
source .venv/bin/activate
```

2. Install `rocm[libraries,devel]`
```
python -m pip install --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/ "rocm[libraries,devel]"
```

3. Run rocm-sdk test
```
rocm-sdk init
rocm-sdk test
```

4. rocm-sdk test error message
```
======================================================================
ERROR: testConsoleScripts (rocm_sdk.tests.core_test.ROCmCoreTest.testConsoleScripts) [Check console-script amdflang]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/rocm/.venv/lib64/python3.12/site-packages/rocm_sdk/tests/core_test.py", line 135, in testConsoleScripts
    output_text = subprocess.check_output(
                  ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3119, opened 2026-01-27
Test `rocm-sdk test` with and without `rocm[devel]` package

ScottTodd:
## Motivation

This `test_rocm_wheels.yml` workflow was added in https://github.com/ROCm/TheRock/pull/3099 as part of https://github.com/ROCm/TheRock/issues/1559, but running the workflow on our GPU runners revealed test failures: https://github.com/ROCm/TheRock/issues/3115. The `rocm-sdk test` tests succeed in our [`test_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml) workflow since that installs `torch` which depends on just `rocm[libraries]`, _not_ `rocm[libraries,devel]`.

This PR switches to first testing 

## Technical Details

This uses a matrix to test across `rocm[libraries]` and `rocm[libraries,devel]` so we can see error messages from both and they each use their own fully isolated environment. I also considered running both tests sequentially in a single job (sharing or not sharing a venv)

## Test Plan and Results

Triggered workflow runs:

* Linux, gfx94X-dcgpu, rocm version `7.12.0a20260129`: https://github.com/ROCm/TheRock/actions/runs/21489047443
* Windows, gfx110X-all, rocm version `7.12.0a20260129`: https://github.com/ROCm/TheRock/actions/runs/21489109174

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4248, opened 2026-03-30
[ci] Add UBI 10 image to rocm wheel test matrix

lucbruni-amd:
Refs: [TheRock#3290](https://github.com/ROCm/TheRock/issues/3290)

## Motivation

Add a minimal RHEL 10 (UBI 10) container image with a matrix alongside the existing `no_rocm_image_ubuntu24_04` entry to the `multi_arch_ci_linux` workflow at the `test_python_packages_per_family` step to increase test coverage across distros. The existing Ubuntu image ships libraries like `libquadmath` pre-installed, masking missing dependency issues that surface on minimal installations.

## Technical Details

`test_rocm_wheels.yml` is now image-agnostic. The container image is supplied by the caller via two inputs available on both `workflow_dispatch` and `workflow_call`.

- `container_image_name`: short label used in job names and for logic.
- `container_image_url`: full image ref, applied only when `test_runs_on` targets Linux. Empty means run directly on the runner natively.

For `workflow_call` both inputs are required with no defaults so callers must state their image explicitly. For `workflow_dispatch` both keep a prefilled Ubuntu 24 default so the "Run workflow" form stays one-click and any new test image URL can be pasted in for a one-off run without editing yaml.

The image matrix now lives in `multi_arch_ci_linux.yml`, which fans `test_python_packages_per_family` across the existing `ghcr.io/rocm/no_rocm_image_ubuntu24_04` and now `ghcr.io/rocm/no_rocm_image_ubi10` images. Nightly multi-arch runs therefore exercise every family on both images and would surface issues like the above-referenced.

## Test Plan

Ensure the job structure is expected and individual jobs test ROCm wheels with `rocm-sdk test` on respective environments.

## Test Result

Each job ran green in conjunction with the prerequisite PRs referenced in events below.

- Linux Ubuntu24: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71890205263?pr=4248
- Linux UBI10: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71890205342?pr=4248
- Windows Native: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71897900838?pr=4248

## Submission Checklist


.github/workflows/test_rocm_wheels.yml
```diff
@@ -86,6 +93,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Install additional packages (UBI 10)
+        if: matrix.image.name == 'ubi10'
+        run: |
+          dnf install -y --nodocs libatomic libquadmath
```
lucbruni-amd: Do we want to follow the `no_rocm_image*` convention more formally here? Such as creating a `no_rocm_image_ubi10.Dockerfile` and publishing it alongside the Ubuntu one. This would avoid the inline `dnf` calls, but I wasn't sure which is preferred.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -76,6 +76,13 @@ jobs:
       fail-fast: false
       matrix:
         packages: ["rocm[libraries]", "rocm[libraries,devel]"]
+        # ubuntu: public Ubuntu 24.04 image
+        # ubi10: public RHEL 10 UBI image
+        image:
+          - name: ubuntu
+            url: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
+          - name: ubi10
+            url: registry.access.redhat.com/ubi10/ubi@sha256:c9c81d3d11bfd56c4bb61a220d4598392b5fbc500df33f2b8e52fdd2cebb4944
```
geomin12: we like to host our own docker images as things may change / images may get deleted in other repos. let's re-create this as no_rocm_image_ubi10 (similar to comment below)

.github/workflows/test_rocm_wheels.yml
geomin12: yes, let's create a docker file with those packages! what if a developer wants to use docker img to debug? would be easier that way :) and cleaner too

geomin12 (commented):
looks like tests are failing with ubi10 img (some py3.12 errors)

.github/workflows/test_rocm_wheels.yml
```diff
@@ -76,6 +76,13 @@ jobs:
       fail-fast: false
       matrix:
         packages: ["rocm[libraries]", "rocm[libraries,devel]"]
+        # ubuntu: public Ubuntu 24.04 image
+        # ubi10: public RHEL 10 UBI image
+        image:
+          - name: ubuntu
+            url: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
```
ScottTodd: Expanding the matrix like this may be the push we needed to choose a different option from https://github.com/ROCm/TheRock/pull/3119#discussion_r2736945862, especially as we add other distros to this list.

Rather than have 2 x 2 jobs:
* test rocm[libraries] on ubuntu
* test rocm[libraries,devel] on ubuntu
* test rocm[libraries] on RHEL
* test rocm[libraries,devel] on RHEL

We could change to 2 jobs with two test steps:
* on ubuntu, test rocm[libraries] then test rocm[libraries,devel]
* on RHEL, test rocm[libraries] then test rocm[libraries,devel]

I don't want too many jobs in the job graph, as that costs more in runner overhead and lower information density in the workflow logs.

.github/workflows/test_rocm_wheels.yml
ScottTodd: This has test failures: https://github.com/ROCm/TheRock/actions/runs/23763117625/job/69259909175?pr=4248

Feel free to test changes like this without creating a PR, by just triggering https://github.com/ROCm/TheRock/actions/workflows/test_rocm_wheels.yml with workflow_dispatch. That skips straight to the test step, without needing to first build all of ROCm (2-4 hours)

.github/workflows/test_rocm_wheels.yml
ScottTodd: Agreed. Let's create a new test dockerfile.

See https://github.com/ROCm/TheRock/tree/main/dockerfiles#no_rocm_image_dockerfile, it's already set up to include other distros.

lucbruni-amd: Resolved by consensus via https://github.com/ROCm/TheRock/pull/4280.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -75,7 +75,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        packages: ["rocm[libraries]", "rocm[libraries,devel]"]
+        image:
+          - name: ubuntu
+            url: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
+          - name: ubi10
+            url: ghcr.io/rocm/no_rocm_image_ubi10:latest # pin digest after first publish
```
lucbruni-amd: To pin once the paired PR is merged and the image is published.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -91,20 +95,37 @@ jobs:
...
+        run: |
+          rocm-sdk test
+
+      - name: Install 'rocm[libraries,devel]' into existing venv
+        run: |
+          python build_tools/setup_venv.py ${VENV_DIR} \
+            --packages "rocm[libraries,devel]==${{ inputs.rocm_version }}" \
+            --index-url=${{ inputs.package_index_url_base }} \
+            --index-subdir=${{ inputs.amdgpu_family }} \
+            --find-links=${{ inputs.package_find_links_url }}
```
pbhandar-amd: I think it might be a good idea to uninstall the existing ROCm installation that was installed when you installed rocm[libraries] _before_ installing rocm[libraries,devel]. I think it's better to have a clean test environment instead of re-using the same venv as before. You can see the instructions for [RHEL uninstall here](https://rocm.docs.amd.com/en/7.12.0-preview/install/rocm.html?fam=instinct&gpu=mi300x&os=rhel&os-version=8.10&i=pip#installation) and for [Ubuntu uninstall here](https://rocm.docs.amd.com/en/7.12.0-preview/install/rocm.html?fam=instinct&gpu=mi300x&os=ubuntu&os-version=24.04&i=pip#installation).

.github/workflows/test_rocm_wheels.yml
ScottTodd: (Let's make any deeper change to the matrix as suggested in this comment in a separate PR, not this PR itself)

lucbruni-amd: Ah. Should I revert back to the 2x2 setup in this PR?

ScottTodd: Yeah. You can either proceed with this using 2x2 or we can do the other work as a prerequisite PR landed first.

lucbruni-amd: Opened https://github.com/ROCm/TheRock/pull/4297 to address this. Converting this one to a draft until it lands.

.github/workflows/test_rocm_wheels.yml
lucbruni-amd: Marking as resolved as I've moved relevant changes and discussion to https://github.com/ROCm/TheRock/pull/4297.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -96,6 +96,12 @@ jobs:
         --user 0:0 # Running as root, by recommendation of GitHub: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support#user
     strategy:
       fail-fast: false
+      matrix:
+        image:
+          - name: ubuntu24.04
+            url: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
+          - name: ubi10
+            url: ghcr.io/rocm/no_rocm_image_ubi10@sha256:a10f34d6006a20d02cf688982de9dea147710927ed405a3b0d5c73b58a6030c0
```
ScottTodd: This doesn't look correct on Windows: https://github.com/ROCm/TheRock/actions/runs/24473053147/job/71556515229?pr=4248.
Two jobs, running the same code:
* `Windows::release / Test Python gfx110X-all / Test ROCm wheels | gfx110X-all | ubuntu24.04`
* `Windows::release / Test Python gfx110X-all / Test ROCm wheels | gfx110X-all | ubi10`

Something like this may work:
```yml
  strategy:
    fail-fast: false
    matrix:
      image:
        - name: ubuntu24.04
          url:
  ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
          os: linux
        - name: ubi10
          url: ghcr.io/rocm/no_rocm_image_ubi10@sha256:a10f34d6006a20d02cf688982de9dea147710927ed405a3b0d5c73b58a6030c0
          os: linux
        - name: native
          url: ""
          os: windows
      exclude:
        - image:
            os: ${{ contains(inputs.test_runs_on, 'linux') && 'windows' || 'linux' }}
```

Otherwise we could try this (per recommendation by Claude Code):
```yml
  strategy:
    fail-fast: false
    matrix:
      image: ${{ fromJSON(
        contains(inputs.test_runs_on, 'linux')
          && '[{"name":"ubuntu24.04","url":"ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d
  4cba4a383c1a7459b28627047bf6302a26"},{"name":"ubi10","url":"ghcr.io/rocm/no_rocm_image_ubi10@sha256:a10f34d6006a20d02c
  f688982de9dea147710927ed405a3b0d5c73b58a6030c0"}]'
          || '[{"name":"native","url":""}]'
      ) }}
```

---

Actually, taking a step back further, we could push the container choice out to the caller of test_rocm_wheels.yml, here alongside the existing `family_info`? https://github.com/ROCm/TheRock/blob/d177c0127c13e9974dd9e4b2a30f39384668f623/.github/workflows/multi_arch_ci_linux.yml#L134-L141 Then this workflow stays modular - it just runs on the container you select. If you want to trigger a one-off test run on a new image that's just a regular workflow_dispatch without needing to edit code.

That would look like:

```diff
  test_python_packages_per_family:
    needs: [build_python_packages]
    name: Test Python ${{ matrix.family_info.amdgpu_family }}
    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).expect_failure == false }}
    strategy:
      fail-fast: false
      matrix:
        family_info: ${{ fromJSON(inputs.build_config).per_family_info }}
+       image:
+         - name: ubuntu24.04
+           url: ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26
+         - name: ubi10
+           url: ghcr.io/rocm/no_rocm_image_ubi10@sha256:a10f34d6006a20d02cf688982de9dea147710927ed405a3b0d5c73b58a6030c0
    uses: ./.github/workflows/test_rocm_wheels.yml
  with:
      amdgpu_family: ${{ matrix.family_info.amdgpu_family }}
      test_runs_on: ${{ matrix.family_info.test-runs-on }}
+     container_image_name: ${{ matrix.image.name }}
+     container_image_url: ${{ matrix.image.url }}
```

lucbruni-amd: Good catch on the Windows side of things, and I agree with what you're saying in regards to pushing the choice outwards + being able to trigger a one-off test with a new image via `workflow_dispatch`. I just pushed https://github.com/ROCm/TheRock/pull/4248/changes/0310320d21097ce53437cc0d3af6794d4614dd58 which hopefully closely matches what you wanted. Let me know if you feel there should be changes.

Main ideas for callers:
- `multi_arch_ci_linux.yml` _should_ produce two wheel-test jobs per family ("... | ubuntu24.04" and "... | ubi10") that both run on Linux
- `multi_arch_ci_windows.yml` produces exactly one wheel-test job per family labeled "... | native" running on Windows, with no erroneous jobs you pointed out
- `ci_linux.yml` produces one wheel-test job on `ubuntu24.04`
- `ci_windows.yml` produces one wheel-test job on `native` Windows
- A manual `workflow_dispatch` run with the prefilled defaults executes on the pinned Ubuntu 24 image (this one I was battling between the SHA256 or `latest` tags, do you have a preference in this case?)
- A manual `workflow_dispatch` _should_ run with any new image URL executes on that image with no code change required to the workflow file itself

ScottTodd: Yep, this looks good! Let's check that it works as we expect.

ScottTodd: * Linux ubuntu: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71890205263?pr=4248
* Linux ubi: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71890205342?pr=4248
* Windows native: https://github.com/ROCm/TheRock/actions/runs/24574808154/job/71897900838?pr=4248

Seems correct in all cases

ScottTodd (approved):
Can you refresh the PR description to reflect the latest code? Otherwise LGTM

PR #4251, opened 2026-03-30
[sysdeps] Fix zstd symbol visibility and -Bsymbolic linking

stellaraccident:
The original -Bsymbolic flag was missing the -Wl, prefix so it never
reached the linker. Additionally, zstd was exporting 391 symbols
(including internals like FSE_*, XXH*, HUF_*) due to default visibility.

This caused crashes on systems (e.g. Debian) where system zstd loaded
alongside ROCm's bundled copy, with system zstd's ZSTD_freeDCtx being
called during ROCm zstd's ZSTD_decompress.

Changes:
- Fix -Bsymbolic to -Wl,-Bsymbolic so the flag reaches the linker
- Build with -fvisibility=hidden and set ZSTDLIB_VISIBLE,
  ZSTDERRORLIB_VISIBLE, ZDICTLIB_VISIBLE to "default" so only the
  public API is exported (391 -> 187 symbols)

Verified locally: readelf shows SYMBOLIC flag present, internal symbols
(FSE_*, XXH*, HUF_*) no longer exported.

Progress on https://github.com/ROCm/TheRock/issues/4211

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/third-party/sysdeps/common/zstd/CMakeLists.txt
+++ b/third-party/sysdeps/common/zstd/CMakeLists.txt
@@ -65,7 +65,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set(CONFIGURE_ARGS
     -DZSTD_BUILD_STATIC=OFF
     -DZSTD_BUILD_SHARED=ON
-    "-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds -Bsymbolic"
+    "-DCMAKE_C_FLAGS=-fvisibility=hidden"
+    -DZSTDLIB_VISIBLE=default
+    -DZSTDERRORLIB_VISIBLE=default
+    -DZDICTLIB_VISIBLE=default
+    "-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/version.lds -Wl,-Bsymbolic"
   )
   set(POST_INSTALL_COMMANDS
     COMMAND


lumachad:
Let me go through this, but something seems seriously broken with the rocm-side of rocgdb tests, as if things are not running properly due to some fundamental breakage.

ScottTodd:
> Let me go through this, but something seems seriously broken with the rocm-side of rocgdb tests, as if things are not running properly due to some fundamental breakage.

This issue maybe?
* https://github.com/ROCm/TheRock/issues/4264

lumachad:
> > Let me go through this, but something seems seriously broken with the rocm-side of rocgdb tests, as if things are not running properly due to some fundamental breakage.
> 
> This issue maybe?
> 
> * [[Issue] CI Nightly: rocgdb tests failing due to missing core dumps + timeouts #4264](https://github.com/ROCm/TheRock/issues/4264)

Yes, thanks for the link. I was chasing it in the background.

ScottTodd:
> I'd trigger CI again just in case. It seems something was off with the sanity checks

This https://github.com/ROCm/TheRock/actions/runs/23771970963/job/69286381964?pr=4251 ? We can re-run just that job and its dependents: 
image

Or we can re-run all failed jobs: 
image

I don't think we need a full refresh of the entire CI run (including builds) for this, if we just suspect machine issues for tests.

lumachad:
I agree re-running just those failed tests works.

lumachad:
This run was OK. rocgdb is supposed to fail due to the compiler bump regression.

stellaraccident:
> This run was OK. rocgdb is supposed to fail due to the compiler bump regression.

Agreed/thanks. Merging.

PR #4253, opened 2026-03-30
[ci] Adding sandbox nodes

geomin12:
Adding sandbox nodes as more capacity is added from OSSCI


geomin12:
need to fix unit tests

geomin12:
sandbox nodes cannot be upgraded, so no need for sandbox nodes from OSSCI, closing

Issue #3517, opened 2026-02-19
[issue] Build failure: GCC 15: missing transitive headers (multiple components)


benrichard-amd:
Split from #3477 

### Environment

* OS: Ubuntu 25.04
* Compiler: GCC 15 (defaults to -std=gnu23)
* Python: 3.13
* GPU target: gfx1151 (AMD Strix Halo iGPU / Radeon 8060S)
* Terminal: Kitty
* Build method: standard cmake + ninja with .venv

### Instructions to reproduce:

On a fresh install of Ubuntu 25.10:
```
cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES=gfx1151 -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
cmake --build build
```

GCC 15 no longer provides headers like `<cstdint>` or `<algorithm>` transitively. This caused errors in several bundled libraries:

| File | Error | Fix |
|------|-------|-----|
| `rocm-systems/projects/rocprofiler-sdk/external/yaml-cpp/src/emitterutils.cpp` | `unknown type name 'uint16_t'` | Add `#include <cstdint>` |
| `rocm-systems/projects/rocprofiler-sdk/external/elfio/elfio/elf_types.hpp` | `unknown type name 'uint16_t'` | Add `#include <cstdint>` |
| `rocm-systems/projects/rocprofiler-systems/source/lib/logger/logger.hpp` | `'any_of' is not a member of 'std'` | Add `#include <algorithm>` |
| `rocm-systems/projects/rocprofiler-systems/external/dyninst/common/h/dyntypes.h` | `'uint32_t' does not name a type` | Add `#include <stdint.h>` |
| `rocm-libraries/projects/hipdnn/plugins/miopen_legacy_plugin/MiopenContainer.hpp` | `unknown type name 'uint32_t'` | Add `#include <cstdint>` |

benrichard-amd:
The build failures with
`rocm-systems/projects/rocprofiler-sdk/external/yaml-cpp/src/emitterutils.cpp`

and

`rocm-systems/projects/rocprofiler-sdk/external/elfio/elfio/elf_types.hpp`

 should be fixed by bumping `elfutil` and `yaml-cpp` submodules.

Note latest `yaml-cpp` requires >= CMake 3.5

marbre:
[...]

PR #4142, opened 2026-03-24
Revert "Reapply "Switch most third-party subprojects to the amd-llvm toolchain" (#3474)"

ScottTodd:
Reverts ROCm/TheRock#4032 to mitigate https://github.com/ROCm/TheRock/issues/4143

## Motivation

This contributed new build failures to the ASan nightly workflow (https://github.com/ROCm/TheRock/actions/workflows/ci_asan.yml?query=branch%3Amain):
* First error before: https://github.com/ROCm/TheRock/actions/runs/23227192107/job/67512361026#step:10:1449
    ```
  [ocl-clr] [203/209] Linking CXX executable tests/ocltst/ocltst
  [ocl-clr] FAILED: tests/ocltst/ocltst 
[...]

PR #4258, opened 2026-03-31
Add third-party/elfio

lamikr:
Add third-party/elfio project for therock
to allow profiler tools like rocprofiler-sdk to
use it from standard location instead of each subproject downloading and building it separately.
their own build phase.

This allows standardizasing to single version of these dependencies and to have a centralized place to fix problems.

In addition latest release has been patched to allow gcc/amd-llvm which uses cpp-header files provided by the never gcc-versions like gcc 15 to solve a following build error.

therock/build/third-party/elfio/build_elfio/
dist/include/elfio/elf_types.hpp:30:20:
error: unknown type name 'uint16_t'
9.5	   30 | using Elf_Half   = uint16_t;

This error happens because in never cpp header files other header files like vector does not anymore
include the cstdint directly and therefore the source code which needs those definitions needs to include the cstdint directly.

Fixes partially: https://github.com/ROCm/TheRock/issues/3517


patches/third-party/elfio/0001-Explicitly-include-cstdint-for-gcc-15.patch
```diff
@@ -0,0 +1,34 @@
+From 884905d589a5eb1d33af5d191b11fc34a47e736d Mon Sep 17 00:00:00 2001
+From: Mika Laitio <mika.laitio@amd.com>
+Date: Mon, 30 Mar 2026 22:00:46 -0700
+Subject: [PATCH] Explicitly-include-cstdint for gcc 15
+
+Fixes a following error on gcc 15:
+
+therock/build/third-party/elfio/build_elfio/
+dist/include/elfio/elf_types.hpp:30:20:
+error: unknown type name 'uint16_t'
```
ScottTodd, line 10: If this subproject/dependency builds with clang, I'd prefer that over carrying a patch for gcc. May need to roll forward https://github.com/ROCm/TheRock/pull/4142 again for that. I can get that ready soon.

third-party/elfio/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+therock_subproject_fetch(therock-elfio-sources
+  CMAKE_PROJECT
+  # Originally mirrored from: "https://github.com/serge1/ELFIO/releases/download/Release_3.12/elfio-3.12.tar.gz"
+  URL "https://github.com/serge1/ELFIO/releases/download/Release_3.12/elfio-3.12.tar.gz"
```
ScottTodd: Let's upload this to our `rocm-third-party-deps` S3 bucket

third-party/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ add_subdirectory(nlohmann-json)
 add_subdirectory(simde)
 add_subdirectory(yaml-cpp)
 add_subdirectory(Catch2)
+add_subdirectory(elfio)
```
ScottTodd: sort alphabetically

profiler/CMakeLists.txt
```diff
@@ -107,6 +108,8 @@ if(THEROCK_ENABLE_ROCPROFV3)
       rocprofiler-sdk-tests
     COMPILER_TOOLCHAIN
       amd-hip
+    BUILD_DEPS
+      therock-elfio
```
ScottTodd: > fixes: #3517

There were a few build errors tracked on that issue, not just elfio. Is it actually fully fixed with this PR?

third-party/elfio/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+)
+
+therock_cmake_subproject_declare(therock-elfio
+  USE_DIST_AMDGPU_TARGETS
+  EXCLUDE_FROM_ALL
+  NO_MERGE_COMPILE_COMMANDS
+  OUTPUT_ON_FAILURE
+  EXTERNAL_SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/source"
+  COMPILER_TOOLCHAIN
+    amd-llvm
```
ScottTodd, line 19: Build error: https://github.com/ROCm/TheRock/actions/runs/23782547594/job/69298327763?pr=4258#step:12:139
```
-- Including subproject therock-elfio (from /__w/TheRock/TheRock/build/third-party/elfio/source)
CMake Error at cmake/therock_subproject.cmake:1221 (get_target_property):
  get_target_property() called with non-existent target "amd-llvm".
```

Actually wait, my other comment about the patch... this _is_ trying to build with LLVM. Huh? Why carry a patch for GCC if not building with GCC?

profiler/CMakeLists.txt
lamikr: No, this is only a partial fix, I changed the wording to "Fixes partially #3517" in the PR description.

Note that.

1) Note that the rocprofiler-sdk changes to support this in 
https://github.com/ROCm/rocm-systems/pull/4579
needs to get in before we can merge this PR. (#4258)

I made now small change remove the rocprofiler-sdk build flag which enables the use of therock-elfio in profiler/CMakeLists.txt.

If it's ok for you, I think it's better to create a separate PR only for that once this and https://github.com/ROCm/rocm-systems/pull/4579 has landed?

2) Yaml-cpp fix is needed also to fix another gcc 15 build error.
https://github.com/ROCm/rocm-systems/pull/4223

PR 4223 review requires some changes before it can be landed. I followed the suggestion of that review on my 4579 PR.  I am not sure should I update the 4223 PR also or wait couple of days if [Muhamed-Husic](https://github.com/Muhamed-Husic) has time to do it.

third-party/elfio/CMakeLists.txt
lamikr: amd-llvm uses header files and libraries that gcc provides also for it's own builds.

Version of header files that older gcc provided were included the cstdint via other header files like vector, so the definitions that cstdint provides got often pulled in inderectly.

With the header files that never gcc version provides this is not the case anymore and therefore the <cstdint> needs to be included more explicitly in places of code that depends from those definitions.

patches/third-party/elfio/0001-Explicitly-include-cstdint-for-gcc-15.patch
lamikr: amd-clang uses in many cases the header files and libraries that gcc provides, it justs builds the new code produced differently. I have longer explanation another reply below.

third-party/CMakeLists.txt
lamikr: Updated the whole list that seemed to be only partially sorted. Change pushed as new patch.

third-party/elfio/CMakeLists.txt
lamikr: Updated, thanks for catching, I needed to have link to official place to test this but planned to do the change before sending the PR but forgot it.

third-party/elfio/CMakeLists.txt
lamikr: I am not sure what is causing this

```
-- Including subproject therock-elfio (from /__w/TheRock/TheRock/build/third-party/elfio/source)
CMake Error at cmake/therock_subproject.cmake:1221 (get_target_property):
  get_target_property() called with non-existent target "amd-llvm".
```

I do not see this error in my local test build, looks like an race condition where the compiler is tried to be used before it's build. I anyway added the amd-llvm to be a BUILD_DEPENDENCY for therock-elfio and at least my local build worked also in that way.

I see this comment on third-party/CMakeLists.txt

```
# Subdirectory load order here is carefully balanced - see usage from the root
# CMakeLists.txt. Notably this file does *not* include the sysdeps subdirectory,
# which is expected to be included *first*. This allows for bootstrapping
# compiler/, so other projects in this directory can use the amd-llvm compiler
# toolchain if they so wish (most should eventually).
```

ScottTodd: See my other comment. We may need to roll forward https://github.com/ROCm/TheRock/pull/4142 before this can use amd-llvm.

third-party/elfio/CMakeLists.txt
lamikr: I also uploaded the package now to 
https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/elfio-3.12.tar.gz

third-party/elfio/CMakeLists.txt
marbre: Either it should be abler to build with amd-llvm in which case we can drop the patch _or_ we land and add the patch for gcc (which seems okay to me) but it would in that case the toolchain line need to go away here.

third-party/elfio/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+  BUILD_DEPS
+    amd-llvm
```
marbre: There shouldn't be a build dep on amd-llvm in any of those cases, otherwise this is serious issue.

BUILD_TOPOLOGY.toml
```diff
@@ -401,6 +401,12 @@ feature_name = "HOST_SUITE_SPARSE"
 feature_group = "HOST_MATH"  # Controlled by THEROCK_ENABLE_HOST_MATH (OFF by default)
 disable_platforms = ["windows"]
 
+[artifacts.elfio]
+artifact_group = "third-party-libs"
+type = "target-neutral"
+artifact_deps = ["amd-llvm"]
```
marbre: See comments above.

BUILD_TOPOLOGY.toml
```diff
@@ -401,6 +401,12 @@ feature_name = "HOST_SUITE_SPARSE"
 feature_group = "HOST_MATH"  # Controlled by THEROCK_ENABLE_HOST_MATH (OFF by default)
 disable_platforms = ["windows"]
 
+[artifacts.elfio]
```
marbre, line 404: Should this be disabled for Windows? profilers only get build on Linux thus we don't need to build this dep on Windows at all.

patches/third-party/elfio/0001-Explicitly-include-cstdint-for-gcc-15.patch
marbre, line 1: Would a newer upstream version fix this?

lamikr: This is latest official release, upstream unreleased git version has also other changes so at least I feel that safest way is at least for now to use the version that has been used earlier with the patch included.

third-party/elfio/CMakeLists.txt
lamikr: Patch is needed because headers from /usr/include/c++/15 are used no matter whether you build with the gcc or with the amd-clang.

third-party/elfio/CMakeLists.txt
lamikr: Patch is needed because headers from /usr/include/c++/15 are used no matter whether you build with the gcc or with the amd-clang.

third-party/elfio/CMakeLists.txt
marbre: Not talking about the patch, talking about the build dep here. There shouldn't be a build dep (linking) to LLVM.

third-party/elfio/CMakeLists.txt
lamikr: Just to show here that error happens with the clang also without the patch when rocprofiler-sdk that uses elfio headers is build if elfio is not patched.

```
therock/build/core/clr/dist/lib/llvm/bin/clang++ -DAMD_INTERNAL_BUILD=1 -DGLOG_USE_GLOG_EXPORT -DROCPROFILER_DL=1 -DROCPROFILER_HAS_GHC_LIB_FILESYSTEM=1 -DROCPROFILER_SDK_USE_SYSTEM_RCCL=0 -DROCPROFILER_SDK_USE_SYSTEM_ROCDECODE=0 -DROCPROFILER_SDK_USE_SYSTEM_ROCJPEG=0 -DUSE_PROF_API=1 -DYAML_CPP_STATIC_DEFINE -D__HIP_PLATFORM_AMD__=1 -I/home/lamikr/therock/build/third-party/sysdeps/linux/libdrm/build/stage/lib/rocm_sysdeps/include -I/home/lamikr/therock/build/third-party/sysdeps/linux/numactl/build/stage/lib/rocm_sysdeps/include -I/home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/source -I/home/lamikr/therock/build/profiler/rocprofiler-sdk/build/source/include -I/home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/source/include -I/home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/external/yaml-cpp/include -I/home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/external/ptl/source -I/home/lamikr/therock/build/profiler/rocprofiler-sdk/build/external/ptl/source -isystem /home/lamikr/therock/build/core/clr/dist/include -isystem /home/lamikr/therock/build/core/ROCR-Runtime/dist/include -isystem /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/external/filesystem/include -isystem /home/lamikr/therock/build/profiler/rocprofiler-sdk/build/external/glog -isystem /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/external/glog/src -isystem /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/external/fmt/include -isystem /home/lamikr/therock/build/third-party/elfio/dist/include -isystem /home/lamikr/therock/build/third-party/sysdeps/linux/elfutils/build/dist/lib/rocm_sysdeps/include -Wno-documentation-unknown-command -Wno-documentation-pedantic -Wno-unused-command-line-argument -Wno-explicit-specialization-storage-class --hip-path=/home/lamikr/therock/build/core/clr/dist --hip-device-lib-path=/home/lamikr/therock/build/core/clr/dist/lib/llvm/amdgcn/bitcode -O3 -DNDEBUG -std=c++17 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -W -Wall -Wno-unknown-pragmas -faligned-new -rdynamic -fstack-protector-strong -Wstack-protector -MD -MT source/lib/common/CMakeFiles/rocprofiler-sdk-common-library.dir/elf_utils.cpp.o -MF source/lib/common/CMakeFiles/rocprofiler-sdk-common-library.dir/elf_utils.cpp.o.d -o source/lib/common/CMakeFiles/rocprofiler-sdk-common-library.dir/elf_utils.cpp.o -c /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/source/lib/common/elf_utils.cpp
[rocprofiler-sdk] In file included from /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/source/lib/common/elf_utils.cpp:23:
[rocprofiler-sdk] In file included from /home/lamikr/therock/rocm-systems/projects/rocprofiler-sdk/source/lib/common/elf_utils.hpp:25:
[rocprofiler-sdk] In file included from /home/lamikr/therock/build/third-party/elfio/dist/include/elfio/elfio.hpp:36:
[rocprofiler-sdk] /home/lamikr/therock/build/third-party/elfio/dist/include/elfio/elf_types.hpp:30:20: error: unknown type name 'uint16_t'
[rocprofiler-sdk]    30 | using Elf_Half   = uint16_t;
[rocprofiler-sdk]       |                    ^
[rocprofiler-sdk] /home/lamikr/therock/build/third-party/elfio/dist/include/elfio/elf_types.hpp:31:20: error: unknown type name 'uint32_t'
[rocprofiler-sdk]    31 | using Elf_Word   = uint32_t;
```

third-party/elfio/CMakeLists.txt
lamikr: I pushed "Revert "elfio add amd-llvm as a build dependency" to fix this. I didn't realie that amd-llvm as a BUILD_DEBS causes the linking to happen. I though I in this way just ensured that the amd-llvm will be build first.

BUILD_TOPOLOGY.toml
lamikr: pushed BUILD_TOPOLOGY changes
    
- removed amd-llvm from artifact_deps
- added disable_platforms = ["windows"]

BUILD_TOPOLOGY.toml
lamikr: added disable_platforms = ["windows"]

third-party/elfio/CMakeLists.txt
marbre: The better fix would be to set the COMPILER_TOOLCHAIN as discussed above.

profiler/CMakeLists.txt
Muhamed-Husic: Sorry for the delay, I hadn’t gotten to those changes yet. I’ll take care of updating PR 4223 now.

third-party/elfio/CMakeLists.txt
lamikr: thirdparty/elfio/CMakeLists.txt has the COMPILER_TOOLCHAIN already defined:

```
therock_cmake_subproject_declare(therock-elfio
  USE_DIST_AMDGPU_TARGETS
  EXCLUDE_FROM_ALL
  NO_MERGE_COMPILE_COMMANDS
  OUTPUT_ON_FAILURE
  EXTERNAL_SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/source"
  COMPILER_TOOLCHAIN
    amd-llvm
  CMAKE_ARGS
    -DELFIO_BUILD_EXAMPLES=OFF
    -DELFIO_BUILD_TESTS=OFF
)
```

third-party/CMakeLists.txt
ScottTodd, line 33: Ok, I remember the details now.

Move elfio down into a block like this if you want to use the amd-llvm toolchain, with a `THEROCK_ENABLE_ELFIO` check.

This is trying to build during the "foundation" stage, where LLVM is not yet available, because the subproject is unconditionally enabled.

lamikr: So this is not enough in the third-party/elfio/CMakeLists.txt?

```
COMPILER_TOOLCHAIN
    amd-llvm
```

I have little trouble understanding why the THEROCK_ENABLE_ELFIO check will help here except that the CI will probably by default not try to skip building the elfio unless we will add the -DTHEROCK_ENABLE_ELFIO=1 builf flag there.

lamikr: Another thing I am not sure is that some packages like third-party/openmpi/CMakeLists.txt have extra "if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)" check in top of the CMakeLists.txt while many packages does not have this.

Is it because of supporting some kind of alternative ways of building the package outside of therock build system. Example for this type definion is in third-party/openmpi/CMakeLists.txt:

```

if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
    set(_source_dir "${CMAKE_CURRENT_BINARY_DIR}/source")
    set(_download_stamp "${_source_dir}/download.stamp")

    therock_subproject_fetch(therock-openmpi-sources
      SOURCE_DIR "${_source_dir}"
      # Originally mirrored from: "https://download.open-mpi.org/release/open-mpi/v5.0/openmpi-5.0.9.tar.bz2"
      URL "https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/openmpi-5.0.9.tar.bz2"
      URL_HASH "SHA256=dfb72762531170847af3e4a0f21d77d7b23cf36f67ce7ce9033659273677d80b"
      TOUCH "${_download_stamp}"
    )

    therock_cmake_subproject_declare(therock-openmpi
        FPRINT_SOURCE_DIR "${_source_dir}"
    ...
endif()

include(CMakePackageConfigHelpers)

cmake_minimum_required(VERSION 3.25)
project(OpenMPI_BUILD VERSION 5.0.9)
...
```

lamikr: I anyway pushed now the change that makes elfio to be build only if the THEROCK_ENABLE_ELFIO flag has been set.

```
if(THEROCK_ENABLE_ELFIO)
  add_subdirectory(elfio)
endif()
```

third-party/elfio/artifact-elfio.toml
```diff
@@ -0,0 +1,4 @@
+# elfio
+[components.dev."third-party/elfio/stage"]
+[components.doc."third-party/elfio/stage"]
+[components.lib."third-party/elfio/stage"]
```
ScottTodd, line 4: Are these artifacts enough / useful as-is? doc and lib are empty from https://therock-ci-artifacts.s3.amazonaws.com/24208372329-linux/index.html (just manifest files), and "dev" is just source files (C++ and CMake)

Seems like this could just be a third party project used entirely internal to the build system without an associated artifact.

BUILD_TOPOLOGY.toml
ScottTodd: Looking at the contents of the generated artifact, I don't think this (yet?) actually needs an artifact at all.

ScottTodd (commented):
Builds look good now.

ScottTodd (approved):
We can audit and refactor some of these third-party-libs and their artifacts as needed later.

lamikr:
Thanks for the review, I push next the patches for rocprofiler-sdk to allow using the elfio version provied by the therock.

PR #4259, opened 2026-03-31
Add a non-blocking build step, which runs COMGR unit tests.

kasaurov:
## Motivation

Add a non-blocking build step, which runs COMGR unit tests.

## Technical Details

COMGR unit tests is a set of small and fast GPU-agnostic compilation only tests. They include LIT tests, which depends on source tree, so to run them on a separate runner will require copying lots of data. The tests are very fast to run (less than 15 secs). All these make they suitable to be run as a step on the build machine. 
The PR:
-  enables running COMGR lit tests and ctests in Linux and Windows CI pipelines (as "Run comgr tests" non-blocking step)
-  adds `THEROCK_BUILD_COMGR_TESTS` flag propagation through the build system
-  enables LLVM tools needed for test execution
-  adds platform-aware `llvm-lit` installation (Windows .cmd wrapper vs Unix bash wrapper).

Screenshot 2026-04-04 163013

## Test Plan

The TheRock CI run. 

## Test Result

Linux COMGR tests pass, Windows - all lit tests fail. It is expected failure -- Windows run requires the changes from https://github.com/ROCm/llvm-project/pull/1860 to be propagated to TheRock `main` branch, which will happen with llvm compiler version bump.
N.B. The tests failures in rocblas, rocgdb, hip-tests and rocprofiler-sdk in some of the architectures are unrelated to this PR.

## Submission Checklist


.github/workflows/build_portable_linux_artifacts.yml
```diff
@@ -130,6 +130,23 @@ jobs:
...
+      - name: Run comgr tests
+        continue-on-error: true
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          rc1=0
+          python build_tools/teatime.py --log-timestamps \
+            --label "comgr lit tests" "${BUILD_DIR}/logs/comgr-lit-tests.log" -- \
+            "${BUILD_DIR}/compiler/amd-llvm/dist/lib/llvm/bin/llvm-lit" \
+            "${BUILD_DIR}/compiler/amd-comgr/build/test-lit" -v || rc1=$?
```
skganesan008: why do we need to do explicit exit code check?  please explain.

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -130,6 +130,24 @@ jobs:
...
+        if: ${{ inputs.stage_name == 'compiler-runtime' }}
+        continue-on-error: true
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          rc1=0
+          python build_tools/teatime.py --log-timestamps \
+            --label "comgr lit tests" "${BUILD_DIR}/logs/comgr-lit-tests.log" -- \
+            "${BUILD_DIR}/compiler/amd-llvm/dist/lib/llvm/bin/llvm-lit" \
+            "${BUILD_DIR}/compiler/amd-comgr/build/test-lit" -v || rc1=$?
```
ScottTodd, line 143: > They include LIT tests, which depends on source tree, so to run them on a separate runner will require copying lots of data.

Which source files specifically? The compiler is fairly special compared to other ROCm subprojects, but the expected way to integrate tests is to include files needed for testing in a "test" artifact that can then be downloaded, packaged, etc. just like other artifacts.

* https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md
* https://github.com/ROCm/TheRock/blob/main/compiler/artifact-amd-llvm.toml
* https://github.com/ROCm/TheRock/blob/main/docs/development/adding_tests.md

> The tests are very fast to run (less than 15 secs). All these make they suitable to be run as a step on the build machine.

While I see that the tests are fast and don't depend on a GPU, I don't approve of the precedent this is trying to set or the integration point here in the workflows. We have other subprojects with similar CPU unit tests that wait until dedicated test jobs to run. Consistency across the superproject is quite important here.

.github/workflows/build_portable_linux_artifacts.yml
ScottTodd, line 1: These non-multi-arch workflow files are scheduled to be deleted (https://github.com/ROCm/TheRock/issues/3340). No new feature work in them please.

build_tools/configure_stage.py
```diff
@@ -161,6 +161,9 @@ def generate_cmake_args(
     for feature in sorted(features):
         args.append(f"-DTHEROCK_ENABLE_{feature}=ON")
 
+    if stage_name == "compiler-runtime":
+        args.append("-DTHEROCK_BUILD_COMGR_TESTS=ON")
```
ScottTodd, line 165: This script is only for CI build stage configuration - generic options only, nothing specific to individual subprojects.

Options like this should go in the subproject build files like `-DBUILD_TEST=${THEROCK_BUILD_TESTING}` here:

https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/math-libs/CMakeLists.txt#L10-L18

compiler/pre_hook_amd-llvm.cmake
```diff
@@ -62,47 +62,65 @@ endif()
...
+"%SCRIPT_DIR%llvm-lit.real.py" %*
+exit /b %errorlevel%
+]=])
+    install(CODE "
+      file(RENAME \"\${CMAKE_INSTALL_PREFIX}/bin/llvm-lit.py\" \"\${CMAKE_INSTALL_PREFIX}/bin/llvm-lit.real.py\")
+      file(COPY \"${CMAKE_CURRENT_BINARY_DIR}/llvm-lit-wrapper.cmd\" DESTINATION \"\${CMAKE_INSTALL_PREFIX}/bin\")
+      file(RENAME \"\${CMAKE_INSTALL_PREFIX}/bin/llvm-lit-wrapper.cmd\" \"\${CMAKE_INSTALL_PREFIX}/bin/llvm-lit.cmd\")
+    ")
+  else()
+    install(PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/bin/llvm-lit" DESTINATION bin)
```
ScottTodd: This is pretty complicated code for a pre_hook. Could more of this code more into LLVM itself? (our ROCm fork or upstream)

ScottTodd (changes requested):

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kasaurov: --- Which source files specifically?
By the nature of LIT unit tests they depend on the source tree and usually are run at the end of the build (when the sources are still available).
So basically we need the source tree (and LLVM one is quite large) or some kind of its stripped version - that was one of the main reasons to run it along with the build process - we don't have to copy sources then, which reduce space and time.
Another reason - these unit tests don't depend on GPU, so they can be run on a general runner to not occupy precious GPU time. Even though it takes little time to actually run the tests, the time will be spent on setting up the runner and copying/unpacking huge source artifacts.
Do we have CPU runners?

.github/workflows/build_portable_linux_artifacts.yml
kasaurov: We are running two sets of unit tests and the resulting exit code for the Step depends on those two runs. It doesn't matter for non-blocking run, except if we want to introduce some kind of diagnostics based on the Step exit code. But in the future if we want to switch to a blocking step this exit code will matter.

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
ScottTodd: > By the nature of LIT unit tests they depend on the source tree and usually are run at the end of the build (when the sources are still available).

We do have one related example in TheRock already: libhipcxx (https://github.com/ROCm/libhipcxx/) puts a source tree into the "test" artifact and uses that to run lit tests:
* https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/math-libs/artifact-libhipcxx.toml#L10-L18
* https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/build_tools/github_actions/test_executable_scripts/test_libhipcxx_hipcc.py#L93-L99
* Sample artifacts from a CI run: https://therock-ci-artifacts.s3.amazonaws.com/24020364898-linux/index.html , [libhipcxx_test_gfx950-dcgpu.tar.zst](https://therock-ci-artifacts.s3.amazonaws.com/24020364898-linux/libhipcxx_test_gfx950-dcgpu.tar.zst) is 1.66MB compressed, 16MB extracted, 3576 files.

I know that the full LLVM is 1-2 orders of magnitude larger, but how much of that is actually needed for just the comgr unit tests?

---

> Do we have CPU runners?

We currently only have a single size of self-hosted CPU runners for Linux and Windows, so the options are:
* GitHub-hosted runners
* Large self-hosted runners (sized for source builds, so 48+ cores, etc.)

There are a few other unit test jobs that would benefit from being self-hosted (with our credentials to upload logs/artifacts to our AWS buckets and being on the same network / cloud provider as the buckets and other build machines) though, so I expect we'll add a middle tier in the future.

kasaurov: > I know that the full LLVM is 1-2 orders of magnitude larger, but how much of that is actually needed for just the comgr unit tests?

The Comgr unit tests is actually the first step - the next step is to execute LIT tests for the whole LLVM project (see my draft https://github.com/ROCm/TheRock/pull/4326 ). I will investigate what is going to be needed for Comgr and LLVM testing.
And thanks for pointing to libhipcxx example.

.github/workflows/build_portable_linux_artifacts.yml
kasaurov: I see, reverted the changes in those files by https://github.com/ROCm/TheRock/pull/4259/changes/48794fa236f6fdc9b521ee818f0c2e72d46778f3

compiler/pre_hook_amd-llvm.cmake
kasaurov: Agreed, I simplified it to focus only on in-tree run by https://github.com/ROCm/TheRock/pull/4259/changes/a9e97aee0544a2493ebd25bda7e52d9dfbc65596

kasaurov: Just FYI @stellaraccident

ScottTodd:
We discussed offline, summarizing some of the conclusions:
* Adding CPU unit tests for components like LLVM in the build jobs is fine (within reason)
* We don't need to bundle test deps in the artifacts for this use case
* There were `compiler/pre_hook_amd-llvm.cmake` changes that adjusted the install structure, that ideally shouldn't be needed to run tests from the existing build directory
* We can add a `amd-llvm-tests` style cmake subproject to TheRock that has a CMake target to run the unit tests. That would allow for overlapping execution of tests while other subproject builds like `hip-tests` run in parallel, saving critical time on CI
* Other subprojects will likely also want to run unit tests in a similar same way, and there is some precedent for running other unit tests as part of build jobs, like TheRock's own `ctest` tests for packaging (which we're currently missing from multi-arch CI, see https://github.com/ROCm/TheRock/issues/4640)

kasaurov:
The PR is closed, the work was moved to https://github.com/ROCm/TheRock/pull/4881

PR #4260, opened 2026-03-31
[debug-tools/rocr-debug-agent] Support both old and new locations for rocm-debug-agent test files

lumachad:
The rocm-debug-agent test binary and run-test.py script have moved from src/rocm-debug-agent-test/ to tests/rocm-debug-agent/ in the artifacts directory.

Update the test harness to check the old location first for backward compatibility, then fall back to the new location where both files now reside in the same directory.

Also update the rocr-debug-agent-tests artifact locations and add the required RPATH entries to the cmake recipe.

As a transitional measure, we put in place a post hook for rocr-debug-agent-tests that adds RPATH entries for both the old and new location of the test executable. This guarantees the test executable will always find its required libraries.

### DIFF
--- a/debug-tools/post_hook_rocr-debug-agent-tests.cmake
+++ b/debug-tools/post_hook_rocr-debug-agent-tests.cmake
@@ -9,7 +9,7 @@
 # This post hook will be removed when the transition is done.
 if(TARGET rocm-debug-agent-test)
   # Use cmake_language(DEFER) to set RPATH after therock_global_post_subproject.cmake runs.
-  # This ensures our custom RPATH isn't overwritten by the automatic processing.
+  # This ensures our custom RPATH isn't overwritten by the automated processing.
   cmake_language(DEFER CALL set_target_properties rocm-debug-agent-test PROPERTIES
     INSTALL_RPATH "$ORIGIN/../lib/rocm_sysdeps/lib:$ORIGIN/../lib/llvm/lib:$ORIGIN/../lib:$ORIGIN/../../lib/rocm_sysdeps/lib:$ORIGIN/../../lib/llvm/lib:$ORIGIN/../../lib"
   )


lumachad:
This is a required change before we merge https://github.com/ROCm/rocm-systems/pull/3527. Otherwise the test executable location change will break CI.

lumachad:
@marbre This PR should allow TheRock to handle both old and new debug agent gracefully build and test wise. Once we merge and pickup https://github.com/ROCm/rocm-systems/pull/3527, we can discard the post hook and the old artifact locations.

lumachad:
Expected rocgdb failure due to compiler bump regression.

build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py
```diff
@@ -135,9 +135,38 @@ def get_default_paths() -> Dict[str, Path]:
...
 
+    # Try the old testing script location first (for backwards compatibility).
+    old_test_bin = therock_bin_dir / "rocm-debug-agent-test"
+    old_test_script = artifacts_dir / "src" / "rocm-debug-agent-test" / "run-test.py"
+
+    if old_test_script.exists():
+        logger.info(f"Using test script from old location: {old_test_script}")
+        logger.info(f"Using test binary from old location: {old_test_bin}")
+        test_bin = old_test_bin
+        test_script = old_test_script
```
marbre:
```suggestion
        test_bin = therock_bin_dir / "rocm-debug-agent-test"
        test_script = artifacts_dir / "src" / "rocm-debug-agent-test" / "run-test.py"
```

build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py
```diff
@@ -135,9 +135,38 @@ def get_default_paths() -> Dict[str, Path]:
...
+    else:
+        # Fall back to new the testing script location (both binary and script
+        # in the same directory).
+        new_test_dir = artifacts_dir / "tests" / "rocm-debug-agent"
+        new_test_bin = new_test_dir / "rocm-debug-agent-test"
+        new_test_script = new_test_dir / "run-test.py"
```
marbre:
```suggestion
        test_dir = artifacts_dir / "tests" / "rocm-debug-agent"
        test_bin = test_dir / "rocm-debug-agent-test"
        test_script = test_dir / "run-test.py"
```

build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py
```diff
@@ -135,9 +135,38 @@ def get_default_paths() -> Dict[str, Path]:
...
+
+        if new_test_script.exists():
```
marbre:
```suggestion
        if not test_script.exists():
```
You could invert the condition here and always fail if `test_script` doesn't exist.

lumachad: Thanks. That was some roundabout way of doing things. I've attempted to simplify it some with the extra commit. Thoughts?

lumachad:
Still needs a few more changes to remove duplicate data.

lumachad:
Should be good for a re-review now.

debug-tools/post_hook_rocr-debug-agent-tests.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+# The rocm-debug-agent-test executable is transitioning from bin/ to
+# tests/rocm-debug-agent-tests/bin/. For compatibility with both locations
+# during the transition, we set RUNPATH entries for both old (bin/) and
+# new (tests/rocm-debug-agent-tests/) directory structures.
+#
+# This post hook will be removed when the transition is done.
+if(TARGET rocm-debug-agent-test)
+  # Use cmake_language(DEFER) to set RPATH after therock_global_post_subproject.cmake runs.
+  # This ensures our custom RPATH isn't overwritten by the automatic processing.
```
marbre:
```suggestion
  # This ensures our custom RPATH isn't overwritten by the automated processing.
```
?

lumachad:
> Not sure about `debug-tools/post_hook_rocr-debug-agent-tests.cmake` but otherwise looks good to me, assuming unit tests on the artifact bundling don't fail: We don't allow to include files in multiple artifacts but I think this isn't an issue here and is rather we have old and new test artifacts (but in a single artifact tarball).

The post hook will vanish as soon as we merge https://github.com/ROCm/rocm-systems/pull/3527. Its purpose is to make sure we don't end up with testing failures while we do the rocr-debug-agent build system update.

lumachad:
Thanks. Merging.

PR #4265, opened 2026-03-31
Make artifact archives reproducible

PeterCDMcLean:
## Motivation

Part of #4202.

Artifacts uploaded to S3 should never silently overwrite a different artifact at the same key. That
check has to be SHA-based, which only works if identical content reliably produces an identical
archive — and today it does not. `tar` records the build time and the building user's uid/gid, and
members were added in filesystem iteration order, so the same artifact built twice got two different
SHA256s. #4202 cannot distinguish "same artifact, rebuilt" from "different content, same key" until
that noise is gone.

## Technical Details

`normalize_tarinfo()` replaces the metadata that varies per build (`mtime`, `uid`/`gid`,
`uname`/`gname`), and members are added in sorted order. Permissions, sizes, symlink targets and
hardlink identity are untouched. Extracting as a normal user still yields files owned by the
extracting user, since `tar` ignores archived ownership without root.

`mtime` is **replaced, not zeroed** — see the timestamp section below.

Three things changed since this PR was first opened, all from @ScottTodd's review:

- **`archive_util.py` now exists.** He floated it in review and then landed it in #5027, so
  `normalize_tarinfo()` goes there rather than staying private to `fileset_tool.py`. That is also
  what the rebase required: `fileset_tool.py` no longer imports `tarfile` at all now that
  `_open_archive()` has been replaced by `open_archive_for_write()`.
- **`py_packaging.py` is included.** He pointed at the `_devel` tarball built for the
  `rocm-sdk-devel` wheel as having the same problem. It does — `os.walk` order and build-time
  metadata leak into it — so the reproducible tree walk is extracted as `archive_util.add_tree()`
  and both callers share one tested implementation instead of repeating the filter/sort wiring.
- **`mtime = 0` is gone.** His
  [analysis](https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_4265.md#-blocking-epoch-mtimes-can-suppress-downstream-rebuilds-after-sdk-upgrades)
  is correct — details below.

Rebased from the original base (`0856406fe`, ~660 commits back) onto current `main`.

### Timestamps

Confirmed empirically that **mtime is restored on extraction**, so this was reaching installed SDKs:

| extraction path | mtime restored? |
|---|---|
| `extractall(filter="tar")` — `artifact_manager.py:434`, `fetch_artifacts.py:203`, `install_rocm_from_artifacts.py:303` | yes |
| `extractall(filter="data")` | yes |
| `tf.extract(ti, ...)` — the devel wheel extractor, `_devel.py:490-498` | yes |
| `tar -xf` as non-root, no flags | yes |

It's *ownership* a non-root extractor drops, not mtime — the original description conflated the two.
It propagates further too: `pattern_match.py` copies with `os.link` (same inode) or `shutil.copy2`,
both of which preserve mtime, so an epoch timestamp entering a stage dir survives into `dist/` and
into whatever is packaged from it.

`get_archive_timestamp()` now resolves, in order:

1. **`SOURCE_DATE_EPOCH`** — the reproducible-builds convention, and the supported override. Set it
   to pin the value explicitly.
2. **The commit time of TheRock's `HEAD`.** TheRock pins every submodule, so its HEAD identifies the
   whole source state. An artifact has no single source submodule to ask: `artifact_subprojects.json`
   maps `sysdeps` to 13 subprojects and `base` to 6, a merged `dist/` tree spans a subproject plus
   its runtime deps, and several subprojects (`zlib`, `zstd`, `bzip2`, `elfutils`) are tarballs
   downloaded from S3 with no git history at all.
3. **The current time**, when neither is available (no git, or an exported source archive). Not
   reproducible, but a working build beats a deterministically broken one.

Parallel jobs building the same commit resolve the same value, so the property #4202 needs is
unchanged.

Two consequences worth deciding on rather than inheriting, both of which I'd rather settle in review
than assume:

- **Over-invalidation.** A docs-only commit here advances the timestamp for every artifact, so every
  archive hash changes even when the content is byte-identical. Harmless for conflict detection
  within a run, but it forecloses using archive hashes to dedupe across commits.
- **Dirty trees** still report the last commit time, which understates a developer's local source
  state. Follow-up — @PeterCDMcLean and I discussed falling back to submodule / working-copy dates
  when the tree is dirty; landing the simple version first so the shape can be reviewed.

This does not by itself give Scott's rebuild-correctness invariant ("a newly installed SDK file is
newer than anything built against the old one"). A commit date can still predate a user's earlier
build — release branches cut weeks before publication, or a user moving from a newer nightly to an
older stable. Guaranteeing that invariant means re-stamping at install time, which is a separate
decision about what each distribution channel should do; this PR just stops the archiver from being
actively wrong.

## Risk Assessment

**Risk level 2.** Metadata-only change to how archives are written; no content, permissions or
layout changes, and readers are unaffected. Blast radius is every artifact archive and the devel
wheel tarball, so it is broad, but the change is uniform and directly covered by tests. The
user-visible differences are that archived files report `root:root` (which only matters when
extracting as root with `-p`) and that every file in an archive now carries the same mtime — the
source commit time — rather than its individual build time.

## Related

- Work tracking: #4202 (SHA-conflict detection on artifact upload). This PR is a prerequisite, not
  the fix — leave #4202 open.
- Builds on: #5027 (`Consolidate zstd/xz archive helpers into archive_util.py`).

None required. Host-side packaging code with no device or architecture dependence; passing PR CI is
sufficient.

## Test Plan

New coverage, all in the existing `pytest build_tools/` lane (Linux + Windows):

`build_tools/tests/fileset_tool_test.py` — end-to-end through the real `fileset_tool artifact-archive`
command, the shape @ScottTodd sketched in review:

- `testArchiveReproducibleDespiteMetadataChanges` — archive the same artifact twice with the file
  mtimes changed in between; the two SHA256s must match.
- `testArchiveMemberMetadataIsNormalized` — with `SOURCE_DATE_EPOCH` set, every member carries that
  mtime, `uid/gid 0` and `root:root`; the manifest is stored first and the remaining members are
  sorted.
- `testArchiveMtimeIsNotTheEpoch` — with nothing configured, every member's mtime is > 0. This is
  the guard against regressing to the behavior Scott flagged.

`build_tools/tests/archive_util_test.py` — unit coverage for the shared helpers:

- `ArchiveTimestampTest` — `SOURCE_DATE_EPOCH` is honored; non-integer and negative values are
  rejected; falls back to the git commit time, and to the current time outside a checkout; and
  resolves to something > 0 in a real checkout with nothing configured.
- `NormalizeTarinfoTest` — the varying metadata is replaced, mode/size/type/name and symlink targets
  survive, and the mtime is never the epoch.
- `AddTreeTest` — member order is pinned, metadata is normalized, and the same tree archives to
  identical bytes after an mtime change.

Both new files' fixtures create entries out of sorted order deliberately, so a passing ordering
assertion cannot be explained by creation order.

## Testing Checklist

- [x] New tests, production change reverted - `git checkout origin/main -- build_tools/fileset_tool.py && pytest tests/fileset_tool_test.py` - Status: Failed as expected (both new tests; digests differed `040a3058…` vs `eb448499…`)
- [x] `add_tree` mutation check (drop the sort + filter) - `pytest tests/archive_util_test.py` - Status: Failed as expected (all 3 `AddTreeTest` cases)
- [x] Epoch-mtime mutation check (set `tarinfo.mtime = 0` again) - `pytest tests/archive_util_test.py tests/fileset_tool_test.py` - Status: Failed as expected (5 cases across both files)
- [x] Extraction behavior verified by experiment (the table above) - Status: mtime restored on all four paths
- [x] New tests, change applied - `cd build_tools && pytest tests/fileset_tool_test.py tests/archive_util_test.py` - Status: Passed
- [x] Full unit-test lane, locally (Linux) - `cd build_tools && pytest` - Status: Passed (2089 passed, 8 skipped)
- [x] Each commit independently - `pytest` at each of the three commits - Status: Passed (2077, 2081, 2089)
- [x] pre-commit - `pre-commit run --files <changed>` - Status: Passed

## Test Result

Both new `fileset_tool` tests fail on unpatched `main` and pass with the change; the reproducibility
test fails with two different digests for byte-identical content, which is exactly the condition
#4202 needs to be able to rely on. All three `AddTreeTest` cases fail when the sort and filter are
removed from `add_tree()`. Full suite green, both commits green in isolation.

### Earlier empirical result, and what it does and does not show

An earlier revision of this branch was run through CI twice with no changes, and the artifacts were
compared file by file
([artifact_comparison_results.txt](https://github.com/user-attachments/files/27369626/artifact_comparison_results.txt)).
57 artifacts still differed between the two runs — `blas_*`, `rccl_lib_*`, `rocprofiler-systems_*`
and others.

That is worth being explicit about: **normalizing archive metadata does not by itself make those
artifacts reproducible.** It removes the archiver as a source of nondeterminism, so any remaining
difference is real content difference produced by the build. Those are separate bugs and are not in
scope here. The value of this PR is that it makes the residue attributable.

## Flags / Guardrails

None. The change is unconditional. It is backward compatible for readers: existing archives still
extract identically, and nothing reads the normalized fields.

## Adjacent Tests Considered

- `py_packaging_test.py` — run, passing. It does not reach `populate_devel_files()`, which needs a
  full artifact catalog fixture, which is why `add_tree()` was extracted and tested directly rather
  than testing the wheel path end to end.
- `artifacts_test.py`, `artifact_manager_tool_test.py`, `build_tarballs_test.py` — the other archive
  readers/writers; run, passing. Readers are unaffected because only metadata changed.
- `install_rocm_from_artifacts_test.py` — exercises extraction; run, passing.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/archive_util.py
+++ b/build_tools/_therock_utils/archive_util.py
@@ -7,46 +7,26 @@ from typing import Callable
 import functools
 import os
 from pathlib import Path
-import subprocess
 import tarfile
 import time
 
-THEROCK_DIR = Path(__file__).resolve().parent.parent.parent
+from .source_date import ENV_VAR, STANDARD_ENV_VAR
 
 
-def _git_commit_timestamp() -> int | None:
-    """Commit time of TheRock's HEAD, or None if not a usable git checkout.
-
-    Deliberately the superproject rather than the submodule an artifact's
-    content came from. TheRock pins every submodule, so its HEAD identifies the
-    whole source state, whereas an artifact has no single source submodule to
-    ask: a merged dist tree spans a subproject and its runtime deps, and several
-    subprojects (zlib, zstd, bzip2, elfutils, ...) are tarballs downloaded from
-    S3 with no git history at all.
-
-    The tradeoff is over-invalidation: a docs-only commit here changes every
-    artifact's hash even when the content is byte-identical. That is fine for
-    detecting conflicting uploads within a run, which is what this is for, but
-    it does mean archive hashes cannot be used to dedupe across commits. Set
-    `SOURCE_DATE_EPOCH` to decide that policy somewhere better informed.
-    """
+def _read_timestamp_env(name: str) -> int | None:
+    value = os.environ.get(name)
+    if not value:
+        return None
     try:
-        result = subprocess.run(
-            ["git", "log", "-1", "--format=%ct"],
-            cwd=THEROCK_DIR,
-            capture_output=True,
-            text=True,
-            check=False,
+        timestamp = int(value)
+    except ValueError:
+        raise ValueError(
+            f"{name} must be an integer number of seconds since the Unix "
+            f"epoch, got {value!r}"
         )
-    except OSError:
-        # git is not installed.
-        return None
-    if result.returncode != 0:
-        # Not a git checkout (e.g. building from an exported source archive) or
-        # no commits yet.
-        return None
-    output = result.stdout.strip()
-    return int(output) if output else None
+    if timestamp < 0:
+        raise ValueError(f"{name} must not be negative, got {timestamp}")
+    return timestamp
 
 
 @functools.cache
@@ -61,36 +41,23 @@ def get_archive_timestamp() -> int:
 
     Resolution order:
 
-    1. `SOURCE_DATE_EPOCH`, the reproducible-builds convention. Set this to pin
-       the timestamp explicitly; it is the supported override.
-    2. The commit time of HEAD. Deterministic for a given revision, so parallel
-       CI jobs building the same commit still produce identical archives, and
-       it advances as the source advances.
-    3. The current time, if neither is available. Not reproducible, but a
-       working build beats a deterministically broken one -- set
-       `SOURCE_DATE_EPOCH` where reproducibility is required.
+    1. `SOURCE_DATE_EPOCH`. Nothing in TheRock sets this unless explicitly asked
+       to, so its presence means somebody deliberately pinned the whole build's
+       timestamp, and that intent wins.
+    2. `THEROCK_SOURCE_DATE_EPOCH`, which orchestrators set for their workers
+       from `source_date.compute_source_date_epoch()`. This is the normal path
+       for a TheRock build.
+    3. The current time, for a worker run directly with neither set. Not
+       reproducible, but a working build beats a deterministically broken one.
 
-    Note that a dirty working tree still reports its last commit time. That is
-    fine for CI, which is always clean; set `SOURCE_DATE_EPOCH` locally if a
-    dirty tree's archives need to look newer than something.
+    Deliberately does no git work of its own: resolving the source timestamp
+    costs several git invocations, and this runs once per archive process.
+    `source_date` owns that, and orchestrators call it once.
     """
-    env_value = os.environ.get("SOURCE_DATE_EPOCH")
-    if env_value:
-        try:
-            timestamp = int(env_value)
-        except ValueError:
-            raise ValueError(
-                "SOURCE_DATE_EPOCH must be an integer number of seconds since "
-                f"the Unix epoch, got {env_value!r}"
-            )
-        if timestamp < 0:
-            raise ValueError(f"SOURCE_DATE_EPOCH must not be negative, got {timestamp}")
-        return timestamp
-
-    commit_timestamp = _git_commit_timestamp()
-    if commit_timestamp is not None:
-        return commit_timestamp
-
+    for name in (STANDARD_ENV_VAR, ENV_VAR):
+        timestamp = _read_timestamp_env(name)
+        if timestamp is not None:
+            return timestamp
     return int(time.time())
 
 
--- a/build_tools/_therock_utils/archive_util.py
+++ b/build_tools/_therock_utils/archive_util.py
@@ -3,6 +3,8 @@
 
 """Utilities for reading and writing zstd/xz compressed tar archives."""
 
+from typing import Callable
+import os
 from pathlib import Path
 import tarfile
 
@@ -23,6 +25,36 @@ def normalize_tarinfo(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo:
     tarinfo.uname = "root"
     tarinfo.gname = "root"
     return tarinfo
+
+
+def add_tree(
+    tf: tarfile.TarFile,
+    source_dir: Path,
+    *,
+    relative_to: Path,
+    on_add: Callable[[str], None] | None = None,
+) -> None:
+    """Adds every file and directory under `source_dir` to `tf` reproducibly.
+
+    Member names are computed relative to `relative_to`. Entries are walked and
+    emitted in sorted order and their metadata is normalized, so the same tree
+    always produces the same archive bytes. `on_add`, if given, is called with
+    each member name as it is added.
+    """
+    for root, dirnames, filenames in sorted(
+        os.walk(source_dir), key=lambda entry: entry[0]
+    ):
+        for name in sorted(list(filenames) + list(dirnames)):
+            file_path = os.path.join(root, name)
+            arcname = os.path.relpath(file_path, relative_to)
+            if on_add is not None:
+                on_add(arcname)
+            tf.add(
+                file_path,
+                arcname=arcname,
+                recursive=False,
+                filter=normalize_tarinfo,
+            )
 
 
 def _get_pyzstd():
--- a/build_tools/_therock_utils/archive_util.py
+++ b/build_tools/_therock_utils/archive_util.py
@@ -7,6 +7,24 @@ from pathlib import Path
 import tarfile
 
 
+def normalize_tarinfo(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo:
+    """Normalizes TarInfo metadata so identical content yields identical archives.
+
+    Pass as the `filter` argument to `TarFile.add()`. Timestamps and the uid/gid
+    of the building user otherwise vary from build to build, giving the same
+    content a different archive hash every time.
+
+    Permissions are preserved. Extracting as a non-root user gives files owned by
+    the extracting user; extracting as root with `-p` gives root:root.
+    """
+    tarinfo.mtime = 0
+    tarinfo.uid = 0
+    tarinfo.gid = 0
+    tarinfo.uname = "root"
+    tarinfo.gname = "root"
+    return tarinfo
+
+
 def _get_pyzstd():
     """Lazy import pyzstd with helpful error message."""
     try:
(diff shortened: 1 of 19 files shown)


PeterCDMcLean:
Full artifact file-by-file comparison of artifacts and problematic files:

[artifact_comparison_results.txt](https://github.com/user-attachments/files/27369626/artifact_comparison_results.txt)

build_tools/fileset_tool.py
```diff
@@ -78,6 +78,22 @@ def do_artifact(args):
         contents.write_artifact(output_dir)
 
 
+def _normalize_tarinfo(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo:
+    """Normalize TarInfo for reproducible archives.
```
ScottTodd: FYI I'm considering setting up a `archive_util.py` which will replace the `_open_archive()` function that is currently inlined into this file with a shared implementation where we can centralize special case handling in https://github.com/ROCm/TheRock/pull/5027. We could also move this `_normalize_tarinfo()` function into that file if we want other scripts like this one to do the same thing: https://github.com/ROCm/TheRock/blob/a16278fe90a2dd0a56a6492e44d52de0f7d667ed/build_tools/_therock_utils/py_packaging.py#L550-L556

(or we could port some of these changes over to that file directly too)

build_tools/fileset_tool.py
```diff
@@ -78,6 +78,22 @@ def do_artifact(args):
...
+    Sets fixed timestamp, uid/gid, and uname/gname for deterministic output.
+    File permissions are preserved. When extracted as a non-root user, ownership
+    defaults to the extracting user. When extracted as root with -p, files are
+    owned by root:root.
+    """
+    tarinfo.mtime = 0
+    tarinfo.uid = 0
+    tarinfo.gid = 0
+    tarinfo.uname = "root"
+    tarinfo.gname = "root"
```
ScottTodd: This seems worth adding new unit test coverage for.

Something of this shape?
```python
  def testArchiveReproducibleDespiteMetadataChanges(self):
      """Archives are identical even when file timestamps differ."""
      input_dir = self.temp_dir / "input"
      descriptor_file = self.temp_dir / "artifact.toml"
      artifact_dir = self.temp_dir / "artifact"
      write_text(descriptor_file, ARTIFACT_DESCRIPTOR_1)

      # Create multiple files (exercises sorting too)
      write_text(input_dir / "example" / "stage" / "share" / "doc" / "z_file.txt", "z")
      write_text(input_dir / "example" / "stage" / "share" / "doc" / "a_file.txt", "a")
      write_text(input_dir / "example" / "stage" / "lib" / "libfoo.so.1", "foo")

      # Build the artifact directory once
      run_command([
          sys.executable, FILESET_TOOL, "artifact",
          "--descriptor", descriptor_file,
          "--artifact-name", "test",
          "--root-dir", input_dir,
          "doc", artifact_dir,
      ])

      # Archive with current timestamps
      archive1 = self.temp_dir / "archive1.tar.xz"
      run_command([
          sys.executable, FILESET_TOOL, "artifact-archive",
          artifact_dir, "-o", archive1,
      ])

      # Simulate a different build time: change mtimes on all files
      for p in artifact_dir.rglob("*"):
          if p.is_file():
              os.utime(p, (1700000000, 1700000000))

      # Archive again with different timestamps
      archive2 = self.temp_dir / "archive2.tar.xz"
      run_command([
          sys.executable, FILESET_TOOL, "artifact-archive",
          artifact_dir, "-o", archive2,
      ])

      hash1 = calculate_hash(archive1, "sha256").hexdigest()
      hash2 = calculate_hash(archive2, "sha256").hexdigest()
      self.assertEqual(hash1, hash2)
```

ScottTodd (dismissed):

ScottTodd:
This is similar in some ways to https://github.com/ROCm/TheRock/pull/7632, so requested from @araravik-psd too

build_tools/_therock_utils/archive_util.py
```diff
@@ -3,10 +3,60 @@
...
+
+    Permissions are preserved. Extracting as a non-root user gives files owned by
+    the extracting user; extracting as root with `-p` gives root:root.
+    """
+    tarinfo.mtime = 0
+    tarinfo.uid = 0
+    tarinfo.gid = 0
+    tarinfo.uname = "root"
+    tarinfo.gname = "root"
+    return tarinfo
```
ScottTodd, line 81: Setting `tarinfo.mtime = 0` on all tar file contents is likely to have unintended side effects for distributed packages (tarball/python/deb/rpm). These packages includes files used by downstream projects to build from source, so any tools that check file timestamps (e.g. ninja) would see input files having older creation dates than they are expecting.

See https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_4265.md#-blocking-epoch-mtimes-can-suppress-downstream-rebuilds-after-sdk-upgrades for a detailed analysis. I think we need to decide on what properties we want at each stage of CI/CD and make some more conscious decisions in the code on whether we want to preserve, strip, or reintroduce metadata at each point.

PeterCDMcLean: We could use the last-modified git commit date, or in a dirty copy the last modified working copy file. That seems like a pretty stable timestamp to use. Thoughts?

I'll see if I can figure out a way to thread that info into the archive_util from the various callpoints.

PeterCDMcLean: working on this, scope is growing considerably 🤯

is control of release package file/dir times a desirable goal?

PeterCDMcLean: Please take a look and advise direct. Made a couple calls but you have a better idea where you want this to head.

build_tools/_therock_utils/source_date.py
```diff
@@ -0,0 +1,506 @@
...
+            paths.extend(_parse_porcelain_z(submodule_output, entry.path))
+    return paths
+
+
+def is_worktree_dirty(
+    repo_dir: Path = THEROCK_DIR,
+    entries: list[SubmoduleEntry] | None = None,
+) -> bool:
+    """Whether anything is uncommitted, in the superproject or any submodule."""
+    return bool(dirty_paths(repo_dir, entries))
```
ScottTodd, line 183: This is returning `true` on the CI run on this PR?
* https://github.com/ROCm/TheRock/actions/runs/33255830639?pr=4265
* https://therock-ci-artifacts.s3.amazonaws.com/33255830639-linux/manifests/therock_manifest.json

I'm not sure how much we'll be able to trust this "is worktree dirty" check for certain CI builds, particularly when TheRock CI is used across repositories, as we do a checkout of TheRock _then_ a checkout of a submodule: https://github.com/ROCm/TheRock/blob/407ca8023c39655487071c96ab000e62d1fce804/.github/workflows/multi_arch_build_portable_linux_artifacts.yml#L119-L133

build_tools/_therock_utils/archive_util.py
ScottTodd: I don't really have enough opinions here, other than knowing that there are lot of edge cases to handle and this feels risky if we do anything other than what the tools are built for by default :/

PR #4275, opened 2026-03-31
[docs] Adding an overarching CI doc page

geomin12:
Lots of developers have asked questions about TheRock CI, how it works, how it differs from math-ci. This document will help the general idea of CI system with links to documentation

This is the first iteration, with many more to come as features are added.

### DIFF
--- a/docs/development/ci_overview.md
+++ b/docs/development/ci_overview.md
@@ -18,10 +18,10 @@ TheRock uses a multi-stage CI pipeline that splits the build into stages (founda
 
 ```mermaid
 graph TD
-    A[generic build] --> B1[arch build \(gfx94X-dcgpu\)]
-    A[generic build] --> B2[arch build \(gfx110X-dgpu\)]
-    B1 --> C[test \(gfx94X-dcgpu\)]
-    B2 --> D[test \(gfx110X-dgpu\)]
+    A[generic build] --> B1[arch build - gfx94X-dcgpu]
+    A[generic build] --> B2[arch build - gfx110X-dgpu]
+    B1 --> C[test - gfx94X-dcgpu]
+    B2 --> D[test - gfx110X-dgpu]
 ```
 
 Each stage runs as a separate job, uploads its artifacts and logs to S3, then downstream stages download and build on top of them. This allows for:


geomin12:
multi-arch CI will be fixed with #4276

amd-justchen (approved):
Glad to have this overview to reference! The mermaid diagrams are very helpful!

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+| **Test Execution**      | Jenkins agents               | GitHub-hosted or self-hosted runners                |
+| **Logs**                | Jenkins UI                   | S3 (see [workflow_outputs.md](workflow_outputs.md)) |
+
+**Migration workflow:**
+
+1. Create test script in `build_tools/github_actions/test_executable_scripts/`
+1. Add entry to `fetch_test_configurations.py`
+1. Ensure artifact dependencies are configured in `install_rocm_from_artifacts.py`
+
+See [adding_tests.md](adding_tests.md) for step-by-step instructions.
```
ScottTodd: I'm not sold on this "MathCI Migration Guide" section.

Comparing two different CI systems is a pretty broad topic that this just scratches the surface of. I'm not sure there's enough information here to be useful to developers, or at least this could be substantially more useful with more concrete details. This sort of content may be easier to put in an internal wiki that can link to the actual sources and include actual screenshots for the groovy scripts, Jenkins UI, etc.

Also, when referencing source files, please include links where possible, to help developers find the files.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+## Quick Summary
+
+TheRock CI follows this workflow:
+
+1. **Build** ROCm from source via TheRock
+1. **Upload** build artifacts to S3 (public-read buckets)
+1. **Test** workflows download those artifacts from S3
+1. **Run** tests against the downloaded artifacts
+
+Instead of Jenkins and Groovy pipelines, TheRock uses **GitHub Actions** workflows defined in YAML files under `.github/workflows/`.
```
ScottTodd, line 13: RE: my other comment about the "MathCI Migration Guide", this page would stand on its own better if it led with "TheRock uses GitHub Actions" and then followed with "not other CI systems such as Jenkins" (we could also mention Buildkite, CircleCI, and other CI providers too...). The way this reads right now, it makes it seem like we only considered Jenkins / Groovy.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+This document provides an overview of how Continuous Integration (CI) works in TheRock, including the build → artifact → test pipeline. If you're migrating from MathCI or Jenkins-based workflows, this guide will help you understand TheRock's GitHub Actions-based approach.
+
+## Quick Summary
+
+TheRock CI follows this workflow:
+
+1. **Build** ROCm from source via TheRock
+1. **Upload** build artifacts to S3 (public-read buckets)
+1. **Test** workflows download those artifacts from S3
+1. **Run** tests against the downloaded artifacts
```
ScottTodd: The "Test" and "Run" steps here are overlapping. Merge? Or rename "Test" to "Download"?

Could also qualify with "CPU machine" and "GPU machine" to highlight why there is an upload/download gap

```suggestion
1. **Build** ROCm from source via TheRock on CPU machines
1. **Upload** build artifacts to S3 (public-read buckets)
1. **Download** artifacts from S3 on GPU machines for testing
1. **Run** tests against the downloaded artifacts
```

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+    B --> C1[Math-Libs gfx1151]
+    B --> C2[Math-Libs gfx110X-all]
+    B --> C3[Math-Libs gfx94X-dcgpu]
+    C1 --> D1[Test gfx1151]
+    C2 --> D2[Test gfx110X-all]
+    C3 --> D3[Test gfx94X-dcgpu]
+    C1 --> E1[Build Python Packages gfx1151]
+    C2 --> E2[Build Python Packages gfx110X-all]
+    C3 --> E3[Build Python Packages gfx94X-dcgpu]
+```
```
ScottTodd: This diagram is not accurate:

* https://github.com/ROCm/TheRock/pull/3928 will remove the foundation stage
* Tests and python package builds in multi-arch CI are triggered after ALL build stages (compiler-runtime generic, matrix of math-libs, iree-libs generic, etc.), not after individual "math-libs" builds

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+Each stage runs as a separate job, uploads its artifacts and logs to S3, then downstream stages download and build on top of them. This allows for:
+
+- **Parallelization:** Multiple GPU families can build math-libs simultaneously once compiler-runtime completes
+- **Incremental builds:** Test-only runs can skip build stages by downloading pre-built artifacts
+- **Flexibility:** Different stages can run on different runner types (e.g., CPU-only for foundation, GPU for tests)
+
+**Key workflow files:**
+
+- [`.github/workflows/multi_arch_ci.yml`](/.github/workflows/multi_arch_ci.yml) - Main entry point
+- [`.github/workflows/multi_arch_build_portable_linux_artifacts.yml`](/.github/workflows/multi_arch_build_portable_linux_artifacts.yml) - Build orchestration
```
ScottTodd: At a minimum, also link the corresponding Windows workflow when linking a Linux workflow.

I'd also link these (and those for Windows):
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci_linux.yml - build rocm, test rocm, build rocm python, build pytorch
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux.yml - stages for "build rocm"

docs/development/ci_overview.md
jayhawk-commits: Not sure this needs to be called out as two separate steps.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+    C1 --> E1[Build Python Packages gfx1151]
+    C2 --> E2[Build Python Packages gfx110X-all]
+    C3 --> E3[Build Python Packages gfx94X-dcgpu]
+```
+
+Each stage runs as a separate job, uploads its artifacts and logs to S3, then downstream stages download and build on top of them. This allows for:
+
+- **Parallelization:** Multiple GPU families can build math-libs simultaneously once compiler-runtime completes
+- **Incremental builds:** Test-only runs can skip build stages by downloading pre-built artifacts
+- **Flexibility:** Different stages can run on different runner types (e.g., CPU-only for foundation, GPU for tests)
```
jayhawk-commits: Might be worth highlighting which stages run on CPU-only and which ones use GPU on the mermaid diagram above.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+```mermaid
+graph LR
+    A[Configure CMake] --> B[Build with Ninja]
+    B --> C[Package Artifacts]
+    C --> D[Upload to S3]
+```
+
+**What gets built:** Compiler (LLVM), core runtime (HIP, ROCr), math libraries (rocBLAS, rocFFT), ML libraries (MIOpen), media libraries (rocDecode, rocJPEG), and more.
+
+**Artifact organization:** Each component is packaged into separate archives by role (libraries, headers, tests, tools). See [artifacts.md](artifacts.md) for complete details on artifact structure and naming conventions.
```
jayhawk-commits: Not a fan of using the term 'role' twice in this section when one instance is for lib, dev, and test and the other is for libraries, headers, tests, and tools.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+TheRock builds ROCm components from source and produces **artifacts** - archive slices of key components organized by role (lib, dev, test, etc.).
+
+```mermaid
+graph LR
+    A[Configure CMake] --> B[Build with Ninja]
+    B --> C[Package Artifacts]
+    C --> D[Upload to S3]
+```
+
+**What gets built:** Compiler (LLVM), core runtime (HIP, ROCr), math libraries (rocBLAS, rocFFT), ML libraries (MIOpen), media libraries (rocDecode, rocJPEG), and more.
```
ScottTodd: This content here is close to technically correct, but the section is missing motivation/context.

* "Build with ninja" is an overly specific detail
* "Package artifacts" isn't a dedicated step, the build itself generates artifact archives that are uploaded as-is
* The lists of libraries are non-exhaustive "(rocBLas, rocFFT)"... could use "etc."?

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+- `therock-ci-artifacts` - Builds from `ROCm/TheRock`
+- `therock-ci-artifacts-external` - Builds from forks and downstream repos (`rocm-libraries`, `rocm-systems`)
+
+**Release Buckets:**
+
+- `therock-nightly-artifacts` - Nightly builds
+- `therock-dev-artifacts` - Development builds
+- `therock-prerelease-artifacts` - Pre-release builds
+
+See [s3_buckets.md](s3_buckets.md) for the complete bucket list and authentication details for uploads.
```
ScottTodd, line 61: This level of detail is not useful for the average user of the CI system here. The important detail is that files are uploaded to _some_ publicly readable location.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+### Accessing Artifacts
+
+Download artifacts using the `install_rocm_from_artifacts.py` script, which handles fetching from S3 and extracting to the correct locations.
+
+See [installing_artifacts.md](installing_artifacts.md) for detailed instructions on:
+
+- Finding GitHub run IDs
+- Selecting components to download
+- Installing from CI runs vs releases
+- Using with different GPU families
```
ScottTodd, line 72: This script is _not_ what I would personally recommend here, at least not on its own.

* https://github.com/ROCm/TheRock/blob/main/build_tools/artifact_manager.py is used by the build system (handles fetching/pushing artifacts for individual stages)
* https://github.com/ROCm/TheRock/blob/main/build_tools/fetch_artifacts.py (can be used to fetch all artifacts and unpack them)
* https://github.com/ROCm/TheRock/blob/main/build_tools/install_rocm_from_artifacts.py is a complicated wrapper on top of fetch_artifacts.py that does additional filtering optimized for running narrow slices of tests

For build system work and while reviewing PRs, I regularly download artifacts with fetch_artifacts.py or from the S3 index page manually (or via wget). I never use `install_rocm_from_artifacts.py`. I see the utility for developers treating the CI system as a black box and just wanting to reproduce what a CI test runner does though.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+
+```mermaid
+graph LR
+    A[Download Artifacts from S3] --> B[Sanity Check]
+    B --> C[Configure Test Matrix]
+    C --> D1[Test rocBLAS]
+    C --> D2[Test rocFFT]
+    C --> D3[Test MIOpen]
+    C --> D4[Test...]
+```
```
ScottTodd, line 85: This diagram is not accurate? The test matrix is configured ahead of time, then each test job fetches artifacts.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Common CI Tasks
+
+### Trigger a CI Run
+
+**For PRs:** CI runs automatically on every push and when labels are added.
+
+**For specific GPU families:** Add a label to your PR (e.g., `gfx1151-linux`, `gfx94X-windows`)
+
+**Manual trigger:** Use the "Run workflow" button on the [Multi-Arch CI workflow page](https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci.yml)
```
ScottTodd: Link to https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md here, then maybe prune the section to just the essentials.

The `** For PRs::**` formatting across seperate lines is also a strange choice. An unordered list would make more sense here.

docs/development/ci_overview.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Download Artifacts from a CI Run
+
+Use `install_rocm_from_artifacts.py` with a run ID from the GitHub Actions UI.
+
+See [installing_artifacts.md](installing_artifacts.md) for detailed instructions on finding run IDs and selecting components.
```
ScottTodd: This section is duplicated with `### Accessing Artifacts` above. Choose one or the other to keep?

docs/development/ci_overview.md
ScottTodd, line 1: I've been wanting to write similar docs, so thanks for starting this. Here are some notes I had:

---

### GitHub Actions users guide

- artifacts and packages -> workflow_dispatch triggers 
- "each CI run produces ..." which can be used for...
  - artifacts, used for: building packages, running unit tests, "installing" [link to install artifacts page]
  - rocm python packages, used for: building pytorch python packages, "installing"
  - pytorch python packages, used for: running pytorch tests
- additional workflows can be triggered starting from artifacts or packages
- complete dev release workflows can be triggered (full build from source, takes hours)

---

I think focusing the documentation on these aspects would help:

* CI system design/architecture (build artifacts --> test artifacts, build packages from artifacts --> test those packages)
* Interacting with the CI system:
  * How to view logs
  * How to download artifacts locally (for inspection/debugging/running tests/building local packages/etc.)
  * How to trigger parts of the CI pipeline using the artifacts or packages produced by workflow runs

ScottTodd (changes requested):
The outline here is decent, but this has a few inaccuracies and shows signs of AI summarization without sufficient input on the architecture / goals of the document.

docs/development/ci_overview.md
geomin12: we can just keep this doc related to TheRock CI, no need to throw in other older CI systems

docs/development/ci_overview.md
geomin12: I'll keep this pretty general, as things will probably be moving around

ScottTodd: Still not quite right. ALL builds complete before tests are triggered. There are substantially more stages than just "math-libs" too.

Currently build has foundation --> compiler --> [comm libs, dc tools, debug tools, profiler, media libs, math libs] --> fusilli libs, and all of those must complete (across all generic/target-specific stages) before tests start
Image

docs/development/ci_overview.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Common CI Tasks
+
+### Viewing logs
+
+After a CI completes, you will be able to see a GitHub Step Summary in the `Summary` page of the CI run.
+
+<img src="./assets/step_summary.png"/>
+
+You are able to find the index pages for artifacts and logs, as well as additional CI information.
```
ScottTodd: The step summary appears at the start, not at the end. Logs and artifacts appear throughout as they are uploaded.

ScottTodd (commented):
Thanks, this is looking better now.

docs/development/ci_overview.md
ScottTodd: Just keep it simple and general :P the topology is going to keep changing in the next few weeks/months and these docs will go stale if they try to encode too much. All jobs upload artifacts as they complete, there isn't a final "upload artifacts" job like the latest diagram suggests.

ScottTodd (approved):
Please 1 more iteration on the diagram, otherwise LGTM

docs/development/ci_overview.md
geomin12: hah when i drew it out, i was like goodness this is really big and idk if folks need this much info

i'll keep it general + simple! thanks

PR #4279, opened 2026-03-31
Use OpenCL ICD from ROCK artifacts instead of system dependency

raramakr:
Remove external ocl-icd package dependencies and use the OpenCL ICD loader from ROCK's core-ocl-icd artifact


arvindcheru (commented):
Looks good to me.
Thanks

build_tools/packaging/linux/package.json
nunnikri, line 1460: dont we need to add the ocl-icd include folder here?

raramakr: The header files are part of the dev components

PR #4281, opened 2026-03-31
ci: Fetch gfx1101 kpack artifacts for Windows gfx110X tests

stellaraccident:
The Windows gfx110X test runners may have gfx1101 GPUs (cloud Radeon Pro VMs) but fetch-gfx-targets only listed gfx1100. With kpack split artifacts, each ISA has its own archive (blas_lib_gfx1100.kpack, blas_lib_gfx1101.kpack, etc.) so the fetch must include all targets the runner might have.

Changes:
- amdgpu_family_matrix.py: add gfx1101 to Windows gfx110X fetch targets
- new_amdgpu_family_matrix.py: same


HereThereBeDragons:
https://github.com/ROCm/TheRock/pull/3653 just for my tracking

PR #4294, opened 2026-04-01
Make artifact-group parameter optional to enable downloading all artifacts

raramakr:
Update fetch_artifacts.py to allow downloading all artifacts when neither --artifact-group nor --amdgpu-targets are specified. This provides more flexibility for users who want to fetch all available artifacts from a run.

### DIFF
--- a/build_tools/fetch_artifacts.py
+++ b/build_tools/fetch_artifacts.py
@@ -13,6 +13,10 @@ Example usage (using https://github.com/ROCm/TheRock/actions/runs/15685736080):
   python build_tools/fetch_artifacts.py \
     --run-id 15685736080 --artifact-group gfx110X-all --output-dir ~/.therock/artifacts_15685736080
 
+Download all artifacts for all GPU architectures:
+  python build_tools/fetch_artifacts.py \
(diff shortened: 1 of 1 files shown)


marbre (dismissed):
Why not just reusing what we already have?

https://github.com/ROCm/TheRock/blob/e5e7d6f7ac1e382de0ad77720680206f3cf8c06d/.github/workflows/build_portable_linux_python_packages.yml#L87-L96

artifact_manager should be the preferred option here.

ScottTodd:
I think support for "download all artifacts" is reasonable in both scripts, at least until we retire one in favor of the other (likely fetch_artifacts in favor of artifact_manager). Good point though - multi-arch workflows should use artifact_manager.py.

ScottTodd (approved):
Sorry for the delay reviewing. I think this looks fine. I spotted a few unrelated issues with artifact discovery and download along the way:

* https://github.com/ROCm/TheRock/pull/5027 to fix Windows extract
* https://github.com/ROCm/TheRock/pull/5037 to update the scripts that can find run ids for commits to use with this script (saving some time looking through the UI and supporting more complex automation)

PR #4296, opened 2026-04-01
[ci] Fixing skip CI `build_variant_label` null value for multi arch CI

geomin12:
during skipped CI runs for multi arch, we get errors such as:

```
evaluate reusable workflow name: Error when evaluating 'name' for job 'windows_build_and_test'. .github/workflows/multi_arch_ci.yml (Line: 88, Col: 11): could not get operand for index access: Error from function 'fromJSON': empty input
evaluate reusable workflow name: Error when evaluating 'name' for job 'linux_build_and_test'. .github/workflows/multi_arch_ci.yml (Line: 69, Col: 11): could not get operand for index access: Error from function 'fromJSON': empty input
```

such as in here: https://github.com/ROCm/TheRock/actions/runs/23817872655?pr=4275

This update fixes it, so inputs are included and not nulled, even if CI is skipped

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -880,8 +880,8 @@ def write_outputs(
     output_vars = {
         # Workflow YAML references this as 'enable_build_jobs'
         "enable_build_jobs": json.dumps(outputs.is_ci_enabled),
-        "linux_build_config": (json.dumps(linux.to_dict()) if linux else ""),
-        "windows_build_config": (json.dumps(windows.to_dict()) if windows else ""),
+        "linux_build_config": json.dumps(linux.to_dict()) if linux else "",
+        "windows_build_config": json.dumps(windows.to_dict()) if windows else "",
         "test_type": test_type,
         "linux_test_labels": outputs.linux_test_labels,
         "windows_test_labels": outputs.windows_test_labels,


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -877,11 +877,23 @@ def write_outputs(
     linux = outputs.builds.linux
     windows = outputs.builds.windows
     test_type = outputs.jobs.test_rocm.test_type if outputs.is_ci_enabled else ""
+    # In the case of skipped CI, downstream workflows expect the build variant label to be set
+    default_build_variant_label = {
+        "build_variant_label": ci_inputs.build_variant,
+    }
```
ScottTodd: https://github.com/ROCm/TheRock/actions/runs/23817872655?pr=4275
`evaluate reusable workflow name: Error when evaluating 'name' for job 'windows_build_and_test'. .github/workflows/multi_arch_ci.yml (Line: 88, Col: 11): could not get operand for index access: Error from function 'fromJSON': empty input evaluate reusable workflow name: Error when evaluating 'name' for job 'linux_build_and_test'. .github/workflows/multi_arch_ci.yml (Line: 69, Col: 11): could not get operand for index access: Error from function 'fromJSON': empty input`

https://github.com/ROCm/TheRock/blob/e5e7d6f7ac1e382de0ad77720680206f3cf8c06d/.github/workflows/multi_arch_ci.yml#L68-L75

ah... we could prune the variant label from the job names, or use some other pattern there.

1 sec, I'll draft some other options

ScottTodd: This probably works:

`name: Linux::${{ fromJSON(needs.setup.outputs.linux_build_config || '{}').build_variant_label || 'skip' }}`

I'd rather not create a "default" config and pass that around when the intent is to skip builds entirely

geomin12: ah good call, i was debating this too and just keeping it empty, but that'll work.

PR #4303, opened 2026-04-01
Adding JAX 0.9.1 build to CI and make jaxlib build conditional

gulsumgudukbay:
## Motivation

Add JAX 0.9.1 wheel build support to TheRock CI. Starting with JAX 0.9.1, jaxlib is no longer built as part of the ROCm wheel pipeline - it is used directly from upstream PyPI (`pip install jaxlib==0.9.1`). Only `jax_rocm7_plugin` and `jax_rocm7_pjrt` wheels are built for 0.9.1.

### Files changed

- **`.github/workflows/release_portable_linux_jax_wheels.yml`**
  Added `rocm-jaxlib-v0.9.1` to the `jax_ref` build matrix.

- **`.github/workflows/build_linux_jax_wheels.yml`**
  - Skip `ROCm/jax` checkout for 0.9.1 (not needed since jaxlib is not built).
  - Skip `--jax-source-dir` argument for 0.9.1 in the `ci_build` command.
  - Make jaxlib `--include` conditional in the S3 release copy step so it does not fail when no jaxlib wheel exists.

- **`build_tools/github_actions/write_jax_versions.py`**
  - Changed jaxlib wheel lookup from a hard failure to a log message when not found, since JAX >= 0.9.1 does not produce a jaxlib wheel.
  - Derive `jax_version` from `jax_rocm7_plugin` version when jaxlib is absent.

- **`.github/workflows/test_linux_jax_wheels.yml`**
  - For JAX >= 0.9.1, install `jax` and `jaxlib` together from upstream PyPI.
  - For JAX <= 0.9.0, behavior is unchanged: jaxlib is installed from the GPU-family index, jax from PyPI.

- **`external-builds/jax/README.md`**
  Added 0.9.1 to the supported versions table with a note that jaxlib is not built and comes from upstream PyPI.

- **`RELEASES.md`**
  Added 0.9.1 to the JAX version table with `upstream (from PyPI)` for jaxlib.

## Test Plan

- [ ] Trigger `build_linux_jax_wheels.yml` with `jax_ref=rocm-jaxlib-v0.9.1` for a single Python version and GPU family to verify:
  - Only `jax_rocm7_plugin` and `jax_rocm7_pjrt` wheels are produced (no jaxlib).
  - `write_jax_versions.py` succeeds with a log message instead of failing.
  - Wheels are uploaded to staging S3 correctly.
- [ ] Verify the test workflow installs `jax` and `jaxlib` from PyPI and `jax_rocm7_plugin`/`jax_rocm7_pjrt` from the GPU-family index.
- [ ] Verify existing versions (0.8.0, 0.8.2, 0.9.0) are unaffected by running the release workflow with those `jax_ref` values.

## Test Result

Pending CI validation. 
Local wheel build with `--therock-path` confirmed that `ci_build dist_wheels` produces only plugin and pjrt wheels for the 0.9.1 branch.

## Submission Checklist

### DIFF
--- a/.github/workflows/ci_nightly_pytorch_full_test.yml
+++ b/.github/workflows/ci_nightly_pytorch_full_test.yml
@@ -105,6 +105,7 @@ jobs:
           # TODO: When adding Windows support, pass --platform win_amd64 for
           # Windows targets or derive the platform from amdgpu_family/runner.
           TORCH_VERSION=$(pip index versions torch \
+            --pre \
             --index-url "${INDEX_URL}" \
             --python-version "${PYTHON_VERSION}" \
             --platform linux_x86_64 2>/dev/null \


.github/workflows/build_linux_jax_wheels.yml
```diff
@@ -328,12 +329,16 @@ jobs:
...
+          if [ -n "${JAXLIB_VERSION}" ]; then
+            JAXLIB_INCLUDE="--include jaxlib-${JAXLIB_VERSION}-${CP_VERSION}-manylinux_2_27_x86_64.whl"
+          fi
           aws s3 cp \
             s3://${S3_BUCKET_PY}/${{ inputs.s3_staging_subdir }}/${{ inputs.amdgpu_family }}/ \
             s3://${S3_BUCKET_PY}/${{ inputs.s3_subdir }}/${{ inputs.amdgpu_family }}/ \
             --recursive \
             --exclude "*" \
-            --include "jaxlib-${JAXLIB_VERSION}-${CP_VERSION}-manylinux_2_27_x86_64.whl" \
+            ${JAXLIB_INCLUDE} \
```
WBobby: Minor: unquoted variable in S3 copy step
${JAXLIB_INCLUDE} on the aws s3 cp line is unquoted.
Consider using a bash array or ${JAXLIB_INCLUDE:+${JAXLIB_INCLUDE}} to make the intent explicit.

.github/workflows/build_linux_jax_wheels.yml
```diff
@@ -175,7 +176,7 @@ jobs:
         run: |
           ls -lah
           pushd rocm-jax
-          [[ "${{ inputs.jax_ref }}" != *"0.8.0"* ]] && SOURCE_ARG="--jax-source-dir=${{ github.workspace }}/jax" || SOURCE_ARG=""
+          [[ "${{ inputs.jax_ref }}" != *"0.8.0"* && "${{ inputs.jax_ref }}" != *"0.9.1"* ]] && SOURCE_ARG="--jax-source-dir=${{ github.workspace }}/jax" || SOURCE_ARG=""
```
WBobby: The workflow uses contains(inputs.jax_ref, '0.9.1') and *"0.9.1"* string matching in multiple places. When 0.9.2, 0.10.0, etc. also skip jaxlib, every new version will require touching multiple workflow files.

Consider introducing an explicit workflow input (e.g. build_jaxlib: true/false) or a version-comparison approach (>=0.9.1) so that future versions only need a matrix entry change.

build_tools/github_actions/write_jax_versions.py
```diff
@@ -82,8 +89,8 @@ def get_all_jax_wheel_versions(
     else:
         raise FileNotFoundError("Did not find jax_rocm7_pjrt wheel")
 
-    # Base JAX version (without +rocm suffix) for installing jax from PyPI.
-    all_versions = all_versions | {"jax_version": jaxlib_version.split("+")[0]}
+    base_version_source = jaxlib_version or jax_plugin_version
+    all_versions = all_versions | {"jax_version": base_version_source.split("+")[0]}
```
WBobby, line 97: base_version_source = jaxlib_version or jax_plugin_version
all_versions = all_versions | {"jax_version": base_version_source.split("+")[0]}
This assumes jax_rocm7_plugin version always matches the jax PyPI package version. If they ever diverge, this silently produces a wrong jax_version. Worth at least a comment documenting the assumption.

WBobby (commented):
Thanks for the clean PR — the description is thorough and the changes are well-scoped. A few suggestions in the comments.

WBobby:
~~Question: 0.9.0 missing from RELEASES.md?~~

~~The matrix already includes rocm-jaxlib-v0.9.0, but the version table in RELEASES.md jumps from 0.8.2 to 0.9.1. Might be a pre-existing gap, but worth confirming whether 0.9.0 should be listed.~~

Skip 0.9.0

.github/workflows/build_linux_jax_wheels.yml
gulsumgudukbay: added workflow input

build_tools/github_actions/write_jax_versions.py
gulsumgudukbay: documented this in this file

.github/workflows/build_linux_jax_wheels.yml
gulsumgudukbay: used the ${JAXLIB_INCLUDE:+${JAXLIB_INCLUDE}} pattern

WBobby:
Thanks for addressing the earlier feedback — the build_jaxlib input, assumption comment, and ${JAXLIB_INCLUDE:+...} fix
  all look good. I traced the full CI pipeline for the build_jaxlib=false path and found two issues:

  Bug 1 (blocking): "Generate JAX manifest" and "Upload JAX manifest" steps fail when build_jaxlib=false

  The "Checkout JAX" step is correctly gated on inputs.build_jaxlib (line 159), so ${{ github.workspace }}/jax is never
  created for 0.9.1. However, the "Generate JAX manifest" step (line 196) still runs unconditionally:

        - name: Generate JAX manifest
          working-directory: build_tools
          run: |
            python3 -m github_actions.generate_jax_manifest \
              --jax-dir "${{ github.workspace }}/jax"   # <-- directory does not exist

  generate_jax_manifest.py → build_sources() → git_head(jax_dir) checks dirpath.exists() and raises FileNotFoundError
  (https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/manifest_utils.py#L90-L94). Since this step runs
  before "Extract JAX versions" (line 205), the entire build job fails and all downstream jobs (test, upload) are skipped.

  The "Upload JAX manifest to artifacts S3" step (line 233) has the same issue.

  Suggested fix — gate both steps:
        - name: Generate JAX manifest
          if: ${{ inputs.build_jaxlib }}
          ...

        - name: Upload JAX manifest to artifacts S3
          if: ${{ github.repository_owner == 'ROCm' && inputs.build_jaxlib }}
          ...

  (Or make --jax-dir optional in generate_jax_manifest.py and derive source info from rocm-jax when the JAX checkout is
  absent, if manifests are needed for 0.9.1+.)

  Bug 2 (latent, will surface once Bug 1 is fixed): Test workflow falls through to wrong version computation

  In test_linux_jax_wheels.yml line 227:
  if [ -n "${{ inputs.jaxlib_version }}" ] && [ -n "${{ inputs.jax_version }}" ]; then

  For 0.9.1, jaxlib_version is empty (no jaxlib wheel built), so this condition fails even though jax_version,
  jax_plugin_version, and jax_pjrt_version are correctly passed from the build workflow. The else branch calls
  compute_jax_package_version.py, which always produces a non-empty JAXLIB_VERSION
  (https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/compute_jax_package_version.py#L77:
  f"{jax_version}+rocm{sanitized_rocm}"). This causes the install step to try pip install --index-url <GPU-index>
  "jaxlib==0.9.1+rocm...", which will fail because no jaxlib wheel exists there.

  Suggested fix — use jax_plugin_version (always set) as the gate instead of jaxlib_version:
  if [ -n "${{ inputs.jax_plugin_version }}" ] && [ -n "${{ inputs.jax_version }}" ]; then

gulsumgudukbay:
> Thanks for addressing the earlier feedback — the build_jaxlib input, assumption comment, and ${JAXLIB_INCLUDE:+...} fix all look good. I traced the full CI pipeline for the build_jaxlib=false path and found two issues:
> 
> Bug 1 (blocking): "Generate JAX manifest" and "Upload JAX manifest" steps fail when build_jaxlib=false
> 
> The "Checkout JAX" step is correctly gated on inputs.build_jaxlib (line 159), so ${{ github.workspace }}/jax is never created for 0.9.1. However, the "Generate JAX manifest" step (line 196) still runs unconditionally:
> 
> ```
>     - name: Generate JAX manifest
>       working-directory: build_tools
>       run: |
>         python3 -m github_actions.generate_jax_manifest \
>           --jax-dir "${{ github.workspace }}/jax"   # <-- directory does not exist
> ```
> 
> generate_jax_manifest.py → build_sources() → git_head(jax_dir) checks dirpath.exists() and raises FileNotFoundError (https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/manifest_utils.py#L90-L94). Since this step runs before "Extract JAX versions" (line 205), the entire build job fails and all downstream jobs (test, upload) are skipped.
> 
> The "Upload JAX manifest to artifacts S3" step (line 233) has the same issue.
> 
> Suggested fix — gate both steps: - name: Generate JAX manifest if: ${{ inputs.build_jaxlib }} ...
> 
> ```
>     - name: Upload JAX manifest to artifacts S3
>       if: ${{ github.repository_owner == 'ROCm' && inputs.build_jaxlib }}
>       ...
> ```
> 
> (Or make --jax-dir optional in generate_jax_manifest.py and derive source info from rocm-jax when the JAX checkout is absent, if manifests are needed for 0.9.1+.)
> 
> Bug 2 (latent, will surface once Bug 1 is fixed): Test workflow falls through to wrong version computation
> 
> In test_linux_jax_wheels.yml line 227: if [ -n "${{ inputs.jaxlib_version }}" ] && [ -n "${{ inputs.jax_version }}" ]; then
> 
> For 0.9.1, jaxlib_version is empty (no jaxlib wheel built), so this condition fails even though jax_version, jax_plugin_version, and jax_pjrt_version are correctly passed from the build workflow. The else branch calls compute_jax_package_version.py, which always produces a non-empty JAXLIB_VERSION (https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/compute_jax_package_version.py#L77: f"{jax_version}+rocm{sanitized_rocm}"). This causes the install step to try pip install --index-url "jaxlib==0.9.1+rocm...", which will fail because no jaxlib wheel exists there.
> 
> Suggested fix — use jax_plugin_version (always set) as the gate instead of jaxlib_version: if [ -n "${{ inputs.jax_plugin_version }}" ] && [ -n "${{ inputs.jax_version }}" ]; then

Thanks for the comments, I addressed these in the latest commit

WBobby:
The code changes look good — both bugs from the previous round are fixed and the orchestration logic is correct. However,
  I'd like to hold off on approving until the 0.9.1 path is actually green end-to-end.

  In https://github.com/ROCm/TheRock/actions/runs/24212432933/job/70695792315, the 0.9.1 build and install steps all
  succeed, but every test fails with:

  RuntimeError: Backend 'rocm' is not in the list of known backends: ['cpu', 'tpu']
  INFO     jax_plugins.xla_rocm7:__init__.py:98  No ROCm wheel installation found
  WARNING  jax_plugins.xla_rocm7:__init__.py:243 rocm_plugin_extension not found

  Let me know if you'd like help drafting the rocm-jax patch.

.github/workflows/test_linux_jax_wheels.yml
```diff
@@ -245,16 +245,35 @@ jobs:
...
+          import importlib, traceback
+          for pkg in ['jax_rocm7_plugin', 'jax_rocm60_plugin', 'jaxlib.rocm']:
+              try:
+                  ext = importlib.import_module(f'{pkg}.rocm_plugin_extension')
+                  print(f'OK: {pkg}.rocm_plugin_extension loaded from {ext.__file__}')
+                  break
+              except Exception as e:
+                  print(f'FAIL: {pkg}.rocm_plugin_extension')
+                  traceback.print_exc()
+          "
```
marbre: Move this to a script please, avoid this kind of inline code. I know, agents love this but we don't.

gulsumgudukbay: > Move this to a script please, avoid this kind of inline code. I know, agents love this but we don't.

This is a temporary snippet to diagnose the plugin issues. I already diagnosed the issue and will revert this change. It will not live in the CI workflow.

WBobby:
The build_jaxlib passthrough expression has a GitHub Actions null-coercion bug:

  build_jaxlib: ${{ matrix.build_jaxlib != false }}

  In GitHub Actions, null == false evaluates to true (loose comparison), so for matrix entries that don't explicitly set
  build_jaxlib (0.8.0, 0.8.2, 0.9.0), matrix.build_jaxlib is null, and null != false → false. This means all versions get
  build_jaxlib: false, not just 0.9.1.

  Evidence from https://github.com/ROCm/TheRock/actions/runs/24212432933, job
  https://github.com/ROCm/TheRock/actions/runs/24212432933/job/70829955703 (0.8.2, py3.11):

  build_jaxlib: false                                          ← should be true
  WARNING: Found no 'jaxlib' wheels matching 'jaxlib-*.whl'   ← no jaxlib built

  Compare with the same version on main (https://github.com/ROCm/TheRock/actions/runs/24327586931/job/71026083281):

  SOURCE_ARG="--jax-source-dir=.../jax"
  Found wheel at 'jaxlib-0.8.2+rocm7.13.0a20260413-...'       ← jaxlib built correctly

  The 0.8.x/0.9.0 tests still pass by accident because the test workflow falls back to installing upstream PyPI
  jaxlib==0.8.2 (without +rocm suffix), which happens to be ABI-compatible. But the GPU index would be missing the
  ROCm-built jaxlib wheel — a silent regression for users.

  Fix: explicitly set build_jaxlib for every jax_ref and pass the value directly:

          include:
            - jax_ref: "rocm-jaxlib-v0.8.0"
              build_jaxlib: true
            - jax_ref: "rocm-jaxlib-v0.8.2"
              build_jaxlib: true
            - jax_ref: "rocm-jaxlib-v0.9.0"
              build_jaxlib: true
            - jax_ref: "rocm-jaxlib-v0.9.1"
              build_jaxlib: false

        # ...
        build_jaxlib: ${{ matrix.build_jaxlib }}

.github/workflows/test_linux_jax_wheels.yml
marbre: If this is not ready for review please mark the PR as a draft.

Issue #4269, opened 2026-03-31
[Issue] Jax job aborts (exit 134) running `jax/tests/multi_device_test.py` with `No ROCm wheel installation found`


chiranjeevipattigidi:
**Summary:**
The job aborts while running a JAX multi-device test

**Effected Archs:** `gfx1151`

**Suspeced root cause:** 
The environment running the test does not have the ROCm-enabled wheel/plugin installed

```
pytest jax/tests/multi_device_test.py -q --log-cli-level=INFO
```
### Error log:
```
jax/tests/multi_device_test.py::MultiDeviceTest::test_broadcast
INFO     jax_plugins.xla_rocm7:__init__.py:98 No ROCm wheel installation found
Fatal Python error: Aborted
...
/__w/_temp/...sh: line 1:   648 Aborted (core dumped) pytest jax/tests/multi_device_test.py -q --log-cli-level=INFO
##[error]Process completed with exit code 134.
```

**Full log:**
- https://github.com/ROCm/TheRock/actions/runs/23786307643/job/69328040537

srinivamd:
## Root Cause Investigation

**TL;DR:** TheRock PR #4012 (merged 2026-03-27) added `JAX_PLATFORMS=rocm`, forcing tests to run on GPU. This exposed a pre-existing gap: the JAX ROCm plugin's `set_rocm_paths()` cannot find HIP device bitcode in the TheRock tarball environment, so `HIP_DEVICE_LIB_PATH` is never set. When `multi_device_test.py` triggers GPU kernel JIT-compilation, HIP can't find `ocml.bc` → `abort()` → SIGABRT (exit 134).

### Root Cause

The installed plugin (`jax_plugins.xla_rocm7`) runs `set_rocm_paths()` at init time, which sets `HIP_DEVICE_LIB_PATH` so HIP can find device bitcode. In `jax_rocm_plugin/pjrt/python/__init__.py` (ROCm/rocm-jax):

```python
def set_rocm_paths(path):
    rocm_lib = None
    try:
        import rocm   # checks for `rocm` Python pip package
        rocm_lib = os.path.join(rocm.__path__[0], "lib")
    except ImportError:
        sp = path.parent.parent.parent   # site-packages root
        maybe_rocm_lib = os.path.join(sp, "rocm/lib")
        if os.path.exists(maybe_rocm_lib):
            rocm_lib = maybe_rocm_lib
    if not rocm_lib:
[...]

Issue #4270, opened 2026-03-31
[Issue] JAX:  `core_test` failures due to XLA autotuning `No valid config found` in portable wheel CI


chiranjeevipattigidi:
**Summary:**
The “Release portable Linux JAX Wheels” CI job fails when running JAX tests on ROCm. Multiple tests in `jax/tests/core_test.py` fail with `jax.errors.JaxRuntimeError: INTERNAL: Autotuning failed ... NOT_FOUND: No valid config found!.`

**Effected Archs:** `gfx1151`

### Failed Tests:
- `jax/tests/core_test.py::CoreTest::test_simple_jit`
- `jax/tests/core_test.py::CoreTest::test_tracer_reprs`
- `jax/tests/core_test.py::CoreTest::test_verbose_tracer_reprs`

**Full Error:**
- https://github.com/ROCm/TheRock/actions/runs/23786307643/job/69331875632

### Error 

```
Autotuning failed for HLO: %input_reduce_fusion = pred[] fusion(...), kind=kInput ... with error: NOT_FOUND: No valid config found!
Autotuning failed for HLO: %loop_reduce_fusion = f32[] fusion(...), kind=kLoop ... with error: NOT_FOUND: No valid config found!

```

srinivamd:
## Root Cause Summary

The CI failure is a two-part problem: a recent CI infrastructure fix exposed a pre-existing XLA architectural gap on gfx1151.

---

## Root Cause

**1. TheRock PR #4012** (merged 2026-03-27) added `JAX_PLATFORMS=rocm` to the JAX wheel test step, along with an `/opt/rocm` symlink for RUNPATH resolution. Before this fix, tests silently fell back to CPU and passed. After the fix, tests correctly run on GPU — and the real failure is exposed on gfx1151.

**2. XLA's GPU reduction autotuner has no valid tile configurations for gfx1151** (STRIX-HALO, RDNA3.5). The underlying cause is in `xla/stream_executor/rocm/rocm_compute_capability.h` and `xla/service/gpu/reduction_utils.cc`:

- gfx1151 uses **32-thread wavefronts** (RDNA) vs 64 threads on CDNA
- `gfx9_mi100_or_later()` returns **false** for gfx1151 — it only covers MI100/MI200/MI300 series
- `has_mfma_instr_support()` returns **false** for gfx1151
[...]

PR #4304, opened 2026-04-01
Add ROCm env vars to JAX test step to fix plugin discovery (#4269)

WBobby:
## Motivation

The JAX CI test job on `gfx1151` (and potentially other architectures) aborts with exit code 134 when running `jax/tests/multi_device_test.py`. The root cause is that the `jax_rocm7_plugin` cannot discover the ROCm device bitcode and LLD linker because the tarball-based ROCm installation at `/opt/rocm` does not match the paths the plugin probes by default.

This is a short-term fix to unblock CI while a proper long-term solution (e.g. embedding correct paths in the wheel or improving plugin auto-discovery) is developed.

Fixes https://github.com/ROCm/TheRock/issues/4269

## Technical Details

Added three environment variables to the "Run JAX tests" step in `.github/workflows/test_linux_jax_wheels.yml`:
HIP_DEVICE_LIB_PATH
JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH
JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH

These variables were already suggested in [#4269](https://github.com/ROCm/TheRock/issues/4269) by @srinivamd as the short-term workaround.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_linux_jax_wheels.yml
+++ b/.github/workflows/test_linux_jax_wheels.yml
@@ -255,6 +255,9 @@ jobs:
       - name: Run JAX tests
         env:
           JAX_PLATFORMS: rocm
+          HIP_DEVICE_LIB_PATH: /opt/rocm/lib/llvm/amdgcn/bitcode
+          JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH: /opt/rocm/lib/llvm/amdgcn/bitcode
+          JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH: /opt/rocm/lib/llvm/bin
         run: |
           pytest jax/tests/multi_device_test.py -q --log-cli-level=INFO
           pytest jax/tests/core_test.py -q --log-cli-level=INFO


srinivamd:
## Code Review

**Verdict: Approve** ✅

### What It Does

A 3-line CI fix adding three env vars to the "Run JAX tests" step in `test_linux_jax_wheels.yml`:

```yaml
HIP_DEVICE_LIB_PATH: /opt/rocm/lib/llvm/amdgcn/bitcode
JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH: /opt/rocm/lib/llvm/amdgcn/bitcode
JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH: /opt/rocm/lib/llvm/bin
```

These tell `jax_rocm7_plugin` where to find HIP device bitcode and `ld.lld`. The plugin's `set_rocm_paths()` auto-discovery only probes for a Python `rocm` pip package or `<site-packages>/rocm/lib` — neither of which exists in the TheRock tarball environment — so it returns early without setting `HIP_DEVICE_LIB_PATH`, causing SIGABRT when the HIP runtime can't find `ocml.bc`.

### Correctness

**Accurate diagnosis.** The fix directly addresses the root cause in `jax_rocm_plugin/pjrt/python/__init__.py` (`set_rocm_paths()` returns at line 98 without setting env vars). These three variables bypass that failure path entirely.

**Correct paths.** Given that `test_linux_jax_wheels.yml` already installs ROCm via `ln -sfn "${DEST}" /opt/rocm` (PR #4012), `/opt/rocm/lib/llvm/amdgcn/bitcode` and `/opt/rocm/lib/llvm/bin` are the correct locations in all TheRock tarballs.

### Minor Notes

- **Hardcoded `/opt/rocm`** — matches the symlink created by PR #4012. The coupling is implicit; a comment linking to that setup step would help future readers.
- **Applies to all arch legs** — env vars are added unconditionally to the full matrix. This is correct; CDNA targets (gfx942, etc.) would hit the same issue.
- **`HIP_DEVICE_LIB_PATH` and `JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH` are redundant but harmless** — the former is HIP's standard variable, the latter is the plugin's internal override. Setting both is the right belt-and-suspenders call.
- **Does not fix #4270** — the companion `NOT_FOUND: No valid config found!` autotuner failure on gfx1151 is a separate XLA gap; intentionally out of scope here.

The fix is minimal, correctly targeted, and matches the short-term workaround recommended in #4269. The proper long-term fix (updating `set_rocm_paths()` in ROCm/rocm-jax to fall back to `/opt/rocm/lib`) can follow as a separate upstream change.

PR #4308, opened 2026-04-02
Add per-ISA device wheels for kpack-split artifacts

stellaraccident:
When KPACK_SPLIT_ARTIFACTS is enabled in the build, the Python packaging pipeline now produces arch-neutral host wheels alongside per-ISA device wheels instead of fat per-family wheels with embedded device code.

Changes:
- New rocm-sdk-device wheel template that overlays .kpack archives and kernel databases into the libraries package's site-packages directory
- New "device" PackageEntry in _dist_info.py
- build_python_packages.py detects kpack-split mode via therock_manifest.json flags and branches into _run_kpack_split() vs _run_legacy()
- populate_device_files() in py_packaging.py does straight file copy for device artifacts (no RPATH/soname processing)
- In kpack-split mode, libraries entry template is overridden to drop the {target_family} suffix, producing a single rocm-sdk-libraries wheel
- Devel package excludes test component in kpack-split mode since generic test binaries cannot run without device code
- Legacy (non-split) packaging path is unchanged

Tested:
* Built local python packages and did install/static verification on them from downloaded artifacts
* Testing that the legacy path remains usable (this PR)

There are additional steps needed to fully comply with the new per-target wheel design, but this first step will get us structurally usable and unblock further testing on the kpack integration branch.

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -42,9 +42,12 @@ def load_therock_manifest(artifact_dir: Path) -> dict:
         / "therock"
         / "therock_manifest.json"
     )
-    if manifest_path.exists():
-        return json.loads(manifest_path.read_text())
-    return {}
+    if not manifest_path.exists():
+        raise FileNotFoundError(
+            f"therock_manifest.json not found at {manifest_path}. "
+            f"Is base_lib_generic present in {artifact_dir}?"
+        )
+    return json.loads(manifest_path.read_text())
 
 
 def run(args: argparse.Namespace):


build_tools/build_python_packages.py
```diff
@@ -22,19 +22,43 @@
...
+        artifact_dir
+        / "base_lib_generic"
+        / "base"
+        / "aux-overlay"
+        / "stage"
+        / "share"
+        / "therock"
+        / "therock_manifest.json"
+    )
+    if manifest_path.exists():
```
stellaraccident: Fail fast if manifest not found. No silent but deadlies please.

stellaraccident:
See:
* Implementation plan: https://github.com/stellaraccident/therock-workspace/blob/main/docs/analysis/kpack-python-packaging.md
* Overall design: https://github.com/stellaraccident/therock-workspace/blob/main/docs/analysis/kpack-wheelnext-proposal.md

build_tools/_therock_utils/py_packaging.py
```diff
@@ -176,6 +189,21 @@ def __init__(
...
+        # setup.py can declare the correct install_requires and overlay dir.
+        if logical_name == "device":
+            libraries_entry = self.params.dist_info.ALL_PACKAGES["libraries"]
+            libraries_dist_name = libraries_entry.get_dist_package_name(
+                target_family=None
+            )
+            libraries_py_package_name = libraries_entry.get_py_package_name(
+                target_family=None
+            )
+            dist_info_contents += f"LIBRARIES_DIST_NAME = {repr(libraries_dist_name)}\n"
```
marbre, line 202: NIT: `LIBRARIES_DIST_NAME` could theoretically be hard coded to `rocm-sdk-libraries` in `setup.py` since it is static. `LIBRARIES_PY_PACKAGE_NAME` cannot however. We can inject both here still.

marbre (approved):
Thanks, this looks good to me. Before you picked this up I was not sure how to overlay the packages (or if we should at all) and discussed this as part of the review with my agent. This might something we might at least release not because shared ownership has some general issues but this should not be concern for now:

---
#### Two wheels sharing the same Python package namespace

**File:** `build_tools/packaging/python/templates/rocm-sdk-device/setup.py:41`  
**Severity:** Known trade-off (acknowledged in PR description)

The overlay design means both `rocm-sdk-libraries` and `rocm-sdk-device-{gfx}`
install into `_rocm_sdk_libraries{nonce}` and each claims it in their
`dist-info/RECORD`. The PR description acknowledges this explicitly as a first
step. Practical consequences to track going forward:

- `pip uninstall rocm-sdk-libraries` removes only its RECORD entries — device
  files stay orphaned in `site-packages/_rocm_sdk_libraries{nonce}/`
- If pip ever adds ownership conflict detection (discussed upstream), this
  pattern would break
---

Another issue might be that `rocm-sdk-devel` so far was kind of device specific as it symlinked to `rocm-sdk-libraries`. It think this is now resolved as the `.so`'s do no longer contain any target specific code and we can simply mark `rocm-sdk-devel` as target neutral now. This will need a change in the workflow as we're currently uploading the `rocm-sdk-devel` to target specific subdirectories (to not need to deal with an `rocm-sdk-devel-gfxXXX` package as an intermediate solution). Anyway, we will need to fix workflows and update for kpack split enabled wheels.

build_tools/packaging/python/templates/rocm-sdk-device/setup.py
```diff
@@ -0,0 +1,74 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""ROCm SDK device package: per-ISA .kpack archives and kernel databases.
+
+Device wheels overlay into the rocm-sdk-libraries package's platform directory
+so the kpack runtime finds .kpack files alongside host shared libraries.
+"""
```
ScottTodd, line 8: (Can be a follow-up)

Can you document this new rocm-sdk-device package in https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md too?

Suggestions: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4308.md#documentation-docspackagingpython_packagingmd

build_tools/packaging/python/templates/rocm-sdk-device/setup.py
```diff
@@ -0,0 +1,74 @@
...
+
+# Collect all files in the platform directory as package_data.
+# include_package_data=True relies on VCS tracking which doesn't apply here.
+# Device files include .kpack archives (in dotfile dirs), kernel databases
+# (.co, .dat, .hsaco), MIOpen perf DBs (.kdb, .db.txt), and ML models.
+package_data_files = []
+if platform_dir.is_dir():
+    for path in platform_dir.rglob("*"):
+        if path.is_file() and path.name != "__init__.py":
+            package_data_files.append(str(path.relative_to(platform_dir)))
```
ScottTodd, line 51: (can be a follow-up)
Can we add some tests to https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests for `device_test.py` that does something with these files?

Suggestions: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_4308.md#sdk-tests-build_toolspackagingpythontemplatesrocmsrcrocm_sdktests

stellaraccident:
All CI failures are known issues (most notably #4311) and inspection of the logs reveals that anything this PR could be impacting are fine.

stellaraccident:
@ScottTodd / @marbre -- I'll address your comments in a followup.

Issue #3329, opened 2026-02-09
Refactor to move native package index file generation server side


marbre:
The script split out in #3328 needs to be refactored to not be called in a workflow and rather to be executed server side via an AWS Lambda.

marbre:
This task was implemented a few weeks ago but the ticket wasn't updated. Closing.

PR #4310, opened 2026-04-02
Multi-arch native package workflow improvements

nunnikri:
### `multi_arch_build_native_linux_packages.yml`

- **Switch artifact fetching to `artifact_manager.py`**: Replaces `fetch_artifacts.py` with `artifact_manager.py fetch` for consistent multi-arch artifact fetching across all GPU families
- **Fix system requirements**: Use `llvm-20` instead of `llvm`; add `pyzstd` for `.tar.zst` artifact decompression
- **Add `package_repository_url` output**: Workflow now exposes the public S3 install URL as an output for downstream consumption (e.g., install test jobs)
- **Add AWS credential guard**: `Configure AWS Credentials` step now only runs for authorized repositories (`ROCm/TheRock`, `ROCm/rockrel`), skipped for forks

### `docs/packaging/native_packaging.md`

- Add S3 bucket/prefix/URL reference tables for all release types (dev, nightly, prerelease, release, CI)
- Separate tables for GFX Specific Packages and Multi-Arch Packages

## Motivation

Update as per the review comments on https://github.com/ROCm/TheRock/pull/3561


build_tools/packaging/linux/tests/get_s3_config_test.py
```diff
@@ -14,6 +14,79 @@
...
+        """Test CI build URL format."""
+        url = get_s3_config.generate_package_repository_url(
+            release_type="ci",
+            pkg_type="deb",
+            yyyymmdd="20260320",
+            artifact_id="12345678",
+        )
+        self.assertEqual(
+            url,
+            "https://therock-ci-artifacts.s3.amazonaws.com/v3/packages/deb/20260320-12345678",
```
marbre: CI artifacts do not go into s3 subfolders, test seems wrong. Artifacts should rather go do GH runid subfolders, see for example where Python packages go.

marbre (commented):
Drive-by comment.

build_tools/packaging/linux/tests/get_s3_config_test.py
nunnikri: Thanks @marbre  for the feedback. I updated the ci prefix area to match artifacts and wheel. 
Now it will be 
https://therock-ci-artifacts.s3.amazonaws.com/99999999-linux/packages/deb
https://therock-ci-artifacts.s3.amazonaws.com/99999999-linux/packages/rpm

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -9,7 +9,7 @@ on:
         type: string
       artifact_group:
         type: string
-        description: "Artifact group (e.g. multi-arch-release)"
+        description: "Artifact group for fetching artifacts (e.g., 'multi-arch-release' for multi-arch builds). Note: This parameter is unused currently"
```
ScottTodd: If unused, remove it.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -132,41 +143,35 @@ jobs:
...
+            --artifacts-dir "${{ env.ARTIFACTS_DIR }}"/artifacts/ \
+            --pkg-type "${{ inputs.native_package_type }}" \
+            --version-suffix "${{ env.ARTIFACT_RUN_ID }}" \
             --enable_kpack=false
 
       - name: Configure AWS Credentials
+        # Only assume OIDC role for authorized repositories (ROCm/TheRock, ROCm/rockrel).
+        # This blocks fork PRs from accessing internal S3 buckets while allowing both
+        # CI builds (TheRock) and release builds (rockrel) to upload packages.
+        if: ${{ (github.repository == 'ROCm/TheRock' || github.repository == 'ROCm/rockrel') && !github.event.pull_request.head.repo.fork }}
```
ScottTodd, line 142: Fork PRs _can't access_ internal S3 buckets. This isn't "blocking" anything, it's just granting credentials where they are necessary. Docs: https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#authentication

build_tools/packaging/linux/get_s3_config.py
```diff
@@ -92,16 +92,59 @@ def extract_date_from_version(version: Optional[str]) -> str:
...
+        artifact_id: Artifact/run ID
+
+    Returns:
+        Public repository URL for package installation instructions
+
+    Examples:
+        CI:         https://therock-ci-artifacts.s3.amazonaws.com/v3/packages/deb/20260320-12345678
+        Nightly:    https://rocm.nightlies.amd.com/deb/20260320-12345678
+        Prerelease: https://rocm.prereleases.amd.com/packages/ubuntu2404 (deb) or .../rhel10 (rpm)
+        Release:    https://repo.amd.com/rocm/packages/ubuntu2404 (deb) or .../rhel10 (rpm)
```
ScottTodd: Information about bucket names, CDN URLs, file paths, etc. should generally move to common code, not be located here in `build_tools/packaging/linux/`. Fine to start here for now, but these branches depending on bucket are pretty concerning.

Some docs to reference:
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#distribution-channels-dev-nightly-release

build_tools/packaging/linux/get_s3_config.py
```diff
@@ -92,16 +92,59 @@ def extract_date_from_version(version: Optional[str]) -> str:
...
+    elif release_type == "release":
+        # Release packages use official repo domain with default OS profile
+        os_profile = "ubuntu2404" if pkg_type == "deb" else "rhel10"
+        return f"https://repo.amd.com/rocm/packages/{os_profile}"
+    elif release_type == "dev":
+        # Dev packages use S3 direct URL
+        return f"https://therock-dev-packages.s3.amazonaws.com/v3/packages/{pkg_type}/{yyyymmdd}-{artifact_id}"
+    else:
+        # CI builds (including empty release_type or 'ci')
+        return f"https://therock-ci-artifacts.s3.amazonaws.com/v3/packages/{pkg_type}/{yyyymmdd}-{artifact_id}"
```
ScottTodd: Echoing the comment from Marius, CI artifacts MUST follow https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md, with a top level path including the workflow run id. Release workflows will copy files from the CI artifacts bucket to a release bucket, and they will choose a subdirectory in an appropriate release bucket that makes sense for the given package type.

nunnikri: Yes i have updated this to top level path with run id. Now it will be
```
https://therock-ci-artifacts.s3.amazonaws.com/<run-id>-linux/packages/deb
https://therock-ci-artifacts.s3.amazonaws.com/<run-id>-linux/packages/rpm
```

build_tools/packaging/linux/get_s3_config.py
nunnikri: Ya I had the same thought. Thinking to move the script to build_tools/ .

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Ack will update the commens

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Ack

build_tools/packaging/linux/tests/get_s3_config_test.py
nunnikri: Done

build_tools/packaging/linux/get_s3_config.py
nunnikri: Done.

build_tools/packaging/linux/get_s3_config.py
```diff
@@ -18,11 +18,11 @@
   │
   ├─ ELSE IF fork PR OR non-ROCm/TheRock repository
   │  └─ Use: therock-ci-artifacts-external bucket (external CI)
-  │     └─ prefix: v3/packages/<pkg_type>/<YYYYMMDD>-<artifact_id>
+  │     └─ prefix: <artifact_id>-<platform>/packages/<pkg_type>
```
marbre: The prefix is not correct. When pushing to `therock-ci-artifacts-external`, there will be an additional prefix to document the repo, e.g. TheRock, rocm-systems etc.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -175,13 +176,9 @@ jobs:
...
           python ./build_tools/packaging/linux/upload_package_repo.py \
-            --pkg-type ${{ inputs.native_package_type }} \
-            --s3-bucket ${{ steps.s3_config.outputs.s3_bucket }} \
-            --artifact-id ${{ env.ARTIFACT_RUN_ID }} \
-            --job ${{ steps.s3_config.outputs.job_type }} \
+            --pkg-type "${{ inputs.native_package_type }}" \
+            --s3-bucket "${{ steps.s3_config.outputs.s3_bucket }}" \
+            --artifact-id "${{ env.ARTIFACT_RUN_ID }}" \
+            --job "${{ steps.s3_config.outputs.job_type }}" \
             --s3-prefix "${{ steps.s3_config.outputs.s3_prefix }}"
```
marbre: As far as I remember `upload_package_repo` calls into code to generate index files. This will interfere and most likely conflicts with the server side index file generation used for artifacts.

Needs #3329 as a prerequisite + most likely a refactoring of `build_tools/generate_s3_index.py` (no ticket created yet, needs a new one + discussion). The latter than needs to be deployed to the server-side Lambda. IMHO this might block this PR to go in as is.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -87,6 +89,11 @@ jobs:
 
       - name: Determine S3 bucket and prefix
         id: s3_config
+        # Selects S3 bucket and public URL based on release type and repository:
+        # - CI builds: therock-ci-artifacts (ROCm/TheRock) or external bucket (forks)
+        # - dev/nightly: therock-{type}-packages with dated subfolder
+        # - prerelease/release: therock-{type}-packages with stable path
+        # See get_s3_config.py for full decision tree
```
marbre: This logic seems outdated and not aligned with what is actually happening in pre-releases. Pre-releases also upload to an artifact bucket and not only to the packages bucket.

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -41,17 +37,23 @@ on:
         description: "Branch, tag or SHA to checkout. Defaults to the reference or SHA that triggered the workflow"
         type: string
         required: false
+    outputs:
+      package_repository_url:
+        description: "Public repository URL for package installation"
+        value: ${{ jobs.build_native_packages.outputs.package_repository_url }}
```
marbre, line 43: Unused?

build_tools/packaging/linux/get_s3_config.py
nunnikri: Good catch @marbre let me fix that and update the PR.

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: To handle the index files issues, upload_package_repo is NOT creating index html files for ci builds. So that it wont conflict with current lamda fns.  That check is already in the current PR. Once the full lambda is done we can disable index for all s3 buckets.

marbre: Does the index files generated via build_tools/generate_s3_index.py fulfill the needs for native packages though?

build_tools/packaging/linux/get_s3_config.py
nunnikri: Done

nunnikri: https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-TheRock/12345678-linux/packages/deb
https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-TheRock/12345678-linux/packages/rpm
will be the new location

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: It will used by the test job to do the install and verify

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Done

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: From the code perspective, build_tools/generate_s3_index.py seems fine for native packaging as well. 

build_tools/packaging/linux/generate_package_indexes.py is used for all other s3 buckets than ci. Once the ci is fully functional and reverified then can decide to use a single index generation logic across the s3 buckets.

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: The pre-release are using 
s3://therock-prerelease-packages/v3/packages/deb
s3://therock-prerelease-packages/v3/packages/rpm

Since its getting bit complicated, have updated the whole list in documentation as well in this PR. Please refer the table added in [native_packaging.md](https://github.com/ROCm/TheRock/pull/4310/changes#diff-baa0dc911001a054ee82109857637239a5946e4a6244aa46d919e165ad98fd2c)

.github/workflows/multi_arch_build_native_linux_packages.yml
marbre: `build_tools/generate_s3_index.py` is used server side, this it is already active and either it or your script wins. If `build_tools/generate_s3_index.py` works, we should stick with it and not override accidentally.

`build_tools/packaging/linux/generate_package_indexes.py` needs to be moved server side as well but that is part of a different issue.

.github/workflows/multi_arch_build_native_linux_packages.yml
marbre: This does not align with what is done in multi-arch with regards to Python packages and tarballs. For any release we produce via CI, packages will get uploaded to `therock-*-artifacts`. What release jobs do in addition to CI, they copy from `therock-*-artifacts` to `therock-*-{tarball,python}`. I think native package should follow the same approach.

nunnikri: Thanks @marbre  for the details. @ScottTodd The current nightly job are uploading the native packages to therock-nightly-packages bucket. Could you reuse the same bucket? Or want to change it? 
If we reuse it, no cloud front changes will be needed.

ScottTodd: I'm setting up the new release workflows to build on top of the CI workflows. ALL outputs from a workflow are uploaded to an "artifacts" bucket _first_, so every workflow run (CI, release, etc.) has the original copies of all files in one bucket + subfolder. A separate release-specific step will then _copy_ files from those artifacts buckets into the corresponding python/package/tarball/etc. buckets.

Docs: https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md

ScottTodd (commented):
Please update the PR title and description to make review easier and commit history understandable.
* This PR has no changes to `get_s3_config`, those changes were moved to https://github.com/ROCm/TheRock/pull/4633?
* The PR title "Streamline multi arch native package workflow" could be more specific now that some changes were moved to other PRs

.github/workflows/multi_arch_build_native_linux_packages.yml
ScottTodd: Passing through around via an explicit output is fine, and consistent with how we build and test rocm python packages: https://github.com/ROCm/TheRock/blob/8c15f74c25160e3db1f73b25a9f022000250a6f2/.github/workflows/ci_linux.yml#L170-L174

I would like to be careful about what was originally in the PR description:

>  Returns release-type-specific domains:
> 
> * CI: [https://therock-ci-artifacts.s3.amazonaws.com/{run_id}-{platform}/packages/{pkg_type}](https://therock-ci-artifacts.s3.amazonaws.com/%7Brun_id%7D-%7Bplatform%7D/packages/%7Bpkg_type%7D)
> * Nightly: [https://rocm.nightlies.amd.com/{pkg_type}/{YYYYMMDD}-{artifact_id}](https://rocm.nightlies.amd.com/%7Bpkg_type%7D/%7BYYYYMMDD%7D-%7Bartifact_id%7D)
> * Dev: [https://therock-dev-packages.s3.amazonaws.com/v3/packages/{pkg_type}/{YYYYMMDD}-{artifact_id}](https://therock-dev-packages.s3.amazonaws.com/v3/packages/%7Bpkg_type%7D/%7BYYYYMMDD%7D-%7Bartifact_id%7D)
> * Prerelease: [https://rocm.prereleases.amd.com/packages/{os-profile}](https://rocm.prereleases.amd.com/packages/%7Bos-profile%7D)
> * Release: [https://repo.amd.com/rocm/packages/{os-profile}](https://repo.amd.com/rocm/packages/%7Bos-profile%7D)

If we keep the package structure predictable enough, downstream jobs _could_ recompute the expected URL themselves. That being said, explicit plumbing is safer in workflows like this, so having predictable and similar URLs would be more useful for users/developers who want to switch channels more easily (particularly between dev/prerelease/release).

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -74,19 +76,24 @@ jobs:
 
       - name: Install Python requirements
         run: |
-          pip install pyelftools boto3 jinja2
+          pip install pyelftools boto3 jinja2 pyzstd
```
ScottTodd, line 79: nit: these deps could all go in https://github.com/ROCm/TheRock/blob/main/requirements.txt, so developers can run the same native packaging scripts themselves without needing to parse through what the workflow code has.

Installing the full requirements list on github runners shouldn't be too much slower than installing this smaller list, so long as we have a well-functioning pip cache (perhaps built into the dockerfile)

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -95,6 +102,7 @@ jobs:
             --pkg-type "${{ inputs.native_package_type }}" \
             --artifact-id "${{ env.ARTIFACT_RUN_ID }}" \
             --rocm-version "${{ inputs.rocm_version }}" \
+            --platform linux \
```
ScottTodd: This argument doesn't exist in https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/linux/get_s3_config.py on the base branch (`main`). Remove from this PR?

(I don't think a `--platform linux` option is appropriate for a script in `build_tools/packaging/linux/`, so I'll give that review feedback on https://github.com/ROCm/TheRock/pull/4633 too)

build_tools/packaging/linux/detect_kpack.py
ScottTodd, line 1: This is a pretty verbose script with some complicated supporting workflow code in `.github/workflows/multi_arch_build_native_linux_packages.yml`

How about inlining the implementation into `build_tools/packaging/linux/build_package.py`? See what we've done for
* `build_python_packages.py`: https://github.com/ROCm/TheRock/blob/8c15f74c25160e3db1f73b25a9f022000250a6f2/build_tools/build_python_packages.py#L66-L69
* `build_tarballs.py`: https://github.com/ROCm/TheRock/blob/8c15f74c25160e3db1f73b25a9f022000250a6f2/build_tools/build_tarballs.py#L102-L108

We could add a common helper that each script uses, but it should go in `build_tools/`, not `build_tools/packaging/linux/`. If we go that route, I'd rather it be generic and support reading _any_ flag value from the manifest, not just `KPACK_SPLIT_ARTIFACTS`. We could then document that module in https://github.com/ROCm/TheRock/blob/main/docs/development/flags.md as one of the supported way to work with flags.

docs/packaging/native_packaging.md
```diff
@@ -89,6 +89,47 @@ Optional Fields
...
+| ---------------------------------- | ------------------------------- | ----------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| **dev**                            | `therock-dev-packages`          | `{pkg_type}/{yyyymmdd}-{artifact_id}`<br>Example: `deb/20260406-12345678`                                         | `https://rocm.devreleases.amd.com/deb/{yyyymmdd}-{artifact_id}`                                                      | `https://rocm.devreleases.amd.com/rpm/{yyyymmdd}-{artifact_id}/x86_64/`                                                      |
```
ScottTodd, line 100: For ROCm packages and tarballs dev versions use commit hashes, not dates (there could be multiple dev releases per day). What was the rationale for `{yyyymmdd}-{artifact_id}`?

References:
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#python-package-versions
* https://github.com/ROCm/TheRock/blob/main/build_tools/compute_rocm_package_version.py (I see that deb and rpm versions in there have dates, so now I'm wondering how important that is in both the file name _and_ the directory name like here)

nunnikri: This is a design decision we took at the beginning of the project. If needed we can revisit later. May not be needed in this PR

build_tools/packaging/linux/detect_kpack.py
nunnikri: Done. For now removed the script and done the same way as in build_python.

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: There is a reason why dev/nightly folder structure is different from prerelease/release. 
In prerelease/release there is only one folder for each package type. This model is error prone for the repo metadata when it grows substantially. Thats why we dont want to use the same single folder structure for dev/nightlies.

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: Done. Thanks

.github/workflows/multi_arch_build_native_linux_packages.yml
nunnikri: This is now aligned with workflow_outputs

.github/workflows/multi_arch_build_native_linux_packages.yml
ScottTodd, line 146: > * **Add AWS credential guard**: `Configure AWS Credentials` step now only runs for authorized repositories (`ROCm/TheRock`, `ROCm/rockrel`), skipped for forks

Please use https://github.com/ROCm/TheRock/blob/main/.github/actions/configure_aws_artifacts_credentials/action.yml instead of this code and the `name: Determine IAM role` step. It should handle all of this for you. See https://github.com/ROCm/TheRock/pull/4386 (in particular the section about migrating native linux packages in a follow-up)

This code is duplicating some policy decisions and I can't easily tell if each is modeled correctly:
* CI runs in other ROCm organization repositories like rocm-libraries, rocm-systems, and llvm-project (which is a fork!) should upload to `therock-ci-artifacts-external`, for which the runner base credentials are sufficient
* CI runs in TheRock for PRs from forks should also use `therock-ci-artifacts-external`
* Release runs should use `therock-{release_type}-artifacts`. Only a few repositories are authorized to assume the roles there, but I've allowed the steps to run in any repository for easier testing in forks (see https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#working-effectively-from-forks)

ScottTodd (approved):
My comment here can be a follow-up task

PR #4314, opened 2026-04-02
Fix external_repo prefix incorrectly applied to release artifact uploads

marbre:
When `RELEASE_TYPE` is set, `_retrieve_bucket_info` already routes uploads to `therock-{dev,nightly,prerelease}-artifacts`. However, `external_repo` was still computed from repository context and returned non-empty for repos other than `ROCm/TheRock` (e.g., `ROCm/rockrel`), causing release artifact paths to be prefixed with `owner-repo/`.

The `external_repo` prefix is meaningful only for
`therock-ci-artifacts-external` (CI builds from forks or non-TheRock repos). All release bucket uploads should use a clean path regardless of source repository.

Fix: force `external_repo = ""` in the `RELEASE_TYPE` branch.

### DIFF
--- a/build_tools/_therock_utils/workflow_outputs.py
+++ b/build_tools/_therock_utils/workflow_outputs.py
@@ -82,14 +82,14 @@ class WorkflowOutputRoot:
     external_repo: str
     """Run path prefix within the bucket (e.g., 'owner-repo/').
 
-    Non-empty only when ``bucket`` is ``therock-ci-artifacts-external``
+    Non-empty only when 'bucket' is 'therock-ci-artifacts-external'
     (i.e., CI builds from a fork or a repository other than ROCm/TheRock).
     Empty for all other buckets, including release buckets
-    (``therock-{dev,nightly,prerelease}-artifacts``) regardless of which
+    ('therock-{dev,nightly,prerelease}-artifacts') regardless of which
     repository triggered the workflow.
 
     This is an implementation detail of path construction - callers should
-    prefer the :attr:`prefix` property or the location methods over reading
+    prefer the 'prefix' property or the location methods over reading
     this field directly.
     """
 


marbre:
If have drafted a follow up which renames `external_repo` as this actually does not reflect what this is used for but let's discuss this separately.

build_tools/_therock_utils/workflow_outputs.py
```diff
@@ -368,6 +379,9 @@ def _retrieve_bucket_info(
                 f"expected one of {sorted(_VALID_RELEASE_TYPES)}"
             )
         _log(f"  (implicit) RELEASE_TYPE: {release_type}")
+        # Release builds always use a clean path - the external_repo prefix is
+        # only meaningful for CI builds going into therock-ci-artifacts-external.
+        external_repo = ""
         bucket = f"therock-{release_type}-artifacts"
```
ScottTodd, line 385: Changing this now will prevent newer code from reading from existing folders in the `therock-prerelease-artifacts` bucket that use the `ROCm-rockrel/` subfolder. I think that's fine, but
* We could maintain backwards compatibility somehow (a date check like above, checking both, etc.)
* We should probably add a way to pull from an explicit index (like on https://github.com/ROCm/TheRock/pull/4088, cc @PeterCDMcLean ) as that will be more robust to changes in paths like this.

Before this change, I'm able to download from https://therock-prerelease-artifacts.s3.amazonaws.com/ROCm-rockrel/19544155346-windows/index-gfx1151.html:
```cmd
set RELEASE_TYPE=prerelease
set GITHUB_REPOSITORY=ROCm/rockrel
python build_tools/fetch_artifacts.py --run-id 19544155346 --artifact-group gfx1151 --dry-run
Retrieving bucket info...
  (implicit) github_repository: ROCm/rockrel
  workflow_run_id             : 19544155346
  head_github_repository      : ROCm/rockrel
  is_pr_from_fork             : False
  (implicit) RELEASE_TYPE: prerelease
Retrieved bucket info:
  external_repo: ROCm-rockrel/
  bucket       : therock-prerelease-artifacts
Retrieving artifacts from 's3://therock-prerelease-artifacts/ROCm-rockrel/19544155346-windows'
Matching artifact target families: ['generic', 'gfx1151']

Filtered artifacts to download:
  s3://therock-prerelease-artifacts/ROCm-rockrel/19544155346-windows/amd-llvm_dbg_generic.tar.xz
  s3://therock-prerelease-artifacts/ROCm-rockrel/19544155346-windows/amd-llvm_dev_generic.tar.xz
```

After, it can't find files:
```cmd
(.venv) λ git diff
diff --git a/build_tools/_therock_utils/workflow_outputs.py b/build_tools/_therock_utils/workflow_outputs.py
index 0ea06668..04604515 100644
--- a/build_tools/_therock_utils/workflow_outputs.py
+++ b/build_tools/_therock_utils/workflow_outputs.py
@@ -368,6 +368,7 @@ def _retrieve_bucket_info(
                 f"expected one of {sorted(_VALID_RELEASE_TYPES)}"
             )
         _log(f"  (implicit) RELEASE_TYPE: {release_type}")
+        external_repo = ""
         bucket = f"therock-{release_type}-artifacts"
     else:
         if external_repo == "":

(.venv) λ python build_tools/fetch_artifacts.py --run-id 19544155346 --artifact-group gfx1151 --dry-run
Retrieving bucket info...
  (implicit) github_repository: ROCm/rockrel
  workflow_run_id             : 19544155346
  head_github_repository      : ROCm/rockrel
  is_pr_from_fork             : False
  (implicit) RELEASE_TYPE: prerelease
Retrieved bucket info:
  external_repo:
  bucket       : therock-prerelease-artifacts
Retrieving artifacts from 's3://therock-prerelease-artifacts/19544155346-windows'
Matching artifact target families: ['generic', 'gfx1151']
Found no artifacts matching ['generic', 'gfx1151'] at 's3://therock-prerelease-artifacts/19544155346-windows'
No matching artifacts for 19544155346 exist. Exiting...
```

build_tools/_therock_utils/workflow_outputs.py
```diff
@@ -80,7 +80,18 @@ class WorkflowOutputRoot:
...
+
+    Non-empty only when ``bucket`` is ``therock-ci-artifacts-external``
+    (i.e., CI builds from a fork or a repository other than ROCm/TheRock).
+    Empty for all other buckets, including release buckets
+    (``therock-{dev,nightly,prerelease}-artifacts``) regardless of which
+    repository triggered the workflow.
+
+    This is an implementation detail of path construction - callers should
+    prefer the :attr:`prefix` property or the location methods over reading
+    this field directly.
```
ScottTodd: nit: Claude Code likes this style of docstring with double backticks and `:attr:`, but we don't feed our docstrings through Sphinx. I've been removing such text from comments when I notice it.

build_tools/_therock_utils/workflow_outputs.py
marbre: I was just considering to move over all prerelease artifacts to the new location.

ScottTodd: That would work too :)

marbre: Moved the files.

build_tools/_therock_utils/workflow_outputs.py
marbre: Fixed.

build_tools/_therock_utils/workflow_outputs.py
ScottTodd: Ah I did include this change as part of https://github.com/ROCm/TheRock/pull/4386

https://github.com/ROCm/TheRock/blob/e699758928c9230bd1b1d0a0a4e05afef271587d/build_tools/_therock_utils/workflow_outputs.py#L310-L321

`external_repo` is only set if the bucket is `therock-ci-artifacts-external`

marbre:
> Ah I did include this change as part of #4386

Closing as @ScottTodd has incorporated this into his PR.

PR #4320, opened 2026-04-02
[ci] Adding clearer target intent for test_filtering.md doc

geomin12:
Users were getting some confusion on test filter intent. Updating docs to reflect this

### DIFF
--- a/docs/development/test_filtering.md
+++ b/docs/development/test_filtering.md
@@ -11,7 +11,7 @@
   - Execution time: < 5 min
   - Target: build system updates, CI system updates, non-component specific updates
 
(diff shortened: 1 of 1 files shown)


docs/development/test_filtering.md
```diff
@@ -8,31 +8,31 @@
   - Runs on: pull requests (if ROCm non-component related change), push to main branch
   - Characteristics: Shallow validation, focus on critical paths, component runs properly
   - Execution time: < 5 min
-  - Example: pull request change to build system, main branch push change for CI
+  - Target: build system updates, CI system updates, non-component specific updates
```
reboss, line 12: What you have is fine, if you wanted to go a step further to make things super clear, it could be helpful to add a note like:
**Note:** not intended to run on component changes, this is for changes to CI that have a very large footprint

Issue #4176, opened 2026-03-26
[Issue] ROCm Sanity Check Aborts with sysfs `topology/nodes` Not Found


chiranjeevipattigidi:
**Description:**
The `Release portable Linux PyTorch Wheels` workflow fails when running the pytorch `nightly` build and sanity check steps. The error is due to the ROCm profiling agent expecting the sysfs path `/sys/class/kfd/kfd/topology/nodes` to exist, but it is missing in the CI environment.

**Effected Gfx:**  `All the Supported Archs`
**Pytorch Version:** `nightly`
**Platform:** `Linux`

**Log Snippet:**
```
F20260326 08:25:46.129966 139771792310848 agent.cpp:638] sysfs nodes path '/sys/class/kfd/kfd/topology/nodes' does not exist
    @     0x7f1f165f9568  rocprofiler::agent::(anonymous namespace)::read_topology()
    @     0x7f1f165f0f10  rocprofiler::agent::(anonymous namespace)::get_agent_topology()
    @     0x7f1f165f0cc6  rocprofiler::agent::get_agents()
    @     0x7f1f165fa693  rocprofiler_query_available_agents
    @     0x7f1c9f5284dc  RocprofLogger::toolInit(void (*)(rocprofiler_client_id_t), void*)
RuntimeError: torch package sanity check failed (see output above)
Process completed with exit code 1.
```
**Full Logs:**
- https://github.com/ROCm/TheRock/actions/runs/23582460453/job/68668020979
- https://github.com/ROCm/TheRock/actions/runs/23582345576/job/68667686944
- https://github.com/ROCm/TheRock/actions/runs/23582060383/job/68666836251
- https://github.com/ROCm/TheRock/actions/runs/23581478231/job/68665094722

**Observed Behavior:**
- The ROCm profiling agent fails because the CI runner does not provide `/sys/class/kfd/kfd/topology/nodes`
- This causes `import torch; print(torch.cuda.is_available())`

amd-chiranjeevi:
comparision between upstream latest and last pytorch `nightly` releases https://github.com/pytorch/pytorch/compare/958d381444ebcad946b965a08545106898420f00...b6107b8b4651745dbd9788e0c17b8bb70b2211d5

amd-chiranjeevi:
as per @jeffdaily this is most likely related to rocprofiler, rocprofiler aborting the build instead of graceful exit when syss paths are not available.

PR #4336, opened 2026-04-03
Fix PyTorch sanity check SIGABRT by disabling ROCPROFILER_BUILD_CI

WBobby:
Fixes: #4176

## Motivation
The post-build PyTorch wheel sanity check (`import torch`) crashes with SIGABRT
on CPU-only CI runners. This is because rocprofiler-sdk, built with
`ROCPROFILER_BUILD_CI=ON`, treats the missing sysfs path
`/sys/class/kfd/kfd/topology/nodes` as a fatal error, aborting the process
before the import can complete.

## Technical Details
Changed `-DROCPROFILER_BUILD_CI=ON` to `-DROCPROFILER_BUILD_CI=OFF` in
`profiler/CMakeLists.txt`. This causes rocprofiler-sdk to log the missing
topology path as a warning (`W`) instead of a fatal error (`F`), allowing
the process to continue normally on machines without AMD GPU hardware.

## Test Plan
Triggered the full release pipeline from the `users/yanywang/temp-check-pytorch`
branch:
1. Built ROCm packages via `release_portable_linux_packages` workflow:
   https://github.com/ROCm/TheRock/actions/runs/23931642087
2. Auto-triggered PyTorch wheel builds via `release_portable_linux_pytorch_wheels`:
   https://github.com/ROCm/TheRock/actions/runs/23934957174

Verified that the **Sanity Check Wheel** step passes across all 19 build
combinations (Python 3.10–3.14 × torch release/2.8, 2.9, 2.10, nightly).

## Test Result

- **Build**: All 19 combinations passed.
- **Sanity Check**: All 19 combinations passed. The log now shows
  `W ... sysfs nodes path does not exist` (warning) instead of
  `F ... ` (fatal/SIGABRT).
- **Test (on GPU)**: Failures on nightly and py 3.14 are pre-existing and unrelated to this change.

## Submission Checklist


jbonnell-amd:
Seems that setting `ROCPROFILER_BUILD_CI=OFF` has no effect on the sdk test run result, so this change appears to be fine (I believe this setting just changes some backend vars for CI usage, otherwise won't have any impacts on the build)

I've added some members of rocprofiler-sdk team (@venkat1361 and @ammarwa) for visibility, unsure if this is a regression at all or expected behavior.

PR #4339, opened 2026-04-03
Make PackageConfig immutable to prevent accidental mutations

raramakr:
Replace mutable PackageConfig with frozen dataclass to prevent bugs
  from accidental state changes during package builds. Config fields
  like gfx_arch and versioned_pkg were previously mutated in-place,
  making the code fragile and error-prone.

  Changes:
  - Add @dataclass(frozen=True) to PackageConfig
  - Change gfxarch_list from list to tuple for immutability
  - Replace all config mutations with dataclasses.replace() calls
  - Extract create_package_config() function from inline code in run()
  - Update all package creation functions to use replace() instead of mutating config fields
  - Improve architecture mode comments for clarity

  Benefits:
  - Prevents accidental config modifications
  - Makes data flow explicit (input config → build_config)
  - Config changes require explicit replace() calls
  - Provides thread-safety foundation for future parallel execution

  Files modified:
  - packaging_utils.py: PackageConfig definition, helper functions
  - build_package.py: Config creation, package build functions

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -31,6 +31,10 @@ GFX_GENERIC = "gfx_generic"
 # versioned_pkg - Used to indicate versioned or non versioned packages
 # enable_kpack - To enable multi-architecture support
 # gfxarch_list - List of all architectures for multi-arch mode
+#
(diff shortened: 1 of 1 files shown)


raramakr:
> Overall looks ok. Add more comments in the code why tuple and frozen is needed

Added the same

PR #4342, opened 2026-04-03
doc: windows build clarifications

jeromew:
## Motivation

I managed after a lot of effort to build TheRock on my Windows box without any failure for my gfx1031 :-)
this PR tries to clarify some gotchas that blocked me along the way in order to help future Windows builders.

## Technical Details

`nasm` is not mentioned in the doc.
The rest is gotchas that can happen on a living machine where the path may have incompatible installs that pollute the pure visual studio 2022 build chain (w64devkit, msys2, python)

## Test Plan

My objective was to achieve a full build.

## Test Result

I achieved a build with -DBUILD_TESTING=OFF as building the tests was taking too much time and I did not have the patience to wait for them. Now that it works, I could start a full build if necessary & if you tell me which tests are supposed to work when building for gfx1031 only.

## Submission Checklist

### DIFF
--- a/docs/development/windows_support.md
+++ b/docs/development/windows_support.md
@@ -209,13 +209,13 @@ If you prefer to install tools manually, you will need:
 
 - gfortran, recommended from Strawberry Perl: https://strawberryperl.com/
 
+- nasm, recommended from Strawberry Perl: https://strawberryperl.com/
+
 - patch, available in Strawberry Perl or Git.
 
 - dvc: https://dvc.org/doc/install/windows
 
 - Python: https://www.python.org/downloads/ (3.11+ recommended)
-
-- nasm: https://www.nasm.us/, needed for BoringSSL which is a dependency of grpc (therock-grpc)
 
 > [!WARNING]
 > Prefer to install Python for the current user only and to a path


docs/development/windows_support.md
```diff
@@ -215,6 +215,8 @@ If you prefer to install tools manually, you will need:
 
 - Python: https://www.python.org/downloads/ (3.11+ recommended)
 
+- nasm: https://www.nasm.us/, needed for BoringSSL which is a dependency of grpc (therock-grpc)
```
ScottTodd: I don't believe we require an explicit install of this, as this is the first I've heard of it. I haven't seen other developers or CI systems need to install it, for example.

Can you share more information here, like error logs from configure/build steps, instructions for installing via a package manager that fixed those errors, etc.?

ScottTodd: nit: pre-commit wants the trailing space at the end of this line removed: https://github.com/ROCm/TheRock/actions/runs/23964756835/job/70111126057?pr=4342#step:4:187

jeromew: yes that's weird
I got that at the `therock-grpc` configure stage

```
[therock-grpc configure] -- The ASM_NASM compiler identification is unknown
[therock-grpc configure] -- Didn't find assembler
[therock-grpc configure] CMake Error at third_party/boringssl-with-bazel/CMakeLists.txt:50 (enable_language):
[therock-grpc configure]   No CMAKE_ASM_NASM_COMPILER could be found.
[therock-grpc configure]
[therock-grpc configure]   Tell CMake where to find the compiler by setting either the environment
[therock-grpc configure]   variable "ASM_NASM" or the CMake cache entry CMAKE_ASM_NASM_COMPILER to the
[therock-grpc configure]   full path to the compiler, or to the compiler name if it is in the PATH.
[therock-grpc configure]
[therock-grpc configure]
[therock-grpc configure] -- Configuring incomplete, errors occurred!
```

I am using scoop as a package installer and simply did a `scoop install nasm` which makes the nasm executable available in the PATH

I have no idea why I would be alone to have that.

jeromew: the only reference i could find is https://stackoverflow.com/questions/73214824/missing-cmake-asm-nasm-compiler-when-compiling-grpc-with-ms-visual-studio

jeromew: done

jeromew:
Note that if nasm is indeed necessary, it seems that it can be installed via winget

https://winget.run/pkg/NASM/NASM

```winget install -e --id NASM.NASM```

but it would probably be necessary that someone other than me tests with a fresh - no nasm - install on windows to check that the dependency is indeed necessary.

I could also remove nasm from the dependency list if you prefer and move it to the troubleshooting section

docs/development/windows_support.md
ScottTodd: Maybe different MSVC versions include different deps... we might run into that when we update the version used by our CI builders?

Do either of these check scripts catch this?
* https://github.com/ROCm/TheRock/blob/main/build_tools/validate_windows_install.ps1
    On my system:
    ```
    3. MSVC Compiler
    ----------------
      [PASS] cl.exe in PATH: C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.42.34433\bin\HostX64\x64\cl.exe
      [info] usage: cl [ option... ] filename... [ /link linkoption... ]
      [PASS] MSVC developer environment activated (x64)
      [PASS] C++ ATL headers found
      [PASS] rc.exe in PATH: C:\Program Files (x86)\Windows Kits\10\bin\10.0.26100.0\\x64\rc.exe
      [PASS] Test compile + link succeeded (um, ucrt, shared, winrt headers; kernel32.lib)
    ```
* https://github.com/ROCm/TheRock/blob/main/build_tools/health_status.py
    On my system:
    ```
    [✓][Visual Studio] Found Visual Studio VS2022.
    [✓][MSVC/VC++] Found MSVC 19.42.34436 (v143) at C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/HostX64/x64/cl.exe
    [✓][MSVC/ATL] Found MSVC ATL.
    [✓][MSVC/ML64] Found Microsoft Macro Assembler 14.42.34436 (v143) at C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/HostX64/x64/ml64.exe
    [✓][MSVC/Archiver] Found MSVC Library Manager lib.exe at C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/HostX64/x64/lib.exe
    [✓][MSVC/Linker] Found Microsoft Incremental Linker link.exe at C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/VC/Tools/MSVC/14.42.34433/bin/HostX64/x64/link.exe
    [✓][Windows SDK/UCRT] Found Microsoft Resource Compiler at C:/Program Files (x86)/Windows Kits/10/bin/10.0.26100.0//x64/rc.exe
    ```

jeromew: No the scripts do not catch the nasm issue, even when i uninstall it.

but reading the result of `validate_windows_install.ps1` regarding gfortran, i understand that you expect `gfortran.exe` to be in the path (winget strawberry perl probably adds a path to its directory of c tools)

my `scoop install perl` installed strawberry but did not add its directory of binaries to the path which is why `gfortran.exe` was not found by cmake and I needed the `FC` (Fortran Compiler) env variable that made cmake find it.

but after a check, `nasm.exe` is also within the tools that strawberry makes available in its directory of binaries so I think what happens is that your `gfortran.exe` dependency install via winget strawberry perl implicitely installs the `nasm.exe` dependency needed for therock-grpc in the path

tell me what you think. I'll push an update with a modification along those lines

ScottTodd: Ah, okay. What you have now is fine.

We'd like to remove the Strawberry Perl requirement at some point. We only intended to include it for gfortran but it includes many other tools too. Originally, we couldn't find an easier way to install a fortran compiler. In the future we may switch to flang that we build from source using LLVM. If/when we do that and drop the Strawberry Perl recommendation, we may need to revisit this and find another source for nasm.

PR #4345, opened 2026-04-04
[hipDNN] Integration Test Project

mousdahl-amd:
## Motivation
We've written a engine-plugin agnostic test harness against hipDNN's frontend.  We want to have this harness built as part of CI and then subsequently used to test against all the engine plugins available (currently miopenprovider, hipblasltprovider, fusilliprovider, and eventually the hip-kernel-provider).  The harness can be configured to target all plugins or a single plugin, and offers services for adjusting the test list or test tolerances.  Using this, we can ensure that the providers support the shapes and operations they claim to, and also that hipDNN as a whole can support all shapes regardless of engine providers.

## Technical Details
- Added hipdnn-integration-tests as a new package / artifact
- Added cmake package export to allow providers access to a cmake function that simplifies integration
- Added unit tests for the hipdnn-integration tests to the CI testing workflows

## Testing strategy
- CI/CD Passes
- Make sure artifacts are built correctly


mousdahl-amd:
The rocblas test failures and the rocfft failure on Windows gfx 120X seem to be unrelated to this work.

ml-libs/CMakeLists.txt
```diff
@@ -284,6 +335,7 @@ if(THEROCK_ENABLE_HIPBLASLTPROVIDER)
     RUNTIME_DEPS
       hipBLAS-common
       hipBLASLt
+      hipdnn_integration_tests
```
adickin-amd, line 338: can you add runtime dep on hip-kernel-provider as well?  Then we can bring it in without needing rock change.

ml-libs/CMakeLists.txt
```diff
@@ -212,6 +212,56 @@ if(THEROCK_ENABLE_HIPDNN)
   )
 endif(THEROCK_ENABLE_HIPDNN)
 
+if(THEROCK_ENABLE_HIPDNN_INTEGRATION_TESTS)
+  ##############################################################################
+  # hipDNN integration tests
+  ##############################################################################
+
+  therock_cmake_subproject_declare(hipdnn_integration_tests
```
adickin-amd, line 220: Lets FF this just like this https://github.com/ROCm/TheRock/pull/4635

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -386,6 +386,18 @@ def _get_script_path(script_name: str) -> str:
             "windows": 1,
         },
     },
+    # hipDNN integration tests (unit tests for the integration test harness)
```
adickin-amd, line 389: will need to comment out this if we FF

BUILD_TOPOLOGY.toml
```diff
@@ -610,15 +610,20 @@ artifact_group = "ml-libs"
...
+[artifacts.hipdnn-integration-tests]
+artifact_group = "ml-libs"
+type = "target-neutral"
+artifact_deps = ["core-runtime", "core-hip", "hipdnn"]
+
 [artifacts.miopenprovider]
 artifact_group = "ml-libs"
 type = "target-neutral"
-artifact_deps = ["core-runtime", "core-hip", "miopen", "hipdnn"]
+artifact_deps = ["core-runtime", "core-hip", "miopen", "hipdnn", "hipdnn-integration-tests"]
```
adickin-amd, line 621: hmm, i wonder how we FF this

ml-libs/CMakeLists.txt
mousdahl-amd: I'll try that next if I can't get this in

geomin12 (commented):
should we just bundle this with hipdnn for cmake? instead of having an entirely new artifact for this

mousdahl-amd:
> should we just bundle this with hipdnn for cmake? instead of having an entirely new artifact for this

It's really a utility for evaluating and testing hipdnn engine plugins.  It doesn't contribute to the core functionality of hipdnn, so we didn't really think it belonged there.

It's also possible I'm not understanding what you're suggesting.  Is there a way to have do that if it's not under `rocm-libraries/projects/hipdnn`, or would we have to move it?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -401,7 +413,7 @@ def _get_script_path(script_name: str) -> str:
     # MIOpen provider tests
     "miopenprovider": {
         "job_name": "miopenprovider",
-        "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
+        "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --hipdnn-integration-tests --tests",
```
geomin12, line 416: i don't think these flags are needed, as miopenprovider test file only runs miopenprovider tests (and needs miopenprovider binaries)

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -412,7 +424,7 @@ def _get_script_path(script_name: str) -> str:
     },
     "fusilliprovider": {
         "job_name": "fusilliprovider",
-        "fetch_artifact_args": "--hipdnn --fusilliprovider --iree-compiler --tests",
+        "fetch_artifact_args": "--hipdnn --fusilliprovider --iree-compiler  --hipdnn-integration-tests --tests",
```
geomin12, line 427: i don't think these flags are needed, as fusilliprovider test file only runs fusilliprovider tests (and needs fusilliprovider binaries)

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -421,7 +433,7 @@ def _get_script_path(script_name: str) -> str:
     # hipBLASLt provider tests
     "hipblasltprovider": {
         "job_name": "hipblasltprovider",
-        "fetch_artifact_args": "--blas --hipdnn --hipblasltprovider --tests",
+        "fetch_artifact_args": "--blas --hipdnn --hipblasltprovider --hipdnn-integration-tests --tests",
```
geomin12, line 436: i don't think these flags are needed, as hipblasltprovider test file only runs hipblasltprovider tests (and needs hipblasltprovider binaries)

geomin12 (commented):
lgtm, just a few comments

geomin12 (approved):

build_tools/github_actions/fetch_test_configurations.py
mousdahl-amd: We're going to add consumption of these binaries in a follow-up PR in rocm-libraries.

build_tools/github_actions/fetch_test_configurations.py
mousdahl-amd: We're going to add consumption of these binaries in a follow-up PR in rocm-libraries.

build_tools/github_actions/fetch_test_configurations.py
mousdahl-amd: We're going to add consumption of these binaries in a follow-up PR in rocm-libraries.

PR #4356, opened 2026-04-06
Fix prebuilt stage missing RUNTIME_DEPS dependencies

### LINKED ISSUE #7690: Prebuilt subprojects ignore RUNTIME_DEPS ordering, silently staging an incomplete dist/
## Summary

When a subproject's stage directory is imported from a prebuilt artifact (a `<stage_dir>.prebuilt`
marker, as created by `artifact_manager.py fetch --bootstrap`), the generated stage rule copies from
its own stage dir **and from every `RUNTIME_DEPS` stage dir** — but it does not depend on those
runtime deps having been staged.

In a hybrid build, where some subprojects are prebuilt and their runtime deps are still built from
source, ninja is free to run that copy before the runtime deps exist. The resulting `dist/` silently
lacks the runtime libraries, and the rule exits 0.

## Root cause

`cmake/therock_subproject.cmake`, in the prebuilt branch of
`therock_cmake_subproject_activate()` (`if(EXISTS "${_prebuilt_file}")`,
[`therock_subproject.cmake#L956`](https://github.com/ROCm/TheRock/blob/main/cmake/therock_subproject.cmake#L956)):

```cmake
add_custom_command(
  OUTPUT "${_stage_stamp_file}"
  # Populate local dist directory with this+all transitive stage installs.
  COMMAND "${Python3_EXECUTABLE}" "${_fileset_tool}" copy ${_fileset_verbose_arg} "${_dist_dir}" ${_dist_source_dirs}
  COMMAND "${CMAKE_COMMAND}" -E touch "${_stage_stamp_file}"
  DEPENDS
    "${_prebuilt_file}"
    "${_fileset_tool}"
)
```

`_dist_source_dirs` comes from `_therock_cmake_subproject_get_stage_dirs(... ${_runtime_deps})`
(`therock_subproject.cmake:947`), so it is *this* target's stage dir plus **each `RUNTIME_DEPS`
stage dir**. The `DEPENDS` list does not mention them, so nothing orders the copy after those deps
are staged.

The non-prebuilt branch never had this problem, but only by accident of the stamp chain:
`stage.stamp` ← `build.stamp` ← `configure.stamp` ← `_configure_dep_stamps`, which includes the
runtime deps' `stage.stamp`. Marking a subproject prebuilt reduces `configure` and `build` to bare
`touch` commands, which deletes that chain — and with it the only thing that was sequencing the
runtime deps.

## Impact

Every staged multi-arch CI build is a hybrid build: `multi_arch_build_portable_linux_artifacts.yml`,
`multi_arch_build_windows_artifacts.yml`, and `multi_arch_build_wsl_rocdxg_artifacts.yml` all run
`artifact_manager.py fetch ... --bootstrap`, which writes `.prebuilt` markers for the inbound
artifacts of that stage while the stage's own subprojects build from source. The same applies to any
local hybrid build.

Because ninja is only *free* to misorder rather than forced to, a full `ninja` run usually gets away
with it and this surfaces as a nondeterministic missing-runtime-library failure. It becomes a
deterministic miss when a target is built that does not otherwise pull in the runtime deps.

The originally observed symptom: `hip-clr` prebuilt with `ROCR-Runtime` built from source produced a
`dist/` with no `libhsa-runtime64.so.1`.

## Reproduction

Environment: Linux, CMake 3.28.3, Ninja 1.11.1, TheRock at `fc11b46e0` (`main`), Python 3.12.

`therock-elfutils` is a convenient stand-in: it declares
`RUNTIME_DEPS therock-bzip2 therock-liblzma therock-zlib therock-zstd`, all of which are small, and
`-DTHEROCK_ENABLE_ALL=OFF` keeps the configure fast. The script fakes the artifact bootstrap by
writing a stage dir and the `.prebuilt` marker next to it, which is exactly what
`ArtifactPopulator` does in `artifact_manager.py`.

```bash
#!/usr/bin/env bash
set -euo pipefail

BUILD=build-repro
ELF=$BUILD/third-party/sysdeps/linux/elfutils/build

rm -rf "$BUILD"

# Simulate `artifact_manager.py fetch --bootstrap` for an inbound prebuilt artifact.
mkdir -p "$ELF/stage/lib/rocm_sysdeps/lib" "$ELF/stage/lib/rocm_sysdeps/include"
echo 'fake prebuilt libelf' > "$ELF/stage/lib/rocm_sysdeps/lib/libelf.so.1"
echo 'fake' > "$ELF/stage/lib/rocm_sysdeps/include/libelf.h"
touch "$ELF/stage.prebuilt"

cmake -B "$BUILD" -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx1100 -DTHEROCK_ENABLE_ALL=OFF > /dev/null

echo "== ninja inputs of the prebuilt stage.stamp =="
ninja -C "$BUILD" -t query third-party/sysdeps/linux/elfutils/build/stamp/stage.stamp \
  | sed -n '/input:/,/outputs:/p' | grep -v outputs:

echo "== ninja therock-elfutils+stage =="
ninja -C "$BUILD" therock-elfutils+stage

echo "== runtime-dep stage dirs (file counts) =="
for d in bzip2 liblzma zlib zstd; do
  printf '  %-8s %s\n' "$d" "$(find "$BUILD/third-party/sysdeps/linux/$d/build/stage" -type f 2>/dev/null | wc -l)"
done

echo "== resulting dist =="
find "$ELF/dist" -type f | sed "s|$ELF/dist/|  |" | sort
```

### Actual (on `main`)

```
== ninja inputs of the prebuilt stage.stamp ==
  input: CUSTOM_COMMAND
    third-party/sysdeps/linux/elfutils/build/stage.prebuilt
    .../build_tools/fileset_tool.py

== ninja therock-elfutils+stage ==
[1/1] Generating build/stamp/stage.stamp

== runtime-dep stage dirs (file counts) ==
  bzip2    0
  liblzma  0
  zlib     0
  zstd     0

== resulting dist ==
  lib/rocm_sysdeps/include/libelf.h
  lib/rocm_sysdeps/lib/libelf.so.1
```

The stage rule has no edge to any runtime dep, so none of the four were built, yet
`therock-elfutils+stage` succeeded and produced a `dist/` containing nothing but the prebuilt
artifact's own two files. `fileset_tool.py copy` does not complain about the missing source dirs, so
the failure is silent.

### Expected

The four runtime deps are staged first and their libraries land in `dist/`:

```
== ninja inputs of the prebuilt stage.stamp ==
  input: CUSTOM_COMMAND
    third-party/sysdeps/linux/elfutils/build/stage.prebuilt
    .../build_tools/fileset_tool.py
    third-party/sysdeps/linux/bzip2/build/stamp/stage.stamp
    third-party/sysdeps/linux/liblzma/build/stamp/stage.stamp
    third-party/sysdeps/linux/zlib/build/stamp/stage.stamp
    third-party/sysdeps/linux/zstd/build/stamp/stage.stamp

== ninja therock-elfutils+stage ==
[37/37] Generating build/stamp/stage.stamp

== runtime-dep stage dirs (file counts) ==
  bzip2    6
  liblzma  55
  zlib     13
  zstd     9

== resulting dist ==
  ...
  lib/rocm_sysdeps/lib/libelf.so.1
  lib/rocm_sysdeps/lib/librocm_sysdeps_bz2.so
  lib/rocm_sysdeps/lib/librocm_sysdeps_liblzma.so.5.8.1
  lib/rocm_sysdeps/lib/librocm_sysdeps_z.so.1.3.2
  lib/rocm_sysdeps/lib/librocm_sysdeps_zstd.so.1.5.7
  ...
```

## Fix

https://github.com/ROCm/TheRock/pull/4356 adds the `RUNTIME_DEPS` stage stamps to the prebuilt stage
command's `DEPENDS`. The "Expected" output above is that PR applied to `fc11b46e0`, using the same
script.

PeterCDMcLean:
## Motivation

Closes #7690.

When a subproject's stage directory is imported from a prebuilt artifact (a
`<stage_dir>.prebuilt` marker, as written by `artifact_manager.py fetch --bootstrap`), the generated
stage rule copies from its own stage dir **and from every `RUNTIME_DEPS` stage dir** — but it does
not depend on those runtime deps having been staged.

In a hybrid build, where some subprojects are prebuilt and their runtime deps are still built from
source, ninja is free to run that copy before the runtime deps exist. The resulting `dist/` silently
lacks the runtime libraries, and the rule exits 0. Originally hit with `hip-clr` prebuilt against a
source-built `ROCR-Runtime`: no `libhsa-runtime64.so.1` in `dist/`.

This is the normal shape of every staged multi-arch CI build —
`multi_arch_build_portable_linux_artifacts.yml`, `multi_arch_build_windows_artifacts.yml` and
`multi_arch_build_wsl_rocdxg_artifacts.yml` all fetch inbound artifacts with `--bootstrap` while the
stage's own subprojects build from source.

## Technical Details

The non-prebuilt branch never had this problem, but only by accident of the stamp chain:
`stage.stamp` ← `build.stamp` ← `configure.stamp` ← `_configure_dep_stamps`, which includes the
runtime deps' `stage.stamp`. Marking a subproject prebuilt reduces `configure` and `build` to bare
`touch` commands, which deletes that chain — and with it the only thing sequencing the runtime deps.

The fix adds the `RUNTIME_DEPS` stage stamps to the prebuilt stage command's `DEPENDS`, so the
`fileset_tool.py copy` cannot run until every directory it reads from has been staged.
`_configure_dep_stamps` moves into the non-prebuilt branch, where it is the only branch that uses it.

## Risk Assessment

**Risk level 2.** Superbuild-only change confined to the prebuilt branch of
`therock_cmake_subproject_activate()`; it adds dependency edges and removes none. Builds with no
prebuilt subprojects generate an identical graph. Builds with prebuilt subprojects gain ordering
edges they should already have had, which can only serialize work that was previously free to race —
no new cycles, since these are the same edges the non-prebuilt path already carries transitively.

## Related

- Work tracking / defect: #7690
- Predecessor: #7551 (`fix(bootstrap): write .prebuilt marker at the subproject stage dir`), which
  put the marker where this code path reads it.

None required. The change is architecture-independent: it only alters CMake dependency-graph
construction, and touches no kernel selection, default, or support surface. Passing PR CI is
sufficient.

## Test Plan

New regression test `build_tools/tests/therock_subproject_prebuilt_test.py`, picked up by the
existing `pytest build_tools/` lane in `unit_tests.yml` (Linux + Windows). It drives
`cmake/therock_subproject.cmake` through a two-subproject miniature superproject — `main_project`
declares `RUNTIME_DEPS dep_project`, and `main_project`'s stage is imported from a `.prebuilt`
marker the way `ArtifactPopulator` writes one under `--bootstrap`. The harness needs no submodules,
no compiler and no bundled sysdeps, and follows the pattern already used by
`build_tools/tests/rocm_build_flags_test.py`.

- `test_prebuilt_subproject_dist_contains_runtime_deps` — the regression: `main_project`'s dist must
  contain `lib/dep.txt`.
- `test_built_subproject_dist_contains_runtime_deps` — control: same assertion for a normally-built
  `main_project`, so the two cases differ only by the prebuilt marker.

Separately verified against real subprojects: `therock-elfutils` marked prebuilt with its
`RUNTIME_DEPS therock-bzip2/liblzma/zlib/zstd` built from source. Full script and both outputs are
in #7690.

## Testing Checklist

- [x] Regression test, fix reverted - `git revert --no-commit HEAD && pytest tests/therock_subproject_prebuilt_test.py` - Status: Failed as expected (`['lib/main.txt'] != ['lib/dep.txt', 'lib/main.txt']`)
- [x] Regression test, fix applied - `cd build_tools && pytest tests/therock_subproject_prebuilt_test.py` - Status: Passed (2 passed in 0.88s)
- [x] Full unit-test lane, locally (Linux) - `cd build_tools && pytest` - Status: Passed (2063 passed, 8 skipped)
- [x] Unit Tests :: ubuntu-24.04 and windows-2022 - GitHub PR checks - Status: Passed (new test green on both; 5-8s per case on Windows)
- [x] `therock-elfutils` end-to-end repro, before/after - Status: Passed (see #7690)
- [x] pre-commit - `pre-commit run --files build_tools/tests/therock_subproject_prebuilt_test.py` - Status: Passed

## Test Result

With the fix reverted, `test_prebuilt_subproject_dist_contains_runtime_deps` fails and the control
passes — the two differ only by the `.prebuilt` marker. With the fix, both pass. The full
`build_tools` suite is green.

On real subprojects, `ninja therock-elfutils+stage` went from one edge and a two-file `dist/` (none
of the four runtime deps built, exit 0) to staging all four deps and landing
`librocm_sysdeps_{bz2,liblzma,z,zstd}.so` in `dist/`.

## Flags / Guardrails

None. The change is unconditional and has no runtime behavior to gate.

## Adjacent Tests Considered

`build_tools/tests/configure_stage_test.py` and `build_topology_test.py` cover which artifacts a
stage declares prebuilt, but not the resulting build graph, so they are unaffected. The Consumer
Graph Drift Check is the other consumer of subproject dependency declarations; it was green on the
fix commit and should stay green — this change adds no edges to the consumer graph, only to the
ninja graph.

## Submission Checklist

### DIFF
--- a/build_tools/tests/therock_subproject_prebuilt_test.py
+++ b/build_tools/tests/therock_subproject_prebuilt_test.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Tests that a prebuilt subproject stage respects RUNTIME_DEPS ordering.
+
+A subproject whose stage directory is imported from an artifact (marked by a
+`<stage_dir>.prebuilt` file) still copies its RUNTIME_DEPS stage directories
+into its own dist directory. Those deps may be built from source in the same
+build -- that is the normal shape of a staged CI build, where
+`artifact_manager.py fetch --bootstrap` marks the inbound artifacts prebuilt
+while the stage's own subprojects build normally.
+
+See https://github.com/ROCm/TheRock/issues/7690.
+
+These tests drive `cmake/therock_subproject.cmake` through a miniature
+superproject rather than TheRock itself, so they need no submodules, no
+compiler, and no bundled sysdeps.
+"""
+
+import subprocess
+import tempfile
+import textwrap
+import unittest
+from pathlib import Path
+
+THEROCK_ROOT = Path(__file__).resolve().parents[2]
+
+# The subset of TheRock's CMake modules that therock_subproject.cmake needs to
+# be usable outside the superproject.
+HARNESS_INCLUDES = (
+    "therock_globals",
+    "therock_sanitizers",
+    "therock_flag_utils",
+    "therock_default_targets",
+    "therock_subproject",
+)
+
+
+def write_file(path: Path, contents: str) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(textwrap.dedent(contents), encoding="utf-8")
+
+
+def write_harness(source_dir: Path) -> None:
+    """Writes a two-subproject superproject into `source_dir`.
+
+    `main_project` declares `RUNTIME_DEPS dep_project`, so its dist directory
+    must contain `lib/dep.txt` (installed by `dep_project`) in addition to
+    whatever is in its own stage directory.
+    """
+    includes = "\n".join(f"include({name})" for name in HARNESS_INCLUDES)
+    write_file(
+        source_dir / "CMakeLists.txt",
+        f"""
+        cmake_minimum_required(VERSION 3.25)
+        project(therock_subproject_prebuilt_harness NONE)
+
+        set(THEROCK_SOURCE_DIR "{THEROCK_ROOT.as_posix()}")
+        set(THEROCK_BINARY_DIR "${{CMAKE_BINARY_DIR}}")
+        list(APPEND CMAKE_MODULE_PATH "${{THEROCK_SOURCE_DIR}}/cmake")
+        find_package(Python3 COMPONENTS Interpreter REQUIRED)
+
+        # therock_subproject.cmake fingerprints this file. The harness declares
+        # no flags, so an empty state file is enough.
+        set(ROCM_BUILD_FLAGS_STATE_FILE "${{CMAKE_BINARY_DIR}}/rocm_build_flags_state.cmake")
+        file(WRITE "${{ROCM_BUILD_FLAGS_STATE_FILE}}" "# no flags\\n")
+
+        {includes}
+
+        therock_cmake_subproject_declare(dep_project
+          EXTERNAL_SOURCE_DIR "${{CMAKE_CURRENT_SOURCE_DIR}}/dep"
+          BINARY_DIR "${{CMAKE_CURRENT_BINARY_DIR}}/dep"
+        )
+        therock_cmake_subproject_activate(dep_project)
+
+        therock_cmake_subproject_declare(main_project
+          EXTERNAL_SOURCE_DIR "${{CMAKE_CURRENT_SOURCE_DIR}}/main"
+          BINARY_DIR "${{CMAKE_CURRENT_BINARY_DIR}}/main"
+          RUNTIME_DEPS dep_project
+        )
+        therock_cmake_subproject_activate(main_project)
+        """,
+    )
+    write_file(
+        source_dir / "dep" / "CMakeLists.txt",
+        """
+        cmake_minimum_required(VERSION 3.25)
+        project(dep_project NONE)
+        file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/dep.txt" "dep\\n")
+        install(FILES "${CMAKE_CURRENT_BINARY_DIR}/dep.txt" DESTINATION lib)
+        """,
+    )
+    write_file(
+        source_dir / "main" / "CMakeLists.txt",
+        """
+        cmake_minimum_required(VERSION 3.25)
+        project(main_project NONE)
+        file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/main.txt" "main\\n")
+        install(FILES "${CMAKE_CURRENT_BINARY_DIR}/main.txt" DESTINATION lib)
+        """,
+    )
+
+
+def mark_prebuilt(build_dir: Path) -> None:
+    """Imports `main_project`'s stage from an "artifact".
+
+    This mirrors what `ArtifactPopulator` in `build_tools/artifact_manager.py`
+    does under `--bootstrap`: populate the stage directory and drop a
+    `<stage_dir>.prebuilt` marker beside it, before CMake configures.
+    """
+    stage_dir = build_dir / "main" / "stage"
+    write_file(stage_dir / "lib" / "main.txt", "main\n")
+    (build_dir / "main" / "stage.prebuilt").touch()
+
+
+def run(*args: str) -> subprocess.CompletedProcess[str]:
+    result = subprocess.run(
+        args,
+        check=False,
+        text=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+    if result.returncode != 0:
+        raise AssertionError(" ".join(args) + "\n" + result.stdout)
+    return result
+
+
+class PrebuiltRuntimeDepsTest(unittest.TestCase):
+    def _stage_main_project(self, *, prebuilt: bool) -> list[str]:
+        """Builds `main_project+stage` and returns its dist contents."""
+        with tempfile.TemporaryDirectory() as temp_dir_str:
+            temp_dir = Path(temp_dir_str)
+            source_dir = temp_dir / "source"
+            build_dir = temp_dir / "build"
+            write_harness(source_dir)
+            if prebuilt:
+                mark_prebuilt(build_dir)
+
+            run("cmake", "-S", str(source_dir), "-B", str(build_dir), "-GNinja")
+            run("cmake", "--build", str(build_dir), "--target", "main_project+stage")
+
+            dist_dir = build_dir / "main" / "dist"
+            return sorted(
+                p.relative_to(dist_dir).as_posix()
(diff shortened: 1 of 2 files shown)


build_tools/tests/therock_subproject_prebuilt_test.py
```diff
@@ -0,0 +1,171 @@
...
+          BINARY_DIR "${{CMAKE_CURRENT_BINARY_DIR}}/dep"
+        )
+        therock_cmake_subproject_activate(dep_project)
+
+        therock_cmake_subproject_declare(main_project
+          EXTERNAL_SOURCE_DIR "${{CMAKE_CURRENT_SOURCE_DIR}}/main"
+          BINARY_DIR "${{CMAKE_CURRENT_BINARY_DIR}}/main"
+          RUNTIME_DEPS dep_project
+        )
+        therock_cmake_subproject_activate(main_project)
```
ScottTodd, line 82: This style of unit testing the CMake functions is a good idea. I think it's fine to land in its current form, but I'd like to follow up with:
* Backfill more general tests for `therock_cmake_subproject_*`
* Move this and the recently added https://github.com/ROCm/TheRock/blob/main/build_tools/tests/rocm_build_flags_test.py next to each other, either via just naming `cmake_*_test.py` or a dedicated tests subdirectory for CMake under:
    * `build_tools/tests/cmake/`
    * `cmake/tests/`
    * `build_tools/cmake/tests/` (we could also move scripts like https://github.com/ROCm/TheRock/blob/main/build_tools/configure_stage.py under `build_tools/cmake/` if we do this)
* Update TESTING.md to officially recognize these tests here: https://github.com/ROCm/TheRock/blob/main/TESTING.md#super-project-cmake-build---validation-methods, updating this comment:
    > We are evaluating adding unit tests for certain features of the CMake
    build system itself, see https://github.com/ROCm/TheRock/pull/6984 for
    example.

    (that PR was merged, see the test file linked above)

ScottTodd (approved):

Issue #4353, opened 2026-04-06
[Issue] Linux PyTorch wheel build fails linking C++ tests: `libtorch_hip.so` undefined reference to `rsmi_init`


chiranjeevipattigidi:
## Summary
The PyTorch wheel build in CI is failing at link time when building multiple C++ test executables. The linker reports an unresolved symbol `rsmi_init` coming from `libtorch_hip.so`, which indicates ROCm SMI (librocm_smi64 / rocm-smi) is not being linked into the final link step.

**Pytorch Version:** `2.8, 2.9, 2.10, 2.11`

## Error log:

```
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: /__w/TheRock/TheRock/external-builds/pytorch/pytorch/build/lib/libtorch_hip.so: undefined reference to `rsmi_init'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

## Full log:
- PR: https://github.com/ROCm/TheRock/pull/4324 (not actually merged, but pytorch build failure still here)
  - https://github.com/ROCm/TheRock/actions/runs/23928422073/job/69806057474
  - https://github.com/ROCm/TheRock/actions/runs/23928422073/job/69806057491
  - https://github.com/ROCm/TheRock/actions/runs/23928422073/job/69806057497
  - https://github.com/ROCm/TheRock/actions/runs/23928422073/job/69806057459

edit 2026-04-06: @amd-justchen - Adding associated PR (was closed) to originally provided logs
edit: originally was a comment here https://github.com/ROCm/TheRock/issues/4311#issuecomment-4182872495

amd-chiranjeevi:
regression after the merge of rocm-systems bump into the rock [Bump rocm-systems from b9e258c to 478a7a4](https://github.com/ROCm/TheRock/commit/3bacd1eb36cf7041081579cc410fbe579f89f0df)

ScottTodd:
https://github.com/ROCm/rocm-systems/pull/3777 looks relevant. cc @oliveiradan @marifamd @gabrpham

amd-aakash:
@amd-justchen Can we test for a revert of this PR and PyTorch build ?

amd-justchen:
[...]

PR #4357, opened 2026-04-06
Revert "Bump rocm-systems from b9e258c to 478a7a4"

amd-justchen:
Reverts ROCm/TheRock#4337

Due to possible regression as seen in CI nightly pytorch builds after bump, test if revert will workaround. Issue noted here https://github.com/ROCm/TheRock/issues/4353.


amd-justchen:
> Try reverting just the suspected culprit commit, not the full rocm-systems submodule update. We can already see from CI history that the submodule update broke the build.

https://github.com/ROCm/TheRock/issues/4353#issuecomment-4193951225
Also realizing this, that we basically know it's the submodule bump, but have to do a revert on the specific commit in rocm-systems

PR #4361, opened 2026-04-06
[ci] Updating submodules to run only on M-F

geomin12:
Getting PR reviews for weekend submodule updates, when no one really observes / checks over weekend in regards to submodule bumps


geomin12:
fair point! closing then

PR #4362, opened 2026-04-06
[ci] Adding option `all` for build stages in workflow_dispatch

geomin12:
During workflow_dispatch runs, it took some time to have to type out all the stages needed for multi-arch. Now, you can specify `all` and get all the packages (parsed from BUILD_TOPOLOGY)

Test run here: https://github.com/ROCm/TheRock/actions/runs/24158875337

### DIFF
--- a/build_tools/_therock_utils/build_topology.py
+++ b/build_tools/_therock_utils/build_topology.py
@@ -14,18 +14,7 @@ from typing import Dict, List, Optional, Set
 
 
 def get_topology(topology_path: Optional[Path] = None) -> "BuildTopology":
-    """Load the BUILD_TOPOLOGY.toml from the default or specified location.
-
-    Args:
-        topology_path: Path to topology file. If None, uses default location
-                      (BUILD_TOPOLOGY.toml in repository root).
-
-    Returns:
-        BuildTopology instance loaded from the specified or default path.
-
-    Raises:
-        FileNotFoundError: If the topology file doesn't exist.
-    """
+    """Load the BUILD_TOPOLOGY.toml from the default or specified location."""
     if topology_path is None:
         # Default: BUILD_TOPOLOGY.toml in repository root
         # This module is at build_tools/_therock_utils/build_topology.py


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -71,6 +91,21 @@ def _parse_comma_list(raw: str) -> list[str]:
...
+    Reads stage names from BUILD_TOPOLOGY.toml when 'all' is specified.
+
+    Example:
+        "foundation,compiler-runtime" → ["foundation", "compiler-runtime"]
+        "all" → ["foundation", "compiler-runtime", "math-libs", ...]
+    """
+    stages = _parse_comma_list(raw)
+    if "all" in stages:
+        return _get_all_build_stages()
+    return stages
```
ScottTodd, line 102: I also want this code to expand the DAG so you can say "compiler-runtime" and that will expand to ["foundation", "compiler-runtime"]. Fine to start like this.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -63,6 +68,21 @@
...
+def _get_all_build_stages() -> list[str]:
+    """Get all build stage names from BUILD_TOPOLOGY.toml.
+
+    Returns:
+        List of stage names in the order they appear in the TOML file
+    """
+    # Path to BUILD_TOPOLOGY.toml relative to this script
+    script_dir = Path(__file__).resolve().parent
+    therock_dir = script_dir.parent.parent
+    topology_path = therock_dir / "BUILD_TOPOLOGY.toml"
```
ScottTodd: Let's add a factory method that loads from the default topology file (I'm not aware of any alternate file locations outside of tests, so being explicit about it like this isn't useful).

Can use that from these other files too:

* https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/build_tools/artifact_manager.py#L108-L125
* https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/build_tools/configure_stage.py#L51-L58
* https://github.com/ROCm/TheRock/blob/a3f13f8c054424ead7c7da52a215d22c0d636155/build_tools/buildctl.py#L236-L247

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: ah smart! can do this in next PR

geomin12:
re-ran tests, works: https://github.com/ROCm/TheRock/actions/runs/24094283978

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Each of those sources can pull from the common factory method. The scripts should NOT import each other like this:

```python
# Add parent directory to path for configure_stage imports
sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
from configure_stage import get_topology
```

build_tools/_therock_utils/build_topology.py
```diff
@@ -13,6 +13,32 @@
...
+        # Default: BUILD_TOPOLOGY.toml in repository root
+        # This module is at build_tools/_therock_utils/build_topology.py
+        script_dir = Path(__file__).parent
+        repo_root = script_dir.parent.parent
+        topology_path = repo_root / "BUILD_TOPOLOGY.toml"
+
+    if not topology_path.exists():
+        raise FileNotFoundError(f"BUILD_TOPOLOGY.toml not found at {topology_path}")
+
+    return BuildTopology(str(topology_path))
```
ScottTodd, line 39: We can optimize this later with a process-wide singleton instead of always constructing a new instance (which loads a file and does some parsing).

.github/workflows/multi_arch_ci.yml
```diff
@@ -37,7 +37,7 @@ on:
       prebuilt_stages:
         type: string
         default: ""
-        description: "Comma-separated build stages to skip; artifacts are copied from baseline_run_id instead"
+        description: "Comma-separated build stages to skip (or 'all' for all stages); artifacts are copied from baseline_run_id instead"
```
ScottTodd, line 40: Ah, we could also document this here: https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#prebuilt-stages-multi-arch-ci

PR #4366, opened 2026-04-07
Add multi-arch JAX release workflow integration (tarball-based)

### LINKED ISSUE #3878: [CI] Build and test JAX Python packages as part of ci.yml

JAX references:

PyTorch code to reference:

### LINKED ISSUE #4367: [CI] Enable JAX builds on bump PR workflows
### Background

TheRock currently supports JAX builds via:

- `.github/workflows/build_linux_jax_wheels.yml`

However, JAX is **not integrated into CI bump PR workflows**, which means:

- Bump PRs (ROCm updates, dependency bumps, etc.) do not validate JAX
- JAX regressions can slip through CI

erman-gurses:
# Motivation

Introduce JAX as a downstream consumer of TheRock multi-arch release artifacts.

This change integrates JAX wheel builds into the multi-arch release pipeline using ROCm tarballs produced within the same release workflow. The goal is to align JAX with the same artifact-driven release model already used by PyTorch.

Instead of relying on the older standalone JAX workflows, JAX now participates directly in the multi-arch release graph.

This PR covers the JAX release build path only.

A prerequisite ROCm/rockrel PR adds the forwarding workflow required by the `workflow-dispatch` architecture used by multi-arch release workflows.

# Multi-arch JAX release workflow integration

Add:

```text
.github/workflows/multi_arch_release_linux_jax_wheels.yml
```

This workflow is the JAX counterpart to:

```text
.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```

The workflow:

* builds JAX wheels per `(python_version, jax_ref)`
* consumes ROCm tarballs generated earlier in the same release pipeline
* builds against the shared multi-arch TheRock tarball
* supports:

  * `rocm-jaxlib-v0.8.2`
  * `rocm-jaxlib-v0.9.0`
  * `rocm-jaxlib-v0.9.1`
  * `rocm-jaxlib-v0.9.2`
* follows the existing multi-arch release workflow structure used by PyTorch

This workflow currently performs build-only validation. Testing and promotion remain follow-up work.

# Tarball URL propagation

Extend:

```text
.github/workflows/multi_arch_build_tarballs.yml
```

to export:

```text
tarball_urls
```

The current consumer is the JAX release workflow, which consumes the shared multi-arch tarball URL through workflow outputs.

Tarball URLs are now constructed directly from:

* `dest.bucket`
* `dest.relative_path`

without reparsing S3 URIs.

Unit tests were added to validate:

* shared tarball URL generation
* compatibility with future archive naming changes discussed in #4438

# Multi-arch release pipeline wiring

Update:

```text
.github/workflows/multi_arch_release_linux.yml
```

to dispatch:

```text
multi_arch_release_linux_jax_wheels.yml
```

from the multi-arch Linux release graph.

Flow:

```text
build_artifacts
    ↓
build_tarballs
    ↓
trigger_release_jax_wheels
```

The JAX workflow consumes the shared multi-arch tarball URL exported by:

```text
build_tools/github_actions/upload_tarballs.py
```

# Validation

Validated successfully with:

## Python

* 3.11
* 3.12
* 3.13
* 3.14

## JAX refs

* rocm-jaxlib-v0.8.2
* rocm-jaxlib-v0.9.0
* rocm-jaxlib-v0.9.1
* rocm-jaxlib-v0.9.2

# Test plan
- Land the prerequisite `ROCm/rockrel` stub workflow first, since the multi-arch release flow uses `workflow-dispatch` and requires a corresponding workflow to exist in `rockrel`.
- Manually dispatch the multi-arch JAX release workflow through `rockrel` and verify that `release_type`, `rocm_version`, `tar_url`, and `ref` are forwarded correctly into `ROCm/TheRock`.
- Verify the workflow builds the expected legacy JAX release branches:
  - `rocm-jaxlib-v0.8.2`
  - `rocm-jaxlib-v0.9.0`
  - `rocm-jaxlib-v0.9.1`
  - `rocm-jaxlib-v0.9.2`
- Verify the workflow consumes the shared multi-arch TheRock tarball produced earlier in the release pipeline.
- Verify the generated run-name uses the generic `Multiarch` label rather than expanded gfx target lists, consistent with the architecture-agnostic nature of the legacy `rocm-jax` release branches.

# Results

In Progress  - `ROCm/rockrel` stub workflow need to be landed first https://github.com/ROCm/rockrel/pull/50

# Follow-up work

Tracked separately:

* Add wrapper workflow as `multi_arch_build_linux_jax_wheels.yml` and JAX wheel testing
* staging bucket upload and promotion
* scheduled/nightly release automation
* migration to ROCm multi-arch Python package consumption (ROCm/jax#781)
* CI-specific JAX build/test gating

## Prerequisite

* ROCm/rockrel PR: *Add stub workflow for multi-arch JAX release dispatch* https://github.com/ROCm/rockrel/pull/50  (must land before this PR)


.github/workflows/ci.yml
```diff
@@ -111,6 +111,7 @@ jobs:
       test_type: ${{ needs.setup.outputs.test_type }}
       sanity_check_only_for_family: ${{ matrix.variant.sanity_check_only_for_family == true }}
       build_pytorch: ${{ matrix.variant.build_pytorch == true }}
+      build_jax: ${{ matrix.variant.build_jax == true }}
```
ScottTodd: ci.yml is scheduled for deletion (https://github.com/ROCm/TheRock/issues/3340). All new feature work must be in multi-arch CI.

build_tools/github_actions/configure_ci.py
ScottTodd, line 1: No new feature work for "CI", only multi-arch CI: https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci.py

ScottTodd (changes requested):

.github/workflows/ci.yml
erman-gurses: Thank you for the clarification -  will update the code.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd, line 25: There are comments up here that will need to be updated, like removing "(future)" from the build-jax line

ScottTodd, line 298: Another comment to update

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -574,12 +578,20 @@ def decide_jobs(
...
         test_rocm=test_rocm,
         build_rocm_python=JobGroupDecision(action=JobAction.RUN),
         build_pytorch=JobGroupDecision(action=JobAction.RUN),
         test_pytorch=JobGroupDecision(action=JobAction.RUN),
+        build_jax=JobGroupDecision(
+            action=JobAction.RUN if enable_jax else JobAction.SKIP
+        ),
+        test_jax=JobGroupDecision(
+            action=JobAction.RUN if enable_jax else JobAction.SKIP
+        ),
```
ScottTodd: > ### Scope
> 
> This PR only introduces control-plane plumbing. Follow-up PRs will:
> 
> * produce JAX-consumable ROCm artifacts
> * invoke the JAX workflow 
> * add validation/tests

The plumbing here looks correct, but I'd prefer to adjust the sequencing so people don't start expecting this to actually do anything (though to that point, we could also remove `test_pytorch` 🤔)

1. Produce ROCm tarballs (I'm sort of adding this as part of https://github.com/ROCm/TheRock/issues/3334), OR switch JAX to build from ROCm python packages
2. Add JAX building to multi-arch CI workflows, together with this plumbing (they can be separate PRs, but I'd get them up for review together so the period between advertising that it is available and it actually working is short)
3. Add JAX testing to multi-arch CI workflows

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -574,12 +578,20 @@ def decide_jobs(
     # Other jobs run unconditionally with no configuration.
     # TODO: job pruning: skip pytorch if only JAX has been edited, etc.
 
+    enable_jax = ci_inputs.is_pull_request and "ci:build-jax" in ci_inputs.pr_labels
```
ScottTodd: Labels should be documented in https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: I expect I'll have ROCm tarballs produced by multi-arch CI workflows within a few days, so this can probably compose pretty easily on top of that 🤞

erman-gurses: That makes sense, thanks.

I’ll wait for (1) to land on your side, then follow up with a PR for (2). We can review this plumbing PR together with the JAX build integration PR to minimize the gap between enabling it and it actually working.

After that, I’ll proceed with (3).

Will address rest of the comments today.

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: Updated

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: Updated

erman-gurses: Updated

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: @ScottTodd, I combined (1) and (2) because once tarball production was available, the remaining JAX build integration became the direct consumer of the new control-plane outputs (`build_jax`, `jax_amdgpu_family`). 

Keeping them separate would leave a no-op state where CI exposes JAX enablement but does not execute any JAX workflow. Combining them makes the path coherent end-to-end, while still keeping the scope limited to build-only execution.

.github/workflows/multi_arch_ci.yml
```diff
@@ -90,6 +90,45 @@ jobs:
...
+      }}
+    uses: ./.github/workflows/multi_arch_build_tarballs.yml
+    secrets: inherit
+    with:
+      artifact_run_id: ${{ github.run_id }}
+      artifact_github_repo: ${{ github.repository }}
+      dist_amdgpu_families: ${{ fromJSON(needs.setup.outputs.linux_build_config).dist_amdgpu_families }}
+      platform: linux
+      package_version: ${{ needs.setup.outputs.rocm_package_version }}
+      release_type: ${{ inputs.release_type || 'dev' }}
```
ScottTodd: Empty string is meaningful here. Only pass "dev" as the release type if actually triggered by a release workflow, otherwise let the default empty string --> "ci" through. See https://github.com/ROCm/TheRock/blob/407bf1b66a2c42f0351b172baa9f618a36da7277/.github/workflows/multi_arch_build_tarballs.yml#L30-L33

build_tools/github_actions/generate_tarball_urls.py
```diff
@@ -0,0 +1,151 @@
...
+    key: str,
+    expires_in: int,
+) -> str:
+    # Fail fast if the uploaded object is not present where we expect it.
+    s3_client.head_object(Bucket=bucket, Key=key)
+    return s3_client.generate_presigned_url(
+        "get_object",
+        Params={"Bucket": bucket, "Key": key},
+        ExpiresIn=expires_in,
+    )
```
ScottTodd: Don't generate a presigned URL. CI buckets are world readable. If release buckets are not readable then don't use S3, use the CDN (cloudfront) URLs.

ScottTodd (commented):
drive-by review pointing out some details that matter significantly (please be careful testing such changes)

erman-gurses:
@ScottTodd, Those comments are helpful thanks. Not tested anything yet - will do after the fixes.

.github/workflows/multi_arch_ci.yml
erman-gurses: Updated

build_tools/github_actions/generate_tarball_urls.py
erman-gurses: Updated

erman-gurses:
@ScottTodd, let me know if you still see any critical issue for the testing.

.github/workflows/build_linux_jax_from_tarball.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Generate tarball download URL
+        id: compute_tarball_url
+        run: |
+          python build_tools/github_actions/generate_tarball_urls.py \
+            --run-id "${{ inputs.artifact_run_id }}" \
+            --platform linux \
+            --release-type "${{ inputs.release_type }}" \
+            --package-version "${{ inputs.package_version }}" \
+            --dist-amdgpu-families "${{ inputs.dist_amdgpu_families }}" \
+            --family "${{ inputs.amdgpu_family }}"
```
ScottTodd: You can pass the URL in to the workflow instead of computing here.

The workflow then can build from any tarball you provide, instead of needing to infer where a tarball _should_ be. If we change the structure of the tarballs or their index pages then this workflow code would break. If a developer wants to build from a dev or release tarball this would not support that.

See how we pass `needs.build_python_packages.outputs.package_find_links_url` and `inputs.rocm_package_version` here:
* https://github.com/ROCm/TheRock/blob/8552a89ae1c25051e32a001dd152e0d9c70c0bd4/.github/workflows/multi_arch_ci_linux.yml#L197-L245
* https://github.com/ROCm/TheRock/blob/8552a89ae1c25051e32a001dd152e0d9c70c0bd4/.github/workflows/build_portable_linux_python_packages.yml#L71-L74
* https://github.com/ROCm/TheRock/blob/8552a89ae1c25051e32a001dd152e0d9c70c0bd4/.github/workflows/build_portable_linux_python_packages.yml#L94-L95
* https://github.com/ROCm/TheRock/blob/8552a89ae1c25051e32a001dd152e0d9c70c0bd4/build_tools/github_actions/upload_python_packages.py#L282-L286

We can set a similar output here:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_tarballs.yml
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/upload_tarballs.py

.github/workflows/build_linux_jax_from_tarball.yml
```diff
@@ -0,0 +1,69 @@
...
+    with:
+      amdgpu_family: ${{ inputs.amdgpu_family }}
+      python_version: "3.12"
+      release_type: ${{ inputs.release_type }}
+      s3_subdir: "v2"
+      s3_staging_subdir: "v2-staging"
+      rocm_version: ${{ inputs.package_version }}
+      tar_url: ${{ needs.generate_tarball_url.outputs.tarball_url }}
+      cloudfront_url: "https://rocm.devreleases.amd.com/v2"
+      cloudfront_staging_url: "https://rocm.devreleases.amd.com/v2-staging"
```
ScottTodd: These URLs should be similarly dynamic.

Some references:
* https://github.com/ROCm/TheRock/blob/main/build_tools/_therock_utils/s3_buckets.py

Looks like https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_linux_jax_wheels.yml is only using `package_index_url: ${{ inputs.cloudfront_staging_url }}` but then it _always_ copies to a release bucket.

We could add a variant similar to https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels_ci.yml that doesn't upload at all (or uploads to the artifacts buckets, not a release bucket), then a release workflow like https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml that uploads

erman-gurses:
Test results:
https://github.com/ROCm/TheRock/actions/runs/25845705421/job/75985177943
https://github.com/ROCm/TheRock/actions/runs/25859429065

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,6 +1,23 @@
...
+# Multi-arch release variant of the JAX wheel build flow.
+#
+# Built shape:
+#   - One build/ci_build invocation per (python_version, jax_ref)
+#     matrix cell.
+#   - Builds directly against the shared TheRock multiarch tarball
+#     passed in via inputs.tar_url.
+#
+# Triggered via workflow_call from
+# multi_arch_release_linux.yml's build_jax_wheels job.
```
ScottTodd: This comment is not accurate - the workflow is triggered using workflow_dispatch

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -200,4 +200,22 @@ jobs:
               "ref": "${{ inputs.ref || '' }}"
             }
 
-  # TODO: trigger_build_jax_wheels (release_portable_linux_jax_wheels.yml or equivalent)
+  build_jax_wheels:
```
ScottTodd: Rename to trigger_release_jax_wheels (I just renamed the others in https://github.com/ROCm/TheRock/pull/5212)

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -56,46 +82,86 @@ permissions:
...
+          - jax_ref: "rocm-jaxlib-v0.8.2"
+            build_jaxlib: true
+          - jax_ref: "rocm-jaxlib-v0.9.0"
+            build_jaxlib: true
+          - jax_ref: "rocm-jaxlib-v0.9.1"
+            build_jaxlib: false
+    permissions:
+      contents: read
+      id-token: write
+      packages: read
```
ScottTodd: Why `packages: read`? We aren't using github packages so this shouldn't be needed?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -651,13 +655,21 @@ def decide_jobs(
 
     # Other jobs run unconditionally with no configuration.
     # TODO: job pruning: skip pytorch if only JAX has been edited, etc.
-
+    enable_jax = ci_inputs.event_name == "workflow_dispatch" or (
+        ci_inputs.is_pull_request and "ci:build-jax" in ci_inputs.pr_labels
+    )
```
ScottTodd: There is a `ci_inputs.is_workflow_dispatch` property you can use instead of inlining the string comparison here

```suggestion
    enable_jax = ci_inputs.is_workflow_dispatch or (
        ci_inputs.is_pull_request and "ci:build-jax" in ci_inputs.pr_labels
    )
```

It looks like you want this to only run on `workflow_dispatch` - what about `schedule` for nightly releases? Future PR?

build_tools/github_actions/upload_tarballs.py
```diff
@@ -99,6 +109,25 @@ def main(argv: list[str]) -> int:
...
+        ):
+            raise ValueError(f"Unexpected tarball name: {name}")
+
+        family_and_version = name[
+            len(f"therock-dist-{args.platform}-") : -len(".tar.gz")
+        ]
+        family = family_and_version.rsplit("-", 1)[0]
+        tarball_urls[family] = f"https://{bucket}.s3.amazonaws.com/{prefix}/{name}"
+
+    gha_set_output({"tarball_urls": json.dumps(tarball_urls)})
```
ScottTodd: What is this parsing trying to do? Some of this data already exists in a usable format and parsing like this is going to be brittle. For example, I don't see any reason to include and run the `parse_s3_uri()` function when you already have `dest.bucket`, `dest.relative_path`, and the list of files

It looks like all that this `tarballs_urls` is currently used for is `"tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",`

Note that the "multiarch" part of the file name will probably be removed entirely, per https://github.com/ROCm/TheRock/pull/4438 . This logic is probably worth adding a unit test for, to make sure it continues to work as we adjust file names.

Outputting the list of URLs and then using them like this is probably fine, I just want to make sure that the URL construction is robust to future refactoring.

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -200,4 +200,22 @@ jobs:
...
+    needs: [build_artifacts, build_tarballs]
+    name: Trigger Release JAX Wheels
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: write
+    steps:
+      - name: Trigger JAX Build
+        uses: benc-uk/workflow-dispatch@7a027648b88c2413826b6ddd6c76114894dc5ec4 # v1.3.1
+        with:
+          workflow: multi_arch_release_linux_jax_wheels.yml
```
ScottTodd, line 252: Let's add a wrapper workflow like https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml (in that repository) so this workflow-dispatch call works when run from nightly releases in https://github.com/ROCm/rockrel

If this code is opt-in via workflow_dispatch then we could land first in TheRock and later add the wrapper workflow (though workflow_dispatch triggers over there are used for dev and prereleases also)

.github/workflows/setup_multi_arch.yml
```diff
@@ -78,6 +78,9 @@ on:
       rocm_package_version:
         description: ROCm package version (primarily for Python packages).
         value: ${{ jobs.setup.outputs.rocm_package_version }}
+      build_jax:
+        description: Whether to run JAX build jobs.
+        value: ${{ jobs.setup.outputs.build_jax }}
```
ScottTodd: It looks like this PR contains changes from two currently disconnected chunks of work:
* setup_multi_arch.yml, configure_multi_arch_ci.py
* multi_arch_release_linux.yml, multi_arch_release_linux_jax_wheels.yml

The `build_jax` output here is not used by any workflows, so it is necessary in this PR?

Release workflows currently unconditionally build all package types, CI workflows are where jobs are conditional conditionally selected. If this PR is just adding the release workflow code then could the CI configuration changes be moved to a separate PR?

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: Updated as the suggested code. 
Plan to do with incremental steps. I will have follow-up PRs for the schedule for nightly releases

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Yes, I can add wrapper workflow at the follow up

.github/workflows/setup_multi_arch.yml
erman-gurses: It is because of my confusion. This PR should only focus on release workflow integration. Will Update the code.

build_tools/github_actions/upload_tarballs.py
erman-gurses: Updated

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Updated

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Updated

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: True, updated

erman-gurses:
Test running currently: https://github.com/ROCm/TheRock/actions/runs/25929965604

ScottTodd (commented):
Please get a code review from another developer before me. I can't be the first pass reviewer on so many PRs.

erman-gurses:
@WBobby, can you please take a look before @ScottTodd's final review.
Latest test is in here:
https://github.com/ROCm/TheRock/actions/runs/26051928050

WBobby:
Blocking: Matrix is missing  rocm-jaxlib-v0.9.2  (added to release_portable_linux_jax_wheels.yml  by #5094 on 2026-05-13).

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -38,64 +62,106 @@ on:
...
-
           python3 build/ci_build \
             --compiler=clang \
             --python-versions="${{ matrix.python_version }}" \
             --rocm-version="${ROCM_VERSION}" \
             --therock-path="${{ inputs.tar_url }}" \
+            ${SOURCE_ARG} \
             dist_wheels
+
+      - name: Extract JAX versions from built wheels
```
WBobby, line 180: `Extract JAX versions` step has no `id:` and no job outputs — the script's outputs are discarded.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
WBobby: `inputs.amdgpu_families` is `required: true` but only used in `run-name`. Consider `required: false` with a comment, since JAX builds use the multiarch tarball and don't take family targets.

build_tools/github_actions/upload_tarballs.py
```diff
@@ -99,6 +101,28 @@ def main(argv: list[str]) -> int:
...
+        url = f"https://{dest.bucket}.s3.amazonaws.com/{dest.relative_path}/{name}"
+
+        if name.startswith(f"therock-dist-{args.platform}-multiarch-"):
+            tarball_urls["multiarch"] = url
+        else:
+            family_and_version = name[
+                len(f"therock-dist-{args.platform}-") : -len(".tar.gz")
+            ]
+
+            # Future-proof shared tarball naming:
```
WBobby: future-proof branch should cite #4438

WBobby:
**Non-blocking:**
PR description should list follow-up TODO links.

**For discussion (not blocking):** 
The new file's 8 build steps are essentially a copy of `build_linux_jax_wheels.yml`'s build job with 0 lines of net-new logic. Same pattern as PyTorch's two parallel yml files. The 0.9.2 miss suggests JAX's faster version cadence may eventually warrant a wrapper refactor or shared matrix source — separate conversation.

erman-gurses:
> Blocking: Matrix is missing rocm-jaxlib-v0.9.2 (added to release_portable_linux_jax_wheels.yml by #5094 on 2026-05-13).

Added rocm-jaxlib-v0.9.2

erman-gurses:
@WBobby , latest tests after the comments were addressed:
https://github.com/ROCm/TheRock/actions/runs/26148631112/job/76956608213
https://github.com/ROCm/TheRock/actions/runs/26162202587

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,101 +1,186 @@
...
+            build_jaxlib: false
+          - jax_ref: "rocm-jaxlib-v0.9.2"
+            build_jaxlib: false
 
-    runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
+    permissions:
+      contents: read
+      id-token: write
+
+    outputs:
```
WBobby: Suggested change (Delete the 4-line outputs: block)
```
    permissions:
      contents: read
      id-token: write

    # Intentionally no job-level `outputs:` here: build_jax_wheels is a
    # matrix job, and matrix cells collide on shared job output keys.
    # The Extract JAX versions step below is retained for per-cell logs
    # and as a sanity check that required wheels were produced. When a
    # downstream consumer is added (test/promote), follow the
    # multi_arch_release_linux_pytorch_wheels.yml pattern and upload
    # inside this same matrix job rather than passing values through
    # job-level outputs.
```

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -200,4 +200,22 @@ jobs:
...
+          inputs: |
+            {
+              "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
+              "release_type": "${{ inputs.release_type }}",
+              "rocm_version": "${{ inputs.rocm_package_version }}",
+              "tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",
```
WBobby, line 257: Fail fast when the multiarch tarball is missing.

fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch silently renders as an empty string if upload_tarballs.py didn't classify any tarball as multiarch. The dispatched JAX workflow then receives tar_url="" and crashes deep inside ci_build with a confusing error.

Please add a guard step before the dispatch so we get an actionable failure on the orchestrator job:

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,101 +1,186 @@
...
+        description: Branch, tag, or SHA to checkout.
         type: string
         default: ""
 
 permissions:
   contents: read
   id-token: write
-  packages: read
 
 run-name: Multi-Arch Release Linux JAX Wheels (${{ inputs.release_type }}, ${{ inputs.rocm_version }}, ${{ inputs.amdgpu_families }})
```
WBobby: Drop amdgpu_families from run-name now that it's informational only.
The input was just downgraded to a documentation-only field defaulting to "multiarch". Showing it in the run title makes it look like a real build axis. Cleaner:
```suggestion
run-name: Multi-Arch Release Linux JAX Wheels (${{ inputs.release_type }}, ${{ inputs.rocm_version }})
```

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,101 +1,186 @@
...
             --python-versions="${{ matrix.python_version }}" \
             --rocm-version="${ROCM_VERSION}" \
             --therock-path="${{ inputs.tar_url }}" \
+            ${SOURCE_ARG} \
             dist_wheels
+
+      - name: Extract JAX versions from built wheels
+        id: write_jax_versions
+        run: |
+          python3 ./build_tools/github_actions/write_jax_versions.py \
```
WBobby, line 183: `PACKAGE_DIST_DIR` is defined but not actually consumed.

```suggestion
          python3 ./build_tools/github_actions/write_jax_versions.py \
            --dist-dir "${{ env.PACKAGE_DIST_DIR }}"
```

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,101 +1,186 @@
...
         type: string
-        default: "gfx94X-dcgpu"
+        default: "multiarch"
+        required: false
       release_type:
         description: 'Release type. All developer-triggered jobs should use "dev"!'
         type: choice
         options:
           - dev
         default: dev
```
WBobby: workflow_dispatch.inputs.release_type is a choice restricted to dev, but trigger_release_jax_wheels in multi_arch_release_linux.yml forwards inputs.release_type from the parent, which can be nightly or prerelease.

```suggestion
      release_type:
        description: 'Release type: "dev", "nightly", or "prerelease". Selects the S3 bucket.'
        type: string
        required: true
```

WBobby (changes requested):
Reminder on my earlier non-blocking note: please expand the PR Motivation with explicit links/issues for the follow-up items already promised in this thread — the `rockrel` wrapper workflow, the `schedule:` trigger for nightly, the JAX test job, the staging-bucket upload, and the carved-out `build_jax` CI gating. The current *"Testing and promotion flows will be added later"* doesn't tell readers what's pending or where to follow it.

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Added into `upload_tarballs.py`

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Ok, I think this should stay as informational. Please see this comment here:
https://github.com/ROCm/TheRock/pull/5127#discussion_r3220907300

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Nice catch, thanks

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Updated, thanks

erman-gurses:
Latest tests: https://github.com/ROCm/TheRock/actions/runs/26232036015

WBobby:
**File:** `build_tools/github_actions/tests/upload_tarballs_test.py`

### What's wrong

Lines 20–23 only add `build_tools/` to `sys.path`, then `import github_actions.upload_tarballs as mod`. Loading that module runs `from github_actions_api import gha_set_output`, which requires `build_tools/github_actions/` on `sys.path` — never added here.

### Why it matters

The test currently passes only because pytest collects `tests/upload_python_packages_test.py` first, and that file inserts `build_tools/github_actions/` into `sys.path`. Run alone from the same cwd CI uses (`build_tools/`) and it fails:

```
$ python -m pytest github_actions/tests/upload_tarballs_test.py
github_actions/upload_tarballs.py:43: in <module>
    from github_actions_api import gha_set_output
E   ModuleNotFoundError: No module named 'github_actions_api'
```

It will also break under collection reordering (e.g. `pytest-randomly`) or `--import-mode=importlib`.

### How to fix

Match the pattern used by sibling tests (`upload_python_packages_test.py`, `post_build_upload_test.py`):

```suggestion
# Add build_tools to path so _therock_utils is importable.
sys.path.insert(0, os.fspath(Path(__file__).parent.parent.parent))
# Add github_actions to path so upload_tarballs is importable.
sys.path.insert(0, os.fspath(Path(__file__).parent.parent))

import upload_tarballs as mod
```

Update the six `@patch("github_actions.upload_tarballs.…")` decorators (3 per test method) to `@patch("upload_tarballs.…")`.

erman-gurses:
> **File:** `build_tools/github_actions/tests/upload_tarballs_test.py`
> 
> ### What's wrong
> Lines 20–23 only add `build_tools/` to `sys.path`, then `import github_actions.upload_tarballs as mod`. Loading that module runs `from github_actions_api import gha_set_output`, which requires `build_tools/github_actions/` on `sys.path` — never added here.
> 
> ### Why it matters
> The test currently passes only because pytest collects `tests/upload_python_packages_test.py` first, and that file inserts `build_tools/github_actions/` into `sys.path`. Run alone from the same cwd CI uses (`build_tools/`) and it fails:
> 
> ```
> $ python -m pytest github_actions/tests/upload_tarballs_test.py
> github_actions/upload_tarballs.py:43: in <module>
>     from github_actions_api import gha_set_output
> E   ModuleNotFoundError: No module named 'github_actions_api'
> ```
> 
> It will also break under collection reordering (e.g. `pytest-randomly`) or `--import-mode=importlib`.
> 
> ### How to fix
> Match the pattern used by sibling tests (`upload_python_packages_test.py`, `post_build_upload_test.py`):
> 
> ```
> # Add build_tools to path so _therock_utils is importable.
> sys.path.insert(0, os.fspath(Path(__file__).parent.parent.parent))
> # Add github_actions to path so upload_tarballs is importable.
> sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
> 
> import upload_tarballs as mod
> ```
> 
> Update the six `@patch("github_actions.upload_tarballs.…")` decorators (3 per test method) to `@patch("upload_tarballs.…")`.

Addressed the comment

erman-gurses:
Test: https://github.com/ROCm/TheRock/actions/runs/26493773169
https://github.com/ROCm/TheRock/actions/runs/26518582682
CC: @WBobby

build_tools/github_actions/upload_tarballs.py
```diff
@@ -99,6 +101,33 @@ def main(argv: list[str]) -> int:
...
+            if "-" not in family_and_version:
+                tarball_urls["multiarch"] = url
+            else:
+                family = family_and_version.rsplit("-", 1)[0]
+                tarball_urls[family] = url
+
+    if not tarball_urls.get("multiarch"):
+        raise ValueError(
+            "multiarch tarball URL was not produced; check tarball naming and upload logic"
+        )
```
WBobby: Blocking — this fail-fast is unconditional and breaks single-family / non-KPACK callers.

Move the assertion to the **consumer** side instead。

WBobby (changes requested):
Reminder (still pending from my earlier note): the PR Motivation has not been updated — please add explicit follow-up links/issues for the `rockrel` wrapper workflow, the `schedule:` trigger for nightly, the JAX test job, the staging-bucket upload, and the carved-out `build_jax` CI gating before merge.

erman-gurses:
> ## Follow-up tracking

Added  Follow-up tracking

build_tools/github_actions/upload_tarballs.py
erman-gurses: Good catch  -  updating

erman-gurses: Just addressed that

erman-gurses:
Latest Test: https://github.com/ROCm/TheRock/actions/runs/26639793144
https://github.com/ROCm/TheRock/actions/runs/26639793144/job/78542081133

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -200,4 +200,33 @@ jobs:
...
+  validate_jax_tarball:
+    needs: [build_tarballs]
+    name: Validate JAX Multiarch Tarball
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Missing multiarch tarball
+        if: ${{ !fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}
+        run: |
+          echo "Expected multiarch tarball URL was not produced for JAX release workflow"
+          exit 1
```
WBobby: **Blocking — skip propagation regression: JAX dispatch will never fire on the happy path.**

`validate_jax_tarball` is gated by `if: ${{ !fromJSON(...).multiarch }}`, so on the healthy case (multiarch URL present) the job is skipped. `trigger_release_jax_wheels` then has `needs: [..., validate_jax_tarball]` with default `if: success()`, and per [GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) a skipped need causes downstream jobs to also be skipped. Net effect: JAX is dispatched **never**.

Please drop `validate_jax_tarball` entirely and inline the check as the first step of `trigger_release_jax_wheels`:

```suggestion
  trigger_release_jax_wheels:
    needs: [build_artifacts, build_tarballs]
    name: Trigger Release JAX Wheels
    runs-on: ubuntu-24.04
    permissions:
      actions: write
    steps:
      - name: Verify multiarch tarball URL
        env:
          MULTIARCH_URL: ${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}
        run: |
          if [ -z "${MULTIARCH_URL}" ]; then
            echo "::error::No 'multiarch' entry in build_tarballs.outputs.tarball_urls; refusing to dispatch JAX build."
            exit 1
          fi

      - name: Trigger JAX Build
        uses: benc-uk/workflow-dispatch@7a027648b88c2413826b6ddd6c76114894dc5ec4 # v1.3.1
        with:
          workflow: multi_arch_release_linux_jax_wheels.yml
          inputs: |
            {
              "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
              "release_type": "${{ inputs.release_type }}",
              "rocm_version": "${{ inputs.rocm_package_version }}",
              "tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",
              "ref": "${{ inputs.ref || '' }}"
            }
```

Inlining avoids the cross-job skip-propagation hazard entirely (bash `exit 1` → step fail → job fail → downstream skip is GitHub Actions' core, well-tested behavior), saves a runner spin-up, and removes the need for a workflow-level dry-run test.

WBobby (changes requested):

erman-gurses:
Test: https://github.com/ROCm/TheRock/actions/runs/26659497017

.github/workflows/multi_arch_release_linux.yml
erman-gurses: The was the older version that I already updated.  But I used your version on the last commit called `[Update tarball check](https://github.com/ROCm/TheRock/pull/4366/commits/8161d2f6b42b710e7cb3f7183cd2086b285962bd)`
This was the run from my version: https://github.com/ROCm/TheRock/actions/runs/26649039750

erman-gurses:
Ping @WBobby

WBobby (approved):
LGTM.

.github/workflows/multi_arch_release_linux.yml
ScottTodd: It needs to be added _before_ merging this PR, or else nightly releases will fail.

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -200,4 +200,30 @@ jobs:
...
+      actions: write
+    steps:
+      - name: Verify multiarch tarball URL
+        env:
+          MULTIARCH_URL: ${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}
+        run: |
+          if [ -z "${MULTIARCH_URL}" ]; then
+            echo "::error::No 'multiarch' entry in build_tarballs.outputs.tarball_urls; refusing to dispatch JAX build."
+            exit 1
+          fi
```
ScottTodd: This step doesn't seem to be adding any value. The output should always be set, right? Not generating a URL should be a failure in the `build_tarballs` job, not this job.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
ScottTodd: Huh? If an input is unused, remove it, don't just change it to do nothing...

This doesn't make sense to me though. Does JAX discover enabled amdgpu targets from the tarball itself somehow? It still should need to know what it is building for.

This test run: https://github.com/ROCm/TheRock/actions/runs/26659497017 built for all of `gfx94X-dcgpu, gfx110X-all, gfx1151, gfx120X-all, gfx950-dcgpu, gfx900, gfx906, gfx908, gfx90a, gfx101X-dgpu, gfx103X-all, gfx1150, gfx1152, gfx1153` but the JAX wheels build has a different target list: https://github.com/ROCm/TheRock/actions/runs/26666509772/job/78600922339#step:8:186
`#10 [stage-0  4/10] RUN printf '%s\n' > /opt/rocm/bin/target.lst gfx908 gfx90a gfx942 gfx950 gfx1030 gfx1100 gfx1101 gfx1151 gfx1200 gfx1201`

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,101 +1,187 @@
...
         working-directory: rocm-jax
         run: |
-
           python3 build/ci_build \
             --compiler=clang \
             --python-versions="${{ matrix.python_version }}" \
             --rocm-version="${ROCM_VERSION}" \
             --therock-path="${{ inputs.tar_url }}" \
+            ${SOURCE_ARG} \
             dist_wheels
```
ScottTodd, line 178: Something is seriously off here for `rocm-jaxlib-v0.9.2`. See the build logs at https://github.com/ROCm/TheRock/actions/runs/26666509772/job/78600922304#step:8:188

```
#8 [stage-0 3/8] RUN --mount=type=cache,mode=0755,target=/root/.cache/pip     /opt/python/cp312-cp312/bin/python3 -m pip install         --pre --index-url "https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/"         "rocm[libraries,devel]" &&     ln -sf /opt/python/cp312-cp312/bin/rocm-sdk /usr/local/bin/rocm-sdk &&     rocm-sdk init &&     ln -sfn "$(rocm-sdk path --root)" /opt/rocm
#8 1.793 Looking in indexes: https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/
#8 1.914 Collecting rocm[devel,libraries]
```

The build should be using _only_ what it is instructed to use (multi-arch rocm tarballs in this case), yet this is downloading from the nightly gfx950 single-arch python packages?)

later in the logs: https://github.com/ROCm/TheRock/actions/runs/26666509772/job/78600922304#step:8:13521
```
2026-05-29 23:15:56,302 - DEBUG - ROCm tookit path: /opt/_internal/cpython-3.12.13/lib/python3.12/site-packages/_rocm_sdk_devel
2026-05-29 23:15:56,302 - DEBUG - ROCm AMD GPU targets: gfx908,gfx90a,gfx9-4-generic,gfx10-3-generic,gfx11-generic,gfx12-generic
2026-05-29 23:15:56,302 - DEBUG - Additional Bazel build options: ['--define=xla_rocm_profiler=v1', '--action_env=HIPCC_COMPILE_FLAGS_APPEND=--offload-compress']
2026-05-29 23:15:56,302 - INFO - Bazel options written to .jax_configure.bazelrc
2026-05-29 23:15:56,302 - DEBUG - Artifacts output directory: /jax_rocm_plugin/dist_manylinux
2026-05-29 23:15:56,302 - INFO - Building jax-rocm-pjrt for linux x86_64...
2026-05-29 23:15:56,302 - INFO - [EXECUTING] ./bazel-7.7.0-linux-x86_64 run --repo_env=HERMETIC_PYTHON_VERSION=3.11 --verbose_failures=true --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang-18" --repo_env=CC="/usr/lib/llvm-18/bin/clang-18" --repo_env=BAZEL_COMPILER="/usr/lib/llvm-18/bin/clang-18" --config=clang --define=xnn_enable_avxvnniint8=false --config=mkl_open_source_only --config=avx_posix --config=rocm_base --config=rocm --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang-18" --action_env=ROCM_PATH="/opt/_internal/cpython-3.12.13/lib/python3.12/site-packages/_rocm_sdk_devel" --action_env=TF_ROCM_AMDGPU_TARGETS=gfx908,gfx90a,gfx9-4-generic,gfx10-3-generic,gfx11-generic,gfx12-generic --define=xla_rocm_profiler=v1 --action_env=HIPCC_COMPILE_FLAGS_APPEND=--offload-compress //pjrt/tools:build_gpu_plugin_wheel -- --output_path="/jax_rocm_plugin/dist_manylinux" --cpu=x86_64 --enable-rocm=True --platform_version=7 --rocm_jax_git_hash=1117030257c2684f6b7017b43b50e57751091755
```

We shouldn't be mixing multiple rocm installs (e.g. single-arch python packages with multi-arch tarballs, each from different versions).

Ideally JAX would build against ROCm multi-arch python packages, as torch does, see https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#building-python-based-projects . I see https://github.com/ROCm/jax/pull/781 from @gulsumgudukbay . Are you coordinating?

build_tools/github_actions/upload_tarballs.py
ScottTodd: This is a little better now but I'm still confused. File name parsing is brittle and we're going to make multiple changes to the tarball/archive names soon (see the linked RFC).

Repeating myself: what is this parsing trying to do? Do we need to support anything beyond that?

ScottTodd (commented):
Please reformat the PR description. There's too much going on there with confusing formatting and I can't easily see important information (test plan? test results? link to associated tracking issue?)

.github/workflows/multi_arch_release_linux_jax_wheels.yml
ScottTodd: Huh, that's at least not a regression, here's a single-arch release: https://github.com/ROCm/TheRock/actions/runs/26948280288/job/79520205806#step:8:242

```
  python3 build/ci_build \
    --compiler=clang \
    --python-versions="3.13" \
    --rocm-version="${ROCM_VERSION}" \
    --therock-path="https://rocm.nightlies.amd.com/tarball/therock-dist-linux-gfx110X-all-7.14.0a20260604.tar.gz" \
    $SOURCE_ARG \
    dist_wheels

...
#8 [stage-0 3/8] RUN --mount=type=cache,mode=0755,target=/root/.cache/pip     /opt/python/cp312-cp312/bin/python3 -m pip install         --pre --index-url "https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/"         "rocm[libraries,devel]" &&     ln -sf /opt/python/cp312-cp312/bin/rocm-sdk /usr/local/bin/rocm-sdk &&     rocm-sdk init &&     ln -sfn "$(rocm-sdk path --root)" /opt/rocm
```

Very strange. We say to build for Python 3.13 using gfx110X packages and it just decides to install Python 3.12 gfx950 packages? I don't think that `build/ci_build` script is something we should be using if the behavior is so opaque, brittle, and broken.

In any case, if that isn't a regression with this PR we can treat it as separable and non-blocking here.

gulsumgudukbay: > Something is seriously off here for `rocm-jaxlib-v0.9.2`. See the build logs at https://github.com/ROCm/TheRock/actions/runs/26666509772/job/78600922304#step:8:188
> 
> ```
> #8 [stage-0 3/8] RUN --mount=type=cache,mode=0755,target=/root/.cache/pip     /opt/python/cp312-cp312/bin/python3 -m pip install         --pre --index-url "https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/"         "rocm[libraries,devel]" &&     ln -sf /opt/python/cp312-cp312/bin/rocm-sdk /usr/local/bin/rocm-sdk &&     rocm-sdk init &&     ln -sfn "$(rocm-sdk path --root)" /opt/rocm
> #8 1.793 Looking in indexes: https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/
> #8 1.914 Collecting rocm[devel,libraries]
> ```
> 
> The build should be using _only_ what it is instructed to use (multi-arch rocm tarballs in this case), yet this is downloading from the nightly gfx950 single-arch python packages?)
> 
> later in the logs: https://github.com/ROCm/TheRock/actions/runs/26666509772/job/78600922304#step:8:13521
> 
> ```
> 2026-05-29 23:15:56,302 - DEBUG - ROCm tookit path: /opt/_internal/cpython-3.12.13/lib/python3.12/site-packages/_rocm_sdk_devel
> 2026-05-29 23:15:56,302 - DEBUG - ROCm AMD GPU targets: gfx908,gfx90a,gfx9-4-generic,gfx10-3-generic,gfx11-generic,gfx12-generic
> 2026-05-29 23:15:56,302 - DEBUG - Additional Bazel build options: ['--define=xla_rocm_profiler=v1', '--action_env=HIPCC_COMPILE_FLAGS_APPEND=--offload-compress']
> 2026-05-29 23:15:56,302 - INFO - Bazel options written to .jax_configure.bazelrc
> 2026-05-29 23:15:56,302 - DEBUG - Artifacts output directory: /jax_rocm_plugin/dist_manylinux
> 2026-05-29 23:15:56,302 - INFO - Building jax-rocm-pjrt for linux x86_64...
> 2026-05-29 23:15:56,302 - INFO - [EXECUTING] ./bazel-7.7.0-linux-x86_64 run --repo_env=HERMETIC_PYTHON_VERSION=3.11 --verbose_failures=true --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang-18" --repo_env=CC="/usr/lib/llvm-18/bin/clang-18" --repo_env=BAZEL_COMPILER="/usr/lib/llvm-18/bin/clang-18" --config=clang --define=xnn_enable_avxvnniint8=false --config=mkl_open_source_only --config=avx_posix --config=rocm_base --config=rocm --action_env=CLANG_COMPILER_PATH="/usr/lib/llvm-18/bin/clang-18" --action_env=ROCM_PATH="/opt/_internal/cpython-3.12.13/lib/python3.12/site-packages/_rocm_sdk_devel" --action_env=TF_ROCM_AMDGPU_TARGETS=gfx908,gfx90a,gfx9-4-generic,gfx10-3-generic,gfx11-generic,gfx12-generic --define=xla_rocm_profiler=v1 --action_env=HIPCC_COMPILE_FLAGS_APPEND=--offload-compress //pjrt/tools:build_gpu_plugin_wheel -- --output_path="/jax_rocm_plugin/dist_manylinux" --cpu=x86_64 --enable-rocm=True --platform_version=7 --rocm_jax_git_hash=1117030257c2684f6b7017b43b50e57751091755
> ```
> 
> We shouldn't be mixing multiple rocm installs (e.g. single-arch python packages with multi-arch tarballs, each from different versions).
> 
> Ideally JAX would build against ROCm multi-arch python packages, as torch does, see https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#building-python-based-projects . I see [ROCm/jax#781](https://github.com/ROCm/jax/pull/781) from @gulsumgudukbay . Are you coordinating?

Hi @ScottTodd this is correct. we are targeting multi-arch packages as well in JAX. That PR will be merged soon. I'm mainly taking into account the latest TheRock wheels to test that PR in our continuous CI workflows. 
cc: @erman-gurses

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Removed this part:
 ```
      - name: Verify multiarch tarball URL
        env:
          MULTIARCH_URL: ${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}
        run: |
          if [ -z "${MULTIARCH_URL}" ]; then
            echo "::error::No 'multiarch' entry in build_tarballs.outputs.tarball_urls; refusing to dispatch JAX build."
            exit 1
          fi
```

Regarding the first comment:
https://github.com/ROCm/TheRock/pull/4366#discussion_r3358022534

This has been added in the latest revision via trigger_release_jax_wheels in multi_arch_release_linux.yml. Were you referring to an additional change needed in rockrel, or was the concern that JAX had not yet been wired into the multi-arch release workflow? Can you please elobrate?

File: .github/workflows/multi_arch_release_linux.yml [hunk 26]
```diff
@@ -200,4 +200,22 @@ jobs:
...
+        uses: benc-uk/workflow-dispatch@7a027648b88c2413826b6ddd6c76114894dc5ec4 # v1.3.1
+        with:
+          workflow: multi_arch_release_linux_jax_wheels.yml
+          inputs: |
+            {
+              "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
+              "release_type": "${{ inputs.release_type }}",
+              "rocm_version": "${{ inputs.rocm_package_version }}",
+              "tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",
+              "ref": "${{ inputs.ref || '' }}"
```
ScottTodd, line 259: This will also need a `repository` input, see my other comment

.github/workflows/multi_arch_release_linux.yml
ScottTodd: > This has been added in the latest revision via trigger_release_jax_wheels in multi_arch_release_linux.yml. Were you referring to an additional change needed in rockrel, or was the concern that JAX had not yet been wired into the multi-arch release workflow? Can you please elobrate?

See https://github.com/ROCm/TheRock/issues/5647 for this exact issue with another workflow. The stub workflow needs to be added to https://github.com/ROCm/rockrel _before_ any code in TheRock's release workflows tries to use workflow_dispatch.

The sequencing must be:
1. Create a stub file for the workflow that is going to be used with workflow-dispatch in rockrel
2. Land the release workflow code in therock that has `uses: benc-uk/workflow-dispatch`

See for example https://github.com/ROCm/rockrel/pull/44 and https://github.com/ROCm/TheRock/pull/5212, how I labeled the first as a prerequisite for the second in "motivation", and how I cross-referenced the first in the second under "technical details"

As for why this is set up this way, we should add some docs/diagrams explaining the architecture... It's roughly:
1. The entry point https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release.yml workflow runs in rockrel with `uses: ROCm/TheRock/.github/workflows/multi_arch_release.yml@main`. That uses the _code_ from TheRock (currently with a floating pin to `main` but we could pin to a specific ref)
2. When that _code_ runs `uses: benc-uk/workflow-dispatch`, the triggered workflow should also run _in rockrel_ (https://github.com/ROCm/rockrel/actions), _not in therock_ (https://github.com/ROCm/TheRock/actions). The default `repo` behavior of that action (https://github.com/benc-uk/workflow-dispatch#repo) is what we want - if we wanted to trigger in another repo then we'd need to specify it and provide a token.
3. For that triggered workflow to run, it needs a file in https://github.com/ROCm/rockrel/tree/main/.github/workflows. We use stub workflows like https://github.com/ROCm/rockrel/blob/main/.github/workflows/test_artifacts.yml that just forward their inputs through to e.g. `uses: ROCm/TheRock/.github/workflows/test_artifacts.yml@main`
4. We pass `repository` and `ref` to each workflow so rockrel can clarify that it wants the actions to checkout `ROCm/TheRock` and some specific ref (branch/tag/etc.) of TheRock, _not_ the repository/ref of the entry point workflow (ROCm/rockrel + some commit there), see https://github.com/ROCm/rockrel/blob/f4db554dd500f9b4e44db33412eb1099e36e80ed/.github/workflows/multi_arch_release.yml#L61

Alternate designs are possible, and we can add unit tests and ref pinning across workflows to make such changes easier/safer.

erman-gurses:
Cannot reply directly this comment on the GitHub so I put my answer here:
https://github.com/ROCm/TheRock/pull/4366#discussion_r3358079445

This is old code - updated - I've simplified the implementation to stop parsing family names from filenames and to construct URLs directly from dest.bucket and dest.relative_path. The output contract is now just the shared tarball URL exposed as multiarch, which avoids depending on the current archive naming scheme and should be more resilient to the naming changes proposed in #4438.

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Updated - it does not show here as outdated I do not know why
Update as below:
```
          inputs: |
            {
              "tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",
              "repository": "${{ inputs.repository || github.repository }}",
              "ref": "${{ inputs.ref || '' }}"
            }
```

.github/workflows/multi_arch_release_linux_jax_wheels.yml
gulsumgudukbay: > Huh, that's at least not a regression, here's a single-arch release: https://github.com/ROCm/TheRock/actions/runs/26948280288/job/79520205806#step:8:242
> 
> ```
>   python3 build/ci_build \
>     --compiler=clang \
>     --python-versions="3.13" \
>     --rocm-version="${ROCM_VERSION}" \
>     --therock-path="https://rocm.nightlies.amd.com/tarball/therock-dist-linux-gfx110X-all-7.14.0a20260604.tar.gz" \
>     $SOURCE_ARG \
>     dist_wheels
> 
> ...
> #8 [stage-0 3/8] RUN --mount=type=cache,mode=0755,target=/root/.cache/pip     /opt/python/cp312-cp312/bin/python3 -m pip install         --pre --index-url "https://rocm.nightlies.amd.com/v2/gfx950-dcgpu/"         "rocm[libraries,devel]" &&     ln -sf /opt/python/cp312-cp312/bin/rocm-sdk /usr/local/bin/rocm-sdk &&     rocm-sdk init &&     ln -sfn "$(rocm-sdk path --root)" /opt/rocm
> ```
> 
> Very strange. We say to build for Python 3.13 using gfx110X packages and it just decides to install Python 3.12 gfx950 packages? I don't think that `build/ci_build` script is something we should be using if the behavior is so opaque, brittle, and broken.
> 
> In any case, if that isn't a regression with this PR we can treat it as separable and non-blocking here.

cc: @magaonka-amd  For the 0.9.2 issue

build_tools/github_actions/tests/upload_tarballs_test.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class TestUploadTarballsMain(unittest.TestCase):
+    @patch("upload_tarballs.gha_set_output")
+    @patch("upload_tarballs.create_storage_backend")
+    @patch("upload_tarballs.WorkflowOutputRoot")
+    @patch("upload_tarballs.logger")
+    def test_main_exports_multiarch_url(
+        self,
+        mock_logger,
```
ScottTodd: Why are you mocking `logger` here?

Generally, logging is not part of the API contract for a function and is not something that needs to be unit tested.

build_tools/github_actions/tests/upload_tarballs_test.py
```diff
@@ -0,0 +1,137 @@
...
+            dest = types.SimpleNamespace(
+                bucket="therock-dev-artifacts",
+                relative_path="25834210506-linux/tarballs",
+                s3_uri="s3://therock-dev-artifacts/25834210506-linux/tarballs",
+            )
+
+            mock_workflow_output_root.from_workflow_run.return_value.tarballs.return_value = (
+                dest
+            )
+            mock_create_storage_backend.return_value.upload_directory.return_value = 1
```
ScottTodd: Huh? The mocking strategy for these unit tests does not make sense. If you mock every function that the file calls then there's no real code for the unit tests left to exercise and you're just testing the implementation instead of the interface.

See tests for similar files like https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/upload_python_packages_test.py and https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/post_stage_upload_test.py

When I added `upload_tarballs.py` in https://github.com/ROCm/TheRock/pull/4448 I skipped unit tests for it since the file was simple - just argparse --> output root --> upload. If that's changing in this PR you should restructure the script so it is easier to write unit tests for, similar to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/upload_python_packages.py
* Pull logic out of `main()` and into a `run()` or other function
* Test that function, with minimal/no mocks

build_tools/github_actions/upload_tarballs.py
```diff
@@ -99,6 +105,42 @@ def main(argv: list[str]) -> int:
...
+
+    for f in tarball_files:
+        name = f.name
+
+        if name.startswith(f"therock-dist-{args.platform}-multiarch-"):
+            shared_tarball_url = _build_s3_url(
+                dest.bucket,
+                dest.relative_path,
+                name,
+            )
```
ScottTodd: Still not using the APIs as designed here.

Don't reimplement what is already part of the API:
```python
def _build_s3_url(bucket: str, relative_path: str, filename: str) -> str:
    return f"https://{bucket}.s3.amazonaws.com/{relative_path}/{filename}"
```

There should be no `.s3.amazonaws.com` mentions outside of https://github.com/ROCm/TheRock/blob/main/build_tools/_therock_utils/storage_location.py. It's an implementation detail of S3 storage, which could be substituted for some other storage (local, another cloud provider, etc.)

1. Add a function here to get a storage location for a specific tarball filename:
https://github.com/ROCm/TheRock/blob/60da64b656140498a9bfeca27e2b575e6c45a786/build_tools/_therock_utils/workflow_outputs.py#L246-L250
2. Call that function with `name` to get a `StorageLocation`
3. Call `https_url()` on that `StorageLocation`: https://github.com/ROCm/TheRock/blob/60da64b656140498a9bfeca27e2b575e6c45a786/build_tools/_therock_utils/storage_location.py#L47-L50

build_tools/github_actions/tests/upload_tarballs_test.py
erman-gurses: Updated

build_tools/github_actions/tests/upload_tarballs_test.py
erman-gurses: I see - I refactored the script to move the behavior into a run() function and updated the tests to call run() directly with minimal mocking. The tests now exercise the real tarball discovery and URL generation paths instead of mocking WorkflowOutputRoot, create_storage_backend, and logger.

build_tools/github_actions/upload_tarballs.py
erman-gurses: Addressed. I added WorkflowOutputRoot.tarball() and updated upload_tarballs.py to use output_root.tarball(name).https_url. The local _build_s3_url() helper was removed, so URL construction now goes through the existing storage abstraction instead of hardcoding S3-specific behavior.

ScottTodd (approved):
Thanks, I support landing this to start producing JAX packages in the multi-arch release pipelines, even if they aren't yet building quite how we expect. We can iterate on the nightly releases ahead of the next stable release. JAX packages have been one of the last missing pieces to turn off the old single-arch release workflows and delete associated code (https://github.com/ROCm/TheRock/issues/3340)

PR #4370, opened 2026-04-07
[Tensilelite] Add TensileLite Python test component

archana-ramalingam:
## Motivation

Adds a tensilelite test component that runs `TensileLite` and `rocisa` Python unit tests via pytest. This is the first Python-environment-based test in the math libraries CI pipeline.

## Technical Details

  - Stage Python source tree for pytest-based CI testing. `TheRock` artifact system bundles these into `share/tensilelite-tests/` so that `test_tensilelite.py` can run `uv run pytest` from the unpacked artifact without requiring a `rocm-libraries` source checkout. 

  Companion PR: https://github.com/ROCm/rocm-libraries/pull/6205

## Test Plan

TBD

## Test Result

TBD

## Submission Checklist


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -135,6 +135,16 @@ def _get_script_path(script_name: str) -> str:
             "windows": 1,
         },
     },
+    "tensilelite": {
+        "job_name": "tensilelite",
+        "fetch_artifact_args": "--blas --tests",
+        "timeout_minutes": 45,
```
geomin12: 45 mins is pretty long, can we shard? are there smoke tests?

build_tools/github_actions/test_executable_scripts/test_tensilelite.py
```diff
@@ -0,0 +1,138 @@
...
+import sys
+from pathlib import Path
+
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+THEROCK_DIR = SCRIPT_DIR.parent.parent.parent
+
+# Paths
+ROCM_LIBRARIES_DIR = THEROCK_DIR / "rocm-libraries"
```
geomin12: this won't work! during `test_component.yml`, we only checkout TheRock and we do not fetch rocm-libraries as it is not needed during tests, meaning this will not work

we'll need to bundle the tensilelite tests artifacts in here: https://github.com/ROCm/TheRock/blob/main/math-libs/BLAS/artifact-blas.toml - that way, when a developer pulls these artifacts, it'll come "out the box"

build_tools/github_actions/test_executable_scripts/test_tensilelite.py
```diff
@@ -0,0 +1,138 @@
...
+        env=env,
+        check=False,
+    )
+    if result.returncode != 0:
+        logging.error(
+            f"TensileLite unit tests failed with return code {result.returncode}"
+        )
+        failed = True
+
+    # Run rocisa tests
```
geomin12: should this be a separate test? or is this directly related to tensiltelite?

geomin12 (changes requested):

build_tools/github_actions/test_executable_scripts/test_tensilelite.py
archana-ramalingam: TensileLite tests are Python/pytest, not compiled gtest binaries — there's nothing to bundle in  artifact-blas.toml. 

The tests require the full Python package tree (Tensile/, rocisa/, pyproject.toml,  uv.lock) for imports and pytest discovery, + we [use uv](https://github.com/ROCm/rocm-libraries/pull/5784) to manage the env. 

So these tests are expected to be unique on TheRockCI because of the necessity of a Python environment.                                                                                    

We'll need a Checkout rocm-libraries step in `therock-test-component.yml` for this component. Can scope it conditionally to tensilelite only if preferred?

archana-ramalingam: Is there a technical limitation preventing a checkout of rocm-libraries during the test phase, or is this a design preference?

build_tools/github_actions/test_executable_scripts/test_tensilelite.py
archana-ramalingam: `rocisa` is a subpackage inside `tensilelite/rocisa/` and a direct dependency of `tensilelite`. They share the same `pyproject.toml` and `uv.lock` — separating them would duplicate the `uv` env setup for 7 test files.

build_tools/github_actions/test_executable_scripts/test_tensilelite.py
ScottTodd: Ideally any user should be able to install ROCm and run project tests from that install, without any extra steps (downloads, configuration, etc.). Many of the subprojects support that already. Requiring a separate source checkout would also mean that any user, CI system, etc. would need to know _which commit_ of that source is associated with the rest of the test libraries/binaries/code/etc.

We have a few subprojects today that include substantial chunks of source code and other project files in their artifacts. These may be useful to reference:
* https://github.com/ROCm/libhipcxx --> https://github.com/ROCm/TheRock/blob/d651c8f488f293badfbf7afe6734211a1348e69d/math-libs/artifact-libhipcxx.toml#L10-L18
* https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler-compute --> https://github.com/ROCm/TheRock/blob/d651c8f488f293badfbf7afe6734211a1348e69d/profiler/artifact-rocprofiler-compute.toml#L6-L10
* https://github.com/ROCm/rocm-systems/tree/develop/projects/rocprofiler-sdk/source/lib/python --> https://github.com/ROCm/TheRock/blob/d651c8f488f293badfbf7afe6734211a1348e69d/profiler/artifact-rocprofiler-sdk.toml#L6-L15

archana-ramalingam: Thanks Scott, this was helpful! I have added tensilelite under hipblaslt's `toml`

archana-ramalingam:
@geomin12 This PR is likely unnecessary, as we chose not to use the artifacts approach, and https://github.com/ROCm/rocm-libraries/pull/6205 appears to pass without it.

geomin12 (changes requested):
i see https://github.com/ROCm/rocm-libraries/pull/6205, but we should bundle it in TheRock together. 

- maintaining another CI workflow `.github/workflows/tensilelite-tests.yml` is a pain point, what `linux-gfx942-1gpu-ossci-rocm` label gets updated? TheRock gets these changes, but tensilelite-tests.yml would need to be another place of maintenance
- how does `.github/workflows/tensilelite-tests.yml` get its artifacts? are they already built or is no build required?
- Adding a test case for `rocm-libraries` makes sense, but we want to test on various levels. rocm-libraries is the first layer of testing in development work, then we want to test in TheRock as changes get moved up. release testing will be done in TheRock. we will miss these tests as we have a separate tensilelite workflow in rocm-libraries
- we will eventually do de-duplication work in TheRock, meaning `therock-ci-*` workflows will pull directly from TheRock.

build_tools/github_actions/fetch_test_configurations.py
archana-ramalingam: Updated to 10 mins

archana-ramalingam:
> i see [ROCm/rocm-libraries#6205](https://github.com/ROCm/rocm-libraries/pull/6205), but we should bundle it in TheRock together.
> 
> * maintaining another CI workflow `.github/workflows/tensilelite-tests.yml` is a pain point, what `linux-gfx942-1gpu-ossci-rocm` label gets updated? TheRock gets these changes, but tensilelite-tests.yml would need to be another place of maintenance
> * how does `.github/workflows/tensilelite-tests.yml` get its artifacts? are they already built or is no build required?
> * Adding a test case for `rocm-libraries` makes sense, but we want to test on various levels. rocm-libraries is the first layer of testing in development work, then we want to test in TheRock as changes get moved up. release testing will be done in TheRock. we will miss these tests as we have a separate tensilelite workflow in rocm-libraries
> * we will eventually do de-duplication work in TheRock, meaning `therock-ci-*` workflows will pull directly from TheRock.

Closing this PR as `TensileLite` tests are [scoped](https://github.com/ROCm/rocm-libraries/pull/6205/) to the `rocm-libraries` repo and will run in isolation, leveraging `TheRock` CI resources only as needed. 
Inclusion in `TheRock`'s release testing pipeline is not necessary for this work.

PR #4302, opened 2026-04-01
[rocgdb] Add --allow-multiple-definition for LLD compatibility

FIM43-Redeye:
## Summary
- LLD is stricter than GNU ld about duplicate symbols from static archives
- rocgdb links libiberty statically, which provides symbols (xmalloc,
  xstrdup, etc.) also defined in other linked archives
- Add `-Wl,--allow-multiple-definition` to LDFLAGS to match GNU ld behavior

Fixes #4223

PR #4380, opened 2026-04-07
[rocgdb] Skip -L/-rpath for system libpython directories

### LINKED ISSUE #4223: rocgdb fails to link with LLD when libiberty-dev is installed
rocgdb fails to build on systems where the compiler defaults to LLD and `libiberty-dev` is installed. GDB's bundled `libiberty.a` conflicts with the system copy -- LLD rejects the duplicate symbols, GNU ld silently resolves them.

```
[rocgdb] ld.lld: error: duplicate symbol: get_DW_TAG_name
>>> defined at dwarfnames.o in archive /usr/lib/x86_64-linux-gnu/libiberty.a
>>> defined at dwarfnames.c in archive ../libiberty/libiberty.a
```

(Same error for `get_DW_FORM_name`, `get_DW_AT_name`, `get_DW_OP_name`, `get_DW_ATE_name`, `get_DW_CFA_name`, `get_DW_IDX_name`, `get_DW_UT_name`.)

## Root cause

The libpython directory gets added to LDFLAGS in `debug-tools/rocgdb/CMakeLists.txt:198-199`:

```cmake
set(LIBPYTHON_LDFLAGS "-L${libpython_dir} -Wl,-rpath,${libpython_dir}")
```

When libpython lives in `/usr/lib/x86_64-linux-gnu`, this puts the entire system lib directory on the search path. GDB's BFD configure then resolves `-liberty` against the system `libiberty.a` in addition to GDB's own bundled copy, causing the duplicate.

Related to #1780 and #2149 -- this will affect anyone building rocgdb once LLD becomes the default system linker.

## Environment

- TheRock main (59d98622), Clang 20.1.8 defaulting to LLD, `libiberty-dev` installed, Linux x86_64

## Reproduction

```bash
cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx1100 -DTHEROCK_ENABLE_ROCGDB=ON
ninja -C build rocgdb
```

Requires: Clang defaulting to LLD (`clang -Wl,--version` shows LLD) and `libiberty-dev` installed.

FIM43-Redeye:
## Summary

- When `sysconfig.get_config_var('LIBDIR')` resolves to a system directory (e.g. `/usr/lib/x86_64-linux-gnu`), the explicit `-L` flag is unnecessary -- the linker already searches there -- and harmful, because it pulls in system `libiberty.a` which conflicts with GDB's bundled copy under LLD.
- GDB's configure discovers system libpython via `--with-python` and its internal `python-config.py` without needing the `-L` hint.
- Non-system libpython paths (`/opt`, `/usr/local`, custom installs) still get the explicit `-L` and `-rpath` flags.
- System paths are identified as under `/usr` but not `/usr/local`, since `/usr/local` is for user-installed software and may not be in the default linker search path on all distributions.

Supersedes #4302, which worked around the symptom with `--allow-multiple-definition`. This addresses the root cause as suggested by @lumachad.

Fixes #4223

## Test plan

- CI builds use bundled Python at `/opt/python-shared/` (non-system path), so this change takes the existing code path there -- CI validates no regression for that case.
- For the system-Python + LLD scenario (the #4223 bug), verified locally that CMake configure completes cleanly and the libpython detection logic is reached.


debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -192,15 +192,30 @@ macro(add_python_variant minor_version configure_flags libpython_dir)
...
+  # the explicit flags since those paths are not in the default search order.
+  # See https://github.com/ROCm/TheRock/issues/4223
   set(ALL_LDFLAGS ${LDFLAGS})
   set(LIBPYTHON_RPATH "")
   if(NOT "${libpython_dir}" STREQUAL "")
-    set(LIBPYTHON_LDFLAGS "-L${libpython_dir} -Wl,-rpath,${libpython_dir}")
-    string(APPEND ALL_LDFLAGS " ${LIBPYTHON_LDFLAGS}")
-    set(LIBPYTHON_RPATH ":${libpython_dir}")
+    cmake_path(IS_PREFIX "/usr" "${libpython_dir}" NORMALIZE _is_under_usr)
+    cmake_path(IS_PREFIX "/usr/local" "${libpython_dir}" NORMALIZE _is_under_usr_local)
```
lumachad: I think these have their arguments reversed.

Should've been like this?

```
cmake_path(IS_PREFIX "${libpython_dir}" "/usr" NORMALIZE _is_under_usr)
cmake_path(IS_PREFIX "${libpython_dir}" "/usr/local" NORMALIZE _is_under_usr_local)
```

FIM43-Redeye: Thanks for reviewing- double-checked the cmake_path docs and I believe the current order is correct? Signature is `cmake_path(IS_PREFIX <prefix> <input> [NORMALIZE] <out_var>)` which would mean this reads as "is `/usr` a prefix of `libpython_dir`?", which was the intent. Referenced https://cmake.org/cmake/help/latest/command/cmake_path.html#is-prefix

lumachad: Yeah, you're right. Though the build system is grumpy at this:

```
[rocgdb configure] CMake Error at CMakeLists.txt:207 (cmake_path):
[rocgdb configure]   cmake_path undefined variable for input path.
[rocgdb configure] Call Stack (most recent call first):
[rocgdb configure]   CMakeLists.txt:424 (add_python_variant)
[rocgdb configure] 
[rocgdb configure] 
[rocgdb configure] CMake Error at CMakeLists.txt:208 (cmake_path):
[rocgdb configure]   cmake_path undefined variable for input path.
[rocgdb configure] Call Stack (most recent call first):
[rocgdb configure]   CMakeLists.txt:424 (add_python_variant)
[rocgdb configure] 
[rocgdb configure] 
```

I think it doesn't like "${libpython_dir}" as the input variable.

FIM43-Redeye: Good catch on the build failure -- `cmake_path(IS_PREFIX)` needs the first argument as a variable name, not a literal. Fixed in the latest push, confirmed it configures cleanly now.

geomin12 (commented):
will let rocgdb folks review, but approved workflow run so CI is running!

lumachad:
re-running rocgdb tests since I spotted one failure there.

lumachad:
@FIM43-Redeye Looks like the other failures are unrelated. Could you please confirm this fixes #4223?

FIM43-Redeye:
@lumachad Fixes it on my end, yup. The system Python and LLD scenario doesn't pull in the conflicting libiberty.a anymore, so it should be all good.

lumachad:
Adding @marbre for a second pair of eyes on the cmake changes. The changes look OK to me, and testing looks good for rocgdb. Could you please double check the cmake changes?

debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -192,15 +192,32 @@ macro(add_python_variant minor_version configure_flags libpython_dir)
...
   set(LIBPYTHON_RPATH "")
   if(NOT "${libpython_dir}" STREQUAL "")
-    set(LIBPYTHON_LDFLAGS "-L${libpython_dir} -Wl,-rpath,${libpython_dir}")
-    string(APPEND ALL_LDFLAGS " ${LIBPYTHON_LDFLAGS}")
-    set(LIBPYTHON_RPATH ":${libpython_dir}")
+    set(_usr_prefix "/usr")
+    set(_usr_local_prefix "/usr/local")
+    cmake_path(IS_PREFIX _usr_prefix "${libpython_dir}" NORMALIZE _is_under_usr)
+    cmake_path(IS_PREFIX _usr_local_prefix "${libpython_dir}" NORMALIZE _is_under_usr_local)
+    if(_is_under_usr AND NOT _is_under_usr_local)
```
marbre, line 217: Isn't `if(_is_under_usr)` sufficient and the whole checking for `/usr/local` could be skipped? We also don't check for `AND NOT _is_under_opt` and I think it wouldn't make a change anyway as we hit `else()` then?

FIM43-Redeye: The `/usr/local` check is necessary as far as I understand it. `IS_PREFIX /usr /usr/local/lib` is true, so without the exclusion it gets classified as a system path and loses the `-L/-rpath` flags. `/usr/local` can't be guaranteed to be in the default linker search path everywhere, either. The `/opt` point is good, but `_is_under_usr` is false for it so they hit the `else()`. Explicit handling shouldn't be needed.

lumachad:
Rebasing to pick up updates.

debug-tools/rocgdb/CMakeLists.txt
marbre: Oh, true! I missed you're checking for `/usr` and not `/usr/lib{64}`. Good as is.

b-sumner:
I am very much in need of a working rocgdb produced by a TheRock build. I would appreciate if this, or another fix, could land very soon.

lumachad:
> I am very much in need of a working rocgdb produced by a TheRock build. I would appreciate if this, or another fix, could land very soon.

If you just need rocgdb, does a nightly tarball help? Or a TheRock build from a manylinux container (where we don't point at the system's libpython directory).

lumachad (approved):
This is OK for me, if the latest CI run is good.

lumachad:
@marbre @geomin12 Any additional thoughts on this one?

FIM43-Redeye:
Quick nudge - @marbre @geomin12, any additional thoughts on the cmake changes? Approved by @lumachad last week, and the CI failures on the latest run are all from the `dvc pull` S3 auth flake hitting math-libs stages (`s3_disable_express_session_auth`, a botocore/aiobotocore version conflict in the CI container), seems unrelated to this PR. Happy to rebase again if that would help move things along.

marbre (approved):
math-libs failures are unrelated, debug stage / tests look good.

lumachad:
Labelled ci:skip as I accidentaly hit update branch, and CI looked good before.

lumachad:
@b-sumner FYI

Issue #2281, opened 2025-11-24
Run ROCm sub-project tests as part of release workflows


araravik-psd:
We presently dont run any component tests on the release workflows in TheRock or [rockrel](https://github.com/ROCm/rockrel). Tests are only run for pytorch as part of [release_portable_linux_pytorch_wheels.yml](https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_portable_linux_pytorch_wheels.yml) and [release_windows_pytorch_wheels.yml](https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_windows_pytorch_wheels.yml).

We would like to plug in rocm component tests as part of [release_portable_linux_packages.yml](https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_portable_linux_packages.yml) and [release_windows_packages.yml](https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_windows_packages.yml) once artifacts are built.

ScottTodd:
I'm considering two main approaches here:

1. After release workflows build artifacts (or publish files), trigger test workflows using `workflow_dispatch`, one workflow run per GPU target/family
    * Test jobs would be their own workflow runs, tracked on separate dashboards
    * Code in TheRock: https://github.com/ScottTodd/TheRock/tree/multi-arch-release-artifact-tests-1
    * Code in rockrel: https://github.com/ScottTodd/rockrel/tree/multi-arch-artifact-tests-1
2. After release workflows build artifacts (or publish files), trigger test workflows using `workflow_call`, one parent job per GPU target/family
    * This would match what multi-arch CI does, particularly when the `ci:run-all-archs` label is applied
    * All test jobs would be visible from a single workflow run (easy to scroll down to see more, but the UI has scaling issues and log downloads can get difficult)
    * If a build job fails, we'd need to wait for all (queued/slow) test jobs to complete to attempt a retry, of we'd need to cancel the job
    * Code in TheRock: https://github.com/ROCm/TheRock/tree/users/scotttodd/multi-arch-release-artifact-tests-2 --> test run: https://github.com/ROCm/TheRock/actions/runs/24917079667
    * (No code in rockrel)

[...]

PR #4384, opened 2026-04-07
[ci] Convert ci_nightly.yml to multi-arch style

geomin12:
Converting ci_nightly.yml to multi-arch

Test ran here: https://github.com/ROCm/TheRock/actions/runs/24106521430 (the failure is due to test_labels bug, but #4383 fixes that)

Adding `ci:skip` as this only impacts ci_nightly.yml

### DIFF
--- a/.github/workflows/ci_nightly.yml
+++ b/.github/workflows/ci_nightly.yml
@@ -21,26 +21,24 @@ on:
         default: ""
       linux_test_labels:
         type: string
-        description: "If enabled, reduce test set on Linux to the list of labels prefixed with 'test:'"
+        description: "If enabled, reduce test set on Linux to the list of labels prefixed with 'test:'. ex: test:rocprim, test:hipcub"
         default: ""
-      linux_use_prebuilt_artifacts:
-        type: boolean
-        description: "If enabled, the CI will pull Linux artifacts using artifact_run_id and only run tests"
       windows_amdgpu_families:
         type: string
         description: "Insert comma-separated list of Windows GPU families to build and test. ex: gfx94X, gfx120X"
         default: ""
       windows_test_labels:
         type: string
-        description: "If enabled, reduce test set on Windows to the list of labels prefixed with 'test:'"
+        description: "If enabled, reduce test set on Windows to the list of labels prefixed with 'test:' ex: test:rocprim, test:hipcub"
         default: ""
-      windows_use_prebuilt_artifacts:
-        type: boolean
-        description: "If enabled, the CI will pull Windows artifacts using artifact_run_id and only run tests"
-      artifact_run_id:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_nightly.yml
```diff
@@ -56,80 +54,63 @@ concurrency:
...
-        variant: ${{ fromJSON(needs.setup.outputs.linux_variants) }}
-    uses: ./.github/workflows/ci_linux.yml
+    uses: ./.github/workflows/multi_arch_ci_linux.yml
     secrets: inherit
     with:
-      amdgpu_families: ${{ matrix.variant.family }}
-      artifact_group: ${{ matrix.variant.artifact_group }}
-      test_runs_on: ${{ matrix.variant.test-runs-on }}
-      benchmark_runs_on: ${{ matrix.variant.benchmark-runs-on }}
-      run_functional_tests: ${{ needs.setup.outputs.run_functional_tests == 'true' }}
```
ScottTodd, line 82: These benchmarks / functional tests that run on a nightly `schedule` are still only supported in `ci_linux.yml` and not `multi_arch_ci_linux.yml`, so doing this switch would stall the work that @lajagapp is doing on https://github.com/ROCm/TheRock/pull/3986

Can we try converting ci_asan.yml first? I think the feature set there is closer to what runs on `push` and `pull_request` events.

geomin12: i was doing both in parallel :) asan is next and coming soon

ScottTodd:
I think we'll want to rework ci_nightly.yml together with these issues:
* https://github.com/ROCm/TheRock/issues/2281

As currently written, once we add tests to the release workflows, we'll be running duplicate builds and tests:

* release_*.yml triggered
  * build artifacts, packages, etc. for all possible targets
  * publish those packages (either before or after tests)
  * (new) test those artifacts, packages, etc.
* ci_nightly.yml triggered
  * (already duplicate work) build artifacts, packages, etc. for all possible targets
  * test those artifacts, packages, etc.

We could restructure ci_nightly to pull artifacts from the latest nightly release build, or just work that trigger directly into the release workflows. Where we currently trigger pytorch/jax/etc. jobs after building rocm artifacts we could also trigger rocm test jobs.

jayhawk-commits:
Is this work still needed?

geomin12:
> Is this work still needed?

yep! now getting around to doing this conversion. although if release pipelines does this, this will be redundant

geomin12:
Closing this PR as release workflows now handle this: https://github.com/ROCm/TheRock/blob/7b7b238e3e0c2e98436ea230b5114a5b6b946abf/.github/workflows/multi_arch_release_linux.yml#L155-L182

Issue #4195, opened 2026-03-27
[CI] Follow-up improvements for Windows ccache setup


subodh-dubey-amd:
Follow-up items from #4161 review comments.

## Tasks

- [ ] **Move local ccache storage to B: drive on Windows runners** — current `CACHE_DIR` is on C: (OS disk, 30GB) which may cause disk space issues with high cache miss rates. Should use the temp disk on B: like the build directory.
  Ref: https://github.com/ROCm/TheRock/pull/4161#discussion_r2117071565

- [ ] **Fix `set` vs `export` shell detection in `setup_ccache.py`** — Windows CI uses bash (not cmd), so `set` is wrong. Detect shell type instead of OS, or remove the output since workflows set `CCACHE_CONFIGPATH` as an env var directly.
  Ref: https://github.com/ROCm/TheRock/pull/4161#discussion_r2116281498

- [ ] **Consider Windows-compatible `compiler_check`** — current POSIX check uses `ldd`/`sha256sum`; a Windows equivalent could improve cache robustness beyond the default `mtime` check.
  Ref: https://github.com/ROCm/TheRock/pull/4161#discussion_r2116294498

- [ ] **Refine ccache preset selection for multi-arch workflows** — split `presubmit`/`postsubmit`/`release` presets based on trigger type (`pull_request` vs `push` vs `schedule`) when multi-arch replaces old CI pipelines.
  Ref: https://github.com/ROCm/TheRock/pull/4161#pullrequestreview-2782133855

- [ ] **Monitor GHA cache usage** — with `actions/cache` removed from Windows builds, check https://github.com/ROCm/TheRock/actions/caches to confirm reduced pressure on the 10GB limit.
  Ref: https://github.com/ROCm/TheRock/pull/4161#pullrequestreview-2782133855

ScottTodd:
Cache hit rates for Windows math-libs stage on multi-arch CI seem pretty low. I'd say that's highest priority to fix compared to the other items.

For example: https://github.com/ROCm/TheRock/actions/runs/23663263447/job/68945135123?pr=4209#step:17:68

```
Cacheable calls:                7636 / 7702 (99.14%)
  Hits:                          123 / 7636 ( 1.61%)
    Direct:                       55 /  123 (44.72%)
    Preprocessed:                 68 /  123 (55.28%)
  Misses:                       7513 / 7636 (98.39%)
Uncacheable calls:                66 / 7702 ( 0.86%)
```

[...]

PR #4419, opened 2026-04-09
Windows reproducible build & ccache hit rate improvement

### LINKED ISSUE #4519: [CI] Windows ccache hit rate improvement
### Suggestion Description

Improve Windows ccache hit rate

### Operating System

Windows

### GPU

_No response_

### ROCm Component

TheRock

# Work Items
1. [x] Add /Brepro for reproducible builds to ensure compiler and components are stable.
    Added to all component via TheRock init.
2. [x] The compiler check on Windows needs work.
    Switch to use `content` compiler check.
3. [x] Improve compiler bootstrap cache hit rate.
4. [x] Documentation update to guide users to enable ccache on Windows
5. [ ] Investigate why the cache rate remains low on mainline CI

amd-nicknick:
## Motivation

Fixes #4519, parent issue #4195

## Technical Details

* Reproducible builds on Windows via `/Brepro` for MSVC toolchain
  Ensure amd-llvm is reproducible & COFF header contains a reproducible hash instead of current timestamp.
* `compiler_check=content` for Windows
  Switch `compiler_check` from `mtime` to `content`. Since the LLVM toolchain on Windows is statically linked, hashing the compiler binary content is sufficient and more reliable for detecting actual compiler changes across CI runs.
* Support PCH
  Add `pch_defines` and `time_micros` to ccache sloppiness. amd-llvm on Windows builds uses precompiled headers by default.
* Misc changes to ccache config
  Max size bump to 10G, rename `secondary_storage` to `remote_storage`

## Test Plan

Validate cache hitrate on Windows following subsequent buil.ds

## Test Result
```
(therock-build-venv) D:\Nick\TheRock\TheRock>ccache -s -v
Cache directory:                D:\Nick\TheRock\TheRock\.ccache\local
Config file:                    D:\Nick\TheRock\TheRock\.ccache\ccache.conf
System config file:
Stats updated:                  04/17/26 05:11:03
Cacheable calls:                28464 /  30423 (93.56%)
  Hits:                         28128 /  28464 (98.82%)
    Direct:                     28128 /  28128 (100.0%)
    Preprocessed:                   0 /  28128 ( 0.00%)
  Misses:                         336 /  28464 ( 1.18%)
Uncacheable calls:               1959 /  30423 ( 6.44%)
  Compilation failed:              13 /   1959 ( 0.66%)
  Compiler output file missing:  1149 /   1959 (58.65%)
  No input file:                  371 /   1959 (18.94%)
  Preprocessing failed:            69 /   1959 ( 3.52%)
  Unsupported code directive:       3 /   1959 ( 0.15%)
  Unsupported compiler option:    354 /   1959 (18.07%)
Successful lookups:
  Direct:                       28128 /  29698 (94.71%)
  Preprocessed:                     0 /   1498 ( 0.00%)
Local storage:
  Cache size (GB):                7.2 / 1024.0 ( 0.70%)
  Files:                        68376
  Hits:                         28128 /  28464 (98.82%)
  Misses:                         336 /  28464 ( 1.18%)
  Reads:                        59324
  Writes:                         672
```

## Submission Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,21 +322,10 @@ for /f "delims=" %i in ('python build_tools/setup_ccache.py') do @%i
 cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx110X-all \
   -DCMAKE_C_COMPILER_LAUNCHER=ccache ^
   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ^
-  -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded ^
   .
 
 cmake --build build
 ```
-
-> [!TIP]
-> ccache [does not support](https://github.com/ccache/ccache/issues/1040)
-> MSVC's `/Zi` flag which may be set by default when a project (e.g. LLVM) opts
-> in to
-> [policy CMP0141](https://cmake.org/cmake/help/latest/policy/CMP0141.html).
-> Setting
-> [`-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html)
-> instructs CMake to compile with `/Z7` or equivalent, which is supported by
-> ccache.
 
 ### Running tests
 
--- a/README.md
+++ b/README.md
@@ -299,8 +299,44 @@ cmake --build build
 
 #### CCache usage on Windows
 
-We are still investigating the exact proper options for ccache on Windows and
-do not currently recommend that end users enable it.
+- You must have a recent ccache (>= 4.13.3 at the time of writing) that contains
+  bug fixes for MSVC and supports proper caching with the `--offload-compress`
+  option used for compressing AMDGPU device code.
+- `export CCACHE_SLOPPINESS=include_file_ctime,pch_defines,time_macros` to
+  support hard-linking and precompiled headers (amd-llvm is built with PCH).
+- Proper setup of the `compiler_check` directive to do safe caching in the
+  presence of compiler bootstrapping.
+- Set the C/CXX compiler launcher options to cmake appropriately.
+
+Since these options are very fiddly and prone to change over time, we recommend
+using the `./build_tools/setup_ccache.py` script to create a `.ccache` directory
+in the repository root with hard coded configuration suitable for the project.
+
+Example (In Command Prompt):
+
+```bat
+# Any command prompt used to build must eval setup_ccache.py to set environment
+# variables.
+for /f "delims=" %i in ('python build_tools/setup_ccache.py') do @%i
+
+cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx110X-all \
+  -DCMAKE_C_COMPILER_LAUNCHER=ccache ^
+  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ^
+  -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded ^
+  .
+
+cmake --build build
+```
+
+> [!TIP]
+> ccache [does not support](https://github.com/ccache/ccache/issues/1040)
+> MSVC's `/Zi` flag which may be set by default when a project (e.g. LLVM) opts
+> in to
+> [policy CMP0141](https://cmake.org/cmake/help/latest/policy/CMP0141.html).
+> Setting
+> [`-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html)
+> instructs CMake to compile with `/Z7` or equivalent, which is supported by
+> ccache.
 
 ### Running tests
 
--- a/build_tools/github_actions/build_configure.py
+++ b/build_tools/github_actions/build_configure.py
@@ -81,10 +81,6 @@ def build_compiler_launcher(
 platform_options = {
     "windows": [
         "-DTHEROCK_BACKGROUND_BUILD_JOBS=4",
-        # ccache does not support MSVC's /Zi flag. Setting Embedded instructs
-        # CMake to compile with /Z7 (or equivalent), which ccache supports.
-        # See: https://github.com/ccache/ccache/issues/1040
-        "-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded",
     ],
 }
 
--- a/build_tools/github_actions/build_configure.py
+++ b/build_tools/github_actions/build_configure.py
@@ -81,6 +81,10 @@ def build_compiler_launcher(
 platform_options = {
     "windows": [
         "-DTHEROCK_BACKGROUND_BUILD_JOBS=4",
+        # ccache does not support MSVC's /Zi flag. Setting Embedded instructs
+        # CMake to compile with /Z7 (or equivalent), which ccache supports.
+        # See: https://github.com/ccache/ccache/issues/1040
+        "-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded",
     ],
 }
 
--- a/build_tools/setup_ccache.py
+++ b/build_tools/setup_ccache.py
@@ -111,16 +111,21 @@ def gen_config(dir: Path, compiler_check_file: Path, args: argparse.Namespace):
         # therefore using content is sufficient to detect changes.
         lines.append(f"compiler_check = content")
 
-    # Slop settings.
-    # Creating a hard link to a file increasing the link count, which triggers
-    # a ctime update (since ctime tracks changes to the inode metadata) for
-    # *all* links to the file. Since we are basically always creating hard
-    # link farms in parallel as part of sandboxing, we have to disable this
-    # check as it is never valid for our build system and will result in
-    # spurious ccache panics where it randomly falls back to the real compiler
-    # if the ccache invocation happens to coincide with parallel sandbox
-    # creation for another sub-project.
-    lines.append(f"sloppiness = include_file_ctime")
+    # Sloppiness settings.
+    # include_file_ctime:
+    #   Creating a hard link to a file increasing the link count, which triggers
+    #   a ctime update (since ctime tracks changes to the inode metadata) for
+    #   *all* links to the file. Since we are basically always creating hard
+    #   link farms in parallel as part of sandboxing, we have to disable this
+    #   check as it is never valid for our build system and will result in
+    #   spurious ccache panics where it randomly falls back to the real compiler
+    #   if the ccache invocation happens to coincide with parallel sandbox
+    #   creation for another sub-project.
+    # pch_defines, time_macros:
+    #   amd-llvm uses PCH on Windows builds by default, CMake will correctly
+    #   use the appropriate compilation flags that ccache understands. See 
+    #   https://ccache.dev/manual/4.7.html#_precompiled_headers for details.
+    lines.append(f"sloppiness = include_file_ctime,pch_defines,time_macros")
 
     # End with blank line.
     lines.append("")
--- a/build_tools/setup_ccache.py
+++ b/build_tools/setup_ccache.py
@@ -40,6 +40,11 @@ POSIX_COMPILER_CHECK_SCRIPT = (
 CACHE_SRV_DEV = "http://bazelremote-svc.bazelremote-ns.svc.cluster.local:8080|layout=bazel|connect-timeout=50"
 CACHE_SRV_REL = "http://bazelremote-svc-rel.bazelremote-ns.svc.cluster.local:8080|layout=bazel|connect-timeout=50"
 
+# Bump this version when making hash-affecting config changes (sloppiness,
+# compiler_check, etc.) to logically isolate new cache entries from stale
+# ones on the shared remote cache server.
+CCACHE_NAMESPACE_VERSION = "v1"
+
 DEFAULT_LOG_DIR = REPO_ROOT / "build" / "logs" / "ccache"
 
 # See https://ccache.dev/manual/4.6.1.html#_configuration
@@ -56,10 +61,12 @@ CONFIG_PRESETS_MAP = {
     "github-oss-dev": {
         "remote_storage": CACHE_SRV_DEV,
         "max_size": "10G",
+        "namespace": f"therock-{CCACHE_NAMESPACE_VERSION}",
     },
     "github-oss-release": {
         "remote_storage": CACHE_SRV_REL,
         "max_size": "10G",
+        "namespace": f"therock-{CCACHE_NAMESPACE_VERSION}",
     },
 }
 
--- a/build_tools/setup_ccache.py
+++ b/build_tools/setup_ccache.py
@@ -47,19 +47,19 @@ DEFAULT_LOG_DIR = REPO_ROOT / "build" / "logs" / "ccache"
 # so that Windows workflows can direct logs to BUILD_DIR/logs/ccache/ (B:\build)
 # instead of REPO_ROOT/build/logs/ccache/ (C: drive).
 CONFIG_PRESETS_MAP = {
-    "local": {},
+    "local": {"max_size": "10G"},
     # Dev and release use separate cache servers to avoid cache pollution.
     # We may later split these further into presubmit (PR) vs postsubmit
     # (post-merge) presets — presubmit serves varied code at mixed trust
     # levels while postsubmit serves a uniform stream of approved commits,
     # so separating them improves both cache hit rates and data integrity.
     "github-oss-dev": {
-        "secondary_storage": CACHE_SRV_DEV,
-        "max_size": "5G",
+        "remote_storage": CACHE_SRV_DEV,
+        "max_size": "10G",
     },
     "github-oss-release": {
-        "secondary_storage": CACHE_SRV_REL,
-        "max_size": "5G",
+        "remote_storage": CACHE_SRV_REL,
+        "max_size": "10G",
     },
 }
 
@@ -123,7 +123,7 @@ def gen_config(dir: Path, compiler_check_file: Path, args: argparse.Namespace):
     #   creation for another sub-project.
     # pch_defines, time_macros:
     #   amd-llvm uses PCH on Windows builds by default, CMake will correctly
-    #   use the appropriate compilation flags that ccache understands. See 
+    #   use the appropriate compilation flags that ccache understands. See
     #   https://ccache.dev/manual/4.7.html#_precompiled_headers for details.
     lines.append(f"sloppiness = include_file_ctime,pch_defines,time_macros")
 
--- a/build_tools/setup_ccache.py
+++ b/build_tools/setup_ccache.py
@@ -98,14 +98,18 @@ def gen_config(dir: Path, compiler_check_file: Path, args: argparse.Namespace):
         local_path.mkdir(parents=True, exist_ok=True)
         lines.append(f"cache_dir = {local_path}")
 
-    # Compiler check: on POSIX we use a custom script that fingerprints the
-    # compiler binary and its shared libraries via ldd + sha256sum. On Windows
-    # (MSVC) those tools don't exist; ccache's default mtime check works well.
+    # Compiler Check
     if not IS_WINDOWS:
+        # On POSIX we use a custom script that fingerprints the
+        # compiler binary and its shared libraries via ldd + sha256sum.
         lines.append(
             f"compiler_check = {sys.executable} {compiler_check_file} "
             f"{dir / 'compiler_check_cache'} %compiler%"
         )
+    else:
+        # On Windows the LLVM toolchain is compiled statically linked,
+        # therefore using content is sufficient to detect changes.
+        lines.append(f"compiler_check = content")
 
     # Slop settings.
     # Creating a hard link to a file increasing the link count, which triggers
--- a/cmake/therock_compiler_config.cmake
+++ b/cmake/therock_compiler_config.cmake
@@ -7,7 +7,7 @@
 # This improves compatibility with ccache and sccache:
 # https://github.com/ccache/ccache/issues/1040
 if(WIN32 AND NOT DEFINED CMAKE_MSVC_DEBUG_INFORMATION_FORMAT)
-  set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>")
+  set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "Embedded")
 endif()
 
 if(WIN32 AND NOT MSVC AND NOT THEROCK_DISABLE_MSVC_CHECK)
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -818,6 +818,14 @@ function(therock_cmake_subproject_activate target_name)
   string(APPEND _init_contents "set(THEROCK_PKG_CONFIG_DIRS \"@_private_pkg_config_dirs@\")\n")
 
   string(APPEND _init_contents "${_compiler_toolchain_init_contents}")
+
+  # Enable reproducible builds on Windows. /Brepro makes the linker (both
+  # MSVC link.exe and lld-link) zero out timestamps in PE headers, producing
+  # deterministic output.
+  if(WIN32)
+    string(APPEND _init_contents "add_link_options(\"LINKER:/Brepro\")\n")
+  endif()
+
   if(_dep_provider_file)
     string(APPEND _init_contents "include(${_dep_provider_file})\n")
   endif()
--- a/.github/workflows/multi_arch_build_windows_artifacts.yml
+++ b/.github/workflows/multi_arch_build_windows_artifacts.yml
@@ -165,7 +165,6 @@ jobs:
             -DTHEROCK_PACKAGE_VERSION=${{ inputs.rocm_package_version }} \
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-            -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded \
             ${{ steps.stage_config.outputs.cmake_args }}
 
       - name: Build stage
--- a/.github/workflows/multi_arch_build_windows_artifacts.yml
+++ b/.github/workflows/multi_arch_build_windows_artifacts.yml
@@ -165,6 +165,7 @@ jobs:
             -DTHEROCK_PACKAGE_VERSION=${{ inputs.rocm_package_version }} \
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded \
             ${{ steps.stage_config.outputs.cmake_args }}
 
       - name: Build stage
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -83,12 +83,28 @@ ctest --test-dir build
 
 **Faster rebuilds with ccache:**
 
+On Linux:
+
 ```bash
+# Any shell used to build must eval setup_ccache.py to set environment variables.
+eval "$(./build_tools/setup_ccache.py)"
 cmake -B build -GNinja \
   -DCMAKE_C_COMPILER_LAUNCHER=ccache \
   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
   -DTHEROCK_AMDGPU_FAMILIES=gfx1100
 ```
+
+On Windows (Command Prompt):
+
+```bat
+for /f "delims=" %i in ('python build_tools/setup_ccache.py') do @%i
+cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx1100 ^
+  -DCMAKE_C_COMPILER_LAUNCHER=ccache ^
+  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ^
+  -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded
+```
+
+See [README.md](README.md#ccache-usage-on-windows) for full ccache setup details.
 
 **Debug build for specific component:**
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,7 +91,7 @@ eval "$(./build_tools/setup_ccache.py)"
 cmake -B build -GNinja \
   -DCMAKE_C_COMPILER_LAUNCHER=ccache \
   -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-  -DTHEROCK_AMDGPU_FAMILIES=gfx1100
+  -DTHEROCK_AMDGPU_FAMILIES=gfx1100 .
 ```
 
 On Windows (Command Prompt):
@@ -100,8 +100,7 @@ On Windows (Command Prompt):
 for /f "delims=" %i in ('python build_tools/setup_ccache.py') do @%i
 cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx1100 ^
   -DCMAKE_C_COMPILER_LAUNCHER=ccache ^
-  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ^
-  -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded
+  -DCMAKE_CXX_COMPILER_LAUNCHER=ccache .
 ```
 
 See [README.md](README.md#ccache-usage-on-windows) for full ccache setup details.
--- a/build_tools/validate_windows_install.ps1
+++ b/build_tools/validate_windows_install.ps1
@@ -505,7 +505,6 @@ $sccacheCmd = Get-Command sccache -ErrorAction SilentlyContinue
 if ($ccacheCmd) {
     $ccVer = ((& ccache --version 2>&1 | Select-Object -First 1) -replace "ccache version ", "").Trim()
     Write-Pass "ccache $ccVer (speeds up rebuilds)"
-    Write-Info "Note: use -DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded with ccache to avoid /Zi conflicts"
 } elseif ($sccacheCmd) {
     $scVer = ((& sccache --version 2>&1) -replace "sccache ", "").Trim()
     Write-Pass "sccache $scVer (speeds up rebuilds)"
--- a/docs/development/development_guide.md
+++ b/docs/development/development_guide.md
@@ -315,14 +315,31 @@ extension can be used to configure the superproject and build individual targets
 Settings for CMake builds can be specified in `.vscode/settings.json` or a
 `.vscode/*.code-workspace` file, like so:
 
+> [!IMPORTANT]
+> If using ccache, first run `setup_ccache.py` to generate the config file:
+>
+> ```bash
+> python build_tools/setup_ccache.py
+> ```
+>
+> Then set the `CCACHE_CONFIGPATH` environment variable in your VSCode
+> settings as shown below. See the
+> [README ccache instructions](../../README.md#ccache-usage-on-linux)
+> for full details.
+
 ```jsonc
 {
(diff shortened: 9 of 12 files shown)


build_tools/setup_ccache.py
```diff
@@ -98,25 +98,34 @@ def gen_config(dir: Path, compiler_check_file: Path, args: argparse.Namespace):
...
+    #   link farms in parallel as part of sandboxing, we have to disable this
+    #   check as it is never valid for our build system and will result in
+    #   spurious ccache panics where it randomly falls back to the real compiler
+    #   if the ccache invocation happens to coincide with parallel sandbox
+    #   creation for another sub-project.
+    # pch_defines, time_macros:
+    #   amd-llvm uses PCH on Windows builds by default, CMake will correctly
+    #   use the appropriate compilation flags that ccache understands. See 
+    #   https://ccache.dev/manual/4.7.html#_precompiled_headers for details.
+    lines.append(f"sloppiness = include_file_ctime,pch_defines,time_macros")
```
astrelsky, line 128: Just a note that this might cause the existing cache to be invalidated (not really, bad terminology, you just won't get a cache hit the first time). When I added it and built from scratch it took the usual amount of time. I'm only bringing it up in case there is a concern later about these changes introducing a regression.


Edit: I might just need to increase my local cache storage, the default appears to be 5GiB 😂

astrelsky: Ok, I just finished confirming that it had to be a cache size problem. 5GB is not enough. I see that this script sets the remote storage to a max size of 5GB as well, this should definitely be increased. If I ran into a problem locally because of it, then it's definitely being encountered with the runners sharing the storage when building for different gpu architectures.

amd-nicknick: Great catch, I have mine overridden via env variable and just completely forgot it.
Pushed update to bump to 10G for now, we could observe CI and adjust if necessary.

astrelsky: > Great catch, I have mine overridden via env variable and just completely forgot it.
> Pushed update to bump to 10G for now, we could observe CI and adjust if necessary.

FWIW my cache is currently at 8.9GB from just one full build of TheRock and pytorch, torchaudio, torchvision for only one GPU architecture. You might want to go more than that and I'm assuming the build speed and hit rates will go up on Linux too.

cmake/therock_subproject.cmake
```diff
@@ -834,6 +834,14 @@ function(therock_cmake_subproject_activate target_name)
   string(APPEND _init_contents "set(THEROCK_PKG_CONFIG_DIRS \"@_private_pkg_config_dirs@\")\n")
 
   string(APPEND _init_contents "${_compiler_toolchain_init_contents}")
+
+  # Enable reproducible builds on Windows. /Brepro makes the linker (both
+  # MSVC link.exe and lld-link) zero out timestamps in PE headers, producing
+  # deterministic output.
+  if(WIN32)
+    string(APPEND _init_contents "add_link_options(\"LINKER:/Brepro\")\n")
+  endif()
```
ScottTodd, line 827: I feel like there should be a better place to put this option 🤔 

How about down here? https://github.com/ROCm/TheRock/blob/faadaa4d93c6096034295700c53add9cf811a376/cmake/therock_subproject.cmake#L1484-L1523

My code review agent also pointed me at LLVM's `HandleLLVMOptions.cmake` where `Brepro` is used, so we might want to try matching that code style: https://github.com/llvm/llvm-project/blob/0d1bf3ac8eb10ecf062a938955637f8f809c0380/llvm/cmake/modules/HandleLLVMOptions.cmake#L685-L706

---

For reference, my prompt:
> Can you review https://github.com/ROCm/TheRock/pull/4419? In particular, I'm wondering if there is a better place to put the `LINKER:/Brepro` option, as the subproject toolchain files are used for different compilers (MSVC and clang from amd-llvm / amd-hip).
 
Response:
> Here's my take on your specific question about `/Brepro` placement:
> 
> The current placement works — both MSVC link.exe and lld-link (used by amd-llvm/amd-hip) support /Brepro, so the unconditional WIN32 guard is functionally correct.
> 
>   But there's a better home for it. The toolchain setup function _therock_cmake_subproject_setup_toolchain already sets `CMAKE_EXE_LINKER_FLAGS_INIT` and `CMAKE_SHARED_LINKER_FLAGS_INIT` (lines 1495-1496), with different paths for MSVC vs amd-llvm toolchains. Adding `/Brepro` there would:
> 
> 1. Co-locate linker config — all linker flags in one place rather than split between the toolchain function and the init file
> 2. Make toolchain-specificity explicit — even though both toolchains support it today, the syntax differs slightly (`/Brepro` for MSVC link.exe, `-Brepro` for lld-link when passed directly)
> 3. Follow LLVM's pattern — LLVM's own `HandleLLVMOptions.cmake` conditionally adds `/Brepro` to compiler flags only for clang-cl, and checks for `/INCREMENTAL` compatibility
> 
> The init file (where the PR puts it) is more of a grab-bag of project-level setup. The toolchain file is where
> compiler/linker identity is established, which is the more natural home.
> 
> One additional note: LLVM adds `/Brepro` as a compiler flag (to `CMAKE_C_FLAGS`/`CMAKE_CXX_FLAGS`), not just a linker flag. This also zeros timestamps in `.obj` files. The PR only does the linker side. For full reproducibility of intermediate artifacts you'd want both, but for the stated goal (ccache hit rate improvement via deterministic final output) the linker flag is the important one.

build_tools/setup_ccache.py
ScottTodd, line 1: > ## Test Plan 
> Validate cache hitrate on Windows following subsequent buil.ds

Would you expect to see cache hits on the PR itself? Maybe after some retries? https://github.com/ROCm/TheRock/actions/runs/24771044846/job/72492570434?pr=4419#step:17:69

I don't think this could make the builds _worse_, but I'd like to see a bit more evidence of it helping before we call it done and move on. We've seen before where ccache works well for local builds but gets low cache hit rates on CI due to a number of factors (between the build system code, workflow code, cache server, runner machines, etc.) so I want to make sure that's all accounted for.

ScottTodd (commented):
Sorry for the delay reviewing, I've been traveling. The approach here seems reasonable to me, just a few questions on some of the details.

cmake/therock_subproject.cmake
ScottTodd: Hmmm... that later location is for flag forwarding, so what you have may be better actually.

Another option would be extending the default flag lists up here to include linker flags too: https://github.com/ROCm/TheRock/blob/faadaa4d93c6096034295700c53add9cf811a376/cmake/therock_subproject.cmake#L59-L88

ScottTodd (approved):
Eh, I'm on the fence about the location in the code for the flag. Seems okay where it is.

build_tools/setup_ccache.py
astrelsky: During testing I noticed that the ccache storage size was very small and a build of TheRock and the pytorch wheels exceeded that size. I can't help but suspect that was the cause of not seeing an improvement in ci, especially with multiple different builds for other gfx going on in parallel whereas my local build was only for one.

Note that because a slopiness change is included for proper precompiled header usage, I suspect it will get worse until all runs include this change and then it will improve. Unless that was due to the cache storage being too small. I'll never know for sure.

cmake/therock_subproject.cmake
amd-nicknick: Yeah, I was debating with myself on this one as well. My arguments for keeping it in init file are:
1. `_therock_cmake_subproject_setup_toolchain` block looks more like dealing with toolchain specific behaviors and forwarding appropriate flags. I think `/Brepro` works more like a "We MUST have this in all components of builds regardless of toolchain", hence in the init file where rest of the "global policies" are placed.
2. As for `THEROCK_AMD_LLVM_DEFAULT_CXX_FLAGS`, that only applies to llvm targets (right?), again we need ALL stuff produced by TheRock to have `/Brepro`, so that wouldn't work.
    * Unless you're proposing we create some sort of `THEROCK_DEFAULT_CXX_FLAGS`, which gets appended to everything including compiler bootstrap? But wouldn't that be exactly where init file should be used?

amd-nicknick: > My code review agent also pointed me at LLVM's `HandleLLVMOptions.cmake` where `Brepro` is used, so we might want to try matching that code style: https://github.com/llvm/llvm-project/blob/0d1bf3ac8eb10ecf062a938955637f8f809c0380/llvm/cmake/modules/HandleLLVMOptions.cmake#L685-L706

I think that's just wrong lol. The `/Brepro` flag is not documented by MSFT, but it's only used in `link.exe` & `lib.exe` linkers.

The reason LLVM can get away with implementing their way is due to the entire toolchain gets built by MSVC, no compiler bootstrapping happens.

In our scenario, the .obj file's timestamp won't be a hashed value (clang doesn't support `/Brepro` unless using clang-cl, which we don't use anyway).
But that doesn't matter at all since we shouldn't care what output files are generated. As long as the artifacts (libs / executable) are stable, it all works well.

amd-nicknick: @ScottTodd any suggestions on the init file & flags are greatly appreciated!

build_tools/setup_ccache.py
amd-nicknick: @astrelsky, the local ccache folder for CI gets cleaned up every time it rebuilds. We have a remote ccache setup in CI.
https://github.com/ROCm/TheRock/blob/1611d700be66fb1fa8b049dbf7f1dfbe8dd2fc4e/build_tools/setup_ccache.py#L40-L63
https://github.com/ROCm/TheRock/blob/1611d700be66fb1fa8b049dbf7f1dfbe8dd2fc4e/.github/workflows/build_windows_artifacts.yml#L116-L122

So the `max_size` shouldn't affect anything, at least on the CI side.

amd-nicknick: @ScottTodd I'm not sure how we're currently trimming the bazel remote cache. From testing it seems like Windows has a significantly larger cache size compared to Linux builds. Can we make sure it's won't run out of space?

Also @astrelsky made a good point on the sloppiness. Unless this change gets merged in & subsequent CI runs all adopt the flag, the caches will be incompatible across dev / this branch. Therefore, reruns will not show any benefit.

astrelsky: Is it the max size for the local or remote cache though? I don't see why there would be both a local and remote cache, it seems redundant. Obviously I understand that local is way faster than remote, but that's not the point.

amd-nicknick: max_size only applies to local cache. It is important for local rebuilds (when you manually clone & iterate with TheRock). 
For CIs these don't help as much but still could reduce remote cache lookups (Some TUs might be duplicated)

astrelsky: Oh ok that's good to know. I'm assuming that the ccache stats are coming from the remote and that the remote cache is larger than 5GB?

ScottTodd: > Note that because a slopiness change is included for proper precompiled header usage, I suspect it will get worse until all runs include this change and then it will improve.

> Also @astrelsky made a good point on the sloppiness. Unless this change gets merged in & subsequent CI runs all adopt the flag, the caches will be incompatible across dev / this branch. Therefore, reruns will not show any benefit.

Can we set `namespace` in ccache's config (or the `CCACHE_NAMESPACE` env var)? https://ccache.dev/manual/4.9.html#_configuration_options

> **namespace (`CCACHE_NAMESPACE`)**
> If set, the namespace string will be added to the hashed data for each compilation. This will make the associated cache entries logically separate from cache entries with other namespaces, but they will still share the same storage space. Cache entries can also be selectively removed from the local cache with the command line option `--evict-namespace`, potentially in combination with `--evict-older-than`.
>
> For instance, if you use the same local cache for several disparate projects, you can use a unique namespace string for each one. This allows you to remove cache entries that belong to a certain project if you stop working with that project.

We've generally used a versioned namespace or cache key that we update when making substantial changes to the cache settings. For example: https://github.com/ROCm/TheRock/blob/ee27e0a7a8b811791cb8c4f759e893083e7c20b6/.github/workflows/build_windows_artifacts.yml#L119-L127

amd-nicknick: Added, this should make eviction in future easier, but won't help this time :)

amd-nicknick:
The last CI looks good on the cache ratio:
https://github.com/ROCm/TheRock/actions/runs/24985166194/job/73156746458
https://github.com/ROCm/TheRock/actions/runs/24985166194/job/73162319806
https://github.com/ROCm/TheRock/actions/runs/24985166194/job/73173840982

Let's wait for this CI round.

build_tools/github_actions/build_configure.py
```diff
@@ -81,6 +81,10 @@ def build_compiler_launcher(
 platform_options = {
     "windows": [
         "-DTHEROCK_BACKGROUND_BUILD_JOBS=4",
+        # ccache does not support MSVC's /Zi flag. Setting Embedded instructs
+        # CMake to compile with /Z7 (or equivalent), which ccache supports.
+        # See: https://github.com/ccache/ccache/issues/1040
+        "-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded",
```
ScottTodd: I don't think this should be needed, see where we already have it here: https://github.com/ROCm/TheRock/blob/e3bef8a43b95addccc3cd9490d5858de3d8ff050/cmake/therock_compiler_config.cmake#L6-L11

README.md
```diff
@@ -299,8 +299,44 @@ cmake --build build
...
+
+> [!TIP]
+> ccache [does not support](https://github.com/ccache/ccache/issues/1040)
+> MSVC's `/Zi` flag which may be set by default when a project (e.g. LLVM) opts
+> in to
+> [policy CMP0141](https://cmake.org/cmake/help/latest/policy/CMP0141.html).
+> Setting
+> [`-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html)
+> instructs CMake to compile with `/Z7` or equivalent, which is supported by
+> ccache.
```
ScottTodd: (same for all these other references to the flag)

build_tools/github_actions/build_configure.py
amd-nicknick: Great catch, removed.

amd-nicknick:
Hm for some reason CI doesn't like that once I removed `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT `, this is from latest CI vs one commit before for compiler:
Before:
```
  Hits:                        19557 / 19609 (99.73%)
    Direct:                    19557 / 19557 (100.0%)
    Preprocessed:                  0 / 19557 ( 0.00%)
  Misses:                         52 / 19609 ( 0.27%)
Uncacheable calls:               258 / 19871 ( 1.30%)
  Compilation failed:             13 /   258 ( 5.04%)
  Preprocessing failed:           28 /   258 (10.85%)
  Unsupported code directive:      2 /   258 ( 0.78%)
  Unsupported compiler option:   211 /   258 (81.78%)
  Unsupported source encoding:     4 /   258 ( 1.55%)
Errors:                            4 / 19871 ( 0.02%)
  Internal error:                  4 /     4 (100.0%)
```

After:
```
Cacheable calls:               10339 / 19871 (52.03%)
  Hits:                         1547 / 10339 (14.96%)
    Direct:                     1388 /  1547 (89.72%)
    Preprocessed:                159 /  1547 (10.28%)
  Misses:                       8792 / 10339 (85.04%)
Uncacheable calls:              9528 / 19871 (47.95%)
  Compilation failed:             13 /  9528 ( 0.14%)
  Preprocessing failed:           28 /  9528 ( 0.29%)
  Unsupported code directive:      2 /  9528 ( 0.02%)
  Unsupported compiler option:  9481 /  9528 (99.51%)
  Unsupported source encoding:     4 /  9528 ( 0.04%)
Errors:                            4 / 19871 ( 0.02%)
  Internal error:                  4 /     4 (100.0%)
```
Checking repro locally

amd-nicknick:
I know what's going on here..
Various subproject & 3rd party libs explicitly inject `/Zi` into its CXX flag.

https://github.com/ROCm/rocm-systems/blob/d39c282dc2e7487612766f02e43f115fb923190a/projects/clr/hipamd/CMakeLists.txt#L34

https://github.com/OpenMathLib/OpenBLAS/blob/10cf63eea44f53176444c3767c0a5a8843e91485/CMakeLists.txt#L234

@ScottTodd, I guess this brings me to a larger feature addition I wanted for TheRock: Windows PDB release support.
We should ditch `RelWithDebInfo`, always build everything with `/Z7`, then split the binaries & PDBs, release both separately.
I see we kind of have this flow for Linux, are you aware if anyone is working on this?

Pros:
* We always get corresponding debug symbols.
* We could host a symbol server & WinDbg automatically grabs correct symbols when doing live & postmortem analysis.

Cons:
* Local build size will increase due to `/Z7`, even if the user doesn't need symbols.
* Cache size will increase.

If we agree on doing this support, I'd prefer to just always set https://github.com/ROCm/TheRock/blob/e3bef8a43b95addccc3cd9490d5858de3d8ff050/cmake/therock_compiler_config.cmake#L6-L11 to `Embedded` for all build types for now.

ScottTodd:
> Various subproject & 3rd party libs explicitly inject `/Zi` into its CXX flag.
> 
> https://github.com/ROCm/rocm-systems/blob/d39c282dc2e7487612766f02e43f115fb923190a/projects/clr/hipamd/CMakeLists.txt#L34
> 
> https://github.com/OpenMathLib/OpenBLAS/blob/10cf63eea44f53176444c3767c0a5a8843e91485/CMakeLists.txt#L234

We should fix those.

> @ScottTodd, I guess this brings me to a larger feature addition I wanted for TheRock: Windows PDB release support.
> We should ditch `RelWithDebInfo`, always build everything with `/Z7`, then split the binaries & PDBs, release both separately.
> I see we kind of have this flow for Linux, are you aware if anyone is working on this?

I think this would be reasonable. I'm not aware of anyone working on that yet.

ScottTodd:
Still just seeing ~5% cache hit rates on CI runs here: https://github.com/ROCm/TheRock/actions/runs/25155042327/job/73739070093?pr=4419. There may be something about the bazelremote cache server that is interfering, if local runs are getting more cache hits. @amd-justchen did someone get around to productionizing the cache server (automating deployment, documenting the setup, writing playbooks for how to check server-side logs and other metrics, etc.)?

amd-justchen:
> Still just seeing ~5% cache hit rates on CI runs here: https://github.com/ROCm/TheRock/actions/runs/25155042327/job/73739070093?pr=4419. There may be something about the bazelremote cache server that is interfering, if local runs are getting more cache hits. @amd-justchen did someone get around to productionizing the cache server (automating deployment, documenting the setup, writing playbooks for how to check server-side logs and other metrics, etc.)?

I think these tasks need looking into still. It might be tracked in the related issue https://github.com/ROCm/TheRock/issues/4195 by @subodh-dubey-amd.

amd-nicknick:
With the flag set & attempt 2 on CI, we can see the eligible cache rate jumped up in CI.
## Foundation
```
Cacheable calls:               1004 / 1128 (89.01%)
  Hits:                         987 / 1004 (98.31%)
    Direct:                     987 /  987 (100.0%)
    Preprocessed:                 0 /  987 ( 0.00%)
  Misses:                        17 / 1004 ( 1.69%)
Uncacheable calls:              124 / 1128 (10.99%)
  Preprocessing failed:           1 /  124 ( 0.81%)
  Unsupported compiler option:  119 /  124 (95.97%)
  Unsupported source encoding:    4 /  124 ( 3.23%)
```
## Compiler
```
Cacheable calls:               19609 / 19871 (98.68%)
  Hits:                        19557 / 19609 (99.73%)
    Direct:                    19557 / 19557 (100.0%)
    Preprocessed:                  0 / 19557 ( 0.00%)
  Misses:                         52 / 19609 ( 0.27%)
Uncacheable calls:               258 / 19871 ( 1.30%)
  Compilation failed:             13 /   258 ( 5.04%)
  Preprocessing failed:           28 /   258 (10.85%)
  Unsupported code directive:      2 /   258 ( 0.78%)
  Unsupported compiler option:   211 /   258 (81.78%)
  Unsupported source encoding:     4 /   258 ( 1.55%)
Errors:                            4 / 19871 ( 0.02%)
  Internal error:                  4 /     4 (100.0%)
```
## Math libs (gfx120X-all)
```
Cacheable calls:               9054 / 9070 (99.82%)
  Hits:                        7611 / 9054 (84.06%)
    Direct:                    1755 / 7611 (23.06%)
    Preprocessed:              5856 / 7611 (76.94%)
  Misses:                      1443 / 9054 (15.94%)
Uncacheable calls:               16 / 9070 ( 0.18%)
  Compilation failed:             2 /   16 (12.50%)
  Preprocessing failed:          13 /   16 (81.25%)
  Unsupported code directive:     1 /   16 ( 6.25%)
```
What remains is the hit rate for repeated builds could be more desirable. My guess is on the Bazel cache server's eviction policy (perhaps it's not hot enough?)
All the code changes we have should be enough (for now). Let's merge this in, that way we could get a better reading of the cache utilization rate.

amd-nicknick:
> > @ScottTodd, I guess this brings me to a larger feature addition I wanted for TheRock: Windows PDB release support.
> > We should ditch `RelWithDebInfo`, always build everything with `/Z7`, then split the binaries & PDBs, release both separately.
> > I see we kind of have this flow for Linux, are you aware if anyone is working on this?
> 
> I think this would be reasonable. I'm not aware of anyone working on that yet.

#5004 to track that work.

cmake/therock_compiler_config.cmake
```diff
@@ -7,7 +7,7 @@
 # This improves compatibility with ccache and sccache:
 # https://github.com/ccache/ccache/issues/1040
 if(WIN32 AND NOT DEFINED CMAKE_MSVC_DEBUG_INFORMATION_FORMAT)
-  set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "$<$<CONFIG:Debug,RelWithDebInfo>:Embedded>")
+  set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT "Embedded")
```
ScottTodd, line 10: We should have thought this through more carefully together with https://github.com/ROCm/TheRock/issues/5004. This _always_ enables debug information through MSVC, even in release builds. This added 10GB of LLVM .lib files to our devel packages (uncompressed, ~3GB compressed).

I don't see how this could be connected to ccache hit rates - release builds would have had debug information disabled and no setting change here would have mattered.

Can we revert this change and only proceed with it when made together with the other ideas in https://github.com/ROCm/TheRock/issues/5004?

amd-nicknick: > Can we revert this change and only proceed with it when made together with the other ideas in #5004?

No unless we're willing to sacrifice cache hit rate:

* Remove hardcoded /Zi from CLR, need this patch: https://github.com/ROCm/rocm-systems/pull/6239
* OpenBLAS does it as well: https://github.com/OpenMathLib/OpenBLAS/blob/develop/CMakeLists.txt#L234

We likely can't do anything with the OpenBLAS one from itself. At most we could do some CMake massage and override it to /Z7 only for `host-blas`.
Without getting rid of these two, `gfx1201` build sits at 66.7% cachable rate (the real rate, not from `ccache -s`)

<details>

<summary>
ccache log diagnostics for gfx1201 local build (latest main + revert this line)
</summary>

```
================================================================================
CCACHE DEBUG LOG ANALYSIS - VERBOSE COMPONENT REPORT
(CMake try-compile / CMakeScratch excluded; use --all to include)
================================================================================

GLOBAL SUMMARY
--------------------------------------------------
  Total compilations:      28062
  (excluded try-compile:     808)
  Hit:                      4199  ( 15.0%)
  Miss:                    14518  ( 51.7%)
  Uncacheable:              9345  ( 33.3%)

  Cacheable rate:          18717  ( 66.7%)  [hit + miss: could benefit from cache]
  Uncacheable rate:         9345  ( 33.3%)  [bypassed cache entirely]

================================================================================
PER-COMPONENT BREAKDOWN
================================================================================

  third-party/host-blas/host-blas
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Total compilations: 6961
    Cache hit:               2  (  0.0%)
    Cache miss:           2324  ( 33.4%)
    Uncacheable:          4635  ( 66.6%)
    Cacheable rate:      33.4%  (2326/6961)
    Uncacheable rate:    66.6%  (4635/6961)

    Uncacheable reasons:
      [unsupported_compiler_option] x4635
        -> /Zi  (4635 occurrences)


  third-party/host-blas/host-blas64
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Total compilations: 6961
    Cache hit:               2  (  0.0%)
    Cache miss:           2324  ( 33.4%)
    Uncacheable:          4635  ( 66.6%)
    Cacheable rate:      33.4%  (2326/6961)
    Uncacheable rate:    66.6%  (4635/6961)

    Uncacheable reasons:
      [unsupported_compiler_option] x4635
        -> /Zi  (4635 occurrences)


  core/clr
  ~~~~~~~~
    Total compilations: 294
    Cache hit:              89  ( 30.3%)
    Cache miss:            165  ( 56.1%)
    Uncacheable:            40  ( 13.6%)
    Cacheable rate:      86.4%  (254/294)
    Uncacheable rate:    13.6%  (40/294)

    Uncacheable reasons:
      [unsupported_compiler_option] x40
        -> /Zi  (40 occurrences)


  compiler/amd-llvm
  ~~~~~~~~~~~~~~~~~
    Total compilations: 3792
    Cache hit:               0  (  0.0%)
    Cache miss:           3790  ( 99.9%)
    Uncacheable:             2  (  0.1%)
    Cacheable rate:      99.9%  (3790/3792)
    Uncacheable rate:     0.1%  (2/3792)

    Uncacheable reasons:
      [unsupported_code_directive] x2
        -> .incbin directive in source code  (2 occurrences)


  core/ocl-clr
  ~~~~~~~~~~~~
    Total compilations: 279
    Cache hit:              89  ( 31.9%)
    Cache miss:            165  ( 59.1%)
    Uncacheable:            25  (  9.0%)
    Cacheable rate:      91.0%  (254/279)
    Uncacheable rate:     9.0%  (25/279)

    Uncacheable reasons:
      [unsupported_compiler_option] x25
        -> /Zi  (25 occurrences)


  math-libs/hipCUB
  ~~~~~~~~~~~~~~~~
    Total compilations: 109
    Cache hit:               0  (  0.0%)
    Cache miss:            106  ( 97.2%)
    Uncacheable:             3  (  2.8%)
    Cacheable rate:      97.2%  (106/109)
    Uncacheable rate:     2.8%  (3/109)

    Uncacheable reasons:
      [preprocessor_error] x3
        -> preprocessor error  (3 occurrences)


  math-libs/rocRAND
  ~~~~~~~~~~~~~~~~~
    Total compilations: 112
    Cache hit:               0  (  0.0%)
    Cache miss:            109  ( 97.3%)
    Uncacheable:             3  (  2.7%)
    Cacheable rate:      97.3%  (109/112)
    Uncacheable rate:     2.7%  (3/112)

    Uncacheable reasons:
      [preprocessor_error] x3
        -> preprocessor error  (3 occurrences)


  ml-libs/MIOpen
  ~~~~~~~~~~~~~~
    Total compilations: 770
    Cache hit:               0  (  0.0%)
    Cache miss:            769  ( 99.9%)
    Uncacheable:             1  (  0.1%)
    Cacheable rate:      99.9%  (769/770)
    Uncacheable rate:     0.1%  (1/770)

    Uncacheable reasons:
      [unsupported_code_directive] x1
        -> .incbin directive in source code  (1 occurrence)


  math-libs/rocThrust
  ~~~~~~~~~~~~~~~~~~~
    Total compilations: 228
    Cache hit:               0  (  0.0%)
    Cache miss:            227  ( 99.6%)
    Uncacheable:             1  (  0.4%)
    Cacheable rate:      99.6%  (227/228)
    Uncacheable rate:     0.4%  (1/228)

    Uncacheable reasons:
      [preprocessor_error] x1
        -> preprocessor error  (1 occurrence)

================================================================================
COMPONENT SUMMARY TABLE
================================================================================

  Component                            Total     Hit    Miss  Uncache  Cache%
  ---------------------------------  ------  ------  ------  -------  ------
  third-party/host-blas/host-blas       6961       2    2324     4635   33.4%
  third-party/host-blas/host-blas64     6961       2    2324     4635   33.4%
  third-party/grpc                      2877    2877       0        0  100.0%
  third-party/fftw3/build_fftw3          465     465       0        0  100.0%
  third-party/fftw3/build_fftw3f         465     465       0        0  100.0%
  core/clr                               294      89     165       40   86.4%
  compiler/amd-llvm                     3792       0    3790        2   99.9%
  core/ocl-clr                           279      89     165       25   91.0%
  core/ocl-icd                            23      23       0        0  100.0%
  compiler/amd-comgr                      18       1      17        0  100.0%
  math-libs/hipCUB                       109       0     106        3   97.2%
  math-libs/rocRAND                      112       0     109        3   97.3%
  third-party/googletest                   7       0       7        0  100.0%
  ml-libs/MIOpen                         770       0     769        1   99.9%
  compiler/hipify                         40       0      40        0  100.0%
  math-libs/hipFFT                        16       0      16        0  100.0%
  math-libs/rocThrust                    228       0     227        1   99.6%
  ml-libs/composable_kernel                4       0       4        0  100.0%
  compiler/hipcc                           4       4       0        0  100.0%
  math-libs/BLAS/hipBLAS                  91       0      91        0  100.0%
  math-libs/BLAS/hipBLASLt               183       1     182        0  100.0%
  math-libs/BLAS/hipSOLVER                53       0      53        0  100.0%
  math-libs/BLAS/rocBLAS                 618       0     618        0  100.0%
  math-libs/BLAS/rocSOLVER               528       2     526        0  100.0%
  ml-libs/hipDNN                         414       0     414        0  100.0%
  ml-libs/hipblasltprovider               29       0      29        0  100.0%
  ml-libs/hipdnn_integration_tests        35       0      35        0  100.0%
  ml-libs/hipdnn_samples                  16       0      16        0  100.0%
  ml-libs/miopenprovider                  61       0      61        0  100.0%
  third-party/spdlog                       6       0       6        0  100.0%
  base/rocm-core                           2       2       0        0  100.0%
  core/hipInfo                             1       0       1        0  100.0%
  core/rocm-kpack                          7       7       0        0  100.0%
  math-libs/rocFFT                        99       2      97        0  100.0%
  third-party/Catch2                     106     106       0        0  100.0%
  math-libs/BLAS/hipSPARSE               191       0     191        0  100.0%
  math-libs/BLAS/rocSPARSE              1119       0    1119        0  100.0%
  math-libs/hipRAND                        6       0       6        0  100.0%
  math-libs/rocPRIM_tests                 84       0      84        0  100.0%
  math-libs/rocWMMA                      944      60     884        0  100.0%
  third-party/flatbuffers                 42       0      42        0  100.0%
  third-party/fmt                          2       2       0        0  100.0%
```
</details>

amd-nicknick: > We should have thought this through more carefully together with #5004. This _always_ enables debug information through MSVC, even in release builds. This added 10GB of LLVM .lib files to our devel packages (uncompressed, ~3GB compressed).
This is an expected increase.

I'm failing to see the problem here..? When we ship the devel packages containing the lib files, aren't we supporting the use case of statically linking amd-llvm in user's application?

In that case, shouldn't we keep the debug information so that the user can actually debug the entire stack? The application could then choose to include / strip / split the symbols depending on their distribution / use case. I'd argue us not shipping the debug symbols makes the libs less useful (vs. user doing a complete rebuild), especially since ROCm is open-source, there's nothing to encapsulate.

After all, the devel packages are for developers, why is size a concern?

Let me know if I misunderstood anything here :)

ScottTodd: > We likely can't do anything with the OpenBLAS one from itself.

We can patch it:
* https://github.com/ROCm/TheRock/tree/main/patches
* https://github.com/ROCm/TheRock/tree/main/patches/third-party

> I'm failing to see the problem here..? When we ship the devel packages containing the lib files, aren't we supporting the use case of statically linking amd-llvm in user's application?

A `Release` build and a `RelWithDebInfo` build are different. The build system should respect what was selected. Also no, we do not support downstream projects statically linking against LLVM (an implementation detail of ROCm).

> After all, the devel packages are for developers, why is size a concern?

Size is always a concern, for developers, CI systems, and users.
* Larger packages increase CI time (build/upload/download)
* Larger packages increase cloud storage costs (even once we institute a retention policy)
* The dev artifacts are included in tarball (archive) releases, and this ~doubled the compressed/uncompressed tarball sizes
  * Compressed: 2.9GB before, 4.8GB after
  * Uncompressed: 13.5GB before, 23.4GB after

ScottTodd: Filed https://github.com/ROCm/TheRock/issues/5341 for the size regression

astrelsky: > I'm failing to see the problem here..? When we ship the devel packages containing the lib files, aren't we supporting the use case of statically linking amd-llvm in user's application?
> 

There should be no debug info for the release builds, so no `/Z*` and thus no problem.

amd-nicknick: Thanks @ScottTodd!
> We can patch it:

PR opened here: https://github.com/ROCm/TheRock/pull/5370

> A `Release` build and a `RelWithDebInfo` build are different. The build system should respect what was selected. 

This is what I've been complaining though, `RelWithDebInfo` is not just "Release but with additional debug info enabled", it is "Release but with **some optimization disabled** and debug info". 

We cannot lose the optimization, but we still need the debug information for release builds.

`RelWithDebInfo` has its own place: I use it for local builds where some issues might have performance implications, but I'd like to have inlining disabled to ease stepping functions. I can trust `RelWithDebInfo` to give me exactly that & we're not breaking it here.

> Also no, we do not support downstream projects statically linking against LLVM (an implementation detail of ROCm).
> Size is always a concern, for developers, CI systems, and users.

Cool, thanks for clarifying for me :) Then I'm not sure about who is exactly consuming these dev artifacts?
Can we just strip out the libs in this case?

I can totally understand about the size, I'm cool with reverting this line after we get the two PRs merged in (or if you believe the size regression outweigh the cache, we could do that now).
* host-blas: https://github.com/ROCm/TheRock/pull/5370
* CLR: https://github.com/ROCm/rocm-systems/pull/6239

I personally think that if we're choosing between size / debug information, we should sway towards better debug experience. Especially since this does not impact the runtime end-user installation whatsoever.

ScottTodd: > Especially since this does not impact the runtime end-user installation whatsoever.

I'm not following here. The settings applied here directly affected end-user installations. The 7.13.0 release we just published (https://repo.amd.com/rocm/tarball-multi-arch/, https://repo.amd.com/rocm/whl-multi-arch/rocm-sdk-devel/) has packages that are ~2x as large as they should be.

astrelsky: > Thanks @ScottTodd!
> > We can patch it:
> 
> PR opened here: https://github.com/ROCm/TheRock/pull/5370
> 
> > A `Release` build and a `RelWithDebInfo` build are different. The build system should respect what was selected. 
> 
> This is what I've been complaining though, `RelWithDebInfo` is not just "Release but with additional debug info enabled", it is "Release but with **some optimization disabled** and debug info". 
> 
> We cannot lose the optimization, but we still need the debug information for release builds.
> 
> `RelWithDebInfo` has its own place: I use it for local builds where some issues might have performance implications, but I'd like to have inlining disabled to ease stepping functions. I can trust `RelWithDebInfo` to give me exactly that & we're not breaking it here.
> 
> > Also no, we do not support downstream projects statically linking against LLVM (an implementation detail of ROCm).
> > Size is always a concern, for developers, CI systems, and users.
> 
> Cool, thanks for clarifying for me :) Then I'm not sure about who is exactly consuming these dev artifacts?
> Can we just strip out the libs in this case?
> 
> I can totally understand about the size, I'm cool with reverting this line after we get the two PRs merged in (or if you believe the size regression outweigh the cache, we could do that now).
> * host-blas: https://github.com/ROCm/TheRock/pull/5370
> * CLR: https://github.com/ROCm/rocm-systems/pull/6239
> 
> I personally think that if we're choosing between size / debug information, we should sway towards better debug experience. Especially since this does not impact the runtime end-user installation whatsoever.

Right and `Release` is full optimizations with no debug info.

amd-nicknick: I wasn't aware the dist tarball & py pkgs are growing as well.
Actions:
1. https://github.com/ROCm/TheRock/issues/5341
    Fix the size regression right now
    https://github.com/ROCm/TheRock/pull/5372
2. CLR & host-blas patch to merge in to bring back the cache ratio
3. Redesign the way we do debug symbol processing for Linux/Windows + Tarball / Python + Symbol server

PR #3973, opened 2026-03-13
Add Windows packaging requirements RFC document

LiamfBerry:
## Motivation

Packaging requirements for windows systems must be updated to align with TheRock's packaging and distribution strategy. The Product Management team is proposing the following changes and requirements for cross-platform delivery of ROCm on Windows.

JIRA ID
ROCM-20689

Proposed Requirements

1. Standardize packaging behavior across the Windows ecosystem. Includes support for MSI installers, and WinGet.
2. Guarantee predictable upgrade behavior, multi-version support/versioning rules, and compatibility with OS package managers
3. Comply with legal, licensing, and redistribution rules
4. Support automated packaging workflows in TheRock. RFC must define productized deliverables, repository layout, artifact naming, versioning scheme, and how SDK components map to TheRock build outputs.
5. Collect public feedback (internal + external). RFC will be published for review before finalizing packaging architecture.

## Technical Details

- The RFC will define packaging for Windows/winget/MSI packaging as well as pip and zip artifacts.
- Folder structures and API search paths must remain stable.

## Test Plan

Documentation-only RFC. No functional tests applicable.

## Test Result

N/A - see test plan

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4438, opened 2026-04-09
ROCm Archive Packaging Requirements

LiamfBerry:
## Motivation

Packaging requirements for ZIP and TAR archives for Windows and Linux systems must be updated to align with TheRock's packaging and distribution strategy. The Product Management team is proposing the following changes and requirements for cross-platform delivery of ROCm ZIP and TAR packages on Windows and Linux.

Proposed Requirements

1. Standardize packaging behavior and installation support for ROCm ZIP and TAR packages.
2. Comply with legal, licensing, and redistribution rules
3. Support automated packaging workflows in TheRock. RFC must define productized deliverables, repository layout, and artifact naming.
4. Collect public feedback (internal + external). RFC will be published for review before finalizing packaging architecture.

## Technical Details

- The RFC will define packaging for Windows ZIP artifacts and TAR artifacts.
- Folder structures and API search paths must remain stable.

**RFC0013 — ROCm SDK archive packaging (ZIP / TAR)**

- One shared contract for both formats; a final section lists the few
  platform-specific differences (`<os>`/`<ext>`/`<format-dir>`,
  registry vs. shell-config passivity).
- Naming `rocm-sdk[-lite][-gfx<target>]-X.Y.Z-<os>.<ext>`, per-target
  (not per-family); optional host-arch suffix defaulting to `x86_64`.
- No wrapping top-level dir — SDK tree at the archive root
  ("extract and point at it"); version carried in the filename.
- Pure extraction only: no env/`PATH`/registry/shell changes, no
  install scripts; SDK-root-portable tools; per-archive `.sha256`.
- Distributed at `<stream>.repo.amd.com/rocm/core/<format-dir>/` across
  `dev`, `nightly`, `weekly`, `rc`, `stable` (`ltsrc`/`lts` reserved).

## Test Plan

Documentation-only RFC. No functional tests applicable.

## Test Result

N/A - see test plan

## Submission Checklist


docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
```diff
@@ -0,0 +1,96 @@
+---
+author: Liam Berry (LiamfBerry), Saad Rahim (saadrahim)
```
saadrahim: We should cover tar files in Linux as well

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
```diff
@@ -0,0 +1,96 @@
...
+ZIP packages must provide a portable file-tree representation of a Windows ROCm installation. The extracted layout must be identical to what a corresponding MSI installation produces at its versioned install root, ensuring interoperability between the two delivery mechanisms.
+
+ZIP archives are intended for power users, CI pipelines, and offline deployments where MSI installation is unavailable or undesirable.
+
+### Naming Convention
+
+ZIP archives must follow the following naming convention:
+
+```
+rocm-<component>-X.Y.Z.zip
```
saadrahim: No zip file granualarity

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
```diff
@@ -0,0 +1,96 @@
...
+- WSL and WSL2 package requirements.
+
+## ZIP Package Requirements
+
+### Package Format
+
+ZIP packages must provide a portable file-tree representation of a Windows ROCm installation. The extracted layout must be identical to what a corresponding MSI installation produces at its versioned install root, ensuring interoperability between the two delivery mechanisms.
+
+ZIP archives are intended for power users, CI pipelines, and offline deployments where MSI installation is unavailable or undesirable.
+
```
saadrahim: Two types:

1. Per GPU architecture zip files
2. Consolidate all GPU architectures in one file

and two variants for the above types: 
1. Contains all build artifacts
2. Cleaned version following content removed: intermediary build files and test files including unit tests,

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
```diff
@@ -0,0 +1,96 @@
...
+- Binaries in `bin\` must not assume a fixed installation prefix set at build time.
+- Path resolution within the SDK must be relative to the extracted root or dynamically resolved at runtime.
+- Tools must remain suitable for CI, offline deployment, and advanced users operating without `ROCM_PATH` or `PATH` pre-configured by the environment.
+
+### Distribution
+
+ZIP archives are hosted on the AMD Official Repository:
+
+```
+repo.amd.com/rocm/windows
```
LiamfBerry: zip files will be in their own folder under rocm-ecosystem

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
LiamfBerry: Added package variants to the RFC

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
LiamfBerry: Removed granularity from package naming formats

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
LiamfBerry: New archive locations has been specified

docs/rfcs/RFC0013-ZIP-Packaging-Requirements.md
LiamfBerry: tar archive requirements have been drafted as well

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,219 @@
...
+
+## TAR Package Requirements
+
+### Package Format
+
+TAR packages must provide a portable file-tree representation of a Linux ROCm installation. The extracted layout must be identical to what a standard installation produces at its versioned install root, ensuring interoperability between the two delivery mechanisms.
+
+TAR archives are intended for power users, CI pipelines, and offline deployments where other installation methods are unavailable or undesirable.
+
+TAR packages will be offered as either a per GPU architecture famiily denoted by `gfx<...>` or contain all GPU families with no denotion is this is the default. Both package types will offer two variants: a full version containing all build artifacts and a cleaned version that has the intermediary build files and test files (including unit tests) removed; this will be denoted with `-lite-`.
```
ScottTodd: cc @marbre 

For the new multi-arch releases we have https://rocm.nightlies.amd.com/tarball-multi-arch/ now, using https://github.com/ROCm/TheRock/blob/main/build_tools/build_tarballs.py. Current behavior is one tarball per GPU _family_ and one for all GPUs.

I'm also considering one tarball per GPU _target_, or one tarball per larger set (e.g. all Instinct, all Radeon).

For example

Tarball name | Currently built? | Notes
-- | -- | --
therock-dist-linux-gfx110X-all | Yes | <ul><li>Contains gfx1100,gfx1101,gfx1102,gfx1103</li><li>Equivalent to `pip install rocm[device-gfx1100,device-gfx1101,device-gfx1102,device-gfx1103]`</li></ul>
therock-dist-linux-gfx1100 | No | 
therock-dist-gfx1151 | Yes | <ul><li>Contains gfx942</li><li>Equivalent to `pip install rocm[device-gfx1151]`</li></ul>
therock-dist-linux-gfx94X-dcgpu | Yes | <ul><li>Contains gfx942</li><li>Equivalent to `pip install rocm[device-gfx942]`</li></ul>
therock-dist-linux-multiarch | Yes | <ul><li>Contains gfx900,gfx908,gfx1010,gfx1100,gfx1200,etc.</li><li>Equivalent to `pip install rocm[device-all]`</li></ul>
therock-dist-linux-instinct | No | Could contain gfx942,gfx950, etc.
therock-dist-linux-radeon | No | Could contain gfx1100,gfx1200, etc.

I think grouping by GPU _family_ as we currently do won't be very useful, as the new python packages and native linux packages are split by _target_, so tarballs would be the only place still using e.g. gfx110X-all instead of 1100,1101,1102,1103.

We could also decide to only distribute tarballs for "all". They'll be quite a bit larger than single-target tarballs (e.g. 3GB for single, 10GB+ for all) though.

Put another way, I'd like to make this change:

```diff
 therock-dist-linux-multiarch
-therock-dist-linux-gfx110X-all
+therock-dist-linux-gfx1100
+therock-dist-linux-gfx1101
+therock-dist-linux-gfx1102
+therock-dist-linux-gfx1103
```

Or simply:
```diff
 therock-dist-linux-multiarch
-therock-dist-linux-gfx110X-all
```

ScottTodd: and we can also drop "multiarch" from the names, so
```diff
-therock-dist-linux-multiarch-7.13.0a20260430.tar.gz
+therock-dist-linux-7.13.0a20260430.tar.gz
```

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+status: draft
+---
+
+# ROCm SDK Archive Packaging for Windows and Linux
+
+With the implementation of TheRock build system, a portable and system-agnostic delivery mechanism is required for the ROCm SDK on Windows to complement MSI-based installation and an equivalent system is required for Linux. This RFC defines the layout, behavioural, and distribution requirements for ZIP and TAR archive packages that deliver the ROCm SDK on Windows and Linux, respectively, aligned with TheRock.
+
+Our goals are to:
+
+1. **Provide a portable, zero-footprint SDK delivery mechanism** suitable for CI pipelines, offline deployments, and advanced users who manage their own SDK roots.
```
ScottTodd: What do you mean by "zero-footprint" here?

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+- Winget package and meta-package requirements.
+- Python pip package requirements.
+- Windows display driver packaging.
+- WSL and WSL2 package requirements.
+
+## ZIP Package Requirements
+
+### Package Format
+
+ZIP packages must provide a portable file-tree representation of a Windows ROCm installation. The extracted layout must be identical to what a corresponding MSI installation produces at its versioned install root, ensuring interoperability between the two delivery mechanisms.
```
ScottTodd: The Windows packaging requirements RFC https://github.com/ROCm/TheRock/pull/3973 says "The installed directory structure must mirror the cross-platform ROCm layout as closely as practical". Should this use similar language instead of "must be identical"?

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+### Directory Layout
+
+The top-level directory inside the archive must be versioned to prevent extraction collisions when multiple versions are present on the same system:
+
+```
+rocm-sdk-X.Y.Z.zip
+  rocm-sdk-X.Y\
+      bin\
```
ScottTodd: You specifically want `X.Y`, not `X.Y.Z` in the inner directory here, right? We don't currently construct an inner directory at all, and this may complicate usage for automation. Where currently you can just extract _any_ given archive (e.g. from https://rocm.nightlies.amd.com/tarball/), here you would need to extract and then also navigate into the subfolder whose name is dynamic (though predictable).

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+### Behavioural Requirements
+
+ZIP packages must be entirely passive with respect to the host system. A ZIP package:
+
+- Must not modify environment variables.
+- Must not append to or modify `PATH`.
+- Must not create, modify, or delete registry entries.
+- Must not execute installation logic, pre/post-install scripts, or elevation prompts.
+
+All system integration, including environment configuration, `PATH` management, and registry entries, is the responsibility of the caller; whether that is the user, a CI script, or a higher-level installer such as the MSI package.
```
ScottTodd: The framing here confuses me. ZIP packages _can't_ do these things, whether we want ours to or not.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+
+Tools and scripts inside ZIP packages must function correctly when the extracted directory is used directly as the SDK root, without any ambient system configuration. Specifically:
+
+- Binaries in `bin\` must not assume a fixed installation prefix set at build time.
+- Path resolution within the SDK must be relative to the extracted root or dynamically resolved at runtime.
+- Tools must remain suitable for CI, offline deployment, and advanced users operating without `ROCM_PATH` or `PATH` pre-configured by the environment.
+
+### Distribution
+
+ZIP archives are hosted on the AMD Official Repository under the structure defined by [RFC0012 Repo Structure](RFC0012-Repo-Structure.md):
```
ScottTodd: Let's get these RFCs reviewed and merged. A few have been open for multiple months and the numbers haven't been "reserved" so these links are going to break depending on which RFC is merged first (https://github.com/ROCm/TheRock/pull/3973 is also trying to use RFC0012)

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+
+This repository serves as the authoritative source for Winget manifest ingestion and internal automation pipelines that consume versioned SDK artifacts directly. ZIP artifacts hosted here must be versioned and integrity-checked.
+
+Additionally the ZIP archive directory will contain a `SHA256SUMS` file that contains hashes for all the provided packages.
+
+```
+<stream>.repo.amd.com/rocm/core/zip/
+  <specified ZIP packages>
+  SHA256SUMS
+```
```
ScottTodd: Hashes for _all_ packages, in a single file? What schema would that file have? How would it be used? What happened to the prior section with one sha256 file per zip file?

```
rocm-sdk-gfx<target>-X.Y.Z.zip
rocm-sdk-gfx<target>-X.Y.Z.zip.sha256
```

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+
+ZIP archives are hosted on the AMD Official Repository under the structure defined by [RFC0012 Repo Structure](RFC0012-Repo-Structure.md):
+
+```
+<stream>.repo.amd.com/rocm/core/zip/
+```
+
+Where `<stream>` is one of `nightly`, `rc`, `stable`, or `lts` as defined by the repository stream model.
+
+This repository serves as the authoritative source for Winget manifest ingestion and internal automation pipelines that consume versioned SDK artifacts directly. ZIP artifacts hosted here must be versioned and integrity-checked.
```
ScottTodd: I don't understand what this is getting at with "Winget manifest ingestion". Seems like that could be out of scope for this RFC, or it could be organized into a single section that talks through what it actually means, instead of being sprinkled throughout like this?

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+All ZIP packages will also have an external sha256 hash validation file for integrity verification. 
+
+### Naming Convention
+
+ZIP archives must follow the following naming convention:
+
+```
+rocm-sdk-X.Y.Z.zip
+rocm-sdk-gfx<target>-X.Y.Z.zip
+```
```
ScottTodd, line 67: We would also need some information about the platform in the filename here, to avoid conflicts between Windows/Linux if they will be placed in the same `<stream>.repo.amd.com/rocm/core/zip/` directory.

Possibly also the CPU architecture (x86/arm/ppc64le/etc.) if we want to be forward-looking with https://github.com/ROCm/TheRock/issues/5518.

For example, see what Python whl files use: https://peps.python.org/pep-0427/#file-name-convention
```
{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl

for example:
rocm_sdk_core-7.13.0a20260511-py3-none-win_amd64.whl
rocm_sdk_core-7.13.0a20260512-py3-none-linux_x86_64.whl
torch-2.10.0+rocm7.14.0a20260525-cp313-cp313-linux_x86_64.whl
torch-2.10.0+rocm7.14.0a20260525-cp313-cp313-win_amd64.whl
```

I don't think just using `.zip` for Windows and `.tar.gz` will be sufficient, as both file formats work on both platforms.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+
+The extracted root of this directory serves as the SDK root for all tools and scripts contained within the archive.
+
+### Behavioural Requirements
+
+TAR packages must be entirely passive with respect to the host system. A TAR package:
+
+- Must not modify environment variables.
+- Must not append to or modify shell configuration files (.bashrc, .zshrc).
+- Must not execute installation logic, pre/post-install scripts, or elevation prompts.
```
ScottTodd: Like above, .tar / .gz files _can't_ do these things, so this framing confuses me.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,227 @@
...
+<stream>.repo.amd.com/rocm/core/zip/
+  <specified ZIP packages>
+  SHA256SUMS
+```
+
+## TAR Package Requirements
+
+### Package Format
+
+TAR packages must provide a portable file-tree representation of a Linux ROCm installation. The extracted layout must be identical to what a standard installation produces at its versioned install root, ensuring interoperability between the two delivery mechanisms.
```
ScottTodd: The RFC structure here with ZIP and the top and TAR at the bottom is confusing. Both are different archives of the same files. What is different? Just which format we use on Linux and Windows?

I think this RFC would be easier to understand if that was outlined at the top and then only platform-specific differences were highlighted, instead of repeating so much information and leaving it up to the reader to spot what is different.

LiamfBerry: Restructured so the requirements are defined once as a shared "Archive Contract", with a "Platform Specifics" section at the bottom that holds the only real differences.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Agreed — reworded to stop disclaiming things an archive can't do.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Same reframe as the ZIP section — see the shared Behavioral Requirements

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Dropped SHA256SUMS entirely — consumption is one-archive-at-a-time, so a combined file was redundant and would have needed its own schema. Kept the per-archive .sha256.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Added a required <os> tag (windows/linux) to the naming convention since the extension alone can't identify the OS. On CPU arch: rather than carry _x86_64 on every name today, the host arch defaults to x86_64 when omitted, and only non-x86 builds get an explicit suffix when #5518 is put into development.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Dropped the inner directory, multi-version extraction will be the user's responsibility to set the destination directory.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Changed language to be softer and in-line with the Windows RFC

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Changed language to be clearer. zero-footprint means it will be self-contained; no trace is left on the hist outside the extracted directory.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: Agreed, Winget package requirements are out of scope in this RFC and is now only mentioned as an example in the distribution section.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: The RFC's per-target scheme (rocm-sdk-gfx<target>-… plus an undenoted "all") matches this now.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: We can consolidate the titles when ready to merge so there is no conflict with other PRs.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,168 @@
...
+  version.txt
+```
+
+This preserves the "extract and point at it" pattern: the extracted contents are the SDK root for all tools and scripts in the archive, with no dynamically-named subfolder to navigate into.
+
+The version is carried in the archive filename rather than an inner directory. When extracting multiple versions side by side, the caller is responsible for choosing distinct destination directories (e.g., `tar -C <dir>` or extracting into a named folder).
+
+### Behavioural Requirements
+
+Extraction is the only action an archive performs. Unpacking an archive must not imply any system integration: all configuration of the host environment is the responsibility of the caller, whether that is the user, a CI script, or a higher-level installer such as the MSI package.
```
ScottTodd: Our archives won't be "self-extracting" after download, so I think we should reword here. The archive itself isn't the thing performing the action. You download a .zip or .tar.gz then extract it yourself (e.g. by double clicking in your OS, opening via 7zip, running through the `tar` command, etc.).

ScottTodd (approved):

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
LiamfBerry: This section has been reworded to be clearer now.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,168 @@
...
+Archives do not contain a wrapping top-level directory. The SDK file-tree sits at the archive root, so extracting an archive directly yields the SDK root:
+
+```
+rocm-sdk-X.Y.Z-<os>.<ext>
+  bin/
+  lib/
+  include/
+  share/
+  tools/
+  version.txt
```
nunnikri, line 105: This version.txt is new, i dont think its available in wheel or native packaging. There is a file .info/version.txt which is available now. thats not needed? or both are needed?
Did we define the format of this file? whether it will be X.Y.X or X.Y.Z-<buildno> or X.Y..Z-<buildno>-<releasetype> or some other format?

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,168 @@
...
+Archives must provide a portable file-tree representation of a ROCm installation. The extracted layout must mirror the platform-native installed layout as closely as practical, ensuring interoperability between the two delivery mechanisms.
+
+Archives are intended for power users, CI pipelines, and offline deployments where platform-native installation is unavailable or undesirable.
+
+### Package Types and Variants
+
+Archives are offered along two axes:
+
+- **Target scope** — either a single GPU target denoted by `gfx<target>` (e.g., `gfx1100`, `gfx1201`), or all GPU targets with no denotation (the default).
+- **Content variant** — either a *full* package containing all build artifacts, or a *lite* package with intermediary build files and test files (including unit tests) removed, denoted by `lite` in the package name.
```
nunnikri: I believe ideally it should be other way. The default package should be compact with all needed libraries, binaries and include files. So that users dont need to install 500+ test binaries who only needed our runtime. Especially because our wheel and native packages doesnt include test packages. 
And an full package option which contains all the unit test and base package. Which we can decide to keep it internal(to match wheel/native packaging)  or to release outside based on demand.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,168 @@
...
+Archives must follow this naming convention, where `<os>` and `<ext>` are platform-specific (see [Platform Specifics](#platform-specifics)):
+
+```
+rocm-sdk-X.Y.Z-<os>.<ext>
+rocm-sdk-gfx<target>-X.Y.Z-<os>.<ext>
+rocm-sdk-lite-X.Y.Z-<os>.<ext>
+rocm-sdk-lite-gfx<target>-X.Y.Z-<os>.<ext>
+```
+
+Where `<target>` is an individual GPU target identifier (e.g., `gfx1100`, `gfx1201`) rather than a GPU family grouping, aligning with Python package and native Linux package distribution conventions.
```
nunnikri, line 69: Currently out tar ball is family based. @ScottTodd any concern on this individual gfx arch tar ball?

ScottTodd: See the prior review thread: https://github.com/ROCm/TheRock/pull/4438#discussion_r3169936871

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
saadrahim: We can remove the version.txt here.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
saadrahim:
```suggestion
- **Content variant** — either a *full* package containing all build artifacts, or a *lite* package with intermediary build files and test files (including unit tests) removed, denoted by `lite` in the package name.
  - <lite> runtimes + compilers only
  - <full> headers, cmake and static libraries, debug, build tools
  - <test> test binaries and test dependencies
```
I think we need to specify these 3 variants, need to name them of course. Follow meta package naming?

LiamfBerry: Removing lite packages and keeping default <name> and <name-test> packages

LiamfBerry: Yes I agree, the RFC has been updated with the new structure for default packages and test packages only

ScottTodd:
Were there still items open for discussion here? I'd like us to finalize and merge some of the active RFCs so we can commit to work on them.

marbre:
Re-requesting a review from @ScottTodd as the RFC was altered after approval (we don't have stale reviews enabled). pre-commit is still failing but this needs to be fixed before it can land @LiamfBerry. Please let's try to finalize and land as this seems to block other RFCs (e.g. #5613).

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Package Types and Variants
+
+Archives are offered along two axes:
+
+- **Target scope** — either a single GPU target denoted by `gfx<target>` (e.g., `gfx1100`, `gfx1201`), or all GPU targets with no denotation (the default).
+- **Content variant** — either a default package containing all build artifacts without the test files, or a *-test* package with basic unit tests, denoted by `-test` in the package name.
+
+- <name> headers, cmake and static libraries, debug, build tools
+- <name-test> test binaries and test dependencies
```
ScottTodd, line 56: nit: Formatting here is not clear on the rendered page, "two axes" but four list items: 

Image

Could reformat using a table or add an intermediate paragraph between the two lists.

docs/rfcs/RFC0013-Archive-Packaging-Requirements.md
ScottTodd: I think we can treat redefining/moving version files/metadata as outside the scope of this RFC. What mostly matters here is the name of the outer archive file itself.

ScottTodd (approved):
This still LGTM from the perspective of TheRock's CI/CD, need more approvals though.

PR #4453, opened 2026-04-10
Update CMakeLists.txt to switch off/on the benchmarks for ASAN build

mdnaveedh:
Disable Google Benchmark in sanitizer builds to fix hipCUB configure failure

Google Benchmark's CMake configuration uses try_run() to detect a regex
backend. Under ASan/TSan builds, the test binaries are compiled with
-fsanitize=address but fail to run because CMake's try_run() does not
set LD_PRELOAD for the sanitizer runtime. This causes all three regex
backends (std::regex, GNU POSIX, POSIX) to fail, and Google Benchmark
aborts with "Failed to determine the source files for the regular
expression backend", breaking the hipCUB configure step.

Guard BUILD_BENCHMARK behind THEROCK_SANITIZER so it is set to OFF
when any sanitizer is active, while preserving the existing behavior
for non-sanitizer builds.

### DIFF
--- a/math-libs/CMakeLists.txt
+++ b/math-libs/CMakeLists.txt
@@ -1,6 +1,15 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
+# Google Benchmark's try_run() regex detection fails under sanitizer builds
+# because the test binaries require LD_PRELOAD for the sanitizer runtime,
+# which CMake's try_run() does not provide.
+if(THEROCK_SANITIZER)
+  set(_THEROCK_BUILD_BENCHMARK OFF)
+else()
+  set(_THEROCK_BUILD_BENCHMARK ${THEROCK_BUILD_TESTING})
+endif()
+
 if(THEROCK_ENABLE_RAND)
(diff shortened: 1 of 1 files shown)


math-libs/CMakeLists.txt
```diff
@@ -1,6 +1,15 @@
...
 # SPDX-License-Identifier: MIT
 
+# Google Benchmark's try_run() regex detection fails under sanitizer builds
+# because the test binaries require LD_PRELOAD for the sanitizer runtime,
+# which CMake's try_run() does not provide.
+if(THEROCK_SANITIZER)
+  set(_THEROCK_BUILD_BENCHMARK OFF)
+else()
+  set(_THEROCK_BUILD_BENCHMARK ${THEROCK_BUILD_TESTING})
+endif()
```
marbre, line 11: Want to add a warning thus a users get's aware that building benchmark will be disabled?

mdnaveedh: It was just a experimental change to check if build works fine with benchmarks disabled. Not planned to merge.
Closing it now.

PR #4457, opened 2026-04-10
Add support for building rocgdb from external source directory

lumachad:
This enables developers to build rocgdb from a custom source location while preserving the TheRock wrapper infrastructure, supporting local development and testing of rocgdb modifications.

Also update README.md to clarify how and what components can be built with external sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,10 +146,8 @@ set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/rocm-systems"
 set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT}" CACHE STRING "Path to rocm-systems superrepo")
 cmake_path(ABSOLUTE_PATH THEROCK_ROCM_SYSTEMS_SOURCE_DIR NORMALIZE)
 
-# Source configuration for ROCgdb.
-set(THEROCK_ROCGDB_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/debug-tools/rocgdb/source")
-set(THEROCK_ROCGDB_SOURCE_DIR "${THEROCK_ROCGDB_SOURCE_DIR_DEFAULT}" CACHE STRING "Path to rocgdb's sources")
-cmake_path(ABSOLUTE_PATH THEROCK_ROCGDB_SOURCE_DIR NORMALIZE)
+# Allow specifying alternative source location for ROCgdb.
+therock_enable_external_source("rocgdb" "${THEROCK_SOURCE_DIR}/debug-tools/rocgdb/source" OFF)
 
 # Allow to specify alternative source locations instead of using
 # repositories tracked in TheRock's `.gitmodules`.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,11 @@ set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/rocm-systems"
 set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT}" CACHE STRING "Path to rocm-systems superrepo")
 cmake_path(ABSOLUTE_PATH THEROCK_ROCM_SYSTEMS_SOURCE_DIR NORMALIZE)
 
+# Source configuration for ROCgdb.
+set(THEROCK_ROCGDB_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/debug-tools/rocgdb/source")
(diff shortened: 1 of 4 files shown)


marbre (commented):
This could use the `therock_enable_external_source` functions or follow how we handle it for rocm-systems as well as rocm-libraries:

https://github.com/ROCm/TheRock/blob/fc07decc78ae4107c3d4b0352e6ea45371b1656d/CMakeLists.txt#L139-L151

Commit 5523139 _removes_ the function use for rccl (which was moved into rocm-systems) but that might give you an idea how this works. Happy to jump on a call if needed.

lumachad:
> This could use the `therock_enable_external_source` functions or follow how we handle it for rocm-systems as well as rocm-libraries:
> 
> https://github.com/ROCm/TheRock/blob/fc07decc78ae4107c3d4b0352e6ea45371b1656d/CMakeLists.txt#L139-L151
> 
> Commit [5523139](https://github.com/ROCm/TheRock/commit/5523139d860b6430aa24baf044e6dfa59921df60) _removes_ the function use for rccl (which was moved into rocm-systems) but that might give you an idea how this works. Happy to jump on a call if needed.

Let me adjust this then.

lumachad:
Updated to use a similar scheme to rocm-systems.

debug-tools/CMakeLists.txt
```diff
@@ -196,6 +196,10 @@ if(THEROCK_ENABLE_ROCGDB)
     message(WARNING "No Python with shared libpython found. rocgdb will be built without Python support.")
   endif()
 
+  if(NOT ${THEROCK_ROCGDB_SOURCE_DIR} STREQUAL ${THEROCK_ROCGDB_SOURCE_DIR_DEFAULT})
+    message(WARNING "Configuring ROCgdb with modified source directory: ${THEROCK_ROCGDB_SOURCE_DIR}")
+  endif()
```
marbre: Seems you'd like to make it more clear to the developer what is happening and think it would be good to warn them. In that case, I'd suggest to rather use `therock_enable_external_source`. The function takes care of some error that can occur and the user explicitly needs to opt into the external source with setting `THEROCK_USE_EXTERNAL_ROCGDB` to `ON`. It will then also error if the external source dir isn't specified or not existing. Also see my code suggestion below. Wdyt?
```suggestion
```

CMakeLists.txt
```diff
@@ -146,6 +146,11 @@ set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/rocm-systems"
 set(THEROCK_ROCM_SYSTEMS_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR_DEFAULT}" CACHE STRING "Path to rocm-systems superrepo")
 cmake_path(ABSOLUTE_PATH THEROCK_ROCM_SYSTEMS_SOURCE_DIR NORMALIZE)
 
+# Source configuration for ROCgdb.
+set(THEROCK_ROCGDB_SOURCE_DIR_DEFAULT "${THEROCK_SOURCE_DIR}/debug-tools/rocgdb/source")
+set(THEROCK_ROCGDB_SOURCE_DIR "${THEROCK_ROCGDB_SOURCE_DIR_DEFAULT}" CACHE STRING "Path to rocgdb's sources")
+cmake_path(ABSOLUTE_PATH THEROCK_ROCGDB_SOURCE_DIR NORMALIZE)
+
 # Allow to specify alternative source locations instead of using
 # repositories tracked in TheRock's `.gitmodules`.
```
marbre:
```suggestion
# Allow to specify alternative source locations instead of using
# repositories tracked in TheRock's `.gitmodules`.
therock_enable_external_source("rocgdb" "${THEROCK_SOURCE_DIR}/debug-tools/rocgdb/sourc" OFF)
```

That should be the only change strictly needed to switch to `therock_enable_external_source`. Adding a note to README.md somewhere around

https://github.com/ROCm/TheRock/blob/371a378c1b0cdaba5c43a1b811263df9f85e70d8/README.md?plain=1#L224

would be welcome.

debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -50,6 +50,12 @@ project(rocgdb C CXX)
 include(ExternalProject)
 include(ProcessorCount)
 
+# Allow external rocgdb source directory.
+if(NOT ROCGDB_SOURCE_DIR)
+  set(ROCGDB_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/source")
+endif()
+message(STATUS "ROCgdb source directory: ${ROCGDB_SOURCE_DIR}")
```
marbre: This is a silent fall back, I'd rather propose to be explicit.

lumachad: I've removed this now.

CMakeLists.txt
lumachad: Thanks. Update coming up shortly.

debug-tools/CMakeLists.txt
lumachad: I see now. Agreed.

lumachad:
Switched to using therock_enable_external_source.

lumachad:
@marbre @geomin12 Ping?

marbre (approved):

PR #4464, opened 2026-04-10
Pass --gpu-targets to artifact splitter

BrianHarrisonAMD:
## Summary
- Passes `THEROCK_AMDGPU_TARGETS` to the artifact splitter via `--gpu-targets` so only per-arch artifacts for the current build's GPU targets are created

## Problem
When multiple CI arch jobs (gfx94X, gfx110X, gfx120X) build MIOpen, the splitter creates per-arch artifacts for ALL architectures found in database filenames. All jobs upload to the same flat S3 prefix, so a later arch job's `miopen_lib_gfx942.tar.zst` (containing only db files) overwrites the earlier gfx94X job's version (which had both CK `.so` files AND db files).

## Changes
Single change in `cmake/therock_artifacts.cmake`: append `--gpu-targets ${THEROCK_AMDGPU_TARGETS}` to the split command args, guarded by the `THEROCK_AMDGPU_TARGETS-NOTFOUND` sentinel.

## Dependencies
- ROCm/rocm-systems#4916 — adds the `--gpu-targets` CLI flag to `split_artifacts.py`

## Test plan
- [x] Dry-run CI with TheRock branch `users/bharriso/dry-run-miopen-multi-arch` (which includes both this change and the rocm-systems bump)


BrianHarrisonAMD:
Leaving as a draft until this can actually land.
Waiting on rocm-systems PR to be merged, and bump to happen.

stellaraccident (commented):
Good catch. I don't think other components have had this kind of overlap/race before.

BradPepersAMD (approved):

BrianHarrisonAMD:
Going to merge.
rocblas test timing out is unrelated, and the sanity test just flagged a node that had GPU issues.

Relevant builds & tests are passing, and the [dry-run draft](https://github.com/ROCm/TheRock/pull/4434) showed another run was passing.

PR #4487, opened 2026-04-13
Fully wire kpack-split into multi-arch Python package CI

marbre:
## Motivation

This adds the missing pieces to produce kpack-split enabled Python packages.
The end-to-end CI produces a flat layout with per-gfx-targets packages rather then the legacy per-family subdirectory layout.

## Technical Details

- Index generation: A flat layout for kpack-split enabled packages is added
  The upload script auto-detects the playout at runtime.
- Device wheels: Fix device extra resolution for kpack-split installs. The rocm meta sdist now exposes explicit device-gfx{N}
  extras per available target alongside the generic `device` extra. On a GPU machine where offload-arch is available the generic extra continues to work as before; on GPU-less machine one can call `device-gfx{N}` directly.
  The extra is passed to `test_rocm_wheels.yml` and `build_portable_linux_pytorch_wheels_ci.yml` to inject the correct per-ISA extras at install time.

Legacy per-family builds are unaffected - all new inputs default to empty/false and leave existing behavior unchanged.

## Test Plan

https://github.com/ROCm/TheRock/actions/runs/24402771972 - CI run with additional change to build with kpack-split enabled.

## Test Result

- ROCm Python packages get successfully produced and installed
- Sanity checks fail
- Legacy builds unaffected (CI on this PR)

## Submission Checklist

### DIFF
--- a/.github/workflows/build_portable_linux_python_packages.yml
+++ b/.github/workflows/build_portable_linux_python_packages.yml
@@ -56,6 +56,9 @@ on:
       package_find_links_url:
         description: URL for pip --find-links to install built packages
         value: ${{ jobs.build_rocm_wheels.outputs.package_find_links_url }}
+      kpack_split:
+        description: "true if this is a kpack-split flat build, false for legacy per-family"
+        value: ${{ jobs.build_rocm_wheels.outputs.kpack_split }}
 
 permissions:
   contents: read
@@ -74,6 +77,7 @@ jobs:
       options: -v /runner/config:/home/awsconfig/
     outputs:
       package_find_links_url: ${{ steps.upload.outputs.package_find_links_url }}
+      kpack_split: ${{ steps.upload.outputs.kpack_split }}
     env:
       AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
       ARTIFACT_RUN_ID: "${{ inputs.artifact_run_id != '' && inputs.artifact_run_id || github.run_id }}"
@@ -98,6 +102,7 @@ jobs:
               --run-github-repo="${{ inputs.artifact_github_repo }}" \
               --stage=all \
               --amdgpu-families="${{ inputs.amdgpu_families }}" \
+              --expand-family-to-targets \
               --output-dir="${{ github.workspace }}"
               # NOTE: artifact_manager.py appends /artifacts to --output-dir
               # in non-flatten mode, so pass workspace root here so that
--- a/.github/workflows/build_portable_linux_pytorch_wheels_ci.yml
+++ b/.github/workflows/build_portable_linux_pytorch_wheels_ci.yml
@@ -44,6 +44,17 @@ on:
         required: false
         type: string
         default: "none"
+      amdgpu_targets:
+        description: >-
+          Comma-separated individual gfx targets for kpack-split builds
+          (e.g. 'gfx942' or 'gfx1200,gfx1201'). Used to install the correct
+          per-target device wheels. Leave empty for legacy per-family builds.
+        type: string
+        default: ""
+      kpack_split:
+        description: "'true' if this is a kpack-split flat build."
+        type: string
+        default: "false"
   workflow_dispatch:
     inputs:
       artifact_group:
@@ -157,6 +168,17 @@ jobs:
           echo "SCCACHE_BUCKET=${SCCACHE_BUCKET}"
           echo "SCCACHE_S3_KEY_PREFIX=${SCCACHE_S3_KEY_PREFIX}"
 
+      - name: Compute ROCm device extras
+        id: device_extras
+        run: |
+          if [ "${{ inputs.kpack_split }}" = "true" ] && [ -n "${{ inputs.amdgpu_targets }}" ]; then
+            extras=$(echo "${{ inputs.amdgpu_targets }}" \
+              | tr ',' '\n' | sed 's/^/device-/' | tr '\n' ',' | sed 's/,$//')
+            echo "value=${extras}" >> "$GITHUB_OUTPUT"
+          else
+            echo "value=" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Build PyTorch wheels
         run: |
           cache_flag=""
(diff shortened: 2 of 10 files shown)


.github/workflows/build_portable_linux_python_packages.yml
ScottTodd, line 1: I'd like to test the equivalent of https://therock-ci-artifacts.s3.amazonaws.com/24402771972-linux/python/index.html from https://github.com/ROCm/TheRock/actions/runs/24402771972 on my Windows dev machine. Can you make the same set of changes to the Windows workflows too, or does that not work yet?

(Though it's generally fine to sequence as Linux first then Windows after during bring-up, as I'm doing on some of my own PRs)

.github/workflows/build_portable_linux_pytorch_wheels_ci.yml
```diff
@@ -157,6 +168,17 @@ jobs:
...
+      - name: Compute ROCm device extras
+        id: device_extras
+        run: |
+          if [ "${{ inputs.kpack_split }}" = "true" ] && [ -n "${{ inputs.amdgpu_targets }}" ]; then
+            # Prefix each comma-separated target with "device-": gfx942,gfx1201 -> device-gfx942,device-gfx1201
+            extras=$(echo "${{ inputs.amdgpu_targets }}" | sed 's/\(^\|,\)/\1device-/g')
+            echo "value=${extras}" >> "$GITHUB_OUTPUT"
+          else
+            echo "value=" >> "$GITHUB_OUTPUT"
+          fi
```
ScottTodd, line 180: We can push this logic down into `build_prod_wheels.py`, pass `--kpack-split` and `--amdgpu-targets` args maybe?

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -144,9 +144,18 @@ jobs:
...
       test_runs_on: ${{ matrix.family_info.test-runs-on }}
-      package_find_links_url: "${{ needs.build_python_packages.outputs.package_find_links_url }}/${{ matrix.family_info.amdgpu_family }}/index.html"
+      package_find_links_url: >-
+        ${{
+          needs.build_python_packages.outputs.kpack_split == 'true'
+          && needs.build_python_packages.outputs.package_find_links_url
+          || format('{0}/{1}/index.html',
+              needs.build_python_packages.outputs.package_find_links_url,
+              matrix.family_info.amdgpu_family)
+        }}
```
ScottTodd, line 156: Could add a TODO to simplify this to just `needs.build_python_packages.outputs.package_find_links_url` once kpack split is always enabled

I hope we don't need this code for long. If we do, some options:
* Do the join in `test_rocm_wheels.yml`
* Do the join `python build_tools/setup_venv.py`

.github/workflows/test_rocm_wheels.yml
```diff
@@ -89,12 +111,27 @@ jobs:
...
       - name: Install rocm[libraries]
         id: install_libraries
         run: |
+          pkg="rocm[libraries]"
+          if [ -n "${{ steps.device_extras.outputs.value }}" ]; then
+            pkg="${pkg%]},${{ steps.device_extras.outputs.value }}]"
+          fi
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages "rocm[libraries]==${{ inputs.rocm_version }}" \
+            --packages "${pkg}==${{ inputs.rocm_version }}" \
```
ScottTodd, line 134: ick, bash logic :/

We could teach setup_venv.py to treat the rocm packages as special and do this expansion in there

build_tools/packaging/python/templates/rocm/setup.py
```diff
@@ -60,6 +60,18 @@ def import_dist_info():
...
+# sdist-build time, which is unreliable since offload-arch (from rocm-sdk-core)
+# is not yet installed at that point. Replace it with explicit per-target entries
+# so callers can request a specific ISA's device wheel unambiguously.
+device_entry = dist_info.ALL_PACKAGES.get("device")
+if device_entry and device_entry.is_target_specific:
+    EXTRAS_REQUIRE.pop("device", None)
+    for _target in dist_info.AVAILABLE_TARGET_FAMILIES:
+        EXTRAS_REQUIRE[f"device-{_target}"] = [
+            device_entry.get_dist_package_require(target_family=_target)
+        ]
```
ScottTodd: Could we switch this from always removing the extra to including it if `offload-arch` is available?
```python
try:
    target = dist_info.determine_target_family()
    ...
except ...:
    EXTRAS_REQUIRE.pop("device", None)
    ...
```

ScottTodd (commented):
Structurally this looks okay. The inlined bash branches are a bit concerning, but once we fully land kpack-split we should be able to simplify substantially, so I'm not sure how much we should prioritize getting that code into scripts.

.github/workflows/build_portable_linux_python_packages.yml
marbre: It should but I you guessed, I was working on Linux first :) Will do the same for Windows of course.

build_tools/packaging/python/templates/rocm/setup.py
marbre: Good point! Implemented that.

.github/workflows/build_portable_linux_pytorch_wheels_ci.yml
marbre: Sure. We talked offline and agreed to address this in a follow up.

.github/workflows/test_rocm_wheels.yml
marbre: We agreed to refactor and clean this up in a follow up.

.github/workflows/multi_arch_ci_linux.yml
marbre: I hope we do neither :/ There are quite some things to simplify with
* non-multi-arch CI abandoned
* always swithting to kpack enabled

Added a TODO note.

ScottTodd (approved):

PR #4489, opened 2026-04-13
[ci] Adding `host-asan` for postsubmit

geomin12:
Adding `host-asan` for push events, while `asan` continues to run on schedule and workflow dispatch events

Changes:
- host-asan variant added
- additional CI logic, where if events match host-asan, we use host-asan variant. otherwise, we use asan variant
- CI asan triggers on push

Ran a test triggered here: https://github.com/ROCm/TheRock/actions/runs/24347848373 (intentionally testing with workflow_dispatch, test commit reverted here: https://github.com/ROCm/TheRock/commit/6e088b77e10b515390c12d3313ff392eb64c85c1)

After this lands and is stable after runs:
- enable host-asan testing
- we will use `host-asan` for presubmit


jayhawk-commits (commented):
`host-asan` still uses the `"asan"` family gate. Since the `build_variants` check happens before `variant_config` is swapped to `host-asan`, a push run can include families that support `asan` but were never opted into `host-asan`. For example, `gfx950` still gets selected here. If `host-asan` is meant to be opt-in per family, this looks wrong.

geomin12:
> `host-asan` still uses the `"asan"` family gate. Since the `build_variants` check happens before `variant_config` is swapped to `host-asan`, a push run can include families that support `asan` but were never opted into `host-asan`. For example, `gfx950` still gets selected here. If `host-asan` is meant to be opt-in per family, this looks wrong.

good point, i think i was trying really hard to not have complex logic in `.yml` files, but just added an update for a really basic conditional

jayhawk-commits (approved):
Thanks for addressing the concern. Looks OK now.

PR #4491, opened 2026-04-13
Build RCCL as single generic Instinct-only artifact in multi-arch CI

marbre:
In multi-arch builds, RCCL was previously built per-arch for every GPU family including consumer RDNA targets (gfx110X, gfx120X, gfx103X, gfx101X). This changes comm-libs to run as a single generic CI job, producing one target-neutral RCCL artifact containing device code for Instinct/CDNA targets only (gfx9xx), consistent with where RCCL is actually tested and supported.

## Changes

Add `USE_FILTERED_AMDGPU_TARGETS REGEX <pattern>` keyword to `therock_cmake_subproject_declare` (cmake/therock_subproject.cmake). This filters `THEROCK_TEST_AMDGPU_TARGETS` by the given regex, enabling a subproject to build for a named subset of all registered targets while still producing a target-neutral (_generic) artifact. Upload races across parallel family jobs are harmless since all jobs would produce identical output.

In comm-libs/CMakeLists.txt, detect generic (multi-arch) builds via `NOT THEROCK_AMDGPU_FAMILIES` (empty in generic stages, set in per-arch stages) and apply `USE_FILTERED_AMDGPU_TARGETS REGEX "^gfx9"` plus `TARGET_NEUTRAL` to both rccl and rccl-tests subproject declarations. Single-family and developer builds are unaffected.

Change BUILD_TOPOLOGY.toml comm-libs build_stage and artifact_group types from per-arch to generic. The artifacts.rccl type remains target-specific, which is what controls kpack split behavior.

Remove the per-arch matrix from the comm-libs CI job in multi_arch_build_portable_linux.yml, replacing it with a single generic job using `amdgpu_family: ""`. rocshmem already uses USE_TEST_AMDGPU_TARGETS and TARGET_NEUTRAL, so it is unaffected by this change.

Update device_artifact_filter in build_python_packages.py to include rccl_lib_generic in device wheels for Instinct families (target.startswith ("gfx9")) while excluding it from RDNA device wheels. Remove rccl from libraries_artifact_filter since it contains device code and must not go into the generic libraries wheel.

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux.yml
+++ b/.github/workflows/multi_arch_build_portable_linux.yml
@@ -121,22 +121,21 @@ jobs:
       id-token: write
 
   # ==========================================================================
-  # STAGE: comm-libs (per-arch, parallel to math-libs)
+  # STAGE: comm-libs (generic, single job parallel to math-libs)
+  # RCCL and rocshmem are both TARGET_NEUTRAL artifacts. RCCL is built for
+  # Instinct/CDNA targets only (^gfx9 regex) via USE_FILTERED_AMDGPU_TARGETS.
+  # Running as a single generic job avoids redundant per-arch builds.
   # ==========================================================================
   comm-libs:
     needs: compiler-runtime
     if: ${{ !cancelled() && !failure() && !contains(inputs.prebuilt_stages, 'comm-libs') }}
-    strategy:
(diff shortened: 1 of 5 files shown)


marbre:
Closing in favor of other alternatives.

PR #4496, opened 2026-04-13
Add a ci_disabled flag for things in the build topology so we can add projects to theRock without submodule bumps

adickin-amd:
## Motivation

We need to be able to add projects to the rock without requiring a submodule bump. Submodule bumps often get delayed for various reasons, and we need a way. to set up our CIs faster. Currently, the hip kernel provider has been waiting for four weeks in order to get in and continuously gets broken on the rocm-libraries side.  This is because we have no CI and the cycle continues.  

This PR will allow us to push project configurations into theRock, test with a submodule bump in the PR, and then revert the submodule change, disable the CI build and then we can merge into TheRock.  From there we can then pull TheRock back to rocm-libraries and enable the CI for the project.  Then once the Ci is up and running and a bump happens, we can come back and enable the CI for the project in the TheRock.

## Submission Checklist


geomin12 (commented):
going to let build folks review, but looks like this worked! yay! hopefully this works in rocm-libraries as intended

adickin-amd:
This was fun but not the right way to do this.  Will use a combination of Feature flags and then a new incoming way of running tests on rocm-libraries.

Issue #4082, opened 2026-03-20
[Feature] pyTorch on TheRock testing request


pruthvistony:
This issue list downs all the testing request of pyTorch on TheRock releases

[PR#3852](https://github.com/ROCm/TheRock/pull/3852) and [PR#3929](https://github.com/ROCm/TheRock/pull/3929) are enabling all config testing of pytorch with Rock builds.
- Default (sGPU runtime - ~12 hrs)
- Distributed (mGPU runtime - ~4 hrs)
- Inductor (sGPU runtime - ~3 hrs)

|pyTorch branch | python version | GPU arch (*add new arch when available) |
|:-------|:-------|:------|
| N  | 3.10, 3.11, 3.12, 3.13   | MI200(gfx90a), MI300(gfx942), MI355(gfx950), gfx110x, gfx115x, gfx120x |
| N-1 | 3.10, 3.11, 3.12, 3.13 | MI200(gfx90a), MI300(gfx942), MI355(gfx950), gfx110x, gfx115x, gfx120x |
| N-2 | 3.10, 3.11, 3.12, 3.13  | MI200(gfx90a), MI300(gfx942), MI355(gfx950), gfx110x, gfx115x, gfx120x |
| upstream/nightly | 3.10, 3.11, 3.12, 3.13   | MI200(gfx90a), MI300(gfx942), MI355(gfx950), gfx110x, gfx115x, gfx120x |

Please note: 
- Frequency of this testing done by QA or automated workflow - **need to be determined**.
- Above pytorch branches needs to be tested across different TheROCk active branches
> https://github.com/ROCm/TheRock/tree/main
> https://github.com/ROCm/TheRock/tree/release/therock-7.11
> https://github.com/ROCm/TheRock/tree/release/therock-7.12

### pyTorch releases
Every 2 months - https://github.com/pytorch/pytorch/blob/main/RELEASE.md#release-cadence.
As soon as release is done, Release branch + TheRock build ready within a week, which will bump the 'N' version is above table.

ethanwee1:
## Address TODOs from PyTorch full unit test PR (#3929)

Tracking issue for follow-up work identified during #3929 (PyTorch full unit test suite in CI).

### Multi-platform / multi-GPU support
- [ ] Expand to additional GPU families (Radeon, gfx950, etc.) and Windows 
https://github.com/ROCm/TheRock/pull/4532/changes

[...]

PR #4499, opened 2026-04-13
[CI] Dispatch PyTorch release branches to run full tests after multi-arch release wheel builds

ethanwee1:
Related: #4082
Related rockrel wrapper: ROCm/rockrel#46

## Summary

- Moves full PyTorch test dispatch to the multi-arch Linux PyTorch wheel release path.
- Runs full tests for scheduled nightly multi-arch Linux PyTorch releases on gfx94X-dcgpu / Python 3.12 against release/2.9, release/2.10, and release/2.11 wheels from the multi-arch release index.
- Runs full tests for PyTorch nightly on Sundays only.
- Keeps full-test failures separate from PyTorch wheel build/upload status by dispatching `test_pytorch_wheels_full.yml` as a downstream workflow.
- Leaves manual/ad-hoc runs opt-in via `run_full_pytorch_tests=false` by default.

## Rollout

- This PR can land before `ROCm/rockrel#46`; existing rockrel wrappers do not pass `run_full_pytorch_tests`, so scheduled releases remain a no-op for full-test dispatch until the companion wrapper is merged.
- `ROCm/rockrel#46` must be merged before enabling scheduled full-test dispatch from rockrel, because it both passes `run_full_pytorch_tests=true` for scheduled releases and adds the required `test_pytorch_wheels_full.yml` wrapper in rockrel.
- Do not pass `run_full_pytorch_tests=true` from any rockrel workflow unless that same repository has the `test_pytorch_wheels_full.yml` wrapper.

## Validation

- Multi-Arch CI passed on the current PR head: https://github.com/ROCm/TheRock/actions/runs/26523375667
- Updated the branch to latest `main` and revalidated the changed workflow YAML locally.
- Checked IDE lints for the touched workflow file.
- A previous direct run of `multi_arch_release_linux_pytorch_wheels.yml` failed before dispatch because it was triggered without the upstream multi-arch ROCm package build output. End-to-end validation should use the top-level multi-arch release path or the rockrel wrapper so `rocm_package_find_links_url` is produced by `build_python_packages`.

The proposed change dispatches full PyTorch tests only from the scheduled nightly multi-arch Linux release path. It runs release/2.9, release/2.10, and release/2.11 nightly on gfx942, and also runs PyTorch nightly once weekly on Sundays.
Per PyTorch branch tested, the full-test workflow creates 11 jobs, and each job takes about 3 hours:

- 6 default shard jobs, each taking 1 linux-gfx942-1gpu-ossci-rocm runner for about 3 hours.
- 3 distributed shard jobs, each taking 1 linux-gfx942-8gpu-ossci-rocm runner for about 3 hours.
- 2 inductor shard jobs, each taking 1 linux-gfx942-1gpu-ossci-rocm runner for about 3 hours.

So each PyTorch branch consumes:

- 8 single-GPU gfx942 runner jobs
- 3 eight-GPU gfx942 runner jobs
- 11 total jobs

Normal nightly load for 3 release branches:

- 24 single-GPU gfx942 runner jobs
- 9 eight-GPU gfx942 runner jobs
- 33 total jobs

Sunday load when PyTorch nightly is included:

- 32 single-GPU gfx942 runner jobs
- 12 eight-GPU gfx942 runner jobs
- 44 total jobs


WBobby:
Concerns with the current approach:
1. Full UT failures will pollute the nightly build's red/green signal (teams using "is the latest nightly green?" as a health check will get noisy readings — especially since upload_pytorch_wheels still publishes the wheel regardless)
2. Dev workflow_dispatch runs of build_portable_linux_pytorch_wheels.yml (gfx94X + py3.12) will be unintentionally slowed down by an appended multi-hour full UT
3. Matrix expansion over pytorch_git_ref amplifies full UT frequency from ~1/day to ~5/day, with limited added signal from older release branches

Would it make sense to trigger full UT as a separate downstream workflow via workflow_run, rather than embedding it inside the build pipeline?

ethanwee1:
> Concerns with the current approach:
> 
> 1. Full UT failures will pollute the nightly build's red/green signal (teams using "is the latest nightly green?" as a health check will get noisy readings — especially since upload_pytorch_wheels still publishes the wheel regardless)
> 2. Dev workflow_dispatch runs of build_portable_linux_pytorch_wheels.yml (gfx94X + py3.12) will be unintentionally slowed down by an appended multi-hour full UT
> 3. Matrix expansion over pytorch_git_ref amplifies full UT frequency from ~1/day to ~5/day, with limited added signal from older release branches
> 
> Would it make sense to trigger full UT as a separate downstream workflow via workflow_run, rather than embedding it inside the build pipeline?

Added workflow_run trigger which fires automatically when Release portable Linux PyTorch Wheels completes (any conclusion: success, failure, etc.)
Added a gate job  that checks github.event.workflow_run.conclusion == 'success' so the full UT suite only runs after a successful release build. If the upstream workflow failed/cancelled, all downstream jobs are skipped.

`release_portable_linux_packages.yml -> release_portable_linux_pytorch_wheels.yml (completes) -> workflow_run triggers ci_nightly_pytorch_full_test.yml.`

.github/workflows/build_portable_linux_pytorch_wheels.yml
```diff
@@ -347,6 +347,29 @@ jobs:
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ inputs.ref || '' }}
 
+  test_pytorch_wheels_full:
```
ScottTodd, line 350: Can you link to a workflow run showing this in action?

.github/workflows/build_portable_linux_pytorch_wheels.yml
```diff
@@ -347,6 +347,29 @@ jobs:
...
+    if: >-
+      ${{
+        inputs.amdgpu_family == 'gfx94X-dcgpu' &&
+        inputs.python_version == '3.12' &&
+        needs.generate_target_to_run.outputs.test_runs_on != ''
+      }}
+    needs: [build_pytorch_wheels, generate_target_to_run, test_pytorch_wheels]
+    permissions:
+      contents: read
+    uses: ./.github/workflows/test_pytorch_wheels_full.yml
```
ScottTodd, line 361: > Would it make sense to trigger full UT as a separate downstream workflow via workflow_run, rather than embedding it inside the build pipeline?

Do you mean workflow_dispatch? Like this: https://github.com/ROCm/TheRock/blob/7b942606d986f47bbd5a8f3b4e8ce41fd7a065e5/.github/workflows/release_portable_linux_packages.yml#L330-L344

.github/workflows/build_portable_linux_pytorch_wheels.yml
ethanwee1: Triggered https://github.com/ROCm/TheRock/actions/runs/24470477684

ScottTodd (commented):
I won't have time to review this. Please find another reviewer.

ethanwee1:
@WBobby Would you mind rereviewing?

.github/workflows/release_portable_linux_pytorch_wheels.yml
ScottTodd, line 1: Sorry for the delayed reviews/responses here. A few things worth being aware of:

* This workflow is for per-family / single-arch releases and is deprecated (scheduled for deletion as soon as multi-arch packages become the only release type, hopefully within 1-2 weeks)
* The replacement is https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
* Scheduled (nightly) runs for multi-arch releases are exclusively triggered over in https://github.com/ROCm/rockrel using a set of wrapper workflows like https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_windows_pytorch_wheels.yml. Any workflow using workflow_dispatch and not workflow_call needs a wrapper like that (see https://github.com/ROCm/rockrel/pull/44 for a recent example)
* We aren't yet running either the quick or the full pytorch tests for multi-arch pytorch releases. I'm working on filling that gap as part of:
  * https://github.com/ROCm/TheRock/pull/5107
  * https://github.com/ROCm/TheRock/issues/5110

ethanwee1: Thanks, moved the implementation off the deprecated per-family release workflows and onto the multi-arch Linux release path.

This now threads an opt-in `run_full_pytorch_tests` input through `multi_arch_release.yml` -> `multi_arch_release_linux.yml` -> `multi_arch_release_linux_pytorch_wheels.yml`, and dispatches `test_pytorch_wheels_full.yml` from the multi-arch PyTorch wheel workflow after the build matrix succeeds. The dispatched tests use the v4 multi-arch staging index with `multi_arch=true`, gfx94X-dcgpu / Python 3.12 / gfx942 runners, nightly release branches daily, and PyTorch nightly on Sundays.

I also reverted the per-family `release_portable_linux_*` workflow changes from the PR diff.

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -1,28 +1,21 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-# Runs the full PyTorch unit-test suite against nightly wheels from v2-staging.
+# Manually dispatchable full PyTorch unit-test suite against wheels from a
+# staging index.
+#
+# Nightly automation is driven by release_portable_linux_pytorch_wheels.yml
+# after wheel builds complete, so scheduled runs here would risk testing stale
+# wheels and duplicating the build-triggered signal.
```
ScottTodd: Stale comment? Switch to `multi_arch_release_linux_pytorch_wheels.yml`?
```suggestion
# Nightly automation is driven by multi_arch_release_linux_pytorch_wheels.yml
# after wheel builds complete, so scheduled runs here would risk testing stale
# wheels and duplicating the build-triggered signal.
```

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -315,3 +323,96 @@ jobs:
...
+          fi
+
+          echo "torch_version=${TORCH_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "Resolved torch_version: ${TORCH_VERSION}"
+
+      - name: Dispatch full PyTorch test workflow
+        if: ${{ steps.cadence.outputs.dispatch == 'true' }}
+        uses: benc-uk/workflow-dispatch@7a027648b88c2413826b6ddd6c76114894dc5ec4 # v1.3.1
+        with:
+          workflow: test_pytorch_wheels_full.yml
```
ScottTodd, line 230: Can you send a PR adding a wrapper `test_pytorch_wheels_full.yml` to https://github.com/ROCm/rockrel/tree/main/.github/workflows , or do you want us to do that? We'll need that for this to be triggered for nightly releases.

* This wrapper workflow gets triggered: https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
* It calls the workflow with the same name from TheRock: https://github.com/ROCm/rockrel/blob/401dfdb4c23dc32a4ecddaed22a924b97985f16a/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml#L47-L49
* That runs _in rockrel_, hence all of the `repository` and `ref` plumbing in other workflows (which we'll need to add to `ci_nightly_pytorch_full_test.yml` too) that say "I know this is running in one repository, but checkout the code from this other repository instead". This would need a PR like https://github.com/ROCm/TheRock/pull/4862
* If any workflow there has `uses: benc-uk/workflow-dispatch`, the `workflow: test_pytorch_wheels_full.yml` code looks for a workflow with that name _in rockrel_, hence the wrapper scripts

It's pretty convoluted, but that gets us isolated security and workflow run history in that repository for releases, while this TheRock repository is more focused on CI.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -315,3 +323,96 @@ jobs:
...
+
+          TORCH_VERSION=$(python -c 'import json, sys; print(json.load(open(sys.argv[1]))["install"][0]["metadata"]["version"])' "${REPORT_PATH}")
+
+          if [ -z "${TORCH_VERSION}" ]; then
+            echo "::error::Failed to resolve ${TORCH_PACKAGE} from ${PACKAGE_INDEX_URL}"
+            exit 1
+          fi
+
+          echo "torch_version=${TORCH_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "Resolved torch_version: ${TORCH_VERSION}"
```
ScottTodd: This "find the version we just built" code is one thing I'm trying to solve as part of https://github.com/ROCm/TheRock/issues/5110. There, we'll generate manifest files with all versions, commit hashes, etc. for each repository/package and then pass that metadata through the workflows.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -315,3 +323,96 @@ jobs:
...
+          - pytorch_git_ref: release/2.11
+            pytorch_git_repo: ROCm/pytorch
+            torch_package: "torch==2.11.*"
+            dispatch_on: daily
+          - pytorch_git_ref: nightly
+            pytorch_git_repo: pytorch/pytorch
+            torch_package: "torch"
+            dispatch_on: sunday
+    env:
+      PACKAGE_INDEX_URL: "https://rocm.nightlies.amd.com/v4/whl-staging"
```
ScottTodd: Hardcoding to `rocm.nightlies.amd.com` is okay since the job has `if ${{ inputs.release_type == 'nightly' }}`, but we may want to support triggering these tests for dev releases as well.

Also note that once https://github.com/ROCm/TheRock/pull/5107 lands this will need to switch to `whl` instead of `whl-staging`

.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -1,28 +1,21 @@
...
-#       workflows (build_portable_linux_pytorch_wheels.yml /
-#       release_portable_linux_packages.yml) so tests only run after a
-#       successful build, or use a workflow_dispatch trigger / Quartz.
 
-name: CI Nightly PyTorch Full Test
+name: CI PyTorch Full Test
 
 on:
-  schedule:
-    - cron: "0 12 * * *" # 12:00 UTC, after nightly builds (~04:00 UTC) have completed
```
ScottTodd: Oh actually, if we drop the `schedule` from here, is this `ci_nightly_pytorch_full_test.yml` serving any purpose?

This workflow with an explicit `pytorch_version` is just a wrapper around `test_pytorch_wheels_full.yml`.

We can make `multi_arch_release_linux_pytorch_wheels.yml` either workflow_dispatch directly to `test_pytorch_wheels_full.yml` (with `workflow`/`ref` plumbing and a wrapper workflow in rockrel per my other comment), OR we can workflow_call (`uses: `) `test_pytorch_wheels_full.yml` directly where the workflows will already run tests (that hasn't been added yet though - I'm working on it)

ScottTodd (commented):
There are a few details here that won't quite work as-is with the cross-repository triggering in https://github.com/ROCm/rockrel/ 🤔 I need at least a few development days (plus time for code reviews) to get the non-full torch tests running as part of multi-arch nightly release workflows, then I think we can slot this in more naturally. Does that work for you?

Otherwise you could switch back to trigger from the non-multi-arch releases for a time.

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: Resolved by removing `.github/workflows/ci_nightly_pytorch_full_test.yml` from this PR. The scheduled/manual wrapper was redundant now that `multi_arch_release_linux_pytorch_wheels.yml` dispatches `test_pytorch_wheels_full.yml` directly.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
ethanwee1: Added the companion rockrel wrapper in ROCm/rockrel#46. I also updated this PR so the dispatch passes `repository`/`ref`, and `test_pytorch_wheels_full.yml` now accepts those inputs for workflow_dispatch. The rockrel scheduled multi-arch wrapper opts into full tests only for scheduled releases, while the PyTorch wheels wrapper passes the flag through.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
ethanwee1: Acknowledged. I left the current pip dry-run resolver as the interim path for this PR and kept it isolated in the dispatch job, so it can be replaced with the manifest output from #5110 once that lands.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
ethanwee1: Kept this scoped to nightly releases for now, so the hardcoded nightlies/staging URL remains intentional. We can switch this from `whl-staging` to `whl` after #5107 lands.

.github/workflows/ci_nightly_pytorch_full_test.yml
ethanwee1: Removed `ci_nightly_pytorch_full_test.yml` instead of keeping it as a second wrapper. The multi-arch PyTorch wheel workflow now dispatches `test_pytorch_wheels_full.yml` directly; rockrel gets the thin wrapper it needs in ROCm/rockrel#46.

ethanwee1:
Addressed the latest review comments in f65b2521 and opened the companion rockrel wrapper PR: ROCm/rockrel#46. This PR now dispatches `test_pytorch_wheels_full.yml` directly with `repository`/`ref` plumbing and removes the redundant `ci_nightly_pytorch_full_test.yml` wrapper. The rockrel PR adds the same-named wrapper workflow and threads `run_full_pytorch_tests` through scheduled multi-arch releases only.

ScottTodd (commented):
Waiting for others to provide review feedback before I take another look.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -311,3 +319,98 @@ jobs:
...
             --release-type="${{ inputs.release_type }}"
+
+  dispatch_pytorch_wheels_full_test:
+    name: Dispatch PyTorch Full Test | gfx94X-dcgpu | torch ${{ matrix.pytorch_git_ref }}
+    if: >-
+      ${{
+        inputs.run_full_pytorch_tests &&
+        inputs.release_type == 'nightly' &&
+        needs.build_pytorch_wheels.result == 'success'
+      }}
```
WBobby, line 159: build_pytorch_wheels is a 5×5 matrix (python_version × pytorch_git_ref, release/2.9 … release/2.12 + nightly) over all amdgpu_families. The full-test job only ever exercises gfx94X-dcgpu / Python 3.12 / release/2.9–2.11 + nightly. With fail-fast: false in the build matrix, a transient failure in any unrelated cell (e.g., Python 3.10 / release/2.12 for a different family) silently kills all full-test dispatches for that night, even though the wheels for the cells we actually test are perfectly fine.

Suggestions:

Use if: always() && ... and check needs.build_pytorch_wheels.result != 'cancelled', then let pip dry-run fail loudly if the specific wheel is missing.

WBobby (changes requested):
Another question: Cross-repo dispatch implicitly depends on rockrel#46 being merged first
When multi_arch_release_linux_pytorch_wheels.yml is invoked via the rockrel wrapper as workflow_call, the job executes inside rockrel. benc-uk/workflow-dispatch (no repo: arg) consequently looks for test_pytorch_wheels_full.yml in rockrel. If this PR lands before ROCm/rockrel#46, every scheduled nightly dispatch will hard-fail with "workflow not found", silently breaking the new signal until the wrapper is added. This should be called out explicitly in the merge ordering (or the dispatch should be gated on something detectable so it no-ops cleanly when the wrapper is absent).

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
ethanwee1: Updated in 09eacd5a. The dispatch job now uses `always()` and only suppresses dispatch when the build matrix is cancelled, so unrelated failed matrix cells no longer prevent the targeted full-test dispatch. The pip dry-run resolver remains the per-branch/per-version check that will fail loudly if the specific wheel we intend to test is missing.

ethanwee1:
Addressed the requested changes in 09eacd5a.

- The full-test dispatch job now uses `always()` and checks `needs.build_pytorch_wheels.result != 'cancelled'`, so unrelated PyTorch wheel matrix failures do not suppress targeted full-test dispatches.
- Added a workflow comment and PR rollout notes explaining the `ROCm/rockrel#46` dependency. Existing rockrel wrappers do not pass `run_full_pytorch_tests`, so this PR can land first as a no-op for scheduled full-test dispatch; `ROCm/rockrel#46` must land before enabling that flag from rockrel.

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -208,6 +212,7 @@ jobs:
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
+              "run_full_pytorch_tests": ${{ inputs.run_full_pytorch_tests }},
```
ScottTodd: Wrap in quotes:
```suggestion
              "run_full_pytorch_tests": "${{ inputs.run_full_pytorch_tests }}",
```

This is causing the unit test failure: https://github.com/ROCm/TheRock/actions/runs/27166371850/job/80194498025?pr=4499
```
==================================== ERRORS ====================================
____ ERROR collecting github_actions/tests/workflow_dispatch_inputs_test.py ____
github_actions/tests/workflow_dispatch_inputs_test.py:213: in <module>
    if not find_dispatch_calls_in_workflow(_workflow):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
github_actions/tests/workflow_dispatch_inputs_test.py:83: in find_dispatch_calls_in_workflow
    passed_inputs=parse_dispatch_inputs_json(
github_actions/tests/workflow_dispatch_inputs_test.py:50: in parse_dispatch_inputs_json
    parsed = json.loads(inputs_raw)
             ^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/json/__init__.py:346: in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/json/decoder.py:338: in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/json/decoder.py:356: in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
E   json.decoder.JSONDecodeError: Expecting value: line 5 column 29 (char 316)
```

I had to printf debug to spot where the issue was 🫠 . We can probably harden the unit tests a bit so that's easier to debug in the future
```
checking D:\projects\TheRock\.github\workflows\multi_arch_ci_linux.yml
checking D:\projects\TheRock\.github\workflows\multi_arch_ci_windows.yml
checking D:\projects\TheRock\.github\workflows\multi_arch_release.yml
checking D:\projects\TheRock\.github\workflows\multi_arch_release_linux.yml
=============================================== short test summary info =============================================== ERROR build_tools\github_actions\tests\workflow_dispatch_inputs_test.py - json.decoder.JSONDecodeError: Expecting value: line 5 column 29 (char 316)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! ================================================== 1 error in 0.69s ===================================================
```

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -135,3 +143,102 @@ jobs:
...
+          - pytorch_git_ref: release/2.11
+            pytorch_git_repo: ROCm/pytorch
+            torch_package: "torch==2.11.*"
+            dispatch_on: daily
+          - pytorch_git_ref: nightly
+            pytorch_git_repo: pytorch/pytorch
+            torch_package: "torch"
+            dispatch_on: sunday
+    env:
+      PACKAGE_INDEX_URL: "https://rocm.nightlies.amd.com/whl-multi-arch/"
```
ScottTodd, line 186: Let's go through and fix up any hardcoded URLs like this after I land https://github.com/ROCm/TheRock/pull/5661

(we need the ability to trigger tests from dev releases not just nightly builds - always start with dev and move forward from there, we should never start directly in "prod" if we can avoid it)

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -135,3 +143,102 @@ jobs:
...
+            --ignore-installed \
+            --no-deps \
+            --only-binary=:all: \
+            --pre \
+            --index-url "${PACKAGE_INDEX_URL}" \
+            --python-version "${PYTHON_VERSION}" \
+            --platform linux_x86_64 \
+            --report "${REPORT_PATH}"
+
+          TORCH_VERSION=$(python -c 'import json, sys; print(json.load(open(sys.argv[1]))["install"][0]["metadata"]["version"])' "${REPORT_PATH}")
```
ScottTodd, line 216: Same down here, cleanup after https://github.com/ROCm/TheRock/pull/5661 lands (then again with https://github.com/ROCm/TheRock/pull/5633). We should know the version _before_ the build, then check the actual version after the build itself. The test jobs should not need to rediscover any of that information. As is, this will install the _latest_ version matching `torch==2.11.*` when test jobs should always test _exactly the version that was just built_. Workflow runs must be reproducible, and this code is not.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -135,3 +143,102 @@ jobs:
...
+
+      - name: Dispatch full PyTorch test workflow
+        if: ${{ steps.cadence.outputs.dispatch == 'true' }}
+        uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
+        with:
+          workflow: test_pytorch_wheels_full.yml
+          inputs: |
+            { "amdgpu_family": "gfx94X-dcgpu",
+              "test_runs_on": "linux-gfx942-1gpu-ossci-rocm",
+              "test_runs_on_multi_gpu": "linux-gfx942-8gpu-ossci-rocm",
```
ScottTodd, line 234: Follow-up: we should have the ability to run "full" tests for any supported GPU and platform, not just Linux gfx942. Probably need to move more of this workflow code into a Python script that generates the matrix, similar to what @geomin12 started in https://github.com/ROCm/TheRock/pull/5326

ScottTodd (approved):
Ok to get started (once the unit tests are fixed). A few things to follow-up on but fine to land this PR ahead of my other PRs that adjust these pytorch workflows.

WBobby (approved):
### Approving as an interim
One change requested before merge, plus a required follow-up.
**Before merge — revert the dispatch gate to `result == 'success'`:**
```yaml
if: >-
  ${{
    inputs.run_full_pytorch_tests &&
    inputs.release_type == 'nightly' &&
    needs.build_pytorch_wheels.result == 'success'
  }}
```
Rationale: the current `always() && result != 'cancelled'` lets the dispatch run even when a build leg failed. The resolve step then re-resolves the version off the live index and picks up *yesterday's* wheel, reporting a misleading green on tonight's run. This is a conscious tradeoff: I prefer **underreport** (no signal — operators know to investigate) over **misreport** (false green — operators misled). Yes, this re-introduces the aggregate-skip behavior (a failing leg skips the whole dispatch); that is accepted here because a trustworthy-but-sparse signal beats a frequent-but-deceptive one. Note `== 'success'` reduces but does not fully eliminate stale-wheel risk (index propagation lag remains).

**Required follow-up PR — decide per-ref whether the build succeeded and test the version actually built this run.**

Root cause: the dispatch job only sees the *aggregate* `needs.build_pytorch_wheels.result`; it cannot tell which refs (2.9/2.10/2.11/nightly) individually passed. Both `== 'success'` (underreport) and `always()` (misreport) are stuck on this. The fix must obtain per-ref status.


### Merge Order — TheRock#4499 + rockrel#46
Cross-repo change. Wrong order breaks the whole nightly release (`Unexpected input(s) 'run_full_pytorch_tests'`).

PR #4508, opened 2026-04-13
TheRock with amd-main LLVM

kasaurov:
## Motivation

$${\color{red}\text{EXPERIMENTAL}}$$: Switch LLVM to `amd-main` from `amd-mainline`

## Technical Details

Changes:
- Submodule pointer & `.gitmodules`: Point `compiler/amd-llvm` to the tip of `amd-main` (b7232468f267) and update the tracked branch.
- GPU runtimes config (`compiler/pre_hook_amd-llvm.cmake`):
--- Remove `flang-rt` from `amdgcn` cross-compilation runtimes (its `CMakeLists.txt` requires `Threads/Backtrace`, unavailable for GPU targets).
--- Fix `compiler-rt` cache file path: `GPU.cmake` was renamed to `AMDGPU.cmake` on `amd-main`.
--- Add install-time symlinks for `amdclang`, `amdclang++`, `amdflang` (wrappers no longer shipped on `amd-main`; needed by downstream components like`rocprofiler-sdk`).
- `CLR` linker fix (`core/CMakeLists.txt`): Pass `--undefined-version` to the linker for `hip-clr` and `ocl-clr`. The `amd-main` `ld.lld` defaults to `--no-undefined-version`, which rejects version-script symbols defined in separate libraries (e.g. `hiprtc`).
- `ROCM_PATH` override (`cmake/therock_subproject.cmake`): Set `ROCM_PATH` in the generated toolchain to ensure TheRock's build tree is found before any system-installed ROCm.
- `hip-tests` patches (`patches/amd-mainline/rocm-systems/`):
--- Guard `SPIR-V` kernel targets behind `ENABLE_SPIRV` (the `spirv64-amd-amdhsa` target hits missing `dpp` feature errors with the newer compiler).
--- Remove obsolete `-fallow-half-arguments-and-returns` flag (removed in newer Clang; `_Float16` is natively supported).
- Drop inapplicable Comgr patch: `0011-Comgr-Fix-memory-leak-...` is already fixed upstream on `amd-main`.

## Submission Checklist


kasaurov:
Closing the PR as there was a decision to drop and deprecate `amd-main` branch.

PR #4520, opened 2026-04-14
[debug-tools] Updated llvm executable/library locations for rocgdb test script

lumachad:
LLVM has switched locations in the ROCm tree from llvm/[bin|lib] to lib/llvm/[bin|lib]. Make sure we use both for now.

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_rocgdb.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rocgdb.py
@@ -805,17 +805,27 @@ def setup_environment(artifacts_dir: Path) -> Dict[str, str]:
     env_vars = os.environ.copy()
 
     # Add ROCgdb and LLVM binaries to PATH.
-    env_vars["PATH"] = f"{artifacts_dir}/bin:{artifacts_dir}/llvm/bin:" + env_vars.get(
-        "PATH", ""
+    #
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/test_executable_scripts/test_rocgdb.py
```diff
@@ -805,17 +805,27 @@ def setup_environment(artifacts_dir: Path) -> Dict[str, str]:
     env_vars = os.environ.copy()
 
     # Add ROCgdb and LLVM binaries to PATH.
-    env_vars["PATH"] = f"{artifacts_dir}/bin:{artifacts_dir}/llvm/bin:" + env_vars.get(
-        "PATH", ""
+    #
```
geomin12, line 808: nit: empty comment line can be removed

geomin12 (approved):

build_tools/github_actions/test_executable_scripts/test_rocgdb.py
lumachad: It was actually a separation so the second block of text doesn't stick together with the first comment. I'm fine either way.

lumachad: Going with the separation.

PR #4531, opened 2026-04-14
add platform input to derive pip platform tag for Windows support

ethanwee1:
Related: #4082

## Summary
- Adds a `platform` workflow input to the full PyTorch UT dispatch path.
- Derives the pip platform tag from that input so Linux uses `linux_x86_64` and Windows uses `win_amd64` during torch version resolution.
- Keeps the existing Linux default while preparing the resolver for Windows wheel testing.

## Validation
- Dispatched CI Nightly PyTorch Full Test on this branch for gfx94X-dcgpu / Python 3.12 / linux / default+distributed+inductor: https://github.com/ROCm/TheRock/actions/runs/25391945103


.github/workflows/ci_nightly_pytorch_full_test.yml
```diff
@@ -61,6 +61,13 @@ on:
         description: "Python version"
         type: string
         default: "3.12"
+      platform:
+        description: "Target platform: linux or windows"
+        type: choice
+        options:
+          - linux
+          - windows
+        default: "linux"
```
ScottTodd, line 70: Normally I'd say explicit is better than implicit, but there's a risk here of this getting out of sync with `test_runs_on`. How about using this `contains(inputs.test_runs_on, 'linux')` code pattern instead?

https://github.com/ROCm/TheRock/blob/1a434283f4e94b97fa70223582d87787228c59fe/.github/workflows/test_pytorch_wheels.yml#L89

Code below that is using the provided runner uses the more reliable `runner.os`: https://github.com/ROCm/TheRock/blob/1a434283f4e94b97fa70223582d87787228c59fe/.github/workflows/test_pytorch_wheels_full.yml#L338

HereThereBeDragons (commented):
i agree with scott. we should not have an extra input parameter.
this can be extracted from the runner name 

in general if i look at the "resolve_torch_version" job: i would prefer if we do this in a python script. i guess you are trying to get the latest torch version we have on the nightly repo?
i dont think this will work as expected as it shouldnt be deterministic which pytorch version you are getting. pytorch builds are started in parallel per nightly we are building 3? versions. so you can get either one of them.

if i were you i would rather hard-code the version and add the part we talked about with having fixed days to test different pytorch versions. in that case you also dont need the platform part.

ethanwee1:
Closing as superseded: ci_nightly_pytorch_full_test.yml was refactored into test_pytorch_wheels_full.yml, which already uses the `contains(inputs.test_runs_on, 'linux')` pattern and no longer derives the pip platform tag in bash. The platform handling this PR added is no longer applicable.

PR #4539, opened 2026-04-14
[ci] Enable `nightly_check_only_for_family` and `run-full-tests-only` for multi arch CI

### LINKED ISSUE #4184: [ci] Multi-arch CI does not honor `run-full-tests-only` flag, while CI does
Currently, multi-arch CI does not support the `run-full-tests-only` option as multi arch runs a different config https://github.com/ROCm/TheRock/blob/59d986221af9310c8543e7476155c64dbe7971e1/build_tools/github_actions/configure_ci.py#L467-L477 than https://github.com/ROCm/TheRock/blob/59d986221af9310c8543e7476155c64dbe7971e1/build_tools/github_actions/configure_ci.py#L710-L712 (uses `matrix_generator`). 

As multi arch migration is occuring, I will hold off this fix for multi arch CI after migration happens (to prevent more delays)

geomin12:
Closes #4184 , covering gaps from `ci.yml` to `multi_arch_ci.yml`. 

Right now, `multi_arch_ci.yml` does not cover `nightly_check_only_for_family` and `run-full-tests-only`, but this PR updates that and runs the architectures during the proper scheduled run or full-test run

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -843,9 +843,6 @@ def _expand_build_config_for_platform(
                     f"disabling tests"
                 )
 
-        # Apply per-family test filtering flags
-        sanity_check_only = platform_info.get("sanity_check_only_for_family", False)
-
         # If run-full-tests-only is set and test_type is "quick", disable testing
         if platform_info.get("run-full-tests-only", False) and test_type == "quick":
             test_runs_on = ""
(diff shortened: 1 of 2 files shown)


jayhawk-commits (commented):
`nightly_check_only_for_family` looks misaligned with previous behaviour. The new `continue` drops the family from `per_family_info` on non-scheduled runs, so we stop building it entirely for PR/push/workflow_dispatch. In the old `configure_ci.py` logic, this flag only downgraded non-nightly runs to `sanity_check_only_for_family`; it didn’t remove the family from the build matrix. Was the intent to skip tests only, rather than skip the whole build?

geomin12:
> `nightly_check_only_for_family` looks misaligned with previous behaviour. The new `continue` drops the family from `per_family_info` on non-scheduled runs, so we stop building it entirely for PR/push/workflow_dispatch. In the old `configure_ci.py` logic, this flag only downgraded non-nightly runs to `sanity_check_only_for_family`; it didn’t remove the family from the build matrix. Was the intent to skip tests only, rather than skip the whole build?

ah yes, just skip tests. updated

jayhawk-commits (approved):
Looks OK to me, if this is the intended skip test behaviour.

PR #4547, opened 2026-04-14
Port kpack-split wiring to Windows workflows

marbre:
Port kpack-split Python packages wiring added with commit 0258e5e9 to the Windows CI equivalents.

Legacy per-family builds are unaffected — all new inputs default to empty/false.

### DIFF
--- a/.github/workflows/build_windows_pytorch_wheels_ci.yml
+++ b/.github/workflows/build_windows_pytorch_wheels_ci.yml
@@ -46,7 +46,7 @@ on:
       amdgpu_targets:
         description: >-
           Comma-separated individual gfx targets for kpack-split builds
-          (e.g. 'gfx942' or 'gfx1200,gfx1201'). Used to install the correct
+          (e.g. 'gfx1151' or 'gfx1200,gfx1201'). Used to install the correct
           per-target device wheels. Leave empty for legacy per-family builds.
         type: string
         default: ""
--- a/.github/workflows/build_windows_pytorch_wheels_ci.yml
+++ b/.github/workflows/build_windows_pytorch_wheels_ci.yml
@@ -192,7 +192,7 @@ jobs:
         id: device_extras
         run: |
           if [ "${{ inputs.kpack_split }}" = "true" ] && [ -n "${{ inputs.amdgpu_targets }}" ]; then
-            # Prefix each comma-separated target with "device-": gfx942,gfx1201 -> device-gfx942,device-gfx1201
+            # Prefix each comma-separated target with "device-": gfx1151,gfx1201 -> device-gfx1151,device-gfx1201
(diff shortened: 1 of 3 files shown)


.github/workflows/build_windows_pytorch_wheels_ci.yml
```diff
@@ -43,6 +43,17 @@ on:
         required: false
         type: string
         default: "none"
+      amdgpu_targets:
+        description: >-
+          Comma-separated individual gfx targets for kpack-split builds
+          (e.g. 'gfx942' or 'gfx1200,gfx1201'). Used to install the correct
+          per-target device wheels. Leave empty for legacy per-family builds.
```
ScottTodd, line 50: Suggest gfx targets that are supported on Windows instead
```suggestion
      amdgpu_targets:
        description: >-
          Comma-separated individual gfx targets for kpack-split builds
          (e.g. 'gfx1151' or 'gfx1200,gfx1201'). Used to install the correct
          per-target device wheels. Leave empty for legacy per-family builds.
```

.github/workflows/build_windows_pytorch_wheels_ci.yml
```diff
@@ -177,6 +188,17 @@ jobs:
...
+      - name: Compute ROCm device extras
+        id: device_extras
+        run: |
+          if [ "${{ inputs.kpack_split }}" = "true" ] && [ -n "${{ inputs.amdgpu_targets }}" ]; then
+            # Prefix each comma-separated target with "device-": gfx942,gfx1201 -> device-gfx942,device-gfx1201
+            extras=$(echo "${{ inputs.amdgpu_targets }}" | sed 's/\(^\|,\)/\1device-/g')
+            echo "value=${extras}" >> "$GITHUB_OUTPUT"
+          else
+            echo "value=" >> "$GITHUB_OUTPUT"
+          fi
```
ScottTodd, line 200: same here
```suggestion
      - name: Compute ROCm device extras
        id: device_extras
        run: |
          if [ "${{ inputs.kpack_split }}" = "true" ] && [ -n "${{ inputs.amdgpu_targets }}" ]; then
            # Prefix each comma-separated target with "device-": gfx1151,gfx1201 -> device-gfx1151,device-gfx1201
            extras=$(echo "${{ inputs.amdgpu_targets }}" | sed 's/\(^\|,\)/\1device-/g')
            echo "value=${extras}" >> "$GITHUB_OUTPUT"
          else
            echo "value=" >> "$GITHUB_OUTPUT"
          fi
```

ScottTodd (approved):

.github/workflows/build_windows_pytorch_wheels_ci.yml
marbre: Good catch.

marbre:
Test failures look infra related.

PR #4548, opened 2026-04-14
Re-enable ASAN build for fusilli and iree

AaronStGeorge:
## Motivation

This PR re-enables the `iree-compiler` build stage and disables the ASAN build for `libIREECompiler.so`. This mirrors AMD LLVM's approach; since IREE vendors its own LLVM copy, the same policy likely applies. IREE runs its own ASAN testing in `iree-org` CI.

## Technical Details

- In ASAN builds `libIREECompiler.so` built in release mode and without sanitizers 

## Test Plan

- IREE should build successfully in multi-arch ASAN build

## Test Result

- [result](https://github.com/ROCm/TheRock/actions/runs/24425589370/job/71361947162)

## Submission Checklist


iree-libs/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ if(THEROCK_ENABLE_IREE_COMPILER)
     NO_MERGE_COMPILE_COMMANDS
     BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/iree-compiler"
     EXTERNAL_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/iree"
+    COMPILER_TOOLCHAIN
```
MaheshRavishankar, line 17: Why do we need the amd-llvm?

AaronStGeorge: IIUC TheRock will use the system compiler otherwise, we likely don't need this but it's probably better to use a newer clang vs older GCC.

MaheshRavishankar: If it becomes an issue maybe, but it creates a dependency between IREEs LLVM build and AMD-LLVM which makes the two LLVM builds sequential instead of allowing them to be (possibly) built in parallel.

AaronStGeorge: It's already dependent on the `hip-clr` to provide the hip headers which is produced in the same [build stage as the compiler](https://github.com/ROCm/TheRock/blob/dba06c75024cf85ae478c11db9c375060f097d7e/BUILD_TOPOLOGY.toml#L137-L148). So it should already need to be sequential.

https://github.com/ROCm/TheRock/blob/dba06c75024cf85ae478c11db9c375060f097d7e/iree-libs/CMakeLists.txt#L36-L41

In terms of latency hiding, currently math libs provides a pretty big shadow for IREE to hide behind 😆.

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -236,16 +224,9 @@ jobs:
...
   fusilli-libs:
     needs: [compiler-runtime, math-libs, iree-compiler]
-    if: >-
-      ${{
-        !cancelled() &&
-        !failure() &&
-        !contains(inputs.prebuilt_stages, 'fusilli-libs') &&
-        inputs.build_variant_label != 'asan'
-      }}
+    if: ${{ !cancelled() && !failure() && !contains(inputs.prebuilt_stages, 'fusilli-libs') }}
```
geomin12: should we keep fusilli-libs? seems like this will only work for iree-compiler (and changes are only applied to iree-compiler)

geomin12 (approved):

geomin12 (changes requested):
oops i didn't mean to approve

.github/workflows/multi_arch_build_portable_linux.yml
AaronStGeorge: I'm happy to leave it off for now and turn it on after some validation either through CI or local testing 👍  

It probably already works. The stage contains fusilli and fusilli provider builds -- fusilli [builds with ASAN in its CI](https://github.com/iree-org/fusilli/actions/runs/24464411962/job/71487350324) and fusilli provider is a standard (and relatively small) C++ project. We can't test in CI ATM as `fusilli-libs` stage depends on `math-libs` [which breaks under ASAN](https://github.com/ROCm/TheRock/actions/runs/24434068103).

geomin12 (approved):

MaheshRavishankar (approved):

PR #4563, opened 2026-04-15
Revert "RFC0010: fetch test scripts from super-repos in CI"

jayhawk-commits:
- Reverts ROCm/TheRock#4045
- checkout times added should have been a blocker


ScottTodd (approved):
Thanks

PR #4570, opened 2026-04-15
[ci] Adding `determine_rocm_test_dependencies.py` script

### LINKED ISSUE #3491: [TheRock CI] upstream component testing run and failures blocks even though unrelated
tests are run for upstream components which can not be affected by downstream:
https://github.com/ROCm/rocm-libraries/pull/456

Failures blocks downstream work for no valid reason.   
Productivity and energy waste is enormous.

geomin12:
For `rocm-libraries`, some components test are run even though upstreams changes to components do not affect downstream components. this results in wasted CI resources and wasted developer time.

Closes #3491

This is not used in TheRock but will be implemented in `rocm-libraries` and will continue to expand as we determine component specific updates.

Originally, we were going to use BUILD_TOPOLOGY, but it would test more than is needed (not granular enough for specific components). With these tests given from component teams, we can target which tests have best value


jayhawk-commits:
I think these new scripts are good candidates to be placed not in `build_tools`, but in the new directory we talked about: `test_tools`.

geomin12:
> I think these new scripts are good candidates to be placed not in `build_tools`, but in the new directory we talked about: `test_tools`.

good call! done :)

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -0,0 +1,143 @@
...
+    """Parse CMakeLists.txt files to extract TEST_SUBPROJECTS declarations.
+
+    Returns dict mapping subproject name (lowercase) -> list of test dependencies (lowercase).
+    """
+    test_deps = {}
+    cmake_files = list(Path(therock_dir).rglob("CMakeLists.txt"))
+
+    for cmake_file in cmake_files:
+        try:
+            content = cmake_file.read_text()
```
jayhawk-commits: If there is an issue with this `try` block, we should not silently ignore the issue. Fail fast on unexpected errors and report.

jayhawk-commits (changes requested):

jayhawk-commits (approved):
Issues I identified previously were addressed. LGTM

TorreZuk:
Thanks @geomin12 for making progress on this.

cmake/therock_subproject.cmake
```diff
@@ -513,6 +513,8 @@ function(therock_cmake_subproject_declare target_name)
     THEROCK_BUILD_DEPS "${ARG_BUILD_DEPS}"
     # Transitive runtime deps.
     THEROCK_RUNTIME_DEPS "${_transitive_runtime_deps}"
+    # Optional override for test dependencies (subproject names to test when this changes).
+    THEROCK_TEST_SUBPROJECTS "${ARG_TEST_SUBPROJECTS}"
```
ScottTodd, line 517: Guys, changes to the build system really need review from build system maintainers.

This approach will not work, see my comments on
* https://github.com/ROCm/TheRock/pull/6646#issuecomment-5008299363
* https://github.com/ROCm/TheRock/pull/6703#discussion_r3617155055

test_tools/determine_rocm_test_dependencies.py
ScottTodd, line 1: * This added a new top level directory, which we want to manage carefully to avoid too much sprawl
* These unit tests are not run on CI without changes to https://github.com/ROCm/TheRock/blob/main/.github/workflows/unit_tests.yml
* The unit tests are slow (10+ seconds for some test cases)

Issue #4441, opened 2026-04-09
[Multi-arch] Build and upload ROCm tarballs as part of CI/CD pipelines


ScottTodd:
New release workflows using multi-arch should upload tarballs to e.g. https://rocm.nightlies.amd.com/tarball/.

Existing release pipelines build `therock-dist` then run `tar cfz` and upload:
* https://github.com/ROCm/TheRock/blob/5250110dbb4e9fbe3e2062f6145734785ff0ea83/.github/workflows/release_portable_linux_packages.yml#L179
* https://github.com/ROCm/TheRock/blob/5250110dbb4e9fbe3e2062f6145734785ff0ea83/.github/workflows/release_portable_linux_packages.yml#L240-L246
* https://github.com/ROCm/TheRock/blob/5250110dbb4e9fbe3e2062f6145734785ff0ea83/.github/workflows/release_portable_linux_packages.yml#L307-L310

I'm planning on adding dedicated scripts and workflows that will repackage artifacts into tarballs, similar to what we do for python packages. Details:
* Without kpack split, produce one tarball per GPU family
* With kpack split, produce one tarball with ALL GPU families per platform (and optionally also include the per-GPU tarballs)
* Later we may want to produce tarballs that include test artifacts and tarballs that don't

We can either include this step just for releases or also include for CI runs (at potentially 10GB+ of extra storage used per run, due to how inefficiency these files are compressed and how they overlap). Building JAX packages on CI will currently need tarballs, as JAX packages do not build against the ROCm Python packages.

ScottTodd:
Status update: 
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_tarballs.yml workflow file has been merged
* It can be triggered manually via workflow_dispatch here: https://github.com/ROCm/TheRock/actions/workflows/multi_arch_build_tarballs.yml, it is not yet connected to any existing workflows
* https://github.com/ROCm/TheRock/pull/4509 will start to use it from new release workflows

Here are two test runs using artifacts from https://github.com/ROCm/TheRock/actions/runs/24364435658 (multi-arch CI, without KPACK split):
1. Linux: https://github.com/ROCm/TheRock/actions/runs/24410162469/job/71304398547 --> https://therock-ci-artifacts.s3.amazonaws.com/24410162469-linux/tarballs/index.html
[...]

PR #4577, opened 2026-04-15
Add publish_to_release_buckets job to multi_arch_release_linux.yml

ScottTodd:
## Motivation

Progress on:
* https://github.com/ROCm/TheRock/issues/4441

## Technical Details

The copy/promote job is local to release workflows, so:
* CI workflows don't risk also affecting release buckets
* All release files are published in one step, instead of having files upload partially as workflow jobs complete
  * Note that this will also prevent publishing a partial release, but the files will be available in the artifacts bucket if we want to manually salvage a run

## Test Plan

* Unit tests
* Tested together with other changes at https://github.com/ROCm/TheRock/actions/runs/24429137657

## Test Result

Test run uploaded files to `therock-dev-artifacts` then copied to `therock-dev-tarball`:

* Build tarballs job: https://github.com/ROCm/TheRock/actions/runs/24429137657/job/71377961777
  * Index for tarballs in `therock-dev-artifacts`: https://therock-dev-artifacts.s3.amazonaws.com/24429137657-linux/tarballs/index.html
* Publish to release buckets job: https://github.com/ROCm/TheRock/actions/runs/24429137657/job/71378733418
  * Published file in `therock-dev-tarball`: https://therock-dev-tarball.s3.us-east-2.amazonaws.com/v3/tarball/therock-dist-linux-gfx1151-7.13.0.dev0%2B492711f25782d286fd9bad6e26bf83909d651e75.tar.gz

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/publish_rocm_to_release_buckets.py
+++ b/build_tools/github_actions/publish_rocm_to_release_buckets.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Publish ROCm release files from an artifacts bucket to release buckets.
+
+These release file types are supported:
+
+- [x] tarballs
+- [ ] python packages
+- [ ] native linux packages
+- [ ] native windows packages
+
+Example with ``--run-id 12345 --platform linux --release-type dev``:
+
+    s3://therock-dev-artifacts/12345-linux/tarballs/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
+      -> s3://therock-dev-tarball/v3/tarball/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
+
+Test usage:
+    python build_tools/github_actions/publish_rocm_to_release_buckets.py \\
+        --run-id 12345 --platform linux --release-type dev --dry-run
+"""
+
+import argparse
+import logging
+import platform as platform_module
+import sys
+from pathlib import Path
+
+_BUILD_TOOLS_DIR = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(_BUILD_TOOLS_DIR))
+
+from _therock_utils.s3_buckets import get_release_bucket_config
+from _therock_utils.storage_backend import StorageBackend, create_storage_backend
+from _therock_utils.storage_location import StorageLocation
+from _therock_utils.workflow_outputs import WorkflowOutputRoot
+
+logger = logging.getLogger(__name__)
+
+
+def publish_tarballs(
+    artifacts_root: WorkflowOutputRoot,
+    release_type: str,
+    backend: StorageBackend,
+) -> int:
+    """Copy tarballs from the artifacts bucket to the release tarball bucket.
+
+    Example:
+        s3://therock-dev-artifacts/12345-linux/tarballs/
+          -> s3://therock-dev-tarball/v3/tarball/
+
+    Returns:
+        Number of tarballs copied.
+    """
+    source = artifacts_root.tarballs()
+    dest_bucket = get_release_bucket_config(release_type, "tarball")
+    dest = StorageLocation(dest_bucket.name, "v3/tarball")
+
+    logger.info("Tarballs: %s -> %s", source.s3_uri, dest.s3_uri)
+    count = backend.copy_directory(source, dest, include=["*.tar.gz"])
+    logger.info("Copied %d tarballs", count)
+    if count == 0:
+        raise FileNotFoundError(f"No tarballs found at {source.s3_uri}")
+
+
+def main(argv: list[str]) -> None:
+    parser = argparse.ArgumentParser(
+        description="Publish ROCm release files to release buckets"
+    )
+    parser.add_argument("--run-id", required=True, help="Source workflow run ID")
+    parser.add_argument(
+        "--platform",
+        default=platform_module.system().lower(),
+        choices=["linux", "windows"],
+        help="Platform (default: current system)",
+    )
+    parser.add_argument(
+        "--release-type",
+        required=True,
+        choices=["dev", "nightly", "prerelease"],
+        help="Release type (determines source and destination buckets)",
+    )
+    parser.add_argument(
+        "--dry-run", action="store_true", help="Print plan without copying"
+    )
+    args = parser.parse_args(argv)
+
+    artifacts_root = WorkflowOutputRoot.from_workflow_run(
+        run_id=args.run_id, platform=args.platform, release_type=args.release_type
+    )
+    backend = create_storage_backend(dry_run=args.dry_run)
+
+    publish_tarballs(artifacts_root, args.release_type, backend)
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+    main(sys.argv[1:])
--- a/build_tools/github_actions/publish_rocm_to_release_buckets.py
+++ b/build_tools/github_actions/publish_rocm_to_release_buckets.py
@@ -14,7 +14,7 @@ These release file types are supported:
 Example with ``--run-id 12345 --platform linux --release-type dev``:
 
     s3://therock-dev-artifacts/12345-linux/tarballs/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
-      -> s3://therock-dev-tarball/v3/tarball/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
+      -> s3://therock-dev-tarball/v4/tarball/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
 
 Test usage:
     python build_tools/github_actions/publish_rocm_to_release_buckets.py \\
@@ -47,14 +47,14 @@ def publish_tarballs(
 
     Example:
         s3://therock-dev-artifacts/12345-linux/tarballs/
-          -> s3://therock-dev-tarball/v3/tarball/
+          -> s3://therock-dev-tarball/v4/tarball/
 
     Returns:
         Number of tarballs copied.
     """
     source = artifacts_root.tarballs()
     dest_bucket = get_release_bucket_config(release_type, "tarball")
-    dest = StorageLocation(dest_bucket.name, "v3/tarball")
+    dest = StorageLocation(dest_bucket.name, "v4/tarball")
 
     logger.info("Tarballs: %s -> %s", source.s3_uri, dest.s3_uri)
     count = backend.copy_directory(source, dest, include=["*.tar.gz"])
--- a/build_tools/github_actions/upload_tarballs.py
+++ b/build_tools/github_actions/upload_tarballs.py
@@ -29,6 +29,7 @@ Usage:
 """
 
 import argparse
+import logging
 import platform as platform_module
 import sys
 from pathlib import Path
@@ -39,13 +40,10 @@ sys.path.insert(0, str(_BUILD_TOOLS_DIR))
 from _therock_utils.storage_backend import create_storage_backend
 from _therock_utils.workflow_outputs import WorkflowOutputRoot
 
-
-def log(*args):
-    print(*args)
-    sys.stdout.flush()
+logger = logging.getLogger(__name__)
 
 
-def main():
+def main(argv: list[str]) -> int:
     parser = argparse.ArgumentParser(description="Upload tarballs to S3")
     parser.add_argument(
         "--input-tarballs-dir",
@@ -53,11 +51,12 @@ def main():
         required=True,
         help="Directory containing .tar.gz tarballs to upload",
     )
+    parser.add_argument("--run-id", required=True, help="Workflow run ID")
     parser.add_argument(
-        "--run-id",
-        type=str,
-        required=True,
-        help="Workflow run ID",
+        "--platform",
+        default=platform_module.system().lower(),
+        choices=["linux", "windows"],
+        help="Platform (default: current system)",
     )
     parser.add_argument(
         "--release-type",
@@ -68,21 +67,12 @@ def main():
         "--output-dir",
         type=Path,
         default=None,
-        help="Output to local directory instead of S3",
+        help="Output to local directory instead of S3 (for testing)",
     )
     parser.add_argument(
-        "--platform",
-        type=str,
-        default=platform_module.system().lower(),
-        choices=["linux", "windows"],
-        help="Platform for the upload path (default: current system)",
+        "--dry-run", action="store_true", help="Print plan without uploading"
     )
-    parser.add_argument(
-        "--dry-run",
-        action="store_true",
-        help="Print what would happen without uploading",
-    )
-    args = parser.parse_args()
+    args = parser.parse_args(argv)
 
     tarballs_dir = args.input_tarballs_dir.resolve()
     if not tarballs_dir.is_dir():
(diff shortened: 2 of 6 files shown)


build_tools/github_actions/upload_tarballs.py
ScottTodd, line 1: The style changes in this file were to make it and the new `build_tools/github_actions/publish_rocm_to_release_buckets.py` look more similar. This script uploads local directory -> S3, that script copies files S3 -> S3.

I don't care too much about the style/logging choices one way or the other, I mostly want it to be easy to compare them.

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -0,0 +1,98 @@
...
+    Example:
+        s3://therock-dev-artifacts/12345-linux/tarballs/
+          -> s3://therock-dev-tarball/v3/tarball/
+
+    Returns:
+        Number of tarballs copied.
+    """
+    source = artifacts_root.tarballs()
+    dest_bucket = get_release_bucket_config(release_type, "tarball")
+    dest = StorageLocation(dest_bucket.name, "v3/tarball")
```
ScottTodd: We may want a different destination subfolder depending on the release bucket and release file type. See `s3_prefix` here for example: https://github.com/ROCm/TheRock/blob/e93933ce4ad6c755a516c86dcbfbb884054f7f6c/build_tools/packaging/linux/get_s3_config.py#L95-L154

ScottTodd: Could also consider `v3/` instead of `v3/tarball/`. IDK. @marbre opinions?

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -0,0 +1,98 @@
...
+
+def publish_tarballs(
+    artifacts_root: WorkflowOutputRoot,
+    release_type: str,
+    backend: StorageBackend,
+) -> int:
+    """Copy tarballs from the artifacts bucket to the release tarball bucket.
+
+    Example:
+        s3://therock-dev-artifacts/12345-linux/tarballs/
```
nunnikri, line 49: Isnt the source bucket therock-ci-artifacts?

build_tools/github_actions/publish_rocm_to_release_buckets.py
nunnikri: Yes instead of hardcoding it here, its might be better to add the prefix/location also to s3_buckets.py

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -0,0 +1,98 @@
...
+    dest = StorageLocation(dest_bucket.name, "v3/tarball")
+
+    logger.info("Tarballs: %s -> %s", source.s3_uri, dest.s3_uri)
+    count = backend.copy_directory(source, dest, include=["*.tar.gz"])
+    logger.info("Copied %d tarballs", count)
+    if count == 0:
+        raise FileNotFoundError(f"No tarballs found at {source.s3_uri}")
+
+
+def main(argv: list[str]) -> None:
```
nunnikri, line 66: might need the package_type(tarball, wheel, packages) also as a input. Based on package_type, release_type and platform make the destination address.

build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd: No, see https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#bucket-inventory. CI workflows upload to `therock-ci-artifacts[-external]`. Release workflows upload to `therock-{release_type}-artifacts`, then copy/promote to `therock-{release_type}-[packages/python/tarball]`

build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd: My current plan with the stack (just added https://github.com/ROCm/TheRock/pull/4625) is to do all uploads here in one step, after waiting for all rocm release files to be ready:
```diff
  publish_to_release_buckets:
-   needs: [build_tarballs]
+   needs: [build_tarballs, build_python_packages]
```
```diff
    publish_tarballs(artifacts_root, args.release_type, backend)
+   publish_python_packages(artifacts_root, args.release_type, backend, kpack_split)
```

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: What we do (when adding origins in CF) is to omit the `v3`. In consequence, the CF will make the file available at https://rocm.*.amd.com/tarballs. If you leave `/tarball` out you won't get it in the URL. Only way around would be a URI rewrite but I think I'd like to avoid this for now (might change with all the CF redesign that is desired). However, if your comment is rather about making this configurable that might be okay, but pushing to the "root" isn't desired?

build_tools/tests/s3_buckets_test.py
```diff
@@ -28,6 +29,10 @@ def test_ci_rocm_therock(self):
             release_type="", repository="ROCm/TheRock", is_pr_from_fork=False
         )
         self.assertEqual(config.name, "therock-ci-artifacts")
+        self.assertEqual(
+            config.write_access_iam_role,
+            "arn:aws:iam::692859939525:role/therock-ci",
```
marbre, line 34: Do we want to hardcode this here in tests?

marbre (approved):
LGTM if testing passes.

build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd: I don't want to push to the root. Some of these files may overlap with files produced by the existing release pipelines and I want an easy way to delete the whole subdirectory and pick a new one if something goes wrong.

I think `v3/` on its own is sufficient, no need for `v3/tarball` when already in `therock-dev-tarball`. For python packages I'm planning on `v3/` and `v3-multi-arch/`: https://github.com/ROCm/TheRock/pull/4625

build_tools/tests/s3_buckets_test.py
ScottTodd: I wanted at least one test for `write_access_iam_role`, and directly checking the expected role is more reliable than doing string manipulation in the test: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

build_tools/github_actions/publish_rocm_to_release_buckets.py
nunnikri: With this flow artifacts buckets have build artifacts, tarballs, wheel and packages, though different buckets are already available.  
Ideal flow should be. 
Release build workflow -> upload build artifacts to therock-{release_type}-artifacts -> Down stream jobs(tarball/wheel/package) fetch artifacts from therock-{release_type}-artifacts -> Upload the tarball/wheel/packages to therock-{release_type}-[packages/python/tarball (Can be first to a staging area and after testing copy to final location)

ScottTodd: > With this flow artifacts buckets have build artifacts, tarballs, wheel and packages, though different buckets are already available.

Yeah we're duplicating files quite a bit here, but I think the benefits are worth it. We still need to roll out retention policies for files in these buckets though, before the ongoing costs grow even larger.

> Release build workflow -> upload build artifacts to therock-{release_type}-artifacts -> Down stream jobs(tarball/wheel/package) fetch artifacts from therock-{release_type}-artifacts -> Upload the tarball/wheel/packages to therock-{release_type}-[packages/python/tarball

This code copies from S3 to S3, no need to download then reupload.

> (Can be first to a staging area and after testing copy to final location)

As for testing, tests can either use files from therock-{release_type}-artifacts and treat that as staging, or we can upload to a dedicated staging directory in the release bucket (e.g. -python). The PyTorch release workflows already do that with v2-staging --> v2. The advantage with a staging directory in a release bucket is that we can have a central index with _untested_ packages that users/developers can install from if they want. There is no central index in the artifacts buckets (there we have self-contained subdirectories, one per workflow run).

For the ROCm packages (tarballs, python packages, native linux/windows packages), I'm leaning towards no staging directory and not waiting for tests to complete before promoting -- at least for tests that depend on GPUs to run. Installation tests and sanity checks would be fine, but I don't want to block releases for multiple hours while we wait for runners to be available. Sometimes runners go offline, in which case then we'd have a stalled release that would never complete.

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: Please note that the `tarball` subdirectory is mainly used to show up in the CloudFront distribution! https://rocm.devreleases.amd.com/tarball/

In general we will restrict access to any non-artifact bucket and you will only be able to access via CF. Thus the subdir matters here. Otherwise we will need an URI rewrite which is something we can do but it will trigger a CF function invocation for _every_ download.

ScottTodd: Ok, so you do want `v3/tarball/`? But just `v3/` and `v3-multi-arch/` for python packages? I'm not really following, especially with the different release buckets and cloudfront mappings :/

build_tools/tests/s3_buckets_test.py
marbre: Okay, so my question should have rather been, so we really want a write test here because that will fail locally if not running on CI.

ScottTodd: All of these tests pass locally and on CI. The tests don't actually assume the role or attempt to upload.

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns:
+        Number of tarballs copied.
+    """
+    source = artifacts_root.tarballs()
+    dest_bucket = get_release_bucket_config(release_type, "tarball")
+    dest = StorageLocation(dest_bucket.name, "v3/tarball")
+
+    logger.info("Tarballs: %s -> %s", source.s3_uri, dest.s3_uri)
+    count = backend.copy_directory(source, dest, include=["*.tar.gz"])
+    logger.info("Copied %d tarballs", count)
```
ScottTodd, line 61: Test run with all archs failed: https://github.com/ROCm/TheRock/actions/runs/24540020624/job/71767976731

Copy code will need to tolerate larger file sizes somehow:
```
INFO:_therock_utils.storage_backend:  therock-dist-windows-gfx908-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-gfx908-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz
INFO:_therock_utils.storage_backend:  therock-dist-windows-gfx90a-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-gfx90a-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz
INFO:_therock_utils.storage_backend:  therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz
WARNING:_therock_utils.storage_backend:S3 copy attempt 1/3 failed for s3://therock-dev-artifacts/24540020624-windows/tarballs/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz: An error occurred (InvalidRequest) when calling the CopyObject operation: The specified copy source is larger than the maximum allowable size for a copy source: 5368709120. Retrying in 1.0s...
WARNING:_therock_utils.storage_backend:S3 copy attempt 2/3 failed for s3://therock-dev-artifacts/24540020624-windows/tarballs/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz: An error occurred (InvalidRequest) when calling the CopyObject operation: The specified copy source is larger than the maximum allowable size for a copy source: 5368709120. Retrying in 2.0s...
  File "/home/runner/work/TheRock/TheRock/build_tools/github_actions/publish_rocm_to_release_buckets.py", line 98, in <module>
    main(sys.argv[1:])
  File "/home/runner/work/TheRock/TheRock/build_tools/github_actions/publish_rocm_to_release_buckets.py", line 93, in main
    publish_tarballs(artifacts_root, args.release_type, backend)
  File "/home/runner/work/TheRock/TheRock/build_tools/github_actions/publish_rocm_to_release_buckets.py", line 60, in publish_tarballs
    count = backend.copy_directory(source, dest, include=["*.tar.gz"])
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/TheRock/TheRock/build_tools/_therock_utils/storage_backend.py", line 178, in copy_directory
    return self.copy_files(pairs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/TheRock/TheRock/build_tools/_therock_utils/storage_backend.py", line 391, in copy_files
    raise RuntimeError(
RuntimeError: Failed to copy 1/13 files. First failure: s3://therock-dev-tarball/v3/tarball/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz: S3 copy failed after 3 attempts: s3://therock-dev-artifacts/24540020624-windows/tarballs/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz -> s3://therock-dev-tarball/v3/tarball/therock-dist-windows-multiarch-7.13.0.dev0+94696237b7f5a5c44ec2d3e367dae21483ec4e87.tar.gz
```

Linux index page: https://therock-dev-artifacts.s3.amazonaws.com/24540020624-linux/tarballs/index.html. 9GB for the multi-arch tarball, 2-3GB for others

Windows index page: https://therock-dev-artifacts.s3.amazonaws.com/24540020624-windows/tarballs/index.html. 5GB for the multi-arch tarball, 2-3GB for others

build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd: Ok I think I understand now.
* This code with `v3/tarball/` is fine
* We can make `v3/tarball/` available as `/tarball-multi-arch` in CloudFront - it'll strip the `/v3` and add expand to `tarball-multi-arch`?
* After the retention policy expires all existing root files we can remap from `/tarball-multi-arch` to just `/tarball`

ScottTodd: We discussed some more and settled on `v4/tarball/` for now, see https://github.com/ROCm/TheRock/pull/4625#issuecomment-4270327982.

The hope is to use a single subfolder structure across dev, nightly, and prerelease buckets. The release bucket will always be a bit special (rocm prefix and such).

Issue #3323, opened 2026-02-09
[Multi-arch] Ship Multi-Architecture Packaging with Kpack


ScottTodd:
Parent issue tracking shipping ROCm multi-architecture packaging with Kpack based on this RFC: https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0008-Multi-Arch-Packaging.md.

## Overview

ROCm currently ships "fat binaries" that embed device kernels for all supported GPU architectures within each binary. This approach has several limitations:

1. **Build scalability**: Each distribution family requires a complete build pipeline, limiting parallelism
2. **Binary bloat**: Users must download and store kernels for architectures they don't use
3. **Architecture support pressure**: Lesser-used architectures face deprecation due to build time and size costs
4. **Compression inefficiency**: No cross-kernel deduplication or advanced compression

This issue tracks implementing the work outlined in the RFC which will improve each of those areas.

## Major milestones

* Integrating rocm-kpack (https://github.com/ROCm/rocm-kpack) into TheRock to enable architecture-separated packaging.
  * Kpack provides a structured format for separating device code from host code at packaging time, enabling parallel builds, selective deployment, and future compression improvements.
* Adding continuous integration (CI) coverage for multi-arch builds, including all existing project unit tests
* Build and distribute multi-arch ROCm Python packages
* Build and distribute multi-arch PyTorch Python packages
* Build and distribute multi-arch ROCm native Linux packages

## Implementation plan and timelines

TBD

kcossett-amd:
Based on the description, I'm assuming that https://github.com/ROCm/TheRock/issues/4848 is expected behavior then?
That is, I should no longer rely on `llvm-objdump --offloading` for final `target-specific` Kpack-split `dist/rocm` binaries and if so, are there any alternatives?

nunnikri:
Multi arch packages are released as part of ROCm7.13 release

ScottTodd:
This issue stays open until all subissues are closed (notably we need to delete the old code to complete the migration).

ScottTodd:
I think enough subissues have been closed that we can close this too, unless we want to keep it open + pinned for longer?

The main thing I was waiting on was deleting the per-family release workflows, which was done in https://github.com/ROCm/TheRock/pull/5790.

marbre:
I think this is fine to close as of today.

PR #4580, opened 2026-04-15
ensure other zen5 chipset architectures are correctly set

paulstuart:
## Motivation

When a cpu architecture is guaranteed, in the case of the Strix Halo chipset (gpu `gfx1151`), the binaries should be compiled to take full advantage of the cpu capabilities.

## Technical Details

This extends the build configuration script to identify and inject the compiler build architecture setting.

## Test Plan

This adds a test file to verify the build architecture is chosen correctly

## Test Result

The 10 tests in the configure test script all pass

## Submission Checklist


build_tools/github_actions/build_configure.py
```diff
@@ -84,6 +84,41 @@ def build_compiler_launcher(
     ],
 }
 
+# Map from individual GPU target tokens to required host CPU march flags.
+#
+# Only list targets here when ALL hardware that ships with that GPU also uses
+# a known, fixed CPU microarchitecture (i.e. unified/integrated chipsets where
+# the GPU and CPU die are the same product line).
```
ScottTodd: Which subprojects specifically benefit from compiling with these flags?

With multi-arch CI/CD (https://github.com/ROCm/TheRock/issues/3323), there is a new hard requirement that all host code (binaries, shared libraries) must be bit-identical no matter which amdgpu targets a subproject is built for, all target-specific files must be distributed as additive sidecar files. We can't set different compilation options depending on GPU target.

build_tools/github_actions/build_configure.py
```diff
@@ -111,6 +146,16 @@ def build_configure(build_dir, manylinux=False):
...
 
+    # Inject host CPU march flags for targets tied to a specific microarchitecture.
+    # MSVC does not accept -march, so skip on Windows.
+    if PLATFORM != "windows":
+        host_march = host_march_for_families(amdgpu_families)
+        if host_march:
+            march_flags = f"-march={host_march} -mtune={host_march}"
+            cmd.append(f"-DCMAKE_C_FLAGS={march_flags}")
+            cmd.append(f"-DCMAKE_CXX_FLAGS={march_flags}")
+            logging.info(f"Injecting host CPU flags: {march_flags}")
```
ScottTodd: This `build_configure.py` is only used for old/deprecated CI pipelines (it's replaced by https://github.com/ROCm/TheRock/blob/main/build_tools/configure_stage.py)

Any build system code like this would need to go closer to files used by the build system directly like these:
* https://github.com/ROCm/TheRock/blob/main/cmake/therock_compiler_config.cmake
* https://github.com/ROCm/TheRock/blob/main/cmake/therock_subproject.cmake
* https://github.com/ROCm/TheRock/blob/main/compiler/pre_hook_amd-llvm.cmake
* https://github.com/ROCm/TheRock/blob/main/CMakePresets.json

build_tools/github_actions/test_build_configure.py
```diff
@@ -0,0 +1,86 @@
...
+    for t in tests:
+        try:
+            t()
+            print(f"  PASS  {t.__name__}")
+            passed += 1
+        except AssertionError as e:
+            print(f"  FAIL  {t.__name__}: {e}")
+            failed += 1
+    print(f"\n{passed} passed, {failed} failed")
+    sys.exit(1 if failed else 0)
```
ScottTodd: We run tests with pytest. No need for custom test runner code like this.

Tests also belong in a test directory like https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions/tests and must be named `*_test.py` not `test_*.py` (see https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#pre-commit-checks, which enforces this)

build_tools/github_actions/build_configure.py
```diff
@@ -111,6 +146,16 @@ def build_configure(build_dir, manylinux=False):
         ]
     )
 
+    # Inject host CPU march flags for targets tied to a specific microarchitecture.
+    # MSVC does not accept -march, so skip on Windows.
+    if PLATFORM != "windows":
+        host_march = host_march_for_families(amdgpu_families)
```
ScottTodd: We build most subprojects using clang from amd-llvm even on Windows, so `-march` should be possible to pass there too. We can't just set top level `CMAKE_C_FLAGS` across all compilers outside of a toolchain like this.

ScottTodd (changes requested):
This approach does not look compatible with multi-arch packaging. Need a different design, based on which subprojects specifically benefit from flags like this, and motivated by benchmarks / data.

build_tools/github_actions/build_configure.py
ScottTodd: (see https://github.com/ROCm/TheRock/blob/main/docs/development/build_system.md, in particular the points about `_toolchain.cmake` generated by the super-project)

paulstuart:
@ScottTodd thanks for taking the time to review and reply.

I haven't done the benchmarks but I hope that my assumption is not unreasonable: that binaries that will only run on targeted hosts can be compiled to make full use of cpu capabilities.

The other was that it's nice to have packages that "just work" rather than manually copying around.

If these goals align with those of the project, I'd be happy to revisit and structure them as per your guidance.
If not, I'll just do source builds and call it a day.

build_tools/github_actions/build_configure.py
paulstuart: @ScottTodd ok, ignoring an earlier reply it looks like the cpu optimization is a no-go. I can understand the reasoning but it does seem unfortunate in a package that is intended for performance. I'll focus on source builds. But os packaging would still be nice.

ScottTodd: Yeah it's tricky. We're shipping one set of binaries/libraries per platform: Linux (RHEL/manylinux), Windows. We don't have a good way to specialize per CPU right now, at least not with fine-grained flags like this. I'd want to understand more about which subprojects / binaries / libraries specifically benefit from CPU-specific compilation flags. We could split those libraries up and ship multiple versions for different CPUs, or something.

Issue #3968, opened 2026-03-13
Extract shared test utilities and eliminate boilerplate from test executable scripts


jayhawk-commits:
**Depends on:** [RFC 0010 — Migrate Test Executable Scripts to Owning Project Repositories](https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0010-Test-Scripts-Migration.md) — this work begins once scripts have landed in their owning repos.

### Summary

[RFC 0010](https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0010-Test-Scripts-Migration.md) migrates the ~35 test runner scripts from `build_tools/github_actions/test_executable_scripts/` into `rocm-systems` and `rocm-libraries`. That migration moves the scripts to the right owners, but it does not fix the underlying structural problems they carry with them:

- Each script duplicates the same 20–30 lines of setup boilerplate.
- The shared helpers scripts depend on live in `build_tools/github_actions/github_actions_utils.py` — a module whose name and location signal "GitHub Actions tooling", not "test utilities". That mismatch will be obvious once the scripts live outside TheRock.
- The scripts cannot be run locally by developers without knowing which environment variables CI sets.

Some of these concerns are also touched on in the [RFC 0010 discussion](https://github.com/ROCm/TheRock/discussions/3964).

---

### Problem 1 — Duplicated boilerplate will move with the scripts

The RFC 0010 migration seeds the owning repos from the current scripts as-is. Every migrated script will arrive carrying the same copy-pasted setup block:

```python
THEROCK_BIN_DIR = os.getenv("THEROCK_BIN_DIR")
logging.basicConfig(level=logging.INFO)

SHARD_INDEX = os.getenv("SHARD_INDEX", 1)
TOTAL_SHARDS = os.getenv("TOTAL_SHARDS", 1)
environ_vars = os.environ.copy()
environ_vars["GTEST_SHARD_INDEX"] = str(int(SHARD_INDEX) - 1)
environ_vars["GTEST_TOTAL_SHARDS"] = str(TOTAL_SHARDS)
```

The `ROCM_PATH` derivation (`Path(THEROCK_BIN_DIR).resolve().parent`) and the `++ Exec [...]$` log line are similarly repeated across ~28 scripts each. Migration moves the duplication to a new address; it does not reduce it.

[...]

PR #4581, opened 2026-04-15
RFC0010: package initial core test scripts in artifacts

jayhawk-commits:
Part of RFC0010 test script migration.

This PR moves an initial core-project slice of TheRock test execution toward
artifact-packaged test scripts instead of relying on repository-local script
execution.

## What changed

- Added subproject support to define packaged test script inputs and install
  destinations:
  - `INSTALL_TEST_SCRIPT_FILES`
  - `INSTALL_TEST_SCRIPT_DESTINATION`
- Installed packaged test scripts under `share/therock-scripts`
- Updated test configuration fetching so packaged scripts resolve from the
  installed artifact tree under `./build/share/therock-scripts`
- Packaged the initial core test scripts used here:
  - `test_hiptests.py`
  - `test_rocrtst.py`
- Packaged `github_actions_api.py` alongside `test_hiptests.py` as a temporary
  compatibility shim for the current `rocm-systems` copy of that script

## Scope

- This PR is intentionally limited to the first core-project slice.
- `test_sanity.py` is deferred to a follow-up because TheRock has not yet picked
  up the required `rocm-systems` content for that packaging path.
- This PR does not yet package scripts for other project groups such as
  math-libs, ml-libs, media-libs, profiler, or debug-tools.

## Follow-ups

- Remove the temporary `hip-tests` dependency on `github_actions_api.py` in
  `rocm-systems`. Today this is only used for a one-time, two-line `is_asan`
  helper path and should be folded into the script so the extra packaged helper
  is no longer needed.
- Package `test_sanity.py` once the required submodule update lands.
- Package scripts for additional project groups in separate PRs.
- Continue RFC0010 migration away from repository-local test script execution.

RFC:
- https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0010-Test-Scripts-Migration.md


jayhawk-commits:
The copy of the sanity test script to rocm-systems was recent. The current SHA of the submodule on TheRock is not recent enough.

jayhawk-commits:
The failing tests are not affected by the changes in the PR.

core/artifact-core-hiptests.toml
```diff
@@ -6,4 +6,6 @@
 [components.test."core/hip-tests/stage"]
 include = [
   "share/hip/**",
+  "share/therock-scripts/github_actions_api.py",
```
agunashe, line 9: does every script need this as well ? cant we somehow have a common package or path for this script as well ?
Ideally the test script should not depend on build scripts. the test script should only depend on test package.
I see hip-tests depends on is_asan flag from github_actions_api , why cant we pass this flag /path as a arg to the script, so we dont have dependencies

jayhawk-commits: Note that one of my follow-ups is to remove this dependency. It is a two-line function that is called once. I don't think this justifies inclusion of a second script.

If there grows a bigger need, then we can revisit the design.

agunashe: okay thanks. Makes sense.

jharryma (approved):
LGTM!

- [install logs](https://therock-ci-artifacts.s3.amazonaws.com/24598871356-linux/logs/compiler-runtime/rocrtst_install.log) verify that the test script is being packaged 
- verified that the [rocrtst_test_generic.tar.zst artifact](https://therock-ci-artifacts.s3.amazonaws.com/24598871356-linux/rocrtst_test_generic.tar.zst) contains the script as well

I see the packaged script is being used to run the test as well: https://github.com/ROCm/TheRock/actions/runs/24598871356/job/72165531710?pr=4581#step:12:3

verified the above for hip-tests aswell.

geomin12 (approved):

jayhawk-commits:
Thanks for the first pass of reviews. Looking for a second tier of reviews for subproject cmake changes before I merge it in.

cmake/therock_subproject.cmake
ScottTodd, line 1: Why are any of the changes to this file needed?

How about having the subprojects install their test scripts (if any) to some directory by convention (e.g. `share/tests/`), then have the CI system look there?

* Any code mentions of "therock" in directory paths, file names, CMake variables, environment variables, etc. are a code smell. This is one build system stacked on top of the underlying subprojects. The subprojects shouldn't need to know about the superproject that wraps them.
* Building the subprojects on their own should install test scripts that work on their own. As written now, the test scripts will only be installed when built using TheRock's build system, not the standalone subproject CMake builds.
* If we later add benchmarks, "extended tests", test dependencies, or other files, it may be tempting to extend `INSTALL_TEST_SCRIPT_FILES` with additional files in spite of those limitations.

I've given similar feedback about the scripts in https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions/test_executable_scripts and https://github.com/ROCm/rocm-libraries/tree/develop/test/therock before. Most of the smells in those files are `THEROCK_DIR` and `THEROCK_BIN_DIR` environment variables, but that directory isn't meaningful, it's just the standard rocm installation structure (produced by flattening TheRock's artifacts, installing a tarball, using a legacy release from `/opt/rocm/`, etc.)

jayhawk-commits: Thanks, I’m going to rework this approach.

I’ve started with ROCm/rocm-systems#5552 for rocrtst. The goal there is to move the runner into the owning subproject, install it from the standalone rocr-runtime build, remove the `test/therock` copy, and make the script use the normal ROCm install layout instead of TheRock-specific paths or environment variables.

While touching the script, I’m also cleaning up issues that came from the original TheRock-side script, and generally aligning it with TheRock’s Python style guide.

Once that pattern is settled and TheRock picks up the updated rocm-systems submodule, I’ll update this PR to consume the installed script instead of packaging scripts centrally through `therock_subproject.cmake`.

ScottTodd: Perfect, thanks! I realize that iterating takes more time, but I think we'll thank ourselves for this later :)

jayhawk-commits:
Closing this as superseded/outdated.

The RFC0010 direction has shifted since this PR was opened. The review discussion here helped clarify that TheRock should not be the long-term owner of installing project-specific test runner scripts through a superproject hook. The direction we are converging on is for the owning projects to install their own runner scripts as normal artifacts, with TheRock consuming those installed runners from fetched artifacts.

I’m tracking the broader migration in ROCm/TheRock#5090, and the shared test utilities / boilerplate cleanup in ROCm/TheRock#3968. Future work should continue through those smaller project-owned runner changes plus the shared utilities work, rather than updating this older `INSTALL_TEST_SCRIPT_FILES` approach.

Thanks for the review feedback here. I’m closing this to avoid keeping an outdated design path open.

PR #4582, opened 2026-04-15
Add multi_arch_release_windows.yml

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/3334, adding Windows multi-arch release workflow code alongside the recently added Linux code.

## Technical Details

The previous configure code would expand empty `linux_amdgpu_families` and `windows_amdgpu_families` to `all_families.keys()`. This introduces explicit `all` and `none` strings and sets the defaults to `gfx94x` (Linux) and `gfx110x` (Windows), matching 
https://github.com/ROCm/TheRock/blob/05aaf931e52461b23d4b3c32c4fdcaba627cbdc3/.github/workflows/release_portable_linux_packages.yml#L53-L56 https://github.com/ROCm/TheRock/blob/05aaf931e52461b23d4b3c32c4fdcaba627cbdc3/.github/workflows/release_windows_packages.yml#L56-L59

## Test Plan

Test dev release at https://github.com/ROCm/TheRock/actions/runs/24478811673

* Job name `Multi-Arch Release (dev) | Linux: none | Windows: gfx110X,gfx1151`
* Configured for `gfx110X-all, gfx1151`
* Jobs should complete, upload artifacts, logs, and tarballs to therock-dev-artifacts, then promote tarballs to therock-dev-tarball

## Submission Checklist


.github/workflows/multi_arch_release.yml
```diff
@@ -27,11 +27,11 @@ on:
         type: string
         default: ""
       linux_amdgpu_families:
-        description: "Comma-separated list of Linux GPU families. Empty = all families."
+        description: 'Comma-separated list of Linux GPU families. "all" = all, "none" or empty = skip.'
```
marbre, line 30: I think if someone types "all, gfx94x", "all" becomes an unknown family name and is filtered downstream, while gfx94x quietly wins?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -680,19 +680,19 @@ def select_targets(ci_inputs: CIInputs) -> TargetSelection:
...
         # Manual trigger: caller specifies exact families per platform.
-        # For CI dispatches, empty input means "no families for that
-        # platform" — the caller has full control over what runs.
-        # For release dispatches, empty input defaults to all families
-        # so that release workflows don't need to enumerate every family.
+        # "all" = all known families. "none" or empty = skip platform.
         linux_names = list(ci_inputs.linux_amdgpu_families)
         windows_names = list(ci_inputs.windows_amdgpu_families)
-        if ci_inputs.release_type and not linux_names and not windows_names:
+        if linux_names == ["all"]:
```
marbre, line 686: Maybe raise an ValueError when all/none appears in multi-entry lists?

.github/workflows/multi_arch_release.yml
```diff
@@ -49,21 +49,21 @@ on:
         type: string
         default: ""
       linux_amdgpu_families:
-        description: "Comma-separated list of Linux GPU families. Empty = all families."
+        description: 'Comma-separated list of Linux GPU families. "all" = all, "none" = skip platform.'
```
marbre, line 52: Keep in sync with the above or drop?

```suggestion
        description: 'Comma-separated list of Linux GPU families. "all" = all, "none" or empty = skip.'
```

ScottTodd: Ah I'll edit the descriptions. This is a bit tricky. Behavior is different between workflow_dispatch and workflow_call.

* workflow_dispatch has `default`, so you can't actually input empty string (github converts that back to the default)
* workflow_call will be used by scheduled jobs for nightly releases, and empty --> default/all. We could make that explicit with code in rockrel that checks event schedule --> "all"

.github/workflows/multi_arch_release.yml
ScottTodd: Edge case but yeah, could do more string validation. Can't quite check for "all" in the input --> only "all" must be there, as someone could try "gfx110X-all,gfx1151".

.github/workflows/multi_arch_release.yml
marbre: Ohhh, I see..

ScottTodd: Added some comments. Hope this is easier to understand now. Note that
* https://github.com/ROCm/TheRock/pull/4619 will also make some changes here, passing `linux_amdgpu_families` and `windows_amdgpu_families` through to `setup_multi_arch.yml` explicitly
* https://github.com/ROCm/rockrel/pull/35 will call this workflow with `linux_amdgpu_families: ${{ inputs.linux_amdgpu_families || 'all' }}` (we can iterate there on defaults for dev/nightly/prerelease)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Can we defer that for later? I think we'll still want to iterate on the defaults for all/none/specific targets and I don't want to fill the configure scripts and tests with too much code yet.

marbre (approved):
LGTM

Issue #4596, opened 2026-04-16
Bump dvc[s3] and pin aiobotocore in build_manylinux_x86_64.Dockerfile


marbre:
## Background

`dockerfiles/build_manylinux_x86_64.Dockerfile` currently installs `dvc[s3]==3.62.0` with a loose upper bound on `aiobotocore`. As a result, image rebuilds can resolve different `aiobotocore` versions, which in turn changes the `botocore` window and silently breaks the `boto3` pin in `requirements.txt`.

This already happened once: image `sha256:d6ae5712...` resolved `aiobotocore 3.1.2` (botocore `<1.42.43`), while a later rebuild `sha256:702a5133...` resolved `aiobotocore 3.4.0` (botocore `>=1.42.79,<1.42.85`). The `boto3>=1.41.0,<1.42.0` pin in `requirements.txt` was compatible with the first but conflicted with the second, causing `pip install -r requirements.txt` to downgrade `botocore` and trigger a
resolver error against `aiobotocore 3.4.0`.

The immediate fix will realign the `boto3` pin in `requirements.txt` to `>=1.42.79,<1.42.85` to match the currently-published image. That fix is reactive - the underlying drift risk remains.

## Proposed work

1. **Bump `dvc[s3]`** from `3.62.0` to the latest stable (3.67.1 at time of writing).
2. **Drop the `pathspec<0.13.0` pin** if the new dvc no longer uses pathspec's private `_DIR_MARK` API. The Dockerfile comment notes this should be possible past dvc ~3.64.0.
3. **Add an explicit `aiobotocore` pin** to the `dvc[s3]` install line so the resolved `botocore` window is deterministic across rebuilds.
4. **Realign the `boto3` pin in `requirements.txt`** to whatever `botocore` window the new `aiobotocore` requires.
5. **Smoke-test `dvc pull`** in `rocm-libraries/` (and `rocm-systems/` on Windows) - this is the only `dvc` command the build invokes (`build_tools/fetch_sources.py:438`).

PR #4572, opened 2026-04-15
Bump manylinux SHA

marbre:
Bumps the manylinux SHA to pick up the latest version including the patched patchelf.

PR #4597, opened 2026-04-16
Re-land bump manylinux SHA (#4572)

marbre:
Relands the manylinux SHA bump which was reverted with commit d90478f4 while adjusting the boto3 version range in `requirements.txt`.

aiobotocore should be pinned in the Dockerimage in a follow up, see issue #4596.


geomin12 (approved):
lgtm if passing and all looks good

marbre:
Ci failures are unrelated, good to land.

PR #4601, opened 2026-04-16
Document patchelf requirements

marbre:
To enable kpack split, the manylinux image is switched to a version building patchelf from source pinned to an upstream version. This version includs a fix to addresses the PHDR-relocation bug that silently corrupts kpack-split libraries.

This adds guidance for developers building outside the container to not let them run into hitting the dlopen failures without a clear pointer to the cause.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,15 +68,10 @@ sudo apt install gfortran git ninja-build cmake g++ pkg-config xxd automake libt
 git clone https://github.com/ROCm/TheRock.git
 cd TheRock
 
-# Install a patched patchelf from source. Ubuntu's packaged patchelf
-# silently corrupts kpack-split libraries; see the environment setup
-# guide for details.
+# Install a patched patchelf from source. For details see
 # https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#patchelf
 sudo apt install curl make
-PATCHELF_GIT_REF=$(sed -n 's/^ENV PATCHELF_GIT_REF="\(.*\)"/\1/p' \
-  dockerfiles/build_manylinux_x86_64.Dockerfile)
-sudo env INSTALL_PREFIX=/usr/local \
-  ./dockerfiles/install_patchelf.sh "${PATCHELF_GIT_REF}"
+sudo env INSTALL_PREFIX=/usr/local ./dockerfiles/install_pinned_patchelf.sh
 
 # Init python virtual environment and install python dependencies
 python3 -m venv .venv && source .venv/bin/activate
--- a/README.md
+++ b/README.md
@@ -62,11 +62,21 @@ instructions and configurations for alternatives.
 ```bash
 # Install Ubuntu dependencies
 sudo apt update
-sudo apt install gfortran git ninja-build cmake g++ pkg-config xxd patchelf automake libtool python3-venv python3-dev libegl1-mesa-dev texinfo bison flex
+sudo apt install gfortran git ninja-build cmake g++ pkg-config xxd automake libtool python3-venv python3-dev libegl1-mesa-dev texinfo bison flex
 
 # Clone the repository
 git clone https://github.com/ROCm/TheRock.git
 cd TheRock
+
+# Install a patched patchelf from source. Ubuntu's packaged patchelf
+# silently corrupts kpack-split libraries; see the environment setup
+# guide for details.
+# https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#patchelf
+sudo apt install curl make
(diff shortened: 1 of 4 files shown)


docs/packaging/python_packaging.md
```diff
@@ -89,6 +89,14 @@ patching via patchelf. On Linux, it is recommended to run this in the same
 portable container as was used to build the SDK (so as to avoid the possibility
 of accidentally referencing too-new glibc symbols).
 
+If running on a host rather than the portable container, make sure the host
+`patchelf` carries the PR #544 PHDR-relocation fix. A stock distro `patchelf`
```
ScottTodd: What is 544 here? Please include the full link `https://github.com/NixOS/patchelf/pull/544`

docs/environment_setup_guide.md
```diff
@@ -30,6 +31,80 @@ There are various, easy ways to acquire specific CMake versions. For Windows and
...
+
+If you see any of these after a local wheel build or `BUNDLE_SYSDEPS`
+build, suspect your host `patchelf`.
+
+#### Fix
+
+The fix is [NixOS/patchelf PR #544](https://github.com/NixOS/patchelf/pull/544)
+("Allocate PHT/SHT at the end of the ELF file"), merged 2025-01-07 to
+master but not yet in a tagged release. Any supported build path needs a
+`patchelf` that includes this commit.
```
ScottTodd: A few points here:

1. This is pretty verbose text to include at the top of this setup guide. We could move "Common Issues" to the bottom of the page,a fter "Reference Build Environments" and "Alternative Configurations"
2. Headers like "Failure mode" and "Fix" aren't descriptive enough. These will appear in the URL when you anchor to them (e.g. https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#resource-utilization). I'd suggest qualifying somehow, e.g. "Common issues with patchelf" and "Compatible versions of patchelf"

README.md
```diff
@@ -62,12 +62,22 @@ instructions and configurations for alternatives.
...
 
+# Install a patched patchelf from source. Ubuntu's packaged patchelf
+# silently corrupts kpack-split libraries; see the environment setup
+# guide for details.
+# https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#patchelf
+sudo apt install curl autoconf make
+PATCHELF_GIT_REF=$(sed -n 's/^ENV PATCHELF_GIT_REF="\(.*\)"/\1/p' \
+  dockerfiles/build_manylinux_x86_64.Dockerfile)
+sudo env INSTALL_PREFIX=/usr/local \
+  ./dockerfiles/install_patchelf.sh "${PATCHELF_GIT_REF}"
```
ScottTodd: I'd really rather not add this much code to the most visible "how do I get started" instructions we have

Image

The "Frequent setup and building problems" section in https://github.com/ROCm/TheRock?tab=readme-ov-file#building-from-source that links to https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#common-issues may be sufficient here (we could rephrase or drop a few hints there, but this `sed -n 's/ ...` code to run an install script is way too much :P

marbre: If we don't grep the version / SHA with some mechanism like this we need to keep it in sync in multiple places. If you don't install any patchelf and already the configure run will fail (we check if patchelf is installed) and there is a high likehood this is missed.

docs/packaging/python_packaging.md
marbre: Done.

docs/environment_setup_guide.md
marbre: Move the section to the bottom, renamed the sections to "Issue with patchelf" and "Compatible patchelf version".

README.md
marbre: Okay, I introduced `install_pinned_patchelf.sh` as an additional wrapper for now.

docs/environment_setup_guide.md
```diff
@@ -67,3 +50,91 @@ Reference image: `ubuntu:22.04`
...
+   indicates a good build. `dnf install patchelf` is sufficient on a
+   release that carries it.
+3. **Any other Linux (Ubuntu, Debian, Arch, openSUSE, ...).** Build
+   `patchelf` from source using the script the manylinux image uses:
+
+   ```bash
+   sudo env INSTALL_PREFIX=/usr/local ./dockerfiles/install_pinned_patchelf.sh
+   ```
+
+   patchelf --version
```
jayhawk-commits, line 129: Looks like the triple back-ticks are missing for this command invocation in this line.

marbre: There was a bit more :)

jayhawk-commits (approved):
Formatting and content of the docs look OK to me

ScottTodd (approved):

PR #4608, opened 2026-04-16
[hipDNN] Make hipdnn_list_engines test less strict.

CMiservaAMD:
## Motivation

Changing the output of the hipdnn_list_engines tool causes unnecessary test failiure.

## Technical Details

Convert test to be more flexible when verifying tool output by using pattern matching instead of strict string comparison.

## Submission Checklist


BrianHarrisonAMD (approved):
LGTM

PR #4616, opened 2026-04-16
[ci] enabling all archs for nightly tests

geomin12:
As mentioned, we are enabling all archs for nightly test runs to determine what is missing and fixing tests

### DIFF
--- a/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
+++ b/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
@@ -770,18 +770,6 @@ class TestExpandBuildConfigs(unittest.TestCase):
         # Windows has no asan variant config at all.
         self.assertIsNone(result.windows)
 
-    def test_test_runner_kernel_overrides_runner_label(self):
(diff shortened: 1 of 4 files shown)


build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -780,7 +780,7 @@ def test_test_runner_kernel_overrides_runner_label(self):
         )
         self.assertIsNotNone(result.linux)
         entry = result.linux.per_family_info[0]
-        self.assertEqual(entry["test-runs-on"], "linux-strix-halo-gpu-rocm-oem")
+        self.assertEqual(entry["test-runs-on"], "")
```
ScottTodd: This change defeats the purpose of the test. Rework?

geomin12: actually, the `nightly` run flag will disable it for all scenarios except scheduled runs. I guess the developers using these machines just invoke the label directly, so this test should be good to remove. 

once we get all the capacity back for linux gfx1151, we can re-enable this PR label

amd-justchen (approved):
Here we go!

PR #4619, opened 2026-04-16
Support cross-repo triggering of multi-arch workflows

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/3334. This will let us create a wrapper `.github/workflows/multi_arch_release.yml` workflow in https://github.com/ROCm/rockrel for "nightly" and "prerelease" runs like https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_windows_packages.yml.

## Technical Details

Three related changes are included here:
1. Pass `repository` and `ref` through to all multi-arch workflows so `actions/checkout` gets the code from TheRock, not rockrel
3. Add explicit inputs to `.github/workflows/setup_multi_arch.yml` instead of reading from `GITHUB_EVENT_PATH`. Inputs actually came from the original workflow_dispatch, which _typically_ lines up, but being explicit is more reliable, especially when triggering across so many repositories and workflows
2. Remove `_ALLOWED_RELEASE_REPOS` check from `s3_buckets.py` for easier testing in forks (runners may not be available and uploads may fail, but workflows can now attempt to configure for releases anyways)

## Test Plan

* A few test runs from my rockrel fork with a new wrapper workflow, like https://github.com/ScottTodd/rockrel/actions/runs/24524144810
* Unit tests + CI workflows on this PR

## Submission Checklist


build_tools/_therock_utils/s3_buckets.py
```diff
@@ -77,10 +77,6 @@ def write_access_iam_role(self) -> str | None:
 
 _ALLOWED_RELEASE_TYPES = {"dev", "nightly", "prerelease"}
 
-# Repositories allowed to use release_type. Only these repositories are trusted
```
HereThereBeDragons, line 80: i wouldnt remove this. at one point we only want to narrow it down to rockrel

build_tools/tests/s3_buckets_test.py
```diff
@@ -63,15 +63,6 @@ def test_release_type_invalid_raises(self):
             )
         self.assertIn("bogus", str(cm.exception))
 
-    def test_release_type_disallowed_repo_raises(self):
```
HereThereBeDragons, line 66: add it back if you add the repo check for releases again

.github/workflows/setup_multi_arch.yml
```diff
@@ -61,16 +93,26 @@ jobs:
       - name: Checking out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || '' }}
```
HereThereBeDragons, line 97: no need for default values

.github/workflows/multi_arch_release.yml
```diff
@@ -72,6 +80,10 @@ jobs:
       build_variant: "release"
       release_type: ${{ inputs.release_type }}
       prerelease_version: ${{ inputs.prerelease_version }}
+      linux_amdgpu_families: ${{ inputs.linux_amdgpu_families }}
```
HereThereBeDragons, line 83: i guess for the scheduled workflow of nightly release etc we will have another wrapper around it - so no need for { ... || <default value>} here?

.github/workflows/multi_arch_release_linux.yml
HereThereBeDragons, line 84: dont forget to also add repo and ref here

.github/workflows/multi_arch_ci.yml
```diff
@@ -64,6 +64,12 @@ jobs:
     uses: ./.github/workflows/setup_multi_arch.yml
     with:
       build_variant: "release"
+      linux_amdgpu_families: ${{ inputs.linux_amdgpu_families || '' }}
```
HereThereBeDragons, line 67: what does empty mean here? run on all archs?

.github/workflows/multi_arch_build_windows_artifacts.yml
```diff
@@ -73,6 +81,9 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || '' }}
```
HereThereBeDragons, line 86: no need for default as only workflow call

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -77,6 +85,9 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || '' }}
```
HereThereBeDragons, line 90: no workflow dispatch, so no need for default value here

HereThereBeDragons (approved):
see comments

build_tools/_therock_utils/s3_buckets.py
ScottTodd: There may also be releases from other downstream repositories (ROCm internal for example). I'd like to support testing of these workflows from forks and rely more on the actual auth checks where they matter.

.github/workflows/setup_multi_arch.yml
ScottTodd: Thanks, removed defaults for cases where workflow_dispatch or workflow_call are the only triggers.

.github/workflows/multi_arch_release.yml
ScottTodd: Yep. The schedule trigger for nightly releases will be in rockrel, see https://github.com/ROCm/rockrel/pull/35

.github/workflows/multi_arch_ci.yml
ScottTodd: Skip linux (or windows). Added to the input description

.github/workflows/multi_arch_release_linux.yml
ScottTodd, line 94: Missed adding `repository` and `ref ` here, fixing in https://github.com/ROCm/TheRock/pull/4792

PR #4620, opened 2026-04-16
Replace `dvc pull` with in-tree boto3-based fetcher

marbre:
## Motivation

TheRock uses `dvc` for exactly one thing: pulling MD5-content-addressed binary artifacts (MIOpen kernel databases, PAL prebuilt libs) from anonymous-public S3 buckets in `fetch_sources.py`. We never use `dvc add`, `push`, or `repro`.

Pulling in `dvc[s3]` drags `aiobotocore` along, which pins a narrow `botocore` window. That forces the `boto3>=1.42.79,<1.42.85` pin in `requirements.txt:14-19` even though nothing in TheRock actually needs that exact version - it exists solely to make dvc's aiobotocore resolve. Plus a `pathspec<0.13.0` workaround pin in the manylinux Dockerfile because dvc 3.62.0 uses pathspec's private API. Every dvc point release re-shifts this window; the alignment has already been reworked twice.

Replacing `dvc pull` with a small in-tree tool that calls boto3 directly removes the aiobotocore/pathspec workarounds. boto3 stays (it's already used in `build_tools/` for release upload and S3 indexing) but no longer needs to satisfy dvc's transitive pin.

## Technical Details

New `build_tools/fetch_dvc_artifacts.py` provides a drop-in subset of `dvc pull` for TheRock's use case:

- Library API `pull(project_dir, *, jobs, cache_dir, log) -> PullResult` and a `pull` CLI subcommand mirroring `dvc pull`.
- Reads the standard `.dvc` pointer YAML and `.dvc/config` INI in each project; expects the dvc 3.x S3 layout `<prefix>/files/md5/<first2>/<rest30>` used by `therock-dvc`.
- Anonymous S3 access (UNSIGNED), matching `allow_anonymous_login`.
- Parallel downloads across pointer files plus boto3's multipart ranged GETs for large objects.
- Atomic write with post-download MD5 verification; mismatches raise `FetchError` and leave no destination file.
- Optional content-addressed cache at `~/.cache/therock-dvc/` with destination-already-correct and cache-hit fast paths; poisoned cache entries are detected and discarded.
- Supports `.dir` directory hashes for forward compatibility (none in TheRock today).
- Stdlib + boto3 only. No PyYAML, no aiobotocore, no s3fs.

`fetch_sources.py:pull_large_files()` switches from a subprocess call to `dvc pull` to an in-process call to `fetch_dvc_artifacts.pull()`. The `shutil.which("dvc")` probe is dropped since the new tool ships with the repo.

## Out of Scope (Follow-Up)

Per a "tool only first, evaluate before deletion" rollout, this change does not yet:

- Drop `dvc[s3]` and the `pathspec<0.13.0` workaround from `dockerfiles/build_manylinux_x86_64.Dockerfile`.
- Loosen the `boto3` pin in `requirements.txt`.
- Remove the `iterative/setup-dvc@v2.0.0` step from the Windows workflows.
- Swap the standalone `dvc pull` calls in the `rocm-libraries` and `rocm-systems` workflows for the new tool.

A follow-up change will do those after the new tool has had a couple of weeks of bake time in CI alongside dvc.

## Test Plan

Unit:
- `python -m pytest build_tools/tests/fetch_dvc_artifacts_test.py`

Integration (manual, against `s3://therock-dvc/rocm-libraries`):
- Cold pull: delete all `.kdb.bz2` and `~/.cache/therock-dvc`, run the tool, cross-check each fetched file's md5 against its `.dvc` pointer.
- Warm pull (destination present): verify zero S3 calls and skip count equals file count.
- Cache hit pull (destination missing, cache present): verify hardlink path and zero S3 calls.
- Cache poisoning: write garbage to a cache entry, delete the destination, re-pull. Tool must detect size mismatch, discard the cache entry, re-download, and re-populate the cache.
- Atomic write: `kill -9` mid-download. Destination must be absent on disk (the canonical-name file); any leftover `.tmp` is cleaned up on the next pull.
- End-to-end via `fetch_sources.pull_large_files(['rocm-libraries'], ...)`.

## Test Result

- 39 unit tests pass.
- Cold pull: 6 files / 1.2 GiB in 39 s.
- Warm pull: 0.75 s, no S3 calls.
- Cache hit pull: 0.75 s, all 6 files hardlinked from cache.
- Cache poisoning: detected and recovered; cache repopulated with correct bytes.
- Atomic write under SIGKILL: destination absent; `.tmp` cleaned on next pull.
- All 6 fetched files md5-match their `.dvc` pointer values.
- `fetch_sources.pull_large_files()` end-to-end works.
- `pre-commit run --files <changed>` is clean.

## Submission Checklist

- [x]  Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/build_tools/fetch_dvc_artifacts.py
+++ b/build_tools/fetch_dvc_artifacts.py
@@ -1,0 +1,592 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Fetch large binary artifacts referenced by .dvc pointer files from an S3 remote.
+
+Minimal in-tree replacement for `dvc pull` covering TheRock's only use of dvc:
+pulling MD5-content-addressed binary artifacts (MIOpen kernel databases, PAL
+prebuilt libs) from anonymous-public S3 buckets. Does not implement add/push/repro.
+
+Pointer file schema (the only one we handle):
+    outs:
+    - md5: <32-char-hex>      # may end with `.dir` for directory hashes
+      size: <bytes>
+      hash: md5
+      path: <relpath-from-pointer-file-dir>
+
+Remote config (.dvc/config, INI):
+    [core]
+        remote = storage
+    ['remote "storage"']
+        url = s3://<bucket>/<prefix>
+        allow_anonymous_login = true
+
+S3 key scheme (dvc 3.x "new" layout, used by therock-dvc):
+    `<prefix>/files/md5/<md5[:2]>/<md5[2:]>` for both files and `.dir` manifests.
+
+Library:
+    from fetch_dvc_artifacts import pull
+    result = pull(Path("/path/to/project"))
+
+CLI:
+    python fetch_dvc_artifacts.py pull [DIR ...] [--jobs N] [--cache-dir PATH]
+                                       [--no-cache] [-v]
+"""
+
+import argparse
+import concurrent.futures
+import configparser
+import hashlib
+import json
+import logging
+import os
+import shutil
+import sys
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Callable
+from urllib.parse import urlparse
+
+import boto3
+from botocore import UNSIGNED
+from botocore.client import Config
+
+
+DEFAULT_JOBS = 8
+DEFAULT_CACHE_DIR = Path.home() / ".cache" / "therock-dvc"
+
+
+class FetchError(Exception):
+    """Raised on missing config, download failure, or MD5 mismatch."""
+
+
+@dataclass
+class PullResult:
+    fetched: int = 0  # downloaded from remote
+    cached: int = 0  # served from local content-addressed cache
+    skipped: int = 0  # already at destination with matching md5
+
+    def __iadd__(self, other: "PullResult") -> "PullResult":
+        self.fetched += other.fetched
+        self.cached += other.cached
+        self.skipped += other.skipped
+        return self
+
+
+@dataclass(frozen=True)
+class _Out:
+    """One entry from a .dvc pointer file's `outs:` list."""
+
+    md5: str
+    size: int
+    path: str
+
+    @property
+    def is_dir(self) -> bool:
+        return self.md5.endswith(".dir")
+
+    @property
+    def bare_md5(self) -> str:
+        return self.md5[:-4] if self.is_dir else self.md5
+
+
+@dataclass(frozen=True)
+class _Remote:
+    bucket: str
+    prefix: str
+    anonymous: bool
+
+
+# --------------------------------------------------------------------------
+# .dvc YAML subset parser
+# --------------------------------------------------------------------------
+
+
+def _parse_dvc_pointer(path: Path) -> list[_Out]:
+    """Parse a .dvc pointer file. Schema is fixed; full YAML is overkill."""
+    text = path.read_text()
+    outs: list[_Out] = []
+    current: dict[str, str] | None = None
+    in_outs = False
+    for raw_line in text.splitlines():
+        line = raw_line.rstrip()
+        if not line or line.lstrip().startswith("#"):
+            continue
+        if not in_outs:
+            if line == "outs:":
+                in_outs = True
+            continue
+        if line.startswith("- "):
+            if current is not None:
+                outs.append(_to_out(current, path))
+            current = {}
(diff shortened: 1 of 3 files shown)


jayhawk-commits:
@litvaOo and @DarylHawkinsAMD have been using `dvc push` for `amdgpu-windows-interop` and `MIOpen` respectively, and @bstefanuk has been investigating for `hipblaslt` use cases. CC'ing them to this PR for awareness.

marbre:
> @litvaOo and @DarylHawkinsAMD have been using `dvc push` for `amdgpu-windows-interop` and `MIOpen` respectively, and @bstefanuk has been investigating for `hipblaslt` use cases. CC'ing them to this PR for awareness.

Well you can still use dvc but it won't be pre-installed in the Docker containers we use for build.

jayhawk-commits (commented):
### Non-blocking robustness note:

`pull()` parallelizes at the `.dvc` pointer-file level, and `_store_in_cache()` writes through a fixed `"<cache_file>.tmp"` path before `os.replace()`.

If we ever end up with two pointers (or `.dir` manifest entries) that reference the same content hash, those workers could race on the same cache temp path. I checked the current `rocm-libraries`/`rocm-systems` `.dvc` pointers and didn’t see duplicate md5s today, so I don’t think this is a current blocker, but it may be worth either documenting that assumption or hardening the cache write path with unique temp names / per-hash locking.

jayhawk-commits (approved):
Even with the suggestion on the potential temp file name collision, the code changes and the shared passing test results look good to me.

stellaraccident (commented):
Non-blocking observation: `fetch_sources.py --jobs` is not passed through to the new DVC artifact fetcher. `run()` builds git submodule update args from `args.jobs`, but `pull_large_files()` calls `fetch_dvc_artifacts.pull(project_dir)` without a `jobs=` argument, so artifact fetching always uses `fetch_dvc_artifacts.DEFAULT_JOBS` (currently 8) even when CI invokes `fetch_sources.py --jobs 12`.

That may be a perfectly reasonable default, but it is a small behavioral mismatch from the user-visible option: callers can tune git fetch parallelism with `--jobs`, while the DVC artifact fetch parallelism is fixed unless they call the new tool directly. If this is intentional, it may be worth documenting; otherwise passing `args.jobs` through to `pull_large_files()` / `pull(..., jobs=...)` would keep the knob consistent.

stellaraccident (approved):
Days since dvc deps killed me: 0

LGTM

marbre:
> Non-blocking observation: `fetch_sources.py --jobs` is not passed through to the new DVC artifact fetcher. `run()` builds git submodule update args from `args.jobs`, but `pull_large_files()` calls `fetch_dvc_artifacts.pull(project_dir)` without a `jobs=` argument, so artifact fetching always uses `fetch_dvc_artifacts.DEFAULT_JOBS` (currently 8) even when CI invokes `fetch_sources.py --jobs 12`.
> 
> That may be a perfectly reasonable default, but it is a small behavioral mismatch from the user-visible option: callers can tune git fetch parallelism with `--jobs`, while the DVC artifact fetch parallelism is fixed unless they call the new tool directly. If this is intentional, it may be worth documenting; otherwise passing `args.jobs` through to `pull_large_files()` / `pull(..., jobs=...)` would keep the knob consistent.

Thanks for reviewing and the catch. I can address and land early next week but feel free to hit merge if you like to have it earlier. Will then fix in a follow up.

marbre:
CI passed with fetch large files:
```

pull: 11 pointer file(s) from s3://therock-dvc/rocm-libraries
  fetching  bnorm_backward.tar.gz (1.3 KiB)
  fetching  bnorm_fwd.tar.gz (1.3 KiB)
  fetching  conv_dgrad.tar.gz (9.3 KiB)
  fetching  conv_fwd.tar.gz (10.6 KiB)
  fetching  gfx1030.kdb.bz2 (381.5 MiB)
  fetching  conv_wgrad.tar.gz (10.4 KiB)
  fetching  gfx900.kdb.bz2 (270.1 MiB)
  fetching  gfx906.kdb.bz2 (318.4 MiB)
  fetching  gfx90a.kdb.bz2 (147.3 MiB)
  fetching  gfx908.kdb.bz2 (104.7 MiB)
  fetching  gfx942.kdb.bz2 (3.2 MiB)
  done: fetched=11 cached=0 skipped=0
```

PR #4624, opened 2026-04-16
Bump dvc to 3.67.1 and pin aiobotocore in manylinux image

marbre:
The old dvc[s3]==3.62.0 install pulled aiobotocore transitively with a loose bound, so each image rebuild could resolve a different aiobotocore and in turn a different botocore window. That silently drifted out of the boto3 pin in requirements.txt, once requiring a reactive realignment. Add an explicit aiobotocore==3.4.0 pin so rebuilds are deterministic. 3.4.0 matches what the current published image already resolves, so the botocore window (and boto3 pin) is unchanged by this commit.

Bump dvc[s3] from 3.62.0 to the current stable 3.67.1 while here.

Drop the pathspec<0.13.0 pin. It was a workaround for dvc 3.62.0 using pathspec's private _DIR_MARK API that later pathspec releases removed. dvc 3.67.1 no longer touches that private API. pathspec itself stays in the image as a transitive dependency of dvc and scmrepo (now at 1.0.4).

Smoke-tested the rebuilt image:
- pip show reports dvc 3.67.1, dvc-s3 3.3.0, aiobotocore 3.4.0, botocore 1.42.84, pathspec 1.0.4
- pip install -r requirements.txt installs cleanly, pip check passes
- dvc pull + dvc status verify correctly against a tracked file in rocm-libraries/

CI workflow sha256 bumps land in a follow-up commit after the new image is published.

Progress on #4596

### DIFF
--- a/dockerfiles/build_manylinux_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_x86_64.Dockerfile
@@ -88,12 +88,13 @@ RUN yum install -y epel-release && \
 # dvc's rpm package includes .so dependencies built against glib 2.29
 # settling for pip install for now, but it installs modules not needed for dvc pull
 # more dvc features may be used in upcoming sequenced builds
-# Also pinning pathspec because a new version of it breaks the private _DIR_MARK
-# API that dvc uses. When upgrading past ~3.64.0, then pin can likely be removed.
+# Pin aiobotocore explicitly so rebuilds resolve a deterministic botocore
+# window. Without this, dvc's loose `aiobotocore` bound lets future pip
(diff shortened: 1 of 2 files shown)


jayhawk-commits (approved):
The version requirements between the different pieces are well documented in the change and match what is being done, so LGTM.

amd-shiraz (approved):
thanks for the quick workaround. pls wait merging before math-libs build passes :)

marbre:
> thanks for the quick workaround. pls wait merging before math-libs build passes :)

This doesn't update the SHA. If merged the "only" thing it will do is to push a new docker image but that won't be used.

Issue #3321, opened 2026-02-09
Refactor `manage.py` to move server side


marbre:
The `build_tools/third_party/s3_management/manage.py` script should be refactored to move it server side in a subsequent step. Currently `manage.py` is called within workflow like
https://github.com/ROCm/TheRock/blob/e267c57b9a21c28fc75872793418c32703e2821d/.github/workflows/build_portable_linux_pytorch_wheels.yml#L222
after a build was pushed. Instead the script should run server side as AWS Lambda.

As part of this issue it must be evaluated how to refactor this script to allow moving it server side. It is not desired to re-generate _all_ index files in the prefixes:
https://github.com/ROCm/TheRock/blob/e267c57b9a21c28fc75872793418c32703e2821d/build_tools/third_party/s3_management/manage.py#L38-L51

Instead only the index files for a specific prefix (`CUSTOM_PREFIX`?) should be regenerated.

PR #4625, opened 2026-04-16
Add python package publishing to release workflows

ScottTodd:
## Motivation

Progress on

This builds on https://github.com/ROCm/TheRock/pull/4577 to additionally build and publish rocm python packages as part of multi-arch releases.

## Technical Details

* With kpack split _disabled_, this should copy from arch-specific subfolders to `v3/`, distinct from `v2/` used for non-multi-arch releases
* With kpack split _enabled_, this should copy from the flat python index to `v3-multi-arch/`
* Once we complete the flag flip and fully land the multi-arch workstream we can redirect our release documentation, test procedures, etc. to use the new flat index

Files will first appear in the S3 buckets (e.g. https://therock-dev-python.s3.amazonaws.com/index.html), then later we can add cloudfront/CDN mappings from the new directories to https://rocm.devreleases.amd.com/. We may also need to make some changes to the server-side index generation and dependency mirroring so that the new pip indexes are complete. See:
* https://github.com/ROCm/TheRock/tree/main/build_tools/third_party/s3_management
* https://github.com/ROCm/TheRock/issues/3321
* https://github.com/ROCm/TheRock/issues/3324

## Test Plan

1. Trigger a dev release in this repository from this branch (kpack disabled): https://github.com/ROCm/TheRock/actions/runs/24540942333
    - [ ] check for packages uploaded to `https://therock-dev-python.s3.amazonaws.com/v3-multi-arch/v3`
3. Trigger a dev release in this repository from this branch rebased onto a kpack integration branch, check for packages uploaded to `https://therock-dev-python.s3.amazonaws.com/v3-multi-arch/`

## Test Result

TBD

## Submission Checklist


build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -63,6 +76,34 @@ def publish_tarballs(
...
+        s3://therock-dev-artifacts/12345-linux/python/gfx110X-all/*.whl
+          -> s3://therock-dev-python/v3/gfx110X-all/*.whl
+
+    With kpack split enabled (flat):
+        s3://therock-dev-artifacts/12345-linux/python/*.whl
+          -> s3://therock-dev-python/v3-multi-arch/*.whl
+    """
+    source = artifacts_root.python_packages()
+    dest_bucket = get_release_bucket_config(release_type, "python")
+    s3_subdir = "v3-multi-arch" if kpack_split else "v3"
```
marbre, line 97: Please let's _not_ go with `v3-multi-arch`. We want to get rid of the `v*` structure and make it invisible for users via the CF distribution. We didn't yet aligned dev and nightlies the way we have it with pre-releases due to the number of files we would need to move but this needs a clean up and `v3-multi-arch` would add more churn here. Rather go with `v3/whl-multi-arch` please.

ScottTodd: Okay, `v3/whl-multi-arch/`
Then `v3/whl/`? For `v3/whl/gfx110X-all/` etc.? Or `v3/gfx110X-all/`?

marbre: Yeah, the cleanup for legacy would move `v2` -> `v2/whl` (which `v2` "hidden"). For newer wheels it depends.

We can serve legacy wheels from `v3/whl` as `whl` for now. If you upload _new_ multi-arch wheels to `v4/whl` we can make them available as `whl-multi-arch` via an URI rewrite and drop this one later so that `whl-multi-arch` becomes `whl` after some time.

marbre: `gfx*` subdirs are not needed at all.

marbre:
We discussed offline and ideally we want to have the ability to produce "legacy" non-multi-arch _and_ multi-arch, kpack-split enabled builds uploaded to release buckets. I'd propose the following subfolders for this:

* `v3/tarball/` for legacy tarballs
* `v4/tarball/` for  kpack-enabled tarballs
* `v3/whl/gfx*/` for legacy Python packages
* `v4/whl/` for kpack-enabled Python packages

What we expose it as in CF is a different thing (can be e.g. https://rocm.nightlies.amd.com/whl/ and https://rocm.nightlies.amd.com/whl-multi-arch) and is actually loosely coupled to the folder structure in S3. This would allow at some point to drop legacy wheels and let https://rocm.nightlies.amd.com/whl point to multi-arch whenever we feel ready.

The above assumes that we land the flag flip. With kpack-split disabled, multi-arch Python packages are actually "replicating" legacy style packages that could go to v3 folders.

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -63,6 +76,34 @@ def publish_tarballs(
...
+        s3://therock-dev-artifacts/12345-linux/python/gfx110X-all/*.whl
+          -> s3://therock-dev-python/v3/whl/gfx110X-all/*.whl
+
+    With kpack split enabled (flat):
+        s3://therock-dev-artifacts/12345-linux/python/*.whl
+          -> s3://therock-dev-python/v4/whl/*.whl
+    """
+    source = artifacts_root.python_packages()
+    dest_bucket = get_release_bucket_config(release_type, "python")
+    s3_subdir = "v4/whl" if kpack_split else "v3/whl"
```
marbre, line 97: Side note: Current nightlies still push to `v2-staging` and `v2`. However we want to change this to `v3/whl/` which is then exposed as `/whl` in CF. Thus for legacy builds upload will work but won't show up in CF as expected!

marbre (approved):
Thanks, lgtm with the updated paths.

build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd: Ack. I haven't fully considered yet what changes we should make to https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml and related workflows for multi-arch. I'd like to continue having "staging" for torch packages so users will still have a central index where they can download packages that may not have passed all tests. Maybe "v4/whl-staging"? We can copy rocm python packages to both subdirectories or gate promotion on the result of package installs + `rocm-sdk test` 🤷

marbre: `v4/whl-staging` works as well!

PR #4629, opened 2026-04-16
[do not merge] core pattern setting test

lumachad:
Testing container core pattern setting.


lumachad:
Test done. Work in progress.

Issue #1877, opened 2025-10-22
libgfortran missing in nightly pytorch job's sanity tests


jayhawk-commits:
rocm-sdk sanity test failures in pytorch nightly job
Example run with failure: https://github.com/ROCm/TheRock/actions/runs/18679287769/job/53258712185

I don't see these failures with the nightly jobs for the ROCm portable packages.

```
OSError: libgfortran.so.5: cannot open shared object file: No such file or directory
testSharedLibrariesLoad (rocm_sdk.tests.libraries_test.ROCmLibrariesTest.testSharedLibrariesLoad) ... 
  testSharedLibrariesLoad (rocm_sdk.tests.libraries_test.ROCmLibrariesTest.testSharedLibrariesLoad) [Check shared library loads] (so_path=PosixPath('/__w/TheRock/TheRock/.venv/lib/python3.11/site-packages/_rocm_sdk_libraries_gfx94X_dcgpu/lib/libhipsolver.so.1')) ... ERROR

======================================================================
ERROR: testPreloadLibraries (rocm_sdk.tests.core_test.ROCmCoreTest.testPreloadLibraries) [Check rocm_sdk.preload_libraries] (shortname='hipsolver')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/TheRock/TheRock/.venv/lib/python3.11/site-packages/rocm_sdk/tests/core_test.py", line 141, in testPreloadLibraries
    rocm_sdk.preload_libraries(lib_entry.shortname)
  File "/__w/TheRock/TheRock/.venv/lib/python3.11/site-packages/rocm_sdk/__init__.py", line 102, in preload_libraries
    cdll = ctypes.CDLL(str(path), mode=mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/_tool/Python/3.11.14/x64/lib/python3.11/ctypes/__init__.py", line 376, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libgfortran.so.5: cannot open shared object file: No such file or directory

======================================================================
[...]

Issue #1418, opened 2025-09-08
rocm-sdk test shouldn't require Fortran


marbre:
Currently `rocm-sdk test` is failing if gfortran isn't installed. This should not be the case.

---

If fortran is still optional (I believe it should be), users should be able to run `rocm-sdk test` and other commands that don't use fortran and see at _most_ warnings, not errors.

The errors observed are `ERROR: testSharedLibrariesLoad (rocm_sdk.tests.libraries_test.ROCmLibrariesTest.testSharedLibrariesLoad) [Check shared library loads] (so_path=PosixPath('/__w/TheRock/TheRock/.venv/lib/python3.11/site-packages/_rocm_sdk_libraries_gfx94X_dcgpu/lib/libhipsolver_fortran.so.1'))`

The requirement comes from `*_fortran.so` files:
```
λ ls C:\Users\Nod-Shark16\Downloads\therock-dist-linux-gfx94X-dcgpu-7.0.0rc20250905\lib | grep solver
libhipsolver.so@
libhipsolver.so.1@
libhipsolver.so.1.0
libhipsolver_fortran.so@
libhipsolver_fortran.so.1@
libhipsolver_fortran.so.1.0
librocsolver.so@
librocsolver.so.0@
librocsolver.so.0.7
```

Here is the test code: https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/libraries_test.py . We could make those fortran tests optional (perhaps conditional on a check for the dependent libraries). See how certain tests are skipped in https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/devel_test.py and https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py

_Originally posted by @ScottTodd in https://github.com/ROCm/TheRock/pull/1416#discussion_r2330601266_

marbre:
Fixed with commit 8ab8580.

PR #4630, opened 2026-04-17
Re-admit hipSOLVER fortran shim in kpack-split devel wheel

marbre:
## Motivation

Downstream CMake consumers, like PyTorch, that `find_package(hipsolver)` against the kpack-split rocm-sdk-devel wheel fail at configure time with:

```
CMake Error at .../cmake/hipsolver/hipsolver-targets.cmake:92:
  The imported target "roc::hipsolver_fortran" references the file
  ".../lib/libhipsolver_fortran.so.1.0" but this file does not exist.
```

`hipsolver-targets-release.cmake` (shipped in the `dev` component) defines `roc::hipsolver_fortran` with an IMPORTED_LOCATION pointing at `libhipsolver_fortran.so.1.0`. However, the lib lives in the `test` artifact because the hipsolver-test binary links against it. The kpack-split path added `exclude_components=["test"]` (commit b6b306b4) to drop generic test. Moving ther shim into the the lib artifact is not a viable solution as this would introduce a runtime dependency on libgfortran for every consumer. Moving to dev would break tests.

## Technical Details

Adds a `component_include_overrides: Mapping[str, Sequence[str]]` parameter to `populate_devel_files` in `py_packaging.py`. For each excluded component with a mapping entry, run a secondary `filter_artifacts` admitting only that component and scoping the PathMatcher to the supplied include globs, then populate through the same `_populate_devel_file` path as the main pass.

With `{"test": ["lib/libhipsolver_fortran.so*"]}` set and wired into `_run_kpack_split` we have a strict match that only matches the hipsolver fortran libraries and nothing else from the test artifacts.

Legacy-mode devel wheel behavior is unchanged (no override passed from `_run_legacy`).

## Test Plan

1. Rebuild rocm Python packages from an existing kpack-split artifact tree.
2. Inspect the devel wheel tarball for the hipsolver fortran variants and absence of other `test`-component files.
3. End-to-end: install `rocm[libraries,devel,device-gfx1201]` from the local dist in a manylinux container, run the PyTorch release/2.10 build via `external-builds/pytorch/build_prod_wheels.py`, and confirm it progresses past the hipsolver-targets.cmake import check.

## Test Result

* Repackaged locally against `/develop/therock/artifacts` (kpack-split, gfx1200/gfx1201) in the manylinux builder image. The devel wheel's `_devel.tar` now contains `_rocm_sdk_devel/lib/libhipsolver_fortran.so{,.1,.1.0}`. No `hipsolver-test` or `hipsolver-bench`.
* PyTorch `release/2.10` (torch-only, cp312, gfx1201) built successfully end-to-end.

## Submission Checklist

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -149,6 +149,15 @@ def _run_kpack_split(
             "nlohmann-json",
         ],
         exclude_components=["test"],
+        # hipSOLVER's fortran interop shim lives in the `test` component
+        # alongside `hipsolver-test`, which links against it at runtime. Devel
+        # consumers also need it: `hipsolver-targets.cmake` exports
+        # `roc::hipsolver_fortran` with an IMPORTED_LOCATION pointing at the
+        # shim, and CMake's generated import-check fails the configure step
+        # of downstream projects (e.g. PyTorch) when the file is absent.
+        component_include_overrides={
+            "test": ["lib/libhipsolver_fortran.so*"],
+        },
         tarball_compression=args.devel_tarball_compression,
     )
     if args.build_packages:
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -3,7 +3,7 @@
 
 """Utilities for producing Python packages."""
 
-from typing import Callable, Sequence
+from typing import Callable, Mapping, Sequence
 
 import importlib.util
 import re
@@ -486,9 +486,19 @@ class PopulatedDistPackage:
         *,
         addl_artifact_names: Sequence[str] = (),
         exclude_components: Sequence[str] = (),
+        component_include_overrides: Mapping[str, Sequence[str]] = {},
         tarball_compression: bool = True,
     ):
-        """Populates all files that have not yet been materialized and symlink the rest."""
+        """Populates all files that have not yet been materialized and symlink the rest.
+
+        ``component_include_overrides`` re-admits narrow file patterns from
+        components listed in ``exclude_components``. Keys are component names;
+        values are glob patterns matched against relpaths. Used when an
+        excluded component is the canonical home for files that are still
+        needed in devel (e.g. the BLAS fortran interop shims live in ``test``
+        because the test binary links against them at runtime, yet devel
+        consumers also need them resolved through exported CMake targets).
+        """
         package_path = self.platform_dir
         # Set up for what artifacts to include in the devel package, including
         # any emitted runtime artifacts plus additional requested.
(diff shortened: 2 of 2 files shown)


build_tools/build_python_packages.py
```diff
@@ -149,6 +149,15 @@ def _run_kpack_split(
...
         exclude_components=["test"],
+        # hipSOLVER's fortran interop shim lives in the `test` component
+        # alongside `hipsolver-test`, which links against it at runtime. Devel
+        # consumers also need it: `hipsolver-targets.cmake` exports
+        # `roc::hipsolver_fortran` with an IMPORTED_LOCATION pointing at the
+        # shim, and CMake's generated import-check fails the configure step
+        # of downstream projects (e.g. PyTorch) when the file is absent.
+        component_include_overrides={
+            "test": ["lib/libhipsolver_fortran.so*"],
+        },
```
ScottTodd, line 160: I don't think this is the right approach. If this library is used by downstream projects (thus outside of tests), it does not belong in a test artifact.

Why are we only spotting this with multi-arch?
> - Devel package excludes test component in kpack-split mode since
generic test binaries cannot run without device code

Okay... but are there other issues, or just this one?

> Moving ther shim into the the lib artifact is not a viable solution as this would introduce a runtime dependency on libgfortran for every consumer. Moving to dev would break tests.

I'm not following the reasoning here. What's the difference between including in `lib` and what this code does with an include override?

ScottTodd (commented):
This needs more careful review.

build_tools/build_python_packages.py
marbre: I think when we packages it _only_ was used for testing, as there have been no PyTorch build done by us. Idk when / why PyTorch picks this up and it definitely needs a rework.

>>   Devel package excludes test component in kpack-split mode since
>>   generic test binaries cannot run without device code
>
>Okay... but are there other issues, or just this one?

Actually, idk and would need to go deeper down the rabbit hole here.



>> Moving ther shim into the the lib artifact is not a viable solution as this would introduce a runtime >>dependency on libgfortran for every consumer. Moving to dev would break tests.
>
>I'm not following the reasoning here. What's the difference between including in lib and what this code >does with an include override?

If you add it to lib, `rocm-sdk test` will fail if you haven't gfortran installed. This broke several times in the past, see #1418 and #1877.


I agree that this is not a viable long term fix but it mimics what is done in the legacy builds. Those actually include the lib in rocm-sdk-devel. I can file an issue to kick of some further analysis and to root cause _why_ we have the CMake entry point and why PyTorch picks it up.

ScottTodd: I did some more digging, @stellaraccident can you help us figure out who / which team to route this to and make a decision on how to unbreak the pytorch build? I'd rather not introduce a new type of artifact filtering for python packaging.

* I don't see any code in pytorch using ROCm's fortran support. I suspect this is just `find_package(hipsolver)` failing as it expects all libraries that hipsolver was _built with_ to be available _at install time_, which is not the case when libraries are split across dev/lib/test artifacts that are only conditionally included in a distribution. Here's the generated `hipsolver-targets.cmake` file: https://gist.github.com/ScottTodd/d5da4f9479cbffc3987efb42139933c7. The generation process has a few layers of indirection in it, but I traced it partially to the `rocm_install_targets` function from this file: https://github.com/ROCm/rocm-cmake/blob/develop/share/rocmcmakebuildtools/cmake/ROCMInstallTargets.cmake 
* A few other subprojects like rocblas include fortran code but I don't see dedicated .so files like `libhipsolver_fortran.so`. For example, rocblas installs `rocblas-example-fortran-axpy` etc. here: https://github.com/ROCm/rocm-libraries/blob/4795d7d62738d57a5682bd518417814d2bede6cf/projects/rocblas/next-cmake/CMakeLists.txt#L344-L354
* https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipsolver/library/src/hipsolver_module.f90 has been deprecated since https://github.com/ROCm/rocm-libraries/commit/687222fca854177142fa6b662e2eb5e48b5f01c5 (August 2024), in favor of https://github.com/ROCm/hipfort. I wonder if we could disable `BUILD_FORTRAN_BUILDINGS` in hipsolver safely? https://github.com/ROCm/rocm-libraries/blob/4795d7d62738d57a5682bd518417814d2bede6cf/projects/hipsolver/CMakeLists.txt#L85 --> https://github.com/ROCm/rocm-libraries/blob/4795d7d62738d57a5682bd518417814d2bede6cf/projects/hipsolver/library/src/CMakeLists.txt#L75-L80

cc a few people from https://github.com/ROCm/hipSOLVER/pull/312 : @cgmb @tfalders @EdDAzevedo

marbre: Another way to mitigate this and an alternative to this PR is https://github.com/ROCm/rocm-libraries/pull/6530.

marbre:
Superseded by #4650.

PR #4635, opened 2026-04-17
[hip-kernel-provider]: Add a FF for hip-kernel-provider

adickin-amd:
## Motivation

Add a FF for the hip-kernel-provider so it can default off for theRock and then I can turn it on in Rocm-libraries

### DIFF
--- a/ml-libs/CMakeLists.txt
+++ b/ml-libs/CMakeLists.txt
@@ -308,45 +308,47 @@ if(THEROCK_ENABLE_HIPKERNELPROVIDER)
   # hipDNN hip kernel provider
   ##############################################################################
 
-  therock_cmake_subproject_declare(hipkernelprovider
-    EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_LIBRARIES_SOURCE_DIR}/dnn-providers/hip-kernel-provider"
-    BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/hipkernelprovider"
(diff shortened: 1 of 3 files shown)


ml-libs/CMakeLists.txt
```diff
@@ -308,45 +308,47 @@ if(THEROCK_ENABLE_HIPKERNELPROVIDER)
...
-      therock-nlohmann-json
-    RUNTIME_DEPS
-      hip-clr
-  )
-  therock_cmake_subproject_glob_c_sources(hipkernelprovider
-  SUBDIRS
-    .
-  )
-  therock_cmake_subproject_activate(hipkernelprovider)
+  if(HIP_KERNEL_PROVIDER_ENABLE)
```
adickin-amd, line 311: as I understand it, I should be using the FF in rocm-libraries in side the project to gate it.  However that requires me to get a rocm-libraries bump.  Just testing this to see if it could work here instead.

adickin-amd: This worked but we do have a build error on windows https://github.com/ROCm/rocm-libraries/pull/6519

adickin-amd: This should be safe to come into TheRock though!

jayhawk-commits (approved):
Thanks for sharing results of this project being built in a rocm-libraries job, while this PR's CI job is not building the project. Since that matches intended behaviour, LGTM.

PR #4636, opened 2026-04-17
Add CUDA containers publishing

litvaOo:
## Motivation
We intend to add NVidia building and testing pipelines. This PR is the first part, providing CUDA images which will be further used in the future pipeline.

## Technical Details
We intend to support two versions of CUDA: 12.9 and 13.2. The CUDA image is based on the build_maylinux_x86_64.Dockerfile, only adding cuda-toolkit package. 

**NOTE:** hip-test build and test enablement will be dependent on availability of Nvidia runner.
## Test Plan
1. Dockerfile build should pass for both versions
2. Hip-tests should be able to compile and run using these docker containers

## Test Result
Docker build and hip-tests compilation pass succesfully locally, hip-test build for Nvidia will be provided in a separate PR

## Submission Checklist

### DIFF
--- a/dockerfiles/build_manylinux_cuda_12_9_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_cuda_12_9_x86_64.Dockerfile
@@ -139,4 +139,3 @@ COPY install_cuda.sh ./
 RUN ./install_cuda.sh "12.9" && rm -rf /install-cuda
 ENV PATH="/usr/local/cuda/bin:${PATH}"
 ENV LD_LIBRARY_PATH="/usr/local/cuda/lib64:${LD_LIBRARY_PATH}"
-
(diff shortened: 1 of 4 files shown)


dockerfiles/build_manylinux_cuda_12_9_x86_64.Dockerfile
marbre, line 1: There should be no need to build those from scratch and it could be additional layers on top of our existing manylinux image.

dockerfiles/build_manylinux_cuda_13.2_x86_64.Dockerfile
marbre, line 1: Same here. This fully duplicated the above where the only change is the CUDA version? This adds an unnecessary maintenance burden.

marbre (changes requested):
I am not sure this should go into TheRock or if another home is a better fit. Eventually let's discuss this offline before moving forward here.

PR #4650, opened 2026-04-17
Don't export hipsolver_fortran via hipSOLVER targets

marbre:
## Motivation

In split-packaging layouts like TheRock's kpack-split rocm-sdk-devel wheel, the exported CMake target
`roc::hipsolver_fortran` breaks `find_package(hipsolver)` for unrelated downstream consumers (e.g. PyTorch). The target references `libhipsolver_fortran.so.1.0`, but the shim lives in the test artifact (where hipsolver-test links it at runtime) and therefore does not land in the devel wheel. CMake's import-check loop asserts the file exists and fails configure.

PyTorch itself never links the Fortran shim - verified against the locally built wheel, no pytorch shared object lists libhipsolver_fortran.so* or libgfortran.so* in DT_NEEDED - so the configure-time error is spurious. Upstream hipSOLVER deprecated hipsolver_module.f90 in Aug 2024 in favour of hipfort, so the dangling public target has no known consumer going forward.

## Technical Details

https://github.com/ROCm/rocm-libraries/pull/6530 introduces a new `EXPORT_FORTRAN_BINDINGS` option to hipSOLVER thsat gates registering hipsolver_fortran in the hipsolver-targets export set; The companion patch part of this commit adds the option to the currently-pinned rocm-libraries submodule. In TheRock, `-DEXPORT_FORTRAN_BINDINGS=OFF` is passed resulting in the devel wheel no longer referencing the dangling target. The `.so` still builds and installs and is used hipsolver-test.

Supersedes the packaging-layer workaround in #4630.

## Test Plan

Build hipSOLVER in Docker with the patch applied; inspect `lib/cmake/hipsolver/hipsolver-targets{,-release}.cmake` to verify no references to `roc::hipsolver_fortran`. Confirm `libhipsolver_fortran.so{,.1,.1.0}` is still installed under `lib/` and that `hipsolver-test`'s DT_NEEDED still includes it.

## Test Result

Verified locally: both cmake files contain zero references to `roc::hipsolver_fortran`, the umbrella foreach lists only `roc::hipsolver`, the `.so` is installed, and
`hipsolver-test` still links `libhipsolver_fortran.so.1` (RPATH-resolved to `$ORIGIN/../lib`).

## Follow-up

Once the upstream option lands in rocm-libraries and TheRock's submodule pin is bumped to include it, the
`0001-Add-EXPORT_FORTRAN_BINDINGS-option.patch` file in `patches/amd-mainline/rocm-libraries/` can be dropped.

## Submission Checklist


ScottTodd (approved):
Land this as a patch is okay to unblock the kpack flag flip for multi-arch.

marbre:
Tested locally, thus we agreed to not wait for the full CI signal. If required, we will revert.

Issue #4561, opened 2026-04-15
[kpack] Patchelf RPATH extend on top of KPACK PHDR relocation breaks ELF


stellaraccident:
Failing job is e.g. https://github.com/ROCm/TheRock/actions/runs/24435014356/job/71411934144?pr=4549
Local log: https://gist.github.com/marbre/8d92dd06ee393f8fb0979c9a17b3bbc2
Claude report (WIP): https://gist.github.com/marbre/ee6fb1d24d05607e3ee1a1608f1c1ca3
https://therock-ci-artifacts.s3.amazonaws.com/24435014356-linux/index.html

stellaraccident:
Confirmed the exact py_packaging.py patchelf:

  - Workflow job: Linux::release / Build Python Packages / Build Python | multi-arch-release
  - Job id: 71410219410
  - Container image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:d6ae5712a9c7e8b88281d021e907b312cd8a26295b95690baef3e8dde4805858
  - py_packaging.py calls bare "patchelf" at build_tools/_therock_utils/py_packaging.py:452, 464, and 481, so it uses PATH.
  - In that image, PATH resolves it to:
      - /usr/local/bin/patchelf
      - version: patchelf 0.17.2

  I also inspected the image layers directly. They show:

  - /usr/local/bin/patchelf
  - Python package metadata: patchelf-0.17.2.2
  - /usr/bin/patchelf also exists in a later layer, but /usr/local/bin wins in PATH, matching your which -a.

  I reproduced the exact failure with upstream patchelf 0.17.2 and the extracted artifact:

  - Input: extracted librocblas.so.5.4
  - Sequence: same as py_packaging.py: --add-rpath, then --set-rpath ... --force-rpath
  - Result with 0.17.2:
      - readelf: Error: the PHDR segment is not covered by a LOAD segment
      - PHDR becomes VirtAddr 0xfffffffffff79040
      - e_phoff is moved from 0x1992000 to 0x40
  - Result with 0.18.0:
      - e_phoff stays at 0x1992000
      - PHDR remains covered by LOAD
      - no readelf PHDR error

  So the precise root here is: the published manylinux build image uses /usr/local/bin/patchelf 0.17.2, and 0.17.2 corrupts the kpack-relocated PHDR layout under the packaging RPATH sequence. 0.18.0 fixes this specific repro.

PR #4656, opened 2026-04-17
[packaging] Reserve RPATH pad to avoid `patchelf` breaking ELF layout on RHEL 8.10 (#4271)

### LINKED ISSUE #4271: [Issue]: CI nightly hip programs fail to compile via pip install on RHEL 8.10 baremetal.
### Problem Description

You won't be able to compile any hip programs If you install the latest nightly on RHEL 8.10 via the pip install method. Here is the error you will get

```
clang++: error: unable to execute command: Segmentation fault
clang++: error: amdgcn-link command failed due to signal (use -v to see invocation)
AMD clang version 22.0.0git (https://github.com/ROCm/llvm-project.git 4adeabb0862ea8119c143d9f8256475b0b687217+PATCHED:f3b5643f91ad4def7b92cd48247bc11f1f39fb5c)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/rocm/.venv/lib/python3.11/site-packages/_rocm_sdk_core/lib/llvm/bin
clang++: error: unable to execute command: Segmentation fault
clang++: note: diagnostic msg: Error generating preprocessed source(s).
gmake[2]: *** [CMakeFiles/hip_hello_world.dir/build.make:75: CMakeFiles/hip_hello_world.dir/main.hip.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/hip_hello_world.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

I initially created a ticket here ROCM-20033 and assigned it to the compiler team. Sam from the compiler team debugged the issue and here is what he said.

> We reproduced the issue on the RHEL 8.10 machine. The crash happens because the pip-packaged clang-offload-bundler binary has a non-standard ELF layout — its base address is at 0x3ff000 instead of the normal 0x400000, and its first LOAD segment is read-write instead of read-only. This is caused by patchelf modifying the binary RPATH during pip package creation in TheRock build system. The RHEL 8.10 kernel (4.18) has a bug in its ELF loader that cannot handle this layout, so execve fails with EEXIST and the process crashes with SIGSEGV. Newer kernels (5.x+) handle it fine, which is why it works in Docker on an Ubuntu host but not on a RHEL 8.10 host. As a workaround, users can invoke the binary through the dynamic linker directly: /lib64/ld-linux-x86-64.so.2 /clang-offload-bundler. To fix this properly, the TheRock packaging should either pre-allocate space in the ELF dynamic section so patchelf does not need to insert a new segment, or ship wrapper scripts for older kernels.

> Hi Parag, I took a look at this and it seems like the issue is coming from TheRock's pip packaging scripts rather than the compiler itself. The patchelf post-processing and the exe stub generator were written by Stella Laurenzo, so it might be worth reaching out to her to see how best to get this ticket reassigned to the TheRock team.

### Operating System

RHEL 8.10

### CPU

12th Gen Intel(R) Core(TM) i7-12700K

### GPU

Navi31 XTX

### ROCm Version

ROCm 7.13 (Nightly)

### ROCm Component

_No response_

### Steps to Reproduce

1. Install RHEL 8.10 on a _baremetal_ system. Running RHEL 8.10 on a docker does _not_ reproduce the error.
2. Install amdgpu version 30.30. This issue was created when the latest version of amdgpu released was 30.30.

**Steps to install amdgpu version 30.30**

```
# Install prerequisites
sudo dnf update --releasever=8.10 --exclude=\*release\*
wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
sudo rpm -ivh epel-release-latest-8.noarch.rpm
sudo dnf config-manager --enable codeready-builder-for-rhel-8-x86_64-rpms
sudo usermod -a -G render,video $LOGNAME # Add the current user to the render and video groups
# Restart machine

# Install amdgpu 30.30
sudo dnf install https://repo.radeon.com/amdgpu-install/7.2/rhel/8/amdgpu-install-7.2.70200-1.el8.noarch.rpm
sudo dnf clean all
sudo dnf install "kernel-headers-$(uname -r)" "kernel-devel-$(uname -r)"
sudo dnf install amdgpu-dkms
# Restart machine again.
```

**Install latest ROCm nightly**

```
sudo dnf install -y python3.11 python3.11-pip libatomic
python3.11 -m venv .venv
source .venv/bin/activate
python -m pip install --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/ "rocm[libraries,devel]"
rocm-sdk init
```

**Download rocm-examples and set environment variables**

```
sudo dnf install -y wget sudo python3 gcc-c++ git cmake glfw-devel vulkan-headers vulkan-loader-devel vulkan-validation-layers mesa-libGL-devel gcc-toolset-11 ninja-build
git clone https://github.com/ROCm/rocm-examples.git -b "release/therock-7.11"
cd rocm-examples/HIP-Basic/hello_world

export ROCM_PATH=/home/rocm/.venv/lib64/python3.11/site-packages/_rocm_sdk_devel
export LD_LIBRARY_PATH=/home/rocm/.venv/lib64/python3.11/site-packages/_rocm_sdk_devel/lib
export HIP_PLATFORM=amd
export HIP_PATH=/home/rocm/.venv/lib64/python3.11/site-packages/_rocm_sdk_devel
export HIP_CLANG_PATH=/home/rocm/.venv/lib64/python3.11/site-packages/_rocm_sdk_devel/llvm/bin
export HIP_DEVICE_LIB_PATH=/home/rocm/.venv/lib64/python3.11/site-packages/_rocm_sdk_devel/lib/llvm/amdgcn/bitcode
export PATH="$ROCM_PATH/bin:$HIP_CLANG_PATH:$PATH"

source /opt/rh/gcc-toolset-11/enable
export CXX=/opt/rh/gcc-toolset-11/root/usr/bin/g++
export CC=/opt/rh/gcc-toolset-11/root/usr/bin/gcc
```

**Build the hello_world example**

```
cmake -B build -DROCM_ROOT=$ROCM_PATH .
cmake --build build
```

You will run into the error above.

Let me know if you need a baremetal RHEL 8.10 system to reproduce this error. I can setup one up for you.

_No response_

### Additional Information

_No response_

lucbruni-amd:
## Motivation

Fixes #4271. `pip install rocm[libraries]` on RHEL 8.10 baremetal failed to
compile any HIP program because `clang-offload-bundler` (and every other
shipped executable) crashed with `SIGSEGV` before `main()` ever ran. The
crash is in the kernel's (4.18) `execve()` path, not in the binary itself.

Root cause: `py_packaging.py` runs `patchelf --add-rpath` + `--set-rpath`
on every shipped ELF to rewrite `$ORIGIN` relative paths into the wheel's
final `_rocm_sdk_core` layout. The new RPATH strings are longer than the
originals, so patchelf has to extend `.dynstr`. To make room, patchelf
prepends a writable `PT_LOAD` segment ahead of the canonical read-only
first segment. RHEL 8.10's 4.18 kernel rejects that layout in
`load_elf_binary()` and signals SIGSEGV to the parent. Newer kernels
(>= 5.4) tolerate it, which is why the nightly only broke on RHEL 8.10.

The issue report frames this in `ET_EXEC` terms (base `0x400000` →
`0x3ff000`), but TheRock ships 100% `ET_DYN` / PIE executables, so the
same kernel bug manifests as a **writable first `PT_LOAD`** (base stays
at `0x0`). Either way the invariant "first `PT_LOAD` is not writable" is
what the kernel's ELF loader checks.

## Technical Details

The fix pre-allocates enough RPATH string space at link time that the
later `patchelf --set-rpath` can overwrite in place without ever needing
to grow `.dynstr`. No patchelf behaviour change, no submodule changes,
no kernel workaround.

**CMake side** (`cmake/therock_subproject_utils.cmake`,
`cmake/therock_subproject.cmake`):

- Define `THEROCK_INSTALL_RPATH_PAD_SIZE` (1024), a marker string
  `__therock_patchelf_pad__`, and both a CMake-list form
  (`THEROCK_INSTALL_RPATH_PAD`) and colon-joined form
  (`..._PAD_COLON`) of the pad.
- Inject all four variables into every subproject's `_init.cmake` so the
  pad propagates to every `ExternalProject_Add` configure step without
  per-component boilerplate.
- `therock_set_install_rpath` appends the pad to the auto-managed
  `INSTALL_RPATH` on non-Darwin ELF platforms.
- Log output is kept clean: `message(STATUS ...)` prints
  `(+patchelf pad)` instead of the full 1 KB filler, so from-source
  build logs stay readable.

**NO_INSTALL_RPATH carve-outs** — three subprojects set their own
linker flags and bypass `therock_set_install_rpath`. They get the pad
from the super-project so no submodule changes are needed:

- `compiler/pre_hook_amd-llvm.cmake` — appends the pad to
  `CMAKE_INSTALL_RPATH` and `LIBOMP_INSTALL_RPATH`.
- `compiler/pre_hook_amd-comgr.cmake` — appends the pad to
  `CMAKE_INSTALL_RPATH`.
- `compiler/post_hook_hipify.cmake` (new) — `hipify-clang` bakes its
  RPATH via `LINK_FLAGS` in the submodule's `CMakeLists.txt`. The
  post-hook appends the colon-joined pad to those `LINK_FLAGS` from the
  super-project, guarded on the exact `-Wl,--rpath,$ORIGIN/../lib` form
  the submodule currently uses. If that form ever changes upstream the
  CI gate below catches it.

**Python packaging side** (`build_tools/_therock_utils/py_packaging.py`):

- `_extend_rpath` + `_normalize_rpath` are collapsed into a single
  `_rewrite_rpath` that reads the current RPATH, strips the pad marker
  and any trailing `$ORIGIN/.__therock_patchelf_pad___XXX...` entries,
  appends the dependency RPATHs computed for the wheel layout, and
  writes the result back with one `patchelf --set-rpath --force-rpath`
  call. Because we never exceed the pre-allocated string space,
  `.dynstr` is overwritten in place and the ELF program headers stay
  exactly as the linker produced them.
- Packaging log output collapses the pad entry to `<pad>` in the
  before/after RPATH line.

## Test Plan

Two CI gates added to `rocm_sdk.tests.core_test` (runs via `rocm-sdk
test` in `test_rocm_wheels.yml` on every PR and nightly):

- `testPatchelfPadStripped` — greps every shipped file for
  `__therock_patchelf_pad__`; asserts zero hits. Catches regressions
  where a newly added target bypasses both the auto-RPATH path and
  `_rewrite_rpath` (pad would leak into the wheel).
- `testExecutableElfLayoutIntact` — parses the program headers of every
  64-bit LE ELF under `_rocm_sdk_core/bin` and asserts the first
  `PT_LOAD` segment is not writable. This is the exact invariant the
  RHEL 8.10 kernel checks; an `RW`-first binary would reproduce #4271.

Local validation against a from-source build and `build_python_packages.py`
output:

- CMake configure: every subproject's `_init.cmake` carries the pad
  definitions.
- Compile + link: every auto-managed target, `amd-llvm`, `amd-comgr`,
  and `hipify-clang` ship with the 1 KB pad in their `DT_RUNPATH` /
  `DT_RPATH` at stage time.
- `build_python_packages.py`: every executable logs `REWRITE_RPATH:
  ... -> <final-rpath>` with no pad entry in the final form.
- `rocm_sdk_core-*.whl`: `__therock_patchelf_pad__` not present in any
  wheel member; census of all 102 shipped ELF executables shows first
  `PT_LOAD` is `R`-only and base vaddr is `0x0` (canonical PIE).

End-to-end on the RHEL 8.10 baremetal system from the original repro:
- `readelf -lW` on the installed `clang-offload-bundler` shows
  first `PT_LOAD flags: R vaddr=0x0000000000000000` and a clean
  `DT_RPATH` with no pad marker.
- `strace -e execve` returns `execve(...) = 0` — the kernel accepts the
  layout. The original bug is cured.

## Test Result

Repro environment:
- RHEL 8.10, kernel `4.18.0-553.117.1.el8_10.x86_64`, glibc 2.28
- Wheel built locally from this branch (from-source, Ubuntu 24.04 host)

Before (baseline nightly build):
- `execve()` → SIGSEGV before `ld.so` runs, `dmesg` shows a segfault
  record for the exec'd binary.

After (this branch):
- `execve()` → `0`, `ld.so` runs, binary proceeds to dynamic linking.
- The installed-on-RHEL-8.10 wheel still errors from the newer host's
  glibc 2.38 / `GLIBCXX_3.4.30` symbol references (unrelated ABI
  mismatch — local build was not a manylinux container). CI's
  `manylinux_2_28` pipeline will produce the ABI-correct wheel; the
  kernel-layout cure demonstrated here is independent of the ABI issue.

Full from-source build and packaging logs available on request.

## Follow-up

- Land this, then run the existing manylinux_2_28 CI wheel on the RHEL
  8.10 box to confirm full end-to-end `pip install rocm[libraries]` +
  HIP compile works in the intended deployment configuration.
- The new `compiler/post_hook_hipify.cmake` is the only place where the
  super-project special-cases hipify's hand-rolled `LINK_FLAGS`. If
  hipify upstream grows richer RPATH handling, the
  `testExecutableElfLayoutIntact` CI gate will flag any regression and
  the post-hook can be dropped.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -52,6 +52,21 @@ ENABLED_VLOG_LEVEL: int = 5
 PATCHELF_PAD_MARKER = "__therock_patchelf_pad__"
 
 
+def _format_rpath_for_log(rpath: str) -> str:
+    """Collapse the pad filler to '<pad>' for readable REWRITE_RPATH logs.
+
+    The pad entry is ~1KB of X's and dumping it unmodified into every log line
+    makes `py_packaging.log` unreadable and spooks users watching the build
+    (issue #4271). The on-disk value is untouched; this only affects logging.
+    """
+    if not rpath:
+        return rpath
+    return ":".join(
+        "<pad>" if PATCHELF_PAD_MARKER in entry else entry
+        for entry in rpath.split(":")
+    )
+
+
 def log(*args, vlog: int = 0, **kwargs):
     if vlog > ENABLED_VLOG_LEVEL:
         return
@@ -499,7 +514,10 @@ class PopulatedDistPackage:
         if new_rpath == existing_rpath:
             return
 
-        log(f"  REWRITE_RPATH: {file_path}: {existing_rpath!r} -> {new_rpath!r}")
+        log(
+            f"  REWRITE_RPATH: {file_path}: "
+            f"{_format_rpath_for_log(existing_rpath)!r} -> {new_rpath!r}"
+        )
         subprocess.check_call(
             [
                 "patchelf",
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -36,6 +36,21 @@ assert DIST_INFO_PATH.exists()
 
 ENABLED_VLOG_LEVEL: int = 5
 
+# Marker substring baked into RPATH pad entries at link time by
+# cmake/therock_subproject_utils.cmake. ELFs TheRock ships are linked with a
+# ~1KB filler RPATH entry containing this marker so that when we later rewrite
+# RPATHs below with `patchelf --set-rpath`, the replacement always fits inside
+# the .dynstr allocation that was reserved at link time. Patchelf therefore
+# overwrites in place and does not need to prepend a new PT_LOAD segment,
+# which is the mutation that crashes execve() on RHEL 8.10 / EL 4.18 kernels.
+# See https://github.com/ROCm/TheRock/issues/4271 for the full diagnosis.
+#
+# The same constant is duplicated in
+# build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
+# because that file ships inside the wheel and cannot import from build-only
+# modules. Keep the two string literals in sync.
+PATCHELF_PAD_MARKER = "__therock_patchelf_pad__"
+
 
 def log(*args, vlog: int = 0, **kwargs):
     if vlog > ENABLED_VLOG_LEVEL:
@@ -428,10 +443,11 @@ class PopulatedDistPackage:
             # Update RPATHs on Linux.
             file_type = get_file_type(dest_path)
             if file_type == "exe" or file_type == "so":
-                self._extend_rpath(dest_path)
-                self._normalize_rpath(dest_path)
+                self._rewrite_rpath(dest_path, self._dep_rpath_entries(dest_path))
 
-    def _extend_rpath(self, file_path: Path):
+    def _dep_rpath_entries(self, file_path: Path) -> list[str]:
+        """$ORIGIN-relative RPATH entries contributed by rpath_deps packages."""
+        entries: list[str] = []
         for dep_project, rpath in self.rpath_deps:
             parent_relpath = self._platform_dir.parent.relative_to(
                 file_path.parent, walk_up=True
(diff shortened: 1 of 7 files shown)


lucbruni-amd:
Please mind the length of this PR in terms of both code changes and description. It is quite large, but we can treat this as a draft/experimental solution to the issue and as a means to discuss the final solution or alternatives (hence my request for more reviewers - feel free to remove yourself if your queue is overloaded).

ScottTodd:
Is this still needed after we upgraded patchelf in https://github.com/ROCm/TheRock/pull/4568 to fix https://github.com/ROCm/TheRock/issues/4561 ? The symptoms look similar.

lucbruni-amd:
Just tested the reproducer in https://github.com/ROCm/TheRock/issues/4271 again with the latest nightly `7.13.0a20260417` which should include the upgraded patchelf, and I no longer hit the issue. `clang-offload-bundler` is invoked during link and the build completes with `[100%] Built target hip_hello_world`, so `execve()` now succeeds on kernel 4.18. Apologies for the noise.

@pbhandar-amd, confirm on your side that this is fixed and we can close both the issue and this PR.

lucbruni-amd:
Closing as this issue has been resolved by https://github.com/ROCm/TheRock/pull/4568.

Issue #3976, opened 2026-03-13
Root out all dependencies on hardcoded system paths like /opt/rocm


ScottTodd:
## Background

Historically, the bulk of ROCm installs were performed system-wide on Linux to a fixed `/opt/rocm/` path. Today, we support installing ROCm on Linux and Windows to a variety of paths via native packages, python packages, build directories, tarballs, and more. However, source code for many subprojects still heavily depends on `/opt/rocm/`. Some uses appear relatively safe on the surface, like as a fallback when no explicit path is chosen, but the pervasiveness of these patterns makes it difficult to spot which uses are truly dangerous.

### Why now?

Now that we've
1. Organized ROCm project sources into superrepos
2. Completed multiple Linux+Windows ROCm releases using TheRock with new packaging methods
3. Made progress turning down "legacy CI" (internal Jenkins etc.) pipelines in favor of open source CI/CD pipelines using GitHub Actions and based on TheRock

I think we're just about in the clear and can pay down this tech debt.

Note that ecosystem projects are harder to update, so we should provide solid guidance for them to follow, but that should be mostly orthogonal to fixing issues in ROCm itself. If an ecosystem project depends on `/opt/rocm/` it may not work with the new package structure / Windows / etc. regardless of whether ROCm itself still has hardcoded paths.

### Related issues / PRs

* https://github.com/ROCm/TheRock/issues/340
* https://github.com/ROCm/TheRock/issues/670
* https://github.com/ROCm/TheRock/issues/683
* https://github.com/ROCm/TheRock/issues/1200
* https://github.com/ROCm/TheRock/issues/1402
  * https://github.com/ROCm/llvm-project/pull/289
  * https://github.com/ROCm/rocm-systems/pull/3150
* https://github.com/ROCm/TheRock/issues/3812
  * https://github.com/ROCm/rocm-systems/pull/3880
* https://github.com/ROCm/TheRock/pull/3825
* https://github.com/ROCm/TheRock/issues/3994
* https://github.com/ROCm/TheRock/issues/4063

## Scope and auditing

I ran an audit rocm-libraries, rocm-systems, and llvm-project here: https://github.com/ScottTodd/claude-rocm-workspace/tree/main/reports/path-audit . That flagged ~371 issues flagged in rocm-libraries, ~490 in rocm-systems, and ~28 in llvm-project.

### Example classes of issues

* Default CMake install paths set to `/opt/rocm` (sometimes with `FORCE`)
[...]

PR #4663, opened 2026-04-17
Fix wrong hip compiler in subproject

macurtis-amd:
## Motivation

If `CMAKE_HIP_COMPILER` is not set, sub-projects will incorrectly find `/opt/rocm*/lib/llvm/bin/clang++`.

This may causes a build failure if the installed compiler is missing features needed by the sub-project.

I was seeing:

```
FAILED: catch_tests/unit/rtc/CMakeFiles/RTC.dir/RtcFunctions.cpp.o
/opt/rocm-7.1.1/lib/llvm/bin/clang++  ...
In file included from .../rocm-systems/projects/hip-tests/catch/unit/rtc/RtcFunctions.cpp:14:
In file included from .../core/clr/dist/include/hip/hip_fp16.h:13:
.../core/clr/dist/include/hip/amd_detail/amd_hip_fp16.h:952:21: error: use of undeclared identifier '__builtin_elementwise_exp10'
  952 |   return __half_raw{__builtin_elementwise_exp10(static_cast<__half_raw>(x).data)};
      |                     ^
```

## Technical Details

Sets HIP-specific variables to the corresponding CXX variable for amd-hip toolchains.

## Submission Checklist

### DIFF
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -1601,6 +1601,8 @@ function(_therock_cmake_subproject_setup_toolchain
     list(APPEND _compiler_toolchain_addl_depends "${_hip_stamp_dir}/stage.stamp")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-path=@_hip_dist_dir@\")\n")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-device-lib-path=@_amd_llvm_device_lib_path@\")\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER \"\${CMAKE_CXX_COMPILER}\")\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_FLAGS_INIT \"\${CMAKE_CXX_FLAGS_INIT}\")\n")
     if(THEROCK_VERBOSE)
       message(STATUS "HIP_DIR = ${_hip_dist_dir}")
     endif()


cmake/therock_subproject.cmake
```diff
@@ -1601,6 +1601,8 @@ function(_therock_cmake_subproject_setup_toolchain
     list(APPEND _compiler_toolchain_addl_depends "${_hip_stamp_dir}/stage.stamp")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-path=@_hip_dist_dir@\")\n")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-device-lib-path=@_amd_llvm_device_lib_path@\")\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER \"\${CMAKE_CXX_COMPILER}\")\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_FLAGS_INIT \"\${CMAKE_CXX_FLAGS_INIT}\")\n")
```
ScottTodd, line 1605: > If `CMAKE_HIP_COMPILER` is not set, sub-projects will incorrectly find `/opt/rocm*/lib/llvm/bin/clang++`.

Including these options here in TheRock is fine, but I suspect we should also fix this somehow in the rocm projects themselves. See https://github.com/ROCm/TheRock/issues/3976 for a description of the broader problem. Discovering a (likely different/mismatched) tool from somewhere on the system other than the current install could be more risky than just raising an exception and requiring setting it explicitly (like what this PR does).

Do you know where this `CMAKE_HIP_COMPILER` is falling back to `/opt/rocm/` if unset?

ScottTodd (approved):

PR #4674, opened 2026-04-20
[cmake] Handle sanitizers for autotools projects

lumachad:
Add ASAN linker flags to CMAKE_EXE_LINKER_FLAGS and CMAKE_SHARED_LINKER_FLAGS

Autotools-based subprojects like rocgdb pass CMAKE_*_LINKER_FLAGS to their configure scripts but don't respect add_link_options(). Add -fsanitize=address to both CMAKE_EXE_LINKER_FLAGS and CMAKE_SHARED_LINKER_FLAGS, with -shared-libsan conditionally for Clang compilers, ensuring ASAN works with autotools builds.

While at it, refactor things so we have less code duplication.

PR #4676, opened 2026-04-20
[debug-tools] Refactor debug tools to handle ASAN/HOST_ASAN/TSAN

lumachad:
This series of 5 commits refactors the debug-tools cmake recipes (mostly rocgdb's) to handle ASAN build properly.

[[debug-tools] Refactor: Use THEROCK_PRIVATE_INSTALL_RPATH_DIRS for rocgdb RPATH patching](https://github.com/ROCm/TheRock/commit/d6aa3702b9cc82cc9d5e9fe3829235a81e843f99) 

Replace hardcoded RPATH segments with dynamic construction from
THEROCK_PRIVATE_INSTALL_RPATH_DIRS variable. Converts paths from
<root>-relative to <root>/bin-relative format for executables.
Falls back to hardcoded lib/rocm_sysdeps/lib for backward compatibility.

[[debug-tools] Refactor: Replace regex-based prefix search with find_package() for rocgdb](https://github.com/ROCm/TheRock/commit/689bd05b4408327670c9e1498617fdb0bd387214) 

Replace fragile fetch_library_prefix() regex pattern matching with
CMake-native find_library_prefix() that uses find_package(). This
properly integrates with TheRock's dependency provider system and
extracts prefixes from imported targets instead of string matching.

Benefits:
- Uses standard CMake primitives (find_package, imported targets)
- Respects CMAKE_PREFIX_PATH and dependency provider
- More robust than regex patterns like "/gmp/build/stage"
- Better integration with generated CMake config files

[[debug-tools] Refactor code to handle ASAN flags](https://github.com/ROCm/TheRock/commit/cf9d19eb2b7c114cf24d9fac03f26f46ad7401fa) 

Right now ROCgdb doesn't handle ASAN flags being set by TheRock. Fix this
so we build using the sanitizer flags passed down by TheRock.

[[debug-tools] Remove post_hook_rocr-debug-agent-tests.cmake](https://github.com/ROCm/TheRock/pull/4676/commits/42aafaf2ef8fe793518f9352cc8dce2ada5f1b1a) 
[42aafaf](https://github.com/ROCm/TheRock/pull/4676/commits/42aafaf2ef8fe793518f9352cc8dce2ada5f1b1a)
Remove the post hook file that is no longer needed.

[Handle sanitizers for autotools projects](https://github.com/ROCm/TheRock/pull/4676/changes/d6df8343e22306537b8f42a77109b2032dbdf4d7)
Autotools-based subprojects like rocgdb pass CMAKE_*_LINKER_FLAGS to
their configure script's LDFLAGS but don't respect add_link_options(),
so sanitizer link flags emitted by therock_sanitizer_configure() never
reach the link line.

Fix this in a rocgdb-local pre_hook that detects THEROCK_SANITIZER and
appends -fsanitize=<address|thread> -shared-libsan to
CMAKE_EXE_LINKER_FLAGS and CMAKE_SHARED_LINKER_FLAGS. Handles ASAN,
HOST_ASAN, and TSAN.

Keeping this in the rocgdb pre_hook (rather than in the generic
sanitizer module) avoids coupling cmake/therock_sanitizers.cmake to a
hardcoded list of gfortran-using subprojects that would silently break
the -shared-libsan flag for math libraries (hipBLASLt, rocBLAS, etc.).


cmake/therock_sanitizers.cmake
```diff
@@ -31,55 +31,51 @@ function(therock_sanitizer_configure
   # Our own toolchains get ASAN enabled consistently.
   # ASAN: Full host+device address sanitizer (xnack+ GPU targets for gfx942, gfx950)
   # HOST_ASAN: Host-only address sanitizer (no device-side instrumentation)
+  # TSAN: Thread sanitizer.
   set(_stanza)
-  if(_sanitizer STREQUAL "ASAN" OR _sanitizer STREQUAL "HOST_ASAN")
+  if(_sanitizer STREQUAL "ASAN" OR _sanitizer STREQUAL "HOST_ASAN" OR _sanitizer STREQUAL "TSAN")
```
geomin12, line 36: should we just try "INCLUDES `ASAN`" ?

geomin12 (commented):
gonna let other folks review further

cmake/therock_sanitizers.cmake
lumachad: I didn't quite get that suggestion. In what context would we use that? Also, this particular patch has its own PR, as stated in the description: https://github.com/ROCm/TheRock/pull/4674

geomin12: actually, sorry I meant to say `includes SAN`, but being explicit is good here

geomin12 (commented):
lgtm but want to let build folks review

lumachad:
This one is unblocked by #4674.

lumachad:
Updated to split out a commit for the removal of the rocr-debug-agent post hook file, which was incorrectly put into another commit.

lumachad:
@marbre Quite a few Windows failures here. But other than that CI looks clean.

lumachad:
Back to a draft as I attempt to merge the more generic TheRock sanitizer changes (https://github.com/ROCm/TheRock/pull/5182) into this one.

geomin12 (commented):
lgtm, but i would recommend doing a manual run of https://github.com/ROCm/TheRock/actions/workflows/ci_tsan.yml and https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci_asan.yml to make this works as expected

lumachad:
> lgtm, but i would recommend doing a manual run of https://github.com/ROCm/TheRock/actions/workflows/ci_tsan.yml and https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci_asan.yml to make this works as expected

Thanks for going through it. I did trigger it here: https://github.com/ROCm/TheRock/actions/runs/25716666037/job/75572999094. It seems OK, apart from packaging failures that seem unrelated.

I'm not too happy with how we have to side-step the hardcoded use of gfortran here. But hopefully the check can be dropped when we switch to flang.

I'd like to have more pairs of eyes looking at the last commit, just to make sure it makes sense.

lumachad:
Just a note. We need to have the debug-tools sanitizer changes alongside the more general change to cmake/therock_sanitizers.cmake. Otherwise, if we only have TheRock's sanitizer change, the debug-tools will run into an error when doing an ASAN build.

lumachad:
Latest ASAN multi-arch run looks clean: https://github.com/ROCm/TheRock/actions/runs/26217092718

geomin12 (approved):
lgtm! seems it passes (i'm working on TSAN multi-arch conversion soon)

lumachad:
Looks like the ASAN CI run was clean-ish: https://github.com/ROCm/TheRock/actions/runs/26666971808. I've been told the sanity checks are expected to fail.

PR #4688, opened 2026-04-20
[DO-NOT-MERGE] Test new core42 gfx942 runners from ossci

amd-justchen:
Using this PR to allow evaluation of new Core42 gfx942 runners.


amd-justchen:
We no longer use this CSP

PR #4695, opened 2026-04-20
Fix UTF-8 when reading GitHub Actions event JSON on Windows

tony-davis:
## Summary

On Windows, `open()` in text mode uses the process default encoding (often cp1252). GitHub writes the workflow event JSON at `GITHUB_EVENT_PATH` as **UTF-8**. If the payload contains UTF-8 text (e.g. curly quotes in a PR body), `json.load()` can fail with `UnicodeDecodeError` when the file is decoded as cp1252. Linux often avoids this because the default encoding is UTF-8.

## Change

- Add `gha_load_github_event()` in `build_tools/github_actions/github_actions_api.py`: reads the event file from the `GITHUB_EVENT_PATH` environment variable only, always with `encoding="utf-8"`.
- Use it from `build_tools/github_actions/configure_multi_arch_ci.py` (`CIInputs.from_environ`) and from `build_tools/_therock_utils/s3_buckets.py` (`_is_current_run_pr_from_fork`), using a **deferred import** in `s3_buckets` so `github_actions` stays an optional dependency for that package.
- Add unit tests in `build_tools/github_actions/tests/github_actions_api_test.py` (UTF-8 payloads, `GITHUB_EVENT_PATH`).

No change to event JSON **semantics**—only how the file is decoded.

## Related

- Ruff **PLW1514** (explicit text encoding) + pre-commit, split per review discussion: [#4907](https://github.com/ROCm/TheRock/pull/4907)


davidd-amd:
Turns out `ruff` will verify this for us. Add `extend-select = ["PLW1514"] under [tool.ruff.lint]` in pyproject.toml. Run `ruff check --select PLW1514 --fix .` to bulk-fix existing offenders. Verify via pre-commit + CI.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -170,7 +170,7 @@ def from_environ() -> "CIInputs":
 
         # Read the full event webhook payload (common to all event triggers).
         event_path = os.environ["GITHUB_EVENT_PATH"]
-        with open(event_path) as f:
+        with open(event_path, encoding="utf-8") as f:
             event = json.load(f)
```
ScottTodd: > If the PR payload contains UTF-8 text (e.g. curly quotes in the PR body), `json.load()` can fail with `UnicodeDecodeError` when the file is decoded as cp1252.

Please add a unit test for this to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/configure_multi_arch_ci_test.py

davidd-amd:
Is it worth creating a function like so the opens don't get out of sync:

```python
import json
import os
from pathlib import Path
from typing import Any

def load_event(path: str | os.PathLike[str] | None = None) -> dict[str, Any]:
    if path is None:
        path = os.environ["GITHUB_EVENT_PATH"]
    with open(path, encoding="utf-8") as f:
        return json.load(f)
```

ScottTodd:
Oh, sure. Could go in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py (then that could get a unit test in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/github_actions_api_test.py and the other modules could just trust it to work)

tony-davis:
> Oh, sure. Could go in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py (then that could get a unit test in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/github_actions_api_test.py and the other modules could just trust it to work)

I can make that change.

davidd-amd:
> Turns out `ruff` will verify this for us. Add `extend-select = ["PLW1514"] under [tool.ruff.lint]` in pyproject.toml. Run `ruff check --select PLW1514 --fix .` to bulk-fix existing offenders. Verify via pre-commit + CI.

This might be a good idea too (*.pre-commit-config.yaml*)

```
- repo: https://github.com/astral-sh/ruff-pre-commit                                                                     
  rev: v0.7.0                                                                                                            
  hooks:                                                                                                                 
    - id: ruff                                                                                                           
      args: [--select, PLW1514]   
 ```

File: build_tools/github_actions/configure_multi_arch_ci.py (hunk 1 above)
Comment (tony-davis) [reply build_tools/github_actions/configure_multi_arch_ci.py]: I centralized the function and made unit tests too. See if that's better.

**Comment by tony-davis** [comment]
I'm going to make a separate PR for the pre-commit changes, just to keep that logically separate. Does that make sense?

**Comment by ScottTodd** [comment]
> I'm going to make a separate PR for the pre-commit changes, just to keep that logically separate. Does that make sense?

Sure, thank you.

File: build_tools/_therock_utils/s3_buckets.py [hunk 2]
```diff
@@ -158,8 +157,11 @@ def _is_current_run_pr_from_fork() -> bool:
     if not event_path:
         return False
```
ScottTodd: We might as well remove this branch and let the helper function read the path from the env var. One less detail for this file to need to be aware of.

build_tools/github_actions/github_actions_api.py
```diff
@@ -379,6 +379,27 @@ def gha_append_step_summary(summary: str):
...
+    path: str | os.PathLike[str] | None = None,
+) -> dict[str, Any]:
+    """Load the GitHub Actions workflow event JSON from disk.
+
+    GitHub writes ``GITHUB_EVENT_PATH`` as UTF-8. On Windows the process
+    default encoding is often not UTF-8, so the file must be opened with
+    ``encoding="utf-8"``.
+
+    Args:
+        path: Event file path. If None, uses :envvar:`GITHUB_EVENT_PATH`.
```
ScottTodd: Related to my other comment, this could just take no arguments and always read from GITHUB_EVENT_PATH. I can't think of a case (outside of unit tests) where a caller of this function would want to provide a different path.

ScottTodd (approved):
This can be simplified a bit. With these comments addressed LGTM

build_tools/github_actions/github_actions_api.py
```diff
@@ -379,23 +379,17 @@ def gha_append_step_summary(summary: str):
...
+    the file must be opened with ``encoding="utf-8"``.
 
     Returns:
         Parsed JSON object (GitHub webhook payloads are JSON objects).
     """
-    if path is None:
-        path = os.environ["GITHUB_EVENT_PATH"]
+    path = os.environ["GITHUB_EVENT_PATH"]
     with open(path, encoding="utf-8") as f:
         return json.load(f)
```
ScottTodd, line 394: Should add error handling here.
* Return `None` if the environment variable is not set
* Check the returned data before trying to use it in `_is_current_run_pr_from_fork`
* Drop the 
    ```python
        if not os.environ.get("GITHUB_EVENT_PATH"):
            return False
    ```
    in `build_tools/_therock_utils/s3_buckets.py`

build_tools/_therock_utils/s3_buckets.py
ScottTodd: The commit did not fully resolve this comment.

PR #4706, opened 2026-04-21
[ncurses] Set default terminfo dir to stage tree

### LINKED ISSUE #3511: [issue] Build failure: ncurses installs terminfo to Kitty's read-only directory
Split from #3477 

### Environment

* OS: Ubuntu 25.04
* Compiler: GCC 15 (defaults to -std=gnu23)
* Python: 3.13
* GPU target: gfx1151 (AMD Strix Halo iGPU / Radeon 8060S)
* Terminal: Kitty
* Build method: standard cmake + ninja with .venv

### Instructions to reproduce:

On a fresh install of Ubuntu 25.10:
```
cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES=gfx1151 -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
cmake --build build
```

third-party/sysdeps/common/ncurses/CMakeLists.txt1

Error:
```
/usr/lib/kitty/terminfo: permission denied (errno 13)
Kitty sets $TERMINFO=/usr/lib/kitty/terminfo in the environment. ncurses configure picks this up as its ticdir install destination.
```

Fix: Add `--unset=TERMINFO` to the `cmake -E env` invocation so ncurses falls back to `${CMAKE_INSTALL_PREFIX}/share/terminfo`.

FIM43-Redeye:
## Summary

- ncurses configure defaults `--with-default-terminfo-dir` (ticdir) to `/usr/share/terminfo`. `make install` then attempts to write the terminfo database there, which fails on unprivileged builds on systems without pre-existing write access.
- Additionally, ncurses configure honors `$TERMINFO` from the environment when choosing ticdir, so terminals like Kitty (which set `$TERMINFO=/usr/lib/kitty/terminfo`) cause ticdir to land in a root-owned path.
- Pass `--with-default-terminfo-dir=${CMAKE_INSTALL_PREFIX}/share/terminfo` so the database lands in the stage tree regardless of environment.

Fixes #3511.

## Test plan

Verified with ncurses-6.6 configure (the exact tarball TheRock uses) in four scenarios, inspecting the reported "default terminfo directory" and the resulting `ticdir` in `Makefile`:

| Scenario | `$TERMINFO` | `--with-default-terminfo-dir` | ticdir |
|---|---|---|---|
| Baseline, unpatched | unset | unset | `/usr/share/terminfo` (would fail unprivileged) |
| Kitty, unpatched | `/usr/lib/kitty/terminfo` | unset | `/usr/lib/kitty/terminfo` (reproduces #3511) |
| Kitty, patched | `/usr/lib/kitty/terminfo` | stage | stage path (patch overrides Kitty env) |
| Baseline, patched | unset | stage | stage path (patch fixes default case too) |

- [x] Patch redirects ticdir to the stage tree even when `$TERMINFO` points at a read-only system path (fixes #3511 directly).
- [x] Patch also fixes the broader unprivileged-build failure where `$TERMINFO` is unset but ncurses' default is `/usr/share/terminfo`.
- [x] Without the patch, `$TERMINFO=/usr/lib/kitty/terminfo` → ticdir becomes Kitty's path (trigger confirmed).

### DIFF
--- a/third-party/sysdeps/common/ncurses/CMakeLists.txt
+++ b/third-party/sysdeps/common/ncurses/CMakeLists.txt
@@ -150,6 +150,9 @@ add_custom_target(
       # by pkgconfig (which is typically the system root, which is not what
       # we want).
       --with-pkg-config-libdir="${CMAKE_INSTALL_PREFIX}/lib/pkgconfig"
+      # Direct terminfo database installation into the stage tree so
+      # make install does not attempt to write to /usr/share/terminfo.
+      --with-default-terminfo-dir="${CMAKE_INSTALL_PREFIX}/share/terminfo"
       # Prevent build-time LDFLAGS (rpath, version-script) from leaking into
       # the .pc files' Libs: line.
       --disable-pkg-ldflags


lumachad:
I'm not quite sure --with-default-terminfo-dir is the right option here. Shouldn't ncurses always be installing to whatever prefix it was told to? Is this a characteristic of ncurses where it does something different if TERMINFO is set?

lumachad:
Interesting. I gave --with-default-terminfo-dir a bogus directory and it does install things to that directory. The option doesn't really make it clear that it does that. So I think this fix is appropriate to address issues around non-conventional terminals.

lumachad:
Let me do a few more checks to make sure our terminfo alternate dirs are still being looked at.

lumachad (approved):
Confirmed OK on local testing. OK if CI is good.

FIM43-Redeye:
Looks like Git infra and gfx94X tests flaking, good to re-run the failed checks?

lumachad:
> Looks like Git infra and gfx94X tests flaking, good to re-run the failed checks?

Re-running.

lumachad:
@marbre I'm happy with this version and CI looks good, despite unrelated failures.

FIM43-Redeye:
CI still seems green, @lumachad does this seem ready to land to you? Pinging since you've owned ncurses changes in merges before.

lumachad:
> CI still seems green, @lumachad does this seem ready to land to you? Pinging since you've owned ncurses changes in merges before.

It's ready to merge from my side, but I'd like @marbre to eyeball this first.

marbre (approved):

marbre:
Let's wait for a fresh CI signal including the latest changes before landing this.

lumachad:
Re-triggering as I saw packaging failures for rpm and deb. I want to make sure those are not related.

lumachad:
@marbre The packaging failures seem to be stable. I can't quite tell if they are a concern though.

lumachad:
Confirmed the packaging failures are unrelated. @FIM43-Redeye Thanks for the PR.

Issue #4527, opened 2026-04-14
[Issue] Pytorch Test failure: `test_graph_make_graphed_callables*` aborts with `hipErrorStreamCaptureUnsupported`


chiranjeevipattigidi:
### Description:
The test runner process aborts (exit code 134) due to a thrown c10::AcceleratorError coming from HIP Graph / CUDA Graph stream capture (“operation not permitted when stream is capturing”; hint hipErrorStreamCaptureUnsupported). 

### Context:
- **Platform:** `Linux`
- **Effected Archs:** `gfx950, gfx94x`
- **Pytorch Versions:** `nightly`
- **Python Versions:** `3.13`

### Full Logs:
- https://github.com/ROCm/TheRock/actions/runs/24385542651/job/71228141198
- https://github.com/ROCm/TheRock/actions/runs/24384283164/job/71218346738

### Error Log:
```
external-builds/pytorch/pytorch/test/test_cuda.py::TestCuda::test_graph_make_graphed_callables_parameterless_nograd_module_without_amp_not_allow_unused_input PASSED [0.0115s] [  9%]
terminate called after throwing an instance of 'c10::AcceleratorError'
  what():  CUDA error: operation not permitted when stream is capturing
Search for `hipErrorStreamCaptureUnsupported' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.
Exception raised from ~CUDAGraph at /__w/TheRock/TheRock/external-builds/pytorch/pytorch/aten/src/ATen/hip/HIPGraph.cpp:367 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x9d (0x7a0d1bed8e2d in /__w/TheRock/TheRock/.venv/lib/python3.13/site-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0x51fcd (0x7a0ad52e1fcd in /__w/TheRock/TheRock/.venv/lib/python3.13/site-packages/torch/lib/libc10_hip.so)
[...]

PR #4709, opened 2026-04-21
[ci] Move test_graph_make_graphed_callables* Pytorch tests to skip_tests/generic.py

lucbruni-amd:
## Motivation

Addresses https://github.com/ROCm/TheRock/issues/4527.

These tests were already skipped for `2.9` and `2.10`, but not elsewhere. Adding them first to `generic.py` to unblock CI.

## Technical Details

See root cause analysis of the test failures in [this comment](https://github.com/ROCm/TheRock/issues/4527#issuecomment-4290763936).

Moving both test names into `generic.py -> common -> cuda` (with the verbatim `hipErrorStreamCaptureUnsupported` comment block retained from the original `2.10` entry) makes the skip apply uniformly across every `pytorch_<ver>.py` (present or not) and every AMDGPU family. The previously-duplicated entries in `pytorch_2.9.py` and `pytorch_2.10.py` are removed so there is a single source of truth.

## Test Plan

Simulate the skip-list loader for every known PyTorch minor version and AMDGPU arches listed in the linked issue, and verify both target test names appear in the computed skip expression.

To observe CI on the upstream-nightly matrix slot.

## Test Result

All combinations reported these tests as freshly skipped, including for `pytorch_version=2.13` where the per-version file (would be `skip_tests/pytorch_2.13.py`, which is warned as missing currently, see https://github.com/ROCm/TheRock/actions/runs/24708625512/job/72280026775) does not exist and only `generic.py` contributes.

## Submission Checklist

### DIFF
--- a/external-builds/pytorch/skip_tests/generic.py
+++ b/external-builds/pytorch/skip_tests/generic.py
@@ -58,6 +58,16 @@ skip_tests = {
             # AttributeError: 'Model' object has no attribute '__annotations__'
             # https://github.com/ROCm/TheRock/issues/2985
             "test_autocast_cat_jit",
+            # what():  HIP error: operation not permitted when stream is capturing
+            # Search for `hipErrorStreamCaptureUnsupported' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__HIPRT__TYPES.html for more information.
+            # HIP kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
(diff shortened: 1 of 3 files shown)


HereThereBeDragons (approved):
if we only see it on gfx94x and gfx950 we might want to put it in the separate section in generic.py (like the very top section in generic.py, for gfx950. for gfx94x use `"gfx94"`)

but i am also fine with skipping this on every gpu

lucbruni-amd:
Merging with approval for tonight's nightly. Failed checks are unrelated.

PR #4710, opened 2026-04-21
[ml-libs][ck] Add missing gfx targets to CK's supported list

evedovelli:
## Motivation

Libraries and projects depending on CK cannot build for gfx110X  when upgrading from ROCm 7.2 because CK isn't include for gfx110X target.

## Technical Details

Include gfx110X targets to the list of CK's supported projects.

## Test Plan

Build with CI and verify CK is properly included for GPU Architecture gfx110X

## Submission Checklist

### DIFF
--- a/ml-libs/CMakeLists.txt
+++ b/ml-libs/CMakeLists.txt
@@ -10,7 +10,7 @@ endif()
 
 # limit use of composable kernel to the GPU families it is valid for
 if(THEROCK_ENABLE_COMPOSABLE_KERNEL)
-  set(_ck_supported_gfx_targets gfx908 gfx90a gfx942 gfx950 gfx1150 gfx1151 gfx1152 gfx1153 gfx1200 gfx1201)
+  set(_ck_supported_gfx_targets gfx908 gfx90a gfx942 gfx950 gfx110X gfx1100 gfx1101 gfx1102 gfx1103 gfx1150 gfx1151 gfx1152 gfx1153 gfx120X gfx1200 gfx1201)
   foreach(_gfx_target ${THEROCK_AMDGPU_TARGETS})
     if(NOT ${_gfx_target} IN_LIST _ck_supported_gfx_targets)
       message(WARNING


geomin12 (commented):
this looks fine, but will wait to see how compile time takes! previously, the issue was compile time doubled after adding this

evedovelli:
> this looks fine, but will wait to see how compile time takes! previously, the issue was compile time doubled after adding this

That's what @brockhargreaves-amd told me as well. @geomin12 can you monitor the build time difference and inform it back here when the CI is done? Hopefully with the build time improvements, it's fast enough now that we can include these plaftorms.

brockhargreaves-amd:
We need to tread lightly here. These were added, and then reverted due to an unacceptable increase in build times: https://github.com/ROCm/TheRock/pull/4022

CK has been putting significant effort into reducing their build times, largely orchestrated by @cgmillette: 

https://github.com/ROCm/rocm-libraries/pull/4447
https://github.com/ROCm/rocm-libraries/pull/4673
https://github.com/ROCm/rocm-libraries/pull/5041
https://github.com/ROCm/rocm-libraries/pull/5031
https://github.com/ROCm/rocm-libraries/pull/4355
https://github.com/ROCm/rocm-libraries/pull/5028
https://github.com/ROCm/rocm-libraries/pull/5938
https://github.com/ROCm/rocm-libraries/pull/5939

There is also ongoing work that hasn't landed yet:

https://github.com/ROCm/rocm-libraries/pull/5945 

We need to be sure we've reached an acceptable threshold to have this land.

evedovelli:
The build time for gfx110X went from 2h~3h to >5h with this change, so I'm going to abandon it.

ScottTodd:
> The build time for gfx110X went from 2h~3h to >5h with this change, so I'm going to abandon it.

With multi-arch CI/CD landing now, we may want to split `gfx110X` out to `[gfx1100, gfx1101, gfx1102, gfx1103]` (and also switch to use LLVM generic targets where possible)

brockhargreaves-amd:
> > The build time for gfx110X went from 2h~3h to >5h with this change, so I'm going to abandon it.
> 
> With multi-arch CI/CD landing now, we may want to split `gfx110X` out to `[gfx1100, gfx1101, gfx1102, gfx1103]` (and also switch to use LLVM generic targets where possible)

Hi @ScottTodd, could you expand on this? Admittedly I haven't quite understood the implementation details of the multi-arch rollout (if you'd rather point me towards some examples): Where should we be using specific targets and where should we be using LLVM generic targets?

ScottTodd:
LLVM generic targets are tangential to multi-arch. We've been wanting to use them for these targets that are very similar. That's part blocked by math library support for targets and part blocked by build system / CI / CD configuration to roll it out.

Non-multi-arch CI used to build the entire stack (LLVM, rocm-systems, rocm-libraries, etc.) separately for each GPU family. Related (and compatible) targets were grouped into families so their packages could use a common index and we didn't need to go too wide on build jobs to support additional targets.

Multi-arch CI builds target-neutral subprojects (LLVM, rocm-systems, etc.) once, then builds target-specific subprojects (rocblas, hipblaslt, etc.) for each GPU _family_, but we could also now efficiently build for each GPU _target_
Here's a dev release workflow run where you can see the current fan-out: https://github.com/ROCm/TheRock/actions/runs/24540020624
image

it would be pretty natural to expand 110x-all, 120x-all, 101x-dgpu, and 103x-dgpu

PR #4714, opened 2026-04-21
[ci] Removing alternative weight runner label logic

geomin12:
Reverting #4381 as we can use same label now across CSPs (worked with OSSCI)


amd-justchen (approved):
That makes it easier!

PR #4718, opened 2026-04-21
[ci] adding three test runner label solution for `gfx94X`

geomin12:
Revert of #4714 , the same label for clusters did not work :( 

We now have to do weight distribution accordingly for single and multi gpu. With `fetch_package_targets.py`, this will also work for superrepos.

Based on count of GPU, I have added the weight distribution!


amd-justchen (approved):
LGTM!

PR #4736, opened 2026-04-22
CI: install Python deps with uv --no-cache to avoid corrupted wheel cache

amd-chiranjeevi:
## Motivation
The Multi-Arch CI job failed during dependency install with:
- `Failed to install: requests-2.33.0...`
- `The wheel is invalid: Metadata field Name not found`
This error is consistent with a corrupted/partial wheel being reused from cache. Passing --no-cache forces uv to install fresh artifacts.

## Test Plan

CI checks running.

## Test Result

TBD

## Submission Checklist

### DIFF
--- a/.github/actions/setup_test_environment/action.yml
+++ b/.github/actions/setup_test_environment/action.yml
@@ -56,7 +56,7 @@ runs:
     - name: Install test requirements
       shell: bash
       run: |
-        uv pip install -r requirements-test.txt
+        uv pip install -r requirements-test.txt --no-cache
         uv pip freeze
 
     - name: Install Choco tools


.github/actions/setup_test_environment/action.yml
```diff
@@ -56,7 +56,7 @@ runs:
     - name: Install test requirements
       shell: bash
       run: |
-        uv pip install -r requirements-test.txt
+        uv pip install -r requirements-test.txt --no-cache
```
ScottTodd, line 59: Please don't turn off caching. Test runner resources are already scarce enough. Need to fix the runners, add cleanup scripts, etc.

PR #4748, opened 2026-04-22
[WIP] [rocprofiler-systems] Build examples as standalone and switch to CTests

kcossett-amd:
## Motivation

We are moving to using CTest as the test executor instead of pytest. 
We also need to build examples as a standalone to ensure it can be architecture specific (this will dratically reduce the build time)

## Technical Details

WIP

## Test Plan

Workflows

## Submission Checklist


kcossett-amd:
And I've broken the branch...

PR #4749, opened 2026-04-22
[profiler] Declare rocprof-trace-decoder as a provided package

ApoKalipse-V:
rocprofiler-sdk's cmake/rocprofiler_config_interfaces.cmake calls
  find_package(rocprof-trace-decoder REQUIRED CONFIG). Without a
  matching therock_cmake_subproject_provide_package declaration,
  THEROCK_PACKAGE_DIR_rocprof-trace-decoder is never set and the
  dep_provider falls through to a system search (line 73), which fails
  in CI:

    Could not find a package configuration file provided by
    "rocprof-trace-decoder" with any of the following names:
        rocprof-trace-decoderConfig.cmake
        rocprof-trace-decoder-config.cmake

  The relpath lib/cmake/rocprof-trace-decoder matches what the
  trace-decoder project installs (TTD_CONFIG_INSTALL_DIR =
  ${CMAKE_INSTALL_LIBDIR}/cmake/rocprof-trace-decoder, with
  CMAKE_INSTALL_LIBDIR forced to "lib"). Same shape as the existing
  declarations for rocdecode, rocjpeg, rdc, etc.

## Submission Checklist

### DIFF
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -11,7 +11,6 @@ if(THEROCK_ENABLE_ROCPROFV3)
   # rocprof-trace-decoder
   # The trace decoder is responsible for decoding advanced thread traces from the
   # GPU and is used at runtime by `rocprofv3 --att`.
-  # See: https://github.com/ROCm/rocprof-trace-decoder/
   ##############################################################################
   therock_cmake_subproject_declare(rocprof-trace-decoder
     # The build infrastructure requires generic stages to pass through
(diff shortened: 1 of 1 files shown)


profiler/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ if(THEROCK_ENABLE_ROCPROFV3)
     BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/rocprof-trace-decoder"
     BACKGROUND_BUILD
   )
+  therock_cmake_subproject_provide_package(rocprof-trace-decoder rocprof-trace-decoder lib/cmake/rocprof-trace-decoder)
```
jbonnell-amd, line 23: While we're adding packages, it might make sense to add some shared lib validation tests for rocprof-trace-decoder shared libraries as well.

You can see rocprofiler-sdk's implementation for reference:
https://github.com/ROCm/TheRock/blob/main/profiler/CMakeLists.txt#L137-L149

jbonnell-amd (approved):
Looks good, assuming that CI passes with the new `therock_test_validate_shared_lib` test

venkat1361 (approved):

PR #4753, opened 2026-04-22
Proposal to include system files in TheRock

nunnikri:
Proposal to include system files in TheRock

### DIFF
--- a/docs/rfcs/RFC0009-OS-Packaging-Requirements.md
+++ b/docs/rfcs/RFC0009-OS-Packaging-Requirements.md
@@ -73,6 +73,44 @@ The soft links allow for an independent directory structure for ROCm expansions,
 /opt/rocm/hpc/ -> /opt/rocm/hpc-26.2.0
 ```
 
+### System Configuration Files (`sys` component)
+
+Some ROCm packages install system-level configuration files that must reside in OS-managed directories outside the ROCm install prefix (e.g. `/etc/ld.so.conf.d/`, `/etc/OpenCL/vendors/`). These are captured in the `sys` artifact component type (see [artifact component types](../development/artifacts.md#component-types)).
+
+Examples:
+
+```
+/etc/ld.so.conf.d/10-rocm7.12-opencl.conf
+/etc/OpenCL/vendors/amdocl64_70200_43.icd
+```
+
+#### ROCm Path References in Config File Contents
+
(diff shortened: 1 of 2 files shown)


docs/rfcs/RFC0009-OS-Packaging-Requirements.md
```diff
@@ -73,6 +73,45 @@ The soft links allow for an independent directory structure for ROCm expansions,
...
+```
+
+#### Handling During Packaging and Installation
+
+When a `sys` component is present, the packaging and install tooling must **copy** these files to the corresponding system directory. Symlinks must not be used because:
+
+- Package managers treat `/etc` files as *conffiles* — preserving user edits on upgrade and removing them cleanly on uninstall. This mechanism is defined for regular files; behavior with symlinks is implementation-dependent and unreliable.
+- `ldconfig` silently ignores broken symlinks in `ld.so.conf.d/`, leaving ROCm library paths absent from `/etc/ld.so.cache` with no error — producing "library not found" failures that are difficult to diagnose.
+- Symlinks are invalidated by any operation that removes or relocates the target: ROCm upgrades, partial uninstalls, or manual cleanup. The broken entry persists in `/etc`, invisible to the package manager.
+- On SELinux-enforcing systems (RHEL and derivatives), symlinks in `/etc` whose targets fall outside the expected security context may trigger policy denials.
```
saadrahim, line 111: Should we revisit the use of a soft link? It seems like using direct paths to the rocm installation may be better.

docs/rfcs/RFC0009-OS-Packaging-Requirements.md
```diff
@@ -73,6 +73,45 @@ The soft links allow for an independent directory structure for ROCm expansions,
...
+/etc/ld.so.conf.d/10-rocm7.12-43-opencl.conf
+/etc/OpenCL/vendors/amdocl64_70200_43.icd
+/usr/lib/systemd/system/amdcuid_daemon.service
+```
+
+#### ROCm Path References in Config File Contents
+
+Any ROCm path referenced inside a `sys` config file must point to the **unversioned symlink** `/opt/rocm/core` rather than a versioned path (e.g. `/opt/rocm/core-7.12`). This ensures the config remains valid across patch updates and supports side-by-side multi-version installs, where the active version is controlled by the symlink.
+
+Example — correct content for `/etc/ld.so.conf.d/10-rocm7.12-43-opencl.conf`:
```
saadrahim, line 92:
```suggestion
Example — correct content for `/etc/ld.so.conf.d/10-rocm-43-opencl.conf`:
```
Should we only allow one opencl.conf file?

saadrahim (commented):
Analysis of ROCm whether we have a soft link or direct link from opencl.conf is needed.

PR #4754, opened 2026-04-22
rocrtst: Added dependency yaml-cpp by rocrtst

cfreeamd:
## Motivation

rocrtst has been updated to use process a YAML file and now requires yaml-cpp to parse it.

## Technical Details

Added a dependency by  _rocrtst_build_deps in core/CMakeLists.txt on therock-yaml-cpp.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -478,7 +478,7 @@ if(THEROCK_BUILD_TESTING)
     set(_rocrtst_build_deps
       amd-llvm
       ocl-clr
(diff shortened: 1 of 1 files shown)


marbre (approved):
If i remember correctly yaml-cpp is statically linked and thus indeed a build time dep. Otherwise (if needed at build and runtime) it would be runtime dep, fyi. LGTM.

marbre:
Btw, please try to stick to branch name rules in the future.

PR #4758, opened 2026-04-23
[WIP] [rocprofiler-systems] Build examples as standalone and have CTest as the test executor

kcossett-amd:
## Motivation

We are moving to using CTest as the test executor instead of pytest. 
We also need to build examples as a standalone to ensure it can be architecture specific (this will dratically reduce the build time)

## Technical Details

WIP

TODO:
 - [x] Follow HIPDNN integration test example for `rocprofiler-systems-examples`
 - [ ] Remove submodule update once PR in rocm-systems is merged and bumped to here
 - [ ] Ensure CTests work

## Test Plan

Workflows

## Submission Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,7 +21,7 @@
 [submodule "rocm-systems"]
 	path = rocm-systems
 	url = https://github.com/ROCm/rocm-systems.git
-	branch = develop
+	branch = users/kcossett-amd/ctests-for-pytests
 [submodule "spirv-llvm-translator"]
 	path = compiler/spirv-llvm-translator
 	url = https://github.com/ROCm/SPIRV-LLVM-Translator.git
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -724,6 +724,13 @@ feature_name = "ROCPROFSYS"
 feature_group = "PROFILER"
 disable_platforms = ["windows"]
 
+[artifacts.rocprofiler-systems-examples]
+artifact_group = "profiler-apps"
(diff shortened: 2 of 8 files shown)


kcossett-amd:
So `ROCPROFSYS_GFX_TARGETS` is working as the `jacobi-usm-sys-run` actually runs (meaning it has offload bundle in it). But transpose is being skipped... as if it itself has no offload bundle...?

kcossett-amd:
Okay... `rocprofiler-systems-examples` is being passed the GPU targets. Some examples are built with the HIP offload bundles for every requested arch, which works fine... 

The problem is downstream in the kpack artifact splitter: it scans every binary for a .hip_fatbin ELF section, extracts the device code into per-arch `.kpack/<artifact>_<arch>.kpack` archives, injects a `.rocm_kpack_ref` marker in the host ELF, and zero-pages the original section. So `llvm-objdump --offloading transpose` reads only the embedded section (now gutted) and reports zero offload bundles, which causes our pytest offload detector to fail.

**EDIT: A fix for this is out of scope for this PR**... I'll just make the examples target neutral :(

(speculative below)
On top of that, install_rocm_from_artifacts.py --amdgpu-family gfx94X-dcgpu only matches artifact names against {"generic", "gfx94X-dcgpu"}, but the per-arch siblings the splitter produced are named _test_gfx942 (individual target), so the .kpack/ companion files are silently dropped at install — meaning even if the static check were to pass, the device code wouldn't be on disk in the first place.

kcossett-amd:
Handled in https://github.com/ROCm/TheRock/pull/4934

PR #4763, opened 2026-04-23
[debug-tools] Enable external source directories for amd-dbgapi and rocr-debug-agent

lumachad:
For convenience of development workflows and CI testing, also enable amd-dbgapi and rocr-debug-agent sources to be picked up from an alternate location.


marbre (approved):
LGTM, assuming you've also verified locally :)

geomin12 (approved):

lumachad:
> LGTM, assuming you've also verified locally :)

I did. In particular the interaction between setting an external rocm-systems AND external dbgapi/debug agent directories.

Issue #4558, opened 2026-04-15
[Issue] Windows benchmark jobs failing in CI Nightly (hipBLASLt/rocSOLVER/rocFFT/rocRAND) - gfx1151


chiranjeevipattigidi:
### Description:
The Windows benchmark stage is failing across multiple library benchmarks.Failures look like GPU runtime / kernel artifact / packaging issues (missing rocBLAS Tensile library; HIP runtime calls failing), not simple benchmark parsing issues (parsing fails only because the benchmark executables error before emitting CSV).

**First Failed Run:** https://github.com/ROCm/TheRock/actions/runs/24273241155
commit history b/w first failed & last success run: https://github.com/ROCm/TheRock/compare/d651c8f488f293badfbf7afe6734211a1348e69d...f61cabb02d4a4cfb4315dc63095ef7799f4138c8

### Context
- Run(15/04) : https://github.com/ROCm/TheRock/actions/runs/24434318178
- **Platform:** `Windows`
- **Effected Archs:** `gfx1151`
- **Release Type:** `nightly`
- **Workflow:** https://github.com/ROCm/TheRock/blob/main/.github/workflows/ci_nightly.yml
### Failing jobs (Windows):
- hipBLASLt benchmark: https://github.com/ROCm/TheRock/actions/runs/24434318178/job/71397285917
- rocSOLVER benchmark: https://github.com/ROCm/TheRock/actions/runs/24434318178/job/71397285928
- rocFFT benchmark: https://github.com/ROCm/TheRock/actions/runs/24434318178/job/71397285936
- rocRAND benchmark: https://github.com/ROCm/TheRock/actions/runs/24434318178/job/71397285937

### Observed Errors:
rocblas_benchmark:
```
rocBLAS error: Cannot read C:\runner\_work\TheRock\TheRock\build\bin\/rocblas/library/TensileLibrary.dat: No such file or directory for GPU arch : gfx1150
```

hipblaslt_benchmark:
```
Repeated:
error: invalid argument (1) at .../hipblaslt/clients/common/include/testing_matmul.hpp:1839
```

rocfft_benchmark:

```
HIP_V_THROWERROR< 1 > (.../rocfft/clients/bench/bench.cpp Line: 315): hipMemGetInfo failed
```

rocrand_benchmark:
```
ROCRAND error: 107 at .../rocrand/benchmark/benchmark_rocrand_host_api.cpp:92
```

astrelsky:
[...]

PR #4764, opened 2026-04-23
Fix GPU_ENABLE_PAL breaking benchmarks on gfx1151 Windows nightly-ci (#3587)

lajagapp:
## Summary

- PR #3593 added `GPU_ENABLE_PAL` to the Test step env block in `test_component.yml`. For non-hip-tests components, the field evaluates to `""`, causing gfx1151 (Strix Halo) Windows benchmarks to use the non-functional ROCR backend instead of PAL.
- Fix: move `GPU_ENABLE_PAL` to a conditional `$GITHUB_ENV` step (same pattern as the existing ASAN env var step) so the env var is only set when the component explicitly provides a value.

### Affected

- hipblaslt_bench: `error: invalid argument (1)`
- rocfft_bench: `hipMemGetInfo failed`
- rocrand_bench: `ROCRAND_STATUS_LAUNCH_FAILURE (107)`
- rocsolver_bench: `hipErrorInvalidKernelFile` / `missing TensileLibrary.dat`

Github-issue: https://github.com/ROCm/TheRock/issues/4558

## Test plan

- Verify gfx1151 Windows benchmarks pass in nightly-CI
- Verify hip-tests PAL/ROCR dual-run still works on Windows

## Test Result
- Without this PR - https://github.com/ROCm/TheRock/actions/runs/24830717391
- With this PR - https://github.com/ROCm/TheRock/actions/runs/24830703374

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -152,6 +152,12 @@ jobs:
           echo "HSA_XNACK=1" >> $GITHUB_ENV
           echo "ASAN_SYMBOLIZER_PATH=${{ env.OUTPUT_ARTIFACTS_DIR }}/llvm/bin/llvm-symbolizer" >> $GITHUB_ENV
 
+      # Set GPU_ENABLE_PAL for Windows hip-tests (PAL=1, ROCR=0) (Issue #3587)
+      - name: Set GPU_ENABLE_PAL for Windows hip-tests
+        if: ${{ fromJSON(inputs.component).gpu_enable_pal != '' }}
+        shell: bash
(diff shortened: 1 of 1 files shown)


geomin12 (commented):
instead of adding a new step, could we just add a ternary value for ` GPU_ENABLE_PAL: ${{ fromJSON(inputs.component).gpu_enable_pal }}` ? or some temporary value that'll fix this? 

not a fan of adding an entire new step for just windows hip tests

ScottTodd:
Do we still need this after https://github.com/ROCm/rocm-systems/pull/5119 ? cc @rakesroy, @gandryey . Seems like the defaults for this flag may warrant further adjustments?

PR #4774, opened 2026-04-23
[CI][Infra] File update for multi-ci on amd-staging

kirthana14m:
validated test link for CI : https://github.com/ROCm/llvm-project/actions/runs/24968388045/job/73113608362?pr=2231


skganesan008:
I would propose the commit title to something like "[CI][Infra] sync multi-ci dependencies from main branch" .
Under description, please mention details on what is synced in case if it is a subset, etc

If there are no code changes at all and it is only a copy of the main branch code, this patch can be merged with the commit title/description change

kirthana14m:
Using - https://github.com/ROCm/TheRock/pull/4897

This PR is no longer used or needed

PR #4778, opened 2026-04-23
[torch] Mirror NumPy 2.2.6 for Python 3.10 dependency sync

ethanwee1:
When trying to pip install torchvision and torchaudio, I run into a pip installation failure because torchvision and torchaudio have a numpy dependency, and expect numpy to be in the same location as where we point the index_url to. https://rocm.nightlies.amd.com/v2-staging/gfx950-dcgpu/numpy/
Numpy built for python 3.10 is missing from this index. 

Following https://github.com/ROCm/TheRock/tree/main/build_tools/third_party/s3_management#adding-a-new-package-dependency


ethanwee1:
Closing as wheels have now been populated.

PR #4786, opened 2026-04-24
Optimize artifact directory lookup with prefix-based indexing

raramakr:
Refactored get_package_list() to improve performance by building a prefix index
   of artifact directories upfront, eliminating redundant directory traversals for
   each package. Changed from nested loop checking every directory entry for each
   artifact to a single pass index creation followed by O(1) lookups.

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -271,13 +271,25 @@ def get_package_list(artifact_dir):
     skipped_list  : list of package names excluded due to missing artifacts
     """
     pkg_list = []
-    skipped = []
+    skipped_list = []
+    artifact_path = Path(artifact_dir)
     data = read_package_json_file()
 
     try:
-        dir_entries = os.listdir(artifact_dir)
+        artifact_dirs = {path.name for path in artifact_path.iterdir() if path.is_dir()}
     except FileNotFoundError:
-        sys.exit(f"{artifact_dir}: Artifactory directory does not exist, Exiting")
+        sys.exit(f"{artifact_dir}: Artifactory directory does not exist, exiting")
+
+    # Create a dictionary
+    prefix_index = {}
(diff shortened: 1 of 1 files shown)


build_tools/packaging/linux/packaging_utils.py
```diff
@@ -271,50 +271,49 @@ def get_package_list(artifact_dir):
...
     try:
-        dir_entries = os.listdir(artifact_dir)
+        artifact_dirs = {path.name for path in artifact_path.iterdir() if path.is_dir()}
     except FileNotFoundError:
-        sys.exit(f"{artifact_dir}: Artifactory directory does not exist, Exiting")
+        sys.exit(f"{artifact_dir}: Artifactory directory does not exist, exiting")
+
+    # Create a dictionary
+    prefix_index = {}
+    SUFFIX_MARKERS = ["_dbg_", "_dev_", "_doc_", "_lib_", "_run_", "_test_"]
```
nunnikri, line 287: What will happen if a new suffix get added? like suppose asan. 
Do we really need a static list here?

nunnikri (approved):
i would suggest to go without using a static list. Otherwise its an improvement

build_tools/packaging/linux/packaging_utils.py
raramakr: Good point. I looked into this - the suffixes are stable and come from artifact.toml definitions ([components.{lib,run,dev,doc,test,dbg}.*]), documented in
  docs/development/artifacts.md. 
  
  Adding a new suffix like asan would require changes to the artifact.toml schema and cmake/therock_artifacts.cmake, so updating this list would naturally be part
   of that work.

  Added a comment to clarify the source:
   Component suffixes from artifact.toml [components.{suffix}.*] definitions
   See docs/development/artifacts.md for artifact naming conventions
  SUFFIX_MARKERS = ["_dbg_", "_dev_", "_doc_", "_lib_", "_run_", "_test_"]
  
  The static list is fine here - it's fast and unlikely to drift since any new component type would be a deliberate infrastructure change. 
The package.json also need to be updated if new components are added

Issue #4789, opened 2026-04-24
Add ability to run build-time tests


stellaraccident:
Historically, most ROCm tests get built standalone and then distributed to GPU runners for execution. But there are a number of projects which (correctly) have CPU-side unit tests intended to be run from the build tree. We want to support these with a new `therock_subproject_build_test()` directive that:

* Adds test stamps and wires up the command to execute in the build graph
* Only runs for built projects (skips for prebuilt)
* Is excluded from ALL and must be explicitly invoked via `build-test-foo` or `build-tests` targets.

While implementing this, it was attempted to enable the KPACK C++ unit tests but they have an issue that was not caught in the standalone pre-submits:

```
  [1/2] Linking CXX executable runtime/tests/test_kpack_runtime
  /usr/bin/ld: runtime/tests/CMakeFiles/test_kpack_runtime.dir/test_isa_target_match.cpp.o:
  undefined reference to `kpack::parse_target_id(std::__cxx11::basic_string<char, ...> const&)'
  /usr/bin/ld: runtime/tests/CMakeFiles/test_kpack_runtime.dir/test_isa_target_match.cpp.o:
  undefined reference to `kpack::strip_target_prefix[abi:cxx11](char const*)'
  collect2: error: ld returned 1 exit status
```

Cause/solution:

* Cause: test_isa_target_match.cpp white-box tests internal helpers from runtime/src/isa_target_match.cpp, but test_kpack_runtime only links runtime/librocm_kpack.so.0.1.0. In TheRock’s shared-library build those internal helper symbols are not available to the test executable.
* Confirmed fix: adding ../src/isa_target_match.cpp to test_kpack_runtime sources makes the test executable link, and then kpack CTest passes 101/101. I backed that fix out per request.

Landing this facility with the KPACK unit tests commented out. We should hold this issue open until that is resolved and they are enabled.

PR #4790, opened 2026-04-24
Add ability to run build-time tests.

stellaraccident:
* Adds a `therock_subproject_build_test(target COMMAND ... [COMMAND ...])` directive that adds test commands to execute in the build graph.
* Only runs for built projects (skips for prebuilt)
* Is excluded from ALL and must be explicitly invoked via build-test-foo or build-tests targets.

Attempted to enable for KPACK which has such tests but encountered an issue so leaving that commented out pending landing of a fix.

Verified:
* Locally patched KPACK and verified that tests run and logs are written.
* Added a temporary second test command and verified those tests run and logs are written.

Progress on #4789


stellaraccident:
@kasaurov review please

kasaurov (commented):
LGTM, I created https://github.com/ROCm/TheRock/pull/4881 based on these changes

kasaurov (approved):
LGTM

PR #4800, opened 2026-04-24
[hipBLAS][rocBLAS][hipblaslt] correct subprocess coding to capture stderr in logs

TorreZuk:
This pull request updates the test scripts for several HIP/ROCm-related executable tests to improve error logging. The main change is to redirect standard error output (`stderr`) to standard output (`STDOUT`) in all subprocess calls, making debugging easier by ensuring all output is captured together.

**Improvements to subprocess error logging:**

* Updated `subprocess.check_call` and `subprocess.run` calls in `test_hipblas.py`, `test_hipblaslt.py`, `test_hipblasltprovider.py`, and `test_rocblas.py` to include `stderr=subprocess.STDOUT`, so that error messages are included in the standard output logs. 

Could be a major flaw in entire theRock infrastructure but only fixing subset in this PR which would need to be replicated across the board if exists everywhere.

[issue-6620] 
ROCM-23617

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_hipblas.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hipblas.py
@@ -57,4 +57,4 @@ else:
 
 
 logging.info(f"++ Exec [{THEROCK_DIR}]$ {shlex.join(cmd)}")
-subprocess.check_call(cmd, cwd=THEROCK_DIR, env=environ_vars)
+subprocess.check_call(cmd, cwd=THEROCK_DIR, env=environ_vars, stderr=subprocess.STDOUT)
--- a/build_tools/github_actions/test_executable_scripts/test_hipblaslt.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hipblaslt.py
@@ -49,4 +49,4 @@ elif test_type == "quick":
 cmd = [f"{THEROCK_BIN_DIR}/hipblaslt-test"] + test_filter
 
 logging.info(f"++ Exec [{THEROCK_DIR}]$ {shlex.join(cmd)}")
-subprocess.run(cmd, cwd=THEROCK_DIR, check=True, env=environ_vars)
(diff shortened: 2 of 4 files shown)


tony-davis:
@davidd-amd are we missing something regarding stderr? I feel like maybe stderr goes to the logs but not to the displayed logs in GitHub? Can you advise us?

geomin12 (approved):
lgtm

we are actually planning to migrate this to rocm-libraries here: https://github.com/ROCm/rocm-libraries/tree/develop/test/therock 

so please update there as well, cc: @jayhawk-commits

build_tools/github_actions/test_executable_scripts/test_hipblas.py
```diff
@@ -57,4 +57,4 @@
 
 
 logging.info(f"++ Exec [{THEROCK_DIR}]$ {shlex.join(cmd)}")
-subprocess.check_call(cmd, cwd=THEROCK_DIR, env=environ_vars)
+subprocess.check_call(cmd, cwd=THEROCK_DIR, env=environ_vars, stderr=subprocess.STDOUT)
```
geomin12, line 60: this is probably a flaw in the gtest executable? as it isn't reporting errors in the right place?

TorreZuk:
When to keep them separate:
Producing machine-readable output | ✔️  
Writing CLI tools for users | ✔️ 

Merge:
Debugging CI logs |   ✔️
Running tests in GitHub Actions |   ✔️

build_tools/github_actions/test_executable_scripts/test_hipblas.py
TorreZuk: No the gtest executables writes to stdout as expected and we may want to output stderr to different files, the test harness in theRock scripts should be fixed as it is for debugging runs and only keeps one log recorded of the run thus merge both.  We can move all the fixes to rocm-libraries but the fixes for other components should be landed in therock if they don't move.

PR #4783, opened 2026-04-23
Build fat PyTorch wheel and kpack-split in multi-arch CI

marbre:
## Motivation

Multi-arch PyTorch CI has been running N per-family builds that produced N identical fat wheels. Every job saw the same superset rocm-sdk-core and build_prod_wheels.py silently fell back to rocm-sdk targets for PYTORCH_ROCM_ARCH. That fallback is also the root cause of #4687 - it misreports targets whenever the installed rocm-sdk diverges from build intent. The per-family matrix was never actually per-family.

This PR replaces the matrix with a single fat build followed by rocm-kpack wheel splitting, plumbs the target list explicitly end-to-end, and uploads the resulting host + per-arch device wheels alongside the ROCm wheels the build_python_packages job already writes, so one --find-links URL installs the whole SDK + PyTorch stack.

## Technical Details

This still makes use of GPU families which are converted into GPU targets. While this helper was previously limited to artifact_manager.py, this is now accessible via a helper script while the code is reused. We want to switch over from families to targets in the long run but the passed targets variables is currently restrict to _test_ targets and misses targets for which no test runner is available.

Intentionally out of scope:
* upstreaming the kpack objcopy/readelf fallback
* making --pytorch-rocm-arch strictly required after the legacy single-arch and release workflows plumb it through

PR #4815, opened 2026-04-24
fix install_rocm_from_artifacts when kpack built family of artifacts

TorreZuk:
This pull request adds logic to support expanding AMDGPU family artifact groups (such as `gfx110X-all`) into their individual GPU targets during artifact installation. This is especially important for split or kpack builds, where artifacts are generated per GPU target rather than as a family archive. The main changes include introducing helper functions to parse and cache the mapping from family names to individual targets, and updating the artifact retrieval logic to use this mapping when needed.

**AMDGPU target expansion improvements:**

* Added imports and helper functions (`parse_amdgpu_targets_cmake`, `build_family_to_targets`) from `_therock_utils.cmake_amdgpu_targets` for parsing the CMake file that defines AMDGPU target families.
* Implemented a caching mechanism and the `get_family_to_targets` function to efficiently parse and cache the family-to-targets mapping from `therock_amdgpu_targets.cmake`.
* Added `resolve_artifact_group_to_targets`, which expands a family artifact group (like `gfx110X-all`) into its individual GPU targets using the cached mapping.

**Artifact retrieval logic updates:**

* Updated `retrieve_artifacts_by_run_id` to automatically expand artifact groups into their individual targets if explicit targets are not provided, ensuring correct artifact downloads for split/kpack builds.

ROCM-23699

### DIFF
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -128,6 +128,11 @@ import sys
 import tarfile
 from typing import Optional
 
+from _therock_utils.cmake_amdgpu_targets import (
+    parse_amdgpu_targets_cmake,
+    build_family_to_targets,
+)
+
 PLATFORM = platform.system().lower()
 s3_client = boto3.client(
     "s3",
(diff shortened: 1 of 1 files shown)


build_tools/install_rocm_from_artifacts.py
```diff
@@ -141,6 +146,61 @@
...
+
+    return _FAMILY_TO_TARGETS_CACHE
+
+
+def resolve_artifact_group_to_targets(artifact_group: str) -> list[str]:
+    """Resolve an artifact group (family name) to individual GPU targets.
+
+    For split/kpack builds, artifacts are created per individual target
+    (e.g., gfx1100, gfx1101) rather than as family archives (e.g., gfx110X-all).
+    This function expands a family name into its constituent targets.
```
ScottTodd: @marbre is adding a common helper for this in https://github.com/ROCm/TheRock/pull/4783 (we can split that into a smaller PR, though he's offline today)

marbre: We have `build_tools/_therock_utils/cmake_amdgpu_targets.py` used by `build_tools/github_actions/expand_amdgpu_families.py` with #4783 merged into main now. I think this might suit your needs.

marbre:
@TorreZuk closing this we extended TheRock in other commits. Please open an issue if you need more.

PR #4817, opened 2026-04-24
CI - Add downstream testing pipeline for targeted CI   

kiritigowda:
## Motivation

A new standalone CI pipeline that tests only the components affected by a change. It figures out what depends on what you changed, and runs just those tests. No existing CI files are modified.

### 3 new files:

  **`.github/workflows/ci_downstream_tests.yml`** — The workflow with two triggers:
  - **Manual** (`workflow_dispatch`): You specify which artifacts changed and which build to test against
  - **Automatic** (`pull_request`): Add the `ci:test-downstream` label to a PR. The pipeline detects which files changed, waits for the PR's CI build to finish, then runs only the downstream tests. Re-runs automatically on new pushes while the label is present.

  **`build_tools/github_actions/configure_downstream_tests.py`** — Python script that does the actual work:
  - Maps changed file paths to artifacts using `.gitmodules` and `BUILD_TOPOLOGY.toml`
  - Walks the dependency graph to find all downstream consumers
  - Translates artifact names to test labels that `test_artifacts.yml` understands
  - Builds per-GPU-family configs for the test matrix

  **`build_tools/tests/configure_downstream_tests_test.py`** — 33 unit tests.

## Test Plan

configure_downstream_tests_test.py - follows the existing test pattern

## Test Result

unit_tests.yml runs python -m pytest across all of build_tools/, which automatically discovers this file and runs with other tests

## Submission Checklist


TorreZuk:
This relates to https://github.com/ROCm/TheRock/pull/4570 and possibly also https://github.com/ROCm/rocm-libraries/pull/6133

.github/workflows/ci_downstream_tests.yml
```diff
@@ -0,0 +1,190 @@
...
+        required: false
+        type: choice
+        options:
+          - full
+          - comprehensive
+          - standard
+          - quick
+        default: "full"
+  pull_request:
+    types: [labeled, synchronize]
```
ScottTodd, line 41: Please be mindful of adding new github actions workflows like this. Pushing a draft PR with a `pull_request` trigger for a new workflow file can be considered as deliberately working around code review and security policies. The https://github.com/ROCm/TheRock/actions page is a shared resource and the list of workflows is flooded with draft workflows that did not get approved/merged (some of which should not have been pushed at all), e.g. 

Image

I'd suggest one of these contributing styles instead:
* Prototype entirely new workflows in a fork (without access to self-hosted runners, secrets, etc.)
* Propose a placeholder workflow via a PR like https://github.com/ROCm/TheRock/pull/1740 or https://github.com/ROCm/TheRock/pull/5127 , get agreement to use that workflow name, then use workflow_dispatch with the placeholder workflow once merged to iterate and get it functional

PR #4818, opened 2026-04-24
users/torrezuk/rocm 23617 test revert

TorreZuk:
You can review this PR test results with submodule bump, but the pure revert is in the linked new PR https://github.com/ROCm/TheRock/pull/4835 so would be mergable without this submodule bump.  Preferred approach, so don't merge this PR unless that submodule commit is needed today.

This pull request updates the default behavior for splitting KPACK artifacts and removes explicit overrides in the build configuration script. It also updates the `rocm-libraries` submodule to a new commit.

**Build configuration and flag management:**

* Changed the default value of the `KPACK_SPLIT_ARTIFACTS` flag from `ON` to `OFF` in `FLAGS.cmake`, meaning artifact splitting is now opt-in by default.
* Removed the explicit `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF` setting from the `build_configure.py` script, as the new default makes this unnecessary.

**Submodule updates:**

* Updated the `rocm-libraries` submodule to a newer commit (`56eee18bfefd703751b5c0371b942c2e00e2dfd6`).

### DIFF
--- a/FLAGS.cmake
+++ b/FLAGS.cmake
@@ -17,7 +17,7 @@ include(therock_flag_utils)
 
 therock_declare_flag(
   NAME KPACK_SPLIT_ARTIFACTS
-  DEFAULT_VALUE ON
+  DEFAULT_VALUE OFF
   DESCRIPTION "Split target-specific artifacts into generic and arch-specific components"
 )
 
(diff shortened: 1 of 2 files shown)


TorreZuk:
This looks like it makes the windows gfx110X all pass again related to the bump PR issues.

HereThereBeDragons:
can we close this? i think some other solution was found for this problem?

TorreZuk:
Yes I think we are still finding new related bugs but will close this test revert.

PR #4844, opened 2026-04-27
[ci] Fix MIOpen build failure on gfx110X by disabling CK-dependent tests

yash-amd:
## Motivation

miopen tests were failing [here](https://github.com/ROCm/TheRock/actions/runs/24944514058/job/73089055608)
with error: use of undeclared identifier 'ck'

## Technical Details

miopen tests are depend on Composable Kernel (CK) headers, but CK only supports certain GPU architectures. When building for unsupported targets like gfx110X (RDNA3/Navi3), the build system correctly disables CK but MIOpen tests were
  still being compiled, causing compilation errors

## changes
This change disables MIOpen tests when THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL is OFF, preventing the CK-dependent test code from being compiled on unsupported architectures.

### DIFF
--- a/ml-libs/CMakeLists.txt
+++ b/ml-libs/CMakeLists.txt
@@ -75,10 +75,18 @@ if(THEROCK_ENABLE_MIOPEN)
     list(APPEND optional_miopen_build_deps composable_kernel)
   endif()
 
-  if(MSVC)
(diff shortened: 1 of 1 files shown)


yash-amd:
upstream fix is already merged in rocm-libraries [here](https://github.com/ROCm/rocm-libraries/pull/6834)

PR #4850, opened 2026-04-27
[debug-tools] Enable configurable debug builds of debug-tools

lumachad:
Enables passing DEBUG_TOOLS_BUILD_TYPE to create a minimal build of TheRock containing debug-tools with debugging symbols.

Also allow specifying C and CXX flags when doing "Debug" builds of debug-tools.

These options make it easy to produce a ROCm stack for development purposes, where debug symbols are necessary and optimization is disabled.

### DIFF
--- a/debug-tools/CMakeLists.txt
+++ b/debug-tools/CMakeLists.txt
@@ -3,6 +3,54 @@
 
 set(DEBUG_TOOLS_DIR "${THEROCK_SOURCE_DIR}/debug-tools")
 
+# If set, propagate DEBUG_TOOLS_BUILD_TYPE to all debug-tools components.
+if(DEBUG_TOOLS_BUILD_TYPE)
+  message(STATUS "Setting all debug-tools components to CMAKE_BUILD_TYPE=${DEBUG_TOOLS_BUILD_TYPE}")
+  if(NOT amd-dbgapi_BUILD_TYPE)
+    set(amd-dbgapi_BUILD_TYPE "${DEBUG_TOOLS_BUILD_TYPE}")
+  endif()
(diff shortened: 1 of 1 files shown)


debug-tools/CMakeLists.txt
```diff
@@ -3,6 +3,54 @@
...
+  endif()
+  if(NOT rocgdb_BUILD_TYPE)
+    set(rocgdb_BUILD_TYPE "${DEBUG_TOOLS_BUILD_TYPE}")
+  endif()
+endif()
+
+# Allow custom debug flags for debug-tools components.
+# Example: -DDEBUG_TOOLS_C_FLAGS_DEBUG="-O0 -g3"
+#          -DDEBUG_TOOLS_CXX_FLAGS_DEBUG="-O0 -g3"
+if(DEBUG_TOOLS_C_FLAGS_DEBUG OR DEBUG_TOOLS_CXX_FLAGS_DEBUG)
```
geomin12, line 23: i am not experienced in cmake, but opportunity to add this here: https://github.com/ROCm/TheRock/blob/main/CMakePresets.json ?

lumachad: The debug-tools already honor the top-level Debug release type (defaults to passing -O0 -g3 for rocgdb and whatever TheRock provides for rocr-debug-agent and amd-dbgapi). These options are specifically for developers to build the debug-tools locally so they can debug problems. I don't think that would apply generically.

But let me go through those presents to see if we can hook debug-tools in there somewhere. Thanks for the pointer.

geomin12 (approved):
lgtm, but would feel more comfortable if debug tools team reviews / or someone from build

PR #4854, opened 2026-04-27
[debug-tools] Add comprehensive developer's guide to debug-tools README.md

lumachad:
Update the debug-tools documentation to include available options and development workflows useful for debugging tools.

### DIFF
--- a/debug-tools/README.md
+++ b/debug-tools/README.md
@@ -382,6 +382,55 @@ Common TSAN_OPTIONS:
 - `report_bugs=0` - Disable bug reporting (useful for performance testing)
 - `log_path=<path>` - Write logs to file instead of stderr
 
+### Testing Debug-Tools
+
+TheRock provides test scripts for ROCgdb and ROCr Debug Agent. Both scripts work with locally-built ROCm trees or downloaded nightly tarballs from https://rocm.nightlies.amd.com/.
+
+**Scripts location:** `build_tools/github_actions/test_executable_scripts/`
+
+#### Setup
+
+Set environment variables pointing to your ROCm installation:
+
+```bash
+# For TheRock builds
+export THEROCK_BIN_DIR=/therock/output/build/dist/rocm/bin
+export OUTPUT_ARTIFACTS_DIR=/therock/output/build/dist/rocm
+
+# For nightly tarballs
+export THEROCK_BIN_DIR=/path/to/rocm-developer-nightly/bin
+export OUTPUT_ARTIFACTS_DIR=/path/to/rocm-developer-nightly
+```
+
+For container testing, use the testing container described in [Container Environments](#testing-container).
+
+#### Running Tests
+
+**ROCgdb:**
+
+```bash
+python3 build_tools/github_actions/test_executable_scripts/test_rocgdb.py
+```
+
+The script runs the GDB test suite against ROCgdb using both GCC and LLVM compilers. By default it will run the gdb.rocm and gdb.dwarf2 test categories.
+
+**ROCr Debug Agent:**
+
+```bash
+python3 build_tools/github_actions/test_executable_scripts/test_rocr-debug-agent.py
+```
+
+The script includes automatic retry logic for flaky GPU-dependent tests.
+
+#### Results
+
+Both scripts report pass/fail status and exit with code 0 on success. ROCgdb tracks expected failures in the script's `XFAILED_TESTS` dictionary and reports:
+
+- Per-compiler results (GCC and LLVM)
+- Flaky tests (passed on retry)
+- Unexpected failures
+- Comparison of compiler-specific failures
+
 ## Additional information
 
 ### ROCgdb dependency on terminfo for TUI mode
--- a/debug-tools/README.md
+++ b/debug-tools/README.md
@@ -328,6 +328,60 @@ For additional information on the testing container image, see [dockerfiles/READ
 
 For testing reproduction steps, see [docs/development/test_environment_reproduction.md](../docs/development/test_environment_reproduction.md).
 
+### Building with Sanitizers
+
+TheRock provides built-in support for compiler sanitizers to help detect memory errors, data races, and other runtime bugs during development. For comprehensive documentation on sanitizer modes (`ASAN`, `HOST_ASAN`, `TSAN`), configuration options, CMake presets, and implementation details, see [docs/development/sanitizers.md](../docs/development/sanitizers.md).
+
+#### Debug-Tools Specific Configuration
+
+Enable sanitizers for debug-tools components using either project-wide or per-component settings:
+
+```bash
+# Project-wide
+-DTHEROCK_SANITIZER=HOST_ASAN
+
+# Per-component
+-Damd-dbgapi_SANITIZER=HOST_ASAN
+-Drocr-debug-agent_SANITIZER=HOST_ASAN
+-Drocgdb_SANITIZER=HOST_ASAN
+
+# Combined with debug builds
+-DDEBUG_TOOLS_BUILD_TYPE=Debug
+-DTHEROCK_SANITIZER=HOST_ASAN
+```
+
+#### Runtime Environment Configuration
+
+**Address Sanitizer (ASAN_OPTIONS):**
+
+```bash
+export ASAN_OPTIONS=detect_leaks=1:verbosity=1:log_path=/tmp/asan_log
+```
+
+Common ASAN_OPTIONS:
+
+- `detect_leaks=1` - Enable memory leak detection (enabled by default)
+- `verbosity=N` - Increase diagnostic verbosity (0-2)
+- `check_initialization_order=1` - Detect initialization order bugs
+- `detect_stack_use_after_return=1` - Detect use-after-return bugs (more expensive)
+- `halt_on_error=1` - Stop on first error
+- `log_path=<path>` - Write logs to file instead of stderr
+
+**Thread Sanitizer (TSAN_OPTIONS):**
+
+```bash
+export TSAN_OPTIONS=verbosity=1:second_deadlock_stack=1:history_size=7
+```
+
+Common TSAN_OPTIONS:
+
+- `verbosity=N` - Increase diagnostic verbosity (0-2)
+- `halt_on_error=1` - Stop on first error
+- `second_deadlock_stack=1` - Show second stack trace for deadlocks
+- `history_size=N` - Size of per-thread history buffer (default 2, max 7)
+- `report_bugs=0` - Disable bug reporting (useful for performance testing)
+- `log_path=<path>` - Write logs to file instead of stderr
+
 ## Additional information
 
 ### ROCgdb dependency on terminfo for TUI mode
--- a/debug-tools/README.md
+++ b/debug-tools/README.md
@@ -248,6 +248,10 @@ docker run --rm -i -t \
 This container includes all necessary build dependencies and toolchains for
 building TheRock components.
 
+**Note:** CI builds use a pinned image version rather than `:latest`. When using
+`:latest`, you may encounter different behavior or updated dependencies compared
+to what CI uses.
+
 For additional information on the manylinux image, see [dockerfiles/README.md](../dockerfiles/README.md#build_manylinux_dockerfile).
 
 **GPU access for testing:**
@@ -350,8 +354,10 @@ required by ROCgdb, such as dejagnu (for the GDB test suite), gfortran (for
 Fortran debugging tests), and other utilities needed for comprehensive debugger
 testing. No builds are performed in this container.
 
-**Note:** This testing container matches the CI environment and can be used to
-locally reproduce CI test failures.
+**Note:** CI tests use a pinned image version rather than `:latest`. To exactly
+reproduce CI test results, use the same pinned version that CI uses. When using
+`:latest`, you may encounter different behavior or updated testing tools compared
+to the CI environment.
 
 For additional information on the testing container image, see [dockerfiles/README.md](../dockerfiles/README.md#no_rocm_image_dockerfile).
 
--- a/debug-tools/README.md
+++ b/debug-tools/README.md
@@ -21,6 +21,19 @@ rocr-debug-agent: The ROCr debug agent source code.
 This guide covers configuration options and development workflows for building
 and debugging the debug-tools components.
 
+**Note:** This guide assumes all commands are executed from within a manylinux
+container. TheRock can also be built outside of a container, but this requires
+distro-specific environment configuration. For container setup, see
+[Container Environments](#build-container).
+
+Outside of the container, TheRock sources can be cloned with the following command:
+
+```bash
+# Clone the repository
+git clone https://github.com/ROCm/TheRock.git
+cd TheRock
+```
+
 ### Quick Start
 
 For users that just need a prebuilt ROCm stack, please refer to the [releases page](../RELEASES.md).
@@ -31,10 +44,14 @@ sources, the default configuration works out of the box:
 ```bash
 cmake -B /therock/output/build -GNinja \
   -DTHEROCK_AMDGPU_FAMILIES="gfx942" \
-  -DTHEROCK_TEST_AMDGPU_FAMILIES="gfx942" \
   /therock/src
 cmake --build /therock/output/build -t rocgdb amd-dbgapi rocr-debug-agent
 ```
+
+**Note:** `-DTHEROCK_TEST_AMDGPU_FAMILIES` can optionally be supplied to restrict
+the number of targets for which rocr-debug-agent-tests is built, as these tests
+are device-specific. If not specified, tests will be built for all available gfx
+targets.
 
 For development and debugging, you may want to build in Debug mode by passing
 the following option:
@@ -222,12 +239,6 @@ For building debug-tools components, use the TheRock build container:
 
 ```bash
 docker run --rm -i -t \
-  --ipc host \
-  --group-add video \
-  --device /dev/kfd \
-  --device /dev/dri \
-  --security-opt seccomp=unconfined \
-  --cap-add=SYS_PTRACE \
   --mount type=bind,src=$HOME/therock/output,dst=/therock/output \
   --mount type=bind,src=$HOME/therock,dst=/therock/src \
   --name $(whoami) ghcr.io/rocm/therock_build_manylinux_x86_64:latest \
@@ -237,10 +248,25 @@ docker run --rm -i -t \
 This container includes all necessary build dependencies and toolchains for
 building TheRock components.
 
-While the build itself does not depend on a GPU, it is worth passing docker
-options to enable the GPU within it so testing can take place.
-
 For additional information on the manylinux image, see [dockerfiles/README.md](../dockerfiles/README.md#build_manylinux_dockerfile).
+
+**GPU access for testing:**
+
+While the build itself does not depend on a GPU, you can enable GPU access within
+the container for testing after a build by adding these options to the `docker run`
+command:
+
+```bash
+  --ipc host \
+  --group-add video \
+  --device /dev/kfd \
+  --device /dev/dri \
+  --security-opt seccomp=unconfined \
+  --cap-add=SYS_PTRACE
+```
+
+**Note:** `--security-opt seccomp=unconfined` and `--cap-add=SYS_PTRACE` are
+required for debug-tools testing, as debuggers need ptrace capabilities.
 
 **Environment setup within the container:**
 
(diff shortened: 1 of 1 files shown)


lumachad:
Companion for #4850. Should go in once that one is merged.

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+and debugging the debug-tools components.
+
+### Quick Start
+
+For most users, the default configuration works out of the box:
+
+```bash
+cmake -B /therock/output/build -GNinja \
+  -DTHEROCK_AMDGPU_FAMILIES="gfx942" \
+  -DTHEROCK_TEST_AMDGPU_FAMILIES="gfx942" \
```
HereThereBeDragons: do we need this? i never build with this

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+
+This guide covers configuration options and development workflows for building
+and debugging the debug-tools components.
+
+### Quick Start
+
+For most users, the default configuration works out of the box:
+
+```bash
+cmake -B /therock/output/build -GNinja \
```
HereThereBeDragons, line 45: i think its not good to start with leading slash `/therock/...`. most people dont have this setup

thats what we have in the top level readme
```
cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES=gfx110X-all
cmake --build build
```

just the `build/` directory under `git clone` therock repo

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+```
+
+Individual component build type variables take precedence over
+`DEBUG_TOOLS_BUILD_TYPE`. For example, if you set
+`-DDEBUG_TOOLS_BUILD_TYPE=Debug` and `-Drocgdb_BUILD_TYPE=Release`, ROCgdb will
+build in Release mode while other components build in Debug mode.
+
+#### DEBUG_TOOLS_C_FLAGS_DEBUG and DEBUG_TOOLS_CXX_FLAGS_DEBUG
+
+Customize debug flags for C and C++ code when building in Debug mode. These
```
HereThereBeDragons, line 116: maybe a comment if those flags are already enabled when building with `-DDEBUG_TOOLS_BUILD_TYPE=Debug`? or why is this extra needed? what does `Debug` than enable?

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+-DDEBUG_TOOLS_C_FLAGS_DEBUG="-Og -ggdb" \
+-DDEBUG_TOOLS_CXX_FLAGS_DEBUG="-Og -ggdb"
+```
+
+These flags are appended to component-specific `CMAKE_C_FLAGS_DEBUG` and
+`CMAKE_CXX_FLAGS_DEBUG` and only apply when the build type is `Debug`.
+
+#### THEROCK_SHARED_PYTHON_EXECUTABLES
+
+ROCgdb supports Python scripting integration, which requires linking against a
```
HereThereBeDragons, line 158: maybe a comment to use a activated venv when building to match it to your favorite python version?

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+  -DTHEROCK_TEST_AMDGPU_FAMILIES="gfx942" \
+  /therock/src
+cmake --build /therock/output/build -t rocgdb
+```
+
+```bash
+# Use external amd-dbgapi source
+cmake -B /therock/output/build -GNinja \
+  -DTHEROCK_USE_EXTERNAL_AMD_DBGAPI=ON \
+  -DTHEROCK_AMD_DBGAPI_SOURCE_DIR=/path/to/my/amd-dbgapi \
```
HereThereBeDragons: could put DTHEROCK_AMD_DBGAPI_SOURCE_DIR and DTHEROCK_ROCGDB_SOURCE_DIR in a single command and not 2 commands? a little bit of deduplication

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+docker run --rm -i -t \
+  --ipc host \
+  --group-add video \
+  --device /dev/kfd \
+  --device /dev/dri \
+  --security-opt seccomp=unconfined \
+  --cap-add=SYS_PTRACE \
+  --mount type=bind,src=$HOME/therock/output,dst=/therock/output \
+  --mount type=bind,src=$HOME/therock,dst=/therock/src \
+  ghcr.io/rocm/therock_build_manylinux_x86_64:latest \
```
HereThereBeDragons: latest might not be the thing you are looking for.

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+```
+
+This container includes all necessary build dependencies and toolchains for
+building TheRock components.
+
+**Environment setup within the container:**
+
+Before building, set up the Python virtual environment and ccache:
+
+```bash
```
HereThereBeDragons, line 279: can we copy 1:1 what we have in the top level readme? or refer to it and that you also need to pip install requirements-test.txt

debug-tools/README.md
lumachad: We have rocr-debug-agent-tests that is gfx-specific. If we don't pass in the TEST family, it will build for all gfx targets. Do we have some sort of auto-discovery now?

debug-tools/README.md
lumachad: We usually do builds from within the manylinux container, where /therock is mapped to some appropriate directory in the host. Would you say the appropriate way to build things is outside of the container?

debug-tools/README.md
lumachad: Yeah, we do enable -O0 -g3 by default. But developers might want more insight when working through, say, rocgdb bugs. That's why these options are made available. I'll add a comment about the default.

debug-tools/README.md
lumachad: That's not quite the workflow we've been following when building within manylinux. I agree you might want to set your environment before building things outside of the container, but within the container we have strict paths to the known-good shared-python versions.

debug-tools/README.md
lumachad: Indeed. Will do. Thanks!

debug-tools/README.md
lumachad: We have had cases where we've been stuck with a very old manylinux image because the rocm-systems CI triggers were using outdated SHA's. For development purposes (as this is meant to be a developer's guide) we'd like to use the latest available image so we know about any issues that may show up. Does that make sense?

debug-tools/README.md
lumachad: We definitely can. I'll add the pointers. But it's worth keeping in mind some of these instructions are tailored towards debug-tools developers, so they have slightly different options. For instance, we never build the full ROCm stack, as it doesn't make sense for our purposes.

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
 ```
 
+## Developer's Guide
+
+This guide covers configuration options and development workflows for building
+and debugging the debug-tools components.
+
+### Quick Start
+
+For most users, the default configuration works out of the box:
```
geomin12: should we also point them to https://github.com/ROCm/TheRock/blob/main/RELEASES.md so they can just install them and use?

then we can add a step below saying `Build from source`?

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+  `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` enables compilation caching for faster
+  rebuilds.
+- **Resource limits**: `-DFLANG_PARALLEL_COMPILE_JOBS=20` and
+  `-DLLVM_PARALLEL_LINK_JOBS=20` limit parallel compilation and linking to
+  prevent memory and CPU exhaustion.
+- **Debug-tools only**: When building only debug-tools, use
+  `-DTHEROCK_ENABLE_ALL=OFF -DTHEROCK_ENABLE_DEBUG_TOOLS=ON` to avoid building
+  unnecessary components.
+
+#### Testing Container
```
geomin12, line 330: could link them here too? https://github.com/ROCm/TheRock/blob/main/docs/development/test_environment_reproduction.md

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+#### Build Container
+
+For building debug-tools components, use the TheRock build container:
+
+```bash
+docker run --rm -i -t \
+  --ipc host \
+  --group-add video \
+  --device /dev/kfd \
+  --device /dev/dri \
```
geomin12: technically for build, you don't need GPU access (only needs CPU for build), so could just run build docker image without any args

debug-tools/README.md
```diff
@@ -16,6 +16,380 @@ rocgdb/source: The ROCgdb source code.
...
+  -DDEBUG_TOOLS_CXX_FLAGS_DEBUG="-O0 -ggdb" \
+  -DTHEROCK_AMDGPU_FAMILIES="gfx942" \
+  -DTHEROCK_TEST_AMDGPU_FAMILIES="gfx942" \
+  /therock/src
+
+# Build debug-tools components
+cmake --build /therock/output/build -t amd-dbgapi rocr-debug-agent rocgdb
+
+# After modifying ROCgdb source, rebuild just that component
+cmake --build /therock/output/build -t rocgdb+build
```
geomin12: is that correct?

```suggestion
cmake --build /therock/output/build -t rocgdb-build
```
?

geomin12 (commented):
overall looks good, i would like to see some debug tool team members review and make sure this is good too

debug-tools/README.md
lumachad: Done

debug-tools/README.md
lumachad: Done.

debug-tools/README.md
lumachad: Added a reference to the top level README.md.

debug-tools/README.md
lumachad: Done.

debug-tools/README.md
lumachad: Right, but this is a developer's guide for debug-tools, so it is beneficial to have GPU access even within the build container. That allows for debugging alongside the built tree instead of having to move to a testing container for that.

debug-tools/README.md
lumachad: Technically it is rocgdb+build. But I have decided to remove these as they seem to be minor details that are up to developers to use or not.

lumachad:
Added a couple of debug-tools folks.

debug-tools/README.md
```diff
@@ -16,6 +16,318 @@ rocgdb/source: The ROCgdb source code.
...
+
+The debug-tools components have dependencies on other ROCm components and system
+libraries. The handling of system library dependencies is controlled by the
+`THEROCK_BUNDLE_SYSDEPS` CMake option.
+
+| Component        | ROCm Dependencies        | Bundled System Dependencies (THEROCK_BUNDLE_SYSDEPS=ON) | System Dependencies (THEROCK_BUNDLE_SYSDEPS=OFF) |
+| ---------------- | ------------------------ | ------------------------------------------------------- | ------------------------------------------------ |
+| amd-dbgapi       | amd-comgr                | libbacktrace                                            | System libbacktrace                              |
+| rocr-debug-agent | amd-dbgapi, ROCR-Runtime | elfutils                                                | System elfutils                                  |
+| rocgdb           | amd-comgr, amd-dbgapi    | bzip2, gmp, mpfr, expat, ncurses, liblzma, zlib, zstd   | System versions of these libraries               |
```
lancesix: Odd to list `amd-comgr` as a dependency for `rcogdb` but not `rocr-debug-agent`.  In both case, comgr is a transitive dependency from comgr.

debug-tools/README.md
```diff
@@ -16,6 +16,318 @@ rocgdb/source: The ROCgdb source code.
...
+
+#### DEBUG_TOOLS_C_FLAGS_DEBUG and DEBUG_TOOLS_CXX_FLAGS_DEBUG
+
+Customize debug flags for C and C++ code when building in Debug mode. These
+provide fine-grained control over optimization levels and debug information.
+
+```bash
+  -DDEBUG_TOOLS_BUILD_TYPE=Debug
+  -DDEBUG_TOOLS_C_FLAGS_DEBUG="-O0 -ggdb"
+  -DDEBUG_TOOLS_CXX_FLAGS_DEBUG="-O0 -ggdb"
```
lancesix, line 122: Is there an equivalent for LDFLAGS?  Could be useful to build with sanitizers.

debug-tools/README.md
```diff
@@ -16,6 +16,318 @@ rocgdb/source: The ROCgdb source code.
...
+```
+
+```bash
+# Multiple Python versions
+  -DTHEROCK_SHARED_PYTHON_EXECUTABLES="/usr/bin/python3.10;/usr/bin/python3.11"
+```
+
+**Requirements:**
+
+The Python executable must provide a shared `libpython` library (e.g.,
```
lancesix: pedantically, the executable does not "provide" a libpython, the python package might.

debug-tools/README.md
lumachad: There isn't. But you can set this TheRock-wide via cmake variables to accomplish the same.

debug-tools/README.md
lumachad: The right word here is support. Let me change it.

debug-tools/README.md
lumachad: Let me remove this.

debug-tools/README.md
lumachad: Sorry, I read it too quickly. For debug-tools sanitizers support see:
- https://github.com/ROCm/TheRock/pull/4674
- https://github.com/ROCm/TheRock/pull/4676

We won't need to handle LDFLAGS explicitly/manually. We'll just need to pass the standard arguments to TheRock's build.

lancesix: If we go that route, are the C_FLAGS / CXX_FLAGS ones necessary as well?  The example is about debug  builds, which we should have with -DCMAKE_BUILD_TYPE=Debug (with less fine grain arguably).  In my mind LDFLAGS sits in the same boat, if we have the C/CXX flags, might as well have the LD one as well.

lumachad: > If we go that route, are the C_FLAGS / CXX_FLAGS ones necessary as well? The example is about debug builds, which we should have with -DCMAKE_BUILD_TYPE=Debug (with less fine grain arguably). In my mind LDFLAGS sits in the same boat, if we have the C/CXX flags, might as well have the LD one as well.

With C and CXX flags you can pass whatever options you see fit for a particular debugging scenario. I'm not sure how much more LDFLAGS would add in the context of trying to make debugging something easier. Any ideas here?

geomin12 (commented):
lgtm, but would prefer some debug folks to approve, and also Laura's comments!

lumachad:
> lgtm, but would prefer some debug folks to approve, and also Laura's comments!

Makes sense. Thanks for the review!

debug-tools/README.md
lancesix: The main use I would have for it would be to set a rpath so I can just use the binary from the GDB build tree and not have to set LD_LIBRARY_PATH.
Having this would also open the door of specifying the sanitizers.  From what I can see for the PRs you have above, it is about ASAN/STAN, but there could be more sanitizers we might want to use (ubsan, memsan).

That being said, it is probably a matter past this doc update to add a setting to set LDFLAGS, I am happy with the doc to document things as they are today.

lancesix (approved):
Had comments for improvements on the infra, but as long as the doc is in line with what is currently available, I am happy with it.

Thanks.

debug-tools/README.md
lumachad: You don't have to set LD_LIBRARY_PATH when you use the rocgdb executables from the artifact tree. The rpath's are already set. The sanitizers are meant to be specified generically for TheRock.

If there is a sanitizer you'd like to add, the correct way is to add it to TheRock in cmake/therock_sanitizers.cmake and have it propagate through the entire build, as opposed to specifically for rocgdb.

lumachad:
@HereThereBeDragons Did you have some thoughts on our comments?

lumachad:
Added sanitizer section for good measure.

lumachad:
Also thrown some testing tips for debug tools.

lumachad:
Unblocked since #4850 is in.

lumachad:
@spatrang Thoughts on this one?

debug-tools/README.md
marbre: It depends. For local dev work I don't build in a container. When I want to produce packages I can ship over to a different machine with potentially a different OS, I build in manylinux.

debug-tools/README.md
marbre: This is fine here but maybe link to the relevant CMake config or add a statement why you're setting this. No auto-config.

debug-tools/README.md
marbre: I see the later argument for why you build in a container and that you also pass the GPU to the container. If that is the workflow, fine but maybe this motivation seems to be unclear to the reviewers. Maybe highlight or rather add the rational close to the start of this doc?

debug-tools/README.md
marbre: True, while it is confusing as next a test container section follows.

debug-tools/README.md
lumachad: Definitely. We do have some level of implicit team-specific workflows here. Let me try to make those clear. Thanks for the heads up.

debug-tools/README.md
lumachad: Clarified this is for rocr-debug-agent-tests. Made it optional.

debug-tools/README.md
lumachad: Clarified the guide assumes building within containers for the quickest setup and cleaner environment. Optionally buildable outside a container.

debug-tools/README.md
lumachad: Clarified this now. And separated the GPU settings into a subsection from the build container.

debug-tools/README.md
lumachad: Clarified this now.

debug-tools/README.md
```diff
@@ -16,6 +16,442 @@ rocgdb/source: The ROCgdb source code.
...
+
+**Environment setup within the container:**
+
+Before building, set up the Python virtual environment and ccache:
+
+```bash
+# Create and activate virtual environment
+python3 -m venv ~/v3
+source ~/v3/bin/activate
+
```
geomin12, line 283: oh let's add a step for git clone THeRock here

geomin12 (approved):
overall lgtm, but i would want a debug team approval too! added one comment

debug-tools/README.md
lumachad: The container instructions assume it is already cloned, but I've added a line at the top on how to clone it outside the container.

debug-tools/README.md
marbre: Maybe add a note that CI does use a pinned version while latest is recommended for dev work but might not match CI?

lumachad: Added text to that effect now, in a new commit:

```
@@ -248,6 +248,10 @@ docker run --rm -i -t \
 This container includes all necessary build dependencies and toolchains for
 building TheRock components.

+**Note:** CI builds use a pinned image version rather than `:latest`. When using
+`:latest`, you may encounter different behavior or updated dependencies compared
+to what CI uses.
+
 For additional information on the manylinux image, see [dockerfiles/README.md](../dockerfiles/README.md#build_manylinux_dockerfile).

 **GPU access for testing:**
@@ -350,8 +354,10 @@ required by ROCgdb, such as dejagnu (for the GDB test suite), gfortran (for
 Fortran debugging tests), and other utilities needed for comprehensive debugger
 testing. No builds are performed in this container.

-**Note:** This testing container matches the CI environment and can be used to
-locally reproduce CI test failures.
+**Note:** CI tests use a pinned image version rather than `:latest`. To exactly
+reproduce CI test results, use the same pinned version that CI uses. When using
+`:latest`, you may encounter different behavior or updated testing tools compared
+to the CI environment.
```

lumachad:
Best if this goes in after https://github.com/ROCm/TheRock/pull/4676, as it contains things added by that PR.

PR #4868, opened 2026-04-27
[hipdnn samples] Increase timeouts so we dont have test timeouts when a sample test takes > 1 min

adickin-amd:
## Motivation
Were seeing variations of test times from 1 to 5x.  See the following issue for more info: https://github.com/ROCm/rocm-libraries/issues/6864


BrianHarrisonAMD (approved):
ty ty

PR #4870, opened 2026-04-27
Add local patch prefix kpack-split host wheel extras

marbre:
Applies https://github.com/ROCm/rocm-systems/pull/5514 locally until landed upstream to have the same extras straight from the start even though it adds a patch to TheRock.


marbre:
Only land if the patch is accepted for upstream inclusion.

ScottTodd (approved):
I've approved https://github.com/ROCm/rocm-systems/pull/5514 also

marbre:
The failures are unrelated.

ScottTodd:
Oops, this patch isn't getting applied during the release workflows, since they just update submodules and don't apply patches: https://github.com/ROCm/TheRock/blob/4dd72055956838eba297e505dd558c7cafce5bee/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml#L253-L267

PR #4877, opened 2026-04-27
Add initial multi-arch PyTorch release workflows

marbre:
## Motivation

This adds initial PyTorch release workflows to be dispatched from the release workflows to build PyTorch wheels without wiring the new workflows into our existing workflows.

## Technical Details

Instead of using workflow calls this explicitly uses workflow dispatch. This tradeoff is chosen as only two runs will be triggered, one for Linux and one for Windows, whereas a workflow call would significantly blow up the workflow graph.

ROCm packages get pulled from the artifact bucket and not the CDN, which allows a more flexible ordering and less dependencies on other workflow runs, once connected. The new workflows mimic the implementation in `build_{linux_portable,windows}_pytorch_wheels_ci.yml` without reusing the existing workflows. What parts can be reused after the new workflows have been extended to build additional packages (torchaudio, torchvideo, trition, apex).

## Test Plan

N/A

## Test Result

N/A
## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
@@ -87,7 +87,7 @@ jobs:
       optional_build_prod_arguments: ""
       SCCACHE_BUCKET: "therock-${{ inputs.release_type }}-pytorch-sccache"
       SCCACHE_REGION: us-east-2
-      SCCACHE_S3_KEY_PREFIX: "linux/multiarch/"
+      SCCACHE_S3_KEY_PREFIX: "/linux/multiarch/v1/"
       SCCACHE_S3_USE_SSL: "true"
       SCCACHE_S3_SERVER_SIDE_ENCRYPTION: "true"
       SCCACHE_IDLE_TIMEOUT: "600"
--- a/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
@@ -1,0 +1,254 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Multi-arch release variant of build_portable_linux_pytorch_wheels_ci.yml.
+#
+# Built shape:
+#   - One fat build_prod_wheels.py invocation per (python_version,
+#     pytorch_git_ref) matrix cell, covering all gfx targets expanded
+#     from the inputs.amdgpu_families set.
+#   - kpack-split the resulting fat torch wheel into a host wheel plus
+#     per-gfx amd-torch-device-* wheels.
+#   - Upload the result to s3://therock-{release_type}-python/v4/whl-staging/
+#     (server-side indexing handles the index page).
+#
+# Triggered exclusively via workflow_dispatch from
+# multi_arch_release_linux.yml's dispatch_pytorch_wheels job.
+#
+# Build phase only: testing and promotion are deferred follow-ups.
+# See https://github.com/ROCm/TheRock/issues/3291 for convergence plans.
+
+name: Multi-Arch Release Linux PyTorch Wheels
+
+on:
+  workflow_dispatch:
+    inputs:
+      amdgpu_families:
+        description: >-
+          Semicolon-separated list of AMD GPU families to build PyTorch for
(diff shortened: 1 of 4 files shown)


.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -0,0 +1,254 @@
...
+      id-token: write
+      contents: read
+    container:
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:702a5133851e6d1daf1207d2c9fbb01c2667914a5b6dc5a01faeb3ce66ea6421
+    env:
+      PACKAGE_DIST_DIR: ${{ github.workspace }}/output/packages/dist
+      optional_build_prod_arguments: ""
+      SCCACHE_BUCKET: "therock-${{ inputs.release_type }}-pytorch-sccache"
+      SCCACHE_REGION: us-east-2
+      SCCACHE_S3_KEY_PREFIX: "linux/multiarch/"
```
ScottTodd: nit: we may want a version in this key prefix, in case we change the cache configuration in the future and want a clean start

build_tools/github_actions/publish_pytorch_to_staging.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Upload PyTorch wheels from a local directory to release-bucket staging.
+
+Used by the multi-arch release Linux PyTorch wheels workflow to push the
+host wheel and per-gfx amd-torch-device-* wheels produced by the kpack
+splitter into the release bucket's staging path. A future test workflow
+consumes this path; a future promotion workflow copies it into v4/whl/.
```
ScottTodd: This comment will go stale quickly.

* The script is used for Windows workflows too
* "A future workflow" style comments are probably better in commit messages and not source files

build_tools/github_actions/publish_pytorch_to_staging.py
```diff
@@ -0,0 +1,79 @@
...
+_BUILD_TOOLS_DIR = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(_BUILD_TOOLS_DIR))
+
+from _therock_utils.s3_buckets import get_release_bucket_config
+from _therock_utils.storage_backend import create_storage_backend
+from _therock_utils.storage_location import StorageLocation
+
+logger = logging.getLogger(__name__)
+
+_STAGING_SUBDIR = "v4/whl-staging"
```
ScottTodd: nit: could move this inline so it's easier to see in context with the code, similar to https://github.com/ROCm/TheRock/blob/a956549ddc7ad8b41f62b52c913421e9c2d318e0/build_tools/github_actions/publish_rocm_to_release_buckets.py#L97-L98

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -0,0 +1,254 @@
...
+        run: |
+          ./external-builds/pytorch/pytorch_torch_repo.py checkout \
+            --repo-hashtag nightly
+
+      - name: Checkout PyTorch source (stable)
+        if: ${{ matrix.pytorch_git_ref != 'nightly' }}
+        run: |
+          ./external-builds/pytorch/pytorch_torch_repo.py checkout \
+            --gitrepo-origin https://github.com/ROCm/pytorch.git \
+            --repo-hashtag ${{ matrix.pytorch_git_ref }}
```
ScottTodd, line 136: TODO: checkout audio,vision,triton,apex?

ScottTodd (approved):
Fine to land and iterate. New workflows are difficult to test (especially on self-hosted runners) until they exist on the default branch in some form.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
marbre: I have this in the commit message right now and hope to follow up quickly.

build_tools/github_actions/publish_pytorch_to_staging.py
marbre: True, thanks for catching this. Updated the comment.

marbre:
Thanks for reviewing. Addressed the review comments and landing as is to further test and iterate.

PR #4878, opened 2026-04-27
CI: gate Windows hip-tests (ROCR) matrix entry behind WINDOWS_HIP_ROCR_TESTS flag

jayhawk-commits:
Adds a CI-only feature flag (`WINDOWS_HIP_ROCR_TESTS` env var) that gates the
informational `hip-tests (ROCR)` matrix entry on Windows. Previously the ROCR
entry was emitted unconditionally alongside the PAL entry; now PAL is always
emitted and ROCR is opt-in via the flag.

Changes:
- `fetch_test_configurations.py`: reads `WINDOWS_HIP_ROCR_TESTS` via `str2bool`
  and wraps the ROCR dict literal in an `if` guard (consistent with how
  `run_extended_tests` is handled nearby)
- `test_artifacts.yml`: exposes `windows_hip_rocr_tests` as a `workflow_input`
  (default `'false'`) in both `workflow_dispatch` and `workflow_call` blocks,
  wired through `${{ inputs.windows_hip_rocr_tests }}`
- Tests: adds `test_windows_hip_tests_default_emits_pal_only` to cover the
  gate-off path (previously untested); updates the two existing ROCR tests to
  set `WINDOWS_HIP_ROCR_TESTS=true` explicitly

All callers of `test_artifacts.yml` (`ci_windows.yml`, `multi_arch_ci_windows.yml`,
and the Linux variants) are unaffected — the input is optional and the `'false'`
default reproduces the intended suppressed-ROCR behavior with no changes required
at call sites.

The flag is expected to be temporary. When ROCR becomes the primary pass/fail
path (tracked in issue #3587), the flag, the PAL-specific entry, and the
`expect_failure: True` ROCR path can all be cleaned up in a single follow-up
commit.

### DIFF
--- a/build_tools/github_actions/tests/fetch_test_configurations_test.py
+++ b/build_tools/github_actions/tests/fetch_test_configurations_test.py
@@ -267,10 +267,26 @@ class FetchTestConfigurationsTest(unittest.TestCase):
     # Output contract
     # -----------------------
 
-    def test_windows_hip_tests_emits_pal_and_rocr_entries(self):
-        """On Windows, hip-tests run twice: PAL (pass/fail) and ROCR (informational)."""
+    def test_windows_hip_tests_default_emits_pal_only(self):
+        """On Windows, hip-tests emits only PAL by default (WINDOWS_HIP_ROCR_TESTS off)."""
         os.environ["RUNNER_OS"] = "Windows"
         os.environ["TEST_LABELS"] = json.dumps(["hip-tests"])
+
+        fetch_test_configurations.run()
+        components = self._get_components()
+
+        hip_jobs = [j for j in components if "hip-tests" in j["job_name"]]
+        self.assertEqual(len(hip_jobs), 1, "Expected only hip-tests (PAL)")
+        self.assertEqual(hip_jobs[0]["job_name"], "hip-tests (PAL)")
+        self.assertNotIn("expect_failure", hip_jobs[0])
+        self.assertEqual(hip_jobs[0]["total_shards"], 4)
+        self.assertEqual(hip_jobs[0]["shard_arr"], [1, 2, 3, 4])
+
+    def test_windows_hip_tests_emits_pal_and_rocr_entries(self):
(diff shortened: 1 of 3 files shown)


jayhawk-commits:
The CI run has Windows hip-tests on ROCR being skipped, so it aligns with desired default behaviour. Other failures should not be related to the changes in this PR and are known issues.

build_tools/github_actions/tests/fetch_test_configurations_test.py
```diff
@@ -267,10 +267,26 @@ def fake_get_all_families(_):
...
+
+        fetch_test_configurations.run()
+        components = self._get_components()
+
+        hip_jobs = [j for j in components if "hip-tests" in j["job_name"]]
+        self.assertEqual(len(hip_jobs), 1, "Expected only hip-tests (PAL)")
+        self.assertEqual(hip_jobs[0]["job_name"], "hip-tests (PAL)")
+        self.assertNotIn("expect_failure", hip_jobs[0])
+        self.assertEqual(hip_jobs[0]["total_shards"], 4)
+        self.assertEqual(hip_jobs[0]["shard_arr"], [1, 2, 3, 4])
```
agunashe, line 283: nit : total shards can be a const and the next arr can be a range with end number of total shards

idass1990: this is only present in the test version of the script (for unit testing CI changes) , the production script uses a different/dynamic method for shard handling

idass1990 (approved):
LGTM

build_tools/github_actions/tests/fetch_test_configurations_test.py
jayhawk-commits: I don't think that level of complexity is required for this unit test.

idass1990: agreed

agunashe (approved):

PR #3429, opened 2026-02-16
[fortran][libc] Enable fortran compile/runtime to use libc on device.

ronlieb:
enable libc to be built for fortran, fixes errors observed in smoke tests in todays 7.12

PR #4881, opened 2026-04-28
Add COMGR build-time unit tests to CI

kasaurov:
## Motivation

Adds build-time Comgr unit tests (LIT + CTest) to CI, building on top of the `therock_subproject_build_test` infrastructure from PR #4790.
N.B. The tests are non-blocking at the moment.

## Technical Details

Based on PR #4790 (`therock_subproject_build_test` mechanism) and PR #4259 (Comgr test integration).

- Comgr test integration (`compiler/CMakeLists.txt`):
  - Adds `build-test-amd-comgr` target running LIT tests (`llvm-lit`) followed by CTest
  - `THEROCK_BUILD_COMGR_TESTS` defaults to `THEROCK_BUILD_TESTING` when not explicitly set
  - Removes option(`THEROCK_BUILD_COMGR_TESTS ...`) from top-level `CMakeLists.txt` to allow proper defaulting
  - Removes `llvm-lit` install/wrapper logic from `pre_hook_amd-llvm.cmake` (tests run from build tree)
- Independent test command execution (`cmake/therock_subproject.cmake`):
  - Modifies `therock_cmake_subproject_build_test()` to generate a CMake runner script that executes each COMMAND via `execute_process` independently -- a failure in one (e.g., LIT) does not prevent others (e.g., CTest) from running
  - Update suites enumeration schema -- for more than one suite enumerate staring from the first one (comgr build-test N)
  - Adds empty line separators between test command outputs
  - Reports combined pass/fail result at the end
  - Improves build test output with per-suite headers and a summary table
- CI integration:
  - Adds "Run build tests" step to multi-arch Linux and Windows workflows for compiler-runtime stage
  - Step is non-blocking (`continue-on-error: true`) with `TEATIME_FORCE_INTERACTIVE: 1` for visible output
  - `configure_stage.py` sets `-DTHEROCK_BUILD_TESTING=ON` for `compiler-runtime`
- CI invocation: `ninja build-test-amd-comgr` or `ninja therock-build-tests`

## Test Plan

The TheRock CI run.

## Test Result

- Linux: All Comgr tests pass -- 20/20 LIT tests, 32/32 CTest tests
- Windows: CTest passes (32/32), LIT tests fail (0/20) due to known upstream path separator issue pending amd-llvm submodule bump ([ROCm/llvm-project#1860](https://github.com/ROCm/llvm-project/pull/1860), fix is in `ww-2026-14+`)

image

## Submission Checklist


cmake/therock_subproject.cmake
ScottTodd, line 1: Please review https://github.com/ROCm/TheRock/pull/4790 first, instead of lifting the changes from that PR to your own branch with different authorship. You could also cherry-pick the commits from that PR branch and stack your changes on top for testing.

ScottTodd (dismissed):

cmake/therock_subproject.cmake
kasaurov: My understanding was that PR #4790 is a prototype to use, if it is going to be merged it makes the things even more simple. Just Approved.

kasaurov: The PR #4790 merged, the upstream changes merged into this PR. Please take another look.

compiler/pre_hook_amd-llvm.cmake
```diff
@@ -72,37 +72,6 @@ endif()
 if(THEROCK_BUILD_LLVM_TESTS OR THEROCK_BUILD_LLVM_TOOLS OR THEROCK_BUILD_COMGR_TESTS)
```
stellaraccident, line 72: Let's not use an implicit passthrough var here. Set what you need explicitly in the subproject. I am completely sure we are buying much by having so many test settings. Maybe just gate comgr tests behind the existing THEROCK_BUILD_LLVM_TESTS

cmake/therock_subproject.cmake
```diff
@@ -23,6 +23,7 @@ set_property(GLOBAL PROPERTY THEROCK_DEFAULT_CMAKE_VARS
   THEROCK_ROCM_LIBRARIES_SOURCE_DIR
   THEROCK_ROCM_SYSTEMS_SOURCE_DIR
   THEROCK_BUILD_TESTING
+  THEROCK_BUILD_COMGR_TESTS
```
stellaraccident: Don't add project specific global passthroughs here. If you need them, set them on the project declaration.

cmake/therock_subproject.cmake
```diff
@@ -1185,15 +1186,20 @@ function(therock_cmake_subproject_build_test target_name)
     message(FATAL_ERROR "Empty COMMAND in build tests for '${target_name}'")
   endif()
 
-  set(_build_test_commands)
+  # Generate a runner script that executes all test commands independently,
+  # so a failure in one does not prevent the others from running.
```
stellaraccident, line 1190: For the record, I started with a cmake script like this and some of the quoting became baroque so I did it how it was originally. Fine to switch to this approach but be warned.

cmake/therock_subproject.cmake
```diff
@@ -1185,15 +1186,20 @@ function(therock_cmake_subproject_build_test target_name)
...
   _therock_cmake_subproject_build_env_pairs(_build_env_pairs)
+  set(_runner_script "${_stamp_dir}/build-test-runner.cmake")
```
stellaraccident: The stamp dir is cleared on clean and other activities. In my prior version where I was using a generated script, this went into the prefix/ dir (adjacent to stamp/).

compiler/CMakeLists.txt
```diff
@@ -180,6 +184,23 @@ if(THEROCK_ENABLE_COMPILER)
   therock_cmake_subproject_provide_package(amd-comgr amd_comgr lib/cmake/amd_comgr)
   therock_cmake_subproject_activate(amd-comgr)
 
+  if(THEROCK_BUILD_COMGR_TESTS)
+    if(WIN32)
+      set(_llvm_lit_script "${CMAKE_BINARY_DIR}/compiler/amd-llvm/build/bin/llvm-lit.py")
```
stellaraccident, line 191: I assume you checked but I'm guessing that the linux version doesn't also have a llvm-lit.py so you could use the same for both? Annoying.

cmake/therock_subproject.cmake
kasaurov: Agreed, this approach may break if an argument contains double quotes, backslashes, semicolons, or other CMake special characters -- the naive quoting will break. We don't have such a case at the moment though. A more robust approach could escape special characters before embedding (e.g., `string(REPLACE)` for backslashes and quotes). Let's address it if we ever encounter such a case.

compiler/pre_hook_amd-llvm.cmake
kasaurov: Removed `THEROCK_BUILD_COMGR_TESTS` from the global passthrough list by https://github.com/ROCm/TheRock/pull/4881/changes/329cee5afcdc9b2229197367b6a6e4b3448415c3 ,  it's now passed explicitly via `CMAKE_ARGS` only to `amd-llvm` and `amd-comgr`. Keep the component level var for flexibility (COMGR is an independent sub-project under LLVM umbrella).

cmake/therock_subproject.cmake
kasaurov: Addressed by https://github.com/ROCm/TheRock/commit/329cee5afcdc9b2229197367b6a6e4b3448415c3

compiler/CMakeLists.txt
kasaurov: Yeah, understand that, but this matches LLVM's own `get_llvm_lit_path` in [AddLLVM.cmake](https://github.com/llvm/llvm-project/blob/436fbb81758bf43bab7d0047633ac8da203b1e57/llvm/cmake/modules/AddLLVM.cmake#L2180-L2185) -- on Windows the generated file is `llvm-lit.py` (needed for Python multiprocessing to find the main module), while on Linux it's `llvm-lit` (with a shebang):
```
  set(lit_file_name "llvm-lit")
  if (CMAKE_HOST_WIN32 AND NOT CYGWIN)
    # llvm-lit needs suffix.py for multiprocess to find a main module.
    set(lit_file_name "${lit_file_name}.py")
  endif ()
  set(${file_name} ${lit_file_name} PARENT_SCOPE)
  ```

File: cmake/therock_subproject.cmake (hunk 4 above)
Comment (kasaurov) [reply cmake/therock_subproject.cmake]: Addressed by https://github.com/ROCm/TheRock/pull/4881/changes/83c34029b3eaab176bb5e0dc282be5693f5f1a13

File: cmake/therock_subproject.cmake
Comment (ScottTodd) [reply cmake/therock_subproject.cmake:1]: Thanks. I dismissed my "request changes" review. I'll try to find time to review this PR soon.

File: .github/workflows/multi_arch_build_portable_linux_artifacts.yml [hunk 6]
```diff
@@ -155,6 +155,14 @@ jobs:
           python build_tools/memory_monitor.py --phase Build -- \
             cmake --build "${BUILD_DIR}" --target stage-${STAGE_NAME} therock-artifacts -- -k 0
 
+      - name: Run build tests
+        if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
+        continue-on-error: true
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
```
ScottTodd, line 164: Can we just add `therock-build-tests` to the same build command? That way the tests will run in parallel to other steps (that was one of the main reasosn to move the tests into the build system instead of running ctest / lit separately).

```yml
     - name: Build stage
       run: |
         python build_tools/memory_monitor.py --phase Build -- \
           cmake \
               --build "${BUILD_DIR}" \
               --target stage-${STAGE_NAME} therock-artifacts therock-build-tests \
               -- -k 0
```

> CI integration: 
> * Adds "Run build tests" step to multi-arch Linux and Windows workflows for compiler-runtime stage
> * Step is non-blocking (`continue-on-error: true`) with `TEATIME_FORCE_INTERACTIVE: 1` for visible output

If we collapse into the prior step then we can't get `--continue-on-error` behavior, but that's probably fine for these compiler tests, right? If those tests fail then we should get a visible error and not proceed.

Looks like test output is already logged to individual files, so finding the output in a sea of other build logs isn't that difficult:
* https://github.com/ROCm/TheRock/actions/runs/25154743498?pr=4881 --> https://therock-ci-artifacts.s3.amazonaws.com/25154743498-linux/logs/compiler-runtime/index.html
* https://therock-ci-artifacts.s3.amazonaws.com/25154743498-linux/logs/compiler-runtime/amd-comgr_build_test_1.log
* https://therock-ci-artifacts.s3.amazonaws.com/25154743498-linux/logs/compiler-runtime/amd-comgr_build_test_2.log

(why are there 2? what's different?)

---

On Linux (not Windows) this also reconfigured CMake: https://github.com/ROCm/TheRock/actions/runs/25154743498/job/73735501275?pr=4881#step:14:1
```
cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
[0/2] Re-checking globbed directories...
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
-- GLOB mismatch!
[1/2] Re-running CMake...
-- Configuring background job pool for 3 concurrent jobs
-- Build flags:
--   * KPACK_SPLIT_ARTIFACTS = ON (-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON)
--   * HIP_KERNEL_PROVIDER_ENABLE = OFF (-DTHEROCK_FLAG_HIP_KERNEL_PROVIDER_ENABLE=OFF)
-- Generated build topology targets from BUILD_TOPOLOGY.toml
-- Enabling building tests
-- ROCm version: 7.13.0
...
```

build_tools/configure_stage.py
```diff
@@ -161,6 +161,13 @@ def generate_cmake_args(
     for feature in sorted(features):
         args.append(f"-DTHEROCK_ENABLE_{feature}=ON")
 
+    # Stage-specific testing flags.
+    if stage_name == "compiler-runtime":
+        if include_comments:
+            args.append("")
+            args.append("# Enable build-time tests for compiler-runtime stage")
+        args.append("-DTHEROCK_BUILD_TESTING=ON")
```
ScottTodd: Is this needed? Doesn't testing default to enabled already?

compiler/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: MIT
 
 if(THEROCK_ENABLE_COMPILER)
+  if(NOT DEFINED THEROCK_BUILD_COMGR_TESTS)
+    set(THEROCK_BUILD_COMGR_TESTS ${THEROCK_BUILD_TESTING})
+  endif()
```
ScottTodd, line 7: We could use `THEROCK_BUILD_TESTING` directly below. If we want to disable tests for individual subprojects then I'd prefer a per-project CMake flag system like we have for `{project}_BUILD_TYPE` and `{project}_CMAKE_ARGS` (docs: https://github.com/ROCm/TheRock/blob/main/docs/development/development_guide.md#per-project-cmake-flags)

Or we could keep this specific, given
```
option(THEROCK_BUILD_LLVM_TESTS "Enables building LLVM LIT tests" OFF)
```

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kasaurov: >Can we just add therock-build-tests to the same build command? That way the tests will run in parallel to other steps (that was one of the main reasosn to move the tests into the build system instead of running ctest / lit separately).
- The main reasons to have unit tests in the build system are threefold: the tests are designed to run in-tree; not to copy huge and changeable amount of build/src artifacts and not to use precious GPU time to run CPU tests. I agree it would be better to run them in parallel on CPU as well.  But it gives very little gain for compiler unit tests. We still have to wait till LLVM is built (along with a few other components) -- which is the main resource eater. The tests themselves take very little time -- if .we don't have to copy artifacts they are pretty fast - 15 secs for COMGR tests and 3 mins for LLVM lit tests (not included in this PR).
- Not having a separate step for them makes them less visible in CI UI (you have to look into specific log files, as the build log is pretty large)
- There are two logs, because we have two sub-suites for COMGR tests, one - LIT tests, another CTests and they run using different test frameworks
- Eventually we want to make these tests blocking, but right now Windows LIT tests are failing and waiting for the fix to makes its way to TheRock (expected to land in the ww-2026-14). So making them blocking will block the entire Windows run at the moment (see https://github.com/ROCm/TheRock/actions/runs/25230470689/job/73985518570?pr=4881 )
- I agree that reconfiguration of CMake on Linux is a bummer, it is a price for having it as a separate Sub-step. This reconfiguration only happens because of GLOB changes between steps -- it's not inherent to having a separate step, but rather a side effect of how ninja detects filesystem changes.

So IMHO having a separate Sub-step gives more pros than cons, especially in terms of readability of the results and flexibility to make tests non-blocking.
I'd prefer to leave it as is (if there are no strong objections) and revisit if the number of tests increased and we would like to exploit benefits of CPU parallelization with the build (which at the moment are negligible).

build_tools/configure_stage.py
kasaurov: Yes, AFAICS it defaults to ON at the moment (hope this won't change) -- removed explicit settings for `compiler-runtime` by https://github.com/ROCm/TheRock/pull/4881/changes/18cf68b5b16988a5f3af9baa944ead033605dd61

compiler/CMakeLists.txt
kasaurov: - At least for now I'd like to keep `THEROCK_BUILD_COMGR_TESTS` it's used in Comgr test infrastructure.
- The CMake flag system doesn't handle the cross-subproject dependency and would require additional plumbing to make it work (enabling Comgr tests needs LLVM tools built in a different subproject (amd-llvm)). 
- The current `THEROCK_BUILD_LLVM_TESTS` var has an issue, setting it ON is not enough to run the tests, `THEROCK_BUILD_TESTING` should also be ON. So having it as a standalone top-level option could be misleading.
- IMHO the current approach for COMGR is a better choice -- we are explicitly defaulting to `THEROCK_BUILD_TESTING`. Even though there is no top-level option to toggle it independently, the variable makes explicit which specific testing it triggers. This improves readability and simplifies adding a separate option in the future if needed. 
- I will have a follow up PR to include LLVM unit testing and will address `THEROCK_BUILD_LLVM_TESTS` var there.

stellaraccident (approved):

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kasaurov: Regarding the GLOB mismatches all 8 are from `rocprofiler` components ( [log with additional debuggibg](https://productionresultssa13.blob.core.windows.net/actions-results/8859a5e8-59c9-47cc-a067-5a0dca2ccba6/workflow-job-run-4f4c79f0-6e20-5109-a506-2eb837ec0662/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-05-02T02%3A14%3A10Z&sig=CbtI2cuGwDFHf5pBeFT2qieYVUUpMJ9dBczQxmnNgbg%3D&ske=2026-05-02T04%3A24%3A51Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-05-02T00%3A24%3A51Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-05-02T02%3A04%3A05Z&sv=2025-11-05)):
```
rocm-systems/projects/rocprofiler-compute/./*.h
rocm-systems/projects/rocprofiler-sdk/./*.c
rocm-systems/projects/rocprofiler-sdk/./*.cc
rocm-systems/projects/rocprofiler-sdk/./*.cpp
rocm-systems/projects/rocprofiler-sdk/./*.h
rocm-systems/projects/rocprofiler-sdk/./*.hh
rocm-systems/projects/rocprofiler-sdk/./*.hpp
rocm-systems/projects/rocprofiler-sdk/./*.inc
```
So 7 of 8 are from `rocprofiler-sdk` and 1 from `rocprofiler-compute`. These are Linux-only components (not built on Windows), which explains why Windows doesn't see the reconfiguration.
The likely cause is that `rocprofiler-sdk` generates source files during its build (e.g., auto-generated headers or source files from code generation), and these new files appear in directories covered by the `therock_cmake_subproject_glob_c_sources(rocprofiler-sdk SUBDIRS .)` glob. When the second ninja invocation runs, it detects these new files.

ScottTodd: Ok for now, but silent failures are dangerous, particularly if we add more tests or other subprojects adopt the same pattern.

I could also see someone extending or copying this for another build stage and then we'd be paying the cost of serializing test execution _and_ the cost of more complex workflow code.

If another stage adds tests:
```diff
      - name: Run build tests
-       if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
+       if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' || inputs.stage_name == 'math-libs' }}
        continue-on-error: true
        env:
          TEATIME_FORCE_INTERACTIVE: 1
        run: |
          cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
```
or

```diff
-     - name: Run build tests
+     - name: Run math-libs build tests
-       if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
+       if: ${{ !cancelled() && inputs.stage_name == 'math-libs' }}
-       continue-on-error: true
        env:
          TEATIME_FORCE_INTERACTIVE: 1
        run: |
          cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
```

ScottTodd (approved):

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kasaurov: I understand your concern, I hope we will switch to blocking run very soon, let it just settle for a week or so.

ScottTodd:
This had an unexpected side-effect - disabling the build and install off `offload-arch`, breaking Windows testing: https://github.com/ROCm/TheRock/issues/5074

---

Agent analysis:

Root Cause: PR #4881 exposed a latent bug from PR #3429

The trigger is PR #4881 (54acd32c5, merged May 4) — "Add COMGR build-time unit tests to CI". It changed how `THEROCK_BUILD_COMGR_TESTS` defaults:

Before #4881: `option(THEROCK_BUILD_COMGR_TESTS ... OFF)` → always `OFF` in CI

After #4881: Removed the option(), replaced with:
```cmake
# compiler/CMakeLists.txt:5-6
if(NOT DEFINED THEROCK_BUILD_COMGR_TESTS)
    set(THEROCK_BUILD_COMGR_TESTS ${THEROCK_BUILD_TESTING})
endif()
```
Since `THEROCK_BUILD_TESTING` defaults to `BUILD_TESTING` which is `ON` after include(CTest), `THEROCK_BUILD_COMGR_TESTS` is now `ON` by default.

The mechanism

In pre_hook_amd-llvm.cmake:144:
```cmake
if(NOT THEROCK_BUILD_LLVM_TESTS AND NOT THEROCK_BUILD_LLVM_TOOLS AND NOT THEROCK_BUILD_COMGR_TESTS)
    # ... calls therock_set_implicit_llvm_options() which FORCE-sets
    # CLANG_TOOL_OFFLOAD_ARCH_BUILD=ON
```

- Before #4881: `THEROCK_BUILD_COMGR_TESTS=OFF` → condition is `TRUE` → `therock_set_implicit_llvm_options()` runs →
FORCE-sets `CLANG_TOOL_OFFLOAD_ARCH_BUILD=ON` → overrides the `CMAKE_ARGS` → offload-arch builds on Windows
- After #4881: `THEROCK_BUILD_COMGR_TESTS=ON` → condition is FALSE → function skipped → the latent `CMAKE_ARGS
-DCLANG_TOOL_OFFLOAD_ARCH_BUILD=OFF` from PR #3429 takes effect → offload-arch NOT built on Windows

The fix

The latent bug is compiler/CMakeLists.txt:104. offload-arch has nothing to do with Fortran/OpenMP offloading — it was incorrectly grouped with those flags. Remove it:

```diff
       -DCLANG_TOOL_CLANG_LINKER_WRAPPER_BUILD=${THEROCK_CONDITION_IS_NON_WINDOWS}
-      -DCLANG_TOOL_OFFLOAD_ARCH_BUILD=${THEROCK_CONDITION_IS_NON_WINDOWS}
       -DOPENMP_ENABLE_LIBOMPTARGET=${THEROCK_CONDITION_IS_NON_WINDOWS}
```

With this line removed, offload-arch will be ON regardless of whether the pre_hook tool-selection runs or not (LLVM defaults all tools to ON).

ScottTodd:
This also significantly increased binary size in the LLVM that we build and install, see
* before: https://github.com/ROCm/TheRock/actions/runs/25347135602 --> https://therock-ci-artifacts.s3.amazonaws.com/25347135602-windows/index.html, `amd-llvm_run_generic.tar.zst	566 MB`
* after: https://github.com/ROCm/TheRock/actions/runs/25348294930 --> https://therock-ci-artifacts.s3.amazonaws.com/25348294930-windows/index.html  --> `amd-llvm_run_generic.tar.zst	970 MB`

Do we want to continue filtering to a more minimal set of tools? This code path is now inactive: https://github.com/ROCm/TheRock/blob/88d5bde079ff7ce49f1d5062dfdbf272beadbb9d/compiler/pre_hook_amd-llvm.cmake#L142-L173

ScottTodd:
I think we should rework that condition and continue to filter the tools that we build. comgr tests shouldn't need much more than llvm-lit?

lamb-j:
> I think we should rework that condition and continue to filter the tools that we build. comgr tests shouldn't need much more than llvm-lit?

For Comgr we only need these:

https://github.com/ROCm/llvm-project/blob/amd-staging/amd/comgr/test-lit/lit.cfg.py#L35-#L41

For LLVM LIT tests, which is on deck to be added afaik, we need quite a few more tools (probably still not all of them though).

PR #4882, opened 2026-04-28
Updating the GitHub Actions outputs using gha_set_output

nunnikri:
## Motivation

Scripts in `build_tools/packaging/linux/` were writing GitHub Actions outputs by printing `key=value` to stdout and relying on callers to redirect with `>> $GITHUB_OUTPUT` or `>> $GITHUB_ENV`. Any unexpected print statement in the call chain can silently corrupt these files, causing hard-to-debug CI failures.

## Technical Details

**`get_url_repo_params.py`**
- Switch `sys.path.insert` to point to `build_tools/` and use `from github_actions.github_actions_api import gha_set_output` — consistent with other `build_tools/` scripts (e.g. `configure_stage.py`).
- Replace all `print(f"key=value")` calls in every `cmd_*` function with `gha_set_output({...})`, which writes directly to the `$GITHUB_OUTPUT` file.
- Add `get-container-image` subcommand with an `_OS_PROFILE_TO_IMAGE` table mapping OS profile prefixes to container images (`sles` → SUSE BCI 16, `ubuntu`/`debian` → `ubuntu:24.04`, default → RHEL UBI 10).

**`setup_python_cmd.sh`**
- In the `github` output format case, write `PYTHON_CMD=...` directly to `$GITHUB_ENV` instead of printing to stdout. Callers no longer need `>> $GITHUB_ENV`.

**Tests**
- `get_url_repo_params_test.py`: Replace `patch("sys.stdout")` in all `MainSubcommandsTest` cases with a `_run_main_with_output()` helper that uses a temp `$GITHUB_OUTPUT` file. Add `GetContainerImageTest` covering `get_container_image()` for all OS profile families and the new subcommand.
- `setup_python_cmd_test.sh`: Add `run_github_format()` helper that sets a temp `$GITHUB_ENV` file and reads from it, replacing direct stdout capture for all `--output-format github` test cases.

## Test Plan

- Run `get_url_repo_params_test.py` unit tests locally.
- Run `setup_python_cmd_test.sh` locally.

## Test Result

All 50 `get_url_repo_params_test.py` tests pass.
`setup_python_cmd_test.sh` github-format mechanism tests pass

### DIFF
--- a/build_tools/packaging/linux/get_url_repo_params.py
+++ b/build_tools/packaging/linux/get_url_repo_params.py
@@ -14,6 +14,7 @@ Subcommands (get operations):
   get-repo-sub-folder  Get repo_sub_folder from an S3 prefix (last segment if YYYYMMDD-<id>, else empty). Prints repo_sub_folder=<value>.
   get-repo-url         Get full repo URL from components(release_type, native_package_type, repo_base_url, os_profile, repo_sub_folder). Prints repo_url=<value>.
   extract-gfx-arch     Extract and normalize GPU architecture from artifact group. Prints gfx_arch=<value>.
+  get-container-image  Get container image for a given OS profile. Prints container_image=<value>.
 
 Usage:
   python build_tools/packaging/linux/get_url_repo_params.py get-base-url --from-url <url>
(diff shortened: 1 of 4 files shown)


arvindcheru (approved):
LGTM

PR #4317, opened 2026-04-02
[torch] Start building from the release/2.11 branch

WBobby:
## Motivation
PyTorch 2.11 is now generally available, and the https://github.com/ROCm/pytorch/tree/release/2.11 branch in the ROCm fork of pytorch is ready for us to build from.

## Technical Details
* Added `release/2.11` to the release wheel build matrices for both [Linux](/.github/workflows/release_portable_linux_pytorch_wheels.yml) and [Windows](/.github/workflows/release_windows_pytorch_wheels.yml)
* Added Python 3.14 include entry for `release/2.11` in the Linux release matrix (Windows already builds 3.14 for all refs)
* Updated the [README version support table](/external-builds/pytorch/README.md) to reflect the new 2.11 stable release:
  - Nightly row updated from "2.11 alpha" to "2.12 alpha"
  - New 2.11 stable row pointing to `ROCm/pytorch release/2.11` branch
  - Previous "2.11 alpha" marked as "Previously built"
  - Updated example checkout command to reference `release/2.11`

This follows the same pattern as https://github.com/ROCm/TheRock/pull/3049 (adding release/2.10 support).

## Test Plan
- [x] Trigger "Build Portable Linux PyTorch Wheels" workflow:
   https://github.com/ROCm/TheRock/actions/runs/23960887139
- [x] Trigger "Build Windows PyTorch Wheels" workflow
   https://github.com/ROCm/TheRock/actions/runs/24045363642  (gfx1152)
   https://github.com/ROCm/TheRock/actions/runs/24062784224/attempts/1  (gfx1151)
- [x] Verify wheels build successfully for at least one GPU family and Python version

## Test Result
- Linux PyTorch:
    + build: passed
    + test: failed  https://github.com/ROCm/TheRock/issues/3939 
- Windows PyTorch:
    + build: passed
    + test: did NOT run

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4894, opened 2026-04-28
[torch] Start building from the release/2.12 branch

ethanwee1:
## Motivation
PyTorch 2.12 is now generally available, and the https://github.com/ROCm/pytorch/tree/release/2.12 branch in the ROCm fork of pytorch is ready for us to build from.

## Technical Details
* Added `release/2.12` to the release wheel build matrices for both [Linux](/.github/workflows/release_portable_linux_pytorch_wheels.yml) and [Windows](/.github/workflows/release_windows_pytorch_wheels.yml)
* Added `release/2.12` to the new multi-arch release wheel build matrices for both [Linux](/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml) and [Windows](/.github/workflows/multi_arch_release_windows_pytorch_wheels.yml)
* Updated PyTorch auxiliary repo checkout helpers to read shared `ubuntu` rows from ROCm/pytorch `related_commits` files, matching the release/2.12 file layout
* Updated the [README version support table](/external-builds/pytorch/README.md) to reflect the new 2.12 stable release:
  - Nightly row updated from "2.12 alpha" to "2.13 alpha"
  - New 2.12 stable row pointing to `ROCm/pytorch release/2.12` branch
  - Previous "2.12 alpha" marked as "Previously built"
  - Updated example checkout command to reference `release/2.12`

This follows the same pattern as https://github.com/ROCm/TheRock/pull/4317 (adding release/2.11 support).

## Test Plan
- [x] Trigger targeted "Build Portable Linux PyTorch Wheels" workflow for `release/2.12`, Python 3.12, gfx94X-dcgpu, using nightly ROCm packages for `rocm==7.13.0a20260428`:
   https://github.com/ROCm/TheRock/actions/runs/25123876760
- [x] Trigger "CI Nightly PyTorch Full Test" with the exact `torch_version` produced by the previous build:
   https://github.com/ROCm/TheRock/actions/runs/25078208628
- [x] Validate `release/2.12` related commit pins resolve for `torchaudio`, `torchvision`, and `apex` using the `ubuntu` rows
We will merge this PR once the core pytorch UT's are passing.


.github/workflows/release_portable_linux_pytorch_wheels.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-        pytorch_git_ref: ["release/2.8", "release/2.9", "release/2.10", "release/2.11", "nightly"]
+        pytorch_git_ref: ["release/2.8", "release/2.9", "release/2.10", "release/2.11", "release/2.12", "nightly"]
```
ScottTodd, line 99: Do you know when we might be able to drop `release/2.8`? The full matrix of [python version] x [pytorch version] x [operating system] here is putting considerable load on our build and test runners and this is going to add more pressure. IIRC we're committed to supporting version N, N-1, N-2? This is now N, N-1, N-2, N-3, and N-4.

ScottTodd (approved):

.github/workflows/release_portable_linux_pytorch_wheels.yml
ScottTodd, line 99: Oh, please also update the new workflows:

* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_windows_pytorch_wheels.yml

Then we also need to update the workflows in rockrel @marbre :

* https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_portable_linux_pytorch_wheels.yml
* https://github.com/ROCm/rockrel/blob/main/.github/workflows/release_windows_pytorch_wheels.yml

external-builds/pytorch/pytorch_apex_repo.py
```diff
@@ -105,7 +105,7 @@ def add_common(command_parser: argparse.ArgumentParser):
         has_related_commit,
     ) = repo_management.read_pytorch_rocm_pins(
         args.torch_dir,
-        os="centos",  # Read pins for "centos" on Linux and Windows
+        os="ubuntu",  # Read shared Linux pins on Linux and Windows
```
HereThereBeDragons, line 108: we build on manylinux container, so centos/rhel based. is using ubuntu here safe?

jithunnair-amd: Honestly, we never really had different content for the Ubuntu vs CentOS rows. It was introduced with that format a long time ago, but never really used to differentiate

jithunnair-amd: Maybe in 7.14, we could remove the Ubuntu part or make it something more generic like "all_os", but I wouldn't call it high priority.

.github/workflows/release_portable_linux_pytorch_wheels.yml
ethanwee1: Updated the two multi arch files here: https://github.com/ROCm/TheRock/pull/4894/changes/b0f45e25d5998b791112bb8891eb4f79455c59f6
@ScottTodd should I make a PR to rockrel?

HereThereBeDragons:
i guess this got a bit lost?
and @ScottTodd  I think then we can drop 2.8 pytorch? but could be a separate pr.

ethanwee1:
We were waiting on 7.13 release otherwise the failures on the hud will be seen as release blockers but if we make that clear, I don’t think we should hold the PR

PR #4904, opened 2026-04-28
Copy ROCm wheels into whl-staging on multi-arch release

marbre:
PyTorch wheels in the multi-arch release flow upload directly to `v4/whl-staging` via `publish_pytorch_to_staging.py`, but each device wheel has a `Requires-Dist` on `rocm-sdk-device-{arch}`. Without ROCm wheels in the same staging index, pip cannot resolve those deps from whl-staging index alone.

This extends `publish_python_packages` to always copy into `whl` and `whl-staging`.

### DIFF
--- a/build_tools/github_actions/publish_rocm_to_release_buckets.py
+++ b/build_tools/github_actions/publish_rocm_to_release_buckets.py
@@ -18,7 +18,7 @@ Example with ``--run-id 12345 --platform linux --release-type dev``:
     s3://therock-dev-artifacts/12345-linux/tarballs/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
       -> s3://therock-dev-tarball/v4/tarball/therock-dist-linux-gfx94X-dcgpu-7.10.0.tar.gz
 
-    python (kpack split enabled):
+    python (kpack split enabled, --copy-to-staging):
 
     s3://therock-dev-artifacts/12345-linux/python/rocm-7.13.0.tar.gz
     s3://therock-dev-artifacts/12345-linux/python/rocm_sdk_core-7.13.0-py3-none-linux_x86_64.whl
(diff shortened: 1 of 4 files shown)


build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -127,6 +149,15 @@ def main(argv: list[str]) -> None:
...
     )
+    parser.add_argument(
+        "--copy-to-staging",
+        action="store_true",
+        default=False,
+        help="Also copy python packages to the corresponding -staging sibling "
+        "(v3/whl-staging or v4/whl-staging). Off by default. Use this when "
+        "PyTorch wheels are uploaded directly to the staging path and need to "
+        "resolve their rocm-sdk-device-* dependencies from the same index.",
+    )
```
ScottTodd: Can we invert this?
* Always upload to staging
* Choose to copy out of staging
  * If we don't want to block "promotion" out of staging on tests, we do it immediately
  * If we _do_ want to block "promotion" out of staging on tests, the test job is responsible for copying

I'd rather "whl" be a subset of "whl-staging", not the other way around.

marbre: In our legacy workflows we always pushed _ROCm_ to whl and whl-staging was rather optional. For PyTorch it was the other way around, so there is indeed some inconsistency. Let's resolve this, yeah.

ScottTodd: The current non-multi-arch release pipelines upload to v2-staging first, then v2: https://github.com/ROCm/TheRock/blob/0ecce00d9852b617473daf1fdc4e9eea0f01f073/.github/workflows/release_portable_linux_packages.yml#L288-L328

marbre: Okay, dropped the option. PTAL.

ScottTodd (approved):

PR #4905, opened 2026-04-28
[ci] Removing ctest timeout

geomin12:
After discussion with some math libs folks, we decided to remove the ctest subtest timeout, as these were hard-coded earlier and each test varies significantly in time.

The global timeout handles the timeout of the job: https://github.com/ROCm/TheRock/blob/e3bef8a43b95addccc3cd9490d5858de3d8ff050/.github/workflows/test_component.yml#L157

Changes:
- removing ctest timeout (per subtest level)


geomin12:
@jayhawk-commits i can send these changes over to monorepos after this lands

tony-davis (approved):
Looks good. Thank you.

PR #4907, opened 2026-04-28
Add Ruff PLW1514 (explicit encoding) to pre-commit and fix call sites

tony-davis:
## Summary

This implements the follow-up suggested on [#4695](https://github.com/ROCm/TheRock/pull/4695): enable Ruff rule **PLW1514** (explicit `encoding` for text-mode `open()` and pathlib text helpers) via `pyproject.toml` and a scoped **ruff-pre-commit** hook, then fix existing call sites under `build_tools/`, `tests/`, and `external-builds/` so checks pass.

## Commits

1. **Add Ruff PLW1514 check to pre-commit and pyproject** — root `pyproject.toml` (`preview` + `extend-select`) and `.pre-commit-config.yaml` hook (`--select PLW1514 --preview`, limited to first-party Python paths).
2. **Apply explicit UTF-8 encoding for text I/O (PLW1514)** — bulk `encoding="utf-8"` (and related) plus Black line wraps where the formatter touched the same files.

## Related

- Discussion and request to use Ruff for this: [#4695](https://github.com/ROCm/TheRock/pull/4695) (split from the UTF-8 GitHub event JSON change as agreed in that thread).

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_libhipcxx_hiprtc.py
+++ b/build_tools/github_actions/test_executable_scripts/test_libhipcxx_hiprtc.py
@@ -31,7 +31,7 @@ def load_rocm_version() -> str:
     """Loads the rocm-version from the repository's version.json file."""
     version_file = THEROCK_DIR / "version.json"
     logging.info(f"Loading ROCm version from: {version_file}")
-    with open(version_file, "rt") as f:
+    with open(version_file, "rt", encoding="utf-8") as f:
         loaded_file = json.load(f)
         return loaded_file["rocm-version"]
 
--- a/build_tools/packaging/python/generate_s3_index.py
+++ b/build_tools/packaging/python/generate_s3_index.py
@@ -91,7 +91,9 @@ def main(args):
     objects = get_objects(args.bucket, args.subdir)
     url = f"https://{args.bucket}.{args.endpoint}/{args.subdir}"
 
-    with sys.stdout if args.output == "-" else open(args.output, "w") as f:
+    with (
+        sys.stdout if args.output == "-" else open(args.output, "w", encoding="utf-8")
+    ) as f:
         f.write(
             textwrap.dedent(
                 """\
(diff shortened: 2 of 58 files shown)


tony-davis:

build_tools/packaging/python/generate_s3_index.py
```diff
@@ -91,7 +91,9 @@ def main(args):
     objects = get_objects(args.bucket, args.subdir)
     url = f"https://{args.bucket}.{args.endpoint}/{args.subdir}"
 
-    with sys.stdout if args.output == "-" else open(args.output, "w") as f:
+    with (
+        sys.stdout if args.output == "-" else open(args.output, "w", encoding="utf-8")
+    ) as f:
```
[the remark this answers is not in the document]
tony-davis: This is outside of the scope of this PR.

pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+    "rocm-systems",
+    "third-party",
+    "dockerfiles",
+]
+
+[tool.ruff.lint]
+# PLW1514: require explicit encoding for open() and pathlib text helpers.
+# Preview until the rule is stabilized; see https://docs.astral.sh/ruff/rules/unspecified-encoding/
+preview = true
+extend-select = ["PLW1514"]
```
ScottTodd, line 29: Can we add this `[tool.ruff.lint]` to new `pyproject.toml` files in `build_tools`, `tests`, and `external-builds` instead of introducing a new top level file with such an extensive deny-list? I don't want to need to edit this list whenever we add a new directory.

build_tools/github_actions/test_executable_scripts/test_libhipcxx_hiprtc.py
ScottTodd, line 1: fyi these `test_executable_scripts/` files are moving to:
* https://github.com/ROCm/rocm-systems/tree/develop/test/therock
* https://github.com/ROCm/rocm-libraries/tree/develop/test/therock

so we may want to make similar changes in those repositories

ScottTodd (commented):
Not thrilled about editing so many call sites, but the lint check looks fine. Once we're up to Python 3.15 as our minimum version (in like... 10 years) we can remove this since utf8 will be the new default :P

https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-utf8-default

PR #3301, opened 2026-02-06
[CI] Add manifest diff workflow (Part 2/3)

amd-hsivasun:
## Motivation

Add a reusable GitHub Actions workflow for generating manifest diff reports on-demand. This workflow enables both manual triggering and integration with other CI workflows.

[**Example Report**](https://therock-ci-artifacts.s3.amazonaws.com/23025588012-linux/logs/manifest-diff/TheRockReport.html)

This is Part 2/3 of the manifest diff report feature (split from #2161).
- Part 1: #3295 - Python scripts and tests
- Part 2: This PR - Reusable workflow
- Part 3: CI nightly integration (depends on this PR)

## Technical details

- **`.github/workflows/manifest-diff.yml`**
  - `workflow_dispatch` for manual runs (GitHub UI or CLI).
  - `workflow_call` for use by other workflows (e.g. `ci_nightly.yml`).
  - Single “Generate manifest diff report” step: all inputs are passed through to the script; the script validates combinations and reports clear errors for invalid input.
  - Supports the same four input combinations (see below); validation and behavior live in the script, not in workflow conditionals.
  - S3 upload of the generated HTML report via `upload_test_report_script.py` (when not from a fork).

- **`build_tools/github_actions/upload_test_report_script.py`**
  - Appends the S3 report URL to the job step summary via `gha_append_step_summary` so the report link appears in the Actions UI. Safe when not in GHA (no-op if `GITHUB_STEP_SUMMARY` is unset).

- **Script and template (supporting changes)**  
  - Script: optional args from the workflow (e.g. empty `--start` / `--find-last-successful`) are normalized and validated so invalid combinations fail with a clear message.  
  - Template: commit range arrow and section icons use HTML entities and `<meta charset="utf-8">` so the report displays correctly and avoids encoding issues (e.g. mojibake in the commit range line).

Input combinations supported:
- `find_last_successful` + `workflow_mode`: Find start from last successful, end as workflow ID
- `find_last_successful` only: Find start from last successful, end as commit SHA
- `workflow_mode` only: Both start and end as workflow IDs
- Neither: Both start and end as commit SHAs

## Test Plan

- [x] Trigger workflow manually with `find_last_successful=ci_nightly.yml` and hardcoded parameters
- [x] Verify report is generated and uploaded to S3
- [x] Verify step summary appears in GitHub Actions UI

## Test Result

[...]

PR #4908, opened 2026-04-28
[CI] Auto-generate manifest-diff report from multi-arch CI on PR/push

amd-hsivasun:
## Motivation

The manifest-diff report exists as a standalone reusable workflow, but nothing fires it automatically. Reviewers and bisecters who want to know "what submodules moved between these two TheRock commits" have to dispatch the workflow by hand or run the script locally.

Wiring it into Multi-Arch CI gives every PR, every push to a CI-relevant branch, and every manual dispatch a per-event report linked from the run's Step Summary. The job runs as a parallel sibling with no `needs:`, so it can never delay or fail the rest of the run; `continue-on-error: true` handles transient API failures.

### `multi_arch_ci.yml`

New top-level sibling job `manifest_diff` calling `./.github/workflows/manifest-diff.yml`. No `needs:`, no `if:`, no `with:` — the reusable workflow derives all refs from the caller's `github.event` directly, so the caller doesn't have to plumb event-specific inputs.

### `manifest-diff.yml`

Follows the same container + AWS-creds-mount pattern as `multi_arch_build_native_linux_packages.yml`:

- `runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}`
- `container: ubuntu:24.04` with `-v /runner/config:/home/awsconfig/`
- `AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini`
- AWS creds via the composite `./.github/actions/configure_aws_artifacts_credentials` (OIDC on base-repo runs, mounted baseline creds on fork PRs)
- `git config --global --add safe.directory $PWD` after checkout — required for `git config --blob` reads inside the container, where the working tree is host-UID-owned
- `continue-on-error: true` on the inner job — Compare/Actions API hiccups surface as a yellow `Manifest Diff` job rather than red-lighting the parent CI run

Job-level permissions: `id-token: write`, `contents: read`. Workflow-level is just `contents: read`.

Per-event ref derivation lives inside the workflow:

| Event               | `end_ref`                            | `start_ref`                                                                |
| ------------------- | ------------------------------------ | -------------------------------------------------------------------------- |
| `pull_request`      | `pull_request.head.sha`              | merge-base of `end_ref` and `pull_request.base.ref` via the Compare API    |
| `push`              | `github.sha`                         | `github.event.before` (previous tip, even on force-push)                   |
| `workflow_dispatch` | dispatcher's `end_ref` input         | one of `pr_base_ref` / `find_last_run` / `start_ref` (precedence below)    |

Input surface: `start_ref`, `end_ref`, `pr_base_ref`, `branch`, `find_last_run`, `workflow_mode`, `repository`, `ref`. All generic strings.

### Script (`build_tools/generate_manifest_diff_report.py`)

`--end` is always required. Start-resolution precedence in `resolve_commits()`:

1. `--pr-base-ref` (highest) — Compare API merge-base.
2. `--find-last-run` — Actions API lookup.
3. `--start` (lowest) — explicit SHA.

`--workflow-mode` is a separate modifier:

- Always re-interprets `--end` as a workflow run ID when set.
- Re-interprets `--start` only when `--pr-base-ref` and `--find-last-run` are both unset (i.e. it composes with the `--start` branch only).

| Flag                          | Role                                                                                                                                                              |
| ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `--end SHA`                   | End commit. Always required.                                                                                                                                     |
| `--pr-base-ref BRANCH`        | Resolves start as `merge_base(end, BRANCH)` via the Compare API. Rebase-safe.                                                                                    |
| `--find-last-run WORKFLOW.yml`| Resolves start as the head SHA of the most recent run of `WORKFLOW.yml` on `--branch` whose conclusion is `success` or `failure`. Exits 0 with no report if no match. |
| `--branch NAME`               | Branch scope for `--find-last-run` (default `main`).                                                                                                             |
| `--start SHA`                 | Explicit start commit.                                                                                                                                            |
| `--workflow-mode`             | See above.                                                                                                                                                        |

CI uses *PR-base mode* on `pull_request` and *explicit-SHA mode* on `push`. The PR/push paths always pass `--find-last-run` and `--branch`, but the script normalizes empty strings to "not provided", so they're effectively no-ops when `--pr-base-ref` is set.

### Path filter

`manifest-diff.yml` added to `_GITHUB_WORKFLOWS_CI_FILENAMES` in `build_tools/github_actions/configure_ci_path_filters.py` so workflow-only PRs still trigger CI.

### Docs

New `docs/development/manifest_diff.md` linked from `docs/development/ci_overview.md`.

```mermaid
flowchart TD
    A[multi_arch_ci.yml fires] --> B{event_name}
    B -- pull_request --> P[pr_base_ref = pull_request.base.ref<br/>end_ref = pull_request.head.sha]
    B -- push --> U[start_ref = github.event.before<br/>end_ref = github.sha]
    P --> G[generate_manifest_diff_report.py]
    U --> G
    G --> R[reports/TheRockReport.html]
    R --> O[upload_test_report_script.py<br/>→ index.html + S3]
    O --> H[step-summary link]
```

### Coverage matrix

| Scenario                                                 | Fires? | Report meaningful? |
| -------------------------------------------------------- | :----: | :----------------: |
| TheRock-internal PR                                      |  Yes   |        Yes         |
| TheRock fork PR (mounted baseline creds, OIDC skipped)   |  Yes   |        Yes         |
| Push to `main` / `multi_arch/**` / `release/therock-*`   |  Yes   |        Yes         |
| `rockrel` / release / nightly (`multi_arch_release.yml`) |   No   |        N/A         |

## Test Plan

**Local:**

```bash
pre-commit run --all-files
python3 -m pytest \
  build_tools/tests/generate_manifest_diff_report_test.py \
  build_tools/github_actions/tests/github_actions_api_test.py \
  build_tools/github_actions/tests/configure_ci_path_filters_test.py
```

pre-commit and pytest clean.

**End-to-end.** All four runs were captured against branch HEAD `17c516fc` (the merge integrating current `main`). The follow-up cleanup commit `b8dac593b` is YAML perms / apt-step grouping / one new unit test / one doc-line edit — no semantic change to the four code paths exercised here.

| Path                                                                                              | Trigger             | Run                                                                                | Resolved (start → end)                                                                                                                       | Report                                                                                                                                              |
| ------------------------------------------------------------------------------------------------- | ------------------- | ---------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| `workflow_call` from `multi_arch_ci.yml` (this PR)                                                | `pull_request`      | [26185715696](https://github.com/ROCm/TheRock/actions/runs/26185715696)            | `7666949b → 17c516fc` (merge-base via Compare API)                                                                                            | [report](https://therock-ci-artifacts.s3.amazonaws.com/26185715696-linux/logs/manifest-diff/TheRockReport.html)                                     |
| `workflow_call` from `multi_arch_ci.yml` (fork PR via [#5373](https://github.com/ROCm/TheRock/pull/5373), since closed) | `pull_request`      | [26183105151](https://github.com/ROCm/TheRock/actions/runs/26183105151)            | Validated against earlier branch tip `2695bbc5`; fork-PR credential path unchanged by the subsequent merge. Mounted baseline creds, not OIDC. | [report](https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-TheRock/26183105151-linux/logs/manifest-diff/TheRockReport.html)               |
| `workflow_dispatch` of `manifest-diff.yml` — explicit `pr_base_ref`                               | `workflow_dispatch` | [26185872602](https://github.com/ROCm/TheRock/actions/runs/26185872602)            | `7666949b → 17c516fc` — byte-identical report to the PR-event run                                                                             | [report](https://therock-ci-artifacts.s3.amazonaws.com/26185872602-linux/logs/manifest-diff/TheRockReport.html)                                     |
| `workflow_dispatch` of `manifest-diff.yml` — `find_last_run`                                      | `workflow_dispatch` | [26185874753](https://github.com/ROCm/TheRock/actions/runs/26185874753)            | `934512e0 → 17c516fc` — script resolved start to head SHA of last terminal-status `multi_arch_ci.yml` run on `main`                          | [report](https://therock-ci-artifacts.s3.amazonaws.com/26185874753-linux/logs/manifest-diff/TheRockReport.html)                                     |

All four reports show **12/12 + 28/28 + 32/32 unchanged** (top-level submodules / `rocm-libraries` components / `rocm-systems` components) — matches git ground truth: this PR only touches workflows, scripts, docs, and tests.

The matrix exercises PR-base merge-base, `find_last_run`, both AWS credential paths, and `workflow_call` vs `workflow_dispatch` wiring. The `push` / explicit-`--start` path is not covered by these four runs.

## Out of scope

Tracked in #5219.

1. **External-repo callers (`rocm-libraries`, `rocm-systems`).** `manifest-diff.yml` references local paths (`./.github/actions/configure_aws_artifacts_credentials`, `build_tools/generate_manifest_diff_report.py`), so a cross-repo `uses:` from rl/rs would also need to checkout TheRock at the right ref. The script also hardcodes the top-level repo as `ROCm/TheRock`, so an `rl`/`rs` PR would get a Compare API 404 or a misleading "no submodule changes" report. Fix: `--repository` plus an override-aware manifest mode that consumes `external_repo_config` from `setup_multi_arch.yml`.

2. **`rockrel` / release / nightly (`multi_arch_release.yml`).** Triggers are `workflow_call` + `workflow_dispatch`; neither has `pull_request.head.sha` or `github.event.before`, and `github.sha` resolves to the *caller's* commit (rockrel's), not a TheRock commit. Wiring requires explicit TheRock `start_ref`/`end_ref` (or a `find_last_run` baseline like "last successful nightly") plumbed through new `workflow_call` inputs.

3. **ASAN (`multi_arch_ci_asan.yml`).** Triggers are `schedule` + `workflow_dispatch` + `pull_request`. The PR arm ports directly; the `schedule` arm has no `github.event.before` or `pull_request.base.ref`, so a literal copy/paste resolves to an empty start. Right shape: PR mapping plus a `find_last_run`-based baseline (e.g. last successful nightly) for the `schedule` arm.

## Submission Checklist

- [x] Looked over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


.github/workflows/manifest-diff.yml
```diff
@@ -43,24 +45,60 @@ jobs:
   generate-report:
     name: Generate Manifest Diff Report
     runs-on: ubuntu-24.04
+    # On PR events, only run when the 'manifest-diff' label is present.
+    # workflow_dispatch / workflow_call events always run.
+    if: >-
+      github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'manifest-diff')
```
ScottTodd: We could also make this part of the pipeline setup that runs on all changes, regardless of labels.

* Call from https://github.com/ROCm/TheRock/blob/main/.github/workflows/setup_multi_arch.yml
* Diffbase against base branch for pull_request/push, previous run on this branch for schedule 
* Upload report to workflow outputs, add link to summary

I'd rather all of our workflows produce a predictable set of outputs.

build_tools/github_actions/bump_automation.py
```diff
@@ -78,6 +78,58 @@ def gh_api(token, endpoint, method="GET", data=None):
...
+    )
+    if response.ok:
+        print(f"[INFO] Created label '{name}' on {repo}")
+        return
+    if response.status_code == 422 and "already_exists" in response.text:
+        return
+    raise RuntimeError(
+        f"Failed to create label '{name}' on {repo}: "
+        f"{response.status_code} {response.text}"
+    )
```
ScottTodd: What are you doing here? Please discuss such changes prior to testing. Automated creation of repository labels (and any other side-effecting operations on the github repository that require elevated permissions) needs strong motivation and careful review. Here all this code should need is to look for an existing label (or possibly fail if that label does not exist).

ScottTodd (changes requested):
Design review before testing of such changes please. I see a few concerning things here (as before on the earlier changes - we should flush all of those during a detailed design review instead of going point by point during draft PR testing like this...)

build_tools/github_actions/bump_automation.py
amd-hsivasun: You're right it is unnecessary, I was initially under the belief I did not have permissions to create labels. But realized that was false and was working on removing this function.

.github/workflows/manifest-diff.yml
amd-hsivasun: Done — latest push does exactly this:
- Called from `setup_multi_arch.yml` as a new `manifest_diff` job (`needs: setup`).
- Diffbase: `pr_base_ref` merge-base via Compare API for  `pull_request`, `github.event.before` for `push`,
  `find_last_run = github.workflow` (success or failure) for `schedule`.
- Uploaded via `upload_test_report_script.py` → S3 + step summary link, same path as test reports.
- No labels, no PR-comment plumbing. Runs on every PR/push/schedule on `ROCm/TheRock`. `continue-on-error: true` so failures never block build/test.
  
There are 2 cases which I have left out and are to be addressed in a future PR. The two cases I left out are external-repo callers (`rocm-libraries` / `rocm-systems`) and the rockrel release driver (calling repo isn't TheRock, plus there's a two-layer ref resolution to figure out). Both need real design thought and as you requested I will have a review of that design before implmentation. I will update this thread with the link to the issue I will make to track the changes.

amd-hsivasun: Here is the new issue: https://github.com/ROCm/TheRock/issues/5219 It is still a WIP, however most of the design is there.

.github/workflows/setup_multi_arch.yml
```diff
@@ -159,3 +159,37 @@ jobs:
...
+    if: >-
+      github.repository == 'ROCm/TheRock' &&
+      inputs.external_repo_config == '' &&
+      (github.event_name == 'pull_request' ||
+       github.event_name == 'push' ||
+       github.event_name == 'schedule')
+    uses: ./.github/workflows/manifest-diff.yml
+    permissions:
+      contents: read
+      id-token: write
```
ScottTodd: This workflow file is not valid: https://github.com/ROCm/TheRock/actions/runs/25767403611

`The workflow is not valid. .github/workflows/multi_arch_ci.yml (Line: 67, Col: 3): Error calling workflow 'ROCm/TheRock/.github/workflows/setup_multi_arch.yml@40da071793a54c999d97a834d67feb607e6a4e96'. The nested job 'manifest_diff' is requesting 'id-token: write', but is only allowed 'id-token: none'.`

.github/workflows/setup_multi_arch.yml
```diff
@@ -159,3 +159,37 @@ jobs:
...
+  #   pull_request → pr_base_ref (merge-base via Compare API, rebase-safe);
+  #                  end_ref is the PR head SHA, not setup.outputs.ref
+  #                  (which is the synthetic refs/pull/N/merge SHA).
+  #   push         → start_ref = github.event.before (previous tip).
+  #   schedule     → find_last_run scoped to github.workflow;
+  #                  accepted_statuses="success,failure" so a failed
+  #                  nightly still has a baseline.
+  manifest_diff:
+    name: Manifest Diff
+    needs: setup
```
ScottTodd: Is this job fast enough to go here? setup currently takes < 5 seconds and this is directly on the critical path for all CI runs: https://github.com/ROCm/TheRock/blob/547ee73eab2373aa88b30702b2950bb976da47ad/.github/workflows/multi_arch_ci.yml#L66-L68

History on https://github.com/ROCm/TheRock/actions/workflows/manifest-diff.yml shows this action taking upwards of 2m30s sometimes. Looking at the logs on https://github.com/ROCm/TheRock/actions/runs/25329303564/job/74258499688, 2 minutes is spent around here (see `04:03` to `06:03`:
```
Mon, 04 May 2026 16:04:03 GMT ++ Exec [/home/runner/work/TheRock/TheRock]$ git ls-tree 64b9f0aeb10d41b16e18039600fe20783fdec898 -- rocm-libraries
Mon, 04 May 2026 16:04:03 GMT ++ Exec [/home/runner/work/TheRock/TheRock]$ git ls-tree 64b9f0aeb10d41b16e18039600fe20783fdec898 -- rocm-systems
Mon, 04 May 2026 16:04:03 GMT ++ Exec [/home/runner/work/TheRock/TheRock]$ git ls-tree 64b9f0aeb10d41b16e18039600fe20783fdec898 -- third-party/sysdeps/linux/amd-mesa/mesa-fork
Mon, 04 May 2026 16:06:03 GMT Found 12 submodules in manifest for 64b9f0ae
Mon, 04 May 2026 16:06:03 GMT 
Mon, 04 May 2026 16:06:03 GMT Found 12 submodules for END commit
```

Some ideas:
* Speed up the workflow/script
* Move this down a level in the workflow job graph to break that dependency link
* Generate a simple commit manifest in setup and upload, then generate the actual diff on the server side with a lambda function

.github/workflows/setup_multi_arch.yml
```diff
@@ -159,3 +159,37 @@ jobs:
...
+    if: >-
+      github.repository == 'ROCm/TheRock' &&
+      inputs.external_repo_config == '' &&
+      (github.event_name == 'pull_request' ||
+       github.event_name == 'push' ||
+       github.event_name == 'schedule')
+    uses: ./.github/workflows/manifest-diff.yml
```
ScottTodd: Every workflow that runs on pull requests *must* be compatible with PRs from personal forks (and soon also workflow calls from the rocm-libraries and rocm-systems repositories). I can't quite tell if https://github.com/ROCm/TheRock/blob/main/.github/workflows/manifest-diff.yml meets that bar yet. For example, the configure credentials step is not using https://github.com/ROCm/TheRock/blob/main/.github/actions/configure_aws_artifacts_credentials/action.yml. Can you test please?

ScottTodd (changes requested):
I can't really review this effectively until the workflows actually parse and run with logs to inspect. Please ensure that this minimum bar is met before requesting a review.

.github/workflows/setup_multi_arch.yml
amd-hsivasun: Yes for sure I will run some tests. Wanted to follow your advice of doing design review before testing.

.github/workflows/setup_multi_arch.yml
amd-hsivasun: I think moving it down a level is the best approach, and it can run independently of the setup as well. This will make sure not to halt CI in anyway.

ScottTodd (commented):
Please get a code review from another developer before me. I can't be the first pass reviewer on so many PRs.

(there are SIX reviewers on this PR)

.github/workflows/manifest-diff.yml
```diff
@@ -32,46 +46,72 @@ on:
...
     name: Generate Manifest Diff Report
-    runs-on: ubuntu-24.04
+    runs-on: azure-linux-scale-rocm
+    # Informational job: never fail the parent caller.
+    continue-on-error: true
     permissions:
       id-token: write
       contents: read
+    container:
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:702a5133851e6d1daf1207d2c9fbb01c2667914a5b6dc5a01faeb3ce66ea6421
```
geomin12, line 87: you can use an ubuntu container as it is smaller

.github/workflows/manifest-diff.yml
```diff
@@ -32,46 +46,72 @@ on:
...
+      options: -v /runner/config:/home/awsconfig/
+    env:
+      AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
geomin12, line 93:
```suggestion
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          repository: ${{ inputs.repository || github.repository }}
          ref: ${{ inputs.ref }}
```

that way, it'll work with other repos

.github/workflows/manifest-diff.yml
```diff
@@ -32,46 +46,72 @@ on:
...
         with:
           fetch-depth: 0
 
+      - name: Adjust git config
+        run: |
+          git config --global --add safe.directory $PWD
+          git config fetch.parallel 10
```
geomin12: this isn't needed, this is relevant to fetch_sources.py. can be removed

.github/workflows/manifest-diff.yml
```diff
@@ -32,46 +46,72 @@ on:
...
+
+      - name: Install python deps
+        run: pip install -r requirements.txt
```
geomin12: i think most of these dependencies are python native. should be able to remove

.github/workflows/multi_arch_ci.yml
```diff
@@ -115,6 +115,22 @@ jobs:
       contents: read
       id-token: write
 
+  manifest_diff:
+    name: Manifest Diff
+    if: github.repository == 'ROCm/TheRock'
```
geomin12: this can be removed! we would want this to work with forked repos as well, otherwise, the "fork-ify" work isn't necessary in `.github/workflows/manifest-diff.yml` (and from PR description)

.github/workflows/multi_arch_ci.yml
```diff
@@ -115,6 +115,22 @@ jobs:
...
+      contents: read
+      id-token: write
+    with:
+      end_ref: >-
+        ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+      start_ref: >-
+        ${{ github.event_name == 'push' && github.event.before || '' }}
+      pr_base_ref: >-
+        ${{ github.event_name == 'pull_request' && github.event.pull_request.base.ref || '' }}
+      branch: ${{ github.event.pull_request.base.ref || github.ref_name }}
```
geomin12: curious if we are able to pull this data from the manifest-diff.yml file itself, rather than this

build_tools/generate_manifest_diff_report.py
geomin12, line 1: i notice that the URL posted in github summary report is a link to aws bucket. could we just link directly to the html?

.github/workflows/manifest-diff.yml
```diff
@@ -16,10 +16,24 @@ on:
...
-        description: 'Workflow file to find last successful run (e.g., ci_nightly.yml)'
+      find_last_run:
+        description: 'Workflow file to find last run on the branch (e.g., ci_nightly.yml). Run status filter is set via accepted_statuses.'
         required: false
         type: string
+      accepted_statuses:
+        description: 'Comma-separated workflow run statuses accepted by --find-last-run (default: success).'
+        required: false
+        type: string
+        default: 'success'
```
geomin12: thinking about this. will this be needed? are the developers using this search for a last pass / fail?

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,109 @@
...
+| `start_ref`         | one of these three     | Start commit SHA (or run ID with `workflow_mode=true`).                                                                             |
+| `find_last_run`     | one of these three     | Workflow filename (e.g. `multi_arch_ci.yml`). The script queries the last run on the branch whose status is in `accepted_statuses`. |
+| `pr_base_ref`       | one of these three     | PR base branch name. The script resolves the merge-base between `end_ref` and `origin/<pr_base_ref>` via the GitHub Compare API.    |
+| `accepted_statuses` | no (default `success`) | Comma-separated workflow run conclusions accepted by `--find-last-run`.                                                             |
+| `branch`            | no (default `main`)    | Branch to scope `--find-last-run` lookups against.                                                                                  |
+| `workflow_mode`     | no (default `false`)   | Treat `start_ref` / `end_ref` as workflow run IDs instead of commit SHAs (resolves each to its `head_sha`).                         |
+
+Exactly one of `start_ref`, `find_last_run`, or `pr_base_ref` is required.
```
geomin12: let's make sure that the workflow dispatch trigger is clear enough for developers! they should be able to understand the fields without needing to use this docs (ideally)

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,109 @@
...
+| `--find-last-run <workflow>`          | Resolve `start` as the most recent run of `<workflow>` on `--branch` whose conclusion is in `--accepted-statuses`. |
+| `--accepted-statuses success,failure` | Comma-separated set of accepted run conclusions for `--find-last-run`. Default: `success`.                         |
+| `--pr-base-ref <branch>`              | Resolve `start` as the merge-base between `end` and `origin/<branch>` via the Compare API.                         |
+| `--workflow-mode`                     | Treat `--start` and `--end` as workflow run IDs and resolve each to its `head_sha`.                                |
+| `--branch <name>`                     | Branch to scope `--find-last-run` to. Default: `main`.                                                             |
+| `--output-dir <path>`                 | Output directory for `index.html`. Default: `reports`.                                                             |
+
+`GITHUB_TOKEN` (or any token with `public_repo` read scope) must be set in the environment for the GitHub API calls used by `--find-last-run`, `--pr-base-ref`, and `--workflow-mode`.
```
geomin12: could we just link them to the script file for the usage instructions?

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Scope
+
+The sibling job in `multi_arch_ci.yml` is gated by `if: github.repository == 'ROCm/TheRock'` and fires on that workflow's events: `pull_request`, `push`, and `workflow_dispatch`. The gate excludes downstream forks of TheRock that have Actions enabled; fork PRs *targeting* `ROCm/TheRock` are unaffected because they run in the base-repo context. External-repo override callers (`rocm-libraries`, `rocm-systems` via `setup_multi_arch.yml`) and the rockrel release-driver flow (via `multi_arch_release.yml`) currently produce no manifest-diff; extending the report to those callers is tracked in #5219.
```
geomin12: as mentioned, some things are contradicting as work in repo is for forked PRs to work, but it still is gated

geomin12 (commented):
also https://github.com/ROCm/TheRock/pull/4636 in Pr description isn't relevant FYI

.github/workflows/manifest-diff.yml
amd-hsivasun: The workflow is running into a a fatal --blob error so I have kept the safe.directory but removed the fetch parallel. https://github.com/ROCm/TheRock/actions/runs/26174698527/job/77001913903?pr=4908#step:7:27

.github/workflows/manifest-diff.yml
amd-hsivasun: Trimmed to the minimum needed to bootstrap a bare `ubuntu:24.04` container:

```yaml
- name: Install bootstrap packages
  run: |
    apt-get update
    apt-get install -y git ca-certificates curl unzip

- name: Install AWS CLI
  run: |
    ./dockerfiles/install_awscli.sh
```

**Bootstrap packages:**

- `git` — bare `ubuntu:24.04` doesn't ship git. `actions/checkout@v6` needs a real git binary to do a proper clone with `fetch-depth: 0`; without it, checkout falls back to a REST tarball and we lose the history needed for the submodule walks `generate_manifest_diff_report.py` does.
- `ca-certificates` — TLS trust for both `git` (HTTPS clone) and `curl` (`awscli.amazonaws.com`).
- `curl` + `unzip` — used by `./dockerfiles/install_awscli.sh` to download and extract the AWS CLI v2 zip.

**AWS CLI:** required because `upload_test_report_script.py` shells out to `aws s3 cp` to push the report. We use the existing repo helper `./dockerfiles/install_awscli.sh` rather than a third-party action so the install path stays consistent with the rest of TheRock's container-based jobs.

No `pip install -r requirements.txt` step — the script and the upload helper only use the Python stdlib + what `actions/setup-python` ships, so the only non-Python dependencies needed inside the container are the four apt packages above plus the AWS CLI.

.github/workflows/manifest-diff.yml
amd-hsivasun: Updated to ubuntu container

.github/workflows/multi_arch_ci.yml
amd-hsivasun: Removed, must have missed it when going over the changes.

.github/workflows/multi_arch_ci.yml
amd-hsivasun: Moved logic into manifest-diff.yml

.github/workflows/manifest-diff.yml
amd-hsivasun: Removed and simplified the logic to only look for pass/fail

docs/development/manifest_diff.md
amd-hsivasun: Provided more clarity on inputs in the workflow dispatch description. Also Simplified this portion of the doc.

docs/development/manifest_diff.md
amd-hsivasun: Removed this portion and linked to the script as reccomended.

build_tools/generate_manifest_diff_report.py
amd-hsivasun: This change would require some code changes to the upload report script which I believe, would be slightly out of scope of this PR. If this is still a request after I can have a follow up PR to address this.

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,76 @@
...
+The multi-arch CI top-level workflow ([`multi_arch_ci.yml`](../../.github/workflows/multi_arch_ci.yml)) hosts the `manifest_diff` job as a top-level sibling. The job has no `needs:`, runs in parallel with `linux_build_and_test` / `windows_build_and_test`, and calls `manifest-diff.yml` with no `with:` block — the reusable workflow derives the **start** and **end** refs from the caller's `github.event` itself, choosing differently depending on what triggered the run.
+
+```mermaid
+flowchart TD
+    A[multi_arch_ci.yml fires] --> B{event_name}
+    B -- pull_request --> P[pr_base_ref = pull_request.base.ref<br/>end_ref = pull_request.head.sha]
+    B -- push --> U[start_ref = github.event.before<br/>end_ref = github.sha]
+    P --> G[generate_manifest_diff_report.py]
+    U --> G
+    G --> R[reports/index.html]
```
rahulc-gh: non-bloking but for acccuracy update it to reports/TheRockReport.html

.github/workflows/multi_arch_ci.yml
```diff
@@ -115,6 +115,13 @@ jobs:
       contents: read
       id-token: write
 
+  manifest_diff:
+    name: Manifest Diff
+    uses: ./.github/workflows/manifest-diff.yml
+    permissions:
+      contents: read
+      id-token: write
+
```
erman-gurses, line 127: `multi_arch_ci.yml` currently triggers on `workflow_dispatch`, but the newly added `manifest_diff.md` explicitly says dispatching `multi_arch_ci.yml` is not a supported way to generate a manifest diff report and that the auto-fired `manifest_diff` job may fail yellow.

Should this job be guarded to only run for `push` / `pull_request` events where the refs are auto-derived correctly?

`if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}`

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,134 @@ on:
...
-        with:
-          aws-region: us-east-2
-          role-to-assume: arn:aws:iam::692859939525:role/therock-ci
+      - name: Configure AWS credentials for artifact uploads
+        if: ${{ always() }}
+        uses: ./.github/actions/configure_aws_artifacts_credentials
 
       - name: Upload Report to S3
-        if: ${{ github.repository == 'ROCm/TheRock' }}
+        if: ${{ always() }}
```
erman-gurses: Because this upload step runs with `always()`, it will also run if report generation fails or exits early before creating `reports/index.html`.

We can guard this on the report file existing. Otherwise the upload step may produce a second, confusing failure after the real failure.

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,134 @@ on:
...
 
 permissions:
   contents: read
 
 jobs:
   generate-report:
     name: Generate Manifest Diff Report
-    runs-on: ubuntu-24.04
+    runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
+    continue-on-error: true
```
erman-gurses: Since `manifest-diff.yml` can also be run directly via `workflow_dispatch`, should `continue-on-error: true` apply unconditionally here?

For auto-generated CI reports, non-blocking behavior makes sense. But for manual runs, this may hide real script/workflow regressions and make validation harder. Should this be limited to the caller/CI path, or handled by guarding the `manifest_diff` job in `multi_arch_ci.yml` instead?

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,134 @@ on:
...
     permissions:
       id-token: write
       contents: read
+    container:
+      image: ubuntu:24.04
+      options: -v /runner/config:/home/awsconfig/
+    env:
+      AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
```
rahulc-gh, line 90: do you actually need this ?? looks like we are using line 129 for aws config . Also we should be running this on AWS scale runner as azure one will be deprecated .

build_tools/generate_manifest_diff_report.py
```diff
@@ -214,42 +248,68 @@ def _optional_str(val: str | None) -> str | None:
...
+        end_sha = end
+
+    # Resolve start, in priority order: --pr-base-ref, --find-last-run,
+    # --workflow-mode, then plain --start.
+    if pr_base is not None:
+        compare_url = (
+            f"https://api.github.com/repos/{therock}/compare/{pr_base}...{end_sha}"
+        )
+        compare = gha_send_request(compare_url)
+        start_sha = compare.get("merge_base_commit", {}).get("sha")
```
erman-gurses, line 288: Has this path been tested with PRs from forks?

For same-repo PRs, `compare/{base}...{head_sha}` should usually work. For fork PRs, the head commit may not always be directly comparable through the base repo API endpoint. If fork PRs are in scope, we may need a fallback using local `git merge-base`, or pass/use the PR head repo/ref instead of only the raw SHA.

erman-gurses (commented):
Please see some added comments:

docs/development/manifest_diff.md
amd-hsivasun: Will update this

.github/workflows/manifest-diff.yml
amd-hsivasun: Yes this workflow is following the same convention as the other workflows for fork prs       AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini.
Regarding the runner all workflows are using the azure runner so I will continue to use the azure runner on this workflow so we can update all workflows at once in a subsequent PR.

.github/workflows/manifest-diff.yml
amd-hsivasun: Switched both the AWS-creds step and the upload step from `if: ${{ always() }}` to `if: ${{ steps.gen-report.conclusion == 'success' }}` (with `id: gen-report` added to the report step). A script failure now surfaces only the real error, no confusing "no such file" tacked on. The step-`conclusion` check is unaffected by the job-level `continue-on-error`.

.github/workflows/manifest-diff.yml
amd-hsivasun: Changed from unconditional `continue-on-error: true` to `continue-on-error: ${{ github.event_name != 'workflow_dispatch' }}`. CI invocations (push / pull_request via `multi_arch_ci.yml`) stay non-blocking; direct `workflow_dispatch` of this workflow is now strict so manual validation runs surface real script regressions as red.

build_tools/generate_manifest_diff_report.py
amd-hsivasun: Verified working end-to-end. Fork-PR test [`#5373`](https://github.com/ROCm/TheRock/pull/5373) run [`26183105151`](https://github.com/ROCm/TheRock/actions/runs/26183105151/job/77031623804) exercised this exact path:

- Script invoked with `--end "2695bbc54..."`, `--start ""`, `--pr-base-ref "main"` (the fork-PR branch).
- Compare API call `GET /repos/ROCm/TheRock/compare/main...2695bbc5` resolved the merge-base to `fc4e6be` (a real `main` commit).
- Log line: `Comparing commits: fc4e6be -> 2695bbc` → report generated successfully → uploaded to `therock-ci-artifacts-external`.

Works because GitHub mirrors fork-PR head SHAs into `refs/pull/N/head` on the base repo, so they're reachable through `ROCm/TheRock`'s compare endpoint as long as the PR is open. No fallback needed.

Please let me know if this is inline with the issue you were referencing. If not then I can look to implement the fall back you are referring to.

.github/workflows/multi_arch_ci.yml
amd-hsivasun: Added `if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}` on the `manifest_diff` caller job.

This is also in line with Scott's [original framing](https://github.com/ROCm/TheRock/pull/4908#discussion_r2092318210) where he enumerated `pull_request` / `push` / `schedule` as the trigger surface for this report — no mention of `workflow_dispatch`. `multi_arch_ci.yml` itself has no `schedule:` trigger, so the `push` / `pull_request` gate covers everything this workflow fires on today. Schedule coverage lives with the other caller workflows (rockrel release nightly, external-repo overrides) and is tracked as follow-up in [#5219](https://github.com/ROCm/TheRock/issues/5219).

If we ever want manifest-diff under `multi_arch_ci.yml` dispatch we'd need to plumb new inputs (start_ref / end_ref) through it; until then, manual runs go through direct `workflow_dispatch` of `manifest-diff.yml` which already has those inputs.

build_tools/generate_manifest_diff_report.py
erman-gurses: Yes, this is exactly what I was asking about.

My concern was whether compare/{base}...{head_sha} would remain valid for fork PRs when only the raw head SHA is available. The fork-PR validation run and the explanation about GitHub exposing the PR head through the base repository's refs/pull/*/head refs address that concern, so I don't think a fallback is necessary at this point.

It may be worth adding a short comment near this code path documenting that assumption, since it's not immediately obvious why fork PR SHAs are resolvable through the base repository compare endpoint.

build_tools/github_actions/github_actions_api.py
```diff
@@ -480,29 +480,32 @@ def gha_query_workflow_runs_for_commit(
...
     """
-    # Use GitHub API query parameters to pre-filter for successful runs on the specified branch
-    url = f"https://api.github.com/repos/{github_repository}/actions/workflows/{workflow_name}/runs?status=success&branch={branch}&per_page=100&sort=created&direction=desc"
+    if accepted_statuses is None:
+        accepted_statuses = {"success"}
+    url = (
+        f"https://api.github.com/repos/{github_repository}"
+        f"/actions/workflows/{workflow_name}/runs"
+        f"?branch={branch}&per_page=100&sort=created&direction=desc"
+    )
```
erman-gurses: Not blocker but suggestion - this only inspects the newest 100 runs. On an active branch, the correct matching run might be older than that, so None can be returned even though a valid run exists. Pagination might make this more reliable. Paginate until a matching run is found, or at least keep fetching pages until the API returns no more runs.
That would make the lookup reliable instead of "reliable only for the last 100 runs."

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,136 @@ on:
...
-        with:
-          aws-region: us-east-2
-          role-to-assume: arn:aws:iam::692859939525:role/therock-ci
+      - name: Configure AWS credentials for artifact uploads
+        if: ${{ steps.gen-report.conclusion == 'success' }}
+        uses: ./.github/actions/configure_aws_artifacts_credentials
 
       - name: Upload Report to S3
-        if: ${{ github.repository == 'ROCm/TheRock' }}
+        if: ${{ steps.gen-report.conclusion == 'success' }}
```
erman-gurses, line 122: One remaining edge case I'm wondering about: generate_manifest_diff_report.py now exits 0 when --find-last-run finds no prior matching run, which means gen-report can succeed without actually generating a report. Since the upload path is gated only on steps.gen-report.conclusion == 'success', could we end up attempting upload in that case as well? Would it make sense to distinguish "report generated" from "nothing to compare" before running the AWS/upload steps?

erman-gurses (commented):
Added extra comments

build_tools/github_actions/github_actions_api.py
amd-hsivasun: `gha_query_last_workflow_run` now paginates with `MAX_PAGES = 5` and `PER_PAGE = 100` (= up to 500 most-recent runs).The cap exists so a filter that matches nothing on a high-volume branch can't fan out into unbounded API calls.

.github/workflows/manifest-diff.yml
amd-hsivasun: Fixed in `generate_manifest_diff_report.py` by having `main()` exit non-zero in the "no comparison performed" branch (was `return 0`, now `return 1` with a clear stderr message). The existing `steps.gen-report.conclusion == 'success'` gate on the AWS-credentials and upload steps then naturally skips them.

erman-gurses (approved):
LGTM as an initial review. Please make sure re-test after the recent changes.

amd-hsivasun:
> LGTM as an initial review. Please make sure re-test after the recent changes.

Performed re-tests to validate the new changes.
Tested the red signal: https://github.com/ROCm/TheRock/actions/runs/26652387467
Tested new pagination changes: https://github.com/ROCm/TheRock/actions/runs/26652483431

Both produced the desired outcome.

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,85 @@
...
+
+The report is generated by [`build_tools/generate_manifest_diff_report.py`](../../build_tools/generate_manifest_diff_report.py) and uploaded to S3 by the [`manifest-diff.yml`](../../.github/workflows/manifest-diff.yml) reusable workflow.
+
+## How it runs in CI
+
+The multi-arch CI top-level workflow ([`multi_arch_ci.yml`](../../.github/workflows/multi_arch_ci.yml)) hosts the `manifest_diff` job as a top-level sibling. The job has no `needs:`, runs in parallel with `linux_build_and_test` / `windows_build_and_test`, and calls `manifest-diff.yml` with no `with:` block — the reusable workflow derives **start** and **end** refs from the caller's `github.event` itself.
```
ScottTodd: The specific code descriptions here are brittle - they would go out of date if the code changes even slightly. For example, ``The job has no `needs:` `` isn't providing any useful information. Same for ``and calls `manifest-diff.yml` with no `with:` block ``

Simplify by just including a link to the workflow where the manifest diff is integrated?

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,85 @@
...
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------ |
+| `pull_request` | `pr_base_ref` (the PR's base branch). The script calls the GitHub Compare API to get the merge-base, which is rebase-safe and works on rewritten PRs. | `pull_request.head.sha` — the tip of the PR's source branch. |
+| `push`         | `github.event.before` — the branch tip before the push.                                                                                               | `github.sha` — the new tip the branch was just moved to.     |
+
+### Trigger gating
+
+The `manifest_diff` caller job is gated on `github.event_name == 'pull_request' || github.event_name == 'push'`, the only two events whose `github.event` payloads carry the refs the reusable workflow auto-derives. A manual `workflow_dispatch` of `multi_arch_ci.yml` therefore skips the job entirely; for manual runs, dispatch [`manifest-diff.yml`](../../.github/workflows/manifest-diff.yml) directly (it exposes the full ref input surface).
```
ScottTodd: Similarly brittle documentation here. I'd restructure to just comment on the design intent and let the code show the specific implementation of that design.

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Failure behavior
+
+The reusable workflow sets `continue-on-error: ${{ github.event_name != 'workflow_dispatch' }}`:
+
+- **Under CI** (`push` / `pull_request` from `multi_arch_ci.yml`): non-blocking — a script or API failure shows yellow on the run summary and never gates the build/test jobs running in parallel.
+- **Under direct `workflow_dispatch`** of `manifest-diff.yml`: strict — real script regressions surface red, since the only reason to dispatch this workflow manually is to validate the script.
```
ScottTodd, line 33: Docs are brittle (and wrong) here too.

To get the job status to show up in the run summary you need to also edit

```yml
  ci_summary:
    name: CI Summary
    if: always()
    needs:
      - setup
      - linux_build_and_test
      - windows_build_and_test
```

See logs on this PR's run: https://github.com/ROCm/TheRock/actions/runs/26650189072/job/78625723984?pr=4908#step:3:33
```
Checking status for 3 job(s):
  setup: success
  linux_build_and_test: failure
  windows_build_and_test: failure

The following job(s) failed:
  linux_build_and_test
  windows_build_and_test
```

If you add the new job to `needs` it should show up as "Checking status for 4 job(s)" instead

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,85 @@
...
+
+The upload and AWS-credentials steps are both gated on the report-generation step succeeding, so a failed report produces only the real error rather than a follow-on "no such file" upload failure.
+
+### Where the report lives
+
+[`upload_test_report_script.py`](../../build_tools/github_actions/upload_test_report_script.py) uploads every `.html` file from `reports/` to S3 under the run's prefix (using `manifest-diff` as the "amdgpu family" segment), keeping filenames — so `TheRockReport.html` lands as `TheRockReport.html`. The script also generates an `index.html` listing page alongside the report and appends a link to it on `$GITHUB_STEP_SUMMARY`, surfacing the report in the run's **Summary** tab.
```
ScottTodd: Docs here are highlighting how the manifest diff is misusing the workflow outputs API. There should be no "amdgpu family segment" at all.

This code should be changed:
* https://github.com/ROCm/TheRock/blob/6fb7b2cca6a7c43b02137b29d9f7c2275fb1ed57/.github/workflows/manifest-diff.yml#L73-L80
* https://github.com/ROCm/TheRock/blob/6fb7b2cca6a7c43b02137b29d9f7c2275fb1ed57/build_tools/github_actions/upload_test_report_script.py#L123-L128

Don't just reuse one flag name for a different purpose, rename the flag name as needed or use a different API. (Here the least invasive change is to rename the flag from `--amdgpu-family` to something more generic like `--subfolder`)

Prior discussion: https://github.com/ROCm/TheRock/pull/3301#discussion_r2931790046

docs/development/manifest_diff.md
```diff
@@ -0,0 +1,85 @@
...
+
+S3 paths (base-repo runs):
+
+```
+s3://therock-ci-artifacts/{run_id}-linux/logs/manifest-diff/TheRockReport.html
+s3://therock-ci-artifacts/{run_id}-linux/logs/manifest-diff/index.html      (listing page; step-summary link target)
+```
+
+For fork PRs the bucket and prefix shift to `therock-ci-artifacts-external` and `<owner>-<repo>/{run_id}-linux/...` per `WorkflowOutputRoot`, and credentials route through [`configure_aws_artifacts_credentials`](../../.github/actions/configure_aws_artifacts_credentials/action.yml) — OIDC for base-repo runs, baseline credentials (mounted via `-v /runner/config:/home/awsconfig/`) for fork PRs on self-hosted runners. Direct `workflow_dispatch` on a fork running on a GitHub-hosted runner has no credentials available; the upload step will fail in that case. See [`workflow_outputs.md`](workflow_outputs.md) for the full S3 layout and [`s3_buckets.md`](s3_buckets.md) for the bucket/role mapping.
```
ScottTodd: More unnecessary detail here (we should be able to edit s3_buckets.md or workflow_outputs.md without also having to update all other files that interface with them, so being overly specific here is harmful)

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,136 @@ on:
...
+    container:
+      image: ubuntu:24.04
+      options: -v /runner/config:/home/awsconfig/
+    env:
+      AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
     steps:
+      - name: Install bootstrap packages
+        run: |
+          apt-get update
+          apt-get install -y git ca-certificates curl unzip
```
ScottTodd: Shouldn't need these "bootstrap packages"? Aren't these already available on either the base runner or in our docker container?

.github/workflows/manifest-diff.yml
```diff
@@ -4,74 +4,136 @@ on:
...
           fetch-depth: 0
 
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: '3.12'
+
+      - name: Install AWS CLI
+        run: |
+          ./dockerfiles/install_awscli.sh
```
ScottTodd: No awscli install. Use boto3 via python.

Update https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/upload_test_report_script.py as needed (or create a new script)

(We should delete that dockerifles/ install script now)

ScottTodd (commented):
User-visible behavior on the PR itself (https://github.com/ROCm/TheRock/actions/runs/26650189072/job/78546455577?pr=4908) looks good. Various comments on script style and repository conventions.

docs/development/manifest_diff.md
amd-hsivasun: Simplified the doc

docs/development/manifest_diff.md
amd-hsivasun: Added to the CI Summary and simplified doc

docs/development/manifest_diff.md
amd-hsivasun: Updated the flag

.github/workflows/manifest-diff.yml
amd-hsivasun: Switched to docker container and removed this step

.github/workflows/manifest-diff.yml
amd-hsivasun: Removed this step but left the switch from awscli to boto3 to a later PR. I believe it is out of scope

rahulc-gh (approved):

amd-hsivasun:
Fork PRs is failing on Upload seems like it is due to the changes in this PR https://github.com/ROCm/TheRock/pull/5450.
Here is the failure https://github.com/ROCm/TheRock/actions/runs/26847489027/job/79170788514?pr=5585#step:6:52 which is only failing after the new changes.

ScottTodd (approved):

PR #4909, opened 2026-04-28
Add local patch to kpack-device

marbre:
Applies https://github.com/ROCm/rocm-systems/pull/5555 locally until we get in the next rocm-systems bump.


ScottTodd (approved):

marbre:
CI failures are unrelated. This won't make anything worse thus not waiting for the final signal.

PR #4910, opened 2026-04-28
Fix RPM symbol version dependency resolution for sysdeps libraries

raramakr:
Fix issue where RPM dependency resolution failed because elfdeps was
   reporting symbol versions using the wrong library name.

   Problem:
     Post-install renaming created symlinks like libhwloc.so -> librocm_sysdeps_hwloc.so.5
     RPM elfdeps used the symlink name for symbol versions:
       libhwloc.so.5(AMDROCM_SYSDEPS_1.0)(64bit)  [WRONG]
     But packages required:
       librocm_sysdeps_hwloc.so.5(AMDROCM_SYSDEPS_1.0)(64bit)  [CORRECT]

   Solution:
     Rename libraries at build-time by patching Makefile.in/meson.build before build.
     Libraries are now built with rocm_sysdeps_ prefix from the start.
     RPM elfdeps now correctly reports:
       librocm_sysdeps_hwloc.so.5(AMDROCM_SYSDEPS_1.0)(64bit)

   Changes:
   - Add patch_source.sh scripts for hwloc and libpciaccess to rename at source level
   - Update CMakeLists.txt to apply source patches before build
   - Extract common utilities to build_tools/patch_linux_so.py
   - Simplify patch_install.py to use shared functions
   
Test Results:
[TestOutput.txt](https://github.com/user-attachments/files/27183341/TestOutput.txt)

### DIFF
--- a/third-party/sysdeps/common/hwloc/patch_source.sh
+++ b/third-party/sysdeps/common/hwloc/patch_source.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/bash
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+set -e
+
+SOURCE_DIR="${1:?Source directory must be given}"
+
+echo "Patching hwloc sources to rename main library..."
+
+# Patch the src/Makefile.in to change library name from libhwloc to librocm_sysdeps_hwloc
+HWLOC_MAKEFILE="$SOURCE_DIR/src/Makefile.in"
+
+if [ -f "$HWLOC_MAKEFILE" ]; then
+  echo "Patching $HWLOC_MAKEFILE..."
+  sed -i 's/libhwloc\.la/librocm_sysdeps_hwloc.la/g' "$HWLOC_MAKEFILE"
+  sed -i 's/libhwloc_la_SOURCES/librocm_sysdeps_hwloc_la_SOURCES/g' "$HWLOC_MAKEFILE"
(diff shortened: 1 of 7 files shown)


third-party/sysdeps/common/hwloc/patch_source.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/bash
```
nunnikri, line 1: Not a recommended choice. Better option is #!/usr/bin/env bash or #!/bin/bash

third-party/sysdeps/common/hwloc/patch_source.sh
```diff
@@ -0,0 +1,30 @@
...
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+set -e
```
nunnikri, line 5: Coding guidelines is asking to use set -euo pipefail

third-party/sysdeps/linux/libpciaccess/patch_source.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+set -e
```
nunnikri, line 5: Coding guidelines is asking to use set -euo pipefail

nunnikri (approved):
Overall looks good. Some small improvement comments

raramakr:
> Overall looks good. Some small improvement comments

In this PR, we added a common function to normalize the libraries. We plan to remove the corresponding logic from all other sysdeps components and use the common function instead. I will address the comments in that PR.

Issue #4848, opened 2026-04-27
Embedded HIP offload information from target-specific `dist/rocm` binaries is not present


kcossett-amd:
**Context**: 

I'm moving the `rocprofiler-systems` example suite to be built as a standalone that is target-specific.
The example binaries are used in our tests. Some of them are built with HIP offload information, and our test logic checks for that using `llvm-objdump --offloading`.

**Problem**: 

When `rocprofiler-systems-examples` is built through the target-specific portable distribution flow, the binaries under:
`output-linux-portable/build/dist/rocm`
no longer report offloading information with `llvm-objdump --offloading`.
For example:

```
(.venv) rocm@rocm:~/therock-branches/test-framework-update$ output-linux-portable/build/dist/rocm/lib/llvm/bin/llvm-objdump --offloading output-linux-portable/build/dist/rocm/share/rocprofiler-systems/examples/transpose

output-linux-portable/build/dist/rocm/share/rocprofiler-systems/examples/transpose:     file format elf64-x86-64
```

But when I check under `output-linux-portable/build/profiler/rocprofiler-systems-examples/...`
```
(.venv) rocm@rocm:~/therock-branches/test-framework-update$ sudo output-linux-portable/build/dist/rocm/lib/llvm/bin/llvm-objdump --offloading output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose

output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose:    file format elf64-x86-64
Extracting offload bundle: output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose.0.host-x86_64-unknown-linux-gnu-
Extracting offload bundle: output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose.0.hipv4-amdgcn-amd-amdhsa--gfx1100
Extracting offload bundle: output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose.0.hipv4-amdgcn-amd-amdhsa--gfx1101
Extracting offload bundle: output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose.0.hipv4-amdgcn-amd-amdhsa--gfx1102
Extracting offload bundle: output-linux-portable/build/profiler/rocprofiler-systems-examples/build/transpose/transpose.0.hipv4-amdgcn-amd-amdhsa--gfx1103
```

[...]

PR #4934, opened 2026-04-29
[rocprofiler-systems] Build rocprofiler-systems examples as standalone and switch to CTests

kcossett-amd:
## Motivation

We are moving to using CTest as the test executor.

We are also opting to build examples as a standalone and break it apart from `rocprofiler-systems`, reducing build times. Once https://github.com/ROCm/TheRock/issues/4848 is addressed, `rocprofiler-systems-examples` can be switched to `TARGET_SPECIFIC`. This should reduce build time.

This also corrects the behavior of tests + examples always being built. We now respect `THEROCK_BUILD_TESTING`

Should an example now need to be removed, append to `ROCPROFSYS_DISABLE_EXAMPLES` in the `CMAKE_ARGS` of `therock_cmake_subproject_declare(rocprofiler-systems-examples)`.

## Technical Details

 - Add `rocprofiler-systems-examples` as a project + add artifacts (`rocprofiler-systems` itself will no longer build examples)
 - Modify test script to use CTest instead
 - Edit the timeout for tests from 15 -> 60. Quick tests still run in ~11 minutes. Full tests run in ~29 minutes.

Once this PR is merged, the test script on `rocm-systems` side will need to be updated. Please see https://github.com/ROCm/rocm-systems/pull/5602

**NOTE A**

With this PR, `ROCPROFSYS_BUILD_FOR_THEROCK`'s new purpose is just to enforce "if `BUILD/INSTALL TESTING`, then do NOT imply build examples (as this is built by the new `rocprofiler-systems-examples` subproject)

**NOTE B**

I cannot have `rocprofiler-systems-examples` be built as target specific for now as the artifact splitter strips HIP offload information, meaning that `llvm-objdump --offloading` cannot detect the architectures that the binary was built for (which our tests rely on). This is not a problem in `TARGET_NEUTRAL` mode, so stick with that for now.

Note B Issue: https://github.com/ROCm/TheRock/issues/4848

## Test Plan

Run on TheRock

## Test Result

Most recent passing run: https://github.com/ROCm/TheRock/actions/runs/26462084819/job/77938442370?pr=4934

You'll notice some tests were skipped:
 - tests with `3.10`, `3.11`, `3.13` are skipped as the build only passed `3.12` to `rocprofiler-systems`.
 - `unit-tests`: For now, it is only ran when the project is built from `rocm-systems`, so this skip is expected. I'm not explicitly excluding this as this test may be enabled to run in install mode (thus enabling it would only require a change in `rocm-systems` side)

Windows workflows are failing, but it cannot be due to this PR as `rocprofiler-systems` can only be built and used on linux.

Using:
```
python3 build_tools/fetch_artifacts.py   --run-id 26462084819  --artifact-group gfx110X-all  --output-dir ./fetch_build   --flatten  '^rocprofiler-systems_(lib|run|dev|test|dbg)_'
```

Examples are not included, as one would expect, as they have been moved to `rocprofiler-systems-examples`. The `test` artifact only contain the generated CTest script, along with the packaged pytest code + some helper scripts used by tests. These take little time to generate and do not depend on the binary existing at **build** time (they need only be there at runtime).

## Submission Checklist


kcossett-amd:
https://github.com/ROCm/rocm-systems/pull/5214 - Blocks openmp-target validation tests
https://github.com/ROCm/rocm-systems/pull/5590 - May resolve certain fork + selective region test failures (may also reduce log size)
AIPROFSYST-418 - Blocks `matrix-exponential` tests

ScottTodd:
We should fix https://github.com/ROCm/TheRock/issues/4948 soon. This PR is running "full" tests for _all_ components, not just rocprofiler-systems, and that's using up lots of our runner capacity.

kcossett-amd:
Quick and full tests both pass

For full tests (https://github.com/ROCm/TheRock/actions/runs/25285876700/job/74138136774?pr=4934), it took 29 minutes. I'll increase timeout to 60 minutes, but quick tests run in ~11 minutes.

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
```diff
@@ -3,57 +3,119 @@
...
-environ_vars = os.environ.copy()
-ld_paths = [
-    # Libraries used by examples
-    rocm_base
-    / "share"
-    / "rocprofiler-systems"
-    / "examples"
-    / "lib",
+# Excluded by default (AIPROFSYST-441)
+EXCLUDED_LABELS = [
```
sputhala-amd, line 36: Adding several sources of truth for “what we skip in Rock CI", doesn't seem like an ideal solution.

Keeping ctest knobs like `--exclude-regex` / `--label-exclude` as optional overrides for users or one-off runs in CI via cmd line is reasonable. However we should not be passing different excludes via different options.

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
```diff
@@ -3,57 +3,119 @@
...
-    str(pytest_package_exec),
-    # TODO: Once the corresponding tests are fixed, remove the lines below
-    "-k",
-    "not TestOpenMPTarget and not (TestTranspose and runtime_instrument) and not TestGPUConnect",
-    "--junit-xml=junit.xml",
-    "--ci-mode",
-    "--log-cli-level=info",
+
+# Limited to 15 minutes
+QUICK_TESTS_REGEX = [
```
sputhala-amd, line 48: I would propose to use something like `THEROCK_ROCPROFSYS_TEST_PROFILE=quick|full`. We can have a curated list of tests in this `QUICK_TESTS_REGEX` which runs when `quick` mode is selected.

Optionally maintainers can exclude from this list using `--exclude-regex` and `--include-regex`.

profiler/CMakeLists.txt
```diff
@@ -306,10 +300,10 @@ if(THEROCK_ENABLE_ROCPROFV3)
         -DROCPROFSYS_BUILD_LIBIBERTY=ON
         -DROCPROFSYS_BUILD_BOOST=ON
         -DROCPROFSYS_BUILD_DYNINST=ON
-        -DROCPROFSYS_BUILD_FOR_THEROCK=ON # TODO: Will be deprecated
+        -DROCPROFSYS_BUILD_FOR_THEROCK=ON
+        -DROCPROFSYS_USE_PYTESTS=OFF
```
sputhala-amd: Do we still need this flag `ROCPROFSYS_USE_PYTESTS`?

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
```diff
@@ -3,57 +3,118 @@
...
+    "transpose.*",
+    "rocprofiler-systems.*",  # Binary tests
+    "config.*",
+    "jpegdecode.*",  # TODO: Binary is not built, so test is skipped
+    "videodeocde.*",  # TODO: Binary is not built, so test is skipped
```
marbre:
```suggestion
    "videodecode.*",  # TODO: Binary is not built, so test is skipped
```

Typo my agent spotted.

profiler/CMakeLists.txt
```diff
@@ -356,5 +351,51 @@ if(THEROCK_ENABLE_ROCPROFV3)
         rocprofiler-sdk
         rocprofiler-systems
       )
+
+
+      # TODO: This should be built as target-specific, but that's not currently possible as
+      # the artifact splitter strips per-arch HIP offload bundles out of the example binaries,
+      # preventing llvm-objdump (used in our tests) from detecting the offload architectures
+      # See https://github.com/ROCm/TheRock/issues/4848
+      if(${THEROCK_BUILD_TESTING})
```
marbre:
```suggestion
      if(THEROCK_BUILD_TESTING)
```

profiler/CMakeLists.txt
```diff
@@ -356,5 +351,51 @@ if(THEROCK_ENABLE_ROCPROFV3)
...
+          CMAKE_ARGS
+            -DROCPROFSYS_GFX_TARGETS="${_ROCPROFSYS_GFX_TARGETS_ESCAPED}"
+            -DROCPROFSYS_DISABLE_EXAMPLES=lulesh
+          INSTALL_RPATH_DIRS
+            "lib"
+            "lib/rocprofiler-systems"
+            "llvm/lib"
+          RUNTIME_DEPS
+            rocprofiler-systems
+            ${_rocprofsys_examples_optional_deps}
```
marbre:
```suggestion
```
This isn't set anywhere.

profiler/CMakeLists.txt
```diff
@@ -356,5 +351,51 @@ if(THEROCK_ENABLE_ROCPROFV3)
...
+        )
+        therock_cmake_subproject_activate(rocprofiler-systems-examples)
+
+        therock_provide_artifact(rocprofiler-systems-examples
+          TARGET_NEUTRAL
+          DESCRIPTOR "${CMAKE_SOURCE_DIR}/profiler/artifact-rocprofiler-systems-examples.toml"
+          COMPONENTS
+            test
+          SUBPROJECT_DEPS
+            rocprofiler-systems
```
marbre:
```suggestion
```
This only packages stuff from the example package does it?

profiler/CMakeLists.txt
```diff
@@ -295,7 +289,7 @@ if(THEROCK_ENABLE_ROCPROFV3)
     endif()
 
     therock_cmake_subproject_declare(rocprofiler-systems
-      USE_TEST_AMDGPU_TARGETS
+      USE_DIST_AMDGPU_TARGETS
```
marbre: Is rocprofiler-systems generic / host-only or does it still bundle GPU-specific code? If the latter, this might break in single-family builds. The artifact is marked `TARGET_NEUTRAL` (and therefore a `_generic` artifact is generated). In a single-family release, e.g. for `gfx942`, `THEROCK_DIST_AMDGPU_TARGETS` is normally not set and will fall back to `THEROCK_AMDGPU_FAMILIES`, here `gfx942`. If a single run builds multiple single-families, which is still the case in the old release workflows there might be a race and latest run wins.

TL;DR this is only safe if rocprofiler-systems is host-only.

marbre (commented):
Some further comments. This will need a rebase and a second round after conflicts have been resolved however.

profiler/CMakeLists.txt
kcossett-amd: With `examples` not being compiled in `rocprofiler-systems` anymore, it is generic. 
I had a commit to set this to `USE_DIST_AMDGPU_TARGETS`. It looks like I might have reverted it by mistake. Fixing...

Edit: I completely misread the git log, I thought this was a suggestion. Ignore this comment. 
But this should be safe because as mentioned, `rocprofiler-systems` is host-only

kcossett-amd:
This PR does not affect windows workflows. 
Windows math lib failure:
`FAILED: math-libs/hipCUB/stamp/stage.stamp B:/build/math-libs/hipCUB/stamp/stage.stamp `

Edit: Passing `rocprofiler-systems` "full" test workflow https://github.com/ROCm/TheRock/actions/runs/25552841928/job/75029215452?pr=4934

ScottTodd:
> This PR does not affect windows workflows.
> Windows math lib failure:
> `FAILED: math-libs/hipCUB/stamp/stage.stamp B:/build/math-libs/hipCUB/stamp/stage.stamp `

That's https://github.com/ROCm/TheRock/issues/5024, fyi

profiler/CMakeLists.txt
```diff
@@ -313,10 +307,10 @@ if (THEROCK_ENABLE_ROCPROFSYS AND THEROCK_SANITIZER STREQUAL "")
       -DROCPROFSYS_BUILD_LIBIBERTY=ON
       -DROCPROFSYS_BUILD_BOOST=ON
       -DROCPROFSYS_BUILD_DYNINST=ON
-      -DROCPROFSYS_BUILD_FOR_THEROCK=ON # TODO: Will be deprecated
+      -DROCPROFSYS_BUILD_FOR_THEROCK=ON
```
marbre, line 310: This drops the `Will be deprecated` comment. Fine for me but just to make sure, this change is intended?

profiler/CMakeLists.txt
```diff
@@ -313,10 +307,10 @@ if (THEROCK_ENABLE_ROCPROFSYS AND THEROCK_SANITIZER STREQUAL "")
...
+      -DROCPROFSYS_USE_PYTESTS=OFF
       -DROCPROFSYS_INSTALL_TESTING="${THEROCK_BUILD_TESTING}"
-      -DROCPROFSYS_INSTALL_EXAMPLES="${THEROCK_BUILD_TESTING}"
```
marbre, line 318: Should we hardcode to `OFF` instead?

profiler/CMakeLists.txt
```diff
@@ -326,8 +320,7 @@ if (THEROCK_ENABLE_ROCPROFSYS AND THEROCK_SANITIZER STREQUAL "")
     INSTALL_RPATH_DIRS
       "lib"
       "lib/rocprofiler-systems"
-      "lib/llvm/lib"
-      "lib/rocm_sysdeps/lib"
+      "llvm/lib"
```
marbre: Are you sure this should be `llvm/lib` and `lib/llvm/lib` as it was before?

profiler/CMakeLists.txt
```diff
@@ -365,4 +359,47 @@ if (THEROCK_ENABLE_ROCPROFSYS AND THEROCK_SANITIZER STREQUAL "")
...
+      EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR}/projects/rocprofiler-systems/examples"
+      BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/rocprofiler-systems-examples"
+      BACKGROUND_BUILD
+      CMAKE_ARGS
+        -DROCPROFSYS_GFX_TARGETS="${_ROCPROFSYS_GFX_TARGETS_ESCAPED}"
+        -DROCPROFSYS_DISABLE_EXAMPLES=lulesh
+      INSTALL_RPATH_DIRS
+        "lib"
+        "lib/rocprofiler-systems"
+        "llvm/lib"
```
marbre: As above, shouldn't that be `lib/llvm/lib`?

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
marbre, line 1: I would appreciate if someone else could review the changes in this file before landing.

marbre (approved):
Build changes look good but I am missing a link to the actual test results + a comparison in the pull request description. E.g. https://github.com/ROCm/TheRock/actions/runs/25552841928/job/75029215452 shows `100% tests passed, 0 tests failed out of 243` but it the test count still the same as before without skipping some due to the change? If links to job runs are provided as part of the PR description this would be way easier to review. Leaving an approval for the build system related changes but please let someone else look at the test script changes before landing. @geomin12 could be pulled in as an addition reviewer for this.

profiler/CMakeLists.txt
kcossett-amd: This is intended. See **NOTE A** in the PR description.

profiler/CMakeLists.txt
kcossett-amd: If you are referring to `ROCPROFSYS_INSTALL_TESTING`, then no. With `ROCPROFSYS_BUILD_FOR_THEROCK=ON`, `INSTALL_TESTING=ON` will only install tests and will not allow `EXAMPLES` to be built/installed.

profiler/CMakeLists.txt
kcossett-amd: Looks to be a mistake from when I was copying my work over from an older branch. Fixing to align with existing implementation.

FIM43-Redeye:
Spotted while merging locally: `${_rocprofsys_examples_optional_deps}` in the new `rocprofiler-systems-examples` `RUNTIME_DEPS` block (`profiler/CMakeLists.txt:382`) is referenced but not defined anywhere on the branch — CMake expands it to empty, so the effective `RUNTIME_DEPS` reduces to just `rocprofiler-systems`. Harmless for environments where the MPI-gated examples soft-skip via the example's own `find_library(roctx64/roctracer64)` → `return()` path (matches your test-plan note that `jacobi-roctx.*` binaries don't build). Reads like a TBD placeholder for the day jacobi-* and friends need real deps; flagging in case you'd like it populated (e.g. with `roctracer`) before merge.

kcossett-amd:
> Spotted while merging locally: `${_rocprofsys_examples_optional_deps}` in the new `rocprofiler-systems-examples` `RUNTIME_DEPS` block (`profiler/CMakeLists.txt:382`) is referenced but not defined anywhere on the branch — CMake expands it to empty, so the effective `RUNTIME_DEPS` reduces to just `rocprofiler-systems`. Harmless for environments where the MPI-gated examples soft-skip via the example's own `find_library(roctx64/roctracer64)` → `return()` path (matches your test-plan note that `jacobi-roctx.*` binaries don't build). Reads like a TBD placeholder for the day jacobi-* and friends need real deps; flagging in case you'd like it populated (e.g. with `roctracer`) before merge.

This is by design, and the undefined placeholder was removed in https://github.com/ROCm/TheRock/pull/4934/changes/3eb1f6c7f7fc8bd1331f4790a558ea88fba2c786.

For example, with `jacobi-hip` that wants to use `roctracer`, we are planning on moving it to use `rocprofiler-sdk-roctx`. Since that was next on my list of things todo, I opted to not add the `roctracer` dep.

Regarding MPI, any test won't work for now as a small change on `rocm-systems` side is required. So again, I opted to not add that dependency.

FIM43-Redeye:
Local test (gfx1102+gfx1103, GCC 15.2 host, system clang-21): `rocprofiler-systems-examples` configure fails on `find_package(OpenMP) ... Could NOT find OpenMP_C (missing: OpenMP_C_LIB_NAMES)`. Root cause is that the new `rocprofiler-systems-examples` subproject declaration (`profiler/CMakeLists.txt:379`) doesn't pass `COMPILER_TOOLCHAIN amd-hip`, so the generated `_toolchain.cmake` falls back to the top-level `/usr/bin/clang`. That clang doesn't have `omp.h` available unless the matching `libomp-N-dev` package is installed for that exact clang version, and it also doesn't pick up the in-tree HIP path additions that the parent `rocprofiler-systems` block (line 318) gets.

Fix is one block — adding it between `CMAKE_ARGS` and `INSTALL_RPATH_DIRS` matches the parent declaration:

```cmake
      COMPILER_TOOLCHAIN
        amd-hip
```

Verified locally: the regenerated `_toolchain.cmake` now overrides `CMAKE_C_COMPILER` to the in-tree amd-llvm clang and appends `--hip-path` / `--hip-device-lib-path` to `CMAKE_CXX_FLAGS_INIT`, identical to the parent rocprofsys toolchain. Examples are HIP/ROCm code so this seems like the architecturally consistent choice.

kcossett-amd:
> Local test (gfx1102+gfx1103, GCC 15.2 host, system clang-21): `rocprofiler-systems-examples` configure fails on `find_package(OpenMP) ... Could NOT find OpenMP_C (missing: OpenMP_C_LIB_NAMES)`. Root cause is that the new `rocprofiler-systems-examples` subproject declaration (`profiler/CMakeLists.txt:379`) doesn't pass `COMPILER_TOOLCHAIN amd-hip`, so the generated `_toolchain.cmake` falls back to the top-level `/usr/bin/clang`. That clang doesn't have `omp.h` available unless the matching `libomp-N-dev` package is installed for that exact clang version, and it also doesn't pick up the in-tree HIP path additions that the parent `rocprofiler-systems` block (line 318) gets.
> 
> Fix is one block — adding it between `CMAKE_ARGS` and `INSTALL_RPATH_DIRS` matches the parent declaration:
> 
> ```cmake
>       COMPILER_TOOLCHAIN
>         amd-hip
> ```
> 
> Verified locally: the regenerated `_toolchain.cmake` now overrides `CMAKE_C_COMPILER` to the in-tree amd-llvm clang and appends `--hip-path` / `--hip-device-lib-path` to `CMAKE_CXX_FLAGS_INIT`, identical to the parent rocprofsys toolchain. Examples are HIP/ROCm code so this seems like the architecturally consistent choice.

Good catch. Locally, this was falling back to GCC, so `find_package(OpenMP)` resolved through GCC’s OpenMP support and configured successfully. That hid the clang/libomp issue you hit. I’ll add `COMPILER_TOOLCHAIN amd-hip` to the examples subproject.

kcossett-amd:
Most recent run had two failure (https://github.com/ROCm/TheRock/actions/runs/25925738549/job/76230639465) followed by a complete timeout of the test run.
 - `roctx-binary-rewrite` (Timeout)
 - `openmp-fortran-offload-runtime-instrument` (Timeout)
 
 Binary-rewrite + instrument tests are taking much longer than expected.
 
 Moving to draft for investigation

kcossett-amd:
Looks like failing tests were due to heavy load on the runner machine. I was unable to reproduce locally and rerunning the tests on CI resulted in success: https://github.com/ROCm/TheRock/actions/runs/26103157485/job/76857850951?pr=4934

I will note that in the failing run, `binary-rewrite` tests took much longer than expected (from the original ~2 seconds to ~50 seconds). Based on this fact, and the fact that I could not reproduce locally or on subsequent CI run, I will assume load issue.

Edit: Moving from Draft to Open and requesting review from @geomin12 for the `test_rocprofiler_systems.py` changes.

profiler/artifact-rocprofiler-systems.toml
```diff
@@ -24,8 +24,8 @@ exclude = [
 include = [
   "share/doc/rocprofiler-systems/**"
 ]
+# Test binaries lie under rocprofiler-systems-examples artifact
```
jbonnell-amd, line 27: `share/rocprofiler-systems/tests/**` is still needed here, it's just the examples that got moved right?

Just wanted to confirm that the test component in `rocprofiler-systems` itself is still needed.

profiler/CMakeLists.txt
```diff
@@ -365,4 +360,49 @@ if (THEROCK_ENABLE_ROCPROFSYS AND THEROCK_SANITIZER STREQUAL "")
...
+    endif()
+
+    therock_cmake_subproject_declare(rocprofiler-systems-examples
+      USE_TEST_AMDGPU_TARGETS
+      EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR}/projects/rocprofiler-systems/examples"
+      BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/rocprofiler-systems-examples"
+      BACKGROUND_BUILD
+      CMAKE_ARGS
+        -DROCPROFSYS_GFX_TARGETS="${_ROCPROFSYS_GFX_TARGETS_ESCAPED}"
+        -DROCPROFSYS_DISABLE_EXAMPLES=lulesh
```
jbonnell-amd, line 384: Why are we disabling lulesh examples here? Just curious

kcossett-amd: This has been here since when examples were built months ago. IIRC it fails to build because of kokkos I believe (I had observed many compilation errors when trying to build this). 

Since it only affects a small subset of tests, I opted to disable it for now until it can be fixed.

profiler/artifact-rocprofiler-systems.toml
kcossett-amd: The test scripts are build as part of `rocprofiler-systems`, and lie in `share/rocprofiler-systems/tests/` folder. 
This includes the `CTesttestfile.cmake`, the `pyz`, and the various scripts.

jbonnell-amd (approved):

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
kcossett-amd: Reviewed by @jbonnell-amd

kcossett-amd:
My attempt to make test execution quicker using CTest `parallel` flag did not work. The system is under too much load and it ends up taking longer than if no parallelism was used in the first place (our tests timeout).

I will keep the `--repeat until-pass:3`. This matches our local CI workflows.

Sharding, for now, cannot be used, as we have a test that depends on the execution of another test, and if they end up in different shards, the dependee will fail. Luckily, its only a single test that does this, and I'm sure this behavior can be changed in the future to allow us to use sharding.

kcossett-amd:
All Linux workflows pass. Our tests took 30 minutes (`full` preset).

Windows workflow failed in the `gfx110X-all` sanity check with error:
```
error: Failed to install: requests-2.33.0-py3-none-any.whl (requests==2.33.0)
  Caused by: The wheel is invalid: Metadata field Name not found
```

This is unrelated to the changes in this PR.

marbre:
@kcossett-amd, fyi, some "comments" sneaked into your commit message:

image

Please make sure to clean those from the PR description and commit message in the future.

kcossett-amd:
> Please make sure to clean those from the PR description and commit message in the future.

My mistake and thank you for informing me.

profiler/CMakeLists.txt
kcossett-amd: It is used to say "don't use pytests directly, instead, go through CTests". Once I remove the code that allows pytest to be called directly in `rocm-systems`, the flag can then be removed.

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
kcossett-amd: As of now, TheRock is the source of truth for tests that can/cannot be ran. I believe once DevOps completes their transition to moving these test scripts to `rocm-systems` side, that will be treated as the source of truth and this file will be deleted.

PR #4923, opened 2026-04-29
Revert "[ml-libs] Remove all-or-nothing CK disable logic; let MIOpen self-filter (#4838)"

marbre:
This reverts commit 20dfb972a5a410e4e6273901f0ee1544538670b0.

The revert breaks the build for gfx900, gfx906 and gfx101X-dgpu on Windows and Linux as observed in the nightly build pipeline:
https://github.com/ROCm/rockrel/actions/runs/25089093160

```
[composable_kernel configure] CMake Error in /__w/rockrel/rockrel/build/ml-libs/composable_kernel/build/CMakeFiles/CMakeScratch/TryCompile-FndcnI/CMakeLists.txt:
[composable_kernel configure]   HIP_ARCHITECTURES is empty for target "cmTC_60590".
[composable_kernel configure] 
[composable_kernel configure] 
[composable_kernel configure] CMake Error in /__w/rockrel/rockrel/build/ml-libs/composable_kernel/build/CMakeFiles/CMakeScratch/TryCompile-FndcnI/CMakeLists.txt:
[composable_kernel configure]   HIP_ARCHITECTURES is empty for target "cmTC_60590".
[composable_kernel configure] 
[composable_kernel configure] 
[composable_kernel configure] CMake Error at /usr/local/therock-tools/share/cmake-3.27/Modules/CMakeDetermineCompilerABI.cmake:57 (try_compile):
[composable_kernel configure]   Failed to generate test project build system.
[composable_kernel configure] Call Stack (most recent call first):
[composable_kernel configure]   /usr/local/therock-tools/share/cmake-3.27/Modules/CMakeTestHIPCompiler.cmake:29 (CMAKE_DETERMINE_COMPILER_ABI)
[composable_kernel configure]   CMakeLists.txt:214 (enable_language)
[composable_kernel configure] 
[composable_kernel configure] 
[composable_kernel configure] -- Configuring incomplete, errors occurred!
```

With the revert, the math-libs stages pass for those targets in a dev build: https://github.com/ROCm/rockrel/actions/runs/25100364100

As an alternative new targets could be enabled in TheRock's filter logic.

PR #4838, opened 2026-04-26
[ml-libs] Remove all-or-nothing CK disable logic; let MIOpen self-filter

BradPepersAMD:
## Summary

Cleans up the TheRock-side CK/MIOpen wiring now that MIOpen self-filters `GPU_TARGETS` per-arch (see ROCm/rocm-libraries#6839).

Previously, if any GPU target in the build lacked CK grouped conv support, TheRock disabled composable kernel entirely for the whole MIOpen build. This silently lost CK kernels on supported arches in mixed multi-arch builds.

**Changes:**
- Remove the all-or-nothing arch validation block that set `THEROCK_ENABLE_COMPOSABLE_KERNEL=OFF` when any target was unsupported
- Remove the now-redundant `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` dependent option
- Replace `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` refs with `THEROCK_ENABLE_COMPOSABLE_KERNEL` in the MIOpen build dep and cmake args
- Pass `-DMIOPEN_USE_COMPOSABLEKERNEL=THEROCK_ENABLE_COMPOSABLE_KERNEL` so MIOpen matches what we've setup in TheRock for CK
- Bump `rocm-libraries` submodule to ROCm/rocm-libraries#6839 branch tip so CI validates both changes together

## Dependency

> [!IMPORTANT]
> This PR depends on ROCm/rocm-libraries#6839 landing first. The submodule will be bumped to the merged SHA before this is merged.

## Test plan

- [ ] `THEROCK_AMDGPU_FAMILIES=gfx942` — `MIOpenCKGroupedConv_gfx942.so` built as before
- [ ] Mixed build with unsupported arch (e.g. `gfx942;gfx1100`) — `MIOpenCKGroupedConv_gfx942.so` built, `gfx1100` skipped with status message, no build failure
- [ ] `THEROCK_ENABLE_COMPOSABLE_KERNEL=OFF` — `-DMIOPEN_USE_COMPOSABLEKERNEL=OFF` forwarded, no CK libs built

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4836, opened 2026-04-26
Stop force-disabling MIOPEN_USE_COMPOSABLEKERNEL when family contains arches without CK support


BrianHarrisonAMD:
## Problem

[`ml-libs/CMakeLists.txt:11-23`](https://github.com/ROCm/TheRock/blob/main/ml-libs/CMakeLists.txt#L11-L23) currently flips **both** `THEROCK_ENABLE_COMPOSABLE_KERNEL` *and* `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` to `OFF` if *any* arch in `THEROCK_AMDGPU_TARGETS` is missing from the CK allowlist:

```cmake
set(_ck_supported_gfx_targets gfx908 gfx90a gfx942 gfx950 gfx1150 gfx1151 gfx1152 gfx1153 gfx1200 gfx1201)
foreach(_gfx_target ${THEROCK_AMDGPU_TARGETS})
  if(NOT ${_gfx_target} IN_LIST _ck_supported_gfx_targets)
    message(WARNING "...Disabling composable kernel and it's use in MIOpen.")
    set(THEROCK_ENABLE_COMPOSABLE_KERNEL OFF)
    set(THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL OFF)
  endif()
endforeach()
```

A single unsupported arch in a multi-arch family disables CK *for the entire MIOpen build* AND skips building the CK subproject altogether — silently dropping CK kernels on supported arches in the same family.

**Both flag flips are wrong:**

- `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL OFF` → tells MIOpen "don't use CK at all," dropping CK kernels even on supported arches.
- `THEROCK_ENABLE_COMPOSABLE_KERNEL OFF` → skips the CK subproject build entirely, even though CK already filters its own input via `rocm_check_target_ids`.

## Concrete impact

| `THEROCK_AMDGPU_FAMILIES` | Supported subset | Current outcome |
|---|---|---|
| `gfx94X-all` | gfx942 | CK on (correct) |
| `gfx110X-all` | none | CK off (correct) |
| **`dgpu-all`** | gfx942/gfx950/gfx1200/gfx1201 | **CK silently off everywhere** |
| **User-set `gfx942;gfx1100`** | gfx942 | **CK silently off for gfx942** |

The `dgpu-all` case is the most realistic in-tree hazard.

## Why no per-arch partition is needed in TheRock

[...]

PR #4941, opened 2026-04-29
[ml-libs] Per-arch CK gating + gfx1100/1101/1102 support

BrianHarrisonAMD:
﻿## Summary

- Replace the all-or-nothing `THEROCK_ENABLE_COMPOSABLE_KERNEL` gate with per-arch filtering. CK is built when at least one requested arch is CK-supported; otherwise the subproject is skipped and MIOpen builds with `MIOPEN_USE_COMPOSABLEKERNEL=OFF`.
- Extend CK support to gfx1100/gfx1101/gfx1102 while keeping gfx1103 excluded until MIOpen adds support.

Motivation: #4838 removed the global flip but left single-arch CK-unsupported shards (gfx906, gfx101X-dgpu, gfx103X, gfx1103) trying to configure CK with an empty `GPU_TARGETS`, which made CK's HIP try-compile fail with `HIP_ARCHITECTURES is empty`. That regression caused revert #4923. This PR lets #4838 stand by both adding the missing arches to `EXCLUDE_TARGET_PROJECTS` and gating both the CK subproject *and* MIOpen's CK references on the post-filter target set, while broadening CK coverage to the new gfx1100/gfx1101/gfx1102 parts.

Closes path to addressing https://github.com/ROCm/TheRock/issues/4836.

## Test plan

- [x] CI green on multi-arch families that include at least one CK-supported arch (CK should build)
- [x] CI green on single-arch CK-unsupported shards (CK should be skipped, MIOpen should build with CK off)
- [x] CI green on existing CK-supported single-arch shards (no behavior change)


BradPepersAMD (approved):

cmake/therock_amdgpu_targets.cmake
```diff
@@ -220,6 +226,7 @@ therock_add_amdgpu_target(gfx1102 "AMD RX 7700S/Framework Laptop 16" FAMILY dgpu
 therock_add_amdgpu_target(gfx1103 "AMD Radeon 780M Laptop iGPU" FAMILY igpu-all gfx110X-all gfx110X-igpu
   EXCLUDE_TARGET_PROJECTS
     hipSPARSELt # https://github.com/ROCm/TheRock/issues/2042
+    composable_kernel # https://github.com/ROCm/TheRock/issues/4836
```
HereThereBeDragons, line 199: gfx1100 - gfx1102 have support for ck?
they dont have ck yet it in `EXCLUDE_TARGET_PROJECTS`
if they have support than the pr description should be adjusted

HereThereBeDragons (commented):
with `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` removed this means that it is impossible to now build miopen with ck disabled on an arch that supports ck.
is that some use case developers need?

cmake/therock_amdgpu_targets.cmake
BrianHarrisonAMD: Yea they have support. 1103 might as well, but we had hit issues with it previously so it's disabled in MIOpen for now.

BrianHarrisonAMD:
> with `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` removed this means that it is impossible to now build miopen with ck disabled on an arch that supports ck. is that some use case developers need?

Well I think you can still bulk disable CK to get this behavior, but if you wanted CK on, and CK off for just MIOpen that no longer exists. I think it would be fine to keep the option though since we can combine it with the other flag.

BrianHarrisonAMD:
@HereThereBeDragons Good call — restored `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` (defaults ON, depends on `THEROCK_ENABLE_COMPOSABLE_KERNEL`) and ANDed it with the new per-arch `_ck_enabled` gate. A user can still build CK for other consumers but disable it just for MIOpen. Pushed in b94800d3.

HereThereBeDragons (approved):
lgtm, but please update the pr description relating to gfx110X (more or less what you added in the comment).

you might want to ask @marbre for a final review

HereThereBeDragons:
and wait for ci of course :)

BrianHarrisonAMD:
> lgtm, but please update the pr description relating to gfx110X (more or less what you added in the comment).
> 
> you might want to ask @marbre for a final review

ty ty will do.

cmake/therock_amdgpu_targets.cmake
BrianHarrisonAMD: Updated now!

marbre (approved):
A rocm-libraries bump landed recently, eventually you want to rebase and run CI again to make sure that nothing interferes but otherwise lgtm.

BrianHarrisonAMD:
We are going to merge this.
Breaks are unrelated to changes, and we have signal that the previous breaks are successful now.

Issue #3326, opened 2026-02-09
Refactor `build_tools/index_generation_s3_tar.py` to move server side


marbre:
The `build_tools/index_generation_s3_tar.py` needs to be refactored to then move server side. Currently the script is called in workflow runs at

https://github.com/ROCm/TheRock/blob/e267c57b9a21c28fc75872793418c32703e2821d/.github/workflows/release_portable_linux_packages.yml#L316-L319

and

https://github.com/ROCm/TheRock/blob/e267c57b9a21c28fc75872793418c32703e2821d/.github/workflows/release_windows_packages.yml#L334-L337

Instead it must be refactored to be executed server side via an AWS Lambda.

ScottTodd:
As part of this, we should ensure that workflow logs no longer include step summary logging that is confusing.

For example: https://github.com/ROCm/TheRock/actions/runs/21851998754

> [Build Logs](https://therock-nightly-artifacts.s3.amazonaws.com/21851998754-windows/logs/gfx110X-all/index.html)
> 
> [Artifacts](https://therock-nightly-artifacts.s3.amazonaws.com/21851998754-windows/index-gfx110X-all.html)
> 
> [TheRock Manifest](https://therock-nightly-artifacts.s3.amazonaws.com/21851998754-windows/manifests/gfx110X-all/therock_manifest.json)
> 
> Detected GPU families (11): gfx101X, gfx103X, gfx110X, gfx1150, gfx1151, gfx1152, gfx1153, gfx120X, gfx90X, gfx94X, gfx950
> 
> Found 3017 .tar.gz files in bucket 'therock-nightly-tarball'.
> 
> index.html generated successfully for bucket 'therock-nightly-tarball'. File saved as index.html
> 
> index.html successfully uploaded. URL: https://therock-nightly-tarball.s3.amazonaws.com/index.html

The last 4 lines there  should be removed (or only be included in server-side logging)

PR #4943, opened 2026-04-29
Decouple index_generation_s3_tar from GitHub Actions

marbre:
Make the script callable from an AWS Lambda by dropping the dependency on the GitHub Actions step-summary helper and switching the upload path to `put_object` so the function does not need a writable local filesystem. The CLI surface is unchanged: arguments, default bucket, and the local-only mode (without `--upload`) all behave as before.

This is a prep step for the tarball index Lambda that lives in the infra repo. The Lambda imports `generate_index_s3` from this module and calls it with `upload=True` on each `.tar.gz` upload event.

Progress on #3326

### DIFF
--- a/build_tools/index_generation_s3_tar.py
+++ b/build_tools/index_generation_s3_tar.py
@@ -3,11 +3,11 @@
 # SPDX-License-Identifier: MIT
 
 """
-Script to generate an index.html listing .tar.gz files in an S3 bucket, performing the following:
- * Lists .tar.gz files in the specified S3 bucket.
- * Generates HTML page with sorting and filtering options
- * Saves the HTML locally as index.html
- * Uploads index.html back to the same S3 bucket
+Generate an index.html listing .tar.gz files in an S3 bucket.
+
+Usable as both a CLI tool (for local inspection) and an importable library
+(used by the tarball-index AWS Lambda handler in TheRock-Infra). The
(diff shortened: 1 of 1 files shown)


cgoea (approved):
LGTM

build_tools/index_generation_s3_tar.py
```diff
@@ -225,15 +214,30 @@ def generate_index_s3(s3_client, bucket_name, prefix: str, upload=False):
...
 
+        region = s3_client.meta.region_name or "us-east-2"
+        if region == "us-east-2":
+            bucket_url = f"https://{bucket_name}.s3.amazonaws.com/{upload_key}"
+        else:
+            bucket_url = f"https://{bucket_name}.s3.{region}.amazonaws.com/{upload_key}"
+        log.info("index.html successfully uploaded. URL: %s", bucket_url)
+        return bucket_url
+
+    # Local-only mode: write next to the caller's working directory.
```
HereThereBeDragons, line 227: from my understanding of the description above i wouldnt have thought that the index.html is always written locally.

With upload=True the index is PUT to s3://bucket_name/<prefix>/index.html
    and the HTTPS URL is returned. >>>Otherwise<<< the index is written to
    ./index.html and the local path is returned

HereThereBeDragons (approved):

PR #4944, opened 2026-04-29
[test-only] Validate CLR comgr unbundling change with re-applied Comgr commit

lamb-j:
## Purpose

**DO NOT MERGE.** This is a test PR to validate that ROCm/rocm-systems#1709 (CLR side) works in combination with re-applying the previously-reverted Comgr commit `eb04fc540`. CI here exercises the combined change against TheRock build/test matrix.

## Background

- `eb04fc540` ([Comgr] Remove silent undbundle path from `LINK_BC_TO_BC`) was merged into `amd-staging`, then reverted in `afd27c2a` because callers (CLR) were passing bundles into `LINK_BC_TO_BC` and depended on the silent unbundle path.
- The revert message explicitly noted the plan: *"We should patch up CLR to forbid passing bundles to LINK_BC_TO_BC, then reapply this."*
- ROCm/rocm-systems#1709 by @shadidashmiz is the CLR-side change. This PR pairs the two so CI can validate them together.

## Patches added

| File | Source | Notes |
|---|---|---|
| `patches/amd-mainline/llvm-project/0013-Comgr-Remove-silent-undbundle-path-from-LINK_BC_TO_BC.patch` | `eb04fc540` from `ROCm/llvm-project` | Re-applies the original commit. Applies cleanly (pin contains both the commit and the revert, net-zero). |
| `patches/amd-mainline/rocm-systems/0003-CLR-Separate-unbundling-from-LINK_BC_TO_BC.patch` | `de92a426dd` from rocm-systems#1709 | Required manual conflict resolution against the rocm-systems pin (file refactored since PR was authored). |

## Conflict resolution detail (rocm-systems patch)

`projects/clr/hipamd/src/hip_comgr_helper.cpp` `LinkProgram::AddLinkerDataImpl` was refactored in the pin since PR #1709 was based:

- `link_data` is now `std::string_view` (was `vector<char>&`)
- New `llvm_code_object_storage` (vector) + `llvm_code_object_view` (string_view) split
- New `is_bundled_` flag from `helpers::CheckIfBundled(link_data)`
- The trailing `else { copy }` was removed (fall-through keeps `llvm_code_object_view = link_data`)

I ported Shadi's new branch as follows:
- Chained as `else if` after the existing SPIRV branch (matching its pattern)
- Writes into `llvm_code_object_storage` (was `llvm_code_object`)
- Sets `llvm_code_object_view` after the unbundle (was missing — needed because the storage/view split is new)
- Reuses the existing `is_bundled_` flag in the predicate
- Dropped the trailing `else { copy }` (no longer needed)

@shadidashmiz please verify the resolved branch matches your intent — see `patches/amd-mainline/rocm-systems/0003-CLR-Separate-unbundling-from-LINK_BC_TO_BC.patch` in this PR.

The CMakeLists conflict was trivial (two list additions in the same place; both kept).

## References

- rocm-systems PR: https://github.com/ROCm/rocm-systems/pull/1709
- Original Comgr commit: https://github.com/ROCm/llvm-project/commit/eb04fc540e7d26dadc91cc5be2fa9bc2404b95a2
- The revert: https://github.com/ROCm/llvm-project/commit/afd27c2a1b6c17de2868c29c409cd49fb1e355f7

cc @shadidashmiz @kjayapra-amd


lamb-j:
CLR Patch landed, need to follow up with the Comgr fix with a PR to ROCm/llvm-project

PR #4946, opened 2026-04-29
Refresh release status for multi-arch python and tarballs

ScottTodd:
## Motivation

We're getting closer to supporting all items in multi-arch releases, this updates the docs with the current status.

Progress on:
* https://github.com/ROCm/TheRock/issues/3323

New content:
* Release status and install instructions for PyTorch packages
* Updated `device-*` extras table referencing https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html
* Documentation for the `device-all` extra (and a warning about the "whl" index vs the "whl-staging" index)
* Documentation for installing multi-arch tarballs
* Placeholders for multi-arch native linux package docs

## Submission Checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -124,36 +124,38 @@ rocm-sdk test
 
 For packages which include device-specific code (such as `rocm`, `torch`, and
 `torchvision`), support for individual devices can be installed using the
-corresponding `device-*` extra from the table below.
+corresponding `device-*` extra from the table below. See also the
+[GPU architecture specs](https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html)
+for a full list of supported AMD GPUs.
 
-| Product Name                       | GFX Target | Device Extra     |
-| ---------------------------------- | ---------- | ---------------- |
-| MI350X/MI355X                      | gfx950     | `device-gfx950`  |
-| MI300A/MI300X                      | gfx942     | `device-gfx942`  |
-| MI210/MI250                        | gfx90a     | `device-gfx90a`  |
-| MI100                              | gfx908     | `device-gfx908`  |
-| Radeon VII / MI50                  | gfx906     | `device-gfx906`  |
-| Vega 10 / MI25                     | gfx900     | `device-gfx900`  |
-| AMD RX 9070 / XT                   | gfx1201    | `device-gfx1201` |
-| AMD RX 9060 / XT                   | gfx1200    | `device-gfx1200` |
-| AMD Radeon 820M iGPU               | gfx1153    | `device-gfx1153` |
-| AMD Krackan 1 iGPU                 | gfx1152    | `device-gfx1152` |
-| AMD Strix Halo APU                 | gfx1151    | `device-gfx1151` |
-| AMD Strix Point iGPU               | gfx1150    | `device-gfx1150` |
-| AMD Radeon 780M Laptop iGPU        | gfx1103    | `device-gfx1103` |
-| AMD RX 7700S / Framework Laptop 16 | gfx1102    | `device-gfx1102` |
-| AMD RX 7800 XT                     | gfx1101    | `device-gfx1101` |
-| AMD RX 7900 XTX                    | gfx1100    | `device-gfx1100` |
-| AMD RX 6800 / XT                   | gfx1030    | `device-gfx1030` |
-| AMD RX 6700 / XT                   | gfx1031    | `device-gfx1031` |
-| AMD RX 6600                        | gfx1032    | `device-gfx1032` |
-| AMD Van Gogh iGPU                  | gfx1033    | `device-gfx1033` |
-| AMD RX 6500 XT                     | gfx1034    | `device-gfx1034` |
-| AMD Radeon 680M Laptop iGPU        | gfx1035    | `device-gfx1035` |
-| AMD Raphael iGPU                   | gfx1036    | `device-gfx1036` |
-| AMD RX 5700                        | gfx1010    | `device-gfx1010` |
-| AMD Radeon Pro V520                | gfx1011    | `device-gfx1011` |
-| AMD RX 5500                        | gfx1012    | `device-gfx1012` |
+| Product Name                                         | GFX Target | Device Extra     |
+| ---------------------------------------------------- | ---------- | ---------------- |
+| AMD Instinct MI355X / MI350X                         | gfx950     | `device-gfx950`  |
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -61,75 +63,185 @@ Table of contents:
...
 
-|                      | Multi-arch releases                    | Per-family releases                    |
-| -------------------- | -------------------------------------- | -------------------------------------- |
-| GPU code separation  | Split into device packages / `.kpack`  | Bundled into each artifact             |
-| GPU selection        | Package extras/variants                | Chosen by index URL or tarball name    |
-| Multiple GPU targets | Install extras / use multiarch tarball | Separate venvs or tarballs per family  |
-| Download size        | Smaller per target                     | Larger (all targets in family bundled) |
+|                      | Multi-arch releases                         | Per-family releases                    |
+| -------------------- | ------------------------------------------- | -------------------------------------- |
+| GPU code separation  | Split into device packages / `.kpack` files | Included directly into main packages   |
```
nunnikri: If we can add a sample device package name it will be more easily understandable to the reader. 

This whole table we are trying to make it very concise. At the same time too many information is there. Having bit more details can help readers i felt. But dont have a good suggestion to make it better

RELEASES.md
```diff
@@ -61,75 +63,185 @@ Table of contents:
...
+<!-- TODO: Advertise wheel variants / WheelNext once available  -->
+
 After installing, verify your installation:
 
 ```bash
 rocm-sdk test
 ```
 
-<!-- ### Installing multi-arch PyTorch Python packages -->
+#### Supported Python `[device-*]` install extras
```
nunnikri, line 143: Should we keep this device list as a top common list. So that all package installation type can refer same

RELEASES.md
```diff
@@ -61,75 +63,185 @@ Table of contents:
...
+> which contain GPU-specific kernels and depend on `rocm-sdk-device-gfx942`.
+> The compatible ROCm packages are installed automatically - you do not need to
+> install ROCm separately:
+>
+> ```bash
+> pip install --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
+>     torch[gfx942]
+>
+> pip freeze | grep rocm
+> # amd-torch-device-gfx942==2.11.0+rocm7.13.0a20260430
```
ScottTodd: **Warning:** installing torch _without_ a `[gfx942]` extra will produce a broken install (until an extra is installed)

We're considering other options to make the ergonomics better, like making `torch` depend on _all_ devices by default... but that would also break the package promotion approach we wanted to use for nightly releases 🙃.

Sample error logs from comfyui:
```
torch.AcceleratorError: CUDA error: device kernel image is invalid
Search for `hipErrorInvalidImage' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect. For debugging consider passing AMD_SERIALIZE_KERNEL=3
Device-side assertion tracking was not enabled by user.
```

RELEASES.md
```diff
@@ -155,16 +267,50 @@ such as setting environment variables.
...
+./install/bin/rocminfo
+ls install/.kpack/
+# blas_lib_gfx1100.kpack  fft_lib_gfx1100.kpack  rand_lib_gfx1100.kpack  ...
+```
+
+> [!TIP]
+> You may want to add the install directory to your environment:
+>
+> ```bash
+> export PATH="$(pwd)/install/bin:$PATH"
```
nunnikri: This is the step in our publlic documentation. Should we use that?

```
# Configure ROCm PATH. Make sure you're in the therock-tarball directory before proceeding.
ROCM_INSTALL_PATH=$(pwd)/install
sudo tee /etc/profile.d/set-rocm-env.sh << EOF
export ROCM_PATH=$ROCM_INSTALL_PATH
export PATH=\$PATH:\$ROCM_PATH/bin
export LD_LIBRARY_PATH=\$ROCM_PATH/lib/rocm_sysdeps/lib:\$ROCM_PATH/lib:\$LD_LIBRARY_PATH
EOF
sudo chmod +x /etc/profile.d/set-rocm-env.sh

source /etc/profile.d/set-rocm-env.sh
```

ScottTodd: That's all very Linux-specific and difficult to translate to Windows (compared to setting two environment variables). We could link to more detailed documentation or include per-platform docs here.

I can also trim this to more closely match what we already had: https://github.com/ROCm/TheRock/blob/main/RELEASES.md#using-installed-tarballs (Claude decided to rephrase and add a code example lol)

RELEASES.md
ScottTodd: Current ordering:

```
### Installing multi-arch ROCm Python packages
<instructions>
#### Supported Python `[device-*]` install extras
#### The Python `[device-all]` install extra
### Installing multi-arch PyTorch Python packages
```

I kept going back and forth on where the list should go. If we went with 

```
### Installing multi-arch ROCm Python packages
#### Supported Python `[device-*]` install extras
#### The Python `[device-all]` install extras
<instructions>
### Installing multi-arch PyTorch Python package
```

Then the `#installing-multi-arch-rocm-python-packages ` anchor (https://github.com/ScottTodd/TheRock/blob/multi-arch-release-docs-progress/RELEASES.md#installing-multi-arch-rocm-python-packages), which is already packed with information, would have the install instructions drop further off the page. See the current page:
image

I also considered putting the list at the top but hiding it in a `<details><summary>` block:

<details><summary>Expand for details</summary>
<p>

like this

</p>
</details>

ScottTodd: We could also move the parts about venvs and package descriptions down into an appendix section 🤔 

The full table is huge though: 
image

RELEASES.md
ScottTodd: I switched this section to what we already had. We can update both together later.

RELEASES.md
ScottTodd: I replaced this section (more user-friendly feature list) and added more specific details down below (file size examples)

nunnikri (approved):

nunnikri:
I think we should look in rewamping the RELEASES.md. One thought i had was to split into two. One RELEASES.md explain what all releases we are making and its path forward and difference(like multi vs single). And second file INSTALL-GUIDE.md

ScottTodd:
I have some ideas in that direction too. I at least want other pages explaining the release process, automation, architecture, etc. for project contributors.

PR #4949, opened 2026-04-30
Fix the PyTorch workflow dispatch

marbre:
`repository` is an unknown argument. The workflow will always dispatched in the repo where the triggering workflow runs.

### DIFF
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -187,7 +187,6 @@ jobs:
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
-              "repository": "${{ inputs.repository || github.repository }}",
               "ref": "${{ inputs.ref || '' }}"
             }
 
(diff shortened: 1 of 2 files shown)


ScottTodd (approved):

ScottTodd:
I guess https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/workflow_dispatch_inputs_test.py didn't catch this? :/

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -187,7 +187,6 @@ jobs:
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
-              "repository": "${{ inputs.repository || github.repository }}",
```
ScottTodd, line 190: > I guess https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/workflow_dispatch_inputs_test.py didn't catch this? :/

Found the issue:

* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml (in TheRock) _does_ have a `repository` input for `workflow_call` and `workflow_dispatch`
* https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml (shim in in rockrel) does _NOT_ have that input, it passes `repository: "ROCm/TheRock"` directly
* When this code trigger the dispatch it gets sent to the shim, which rejects the call since there is an unexpected input
* The tests here in TheRock only test the workflow code in TheRock, not the shims in rockrel

If we dispatch here without the input, workflow runs in TheRock should use the default of `github.repository` and workflow runs in rockrel will use the hardcoded `"ROCm/TheRock"`

We could add unit tests to rockrel that enforce symmetry with the corresponding branch
* `@main` is the default pin
* release branches, if we create them, would pin to a release branch ref

PR #4953, opened 2026-04-30
Add optional hrx component to the core runtime group.

stellaraccident:
* Wires the external [HRX](https://github.com/ROCm/hrx) repository into the core-runtime artifact group. This is flag guarded under INCLUDE_HRX. Note that this component is experimental and undergoing rapid development. Having it wired in properly with a flag prior to landing in rocm-systems will facilitate iteration on an integration branch.
* Most of this PR is about generalizing the way we can add such optional components:
  * Allows a new BRANCH_CONFIG.json that can be configured to flip flags and control optional source checkout/splicing. This is intended to be used on integration branches prior to full integration of a component into the source tree.
  * Extends BUILD_TOPOLOGY source_sets to allow sources to be pulled from a pinned commit in an external repository without the overhead of a submodule (which would impact everyone with its presence). All source sources go under the new .gitignored optional-sources/ and can be used for all early prototyping of new components.
  * Adds flags to fetch_sources.py to explicitly fetch source_sets (which can be used for any source sets but is intended for optional ones like this).
  * Updates the flags.md docs to describe the facility.

We are aiming to enable this by default following the 7.13 build so that it is available for lightweight consumers of ROCm which do not need the full overhead of the HIP runtime.


stellaraccident:
@amd-aakash please assign an appropriate reviewer

jammm (approved):

BUILD_TOPOLOGY.toml
```diff
@@ -94,6 +97,12 @@ submodules = ["rocgdb"]
 description = "ROCm systems monorepo (runtime, profiler, etc.)"
 submodules = ["rocm-systems"]
 
+[source_sets.optional-hrx]
+description = "Optional HRX source checkout"
+external_git_sources = [
+    { name = "hrx", origin = "https://github.com/ROCm/hrx.git", commit = "56e60cfdc8df89edd43f5ae507cd3fd7f5aafd03", path = "optional-sources/hrx" },
```
jammm, line 103: nit: parameterize `commit` to switch between commits ?

stellaraccident: Not sure what mechanism you mean to do so? We can add something on top of this.

stellaraccident:
Verified that test failures are known / flakes not related to this PR.

PR #4960, opened 2026-04-30
[debug-tools] Install rocgdb CI test script from component source

lumachad:
Installs test_rocgdb.py from the rocgdb component's .github/scripts/
directory into tests/rocgdb so it can be invoked by TheRock CI. This
aligns test script management with component ownership.

We don't need extra changes for rocr-debug-agent because its build system was already modified in https://github.com/ROCm/rocm-systems/pull/5651, such that test_rocr-debug-agent.py is sourced from rocm-systems/projects/rocr-debug-agent/.github/scripts and installed into tests/rocm-debug-agent.

### DIFF
--- a/debug-tools/rocgdb/CMakeLists.txt
+++ b/debug-tools/rocgdb/CMakeLists.txt
@@ -705,6 +705,14 @@ if(BUILD_TESTING)
       COMPONENT tests
       USE_SOURCE_PERMISSIONS)
   endforeach()
+
+  # Also install the CI wrapper script to invoke rocgdb tests.
+  # This is used in the context of TheRock CI.
+  install(
+    PROGRAMS ${ROCGDB_SOURCE_DIR}/.github/scripts/test_rocgdb.py
+    DESTINATION tests/rocgdb
+    COMPONENT tests
+  )
 else()
   message(STATUS "rocgdb tests: BUILD_TESTING not set. Skipping installation integrity tests.")
 endif()


lumachad:
Still needs some work. Not to be merged yet.

lumachad:
Not strictly blocked by, but would be nice to get these updates from:
- https://github.com/ROCm/ROCgdb/pull/117

lumachad:
@jayhawk-commits This one should allow us to remove both rocr-debug-agent and rocgdb test scripts from TheRock.

lumachad:
Updated the PR so we don't remove the testing scripts from TheRock just yet. We will do so in https://github.com/ROCm/TheRock/pull/5385.

HereThereBeDragons (commented):
ah i forgot to send my review.

where do we get `test_rocr-debug-agent.py` now from?
these changes here only talk about `.github/scripts/test_rocgdb.py` while https://github.com/ROCm/TheRock/pull/5385 (and the old pr version here) removed both

lumachad:
> ah i forgot to send my review.
> 
> where do we get `test_rocr-debug-agent.py` now from? these changes here only talk about `.github/scripts/test_rocgdb.py` while #5385 (and the old pr version here) removed both

We get it from its submodule. The new location for the rocr-debug-agent CI script is rocm-systems/projects/rocr-debug-agent/.github/scripts/test_rocr_debug_agent.py, and it gets installed into tests/rocm-debug-agent. The PR doing this change was https://github.com/ROCm/rocm-systems/pull/5651, and it's been picked up already by the rocm-systems bump.

So once this PR is merged, we will be free to remove both rocgdb's and rocr-debug-agent's CI script from TheRock in https://github.com/ROCm/TheRock/pull/5385.

Edit 1: The reason we need this PR is because rocgdb uses a cmake wrapper that lives in TheRock (which this PR updates).

HereThereBeDragons:
thanks for clarifying :)

maybe you can update the pr description here and the other pr to reflect this kinda and link those 2 together?

lumachad:
> thanks for clarifying :)
> 
> maybe you can update the pr description here and the other pr to reflect this kinda and link those 2 together?

Thanks for the feedback.

I've extended the PR description a bit. I've also extended the description for https://github.com/ROCm/TheRock/pull/5385. Does that look like what you had in mind?

lumachad:
@HereThereBeDragons @jayhawk-commits @geomin12 Is this latest version OK?

marbre (commented):
A few questions + comments on issues I see with this approach:

* Why should the script not be available in the checked our source?
* If the file is indeed not present this will break hermetic builds and is incompatible with that approach at all.
* The test script is not pinned. It is from our repo, thus security might not be the main issue but how do you make sure we fetch a compatible version?

lumachad:
> A few questions + comments on issues I see with this approach:

Thanks. Pertinent questions.

> 
> * Why should the script not be available in the checked our source?

For our own builds it will always be available. We plan on doing upstream master gdb builds within TheRock as well (for our own ROCgdb-specific CI), and that repo won't have any of these github files and workflows. Therefore this option is geared towards supporting that particular use case. 

Short answer, having a fallback to grab the testing script from the repo means we can build any branch of gdb, not just amd-staging or amd-staging-rocgdb-16.

> * If the file is indeed not present this will break hermetic builds and is incompatible with that approach at all.

The only cases where this file isn't present are when we use branches/repos of gdb not used in TheRock, mostly for internal CI purposes.

> * The test script is not pinned. It is from our repo, thus security might not be the main issue but how do you make sure we fetch a compatible version?

Since the script is a wrapper around gdb's testsuite, it is very difficult to be incompatible. But to answer the question, we plan to keep things compatible for our own validation's sake.

lumachad:
> > A few questions + comments on issues I see with this approach:
> 
> Thanks. Pertinent questions.
> 
> > * Why should the script not be available in the checked our source?
> 
> For our own builds it will always be available. We plan on doing upstream master gdb builds within TheRock as well (for our own ROCgdb-specific CI), and that repo won't have any of these github files and workflows. Therefore this option is geared towards supporting that particular use case.
> 
> Short answer, having a fallback to grab the testing script from the repo means we can build any branch of gdb, not just amd-staging or amd-staging-rocgdb-16.
> 
> > * If the file is indeed not present this will break hermetic builds and is incompatible with that approach at all.
> 
> The only cases where this file isn't present are when we use branches/repos of gdb not used in TheRock, mostly for internal CI purposes.
> 
> > * The test script is not pinned. It is from our repo, thus security might not be the main issue but how do you make sure we fetch a compatible version?
> 
> Since the script is a wrapper around gdb's testsuite, it is very difficult to be incompatible. But to answer the question, we plan to keep things compatible for our own validation's sake.

@marbre Alternatively I can leave the upstream gdb build support bit out and have that as a separate PR.

lumachad:
As discussed with @marbre, moved the CI script download logic to a more appropriate PR here: https://github.com/ROCm/TheRock/pull/5123.

lumachad:
Looking clean CI-wise.

marbre (approved):

jayhawk-commits (approved):
Thanks for this effort. CC'ing @dileepr1 for awareness.

lumachad:
Letting CI run this one more time for good measure.

lumachad:
Unrelated failures. Merging.

PR #4961, opened 2026-04-30
[PyTorch] move UTs requiring hipblas headers to generic.py

pragupta:
## Motivation
Expandable Segments feature got backported to 2.9 and 2.10 release branches as well due to a customer ask. Therefore, moving the associated UTs to generic.py to get skipped across all release branches. These UTs require devel package to be installed at runtime. We will do a thorough analysis and try to run UTs that require devel package in "core" UTs as a separate step/job in our nightly builds

UTs impacted:
            "test_tensor_delete_after_allocator_delete",
            "test_deleted_mempool_not_used_on_oom",
            "test_mempool_expandable",
            
## Submission Checklist

### DIFF
--- a/external-builds/pytorch/skip_tests/generic.py
+++ b/external-builds/pytorch/skip_tests/generic.py
@@ -48,6 +48,13 @@ skip_tests = {
             # NEW ERROR
             # RuntimeError: Error building extension 'dummy_allocator'
             "test_mempool_with_allocator",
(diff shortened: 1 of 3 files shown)


HereThereBeDragons (approved):
would be nice if you could also put those 3 names in the pr description. 

while the comment says it has started with 2.11,  i now also see it failing in 2.9 - so i guess something got backported that introduced at least failing `test_mempool_expandable` to the older versions

PR #4962, opened 2026-04-30
[ml-libs] Remove all-or-nothing CK disable logic; let MIOpen self-filter

BradPepersAMD:
## Summary
- Reapplies #4838 which was reverted in #4923 due to CK failing when `GPU_TARGETS` was empty
- The root cause is now fixed in rocm-libraries: CK's CMakeLists.txt gracefully skips its build when `GPU_TARGETS` is explicitly empty (ROCm/rocm-libraries#6960)
- Removes the all-or-nothing `_ck_supported_gfx_targets` gate that disabled CK entirely when any unsupported arch was in the target list
- Removes `THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL` option — `THEROCK_ENABLE_COMPOSABLE_KERNEL` is sufficient since MIOpen's ck_impl self-filters GPU_TARGETS
- Updates rocm-libraries submodule to include the CK empty GPU_TARGETS fix

## Test plan
- [ ] Build with `THEROCK_AMDGPU_FAMILIES=gfx1100` — CK should configure with empty targets and skip gracefully, MIOpen should build with `MIOPEN_USE_COMPOSABLEKERNEL=ON` and its ck_impl should handle the unsupported arch
- [ ] Build with `THEROCK_AMDGPU_FAMILIES=gfx942` — CK and MIOpen should build normally with full CK support
- [ ] Build with mixed targets (e.g., `gfx942;gfx1100`) — CK should build for gfx942 only, MIOpen ck_impl should produce real library for gfx942 and skip gfx1100

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,10 +269,6 @@ include("${CMAKE_CURRENT_BINARY_DIR}/cmake/therock_topology.cmake")
 # which filters out CORE_RUNTIME dependency for HIP_RUNTIME on Windows.
 
 # Optional sub-features that control behavior within artifacts
-cmake_dependent_option(THEROCK_MIOPEN_USE_COMPOSABLE_KERNEL
-  "Enables composable kernel in MIOpen" ON
-  "THEROCK_ENABLE_COMPOSABLE_KERNEL" OFF)
-
 cmake_dependent_option(THEROCK_ROCWMMA_ENABLE_BENCHMARKS
   "Enables building rocWMMA benchmarks" OFF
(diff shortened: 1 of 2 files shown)


rocm-libraries
marbre, line 1: rocm-systems bump need to go in separately and should not be part of this PR. Maybe we land #4941 first, bump the submodule and then apply the changes here? @BrianHarrisonAMD wdyt?

BrianHarrisonAMD: Yea we want to land #4941 first, and follow-up with this once we get the CK fixes in to allow us to remove the CK skip for empty.

Sorry for the confusion, this should be a draft for now.

PR #4975, opened 2026-04-30
Add new TheRock flag HIP_TESTS_MONITOR_OPS_ARE_ATOMIC

kzhuravl:
- If ON, then monitor builtins are implemented as atomic operations
  - If OFF, then monitor builtins are not an atomic operations
  - Defaults to OFF
  - Flag is used in hip-tests

### DIFF
--- a/FLAGS.cmake
+++ b/FLAGS.cmake
@@ -32,11 +32,11 @@ therock_declare_flag(
 )
 
 therock_declare_flag(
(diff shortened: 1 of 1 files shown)


kzhuravl:
Changed the flag name to better reflect reality (all monitor operations are atomic in our compiler)

rahulc-gh (approved):
LGTM

PR #4981, opened 2026-05-01
[compiler] Add THEROCK_FLAG_LLVM_DYLIB; default LLVM build to fully static

lamb-j:
## Summary

Adds a new build flag `THEROCK_FLAG_LLVM_DYLIB` (defaulting `OFF`) that controls whether LLVM is built as a shared dylib (`libLLVM.so`/`libclang-cpp.so`) and whether tools link against it. With the flag `OFF` (the new default on Linux), LLVM is fully static: no `libLLVM.so` is produced and tools statically link LLVM. To restore the previous behavior:

```
cmake -DTHEROCK_FLAG_LLVM_DYLIB=ON ...
```

Windows is unaffected — the amd-llvm pre-hook already hardcoded the dylib OFF there regardless of this flag.

## Why

When `libLLVM.so` is in the install, processes that also load another LLVM (e.g. rocMLIR statically embedded into MIGraphX, plus Comgr loaded via `DT_NEEDED` through `libamd_comgr.so`) end up with two LLVM instances sharing the `cl::opt`/pass-registry singletons. They crash during `_dl_init` with:

```
rocmlir-gen: ADT/SmallPtrSet.h:301:
  void *llvm::SmallPtrSetIteratorImpl::dereference() const:
  Assertion `Bucket < End' failed.
Aborted (core dumped)
```

The Comgr-side `COMGR_STATIC_LLVM` option does not fully prevent this: `libamd_comgr.so` still ends up with `DT_NEEDED libLLVM.so` because (a) clang/lld static libs carry `"LLVM"` in their installed `INTERFACE_LINK_LIBRARIES`, and (b) symbol versioning on `libLLVM.so` causes the linker to prefer dylib-versioned defs over identical unversioned static-lib defs. As long as `libLLVM.so` exists in the install, downstream consumers can hit the crash. Defaulting the LLVM build to fully static eliminates the conflict by ensuring there is no `libLLVM.so` to load.

The reproducer from LCOMPILER-2156 (`rocmlir-gen --operation gemm --arch gfx950 -m 10 -n 10 -k 10`) was confirmed to no longer crash when the patched `libamd_comgr.so` from this PR's CI is dropped into a nightly install.

This default is intended as an interim workaround — the goal is to revert it back to `ON` once the underlying issues are addressed in Comgr's CMake and/or LLVM's `libLLVM.so` symbol versioning.

## Tradeoff: install size

The LLVM portion of the install (`/opt/rocm/lib/llvm/`) grows by **~1 GB** because every tool (clang, lld, opt, mlir-opt, flang, clang-tidy, ...) now embeds its own copy of LLVM instead of sharing a dylib.

Apples-to-apples comparison of `lib/llvm/{bin,lib}` between the previous DYLIB build (nightly) and this PR's CI artifacts (static):

|                              | Baseline (DYLIB) | Patched (static) | Delta            |
| ---------------------------- | ---------------: | ---------------: | ---------------: |
| `lib/llvm/bin/` (62 files)   |           267 MB |         2,968 MB | **+2,701 MB (+10x)** |
| `lib/llvm/lib/` (libs)       |         2,234 MB |           529 MB | -1,705 MB (-76%) |
| **`lib/llvm/` total**        |     **2,501 MB** |     **3,497 MB** | **+996 MB (+40%)** |

The bin/ growth dominates because tools mushroom from kilobyte-sized stubs (e.g. baseline `mlir-opt` = 27 KB, `clang-23` = 200 KB) into multi-hundred-MB static-linked binaries (patched `mlir-opt` = 268 MB, `clang-23` = 174 MB). That is partially offset by `lib/` shrinking — `libLLVM.so` (130 MB) and `libclang-cpp.so` (89 MB) are no longer produced.

The rest of `/opt/rocm` (HIP runtimes, math libs, etc.) is unaffected. Sizes above are uncompressed; compressed tarballs grow by less. Build time also goes up because each tool now does its own full LLVM static link, but I haven't measured by how much.

## Tracking

LCOMPILER-2156 (https://amd-hub.atlassian.net/browse/LCOMPILER-2156)

## Test plan

- [x] CI configure + build passes with the flag OFF (new default)
- [ ] CI configure + build passes with `-DTHEROCK_FLAG_LLVM_DYLIB=ON` (verifies the previous behavior is still reachable)
- [x] Resulting `libamd_comgr.so` no longer has `libLLVM.so` in `DT_NEEDED` when flag is OFF
- [x] Reproducer from LCOMPILER-2156 (`rocmlir-gen --operation gemm --arch gfx950 -m 10 -n 10 -k 10`) no longer crashes against an install built with the flag OFF (confirmed by reporter)
- [ ] `share/therock/therock_manifest.json` records `"LLVM_DYLIB": false` in the `flags` block


stellaraccident (changes requested):
We're not going to land this because one of our first party projects isn't built correctly and trade off a 1GB binary size regression. Fine to add the flag but not to default it to ON.

Fix migraphx/rocmlir per the usual options for doing so. We shouldn't be shipping 1P components that are linked or versioned incorrectly. If they are embedding a fork of MLIR statically, they need to link their libraries with -Bsymbolic. If they are depending on a dynamic library that uses LLVM, they need to either use their own private version suffix and/or build with hidden visibility.

Static linking of LLVM libraries without -Bsymbolic (or careful versioning) is very unsupported and is usually the cause of co-existence issues like this. We've fixed others on the internet and I'm not going to regress ROCm in this way for a 1P library.

causten:
Thanks @stellaraccident rocMLIR will approach the solution differently

lamb-j:
Closing this one (mostly created to generate some artifacts for testing, thanks TheRock CI!)

PR #4985, opened 2026-05-01
[ci] Removing memory monitor script

geomin12:
As the ASAN resource was resolved and with `actions-prolense` scaling onto GPU machines, we are removing the use of memory monitor script across the build step and test step to lessen logging bloat

### DIFF
--- a/.github/workflows/build_portable_linux_artifacts.yml
+++ b/.github/workflows/build_portable_linux_artifacts.yml
@@ -126,8 +126,7 @@ jobs:
 
       - name: Build therock-archives and therock-dist
         run: |
-          python build_tools/memory_monitor.py --phase Build -- \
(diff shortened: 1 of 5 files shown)


amd-justchen (approved):
LGTM

ScottTodd:
This logging was actually growing on me. See Windows build logs like https://github.com/ROCm/TheRock/actions/runs/25348294930/job/74329922150#step:16:780 where there would otherwise be no log output for 15+ minutes . This logging lets us see that composable_kernel is still building... just very inefficiently (0-5% CPU usage)

Or https://github.com/ROCm/TheRock/actions/runs/25422250417/job/74594699712#step:16:890 where there would be a few periods of 50+ minutes of silence.

PR #4993, opened 2026-05-01
fix rccl ASAN build due to missing hip flags in device ASAN compilation

prasanna-amd:
## Motivation

Fix rccl build failure in TheRock ASAN CI pipeline

## Technical Details

* Summary

The ASAN CI build for TheRock (gfx94X-dcgpu) fails during RCCL device
compilation with:

#+begin_example
clang++: error: cannot find HIP runtime; provide its path via '--rocm-path',
or pass '-nogpuinc' to build without HIP runtime
#+end_example

PR #5569 (rocm-systems) successfully removed the single-GPU-target
restriction for ASAN builds, but exposed a latent variable-shadowing
bug in TheRock's sanitizer toolchain integration.

The ASAN sanitizer stanza in the generated toolchain file
(=cmake/therock_sanitizers.cmake= line 39-40) uses:

#+begin_src cmake
string(APPEND CMAKE_CXX_FLAGS " -fsanitize=address -fno-omit-frame-pointer -g")
string(APPEND CMAKE_C_FLAGS " -fsanitize=address -fno-omit-frame-pointer -g")
#+end_src

Meanwhile, the HIP path flags are set correctly via the =_INIT= mechanism
(=cmake/therock_subproject.cmake= line 1727-1728):

#+begin_src cmake
string(APPEND CMAKE_CXX_FLAGS_INIT " --hip-path=@_hip_dist_dir@")
string(APPEND CMAKE_CXX_FLAGS_INIT " --hip-device-lib-path=@_amd_llvm_device_lib_path@")
#+end_src

** Why this causes the failure

1. The toolchain file runs *before* =project()=.
2. =CMAKE_CXX_FLAGS_INIT= is set with =--hip-path=...= -- this is correct.
3. =string(APPEND CMAKE_CXX_FLAGS ...)= creates a *normal (non-cache) variable*
   containing only =-fsanitize=address ...= (no =--hip-path=).
4. =project()= initializes the *cache variable* =CMAKE_CXX_FLAGS= from
   =CMAKE_CXX_FLAGS_INIT= (which has =--hip-path=).
5. But the normal variable from step 3 *shadows* the cache variable at
   configure time.
6. When =DeviceLinker.cmake= reads =${CMAKE_CXX_FLAGS}= to extract
   =--hip-path=, it gets the normal variable -- which lacks the flag.
7. =rccl-device-compile= is invoked without =--hip-path=, so =amdclang++=
   cannot find HIP runtime headers in TheRock's split directory layout.

** Why non-ASAN builds are unaffected

No sanitizer stanza means no normal variable is created. The cache variable
(with =--hip-path=) is the only =CMAKE_CXX_FLAGS=, and =DeviceLinker.cmake=
reads it correctly.

** Why host compilation works even with ASAN

CMake's built-in compile rules use the *cache variable* for build-system
generation, not the normal variable. The shadowing only affects configure-time
CMake code that reads =${CMAKE_CXX_FLAGS}= directly -- which is exactly what
=DeviceLinker.cmake= does.

** Why this was never caught before

The old RCCL code had a =FATAL_ERROR= guard that rejected multi-target ASAN
builds at configure time. The DeviceLinker pipeline was never reached in ASAN
CI. PR #5569 removed the guard, allowing configure to pass, but exposing the
latent =--hip-path= issue in device compilation.

## Test Plan

Run TheRock CI ASAN

## Test Result

testing-in-progress

### DIFF
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -36,8 +36,8 @@ function(therock_sanitizer_configure
     string(APPEND _stanza "set(THEROCK_SANITIZER \"${_sanitizer}\")\n")
     # TODO: Support ASAN_STATIC to use static ASAN linkage. Shared is almost always the right thing,
     # so make "ASAN" imply shared linkage.
-    string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS \" -fsanitize=address -fno-omit-frame-pointer -g\")\n")
-    string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS \" -fsanitize=address -fno-omit-frame-pointer -g\")\n")
+    string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -fsanitize=address -fno-omit-frame-pointer -g\")\n")
+    string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -fsanitize=address -fno-omit-frame-pointer -g\")\n")
     # Sharp edge: The -shared-libsan flag is compiler frontend specific:
     #   gcc (and gfortran): defaults to shared sanitizer linkage
     #   clang: defaults to static linkage and requires -shared-libsan to link shared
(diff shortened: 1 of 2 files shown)


sa-faizal:
Thank you for tracing this down.  The write-up on `CMAKE_CXX_FLAGS` vs `CMAKE_CXX_FLAGS_INIT` and configure-time reads is convincing.

The RCCL-only fix in `pre_hook_rccl.cmake` will likely unblock CI, but it’s pretty specialized (regex over _INIT). Can we switch to fixing this in [cmake/therock_sanitizers.cmake](https://github.com/ROCm/TheRock/blob/31245e6b487bde7cf3a493d9a6fabf821fd096e9/cmake/therock_sanitizers.cmake#L39) instead by having the generated toolchain append sanitizer compile flags to `CMAKE_CXX_FLAGS_INIT` / `CMAKE_C_FLAGS_INIT` rather than `CMAKE_CXX_FLAGS` / `CMAKE_C_FLAGS`? 

That can address the shadowing at the source for every subproject that might read `${CMAKE_CXX_FLAGS}` at configure time, not just RCCL, and avoids maintaining a second place that has to stay in sync with how HIP flags are injected.

I have tested out the same on my branch - https://github.com/ROCm/TheRock/actions/runs/25259761925

geomin12 (approved):

sa-faizal (approved):
lgtm! TSAN also unaffected https://github.com/ROCm/TheRock/actions/runs/25261454530/job/74069364187

PR #4995, opened 2026-05-01
[ci] Removing gfx950 from pool

geomin12:
Due to a partial power failure for CCS cluster, these machines are down for maintenance until May 8th. 

We are disabling machines until back as no gfx950 machines are available

### DIFF
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -246,8 +246,8 @@ amdgpu_family_info_matrix_postsubmit = {
         "linux": {
             # Disabled on 5/1/2026 due to power failures in CCS cluster
             # Expected to be restored 5/8/2026
-            "test-runs-on": "", # label is linux-gfx950-1gpu-ccs-ossci-rocm
-            "test-runs-on-multi-gpu": "", # label is linux-gfx950-8gpu-ccs-ossci-rocm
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -244,8 +244,10 @@ def select_build_runner(platform: str, build_variant: str) -> str:
 amdgpu_family_info_matrix_postsubmit = {
     "gfx950": {
         "linux": {
-            "test-runs-on": "linux-gfx950-1gpu-ccs-ossci-rocm",
-            "test-runs-on-multi-gpu": "linux-gfx950-8gpu-ccs-ossci-rocm",
+            # Disabled on 5/1/2026 due to power failures in CCS cluster
+            # Expected to be restored 5/8/2026
```
ScottTodd, line 248: oof, expected 1 week with no gfx950 signal 😣

ScottTodd (approved):

build_tools/github_actions/amdgpu_family_matrix.py
geomin12: yes very unfortunate, OSSCI is working on getting replacement capacity but the power outage is no bueno. landing this to avoid queues

geomin12: there was a spare node, so we have some signal still

Issue #3987, opened 2026-03-16
[Issue]  Windows Pytorch test failures with `Segmentation fault` while running `TestNN::test_CTCLoss_critical_target_len` - gfx110x, gfx1151


chiranjeevipattigidi:
**Platform:** `Windows`
**Effected Archs:** `gf110x-all, gfx1151`
**Pytorch Version:** `nightly`
**Python Versions:** `3.10, 3.11, 3.12, 3.13`

### Run:
- https://github.com/ROCm/TheRock/actions/runs/23141038322/job/67226511146
- https://github.com/ROCm/TheRock/actions/runs/23136265665/job/67209632758

```
Windows fatal exception: access violation

Thread 0x00001fd8 (most recent call first):
  <no Python frame>

Current thread 0x0000248c (most recent call first):
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\torch\nn\functional.py", line 3111 in ctc_loss
  File "B:\runner\_work\TheRock\TheRock\external-builds\pytorch\pytorch\test\test_nn.py", line 2845 in test_CTCLoss_critical_target_len
  File "B:\runner\_work\_tool\Python\3.12.10\x64\Lib\unittest\case.py", line 589 in _callTestMethod
  File "B:\runner\_work\_tool\Python\3.12.10\x64\Lib\unittest\case.py", line 634 in run
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\torch\testing\_internal\common_utils.py", line 3514 in _run_custom
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\torch\testing\_internal\common_utils.py", line 3545 in run
  File "B:\runner\_work\_tool\Python\3.12.10\x64\Lib\unittest\case.py", line 690 in __call__
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\_pytest\unittest.py", line 351 in runtest
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\_pytest\runner.py", line 174 in pytest_runtest_call
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\pluggy\_callers.py", line 121 in _multicall
  File "B:\runner\_work\TheRock\TheRock\.venv\Lib\site-packages\pluggy\_manager.py", line 120 in _hookexec
[...]

PR #4996, opened 2026-05-01
Add multi-arch support to PyTorch test workflows

ScottTodd:
## Motivation

This lets our test workflows install multi-arch packages to make progress on

The next step will be to connect trigger these tests as part of the new multi-arch releases here: https://github.com/ROCm/TheRock/blob/6d964adcf0b9194349a4c435f780dfa27130192f/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml#L313-L317

similar to https://github.com/ROCm/TheRock/blob/6d964adcf0b9194349a4c435f780dfa27130192f/.github/workflows/build_portable_linux_pytorch_wheels.yml#L311-L348

## Technical Details

Package install steps differ between single-family and multi-arch packages. Rather than use a separate _index_ per GPU family they use a _device extra_ in a shared index:

Package type | Install command
-- | --
Single family | `pip install --index-url=https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu torch==2.11.0+rocm7.13.0a20260501`
Multi-arch | `pip install --index-url=https://rocm.nightlies.amd.com/whl-staging-multi-arch/ 'torch[device-gfx942]==2.11.0+rocm7.13.0a20260501'`

I'm not thrilled about the bash branches in this code (https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash), but we have worse offenders where we need to support both single-family and multi-arch (or kpack disabled/enabled) paths. Once we fully complete the migration to multi-arch we can go through and prune those branches down to just the new code.

## Test Plan

Triggered a few test jobs that show no obvious regressions between single-family and multi-arch torch packages:

Job description | Link | Results
-- | -- | --
Linux gfx942 | [run-25233709159](https://github.com/ROCm/TheRock/actions/runs/25233709159) | `= 39604 passed, 1986 skipped, 151 deselected, 48 xfailed, 6 subtests passed in 1052.95s (0:17:32) =`
Windows gfx1151 | [run-25232905677](https://github.com/ROCm/TheRock/actions/runs/25232905677) | segfault (matches https://github.com/ROCm/TheRock/issues/3987)
Windows gfx110X | [run-25234965314](https://github.com/ROCm/TheRock/actions/runs/25234965314) | pending

I did **not** trigger `test_pytorch_wheels_full.yml`.

We'll get more complete signal once these test jobs are integrated into the multi-arch release workflows.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -132,7 +132,8 @@ jobs:
             --torch-version=${{ inputs.torch_version }} \
             --pytorch-git-ref=${{ inputs.pytorch_git_ref }} \
             --index-url=${{ inputs.package_index_url }} \
-            ${{ inputs.multi_arch && format('--device-extras={0}', steps.expand.outputs.device_extras) || format('--index-subdir={0}', inputs.amdgpu_family) }}
+            --index-subdir=${{ inputs.amdgpu_family }} \
+            --device-extras="${{ steps.expand.outputs.device_extras }}"
 
       - name: Set git options
         run: |
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -121,11 +121,9 @@ jobs:
         if: ${{ inputs.multi_arch }}
         id: expand
         run: |
-          device_extras=$(python build_tools/github_actions/expand_amdgpu_families.py \
+          python build_tools/github_actions/expand_amdgpu_families.py \
             --amdgpu-families "${{ inputs.amdgpu_family }}" \
-            --prefix "device-")
-          echo "Expanded '${{ inputs.amdgpu_family }}' -> device extras '${device_extras}'"
-          echo "device_extras=${device_extras}" >> "$GITHUB_OUTPUT"
+            --output-mode=device-extras
 
       # TODO: also upload and reference test report together with this logging?
       - name: Summarize workflow inputs
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -17,7 +17,10 @@ on:
         type: string
         default: "linux-mi325-1gpu-ossci-rocm"
       package_index_url:
-        description: Base Python package index URL to test, typically nightly/dev URL with a "v2" or "v2-staging" subdir (without a GPU family subdir)
+        description: >-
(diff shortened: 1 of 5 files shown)


.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 1: > The next step will be to connect trigger these tests as part of the new multi-arch releases here:

I've started design work for this. Here are my scratch notes:
* https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/multi-arch-pytorch-testing.md
* https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/multi-arch-pytorch-testing.md#test-matrix-generation

(This PR should be ready to review independently of how we end up trigger the test workflows)

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -15,9 +15,12 @@ on:
         description: Runner label to use. The selected runner should have a GPU supported by amdgpu_family
         required: true
         type: string
-        default: "linux-gfx942-1gpu-ossci-rocm"
+        default: "linux-mi325-1gpu-ossci-rocm"
       package_index_url:
-        description: Base Python package index URL to test, typically nightly/dev URL with a "v2" or "v2-staging" subdir (without a GPU family subdir)
+        description: >-
+          Base Python package index URL to test.
+          Per-family: nightly/dev URL with a "v2" or "v2-staging" subdir (without a GPU family subdir).
```
marbre, line 22: How do you think about me applying a trick on the CF side to make this `whl` and `whl-staging`. In the multi-arch case you would still need to append `-multi-arch` only. However, single-family also needs the family, thus idk how much this really would help.

build_tools/github_actions/expand_amdgpu_families.py
marbre, line 1: Nice extension!

marbre (approved):

.github/workflows/test_pytorch_wheels.yml
ScottTodd: Eh I'd rather not rock the boat too much on the single-family releases. If you think it can be done pretty transparently/safely then sure. Otherwise I'd say we should just get the multi-arch releases fully working, migrate all code to them, then delete code paths for the older releases (while keeping the uploaded files available within our retention policies)

PR #5028, opened 2026-05-04
Enable multi-arch package download and filtering in download_prerelease_packages.py

araravik-psd:
Adds support for downloading and listing multi-arch packages from flat S3 layouts (e.g. v4/whl/) and introduces consistent filtering using both existing and multi-arch package allow lists.

**Key Changes**

- Multi-arch support

1. Added --multi-arch mode for flat S3 layouts
2. Supports both listing (--list-packages-multi-arch) and download flows
3. Removes reliance on architecture-based directory structure
4. Added rocm_profiler to promotion list
5. Updated how_to_do_release.md with script naming change and script usage

**Package filtering**
1. Introduced PACKAGES_TO_PROMOTE_MULTI_ARCH for new packages introduced for multi arch
2. Added is_allowed_multi_arch_package() to unify filtering across:

- PACKAGES_TO_PROMOTE
- PACKAGES_TO_PROMOTE_MULTI_ARCH

Package filename filtering

Added repeatable package filename filter options to the download script:

- `--include-package-glob`
- `--exclude-package-glob`

These filters are applied after package discovery and before listing/downloading packages. This allows release flows to exclude specific package variants at download time, for example filtering out PyTorch 2.8.0-related packages from multi-arch staging packages.

The filtering applies to:

- single-arch package downloads
- multi-arch package listing
- multi-arch package downloads
- `--list-packages-per-arch`

Example:

```bash
python build_tools/packaging/download_python_packages.py \
  --version=7.13.0rc1 \
  --multi-arch \
  --list-multi-arch-packages \
  --bucket-prefix=v4/whl-staging/ \
  --bucket therock-prerelease-python \
  --exclude-package-glob "*-2.8.0+rocm*"
```

Test logs have been added to the end of the PR description as a Github gist

**Multi-arch arch filtering**

- Extended existing --arch filtering support to multi-arch flows
- Added filtering for both:
   multi-arch package listing
   multi-arch package downloads
- Generic packages continue to be included
- Device-specific packages are filtered based on requested gfx architectures

3. Ensures only relevant packages (including device-specific ones) are processed

**New functionality**

1. list_packages_multi_arch_verbose() for listing flat-layout packages
2. download_multi_arch_packages() for downloading filtered packages
3. has_version_in_directory() to validate package availability

CLI updates and new Variables added:

Added flags:
--multi-arch
--list-packages-multi-arch

Notes
Multi-arch flow assumes a flat prefix layout (no per-arch directories)
Single-arch flow remains unchanged and continues to use existing logic
Filtering is now consistent across listing and download flows

Testing

Listing functionality for both multi-arch and single arch are added below:
https://gist.github.com/araravik-psd/58a570bccd944fc8a5df7a96d6c6abe3

Downloading multi-arch packages
https://gist.github.com/araravik-psd/38682418ed118f8aa06d023b35d44911

Filtering download script based on version number and include exclude package globs
https://gist.github.com/araravik-psd/3b705512e8a86bc0a98cb634cb88da79


build_tools/packaging/download_prerelease_packages.py
```diff
@@ -140,6 +141,24 @@
...
+    "rocm_sdk_core",
+    "rocm_sdk_devel",
+    "rocm_sdk_libraries",
+    "rocm_profiler",
+    "rocm_bootstrap",
+    # device packages (IMPORTANT)
+    "amd_torch_device",
+    "amd_torchvision_device",
+    "rocm_sdk_device",
+}
```
ScottTodd, line 173: drive-by: we may want a placeholder here for `amd_torchaudio_device` (and in a few other scripts/workflows) so we remember if https://github.com/pytorch/audio/pull/4180 gets further along

build_tools/packaging/download_prerelease_packages.py
```diff
@@ -632,6 +800,20 @@ def parse_arguments(argv):
...
 
+    parser.add_argument(
+        "--multi-arch",
+        action=argparse.BooleanOptionalAction,
+        default=False,
+        help="--multi-arch requires prefix like v4/whl/",
+    )
+
+    parser.add_argument(
+        "--list-packages-multi-arch",
```
marbre:
```suggestion
        "--list-multi-arch-packages",
```
?

build_tools/packaging/download_prerelease_packages.py
```diff
@@ -632,6 +800,20 @@ def parse_arguments(argv):
         help="List all packages per architecture, do not download",
     )
 
+    parser.add_argument(
+        "--multi-arch",
+        action=argparse.BooleanOptionalAction,
+        default=False,
+        help="--multi-arch requires prefix like v4/whl/",
```
marbre: This help is confusing since the option is boolean.

build_tools/packaging/download_prerelease_packages.py
marbre, line 1: We may want to rename this script?

araravik-psd: I suggest renaming to download_release_artifacts.py. As this is not only for prereleases.

build_tools/packaging/download_prerelease_packages.py
araravik-psd: Updated with a placeholder

build_tools/packaging/download_prerelease_packages.py
araravik-psd: Done updated

build_tools/packaging/download_prerelease_packages.py
araravik-psd: Updated the help text here.

build_tools/packaging/download_prerelease_packages.py
araravik-psd: Also updated the tests with the argument change

https://gist.github.com/araravik-psd/58a570bccd944fc8a5df7a96d6c6abe3

build_tools/packaging/download_prerelease_packages.py
marbre: Well it's not artifacts, those are rather the tarballs we push to the artifact buckets. It is rather download_python_packages isn't it?

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +189,64 @@
 }
 
 
+def is_allowed_multi_arch_package(filename: str) -> bool:
```
marbre: Why is this distinction important here? Multi-arch will always be in a different index / suffic.

build_tools/packaging/download_python_packages.py
marbre, line 1: I think this could actually reuse quite some code from other scripts, see my comments in the upload script PR.

araravik-psd: Ok I will create an issue to reuse build_tools/github_actions/publish_rocm_to_release_buckets.py functionality in the next version of the script and update this PR with the issue.

build_tools/packaging/download_python_packages.py
araravik-psd: The distinction is not for separating storage layouts, multi-arch already uses a dedicated prefix.

The filtering here is used to restrict downloads/listings to promotable package families only, while excluding unrelated artifacts that may exist in the same bucket/prefix.

In particular, multi-arch introduces additional dynamic package families (for example amd_torch_device_* and rocm_sdk_device_*) that are not part of the legacy single-arch allow-list, so the helper centralizes that filtering logic for the multi-arch flow.

I am mainly trying to keep multi-arch package selection independent from the legacy single-arch categorization logic and not overload categorize_package() with different semantics used. I hope this is good, if you would like me to consolidate all together I can do that too

build_tools/packaging/download_prerelease_packages.py
araravik-psd: yes that is more accurate, changing the script name to download_python_packages.py

build_tools/packaging/download_python_packages.py
```diff
@@ -80,7 +80,8 @@
       <arch2>/
         package1.whl
         ...
-      tarball/  (if --include-tarballs is specified)
+      tarball/              (single-arch tarballs)
```
HereThereBeDragons: you still need `--include-tarballs` - so might be good to have this still in the comment?

build_tools/packaging/download_python_packages.py
```diff
@@ -140,6 +142,27 @@
     "jaxlib",
 }
 
+PACKAGES_TO_PROMOTE_MULTI_ARCH = {
+    "torch",
```
HereThereBeDragons: can we get here some ordering in:
rocm, torch, jax? (currently PACKAGES_TO_PROMOTE) or alphabetically? but then the same should be also done to PACKAGES_TO_PROMOTE

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +190,64 @@
...
+    def matches(pkg: str) -> bool:
+        # Handle wildcard like "rocm_sdk_libraries-*"
+        if pkg.endswith("*"):
+            return base.startswith(pkg[:-1])
+
+        # Exact match OR prefix match (for device families)
+        return base == pkg or base.startswith(pkg)
+
+    return any(
+        matches(pkg) for pkg in (PACKAGES_TO_PROMOTE | PACKAGES_TO_PROMOTE_MULTI_ARCH)
```
HereThereBeDragons:
```suggestion
        matches(pkg) for pkg in PACKAGES_TO_PROMOTE_MULTI_ARCH
```

you already list all packages in PACKAGES_TO_PROMOTE_MULTI_ARCH. dont poison it with PACKAGES_TO_PROMOTE .

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +190,64 @@
...
+    Supports:
+      - exact match
+      - wildcard '*' suffix (legacy)
+      - prefix match (for device packages)
+    """
+    base = filename.split("-")[0]
+
+    def matches(pkg: str) -> bool:
+        # Handle wildcard like "rocm_sdk_libraries-*"
+        if pkg.endswith("*"):
```
HereThereBeDragons: this is not needed for mutlliarch. for multiarch you can match it with `device` to know that it needs to be followed by gfx.

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +190,64 @@
...
+            return base.startswith(pkg[:-1])
+
+        # Exact match OR prefix match (for device families)
+        return base == pkg or base.startswith(pkg)
```
HereThereBeDragons: as comment above: please do it more correctly as you have a separate list of multiarch and single arch anyway. so just match either correctly or if contains device check that it startswtih and contains some gfx.

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +190,64 @@
...
+
+    return any(
+        matches(pkg) for pkg in (PACKAGES_TO_PROMOTE | PACKAGES_TO_PROMOTE_MULTI_ARCH)
+    )
+
+
+def matches_requested_arches(filename: str, requested_arches: list[str]) -> bool:
```
HereThereBeDragons: i think you can integrate this function into `is_allowed_multi_arch_package`.
`if package name contains device` check `if gfx in requested_arches`

build_tools/packaging/download_python_packages.py
```diff
@@ -167,6 +190,64 @@
...
+    """
+    Check whether a multi-arch package matches one of the requested gfx arches.
+
+    Rules:
+      - If no requested arches are provided -> allow
+      - If filename does not contain gfx -> allow (generic package)
+      - If filename contains gfx -> require one requested arch match
+    """
+
+    if not requested_arches:
```
HereThereBeDragons: why would we even end up here?

build_tools/packaging/download_python_packages.py
```diff
@@ -816,13 +1133,68 @@ def download_prerelease_packages(
...
     print(f"Version: {version}")
+
     if architectures:
         print(f"Architectures: {architectures} (user-specified)")
     else:
         print(f"Architecture: ALL")
     print("=" * 80)
 
     s3_client = boto3.client("s3")
+    if multi_arch:
```
HereThereBeDragons, line 1287: considering how long multi arch and single arch sections are: lets put them in separate functions

build_tools/packaging/download_python_packages.py
```diff
@@ -816,13 +1133,68 @@ def download_prerelease_packages(
...
+        # Validate that packages exist
+        if not has_version_in_directory(
```
HereThereBeDragons: now it makes sense to rename that stuff:
`has_version_in_directory` -> `exists_version_mutli_arch`
`has_version_in_arch` -> `exists_version_single_arch`

or something like this?

build_tools/packaging/download_python_packages.py
```diff
@@ -263,6 +344,78 @@ def has_version_in_arch(
...
+            continue
+
+        for obj in page["Contents"]:
+            if version in obj["Key"]:
+                return True
+
+    return False
+
+
+def list_packages_multi_arch_verbose(
```
HereThereBeDragons:
```suggestion
def list_packages_multi_arch(
```
why the verbose?

build_tools/packaging/download_python_packages.py
```diff
@@ -816,13 +1133,68 @@ def download_prerelease_packages(
...
+                bucket_prefix,
+                version,
+                architectures,
+            )
+            return
+
+        output_dir.mkdir(parents=True, exist_ok=True)
+        print(f"\nOutput directory: {output_dir.absolute()}")
+
+        success, fail = download_multi_arch_packages(
```
HereThereBeDragons: keep it consistent with single arch and also put out the tarball download directory

```
 if include_tarballs:
        tarball_output_dir.mkdir(parents=True, exist_ok=True)
        print(f"Tarball output directory: {tarball_output_dir.absolute()}")
```

build_tools/packaging/download_python_packages.py
```diff
@@ -816,13 +1133,68 @@ def download_prerelease_packages(
...
+                output_dir,
+                architectures,
+            )
+
+            success += tar_success
+            fail += tar_fail
+
+        print("\n" + "=" * 80)
+        print("DOWNLOAD COMPLETE (MULTI-ARCH)")
+        print("=" * 80)
```
HereThereBeDragons: in single arch we list the architectures we downloade.d might also want to do it here.

`print(f"Total architectures: {len(architectures)}")`
or maybe rewrite it to name all the architectures? might be nicer

build_tools/packaging/download_python_packages.py
```diff
@@ -816,13 +1133,68 @@ def download_prerelease_packages(
...
+        print(f"Total successful downloads: {success}")
+        print(f"Total failed downloads: {fail}")
+
+        if fail > 0:
```
HereThereBeDragons:
```
if total_fail > 0:
        print("\nWARNING: Some downloads failed!")
        sys.exit(1)
```

build_tools/packaging/download_python_packages.py
HereThereBeDragons: maybe lets rename it to have the same order like the donwload func `download_multi_arch_packages`: `list_multi_arch_packages`?

build_tools/packaging/download_python_packages.py
```diff
@@ -535,28 +827,28 @@ def parse_arguments(argv):
...
-  python download_prerelease_packages.py --version=7.10.0rc2 --list-archs
+  python download_python_packages.py --version=7.10.0rc2 --list-archs
 
   # List all packages per architecture with sizes
-  python download_prerelease_packages.py --version=7.10.0rc2 --list-packages-per-arch
+  python download_python_packages.py --version=7.10.0rc2 --list-packages-per-arch
 
   # List packages and tarballs with sizes
-  python download_prerelease_packages.py --version=7.10.0rc2 --list-packages-per-arch --include-tarballs
+  python download_python_packages.py --version=7.10.0rc2 --list-packages-per-arch --include-tarballs
```
HereThereBeDragons, line 942: lets add soem examples for multiarch

build_tools/packaging/download_python_packages.py
```diff
@@ -387,6 +540,145 @@ def download_file(s3_client, bucket_name: str, key: str, local_path: Path) -> bo
...
+            )
+
+            if download_file(s3_client, bucket, key, local_path):
+                total_success += 1
+            else:
+                total_fail += 1
+
+    return total_success, total_fail
+
+def download_multi_arch_tarballs(
```
HereThereBeDragons: multi arch and single arch tarballs have the same s3 layout.

how about we reuse it and just rename and have a thing 
```
def list_tarball_for_arch(
    s3_client, bucket_name: str, bucket_prefix: str, arch: str, version: str
) -> List[Tuple[str, int]]:
```
to 
```
def list_tarball_for_package(
    s3_client, bucket_name: str, bucket_prefix: str, package: str, version: str
) -> List[Tuple[str, int]]:
```

i think i remember faintly we had talked about this?

you then just need to have an extra look up for the multiarch tarball.
could maybe allow 
```
def download_tarball(
    s3_client,
    bucket_name: str,
    bucket_prefix: str,
    arch: str,
    version: str,
    output_dir: Path,
) -> Tuple[int, int]:
```
to become a list `arch: str` -> `packages: List[str]` to work for both

HereThereBeDragons: i will not review this function any further, as i think this is a duplicate.

build_tools/packaging/download_python_packages.py
```diff
@@ -263,6 +344,78 @@ def has_version_in_arch(
...
+            continue
+
+        for obj in page["Contents"]:
+            key = obj["Key"]
+            filename = key.split("/")[-1]
+
+            if not filename or filename == "index.html":
+                continue
+
+            if version in filename and is_allowed_multi_arch_package(filename):
```
HereThereBeDragons: as said, better combine `is_allowed_multi_arch_package` and `matches_requested_arches`

build_tools/packaging/download_python_packages.py
```diff
@@ -387,6 +540,145 @@ def download_file(s3_client, bucket_name: str, key: str, local_path: Path) -> bo
...
+
+    total_success = 0
+    total_fail = 0
+
+    pages = paginator.paginate(
+        Bucket=bucket,
+        Prefix=prefix,
+    )
+
+    print("\nDownloading packages")
```
HereThereBeDragons: can we maybe match this more what single arch does?
use `list_packages_multi_arch_verbose` to create the list 

```
packages_to_promote, dependencies, unknown = list_packages_for_arch(
        s3_client, bucket_name, bucket_prefix, arch, version
    )
```

have more verbose output and then just download `packages_to_promote`?

HereThereBeDragons (commented):
please add comments to all the functions like already available for the single arch functions

please introduce type hints as we want per our code style

further comments from claude:
- Minor — Naming inconsistency. New functions use bucket/prefix; the rest of the file uses bucket_name/bucket_prefix. Stay consistent.
- Minor — [download_python_packages.py:609-680](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L609-L680) — download_multi_arch_tarballs writes to output_dir / "tarball-multi-arch" regardless of the user's --tarball-output-dir. The single-arch path honors that flag; the multi-arch path silently ignores it.
- Minor — [download_python_packages.py:609-680](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L609-L680) — tarball path applies matches_requested_arches but not is_allowed_multi_arch_package. If that's intentional (tarball naming differs from wheel naming), worth a one-line comment saying so; otherwise it's a bug.
- Minor — [download_python_packages.py:1078-1118](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L1078-L1118) — download_prerelease_packages docstring is out of sync: no Args entries for the new multi_arch and list_multi_arch_packages parameters.
- Minor — Default --bucket-prefix=v3/whl/ is a single-arch layout. Passing --multi-arch without overriding the prefix will fail confusingly inside has_version_in_directory. Either auto-error early ("--multi-arch requires a flat prefix like v4/whl/") or default the prefix when --multi-arch is set.
- Nit — [download_python_packages.py:157](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L157) — # device packages (IMPORTANT) doesn't say why they're important. Either drop the parenthetical or replace it with the actual reason (e.g. "must remain in allow-list because they're per-gfx and not covered by PACKAGES_TO_PROMOTE").

Nit — [download_python_packages.py:543-607](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L543-L607) and [:609-680](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/build_tools/packaging/download_python_packages.py#L609-L680) — print(f"  Downloading: {filename} " f"({size / BYTES_TO_MB:.2f} MB)") is split across two adjacent f-strings; collapse to one line.

ScottTodd:
I think others have provided enough review feedback here already. LMK if you want me to look at something specific.

build_tools/packaging/download_python_packages.py
araravik-psd: Issue created to track this: https://github.com/ROCm/TheRock/issues/5187

build_tools/packaging/download_python_packages.py
araravik-psd: I reordered both package lists consistently (alphabetically) for readability.

build_tools/packaging/download_python_packages.py
araravik-psd: Thanks for catching this made the changes.

build_tools/packaging/download_python_packages.py
araravik-psd: wildcard handling was inherited from the single-arch flow and is unnecessary for the multi-arch allowlist. I’ll simplify the matching logic and update the comments/docstring accordingly.

build_tools/packaging/download_python_packages.py
araravik-psd: I’ll tighten the matching logic so regular packages use exact matching, while device packages explicitly require both the device prefix and a gfx target in the filename.

build_tools/packaging/download_python_packages.py
araravik-psd: I’ll fold the gfx/device arch filtering into is_allowed_multi_arch_package() so the multi-arch filtering logic stays centralized in a single helper

build_tools/packaging/download_python_packages.py
araravik-psd: The multi-arch flow has grown quite a bit now, so I’ll extract the multi-arch handling into a separate helper function(handle_multi_arch_downloads) to keep download_prerelease_packages() easier to follow.

build_tools/packaging/download_python_packages.py
araravik-psd: Yeah, that makes sense now that the flows are more separated. I’ll rename them to make the intent clearer between single-arch and multi-arch handling.

build_tools/packaging/download_python_packages.py
araravik-psd: Makes sense, I’ll rename it to `list_multi_arch_packages()` for consistency with the download helper.

build_tools/packaging/download_python_packages.py
araravik-psd: I’ll add the downloaded architectures to the multi-arch summary as well for consistency with the single-arch flow.

build_tools/packaging/download_python_packages.py
araravik-psd: I combined the requested-arch filtering into is_allowed_multi_arch_package() now, so the separate helper is no longer needed.

build_tools/packaging/download_python_packages.py
araravik-psd: Done. I refactored the multi-arch flow to follow the single-arch pattern more closely now. `list_packages_multi_arch()` builds and returns the filtered package list first, and `download_multi_arch_packages()` just consumes that list for downloading with more verbose output.

araravik-psd:
@HereThereBeDragons Addressed your comments and have updated the tests. Please take a look again when you get a chance.

build_tools/packaging/download_python_packages.py
```diff
@@ -126,18 +137,39 @@
...
+
+PACKAGES_TO_PROMOTE_MULTI_ARCH = {
+    "amd_torch_device",
+    "amd_torchvision_device",
     "apex",
-    "jax_rocm7_pjrt",
-    "jax_rocm7_plugin",
-    "jaxlib",
+    "rocm",
+    "rocm_bootstrap",
```
marbre:
```suggestion
```
We currently don't build `rocm-bootstrap` via TheRock and it gets instead pulled from PyPI where it is already published. So nothing to promote here.

build_tools/packaging/download_python_packages.py
```diff
@@ -25,51 +25,60 @@
...
     --include-tarballs
 
   # Download packages to promote and all known PyPI dependencies:
-  python ./build_tools/packaging/download_prerelease_packages.py \
+  python ./build_tools/packaging/download_python_packages.py \
     --version=7.10.0rc2 \
     --output-dir=./downloads \
     --include-dependencies
 
+  # Download multi-arch packages
```
JeniferC99, line 73: cosmetic, but the download instruction formatting is not consistent with above. 

and maybe it's worth mentioning that --bucket-prefix=v4/whl/ is added for multi-arch because we are extracting from whl-staging on multi-arch due to test instability right now

JeniferC99 (approved):
I tested it for 7.13 release. it worked as described.

build_tools/packaging/download_python_packages.py
ScottTodd: Multi-arch does not use whl-staging as of https://github.com/ROCm/TheRock/pull/5107

araravik-psd: Example was taken from whl-staging for rc1 in 7.13.0rc1, will update download instructions with more examples

Issue #5029, opened 2026-05-04
[Issue]: Windows build failure `[hip-clr] wkmi.lib(wkmi.obj) : error LNK2019: unresolved external symbol __std_search_1`


ScottTodd:
My local builds on Windows started failing recently, consistently:
```
[build] [hip-clr] [140/141] Building CXX object hipamd\src\CMakeFiles\amdhip64.dir\hip_graph.cpp.obj
[build] [hip-clr] cl : Command line warning D9025 : overriding '/MD' with '/MT'
[build] [hip-clr] [141/141] Linking CXX shared library hipamd\src\amdhip64_7.dll
[build] [hip-clr] FAILED: [code=4294967295] hipamd/src/amdhip64_7.dll hipamd/lib/amdhip64.lib 
[...]

PR #5031, opened 2026-05-04
Bump minimum MSVC version to 19.43 in docs and validate scripts

ScottTodd:
## Motivation

Patch for https://github.com/ROCm/TheRock/issues/5029 . Source builds since ~April 7 have required version 19.43+, so my builds using 19.42 were failing.

I also removed checks for CMake <4.0 given that https://github.com/ROCm/TheRock/issues/318 has been marked fixed for months.

## Test Plan

* Source builds of `hip-clr` failed with the older version installed
* Source builds of `hip-clr` succeeded with the newer version installed
* Both validate scripts produced errors with an older version installed
* Both validate scripts passed with a newer version installed

## Submission Checklist

### DIFF
--- a/build_tools/validate_windows_install.ps1
+++ b/build_tools/validate_windows_install.ps1
@@ -219,12 +219,34 @@ try {
 # ============================================================================
 Write-Section "3. MSVC Compiler"
 
+$MinMsvcVersion = [Version]"19.43"
+
 $clCmd = Get-Command cl.exe -ErrorAction SilentlyContinue
 if ($clCmd) {
-    # cl.exe prints its version banner to stderr
-    $clBanner = (& cl.exe 2>&1 | Select-Object -First 1).ToString().Trim()
+    # cl.exe prints its version banner to stderr; capture it with
+    # ErrorAction Continue so SilentlyContinue doesn't swallow error records.
+    $prevEAP = $ErrorActionPreference
+    $ErrorActionPreference = "Continue"
+    $clOutput = (& cl.exe 2>&1) | ForEach-Object { $_.ToString().Trim() }
+    $ErrorActionPreference = $prevEAP
+    $clBanner = ($clOutput | Select-Object -First 1)
     Write-Pass "cl.exe in PATH: $($clCmd.Source)"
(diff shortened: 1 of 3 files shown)


build_tools/validate_windows_install.ps1
```diff
@@ -219,12 +219,34 @@ try {
...
+    $clBanner = ($clOutput | Select-Object -First 1)
     Write-Pass "cl.exe in PATH: $($clCmd.Source)"
     Write-Info $clBanner
+
+    # Check MSVC version meets minimum requirement.
+    # Prebuilt libraries (e.g. wkmi.lib) may reference STL internals introduced
+    # in newer MSVC versions, causing LNK2019 errors on older toolchains.
+    # See https://github.com/ROCm/TheRock/issues/5029
+    $clVersionLine = $clOutput | Where-Object { $_ -match "Version \d+\.\d+" } | Select-Object -First 1
+    if ($clVersionLine -match "Version (\d+\.\d+)") {
```
erman-gurses, line 241: One concern here: What happens if this condition is not true - in case of `$clVersionLine` could be null or not matching with the regex.
It looks this block is skipped entirely and no version check is performed.

erman-gurses (commented):
LGTM overall - pointed one concern.

build_tools/validate_windows_install.ps1
ScottTodd: The output of `cl` has been pretty standard over the years, so I'm not too concerned about this matching code:
```
λ cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35226 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]
```

This script is also just informational for developers to check their environment, so if something is overlooked that doesn't have too severe of an impact (especially given that this script was not checking the version before, so it missed this failure until I noticed :P)

erman-gurses: Sounds Good :)

erman-gurses (approved):

PR #5035, opened 2026-05-04
[ci] Adding test reporting from gtest and ctest

geomin12:
For users (#3492) and actions-prolense, we will report test failures from ctest and gtest! This script will write to a file and print to console.

This is v1 as this covers only gtest and ctest, with future PRs covering additional test frameworks

I only covered hip-tests, hipblaslt and rocblas for testing purposes, as we are doing test script migration work. Will start rolling this out and applying it to rocm-libraries test scripts + rocm-systems test scripts while integrating with actions-prolense

Progress on #3492


geomin12:
testing more things - this may not work as I expected

jayhawk-commits:
I think it would be worth it to upload the generated report logs, similar to build logs.

Issue #4755, opened 2026-04-22
[issue] `offload-arch.exe` error-ing with `amdgpu64.dll` not found in `gfx1151` windows machines


geomin12:
When running `offload-arch.exe` with no existing `amdgpu64_*.dll` on the system and with `amdgpu64_7.dll` existing under `TheRock/therock-build/bin`, it still cannot find the amdhip64.dll with error: `Failed to load amdhip64.dll: Can't open: The specified module could not be found`.

When amdgpu64_7.dll did exist, offload-arch.exe worked fine! but it did produce the wrong architecture of gfx1150

This occurs on a Windows strix-halo machine (gfx1151)

Repro steps (on powershell admin):
```
# remove existing amdgpu64_*.dll files
$ takeown /f C:\Windows\System32\amdhip64_7.dll; icacls C:\Windows\System32\amdhip64_7.dll /grant Administrators:F; Remove-Item C:\Windows\System32\amdhip64_7.dll -Force
$ takeown /f C:\Windows\System32\amdhip64_6.dll; icacls C:\Windows\System32\amdhip64_6.dll /grant Administrators:F; Remove-Item C:\Windows\System32\amdhip64_6.dll -Force
$ cd ~; git clone https://github.com/ROCm/TheRock.git; cd TheRock; pip install boto3; python build_tools/install_rocm_from_artifacts.py --run-id 24434318178 --amdgpu-family gfx1151 --base-only
$ cd therock-build/lib/llvm/bin
$ ./offload-arch.exe --verbose
```

Please contact @geomin12 or @amd-justchen for access to strix-halo windows machine

ScottTodd:
This may be fixed by https://github.com/ROCm/TheRock/pull/5279 which cherry-picked https://github.com/ROCm/llvm-project/commit/fa0bf9b9c72e

ScottTodd:
This should be resolved now with https://github.com/ROCm/TheRock/pull/6902

Issue #4617, opened 2026-04-16
[Issue]: Windows CI: Test libhipcxx_hipcc failing with "error: unsupported HIP gpu architecture" on gfx110X"


idass1990:
### Problem Description

the **Test libhipcxx_hipcc** test is failing with the following error: "error: unsupported HIP gpu architecture" on gfx110X. 

hipinfo is run prior to the test returning the correct gfx arch (gfx1101), however this test uses offload-arch to determine the gfx arch and it is returning 'none' 

<img width="968" height="83" alt="Image" src="https://github.com/user-attachments/assets/8371a0e6-e54a-4b87-b7fb-a86408758405" />

The 'none' value is being passed into build command resulting in failure.

<img width="1239" height="144" alt="Image" src="https://github.com/user-attachments/assets/494adee1-d986-4c12-a42b-85566cdabcb2" />

**Error log:**
https://github.com/ROCm/TheRock/actions/runs/24493072786/job/71665146481#step:11:34

could potentially be related to offload-arch behaviour seen here: https://github.com/ROCm/TheRock/issues/4604

### Operating System

Windows

### CPU

AMD EPYC 9V64 32-Core Processor 

### GPU

gfx110x

### ROCm Version

7.13

AmosLewis:
still block https://github.com/ROCm/TheRock/pull/4761 and https://github.com/ROCm/TheRock/pull/4760

amd-chiranjeevi:
Hi @jharryma @geomin12 
Just wanted to check current status of this issue, can you please update latst status here ?

ScottTodd:
Guys we can't just ignore test failures like this for so long. offload-arch is actually broken now: https://github.com/ROCm/TheRock/issues/6571. This isn't just a machine/runner issue.

ScottTodd:
I have a fix for offload-arch up for review in LLVM: https://github.com/llvm/llvm-project/pull/209898. I've verified that at least our sanity check tests can be re-enabled with it: https://github.com/ROCm/TheRock/blob/a20ab791d74c94354704b791a85343babb83409b/tests/test_rocm_sanity.py#L89-L94

Hopefully that will also fix libhipcxx_hipcc tests on windows too.

ScottTodd:
As on https://github.com/ROCm/TheRock/actions/runs/30301391987/job/90131237052?pr=6902, tests should now have this startup behavior instead of errors:
```
[...]

Issue #4604, opened 2026-04-16
[Issue] Windows CI: test_rocm_sanity test_hip_printf fails with “Pcie atomics not enabled, hostcall not supported” and hipErrorIllegalState - gfx1151


chiranjeevipattigidi:
### Summary:
On Windows CI, tests/test_rocm_sanity.py::TestROCmSanity::test_hip_printf fails when running the compiled hipcc_check binary. The runtime reports hostcall/printf support is unavailable (“Pcie atomics not enabled, hostcall not supported”), the AQL dispatch fails, and the output buffer remains zeros, producing many mismatches.

### Context:
**Platform:**  Windows
**Arch:** `gfx1151`

### Full logs:
- https://github.com/ROCm/TheRock/actions/runs/24493072786/job/71645576215?pr=4592
- https://github.com/ROCm/TheRock/actions/runs/24493172287/job/71660041932?pr=4593

### Error log:
```
2026-04-16T13:12:43.9435792Z INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [None]$ 'C:\runner\_work\TheRock\TheRock\build\lib\llvm\bin\offload-arch.exe'
2026-04-16T13:12:43.9437476Z INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [C:\runner\_work\TheRock\TheRock\build\bin]$ 'C:\runner\_work\TheRock\TheRock\build\bin/hipcc' 'C:\runner\_work\TheRock\TheRock\tests\hipcc_check.cpp' -Xlinker '-rpath=C:\runner\_work\TheRock\TheRock\build\bin/../lib/' --offload-arch=gfx1150 -o hipcc_check.exe
2026-04-16T13:12:43.9439024Z INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [C:\runner\_work\TheRock\TheRock\build\bin]$ hipcc_check
2026-04-16T13:12:43.9439666Z ERROR    test_rocm_sanity:test_rocm_sanity.py:38 Command failed!
2026-04-16T13:12:43.9440122Z ERROR    test_rocm_sanity:test_rocm_sanity.py:39 command stdout:
2026-04-16T13:12:43.9440571Z ERROR    test_rocm_sanity:test_rocm_sanity.py:42 command stderr:
[...]

PR #5040, opened 2026-05-05
Fallback libhipcxx tests to AMDGPU_TARGETS

stellaraccident:
Partial landing split from #3928.

Changes:

- Make the libhipcxx test helper fall back to AMDGPU_TARGETS when offload-arch is missing, fails, or reports no architecture.

- Normalize comma and semicolon separated AMDGPU_TARGETS values to the semicolon-separated form expected by the test scripts.

- Fail early in the hipcc and hiprtc libhipcxx tests when no architecture can be determined instead of continuing with an empty arch.

- Add unit coverage for missing offload-arch (as that will soon come from its own artifact), empty offload-arch output, and no fallback targets.

Rationale:

- Artifact-based CI jobs may not always have a usable offload-arch in the output tree, while the job already knows the intended GPU targets through AMDGPU_TARGETS.

Validation:

- python -m unittest build_tools.github_actions.tests.libhipcxx_utils_test

- python -m py_compile build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py build_tools/github_actions/test_executable_scripts/test_libhipcxx_hipcc.py build_tools/github_actions/test_executable_scripts/test_libhipcxx_hiprtc.py build_tools/github_actions/tests/libhipcxx_utils_test.py

- git diff --cached --check

Generated with Codex

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
+++ b/build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
@@ -3,6 +3,7 @@ import platform
 import subprocess
 import sys
 import logging
+import re
 
 
 def prepend_env_path(env: dict, var_name: str, new_path: str):
@@ -25,6 +26,23 @@ def get_gpu_architecture_portable(therock_build_dir):
     """
     therock_build_dir = str(therock_build_dir)
     file_ending = ".exe" if platform.system() == "Windows" else ""
+
+    def get_architectures_from_targets() -> str | None:
+        amdgpu_targets = os.getenv("AMDGPU_TARGETS", "")
+        targets = [
+            target.strip()
+            for target in re.split(r"[,;]", amdgpu_targets)
+            if target.strip()
+        ]
+        if targets:
+            fallback_archs = ";".join(targets)
+            logging.info(
+                "Falling back to AMDGPU_TARGETS for HIP architectures: %s",
+                fallback_archs,
+            )
+            return fallback_archs
+        return None
+
     try:
         executable = therock_build_dir + f"/lib/llvm/bin/offload-arch{file_ending}"
         result = subprocess.run(
@@ -32,12 +50,14 @@ def get_gpu_architecture_portable(therock_build_dir):
         )
         lines = result.stdout.strip().split("\n")
         logging.info(f"DEBUG:{lines}")
-        return lines[-1]
+        if lines and lines[-1]:
+            return lines[-1]
+        return get_architectures_from_targets()
 
     except subprocess.CalledProcessError as e:
         print(f"Error executing offload-arch: {e}", file=sys.stderr)
         print(f"stderr: {e.stderr}", file=sys.stderr)
-        return None
+        return get_architectures_from_targets()
     except FileNotFoundError:
         print("Error: offload-arch command not found", file=sys.stderr)
-        return None
+        return get_architectures_from_targets()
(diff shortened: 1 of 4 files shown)


build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
```diff
@@ -25,19 +26,38 @@ def get_gpu_architecture_portable(therock_build_dir):
...
+                fallback_archs,
+            )
+            return fallback_archs
+        return None
+
     try:
         executable = therock_build_dir + f"/lib/llvm/bin/offload-arch{file_ending}"
         result = subprocess.run(
             [executable], capture_output=True, text=True, check=True
         )
```
ScottTodd, line 50: Using explicit targets from the environment is fine, however...

> Rationale: 
> * Artifact-based CI jobs may not always have a usable offload-arch in the output tree

They should though. We build `offload-arch` and it should work on all of our test runners. Is this for these issues? (none of which have been fully resolved yet)

* https://github.com/ROCm/TheRock/issues/4617
* https://github.com/ROCm/TheRock/issues/4604
* https://github.com/ROCm/TheRock/issues/4755

That being said, we should have test coverage for `offload-arch` itself and libhipcxx should be able to trust it based on those prior test results.

build_tools/github_actions/tests/libhipcxx_utils_test.py
```diff
@@ -0,0 +1,55 @@
...
+from pathlib import Path
+import subprocess
+import sys
+import unittest
+from unittest import mock
+
+sys.path.insert(
+    0,
+    os.fspath(Path(__file__).parents[1] / "test_executable_scripts"),
+)
```
ScottTodd, line 15: These `test_executable_scripts` are moving closer to project code in https://github.com/ROCm/rocm-libraries and https://github.com/ROCm/rocm-systems. @jayhawk-commits / @geomin12 what do you think about this adding a unit test for a test script itself?

ScottTodd (approved):
Generally LGTM. A few details relating to ongoing issue resolution and script migration work though.

Windows CI is still running on this PR _18 hours_ after starting:
* https://github.com/ROCm/TheRock/actions/runs/25351426613/job/74336403226?pr=5040 5h30m math-libs build for gfx110X-all
* https://github.com/ROCm/TheRock/actions/runs/25351426613/job/74365539627?pr=5040 configure test matrix was queued for _9h30m_
* https://github.com/ROCm/TheRock/actions/runs/25351426613/job/74458261377?pr=5040 sanity check is queued for _2h45m+_

build_tools/github_actions/tests/libhipcxx_utils_test.py
jayhawk-commits: `libhipcxx` is a special case as it is in neither super-repo, so these changes look OK to me for now until we get to tackling the `libhipcxx` repo.

Having unit tests for the test script itself is fine to me. We can eventually move them to their new locations.

jayhawk-commits (approved):
No blocking concerns from me.

We are iteratively moving the project-specific test scripts to their source repos (e.g., super-repos). We will be moving common, test utility functions that multiple scripts can leverage away from `build_tools` into a more appropriately-named directory, but that will happen independently of this PR.

geomin12 (approved):

build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
ScottTodd: offload-arch missing entirely was https://github.com/ROCm/TheRock/issues/5074. That should be fixed now

obersteiner:
Thanks for resolving this issue. I have noticed this issue myself but have not seen this PR here. One aspect that I would like to add is that ideally we want only the actual architecture inside of the AMDGPU_TARGET as compiling for multiple architectures can significantly increase the test time. Is it possible to set the AMDGPU_TARGETS explicitly to only the architecture of the family where it is actually executed on?

PR #5055, opened 2026-05-05
[ci] Removing scheduled CI ASAN

geomin12:
As multi arch CI ASAN is green, we are turning off CI ASAN scheduled. workflow dispatch is still available

No workflow errors here: https://github.com/ROCm/TheRock/actions/runs/25390129298 (but cancelled to save resources)


sa-faizal (approved):
lgtm, we can avoid redundancy and save resources this way

PR #5058, opened 2026-05-05
Add multi-arch (flat) upload mode for Python packages and tarballs

araravik-psd:
This PR mainly adds

- Add --multi-arch flag to support flat S3 upload layout
- Apply layout changes to both Python packages and tarballs via prefix selection
- Preserve existing arch-based behavior by default

1. Multi-arch upload mode (flat layout)

Introduced --multi-arch flag
Changes S3 key construction:

Default:

`<prefix>/<arch>/<package>`

Multi-arch:

`<prefix>/<package>`

2. Prefix selection for both wheels and tarballs

Updated --use-release-buckets handling:
Default → v3/whl/, v3/tarball/
Multi-arch → v4/whl/, v4/tarball/
Ensures both Python packages and tarballs are uploaded to the correct layout

3. Tarball uploads

Upload logic remains flat:

`<prefix>/<tarball>`

Destination changes based on selected prefix (standard vs multi-arch)

Testing:

https://gist.github.com/araravik-psd/904784e6f08af6c1fd79aab6b5d16da5

### DIFF
--- a/build_tools/packaging/upload_release_packages.py
+++ b/build_tools/packaging/upload_release_packages.py
@@ -35,6 +35,20 @@ TYPICAL USAGE (Command Line):
     --execute \
     --use-release-buckets
 
+  # Upload wheels only using multi-arch (flat layout)
+  python ./build_tools/packaging/upload_release_packages.py \
+    --input-dir=./downloads \
+    --multi-arch \
+    --execute
+
+  # Upload wheels and tarballs using multi-arch (flat layout, production buckets)
+  python ./build_tools/packaging/upload_release_packages.py \
+    --input-dir=./downloads \
+    --upload-tarballs \
+    --multi-arch \
+    --execute \
+    --use-release-buckets
+
 DIRECTORY STRUCTURE:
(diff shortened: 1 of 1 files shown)


marbre (approved):
Approving as I assume the functionality has been validated manually. @HereThereBeDragons please look more closely.

Solely looking at the changes, the functionality added seems mostly covered with `build_tools/github_actions/publish_rocm_to_release_buckets.py`. It should at least be possible to reuse code `_therock_utils.storage_location` and `_therock_utils.storage_backend` in the script. This script is also lacking tests and I am only approving as this is time sensitive. Before landing I expect to

1. Evaluate if reusing the functions pointed to above is reasonable. If it is, file an issue for a future refactoring.
2. If it is not, file an issue to add tests specifically as this is lacking any test so far.

Issues should be linked and PR description can be tidied up a bit further.

marbre:
Let's land as this is in production now @araravik-psd. This should give us enough confidence without needing to wait for a second review by @HereThereBeDragons.

PR #4771, opened 2026-04-23
Switch CI to `artifact_manager.py push`, remove `therock-archives`

ScottTodd:
## Motivation

Progress on
* https://github.com/ROCm/TheRock/issues/4769
* https://github.com/ROCm/TheRock/issues/3340
* https://github.com/ROCm/TheRock/issues/3343

We flipped the `THEROCK_FLAG_KPACK_SPLIT_ARTIFACTS` flag for multi-arch CI in TheRock, but non-multi-arch CI in TheRock, rocm-libraries, rocm-systems, etc. continues to keep the flag disabled. This unblocks part of what those workflows will need to switch.

> [!WARNING]
> When other repositories include this change, they will need to make similar changes to their workflow files. 

## Technical Details

* Non-multi-arch CI was using `post_build_upload.py`, which had `include=["*.tar.xz*"]` - that excluded the `.tar.zst` files that kpack split produces (see how https://therock-ci-artifacts.s3.amazonaws.com/24579570384-linux/index.html from https://github.com/ROCm/TheRock/actions/runs/24579570384 only contains generic artifacts). The `artifact_manager.py` script handles artifact uploading now and supports both `.tar.xz` and `.tar.zst`.

    We could also update `post_build_upload.py` to consider both extensions, but I'd continue chipping away at the old scripts so they'll be easier to remove once the migration to multi-arch and kpack split is complete.

* The `artifact_manager.py push` command generates its own compressed archives, so the `therock-archives` CMake target that invokes `fileset_tool artifact-archive` on its own can now be removed too. This resolves a comment:
    ```cmake
      # Generate artifact archive commands and save fingerprints.
      #
      # NOTE: In the multi-arch CI flow that kpack splitting enables, archive
      # generation moves out of the build system and into the upload phase.
      # Once fully transitioned to that model, archive generation logic here
      # can be removed entirely.
    ```

## Test Plan

* New unit tests
* Multi-arch CI and non-multi-arch CI on this PR
* Test run with `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON`

## Test Result

Non-multi-arch CI run https://github.com/ROCm/TheRock/actions/runs/25464518294?pr=4771 uploaded the expected files and tests ran. There are some test failures but these appear unrelated to which files are built/uploaded/downloaded/extracted.

The test run with the flag flip (https://github.com/ROCm/TheRock/actions/runs/24804899907) had a few issues so I'm sending this PR on its own without the flag flip.
[...]

Issue #1236, opened 2025-08-12
Release workflows do not freeze the commits they select


ScottTodd:
At the moment, our PyTorch release workflows use floating branch or tag names like
* `release/2.7`: https://github.com/ROCm/TheRock/blob/11739e9aed1c07d4b3ceba65c78dcaecb37eaf0c/.github/workflows/build_portable_linux_pytorch_wheels.yml#L65-L69
* `main`: https://github.com/ROCm/TheRock/blob/11739e9aed1c07d4b3ceba65c78dcaecb37eaf0c/.github/workflows/build_windows_pytorch_wheels.yml#L105-L116

These are the _versions_ that we want, but if a workflow fails and we re-run it, the workflow can pick a different commit. Furthermore, if multiple workflows are each schedule at slightly different times, each can use a different commit even without any re-runs - and that is what happens with one run of [`release_portable_linux_packages.yml`](https://github.com/ROCm/TheRock/actions/workflows/release_portable_linux_packages.yml) -> multiple runs of [`release_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/actions/workflows/release_portable_linux_pytorch_wheels.yml). Windows is especially prone to this because it uses pytorch `main`, which receives 10s of commits per day.

See for example:
* https://github.com/ROCm/TheRock/actions/runs/16909060869/job/47905678539#step:7:70 `HEAD is now at f33ce40bc0 [bucketing] Bucket only adjacent collectives to prevent reordering (#159983)`
* (two hours later, for the same nightly release) https://github.com/ROCm/TheRock/actions/runs/16910148775/job/47909418218#step:7:70 `HEAD is now at 7fbc22855c Update triton xpu commit to support python 3.14 (#160183)`

I think we should introduce a workflow like https://github.com/iree-org/iree/blob/main/.github/workflows/schedule_candidate_release.yml that chooses specific commits to release from and then triggers independent workflow runs using workflow_dispatch. We would be intentionally breaking the link between workflows that way, such that we could re-run the child workflows and have them retain the frozen commits determined during the "schedule" workflow run. In the child workflows, we would checkout with commands like `pytorch_torch_repo.py checkout --repo-hashtag EXACT_COMMIT --patchset main`

---

[...]

PR #5059, opened 2026-05-05
[ci] Adding external repo support for rocm-libraries multi-arch

geomin12:
Got a request to enable multi-arch CI workflow_dispatch in rocm-libraries. They were looking to test everything and build everything to make sure it works (with kpack)

I am running test here: https://github.com/ROCm/rocm-libraries/actions/runs/25452223287/job/74672238697   with branch changes here: https://github.com/ROCm/rocm-libraries/compare/develop...users/geomin12/multi-arch-ci

This is a good v1, where we can iterate as we migrate to multi-arch CI. somewhat relevant to de-dup work but fulfilling this request as developers want to get kpack tests ASAP

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux.yml
+++ b/.github/workflows/multi_arch_build_portable_linux.yml
@@ -58,6 +58,10 @@ on:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
         default: ""
+      external_repo_config:
+        description: "JSON config for external repo checkout (repository, ref, checkout_path, cmake_source_dir_var)"
+        type: string
+        default: ""
 
 permissions:
   contents: read
@@ -82,6 +86,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -105,6 +110,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -138,6 +144,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -165,6 +172,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -188,6 +196,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -211,6 +220,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -234,6 +244,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -257,6 +268,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -287,6 +299,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -310,6 +323,7 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
+      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/multi_arch_build_portable_linux.yml
+++ b/.github/workflows/multi_arch_build_portable_linux.yml
@@ -58,10 +58,6 @@ on:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
         default: ""
-      external_repo_config:
-        description: "JSON config for external repo checkout (repository, ref, checkout_path, cmake_source_dir_var)"
-        type: string
-        default: ""
 
 permissions:
   contents: read
@@ -86,7 +82,6 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
-      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -110,7 +105,6 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
-      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
@@ -144,7 +138,6 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository }}
       ref: ${{ inputs.ref }}
-      external_repo_config: ${{ inputs.external_repo_config }}
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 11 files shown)


.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -148,21 +160,14 @@ jobs:
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             ${{ inputs.build_variant_cmake_preset && format('--preset={0}', inputs.build_variant_cmake_preset) }} \
+            ${{ inputs.external_repo_config != '' && format('-D{0}={1}', fromJSON(inputs.external_repo_config).cmake_source_dir_var, fromJSON(inputs.external_repo_config).checkout_path) || '' }} \
```
ScottTodd: This does look like a pretty minimal change, but I'm concerned about how this lets you inject arbitrary `-D{ANY_CMAKE_OPTION_YOU_WANT}={ANY_VALUE_YOU_WANT}` into the configure command while being advertised as just for external repository configuration 🤔. If we really want to support arbitrary options, we could pick names that reflect what it is actually enabling like `extra_cmake_options`.

geomin12: ah i would want to keep it to external repo specific. not a fan of custom cmake options going in! i'll update, thanks

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -93,6 +97,14 @@ jobs:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref }}
 
+      - name: Checkout external repository
+        if: ${{ inputs.external_repo_config != '' }}
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ fromJSON(inputs.external_repo_config).repository }}
+          ref: ${{ fromJSON(inputs.external_repo_config).ref }}
+          path: ${{ fromJSON(inputs.external_repo_config).checkout_path }}
```
ScottTodd, line 107: Are we only ever going to have one external repository? There are multiple submodules in TheRock. We can certainly start here, but I wonder if transitioning to a manifest model will help - pass the manifest (a file listing each submodule along with commits, repository URLs, etc.) as an input and then the project fetches all sources based on its contents. Then we can model this more natively in scripts instead of relying on github actions workflow code to know "this JSON --> run actions/checkout --> configure CMake with conditional options".

I'm planning on picking up https://github.com/ROCm/TheRock/issues/1236 soon to at least freeze the commits we use for pytorch and I'll keep this in mind for the design there.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1055,7 +1055,13 @@ def configure(ci_inputs: CIInputs, git_context: GitContext) -> CIOutputs:
...
+    # Skip path filtering for external repos (e.g., rocm-libraries calling TheRock workflows)
+    skip_path_filters = os.environ.get("SKIP_PATH_FILTERS", "").lower() == "true"
+
+    if skip_path_filters:
+        # External repo: skip path filtering, run everything
+        git_context = GitContext.empty()
+    elif ci_inputs.is_pull_request or ci_inputs.is_push:
         # 'pull_request' and 'push' events can use the list of changed files
         # compared to the "prior commit" to affect job selections/options.
         git_context = GitContext.from_repo(base_ref=ci_inputs.base_ref)
```
ScottTodd, line 1070: Can you elaborate on the new comments here? I thought the intent was for rocm-libraries calling these workflows to provide its own path filters (and perhaps its own entire set of decision logic, not just filtering "should or shouldn't run")? Is this "run everything" the expected end state? If not, can you add a TODO?

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -93,6 +97,14 @@ jobs:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref }}
 
+      - name: Checkout external repository
+        if: ${{ inputs.external_repo_config != '' }}
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ fromJSON(inputs.external_repo_config).repository }}
```
ScottTodd, line 105: > I am running test here: https://github.com/ROCm/rocm-libraries/actions/runs/25404205067/job/74511163703 with branch changes here: [ROCm/rocm-libraries@develop...users/geomin12/multi-arch-ci](https://github.com/ROCm/rocm-libraries/compare/develop...users/geomin12/multi-arch-ci)

I'm noticing a few things in the logs. Let's triple check that this is actually checking out and building what we expect, with clear logs.

* When run in https://github.com/ROCm/rocm-libraries, we get a triply nested directory structure:
    ```
    /__w/rocm-libraries/                 ←  runner work dir (named after calling repo)
    /__w/rocm-libraries/rocm-libraries/  ← GitHub workspace (always <work>/<repo>), for TheRock
    /__w/rocm-libraries/rocm-libraries/rocm-libraries/  ← checkout_path="rocm-libraries" subdirectory
    ```
* the compiler-runtime stage spends 4+ minutes checking out rocm-libraries when it doesn't need it
* https://github.com/ROCm/rocm-libraries/actions/runs/25404205067/job/74517360333 this runs fetch_sources.py and tries to init the rocm-libraries submodule on top of the existing rocm-libraries checkout. We could choose a non-colliding path and then have fetch_sources.py skip the submodule when it will be provided by another directory already on disk

I'm scared of how brittle this is. The manifest approach is looking more attractive now...

.github/workflows/multi_arch_build_portable_linux.yml
davidd-amd, line 1: I don't quite understand what this file does but it seems like there is considerable code dup that factoring out could help avoid a consistency bug.

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: indeed it is not the final state! it'll be paired with #4570 . I'll add a better comment for this

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
geomin12: good point (for previous manifest comment and here)! I made some additional changes to make sure we are checking out the right place. this is a v1 request to run multi-arch on workflow dispatch only. 

once the run https://github.com/ROCm/rocm-libraries/actions/runs/25450700813 confirms all is good, we can work towards using a manifest and optimizing / improving with the inclusion of path filtering / component-based building in a follow up PR!

.github/workflows/multi_arch_build_portable_linux.yml
geomin12: this is initial work towards that multi-arch CI workflow dispatch. as mentioned above, once this lands and gets working, i'll work on a manifest file and optimizing so it'll be less de-dup / faster runs

ScottTodd:
> but fulfilling this request as developers want to get kpack tests ASAP

Note that kpack and multi-arch are not tightly coupled. We can build with kpack without migrating to multi-arch. I have a few PRs building towards letting us flip the flag for the non-multi-arch CI workflows:
* https://github.com/ROCm/TheRock/pull/5027
* https://github.com/ROCm/TheRock/pull/4771

ScottTodd:
> I am running test here: https://github.com/ROCm/rocm-libraries/actions/runs/25450700813 with branch changes here: [ROCm/rocm-libraries@develop...users/geomin12/multi-arch-ci](https://github.com/ROCm/rocm-libraries/compare/develop...users/geomin12/multi-arch-ci)

That looks like it should have worked. Upload failures due to permissions though: https://github.com/ROCm/rocm-libraries/actions/runs/25450700813/job/74666864425#step:18:66 (I've seen that a few times on other PRs... not sure what is breaking where, seems transient :/)

ScottTodd (approved):

geomin12:
> > I am running test here: https://github.com/ROCm/rocm-libraries/actions/runs/25450700813 with branch changes here: [ROCm/rocm-libraries@develop...users/geomin12/multi-arch-ci](https://github.com/ROCm/rocm-libraries/compare/develop...users/geomin12/multi-arch-ci)
> 
> That looks like it should have worked. Upload failures due to permissions though: https://github.com/ROCm/rocm-libraries/actions/runs/25450700813/job/74666864425#step:18:66 (I've seen that a few times on other PRs... not sure what is breaking where, seems transient :/)

<s>i saw it :) i am fixing, it's due to role not working in rocm-libraries as we should be using external role</s> - never mind, i see the cool composite action already doing this

build_tools/_therock_utils/s3_buckets.py
```diff
@@ -48,7 +48,7 @@ def write_access_iam_role(self) -> str | None:
 s3_bucket_configs = [
     # CI (self-hosted runners include credentials for therock-ci-artifacts-external)
     S3BucketConfig("therock-ci-artifacts", iam_role="therock-ci"),
-    S3BucketConfig("therock-ci-artifacts-external", iam_role=None),
+    S3BucketConfig("therock-ci-artifacts-external", iam_role="therock-ci-external"),
```
ScottTodd: I thought this shouldn't be needed, since the runners already have those base credentials? Or do the base credentials not work in rocm-libraries for some reason? PRs from forks over there have been working just fine (e.g. https://github.com/ROCm/rocm-libraries/actions/runs/25405646332/job/74515900277?pr=7087#step:16:13)

geomin12: ah so the base runners do have them for forked repos (this is where we do the docker mount for secrets), but for non-forked runs, it still behaves the same with configure AWS action: https://github.com/ROCm/rocm-libraries/blob/1f8654f9532412edd2a5014a58eb6b45d0d1204a/.github/workflows/therock-ci-linux.yml#L119-L124

we just didn't know this as TheRock is the only repo that uses multi-arch (until now!)

ScottTodd: Windows was able to upload just fine: https://github.com/ROCm/rocm-libraries/actions/runs/25450700813/job/74671279449#step:21:67. Maybe retry the Linux to see if it was a flake?

I don't think we should need to explicitly assume the therock-ci-external role (we can't assume it for fork PRs, hence why those credentials are on the runners directly)

geomin12: huh it did work. https://github.com/ROCm/rocm-libraries/actions/runs/25455406889/job/74683453536#step:17:39, what a strange flaky error.

i'll try to get to the root cause for that though

PR #2872, opened 2026-01-12
[ci] Add resource spec file generation before running rocThrust ctest

amd-hsong:
## Motivation

When running ctest in parallel mode, a resource spec file must be provided. This PR adds the missing step of generating the resource spec file before running rocThrust ctest.

## Technical Details

The rocThrust library recently added an executable called "**generate_resource_spec**" to generate the resource spec file for ctest. This file describes the GPUs available on the system and it should be passed to ctest through the **--resource-spec-file** option whenever the --parallel or the -j option is used. Otherwise tests may fail.

With this fix, we no longer need the ctest option **--repeat until-pass:6**. Test results should be deterministic.

## Test Plan

Verify that the script of running ctest on rocThrust now calls "generate_resource_spec" and passes the genereated file to the --resource-spec-file option of ctest.

## Test Result

Test output shows that "generate_resource_spec" is executed and the --resource-spec-file option is added when running test_rocthrust.py.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5060, opened 2026-05-05
[ci] Add resource spec file generation before running rocPrim ctest

amd-jmahovsky:
## Motivation

When running ctest in parallel mode, a resource spec file must be provided. This PR adds the missing step of generating the resource spec file before running rocPrim ctest.

## Technical Details

The rocPrim library recently added an executable called "generate_resource_spec" to generate the resource spec file for ctest. This file describes the GPUs available on the system and it should be passed to ctest through the --resource-spec-file option whenever the --parallel or the -j option is used. Otherwise tests may fail.

With this fix, we no longer need the ctest option --repeat until-pass:6. Test results should be deterministic.

## Test Plan

Verify that the script of running ctest on rocPrim now calls "generate_resource_spec" and passes the generated file to the --resource-spec-file option of ctest.

## Test Result

Test output shows that "generate_resource_spec" is executed and the --resource-spec-file option is added when running test_rocprim.py.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_rocprim.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rocprim.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+import platform
 import shlex
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


stanleytsang-amd (approved):

build_tools/github_actions/test_executable_scripts/test_rocprim.py
```diff
@@ -101,16 +102,49 @@
...
+rocm_base = Path(THEROCK_BIN_DIR).resolve().parent
+ld_paths = [
+    rocm_base / "lib",
+]
+ld_paths_str = os.pathsep.join(str(p) for p in ld_paths)
+existing_path = os.environ.get("PATH", "")
+existing_ld_path = os.environ.get("LD_LIBRARY_PATH", "")
+env_vars = os.environ.copy()
+env_vars["PATH"] = (
+    f"{THEROCK_BIN_DIR}{os.pathsep}{existing_path}"
```
geomin12, line 115: do we assume that there are dashes in between? is this correct?

geomin12 (commented):
rocprim tests are not working:
```
Internal ctest changing into directory: /__w/TheRock/TheRock/build/bin/rocprim
Test project /__w/TheRock/TheRock/build/bin/rocprim
Could not read/parse resource spec file resources.json: 
File not found: resources.json
No tests were found!!!
```

amd-jmahovsky:
@geomin12 Where are you seeing that error with the resources.json not being found?  This might be expected behavior because the utility that generates that file hasn't yet been merged into rocPrim as it's another PR right now.  I don't know if the TheRock is smart enough to know to pull down that branch or if it's just pulling down the develop branch.  Maybe I need to merge that work before this PR can complete.

build_tools/github_actions/test_executable_scripts/test_rocprim.py
amd-jmahovsky: I'm not sure.  This work was ported from https://github.com/ROCm/TheRock/pull/2872/changes.  I had assumed rocPrim and rocThrust would be consistent with each other.

amd-jmahovsky:
@geomin12 I confirmed with the rocPrim team that this is not expected to work until the rocPrim changes have been merged.  I'll let you know when that happens so you can revisit this PR.

amd-jmahovsky:
@geomin12 I think I need to merge both the rocprim changes and this PR at the same time because these changes depend on the rocprim changes, and TheRock may be broken without them. I.e. if I merge just the rocprim PR, it'll break TheRock.

I'm just awaiting a documentation clarification on the rocprim PR.  Then I'll merge both PRs... if there is an unforeseen issue with this PR, I can always revert the changes - it's fairly simple.

ScottTodd:
> @geomin12 I think I need to merge both the rocprim changes and this PR at the same time because these changes depend on the rocprim changes, and TheRock may be broken without them. I.e. if I merge just the rocprim PR, it'll break TheRock.

Both repositories use pinned commit refs to avoid such issues (TheRock has rocm-libraries as a git submodule, rocm-libraries uses a pinned ref in github actions workflows to load other submodules, test scripts, workflow code, etc.).

PR #5066, opened 2026-05-05
[ci] Fixing dep rpm packages for forked upload

geomin12:
Noticied that RPM and DEB packages fail to upload: https://github.com/ROCm/TheRock/actions/runs/25237692770/job/74244690857?pr=4998 due to the build image not correctly configured (or even used!).

This change uses a proper docker image as well as proper credentials for forked repos


ScottTodd (commented):
Oh. That would do it... Thanks.

Please wait for CI to complete on this. I have some doubts about the "Install Python requirements" and "Install System requirements" steps working in the container 😅

ScottTodd:
Oops, CI didn't run on this. Need to edit https://github.com/ROCm/TheRock/blob/d41dc8475f7dcb0d17d74e036ff9c4be9dbf8f48/build_tools/github_actions/configure_ci_path_filters.py#L165-L179 too

(or you can trigger the workflow directly to skip past the rest of the CI jobs)

.github/workflows/build_native_linux_packages.yml
```diff
@@ -91,8 +91,10 @@ jobs:
     runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
     outputs:
       package_install_url: ${{ steps.upload-packages.outputs.package_repository_url }}
+    container:
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:702a5133851e6d1daf1207d2c9fbb01c2667914a5b6dc5a01faeb3ce66ea6421
```
nunnikri: Current job needs to be run on Ubuntu. Ubuntu2404/Ubuntu2204

ScottTodd: Ah. Then this job might need to set the `AWS_SHARED_CREDENTIALS_FILE` env var per https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#authentication ... though I would expect the credential file on the runners to already be used by default without that 🤔

geomin12: ah it isn't as it's only used to docker volume mounts, i'll fix

build_tools/github_actions/configure_ci_path_filters.py
```diff
@@ -176,6 +176,7 @@ def is_ci_run_required(paths: Optional[Iterable[str]]) -> bool:
     "test_rocm_wheels.yml",
     "test_sanity_check.yml",
     "test_component.yml",
+    "build_native*packages.yml",
```
ScottTodd: > Oops, CI didn't run on this.

Thanks for the fix here. I also just sent https://github.com/ROCm/TheRock/pull/5089 for a general solution

geomin12:
it works yay! sorry it took a while - will fix build pytorch wheels next after this PR

.github/workflows/multi_arch_build_native_linux_packages.yml
```diff
@@ -53,15 +53,21 @@ jobs:
   build_native_packages:
     name: Build ${{ inputs.native_package_type }} packages
     runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
+    container:
+      image: ubuntu:24.04
+      options: -v /runner/config:/home/awsconfig/
     outputs:
       package_repository_url: ${{ steps.upload-packages.outputs.package_repository_url }}
     env:
+      AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
```
ScottTodd, line 62: Why do we need to run in a container and inject this credentials file if the credentials are already on the host?

I don't mind running these in an explicit container (looks like it just takes ~2 seconds to init: https://github.com/ROCm/TheRock/actions/runs/25897755289/job/76127352092?pr=5066)... I'm just curious why this is needed at all, as we might miss it for other workflows in the future.

ScottTodd (approved):

.github/workflows/multi_arch_build_native_linux_packages.yml
geomin12: yeah so i thought it mounted to file path normally, but it is actually not accessible with regular directories and must be mounted via a `-v` option. i didn't want to adjust the config in the runner setup and we should run in containers anyway!

PR #5082, opened 2026-05-06
Rework concurrency groups so workflow_dispatch and push are independent

ScottTodd:
## Motivation

Runs like https://github.com/ROCm/TheRock/actions/runs/25444009603 on `push` get cancelled when someone triggers a run on `main` (the default) using `workflow_dispatch`, even if the inputs are different: https://github.com/ROCm/TheRock/actions/runs/25444348607.

This changes the concurrency groups so that `workflow_dispatch` and `push` events do not overlap concurrency groups with each other or `pull_request` events while keeping the same behavior as before on pull requests.

## Technical Details

See https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

Note that this does _not_ fix a race between the `opened` and `labeled` events when creating a PR with labels already attached. For that we'd need something like https://github.com/iree-org/iree/blob/089f3d60d6944f6868d4c41d841b3d15eee30f6e/.github/workflows/benchmark_trigger.yml to look for the labels and choose whether or not to re-run the workflows.

## Test Plan

* Test `pull_request`
  * PR 1, `opened` event: https://github.com/ScottTodd/TheRock/actions/runs/25447478311
  * PR 1, `labeled` event: https://github.com/ScottTodd/TheRock/actions/runs/25447481616 (cancelled `opened` run, as expected) 
  * PR 2, `opened` event: https://github.com/ScottTodd/TheRock/actions/runs/25448478142
  * PR 2, `synchronize` event: https://github.com/ScottTodd/TheRock/actions/runs/25448564514 (cancelled `opened` run, as expected)
* Test `workflow_dispatch`
  * Trigger 1: https://github.com/ScottTodd/TheRock/actions/runs/25447052347
  * Trigger 2: https://github.com/ScottTodd/TheRock/actions/runs/25447192809 (same branch as other workflow_dispatch, did not cancel 1)
  * Trigger 3: https://github.com/ScottTodd/TheRock/actions/runs/25449341353 (same branch as PR, did not cancel that run)
* `push` untested (should be the same as `workflow_dispatch`, can watch postsubmit)

## Submission Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,11 +60,15 @@ permissions:
   contents: read
 
 concurrency:
-  # A PR number if a pull request and otherwise the commit hash. This cancels
-  # queued and in-progress runs for the same PR (presubmit) or commit
-  # (postsubmit). The workflow name is prepended to avoid conflicts between
-  # different workflows.
-  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  # See https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
+  #
+  # Only allow one run at a time for 'pull_request' events.
+  # Allow multiple runs for 'push' and 'workflow_dispatch'.
+  #
+  # `github.workflow`: each workflow gets its own concurrency group
+  # `github.head_ref`: the branch name (only defined for pull_request events)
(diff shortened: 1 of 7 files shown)


geomin12 (approved):
lgtm, the tests in PR description looks to cover the scenarios and this is great! we can observe on post-submit as well

PR #5087, opened 2026-05-06
[rocSOLVER] Increase test timeout

tfalders:
## Motivation

Nightly rocSOLVER tests intermittently time out on theRock CI. The variance in test time will need to be investigated, but increasing the time out value was suggested as a short-term solution.

## Technical Details

Increased test time out for rocSOLVER from 60 mins to 120 mins.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -154,9 +154,8 @@ test_matrix = {
     "rocsolver": {
         "job_name": "rocsolver",
         "fetch_artifact_args": "--blas --tests",
-        # 68350(approx) tests needs 48 mins, so 48 mins / 2 shards = 24 mins per shard
-        # 24 mins + 20% margin = 30 mins => ~40 mins (considering gpu delays and lags)
-        "timeout_minutes": 60,
+        # Extended tests on math-ci take approx 5 hrs (as of May 5, 2026)
+        "timeout_minutes": 120,
         "test_script": f"python {_get_script_path('test_rocsolver.py')}",
         # Issue for adding windows tests: https://github.com/ROCm/TheRock/issues/1770
         "platform": ["linux"],


TorreZuk (approved):
New tests have been added, and can reduce after variability of runs is reduced.

tony-davis (approved):
Looks good.

jayhawk-commits:
Please add some details on how many more tests were added and how much longer the execution time needs. Eventually, this can be captured automatically with prolense. CC @amd-shiraz

tfalders:
> Please add some details on how many more tests were added and how much longer the execution time needs. Eventually, this can be captured automatically with prolense. CC @amd-shiraz

The newest tests are for SYTRS and SYTRS2; there are 3456 new test cases. I was able to run these tests in about 47 seconds on gfx942. But because of the test variance on TheRock CI, we also saw individual test cases taking multiple seconds in the nightly tests.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -154,9 +154,8 @@ def _get_script_path(script_name: str) -> str:
     "rocsolver": {
         "job_name": "rocsolver",
         "fetch_artifact_args": "--blas --tests",
-        # 68350(approx) tests needs 48 mins, so 48 mins / 2 shards = 24 mins per shard
-        # 24 mins + 20% margin = 30 mins => ~40 mins (considering gpu delays and lags)
-        "timeout_minutes": 60,
+        # Extended tests on math-ci take approx 5 hrs (as of May 5, 2026)
+        "timeout_minutes": 120,
```
geomin12, line 158: let's work on test standardization for rocsolver soon! https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md

having a developer wait 360 minutes for tests to run in a PR is not good

perhaps, could we do just standard tests instead of extended? and if nightly, we run extended?

jayhawk-commits (approved):
Thanks for providing details on the added tests.

tfalders:
Thanks @jayhawk-commits. Can I merge?

Issue #5021, opened 2026-05-04
[Issue] Linux PyTorch : segfault (exit 139) in `test_nn.py Conv1d CUDA tests`


amd-chiranjeevi:
## Context:
- Platform: Linux
- Impacted Archs: `gfx94x, gfx950, gfx120x, gfx110x`
- Pytorch Versions: `2.9, 2.10, 2.11, nightly`
- Rocm Version: `7.13.0a20260504`

## Tests Failed:
```
test_Conv1d_circular_stride2_pad2 FAILED
test_Conv1d_cuda_fp32 FAILED
test_Conv1d_cuda_tf32 FAILED
test_Conv1d_dilated_cuda_fp32 FAILED
test_Conv1d_pad2 FAILED
test_Conv1d_pad2_cuda_fp32 FAILED
test_Conv1d_pad2_cuda_tf32 FAILED
test_Conv1d_pad_same_cuda_fp32 FAILED
test_Conv1d_pad_same_cuda_tf32 FAILED
test_Conv1d_reflect_stride2_pad2 FAILED
test_Conv1d_replicate_stride2_pad2_cuda_fp32 FAILED
test_Conv1d_stride_cuda_tf32 FAILED
```

## Run Links:
- gfx950-dcgpu: https://github.com/ROCm/TheRock/actions/runs/25311104571
- gfx94x: https://github.com/ROCm/TheRock/actions/runs/25312522580
- gfx120x: https://github.com/ROCm/TheRock/actions/runs/25312548814
- https://github.com/ROCm/TheRock/actions/runs/25316706893

### Error snippet:
```
2026-05-04T11:41:46.6847781Z Fatal Python error: Segmentation fault
2026-05-04T11:41:46.6847971Z 
2026-05-04T11:41:46.6848182Z Thread 0x000076d895a006c0 (most recent call first):
2026-05-04T11:41:46.6848512Z   <no Python frame>
2026-05-04T11:41:46.6848647Z 
2026-05-04T11:41:46.6848799Z Current thread 0x000076e23ea53b80 (most recent call first):
2026-05-04T11:41:46.6849893Z   File "/__w/TheRock/TheRock/.venv/lib/python3.10/site-packages/torch/autograd/gradcheck.py", line 822 in _get_analytical_vJu_backward_mode
[...]

PR #5091, opened 2026-05-06
[CI] Skip test_Conv1d_zero_batch

rraminen:
Skip the below unit test across all the archs for a cleaner run

`external-builds/pytorch/pytorch/test/test_nn.py::TestNN::test_Conv1d_zero_batch`

This test fails with a `Segmentation fault (core dumped)` error, which isn’t related to PyTorch and may be due to changes in the ROCm version. (https://github.com/ROCm/TheRock/actions/runs/25273411730/job/74102474492)

This test passed with torch==2.11.0+rocm7.13.0a20260430 (gfx94x) (APR 30) - https://github.com/ROCm/TheRock/actions/runs/25150366706/job/73730936029

First failure occured with torch==2.11.0+rocm7.13.0a20260502 (gfx94X) (MAY 2) - https://github.com/ROCm/TheRock/actions/runs/25312522580/job/74208101156

### DIFF
--- a/external-builds/pytorch/skip_tests/generic.py
+++ b/external-builds/pytorch/skip_tests/generic.py
@@ -99,6 +99,11 @@ skip_tests = {
             # to be on the same device, but got weight is on cpu, different from other tensors on cuda:0 (when checking
             # argument in method wrapper_CUDA__miopen_rnn)"
             "test_rnn_check_device",
+            # This test passed with torch==2.11.0+rocm7.13.0a20260430 (gfx94x) (APR 30) - https://github.com/ROCm/TheRock/actions/runs/25150366706/job/73730936029
+            # First failure occured with torch==2.11.0+rocm7.13.0a20260502 (gfx94X) (MAY 2) - https://github.com/ROCm/TheRock/actions/runs/25312522580/job/74208101156
+            # This Segmentation fault not related to pytorch version
+            # Skipping this test for now to get a cleaner run
+            "test_Conv1d_zero_batch",
         ],
         "torch": [
             # FLAKY!! AssertionError: 'tensor([2.3000+4.j, 7.0000+6.j])' != 'tensor([2.30000+4.j, 7.00000+6.j])'


external-builds/pytorch/skip_tests/generic.py
```diff
@@ -99,6 +99,11 @@
             # to be on the same device, but got weight is on cpu, different from other tensors on cuda:0 (when checking
             # argument in method wrapper_CUDA__miopen_rnn)"
             "test_rnn_check_device",
+            # This test passed with torch==2.11.0+rocm7.13.0a20260430 (gfx94x) (APR 30) - https://github.com/ROCm/TheRock/actions/runs/25150366706/job/73730936029
+            # First failure occured with torch==2.11.0+rocm7.13.0a20260502 (gfx94X) (MAY 2) - https://github.com/ROCm/TheRock/actions/runs/25312522580/job/74208101156
+            # This Segmentation fault not related to pytorch version
+            # Skipping this test for now to get a cleaner run
+            "test_Conv1d_zero_batch",
```
ScottTodd, line 106: If ROCm code is causing a segfault, that needs to get fixed. That's exactly what these tests are meant to exercise (pytorch is already tested against _stable rocm_ releases upstream, here is where we test pytorch against _latest rocm_)

ScottTodd (changes requested):

external-builds/pytorch/skip_tests/generic.py
pragupta: @ScottTodd  -- Yeah true, our builds were not getting promoted so this was just a desperate attempt to get us to green. But, yes, on a second thought, we should hold on these and wait for proper rocm fix to come in.

ScottTodd: Note that for multi-arch packages we're planning on removing the "promoted" concept entirely, since we haven't found a good way to make that work with the unified release index. See https://github.com/ROCm/TheRock/pull/5107. More reason to get tests passing :)

For segfaults though, we can blur the lines a bit... Windows tests for some torch versions in particular have been failing due to segfaults and that prevents getting clear test signal since the test runner crashes and does not run all expected tests.

In this case I'd like to see some triage performed and if we can identify a culprit for reverting quickly, especially if we already have a precise window where this regressed (April 30 - May 2).

HereThereBeDragons:
lets connect it with the issue #5021

Issue #4687, opened 2026-04-20
[Issue] Multi-arch CI prebuilt artifacts can use misaligned dist_amdgpu_targets


ScottTodd:
Reported here:
* https://github.com/ROCm/TheRock/issues/4683
* https://github.com/ROCm/TheRock/pull/4646#issuecomment-4282369045

https://github.com/ROCm/TheRock/actions/runs/24532353196
1. This built rocm python packages with `Fetching artifacts for multiple families: gfx900;gfx906;gfx908;gfx90a` then
    ```
    Uploading packages
    ------------------
    [INFO] Found 5 top-level package files in C:\home\runner\_work\TheRock\TheRock\packages\dist:
      - rocm_sdk_core-7.13.0.dev0+74a62cb9bc9c95b6ea37e9b5b8d366cf00ec62ee-py3-none-win_amd64.whl
      - rocm_sdk_libraries_gfx900-7.13.0.dev0+74a62cb9bc9c95b6ea37e9b5b8d366cf00ec62ee-py3-none-win_amd64.whl
      - rocm_sdk_libraries_gfx906-7.13.0.dev0+74a62cb9bc9c95b6ea37e9b5b8d366cf00ec62ee-py3-none-win_amd64.whl
      - rocm_sdk_libraries_gfx908-7.13.0.dev0+74a62cb9bc9c95b6ea37e9b5b8d366cf00ec62ee-py3-none-win_amd64.whl
      - rocm_sdk_libraries_gfx90a-7.13.0.dev0+74a62cb9bc9c95b6ea37e9b5b8d366cf00ec62ee-py3-none-win_amd64.whl
    ```
2. Building pytorch then used different gpu targets:
    ```
    Installing collected packages: rocm-sdk-libraries-gfx90a, rocm-sdk-devel, rocm-sdk-core, rocm
    ...
    ++ Capture [C:\home\runner\_work\TheRock\TheRock]$ 'C:\home\runner\_work\_tool\Python\3.12.10\x64\python.exe' -m rocm_sdk targets
      Using default PYTORCH_ROCM_ARCH from rocm-sdk targets: gfx1100;gfx1101;gfx1102;gfx1103;gfx1151;gfx1200;gfx1201
    ...
    +++ Building pytorch:
    ++ Exec [B:\src\torch]$ 'C:\home\runner\_work\_tool\Python\3.12.10\x64\python.exe' setup.py bdist_wheel
    :: Env:
      PYTHONUTF8=1
[...]

PR #5095, opened 2026-05-06
Add release prebuilt artifact reuse mode

marbre:
## Motivation

In rare cases release re-runs need to reuse already-built artifacts. Those might be manually patched or we restrict the re-run to a subset of targets and solely use the release run to repackage for this limited subset. The release workflows will build tarballs, Python packages and native Linux packages without starting a source build.

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -29,13 +29,20 @@ on:
       ref:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
+      prebuilt_prefix:
+        description: >-
+          Optional fixed prefix to copy artifacts from instead of building
+          from source. Empty selects normal source-build mode.
+        type: string
+        default: ""
 
 permissions:
   contents: read
 
 jobs:
-  build_artifacts:
-    name: Build Artifacts
+  build_source:
+    name: Build Artifacts (source)
+    if: ${{ inputs.prebuilt_prefix == '' }}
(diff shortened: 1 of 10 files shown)


ScottTodd (commented):
We discussed this PR offline. Some ideas:

* Frame as "release repackaging" or "release promotion" instead of "prebuilt artifacts"
* Allow CI workflows to mix prebuilt and built-from-source artifacts, draw a harder line in release workflows to only either use _all_ prebuilt (for repacking) or _all_ built-from-source
* Move to https://github.com/ROCm/rockrel if focused on promotion

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -53,6 +60,41 @@ jobs:
       contents: read
       id-token: write
 
+  copy_prebuilt:
+    name: Copy Prebuilt Artifacts
+    uses: ./.github/workflows/copy_prebuilt_artifacts.yml
```
ScottTodd, line 65: Another thing to watch for:

the `base_lib_generic.tar.zst` contains `dist_info.json` with contents like
```json
{
  "dist_amdgpu_targets" : "gfx942;gfx1100;gfx1101;gfx1102;gfx1103;gfx1151;gfx1200;gfx1201;gfx950"
}
```

just repackaging a subset of existing artifacts (e.g. choosing to release a subset of the targets that were built in a nightly release for a stable release) with a changed version will include the original list there, resulting in tools like `rocm-sdk targets` returning a list that may not match expectations (see also https://github.com/ROCm/TheRock/issues/4687)

PR #5100, opened 2026-05-07
[CI] Skip failing tests 

rraminen:
## Motivation

- Skip the tests from the below modules for release/2.11 for a clean PyTorch Wheels (Full Suite) CI job

```
- test_utils
- test_multiprocessing
- test_serialization
- test_modultes
- test_export
- test_inductor
- test_functorch
- test_jit_fuser_te
- torch_config_hash_determinism
- test_spectral_ops
```

- Adds code to Install C++ build tools for PyTorch tests


external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,11 +67,94 @@
...
+            "test_debug_set_in_ci",
+        ],
+        "modules": [
+            # TestModuleCUDA - CTCLoss cpu/gpu parity scalar mismatch
+            "test_cpu_gpu_parity_nn_CTCLoss_cuda_float32",
+            # TestModuleCUDA - CTCLoss forward scalar mismatch
+            "test_forward_nn_CTCLoss_cuda_float32",
+        ],
+        "export": [
+            # TestExportOnFakeCudaCUDA - subprocess import fails: missing librocm_sysdeps_liblzma.so.5
```
marbre, line 96: This rather sounds like a bug that should be fixed.

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,11 +67,94 @@
...
+        ],
+        "inductor": [
+            # inductor/test_aot_inductor_package: AOTI C++ package tests need
+            # more complete CMake/runtime library-path handling in the wheel CI lane.
+            "test_compile_after_package_multi_arch",
+            "test_compile_after_package_static",
+            "test_compile_standalone_cos",
+            "test_compile_with_exporter",
+            "test_compile_with_exporter_weights",
+            #Also failed on https://github.com/ROCm/TheRock/actions/runs/24898379109
```
marbre, line 115: Logs will expire, so this needs an issue to track.

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,11 +67,94 @@
...
+            "test_fake_export_nn_functional_conv2d_cuda_float32",
+            "test_fake_export_nn_functional_instance_norm_cuda_float32",
+            "test_fake_export_nn_functional_multi_margin_loss_cuda_float32",
+            "test_fake_export_nn_functional_scaled_dot_product_attention_cuda_float32",
+            "test_fake_export_nonzero_cuda_float32",
+            "test_preserve_original_behavior_cuda",
+        ],
+        "inductor": [
+            # inductor/test_aot_inductor_package: AOTI C++ package tests need
+            # more complete CMake/runtime library-path handling in the wheel CI lane.
```
marbre, line 109: As the comment already says this might need proper handling in the wheel, why are tests skipped instead of aiming for a fix?

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,11 +67,94 @@
...
+            #Also failed on https://github.com/ROCm/TheRock/actions/runs/24898379109
+            "test_flex_attention_logging_cuda",
+            "test_linalg_eig_stride_consistency_cuda",
+            "test_linalg_eig_stride_consistency_cuda",
+            "test_linalg_eig_stride_consistency_dynamic_shapes_cuda",
+            "test_repeated_calling_cuda",
+            #new test
+            "test_run2run_determinism_model_name_DistillGPT2_training_or_inference_inference_precision_amp",
+            #Passed in https://github.com/ROCm/TheRock/actions/runs/24898379109
+            "test_return_aux_deprecation_warnings_cuda_float16",
```
marbre, line 124: If this is is passing, why is it on the exclude list? As above, logs will expire and this might need a proper issue.

marbre (changes requested):
GitHub action logs will expire and as there are no issues filed providing further details. What is the plan to fix those test failures? The plan cannot be to just skip tests.

pragupta:
@marbre  -- We are trying to get some green state for full pytorch UT testing and then tackle them one by one by root causing them. With so many failures currently in pytorch full testing and with constantly moving rocm, it's been hard to achieve stability. So, yes, this is a hammer. But, we need to get this in and then start working on unskipping them, @rraminen  has another task on her scrum board to start to root cause each of these skips.

external-builds/pytorch/skip_tests/pytorch_2.11.py
HereThereBeDragons: i agree librocm_sysdebs_liblzma must be available otherwise it is a bug

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -275,6 +275,11 @@ jobs:
           echo "CC=${ROCM_HOME}/lib/llvm/bin/clang" >> "${GITHUB_ENV}"
           echo "CXX=${ROCM_HOME}/lib/llvm/bin/clang++" >> "${GITHUB_ENV}"
 
+      - name: Install C++ build tools for PyTorch tests
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends build-essential
```
ScottTodd, line 281: This workflow makes an attempt to be portable to Windows, see the `if: ${{ runner.os == 'Windows'` steps below. These commands that will only ever work on Linux should either be
1. Moved into a test dockerfile used on Linux so they aren't downloaded from apt every workflow run (canonical's servers have gone offline a few times recently, leading to workflow instability that would be avoided if all deps were in our own dockerfiles)
2. Gated by a `if: ${{ runner.os == 'Linux' }}` condition

That being said, we may also want to segment the tests that need `build-essential` somehow, similar to tests that need rocm development headers/files. Users should get a functional install of pytorch with just `pip install torch`, and anything that needs to be separately installed at the operating system level should be treated with care. If only 1% of the software/tests need additional packages we don't want to miss that suddenly growing to 50%+ since all of our test workflows preinstall packages.

external-builds/pytorch/skip_tests/pytorch_2.11.py
ScottTodd, line 1: Are these test failures/skips unique to torch version 2.11? What about 2.12: https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/skip_tests/pytorch_2.12.py ?

ScottTodd (commented):
+1 to the other comments, though I don't want to block this for too long, since all effort stabilizing these tests is appreciated

.github/workflows/test_pytorch_wheels_full.yml
HereThereBeDragons: agree. my old comment (which i never sent) was "why is this needed?".

definitely needs to be flagged behind linux and should have some explanation why it is needed in the first plac.e

external-builds/pytorch/skip_tests/pytorch_2.11.py
HereThereBeDragons: and please add a proper error message here - so we know why it was added to the skip list

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,7 +67,95 @@
...
+        ],
+        "functorch": [
+            #passed on https://github.com/ROCm/TheRock/actions/runs/24898379109
+            "test_torch_return_types_returns_cuda",
```
HereThereBeDragons, line 128: so are those tests flaky or why are the passing in one but are here added to the skip list?

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,7 +67,95 @@
...
+        ],
+        "jit_fuser_te": [
+            #passed on https://github.com/ROCm/TheRock/actions/runs/24898379109
```
HereThereBeDragons, line 131: same

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,7 +67,95 @@
...
+            "test_binary_div_ops",
+            "test_binary_ops",
+            "test_binary_tensor_scalar_ops",
+            "test_ternary_norm_ops",
+            "test_ternary_ops",
+            "test_unary_ops",
+            "test_where_ops",
+        ],
+        "torch_config_hash_determinism": [
+            #passed on https://github.com/ROCm/TheRock/actions/runs/24898379109
```
HereThereBeDragons, line 148: same

external-builds/pytorch/skip_tests/pytorch_2.11.py
```diff
@@ -67,7 +67,95 @@
...
+            "test_inductor_config_hash_portable_deterministic",
+            "test_inductor_config_hash_portable_without_ignore",
+        ],
+        "spectral_ops": [
+            #new failures
```
HereThereBeDragons, line 153: new failures - why? rocm or pytorch? would be good to triage a bit if it is a new failure to know where the regresison comes from

and please again: more error messages as comments here so we can better categorize the test failure

HereThereBeDragons (commented):
sorry apparently i forgot to post the review...

please provide proper error messages why the tests are failing

PR #5106, opened 2026-05-07
build(rocjitsu): add rocjitsu to therock build system

amd-arosa:
## Motivation

Integrate rocjitsu (ROCm Just-in-Time Suite) into TheRock's CI build system. rocjitsu is a simulation, virtualization, and instrumentation toolset for AMD GPU applications. It provides ISA-level execution of AMD GPU and RISC-V machine code from HIP fat binaries without requiring GPU hardware. Until now it had only a standalone build and a separate CI workflow in rocm-systems; this PR brings it into the TheRock super-project so it participates in the unified build, artifact distribution, and test infrastructure.

## Technical Details

followed https://github.com/ROCm/TheRock/blob/main/docs/development/build_system.md#adding-sub-projects
also added a flag.

## Test Plan

The CI should run and pass.

## Test Result

Worked on my machine.

## Submission Checklist


README.md
```diff
@@ -208,6 +209,7 @@ minimal build):
 | `-DTHEROCK_ENABLE_SYSDEPS_AMD_MESA=ON` | Enables AMD Mesa for media libs (Linux only)        |
 | `-DTHEROCK_ENABLE_ROCDECODE=ON`        | Enables rocDecode video decoder (Linux only)        |
 | `-DTHEROCK_ENABLE_ROCJPEG=ON`          | Enables rocJPEG JPEG decoder (Linux only)           |
+| `-DTHEROCK_ENABLE_ROCJITSU=ON`         | Enables ROCm emulation tools (Linux only)           |
```
ScottTodd, line 212: We also have a list of projects that are supported on Windows at https://github.com/ROCm/TheRock/blob/main/docs/development/windows_support.md#supported-subprojects. Could you an an entry there too?

I think we could also split 'debug-tools', 'profiler', 'media-libs', and now 'emulator' out into a separate table to group the component groups that _are_ supported on Windows more closely together. (then later we could move debug-tools back into the main group if/when the release at https://rocm.docs.amd.com/projects/install-on-windows/en/latest/how-to/debugger-windows.html gets included in TheRock / open source releases)

emulation/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+      lib
+      tools
+      tests
+  )
+
+  therock_cmake_subproject_activate(rocjitsu)
+
+  therock_cmake_subproject_build_test(rocjitsu
+    COMMAND "${CMAKE_CTEST_COMMAND}" --output-on-failure --no-tests=error
+  )
```
ScottTodd, line 44: Do the tests in https://github.com/ROCm/rocm-systems/tree/develop/emulation/rocjitsu/tests run on CPU only machines, or do they require a GPU? If they require a GPU, tests will need to follow https://github.com/ROCm/TheRock/blob/main/docs/development/adding_tests.md. This `therock_cmake_subproject_build_test` was added recently for LLVM lit tests (unit tests that run in ~seconds on CPU machines), so they can execute during the build on CPU runners while the full build directory is already on disk (and not packaged/uploaded as artifacts).

Note that we aren't yet "building" (running) `therock-build-tests` in CI for all build stages here, see https://github.com/ROCm/TheRock/pull/4881#discussion_r3174362336 (cc @kasaurov, other projects wanting to run tests as part of the build is why I wanted a general solution there and not just `if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime'`)

ScottTodd (commented):
This looks correct/safe, especially with the build defaulting to disabled. Once we add the build stage to https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux.yml it will start getting built and then we can go further with the CI/CD integration in this repository.

I see that you're also planning to opt-in to enabling the new component via `-DTHEROCK_ENABLE_EMULATION=ON` in https://github.com/ROCm/rocm-systems/pull/6044. That's fine (though note that the `therock_cmake_subproject_build_test` code won't run without also having cmake build `therock-build-tests`, and even if it does, that will be on CPU machines that may not run the tests correctly, see my other comment)

README.md
amd-arosa: we don't support windows.
I agree we should split it out to own stage but aren't doing that quite yet.

emulation/CMakeLists.txt
amd-arosa: tests run on cpu.
the test runs each take seconds, disabled the slow tests.

we are in the compiler-runtime stage

ScottTodd: > tests run on cpu.
> the test runs each take seconds, disabled the slow tests.

Ah, good. So even tests like https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/tests/hip_gemm_test.cpp are just running on a simulated GPU? (I see the comments but want to verify that none of the tests are _also_ expecting to compare simulated vs real)

---

The PR I linked was reverted, we actually aren't currently building/running `therock-build-tests` as part of _any_ build stage. I expect it would be simple enough to add though, and @kasaurov was working on a roll forward.

ScottTodd (approved):

geomin12 (commented):
i re-triggered CI, let's wait until build shows passing

amd-arosa:
> i re-triggered CI, let's wait until build shows passing

re-triggered failed job a few more times and it's still failing due to unrelated failures. what is protocol for this?

ScottTodd:
This should have no impact on CI workflows here yet, as the new feature defaults to disabled and it does not yet have a build stage in multi-arch CI here. I'd say it's fine to ignore the failures (which look like AWS runner configuration issues) and merge. If we notice failures after merge (for CI or releases) we can just revert as needed.

Issue #1072, opened 2025-07-17
[CI] Add gating mechanism to ensure PyTorch wheels pass tests before uploading to S3 release bucket


araravik-psd:
- Purpose
We aim to introduce gating checks for the PyTorch build and release process by refactoring the current workflow to ensure that tests are executed and pass successfully before artifacts are uploaded to the release S3 bucket.

- Proposed Approach
The idea is to reuse the existing S3 bucket but omit a versioned or release-specific index during initial upload. This prevents direct exposure of unvalidated artifacts as release assets. Artifacts will only be promoted to the release path if all validation steps succeed.

Refer to the existing workflow:

[build_portable_linux_pytorch_wheels.yml](https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml)

- Current Workflow Design
Job 1: Checks out the PyTorch source repositories, builds the wheels, and uploads artifacts directly to the release S3 bucket.

Job 2: Dynamically determines the test targets to run.

Job 3: Executes PyTorch test suite on appropriate targets.

- Proposed Workflow Refactor
Job 1: Clone the PyTorch source repositories and build the wheels. Instead of uploading directly to the release path, the artifacts are temporarily stored in a staging directory within the same S3 bucket.

Job 2: Dynamically determine the appropriate test targets by running a configuration script.

Job 3: Execute tests against the generated wheels across the identified targets.

Job 4: Upon successful completion of all test jobs, promote the staged artifacts by copying them to the official release directory in the S3 bucket.

ScottTodd:
I'm skeptical of using a subdirectory in the existing "nightly" bucket and then moving. That has the potential to leave us with files in an unstable state and may make it harder to have a predictable way to install releases that are either still being tested or failed the tests.

Here are some alternatives to consider:

## A) Upload to "nightly staging" then copy to "nightly"

Always upload to a separate "nightly staging" bucket. After tests pass, _copy_ to the "nightly" bucket.

There would be no destructive moves but files would be stored twice. We could have separate index pages for each, with the ability to perform uninterrupted bisects across "nightly staging" and we could always copy a release across manually if we want to override the CI results.

## B) Upload to "dev" then copy to "nightly"

Always upload to the same "dev" bucket as test runs from `workflow_dispatch` testing, `pull_request` presubmit runs, and `push` postsubmit runs. After tests pass, copy to the "nightly" bucket.

[...]

PR #5107, opened 2026-05-07
Publish multi-arch PyTorch wheels directly to whl-multi-arch

ScottTodd:
## Motivation

Context:
* https://github.com/ROCm/TheRock/issues/1072

We currently publish nightly rocm packages to `whl-multi-arch` and `whl-staging-multi-arch` but only publish nightly torch packages to `whl-staging-multi-arch`. The plan was to match the behavior of the non-multi-arch releases by running pytorch tests then choosing to promote out of staging based on the results. However, we can't see a reliable way to make this strategy work with the multi-arch packages so the new plan is to always upload directly to `whl-multi-arch` and remove the "staging" index entirely for the new packages.

## Technical Details

See https://github.com/ROCm/TheRock/issues/3332#issuecomment-4361527784 for other options that we considered. Some highlights
* If we partially promote packages (e.g. as test results complete), commands like `pip install torch[device-gfx1100]` will either download the extra as expected or, if the packages are not available, succeed (!) with a warning like `WARNING: torch 2.11.0+rocm7.13.0a20260502 does not provide the extra 'device-gfx1100`. There is no easy way to tell pip to "install the most recent torch package that includes this extra"
* If we want to promote packages until _all_ tests succeed, we don't have test runners for all GPUs and some GPUs can have 12h+ queues
* This does increase the risk of publishing broken packages. We'll address that via better (and more visible) monitoring as well as earlier testing of pytorch during development (https://github.com/ROCm/TheRock/issues/3291)

## Test Plan

Test runs of `.github/workflows/multi_arch_release_linux_pytorch_wheels.yml`:
* https://github.com/ROCm/TheRock/actions/runs/25513057795 (before script rename) published only to https://rocm.devreleases.amd.com/whl-multi-arch/torch/
* https://github.com/ROCm/TheRock/actions/runs/25516361166 (after script rename)

## Submission Checklist


ScottTodd:
Ping?

araravik-psd (approved):
See that the test runs have passed fine both before and after the renames. Also no traces for staging bucket in the readme file. Approving this PR

build_tools/github_actions/tests/publish_pytorch_to_release_bucket_test.py
```diff
@@ -10,7 +10,7 @@
 
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent.parent))
 
-from github_actions.publish_pytorch_to_staging import main
+from github_actions.publish_pytorch_to_release_bucket import main
 
 
 class TestPublishPytorchToStaging(unittest.TestCase):
```
marbre: NIT:
```suggestion
class TestPublishPytorchToReleaseBucket(unittest.TestCase):
```

marbre (approved):
Just one minor NIT.

+ should we manually move `v4/whl-staging/*torch*.whl` to `v4/whl` and remove staging from the index.html files?

ScottTodd:
> should we manually move `v4/whl-staging/*torch*.whl` to `v4/whl` and remove staging from the index.html files?

Yeah, we could move existing files over so they're not orphaned once the docs no longer mention the staging directory. Do you have a recipe for doing that with the s3 bucket handy, or want me to handle that?

RELEASES.md
```diff
@@ -187,37 +207,29 @@ for a full list of supported AMD GPUs.
...
->
-> We also publish **untested** packages to the nightly "whl-staging-multi-arch"
-> index which is not affected by this limitation.
->
-> | Package index                                          | Safe to use `[device-all]`?                              |
-> | ------------------------------------------------------ | -------------------------------------------------------- |
-> | https://rocm.nightlies.amd.com/whl-multi-arch/         | ❌ No (some packages may not be available)               |
-> | https://rocm.nightlies.amd.com/whl-staging-multi-arch/ | ✅ Yes (index includes all packages, even if tests fail) |
-
 <!-- TODO: add repo.amd.com URL to the list of package indexes once we publish a stable release? -->
```
HereThereBeDragons: this we can do now?

ScottTodd: Oh, good catch. I'll remove the TODO for now. I still want us to add more of an overview of the different release channels to the top of the page, where this could go then.

PR #5108, opened 2026-05-07
[rocprofiler-compute] Fix RPATH for rocprofiler-compute test binary in non-default install location

vedithal-amd:
## Motivation

The C++ unit test binary `test-rocprofiler-compute-tool` (registered with CTest via `TEST_FROM_INSTALL=ON`) fails on TheRock builds with:

```
test-rocprofiler-compute-tool: error while loading shared libraries:
librocprofiler-compute-tool.so: cannot open shared object file
```

The library is installed to `lib/rocprofiler-compute/` and the binary to `libexec/rocprofiler-compute/tests/`. The per-target `INSTALL_RPATH` set upstream resolves correctly in standalone CMake builds, but TheRock's global post-subproject hook overwrites it with paths computed relative to the default executable origin (`bin/`), producing an RPATH that does not reach `lib/rocprofiler-compute/`.

This unblocks the upstream rocprofiler-compute PR https://github.com/ROCm/rocm-systems/pull/5721, which adds CTest registration of the native tool test and surfaces the failure on TheRock CI.

## Technical Details

Two changes:

1. `profiler/CMakeLists.txt`: add `INSTALL_RPATH_DIRS "lib/rocprofiler-compute"` to the rocprofiler-compute subproject declaration so the project's own lib subdir is part of the RPATH set TheRock plumbs into every binary in the subproject.

2. `profiler/post_hook_rocprofiler-compute.cmake` (new): annotate `test-rocprofiler-compute-tool` with `THEROCK_INSTALL_RPATH_ORIGIN libexec/rocprofiler-compute/tests` so TheRock computes RPATH relative to the actual install location. Auto-discovered via the `post_hook_<name>.cmake` naming convention. Mirrors the pattern in the adjacent `profiler/post_hook_rocprofiler-sdk.cmake`. Guarded with `if(TARGET ...)` so it is a no-op when `THEROCK_BUILD_TESTING=OFF`.

No upstream changes required. The existing `INSTALL_RPATH` in the rocprofiler-compute project continues to handle standalone builds.

## Test Plan

* TheRock build with `THEROCK_BUILD_TESTING=ON`: confirm `readelf -d` on the installed `test-rocprofiler-compute-tool` shows RUNPATH containing a relative path that resolves to `lib/rocprofiler-compute/` from `libexec/rocprofiler-compute/tests/`.
* CTest runs `test-rocprofiler-compute-tool` end-to-end without setting `LD_LIBRARY_PATH`.
* TheRock build with `THEROCK_BUILD_TESTING=OFF`: build still configures cleanly (post-hook guard prevents touching the missing target).

## Submission Checklist

### DIFF
--- a/profiler/post_hook_rocprofiler-compute.cmake
+++ b/profiler/post_hook_rocprofiler-compute.cmake
@@ -1,0 +1,11 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# test-rocprofiler-compute-tool installs to libexec/rocprofiler-compute/tests/
+# instead of the default bin/. Tell TheRock the actual origin so it can compute
+# correct relative RPATH entries.
+if(TARGET test-rocprofiler-compute-tool)
+  set_target_properties(test-rocprofiler-compute-tool PROPERTIES
+    THEROCK_INSTALL_RPATH_ORIGIN libexec/rocprofiler-compute/tests
+  )
+endif()
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -233,6 +233,8 @@ if(THEROCK_ENABLE_ROCPROFILER_COMPUTE)
       therock_explicit_finders.cmake
     COMPILER_TOOLCHAIN
       amd-hip
+    INSTALL_RPATH_DIRS
+      "lib/rocprofiler-compute"
     RUNTIME_DEPS
       rocprofiler-sdk
       aqlprofile


vedithal-amd:
FYI @svolkov-amd and @feizheng10  for future native collector changes

profiler/post_hook_rocprofiler-compute.cmake
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: MIT
+
+# test-rocprofiler-compute-tool installs to libexec/rocprofiler-compute/tests/
+# instead of the default bin/. Tell TheRock the actual origin so it can compute
+# correct relative RPATH entries.
+if(TARGET test-rocprofiler-compute-tool)
+  set_target_properties(test-rocprofiler-compute-tool PROPERTIES
+    THEROCK_INSTALL_RPATH_ORIGIN libexec/rocprofiler-compute/tests
+  )
+endif()
```
ScottTodd, line 11: FYI I ran an audit across the project for other targets that might need similar treatment: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_5108.md#broader-analysis-other-targets-that-may-be-affected

potentially missing:
* roctracer `codeobj_test` and test executables
* aqlprofile test executables
* rocprofiler (v1), though we don't build it in TheRock and probably don't plan to in the future?

I also checked if we could validate RPATH correctness through unit tests to ensure we don't regress here. See https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_TheRock_5108.md#unit-test-coverage-for-rpath-correctness for a ideas there

ScottTodd (approved):
Thanks, this looks correct based on existing patterns. Are there tests that would run on this PR itself that would confirm the fix, or just the separate PR in rocm-systems that will start using this?

profiler/post_hook_rocprofiler-compute.cmake
ScottTodd: Filed https://github.com/ROCm/TheRock/issues/5114 to follow up so we don't regress and other subprojects getting onboarded can have more confidence that the packaging will work

vedithal-amd:
>Thanks, this looks correct based on existing patterns. Are there tests that would run on this PR itself that would confirm the fix, or just the separate PR in rocm-systems that will start using this?

The test-rocprofiler-compute-tool PR merge into rocm-systems is blocked by this.
So, once we merge this PR - TheRock CI on that PR will tell whether the fix worked.

vedithal-amd:
Its allowing me to merge even when things are red, can i go ahead ?

The rocprofiler-compute test error is due to flaky test_L1_cache_counter metric validation test which i just disabled in develop to fix it later, so that is not related to this PR

image

ScottTodd:
If the builds and tests you expect to pass do pass then yes, safe to merge.

PR #5109, opened 2026-05-07
Generalize promotion script and add multi-arch target filter

HereThereBeDragons:
Renames `promote_from_rc_to_final.py` -> `promote_packages.py` to reflect its broader scope. 

The script no longer assumes RC → release. It now handles the common transitions: `rc → release` (the existing default), `a → release`, `a → rc<N>`.

Changes of the CLI args:
- `--src-version-type {rc,a}` — replaces `--prerelease-type` (single-value; `dev` dropped). Defaults to `rc`.
- `--dest-version` — `release` | `rc<N>` | `a<YYYYMMDD>`. Defaults to `release`.
- `--multi-arch-targets` — **multi-arch packages only.** Positive list of gfx targets to retain. When unset (default), no arch filtering happens; single-arch packages and tarballs ignore this flag entirely.
- `--skip-version-promotion` — run only the multi-arch arch-trim pass without touching versions. Mutex with the version flags; requires `--multi-arch-targets`.

`--multi-arch-targets`: Applies both to aggregator wheels (rocm, torch, torchvision, ...) and per-gfx wheels:
  - Per-gfx wheels for non-kept archs are skipped (and deleted with --delete-old-on-success).
  - METADATA / PKG-INFO loses `Provides-Extra: device-gfx<N>` and matching `Requires-Dist: ...-gfx<N>` lines for non-kept archs.
  - requires.txt loses `[device-gfx<N>]` sections and `[device-all]` body deps for non-kept archs.
  - _dist_info.py loses `AVAILABLE_TARGET_FAMILIES.append('gfx<N>')` for non-kept archs.
  - When the package's *default* arch is dropped, three sites are repointed to the best fitting requested arch (so the [device] extra still resolves, see `_repoint_priority()`): `DEFAULT_TARGET_FAMILY` in _dist_info.py, the bare `extra == "device"` line in METADATA, and the `[device]` body in requires.txt.
  - Single-arch packages are auto-detected and pass through unchanged.
  - Empty intersection of requested targets and available targets raises ValueError.

Other changes:
  - New `tests/promote_keep_list_test.py`: unit tests for the keep-list helpers, version-string transitions, and argparse mutex/validation.
  - `tests/promote_from_rc_to_final_test.py` -> `promote_packages_test.py`.
  - third_party/change_wheel_version: skip dir / dist-info / data renames when old_slug == new_slug, so `--skip-version-promotion` can repack a wheel without touching its version. Same-version guard tightened to use the resolved `new_version`.
  - upload_release_packages.py docstring updated to point at the new script name.
  - how_to_do_release.md: new subsections for nightly (`a`) promotion (a -> release, a -> rc) and the multi-arch `--multi-arch-targets` / `--skip-version-promotion` workflows.


HereThereBeDragons:
for whatever reason git does not recognize the renaming of `promote_from_rc_to_final.py` -> `promote_packages.py`.
you might therefore want to checkout `main` branch and my `promote_packages.py` and just copy it into `promote_from_rc_to_final.py` to just see the partial changes.

build_tools/packaging/promote_packages.py
marbre, line 1: The scripts grows from ~ 400 LOC to more than 1000 lines. Do you have any thoughts already on how to improve the structure? Seems GH not even realizes this is a re-write and doesn't offer a diff.

build_tools/packaging/promote_packages.py
```diff
@@ -0,0 +1,1090 @@
...
+        help=(
+            "Comma-separated positive list of gfx targets to retain "
+            "(e.g. 'gfx1201,gfx1010,gfx11'). Everything not in this list is dropped: "
+            "multi-arch aggregator wheels (rocm, torch, torchvision) lose matching "
+            "Provides-Extra/Requires-Dist entries, multi-arch _dist_info.py loses "
+            "matching AVAILABLE_TARGET_FAMILIES entries (and DEFAULT_TARGET_FAMILY "
+            "is repointed at the first kept arch if it referenced a dropped one), "
+            "and per-gfx wheels for non-kept archs are skipped (or deleted with "
+            "--delete-old-on-success). Single-arch packages are detected and pass "
+            "through untouched."
```
marbre: This is pretty long description. I was about to ask about single-family packages until I found the last sentence providing the information I was looking for. Make rename to make this more clear from the start, e.g. what about the following:

* `--multi-arch-gfx`
* `--mutli-arch-gfx-list`
* `--multi-arch-devices`
* `--multi-arch-targets`

Could then simply state that every _mutli-arch_ `_device_` package not in the list is dropped from extras.

build_tools/packaging/promote_packages.py
```diff
@@ -0,0 +1,1090 @@
...
+        ),
+        default=None,
+    )
+    parser.add_argument(
+        "--skip-version-promotion",
```
marbre, line 172: Why is this needed?

build_tools/packaging/how_to_do_release.md
```diff
@@ -34,27 +34,99 @@ python build_tools/packaging/download_prerelease_packages.py --version=7.10.0rc2
...
+
+### Multi-arch packages: restricting which gfx targets ship
+
+Multi-arch aggregator wheels (`rocm`, `torch`, `torchvision`, …) reference
+several gfx targets via `Provides-Extra` / `Requires-Dist` entries, and the
+download directory may contain per-gfx wheels (`rocm_sdk_device_gfx1010-…`,
+`amd_torch_device_gfx1010-…`) for each of those targets.
+
+If a release should only ship a subset of those archs, pass
+`--keep-gfx-archs` (positive list — everything else is dropped):
```
marbre: This reads like nothing is dropped if the list isn't provided, in the arg description it read like this flag is mandatory which would imply everything is dropped if the arg is not passed.

build_tools/packaging/how_to_do_release.md
```diff
@@ -34,27 +34,99 @@ python build_tools/packaging/download_prerelease_packages.py --version=7.10.0rc2
...
+   --src-version-type=a \
+   --dest-version=rc1 \
+   --delete-old-on-success
+```
+
+`--dest-version` accepts `release`, `rc<N>` (e.g. `rc1`, `rc2`), or
+`a<YYYYMMDD>` (e.g. `a20260501`). The downstream RC -> release flow above is
+unchanged.
+
+### Multi-arch packages: restricting which gfx targets ship
```
marbre:
```suggestion
### Multi-arch packages: restricting which gfx targets to ship
```

marbre (commented):
Not a full review but some initial comments. Needs a more careful look.

build_tools/packaging/promote_packages.py
HereThereBeDragons: claude suggested rewriting the history: first just renaming in a single commit and then applying the changes in a new commit. i wil ltry this but will result in a force push.

most new stuff is the code to rewrite which archs to keep for the different types of file.

build_tools/packaging/promote_packages.py
HereThereBeDragons: good idea thanks.

build_tools/packaging/promote_packages.py
HereThereBeDragons: because you might want to use `--keep-gfx-archs` without changing the version.

build_tools/packaging/how_to_do_release.md
HereThereBeDragons: will clarify it. if `--keep-gfx-archs` is not given, the archs are not touched and everything is kept as-is. only the version of the package is changing

build_tools/packaging/promote_packages.py
HereThereBeDragons: the splitting sadly did not really help.

i renamed it now for the meantime back to `promote_from_rc_to_final.py`. until the review is over and then back to its final name `promote_packages.py`

HereThereBeDragons:
note: `promote_from_rc_to_final.py` will be renamed after the final review to `promote_packages.py` 
keeping it on the old name for easier reviewing

HereThereBeDragons:
## Summary: `promote_from_rc_to_final.py` → `promote_packages.py`

Here an overview of what all changed. Especially a table with what functions were added (majority of code addition due to new  `--multi-arch-targets`.)

### What's new

The script no longer assumes RC → release. It now handles three orthogonal axes:

- `--src-version-type {rc,a}` — replaces `--prerelease-type` (single-value; `dev` dropped). Defaults to `rc`.
- `--dest-version` — `release` | `rc<N>` | `a<YYYYMMDD>`. Defaults to `release`.
- `--multi-arch-targets` — **multi-arch packages only.** Positive list of gfx targets to retain. When unset (default), no arch filtering happens; single-arch packages and tarballs ignore this flag entirely.
- `--skip-version-promotion` — run only the multi-arch arch-trim pass without touching versions. Mutex with the version flags; requires `--multi-arch-targets`.

Common transitions covered: `rc → release` (the existing default), `a → release`, `a → rc<N>`.

### Multi-arch keep-list pass (only fires on multi-arch wheels)

For aggregator wheels (`rocm`, `torch`, `torchvision`, …) and their per-gfx companion wheels (`rocm_sdk_device_gfx<N>`, `amd_torch_device_gfx<N>`):

- Per-gfx companion wheels for non-kept archs are skipped (deleted with `--delete-old-on-success`).
- METADATA / `requires.txt` / `_dist_info.py` lose `Provides-Extra` / `Requires-Dist` / `[device-gfx<N>]` / `AVAILABLE_TARGET_FAMILIES` entries for non-kept archs.
- If the package's *default* arch is dropped, three sites are repointed at the highest-priority kept arch (specific `gfx<NNN+>` preferred over family prefix `gfx<NN>`): `DEFAULT_TARGET_FAMILY`, the bare `extra == "device"` line, and the `[device]` body.
- Single-arch wheels and tarballs are auto-detected and pass through unchanged.

### Functions in `promote_packages.py`

**New (multi-arch keep-list machinery):**

| Function | Purpose |
|---|---|
| `compute_new_version_str(version_str, src_version_type, dest_version)` | Pure version-string transition (extracted from inline logic in `promote_wheel`). |
| `_scan_multiarch_metadata(path)` | Detect multi-arch shape in METADATA / PKG-INFO; enumerate present archs. |
| `_scan_multiarch_requires_txt(path)` | Same, for `requires.txt`. |
| `_scan_multiarch_dist_info_py(path)` | Same, for `_dist_info.py`. |
| `_apply_keep_arch_list_to_metadata(path, keep)` | Drop / repoint METADATA entries. |
| `_apply_keep_list_to_requires_txt(path, keep)` | Drop / repoint `[device-gfx<N>]` sections and `[device]` body. |
| `_apply_keep_list_to_dist_info_py(path, keep)` | Drop / repoint `_dist_info.py` entries. |
| `_assert_keep_overlaps_found(label, found, keep)` | Guard; makse sure that in the package we have at least on arch we want to keep. |
| `_repoint_priority(keep_archs)` | Order keep list for default-arch repointing — specific archs first, family prefixes second; user order preserved within each tier. Needed when the default family is being dropped. Choose the next eligble arch from the keep list |
| `wheel_apply_gfx_keep_list(extracted_wheel_dir, keep)` | Orchestrates the three per-file passes against an unpacked wheel; no-op on single-arch. |

**Modified:**

| Function | Change |
|---|---|
| `parse_arguments` | Typed signature; new flags above; sentinel `None` defaults so the mutex check distinguishes "user passed `--dest-version=release`" from "unset"; calendar validation for `a<YYYYMMDD>`. |
| `wheel_change_extra_files` | Gains `multi_arch_targets` kwarg; runs the keep-list pass on the unpacked wheel before re-packing. No-op on single-arch. |
| `promote_wheel` | Version logic moved out to `compute_new_version_str`; gains `multi_arch_targets` kwarg; handles `--skip-version-promotion` (keep-list pass only, no version rewrite). |
| `promote_targz_sdist` | Same shape as `promote_wheel`. |
| `main` | Accepts `dest_version`, `multi_arch_targets`, `skip_version_promotion`; threads them through. |

**Unchanged (carried over):** `update_metadata_rocm_requires_dist`, `promote_targz_tarball`.

### Companion change

- `build_tools/third_party/change_wheel_version/change_wheel_version.py`: skips dir / dist-info / data renames when `old_slug == new_slug`, so `--skip-version-promotion` can repack without touching the version. Same-version guard tightened to use the resolved `new_version`.

### Tests

- `tests/promote_keep_list_test.py` (new, ~430 lines, 27 tests): every keep-list helper, single-arch passthrough, version-string transitions, `_repoint_priority` ordering, argparse mutex/validation.
- `tests/promote_packages_test.py` (renamed from `promote_from_rc_to_final_test.py`; minor adjustments — git did detect this rename at R096).

### Docs

- `how_to_do_release.md`: new "Promoting nightly (`a`) builds" subsection; new "Multi-arch packages: restricting which gfx targets to ship" subsection covering both `--multi-arch-targets` and `--skip-version-promotion`; explicit default-behavior callout.

### Backwards compatibility

Defaults reproduce the old behaviour exactly. CI invocations passing only `--input-dir` and `--delete-old-on-success` continue to work — including single-arch and tarball pipelines, since `--multi-arch-targets` is multi-arch-only and gated on auto-detection. Only breaking surface: renamed flag (`--prerelease-type` → `--src-version-type`) and dropped `dev` choice.

build_tools/packaging/promote_packages.py
marbre: Well, in that case the script could just not change the version if you provide `--keep-gf-archs` but no version?

HereThereBeDragons: as the default is to do promotion from rc/a -> release, if we just want to strict the archs we need a flag to explicitly say that we do not want to do any promotion

(we can also change the default, but i think the default right now is still more common)

araravik-psd:
This script was used to promote multi arch rc candidates for 7.13 and the multi arch packages were promoted fine without any failures. Will approve this PR

HereThereBeDragons:
@araravik-psd  please review :) we probably need to add stuff to it soon so it would be good to have this version merged. and i also want to finally rename it to `promote_packages.py`

build_tools/packaging/promote_packages.py
```diff
@@ -115,20 +245,534 @@ def update_metadata_rocm_requires_dist(
...
+    The source prerelease segment (`src_version_type` plus any trailing digits) is
+    either stripped or replaced:
+      - `dest_version == "release"`: strip it
+        (`7.13.0a20260430` -> `7.13.0`, `2.10.0+rocm7.13.0a20260430` -> `2.10.0+rocm7.13.0`).
+      - otherwise: replace it with `dest_version`
+        (`7.13.0a20260430` -> `7.13.0rc1` when called with src='a', dest='rc1').
+    """
+    # src_version_type is constrained to "rc" or "a" by argparse; regex-safe.
+    replacement = "" if dest_version == "release" else dest_version
+    return re.sub(rf"{src_version_type}\d*", replacement, version_str)
```
araravik-psd, line 280: If version_str has no prerelease segment matching src_version_type, this silently returns the version unchanged. the user won't know their input directory contained the wrong version type. There should be a check that at least one substitution occurred, otherwise you silently produce the same-versioned file alongside the original.

build_tools/packaging/promote_packages.py
```diff
@@ -144,51 +788,61 @@ def wheel_change_extra_files(new_dir_path: pathlib.Path, old_version, new_versio
...
+    # rocm-* package to the build's rocm version. Rewrite those, then return.
+    if "device_gfx" in new_dir_path.name:
+        return
+    if "jax_rocm7_plugin" in package_name_no_version or "jax_rocm7_pjrt" in package_name_no_version:
+        return
+
+    # rocm packages needing extra handling
     if new_dir_path.name.startswith("rocm"):
+        if "rocm_sdk_libraries" in new_dir_path.name and "gfx" not in new_dir_path.name:
+            package_name_no_version = "rocm_sdk_libraries_None"
```
araravik-psd, line 829: This looks like a workaround for a path construction issue, and the string "None" in a package name? Can we add a comment explaining this

build_tools/packaging/promote_packages.py
```diff
@@ -2,50 +2,85 @@
...
+  python ./build_tools/packaging/promote_packages.py --input-dir=./release_candidates/rc1/ --delete-old-on-success
 
   # Promote only specific files matching a pattern:
-  python ./build_tools/packaging/promote_from_rc_to_final.py --input-dir=./release_candidates/ --match-files='*rc2*'
+  python ./build_tools/packaging/promote_packages.py --input-dir=./release_candidates/ --match-files='*rc2*'
+
+  # Promote some nightly to release (e.g. 7.13.0a20260430 -> 7.13.0):
+  python ./build_tools/packaging/promote_packages.py --input-dir=./release_candidates/ --src-version-type=a
+
+  # Promote alpha to release candidate (e.g. 7.13.0a20260430 -> 7.13.0rc1)
```
araravik-psd, line 59: The new compute_new_version_str function handles a -> rc, do we have a test for this?

build_tools/packaging/promote_packages.py
araravik-psd: Acknowledged, reviewing under promote_from_rc_to_final.py for now. Please do the rename to promote_packages.py as a clean separate commit before or immediately after approval so the merge commit has the final name

build_tools/packaging/promote_packages.py
```diff
@@ -115,20 +245,534 @@ def update_metadata_rocm_requires_dist(
...
+    once. Returns the set of all appended archs (regardless of count).
+    If it only occurs once it is a single-arch package.
+    """
+    text = path.read_text(encoding="utf-8")
+    append_re = re.compile(
+        rf"^\s*AVAILABLE_TARGET_FAMILIES\.append\(['\"]({_GFX_ARCH})['\"]\)\s*$",
+        re.MULTILINE,
+    )
+    archs = [m.group(1) for m in append_re.finditer(text)]
+    if len(archs) <= 1:
```
araravik-psd, line 642: A single-arch package legitimately has exactly one AVAILABLE_TARGET_FAMILIES.append(...) line. But this means a package that happens to have two appends for the same arch (a bug in the input) would be classified as multi-arch. The check should probably be len(set(archs)) <= 1 (to filter out of duplicates of the same arch getting flagged as multi arch)?

araravik-psd (commented):
HI @HereThereBeDragons please take a look at the review comments

build_tools/packaging/promote_packages.py
HereThereBeDragons: this was fixed in https://github.com/ROCm/TheRock/pull/5984
@marbre do we need to support those old multiarch packages or just remove this workaround?

HereThereBeDragons: (i also double checked in the latest nightly this is fixed in the wheel)

build_tools/packaging/promote_packages.py
HereThereBeDragons: this is just a helper function.
the guard is done here:
https://github.com/ROCm/TheRock/blob/a20c4ccb7b503f7740a9bcb9c9e7d51a4baa36cf/build_tools/packaging/promote_from_rc_to_final.py#L904-L924


>  otherwise you silently produce the same-versioned file alongside the original.

this would result in an overwrite of the package after applying everything else

build_tools/packaging/promote_packages.py
HereThereBeDragons: i would argue we are using the exact same logic for it as for the other promotion and it is provided by an external library.

```
new_wheel_path = change_wheel_version.change_wheel_version(
        filename,
        new_base_version,
        new_local_version,
        callback_func=callback,
    )
````

that being said: the entire test stuff needs to be reworked - as it is still using old 7.9 single arch packages. 

for the sake of time i would rather put this in a new pr that then also includes testing after promotion of live packages. (like the e2e testing we had talked about before)

File: build_tools/packaging/promote_packages.py (hunk 6 above)
Comment (marbre) [reply build_tools/packaging/promote_packages.py:829]: Two options. Land as is and delete and refactor as a follow up, in case we need to promote a patch release for e.g. 7.13. No need to support both.

Comment (HereThereBeDragons) [reply build_tools/packaging/promote_packages.py:829]: created an issue for it https://github.com/ROCm/TheRock/issues/6266

File: build_tools/packaging/promote_packages.py (hunk 7 above)
Comment (HereThereBeDragons) [reply build_tools/packaging/promote_packages.py:59]: https://github.com/ROCm/TheRock/issues/6266 issue to track

**Review by araravik-psd — APPROVED** [review]
Approving! One minor note before merging: the _None workaround at line 822–823 (rocm_sdk_libraries_None) would benefit from a brief inline comment referencing issue #6266, just so it's not a head-scratcher for anyone reading the code later. Everything else looks good.

File: build_tools/packaging/promote_packages.py [hunk 9]
```diff
@@ -0,0 +1,1176 @@
...
+    base_dir = filename.parent
+    package_name = filename.name.removesuffix(".tar.gz")  # removes .tar.gz
+
+    with tempfile.TemporaryDirectory(prefix=package_name + "-") as tmp_dir:
+        print(f"  Extracting tar file to {tmp_dir}", end="")
+
+        tmp_path = pathlib.Path(tmp_dir)
+
+        targz = tarfile.open(filename)
+        targz.extractall(tmp_path)
```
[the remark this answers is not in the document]
HereThereBeDragons: i will merge as-is as i didnt touch that part of the code. this is a follow up. most likely we want `targz.extractall(tmp_path, filter="tar")` but needs testing that it works.

will write it down in #6266

PR #5122, opened 2026-05-08
Fix jax dispatch when passing a ref for TheRock checkout

HereThereBeDragons:
For jax dispatch the ref must be past as a inputs parameter, not as part of `with:`


marbre (approved):
Thanks!

Can you manually trigger the JAX build in rockrel?

PR #5123, opened 2026-05-08
feat(rocgdb): add upstream build and download CI script flags

lumachad:
Adds `THEROCK_ROCGDB_UPSTREAM_BUILD` and `THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT` flags (both default `OFF`) to support building rocgdb from upstream GDB master. When enabled, the build skips AMD-specific install artifacts (`NOTICES.txt`, `roccoremerge`) and optionally downloads `test_rocgdb.py` from `ROCm/ROCgdb` amd-staging when absent from the source tree.

Both flags are declared via `option()` so they get proper `BOOL` cache entries and help strings visible in `ccmake`/`cmake-gui`/`cmake -LH`.

The `test_rocgdb.py` download uses `execute_process(curl --fail ...)` rather than `file(DOWNLOAD)` so HTTP 4xx/5xx errors actually fail the retry loop instead of silently writing an error page to disk.

Depends on:

- [x] https://github.com/ROCm/TheRock/pull/4960
- [x] https://github.com/ROCm/TheRock/pull/5720
- [x] https://github.com/ROCm/ROCgdb/pull/117

Related: #4960

### DIFF
--- a/debug-tools/CMakeLists.txt
+++ b/debug-tools/CMakeLists.txt
@@ -281,18 +281,18 @@ if(THEROCK_ENABLE_ROCGDB)
   #   - roccoremerge man page
   # set: -DTHEROCK_ROCGDB_UPSTREAM_BUILD=ON
   # Note: upstream master will eventually include roccoremerge.
-  if(NOT DEFINED THEROCK_ROCGDB_UPSTREAM_BUILD)
-    set(THEROCK_ROCGDB_UPSTREAM_BUILD OFF)
-  endif()
+  option(THEROCK_ROCGDB_UPSTREAM_BUILD
+    "Build rocgdb against upstream GDB master (skips AMD-specific files like NOTICES.txt, roccoremerge)"
+    OFF)
 
   # When ON, download the test_rocgdb.py CI wrapper script from ROCm/ROCgdb
   # amd-staging if it is not already present in the rocgdb source tree, and
   # install it into tests/rocgdb. Required by TheRock CI when building a
   # rocgdb source tree that is not part of the branches exercised by CI.
   # Off by default to avoid a network fetch at configure time.
-  if(NOT DEFINED THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT)
-    set(THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT OFF)
-  endif()
+  option(THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT
+    "Download test_rocgdb.py from ROCm/ROCgdb amd-staging if missing in rocgdb source tree"
+    OFF)
 
   therock_cmake_subproject_declare(rocgdb
     USE_DIST_AMDGPU_TARGETS
--- a/debug-tools/CMakeLists.txt
+++ b/debug-tools/CMakeLists.txt
@@ -274,6 +274,26 @@ if(THEROCK_ENABLE_ROCGDB)
     message(WARNING "Configuring ROCgdb with modified source directory: ${THEROCK_ROCGDB_SOURCE_DIR}")
   endif()
 
+  # Default to building ROCm's rocgdb branch (with NOTICES.txt, roccoremerge, etc.).
+  # For upstream master builds, which currently lack AMD-specific files:
+  #   - NOTICES.txt file
+  #   - roccoremerge program
+  #   - roccoremerge man page
+  # set: -DTHEROCK_ROCGDB_UPSTREAM_BUILD=ON
+  # Note: upstream master will eventually include roccoremerge.
+  if(NOT DEFINED THEROCK_ROCGDB_UPSTREAM_BUILD)
+    set(THEROCK_ROCGDB_UPSTREAM_BUILD OFF)
+  endif()
+
+  # When ON, download the test_rocgdb.py CI wrapper script from ROCm/ROCgdb
+  # amd-staging if it is not already present in the rocgdb source tree, and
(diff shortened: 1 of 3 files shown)


debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -705,6 +736,50 @@ if(BUILD_TESTING)
...
+        message(STATUS "Attempting to download test_rocgdb.py (attempt ${attempt}/3)")
+        file(DOWNLOAD
+          "${TEST_ROCGDB_URL}"
+          "${TEST_ROCGDB_SCRIPT}"
+          STATUS download_status
+          TIMEOUT 60
+        )
+        list(GET download_status 0 status_code)
+        if(status_code EQUAL 0)
+          set(download_success TRUE)
```
marbre: From my review agent:
> CMake's `file(DOWNLOAD)` status code is the libcurl transfer code and not the HTTP status. As  404/500 from raw.githubusercontent.com still returns `status_code == 0` and writes the error-page body to disk. The build would then install a bogus `test_rocgdb.py` containing 404: Not Found and CI would fail later with a confusing error. The retry loop and FATAL_ERROR guard would never trigger.

The agent proposed to introduce a `EXPECTED_HASH_SHA256` but I think this isn't desired here. Anyway, we need another solution here.

debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -705,6 +736,50 @@ if(BUILD_TESTING)
...
+  if(NOT EXISTS "${TEST_ROCGDB_SCRIPT}")
+    if(ROCGDB_DOWNLOAD_CI_SCRIPT)
+      set(TEST_ROCGDB_SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/test_rocgdb.py")
+      # The source of truth for the CI script is the one in amd-staging.
+      set(TEST_ROCGDB_URL "https://raw.githubusercontent.com/ROCm/ROCgdb/amd-staging/.github/scripts/test_rocgdb.py")
+
+      set(download_success FALSE)
+      foreach(attempt RANGE 1 3)
+        message(STATUS "Attempting to download test_rocgdb.py (attempt ${attempt}/3)")
+        file(DOWNLOAD
```
marbre: I think we should set `TLS_VERIFY ON` as we are downloading an executable and we don't hash.

debug-tools/CMakeLists.txt
```diff
@@ -268,6 +268,26 @@ if(THEROCK_ENABLE_ROCGDB)
...
+  endif()
+
+  # When ON, download the test_rocgdb.py CI wrapper script from ROCm/ROCgdb
+  # amd-staging if it is not already present in the rocgdb source tree, and
+  # install it into tests/rocgdb. Required by TheRock CI when building a
+  # rocgdb source tree that is not part of the branches exercised by CI.
+  # Off by default to avoid a network fetch at configure time.
+  if(NOT DEFINED THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT)
+    set(THEROCK_ROCGDB_DOWNLOAD_CI_SCRIPT OFF)
+  endif()
```
marbre: Why not using options here as well?

lumachad: No reason. Just slipped by. Converted to using option now.

debug-tools/rocgdb/CMakeLists.txt
lumachad: No. The script may change and we don't have a stable hash then. Switched the download to curl --fail, which verifies TLS by default — so the explicit TLS_VERIFY is no longer needed on this path.

debug-tools/rocgdb/CMakeLists.txt
lumachad: Switched to curl directly, as it covers a false success on 404/500 replies.

lumachad:
Lift ci:skip label.

debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -706,10 +737,53 @@ if(BUILD_TESTING)
...
+    if(ROCGDB_DOWNLOAD_CI_SCRIPT)
+      set(TEST_ROCGDB_SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/test_rocgdb.py")
+      # The source of truth for the CI script is the one in amd-staging.
+      set(TEST_ROCGDB_URL "https://raw.githubusercontent.com/ROCm/ROCgdb/amd-staging/.github/scripts/test_rocgdb.py")
+
+      # Use curl rather than file(DOWNLOAD): file(DOWNLOAD)'s STATUS is the
+      # libcurl transfer code, so HTTP 404/500 responses are reported as
+      # success and the error-page body is written to disk. curl --fail
+      # returns non-zero on HTTP >= 400, and verifies TLS by default.
+      find_program(CURL_EXECUTABLE curl REQUIRED)
```
marbre, line 754: `curl` is of course pretty standard but I am not sure we should add it at least to the docs somewhere as one might no have it for reasons?

marbre (approved):

debug-tools/rocgdb/CMakeLists.txt
lumachad: The general docs or the debug-tools-specific docs?

lumachad: Documented in the debug-tools-specific docs.

lumachad:
All dependencies now merged. I'll give this another go in the CI and will merge it if good.

Issue #4987, opened 2026-05-01
[Multi-arch] Quality of life improvements for multi-arch release pipelines


ScottTodd:
Collecting a few follow-up tasks found during development

- [ ] https://github.com/ROCm/TheRock/issues/1236
- [x] Reorder and trim run names in https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_linux_pytorch_wheels.yml since the targets list is long so move it to the back, start of the name overlaps with the workflow name "Multi-Arch Release Linux PyTorch Wheels"
    * from e.g. `Multi-Arch Release Linux PyTorch Wheels (gfx94X-dcgpu;gfx110X-all;gfx1151;gfx120X-all;gfx950-dcgpu;gfx900;gfx906;gfx908;gfx90a;gfx101X-dgpu;gfx103X-all;gfx1150;gfx1152;gfx1153, nightly, 7.13.0a20260501)`
    * to more like `Release PyTorch (nightly, 7.13.0a20260501, gfx94X-dcgpu;gfx110X-all;gfx1151;gfx120X-all;gfx950-dcgpu;gfx900;gfx906;gfx908;gfx90a;gfx101X-dgpu;gfx103X-all;gfx1150;gfx1152;gfx1153)`
- [ ] Add summary output to jobs like https://github.com/ROCm/rockrel/actions/runs/25207747221
    * Show versions and link to commit hashes used
    * Link to the release index pages the jobs will upload to
    * Document where pytorch manifest files are stored
    * Similar to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/summarize_test_pytorch_workflow.py for the reports on e.g. https://github.com/ROCm/TheRock/actions/runs/25150366706
- [ ] Upload pytorch build logs to AWS somewhere in https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md
- [x] When building pytorch don't commit the hipify changes, so the embedded git version in the packages matches the release branch commits
- [x] Dispatch test jobs instead of using workflow_call, so we can have separate status badges for release _builds_ and release _tests_ and release jobs don't remain queued for 15h+ (e.g. https://github.com/ROCm/rockrel/actions/runs/25200531110)
- [ ] Add the package versions to the setup output in https://github.com/ROCm/TheRock/blob/main/.github/workflows/setup_multi_arch.yml / https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci_summary.py
[...]

PR #5127, opened 2026-05-08
Add placeholder multi-arch JAX release workflow

erman-gurses:
Placeholder for `multi_arch_release_linux_jax_wheels.yml` for https://github.com/ROCm/TheRock/pull/4366


.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,129 @@
...
+run-name: Multi-Arch Release Linux JAX Wheels (${{ inputs.amdgpu_family }}, ${{ inputs.release_type }}, ${{ inputs.rocm_version }})
+
+jobs:
+  build_jax_wheels:
+    name: Build Linux JAX Wheels (${{ inputs.amdgpu_family }}, py${{ matrix.python_version }}, ${{ inputs.release_type }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python_version: ["3.11", "3.12"]
```
marbre, line 66: I think this needs further Python version.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,129 @@
...
+
+name: Multi-Arch Release Linux JAX Wheels
+
+on:
+  workflow_call:
+    inputs:
+      amdgpu_family:
+        type: string
+        required: true
+
```
marbre: You can drop those empty lines, look at the other PRs.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,129 @@
...
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
+        with:
+          python-version: ${{ matrix.python_version }}
+
+      - name: Build JAX wheels
+        env:
+          ROCM_VERSION: ${{ inputs.rocm_version }}
+        run: |
+          pushd rocm-jax
+
```
marbre: I think you could either set the workdir for this step or just `cd` into the directory as you don't call any other command after `popd`.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
marbre, line 1: The workflow misses to upload :(

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,129 @@
...
+        type: string
+        required: true
+
+      rocm_version:
+        type: string
+        required: true
+
+      tar_url:
+        type: string
+        required: true
```
marbre, line 21: Has there been any effort so far to switch over to Python packages as a base to build jax upon? @charleshofer can you maybe comment on this?

marbre (changes requested):

erman-gurses:
@marbre / @ScottTodd, this is just placeholder for this PR: https://github.com/ROCm/TheRock/pull/4366. Otherwise, I cannot do my tests.

marbre:
> @marbre / @ScottTodd, this is just placeholder for this PR: #4366. Otherwise, I cannot do my tests.

This makes more sense now with the updated PR title, thanks. I assume it is not triggered automatically anywhere yet? If it is not we can land as is but if it is, it should be closer to a no-op (and skip the build). Please confirm.

erman-gurses:
> > @marbre / @ScottTodd, this is just placeholder for this PR: #4366. Otherwise, I cannot do my tests.
> 
> This makes more sense now with the updated PR title, thanks. I assume it is not triggered automatically anywhere yet? If it is not we can land as is but if it is, it should be closer to a no-op (and skip the build). Please confirm.

It did not trigger any builds/test. But I added just in case `ci:skip`

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,118 @@
...
+          - gfx94X-dcgpu
+          - gfx950-dcgpu
+        default: gfx94X-dcgpu
+      release_type:
+        type: choice
+        options:
+          - dev
+          - nightly
+          - prerelease
+        default: dev
```
ScottTodd, line 39: Let's start with only `dev` in `workflow_dispatch` here (TheRock), see https://github.com/ROCm/TheRock/blob/b7abaa48d2390ad1dee8a342bf7c44d8e2d48cb3/.github/workflows/multi_arch_release.yml#L48-L53

If we add a wrapper workflow in https://github.com/ROCm/rockrel we can give more choices there, like in https://github.com/ROCm/rockrel/blob/45223d0d4709b5ec27afa0a0efb53faf2d3502a3/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml#L20-L27

```suggestion
      release_type:
        type: choice
        options:
          - dev
        default: dev
```

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,118 @@
...
+          - gfx1152
+          - gfx1153
+          - gfx120X-all
+          - gfx900
+          - gfx906
+          - gfx908
+          - gfx90a
+          - gfx94X-dcgpu
+          - gfx950-dcgpu
+        default: gfx94X-dcgpu
```
ScottTodd: Let's copy this pattern for gpu family selection:
https://github.com/ROCm/TheRock/blob/b7abaa48d2390ad1dee8a342bf7c44d8e2d48cb3/.github/workflows/multi_arch_release.yml#L58-L67

The point of "multi arch release" is that we build for multiple GPU targets/families at once, so the inputs here should allow for that (i.e. use a list of targets/families and not a fixed single "choice")

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -0,0 +1,118 @@
...
+      ref:
+        type: string
+        default: ""
+
+permissions:
+  contents: read
+  id-token: write
+  packages: read
+
+run-name: Multi-Arch Release Linux JAX Wheels (${{ inputs.amdgpu_family }}, ${{ inputs.release_type }}, ${{ inputs.rocm_version }})
```
ScottTodd: Let's keep an eye on the `run-name` here. See https://github.com/ROCm/TheRock/pull/5156 which proposes a reordering.

When viewed in history like https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_linux_pytorch_wheels.yml some of the more useful information about the run is pushed back to the end while the start is repeated 

Image

From that PR it would be:
> Before: Multi-Arch Release Linux PyTorch Wheels (gfx94X-dcgpu;..., nightly, 7.x.x)
> After: Release Linux PyTorch Wheels (nightly, 7.x.x, gfx94X-dcgpu;...)

```suggestion
run-name: Multi-Arch Release Linux JAX Wheels (${{ inputs.release_type }}, ${{ inputs.amdgpu_family }}, ${{ inputs.rocm_version }})
```

From https://github.com/ROCm/TheRock/issues/4987, I think we should also move `inputs.rocm_version` to the front while we're at it too.

ScottTodd: I just merged that PR. Let's go with this:

```suggestion
run-name: Release Linux JAX Wheels (${{ inputs.release_type }}, ${{ inputs.rocm_version }}, ${{ inputs.amdgpu_family }})
```

(swap `amdgpu_family` for `amdgpu_families` when making the other changes requested)

```suggestion
run-name: Release Linux JAX Wheels (${{ inputs.release_type }}, ${{ inputs.rocm_version }}, ${{ inputs.amdgpu_families }})
```

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Updated.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Updated

ScottTodd (approved):

erman-gurses:
@marbre, Let me know if we can land this as is. If so, can you please remove the blocker.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
charleshofer: Yes. There has been some. We've got a Docker image that we've been building against. We're kind of cheating by creating a symlink to ROCm path that we get out of `rocm-sdk`: https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/Dockerfile.base-therock-ubu24#L54

If you do a similar trick with the symlink, we can build against the wheels.

marbre: @charleshofer thanks for the reply. @erman-gurses maybe consider that for multi-arch builds :)

marbre (approved):

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Sounds good @marbre.

PR #5128, opened 2026-05-08
[CI] Pin distributed jobs to 4 GPUs and disable stepcurrent retries

chinmaydk99:
Pin `HIP_VISIBLE_DEVICES=0,1,2,3` for `test_config == 'distributed'` jobs. Distributed tests assume a fixed world size, and the 8-GPU runner was giving each shard a different rank count, surfacing as flaky multi-rank failures.

Disable PyTorch's stepcurrent retry for distributed tests in the wrapper. `pytest-rerunfailures` is already off here, but stepcurrent still relaunches each failed distributed test in a fresh subprocess. Most distributed failures are deterministic multi-rank timeouts, so retrying just wastes 5+ minutes per rank and clutters `--keep-going` output during triage.

### DIFF
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -277,6 +277,17 @@ jobs:
           echo "CC=${ROCM_HOME}/lib/llvm/bin/clang" >> "${GITHUB_ENV}"
           echo "CXX=${ROCM_HOME}/lib/llvm/bin/clang++" >> "${GITHUB_ENV}"
 
+      # Restrict distributed jobs to 4 GPUs (0,1,2,3) on the 8-GPU runner.
+      # Several FSDP / DTensor / pipelining suites assume world_size <= 4 or
+      # are flaky/slow at world_size=8 on gfx94X; pinning to 4 GPUs gives
+      # more deterministic, faster runs. set_gpu_execution_policy() in
+      # run_pytorch_tests_full.py respects a pre-set HIP_VISIBLE_DEVICES,
+      # so policy="all" will only pick from the 4 we expose here.
+      - name: Restrict distributed jobs to 4 GPUs
+        if: matrix.test_config == 'distributed'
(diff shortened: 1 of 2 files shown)


pruthvistony (approved):
We want to remove retries and get failures fast and debug these flaky issues.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -277,6 +277,11 @@ jobs:
           echo "CC=${ROCM_HOME}/lib/llvm/bin/clang" >> "${GITHUB_ENV}"
           echo "CXX=${ROCM_HOME}/lib/llvm/bin/clang++" >> "${GITHUB_ENV}"
 
+      - name: Restrict distributed jobs to 4 GPUs
```
HereThereBeDragons, line 280: you cannot do this. you dont even know if the runner is a multi gpu or a single gpu runner.

also in the container HIP_VISIBLE_DEVICES is already pre-set. so you cant just overwrite it with maybe wrong values (e.g. a single-gpu runner could have HIP_VISIBLE_DEVICES=6 to just use the 7th gpu on the machine). even if there are more gpus they could be of the wrong architecture.

if you want to get a bit more context about the problems have a look here: https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/pytorch_utils.py

you could also exten there the script - based on the available gpus and supported by pytroch - to set HIP_VISIBLE_DEVICES to only N devices (if possible)

external-builds/pytorch/run_pytorch_tests_full.py
```diff
@@ -122,6 +122,19 @@
     "functorch/test_control_flow",
 ]
 
+
+def disable_distributed_stepcurrent_retries(pytorch_dir: Path) -> None:
```
HereThereBeDragons, line 126: what does this exactly do?
i am wondering why this needs a function and canot done either via config or skip_tests/

PR #5132, opened 2026-05-08
[debug-tools] Refactor container options and split rocgdb tests into CPU and GPU runners

lumachad:
[Refactor container options to use arrays in Python](https://github.com/ROCm/TheRock/pull/5132/commits/6e286f40b97c3662e188c53e16fd3c1a33c976b0) 
[6e286f4](https://github.com/ROCm/TheRock/pull/5132/commits/6e286f40b97c3662e188c53e16fd3c1a33c976b0)
Moves hardcoded docker container flags from the GitHub Actions workflow
YAML to the Python configuration script (fetch_test_configurations.py),
using arrays instead of concatenated strings for better maintainability.

This centralized approach:
- Simplifies workflow maintenance by keeping all container configuration
  in one place
- Makes it easier to update flags globally by managing them as arrays
- Reduces YAML complexity
- Enables conditional container options (e.g., CPU vs GPU runners)

Container options are now built in three layers using array operations:
1. Base options (--ipc, --user, --ulimit, --seccomp) - always applied
2. GPU options (--device, --group-add) - applied unless cpu_runner=True
3. Job-specific options - from individual test configurations

Arrays are joined with spaces to produce the final container_options
string passed to the workflow.

[Split rocgdb tests into CPU and GPU runners](https://github.com/ROCm/TheRock/pull/5132/commits/5169ce35a23581a87752f324dfbe79542561adde) 
[5169ce3](https://github.com/ROCm/TheRock/pull/5132/commits/5169ce35a23581a87752f324dfbe79542561adde)
Separates rocgdb test suite into two jobs:
- rocgdb-cpu: runs gdb.dwarf2 tests on standard ubuntu-24.04 runners
- rocgdb-gpu: runs gdb.rocm tests on GPU runners

CPU tests no longer require GPU runners, improving CI resource utilization.
The cpu_runner flag triggers runner selection and container option handling
(via the refactored _build_container_options function).

[Add common_settings for all test jobs](https://github.com/ROCm/TheRock/pull/5132/commits/5de9b6e3589ac2497d6a31f10b7cebde8992ccba) 
[5de9b6e](https://github.com/ROCm/TheRock/pull/5132/commits/5de9b6e3589ac2497d6a31f10b7cebde8992ccba)
Introduces _common_settings dictionary that is applied to all jobs in
the test matrix. This provides a central location for settings that
should be shared across all test configurations.

Currently initialized as empty after analysis showed no settings are
truly common across all jobs. Each job has unique timeout, script,
and artifact requirements. This structure is ready for future common
defaults if needed (e.g., retry policies, common environment variables).

### DIFF
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -187,7 +187,8 @@ jobs:
       # Route to the appropriate runner:
       #   1. Benchmark components use the dedicated benchmark runner
       #   2. Multi-GPU components use the multi-GPU runner
-      #   3. Everything else uses the standard test runner
+      #   3. CPU-only components use standard GitHub runners (azure-linux-scale-rocm)
+      #   4. Everything else uses the standard test runner (GPU runners)
       test_runs_on: >-
         ${{
           (matrix.components.is_benchmark == true &&
@@ -195,6 +196,8 @@ jobs:
             inputs.benchmark_runs_on) ||
           (matrix.components.multi_gpu_runner != '' &&
             matrix.components.multi_gpu_runner) ||
+          (matrix.components.linux_cpu_runner == true &&
+            'azure-linux-scale-rocm') ||
           inputs.test_runs_on
         }}
       platform: ${{ needs.configure_test_matrix.outputs.platform }}
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -57,22 +57,10 @@ jobs:
             (fromJSON(inputs.component).container_image || inputs.default_container_image)
             || null
         }}
-      # --ulimit memlock=-1:-1 - Prevents memory allocation issues with ROCm inside container
-      # --security-opt seccomp=unconfined - enables memory mapping, and is recommended for containers running in HPC environments
-      # --env-file /etc/podinfo/gha-gpu-isolation-settings - Required for GPU isolation on OSSCI MIXXX runners
-      # --user 0:0 - Running as root, by recommendation of GitHub: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support#user
-      options: --ipc host
-        --group-add video
-        --device /dev/kfd
-        --device /dev/dri
-        --group-add 993
-        --group-add 992
-        --group-add 110
-        --ulimit memlock=-1:-1
-        --security-opt seccomp=unconfined
-        --env-file /etc/podinfo/gha-gpu-isolation-settings
-        --user 0:0
-        ${{ fromJSON(inputs.component).container_options }}
+      # Container options are configured in build_tools/github_actions/fetch_test_configurations.py
+      # and include base options (ipc, user, ulimit, seccomp), GPU options (devices, groups),
+      # and any job-specific options
+      options: ${{ fromJSON(inputs.component).container_options }}
     strategy:
       fail-fast: false
       matrix:
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -60,7 +60,7 @@ _BASE_CONTAINER_OPTIONS = [
     "--security-opt seccomp=unconfined",
 ]
 
-# GPU-specific container options (only applied when cpu_runner != True)
+# GPU-specific container options (only applied when linux_cpu_runner != True)
 # --group-add video - Grants access to GPU video group
 # --device /dev/kfd - AMD KFD device for GPU compute
 # --device /dev/dri - Direct Rendering Infrastructure devices
@@ -96,7 +96,7 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
     options_parts = _BASE_CONTAINER_OPTIONS.copy()
 
     # Add GPU-specific options unless this is a CPU-only runner
-    if not job_config.get("cpu_runner", False):
+    if not job_config.get("linux_cpu_runner", False):
         options_parts.extend(_GPU_CONTAINER_OPTIONS)
 
     # Add any job-specific container options
@@ -108,6 +108,16 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
 
     return job_config
 
+
+# Common settings for rocgdb jobs
+_rocgdb_common = {
+    "fetch_artifact_args": "--debug-tools --tests",
+    "timeout_minutes": 30,
+    "platform": ["linux"],
+    "total_shards": 1,
+    "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_rocgdb@sha256:7063e922b4b9145c92f20011674571f1c97b8fad6faaeb0b7d2d165b0bd9ae8b",  # 2026-04-02T21:47:07.506375216Z
+    "container_options": ["--cap-add=SYS_PTRACE"],
+}
 
 test_matrix = {
     # Sanity tests - always run first as a prerequisite for other component tests
@@ -249,15 +259,16 @@ test_matrix = {
             "windows": 1,
         },
     },
-    "rocgdb": {
-        "job_name": "rocgdb",
-        "fetch_artifact_args": "--debug-tools --tests",
-        "timeout_minutes": 45,
-        "test_script": f"python {_get_script_path('test_rocgdb.py')}",
-        "platform": ["linux"],
-        "total_shards": 1,
-        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_rocgdb@sha256:7063e922b4b9145c92f20011674571f1c97b8fad6faaeb0b7d2d165b0bd9ae8b",  # 2026-04-02T21:47:07.506375216Z
-        "container_options": "--cap-add=SYS_PTRACE",
+    "rocgdb-cpu": {
+        **_rocgdb_common,
+        "job_name": "rocgdb-cpu",
+        "test_script": f"python {_get_script_path('test_rocgdb.py')} --tests gdb.dwarf2",
+        "linux_cpu_runner": True,
+    },
+    "rocgdb-gpu": {
+        **_rocgdb_common,
+        "job_name": "rocgdb-gpu",
+        "test_script": f"python {_get_script_path('test_rocgdb.py')} --tests gdb.rocm",
     },
     "rocr-debug-agent": {
         "job_name": "rocr-debug-agent",
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -48,6 +48,67 @@ def _get_script_path(script_name: str) -> str:
     return str(posix_path)
 
 
+# Base container options applied to all Linux containers
+# --ipc host - Allows shared memory between host and container
+# --user 0:0 - Running as root, by recommendation of GitHub: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support#user
+# --ulimit memlock=-1:-1 - Prevents memory allocation issues with ROCm inside container
+# --security-opt seccomp=unconfined - enables memory mapping, and is recommended for containers running in HPC environments
+_BASE_CONTAINER_OPTIONS = [
+    "--ipc host",
+    "--user 0:0",
+    "--ulimit memlock=-1:-1",
+    "--security-opt seccomp=unconfined",
+]
+
+# GPU-specific container options (only applied when cpu_runner != True)
+# --group-add video - Grants access to GPU video group
+# --device /dev/kfd - AMD KFD device for GPU compute
+# --device /dev/dri - Direct Rendering Infrastructure devices
+# --group-add 993,992,110 - Additional GPU-related groups
+# --env-file /etc/podinfo/gha-gpu-isolation-settings - Required for GPU isolation on OSSCI MIXXX runners
+_GPU_CONTAINER_OPTIONS = [
+    "--group-add video",
+    "--device /dev/kfd",
+    "--device /dev/dri",
+    "--group-add 993",
+    "--group-add 992",
+    "--group-add 110",
+    "--env-file /etc/podinfo/gha-gpu-isolation-settings",
+]
+
+
+def _build_container_options(job_config: dict, platform: str) -> dict:
+    """
+    Build the final container_options string by concatenating base, GPU, and job-specific options.
+
+    Args:
+        job_config: The job configuration dictionary
+        platform: The platform (e.g., "linux", "windows")
+
+    Returns:
+        The modified job_config with updated container_options
+    """
+    # Only apply container options for Linux platforms
+    if platform != "linux":
+        return job_config
+
+    # Start with base options (always applied on Linux)
+    options_parts = _BASE_CONTAINER_OPTIONS.copy()
+
+    # Add GPU-specific options unless this is a CPU-only runner
+    if not job_config.get("cpu_runner", False):
+        options_parts.extend(_GPU_CONTAINER_OPTIONS)
+
+    # Add any job-specific container options
+    if "container_options" in job_config:
+        options_parts.extend(job_config["container_options"])
+
+    # Concatenate all parts with a space separator
+    job_config["container_options"] = " ".join(options_parts)
+
+    return job_config
+
+
 test_matrix = {
     # Sanity tests - always run first as a prerequisite for other component tests
     "sanity": {
@@ -62,7 +123,7 @@ test_matrix = {
         },
         # Running docker with cap-add and -v /lib/modules, by recommendation of GitHub:
         # https://rocm.docs.amd.com/projects/amdsmi/en/amd-staging/how-to/setup-docker-container.html
-        "container_options": "--cap-add SYS_MODULE -v /lib/modules:/lib/modules",
+        "container_options": ["--cap-add SYS_MODULE", "-v /lib/modules:/lib/modules"],
     },
     # hip-tests
     "hip-tests": {
@@ -354,7 +415,7 @@ test_matrix = {
         ],
         "test_script": f"python {_get_script_path('test_rocprofiler_sdk.py')}",
         "platform": ["linux"],
-        "container_options": "--cap-add=SYS_PTRACE",
+        "container_options": ["--cap-add=SYS_PTRACE"],
         "total_shards_dict": {
             "linux": 1,
         },
@@ -723,6 +784,9 @@ def run():
                     continue
 
             all_components.append(job_config_data)
+
+    # Build container options for all components (concatenates base, GPU, and job-specific options)
+    all_components = [_build_container_options(c, platform) for c in all_components]
 
     # Separate sanity (always a prerequisite) from the regular component matrix.
     sanity_component = next(
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -109,6 +109,9 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
     return job_config
 
 
+# Common settings applied to all jobs
+_common_settings = {}
+
 # Common settings for rocgdb jobs
 _rocgdb_common = {
     "fetch_artifact_args": "--debug-tools --tests",
@@ -711,6 +714,7 @@ def run():
                 shard_arr = list(range(1, total_shards + 1))
 
                 pal_entry = {
+                    **_common_settings,
                     "job_name": "hip-tests (PAL)",
                     "fetch_artifact_args": base["fetch_artifact_args"],
                     "timeout_minutes": base["timeout_minutes"],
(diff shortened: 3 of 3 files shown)


lumachad:
Experimental for now.

lumachad:
I'd like second opinions on this one. We should be able to free up GPU runners a bit more by doing this. But I want to make sure the strategy is sane.

lumachad:
Adding @jayhawk-commits for feedback.

.github/workflows/test_artifacts.yml
```diff
@@ -185,14 +185,17 @@ jobs:
       # Route to the appropriate runner:
       #   1. Benchmark components use the dedicated benchmark runner
       #   2. Multi-GPU components use the multi-GPU runner
-      #   3. Everything else uses the standard test runner
+      #   3. CPU-only components use standard GitHub runners (ubuntu-24.04)
```
marbre: Are ubuntu 24.04 runners powerful enough or should we rather set up CPU test runners for this on our side, preferably in AWS? @amd-shiraz any thought regarding infra?

.github/workflows/test_component.yml
```diff
@@ -61,18 +61,17 @@ jobs:
...
         --ulimit memlock=-1:-1
         --security-opt seccomp=unconfined
-        --env-file /etc/podinfo/gha-gpu-isolation-settings
-        --user 0:0
-        ${{ fromJSON(inputs.component).container_options }}
+        ${{
+          fromJSON(inputs.component).cpu_runner != true &&
+          '--group-add video --device /dev/kfd --device /dev/dri --group-add 993 --group-add 992 --group-add 110 --env-file /etc/podinfo/gha-gpu-isolation-settings '
+          || ''
+        }}${{ fromJSON(inputs.component).container_options }}
```
marbre: Should we rather move this logic to Python instead of inlining here?

.github/workflows/test_artifacts.yml
lumachad: I've done some test runs and the 24.04 CPU runners (if they're all the same spec) seem to be 4-core systems. Not a lot of processing power, but they get the job done. Obviously if we could have more processing power we could be done with these tests quicker via more concurrency.

.github/workflows/test_component.yml
lumachad: Most probably. Let me look into that.

.github/workflows/test_artifacts.yml
lumachad: From a previous run, I seem to recall we took about 5 to 10 minutes to run things at -j8 for the cpu tests.

.github/workflows/test_component.yml
lumachad: Reworked this now into a refactor commit before the split.

.github/workflows/test_artifacts.yml
```diff
@@ -185,14 +185,17 @@ jobs:
...
+      #   4. Everything else uses the standard test runner (GPU runners)
       test_runs_on: >-
         ${{
           (matrix.components.is_benchmark == true &&
             inputs.benchmark_runs_on != '' &&
             inputs.benchmark_runs_on) ||
           (matrix.components.multi_gpu_runner != '' &&
             matrix.components.multi_gpu_runner) ||
+          (matrix.components.cpu_runner == true &&
+            'ubuntu-24.04') ||
```
geomin12: as Marius mentioned, if the job does finish within a decent time and doesn't crash due to resources, it's good to use this ubuntu24-04.

we can look into shifting into aws machine if the ubuntu machine is not sufficient

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,66 @@ def _get_script_path(script_name: str) -> str:
     return str(posix_path)
 
 
+# Base container options applied to all Linux containers
+# --ipc host - Allows shared memory between host and container
+# --user 0:0 - Running as root, by recommendation of GitHub: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support#user
+# --ulimit memlock=-1:-1 - Prevents memory allocation issues with ROCm inside container
+# --security-opt seccomp=unconfined - enables memory mapping, and is recommended for containers running in HPC environments
+_BASE_CONTAINER_OPTIONS = (
+    "--ipc host --user 0:0 --ulimit memlock=-1:-1 --security-opt seccomp=unconfined"
```
geomin12: should we use arrays? then do a "join" later? could be easier to visualize and add new options (less error prone)

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,66 @@ def _get_script_path(script_name: str) -> str:
...
+)
+
+# GPU-specific container options (only applied when cpu_runner != True)
+# --group-add video - Grants access to GPU video group
+# --device /dev/kfd - AMD KFD device for GPU compute
+# --device /dev/dri - Direct Rendering Infrastructure devices
+# --group-add 993,992,110 - Additional GPU-related groups
+# --env-file /etc/podinfo/gha-gpu-isolation-settings - Required for GPU isolation on OSSCI MIXXX runners
+_GPU_CONTAINER_OPTIONS = "--group-add video --device /dev/kfd --device /dev/dri --group-add 993 --group-add 992 --group-add 110 --env-file /etc/podinfo/gha-gpu-isolation-settings"
```
geomin12: same here for arrays

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -48,6 +48,66 @@ def _get_script_path(script_name: str) -> str:
...
+# Common settings for rocgdb jobs
+_rocgdb_common = {
+    "fetch_artifact_args": "--debug-tools --tests",
+    "timeout_minutes": 30,
+    "platform": ["linux"],
+    "total_shards": 1,
+    "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_rocgdb@sha256:939b8e35887144d1ca4eca928dc2869991339cab869168790e495fc0a5907bbb",
+    "container_options": "--cap-add=SYS_PTRACE",
+}
+
```
geomin12, line 124: could we actually add a "common_settings" array outside of this? in case other components need to add to it

.github/workflows/test_artifacts.yml
lumachad: Just so things are clear, is this CPU runner running other CPU-only tests? Is there a more powerful pool of CPU runners available, like maybe the builders?

lumachad: I do not need them to be ubuntu-24.04 to be clear.

build_tools/github_actions/fetch_test_configurations.py
lumachad: What exactly would you like to share on those common_settings?

lumachad: I've implemented this in a separate commit, but I couldn't find any common ground across all test jobs. I can see the value for future use, but for now I've set it to empty.

build_tools/github_actions/fetch_test_configurations.py
lumachad: Done for both _BASE and _GPU.

.github/workflows/test_artifacts.yml
lumachad: @amd-shiraz Any thoughts here?

.github/workflows/test_artifacts.yml
marbre: I think right now we only have the builder pool and those are likely oversized for testing. If we can go with the GH runners, that's okay, if they are not sufficient, we would need to set up another pool IMHO.

.github/workflows/test_artifacts.yml
marbre: Yes, these have 4 cores and 16 GB memory, see https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories. If that is sufficient, we can go with it. 5-10 min would below other test shards.

.github/workflows/test_artifacts.yml
lumachad: I'll run some checks and see how it behaves first.

lumachad:
Latest serial CPU run took about 10 minutes. Let me try to parallelize that and see how much we can reduce that.

The GPU run took roughly 20 minutes. Can't be parallelized yet.

lumachad:
> Latest serial CPU run took about 10 minutes. Let me try to parallelize that and see how much we can reduce that.
> 
> The GPU run took roughly 20 minutes. Can't be parallelized yet.

Nevermind. We don't have a setting to control the number of parallel jobs yet in our rocgdb CI script. So it's either nproc or serial. If we have a few CPU's, that's fine. But if we have a bigger system, that is a lot of processes.

Given we're migrating the CI scripts to their own components (away from TheRock), I think this improvement can wait until the scripts are being called from their new locations. Therefore I think this PR is fine as-is.

.github/workflows/test_artifacts.yml
geomin12: let us know!

.github/workflows/test_artifacts.yml
```diff
@@ -187,14 +187,17 @@ jobs:
...
+      #   3. CPU-only components use standard GitHub runners (ubuntu-24.04)
+      #   4. Everything else uses the standard test runner (GPU runners)
       test_runs_on: >-
         ${{
           (matrix.components.is_benchmark == true &&
             inputs.benchmark_runs_on != '' &&
             inputs.benchmark_runs_on) ||
           (matrix.components.multi_gpu_runner != '' &&
             matrix.components.multi_gpu_runner) ||
+          (matrix.components.cpu_runner == true &&
```
geomin12: oh also, can we specify `linux_cpu_runner`?

geomin12 (commented):
i see the tests pass! although takes ~ 20 mins. feel free to use azure-linux-scale-rocm as the CPU runner as this has more resources

jayhawk-commits (commented):
I do not have any concerns outside of what was discussed already in the comments in this PR. In the future, we may enable our own set of lower-powered CPU-only runners that may be better than the GitHub-provided option for use cases like emulation, but there is no timeline on that.

.github/workflows/test_artifacts.yml
lumachad: Switched to azure's to see how it goes.

lumachad:
> i see the tests pass! although takes ~ 20 mins. feel free to use azure-linux-scale-rocm as the CPU runner as this has more resources

@geomin12 I flipped the code to use the azure cpu runners. Curiously enough, it takes longer than the gh ubuntu-24.04 runners. I'm fine either way. Unless we want to define a pool of linux-cpu-runners, I'm fine with the azure ones.

Later on after rocgdb script updates we will parallelize these tests anyway, so the runtime will fall for cpu tests.

lumachad:
> > i see the tests pass! although takes ~ 20 mins. feel free to use azure-linux-scale-rocm as the CPU runner as this has more resources
> 
> @geomin12 I flipped the code to use the azure cpu runners. Curiously enough, it takes longer than the gh ubuntu-24.04 runners. I'm fine either way. Unless we want to define a pool of linux-cpu-runners, I'm fine with the azure ones.
> 
> Later on after rocgdb script updates we will parallelize these tests anyway, so the runtime will fall for cpu tests.

I think I actually had a bug that ran things on the gh runners. Trying again...

lumachad:
Ok. Confirmed things ran on the azure ones now. Serially it still takes about 10 minutes. But let's stick to the azure ones since they have more cores and memory. When I enable parallel runs on cpu, it will be much faster.

lumachad:
Re-triggering to make sure things are still sane.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -188,15 +262,16 @@ def _get_script_path(script_name: str) -> str:
...
-        "timeout_minutes": 45,
-        "test_script": f"python {_get_script_path('test_rocgdb.py')}",
-        "platform": ["linux"],
-        "total_shards": 1,
-        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_rocgdb@sha256:7063e922b4b9145c92f20011674571f1c97b8fad6faaeb0b7d2d165b0bd9ae8b",  # 2026-04-02T21:47:07.506375216Z
-        "container_options": "--cap-add=SYS_PTRACE",
+    "rocgdb-cpu": {
+        **_rocgdb_common,
+        "job_name": "rocgdb-cpu",
+        "test_script": f"python {_get_script_path('test_rocgdb.py')} --tests gdb.dwarf2",
```
geomin12, line 268: for a future PR, but we could update the filter to something better :)

geomin12 (approved):
lgtm thanks for update

build_tools/github_actions/fetch_test_configurations.py
lumachad: How so? Do you mean the rocgdb-cpu/rocgdb-gpu names?

marbre (approved):
Is 20 vs 10 minutes really that bad? The debug-tools stage is kicked off in parallel to math-libs but will, assuming it gets picked up at the same time, finishing compilation earlier. Thus debug-tools tests should finish before some of the math-libs related tests even get kicked off?

@geomin12, wdyt? I fear we otherwise "forget" about moving these tests to a lower CPU core count pool and running single-threaded on the builders (which have high pressure) feels lake a waste of resources. It is not as bad a running CPU-only tests on GPU-equipped machines, therefore approving but we should still consider running on the free runners.

lumachad:
> Is 20 vs 10 minutes really that bad? The debug-tools stage is kicked off in parallel to math-libs but will, assuming it gets picked up at the same time, finishing compilation earlier. Thus debug-tools tests should finish before some of the math-libs related tests even get kicked off?

Just because I didn't quite get the context, is this in relation to running things on the azure runners as opposed to GH runners? To be honest the GH runners seem fairly OK for our current CPU test demands to be honest. If/When we extend the CPU tests in the future, we could adjust this if we hit a bottleneck. So do let me know if you want me to switch back to the GH runners.

> 
> @geomin12, wdyt? I fear we otherwise "forget" about moving these tests to a lower CPU core count pool and running single-threaded on the builders (which have high pressure) feels lake a waste of resources. It is not as bad a running CPU-only tests on GPU-equipped machines, therefore approving but we should still consider running on the free runners.

The CPU tests can be parallelized quite heavily (just not at the moment). Once the testing script becomes smarter about that, we can size the concurrency based on the number of available cores.

marbre:
Oh I think I am actually wrong. In multi-arch tests get kicked off after all build stages passed and not after an individual stage passed. So 20 min is indeed long. Let's merge as is.

geomin12:
> Is 20 vs 10 minutes really that bad? The debug-tools stage is kicked off in parallel to math-libs but will, assuming it gets picked up at the same time, finishing compilation earlier. Thus debug-tools tests should finish before some of the math-libs related tests even get kicked off?
> 
> @geomin12, wdyt? I fear we otherwise "forget" about moving these tests to a lower CPU core count pool and running single-threaded on the builders (which have high pressure) feels lake a waste of resources. It is not as bad a running CPU-only tests on GPU-equipped machines, therefore approving but we should still consider running on the free runners.

I think 20 mins on free runners are fine!

build_tools/github_actions/fetch_test_configurations.py
lumachad: Will be handled in the future when the update test_rocgdb.py script lands.

ScottTodd:
Please prefer "squash and merge" over "rebase and merge" for future PR merges. This added 3 commits to `main`, and it looks like our CI configure script didn't handle that git diff correctly: https://github.com/ROCm/TheRock/actions/runs/26232906428/job/77197971377
```
fatal: bad object bb1c030b66709b958a09596ff1bc76923024d497
  File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 1119, in <module>
    main()
  File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 1108, in main
    git_context = GitContext.from_repo(base_ref=ci_inputs.base_ref)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 250, in from_repo
    changed_files = get_git_modified_paths(base_ref)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/TheRock/TheRock/build_tools/github_actions/configure_ci_path_filters.py", line 43, in get_git_modified_paths
    return subprocess.run(
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'diff', '--name-only', 'bb1c030b66709b958a09596ff1bc76923024d497']' returned non-zero exit status 128.
```
(we should fix that script, but it looks like a squash commit would have been more appropriate here?)

lumachad:
> Please prefer "squash and merge" over "rebase and merge" for future PR merges. This added 3 commits to `main`, and it looks like our CI configure script didn't handle that git diff correctly: https://github.com/ROCm/TheRock/actions/runs/26232906428/job/77197971377
> 
> ```
> fatal: bad object bb1c030b66709b958a09596ff1bc76923024d497
> Traceback (most recent call last):
>   File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 1119, in <module>
>     main()
>   File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 1108, in main
>     git_context = GitContext.from_repo(base_ref=ci_inputs.base_ref)
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/runner/work/TheRock/TheRock/./build_tools/github_actions/configure_multi_arch_ci.py", line 250, in from_repo
>     changed_files = get_git_modified_paths(base_ref)
>                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/runner/work/TheRock/TheRock/build_tools/github_actions/configure_ci_path_filters.py", line 43, in get_git_modified_paths
>     return subprocess.run(
>            ^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/subprocess.py", line 571, in run
>     raise CalledProcessError(retcode, process.args,
> subprocess.CalledProcessError: Command '['git', 'diff', '--name-only', 'bb1c030b66709b958a09596ff1bc76923024d497']' returned non-zero exit status 128.
> ```
> 
> (we should fix that script, but it looks like a squash commit would have been more appropriate here?)

@ScottTodd Thanks. Noted. Any fallout I need to take care of? Or is this a transient failure?

ScottTodd:
No lasting impact, just a few extra commits in history (and those aren't too bad, we've seen stacks of 20+ commits with things like "typo fix" in them before which is why I mention it)

jayhawk-commits:
Looking at the Multi-arch CI runs on this PR plus all the commits on main after, the Windows sanity check is failing a template check. See https://github.com/ROCm/TheRock/actions/runs/26146790105/job/76950097541

```
Error: The template is not valid. ROCm/TheRock/.github/workflows/test_component.yml@9ef51322a8468ef0f65690f7cfd8786ce527a7f1 (Line: 63, Col: 16): A sequence was not expected
```

I don't see the issue on the commit on main before this PR was merged. https://github.com/ROCm/TheRock/actions/runs/26231193664/job/77302432578

I am seeing the invalid template issue on the submodule bump PRs. I will use GitHub's revert button shortly.

jayhawk-commits:
> Looking at the Multi-arch CI runs on this PR plus all the commits on main after, the Windows sanity check is failing a template check. See https://github.com/ROCm/TheRock/actions/runs/26146790105/job/76950097541
> 
> ```
> Error: The template is not valid. ROCm/TheRock/.github/workflows/test_component.yml@9ef51322a8468ef0f65690f7cfd8786ce527a7f1 (Line: 63, Col: 16): A sequence was not expected
> ```
> 
> I don't see the issue on the commit on main before this PR was merged. https://github.com/ROCm/TheRock/actions/runs/26231193664/job/77302432578
> 
> I am seeing the invalid template issue on the submodule bump PRs. I will use GitHub's revert button shortly.

I see you have a fix-forward in #5413. Since Windows sanity checks are going through on that PR, I will approve and merge it to unblock submodule bump PRs.

lumachad:
> Looking at the Multi-arch CI runs on this PR plus all the commits on main after, the Windows sanity check is failing a template check. See https://github.com/ROCm/TheRock/actions/runs/26146790105/job/76950097541
> 
> ```
> Error: The template is not valid. ROCm/TheRock/.github/workflows/test_component.yml@9ef51322a8468ef0f65690f7cfd8786ce527a7f1 (Line: 63, Col: 16): A sequence was not expected
> ```
> 
> I don't see the issue on the commit on main before this PR was merged. https://github.com/ROCm/TheRock/actions/runs/26231193664/job/77302432578
> 
> I am seeing the invalid template issue on the submodule bump PRs. I will use GitHub's revert button shortly.

Yeah, sorry. I spotted that one afterwards. Hopefully did not disrupt things too much. Moving forward I'll have to a be a bit more mindful of the Windows workflows. It shows a few unrelated failures, but this one was clearly off.

PR #5134, opened 2026-05-08
Revert "[cmake] Handle sanitizers for autotools projects"

lumachad:
Reverts ROCm/TheRock#4674 due to https://github.com/ROCm/TheRock/pull/5129.

gfortran seems to be getting through the system compiler guard in the cmake asan logic in cmake/therock_sanitizers.cmake.

ROCM-24371


geomin12 (approved):

tony-davis (approved):
LGTM. Thank you!

PR #5138, opened 2026-05-08
Bump fusilli and iree sublibraries

rsuderman:
Bump fusilli from aeed46fe to 90053d0b.
Bump iree from 192602f8 to b1ddcab6.

## Motivation

Need to update the fusilli and iree deps to extend fusilli plugin support

## Technical Details

Bumping forward to current HEAD.

## Test Plan

Existing CI should catch issues.

## Test Result

CI should identify any majori ssues.

## Submission Checklist


AaronStGeorge:
If I'm not mistaken this will contain the recent breaking changes in https://github.com/iree-org/fusilli/pull/393 and we'll get a build error.

Unfortunately at this point CI is deadlocked, the dependency flow between TheRock submodules (iree-libs/iree, iree-libs/fusilli) and the fusilli-provider library inside rocm-libraries doesn't allow us to get a breaking change through without turning off the plugin build: 

- TheRock's fusilli pin can't move forward until rocm-libraries' fusilli-provider is updated.
- rocm-libraries' fusilli-provider can't be updated until TheRock provides the new fusilli, because rocm-libraries CI builds fusilli-provider against TheRock's pinned fusilli.

Here's the proposed fix -> https://github.com/ROCm/TheRock/pull/5133

You could update the PR to only bump to https://github.com/iree-org/fusilli/commit/dc198af09e4b1d5f19937fbb3d63c23525dafc38, we should be free to jump to that point.

Issue #5139, opened 2026-05-08
Request: mirror elfutils-0.195.tar.bz2 to rocm-third-party-deps S3


FIM43-Redeye:
## Request
TheRock's `third-party/sysdeps/linux/elfutils/CMakeLists.txt` currently uses
`https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/elfutils-0.192.tar.bz2`
as the source URL.

Could the elfutils 0.195 tarball be mirrored to the same S3 bucket?

## Why
Planning a bump PR (0.192 -> 0.195) to fix #5120 at source. Since the S3
mirror does not yet host 0.195, the PR would either need a placeholder
sourceware URL (and a follow-up commit to swap to S3 once mirrored), or
this issue can be resolved first so the PR lands cleanly with the proper
S3 URL.

## Tarball details
- Upstream: https://sourceware.org/elfutils/ftp/0.195/elfutils-0.195.tar.bz2
- SHA512: `e1e1fdf4f7f72bf520deb0103bb8fd208dc247bab14af100c6fb56d38c0ef6c6d54ff5bd15b84e4f70e2b2ea481e5999fea2842360f8932a861122df79b5fc8f`
- Size: 12,032,640 bytes

If there is a documented process for requesting third-party mirror
additions that I should follow instead of an issue, happy to redirect.

ScottTodd:
The steps for updating are at https://github.com/ROCm/TheRock/blob/main/docs/development/git_chores.md#updating-a-third-party-mirror. Just needs someone on the team with access to the AWS project to follow the steps.

We were talking about creating some more hands-free process for requesting mirrored deps (cc @amd-shiraz ). I think the current thinking there would involve a private repository though... supporting requests via public issues would be good too.

ScottTodd:
Here you go: https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/elfutils-0.195.tar.bz2

FIM43-Redeye:
Verified — S3 mirror is serving `elfutils-0.195.tar.bz2` and SHA512 matches the canonical sourceware tarball. Filing the bundled bump PR now. Thanks @ScottTodd!

marbre:
Closing the issue as the source tarball got uploaded.

PR #5142, opened 2026-05-09
[elfutils] Bump bundled elfutils 0.192 -> 0.195

### LINKED ISSUE #5120: [issue] elfutils 0.192 fails to build under amd-llvm/clang due to -Werror=incompatible-pointer-types-discards-qualifiers
## Summary

The bundled elfutils sysdep (third-party/sysdeps/linux/elfutils, elfutils 0.192) fails to compile under the amd-llvm/clang toolchain that #4032 routed third-party builds through. The failure is a `-Werror`-promoted warning in `libcpu/riscv_disasm.c:1259` that gcc was lenient about but clang treats as a hard error.

This is **adjacent to** but distinct from #3477's catalog item 6 (which is the rocprofiler-systems-vendored elfutils with a different GCC-15 warning). The failure described here is in the TheRock-bundled elfutils, under clang.

## Reproduction

A clean build with the amd-llvm/clang toolchain (post-#4032) on Ubuntu 26.04 + clang from the amd-llvm tree, configuring elfutils via `third-party/sysdeps/linux/elfutils/CMakeLists.txt`, hits:

```
clang -D_GNU_SOURCE -DHAVE_CONFIG_H ... -std=gnu99 -Wall -Wshadow -Wformat=2 -Wold-style-definition
      -Wstrict-prototypes -Wnull-dereference -Wimplicit-fallthrough -Werror -Wunused -Wextra ...
      -fdollars-in-identifiers -D_FORTIFY_SOURCE=3 -g -O2 -fno-addrsig
      -c .../libcpu/riscv_disasm.c

riscv_disasm.c:1259:24: error: initializing 'struct known_csrs *' with an expression of type 'const void *'
                       discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
 1259 |                   struct known_csrs *found = bsearch (&key, known,
      |                                      ^       ~~~~~~~~~~~~~~~~~~~~~
```

Cause: `bsearch` returns `const void *` (C11), and `known` is `static const struct known_csrs known[]`. Assigning the return to non-const `struct known_csrs *` discards the const qualifier. Clang's `-Werror -Wincompatible-pointer-types-discards-qualifiers` rejects it; gcc was lenient.

## Workaround

Adding a `CFLAGS` override to the configure invocation in `third-party/sysdeps/linux/elfutils/CMakeLists.txt` is sufficient:

```cmake
"CFLAGS=-g -O2 -Wno-error=incompatible-pointer-types-discards-qualifiers"
```

(The explicit `-g -O2` preserves configure's defaults, which would otherwise be dropped when `CFLAGS` is set explicitly. The `-Wno-error=...` demotes the specific clang error to a warning while keeping `-Werror` for everything else.)

This unblocks the build; verified on a local TheRock build.

## Suggested resolution

Two paths, both viable:

1. **Upstream fix**: file an elfutils bug at `libcpu/riscv_disasm.c:1259` and patch the assignment to `const struct known_csrs *found = bsearch(...)` (or add an explicit cast if the const propagation is invasive).
2. **TheRock-side workaround**: land the `CFLAGS` patch above in `third-party/sysdeps/linux/elfutils/CMakeLists.txt` with a comment pointing at the upstream issue.

Doing both in parallel (file upstream, land workaround as a bridge) seems reasonable. Happy to file a PR for the workaround if that's the preferred path.

FIM43-Redeye:
## Summary

Bumps the TheRock-bundled elfutils from 0.192 to 0.195 (released 2026-04-15).

## Why

elfutils 0.192's `libcpu/riscv_disasm.c:1259` discards `const` from a `bsearch` return, tripping amd-llvm clang's `-Werror=incompatible-pointer-types-discards-qualifiers` (#5120) and GCC 15's `-Werror=discarded-qualifiers`. 0.195 fixes this at source — the result is correctly typed `const struct known_csrs *` (now at `:1262`).

Same root-cause class as the rocprofsys-vendored elfutils 0.188 backport that landed in rocm-systems#5574 (2026-04-30), but applied at the more current bundled-side version where elfutils carries the fix natively rather than as a downstream patch.

## What changes

Single-file: `third-party/sysdeps/linux/elfutils/CMakeLists.txt`
- URL bumped to the S3 mirror at `elfutils-0.195.tar.bz2` (uploaded by @ScottTodd via #5139)
- SHA512 verified against the canonical sourceware tarball

## Verification

- Full integrate-tip build (Ubuntu 26.04, GCC 15.2.0, amd-llvm clang) — green
- Bundled elfutils builds clean — riscv_disasm.c compiles without the const-discard cascade
- amd-llvm, debug-tools, profiler, core consumers re-link without recompilation, confirming 0.192 → 0.195 is libdw/libelf ABI-compatible
- `libdw.so` / `libelf.so` soname unchanged

Closes #5120.


scottt:
@marbre this PR looks good to me.

I've verified the SHA512 sum of the upstream elfutils-0.195 tarball is `e1e1fdf4f7f72bf520deb0103bb8fd208dc247bab14af100c6fb56d38c0ef6c6d54ff5bd15b84e4f70e2b2ea481e5999fea2842360f8932a861122df79b5fc8f`

I've created another PR to disable treating compiler warnings as errors by default in #5610. Landing that would make elfutils build failure happen a lot less often as compilers get updated.

ScottTodd (approved):

PR #5164, opened 2026-05-11
Add --include-intermediate-dirs flag to S3 index generator

marbre:
## Motivation

generate_s3_index.py only discovered directories that directly contain files. Directories whose contents were exclusively subdirectories (e.g. `<run>/logs/`, `<run>/logs/math-libs/`, `<run>/packages/deb/dists/`) had no index.html generated, so clicking through from a parent index 404s on the static-hosted bucket. While this is not an issue on server-side (as the lambda handler takes care), this is hit when executing the script manually.

## Technical Details

Add an opt-in `--include-intermediate-dirs` flag that expands the discovered set with all ancestors up to the run root. The existing `_list_files_s3` already lists` CommonPrefixes` correctly, so no listing-side change is needed.

## Test Plan and Results

Default behavior is unchanged. Verified against run 25528390694 (28 dirs without the flag, 36 with the flag, adding the 8 intermediate-only directories).

## Submission Checklist

### DIFF
--- a/build_tools/generate_s3_index.py
+++ b/build_tools/generate_s3_index.py
@@ -200,6 +200,24 @@ def _discover_dirs_with_files_s3(s3_client, bucket: str, run_prefix: str) -> lis
     return sorted(dirs)
 
 
+def _expand_with_intermediate_dirs(dirs: list[str], run_prefix: str) -> list[str]:
+    """Add every ancestor directory of each entry, up to and including run_prefix.
+
+    Without this, intermediate directories that contain only subdirectories
+    (e.g. ``<run>/logs/`` when only ``<run>/logs/<component>/`` has files)
+    have no index.html generated, so links from a parent index 404.
+    """
+    all_dirs: set[str] = set(dirs)
+    all_dirs.add(run_prefix)
+    for d in dirs:
+        while "/" in d and d != run_prefix:
+            d = d.rsplit("/", 1)[0]
+            all_dirs.add(d)
+            if d == run_prefix:
+                break
+    return sorted(all_dirs)
+
+
 # ---------------------------------------------------------------------------
 # Local listing helpers (for --output-dir mode)
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


build_tools/generate_s3_index.py
```diff
@@ -386,5 +412,14 @@ def run(args) -> None:
...
     )
+    parser.add_argument(
+        "--include-intermediate-dirs",
+        action="store_true",
+        help=(
+            "Also generate index.html for directories whose contents are "
+            "only subdirectories (no direct files). Without this, links to "
+            "such directories from a parent index 404."
+        ),
+    )
```
ScottTodd, line 423: Why make this optional and opt-in? Is there a case where we wouldn't want intermediate directories to also include index pages for complete navigation?

I think the code would be easier to follow if the `_discover_dirs_with_files_s3` and `_discover_dirs_with_files_local` helper functions just returned a complete list without needing the additional run of `_expand_with_intermediate_dirs`.

build_tools/generate_s3_index.py
```diff
@@ -386,5 +412,14 @@ def run(args) -> None:
         action="store_true",
         help="Print what would be uploaded without actually uploading",
     )
+    parser.add_argument(
+        "--include-intermediate-dirs",
+        action="store_true",
```
ScottTodd, line 417: nit: could use BooleanOptionalAction instead of `action="store_true"` here: https://docs.python.org/3/library/argparse.html#argparse.BooleanOptionalAction, so an explicit `--no-include-intermediate-dirs` option is generated too

(the `--dry-run` argument above doesn't use that, though I guess we don't really have a convention of providing a `--no-dry-run` mode)

ScottTodd (approved):
Could go another direction with this (see comments) but this is also fine-as is. Up to you.

build_tools/generate_s3_index.py
marbre: Yeah, the lambda handler uses this differently and it is less risk to make this `opt-in`, refactor the lambda handler eventually and switch over after this.

build_tools/generate_s3_index.py
marbre: Filed #5237 to keep track as I run out of dev time for today.

PR #5168, opened 2026-05-11
Drop `generate_flat_index`

marbre:
Flat indexes for multi-arch, kpack-split enabled packages are generated server side via `generate_s3_index.py`. The function is no longer needed and can be dropped.

### DIFF
--- a/build_tools/packaging/tests/generate_local_index_test.py
+++ b/build_tools/packaging/tests/generate_local_index_test.py
@@ -19,7 +19,6 @@ from pathlib import Path
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent / "python"))
 
 from generate_local_index import (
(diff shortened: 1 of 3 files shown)


build_tools/packaging/tests/generate_local_index_test.py
HereThereBeDragons: needs to be removed

HereThereBeDragons (approved):
see comment

cgoea (approved):

araravik-psd (approved):
LGTM! Checked from the lambda logs that with uploads the indexes are getting refreshed fine.

build_tools/packaging/tests/generate_local_index_test.py
marbre: Good catch.

PR #5084, opened 2026-05-06
Revert "Add COMGR build-time unit tests to CI (#4881)"

skganesan008:
This reverts commit 54acd32c58aa0b5c2911b42b53bc23b0e0ddd6d6.

## Motivation

Fixes https://github.com/ROCm/TheRock/issues/5074. This commit inadvertently disabled building `offload-arch` on Windows, which caused sanity check tests to fail (among other downstream effects).

## Technical Details

See details at https://github.com/ROCm/TheRock/pull/4881#issuecomment-4390716989. 

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5197, opened 2026-05-12
Add COMGR build-time unit tests to CI (re-land).

kasaurov:
## Motivation

Adds build-time Comgr unit tests (LIT + CTest) to CI, building on top of the `therock_subproject_build_test` infrastructure from PR #4790.

N.B. (1) The tests are non-blocking at the moment
N.B. (2) The PR is a re-land of reverted PR https://github.com/ROCm/TheRock/pull/4881 with:
* fix for offload-arch Windows regression (remove incorrect non-Windows guard)
* narrow tool-filtering condition to allow minimal builds with Comgr-only tests

## Technical Details

Based on PR #4790 (`therock_subproject_build_test` mechanism) and PR #4259 (Comgr test integration).

- Comgr test integration (`compiler/CMakeLists.txt`):
  - Adds `build-test-amd-comgr` target running LIT tests (`llvm-lit`) followed by CTest
  - `THEROCK_BUILD_COMGR_TESTS` defaults to `THEROCK_BUILD_TESTING` when not explicitly set
  - Removes option(`THEROCK_BUILD_COMGR_TESTS ...`) from top-level `CMakeLists.txt` to allow proper defaulting
  - Removes `llvm-lit` install/wrapper logic from `pre_hook_amd-llvm.cmake` (tests run from build tree)
- Independent test command execution (`cmake/therock_subproject.cmake`):
  - Modifies `therock_cmake_subproject_build_test()` to generate a CMake runner script that executes each COMMAND via `execute_process` independently -- a failure in one (e.g., LIT) does not prevent others (e.g., CTest) from running
  - Update suites enumeration schema -- for more than one suite enumerate staring from the first one (comgr build-test N)
  - Adds empty line separators between test command outputs
  - Reports combined pass/fail result at the end
  - Improves build test output with per-suite headers and a summary table
- CI integration:
  - Adds "Run build tests" step to multi-arch Linux and Windows workflows for compiler-runtime stage
  - Step is non-blocking (`continue-on-error: true`) with `TEATIME_FORCE_INTERACTIVE: 1` for visible output
  - `configure_stage.py` sets `-DTHEROCK_BUILD_TESTING=ON` for `compiler-runtime`
- CI invocation: `ninja build-test-amd-comgr` or `ninja therock-build-tests`

## Test Plan

The TheRock CI run.

## Test Result

- Linux: All Comgr unit tests pass -- 23/23 LIT tests, 31/31 CTest tests
- Windows: All Comgr unit tests pass - 23/23 LIT Tests, 31/31 CTest tests

image

## Submission Checklist


.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -168,6 +168,14 @@ jobs:
...
           cmake --build "${BUILD_DIR}" --target stage-${STAGE_NAME} therock-artifacts -- -k 0
 
+      - name: Run build tests
+        if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
+        continue-on-error: true
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
+
```
geomin12, line 178: can we include this with tests below?

```
- name: Test Packaging
        run: |
          ctest --test-dir ${BUILD_DIR} --output-on-failure
```

.github/workflows/multi_arch_build_windows_artifacts.yml
```diff
@@ -190,6 +190,14 @@ jobs:
         run: |
           cmake --build "${BUILD_DIR}" --target stage-${STAGE_NAME} therock-artifacts -- -k 0
 
+      - name: Run build tests
+        if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
+        continue-on-error: true
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
```
geomin12, line 199: same here, seems duplicated

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
geomin12: we want to try to keep component specific changes outside of yml files so it stays generic enough. maybe we can work some ctest logic so it runs only on compiler-runtime?

compiler/CMakeLists.txt
```diff
@@ -74,6 +78,7 @@ if(THEROCK_ENABLE_COMPILER)
 
       # Features
       -DLLVM_INCLUDE_TESTS=${THEROCK_ENABLE_LLVM_TESTS}
+      -DTHEROCK_BUILD_COMGR_TESTS=${THEROCK_BUILD_COMGR_TESTS}
```
geomin12, line 81: i know these are build tests (CPU-only) from what i see. maybe we can include these in the testing phase? i know rocgdb is introducing cpu only tests alongside gpu only here #5132

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
ScottTodd: See prior discussion at https://github.com/ROCm/TheRock/pull/4881#discussion_r3174362336. My preference is still to unconditionally build `therock-build-tests` at the same time as `therock-artifacts` above, so we overlap builds and unit tests and don't need to make any changes to these workflows when other components include their own unit tests using `therock_cmake_subproject_build_test`

compiler/CMakeLists.txt
kasaurov: The short answer is No, because comgr tests run from the build tree and would require the LLVM build tree to be available in the testing phase, which only has access to staged/installed artifacts.
Comgr unit tests are build-tree tests -- they verify the build artifacts in the build tree itself (llvm-lit runs on the build directory, comgr CTest runs from the build directory). They don't need installed packages, GPUs, or the full ROCm stack, but they do need many LLVM build tree components. This is fundamentally different from what PR #5132 does -- rocgdb tests are runtime tests that run compiled binaries against an installed debugger, while comgr tests validate the compiler/comgr build itself.
I have a follow-up PR to enable LLVM unit tests, which require even more build tree components. That's why @stellaraccident proposed a dedicated mechanism for such case in #4790.

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kasaurov: These are two different test mechanisms that can't be combined:
- `Test Packaging` would run ctest on the staged/installed artifacts (e.g., therock-examples-cpp-sdk-user which validates the installed SDK)
- `Run build tests` runs `therock-build-tests` -- build-tree tests (llvm-lit, subproject ctest) that operate on the build tree itself. They also need `continue-on-error: true` (non-blocking during bringup) and `TEATIME_FORCE_INTERACTIVE: 1` for streaming output.
Also worth noting that on Windows, "Test Packaging" is currently commented out (`TODO(scotttodd) -- hipsparselt-config.cmake issue`), so merging would silently disable build tests there too.

Regarding keeping the YAML generic: the step already calls `therock-build-tests`, which is a single aggregation target -- any component that registers tests via `therock_cmake_subproject_build_test()` (e.g., rocjitsu already does) automatically appears under this target with zero workflow changes. The `stage_name == 'compiler-runtime'` guard is a temporary optimization since build tests currently only exist for that stage; it can be removed as more components add tests.

Regarding merging into the Build stage step -- this was attempted in #4881 (commit a6cb2d2) and reverted (4dc8ac8) because if tests fail, ninja returns non-zero and kills the entire step, preventing artifact upload even when the build itself succeeded. The `continue-on-error` at the step level allows flexibility. The separate step gives clean separation: build failures are hard failures, test failures are visible but non-blocking (at least at the moment). It also improves readability of test results in CI UI.

.github/workflows/multi_arch_build_windows_artifacts.yml
kasaurov: Same step, same rationale as above -- the Linux and Windows workflows are separate files so the step appears in both.

ScottTodd (approved):
I checked that binary size and built binaries look correct here (those were what prompted the revert at https://github.com/ROCm/TheRock/pull/5084)

geomin12 (approved):
lgtm, thanks for the clarification

lumachad:
@kasaurov @ScottTodd @geomin12

There seems to be some bad interaction between these changes and doing a build outside of the manylinux container.

```
  [amd-comgr configure] -- Resolving system find_package(Threads) (not found in super-project ZLIB;zstd;NUMA;BZip2;liblzma;LibElf;libdw;AMDDeviceLibs;Clang;LLD;LLVM;ROCmCMakeBuildTools;ROCM)
  [amd-comgr configure] -- Resolving system find_package(hip) (not found in super-project ZLIB;zstd;NUMA;BZip2;liblzma;LibElf;libdw;AMDDeviceLibs;Clang;LLD;LLVM;ROCmCMakeBuildTools;ROCM)
  [amd-comgr configure] -- Resolving super-project find_package(AMDDeviceLibs BYPASS_PROVIDER NO_DEFAULT_PATH PATHS /develop/work/repos/TheRock/build/compiler/amd-llvm/dist/lib/llvm/lib/cmake/AMDDeviceLibs)
  [amd-comgr configure] CMake Error at /develop/work/repos/TheRock/cmake/therock_subproject_dep_provider.cmake:68 (message):
  [amd-comgr configure]   Project contains find_package(amd_comgr) for a package availabe in the
  [amd-comgr configure]   super-project but not declared: Add a BUILD_DEPS or RUNTIME_DEPS
  [amd-comgr configure]   appropriately
  [amd-comgr configure] Call Stack (most recent call first):
  [amd-comgr configure]   /usr/share/cmake-3.28/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
  [amd-comgr configure]   /opt/rocm/lib/cmake/hip/hip-config-amd.cmake:125 (find_dependency)
  [amd-comgr configure]   /opt/rocm/lib/cmake/hip/hip-config.cmake:138 (include)
  [amd-comgr configure]   /develop/work/repos/TheRock/cmake/therock_subproject_dep_provider.cmake:73 (find_package)
  [amd-comgr configure]   test/CMakeLists.txt:214 (find_package)
  [amd-comgr configure]
  [amd-comgr configure]
  [amd-comgr configure] -- Configuring incomplete, errors occurred!
  [amd-comgr configure FAILED WITH CODE 1 in 1 seconds]
  ninja: build stopped: subcommand failed.
```

It seems we fallback to trying to find hip in /opt/rocm. A brief analysis:

---

The error is a chain reaction from -DTHEROCK_BUILD_COMGR_TESTS=ON on a machine that has a system ROCm installed at /opt/rocm:

  1. test/CMakeLists.txt:214 calls find_package(hip)
  2. hip is not a declared super-project dependency for amd-comgr, so the dep provider falls through to the system ROCm at /opt/rocm
  3. The system /opt/rocm/lib/cmake/hip/hip-config-amd.cmake calls find_dependency(amd_comgr) as part of its own config
  4. The super-project dep provider intercepts that call, recognizes amd_comgr as a super-project package, and fatally errors because it wasn't declared as a BUILD_DEPS or RUNTIME_DEPS for the comgr component

  The root cause is that the comgr tests are pulling in system HIP (not the in-tree build), and system HIP's cmake config has a transitive dependency on amd_comgr that trips the super-project's guard.

  The fix options are:

  - Preferred: Declare hip as a test-only dependency in amd-comgr's super-project CMake, so the dep provider resolves it from the in-tree build rather than /opt/rocm
  - Simpler workaround: Have the comgr test CMake guard the find_package(hip) so it only runs when HIP is actually available from the super-project dist (i.e., after core/clr is built), not from the system

  This wouldn't happen on a clean machine without /opt/rocm, which is why it likely passed CI but failed on your developer workstation.

ronlieb:
AMD General

kasaurov:
1. The problem manifests itself only in builds outside of the manylinux container, on a Ubuntu system with a system-wide /opt/rocm
2. @lamb-j has already proposed a simple fix in `llvm-project` repo https://github.com/ROCm/llvm-project/pull/3020

PR #5223, opened 2026-05-12
[ci] Skip native packages and PyTorch builds for ASAN variant

geomin12:
Previously in Multi Arch CI ASAN (https://github.com/ROCm/TheRock/actions/runs/25710576013), pytorch and package builds fail as we have not worked on support for those. To keep green signal, we are removing those jobs until we provide support and have a green signal

working here on test: https://github.com/ROCm/TheRock/actions/runs/25811658440

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -143,11 +143,7 @@ jobs:
   build_native_deb_packages:
     needs: [build_multi_arch_stages]
     name: Build DEB Packages
-    if: >-
-      ${{
-        !failure() && !cancelled() &&
-        fromJSON(inputs.build_config).build_native_linux == true
-      }}
+    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_native_linux == true }}
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
     with:
       dist_amdgpu_families: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
@@ -160,11 +156,7 @@ jobs:
   build_native_rpm_packages:
     needs: [build_multi_arch_stages]
     name: Build RPM Packages
-    if: >-
-      ${{
-        !failure() && !cancelled() &&
-        fromJSON(inputs.build_config).build_native_linux == true
-      }}
+    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_native_linux == true }}
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
     with:
       dist_amdgpu_families: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
@@ -177,11 +169,7 @@ jobs:
   test_native_packages_ubuntu2404:
     needs: [build_native_deb_packages]
     name: Test DEB Install - ubuntu2404
-    if: >-
-      ${{
-        !failure() && !cancelled() &&
-        fromJSON(inputs.build_config).build_native_linux == true
-      }}
+    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_native_linux == true }}
     uses: ./.github/workflows/test_native_linux_packages_install.yml
     with:
       package_install_url: ${{ needs.build_native_deb_packages.outputs.package_repository_url }}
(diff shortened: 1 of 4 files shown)


.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -140,10 +140,16 @@ jobs:
...
     needs: [build_multi_arch_stages]
     name: Build DEB Packages
-    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).expect_failure == false }}
+    if: >-
+      ${{
+        !failure() && !cancelled() &&
+        fromJSON(inputs.build_config).expect_failure == false &&
+        fromJSON(inputs.build_config).build_variant_suffix != 'asan'
+      }}
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
```
ScottTodd: The designed way for this plumbing is via configure_multi_arch_ci.py, see "build-native-linux" in https://github.com/ROCm/TheRock/blob/b296f29cddce6316d52ba0b2b427ee152bcbebf5/build_tools/github_actions/configure_multi_arch_ci.py#L19-L25

If we complete the plumbing there these jobs can be opt-in/out depending on other details on a PR or other CI/CD run. I'd rather not code anything like "asan can't do this or that" directly into the workflow files.

ScottTodd: For example, if a PR _only_ modifies native Linux packaging code, we should be able to use prebuilt artifacts for the whole rocm build, skip building rocm python, and choose to build rocm native linux packages.

geomin12:
re-running a test here: https://github.com/ROCm/TheRock/actions/runs/25811658440

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -270,7 +290,11 @@ jobs:
   build_pytorch_wheel_fat:
     needs: [build_python_packages]
     name: Build PyTorch (fat + split)
-    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_pytorch == true }}
+    if: >-
+      ${{
+        !failure() && !cancelled() &&
+        fromJSON(inputs.build_config).build_pytorch == true
+      }}
```
ScottTodd: nit: Why the formatting change? I think the ~100 character lines are fine here (120+ would be pushing it though). I'd rather these workflows be easier to scan vertically to see the job graph, the individual conditions on each job can have some slightly long lines IMO.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -935,7 +927,10 @@ def _expand_build_config_for_platform(
         build_variant_suffix=suffix,
         build_variant_cmake_preset=variant_config["build_variant_cmake_preset"],
         expect_failure=expect_failure,
-        build_pytorch=not expect_failure and not expect_pytorch_failure,
+        build_pytorch=(
+            not expect_failure and not expect_pytorch_failure and suffix != "asan"
+        ),
+        build_native_linux=(not expect_failure and suffix != "asan"),
```
ScottTodd: nit: sort "native linux" before "pytorch" to match the ordering in the workflow? (rocm packages before framework packages built on top of rocm)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd, line 25: you can remove the "(future)" here now
```suggestion
The CI pipeline is a DAG of job groups:

   build-rocm → test-rocm
              → build-rocm-python → build-pytorch → test-pytorch
                                  → build-jax     → test-jax (future)
              → build-native-linux   → test-native-linux
              → build-native-windows → test-native-windows (future)
```

ScottTodd, line 652: Hmmm... technically we should keep the build decisions up here in `decide_jobs` instead of computing down at the bottom of the file with 

```python
    return BuildConfig(
...
        build_pytorch=(
            not expect_failure and not expect_pytorch_failure and suffix != "asan"
        ),
        build_native_linux=(not expect_failure and suffix != "asan"),
```

So here we'd add `build_native_linux=JobGroupDecision(action=JobAction.RUN)` by default, and set `build_pytorch` / `test_pytorch` / `build_native_linux` to `JobAction.SKIP` for ASan, then read those decisions in `_expand_build_config_for_platform`

We can defer that until later though. Right now it's just extra complexity for forward-looking features. I can also go through after this PR and refactor in that way (I see that some function args are missing plumbing to make that easy, then a whole bunch of unit tests will need updating)

.github/workflows/multi_arch_ci_linux.yml
geomin12: oh! this was to accomodate for additional flags, but it can be reverted

ScottTodd (approved):

PR #5229, opened 2026-05-13
[ci] Allow rocBLAS to keep its 6 shards on quick test runs

FIM43-Redeye:
## Summary

`fetch_test_configurations.py` unconditionally collapses `total_shards` to 1 when `test_type == "quick"`. The override is right for most components — their quick suites are quick — but it's incorrect for rocBLAS, where the "quick" CTest entry (`rocblas-test_quick_suite`) is a single non-shardable target whose gtest filter (`*quick*`) now matches enough cases to exceed the 600s CTest TIMEOUT budget when run unsharded.

Evidence: https://github.com/ROCm/TheRock/actions/runs/25753877460/job/75666293964 — `0% tests passed, 1 tests failed out of 1: rocblas-test_quick_suite (Timeout)` at exactly `600.29 sec`, on a job named `(shard 1/1)`. The rocBLAS config (line 83-84) already documents an assumption of 6-way sharding (`24h / 6 shards = 4h per shard`), so the override silently breaks that intent for the quick path.

## Fix

Add an opt-in flag `shard_quick_tests: True` to the component config. When set, the component keeps its `total_shards_dict` value even for `test_type=quick`; when unset (the existing default), behavior is unchanged. rocBLAS opts in.

## Why opt-in rather than removing the override

The override was added deliberately in 2025-10 (then for "smoke" tests, renamed when test types were unified) to save CI machine time on fast suites. That assumption is still correct for most components; only rocBLAS has outgrown a single-shard budget so far. An opt-in keeps everyone else's behavior identical and minimizes blast radius.

## Changes

| File | Edit |
|---|---|
| `fetch_test_configurations.py` | `shard_quick_tests: True` on rocblas; gate override on the flag |
| `tests/fetch_test_configurations_test.py` | existing `test_quick_test_forces_single_shard` skips opted-in components; new `test_quick_test_respects_shard_quick_tests_opt_in` asserts rocblas keeps its 6 shards |

## Test plan

- [x] `python -m unittest build_tools.github_actions.tests.fetch_test_configurations_test -v` — 21/21 pass
- [x] `pre-commit run` clean (black formatted)
- [ ] CI on this PR exercises `quick` test_type on rocBLAS; expected outcome is six parallel `Test rocblas (shard N/6)` jobs each finishing well under 600s, where the previous behavior had one `(shard 1/1)` job hitting the 600s timeout

Generated using Claude Code.

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -89,6 +89,13 @@ test_matrix = {
             "linux": 6,
             "windows": 6,
         },
+        # rocBLAS's "quick" suite is a single CTest target (rocblas-test_quick_suite)
+        # with a 600s CTest TIMEOUT. The unsharded gtest invocation runs every
+        # *quick* test case in one process and now exceeds 600s on CI, so the
+        # one-shard default for test_type=quick must be overridden here. With
+        # 6 shards, each gtest invocation handles ~1/6 of the cases via
+        # GTEST_TOTAL_SHARDS/GTEST_SHARD_INDEX and finishes well under 600s.
+        "shard_quick_tests": True,
     },
     "rocroller": {
         "job_name": "rocroller",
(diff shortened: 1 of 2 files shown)


FIM43-Redeye:
@geomin12 available when convenient, would appreciate eyes — does this look ready to you? Adds a `shard_quick_tests: True` opt-in so rocBLAS's quick suite (single CTest target, 600s TIMEOUT) keeps its 6-shard fanout instead of collapsing to one process. Same per-component sharding area as your #4890.

geomin12 (commented):
quick tests should not need shards, see https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md !

Quick tests are sanity checks, they shouldn't take 60 mins

FIM43-Redeye:
Fair point, the contract is clear. The bloat lives in rocBLAS's own `test_categories.yaml` (defines quick as 10-min "smoke" with 506 entries), so the fix belongs upstream. Closing and filing against rocm-libraries.

PR #5243, opened 2026-05-13
docs/rfcs/packaging: Refine debugger related packages

lancesix:
## Motivation

The debugging tools are currently all grouped under the "amdrocm-debugger" package.  This does not match how the various tools are used and deployed.

Some can decide to include the rocr-debug-agent with production workloads, so having a minimal package for this use-case would be preferred.  Pulling the full debugger un-necessarily increases the size of the dependencies.

Development or post-mortem analysis scenario will use rocgdb, and probably won't need the debug agent.

Both the rocgdb and debug agent have a shared dependency to dbgapi, so it would make sense to have 3 distinct packages:
- amdrocm-dbgapi
- amdrocm-gdb (or amdrocm-rocgdb, depends on amdrocm-dbgapi)
- amdrocm-debug-agent (depends on amdrocm-dbgapi)

This patch proposes to change the policy outlined in RFC0009-OS-Packaging-Requirements.md to use those 3 packages.


LiamfBerry (approved):
Justification makes sense to me, approved on my end.

PR #5244, opened 2026-05-13
[ci] Add more test dependency for test subprojects

geomin12:
After reaching out to component teams in regards to what components are dependent, we have determined the best subcomponents to test for each project.

Projects with `TEST_SUBPROJECTS` only need to test itself


jayhawk-commits (approved):
No blocking findings from me. My recommendation in the future for test dependency updates is to get the reviews done by the project teams confirming those dependencies, so we have record for reference.

PR #5245, opened 2026-05-14
Remove two patches from compiler/amd-staging branch

ronlieb:
Remove two patches from compiler/amd-staging
patches/amd-mainline/llvm-project/0002-hipcc-fix-default-include-path-on-Windows-and-adapt-.patch
patches/amd-mainline/llvm-project/0006-Rework-constructHipPath-so-HIP_PATH-env-var-is-lower.patch

landing this first to allow equivalent patches to land in llvm-project branch amd-staging


ScottTodd (commented):
Why do you need this PR? Can you sequence the changes differently, so the patches are deleted at the same time as the changes land? See https://github.com/ROCm/TheRock/tree/main/patches#using-patches-from-rocm-libraries-rocm-systems-and-other-repositories

ronlieb:
> Why do you need this PR? Can you sequence the changes differently, so the patches are deleted at the same time as the changes land? See https://github.com/ROCm/TheRock/tree/main/patches#using-patches-from-rocm-libraries-rocm-systems-and-other-repositories

we are landing the patches in amd-staging branch of llvm-project, and removing the two patches is easier to land in compiler/amd-staging branch first. then when we land the two patches in amd-staging it will not conrflict with the patch apply.   

our normal flow to land content in llvm-project:amd-staging  is simply open a PR and let the compiler/amd-staging version of the rock validate it.

ScottTodd (approved):

skganesan008:
@kasaurov, please help review this so we can land it.

kasaurov (approved):
LGTM as preparation for merge from upstream

PR #5248, opened 2026-05-14
Add platform override for stage log uploads

jayhawk-commits:
## Summary

- Adds an optional `--platform` argument to `post_stage_upload.py`.
- Preserves the existing default behavior: when `--platform` is omitted, the script uses `platform.system().lower()`.

## Motivation

- Stage logs are uploaded under `{run_id}-{platform}/logs/...`, so workflows sometimes need to choose the output platform namespace explicitly.
- This supports #4541, where the WSL ROCDXG job starts on a Windows runner but produces Linux/WSL artifacts that should share the Linux output namespace with their logs.


build_tools/github_actions/tests/post_stage_upload_test.py
ScottTodd, line 240: `platform="windows"` is already covered here. I don't think new unit tests for the one line `default=platform.system().lower(),` are necessary.

ScottTodd (approved):
The new unit tests aren't pulling their weight so I'd leave them off. Otherwise LGTM

build_tools/github_actions/tests/post_stage_upload_test.py
jayhawk-commits: Unit tests removed. I will wait for CI results to go through overnight.

jayhawk-commits:
Test failures are unrelated. Will merge.

Issue #5076, opened 2026-05-06
[Issue] DEB packaging fails: `dh_dwz/dwz cannot handle DWARF v5 .debug_str_offsets (amdrocm-amdsmi)`


amd-chiranjeevi:
## Description:
The Debian packaging step fails when dpkg-buildpackage runs dh_dwz, which invokes dwz. dwz errors out because it does not recognize the DWARF v5 debug section .debug_str_offsets. 

## Context
- Workflow: Multi-Arch CI ASAN
- Platform: Linux

## Full log:
- https://github.com/ROCm/TheRock/actions/runs/25414061404/job/74562173461

## Error log:
```
2026-05-06T08:21:48.3618878Z install -d /home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506/debian/amdrocm-amdsmi7.13/opt/rocm/core-7.13/share/doc/amdrocm-amdsmi7.13
2026-05-06T08:21:48.3644906Z if [ -f /home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506/debian/changelog ]; then \
2026-05-06T08:21:48.3646042Z 	install -m 644 /home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506/debian/changelog \
2026-05-06T08:21:48.3647434Z 		/home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506/debian/amdrocm-amdsmi7.13/opt/rocm/core-7.13/share/doc/amdrocm-amdsmi7.13/changelog.Debian; \
2026-05-06T08:21:48.3649148Z 	gzip -9 -f /home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506/debian/amdrocm-amdsmi7.13/opt/rocm/core-7.13/share/doc/amdrocm-amdsmi7.13/changelog.Debian; \
2026-05-06T08:21:48.3650171Z fi
2026-05-06T08:21:48.3685689Z make[1]: Leaving directory '/home/runner/_work/TheRock/TheRock/output/packages/deb/amdrocm-amdsmi7.13.0~dev20260506'
2026-05-06T08:21:48.3719963Z    dh_installsystemduser -O--no-parallel
[...]

PR #5250, opened 2026-05-14
fix(ci): Use DWARF4 for ASAN builds to fix dwz compatibility

raramakr:
dwz version 0.14/0.15 (used in Debian packaging) doesn't support DWARF5
  debug info sections like .debug_str_offsets, causing dh_dwz to fail
  when building ASAN-enabled packages.

  This change adds -gdwarf-4 flag for ASAN/TSAN builds to ensure
  DWARF4 format is used:

  - CMakePresets.json: Add -gdwarf-4 to sanitizer compile flags
  - compiler/CMakeLists.txt: Pass THEROCK_SANITIZER to amd-llvm subproject
  - compiler/pre_hook_amd-llvm.cmake: Add -gdwarf-4 to RUNTIMES_CMAKE_ARGS for ASAN builds (compiler-rt/ASAN runtime)
  
Jira ID : https://github.com/ROCm/TheRock/issues/5076

Passed CI build : https://github.com/ROCm/TheRock/actions/runs/26191034841


nunnikri (approved):
Changes looks ok. Need to test few times and confirm its not increasing the asan build time

compiler/pre_hook_amd-llvm.cmake
```diff
@@ -39,6 +39,11 @@ else()
     # CONFIG mode.
     set(RUNTIMES_CMAKE_ARGS "-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON")
 
+    # Use DWARF4 for ASAN builds - compatibility with dwz < 0.15 which doesn't support DWARF5.
+    if(THEROCK_SANITIZER STREQUAL "ASAN" OR THEROCK_SANITIZER STREQUAL "HOST_ASAN")
+        string(APPEND RUNTIMES_CMAKE_ARGS ";-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS} -gdwarf-4;-DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS} -gdwarf-4")
+    endif()
```
ScottTodd: I'd rather we not unconditionally limit ourselves to old versions of software (dwz 0.14 was released 2021-03-8, over 5 years ago: https://sourceware.org/pipermail/dwz/2021q1/001066.html). Someone may want an ASan build that will be used apart from our debian packaging CI/CD where we currently have an older dwz version.

Also, could we update the dwz version? Is that under our control or something in the package manager ecosystem itself?

If keeping this, let's add more comments about what the requirements are and when we might be able to update.

compiler/pre_hook_amd-llvm.cmake
```diff
@@ -39,6 +39,11 @@ else()
     # CONFIG mode.
     set(RUNTIMES_CMAKE_ARGS "-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON")
 
+    # Use DWARF4 for ASAN builds - compatibility with dwz < 0.15 which doesn't support DWARF5.
+    if(THEROCK_SANITIZER STREQUAL "ASAN" OR THEROCK_SANITIZER STREQUAL "HOST_ASAN")
```
ScottTodd: If keeping this, should it look for TSan also? The changes in `CMakePresets.json` are for `ASAN`, `HOST_ASAN`, and `TSAN`

raramakr: Done

compiler/pre_hook_amd-llvm.cmake
raramakr: I tested with dwz 0.16 on Ubuntu 26.04 and it still fails with Unknown debugging section .debug_str_offsets. This appears to be an ongoing limitation in dwz's  DWARF5 support, not just an old-version issue.
Until dwz properly handles .debug_str_offsets and other DWARF5 sections, 
DWARF4 is the pragmatic choice for sanitizer builds that go through the packaging pipeline.

Updating dwz isn't within our control since it's a system package managed by the distro

Added more comments in the source code as well

geomin12 (commented):
gonna let build folks handle this

Issue #1781, opened 2025-10-14
Building with --preset linux-release-package has various issues


stellaraccident:
* Seems to push us over some capacity threshold that makes building unreliable.
* Possibly triggering some other observed issues in the build so not enabling in the broader pipeline refactor until after.

PR #5267, opened 2026-05-15
Add Windows debug PDB support.

### LINKED ISSUE #5004: [Feature]: TheRock Debug Symbol Support
# Summary
Add & unify debug symbol support in TheRock

### Operating System

Windows

### GPU

_No response_

### ROCm Component

_No response_

amd-nicknick:
## Motivation
Closes #5004

Enable release PDB generation and packaging for Windows builds when `THEROCK_SPLIT_DEBUG_INFO` is enabled. This gives Windows release artifacts the same practical split-debug support as Linux build-id debug files, while preserving ccache/sccache-friendly embedded debug information during compilation.

## Technical Details

- Enables `THEROCK_SPLIT_DEBUG_INFO` for the `windows-release` preset.
- Adds Windows PDB collection in therock_global_post_subproject.cmake for Windows MSVC/Clang toolchains.
  - Applies `/DEBUG:FULL` so linkers emit final PDBs.
  - Applies `/PDBALTPATH:<filename>` to avoid embedding stale absolute build-tree PDB paths.
  - Sets per-target `PDB_OUTPUT_DIRECTORY` under `therock-pdb/<target>/`.
- Adds therock_install_windows_pdb_files.cmake to copy generated linker PDBs into:
  ```text
  .debug/pdb/<subproject>/<target>/<file>.pdb
  ```
- Makes `cmake --install --strip` Linux-only for split-debug builds, since Windows PDBs are separate linker outputs rather than post-link stripped debug sections.
- Propagates `CMAKE_POLICY_DEFAULT_CMP0141=NEW` and `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded` into generated subproject toolchains so MSVC-ABI compilers, including Clang, use embedded CodeView debug info instead of compile-time PDBs.
- Extends `dbg` artifact defaults to include Windows PDBs:
  ```text
  .debug/pdb/**/*.pdb
  ```
- Prevents catch-all `run` artifacts from claiming debug-symbol trees before `dbg` artifacts can route them.
- Updates Windows release packaging to:
  - exclude `.debug/**` and `*.pdb` from the main dist archive,
  - create a separate symbols archive from `.debug`,
  - upload the symbols archive when present.
- Updates artifact documentation for Windows PDB debug-symbol routing.

## Test Plan

- Ran focused artifact routing tests:
  ```text
  python.exe -m unittest build_tools.tests.artifacts_test
  ```
- Validated the Windows PDB install helper parses:
  ```text
  cmake -P cmake/therock_install_windows_pdb_files.cmake
  ```
- Inspected a completed and verified:
  - PDBs were generated under component `therock-pdb/<target>/` directories.
  - Existing source PDBs were copied into component stage `.debug/pdb/...` trees.
  - Final dist PDBs were present only under `dist/rocm/.debug/pdb/...`.
  - Bootstrapped Clang/LLD subprojects had `/DEBUG:FULL`, `/PDBALTPATH`, and `-fuse-ld=lld-link` in generated build rules.

## Test Result

- Artifact tests passed:
  ```text
  Ran 24 tests in 0.090s

  OK
  ```
- `cmake -P cmake/therock_install_windows_pdb_files.cmake` completed successfully.
- `git diff --check` passed.
- Completed Windows build inspection found:
  ```text
  Final debug PDB count: 1329
  Final debug PDB total size: 17,683,247,104 bytes
  Malformed final layout count: 0
  PDBs in dist/rocm outside .debug/pdb: 0
  Source exists but stage missing: 0
  ```

https://github.com/ROCm/TheRock/actions/runs/25858256113 contains the correct artifacts and matching PDBs in dbg artifacts.
## Submission Checklist


amd-nicknick:
Looking through the current TheRock implementation on debug symbols, Linux is weird that:
```
CMAKE_BUILD_TYPE=Release
THEROCK_MINIMAL_DEBUG_INFO=ON
THEROCK_SPLIT_DEBUG_INFO=ON
```
Still yields no symbol output, since it only applies to Debug / RelWithDebInfo
https://github.com/ROCm/TheRock/blob/cda2792eef5fcea0b2d7f718b2b4d21be6e3d4ca/cmake/therock_subproject.cmake#L1653-L1656

This means there's no conceivable way to build Release (with all optimizations) + debug symbols for Linux, whereas on current Windows impl it's always the case when `THEROCK_SPLIT_DEBUG_INFO=ON` is set.


I propose that rename the flag `THEROCK_MINIMAL_DEBUG_INFO` to `THEROCK_DEBUG_INFO_LEVEL`, this variable has different behavior on Linux / Windows:
* Linux
    * It populates `-gLEVEL` with the selected level **to all build types**. 
    * Eg.  `THEROCK_DEBUG_INFO_LEVEL=0`->`-g0` (off)
    `THEROCK_DEBUG_INFO_LEVEL=1`->`-g1` (minimal)
    `THEROCK_DEBUG_INFO_LEVEL=2`->`-g2` (full)
* Windows
    * Sets `/DEBUG:FULL` flag regardless of `THEROCK_DEBUG_INFO_LEVEL` value, as long as it is not `0`.
    * If `THEROCK_DEBUG_INFO_LEVEL=1` (Let's call this "minimal" equivalent for Windows), we perform additional PDB stripping.
    * If `THEROCK_DEBUG_INFO_LEVEL=2`, we leave the PDB alone without stripping.

This will make `THEROCK_DEBUG_INFO_LEVEL` turn on debug symbols for all OS, regardless of `CMAKE_BUILD_TYPE`, leaving build type to specify for optimization purpose.

For `THEROCK_SPLIT_DEBUG_INFO`, reduce its current functionality:
* Linux
    * Keep current behavior, split out the debug information from binaries without affecting if the binaries contain debug information at all.
* Windows
    * Add linker flag `/PDBALTPATH` to remove PDB full path from binaries.
    * Copy PDB to .debug.

A note: this will mean anyone expecting a reproducible build on Windows + symbols enabled (either with `THEROCK_DEBUG_INFO_LEVEL` or `RelWithDebInfo`) **will need to set `THEROCK_SPLIT_DEBUG_INFO` to achieve this** (since the alt path is only stripped with split ON), we could address this in documents.

@stellaraccident, @ScottTodd, any feedbacks / suggestions are greatly appreciated. I'm not quite familiar on why it was designed like this for Linux, and why Multi-Arch CI is not using CMakePreset, and why it's not producing split dbg info now.

ScottTodd:
Sorry for the slow response here - my time is pretty limited to review draft PRs (in fact, I filter them all out from my review queue entirely).

A few quick points:

* If you're proposing substantial changes to the build and packaging, an RFC (https://github.com/ROCm/TheRock/tree/main/docs/rfcs) may be a better place for design discussion
* Linux is not using presets due to https://github.com/ROCm/TheRock/issues/1781 (an old issue, which may be fixed now), which also overlaps with some other necessary work being addressed in https://github.com/ROCm/TheRock/pull/5191 or https://github.com/ROCm/TheRock/pull/5192

We just had a user report on Discord at https://discord.com/channels/1239631572886491286/1375223676659564546/1506046515347587163 a 10GB binary size increase in our release packages due to https://github.com/ROCm/TheRock/pull/4419#discussion_r3262592363 and maintaining healthy releases is my first priority ahead of feature work like this.

amd-nicknick:
> * If you're proposing substantial changes to the build and packaging, an RFC (https://github.com/ROCm/TheRock/tree/main/docs/rfcs) may be a better place for design discussion

Yeah let me do that. This has grown a bit after I initially thought about it. Let me draft an RFC and initiate the discussion.

> We just had a user report on Discord at https://discord.com/channels/1239631572886491286/1375223676659564546/1506046515347587163 a 10GB binary size increase in our release packages due to [#4419 (comment)](https://github.com/ROCm/TheRock/pull/4419#discussion_r3262592363) and maintaining healthy releases is my first priority ahead of feature work like this.

Let's address that one on #4419 & #4519.

amd-nicknick:
Closing this first to de-clutter,
1. Get the artifact size discussion addressed.
2. Raise RFC with full design

PR #5277, opened 2026-05-15
[ci] fixing pytorch build wheels for forked repos

geomin12:
PyTorch build wheels was not working for forked repos. Fixing it with changes below:

Mount /runner/config to /home/awsconfig/ inside the container and set AWS_SHARED_CREDENTIALS_FILE to enable S3 uploads from within the container.


build_tools/github_actions/tests/configure_ci_path_filters_test.py
```diff
@@ -57,8 +57,11 @@ def test_run_ci_if_source_file_and_unrelated_workflow_file_edited(self):
         self.assertTrue(run_ci)
 
     def test_ci_workflow_filenames_cover_all_transitive_uses(self):
-        """_GITHUB_WORKFLOWS_CI_FILENAMES must exactly match the set of
-        workflows transitively called by ci.yml and multi_arch_ci.yml.
+        """_GITHUB_WORKFLOWS_CI_FILENAMES must include all workflows
+        transitively called by ci.yml and multi_arch_ci.yml.
+
+        Additional release workflows may be included to ensure CI runs
+        when those workflows are modified.
```
ScottTodd: None of the changes to `build_tools/github_actions/` look correct here?

If a workflow is not referenced by ci.yml/multi_arch_ci.yml there is no reason to run those workflows when only modifying that workflow in a PR.

ScottTodd (changes requested):

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -117,7 +117,9 @@ jobs:
       contents: read
     container:
       image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:702a5133851e6d1daf1207d2c9fbb01c2667914a5b6dc5a01faeb3ce66ea6421
+      options: -v /runner/config:/home/awsconfig/
     env:
+      AWS_SHARED_CREDENTIALS_FILE: /home/awsconfig/credentials.ini
```
ScottTodd, line 122: Do you want to edit https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels_ci.yml also? That is used by CI.

ScottTodd (approved):

PR #4410, opened 2026-04-08
Adding TBB to sysdeps

ajanicijamd:
## Motivation

- Provide **Threading Building Blocks (TBB)** as an **optional Linux sysdep** under `lib/rocm_sysdeps`, consistent with other bundled third-party sysdeps.
- **rocprofiler-systems** previously built its own TBB (`ROCPROFSYS_BUILD_TBB=ON`). That can duplicate work and does not align with TheRock’s portable layout. This PR **turns that off** and consumes **`therock-tbb`** via `${THEROCK_BUNDLED_TBB}`.
- Aligns with tracking work: **AIPROFSYST-22**; see also [TheRock#1902](https://github.com/ROCm/TheRock/issues/1902), [TheRock#1905](https://github.com/ROCm/TheRock/issues/1905).
- Note:  Also needed by hipFile library:  ROCM-1261, AIHIPFILE-27

## Technical Details

- **Topology:** New artifact `sysdeps-tbb` in `BUILD_TOPOLOGY.toml` (optional third-party sysdep pattern). `rocprofiler-systems` **depends on** `sysdeps-tbb` so `ROCPROFSYS` implicitly requires **`THEROCK_ENABLE_SYSDEPS_TBB`** when features are finalized.
- **oneTBB build:** `third-party/sysdeps/linux/tbb/` fetches **oneTBB 2022.3.0** (mirrored tarball + `URL_HASH`). Custom install uses Ninja, `TBB_TEST=OFF`, install under `lib/rocm_sysdeps`.
- **Namespacing:** `patch_source.sh` rewrites upstream CMake so shared libraries use **`rocm_sysdeps_*` `OUTPUT_NAME`** to avoid clashing with system TBB.
- **Profiler integration:** `profiler/CMakeLists.txt` — `ROCPROFSYS_BUILD_TBB=OFF`, `${THEROCK_BUNDLED_TBB}` in `RUNTIME_DEPS`, **`INSTALL_RPATH_DIRS`** include `lib/rocm_sysdeps/lib`. **`post_hook_rocprofiler-systems.cmake`** sets **`THEROCK_INSTALL_RPATH_ORIGIN`** for Dyninst-related targets so `$ORIGIN/../rocm_sysdeps/lib` resolves correctly from `lib/rocprofiler-systems`.
[...]

PR #5287, opened 2026-05-15
[rocprofiler-systems] Patch library files to set the correct RPATH

ajanicijamd:
## Motivation

Patching rocprofiler-systems libraries' RPATH.

## Technical Details

The issue in ROCM-24678 is that `rocprof-sys-instrument` cannot be run in the `dist` directory because the loader cannot find `libomp.so`.
- `rocprof-sys-instrument` depends on `libinstructionAPI.so`
- `libinstructionAPI.so` depends on `libomp.so`

The root cause is that RPATH in `rocprof-sys-instrument` is correctly pointing to
the location of `libinstructionAPI.so`, but RPATH in that library is wrong, so loader
can't find `libomp.so`.

This PR patches all libraries in `dist/rocm/lib/rocprofiler-systems` so that their RPATH
contains entries
`$ORIGIN/..:$ORIGIN:$ORIGIN/../llvm/lib:$ORIGIN/../rocm_sysdeps/lib`.

## Test Plan

- Build TheRock.
- In the directory `/therock/output/build/dist/rocm/lib/rocprofiler-systems` in the manylinux container, inspect each `.so` file with `patchelf`, for example:
  `patchelf --print-rpath libinstructionAPI.so`

## Test Result

- The output of the `patchelf` in the example above is:
  `$ORIGIN/..:$ORIGIN:$ORIGIN/../llvm/lib:$ORIGIN/../rocm_sysdeps/lib`

## Submission Checklist

### DIFF
--- a/profiler/post_hook_rocprofiler-systems.cmake
+++ b/profiler/post_hook_rocprofiler-systems.cmake
@@ -1,0 +1,29 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Dyninst and rocprofiler-systems internal libraries live in lib/rocprofiler-systems.
+# Set the RPATH origin per-target so that (for example) lib/rocm_sysdeps/lib resolves as
+# $ORIGIN/../rocm_sysdeps/lib rather than $ORIGIN/rocm_sysdeps/lib.
+set(_rocprofsys_lib_targets
+  common
+  dynElf
+  dynDwarf
+  dyninstAPI
+  dyninstAPI_RT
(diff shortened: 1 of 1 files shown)


dgaliffiAMD (approved):

dgaliffiAMD:
These are the post hook changes from https://github.com/ROCm/TheRock/pull/4410, pulled out into their own pull request.

marbre:
Let's rebase or merge main into this to make sure the packaging failures are unrelated.

jbonnell-amd (approved):

PR #5288, opened 2026-05-15
[ci] Enabling ASAN test runners

geomin12:
As we have capacity from conductor to enable machines, we are running ASAN tests now.

These tests will only run on nightlies/workflow_dispatch and will not run on post-submit, presubmit.

Test runs here: https://github.com/ROCm/TheRock/actions/runs/26608276736/job/78408321807#step:3:51 (cancelled to save resources)


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -875,17 +875,25 @@ def _expand_build_config_for_platform(
...
+                if can_run_asan:
+                    print(
+                        f"  {family_name}: no ASAN sandbox runner available, "
+                        f"disabling tests"
+                    )
+                else:
+                    print(
+                        f"  {family_name}: ASAN tests only run on nightly/workflow_dispatch, "
+                        f"disabling tests"
+                    )
```
ScottTodd: We should move more of this logic into `decide_jobs()`, here: https://github.com/ROCm/TheRock/blob/37041cf203938471004cee945ae5691ce73c4e73/build_tools/github_actions/configure_multi_arch_ci.py#L641-L650

This `_expand_build_config_for_platform` function should just transforming data, not making decisions. That way we can test the decision making and the data transformation code separately. If this function does both then reasoning about the behavior (and testing it) gets harder.

ScottTodd: Specifically what I had in mind was 

```python
if build_variant == "asan":
  test_rocm = TestRocmDecision(
    action=JobAction.SKIP
    ...
```

Choosing whether or run tests or skip them based on the configuration should happen up there, while the code down here would then just see "oh, test should be skipped - okay!" and not try to make that decision on its own.

I see from the code here that the decision is actually based on whether a given GPU has "sandbox runners", so the aggregate "should tests run at all" decision is fuzzier...

I guess another data transformation we could make would be to explicitly choose which GPU targets to build and test for each configuration:
```
gfx1100
  ci: build, test
  ci_asan: build

gfx942
  ci: build, test
  ci_asan: build, test

etc.
```
right now this hides the decision a bit under "if there's a sandbox runner, test on asan".

ScottTodd (commented):
I'm still indecisive about the configure code in use here. Worth a fresh look on Monday.

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: fair point, we can re-review on monday indee!d! that sounds good though i can incorporate those

ScottTodd: Here's an idea: move the decision for which targets to build by default for each "build variant" ('ci', 'release', 'asan', 'tsan', etc.) further into amdgpu_family_matrix.py (or a similar file -- I still want to remove the concept of a "gpu family" from the CI at this level).

So this code: https://github.com/ROCm/TheRock/blob/37041cf203938471004cee945ae5691ce73c4e73/build_tools/github_actions/configure_multi_arch_ci.py#L961-L984

would ask "I'm a CI ASan workflow run, which GPU targets should I build and test for?" and it would get back e.g. "just gfx942 on a sandbox runner". Then another run through the file would ask "I'm a release workflow run, which should _I_ build and test for?" and the response would be "this full list of targets that build successfully and the test runners that we currently have for each of them".

Right now the code in this PR has some deep nesting:
```python
        if build_variant == "asan":
            if use_sandbox_runners and "test-runs-on-sandbox" in platform_info:
                ...
            else:
                if use_sandbox_runners:
                    ...
                else:
                    ...
```
and the logic for what is actually going to get selected is non-obvious - do we build gfx1100 and gfx1200 even if we aren't going to test it? Why? Should we just limit to building gfx942 since only that has test runners? etc. If we extract just the explicit list of what we want to build and test and load from that list in an obvious way then the CI system will be easier to understand for contributors and maintainers.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -643,11 +647,23 @@ def decide_jobs(
...
+    # TODO(#3433): Remove once ASAN tests pass on production runners.
+    use_sandbox_runners = ci_inputs.is_schedule or ci_inputs.is_workflow_dispatch
+    is_sandbox_variant = ci_inputs.build_variant == "asan"
+    if is_sandbox_variant and not use_sandbox_runners:
+        test_rocm = TestRocmDecision(
+            action=JobAction.SKIP,
+            test_type=test_type,
+            test_type_reason="sandbox variant on non-nightly trigger",
+            use_sandbox_runners=False,
+        )
```
ScottTodd: This logic is far from obvious. What is a "sandbox runner" and how does it differ from a "production runner"? Why would some jobs use them but not others? The layers of indirection here are not helping us.

The PR description is also missing critical details here.

ScottTodd (changes requested):
I'm not able to follow the changes here. Needs more explanation + a first level review from someone else on the team before me.

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: i'll need to re-work! thanks for the review, i'll add more details tmrw

ndhillon07 (approved):
as long previous comments are resolved, good on my end. Label might have to be updated later

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -649,6 +649,18 @@ def decide_jobs(
...
+    # During ASAN testing, machines would crash resulting in impact for production CI jobs.
+    # As we get ASAN tests to green, we use a separate conductor pool for ASAN testing
+    if ci_inputs.build_variant == "asan":
+        # Only run ASAN tests on scheduled or workfow dispatch runs, to avoid impact on submodule bumps
+        if not (ci_inputs.is_schedule or ci_inputs.is_workflow_dispatch):
+            test_rocm = TestRocmDecision(
+                action=JobAction.SKIP,
+                test_type=test_type,
+                test_type_reason="ASAN tests skipped due to non-nightly trigger",
+            )
```
ScottTodd, line 664: This JobAction is not yet plumbed through, so this does nothing. Do you have a workflow run showing otherwise?

Review details + some implementation options to consider: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_5288.md#blocking-testrocmdecisionactionskip-does-not-suppress-downstream-test-jobs

ScottTodd (changes requested):

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Otherwise, the code looks clear enough now. Sorry that this was partially implemented - the multi-arch CI configure script has a few forward looking features to ensure that we could build what we knew we'd need without needing to make deeper refactoring changes later.

geomin12: oops good call, i thought that was a perfect place to put it but I see it isn't plumbed! no problem, we can def do some workaround until that plumbing goes into place!

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -899,6 +948,46 @@ def test_no_test_runner_label_uses_default(self):
...
+        result = cm.expand_build_configs(
+            targets=targets,
+            ci_inputs=self._inputs(event_name="schedule", build_variant="release"),
+            test_type="quick",
+        )
+        self.assertIsNotNone(result.linux)
+        entry = result.linux.per_family_info[0]
+        # Release tests should use normal runner, not sandbox
+        self.assertNotEqual(entry["test-runs-on"], "")
+        self.assertNotEqual(entry["test-runs-on"], "linux-mi325-gpu-rocm-cpu-sandbox")
```
ScottTodd: This test case overlaps with existing test cases - looks like all it's really adding is the `assertNotEqual(...-sandbox)`. I'd merge it into others instead of adding a new test (coding agents love to just append new tests instead of editing, gotta be careful with that).

Code coverage reports should also show overlap too I think.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -434,6 +434,55 @@ def test_build_rocm_stage_partitioning(self):
...
+        )
+        self.assertEqual(result.test_rocm.action, cm.JobAction.RUN)
+
+    def test_release_pr_runs_tests(self):
+        """Release builds on PR should run tests (not skip)."""
+        result = cm.decide_jobs(
+            self._inputs(event_name="pull_request", build_variant="release"),
+            git_context=cm.GitContext(),
+        )
+        self.assertEqual(result.test_rocm.action, cm.JobAction.RUN)
```
ScottTodd: This overlaps with other existing test cases. Common inputs -> check all outputs, not one test case per checked output.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -434,6 +434,55 @@ def test_build_rocm_stage_partitioning(self):
...
+        """ASAN builds on PR should skip tests entirely."""
+        result = cm.decide_jobs(
+            self._inputs(event_name="pull_request", build_variant="asan"),
+            git_context=cm.GitContext(),
+        )
+        self.assertEqual(result.test_rocm.action, cm.JobAction.SKIP)
+        self.assertEqual(
+            result.test_rocm.test_type_reason,
+            "ASAN tests skipped due to non-nightly trigger",
+        )
```
ScottTodd: Change detector test (just checks that one string exactly matches another) - remove or simplify. Generally, strings are user/developer-facing and not part of the API contract that is worth unit testing. Structured Exception subclasses are for code and _are_ worth testing.

See other test cases for example: https://github.com/ROCm/TheRock/blob/1f9923535be30c073b2f5d7152ccd8a3cddac367/build_tools/github_actions/tests/configure_multi_arch_ci_test.py#L364

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -434,6 +434,55 @@ def test_build_rocm_stage_partitioning(self):
...
+        """ASAN builds on push should skip tests entirely."""
+        result = cm.decide_jobs(
+            self._inputs(event_name="push", build_variant="asan"),
+            git_context=cm.GitContext(),
+        )
+        self.assertEqual(result.test_rocm.action, cm.JobAction.SKIP)
+        self.assertEqual(
+            result.test_rocm.test_type_reason,
+            "ASAN tests skipped due to non-nightly trigger",
+        )
```
ScottTodd: Change detector test

ScottTodd (commented):
Logic looks okay for now. Tests are too verbose / redundant.

ScottTodd (approved):

PR #5303, opened 2026-05-18
Add flydsl to S3 dependency mirror and index allow list

subodh-dubey-amd:
## Summary

- Add `flydsl` to `PACKAGES_PER_PROJECT` in `update_dependencies.py` (under the `rocm` project) so the S3 mirror script pulls flydsl wheels from PyPI
- Add `flydsl` to `PACKAGE_ALLOW_LIST` in `manage.py` so flydsl appears on the PEP 503 index pages

## Context

Resolves the TheRock-side infrastructure gap for [**ROCM-23479**](https://amd-hub.atlassian.net/browse/ROCM-23479)  ("TheRock Pytorch 3.13 index is missing flydsl, blocking model to run").

`flydsl` is a dependency of `amd-aiter`, which is pulled in by `flash-attention` during Docker image builds. It was completely absent from TheRock's S3 package index, causing `pip install flydsl` to fail with "No matching distribution found" when using the nightlies index.

The upstream `cp313` wheel gap was separately resolved in [ROCm/FlyDSL#440](https://github.com/ROCm/FlyDSL/pull/440). With both this PR and that fix, `flydsl` (including cp313) is now available on all nightlies index prefixes.

## Test plan

- [x] Dry-run of `update_dependencies.py --package rocm --dry-run` confirms flydsl wheels are picked up
- [x] Wheels (cp310/cp311/cp312/cp313/cp314) uploaded and index pages regenerated for both dev and nightly S3 buckets
- [x] Verified live: https://rocm.nightlies.amd.com/v2-staging/gfx1151/flydsl/ serves cp313 wheel


marbre (changes requested):
I am not convinced that this should go in. Currently our indexes only mirror dependencies which are required for those packages published via the index. aiter is not yet part of those packages. Furthermore, looking at https://github.com/ROCm/aiter/blob/main/requirements.txt it seems that even more deps are required. I'd suggest to instead install those from PyPI or use `--extra-index-url` when pointing at our index. Keeping track of the dependency chain for packages we're not publishing via the indexes seems not feasible.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -64,6 +64,7 @@
     "typing-extensions": {"versions": ["latest"], "project": "torch"},
     "rocm-bootstrap": {"versions": ["latest"], "project": "torch"},
     "setuptools": {"versions": ["81.0.0"], "project": "rocm"},
+    "flydsl": {"versions": ["latest"], "project": "rocm"},
```
ScottTodd, line 67: What **specifically** failed? We're not going to provide arbitrary additional packages in our index - it's either part of our support area or it isn't, and this package has not shown up as required before.

ScottTodd (changes requested):
+1 to the other comment.

PR #5316, opened 2026-05-18
Pin ROCm devel packages for full PyTorch tests

ethanwee1:
Test run: https://github.com/ROCm/TheRock/actions/runs/26044789530

Result: failed. Several gfx94X shards were blocked while installing packages because `rocm.nightlies.amd.com` could not be resolved (`Temporary failure in name resolution`). The install command shape is correct, and later shards that reached the package index got past virtualenv setup.

Likely blocker/fix: https://github.com/ROCm/TheRock/pull/5324 adds a DNS fallback for `rocm.nightlies.amd.com` to the S3 bucket and appears to target the same runner/index failure mode.

Old-version validation run: https://github.com/ROCm/TheRock/actions/runs/27284482421

Dispatched with `torch_version=2.13.0a0+rocm7.14.0a20260518` to validate that `rocm[devel]` does not float to the latest nightly. Cancelled after the relevant setup validation completed. Evidence from https://github.com/ROCm/TheRock/actions/runs/27284482421/job/80588048888:

```text
pip install --index-url=https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu torch==2.13.0a0+rocm7.14.0a20260518 'rocm[devel]'
Collecting rocm==7.14.0a20260518 (from rocm[libraries]==7.14.0a20260518->torch==2.13.0a0+rocm7.14.0a20260518)
Collecting rocm-sdk-devel==7.14.0a20260518 (from rocm[devel])
Successfully installed ... rocm-7.14.0a20260518 rocm-sdk-core-7.14.0a20260518 rocm-sdk-devel-7.14.0a20260518 rocm-sdk-libraries-gfx94X-dcgpu-7.14.0a20260518 ... torch-2.13.0a0+rocm7.14.0a20260518
```

### DIFF
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -225,14 +225,21 @@ jobs:
               --activate-in-future-github-actions-steps
           fi
 
+      - name: Resolve ROCm package version
+        id: rocm_package
+        run: |
+          ROCM_PACKAGE_VERSION="$(python -c 'import importlib.metadata; print(importlib.metadata.version("rocm"))')"
+          echo "rocm_package_version=${ROCM_PACKAGE_VERSION}" >> "${GITHUB_OUTPUT}"
+          echo "Resolved ROCm package version: ${ROCM_PACKAGE_VERSION}"
+
       - name: Install ROCm devel packages
         timeout-minutes: 15
         run: |
           if [ "${{ inputs.multi_arch }}" = "true" ]; then
-            pip install "rocm[devel,${{ steps.expand.outputs.device_extras }}]" \
+            python -m pip install "rocm[devel,${{ steps.expand.outputs.device_extras }}]==${{ steps.rocm_package.outputs.rocm_package_version }}" \
               --index-url=${{ inputs.package_index_url }}
           else
-            pip install "rocm[devel]" \
+            python -m pip install "rocm[devel]==${{ steps.rocm_package.outputs.rocm_package_version }}" \
               --index-url=${{ inputs.package_index_url }}/${{ inputs.amdgpu_family }}
           fi
 
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -214,33 +214,15 @@ jobs:
         run: |
           if [ "${{ inputs.multi_arch }}" = "true" ]; then
             python build_tools/setup_venv.py ${VENV_DIR} \
-              --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }}" \
+              --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }} rocm[devel,${{ steps.expand.outputs.device_extras }}]" \
               --index-url=${{ inputs.package_index_url }} \
               --activate-in-future-github-actions-steps
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -214,28 +214,17 @@ jobs:
         run: |
           if [ "${{ inputs.multi_arch }}" = "true" ]; then
             python build_tools/setup_venv.py ${VENV_DIR} \
-              --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }}" \
+              --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }} rocm[devel,${{ steps.expand.outputs.device_extras }}]" \
```
HereThereBeDragons, line 217: did you test if the correct version of rocm is being pulled? (and that isnt the latest nightly)
i would think you will also need to fix the rocm version as rocm devel e.g. is independent of torch and torch dependencies.

HereThereBeDragons (commented):
see comment, otherwise good idea to integrate it in the same command

.github/workflows/test_pytorch_wheels_full.yml
ethanwee1: Validated in this job that `rocm[devel]` resolves to the torch-matched ROCm version: https://github.com/ROCm/TheRock/actions/runs/26044789530/job/76566125184

```text
Collecting rocm-sdk-devel==7.14.0a20260518 (from rocm[devel])
rocm==7.14.0a20260518

Comment (HereThereBeDragons) [reply .github/workflows/test_pytorch_wheels_full.yml:217]: i would not trust this run as this just pulled the most recent nightly - which by default should be installed if you just say pip install rocm from our nightly page.

can you please run the full pytorch test suite but use some old nightly rocm version? e.g. could use `7.14.0a20260518` like just in the workflow above

Comment (ethanwee1) [reply .github/workflows/test_pytorch_wheels_full.yml:217]: Good point. Re-ran the full suite pinned to an older nightly to prove the devel pin actually resolves the matching ROCm version (not just the latest): torch `2.10.0+rocm7.14.0a20260518` from v2-staging/gfx94X-dcgpu. Run: https://github.com/ROCm/TheRock/actions/runs/28261891938 — I'll confirm here once it finishes that the installed rocm-sdk reports 7.14.0a20260518 rather than the current nightly.

Comment (ethanwee1) [reply .github/workflows/test_pytorch_wheels_full.yml:217]: Validation confirms the pin works as intended. With `torch==2.10.0+rocm7.14.0a20260518` the setup step resolved the matching old ROCm, not the latest:

```
pip install ... torch==2.10.0+rocm7.14.0a20260518 'rocm[devel]'
Collecting rocm==7.14.0a20260518 (from rocm[libraries]==7.14.0a20260518->torch==2.10.0+rocm7.14.0a20260518)
```

Both 'Set up virtual environment' and 'Initialize ROCm SDK' passed. The run (https://github.com/ROCm/TheRock/actions/runs/28261891938) failed later at the unrelated 'Install test requirements' step — building pandas hit `ModuleNotFoundError: No module named 'pkg_resources'` (a setuptools/build-env issue on the old release branch), not related to this PR's devel pin.

HereThereBeDragons (approved):
please double check if there are any merge conflicts as this is an older pr but other than that lgtm. 
thanks for the patience

Issue #5285, opened 2026-05-15
[Issue]: linux-gfx942-1gpu-ossci runners `Failed to resolve 'rocm.nightlies.amd.com'` errors during torch tests


ScottTodd:
I couldn't find an existing github issue for this, just a few chat threads and internal tickets.

For example: https://github.com/ROCm/TheRock/actions/runs/25911561475/job/76174127864

```
++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/python -m pip install --index-url=https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu torch==2.8.0+rocm7.13.0a20260515
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
Looking in indexes: https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu
WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NameResolutionError("HTTPSConnection(host='rocm.nightlies.amd.com', port=443): Failed to resolve 'rocm.nightlies.amd.com' ([Errno -3] Temporary failure in name resolution)")': /v2-staging/gfx94X-dcgpu/torch/
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NameResolutionError("HTTPSConnection(host='rocm.nightlies.amd.com', port=443): Failed to resolve 'rocm.nightlies.amd.com' ([Errno -3] Temporary failure in name resolution)")': /v2-staging/gfx94X-dcgpu/torch/
WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NameResolutionError("HTTPSConnection(host='rocm.nightlies.amd.com', port=443): Failed to resolve 'rocm.nightlies.amd.com' ([Errno -3] Temporary failure in name resolution)")': /v2-staging/gfx94X-dcgpu/torch/
```

https://rocm.nightlies.amd.com/v2-staging/gfx94X-dcgpu/ appears online and accessible from other machines, so this seems like a DNS issue either for rocm.nightlies.amd.com or the ossci gfx942 nodes.

[...]

PR #5038, opened 2026-05-05
Honor artifact-only Python package installs

stellaraccident:
Partial landing split from #3928.

Changes:

- Keep setuptools and wheel available when setup_venv updates a venv so ROCm sdists can build without resolving build dependencies from the package index under test.

- Treat an explicit empty --index-url as an artifact-only install request and pass pip --no-index --no-build-isolation while preserving --find-links.

- Add unit coverage for the empty-index command shape.

Validation:

- python -m unittest build_tools.tests.setup_venv_test

- git diff --cached --check

Generated with Codex

PR #5324, opened 2026-05-18
[ci] Add DNS fallback for rocm.nightlies.amd.com to S3 bucket

geomin12:
When DNS resolution fails for rocm.nightlies.amd.com, automatically fall back to therock-nightly-python.s3.amazonaws.com. This mitigates DNS issues seen on gfx94X runners where the CDN domain is unreachable but direct S3 access works.

Progress on #5285, as this is a temporary workaround

Tested here with same error scenario: https://github.com/ROCm/TheRock/actions/runs/27578771062/job/81533725193 (cancelled to save resources)

### DIFF
--- a/build_tools/setup_venv.py
+++ b/build_tools/setup_venv.py
@@ -50,10 +50,12 @@ from pathlib import Path
 import platform
 import shlex
 import shutil
+import socket
 import subprocess
 import sys
 import re
 import time
+from urllib.parse import urlparse
 
 from github_actions.github_actions_api import *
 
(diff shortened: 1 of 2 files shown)


build_tools/setup_venv.py
```diff
@@ -188,12 +254,27 @@ def install_packages_into_venv(
...
+            pkg_names = list(ROCM_PACKAGE_NAMES)
+            # Add arch-specific libraries package (e.g., rocm-sdk-libraries-gfx94x-dcgpu).
+            if index_subdir:
+                arch = index_subdir.strip("/").lower()
+                pkg_names.append(f"rocm-sdk-libraries-{arch}")
+            for pkg_name in pkg_names:
+                pkg_find_links = f"{index_url.rstrip('/')}/{pkg_name}/index.html"
+                pip_install_cmd.append(f"--find-links={pkg_find_links}")
+        else:
+            pip_install_cmd.append(f"--index-url={index_url}")
```
ScottTodd, line 359: What do you mean "S3 doesn't serve root index"? It does: https://therock-nightly-python.s3.amazonaws.com/v2-staging/gfx94X-dcgpu/index.html

(Please no rewriting `--find-links` with `--index-url`, they have different behavior)

build_tools/setup_venv.py
```diff
@@ -53,12 +55,76 @@
     "dev": "https://rocm.devreleases.amd.com/v2",
 }
 
+# Fallback mapping from CDN domains to direct S3 bucket URLs.
+# Used when DNS resolution fails for the CDN.
+CDN_TO_S3_FALLBACK_MAP = {
+    "rocm.nightlies.amd.com": "therock-nightly-python.s3.amazonaws.com",
+}
```
ScottTodd, line 76: We should probably also include `rocm.devreleases.amd.com` here too (https://therock-dev-python.s3.amazonaws.com/)

```suggestion
# Fallback mapping from CDN domains to direct S3 bucket URLs.
# Used when DNS resolution fails for the CDN.
CDN_TO_S3_FALLBACK_MAP = {
    "rocm.devreleases.amd.com": "therock-dev-python.s3.amazonaws.com",
    "rocm.nightlies.amd.com": "therock-nightly-python.s3.amazonaws.com",
}
```

build_tools/setup_venv.py
geomin12: ah i need to clarify that comment

ScottTodd: Ah, I see... yeah, that's a tough limitation to work within.

Well, if we're going to use `--find-links` we should make it work as `--index-url` does, as https://github.com/ROCm/TheRock/pull/5038 proposed and add `--no-index -no-build-isolation` to the install command. Otherwise these could install packages from pypi when we _only_ want packages from our index.

geomin12:
re-running test here: https://github.com/ROCm/TheRock/actions/runs/26067154254

build_tools/setup_venv.py
ScottTodd: Needed the other changes (`pip install setuptools`) from that PR too, see the failures here: https://github.com/ROCm/TheRock/actions/runs/26067134118/job/76660785531?pr=5324#step:5:123
```
pip._vendor.pyproject_hooks._impl.BackendUnavailable: Cannot import 'setuptools.build_meta'
```

build_tools/setup_venv.py
```diff
@@ -53,12 +55,79 @@
...
+# Used when DNS resolution fails for the CDN.
+CDN_TO_S3_FALLBACK_MAP = {
+    "rocm.devreleases.amd.com": "therock-dev-python.s3.amazonaws.com",
+    "rocm.nightlies.amd.com": "therock-nightly-python.s3.amazonaws.com",
+}
+
+# ROCm packages for S3 fallback. We use --find-links instead of --index-url because
+# S3 doesn't auto-serve index.html for directory requests (pip requests /pkg/, not
+# /pkg/index.html). Each package's index.html must be specified explicitly.
+ROCM_PACKAGE_NAMES = [
```
jayhawk-commits: I'm not a fan of having to maintain an explicit list here. The list can be dynamically generated, without having to use `pip`, from scraping the HTML content of `index.html` in the fallback URL. e(.g., `{fallback_url}/v2/{arch}/index.html`)

build_tools/setup_venv.py
```diff
@@ -188,15 +271,36 @@ def install_packages_into_venv(
...
+                pip_install_cmd.append(f"--find-links={pkg_find_links}")
+        else:
+            pip_install_cmd.append(f"--index-url={index_url}")
 
     if find_links:
         pip_install_cmd.append(f"--find-links={find_links}")
+        # When using find-links without index-url, prevent fallback to PyPI.
+        if not index_url:
+            pip_install_cmd.append("--no-index")
+            pip_install_cmd.append("--no-build-isolation")
```
jayhawk-commits: The guarding `if find_links:` is only set from arguments, when we want these appended to the install command also when `using_s3_fallback` is set.

jayhawk-commits: A unit test to mock going through the fallback path would be good, versus the unit test that sets `find_links`.

jayhawk-commits (approved):
Please avoid using force push after a PR has been reviewed, so that we can use GitHub's feature of only looking at commits since the last review.

However, I see that my previous concerns were addressed and I do not see anything blocking from the new code.

PR #5326, opened 2026-05-18
[ci] Add optional PyTorch build toggle for workflow_dispatch

geomin12:
As discussed in standup, `workflow_dispatch` release jobs take up lots of capacity, even if not intended (most jobs don't require pytorch builds). Now, the workflow dispatch trigger will have additional options to prevent unnecessary builds and if pytorch is needed, specific python versions to limit.

Changes:
- Adding build_pytorch check mark to build pytorch for workflow dispatch (default off)
- If above option is enabled, specify the python version to build for pytorch
- Creating `build_tools/github_actions/fetch_pytorch_matrix.py` script as logic is introduced in matrix selection

Tested (but cancelled to save resources) here: https://github.com/ROCm/TheRock/actions/runs/26059852448 / https://github.com/ROCm/TheRock/actions/runs/26065169000/job/76634242693 to ensure no syntax issues

Adding `ci:skip` as this only impacts release workflows

In order to avoid logic in YML, we are converting version selection /ignore via python files

### DIFF
--- a/.github/workflows/multi_arch_release.yml
+++ b/.github/workflows/multi_arch_release.yml
@@ -65,6 +65,14 @@ on:
         description: 'Comma-separated list of Windows GPU families. "all" = all, "none" = skip platform.'
         type: string
         default: "gfx110x"
+      build_pytorch:
+        description: "Build PyTorch wheels"
+        type: boolean
+        default: false
+      python_version:
+        description: "Single Python version for wheel builds (empty for full matrix)"
+        type: string
+        default: ""
 
 permissions:
   contents: read
@@ -101,6 +109,8 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ needs.setup.outputs.ref }}
+      build_pytorch: ${{ inputs.build_pytorch }}
+      python_version: ${{ inputs.python_version }}
     permissions:
       contents: read
       actions: write  # Added permission to trigger workflows
@@ -119,6 +129,8 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ needs.setup.outputs.ref }}
+      build_pytorch: ${{ inputs.build_pytorch }}
+      python_version: ${{ inputs.python_version }}
     permissions:
       contents: read
       actions: write  # Added permission to trigger workflows
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -29,6 +29,14 @@ on:
       ref:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
+      build_pytorch:
+        description: "Build PyTorch wheels (defaults to true for workflow_call)"
+        type: boolean
+        default: true
+      python_version:
+        description: "Single Python version for wheel builds (empty for full matrix)"
+        type: string
+        default: ""
 
 permissions:
   contents: read
@@ -186,6 +194,7 @@ jobs:
   trigger_release_pytorch_wheels:
     needs: [build_artifacts, build_python_packages]
     name: Trigger Release PyTorch Wheels
+    if: ${{ inputs.build_pytorch != false }}
     runs-on: ubuntu-24.04
     permissions:
       actions: write
@@ -199,7 +208,8 @@ jobs:
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
-              "ref": "${{ inputs.ref || '' }}"
+              "ref": "${{ inputs.ref || '' }}",
+              "python_version": "${{ inputs.python_version }}"
             }
 
   # TODO: trigger_build_jax_wheels (release_portable_linux_jax_wheels.yml or equivalent)
--- a/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
@@ -44,6 +44,10 @@ on:
         description: "Compiler cache type ('none', 'sccache', or 'ccache')"
         type: string
         default: "none"
+      python_version:
+        description: "Single Python version to build (empty for full matrix)"
+        type: string
+        default: ""
   workflow_dispatch:
     inputs:
       amdgpu_families:
@@ -81,6 +85,10 @@ on:
           - sccache
           - ccache
         default: none
+      python_version:
+        description: "Single Python version to build (empty for full matrix)"
+        type: string
+        default: ""
 
 run-name: Release Linux PyTorch Wheels (${{ inputs.release_type }}, ${{ inputs.rocm_version }}, ${{ inputs.amdgpu_families }})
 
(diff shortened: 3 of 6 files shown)


.github/workflows/release_portable_linux_packages.yml
ScottTodd, line 1: These workflows are deprecated and are scheduled to be deleted ASAP (https://github.com/ROCm/TheRock/issues/3340). Please no new feature work in them. Everything must be migrated to multi-arch CI/CD.

geomin12: oh right! i'll move those over, wrong place to be

.github/workflows/multi_arch_release.yml
```diff
@@ -65,6 +65,22 @@ on:
...
+        type: string
+        default: ""
+      build_jax:
+        description: "Build JAX wheels"
+        type: boolean
+        default: true
+      jax_python_version:
+        description: "Single Python version for JAX builds (empty for full matrix)"
+        type: string
+        default: ""
```
ScottTodd: I doubt someone would want to build PyTorch for one set of Python versions and JAX for a different set. Should we collapse down to a single comma-separated list of `python_versions`?

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -197,7 +214,49 @@ jobs:
...
-              "ref": "${{ inputs.ref || '' }}"
+              "ref": "${{ inputs.ref || '' }}",
+              "python_version": "${{ inputs.pytorch_python_version }}"
             }
 
-  # TODO: trigger_build_jax_wheels (release_portable_linux_jax_wheels.yml or equivalent)
+  trigger_release_jax_wheels:
+    needs: [publish_to_release_buckets]
+    name: Trigger Release JAX Wheels
+    if: ${{ inputs.build_jax != false }}
```
ScottTodd: I don't think this multi_arch_release_linux_jax_wheels.yml workflow works yet. See @erman-gurses 's #4366

Let's keep this PR focused on changing what already exists, not adding something new + untested

geomin12:
In order for folks to stop using old release pipelines, we just need to remove those workflows if we don't future work in there.

However, if they need to stay for a bit, I can do some quick tweaks to avoid big queues until they are officially deleted. This work is primarily for multi arch PyTorch as JAX is not implemented yet in multi arch

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+) -> list[dict]:
+    versions = [python_version] if python_version else PYTHON_VERSIONS
+    pytorch_refs = PYTORCH_REFS_WINDOWS if platform == "windows" else PYTORCH_REFS_LINUX
+    matrix = []
+
+    for py in versions:
+        for ref in pytorch_refs:
+            # Python 3.14 support added in PyTorch 2.9
+            if py == "3.14" and ref == "release/2.8":
+                continue
```
ScottTodd: We don't need this check, release/2.8 support was dropped in https://github.com/ROCm/TheRock/pull/5397

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+            # gfx1153 support added in PyTorch 2.10
+            if amdgpu_family == "gfx1153" and ref in ["release/2.8", "release/2.9"]:
+                continue
```
ScottTodd: We don't need this check either (multi-arch releases build for all targets regardless of release version)

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+            matrix.append({"python_version": py, "pytorch_git_ref": ref})
+
+    return matrix
+
+
+def main():
+    python_version = os.getenv("PYTHON_VERSION") or None
+    amdgpu_family = os.getenv("AMDGPU_FAMILY", "")
```
ScottTodd: Remove AMDGPU_FAMILY env var and variable throughout, it isn't used.

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+            # gfx1153 support added in PyTorch 2.10
+            if amdgpu_family == "gfx1153" and ref in ["release/2.8", "release/2.9"]:
+                continue
+            matrix.append({"python_version": py, "pytorch_git_ref": ref})
+
+    return matrix
+
+
+def main():
+    python_version = os.getenv("PYTHON_VERSION") or None
```
ScottTodd: Let's use argparse for this? Accept a comma-delimited or semicolon-delimited list of python versions to generate the matrix for? Then developers could choose to trigger for python `3.12,3.13` and this script wouldn't need to special case so much on "env var == only one version, otherwise use a list"

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+    amdgpu_family = os.getenv("AMDGPU_FAMILY", "")
+    platform = os.getenv("PLATFORM", "linux")
```
ScottTodd: argparse + default to `platform_module.system().lower()` for consistency with other scripts? That would also give this the expected behavior when run on a Windows host

build_tools/github_actions/fetch_pytorch_matrix.py
ScottTodd, line 1: "fetch" is a bit misleading of a name as this is not acquiring any resource from a remote source (e.g. downloading). I was going to call this `build_tools/github_actions/configure_pytorch_build_matrix.py` or `build_tools/github_actions/configure_pytorch_release_matrix.py`

Looking through https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions, we may later want to sort into some more subfolders or group alphabetically somehow. We have a few related scripts:
* Manifest generation
* Matrix generation
* Uploading
* 'amdgpu' family/target/matrix

build_tools/github_actions/fetch_pytorch_matrix.py
```diff
@@ -0,0 +1,60 @@
...
+
+import json
+import os
+
+from github_actions_api import gha_set_output
+
+PYTHON_VERSIONS = ["3.10", "3.11", "3.12", "3.13", "3.14"]
+
+PYTORCH_REFS_LINUX = [
+    "release/2.8",
```
ScottTodd: remove release/2.8

.github/workflows/multi_arch_release_linux.yml
ScottTodd: (fine to ignore, I see your more recent comment + updates)

ScottTodd:
> In order for folks to stop using old release pipelines, we just need to remove those workflows if we don't future work in there.

Yeah, see https://github.com/ROCm/TheRock/issues/3340 for the cleanup (blocked by JAX not being included in multi-arch releases at all, and PyTorch tests not running there yet either)

.github/workflows/multi_arch_release.yml
ScottTodd: Thinking about this again, we might choose to release JAX and PyTorch for different version ranges. Not ideal, since users should be able to use all supported frameworks on all supported versions, but it might come up...

However, for the workflow_dispatch case of explicitly setting a version [list], it's simpler to have a single input.

ScottTodd (approved):

amd-chiranjeevi:
@geomin12 @ScottTodd 
It broke the multi arch release workflow ([log](https://github.com/ROCm/rockrel/actions/runs/27175495601)) , reverted the change
revert pr: https://github.com/ROCm/TheRock/pull/5710

.github/workflows/multi_arch_release.yml
```diff
@@ -101,6 +109,8 @@ jobs:
       release_type: ${{ inputs.release_type }}
       repository: ${{ inputs.repository || github.repository }}
       ref: ${{ needs.setup.outputs.ref }}
+      build_pytorch: ${{ inputs.build_pytorch }}
+      python_version: ${{ inputs.python_version }}
```
ScottTodd, line 113: errors: https://github.com/ROCm/rockrel/actions/runs/27175495601
`evaluate reusable workflow inputs: ROCm/TheRock/.github/workflows/multi_arch_release.yml@main (Line: 112, Col: 22): Unexpected value ''`

probably due to only adding these inputs to `workflow_dispatch`, so they were undefined for `workflow_call`?

geomin12: yeah that's what it seems like, i'll fix

Issue #3343, opened 2026-02-09
[Multi-arch] Enable multi-arch CI in rocm-libraries and rocm-systems (optional)


marbre:
Note: This is not strictly required, since the current single-arch build structure will remain fully supported.

jayhawk-commits:
I'll try to pick this up after we finalize some of the changes to the build/test matrix on rocm-systems, and when you think multi-arch CI is in a good state to start doing this. I'll try to incorporate some of the ideas I had with the deduplication effort to this.

ScottTodd:
rocm-libraries and rocm-systems are still using workflows based on the non-multi-arch CI workflows, but with the `KPACK_SPLIT_ARTIFACTS` flag flip at https://github.com/ROCm/TheRock/pull/4652 we opted out those workflows as there were artifact upload and test failures.

The `post_build_upload.py` script is currently only uploading `.tar.xz`: https://github.com/ROCm/TheRock/blob/5c2194a0c541b7a89979556c527c5ee4ca6419b2/build_tools/github_actions/post_build_upload.py#L135-L139 while kpack split artifacts use `.tar.zst`: https://github.com/ROCm/TheRock/blob/5c2194a0c541b7a89979556c527c5ee4ca6419b2/build_tools/_therock_utils/artifact_backend.py#L36-L37

I'm seeing if switching from `post_build_upload.py` to `artifact_manager.py push` will pick up all .tar.zst artifacts and use them with this branch: https://github.com/ROCm/TheRock/tree/refs/heads/users/scotttodd/ci-kpack-artifacts, test non-multi-arch CI run: https://github.com/ROCm/TheRock/actions/runs/24795375091. We'll then see if there are other issues (test failures etc.) associated with the flag flip for non-multi-arch CI.

lumachad:
rocgdb also uses the single arch build and will need to migrate. I can do the work, but I'd like rocgdb to be considered/mentioned for this switch so we don't get left behind.

lumachad:
Some thoughts...

Looks like we might need, at a minimum:
- A way to determine which stages we need (or don't need) to build, so a component can do a quick build with only the required bits. For instance debug-tools doesn't need math-libs. rocdbgapi and rocr-debug-agent (part of rocm-systems) are don't need math-libs or comm-libs.
[...]

PR #5332, opened 2026-05-19
[ci] Register rocgdb as a multi-arch external repo

lumachad:
Adds rocgdb to REPO_CONFIGS in detect_external_repo_config.py so the multi-arch CI pipeline can substitute an external rocgdb checkout into the debug-tools build stage via the external_repo mechanism.

Also fixes the CMake invocation in
multi_arch_build_portable_linux_artifacts.yml to emit -DTHEROCK_USE_EXTERNAL_<PKG>=ON alongside
-DTHEROCK_<PKG>_SOURCE_DIR=<path>. rocgdb's source is declared via therock_enable_external_source(... OFF) which uses CACHE STRING FORCE, so a SOURCE_DIR override is silently discarded unless USE_EXTERNAL is also turned on. The extra option is harmless for rocm-libraries / rocm-systems (no USE_EXTERNAL gate defined there).

The rocgdb counterpart, still very WIP:
- https://github.com/ROCm/ROCgdb/pull/130

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -161,7 +161,7 @@ jobs:
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             ${{ inputs.build_variant_cmake_preset && format('--preset={0}', inputs.build_variant_cmake_preset) }} \
-            ${{ inputs.external_repo_config != '' && format('-DTHEROCK_{0}_SOURCE_DIR={1}', fromJSON(inputs.external_repo_config).source_package, fromJSON(inputs.external_repo_config).checkout_path) || '' }} \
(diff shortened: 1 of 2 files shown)


lumachad:
First go at transitioning rocgdb to multi-arch CI to prepare for:
- https://github.com/ROCm/TheRock/issues/3340

And follow:
- https://github.com/ROCm/TheRock/issues/3343

lumachad:
Marking ci:skip for now.

lumachad:
I'm thinking we might need a way to communicate to the workflow that configuring an external rocgdb (or, more generically, any component's) repo can be stage-specific. In that sense, the current change doesn't work quite right as it passes the rocgdb external repo to every stage configure.

lumachad:
We might also want to pass `-DTHEROCK_USE_EXTERNAL_<pkg>=ON -DTHEROCK_<pkg>_SOURCE_DIR=.` to the right stage's configure.

lumachad:
Superseded by https://github.com/ROCm/TheRock/pull/5405.

PR #5217, opened 2026-05-12
[hipfile] Add initial hipFile support

derobins:
## Motivation

Add hipFile (from rocm-systems) to TheRock

## NOTE
This PR depends upon a third-party sysdeps dependency PRs:
* util-linux (for libmount): https://github.com/ROCm/TheRock/pull/5337

## Technical Details

* Add a new storage-libs stage/grouping
* Add hipFile to CLAUDE.md
* Add storage-libs to root CMakeLists.txt
* Add THEROCK_ENABLE_HIPFILE CMake option (default ON)
* Add THEROCK_ENABLE_STORAGE_LIBS CMake option (default ON) 
* Add storage-libs to build_tools/analyze_build_times.py
* Add storage-libs to bump list in build_tools/bump_submodules.py
* Add storage-libs and hipfile info to:
  * docs/development/build_system.md
  * docs/development/windows_support.md
* Add new file: storage-libs/CMakeLists.txt
* Add new file: storage-libs/artifact-hipfile.toml

Packaging will be added in a follow-up PR

This PR needed for rocm-examples' https://github.com/ROCm/rocm-examples/pull/465

## Jira ID

ROCM-1261
AIHIPFILE-170

## Test Plan

Verify TheRock CI passes

## Test Result

TBD

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5337, opened 2026-05-19
[hipfile] Add a util-linux dependency for hipFile

derobins:
## Motivation

hipFile has a dependency on libmount, which in turn requires libblkid, both of which are built from the util-linux repository. This PR adds the dependency for both.

## NOTE

This PR is required for getting hipFile into TheRock. The hipFile --> TheRock PR is here:
https://github.com/ROCm/TheRock/pull/5217

hipFile also has a second dependency on oneTBB, which can be found here:
https://github.com/ROCm/TheRock/pull/5660

## Jira ID

ROCM-1261
AIHIPFILE-170

## Technical Details

Adds a third-party dependency via the usual mechanism. util-linux is pulled from a GitHub tarball and configured (via meson) and built for packaging with TheRock.

The feature is gated on `-DTHEROCK_ENABLE_SYSDEPS_UTIL_LINUX`, which defaults to OFF. When this PR is merged, #5217 will be updated to only build this dependency when hipFile is built.

The util-linux repository can be found here:
https://github.com/util-linux/util-linux

## Test Plan

CI passes, manual inspection of artifacts. The project that needs this dependency is not yet in TheRock.

Everything passes, artifacts are named appropriately, RPATHs are correct.


derobins:
Still needs patch fixes. Won't work properly until that is done.

third-party/sysdeps/linux/artifact-libmount.toml
```diff
@@ -0,0 +1,15 @@
...
+unmatched_exclude = [
+  "lib/rocm_sysdeps/share/locale/**",
+  "lib/rocm_sysdeps/share/man/**",
+]
+
+# libmount
+[components.lib."third-party/sysdeps/linux/libmount/build/stage"]
+[components.dev."third-party/sysdeps/linux/libmount/build/stage"]
+
+[options]
```
lumachad: Two [options] entries is invalid. This is probably a mistake.

third-party/sysdeps/linux/libmount/patch_install.py
```diff
@@ -0,0 +1,82 @@
...
+    # Set RPATH on all prefixed shared libraries (already named correctly from build).
+    for lib_path in lib_dir.glob("librocm_sysdeps_*.so*"):
+        if lib_path.is_symlink():
+            continue
+        try:
+            subprocess.run(
+                [
+                    patchelf_exe,
+                    "--set-rpath",
+                    "$ORIGIN:$ORIGIN/rocm_sysdeps/lib",
```
lumachad: This sysdeps will already be installed to lib/rocm_sysdeps/lib. That's the $ORIGIN. The extra rocm_sysdeps/lib will therefore point at an invalid path.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    # When included in TheRock, we download sources and set up the sub-project.
+    set(_source_dir "${CMAKE_CURRENT_BINARY_DIR}/source")
+    set(_download_stamp "${_source_dir}/download.stamp")
+
+    therock_subproject_fetch(therock-libmount-sources
+      SOURCE_DIR "${_source_dir}"
+      URL "https://github.com/util-linux/util-linux/archive/refs/tags/v2.42.1.tar.gz"
+
```
lumachad, line 15: I know this is temporary, but let's add a URL_HASH in here so we don't forget about it later when we switch to downloading from the s3 bucket.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+include("${THEROCK_SOURCE_DIR}/cmake/therock_meson_env.cmake")
+therock_get_meson_compiler_env(_meson_cc _meson_cxx)
+
+set(patch_source_commands)
+list(APPEND patch_source_commands   COMMAND
+    bash "${CMAKE_CURRENT_SOURCE_DIR}/patch_source.sh" "${SOURCE_DIR}")
+
+add_custom_target(
+  meson_build ALL
```
lumachad, line 82: Nit: Other similar sysdeps call this step "build". Might want to keep it that way for consistency.

third-party/sysdeps/linux/libmount/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+        COMMAND ls "${CMAKE_CURRENT_BINARY_DIR}/build/dist/lib/rocm_sysdeps/include/${libmount_header}"
+      )
+    endforeach()
+
+    return()
+endif()
+
+# Otherwise, this is the sub-project build.
+cmake_minimum_required(VERSION 3.25)
+project(LIBMOUNT_BUILD)
```
lumachad: It would be nice to add the version here.

lumachad (commented):
Let's give this another spin after the duplicate [options] is fixed.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
derobins: Grepping through the other libraries, it's split between meson_build and build and it looks like meson_build is consistently used when meson is the build system.

third-party/sysdeps/linux/libmount/CMakeLists.txt
derobins: Done

third-party/sysdeps/linux/artifact-libmount.toml
derobins: Fixed

third-party/sysdeps/linux/util-linux/CMakeLists.txt
lumachad: That's fine then.

third-party/sysdeps/linux/libmount/patch_install.py
derobins: Fixed. It's just $ORIGIN now.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
derobins: Done

lumachad:
Let's see what CI brings back to us.

lumachad (commented):
It would be nice to add some text to docs/development/dependencies.md in a similar fashion as the other sysdeps. Otherwise this looks OK to me. I've added @marbre for a TheRock-side review.

lumachad:
CI looks OK as far as I can tell. Unrelated failures. I checked the versioning and it seems correct. RPATH is also correct.

third-party/sysdeps/linux/util-linux/patch_source.sh
marbre, line 1: Let us use a proper patch file instead as already done in some other third-party deps. The approach to replace via sed might be kind of reasonable for files created during the build but for everything in the source tree we should go and apply which `patch`. For third-party projects we have patches for see https://github.com/ROCm/TheRock/tree/main/patches/third-party.

CMakeLists.txt
```diff
@@ -424,6 +425,9 @@ if(THEROCK_BUNDLE_SYSDEPS)
     if(THEROCK_ENABLE_SYSDEPS_HWLOC)
       set(THEROCK_BUNDLED_HWLOC therock-hwloc)
     endif()
+    if(THEROCK_ENABLE_SYSDEPS_LIBMOUNT)
```
marbre: Why is this gated?

third-party/sysdeps/linux/libmount/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+# SPDX-License-Identifier: MIT
+
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    # When included in TheRock, we download sources and set up the sub-project.
+    set(_source_dir "${CMAKE_CURRENT_BINARY_DIR}/source")
+    set(_download_stamp "${_source_dir}/download.stamp")
+
+    therock_subproject_fetch(therock-libmount-sources
+      SOURCE_DIR "${_source_dir}"
+      URL "https://github.com/util-linux/util-linux/archive/refs/tags/v2.42.1.tar.gz"
```
marbre:
```suggestion
      # Originally mirrored from: https://github.com/util-linux/util-linux/archive/refs/tags/v2.42.1.tar.gz
      URL "https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/util-linux-2.42.1.tar.gz"
```

third-party/sysdeps/linux/util-linux/CMakeLists.txt
marbre, line 1: Why don't we call this `util-linux`? With `libblkid` this already provides more than only `libmount` and what if someone wants to enable further libs?

CMakeLists.txt
derobins: It's only needed for hipFile at this time. When that library is more established, we could build it unconditionally.

third-party/sysdeps/linux/libmount/CMakeLists.txt
derobins: Done. Also updated the URL to use the S3 bucket.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
derobins: Done

derobins:
> It would be nice to add some text to docs/development/dependencies.md in a similar fashion as the other sysdeps. Otherwise this looks OK to me. I've added @marbre for a TheRock-side review.

I added util-linux to docs/development/dependencies.md

third-party/sysdeps/linux/util-linux/patch_source.sh
derobins: I did this for the meson.build patching since that should be more robust. I left the file clobber for the versions, though, since that's more "single source of truth"-y.

third-party/sysdeps/linux/util-linux/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+  WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
+
+  COMMAND
+    ${patch_source_commands}
```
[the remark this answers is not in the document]
derobins: Copilot is imagining things since it DOES work, but I'll look into the suggestion.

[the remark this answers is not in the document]
derobins: Worthwhile cleanup. Fixed.

lumachad:
This next iteration still looks OK to me.

docs/development/dependencies.md
```diff
@@ -201,6 +202,27 @@ SIMDe (SIMD Everywhere) is a header-only portability library for SIMD intrinsics
...
+
+- Canonical method: `pkg_check_modules(LIBMOUNT REQUIRED IMPORTED_TARGET mount)`
+- Import library: `PkgConfig::LIBMOUNT`
+- Alternatives: none
+
+### libblkid
+
+- Canonical method: `pkg_check_modules(LIBBLKID REQUIRED IMPORTED_TARGET blkid)`
+- Import library: `PkgConfig::LIBBLKID`
+- Alternatives: none
```
marbre, line 224: We're trying to avoid `pkg_check_modules` and rather rely on `find_package`. What is the project you need util-linux for using? Should we switch to `find_package` or eventually even teach it to util-linux?

third-party/sysdeps/linux/util-linux/patch_source.sh
marbre, line 1: Can't this be dropped now and rather go into the patch files?

CMakeLists.txt
```diff
@@ -433,6 +434,9 @@ if(THEROCK_BUNDLE_SYSDEPS)
     if(THEROCK_ENABLE_SYSDEPS_AMD_MESA)
       set(THEROCK_BUNDLED_AMDMESA therock-amd-mesa)
     endif()
+    if(THEROCK_ENABLE_SYSDEPS_UTIL_LINUX)
+      set(THEROCK_BUNDLED_UTIL_LINUX therock-util-linux)
+    endif()
```
marbre, line 439: Seems I can't resolve the prior discussion but I think this should not have been resolved, I still have questions :D

I think the gating here is not doing what you think and it doesn't help to only build it for hipfile.

docs/development/dependencies.md
derobins: util-linux is not a CMake package. It builds with Meson or the Autotools and uses pkg-config. You need pkg_check_modules for this. I need this for hipFile, which is a 7.14 feature.

third-party/sysdeps/linux/util-linux/patch_source.sh
derobins: I could turn the file clobber into a "patch", but that seems a bit weird since we're replacing an entire file.

docs/development/dependencies.md
lumachad: It doesn't need to be a cmake package. See https://github.com/ROCm/TheRock/blob/main/third-party/sysdeps/common/ncurses/ncurses-config.cmake.in for example. We create our own reference and then subcomponents call find_package for that particular reference to find the dependency within TheRock.

That is for the case of bundled sydeps. For unbundled, where things are fetched from the system, then we'll find util-linux some other way.

CMakeLists.txt
derobins: Nobody else is using it, so it only needs to be built for hipFile. If you prefer to build it unilaterally, that's fine as well. hipFile is not in TheRock yet and won't be until this PR is in, so I can't gate it on anything that currently exists.

ScottTodd (commented):
This has enough engaged reviewers already but I can't unrequest myself due to codeowners (the `patches/third-party/` files are fine to edit, no concerns from me). Marking as 'reviewed' to drop it from my queue.

CMakeLists.txt
marbre: The gating here doesn't help and only makes the dep on `therock-util-linux` "conditional". Without a consumer this does nothing anyway so you could just drop it but you also leave as is.

third-party/sysdeps/linux/util-linux/patch_source.sh
marbre: Fair enough to not replicate the `version.lds` file twice.

marbre (approved):

PR #5349, opened 2026-05-19
[hipDNN] ALMIOPEN-1869 Enable hipDNN Python bindings build and add wheel validation test

tvy-amd:
## Summary
- Enable `HIPDNN_BUILD_PYTHON_BINDINGS=ON` for the hipDNN subproject with `therock-nanobind` as a build dependency
- For testing Bump rocm-libraries submodule to include hipDNN Python bindings support (ROCm/rocm-libraries#7352)
- Add Python binding files (`share/hipdnn/python/**`) to the hipDNN test artifact component
- Add `pack_python_wheel.py`: lightweight PEP 427 wheel packer that infers standard wheel tags from extension module filenames
- Add `test_install_hipdnn_python_bindings.py`: end-to-end test that builds a wheel, installs it in a temp venv, validates import, and runs the upstream pytest suite

## Test plan
- [ ] Verify hipDNN builds successfully with Python bindings enabled
- [ ] Confirm `test_install_hipdnn_python_bindings.py` passes on a GPU host with `OUTPUT_ARTIFACTS_DIR` set
- [ ] Confirm wheel installs cleanly and `import hipdnn_frontend` succeeds


build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/libraries_test.py
```diff
@@ -90,7 +90,7 @@ def testSharedLibrariesLoad(self):
                 # in '_rocm_sdk_core/bin'
                 # TODO(#996): track deps in libraries then have the preloader
                 #   recursively get deps instead of hardcoding like this
-                preload_command = "import rocm_sdk; rocm_sdk.preload_libraries('amd_comgr', 'amdhip64', 'hiprtc');"
+                preload_command = "import rocm_sdk; rocm_sdk.preload_libraries('amd_comgr', 'amdhip64', 'hiprtc', 'hipdnn');"
```
tvy-amd, line 93:
```
utils.py:40-49. get_module_shared_libraries() — globs **/*.so.* and **/*.so recursively from package root.
Called at module level in libraries_test.py:30:
```

TIL, this test goes and looks for all the shared objects. The hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so gets discovered so it gets loaded but during the dependency check it doesn't find libhipdnn_backend.so so the tests fail. 
By adding this `hipdnn` preload, it loads the libhipdnn_backend.so into the RTLD_GLOBAL so it can be discovered by hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so.

PR #5360, opened 2026-05-20
Revert "Enable amd-dbgapi on Windows"

marbre:
This reverts commit 41b84b35fe5f77aa0084dc58db169cd9dea1c826. The commit introduced the debug tools stage to Windows, which is broken in multi-arch nightly releases:
https://github.com/ROCm/rockrel/actions/runs/26139063516/job/76887129703

PR #5361, opened 2026-05-20
Forward checkout inputs to Windows debug-tools stage

marbre:
The debug-tools stage in multi_arch_build_windows.yml was added in 41b84b35 without forwarding the repository, ref, and external_repo_config inputs to the reusable multi_arch_build_windows_artifacts.yml workflow. The other three stages (foundation, compiler-runtime, math-libs) all forward those inputs. This broke multi-arch nightly releases triggered in rockrel, which failed during the checkout step.

This adds the missing inputs. Superseds the revert #5360.


lumachad (approved):
LGTM. We should trigger a rocrel job to make sure.

marbre:
> LGTM. We should trigger a rocrel job to make sure.

Triggered https://github.com/ROCm/rockrel/actions/runs/26161538153.

PR #5366, opened 2026-05-20
TheRock: Make rocminfo use correct ROCr lib

cfreeamd:
Some hip tests seem to be failing because the incorrect version of ROCr (libhsa-runtime) is being used.

### DIFF
--- a/build_tools/print_driver_gpu_info.py
+++ b/build_tools/print_driver_gpu_info.py
@@ -34,7 +34,9 @@ def log(*args, **kwargs):
     sys.stdout.flush()
 
 
-def run_command(args: List[str | Path], cwd: Optional[Path] = None) -> None:
+def run_command(
+    args: List[str | Path], cwd: Optional[Path] = None, env: Optional[dict] = None
+) -> None:
     args = [str(arg) for arg in args]
     if cwd is None:
         cwd = Path.cwd()
@@ -50,6 +52,7 @@ def run_command(args: List[str | Path], cwd: Optional[Path] = None) -> None:
             text=True,
             check=False,
             stdin=subprocess.DEVNULL,
+            env=env,
         )
         log(proc.stdout.rstrip())
     except FileNotFoundError:
@@ -61,6 +64,7 @@ def run_command_with_search(
     command: str,
     args: List[str],
     extra_command_search_paths: List[Path],
+    env: Optional[dict] = None,
 ) -> None:
     """
     Run a command, searching in extra paths first, then PATH.
@@ -78,14 +82,14 @@ def run_command_with_search(
         candidate = base / command
         if candidate.exists():
             log(f"\n=== {label} ===")
-            run_command([candidate] + args)
+            run_command([candidate] + args, env=env)
             return
 
     # Then fall back to PATH
     resolved = shutil.which(command)
     if resolved:
         log(f"\n=== {label} ===")
-        run_command([resolved] + args)
+        run_command([resolved] + args, env=env)
         return
 
     # Nothing found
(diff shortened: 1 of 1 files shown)


cfreeamd:
This PR was generated by Claude. 

Background:
In https://amd-hub.atlassian.net/browse/ROCM-23909, the complaint was that rocminfo was returning some incorrect info. A fix to ROCr (libhsa-runtime) was made to correct the rocminfo output. However, the change cannot be merged because it fails CI. Some hiptests are failing because (apparently) they are picking up rocminfo without the fix. So this PR is to try to make the newly built rocr get picked up.
Please see the ticket for more info.

erman-gurses (commented):
I will check in detail but for now pre-commit is failing because of formatting.

cfreeamd:
> I will check in detail but for now pre-commit is failing because of formatting.

Pushed a fix for the Black issues.

build_tools/print_driver_gpu_info.py
```diff
@@ -98,6 +102,17 @@ def run_sanity(os_name: str) -> None:
...
+    env = os.environ.copy()
+    if os_name.lower() != "windows":
+        lib_dir = bin_dir.parent / "lib"
+        if lib_dir.exists():
+            existing_ld_path = env.get("LD_LIBRARY_PATH", "")
+            env["LD_LIBRARY_PATH"] = (
+                f"{lib_dir}:{existing_ld_path}" if existing_ld_path else str(lib_dir)
+            )
+            log(f"=== Setting LD_LIBRARY_PATH={env['LD_LIBRARY_PATH']} ===\n")
+
```
marbre, line 115: I am not confident that this is the right fix. Setting `LD_LIBRARY_PATH` is a big hammer and this might shadow that an rpath is not properly set. @nunnikri, @ScottTodd any thoughts?

nunnikri: Yes using LD_LIBRARY_PATH is not a solution. All interdependent ROCm dependency is supposed to be taken cared with RPATH.  @cfreeamd  what issue you are seeing with rocminfo.

nunnikri (changes requested):

build_tools/print_driver_gpu_info.py
cfreeamd: @nunnikri The output of the CI tests (https://github.com/ROCm/rocm-systems/actions/runs/25928259790/job/76215239429?pr=5676) indicates that the old version of libhsa-runtime is being picked up from somewhere, not the version that is fixed by the PR that is being tested in CI (https://github.com/ROCm/rocm-systems/pull/5676). 

I am not knowledgeable about how TheRock does CI testing for newly built components, or when/how the python scripts are used. Does TheRock set up a test directory where libhsa-runtime is ../lib, relative to rocminfo, in the bin/ directory? If so, then maybe setting rpath would work. Otherwise, I'm not sure how to do it.

Normally, I wouldn't have a problem using LD_LIBRARY_PATH to override other systems paths for testing a library such as this. But if we don't want to use LD_LIBRARY_PATH in CI tests to make sure the newly built component is tested, perhaps rocminfo could be updated. Of course, it needs to work outside of TheRock env. as well. Claude proposed this change to the rocminfo CMakefile.txt:

```
  ###########################`
  # RPATH Configuration
  ###########################
  # Set RPATH for installed binary to find libraries relative to its location
  if(NOT WIN32)
    # Don't skip RPATH for the build tree
    set_target_properties(${ROCMINFO_EXE} PROPERTIES
      SKIP_BUILD_RPATH FALSE
    )

    # When building, use the build RPATH (absolute paths for development)
    set_target_properties(${ROCMINFO_EXE} PROPERTIES
      BUILD_WITH_INSTALL_RPATH FALSE
    )

    # Add the automatically determined parts of the RPATH
    # which point to directories outside the build tree to the install RPATH
    set_target_properties(${ROCMINFO_EXE} PROPERTIES
      INSTALL_RPATH_USE_LINK_PATH TRUE
    )

    # Set the install RPATH to look for libraries in ../lib relative to the binary
    # $ORIGIN is a special token that expands to the directory containing the executable
    set_target_properties(${ROCMINFO_EXE} PROPERTIES
      INSTALL_RPATH "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR};\$ORIGIN/../lib"
    )
  endif()
```

cfreeamd:
I have an alternate PR here that adjust the RPATH imposed by TheRock for rocminfo:

https://github.com/ROCm/TheRock/pull/5410

PR #5367, opened 2026-05-20
add gitleaks shared workflow and callers

cgoea:
## Motivation

Enablement of gitleak secrets scanner for both PRs and post-merge events (decide if we go for schedule instead)

### Architecture: one reusable workflow + two entry points

- **`build_tools/scan_tools/github_actions/gitleaks.py`** — single-entry scanner driver. Downloads + caches the pinned gitleaks binary under `$RUNNER_TEMP`, validates inputs, derives a `--log-opts` git range from the GHA event context, runs `gitleaks detect` once per requested report format, and writes `sarif_path` / `non_sarif_paths` / `config_template_path` to `$GITHUB_OUTPUT` for the calling workflow to consume.
- **`.github/workflows/gitleaks.yml`** — reusable `workflow_call`-only callee. Drives the script with declarative inputs (`scan_mode`, `report_formats`, `scan_path`, `gitleaks_version`), uploads SARIF to Code Scanning under `category: gitleaks`, uploads non-SARIF reports as build artifacts, and mirrors any non-SARIF report into the job log + `$GITHUB_STEP_SUMMARY`.
- **`.github/workflows/gitleaks_main.yml`** — Fires on `schedule` + `workflow_dispatch`, runs `scan_mode: all` (full history), uploads SARIF. Grants `security-events: write` at the caller (see Permissions below).
- **`.github/workflows/pre-commit-security.yml`** — PR entry. Fires on `pull_request`, runs the default `scan_mode: changed` (only commits introduced by the PR via `--log-opts`), produces a CSV artifact. Designed as the PR-side security counterpart to `pre-commit.yml`; sibling scanners can be added as additional jobs in this file as they come online.

### Permission model

`gitleaks.yml` deliberately declares only `permissions: contents: read` and does **not** request `security-events: write` at the workflow level. Each caller grants what it actually needs:

- `gitleaks_main.yml` grants `security-events: write` on its `uses:` job → SARIF reaches the Security tab.
- `pre-commit-security.yml` grants no extra permissions → CSV artifact only.

### Severity classification (gitleaks defaults → High)

Gitleaks's native SARIF formatter leaves `result.level` unset (SARIF default `warning` → GitHub Medium tier) and emits no `properties.security-severity`. The Python driver post-processes the SARIF to backfill both uniformly:

- `result.level = "error"`
- `result.properties.security-severity = 8.5` (CVSS-like; lands in GitHub Code Scanning's **High** tier and makes findings filterable in the Security tab's severity dropdown the same way CodeQL alerts are).

### Dismissal persistence

Alerts dismissed in the Security tab stay dismissed across subsequent scans.

Re-opens only happen on real triggers: the secret was removed and re-introduced (different line hash), the surrounding code changed enough to shift the hash, the gitleaks rule was renamed, or someone uploads under a different `category:`.

### Dependencies

https://github.com/ROCm/TheRock/pull/5643
https://github.com/ROCm/TheRock/pull/5642

## Submission Checklist


.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,159 @@
...
+          values: 'sarif' (default), 'json', 'csv', 'junit'. SARIF outputs are
+          uploaded to GitHub code scanning via github/codeql-action/upload-sarif;
+          all other formats are uploaded as build artifacts via
+          actions/upload-artifact. Example: 'sarif,json'.
+        required: false
+        type: string
+        default: sarif
+      config_path:
+        description: >-
+          Path (relative to the repository root) to the gitleaks config TOML.
```
HereThereBeDragons: dont think we need to expose this as an input.
the config_path should be always the same inside therock

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,159 @@
...
+          missing file gracefully falls back to gitleaks's built-in default
+          rules so the workflow works in repos without a custom config. When
+          set to any non-default value the file MUST exist or the scan job
+          fails (so a typo can't silently disable custom rules).
+        required: false
+        type: string
+        default: gitleaks.toml
+      scan_path:
+        description: >-
+          Path (relative to the repository root) to scan. Defaults to '.' so
```
HereThereBeDragons, line 31: how are sub modules treated? are they part of the full scan from "."?

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,159 @@
...
+        type: string
+        default: "."
+      gitleaks_version:
+        description: Gitleaks release version to download.
+        required: false
+        type: string
+        default: 8.30.1
+      generate_config_template:
+        description: >-
+          When true, write a starter 'gitleaks.toml.template' alongside the
```
HereThereBeDragons: do we need this?
i would expect that the gitleaks.toml in therock will contain all the things we are of interest - and if we want to try something one modifies this on his branch?

.github/workflows/gitleaks_main.yml
HereThereBeDragons, line 1: do we really need this if we have the per-PR check?

its not like with software code where you merge 2 independent prs you can suddenly break the code. you either have secrets in a PR or you dont.

if you want to have this i would opt for a weekly scan - and best would be if that could be just part of the `gitleaks.yml`, or is the problem the `security-events: write`?

.github/workflows/pre_commit_security.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  # Scan commits introduced by this PR (`scan_mode: changed`, the
+  # reusable workflow's default). CSV report is uploaded as a build
+  # artifact and printed to the job summary so reviewers can browse
+  # findings directly from the PR's Checks tab.
+  gitleaks:
+    uses: ./.github/workflows/gitleaks.yml
+    with:
+      report_formats: csv
```
HereThereBeDragons, line 26: why csv if we have the sarif/json everywhere else?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""Run gitleaks against the current repository checkout.
+
+This script is the implementation behind ``.github/workflows/gitleaks.yml``
```
HereThereBeDragons: i dont think this paragraph is needed.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+and any other reusable callers. It performs every step the CI workflow
+needs in a single invocation:
+
+* Downloads + caches the gitleaks release binary at the requested version
+  under ``$RUNNER_TEMP`` and smoke-tests it with ``gitleaks version``
```
HereThereBeDragons: please remove all the double `

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""Run gitleaks against the current repository checkout.
+
+This script is the implementation behind ``.github/workflows/gitleaks.yml``
+and any other reusable callers. It performs every step the CI workflow
+needs in a single invocation:
+
+* Downloads + caches the gitleaks release binary at the requested version
```
HereThereBeDragons: we probably want this in our s3 third party bucket?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+  under ``$RUNNER_TEMP`` and smoke-tests it with ``gitleaks version``
+  before any scan runs.
+* Validates the caller-supplied inputs (scan mode, report formats,
```
HereThereBeDragons: i think all those points should be shorten. too much to read and too little good information 

still reads more like the AI instruction summary :)

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+# Mapping of the gitleaks `--report-format` values we expose to the file
+# extension we use for the on-disk artifact. Keep this list in sync with
+# the `report_formats` input documented in `.github/workflows/gitleaks.yml`.
+_SUPPORTED_FORMATS: dict[str, str] = {
+    "sarif": "sarif",
+    "json": "json",
+    "csv": "csv",
+    "junit": "xml",
+}
+_DEFAULT_GITLEAKS_VERSION = "8.30.1"
```
HereThereBeDragons: just have all those values as argparse default values.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+_LEAK_EXIT_CODE = 1
+_DOWNLOAD_TIMEOUT_SECONDS = 60
+# Numeric `security-severity` (CVSS-like 0.1-10.0) injected into every
+# SARIF result so the GitHub code-scanning Security tab tiers gitleaks
+# findings uniformly at *High*. Unlike a code-quality scanner (which
+# can have low/medium/high tiers per rule), every gitleaks hit is a
+# leaked secret -- there is no useful "low-severity secret" case.
+# 8.5 lands squarely inside GitHub's High tier (7.0-8.9). Bump to 9.0+
+# if exposed secrets should appear under the *Critical* tier instead.
+_LEAK_SECURITY_SEVERITY = "8.5"
```
HereThereBeDragons:
```suggestion
_LEAK_SECURITY_SEVERITY_HIGH = "8.5"
```

than you dont need the blabla above :)

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+
+def _render_config_template(gitleaks_version: str) -> str:
+    """Return the contents of a starter ``gitleaks.toml`` for review.
+
+    The template extends gitleaks's built-in default ruleset and seeds
+    common allowlist patterns + a worked custom-rule example, all
+    commented out. The intent is to give the reviewer a working starting
+    point rather than a wall of opinionated defaults.
+    """
+    return f"""\
```
HereThereBeDragons: can we please have this as a proper file?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+    return raw.strip().lower() in ("true", "1", "yes", "on")
+
+
+def _render_config_template(gitleaks_version: str) -> str:
+    """Return the contents of a starter ``gitleaks.toml`` for review.
+
+    The template extends gitleaks's built-in default ruleset and seeds
+    common allowlist patterns + a worked custom-rule example, all
+    commented out. The intent is to give the reviewer a working starting
+    point rather than a wall of opinionated defaults.
```
HereThereBeDragons: "opinionated"? thats quite some prosa :-D

thinking about it i dont think we need this template function at all. we should provide one good gitleaks.toml we use for therock and thats it.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+class _ReportTarget:
+    """A single ``(format, on-disk path)`` pair the runner will produce."""
+
+    fmt: str
+    path: Path
+
+
+def _gitleaks_release_url(version: str) -> str:
+    return (
+        f"https://github.com/gitleaks/gitleaks/releases/download/v{version}/"
```
HereThereBeDragons: @marbre can maybe tell us better if we want this in our s3 bucket

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+        f"gitleaks_{version}_linux_x64.tar.gz"
+    )
+
+
+def _ensure_gitleaks(version: str) -> Path:
+    """Return the path to a gitleaks binary, downloading it if needed.
+
+    The binary is cached under ``$RUNNER_TEMP`` (when running on a GitHub
```
HereThereBeDragons: "caching" might be a bit misleading on runners as they need to download it everytime anew

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+        return binary
+
+    install_dir.mkdir(parents=True, exist_ok=True)
+    url = _gitleaks_release_url(version)
+    log.info("Downloading gitleaks v%s from %s", version, url)
+    with tempfile.NamedTemporaryFile(suffix=".tar.gz", delete=False) as tmp:
+        tarball_path = Path(tmp.name)
+    try:
+        with (
+            urlopen(Request(url), timeout=_DOWNLOAD_TIMEOUT_SECONDS) as resp,
```
HereThereBeDragons: just put the timeout hardcoded here. its only used here. 
i atl east dont remember its value being set 130 lines ago :)

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,159 @@
...
+    inputs:
+      scan_mode:
+        description: >-
+          'changed' (default) scans only commits introduced by the calling
+          event (PR commits or push range). 'all' scans the full repository
+          history.
+        required: false
+        type: string
+        default: changed
+      report_formats:
```
HereThereBeDragons, line 19: any good reason why we should support writing the same report in multiple formats?
i would rather make it a choice. if we even want this. we could also just say: its always sarif.

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+        )
+    except (OSError, subprocess.CalledProcessError) as exc:
+        raise RuntimeError(
+            f"gitleaks at {binary} failed to execute after install: {exc}"
+        ) from exc
+    log.info("Installed gitleaks %s at %s", result.stdout.strip(), binary)
+    return binary
+
+
+def _parse_report_formats(raw: str) -> list[_ReportTarget]:
```
HereThereBeDragons, line 167: same question as in the workflow: do we need to support writing multiple formats in the same run? especially if underlying we are running gitleaks multiple times anyway.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            "report_formats is empty (expected one or more of: "
+            f"{', '.join(sorted(_SUPPORTED_FORMATS))})"
+        )
+    return targets
+
+
+def _resolve_config_path(raw: str, *, is_default: bool) -> str | None:
+    """Return the existing config path, or ``None`` to use gitleaks defaults.
+
+    When the caller did not override the default path, a missing file is
```
HereThereBeDragons: no graceful fallback. have it as default in argparse and otherwise hard fail.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            data = json.load(f)
+    except json.JSONDecodeError as exc:
+        raise ValueError(
+            f"GITHUB_EVENT_PATH '{event_path}' contains invalid JSON: {exc}"
+        ) from exc
+    except OSError as exc:
+        raise RuntimeError(
+            f"Cannot read GITHUB_EVENT_PATH '{event_path}': {exc}"
+        ) from exc
+    if not isinstance(data, dict):
```
HereThereBeDragons: should this be captured by `xcept json.JSONDecodeError as exc:`?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            f"got {type(data).__name__}"
+        )
+    return data
+
+
+def _determine_log_opts(scan_mode: str, event_name: str, event: dict[str, Any]) -> str:
+    """Build the ``--log-opts`` value for ``gitleaks detect``.
+
+    Returns an empty string to indicate "scan the entire history" (i.e.
+    don't pass ``--log-opts`` at all). Unknown event types fall back to
```
HereThereBeDragons: why would we support unkown even types? i would just make them fail. obviously the user entered something wrong and should get notified about it.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            ["git", "fetch", "--no-tags", "--depth=1", "origin", base_sha],
+            check=False,
+            capture_output=True,
+        )
+        return f"--no-merges {base_sha}..{head_sha}"
+
+    if event_name == "push":
+        before = event.get("before") or ""
+        after = event.get("after") or ""
+        if not before or not after:
```
HereThereBeDragons: is this realistic? 
the api says both are required.

https://docs.github.com/en/webhooks/webhook-events-and-payloads?apiVersion=2026-03-10&versionId=free-pro-team%40latest&category=actions&subcategory=workflow-runs&restPage=github-event-types#push

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            log.warning(
+                "Push event missing before/after SHA; falling back to full history"
+            )
+            return ""
+        # GitHub uses a 0-only SHA for "no previous commit" (new ref);
+        # there's nothing to diff against so we scan everything.
+        if set(before) <= {"0"}:
```
HereThereBeDragons: if i understand correctly this means a pr got opened without any code changes?
then i would not run it all?
as before it is stated that it runs on the diff for prs

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+    Security tab still ingests the SARIF), so we'd rather emit a
+    slightly-less-rich SARIF than fail the scan job.
+    """
+    try:
+        with open(sarif_path, encoding="utf-8") as f:
+            data = json.load(f)
+    except (OSError, json.JSONDecodeError) as exc:
+        log.warning("SARIF severity enrichment skipped (%s): %s", sarif_path, exc)
+        return
+    if not isinstance(data, dict):
```
HereThereBeDragons: again dont think we need this

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            sarif_path,
+        )
+        return
+
+    enriched_level = 0
+    enriched_score = 0
+    preserved_level = 0
+    preserved_score = 0
+    for run in data.get("runs") or []:
+        if not isinstance(run, dict):
```
HereThereBeDragons, line 301: is this possible that it isnt a dict?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            continue
+        for result in run.get("results") or []:
+            if not isinstance(result, dict):
```
HereThereBeDragons, line 313: same here

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+                continue
+            if result.get("level") is None:
+                result["level"] = "error"
+                enriched_level += 1
```
HereThereBeDragons: so these are counters?
might be good to make this more clear

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            else:
+                preserved_level += 1
+            props = result.get("properties")
+            if not isinstance(props, dict):
```
HereThereBeDragons, line 325: same here

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            if tgt.fmt == "sarif" and tgt.path.is_file():
+                _enrich_sarif_with_security_severity(tgt.path)
+            continue
+        raise RuntimeError(
+            f"gitleaks exited unexpectedly with code {rc} for format '{tgt.fmt}'"
+        )
+    return leaks_found
+
+
+def _write_github_output(**values: str) -> None:
```
HereThereBeDragons: please use build_tools/github_actions/github_actions_api.py
that already provides functionality e.g. for gha_append_step_summary

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+    with open(out_path, "a", encoding="utf-8") as f:
+        for key, raw in values.items():
+            value = "" if raw is None else str(raw)
+            if "\n" in value:
+                f.write(f"{key}<<EOF\n{value}\nEOF\n")
+            else:
+                f.write(f"{key}={value}\n")
+
+
+def build_parser() -> argparse.ArgumentParser:
```
HereThereBeDragons, line 463: description need then adjustments for whatever changes are applied

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            "No config found at default path '%s'; using gitleaks built-in default rules",
+            raw,
+        )
+        return None
+    raise FileNotFoundError(
+        f"config_path '{raw}' was explicitly set but no such file exists"
+    )
+
+
+def _load_github_event() -> dict[str, Any]:
```
HereThereBeDragons: build_tools/github_actions/github_actions_api.py has gha_load_github_event

HereThereBeDragons (commented):
overall not bad. do you have a n example run to see how the output looks like?

some general comments
- i think you are exposing to many features to the user - which makes sense as standalone and for development, but not here in therock where we want a static working version where no one needs and also should not fiddle around with it.

- for me again a bit too much prosa/blabla. would be nicer to have a higher ratio information content : text length.

- now in therock you can also take advantage of build_tools/github_actions/github_actions_api.py to use functions from there so you dont need to write your own (and you will see your code is more defensively written than that script).

.github/workflows/pre-commit-security.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# Pre-commit security gate: fast, PR-side security scanners that run on
+# every pull request. The security counterpart to `pre-commit.yml`
```
marbre:
```suggestion
# every pull request. The security counterpart to `pre-commit.yml`.
```

.github/workflows/pre-commit-security.yml
```diff
@@ -0,0 +1,26 @@
...
+# sibiling scans can be added as additional jobs in this file as they
```
marbre:
```suggestion
# Sibling scanners can be added as additional jobs in this file as they
```
?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+
+
+def _gitleaks_release_url(version: str) -> str:
+    return (
+        f"https://github.com/gitleaks/gitleaks/releases/download/v{version}/"
+        f"gitleaks_{version}_linux_x64.tar.gz"
+    )
+
+
+def _ensure_gitleaks(version: str) -> Path:
```
marbre: We don't  check integrity it seems. After downloading the SHA-256 isn't ever verified against the published `gitleaks_*_checksums.txt`. For a security scanner this is the exact threat model the rest of the PR is set up to defend against. If a release tarball were ever swapped, every TheRock PR would silently execute it under the GHA token. Let's pin to one checksum and verify before `tar.extract`. This is what we do with every third-party dep, even though we mirror those!

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,159 @@
...
+            cat -- "$report_path"
+            echo "::endgroup::"
+            {
+              echo "### Gitleaks report: \`${report_path}\`"
+              echo
+              echo '```'
+              cat -- "$report_path"
+              echo '```'
+            } >> "$GITHUB_STEP_SUMMARY"
+          done <<< "$NON_SARIF_PATHS"
```
marbre: No inline code in workflows please. This needs to go into a script, see our design guidelines.

build_tools/scan_tools/github_actions/gitleaks.py
marbre, line 1: As this is adding ~700 lines of Python code with branching logic, it deserves tests. Please see the Python style-guide we're following. Seems most functions are pure and trivially testable with synthetic input.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,712 @@
...
+            log.warning(
+                "Push event missing before/after SHA; falling back to full history"
+            )
+            return ""
+        # GitHub uses a 0-only SHA for "no previous commit" (new ref);
+        # there's nothing to diff against so we scan everything.
+        if set(before) <= {"0"}:
+            log.info("Push created a new ref; falling back to full history scan")
+            return ""
+        return f"--no-merges {before}..{after}"
```
marbre: Spotted by my review agent:

Both PR and push log-opts include --no-merges. If a contributor introduces a secret via a merge commit (not unheard of on long-running feature branches that get conflict-resolved at merge), the changed-mode scan won't see it. The post-merge scan_mode: all workflow catches it eventually, but the PR gate won't. Worth either removing the flag or documenting the gap.

scan_tools/github_actions/gitleaks.py
marbre: We want to! :)

.github/workflows/gitleaks.yml
cgoea: sarif is not very human friendly, that's why I kept all the options open. I would suggest for PR using the csv and for whole repo scan sarif as it's the only format github security accepts.

.github/workflows/gitleaks_main.yml
cgoea: was thinking initially to have the overall status in the security tab. there everyone could disposition their findings or fix them and so on. once we are all green it can be disabled since the PR check will safeguard the addition of any other secrets. weekly scan sounds good and yes; I kept it separate so only this workflow has security-events: write access.

.github/workflows/gitleaks.yml
cgoea: ok, removing this if there's no use case for it.

.github/workflows/gitleaks.yml
cgoea: no. checkout is only done for the current super repo. repo, no submodules are pulled. The way I see it those repos should be scanned at their level not here. I'll add a small explanation so that it's clear submodules don't get scanned.

.github/workflows/pre_commit_security.yml
cgoea: it was my own preference as csv could pe opened in excel and thought is more human readable

scan_tools/github_actions/gitleaks.py
cgoea: ok, I'll take a look how that can be done.

scan_tools/github_actions/gitleaks.py
cgoea: changed to use the S3

scan_tools/github_actions/gitleaks.py
cgoea: Seems a bit excessive, especially if we want to share this with other repos. let's discuss on it.

scan_tools/github_actions/gitleaks.py
cgoea: ok, guess we can go that way.

scan_tools/github_actions/gitleaks.py
cgoea: added count suffix

scan_tools/github_actions/gitleaks.py
HereThereBeDragons: why? in therock it will take the default and if we have anywhere some other path i would expect that you give the correct path via cli args

build_tools/github_actions/github_actions_api.py
```diff
@@ -379,6 +379,62 @@ def gha_append_step_summary(summary: str):
...
+        raise RuntimeError(
+            f"Cannot read GITHUB_EVENT_PATH '{event_path}': {exc}"
+        ) from exc
+    if not isinstance(data, dict):
+        raise ValueError(
+            f"GITHUB_EVENT_PATH '{event_path}' must contain a JSON object, "
+            f"got {type(data).__name__}"
+        )
+    return data
+    """Load the GitHub Actions workflow event JSON from disk.
```
HereThereBeDragons: this code section is never reached? and it semes like this is the original `gha_load_github_event`

build_tools/github_actions/github_actions_api.py
ScottTodd: What's going on with the changes in this file? Can you move these to their own PR to keep this PR focused? Or mark as a draft PR so it doesn't trigger codeowners if this is not actually ready for review/merge?

This code looks wrong:
```python
def function(args):
  """docstring 1"""
  return 1
  """docstring 2"""
  return 2  # never hit, as the previous return replaces it)
```

.github/workflows/gitleaks_main.yml
HereThereBeDragons: if the gitleaks csv you shared with me was over the entire therock code base i think we are only left with false positives. how do we remove those that they are not raised every time?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+import tarfile
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from urllib.request import Request, urlopen
+
+# Reach the shared GitHub Actions helpers under `build_tools/`. Mirrors
+# the import pattern used by `build_tools/generate_manifest_diff_report.py`.
+_REPO_ROOT = Path(__file__).resolve().parents[2]
```
HereThereBeDragons: existing scripts tend to use 

```
THEROCK_DIR = Path(__file__).resolve().parent.parent.parent

# Add build_tools to path for _therock_utils imports.
sys.path.insert(0, str(THEROCK_DIR / "build_tools"))
```

.github/workflows/pre_commit_security.yml
HereThereBeDragons, line 1: file name should use _

.github/workflows/gitleaks_main.yml
```diff
@@ -0,0 +1,29 @@
...
+name: Gitleaks_main
+
+on:
+  # Run every Saturday at 10:00 UTC.
+  schedule:
+    - cron: "0 10 * * 6"
+  # Allow re-running the weekly scan on demand, e.g. after
+  # tweaking `gitleaks.toml` or rotating the gitleaks version pin, so
+  # the Security tab can be refreshed without an unrelated commit.
+  workflow_dispatch:
```
HereThereBeDragons, line 15: claude:
Major — [.github/workflows/gitleaks_main.yml:114-120](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/.github/workflows/gitleaks_main.yml#L114-L120) — PR description / file disagree
PR body says gitleaks_main.yml fires on push: main + workflow_dispatch. File actually has schedule: "0 10 * * 6" + workflow_dispatch — no push trigger. Update one to match the other.

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+  or gitleaks itself errored.
+
+Inputs come from CLI flags or matching `GITLEAKS_*` env vars set by the
+workflow.
+"""
+
+import argparse
+import json
+import logging
+import os
```
HereThereBeDragons, line 21: please alphabetical ordering

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+        )
+    return targets
+
+
+def _resolve_config_path() -> str | None:
+    """Return `_CONFIG_PATH` if it exists, else `None` (gitleaks built-ins)."""
+    if Path(_CONFIG_PATH).is_file():
+        log.info("Using gitleaks config: %s", _CONFIG_PATH)
+        return _CONFIG_PATH
+    log.warning(
```
HereThereBeDragons: please let us have it a hard fail. if the config path is wrong we also dont want to use the built-in rules - as they will be either different or too little of what the repo would expect.

if you want to support it just for testing, then have it as an cli arg. but it shouldnt be silent

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,93 @@
...
+  contents: read
+
+jobs:
+  scan:
+    name: Gitleaks scan
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
```
HereThereBeDragons: fetch-depth 0 is not the cheapest. i wonder if dependent on if it is a pr or a full run we can change the depth?

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+        # rather than something we should try to paper over.
+        pr = event["pull_request"]
+        base_sha = pr["base"]["sha"]
+        head_sha = pr["head"]["sha"]
+        # Best-effort fetch of the base SHA so the range is reachable.
+        # `actions/checkout` with `fetch-depth: 0` fetches HEAD's lineage,
+        # but the PR base may live on a ref that wasn't explicitly fetched.
+        subprocess.run(
+            ["git", "fetch", "--no-tags", "--depth=1", "origin", base_sha],
+            check=False,
```
HereThereBeDragons, line 223: why check false?
we should print at leas tsome log message that the base is not reachable

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+        # leave that class of leak undetected on long-running branches.
+        return f"{base_sha}..{head_sha}"
+
+    if event_name == "push":
+        # GitHub guarantees `before` and `after` on push events.
+        before = event["before"]
+        after = event["after"]
+        # GitHub uses a 0-only SHA for "no previous commit" (new ref);
+        # there's nothing to diff against so we scan everything.
+        if set(before) <= {"0"}:
```
HereThereBeDragons: zero sha should be `Z40 = '0' * 40` as e.g. used by https://github.com/pre-commit/pre-commit/blob/main/pre_commit/commands/hook_impl.py

scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            f.write(payload)
+            path = Path(f.name)
+        self.addCleanup(path.unlink, missing_ok=True)
+        self.assertEqual(_sha256_of(path), hashlib.sha256(payload).hexdigest())
+
+    def test_pinned_constant_is_a_valid_sha256_hex_string(self):
+        # Guards against typos / accidental truncation in the constant.
+        self.assertEqual(len(_GITLEAKS_TARBALL_SHA256), 64)
+        int(_GITLEAKS_TARBALL_SHA256, 16)  # raises ValueError on non-hex
```
HereThereBeDragons: claude:
Nit — [scan_tools/github_actions/tests/gitleaks_test.py:1043-1045](vscode-webview://0hqkdvqoea7nkvc21be0sti4vied45o2ih50e8i0kish8ei0jl2c/scan_tools/github_actions/tests/gitleaks_test.py#L1043-L1045)
int(_GITLEAKS_TARBALL_SHA256, 16) works, but self.assertRegex(_GITLEAKS_TARBALL_SHA256, r"^[0-9a-f]{64}$") states the intent (lowercase hex, length 64) in one line.

build_tools/scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertEqual(
+            [t.path for t in targets],
+            [
+                Path("gitleaks-report.sarif"),
+                Path("gitleaks-report.json"),
+                Path("gitleaks-report.csv"),
+            ],
+        )
+
+    def test_junit_uses_xml_extension(self):
```
HereThereBeDragons, line 45: why exactly this test?

build_tools/scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertEqual(_determine_log_opts("all", "pull_request", {}), "")
+        self.assertEqual(_determine_log_opts("all", "release", {"unrelated": 1}), "")
+
+    def test_pull_request_returns_sha_range_without_no_merges(self):
+        event = {"pull_request": {"base": {"sha": "aaa"}, "head": {"sha": "bbb"}}}
+        log_opts = _determine_log_opts("changed", "pull_request", event)
+        self.assertEqual(log_opts, "aaa..bbb")
+        self.assertNotIn("--no-merges", log_opts)
+
+    def test_pull_request_target_is_explicitly_rejected(self):
```
HereThereBeDragons, line 84: why is this suppose to be rejected?

scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+            "3.5",
+        )
+
+    def test_empty_runs_is_a_noop(self):
+        path = self._write_sarif({"runs": []})
+        original = path.read_text()
+        _enrich_sarif_with_security_severity(path)
+        self.assertEqual(path.read_text(), original)
+
+    def test_malformed_top_level_is_skipped(self):
```
HereThereBeDragons: a malformed sarif should fail

scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+        path = self._write_sarif(["not", "a", "dict"])
+        original = path.read_text()
+        _enrich_sarif_with_security_severity(path)
+        # File should be left unchanged when payload is unexpectedly shaped.
+        self.assertEqual(path.read_text(), original)
+
+    def test_missing_file_is_skipped(self):
```
HereThereBeDragons: i think this should also fail.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,548 @@
...
+                    levels_set_count += 1
+                else:
+                    levels_kept_count += 1
+                props = result.setdefault("properties", {})
+                if props.get("security-severity") is None:
+                    props["security-severity"] = _LEAK_SECURITY_SEVERITY_HIGH
+                    scores_set_count += 1
+                else:
+                    scores_kept_count += 1
+    except (OSError, json.JSONDecodeError, AttributeError, KeyError, TypeError) as exc:
```
HereThereBeDragons: this should be a hard fail not just a logging a warning. 
sarif file must exist if _enrich_sarif_with_security_severity is called otherwise it should hard fail. the check should be done in the outside function to ensure that _enrich_sarif_with_security_severity is only called if sarif file exist

scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def tearDown(self):
+        os.chdir(self._original_cwd)
+        self._tmp.cleanup()
+
+    def test_returns_config_path_when_present(self):
+        Path(_CONFIG_PATH).write_text("# stub config", encoding="utf-8")
+        self.assertEqual(_resolve_config_path(), _CONFIG_PATH)
+
+    def test_returns_none_when_missing(self):
```
HereThereBeDragons: as i said i think it is better if it fails in case the config file does not exist

scan_tools/github_actions/tests/gitleaks_test.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertIsNone(_resolve_config_path())
+
+
+class Sha256OfTest(unittest.TestCase):
```
HereThereBeDragons: i dont think this test is needed

HereThereBeDragons (commented):
are we planning to add more tools in scan_tools/?  otherwise i think it should be in build_tools/github_actions/ like we have all the others.

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          python-version: "3.12"
+
+      - name: Run gitleaks
+        id: scan
+        env:
+          GITLEAKS_SCAN_MODE: ${{ inputs.scan_mode }}
+          GITLEAKS_REPORT_FORMATS: ${{ inputs.report_formats }}
+          GITLEAKS_SOURCE_DIR: ${{ inputs.scan_path }}
+        run: python scan_tools/github_actions/gitleaks.py
```
haribabug: Should we need to include "timeout-minutes:" field to avoid run forever?

.github/workflows/gitleaks.yml
```diff
@@ -0,0 +1,93 @@
...
+
```
haribabug, line 80: Hope its configured to redact the secrets.

.github/workflows/gitleaks_main.yml
```diff
@@ -0,0 +1,29 @@
...
+
```
haribabug, line 16: Should we need to include "timeout-minutes:" field to avoid run forever?

.github/workflows/pre_commit_security.yml
```diff
@@ -0,0 +1,26 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# Pre-commit security gate: fast, PR-side security scanners that run on
+# every pull request. The security counterpart to `pre-commit.yml`.
+# Sibling scanners can be added as additional jobs in this file as they
+# come online, keeping all PR-side security gates discoverable in one
+# place.
+
```
haribabug, line 9: Should we need to include "timeout-minutes:" field to avoid run forever?

.github/workflows/gitleaks.yml
ScottTodd: Yeah I was wondering about this style of integration. If the leak checker runs _after a commit is already pushed_, doesn't that defeat the purpose? Or at least wouldn't it be strictly better to catch such leaks prior to push? Once a commit is pushed, sensitive data is out there ~forever, so at _best_ what we can do it flag that and advise the developer to revoke/rotate the secret or do damage control for the leak. It certainly doesn't help if the leak checker itself draws more (public) attention to the leak via a failing workflow run or logs highlighting what was leaked.

gitleaks seems to have a pre-commit hook: https://github.com/gitleaks/gitleaks#pre-commit-hook that could run on developer machines prior to push, and github also has push protection for developers using the common repository here (and not personal forks).

The PR description has a lot of words in it but I can't tell what problem this is actually solving and if it solves that problem effectively compared to other possible implementations...

cgoea: @haribabug  Yes, secrets are redacted.  @ScottTodd  is more of last line of defense. Ideally everyone should know not to commit secrets but if someone does by accident this should flag it and the owner should quickly revoke it.

build_tools/github_actions/github_actions_api.py
cgoea: yes, forgotten from the initial implementation.

scan_tools/github_actions/gitleaks.py
cgoea: thought it should be agnostic to the rock, but I get the standardization so changing to this.

scan_tools/github_actions/gitleaks.py
cgoea: ok, going with this approach

.github/workflows/gitleaks.yml
cgoea: something like this  "fetch-depth: ${{ github.event.pull_request.commits && github.event.pull_request.commits + 1 || 0 }}" as checkout is capable to evaluate the expression.

build_tools/scan_tools/github_actions/tests/gitleaks_test.py
cgoea: cause junit produces and xml. the rest of the formats are covered by test_multiple_formats_with_whitespace_and_dedup

build_tools/scan_tools/github_actions/tests/gitleaks_test.py
cgoea: for security reasons. in case of forks this runs with access to secrets. tokens and so on but the payload will have commits from the fork. https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target  comes with  a warning

scan_tools/github_actions/gitleaks.py
cgoea: ok. going for hard failure everywhere if sarif is not created

scan_tools/github_actions/tests/gitleaks_test.py
cgoea: changed in gitleaks to use your approach. tests behave the same now

build_tools/scan_tools/github_actions/gitleaks.py
cgoea: I would vote yes as this will be used by other repos with and dev teams there might prefer different report types. but this is just my view. @haribabug what do you think?

cgoea:
> are we planning to add more tools in scan_tools/? otherwise i think it should be in build_tools/github_actions/ like we have all the others.

I think so. Hari had a list of tools this being just the first.

.github/workflows/gitleaks_main.yml
cgoea: @HereThereBeDragons  removed them via the config toml. enhanced the allow list to match our current false positives. also put a lot of explanations so it's clear what each config does.

.github/workflows/gitleaks.yml
cgoea: @HereThereBeDragons figured out a way to get the fetch depth dynamically.

haribabug (approved):

.github/workflows/gitleaks.yml
```diff
@@ -41,27 +41,28 @@ on:
...
+          fetch-depth: 1
+
+      - name: Compute checkout fetch-depth
+        id: depth
+        uses: ./.github/actions/compute_pr_fetch_depth
+
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          fetch-depth: 0
+          fetch-depth: ${{ steps.depth.outputs.value }}
```
HereThereBeDragons, line 65: why are we not using the version you suggested? thats also dynamic and does not need any extra compute steps:
`${{ github.event.pull_request.commits && github.event.pull_request.commits + 1 || 0 }}`

.github/workflows/gitleaks_main.yml
```diff
@@ -0,0 +1,29 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# Weekly gitleaks scan: runs on a fixed cadence and pushes
+# SARIF findings to the repository's code-scanning Security tab.
+name: Gitleaks_main
```
HereThereBeDragons: can we name this nicer without the underscore?

gitleaks.toml
```diff
@@ -0,0 +1,74 @@
...
+
+
+# Allowlist field-vs-target reference:
+#   stopwords         -> always matched against the SECRET value.
+#   regexes           -> matched against `regexTarget` (default: "secret").
+#   regexTarget       -> "secret" | "match" | "line".
+# Use "match" or "line" when the "tell" is the variable name or
+# surrounding syntax, not the value itself.
+
+[[allowlists]]
```
HereThereBeDragons, line 28: do we really want to not check all files? unless we explicitly exclude it i would say we should always scan everything? like secrets can also be leaked through markdown documents.

gitleaks.toml
```diff
@@ -0,0 +1,74 @@
...
+  '''(?i)docs?/.*\.md$''',
+  '''third-party/''',
+  '''.*\.lock$''',
+]
+
+[[allowlists]]
+description = "Dummy secret values (matched against the secret itself)"
+# Default regexTarget = "secret". Keep this list focused on literal
+# placeholder values that may legitimately appear as a token.
+stopwords = [
```
HereThereBeDragons: should this be empty for now? until we find one?

gitleaks.toml
```diff
@@ -0,0 +1,74 @@
...
+  "your-token-here",
+  "REPLACE_ME",
+]
+regexes = [
+  '''(?i)example[_-]?key''',
```
HereThereBeDragons: same here

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,602 @@
...
+        _GITLEAKS_TARBALL_URL,
+    )
+    with tempfile.NamedTemporaryFile(suffix=".tar.gz", delete=False) as tmp:
+        tarball_path = Path(tmp.name)
+    try:
+        with (
+            urlopen(Request(_GITLEAKS_TARBALL_URL), timeout=60) as resp,
+            open(tarball_path, "wb") as out,
+        ):
+            shutil.copyfileobj(resp, out)
```
HereThereBeDragons: something from claude to think about (no strong opinion from my side) - i asked to consider security implications more during the review:
Security: download body is unbounded. shutil.copyfileobj(resp, out) writes whatever the S3 endpoint sends. If the bucket object is ever swapped for a multi-GB blob (compromise or accident), the runner fills its disk before the SHA-256 check rejects it. Cheap mitigation: read in a loop with a hard cap (e.g. 100 MB — gitleaks tarball is ~5 MB) and raise before hashing.

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,602 @@
...
+        tarball_path.unlink(missing_ok=True)
+
+    if not binary.is_file():
+        raise RuntimeError(
+            f"gitleaks tarball for v{_GITLEAKS_VERSION} did not contain "
+            f"a 'gitleaks' file at {binary}"
+        )
+    binary.chmod(0o755)
+
+    try:
```
HereThereBeDragons, line 146: claude:
Smoke-test runs gitleaks version but doesn't assert it matches _GITLEAKS_VERSION.

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,602 @@
...
+            scores_kept_count,
+            sarif_path,
+        )
+        return
+
+    try:
+        with open(sarif_path, "w", encoding="utf-8") as f:
+            json.dump(data, f, indent=2)
+    except OSError as exc:
+        log.warning("Failed to write enriched SARIF to %s: %s", sarif_path, exc)
```
HereThereBeDragons: raise? better to fail here hard or not?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,602 @@
...
+    have to download the `gitleaks-report` artifact, unzip it, and open
+    the file locally:
+
+    1. stdout, wrapped in `::group::`/`::endgroup::` workflow
+       commands so the live job log stays scannable.
+    2. `$GITHUB_STEP_SUMMARY` via `gha_append_step_summary` so the
+       report is one click away from the run page.
+
+    Missing files are skipped with a warning rather than failing the
+    job; they typically mean gitleaks exited before producing that
```
HereThereBeDragons: how can gitleak exit before producing output? that should then be some hard failure?

scan_tools/github_actions/gitleaks.py
```diff
@@ -0,0 +1,602 @@
...
+            log.warning(
+                "non-SARIF report '%s' missing; skipping log + summary emission",
+                path,
+            )
+            continue
+        content = path.read_text(encoding="utf-8", errors="replace")
+        print(f"::group::Gitleaks report: {path}")
+        print(content)
+        print("::endgroup::")
+        summary_chunks.append(f"### Gitleaks report: `{path}`\n\n'`\n{content}\n'`")
```
HereThereBeDragons: claude things this si broken:
Broken code fence in step summary. The literal is '\...'`` (single-quote + backtick), not a triple-backtick fence. The CSV/JSON report renders as inline-code soup in the GitHub step summary, defeating the point of mirroring it there.


do yoiu have some current output?

build_tools/github_actions/github_actions_api.py
```diff
@@ -383,18 +392,47 @@ def gha_append_step_summary(summary: str):
 
 
 def gha_load_github_event() -> dict[str, Any]:
-    """Load the GitHub Actions workflow event JSON from disk.
+    """Loads the JSON event payload pointed to by $GITHUB_EVENT_PATH.
```
HereThereBeDragons, line 395: in the old version this was a hard fail if not set:
`path = os.environ["GITHUB_EVENT_PATH"]`
now it returns `{}`


did you double check that existing code handles this correctly?

@ScottTodd  any opinion on this?

HereThereBeDragons (commented):
still some few more changes but slowly getting there

build_tools/github_actions/github_actions_api.py
ScottTodd: Returning empty if the env var is unset seems fine, though I've been solving this by having
* unit tests mock the related github actions APIs to simulate what is needed
* scripts avoid github-related paths when run locally, or skip/warn

These changes to `github_actions_api.py` could also move to a separate PR. Keep PRs as small and focused as possible for significantly faster reviews.

.github/workflows/gitleaks.yml
cgoea: the "+1" crashes the precomit check as illegal.

build_tools/github_actions/github_actions_api.py
cgoea: @ScottTodd  yes, it's safe, checked it before, but if the preffered way is hard fail I can change it. I'll move this to a separate PR.

gitleaks.toml
cgoea: ok. removed markdown and fixtures since they are handled more precisely by value based and variable name allowlists. leaving the rest especially lock files which cause a lot of false positives

gitleaks.toml
cgoea: removing these

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -56,18 +42,16 @@
...
     "sarif": "sarif",
     "json": "json",
     "csv": "csv",
     "junit": "xml",
 }
-# Gitleaks release mirrored to the rocm-third-party-deps S3 bucket (see
-# docs/development/git_chores.md "Updating a third-party mirror") so the
-# CI runner doesn't depend on github.com/gitleaks/gitleaks being reachable.
+# Mirrored to the rocm-third-party-deps S3 bucket (see
+# docs/development/git_chores.md) so CI doesn't depend on github.com.
```
HereThereBeDragons, line 54: maybe add the original URL so one can easily update it to a newer version

scan_tools/github_actions/gitleaks.py
```diff
@@ -76,22 +60,15 @@
 _GITLEAKS_TARBALL_SHA256 = (
     "551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb"
 )
-# Repository-root config path. Hardcoded (no input override): every
-# caller of this script points at the same file, so exposing a path
-# input just invited drift between callers. A missing file gracefully
-# falls back to gitleaks's built-in default rules so the workflow keeps
-# working in fresh checkouts before a config has been committed.
+# Hardcoded (no input override) so every caller uses the same file and
```
HereThereBeDragons: think this comment is a bit redundant?

scan_tools/github_actions/gitleaks.py
```diff
@@ -110,17 +87,6 @@ def _sha256_of(path: Path) -> str:
 
 
 def _ensure_gitleaks() -> Path:
-    """Return the path to a verified gitleaks binary, downloading it if needed.
```
HereThereBeDragons: i would keep this line + the next sentence as comment

`_ensure_gitleaks` is very ambigous what it means. maybe `get_gitleaks_binary()` might be easier to understand. in any case the install location etc is not bad to have it in the description

scan_tools/github_actions/gitleaks.py
```diff
@@ -175,16 +140,17 @@ def _ensure_gitleaks() -> Path:
...
+        raise RuntimeError(
+            f"gitleaks at {binary} reports version {installed_version!r}, "
+            f"expected {_GITLEAKS_VERSION!r}"
+        )
+    log.info("Installed gitleaks %s at %s", installed_version, binary)
     return binary
 
 
 def _parse_report_formats(raw: str) -> list[_ReportTarget]:
-    """Parse a comma-separated `report_formats` value into report targets.
```
HereThereBeDragons: some comment here is not bad

scan_tools/github_actions/gitleaks.py
```diff
@@ -219,14 +184,7 @@ def _resolve_config_path() -> str:
 
 
 def _determine_log_opts(scan_mode: str, event_name: str, event: dict[str, Any]) -> str:
-    """Build the `--log-opts` value for `gitleaks detect`.
-
-    Returns an empty string to indicate "scan the entire history" (i.e.
```
HereThereBeDragons: the return types are not bad to have. maybe as a bulletpoint list for easier reading?

scan_tools/github_actions/gitleaks.py
```diff
@@ -289,27 +247,8 @@ def _determine_log_opts(scan_mode: str, event_name: str, event: dict[str, Any])
 def _enrich_sarif_with_security_severity(sarif_path: Path) -> None:
     """Mark every gitleaks SARIF result as High severity for code scanning.
 
-    Gitleaks's native SARIF formatter leaves `result.level` unset
-    (defaults to SARIF's `warning` -> GitHub's Medium tier) and does
-    not emit `properties.security-severity`. Every gitleaks hit is a
-    leaked secret -- there's no useful low/medium tier for that -- so
-    we backfill both fields uniformly:
-
-    * `level = "error"` (so SARIF viewers that don't read
```
HereThereBeDragons: one small comment about this would make sense. otherwise the "pre-existing values.." is out of context.

build_tools/scan_tools/github_actions/gitleaks.py
```diff
@@ -386,12 +325,25 @@ def _enrich_sarif_with_security_severity(sarif_path: Path) -> None:
         )
         return
 
+    tmp_path: Path | None = None
     try:
-        with open(sarif_path, "w", encoding="utf-8") as f:
-            json.dump(data, f, indent=2)
+        with tempfile.NamedTemporaryFile(
```
HereThereBeDragons, line 349: why do you think this extra step is necessary?

scan_tools/github_actions/gitleaks.py
```diff
@@ -458,22 +409,14 @@ def _run_gitleaks(
...
 
 
-def _emit_non_sarif_reports(non_sarif: list[_ReportTarget]) -> None:
-    """Surface each non-SARIF report in the workflow run.
-
-    Writes each report's content to two places so PR reviewers don't
-    have to download the `gitleaks-report` artifact, unzip it, and open
-    the file locally:
+def _md_code_fence(content: str) -> str:
+    """Return a backtick fence longer than any backtick run in `content`."""
```
HereThereBeDragons: why is this needed? would be good to add a sentence - as right now this is not obvious

gitleaks.toml
```diff
@@ -26,28 +26,15 @@ useDefault = true
...
 [[allowlists]]
-description = "Structurally non-sensitive paths"
+# Allowlist by location only where a real first-party secret structurally
+# can't live: vendored third-party trees and dependency lock files.
+description = "Vendored and generated paths"
 paths = [
-  '''(?i)tests?/fixtures/''',
-  '''(?i)docs?/.*\.md$''',
   '''third-party/''',
+  '''third_party/''',
```
HereThereBeDragons: its now double?
and we also provide cmake code/stuff in third_party

HereThereBeDragons (commented):
getting close

build_tools/scan_tools/github_actions/gitleaks.py
cgoea: if the process crashes mid-write or the filesystem fills up, we avoid leaving a truncated/corrupted SARIF

gitleaks.toml
cgoea: thought third party is fully "outside" stuff. ok, then I'll remove it

build_tools/scan_tools/github_actions/gitleaks.py
marbre, line 1: @ScottTodd any objections about adding a new top level directory or do you feel this should go into an existing subdirectory?

ScottTodd: Yes, I'd prefer a subfolder under https://github.com/ROCm/TheRock/tree/main/build_tools

gitleaks.toml
HereThereBeDragons: no sadly not. we build it with some modifications to rename the symbols to not clash with the default systemwide installations

HereThereBeDragons (commented):
lgtm

please adjust the location as scott said and also update the pr description

build_tools/scan_tools/github_actions/gitleaks.py
cgoea: @ScottTodd  moved into build_tools

cgoea:
> lgtm
> 
> please adjust the location as scott said and also update the pr description

@HereThereBeDragons  moved into build_tools and adapted the description.

PR #5370, opened 2026-05-20
Add host-blas patch to remove hardcoded Zi and DEBUG flags from CMake

amd-nicknick:
## Motivation

Upstream openblas contains hardcoded /Zi and /DEBUG flags in their CMakeLists.txt when building on Windows, this makes it incompatible with ccache.

This together with https://github.com/ROCm/rocm-systems/pull/6239 should cover all scenarios where the /Zi flag is still leaking into TheRock builds.

## Technical Details

Add a patch to remove the hardcoded line in ther CMakeLists.txt. 

## Test Plan

Build passes with ccache eligible rate.

## Test Result

Build passes with ccache eligible rate.

## Submission Checklist


marbre (approved):
Looks reasonable to me but Scott may want to have a second look before landing.

ScottTodd (approved):

PR #5377, opened 2026-05-20
Adding kfdtest to The Rock CI.

andmar-amd:
Add kfdtest to TheRock CI
    
    - Add kfdtest subproject with correct dependencies (statically links libhsakmt and LLVM)
    - Use run_kfdtest.sh script which handles platform-specific test exclusions
    - Add --kfdtest flag to install_rocm_from_artifacts.py
    - Add kfdtest to BUILD_TOPOLOGY.toml (artifact group, build stage, and artifact)
    - Add sysdeps to kfdtest artifact dependencies for bundled libraries
    - Use get_target_property to dynamically locate ROCR-Runtime binary directory
    - Set RPATH to lib/rocm_sysdeps/lib only (kfdtest statically links main deps)
    - Support both standard and quick test modes via TEST_TYPE environment variable

## Submission Checklist

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -632,3 +632,63 @@ if(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_RUNTIME_TESTS)
       rocrtst
   )
 endif(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_RUNTIME_TESTS)
+
+if(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_KFDTESTS)
+
+    # kfdtest statically links libhsakmt and LLVM (AMDGPUAsmParser, Core, Support).
+    # It needs LLVM headers at build time and libhsakmt.a for linking.
+    set(_kfdtest_build_deps
+      amd-llvm
+      ROCR-Runtime
+      therock-yaml-cpp
+    )
+
+    # Get the ROCR-Runtime binary directory to locate libhsakmt
+    get_target_property(_rocr_binary_dir ROCR-Runtime THEROCK_BINARY_DIR)
+
+    therock_cmake_subproject_declare(kfdtest
+      USE_TEST_AMDGPU_TARGETS
+      EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_SYSTEMS_SOURCE_DIR}/projects/rocr-runtime/libhsakmt/tests/kfdtest"
+      BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/kfdtest"
+      BACKGROUND_BUILD
+      CMAKE_ARGS
+        "-DCMAKE_PREFIX_PATH="
+        "-DLLVM_DIR="
+        "-DROCM_DIR="
+        "-DLIBHSAKMT_PATH=${_rocr_binary_dir}/libhsakmt"
+        "-DCMAKE_EXE_LINKER_FLAGS=-ldl"
+        "-DCMAKE_BUILD_WITH_INSTALL_RPATH=ON"
+      COMPILER_TOOLCHAIN
+        "${_system_toolchain}"
+      BUILD_DEPS
+        ${_kfdtest_build_deps}
+      RUNTIME_DEPS
+        amd-llvm
+        ${THEROCK_BUNDLED_LIBDRM}
+        ${THEROCK_BUNDLED_NUMACTL}
+        ${THEROCK_BUNDLED_ZLIB}
+        ${THEROCK_BUNDLED_ZSTD}
+      INTERFACE_LINK_DIRS
+        "lib"
+        "lib/rocm_sysdeps/lib"
+      INTERFACE_INSTALL_RPATH_DIRS
+        "lib"
+        "lib/rocm_sysdeps/lib"
+    )
+    therock_cmake_subproject_glob_c_sources(kfdtest SUBDIRS .)
+    therock_cmake_subproject_activate(kfdtest)
+
+    therock_provide_artifact(kfdtest
+      TARGET_NEUTRAL
+      DESCRIPTOR artifact-core-kfdtest.toml
+      COMPONENTS
+        dbg
+        dev
+        doc
+        lib
+        run
+        test
+      SUBPROJECT_DEPS
+        kfdtest
+    )
+endif(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_KFDTESTS)
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -151,6 +151,7 @@ artifact_groups = [
     "hip-runtime",
     "opencl-runtime",
     "profiler-core",
+    "kfdtest",
     "rocjitsu"
 ]
 
@@ -268,6 +269,12 @@ source_sets = ["rocm-systems"]  # clr is in rocm-systems
 description = "Runtime tests (hip-tests, rocrtst)"
 type = "generic"
 artifact_group_deps = ["hip-runtime", "opencl-runtime", "core-runtime"]
+source_sets = ["rocm-systems"]
+
+[artifact_groups.kfdtest]
+description = "KFD kernel driver tests"
+type = "generic"
+artifact_group_deps = ["core-runtime"]
 source_sets = ["rocm-systems"]
 
 [artifact_groups.math-libs]
@@ -563,6 +570,14 @@ artifact_group = "runtime-tests"
 type = "target-neutral"
 artifact_deps = ["core-runtime", "core-ocl", "core-amdsmi", "sysdeps-hwloc"]
 feature_name = "CORE_RUNTIME_TESTS"
+feature_group = "CORE"
+disable_platforms = ["windows"]
+
+[artifacts.kfdtest]
+artifact_group = "kfdtest"
+type = "target-neutral"
+artifact_deps = ["core-runtime", "amd-llvm", "sysdeps"]
+feature_name = "CORE_KFDTESTS"
 feature_group = "CORE"
 disable_platforms = ["windows"]
 
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -43,6 +43,7 @@ python build_tools/install_rocm_from_artifacts.py
     [--rocprofiler-systems-examples | --no-rocprofiler-systems-examples]
     [--rocrtst | --no-rocrtst]
     [--rocalution | --no-rocalution]
+    [--kfdtest | --no-kfdtest]
     [--rocwmma | --no-rocwmma]
     [--libhipcxx | --no-libhipcxx]
     [--tests | --no-tests]
@@ -386,6 +387,7 @@ def retrieve_artifacts_by_run_id(args):
             args.rocprofiler_systems_examples,
             args.rocrtst,
             args.rocalution,
+            args.kfdtest,
             args.rocwmma,
             args.libhipcxx,
         ]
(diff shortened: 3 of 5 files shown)


andmar-amd:

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
```diff
@@ -0,0 +1,83 @@
...
+test_type = os.getenv("TEST_TYPE", "full")
+
+if test_type == "quick":
+    # For quick tests, use the core_sws test suite which is a minimal set
+    # that tends to succeed consistently
+    cmd.extend(["-p", "core_sws"])
+
+# Run on node 7 only to avoid flaky SDMA queue tests on other nodes
+# Node 7 has shown consistent test success in CI
+cmd.extend(["-n", "7"])
```
saadrahim: We should not hardcode this. Flaky SDMA queue tests are symptom of other underlying issues that need to be resolved.

saadrahim (changes requested):

saadrahim:
PLATFORM_MAP is defined but never referenced (the comment says detection is automatic) — dead code.
The large commented-out ADDITIONAL_EXCLUDE block is dead code; if container exclusions aren't needed in TheRock CI, delete it rather than ship it commented.
SHARD_INDEX/TOTAL_SHARDS defaults are ints (os.getenv(..., 1)) while env values are strings; it works because of later int(...) casts, but it's inconsistent.
In core/CMakeLists.txt, the existing rocrtst block gets re-indented (whitespace-only churn) just to nest the new if(... CORE_KFDTESTS) beside it. That inflates the diff and makes blame noisier — better to add the new block without reformatting the old one.

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
andmar-amd: Greetings @saadrahim 

Thank for the comment.  This was just an attempt to see it would succeed here.  Is there a way to get a node exclusively for kfdtest?  Is does not play well with sharing the GPU, it has to have the node for itself .  As for how long it would take to run this full test, it averages to about  20 minutes.

saadrahim: I believe you should consider it a requirement to make the tests work on a shared GPU. What feature does the test exercise that all GPUs are impacted? Can you separate the ones that cannot share GPUs because they test a feature that impacts all GPUs.

andmar-amd: @kentrussell could articulate better on this.

But it is not that it need all of the GPU nodes, it just needs 1 in the set, and it need to have all by itself or it may crash.  FTR it does run on all the GPU node available, if it is not controlled with a '-n', and this usually will be green if there is no other process sharing that specific node.  Since it looks like you have 8+ nodes on this ASIC , just show me how I can lock 1 exclusively for kfdtest and then I would release it when the test is complete.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -672,6 +672,17 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
             "windows": 1,
         },
     },
+    # kfdtest tests
+    "kfdtest": {
+        "job_name": "kfdtest",
+        "fetch_artifact_args": "--kfdtest --tests",
+        "timeout_minutes": 15,
+        "test_script": f"python {_get_script_path('test_kfdtest.py')}",
+        "platform": ["linux"],
```
geomin12: please add:

```
# Architectures that we have multi GPU setup for testing
"multi_gpu": {"linux": ["gfx94X-dcgpu", "gfx950-dcgpu"]},
```

like here:
https://github.com/ROCm/TheRock/blob/bbee0509e381741847b82d78ce4a990f4c9f6208/build_tools/github_actions/fetch_test_configurations.py#L441-L453

this will ensure that these tests will run on a single node, so the tests can be done without impacting others

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
geomin12: let's avoid hard-coding, what if we want to test on 4-gpu?

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
```diff
@@ -0,0 +1,83 @@
...
+#     "gfx90a": {
+#         "linux": [
+#             "KFDSVMRangeTest.HMMProfilingEvent*",
+#         ]
+#     },
+#     ...
+# }
+
+# Check if quick tests are requested
+test_type = os.getenv("TEST_TYPE", "full")
```
geomin12: let's make default as standard pls

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
```diff
@@ -0,0 +1,83 @@
...
+    "gfx941": "gfx942",
+    "gfx942": "gfx942",
+    "gfx950-dcgpu": "gfx950",
+    "gfx1100": "gfx1100",
+    "gfx1101": "gfx1101",
+    "gfx1102": "gfx1102",
+    "gfx1103": "gfx1103",
+    "gfx1150": "gfx1150",
+    "gfx1151": "gfx1151",
+}
```
geomin12: remove if not used

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
```diff
@@ -0,0 +1,83 @@
...
+cmd.extend(["-n", "7"])
+
+# Pass through any gtest sharding args
+gtest_args = []
+if int(TOTAL_SHARDS) > 1:
+    gtest_args.append(f"--gtest_total_shards={TOTAL_SHARDS}")
+    gtest_args.append(f"--gtest_shard_index={int(SHARD_INDEX) - 1}")
+
+if gtest_args:
+    cmd.extend(gtest_args)
```
geomin12: do we use gtest sharding? i see it's a bash script for tests

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
```diff
@@ -0,0 +1,83 @@
...
+logging.basicConfig(level=logging.INFO)
+
+# GTest sharding
+SHARD_INDEX = os.getenv("SHARD_INDEX", 1)
+TOTAL_SHARDS = os.getenv("TOTAL_SHARDS", 1)
+environ_vars = os.environ.copy()
+# For display purposes in the GitHub Action UI, the shard array is 1th indexed.
+# For shard indexes, we convert to 0th index.
+environ_vars["GTEST_SHARD_INDEX"] = str(int(SHARD_INDEX) - 1)
+environ_vars["GTEST_TOTAL_SHARDS"] = str(TOTAL_SHARDS)
```
geomin12: we can remove if not used

core/CMakeLists.txt
```diff
@@ -618,17 +618,78 @@ if(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_RUNTIME_TESTS)
...
+        doc
+        lib
+        run
+        test
+      SUBPROJECT_DEPS
+        rocrtst
+    )
+
+  endif(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_RUNTIME_TESTS)
+  if(THEROCK_BUILD_TESTING AND THEROCK_ENABLE_CORE_KFDTESTS)
```
geomin12: nit: space in between

geomin12 (commented):
works! https://github.com/ROCm/TheRock/actions/runs/27444211651/job/81144051501?pr=5377

but have some comments

geomin12 (commented):
please provide better PR description

build_tools/github_actions/fetch_test_configurations.py
andmar-amd: @geomin12 , thanks. 

For my learning, just give me a thumbs up if these thoughts are in alignment.  The logic here seem to be in reverse, b/c when I saw the "multi_gpu" line I was thought it meant that I was forcing kfdtest to run on multiple gpu, that is why it was left out,  but now it seem to work the opposite way?  I guess the  "rccl" example runs also on single GPU also?

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
kentrussell: While kfdtest can benefit from running on all GPUs, there is a lot of redundancy. The majority of subtests are testing the thunk+kernel interface, and it's not designed to be a HW validation tool like agfhc. Yes certain GPUs could throw failures in certain cases where previous ones did not, but it's rare. For TheRock CI, a single GPU is sufficient, since that covers the interfaces well enough, and speeds up the test by 7x
The big issue is that a number of subtests don't play well with others. Allocating all VRAM (or more), triggering faults, triggering GPU resets, etc. These all conflict with running other tests simultaneously

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
andmar-amd: Will do, this PR is Clade Code Assisted, so it was following the `rocrtst` example (I was told by Luis M) which means that some of things it mimic was not needed.  I guess it would have been more accurate to follow the  "rccl" example?

build_tools/github_actions/test_executable_scripts/test_kfdtest.py
saadrahim: kfdtest is designed to test the driver. TheRock CI does not gate the driver releases. Therefore, updating the driver may break TheRock builds if a failure is introduced in the driver and is caught in TheRock CI through kfdtests. From a testability perspective, I do not support its enablement in TheRock. I think you need to consult with leadership on the CI side. Fixing a defect in the driver has no associated source code in TheRock as well. 

@amd-aakash @suchitravemuri1 @ScottTodd @marbre 

We also cannot publish kfdtests as a component versioned with ROCm, it needs to be versioned with the AMD GPU driver.

nunnikri (changes requested):
kfdtest is an internal amdgpu driver test. Its not testing user space ROCm. It need not be part of TheRock/ROCm ci. Or I would suggest to make it as 2 part change. bringing the  kfdtest artifact to ROCm as first step and adding to TheRock ci be a different PR which need more consideration. 
@saadrahim @amd-aakash   could you clarify?

saadrahim:
> kfdtest is an internal amdgpu driver test. Its not testing user space ROCm. It need not be part of TheRock/ROCm ci. Or I would suggest to make it as 2 part change. bringing the kfdtest artifact to ROCm as first step and adding to TheRock ci be a different PR which need more consideration. @saadrahim @amd-aakash could you clarify?

Yes, adding it as a build artifact is the best approach. There is no need to add to CI for testing. kfdtests is only being built for historic purposes and should be removed in the future.

nunnikri:
> > kfdtest is an internal amdgpu driver test. Its not testing user space ROCm. It need not be part of TheRock/ROCm ci. Or I would suggest to make it as 2 part change. bringing the kfdtest artifact to ROCm as first step and adding to TheRock ci be a different PR which need more consideration. @saadrahim @amd-aakash could you clarify?
> 
> Yes, adding it as a build artifact is the best approach. There is no need to add to CI for testing. kfdtests is only being built for historic purposes and should be removed in the future.

@andmar-amd do you want to remove build_tools/github_actions/test_executable_scripts/test_kfdtest.py based on this?

andmar-amd:
Greetings @nunnikri @saadrahim 

Yes I have completely remove the build_tools/github_actions/test_executable_scripts/test_kfdtest.py and other references to testing.  As you can see AI review is now failing for missing unit test.

saadrahim (approved):
This is needed temporarily until we find a new home for kfdtests.

nunnikri (approved):
Discussed with @saadrahim. BU is good to release kfdtest as it is. So approving.

jayhawk-commits:
Latest ASAN builds are failing on main due to kfdtest, I will revert this commit.
Example log: https://github.com/ROCm/TheRock/actions/runs/28172850345/job/83441457289

stellaraccident:
> Latest ASAN builds are failing on main due to kfdtest, I will revert this commit. Example log: https://github.com/ROCm/TheRock/actions/runs/28172850345/job/83441457289

```
[kfdtest] ld.lld: error: unable to find library -lhsakmt
[kfdtest] ld.lld: error: unable to find library -lnuma
[kfdtest] clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

This doesn't look like it would be as an specific and I would expect that it would impact the non asan builds too. It will need cmake plumbing changes to detect the proper in tree dependencies for those two deps.

PR #5258, opened 2026-05-14
[ci] Use packaged runners for selected rocm-systems tests

jayhawk-commits:
## Summary

Use packaged rocm-systems test runners from the built ROCm payload for selected TheRock test jobs.

This removes TheRock-side wrappers for:

- `hip-tests`
- `rocr-debug-agent`
- `rocdecode`
- `rocjpeg`

## Related Issues

- #3968
- #5090 

## Validation

- `python -m py_compile build_tools/github_actions/fetch_test_configurations.py`
- `python -m pytest -q build_tools/github_actions/tests/fetch_test_configurations_test.py`
- `git diff --check`

PR #5385, opened 2026-05-21
refactor(debug-tools): remove test scripts now owned by component submodules

lumachad:
Removes test_rocgdb.py and test_rocr-debug-agent.py from build_tools/github_actions/test_executable_scripts/ and points to the component-owned testing scripts from the build directory.

The change in #4960 makes it so we can fetch test_rocgdb.py from rocgdb's submodule's build tree.

The change in https://github.com/ROCm/rocm-systems/pull/5651 makes it so we can fetch test_rocr-debug-agent.py from rocr-debug-agent's submodule (rocm-systems).

With both PRs above merged and this PR pointing at the new locations, the test_rocgdb.py and test_rocr-debug-agent.py scripts in TheRock serve no purpose anymore.

Related: #4960


lumachad:
One thing I need your feedback in @jayhawk-commits is the CI script location information. At least for rocgdb we don't have a single build directory as we can build multiple rocgdb's each linked against a particular version of python.

Would it make sense, in this case, to point at the install tree location?

I see you've pointed at the build tree in this PR (https://github.com/ROCm/TheRock/pull/5258), so I did that initially in this PR. But I think I'd like both to point at the install tree. Would that make sense?

lumachad:
Updated to match the split rocgdb-cpu and rocgdb-gpu changes from https://github.com/ROCm/TheRock/pull/5132.

lumachad:
@jayhawk-commits @dileepr1 You might also want to track this one, when we finally remove TheRock's CI scripts for rocgdb and rocr-debug-agent.

lumachad:
Lifting ci:skip label.

lumachad:
I see both rocr-debug-agent and rocgdb CI scripts were invoked from their new build directory locations. I wonder if we should instead invoke them from the install tree.

lumachad:
> I see both rocr-debug-agent and rocgdb CI scripts were invoked from their new build directory locations. I wonder if we should instead invoke them from the install tree.

Nevermind. I misread. The scripts are being invoked from the ROCm tree where the artifacts get installed.

jayhawk-commits (approved):
LGTM. I see the CI for this PR is using the new script paths.

lumachad:
Thanks. I'll hold off merging this one until it's appropriate.

lumachad:
@lancesix  Looks like I need an approval from the debug-tools side as well.

lancesix (approved):
LGTM, thanks.

Issue #5110, opened 2026-05-07
[Multi-arch] PyTorch release workflows should generate and upload manifest files


ScottTodd:
Non-multi-arch releases like https://github.com/ROCm/TheRock/actions/runs/25482880855 generate a set of manifest files like https://therock-nightly-artifacts.s3.amazonaws.com/25482880855-linux/manifests/gfx950-dcgpu/index.html with contents like
```json
{
  "pytorch": {
    "commit": "1a2700743c0a51bb12e51ca66c4b33e34e8640cf",
    "repo": "https://github.com/ROCm/pytorch.git",
    "branch": "release/2.10"
  },
  "pytorch_audio": {
    "commit": "5047768f2447d963dffc250b64a5d6c01afc84fa",
    "repo": "https://github.com/pytorch/audio"
  },
  "pytorch_vision": {
    "commit": "82df5f599578b383987510836bb05ea97dcc9669",
    "repo": "https://github.com/pytorch/vision"
  },
  "triton": {
    "commit": "4ed888920c5a0871957f1cf912e557bc79fbe56c",
    "repo": "https://github.com/ROCm/triton.git"
  },
  "apex": {
    "commit": "751f5dd5ae26adedfd8bd0f39c3aa46a8fdcb4d8",
    "repo": "https://github.com/ROCm/apex"
  },
  "therock": {
    "commit": "6cde06cfe0a2bd41f2070c531ff24c2e939f40ed",
    "repo": "https://github.com/ROCm/TheRock.git",
    "branch": "main"
  }
}
```

The new https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml workflow does not include this code: https://github.com/ROCm/TheRock/blob/360c9753a808d650531afc97d59a78e78d5c2bb7/.github/workflows/build_portable_linux_pytorch_wheels.yml#L268-L278 https://github.com/ROCm/TheRock/blob/360c9753a808d650531afc97d59a78e78d5c2bb7/.github/workflows/build_portable_linux_pytorch_wheels.yml#L293-L301

I don't think a direct translation is necessarily the way to go though. The design all along should have been
1. Compute manifests (see https://github.com/ROCm/TheRock/issues/1236)
2. Build based on those manifests

Instead of the current
1. Build based on floating refs (these can differ across workflows running concurrently)
2. Generate manifests based on what was build

[...]

PR #5397, opened 2026-05-21
Remove pytorch 2.8

HereThereBeDragons:
With TheRock release 7.13 we stopped providing pytorch 2.8 packages, so also drop it for the nightlies.

Removes also our pytorch test filtering files: 
- external-builds/pytorch/skip_tests/pytorch_2.7.py
- external-builds/pytorch/skip_tests/pytorch_2.8.py


.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
marbre, line 1: We actually did for multi-arch, see https://repo.amd.com/rocm/whl-multi-arch/torch/, but didn't for single-arch. This was most likely because multi-arch always builds the matrix defined here and is not controllable from rockrel. So this seems to be the right fix for me here.

Regarding the downloading, promotion or uploading scripts, we could add a flag to filter. Makes most sense at downloading time. @HereThereBeDragons, @araravik-psd if this makes sense to both of you, please file an issue. Either @araravik-psd or @cgoea might be able to pick this up.

marbre (approved):

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -104,11 +104,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-        pytorch_git_ref: ["release/2.8", "release/2.9", "release/2.10", "release/2.11", "release/2.12", "nightly"]
-        exclude:
-          # Python 3.14 support was added in PyTorch 2.9.
-          - pytorch_git_ref: "release/2.8"
-            python_version: "3.14"
```
ScottTodd, line 111: Oh nice! If we remove this `exclude` it will simplify some of the work I'm doing on https://github.com/ROCm/TheRock/issues/5110 😅 (I'm moving the matrix of versions into a script)

ScottTodd (approved):

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
HereThereBeDragons: nice! yeah we should move more of those things in scripts.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
araravik-psd: Thanks, that makes sense. I filed an issue to track this and I have picked it up. I’ll add package filtering at download time so we can exclude specific package variants before promotion/upload.

https://github.com/ROCm/TheRock/issues/5443

ScottTodd (approved):

PR #5045, opened 2026-05-05
Collapse foundation and split runtime tests

stellaraccident:
Partial landing split from #3928.

Changes:

- Remove the foundation stage and fold sysdeps/base into compiler-runtime.

- Add a runtime-tests stage for hip-tests and rocrtst so runtime test builds no longer sit on the compiler-runtime critical path.

- Update multi-stage workflow dependencies, helper text, docs, and CI unit-test expectations for the new stage layout.

PR #5405, opened 2026-05-21
[debug-tools][don't merge] rocgdb multi-arch CI

lumachad:
Draft to implement rocgdb multi-arch CI on TheRock's side

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux.yml
+++ b/.github/workflows/multi_arch_build_portable_linux.yml
@@ -43,6 +43,10 @@ on:
         type: string
         default: ""
         description: "Comma-separated build stages to skip (artifacts already copied by the orchestrator)"
+      skip_stages:
+        type: string
+        default: ""
+        description: "Comma-separated build stages to skip entirely (no build, no prebuilt fetch). Used by external repos that only need a subset of the topology."
       build_runs_on:
         type: string
         default: "azure-linux-scale-rocm"
--- a/.github/workflows/multi_arch_build_portable_linux.yml
+++ b/.github/workflows/multi_arch_build_portable_linux.yml
@@ -77,7 +77,7 @@ jobs:
   # STAGE: foundation (generic)
   # ==========================================================================
   foundation:
-    if: ${{ !cancelled() && !failure() && !contains(inputs.prebuilt_stages, 'foundation') }}
+    if: ${{ !cancelled() && !failure() && !contains(inputs.prebuilt_stages, 'foundation') && !contains(inputs.skip_stages, 'foundation') }}
     uses: ./.github/workflows/multi_arch_build_portable_linux_artifacts.yml
     secrets: inherit
     with:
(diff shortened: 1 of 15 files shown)


.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -43,6 +43,10 @@ on:
         type: string
         default: ""
         description: "Comma-separated build stages to skip (artifacts already copied by the orchestrator)"
+      skip_stages:
+        type: string
+        default: ""
+        description: "Comma-separated build stages to skip entirely (no build, no prebuilt fetch). Used by external repos that only need a subset of the topology."
```
geomin12, line 49: oh... i think some of these are needed as dependencies to build? more for the `no prebuilt fetch` comment

.github/workflows/multi_arch_build_portable_linux.yml
```diff
@@ -73,7 +77,7 @@ jobs:
   # STAGE: foundation (generic)
   # ==========================================================================
   foundation:
```
geomin12, line 79: I believe stella removed foundation #5045

.github/workflows/multi_arch_build_portable_linux.yml
geomin12, line 1: joseph and i discussed a way to fetch prebuilt steps (to avoid rebuilding). opportunity to combine? i like the idea of skipping stages, but probably would prefer to do this more dynamically in a python script

geomin12, line 1: i guess another question: what cases would we have for skipping steps? wouldn't we want to prefetch if possible instead?

.github/workflows/multi_arch_build_portable_linux.yml
lumachad: FTR, I'm not actively updating this PR as I'm following what you're doing before I plan the debug-tools side. With that said, I'm happy with a mechanism that calculates the minimum amount of stages to build for a particular component, and then we can drop skip_stages.

.github/workflows/multi_arch_build_portable_linux.yml
lumachad: True. Dead code for now. I'll remove it when I update this PR with whatever is left to do specifically for debug-tools.

.github/workflows/multi_arch_build_portable_linux.yml
lumachad: That would be really nice. If we can fingerprint each component in a build and identify when pieces of pre-built components can be reused, that would speed things up quite a bit. For ROCgdb CI, we only change ROCgdb's sources, so the rest of TheRock is the same.

lumachad: Yes, for sure. I mean, this is a prototype PR before I was made aware of your plans to touch this area. So I'm all for combining your work and debug-tools needs such that we don't duplicate effort or overlap functionality.

lumachad:
Needs a rework. Closing.

PR #5406, opened 2026-05-21
Add GitHub API 'resolve_git_ref' and 'fetch_file_contents' helpers

ScottTodd:
## Motivation

These helpers will be used for pytorch manifest generation in multi-arch CI/CD as part of https://github.com/ROCm/TheRock/issues/5110 by letting us
1. Resolve refs such as:
    * `release/2.10` https://github.com/ROCm/pytorch/tree/release/2.10 --> https://github.com/ROCm/pytorch/commit/3d3aa833db84eed6b7f5595cb5f162c2f78300a4
2. Fetch individual files such as:
    * https://github.com/ROCm/pytorch/blob/3d3aa833db84eed6b7f5595cb5f162c2f78300a4/related_commits
    * https://github.com/ROCm/pytorch/blob/3d3aa833db84eed6b7f5595cb5f162c2f78300a4/version.txt

These APIs are usable without needing to clone the repositories, so manifest generation remains fast (~10 seconds with ~30 API calls and some other overhead).

## Test Plan

* New unit tests
* Used as part of my larger feature branch, test workflow run that calls these new functions: https://github.com/ROCm/TheRock/actions/runs/26252181316/job/77265866122#step:6:32

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/github_actions_api.py
+++ b/build_tools/github_actions/github_actions_api.py
@@ -435,6 +435,9 @@ def gha_query_workflow_run_by_id(github_repository: str, workflow_run_id: str) -
     return gha_send_request(url)
 
 
+# TODO: Consider accepting a git ref (branch/tag) here and resolving it
+# to a SHA via gha_resolve_git_ref, so callers don't need to do that
+# themselves. Same for other functions that take a commit SHA.
 def gha_query_workflow_runs_for_commit(
     github_repository: str,
     workflow_file_name: str,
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/github_actions_api.py
```diff
@@ -531,6 +534,53 @@ def gha_query_recent_branch_commits(
...
+
+    Returns:
+        The full 40-character commit SHA.
+
+    Raises:
+        GitHubAPIError: If the ref cannot be resolved.
+
+    See: https://docs.github.com/en/rest/commits/commits#get-a-commit
+    """
+    url = f"https://api.github.com/repos/{github_repository}/commits/{ref}"
```
erman-gurses: Would it be helpful if we add URL Encoding to prevent to parse extra path segments?

build_tools/github_actions/github_actions_api.py
```diff
@@ -531,6 +534,53 @@ def gha_query_recent_branch_commits(
...
+    Raises:
+        GitHubAPIError: If the file cannot be fetched (not found, API error, etc.).
+
+    See: https://docs.github.com/en/rest/repos/contents#get-repository-content
+    """
+    import base64
+
+    url = f"https://api.github.com/repos/{github_repository}/contents/{path}?ref={ref}"
+    response = gha_send_request(url)
+    return base64.b64decode(response["content"]).decode("utf-8")
```
erman-gurses: The Contents API can return different payloads depending on the path/type, and non-text files or unexpected responses will currently fail with a KeyError or decode error that is harder to diagnose. Might be considered checking fields like "type" / "encoding" and raising a clearer GitHubAPIError when the response is not a regular text file.

ScottTodd: Changed to return the bytes without decoding, letting callers choose how to decode.

build_tools/github_actions/github_actions_api.py
ScottTodd: Sure, added some encoding for extra safety.

erman-gurses (approved):
LGTM

PR #5407, opened 2026-05-21
Expose additional outputs in `build_tools/github_actions/determine_version.py`

ScottTodd:
## Motivation

I want to use this data from some new pytorch manifest generation code in multi-arch CI/CD as part of https://github.com/ROCm/TheRock/issues/5110, so exposing the underlying `version_suffix` in addition to `optional_build_prod_arguments`.

## Test Plan

* New and existing unit tests
* Python package builds/tests on this PR should continue to pass

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/determine_version.py
+++ b/build_tools/github_actions/determine_version.py
@@ -24,60 +24,42 @@ Writing the output to the "GITHUB_ENV" file can be suppressed by passing
 #       together and follow a similar style.
 # TODO: Or share with JAX builds? Could use a different name in that case too.
 
-from packaging.version import parse
-
 import argparse
 import sys
+from pathlib import Path
 
-from github_actions_api import *
+from packaging.version import parse
+
+_BUILD_TOOLS_DIR = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(_BUILD_TOOLS_DIR))
+
+from github_actions.github_actions_api import gha_set_env
 
 
 def derive_version_suffix(rocm_version: str) -> str:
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/determine_version.py
```diff
@@ -102,11 +84,28 @@ def main(argv: list[str]):
...
     print(f"{optional_build_prod_arguments}")
 
     if args.write_env_file:
-        gha_set_env({"optional_build_prod_arguments": optional_build_prod_arguments})
+        gha_set_env(
+            {
+                "optional_build_prod_arguments": optional_build_prod_arguments,
+                "version_suffix": version_suffix,
+            }
+        )
```
erman-gurses, line 108: The existing tests in `determine_version_test.py` only validate `derive_version_suffix()` via:
```
test_dev_version_suffix
test_nightly_version_suffix
test_version_suffix_sorting
```
https://github.com/ROCm/TheRock/blob/ea914cbdb0c680c21c149ddc1d985e5ef67840b2/build_tools/github_actions/tests/determine_version_test.py#L14

Currently there is no assertion verifying that `gha_set_env()` is called with both `optional_build_prod_arguments` and `version_suffix`. Expanding the unit tests to cover the `--write-env-file` path would be good here.

ScottTodd: Added a test (generally I've been avoiding adding tests for github actions side effects in `main()` as that code is deliberately as simple as possible and can only be safely tested on github actions via mocks).

erman-gurses (approved):
LGTM

PR #5410, opened 2026-05-22
Explicitly set INSTALL_RPATH_DIRS for rocminfo

cfreeamd:
Configure rocminfo to explicitly use TheRock's RPATH infrastructure by specifying INSTALL_RPATH_DIRS. This ensures rocminfo binaries have RPATH set to find libraries relative to their installation location.

Without this explicit configuration, rocminfo (which has its own standalone CMakeLists.txt in rocm-systems submodule) doesn't participate in TheRock's automatic RPATH post-processing, potentially causing the binary to search for libraries in system paths instead of the artifact's lib/ directory.

This fix ensures that:
1. rocminfo has RPATH set to $ORIGIN/../lib and $ORIGIN/../lib/rocm_sysdeps/lib
2. The binary will find the correct libhsa-runtime64.so from the same artifact
3. CI tests use the newly built runtime with fixes, not system libraries
4. The solution doesn't rely on LD_LIBRARY_PATH environment configuration

Fixes: ROCM-23909

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -171,6 +171,8 @@ if(THEROCK_ENABLE_CORE_RUNTIME)
       "${_system_toolchain}"
     RUNTIME_DEPS
       ROCR-Runtime
+    INSTALL_RPATH_DIRS
+      lib
   )
   therock_cmake_subproject_glob_c_sources(rocminfo SUBDIRS .)
   therock_cmake_subproject_activate(rocminfo)


cfreeamd:
This is an alternative to the LD_LIBRARY_PATH solution in this PR:
https://github.com/ROCm/TheRock/pull/5366

cfreeamd:
I have verified that rocminfo executable will look in the relative paths that are added with this change.

cfreeamd:
@marbre @nunnikri Please review. This is in response to feedback on https://github.com/ROCm/TheRock/pull/5366.

nunnikri:
> I have verified that rocminfo executable will look in the relative paths that are added with this change.

Can you add the output of readelf -d rocminfo ( generated with this patch). 
The RPATH should be added with relative path TO $ORIGIN. It shouldnt have hard coded path. If its relative path, i think it should be good.

cfreeamd:
> > I have verified that rocminfo executable will look in the relative paths that are added with this change.
> 
> Can you add the output of readelf -d rocminfo ( generated with this patch). The RPATH should be added with relative path TO $ORIGIN. It shouldnt have hard coded path. If its relative path, i think it should be good.

Thanks @nunnikri. See below. I'm not sure why the CI is failing for this patch. I don't think it's related. 

amd@splinter-ctr-av78-4:~/cfreehil/dbg_binaries/bin$  readelf -d ./rocminfo

Dynamic section at offset 0xaa60 contains 31 entries:
  Tag        Type                         Name/Value
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../lib:$ORIGIN/../lib/rocm_sysdeps/lib]
 0x0000000000000001 (NEEDED)             Shared library: [libhsa-runtime64.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000006ffffffb (FLAGS_1)            Flags: PIE
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000007 (RELA)               0x1428
 0x0000000000000008 (RELASZ)             312 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000006ffffff9 (RELACOUNT)          3
 0x0000000000000017 (JMPREL)             0x1560
 0x0000000000000002 (PLTRELSZ)           1488 (bytes)
 0x0000000000000003 (PLTGOT)             0xdcb0
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000006 (SYMTAB)             0x2e8
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0xb24
 0x000000000000000a (STRSZ)              2302 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0xb08
 0x0000000000000019 (INIT_ARRAY)         0xca50
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0xca58
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x000000000000000c (INIT)               0xb634
 0x000000000000000d (FINI)               0xb650
 0x000000006ffffff0 (VERSYM)             0x978
 0x000000006ffffffe (VERNEED)            0xa04
 0x000000006fffffff (VERNEEDNUM)         4
 0x0000000000000000 (NULL)               0x0

core/CMakeLists.txt
```diff
@@ -171,6 +171,9 @@ if(THEROCK_ENABLE_CORE_RUNTIME)
       "${_system_toolchain}"
     RUNTIME_DEPS
       ROCR-Runtime
+    INSTALL_RPATH_DIRS
+      lib
+      lib/rocm_sysdeps/lib
```
marbre: As there is only a runtime dep on rocr-runtime, are sysdeps needed here at all or can we skip adding thos to the rpath?

cfreeamd: The ones in lib/rocm_sysdeps/lib are transitive. They are needed by rocr-runtime.
I suppose they should be handled by rocr and not rocminfo. I can remove them.

cfreeamd: @marbre @nunnikri I made that change.

rocm-systems
ScottTodd, line 1: Careful with your git operations. This submodule should not be modified (much less deleted) in this PR.

cfreeamd: Fixed

nunnikri (approved):

PR #5413, opened 2026-05-22
fetch_test_configurations: coerce container_options to string on non-linux

lumachad:
Introduced in https://github.com/ROCm/TheRock/pull/5132:

_build_container_options early-returned on non-Linux platforms without collapsing the list-form container_options from test_matrix entries (e.g. sanity's ["--cap-add SYS_MODULE", "-v /lib/modules:/lib/modules"]). The raw list survived JSON serialization into the workflow matrix, and test_component.yml's `options: ${{ fromJSON(...).container_options }}` then evaluated to a YAML sequence, failing template parsing:

  test_component.yml (Line: 63, Col: 16): A sequence was not expected

Containers are Linux-only (container.image is already gated on platform == 'linux'), so on Windows the field is dead weight. Collapse it to an empty string so the workflow expression yields a harmless empty options value.

Fixes the following error on the Windows sanity test job:

Error: The template is not valid. ROCm/TheRock/.github/workflows/test_component.yml@ref (Line: 63, Col: 16): A sequence was not expected

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -88,8 +88,12 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
     Returns:
         The modified job_config with updated container_options
     """
-    # Only apply container options for Linux platforms
+    # Containers are Linux-only (test_component.yml gates container.image on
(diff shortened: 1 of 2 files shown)


jayhawk-commits (approved):
lgtm as Windows sanity test template issue is resolved

PR #5422, opened 2026-05-24
exe_stub_gen: use /proc/self/exe instead of dladdr() to locate stub binary

eppaneamd:
## Problem

The POSIX exe stub uses `dladdr(main)` to locate its own path, then resolves the real target binary at a relative path from that directory. This fails when the stub is invoked with a bare filename as argv[0] (no '/' separator): `dladdr()` returns `dli_fname` without a path component, `strrchr` returns NULL, and the stub exits 1:

  could not find path component of main program: 'ld.lld'

The concrete trigger is MLIR's ROCDL `gpu-module-to-binary` pass, which invokes `ld.lld` via `llvm::sys::ExecuteAndWait()` with the absolute path as the executable but hardcodes the bare string `"ld.lld"` as argv[0]: https://github.com/llvm/llvm-project/blob/e4716e0c1155/mlir/lib/Target/LLVM/ROCDL/Target.cpp#L364-L366

This causes every FlyDSL AOT kernel compilation to fail during `setup.py` when building AITER v0.1.14 from source against TheRock wheels, blocking the wheel build entirely with 400+ failed kernels.

## Fix

The fix splits the stub into two platform-specific templates. On Linux, `dladdr(main)` is replaced with `readlink("/proc/self/exe")`, which always returns an absolute path regardless of how the process was invoked. On other POSIX platforms (macOS, BSD), the original `dladdr(main)` path is preserved unchanged, as `/proc/self/exe` is Linux-specific and TheRock supports macOS. Neither branch requires `-ldl`: the Linux template does not use `dladdr`, and macOS/BSD expose `dladdr` in the system library. A truncation check (`len == (ssize_t)sizeof(main_path)`) is added to the Linux path.

## Companion fix

The root cause in MLIR should also be fixed in llvm/llvm-project: passing `lldPath` (already computed as an absolute path at line 393) as argv[0] instead of the bare `"ld.lld"` literal. This PR makes the stub robust against that class of caller regardless of whether the MLIR fix lands first.

### DIFF
--- a/build_tools/_therock_utils/exe_stub_gen.py
+++ b/build_tools/_therock_utils/exe_stub_gen.py
@@ -89,8 +89,9 @@ POSIX_EXE_STUB_TEMPLATE = r"""#define _GNU_SOURCE
 static const char EXEC_RELPATH[] = "@EXEC_RELPATH@";
 
 int main(int argc, char** argv) {
-    // Use the Dl_info of the main program to get the path. This is only valid
-    // because we linked as a PIE executable and the cwd has not been changed.
+    // Use the Dl_info of the main program to get the path. -fPIE is required
+    // so the dynamic linker loads the binary as a position-independent
+    // executable, which allows dladdr() to resolve dli_fname.
     Dl_info info;
     if (!dladdr(main, &info)) {
         fprintf(stderr, "could not get dl info for main: %s\n", dlerror());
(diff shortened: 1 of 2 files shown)


marbre:
@dan-garvey, I haven't look closely but seems this is a duplicate to what you just landed with ebd468c947b3f72e43006b9a9496e01c90924a8e?

dan-garvey:
> @dan-garvey, I haven't look closely but seems this is a duplicate to what you just landed with [ebd468c](https://github.com/ROCm/TheRock/commit/ebd468c947b3f72e43006b9a9496e01c90924a8e)?

It is largely a duplicate. Things it does mine didnt: 
  - splitting Linux vs generic POSIX templates
  - adding relative_link_to validation
  - handling multi-word CC via shlex.split
  - adding different tests around generated source and bare argv[0]

PR #5429, opened 2026-05-25
[hipDNN] ALMIOPEN-1869 Enable hipDNN Python bindings build, wheel validation test, and CI matrix entry

tvy-amd:
## Summary

Enables hipDNN Python bindings+ CI wheel test in TheRock.

## Changes

**Build / packaging**
- `pack_frontend_wheel.py` stages pre-built `hipdnn_frontend/` package next to template `pyproject.toml` + `setup.py`, delegates to `uv pip wheel` (uv-created venvs skip ensurepip).
- `pack_frontend_wheel_setup.py` forces `Distribution.has_ext_modules() -> True` so bdist_wheel tags the wheel `cp{X}{Y}-cp{X}{Y}-<plat>` (nanobind `.so` is ABI-locked, unlike `py3-none-<plat>` rocm-sdk-* templates).

**Artifact split**
- `share/hipdnn/python/**` excluded from lib component (Python tree only ships via wheel, not the shared-library tarball).
- python bindings wheels is not shipped. Only built in tests to test out the bindings

**CI test**
- New `test_hipdnn_frontend_python.py`: builds wheel → creates venv → installs wheel + pytest → runs upstream `share/hipdnn/tests/python` suite.
- Per-step timeouts: wheel build 5m, pip install 5m, pytest 20m.
- Reuses `setup_venv.create_venv` / `find_venv_python_exe` (no hand-rolled venv).
- Reuses new shared `build_rocm_loader_env(artifacts_path)` helper in `libhipcxx_utils.py` (Linux: `LD_LIBRARY_PATH`; Windows: `PATH`).
- pytest collection imports the package itself → no separate import smoke-test needed.
- Waiting for rocm-libraries bump so the CI passes 
- 
**CI matrix**
- New entry runs the test against the merged artifact tree on GPU runners.


build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/libraries_test.py
```diff
@@ -90,7 +90,7 @@ def testSharedLibrariesLoad(self):
                 # in '_rocm_sdk_core/bin'
                 # TODO(#996): track deps in libraries then have the preloader
                 #   recursively get deps instead of hardcoding like this
-                preload_command = "import rocm_sdk; rocm_sdk.preload_libraries('amd_comgr', 'amdhip64', 'hiprtc');"
+                preload_command = "import rocm_sdk; rocm_sdk.preload_libraries('amd_comgr', 'amdhip64', 'hiprtc', 'hipdnn');"
```
tvy-amd:
```
utils.py:40-49. get_module_shared_libraries() — globs **/*.so.* and **/*.so recursively from package root.
Called at module level in libraries_test.py:30:
```

TIL, this test goes and looks for all the shared objects. The hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so gets discovered so it gets loaded but during the dependency check it doesn't find libhipdnn_backend.so so the tests fail. 
By adding this `hipdnn` preload, it loads the libhipdnn_backend.so into the RTLD_GLOBAL so it can be discovered by hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so.

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,168 @@
...
+    pkg_dir = find_pkg_dir(artifacts_path)
+    logging.info(f"Found hipdnn_frontend at: {pkg_dir}")
+
+    with tempfile.TemporaryDirectory() as tmp:
+        tmp_path = Path(tmp)
+        wheel_dir = tmp_path / "wheels"
+        wheel_dir.mkdir()
+        venv_dir = tmp_path / "venv"
+
+        wheel_path = build_wheel(pkg_dir, wheel_dir)
```
tvy-amd, line 145: for now, we will only build the python wheels in this test - just so we can test them. We won't worry about releasing them yet

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/devel_test.py
```diff
@@ -192,8 +192,14 @@ def testSharedLibrariesLoad(self):
                 # Load each in an isolated process because not all libraries in the tree
                 # are designed to load into the same process (i.e. LLVM runtime libs,
                 # etc).
-                command = (
-                    extra_setup + "import ctypes; import sys; ctypes.CDLL(sys.argv[1])"
+                command = extra_setup + "; ".join(
```
tvy-amd: this specific change may not be needed but AI wanted to do it so I let it

rocm-libraries
tvy-amd, line 1: Undo this bump later when we can confirm the tests can run well

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,161 @@
...
+    / "python"
+    / "tests"
+)
+
+logging.basicConfig(level=logging.INFO)
+
+
+def find_pkg_dir(artifacts_path: Path) -> Path:
+    """Locate the hipdnn_frontend package directory in the test artifact."""
+    candidate = artifacts_path / "share" / "hipdnn" / "python" / "hipdnn_frontend"
```
tvy-amd: hipdnn will build bindings and install them in "share/hipdnn/python/hipdnn_frontend" folder.

__init__.py and hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so are the binding files in that folder

rocm-libraries
tvy-amd: rocm-libraries has been bumped in TheRock 🥳 I will undo mine now

ml-libs/CMakeLists.txt
```diff
@@ -175,6 +176,7 @@ if(THEROCK_ENABLE_HIPDNN)
       therock-googletest
       therock-fmt
       therock-nanobind
+      therock-robin-map
```
SamuelReeder, line 179: This didn't come in transitively?

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
SamuelReeder, line 1: Maybe this file should be named `test_hipdnn_python.py`. It ultimately runs the tests and the wheel build is sort of just a prereq step?

ml-libs/CMakeLists.txt
tvy-amd: It did not.  I thought it would've but didn't - There's seems to be strict dependency define here if it's defined in the subproject eg. since we do an explicit robin-map find in hipdnn

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: Yeah I like that. I'll apply that change

build_tools/github_actions/test_executable_scripts/hipdnn/pack_frontend_wheel.py
SamuelReeder, line 1: Nit: thinking the packing scripts should go in a sub-folder of `test_executable_scripts`.

SamuelReeder, line 1: This name should potentially be less generic too. Since it's for hipdnn.

tvy-amd:
FYI I'm running the tests in this PR https://github.com/ROCm/rocm-libraries/pull/7600

Edit: And They look good 🥳 

https://github.com/ROCm/rocm-libraries/actions/runs/26894141207/job/79368556758?pr=7600
```
====================== 103 passed, 17 warnings in 17.05s =======================
INFO:root:All hipDNN Python bindings tests passed!

======================================================================
Resource Summary - Test hipdnn_python_bindings
======================================================================
Duration:     0.5 min (1 samples)
Memory:       8% peak (156.9 GB), 8% avg
CPU:          112/112 cores avg, 112 peak
Storage:      11134 GB min free
======================================================================
Max memory usage was 8%
======================================================================
```

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,191 @@
...
+    validates exactly the artifacts it downloaded. Honors
+    HIPDNN_PYTHON_TESTS_DIR / ROCM_LIBRARIES_DIR for developer overrides
+    pointing at a rocm-libraries checkout.
+    """
+    override = os.getenv("HIPDNN_PYTHON_TESTS_DIR")
+    if override:
+        return Path(override).resolve()
+    rocm_libraries_dir = os.getenv("ROCM_LIBRARIES_DIR")
+    if rocm_libraries_dir:
+        return (Path(rocm_libraries_dir) / _HIPDNN_TESTS_SOURCE_RELPATH).resolve()
```
SamuelReeder: I'm wondering if this path is ever exercised?

SamuelReeder (approved):
LGTM! Just one question.

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: It never did. I remove

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -461,6 +461,18 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
             "windows": 1,
         },
     },
+    # hipDNN Python bindings wheel build + install + pytest
+    "hipdnn_python_bindings": {
+        "job_name": "hipdnn_python_bindings",
+        "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
```
BrianHarrisonAMD, line 490: If we have integration tests, then I suppose this makes sense.

As it stands though, I think you could get by with just --hipdnn --tests.
I don't think we have any E2E tests that require the providers.

tvy-amd: I'm about to land some E2E tests in a rocm-libraries PR so I want to leave this in for now

BrianHarrisonAMD (approved):
LGTM assuming CI passes in the various environments.

build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
```diff
@@ -16,6 +17,22 @@ def prepend_env_path(env: dict, var_name: str, new_path: str):
...
+    ``bin/`` on Windows).
+    """
+    env = os.environ.copy()
+    if platform.system() == "Windows":
+        prepend_env_path(env, "PATH", str(artifacts_path / "bin"))
+    else:
+        prepend_env_path(env, "LD_LIBRARY_PATH", str(artifacts_path / "lib"))
+    return env
+
+
```
geomin12, line 35: is this in the right place?

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+# Copyright (c) Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""
+hipDNN Python bindings wheel test.
+
+This test validates the hipDNN Python wheel packaging pipeline:
+1. Builds a wheel from the staged Python bindings in the test artifact
```
geomin12, line 9: i know this is a test, but these scripts are really designed to be just test wrappers for CI. do we want to include this package into TheRock's build (for future use for hipdnn)?

or is this meant to be tested via wrappers and will not be a part of TheRock build

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,159 @@
...
+
+OUTPUT_ARTIFACTS_DIR = os.getenv("OUTPUT_ARTIFACTS_DIR")
+SCRIPT_DIR = Path(__file__).resolve().parent
+PACK_WHEEL_SCRIPT = SCRIPT_DIR / "hipdnn" / "pack_frontend_wheel.py"
+
+# build_tools/ on sys.path so we can reuse the shared venv helpers instead of
+# re-implementing platform-specific python-exe discovery here.
+_BUILD_TOOLS_DIR = SCRIPT_DIR.parent.parent
+sys.path.insert(0, str(_BUILD_TOOLS_DIR))
+from setup_venv import create_venv, find_venv_python_exe  # noqa: E402
```
geomin12, line 33: `noqa: E402`, clarification?

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
```diff
@@ -0,0 +1,159 @@
...
+    wheel install must be isolated from the test-dependency resolution.
+    numpy is required by the pytest suite for array construction.
+    """
+    subprocess.run(
+        [str(python), "-m", "pip", "install", "--no-deps", str(wheel_path)],
+        check=True,
+        timeout=_TIMEOUT_PIP_INSTALL,
+    )
+    subprocess.run(
+        [str(python), "-m", "pip", "install", "pytest>=7,<9", "numpy"],
```
geomin12, line 96: requirements-test.txt handles? don't think this is needed

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
geomin12, line 9: we also like to avoid building here, as this is on a GPU... (taking up resources for just building which can be done on CPU and GPU resources can be used elsewhere)

geomin12: although i see the tests are minimal time. just a thought :)

tvy-amd: For now, we don't want to package the hipdnn frontend wheels in TheRock - I'm only building them to enable testing.

There are later plans to include them in TheRock build, but there will need to be other wheel considerations when that happens

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: That's from AI asking a question, I need to remove

build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
tvy-amd: It's a general rocm load into your system paths so this doesn't seem out of place.
Only the testing python bindings script uses it atm, I could put it there.
I'm easy on either ideas

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: I stayed away from using the requirements-tests.txt because of download/import speed. There's a lot of unnecessary dependencies that will be pulled in if I use the requirements-tests.txt

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: Yeah, that is another con on doing it this way.
We are only doing it shortterm until we can confirm the bindings are stable and we decide on how we release the frontend python wheels in TheRock

geomin12 (approved):
thanks for clarifying! lgtm

build_tools/github_actions/test_executable_scripts/test_hipdnn_frontend_python.py
tvy-amd: Edit: Ahh It's a linter suppression 
```
  ▎ E402 = "import not at top of file". setup_venv is in
  ▎ build_tools/, not an installed module, so the import only
  ▎ resolves after sys.path.insert(...) runs above it. The
  ▎ import must come after the path mutation, which trips E402
  ▎  — noqa suppresses that single expected case. Comment
  ▎ above already notes why we patch sys.path; can expand it
  ▎ to mention the import ordering if you'd prefer.
```

Issue #5433, opened 2026-05-25
Track targeted CI automation for TheRock submodule bumps


jayhawk-commits:
## Summary

Track the work needed to make TheRock CI more targeted for submodule bump PRs such as `rocm-libraries` and `rocm-systems`.

The automation should use `BUILD_TOPOLOGY.toml` to determine which build stages are affected by source changes, which stages can reuse prebuilt artifacts, and eventually which test components should run.

### 1. Build topology reverse indexes #5434 

Add reusable helpers for mappings such as:

- source set -> artifact groups
- artifact group -> artifacts
- artifact group -> build stages
- artifact -> build stages
- stage -> source sets
- source set -> stages

### 2. Changed path classification

Map changed files to affected source sets using `.gitmodules` and `BUILD_TOPOLOGY.toml`.

Also classify changes that should force a full rebuild, such as topology, build tooling, CMake, workflow, or packaging changes.

### 3. Stage impact planning

Compute:

- stages that must rebuild
- stages eligible for prebuilt artifact reuse
- reasons for each decision

### 4. Baseline run selection #5435

Select and validate a compatible successful baseline run before copying prebuilt artifacts.

Validation should cover workflow success, required artifacts, platform/family compatibility, and stale or ambiguous baselines.

### 5. Dry-run CI integration

Expose computed targeted CI decisions from `configure_multi_arch_ci.py` without changing build behavior yet.

Example outputs:

- `computed_rebuild_stages`
- `computed_prebuilt_stages`
- `computed_baseline_run_id`
- `targeted_ci_reason`

### 6. Opt-in prebuilt stage reuse

Enable stage reuse behind an explicit feature flag, workflow input, or repository variable.

Fallback to full rebuild when no valid baseline is available.

### 7. Test impact mapping

Map impacted topology artifacts/stages to test matrix components and validate the mapping against existing test configuration.

### 8. rocm-libraries and rocm-systems adapters

Connect existing submodule/project classification logic in `rocm-libraries` and `rocm-systems` to the shared TheRock targeting logic.

## Acceptance Criteria

- Targeted CI decisions are visible in dry-run mode before affecting builds.
- CI explains why each stage is rebuilt or reused.
- Prebuilt reuse only happens with a validated successful baseline.
- Missing or invalid baseline data falls back to full rebuild.
- The approach supports both `rocm-libraries` and `rocm-systems` submodule bumps.

PR #5435, opened 2026-05-25
Add baseline run selection helpers

jayhawk-commits:
## Summary

Add behavior-neutral helpers for selecting a successful baseline workflow run with reusable artifacts.

This is groundwork for the targeted CI work tracked in #5433. No workflow behavior changes in this PR.

## Flow

```mermaid
flowchart TD
    A[Required artifact/family pairs] --> B[Query successful workflow runs]
    B --> C[Skip excluded runs]
    C --> D[Create artifact backend for candidate run]
    D --> E[List available artifacts]
    E --> F{All required pairs present?}
    F -- No --> B
    F -- Yes --> G[Return BaselineRun]
    B --> H[No valid candidate]
    H --> I[Return None]
```

## Changes

- Add `RequiredArtifact`, `ArtifactAvailability`, and `BaselineRun` data structures.
- Add helper to query recent successful workflow runs for a workflow and branch.
- Add artifact availability validation for required artifact/family pairs.
- Add baseline selection that skips excluded runs and returns the newest successful run with all required artifacts.
- Add mocked unit tests for run filtering, artifact availability, missing families, and baseline selection.

## Testing

- `python -B build_tools/github_actions/tests/baseline_runs_test.py`
- `pre-commit run --files build_tools/github_actions/baseline_runs.py build_tools/github_actions/tests/baseline_runs_test.py`

### DIFF
--- a/build_tools/github_actions/baseline_runs.py
+++ b/build_tools/github_actions/baseline_runs.py
@@ -2,7 +2,22 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""Helpers for selecting baseline workflow runs for prebuilt artifact reuse."""
+"""Helpers for selecting baseline workflow runs for prebuilt artifact reuse.
+
+Example:
+
+    baseline = select_baseline_run(
+        required_artifacts=[
+            RequiredArtifact("blas", "gfx94X-dcgpu"),
+            RequiredArtifact("base", "generic"),
+        ],
+        platform="linux",
+        exclude_run_ids=[os.environ["GITHUB_RUN_ID"]],
+    )
+    if baseline:
+        # Pass baseline.run_id as the multi-arch CI baseline_run_id.
+        ...
+"""
 
 from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass
@@ -46,8 +61,22 @@ class ArtifactAvailability:
 
 
 @dataclass(frozen=True)
+class WorkflowJobHealth:
+    """Result of checking required workflow jobs in a candidate run."""
+
+    required_name_substrings: tuple[str, ...]
+    matched_job_names: tuple[str, ...]
+    failed_job_names: tuple[str, ...]
+    missing_name_substrings: tuple[str, ...]
+
+    @property
+    def is_valid(self) -> bool:
+        return not self.failed_job_names and not self.missing_name_substrings
+
+
+@dataclass(frozen=True)
 class BaselineRun:
-    """Successful workflow run with artifacts suitable for reuse."""
+    """Workflow run with build jobs and artifacts suitable for reuse."""
 
     run_id: str
     html_url: str
@@ -55,10 +84,12 @@ class BaselineRun:
     branch: str
     workflow_name: str
     platform: str
+    job_health: WorkflowJobHealth
     artifact_availability: ArtifactAvailability
 
 
 ArtifactBackendFactory = Callable[[dict, str, str], ArtifactBackend]
+WorkflowJobsFetcher = Callable[[dict, str], Sequence[dict]]
 
 
 def _dedupe_required_artifacts(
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/baseline_runs.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def query_successful_workflow_runs(
+    *,
+    github_repository: str = "ROCm/TheRock",
+    workflow_name: str = "multi_arch_ci.yml",
+    branch: str = "main",
+    max_runs: int = 20,
+) -> list[dict]:
+    """Query recent successful workflow runs for a workflow and branch."""
```
ScottTodd, line 195: We have a util for this already: https://github.com/ROCm/TheRock/blob/b2082a7fcb5350b28f0685e5057df89ba7f8c7ba/build_tools/github_actions/github_actions_api.py#L477-L500

(would prefer to expand on what is there)

lumachad:
Do you have a practical example of a baseline run being reused?

build_tools/github_actions/baseline_runs.py
```diff
@@ -0,0 +1,265 @@
...
+    if max_runs < 1:
+        raise ValueError("max_runs must be at least 1")
+
+    per_page = min(max_runs, 100)
+    workflow_path = quote(workflow_name, safe="")
+    query = urlencode(
```
geomin12, line 201: this will actually be tricky, as tests may be flaky and builds pass, but the workflow itself may mark as "failed" (and we probably haven't a good run in a while)

geomin12: what i have had to do was query, then dig into jobs for workflow for builds and see health, we can try that and parse for "Build"

build_tools/github_actions/baseline_runs.py
```diff
@@ -0,0 +1,265 @@
...
+    url = (
+        f"https://api.github.com/repos/{github_repository}"
+        f"/actions/workflows/{workflow_path}/runs?{query}"
+    )
+    response = gha_send_request(url)
+    workflow_runs = response.get("workflow_runs", [])
+    return workflow_runs[:max_runs]
+
+
+def create_artifact_backend_for_workflow_run(
```
geomin12, line 265: i know Scott mentioned that we may want to use nightly releases for CI nightly. are we able to add that here to query for both last good known workflow run and also get the most recent good nightly run? maybe with the --latest-release (actually as i am typing this out, --latest-release will work fine)

geomin12 (commented):
few comments

build_tools/github_actions/baseline_runs.py
jayhawk-commits: Good point. I changed the selector to query completed runs and validate the required build jobs plus required artifacts, instead of requiring the whole workflow conclusion to be success. That lets us reuse artifacts from a run where unrelated tests failed, while still rejecting runs where the relevant build jobs failed.

jayhawk-commits: Added workflow job querying plus `required_successful_job_name_substrings`, defaulting to `("Build",)`. The tests now cover selecting a failed workflow with healthy build jobs and skipping a run where the required build job failed.

build_tools/github_actions/baseline_runs.py
jayhawk-commits: I left latest-release selection out of this helper for now since the existing artifact-manager path already covers that source. This PR focuses on workflow-run baselines; we can add a common selector later if we want one API across workflow runs and release artifacts.

jayhawk-commits:
> Do you have a practical example of a baseline run being reused?

One example would be to reuse built artifacts of compilers, runtimes, profilers, and debuggers for a `rocm-libraries` submodule bump PR.

I will kick off a workflow dispatch using this PR's branch with an example. Nevermind. Having trouble getting a dispatch run not skip CI.

geomin12 (approved):
thanks for clarification. lgtm

erman-gurses (approved):
LGTM!

ammallya (approved):
LGTM

build_tools/github_actions/baseline_runs.py
```diff
@@ -0,0 +1,485 @@
...
+    *,
+    backend: ArtifactBackend,
+    required_artifacts: Iterable[RequiredArtifact],
+) -> ArtifactAvailability:
+    """Validate that a backend has archives for each artifact/family pair.
+
+    This mirrors the artifact filename matching used by ``artifact_manager.py``
+    copy/fetch operations. It validates artifact/family presence, not a
+    complete per-component manifest.
+    """
```
ScottTodd, line 308: This could go in https://github.com/ROCm/TheRock/blob/main/build_tools/find_artifacts_for_commit.py instead, then this file could use that module instead of reimplementing its own artifact discovery. See the TODO: https://github.com/ROCm/TheRock/blob/45b8473f3c4143234e8872e52842bf97a7398087/build_tools/find_artifacts_for_commit.py#L107-L110

ScottTodd:
Thanks for this :) tagging https://github.com/ROCm/TheRock/issues/3399 here also

erman-gurses:
> We have a util for this already:
> 
> https://github.com/ROCm/TheRock/blob/b2082a7fcb5350b28f0685e5057df89ba7f8c7ba/build_tools/github_actions/github_actions_api.py#L477-L500
@jayhawk-commits handed over this task to me and @ammallya . I will raise a PR for follow up for the comments.

PR #5440, opened 2026-05-26
Fix pip install failure for rocm metapackage

### LINKED ISSUE #5396: [Issue]: ROCm pip install instruction errors for single arch gfx103x
### Problem Description

When installing ROCm via pip for single arch gfx103X in a docker, the following error occurs:

`(.venv) [root@fa0a90313715 mnt]# python -m pip install --index-url https://repo.amd.com/rocm/whl/gfx103X-all/ "rocm[libraries,devel]"
Looking in indexes: https://repo.amd.com/rocm/whl/gfx103X-all/
Collecting rocm[devel,libraries]
  Downloading https://repo.amd.com/rocm/whl/gfx103X-all/rocm-7.13.0.tar.gz (18 kB)
  Installing build dependencies ... error
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [4 lines of output]
      Looking in indexes: https://repo.amd.com/rocm/whl/gfx103X-all/
      ERROR: Could not find a version that satisfies the requirement setuptools>=70.2.0 (from versions: none)
      ERROR: No matching distribution found for setuptools>=70.2.0
      WARNING: There was an error checking the latest version of pip.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× pip subprocess to install build dependencies did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
WARNING: There was an error checking the latest version of pip.`

### Operating System

RHEL 9.7

### CPU

11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz

### GPU

n/a

### ROCm Version

ROCm 7.13

### ROCm Component

_No response_

### Steps to Reproduce

dnf install sudo wget
sudo dnf update --releasever=9.7 --exclude=\*release\*
sudo dnf install libatomic libquadmath
sudo dnf install python3.11 python3.11-pip
python3.11 -m venv .venv
source .venv/bin/activate
python -m pip install --index-url https://repo.amd.com/rocm/whl/gfx103X-all/ "rocm[libraries,devel]"

_No response_

### Additional Information

This same error appears for the following OSes (docker):
- Ubuntu 24.04.4 
- Ubuntu 22.04.5
- RHEL 10.1
- RHEL 9.7

Muhamed-Husic:
## Motivation

Fixes: #5396 

When users follow the official installation instructions to install `rocm` metapackages using `pip install --index-url https://repo.amd.com/rocm/whl/gfx103X-all/ "rocm[libraries,devel]"` the installation fails with an error stating that `setuptools>=70.2.0` can't be found. This happends because pip creates a temporary virtual environment and attempts to download setuptools from the configured index. However since `--index-url` restricts pip to only search AMD's package index (which does not host standard PyPI build tools like setuptools) the build fails before installation even begins.

**EDIT:** Upon further investigation I noticed that this only affects certain GPU families. For example, installing from gfx90a index succeeds while gfx103X-all fails. The root cause is that for example enterprise gfx90a index mirrors standard PyPI dependencies, allowing pip to download them during build, while the consumer gfx103X-all index strictly hosts AMD pacakges and omits PyPI dependencies.

## Technical Details

This fix ensures that build dependencies are available locally, and tell pip to use those instead of trying to download them.

In `RELEASES.md` the virtual environment setup instructions now include `pip install --upgrade pip setuptools wheel` immediatelly after  creating and activating venv, because [python 3.12+ no longer bundles setuptools](https://docs.python.org/3/whatsnew/3.12.html#:~:text=gh%2D95299%3A%20Do%20not%20pre%2Dinstall%20setuptools%20in%20virtual%20environments%20created%20with%20venv.%20This%20means%20that%20distutils%2C%20setuptools%2C%20pkg_resources%2C%20and%20easy_install%20will%20no%20longer%20available%20by%20default%3B%20to%20access%20these%20run%20pip%20install%20setuptools%20in%20the%20activated%20virtual%20environment.) in newly created venv. Additionally, all `pip install` commands for `rocm` metapackages now include `--no-build-isolation` flag, which instructs pip to use locally installed `setuptools` rather than attempting to fetch it from AMD index.

In `build_tools/setup_venv.py` the `update_venv` function now installs setuptools and wheel alongside the existing pip upgrade. `install_packages_into_venv` function was updated to automatically append `--no-build-isolation` to the pip command whenever a package name statring with `rocm` is detected.

In `build_tools/github_actions/upload_python_packages.py` the string templates that generate GitHub Actions step summaries were updated to include `--no-build-isolation` in all generated `pip install` snippets.

## Test Plan

I set up clean Ubuntu DOcker container and ran the installation. The test was performed in fresh venv with no pre-existing packages (except for those that `python3 -m venv` provides).

## Test Result

Without the fix we get immediate error: `ERROR: Could not find a version that satisfies the requirement setuptools>=70.2.0`. With the fix applied, running the updated sequence (`pip install --upgrade pip setuptools wheel` followed by `pip install --no-build-isolation --index-url https://repo.amd.com/rocm/whl/gfx103X-all/ "rocm[libraries,devel]"`) completes successfully, with the following result: `Successfully installed rocm-7.13.0 rocm-sdk-core-7.13.0 rocm-sdk-devel-7.13.0 rocm-sdk-libraries-gfx103X-all-7.13.0`.

## Submission Checklist

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -101,6 +101,7 @@ index at https://rocm.nightlies.amd.com/whl-multi-arch/.
 > ```bash
 > python -m venv .venv
 > source .venv/bin/activate
+> pip install --upgrade pip setuptools wheel
 > ```
 >
 > Multiple virtual environments can be present on a system at a time, allowing you to switch between them at will.
@@ -139,7 +140,7 @@ Install ROCm with device support for your GPU using the unified index:
 
 ```bash
 # Replace device-gfx942 with your GPU, see the section below for details
-pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ "rocm[libraries,device-gfx942]"
+pip install --no-build-isolation --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ "rocm[libraries,device-gfx942]"
 ```
 
 <!-- TODO: Advertise wheel variants / WheelNext once available  -->
@@ -215,7 +216,7 @@ Install PyTorch with ROCm support using the same unified index:
 # Replace device-gfx942 with your GPU, see the section above for details
 # Note: we'll recommend 'whl-multi-arch' instead of 'whl-staging-multi-arch'
 #       as soon as we test run automate tests on these packages
-pip install --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
+pip install --no-build-isolation --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
     "torch[device-gfx942]" "torchvision[device-gfx942]" torchaudio
 
 # Optional additional packages on Linux:
@@ -229,7 +230,7 @@ pip install --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
 > install ROCm separately:
 >
 > ```bash
-> pip install --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
+> pip install --no-build-isolation --index-url https://rocm.nightlies.amd.com/whl-staging-multi-arch/ \
 >     "torch[device-gfx1100]"
 >
 > pip freeze  # with approximate download sizes:
@@ -424,6 +425,7 @@ We currently support Python 3.10, 3.11, 3.12, 3.13, and 3.14 (PyTorch 2.9+ only)
 > ```bash
 > python -m venv .venv
 > source .venv/bin/activate
+> pip install --upgrade pip setuptools wheel
 > ```
 >
 > Multiple virtual environments can be present on a system at a time, allowing you to switch between them at will.
@@ -494,7 +496,7 @@ A new optional package `rocm-profiler` is available, providing ROCm profiling to
 Install profiling tools via the meta package:
 
 ```bash
-pip install "rocm[profiler]"
+pip install --no-build-isolation "rocm[profiler]"
 ```
 
 This will install:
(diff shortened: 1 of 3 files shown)


marbre (commented):
Hi @Muhamed-Husic,

Thanks for flagging this and raising a PR. What you propose would work but the actual fix is to update the indexes on our side and make sure the dependencies got properly uploaded,

Muhamed-Husic:
> Hi @Muhamed-Husic,
> 
> Thanks for flagging this and raising a PR. What you propose would work but the actual fix is to update the indexes on our side and make sure the dependencies got properly uploaded,

Hi, thanks for clarification, that makes sense. Should I close this PR then, or do you think it is still worth keeping `--no-build-isolation` and setuptools pre-install in the venv setup instructions as a safeguard?

marbre:
> > Hi @Muhamed-Husic,
> > Thanks for flagging this and raising a PR. What you propose would work but the actual fix is to update the indexes on our side and make sure the dependencies got properly uploaded,
> 
> Hi, thanks for clarification, that makes sense. Should I close this PR then, or do you think it is still worth keeping `--no-build-isolation` and setuptools pre-install in the venv setup instructions as a safeguard?

@ScottTodd wdyt?

ScottTodd:
Let's leave off additional flags like `--no-build-isolation` or steps like `pip install --upgrade pip setuptools wheel`. The simplest install instructions must work. If they don't (such as when packages are missing from our of our indices), that's the issue that needs _fixing_, not working around.

marbre:
Agreeing with @ScottTodd here. Anyway, thanks for proposing a fix @Muhamed-Husic, apprechiated! @araravik-psd is on it to upload the missing deps.

PR #5445, opened 2026-05-26
[ck][cmake] Suppress all lifetime warnings as errors

ronlieb:
adds
  -Wno-lifetime-safety-intra-tu-suggestions
  -Wno-lifetime-safety-cross-tu-suggestions
  -Wno-lifetime-safety-lifetimebound-violation
  -Wno-unknown-warning-option

## Motivation

suppresses the lifetime-safety warnings to allow newer compiler SMP bumps to not break when compiling somwhat older CK


patches/amd-mainline/rocm-libraries/0001-ck-cmake-Suppress-all-lifetime-warnings-as-errors.patch
ScottTodd, line 1: Is this commit already up for review in rocm-libraries? New patches to the default `main` branch are strongly discouraged per https://github.com/ROCm/TheRock/blob/main/patches/README.md. You can carry patches as needed on non-default branches (e.g. branches for the compiler) but the costs of adding patches here on the default branch are quite high.

ronlieb: illia has commit up for review here: https://github.com/ROCm/rocm-libraries/pull/7863

PR #5449, opened 2026-05-26
[ci] Introducing specific build selection with `--projects` flag

geomin12:
As we roll out multi-arch CI to superrepos and continue to improve, we require a "build selection" tool in cases where we do not have to build everything (particular scenarios are superrepos).

Changes:
- Additional input of `--projects` flag in `build_tools/configure_stage.py`, where if --projects is added with --stage, it will output the necessary build flags (ex: `configure_stage.py --stage math-libs --projects rocblas => -DENABLE_BLAS=ON -DENABLE_ALL=OFF`)
- Adding a cmake parse section where it parses https://github.com/ROCm/TheRock/blob/ee810570e9c330241e88e0b2370ef6b8d08ff18f/math-libs/CMakeLists.txt#L221 to its corresponding artifact
- a python script that generates the cmake parsing -> JSON subproject mapping (we also add unit tests to ensure it doesn't break when projects are added / removed)
- this manifest is generated and not frequently updated, so the test checks this (while we continue to use it). if we need to update, we just re-generate
- Additional methods in `build_tools/_therock_utils/build_topology.py` to pull data from bullet point above

This is just the introduction of the script, where next steps are to 
- check this into superrepo multi arch CI and test it (i've only tested locally so far and unit tests)
- after it works with tests properly, we will introduce the "last known good run" script to help with `copy_prebuilt_stages`
- then, we can use this with submodule bumps to determine what needs to build + test (so we don't waste unnecessary resources)

Progress on #3343


geomin12:
going to get reviews from Joseph first, then later add on Scott / Marius after first pass

build_tools/tests/configure_stage_test.py
```diff
@@ -22,27 +22,225 @@ class ProjectResolutionTest(unittest.TestCase):
...
+        args = self._get_flags(["cuid"])
+        self.assertIn("-DTHEROCK_ENABLE_RDC=ON", args)
+
+    def test_rocdbgapi(self):
+        args = self._get_flags(["rocdbgapi"])
+        self.assertIn("-DTHEROCK_ENABLE_AMD_DBGAPI=ON", args)
+
+    def test_amd_dbgapi(self):
+        args = self._get_flags(["amd-dbgapi"])
+        self.assertIn("-DTHEROCK_ENABLE_AMD_DBGAPI=ON", args)
```
lumachad: Why would we need this seemingly duplicate entry?

build_tools/_therock_utils/build_topology.py
```diff
@@ -865,3 +869,33 @@ def get_python_requires_for_stage(self, build_stage: str) -> List[str]:
...
+
+    def get_alias_to_artifact_map(self) -> Dict[str, str]:
+        """Map subproject names to artifact names (e.g., 'rocblas' -> 'blas')."""
+        alias_map: Dict[str, str] = {}
+        for artifact in self.artifacts.values():
+            alias_map[artifact.name.lower()] = artifact.name
+            for alias in artifact.subprojects:
+                alias_map[alias.lower()] = artifact.name
+            for db_name in artifact.split_databases:
+                alias_map[db_name.lower()] = artifact.name
```
jayhawk-commits, line 905: There is a flaw in this logic when it comes to `hipsparselt`.
- `hipsparselt` is guarded by `THEROCK_ENABLE_SPARSE` at https://github.com/ROCm/TheRock/blob/main/math-libs/BLAS/CMakeLists.txt#L249C4-L249C25 
-  However, `hipsparselt` is a `split_database` entry for `blas`.
- Thus, `configure_stage.py --stage math-libs --projects hipsparselt --platform linux --oneline` would output:
`-DTHEROCK_ENABLE_ALL=OFF -DTHEROCK_ENABLE_BLAS=ON` and this is missing `THEROCK_ENABLE_SPARSE` to actually build hipsparselt.

build_tools/configure_stage.py
```diff
@@ -246,11 +278,27 @@ def main(argv: List[str] = None):
...
+    if args.stage and args.stage not in topology.build_stages:
         available = ", ".join(s.name for s in topology.get_build_stages())
         parser.error(f"Unknown stage '{args.stage}'. Available stages: {available}")
 
+    # Validate projects if provided
+    if args.projects:
+        alias_map = topology.get_alias_to_artifact_map()
+        unknown = [p for p in args.projects if p.lower() not in alias_map]
+        if unknown:
+            log(f"Warning: Unknown project(s): {', '.join(unknown)}")
```
jayhawk-commits: I don't think `unknown` case should be a warning, but instead a fast fail, as a green CI signal for the step could be misleading.

jayhawk-commits (commented):
Commented on a few issues.

build_tools/_therock_utils/build_topology.py
geomin12: we actually added it here: https://github.com/ROCm/TheRock/commit/57ed73e30e7ad81f14b3a5d4b6e69241b0c65e72 to fix a test issue, but I think we added it to the wrong section (despite it working). Moving it to the correct section and going to make sure tests pass

build_tools/tests/configure_stage_test.py
geomin12: not duplicate! they testing two names and making sure they map to correct cmake flag

lumachad: Got it. Thanks for clarifying it.

build_tools/tests/configure_stage_test.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+    @unittest.skipIf(IS_WINDOWS, "rocprofiler-compute disabled on Windows")
+    def test_rocprofiler_compute(self):
+        self.assertIn(
+            "-DTHEROCK_ENABLE_ROCPROFILER_COMPUTE=ON",
+            self._get_flags(["rocprofiler-compute"]),
+        )
+
+    # rocm-systems: debug/dc tools
```
lumachad: Not sure dc tools are related to this. Do we also need to include rocr-debug-agent and rocr-debug-agent-tests?

BUILD_TOPOLOGY.toml
lumachad, line 762: Is there a reason why we don't need a subprojects entry in rocr-debug-agent-tests?

BUILD_TOPOLOGY.toml
```diff
@@ -744,13 +778,15 @@ artifact_group = "debug-tools"
 type = "target-neutral"
 artifact_deps = ["amd-llvm", "sysdeps"]
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
+subprojects = ["amd-dbgapi", "rocdbgapi"]
```
lumachad: I still find this somewhat confusing. We should probably clarify why we have two names here.

BUILD_TOPOLOGY.toml
```diff
@@ -765,6 +801,7 @@ type = "target-neutral"
 artifact_deps = ["amd-llvm", "amd-dbgapi", "sysdeps-gmp", "sysdeps-mpfr", "sysdeps-expat", "sysdeps-ncurses"]
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
 disable_platforms = ["windows"]
+subprojects = ["rocgdb"]
```
lumachad: Seeing how these names are mostly repetitions of the artifacts, could we figure these out automagically as opposed to having these fields? Maybe make sure things are named correctly so we can just auto-discover the subprojects.

geomin12: good point! I think subprojects is a good way to determine but I am looking into a way to dynamically generate this (as I am also not a fan of having these fields)

geomin12:
doing some testing, don't want to take up resources so closing PR

BUILD_TOPOLOGY.toml
geomin12: I added the automagically generation. take a look!

lumachad: Thanks. The output in the json file looks sane to me.

build_tools/configure_stage.py
jayhawk-commits: Thanks for changing this to fail-fast.

build_tools/_therock_utils/build_topology.py
jayhawk-commits: hipsparselt seems to be producing the correct output with the script, but rocsparse and hipsparse do not appear to be exhibiting the correct behaviour for their build flags.

geomin12:
Handled in #5626 . closing

PR #5450, opened 2026-05-26
[ci] Improving multi arch CI for external repos 

geomin12:
As the multi arch CI fails in rocm-libraries from other workflows (such as validate package, python builds, etc), this PR addresses those errors. Also in order for `Copy prebuilt stage artifacts` to work for external, we require IAM role ([errors from sample run](https://github.com/ROCm/rocm-libraries/actions/runs/26469059871/job/77937255846#step:6:597))

Tested here: https://github.com/ROCm/rocm-libraries/actions/runs/26520802785

This is another iteration as we continue to find and fix errors

### DIFF
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -46,9 +46,9 @@ class S3BucketConfig:
 
 
 s3_bucket_configs = [
-    # CI (self-hosted runners include credentials for therock-ci-artifacts-external)
+    # CI
     S3BucketConfig("therock-ci-artifacts", iam_role="therock-ci"),
-    S3BucketConfig("therock-ci-artifacts-external", iam_role=None),
+    S3BucketConfig("therock-ci-artifacts-external", iam_role="therock-ci-external"),
     # Release type "dev"
     S3BucketConfig("therock-dev-artifacts", iam_role="therock-dev"),
     S3BucketConfig("therock-dev-packages", iam_role="therock-dev"),
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -48,7 +48,7 @@ class S3BucketConfig:
 s3_bucket_configs = [
     # CI (self-hosted runners include credentials for therock-ci-artifacts-external)
     S3BucketConfig("therock-ci-artifacts", iam_role="therock-ci"),
-    S3BucketConfig("therock-ci-artifacts-external", iam_role=None),
+    S3BucketConfig("therock-ci-artifacts-external", iam_role="therock-ci-external"),
     # Release type "dev"
     S3BucketConfig("therock-dev-artifacts", iam_role="therock-dev"),
     S3BucketConfig("therock-dev-packages", iam_role="therock-dev"),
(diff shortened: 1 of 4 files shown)


build_tools/_therock_utils/s3_buckets.py
```diff
@@ -48,7 +48,7 @@ def write_access_iam_role(self) -> str | None:
 s3_bucket_configs = [
     # CI (self-hosted runners include credentials for therock-ci-artifacts-external)
     S3BucketConfig("therock-ci-artifacts", iam_role="therock-ci"),
-    S3BucketConfig("therock-ci-artifacts-external", iam_role=None),
+    S3BucketConfig("therock-ci-artifacts-external", iam_role="therock-ci-external"),
     # Release type "dev"
```
ScottTodd, line 52: The comment here was explaining why `iam_role=None` was there. Either change the comment or revert this change.

I bet it's because "Copy Prebuilt Stages" runs outside of Docker and doesn't have the credentials file mounted / at a discoverable path...

geomin12: oh I can probably fix that copy prebuilt stages item then

tony-davis (approved):
Looks good for the external-repo plumbing layer. The copy-prebuilt OIDC fix is validated (221/221 artifacts copied on the rocm-libraries run vs. anonymous AccessDenied before), and the `repository`/`ref` propagation matches what downstream workflows need.

+1 on Scott's note about the `s3_buckets.py` comment — a short update on why external copy-prebuilt uses OIDC now would help the next person. Not blocking from my side.

tony-davis:
Just want to point out that this is in service of Issue https://github.com/ROCm/TheRock/issues/3343.

build_tools/_therock_utils/s3_buckets.py
ScottTodd: This change probably broke CI runs here in TheRock from forks, see https://github.com/ROCm/TheRock/pull/5531 and https://github.com/ROCm/TheRock/actions/runs/26984086713?pr=5531

Linux AWS: `Error: Could not assume role with user credentials: User: arn:aws:sts::324352301041:assumed-role/therock-runners-prod-runner-s3-external/aws-sdk-js-session-1780614257367 is not authorized to perform: sts:TagSession on resource: arn:aws:iam::692859939525:role/therock-ci-external`

Windows Azure: `Error: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::692859939525:role/therock-ci-external`

The fix I was suggesting was to mount the credentials file in https://github.com/ROCm/TheRock/blob/ee2040338f111a54d5bffa9335ee28114ec2b2f2/.github/workflows/multi_arch_ci_linux.yml#L50-L85 , not to change the code here.

ScottTodd: Filed https://github.com/ROCm/TheRock/issues/5654 to follow up

ScottTodd: PRs from forks are still broken. Sent https://github.com/ROCm/TheRock/pull/5698 to revert.

PR #5458, opened 2026-05-27
[miopen-provider]: always print test output for miopen provider

adickin-amd:
## Motivation

There appears to be a random issue where were timing out miopen tests on CI.  We want to always print the output so we can try and track this down as we cant reproduce locally.


adickin-amd:
we dont need this anymore

PR #5399, opened 2026-05-21
[rocprim] change incorrect rocthrust -> rocprim in resource spec executable path

amd-jmahovsky:
## Motivation

The rocprim tests were looking in the rocthrust directory for the resource spec executable.

## Technical Details

Change incorrect rocthrust -> rocprim in resource spec executable path in test_rocprim.py.

## Test Plan

Verify the rocprim unit tests are running.

## Test Result

rocprim unit tests should run instead of silently failing.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5470, opened 2026-05-27
[ci] Add resource spec file generation before running rocPrim ctest

amd-jmahovsky:
## Motivation

When running ctest in parallel mode, a resource spec file must be provided. This PR adds the missing step of generating the resource spec file before running rocPrim ctest.

## Technical Details

The rocPrim library recently added an executable called "generate_resource_spec" to generate the resource spec file for ctest. This file describes the GPUs available on the system and it should be passed to ctest through the --resource-spec-file option whenever the --parallel or the -j option is used. Otherwise tests may fail.

With this fix, we no longer need the ctest option --repeat until-pass:6. Test results should be deterministic.

## Test Plan

Verify that the script of running ctest on rocPrim now calls "generate_resource_spec" and passes the generated file to the --resource-spec-file option of ctest.

## Test Result

Test output shows that "generate_resource_spec" is executed and the --resource-spec-file option is added when running test_rocprim.py.

## Submission Checklist


amd-jmahovsky:
This PR consolidates two other PRs:
https://github.com/ROCm/TheRock/pull/5060 (which was reverted)
https://github.com/ROCm/TheRock/pull/5399 (which fixes a bug in 5060)
This allows me to push the entire correct code through CI at once.

geomin12 (commented):
the core42 runners ran into a regression. can you rebase and re-run tests pls?

geomin12 (approved):
the windows gfx110X runners are flaky, so that error is fine

lgtm

Issue #5459, opened 2026-05-27
rocshmem warnings on torch import


stellaraccident:
Clean install on gfx1151/linux and import torch prints the following warnings/errors to the console (probably not at all related to which gfx arch is used):

```
>>> import torch
W-001h rocSHMEM Could not open libibverbs. Disabled.    IBVWrapper@src/gda/ibv_wrapper.cpp:51
E-001h rocSHMEM Could not open libnuma. Returning       NUMAWrapper@src/gda/numa_wrapper.cpp:48
```

Repro:

```
$ pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ "torch[device-gfx1151]"
$ python -c "import torch"
```

stellaraccident:
@ROCm/rocm-genies can one of you assign this to the rocshmem team to get fixed. Two issues:

1. Shouldn't be dumping console logs like this on library load.
2. Probably should be able to find libnuma by rpath (the second line)

stellaraccident:
Seems related to #5355 but that commentary looks: a) wrong/incomplete, and b) found in automation vs an end-user report, so please make this the primary issue.

ScottTodd:
See also https://github.com/ROCm/rocm-systems/pull/5918, cc @Yiltan (author), @abouteiller (reviewer), @edgargabriel (reviewer).

* libnuma should come from our bundled https://github.com/ROCm/TheRock/tree/main/third-party/sysdeps/linux/numactl (see https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md#numactl), without relying on any form of system install
* library code should not report to stdout/stderr like this unless some form of verbose logging is enabled

abouteiller:
The warnings and error messages can be disabled by changing ROCSHMEM_DEBUG_LEVEL=none.
* the default is warn
* We could change the default to error
* I would recommend we do not change the default to none as that will make understanding why their program stopped difficult for users.
* Will need to investigate if missing libnuma during init is a condition that justify erroring rather than just warning and not initializing, letting the rest of the program continue without rocshmem functionality.

edgargabriel:
We already made a change in rocSHMEM that turned not finding libnuma into a warning instead of a fatal error, similarly to how the libibverbs library is treated. If libnuma is not found, the gda backend is disabled, but single-node jobs using ROCm IPC would still be able to work correctly. 

This change is already in rocSHMEM develop as of today, but it will probably be 7.14 until its in a public release.

[...]

Issue #2518, opened 2025-12-11
[Issue]: libatomic installation needed on certain distros for rocminfo to work following tarball install for ROCm 7.10


lsudarsh-amd:
### Problem Description

On clean installations of distros like RHEL 8.10, 10.0 and SLES 15.7, after installing ROCm 7.10 using tarball method, commands like rocminfo do not work due to a missing libatomic dependency. If libatomic is installed after the fact, then rocminfo works. This only appears to affect these 3 distros, the other supported distros do not need libatomic to be explicitly installed for things to work as expected.

Here is a log of the failure:
rocminfo
rocminfo: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

### Operating System

RHEL 8.10, 10.0 and SLES 15.7

### CPU

Intel Core i7-10700

### GPU

AMD Radeon RX 7900 XT

### ROCm Version

ROCm 7.10 Prerelease

### Steps to Reproduce

1. Clean OS install of RHEL 8.10, 10.0 or SLES 15.7 from distro ISO

2. Follow the steps to install tarball
sudo usermod -a -G render,video $LOGNAME
mkdir therock-tarball && cd therock-tarball

wget https://repo.amd.com/tarball/therock-dist-linux-gfx950-dcgpu-7.10.0rc2.tar.gz

mkdir install
tar -xf *.tar.gz -C install

cd therock-tarball
ROCM_INSTALL_PATH=$(pwd)/install
sudo tee /etc/profile.d/set-rocm-env.sh << EOF
export ROCM_PATH=$ROCM_INSTALL_PATH
export PATH=$PATH:$ROCM_PATH/bin
export LD_LIBRARY_PATH=$ROCM_PATH/lib
EOF
sudo chmod +x /etc/profile.d/set-rocm-env.sh

3. Reboot system for all changes to take effect

4. Run rocminfo

stellaraccident:
Thanks. We will need to track this down. @ROCm/rocm-genies for attention

darren-amd:
Hi @lsudarsh-amd,

Thanks for reporting this issue! I was able to reproduce the issue both on RHEL 8.10 and inside of a Ubuntu 22.04 and 24.04 container. `libatomic1` is pulled in when `gcc` is installed; I had initially assumed this was bundled with Ubuntu by default but it seems it isn't.

[...]

PR #5474, opened 2026-05-27
Move rocshmem numactl from BUILD_DEPS to RUNTIME_DEPS

harkgill-amd:
## Motivation

1/2 fix for https://github.com/ROCm/TheRock/issues/5459. 

libnuma should be bundled in the dist artifact so rocSHMEM can find it at runtime via rpath, preventing the below error on systems that don't have system installed numa 
```
E-001h rocSHMEM Could not open libnuma. Returning       NUMAWrapper@src/gda/numa_wrapper.cpp:48
```

## Technical Details

Move rocshmem numactl from BUILD_DEPS to RUNTIME_DEPS so it's shipped and found correctly.

## Test Plan

- Copy libnuma.so.1 into torch/lib/ and create the unversioned libnuma.so symlink, mimicking how `RUNTIME_DEP` shipped numa would exist 
- Confirm dlopen("libnuma.so") then succeeds and the E-001h error disappears

## Test Result

- Error is no longer seen as libnuma is found correctly.

## Submission Checklist

### DIFF
--- a/comm-libs/CMakeLists.txt
+++ b/comm-libs/CMakeLists.txt
@@ -164,9 +164,9 @@ if(THEROCK_ENABLE_ROCSHMEM)
       rocm-cmake
       therock-fmt
       therock-googletest
-      ${THEROCK_BUNDLED_NUMACTL}
     RUNTIME_DEPS
       hip-clr
+      ${THEROCK_BUNDLED_NUMACTL}
   )
   therock_cmake_subproject_glob_c_sources(rocshmem
     SUBDIRS


abouteiller (approved):
Assuming RUNTIME_DEPS are also avail as BUILD_DEPS.

marbre:
> Assuming RUNTIME_DEPS are also avail as BUILD_DEPS.

https://github.com/ROCm/TheRock/blob/main/docs/development/build_system.md#terminology

ScottTodd:
> ## Test Plan
> 
> * Copy libnuma.so.1 into torch/lib/ and create the unversioned libnuma.so symlink, mimicking how `RUNTIME_DEP` shipped numa would exist
> * Confirm dlopen("libnuma.so") then succeeds and the E-001h error disappears

Can you also check with the rocm + torch packages built by the CI run on this PR? From https://github.com/ROCm/TheRock/actions/runs/26521858577/job/78147299537?pr=5474, the index is https://therock-ci-artifacts.s3.amazonaws.com/26521858577-linux/python/index.html

harkgill-amd:
> Can you also check with the rocm + torch packages built by the CI run on this PR?

`python -c "import torch"` is passing with no warnings/errors but that shouldn't be the case as this PR only fixes 1/2 of the issues. 

Upon closer inspection, `torch/lib/libtorch_rocshmem.so ` is missing from the wheels entirely, looks like rocshmem isn't enabled in the CI build. Does it make sense to just run a local build to test at this point?

ScottTodd:
> Upon closer inspection, `torch/lib/libtorch_rocshmem.so ` is missing from the wheels entirely, looks like rocshmem isn't enabled in the CI build.

Ah, I bet since we're building release/2.10 and https://github.com/pytorch/pytorch/commit/e2584b2554d11fda4998a8d2be6145b0eded5049 only just landed in 2.12. We may want to switch to building a more recent stable torch release on CI (with opt-in for older releases).

> Does it make sense to just run a local build to test at this point?

Ideally we'd be able to rely on presubmit coverage here. I'd say if the rocshmem tests are still passing that's better signal than nothing, but still incomplete. Your local tests may be sufficient.

ScottTodd:
Well I just landed a new workflow that can build a single multi-arch torch package on demand (instead of needing to build for all python and pytorch versions or being limited to just torch and not torchaudio/torchvision/etc.): https://github.com/ROCm/TheRock/actions/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml. I want to test it anyways, so I'll trigger a release/2.12 build from the packages on this PR: https://github.com/ROCm/TheRock/actions/runs/26540807233 (this'll be for gfx1151 and gfx942)

ScottTodd:
How does this package from https://github.com/ROCm/TheRock/actions/runs/26540807233/job/78181525788 look? (that should be testing this PR if I triggered the jobs correctly)
`pip install --index-url https://rocm.devreleases.amd.com/whl-multi-arch/ --find-links https://therock-ci-artifacts.s3.amazonaws.com/26521858577-linux/python/index.html torch==2.12.0+devrocm7.14.0.dev0.44d3df046ac29fb3338afa0dad3ab781e9e470e6`
(I'm on Windows and thus can't repro myself easily)

harkgill-amd:
Still seeing the warning/errors with the packages from https://github.com/ROCm/TheRock/actions/runs/26540807233/job/78181525788. It was run against [main](https://github.com/ROCm/TheRock/commit/fcdcea4cf9beb4e984359979d549aaab81181491) so I don't think this would have this PR change in it. 

I can rebase and pull in the workflow to run against this branch but wanted to confirm - the workflow pulls in prebuilt ROCm wheels to build torch packages but to my understanding, this PR would require a fresh rebuild of TheRock/rocshmem then torch ontop in order to be tested. Is this correct or is just the torch build ok for testing?

ScottTodd:
The workflow run I linked used rocm python packages from this PR, the commit used would just affect the workflow code and pytorch build script.

harkgill-amd:
Ok so while this change is technically correct, it's not enough for libnuma to be found at runtime. 

libtorch_rocshmem.so rpath points to `torch/lib/` but TheRock bundled numa library exists at `_rocm_sdk_core/lib/rocm_sysdeps/lib/librocm_sysdeps_numa.so.1`. There's a couple different ways we can go about finding the library but the easiest/dirtiest would be a fallback in rocSHMEM to dlopen librocm_sysdeps_numa.so in case libnuma.so fails. Worth noting that the libnuma.so symlink from [numactl/patch_install.sh](https://github.com/ROCm/TheRock/blob/main/third-party/sysdeps/linux/numactl/patch_install.sh#L57) get's dropped during wheel packaging so librocm_sysdeps_numa.so.1 is all we have without creating a new symlink.

@abouteiller @ScottTodd, any preference on how we should handle this case?

ScottTodd:
> libtorch_rocshmem.so rpath points to `torch/lib/` but TheRock bundled numa library exists at `_rocm_sdk_core/lib/rocm_sysdeps/lib/librocm_sysdeps_numa.so.1`.

Would a preload here help? https://github.com/ROCm/TheRock/blob/2555d629204cb066d6668823ad6241d2fcf250e4/external-builds/pytorch/build_prod_wheels.py#L161-L184

Similar to what we already have for `rocm_sysdeps_liblzma`

harkgill-amd:
Preload helps but we'll still run into the same issue where rocSHMEM is looking for libnuma.so but we have librocm_sysdeps_numa. If we pair the preload with the dlopen fallback, that should work.

abouteiller:
We may also consider no dlopening libnuma at all and rely on normal linking then since its part of the rocm package anyhow. Let me gather feedback from the team.

marbre:
Regarding the libnuma inclusion, please also see the discussion related to the underlying deps: https://github.com/ROCm/TheRock/issues/2518

harkgill-amd:
@ScottTodd, I think this PR should still go in while https://github.com/ROCm/rocm-systems/pull/6640 tackles the rocshmem library linking behavior in parallel.

ScottTodd (approved):

PR #5479, opened 2026-05-27
Generate sqlite3 package version file for find_package

marbre:
The sqlite3 3rd party dep installed `sqlite3-config.cmake` but no companion `sqlite3-config-version.cmake`. Without the version file, any consumer calling find_package(SQLite3 <ver>) with any version constraint silently rejects the package. CMake requires both files to satisfy a versioned find.

This patch introduces SQLITE3_VERSION (semantic) and SQLITE3_AMALGAMATION_VERSION (the MMNNPP00 from the download URL), which both need to be touched on a version bump. The tradeoff seems acceptable as the logic to derive the amalgamation version is rather bloated.


ScottTodd (approved):

marbre:
Verified thar https://therock-ci-artifacts.s3.amazonaws.com/26539047906-linux/sysdeps_dev_generic.tar.zst now contains a valid version file.

Issue #102, opened 2025-02-18
CMake HIP language support requires undocumented configuration


stellaraccident:
The documentation at https://rocm.docs.amd.com/en/latest/conceptual/cmake-packages.html#using-hip-in-cmake is incomplete for an SDK not installed at /opt/rocm.

I experimentally determined that the following three cache variables must be defined to function:

```
-DCMAKE_HIP_PLATFORM=amd
"-DCMAKE_HIP_COMPILER=${THEROCK_BINARY_DIR}/dist/rocm/lib/llvm/bin/clang++"
"-DCMAKE_HIP_COMPILER_ROCM_ROOT=${THEROCK_BINARY_DIR}/dist/rocm"
```

The documentation does call out that the third should be a catch-all, but I found that the others were required in practice.

lumachad:
Fix candidate in #5489. It's been approved, but I'd like second opinions.

PR #5489, opened 2026-05-28
fix(cmake): pin HIP compiler variables in amd-hip toolchain file

lumachad:
## Motivation

Guarantee that subprojects using `COMPILER_TOOLCHAIN amd-hip` always build
with TheRock's own HIP compiler rather than a system-discovered one.

## Technical Details

Subprojects that call `project(... LANGUAGES HIP)` or `enable_language(HIP)`
trigger `CMakeDetermineHIPCompiler`, which scans the system and silently picks
up `/opt/rocm-*/bin/*` when a system ROCm install is present. Inside the
manylinux build container this was latent; on a developer machine with
`/opt/rocm-7.3` installed it caused `rocr-debug-agent-tests` and `libhipcxx`
to compile against the wrong toolchain.

Emit `CMAKE_HIP_PLATFORM`, `CMAKE_HIP_COMPILER_ROCM_ROOT`, `CMAKE_HIP_COMPILER`,
and `CMAKE_HIP_COMPILER_LAUNCHER` into the generated toolchain file using
`CACHE ... FORCE`. The `FORCE` is required because `CMakeDetermineHIPCompiler`
runs *after* the toolchain file is processed and would override plain `set()`
calls — forcing into the cache makes the values visible to the determine module
so it skips detection entirely. This mirrors the pattern in
`examples/CMakeLists.txt` (ROCm/TheRock#102).

`CMAKE_HIP_COMPILER_LAUNCHER` is propagated from `CMAKE_CXX_COMPILER_LAUNCHER`
so that HIP compilation benefits from ccache, consistent with C/CXX targets.

## Test Plan

Tested by observing the HIP compiler used to build the mentioned subprojects.
Also validated with a local script that confirms the CMake variables point at
TheRock's compiler.

## Test Result

Confirmed the correct compiler is being used.

ISSUE ID : #102

### DIFF
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -1819,9 +1819,14 @@ function(_therock_cmake_subproject_setup_toolchain
     # CMakeDetermineHIPCompiler, which scans the system and silently picks up
     # /opt/rocm-*/bin/* when present. See ROCm/TheRock#102 and the matching
     # pattern in examples/CMakeLists.txt.
+    # CACHE ... FORCE is required because CMakeDetermineHIPCompiler runs after
+    # the toolchain file is processed and would override plain set() calls.
+    # Forcing into the cache makes the values visible to the determine module
+    # so it skips detection entirely.
     string(APPEND _toolchain_contents "set(CMAKE_HIP_PLATFORM \"amd\" CACHE STRING \"\" FORCE)\n")
     string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER_ROCM_ROOT \"@_hip_dist_dir@\" CACHE PATH \"\" FORCE)\n")
     string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER \"@AMD_LLVM_CXX_COMPILER@\" CACHE FILEPATH \"\" FORCE)\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER_LAUNCHER \"@CMAKE_CXX_COMPILER_LAUNCHER@\" CACHE STRING \"\" FORCE)\n")
 
     if(THEROCK_VERBOSE)
       message(STATUS "HIP_DIR = ${_hip_dist_dir}")
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -1830,7 +1830,7 @@ function(_therock_cmake_subproject_setup_toolchain
 
     if(THEROCK_VERBOSE)
       message(STATUS "HIP_DIR = ${_hip_dist_dir}")
-      message(STATUS "CMAKE_HIP_COMPILER = ${AMD_LLVM_CXX_COMPILER}")
+      message(STATUS "CMAKE_HIP_COMPILER = ${CMAKE_HIP_COMPILER}")
     endif()
   endif()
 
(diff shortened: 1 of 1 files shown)


geomin12 (approved):

cmake/therock_subproject.cmake
```diff
@@ -1795,8 +1795,19 @@ function(_therock_cmake_subproject_setup_toolchain
...
+    # CMakeDetermineHIPCompiler, which scans the system and silently picks up
+    # /opt/rocm-*/bin/* when present. See ROCm/TheRock#102 and the matching
+    # pattern in examples/CMakeLists.txt.
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_PLATFORM \"amd\" CACHE STRING \"\" FORCE)\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER_ROCM_ROOT \"@_hip_dist_dir@\" CACHE PATH \"\" FORCE)\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER \"@AMD_LLVM_CXX_COMPILER@\" CACHE FILEPATH \"\" FORCE)\n")
+
     if(THEROCK_VERBOSE)
       message(STATUS "HIP_DIR = ${_hip_dist_dir}")
+      message(STATUS "CMAKE_HIP_COMPILER = ${AMD_LLVM_CXX_COMPILER}")
```
Naraenda: Non-blocking nit-pick: Wouldn't it be better to echo `CMAKE_HIP_COMPILER` here, in the case that the preceding variable derivation (in the very unlikely case) gets changed?


```suggestion
      message(STATUS "CMAKE_HIP_COMPILER = ${CMAKE_HIP_COMPILER}")
```

Naraenda (commented):
This looks nice! Can't wait to use this in some of the math/primitives libraries 😄.

Silent approval, since I'm not a main contributor here.

cmake/therock_subproject.cmake
lumachad: Yes, sounds reasonable. Thanks.

lumachad:
> This looks nice! Can't wait to use this in some of the math/primitives libraries 😄.
> 
> Silent approval, since I'm not a main contributor here.

Given its global impact, I'm waiting for a few more opinions before I merge this. Otherwise this is a silent bug.

lumachad:
@marbre ping?

lumachad:
@marbre Just checking, did you have a chance to go throw this one? I'm looking for second opinions, as I don't have the full picture for HIP support in CMAKE.

umfranzw:
Some folks have expressed some concerns about the impact of this change. Until there's time to address them, we may need to hold off on this one. @marbre when you have a chance, can you provide some additional details?

lumachad:
> Some folks have expressed some concerns about the impact of this change. Until there's time to address them, we may need to hold off on this one. @marbre when you have a chance, can you provide some additional details?

Could you please expand on what folks and what concerns? This change has been sitting here waiting for feedback for a while, so it would be nice to have that feedback.

Meanwhile we're potentially silently using the wrong toolchain to build HIP cmake projects if there is a system ROCm stack. If that's the case, we're just getting lucky the picked up toolchain does the right thing.

lumachad:
@umfranzw ping?

lumachad:
Drop this one, as we seem to have merged a parallel fix here: https://github.com/ROCm/TheRock/pull/6822

PR #5491, opened 2026-05-28
[ci] Adding GPU resource release script 

geomin12:
After component tests run on gfx942, some processes continue to hold the GPU after tests complete resulting in orphan GPU processes. This results in performance regression and causes test slowdowns.

Of course, the component tests should release GPU resources properly, so I have opened internal tix ROCM-25255 and ROCM-25274 but this will provide immediate relief

Working here: https://github.com/ROCm/TheRock/actions/runs/26543660777/job/78191582156#step:14:26

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -98,6 +98,11 @@ jobs:
         shell: powershell
         run: . '${{ github.workspace }}\prejob\build_tools\github_actions\cleanup_processes.ps1'
 
+      - name: Pre-job cleanup GPU processes on Linux
+        if: ${{ runner.os == 'Linux' }}
+        timeout-minutes: 5
+        run: bash ./prejob/build_tools/github_actions/cleanup_processes.sh
+
       - name: Checking repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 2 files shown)


amd-justchen (approved):
Agree that the GPU should be released correctly!

.github/workflows/test_component.yml
```diff
@@ -98,6 +98,11 @@ jobs:
         shell: powershell
         run: . '${{ github.workspace }}\prejob\build_tools\github_actions\cleanup_processes.ps1'
 
+      - name: Pre-job cleanup GPU processes on Linux
+        if: ${{ runner.os == 'Linux' }}
+        timeout-minutes: 5
+        run: bash ./prejob/build_tools/github_actions/cleanup_processes.sh
```
ScottTodd, line 104: Are we using pre- and post-job scripts yet? https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/run-scripts

I'd really rather not sprinkle so much cleanup code directly into workflow files (as then pytorch tests, jax tests, tests in other repositories, etc. would all want the same code)

geomin12: no we are not. let me look into this actually

geomin12: i'll land this for immediate value (and testing with core42), then I'll deploy this on ossci's docker images and get this rolling! thanks though, this is pretty neat

Issue #5496, opened 2026-05-28
[Multi-arch] Run PyTorch tests after building packages


ScottTodd:
In https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml, 
similar to how https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml runs tests.

Triggering will be different since the build is for all amdgpu targets and then tests will be for individual amdgpu targets (1:many), while the previous single-arch workflow was 1:1. I'm planning on solving this with the help of manifest files tracking package versions, git refs, etc. for each project (https://github.com/ROCm/TheRock/issues/5110).

PR #5503, opened 2026-05-28
Add new PyTorch manifest generation and checkout scripts

ScottTodd:
## Motivation

Progress on
* https://github.com/ROCm/TheRock/issues/5110 (These new scripts will be used for multi-arch instead of the existing scripts)
* https://github.com/ROCm/TheRock/issues/1236 (Generating manifests _before_ building will let us freeze commits for repeatable release workflows) 
* https://github.com/ROCm/TheRock/issues/5496 (Having these records of what packages were built will let us trigger torch tests from multi-arch releases)

## Overview

This adds two new scripts:

* `build_tools/github_actions/generate_pytorch_source_manifest.py`
  * replacement for [`generate_pytorch_manifest.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/generate_pytorch_manifest.py) (to be deleted as part of https://github.com/ROCm/TheRock/issues/3340)
* `external-builds/pytorch/checkout_from_manifest.py`
  * wrapper around [`pytorch_torch_repo.py`](https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/pytorch_torch_repo.py), [`pytorch_audio_repo.py`](https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/pytorch_audio_repo.py), etc.

This will change our torch checkout workflows like so:

```mermaid
  flowchart LR
      C1["Current: checkout PyTorch from floating pin"]
      C2["Checkout related projects from PyTorch pins"]
      C3["Build"]
      C4["Record versions after build"]
      C5["Upload/test"]

      N1["New: resolve commits and versions into manifest"]
      N2["Checkout using manifest"]
      N3["Build"]
      N4["Upload/test"]

      C1 --> C2 --> C3 --> C4 --> C5
      N1 --> N2 --> N3 --> N4
```

This PR is just the new scripts, future PRs will use them from [`multi_arch_build_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml) and [`multi_arch_build_windows_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml).

### Technical Details

Highlighted design decisions:
* Commit hash and version discovery fetches individual files using the GitHub API, rather than clone the repositories. This keeps manifest generation fast, but it does risk API-related failures (e.g. rate limits) earlier in jobs that might not appear with the more natural `git clone` code.
* Version fields in the generated manifests are _expected_ versions, and those versions can then be checked against the _actual_ versions later. There is a risk that these diverge if we don't check.
* There is special case handling for Triton across Linux and Windows since support is still pending on Windows.
* I considered adding more metadata to the manifest schema (e.g. generation date, schema version, etc.) but decided to defer that until later and to start simple.
* The generator lives in `build_tools/github_actions/` as it uses utils from TheRock and references ROCm/pytorch release branches. The checkout script lives in `external-builds/pytorch/` as it connects to the existing checkout scripts and _could_ be used standalone from hand crafted manifest files. If we later want to move more of this code to ROCm/pytorch or pytorch/pytorch we can revisit the split.

## Test Plan

* Added new unit tests
* Ran manifest generation and checkout scripts manually on Windows
  1. These files were generated: [torch_manifest_examples (gist)](https://gist.github.com/ScottTodd/e9e2f9f2622ecb526ec4de478fe32b12)
  2. Spot checked manifest contents for release branches, e.g. https://github.com/ROCm/pytorch/blob/release/2.10/related_commits --> https://github.com/pytorch/audio/blob/5047768f2447d963dffc250b64a5d6c01afc84fa/version.txt
* Tested on github actions as part of a larger feature branch. Sample run: https://github.com/ROCm/TheRock/actions/runs/26195227273
  1. Generated and uploaded https://therock-dev-artifacts.s3.amazonaws.com/26195227273-linux/manifests/pytorch/index.html
  2. Build jobs each receive a single manifest and use it to checkout: https://github.com/ROCm/TheRock/actions/runs/26195227273/job/77073137433#step:10:25
  3. Test jobs also checkout from the manifest: https://github.com/ROCm/TheRock/actions/runs/26195227273/job/77077733178#step:8:19

## Submission Checklist

Assisted-by: Codex

### DIFF
--- a/build_tools/github_actions/generate_pytorch_source_manifest.py
+++ b/build_tools/github_actions/generate_pytorch_source_manifest.py
@@ -45,6 +45,7 @@ DEFAULT_PYTORCH_GIT_REFS = [
     "release/2.12",
     "nightly",
 ]
+SCHEMA_VERSION = 1
 
 THEROCK_DIR = Path(__file__).resolve().parents[2]
 TRITON_WINDOWS_REPO = "triton-lang/triton-windows"
@@ -92,6 +93,8 @@ REPOS: dict[str, RepoConfig] = {
     "triton": RepoConfig(
         stable_repo="ROCm/triton",
         nightly_repo="ROCm/triton",
+        # Triton does not use a floating nightly branch. PyTorch's pin files
+        # resolve the exact commit and the base package version.
         # Windows release Triton is not enabled by default until PyTorch repos
         # publish a shared pin format for release branches.
         exclude_platforms=("windows",),
@@ -130,7 +133,10 @@ def manifest_filename(*, platform: str, pytorch_git_ref: str) -> str:
 
 
 def write_manifest_file(path: Path, manifest: Manifest) -> None:
-    serialized_manifest = {name: entry.to_dict() for name, entry in manifest.items()}
+    serialized_manifest: dict[str, object] = {"schema_version": SCHEMA_VERSION}
+    serialized_manifest.update(
+        {name: entry.to_dict() for name, entry in manifest.items()}
+    )
     path.parent.mkdir(parents=True, exist_ok=True)
     path.write_text(
         json.dumps(serialized_manifest, indent=2, sort_keys=False) + "\n",
@@ -229,7 +235,9 @@ def _resolve_triton(
         )
 
     config = REPOS["triton"]
-    triton_repo = config.stable_repo
+    triton_repo = (
+        config.nightly_repo if pytorch_ref == "nightly" else config.stable_repo
+    )
     pin_file = ".ci/docker/ci_commit_pins/triton.txt"
 
     # Use PyTorch's explicit Triton pin. triton_version.txt only provides the
@@ -256,15 +264,16 @@ def default_projects_for_pytorch_ref(platform: str, pytorch_ref: str) -> list[st
     """Return default projects for a platform and PyTorch ref."""
     projects = default_projects_for_platform(platform)
 
-    # TODO: Flip this once Windows Triton nightly builds are ready by default.
-    # Release branches still need a shared PyTorch-hosted pin format.
-    if (
-        False
-        and platform == "windows"
-        and pytorch_ref == "nightly"
-        and "triton" not in projects
-    ):
-        projects.append("triton")
+    if False:
+        # TODO: Flip this once Windows Triton nightly builds are ready by
+        # default. Release branches still need a shared PyTorch-hosted pin
+        # format.
+        if (
+            platform == "windows"
+            and pytorch_ref == "nightly"
+            and "triton" not in projects
+        ):
+            projects.append("triton")
     return projects
 
 
--- a/build_tools/github_actions/generate_pytorch_source_manifest.py
+++ b/build_tools/github_actions/generate_pytorch_source_manifest.py
@@ -1,0 +1,517 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Generate PyTorch source manifests.
+
+Resolves PyTorch ecosystem refs and pin files to exact commit SHAs and records
+expected package versions in manifest JSON files.
+
+Usage::
+
+    python generate_pytorch_source_manifest.py \
+        --rocm-version 7.13.0a20260501 \
+        --version-suffix "+rocm7.13.0a20260501" \
+        --manifest-dir /tmp/manifests \
+        --pytorch-git-refs "release/2.11"
+"""
+
+import argparse
+import json
+import platform as platform_module
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+_BUILD_TOOLS_DIR = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(_BUILD_TOOLS_DIR))
+
+from github_actions.github_actions_api import (
+    gha_fetch_text_file_contents,
+    gha_resolve_git_ref,
+)
+from github_actions.determine_version import derive_version_suffix
+from github_actions.manifest_utils import (
+    GitSourceInfo,
+    detect_therock_source_info,
+    log,
+    normalize_ref_for_filename,
+)
+
+DEFAULT_PYTORCH_GIT_REFS = [
+    "release/2.9",
+    "release/2.10",
+    "release/2.11",
+    "release/2.12",
+    "nightly",
+]
+
+THEROCK_DIR = Path(__file__).resolve().parents[2]
+TRITON_WINDOWS_REPO = "triton-lang/triton-windows"
+
+Manifest = dict[str, GitSourceInfo]
+
+
+@dataclass(frozen=True)
+class RepoConfig:
+    """Configuration for a pytorch ecosystem repository."""
+
+    stable_repo: str
+    nightly_repo: str
+    nightly_branch: str | None = None
+    version_file: str | None = None
+    # Key in ROCm/pytorch's ``related_commits`` file (e.g. "torchaudio").
+    # When set, stable builds resolve from related_commits; when None,
+    # the repo uses custom resolution logic (pytorch itself, triton).
+    related_commits_key: str | None = None
+    # Platforms this repo is excluded from. Empty means all platforms.
+    exclude_platforms: tuple[str, ...] = ()
+
+
+REPOS: dict[str, RepoConfig] = {
+    "pytorch": RepoConfig(
+        stable_repo="ROCm/pytorch",
+        nightly_repo="pytorch/pytorch",
+        nightly_branch="nightly",
+        version_file="version.txt",
+    ),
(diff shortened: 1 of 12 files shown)


build_tools/github_actions/generate_pytorch_source_manifest.py
```diff
@@ -0,0 +1,517 @@
...
+
+    # TODO: Flip this once Windows Triton nightly builds are ready by default.
+    # Release branches still need a shared PyTorch-hosted pin format.
+    if (
+        False
+        and platform == "windows"
+        and pytorch_ref == "nightly"
+        and "triton" not in projects
+    ):
+        projects.append("triton")
```
ScottTodd: @tvukovic-amd here's the new torch checkout code I've been working on.

* For https://github.com/ROCm/TheRock/pull/4205, we would drop the `False` from the condition here to start including triton for the "nightly" pytorch ref release builds
* Later on we would change `_resolve_triton()` and `read_triton_windows_pin()` above to read from a file like `.ci/docker/triton_version.txt` (perhaps `.ci/docker/triton_windows_version.txt`) from the torch repository itself

How it will work in the workflows will be:
* The manifests that get generated at the start of the workflows like https://gist.github.com/ScottTodd/e9e2f9f2622ecb526ec4de478fe32b12 will start to include `triton` for select release versions on Windows
* The checkout step during build workflows will pick up whatever projects are in the manifest
* The build step can use an explicit `--triton-dir B:/src/triton` or an implicit checkout root location + an expected directory name

external-builds/pytorch/pytorch_manifest_nightly.example.json
```diff
@@ -0,0 +1,37 @@
...
+    "commit": "88b227e23f0445f3f695bad05bbf1a363b4f50e0",
+    "version": "3.7.0+rocm7.14.0"
+  },
+  "apex": {
+    "repo": "https://github.com/ROCm/apex",
+    "commit": "af25af4e26f0546d897c3e2ff31502ef50530ecc",
+    "branch": "master",
+    "version": "1.11.0+rocm7.14.0"
+  },
+  "therock": {
```
marbre, line 32: Should this be `rocm` instead of `therock`?

external-builds/pytorch/README.md
```diff
@@ -454,26 +459,90 @@ the fork at https://github.com/ROCm/pytorch.
...
+# Multiple versions (computed filenames in a directory):
+python build_tools/github_actions/generate_pytorch_source_manifest.py \
+    --rocm-version 7.13.0a20260501 \
+    --manifest-dir external-builds/pytorch/manifests/ \
+    --pytorch-git-refs "release/2.10 release/2.11 nightly"
+
+# Explicit platform selection (Windows may include different packages):
+python build_tools/github_actions/generate_pytorch_source_manifest.py \
+    --rocm-version 7.13.0a20260501 \
+    --platform windows \
```
marbre, line 516: Only looking at the README, does the platform otherwise defaults to Linux?

build_tools/github_actions/generate_pytorch_source_manifest.py
```diff
@@ -0,0 +1,517 @@
...
+        nightly_repo="pytorch/vision",
+        nightly_branch="nightly",
+        version_file="version.txt",
+        related_commits_key="torchvision",
+    ),
+    "triton": RepoConfig(
+        stable_repo="ROCm/triton",
+        nightly_repo="ROCm/triton",
+        # Windows release Triton is not enabled by default until PyTorch repos
+        # publish a shared pin format for release branches.
```
marbre, line 99: Should `nightly_branch` be set here for consistency as well?

build_tools/github_actions/generate_pytorch_source_manifest.py
ScottTodd: 🤔 might want to rename "triton" to "triton-windows" in some of these places though...

build_tools/github_actions/generate_pytorch_source_manifest.py
```diff
@@ -0,0 +1,517 @@
...
+
+
+def default_projects_for_pytorch_ref(platform: str, pytorch_ref: str) -> list[str]:
+    """Return default projects for a platform and PyTorch ref."""
+    projects = default_projects_for_platform(platform)
+
+    # TODO: Flip this once Windows Triton nightly builds are ready by default.
+    # Release branches still need a shared PyTorch-hosted pin format.
+    if (
+        False
```
marbre: Maybe use a single `if False:` which the surrounding explanatory comment?

build_tools/github_actions/generate_pytorch_source_manifest.py
```diff
@@ -0,0 +1,517 @@
...
+        pin = read_triton_windows_pin()
+        log(f"  triton-windows pin: {pin[:12]}")
+        return GitSourceInfo(
+            commit=pin,
+            repo=f"https://github.com/{TRITON_WINDOWS_REPO}",
+            version=version,
+        )
+
+    config = REPOS["triton"]
+    triton_repo = config.stable_repo
```
marbre:
```suggestion
    triton_repo = config.nightly_repo if nightly else config.stable_repo
```
?

This is always the stable repo which works because `stable_repo` and `nightly_repo` are both set to `ROCm/trition`. This might break in the future though. Either mirror the selection logic used elsewhere or drop `nightly_repo`?

build_tools/github_actions/manifest_utils.py
```diff
@@ -132,6 +135,22 @@ def resolve_branch(*, inferred: str | None, provided: str | None) -> str | None:
     return None
 
 
+def detect_therock_source_info(repo_root: Path) -> GitSourceInfo:
+    """Detect TheRock commit, repo, and branch from a local git checkout.
+
+    Falls back to "unknown" for fields that cannot be determined (e.g. when
```
marbre, line 147: Would it make sense to warn on stderr when falling back to unknown?

external-builds/pytorch/checkout_from_manifest.py
```diff
@@ -0,0 +1,250 @@
...
+        "--checkout-dir",
+        str(checkout_dir),
+    ]
+
+    # pytorch_torch_repo.py doesn't have --torch-dir, but the others do.
+    # When checking out from a manifest we pass explicit origin/hashtag,
+    # so related_commits is not needed. We still pass --torch-dir pointing
+    # to the pytorch checkout so the scripts can find it if needed.
+    if name != "pytorch":
+        cmd.extend(["--torch-dir", str(checkout_root / "pytorch")])
```
marbre: If this isn't needed I think we shouldn't pass it?

marbre (approved):
Just some minor things. Please address before landing.

> I considered adding more metadata to the manifest schema (e.g. generation date, schema version, etc.) but decided to defer that until later and to start simple.

How about a simple `"schema_version": 1`?

external-builds/pytorch/README.md
ScottTodd: It defaults to `platform.system().lower()`, so developers on Windows/Linux both get a manifest for their platform (e.g. including apex/triton on Linux, including triton-windows on Windows) by default. CI jobs that use Linux runners to generate manifests and unit tests can both set the platform explicitly.

external-builds/pytorch/pytorch_manifest_nightly.example.json
ScottTodd: I could go either way on the name here.

* I wanted to keep parity with the existing manifest format, e.g. https://therock-nightly-artifacts.s3.amazonaws.com/26623058753-linux/manifests/gfx950-dcgpu/therock-manifest_torch_py3.10_nightly.json from https://github.com/ROCm/TheRock/actions/runs/26623058753 (single-arch nightly release)
* The "project" name here for each doesn't directly map to a package or repository name. For example, `pytorch_vision` is for https://github.com/pytorch/vision and https://pypi.org/project/torchvision/

external-builds/pytorch/checkout_from_manifest.py
ScottTodd: Good catch. Removed `--torch-dir` and fixed the apex checkout script to tolerate it being omitted.

build_tools/github_actions/generate_pytorch_source_manifest.py
tvukovic-amd: @ScottTodd thank you for explaining, I pushed the following changes to https://github.com/ROCm/TheRock/pull/4205: 

- Removed the if False guard in default_projects_for_pytorch_ref() so Windows nightly manifests include triton by default.
- Re-enabled test_windows_nightly_default_projects_include_triton

PR #5507, opened 2026-05-29
[ci] Per-category timeouts for test matrix; migrate MIOpen

JonathanLichtnerAMD:
## Motivation

Tracking: ALMIOPEN-1994.

The test filter standardization feature provides four classes of testing
(quick, standard, comprehensive, full), each with its own idealized time
budget. But TheRock's matrix generator only accepts a single
`timeout_minutes` per component, which means every component's timeout
has to be sized for its longest-running category — even quick CI runs
inherit the worst-case ceiling.

## Overview

Two commits:

1. **`[ci] Allow per-category timeouts in test matrix generator`**
   - `_resolve_timeout` accepts either a scalar `int` (legacy) or a
     dict keyed by `test_type` with an optional `default` fallback;
     `smoke` aliases to `quick`.
   - `_validate_timeout_minutes` runs at module import over
     `test_matrix`, `functional_matrix`, and `benchmark_matrix`.
     Rejects unknown keys (typo guard), partial dicts without
     `default`, non-positive values, and `bool` values (which would
     otherwise sneak past `isinstance(x, int)`).
   - `run()` call site is guarded by an `in`-check so components that
     omit `timeout_minutes` keep working.
   - 21 new unit tests in
     `build_tools/github_actions/tests/fetch_test_configurations_test.py`
     covering both helpers and the end-to-end resolution path through
     `run()` (Linux + Windows hip-tests branches).

2. **`[ci] Raise MIOpen comprehensive/full timeouts to 120 min, drop quick to 30`**
   - Migrates the `miopen` entry only: `quick=30`, `standard=60`,
     `comprehensive=120`, `full=120`, `default=60`.

No other components are migrated.

## Risk Assessment

**Rating: Low.**

CI-only change, no runtime code path. Scalar `timeout_minutes` entries
keep exact prior behavior; the MIOpen dict is the only migrated entry.
The new import-time validator pins the shipped MIOpen dict and blocks
malformed configs before any job is scheduled. The only behavioral
delta to watch is MIOpen `quick` dropping 60→30 min — generous against
quick's 5-minute target, and reverting is a one-line dict edit.

## Test Plan

- [x] `python -m unittest build_tools.github_actions.tests.fetch_test_configurations_test` — 43 / 43 pass locally.
- [x] Confirm 30 minutes used for `quick`
- [x] Confirm 60 minutes used for `standard`
- [x] Confirm 120 minutes used for `comprehensive`
- [x] Spot-check one MIOpen quick job's emitted timeout in the workflow
      run logs.


JonathanLichtnerAMD:
Okay, thanks to core42 pool slowness, I see that the timeouts in the test_categories.yaml are actually honoured, so the quick tests time out in 10 minutes rather than 30 minutes.  This means this PR is overkill:  I can just increase the MIOpen default timeout to the largest possible value and rely on the timeout in the test_categories.yaml file.

JonathanLichtnerAMD:
https://github.com/ROCm/TheRock/pull/5572 raised instead

PR #5508, opened 2026-05-29
[rocgdb] XFAIL multi-inferior tests on mixed-GPU hosts with gfx1036

lumachad:
These tests fail when run on machines hosting mixed GPUs where one of them is the unsupported gfx1036.


marbre (approved):
Can we file a tracking issue to come up with a proper solution for this and link it here in the comment?

lumachad:
> Can we file a tracking issue to come up with a proper solution for this and link it here in the comment?

Tracking as AIROCGDB-603.

PR #5521, opened 2026-05-29
Add test filter standardisation for rocwmma

dileepr1:
## Motivation

Add test filter standardisation for rocwmma.
This works with the companion PR https://github.com/ROCm/rocm-libraries/pull/7535

## Technical Details

- fetch_test_configurations.py: rocwmma test_script -> test_runner.py so the generic runner drives ctest from THEROCK_BIN_DIR/rocwmma using the quick/standard/comprehensive/full category labels.

## Test Plan

Confirm that the CI tests for rocwmma works good once https://github.com/ROCm/rocm-libraries/pull/7535 is merged.

## Test Result

Test Run: https://github.com/ROCm/TheRock/actions/runs/27280235753/job/80607463646?pr=5521

## Submission Checklist


tony-davis:
This swaps rocwmma to `test_runner.py`, which always runs `ctest --test-dir $THEROCK_BIN_DIR/rocwmma`. Legacy `test_rocwmma.py` used `--test-dir …/rocwmma/regression` for `TEST_TYPE=quick` (and `regression`). The install tree ships three independent fragments — `rocwmma/`, `rocwmma/smoke/`, `rocwmma/regression/` — and the root `CTestTestfile.cmake` has no `subdirs()` into `regression/`.

This won't error: the root tree includes a `regression_tests` entry that carries the `quick` label, so `ctest -L quick` finds tests and the job passes. But it silently reduces quick coverage — the per-target emulation `"<target> regression"` runs (gemm/unit/dlrm) live only under `…/rocwmma/regression` and would no longer run in quick.

Two options: (1) add `TEST_TYPE`-aware test-dir selection for rocwmma (quick/regression → `…/rocwmma/regression`, else `…/rocwmma`), mirroring legacy; or (2) deliberately accept `regression_tests`-only as the new quick scope and update rocm-libraries#7535's `test_categories.yaml` comment to match. Happy to pair on where (1) fits cleanest in `test_runner.py`.

tony-davis (changes requested):
I left a comment, and I think it's important to address that subdirectory situation for rocwmma. Thanks!

dileepr1:
> I left a comment, and I think it's important to address that subdirectory situation for rocwmma. Thanks!

Thank you for pointing it out.
This is implemented.

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -174,7 +214,12 @@ def apply_component_overrides(job_name, rocm_path, therock_dir, default_test_dir
 
 TEST_DIR = str(Path(THEROCK_BIN_DIR) / TEST_COMPONENT)
 TEST_DIR = apply_component_overrides(
-    test_component_job_name, ROCM_PATH, THEROCK_DIR, TEST_DIR, environ_vars
+    test_component_job_name,
+    (TEST_TYPE or "").lower(),
```
geomin12: can we add a default instead of empty?

geomin12 (commented):
ci doesn't run. can we get CI to run? maybe a restart will do

build_tools/github_actions/test_executable_scripts/test_runner.py
dileepr1: This is done, handled at high level instead of doing at at here.

dileepr1:
> ci doesn't run. can we get CI to run? maybe a restart will do

Restarted, https://github.com/ROCm/rocm-libraries/actions/runs/27217781805/job/80423012942?pr=7037

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -118,6 +142,26 @@
...
+    # TEST_TYPE=regression, which the module-level validator now folds
+    # back to "quick") to the regression fragment so the per-target
+    # emulation regression runs (gemm/unit/dlrm) were exercised. Mirror
+    # that here so swapping to test_runner.py preserves coverage. Pairs
+    # with the rocm-libraries PR that tags the "<target> regression"
+    # entries with the `quick` label in bin/rocwmma/regression/
+    # CTestTestfile.cmake.
+    # Other TEST_TYPEs (standard/comprehensive/full) fall through to the
+    # default bin/rocwmma/ fragment, matching legacy behaviour.
+    "rocwmma": {
```
geomin12, line 160: can we get rocwmma folks to review this too?

geomin12 (approved):
lgtm, just want an approval from rocwmma team too

tony-davis (changes requested):
**Requesting changes — thanks for addressing the subdirectory routing; the new issue is the global `--no-tests=error`.**

First, the good news: the `test_dir_by_type` route for rocwmma `quick` → `bin/rocwmma/regression` correctly resolves my earlier comment — the per-target emulation regression entries (gemm/unit/dlrm) are preserved through the migration. 👍 And the rocwmma test job itself is green.

**Blocking — the global `--no-tests=error` breaks CI for other components.** This flag is added in `build_ctest_command()`, which is shared by *every* component that runs through `test_runner.py`, not just rocwmma. On the current tip commit (`e3f2558`) it turns a previously-silent-green "No tests were found" into a hard failure. From the rocprim job:

```
# Running: ctest -L quick ... --no-tests=error ... --test-dir build/bin/rocprim
No tests were found!!!
Errors while running CTest
##[error]Process completed with exit code 8.
```

Any component whose install tree has tests but none carry the requested category label now fails. Making zero-selected-tests an error may well be the right end state, but it's a repo-wide CI-policy change riding inside a rocwmma migration PR. Could we either (a) gate it per-component (opt in once a component is migrated to category labels), or (b) split it into its own PR and wire labels for the affected components first? As-is it expands the blast radius well beyond rocwmma.

**Important — please refresh the PR description.** The Motivation and Test Plan still say "hipdnn" (this is the rocwmma PR), and the Technical Details only mention the `test_script` swap — they don't cover the TEST_TYPE normalize/validate change or the global `--no-tests=error` policy, which together are most of the diff. Updating it will help the next reviewer see the real scope.

Thanks again!

dileepr1:
> **Requesting changes — thanks for addressing the subdirectory routing; the new issue is the global `--no-tests=error`.**
> 
> First, the good news: the `test_dir_by_type` route for rocwmma `quick` → `bin/rocwmma/regression` correctly resolves my earlier comment — the per-target emulation regression entries (gemm/unit/dlrm) are preserved through the migration. 👍 And the rocwmma test job itself is green.
> 
> **Blocking — the global `--no-tests=error` breaks CI for other components.** This flag is added in `build_ctest_command()`, which is shared by _every_ component that runs through `test_runner.py`, not just rocwmma. On the current tip commit (`e3f2558`) it turns a previously-silent-green "No tests were found" into a hard failure. From the rocprim job:
> 
> ```
> # Running: ctest -L quick ... --no-tests=error ... --test-dir build/bin/rocprim
> No tests were found!!!
> Errors while running CTest
> ##[error]Process completed with exit code 8.
> ```
> 
> Any component whose install tree has tests but none carry the requested category label now fails. Making zero-selected-tests an error may well be the right end state, but it's a repo-wide CI-policy change riding inside a rocwmma migration PR. Could we either (a) gate it per-component (opt in once a component is migrated to category labels), or (b) split it into its own PR and wire labels for the affected components first? As-is it expands the blast radius well beyond rocwmma.
> 
> **Important — please refresh the PR description.** The Motivation and Test Plan still say "hipdnn" (this is the rocwmma PR), and the Technical Details only mention the `test_script` swap — they don't cover the TEST_TYPE normalize/validate change or the global `--no-tests=error` policy, which together are most of the diff. Updating it will help the next reviewer see the real scope.
> 
> Thanks again!

That was the exact intention of adding --no-tests=error, so that we won't have false positives - a pass with no tests run. The rocprim failure is because the component side PR for rocprim is not merged.
But you are right - it is better not to piggyback that change on top of rocwmma. I've reverted it.
Also corrected the description.

tony-davis (approved):
Approving — reverting the `--no-tests=error` commits resolves my blocking concern, and the diff is now cleanly scoped to the rocwmma migration: the swap to `test_runner.py`, the `test_dir_by_type` quick→`bin/rocwmma/regression` routing (the fix for my earlier comment), and the TEST_TYPE normalize/validate cleanup all look good. Thanks!

PR #5522, opened 2026-05-29
[rocprofiler-systems] Route test job through test_runner.py

dileepr1:
The literal-value env vars surface a new, generic `env` field on `COMPONENT_OVERRIDES`. Values are formatted via `str.format()` with a single `{rocm_path}` placeholder, which keeps the existing path-prepend mechanism untouched and is purely additive for other components.

## Motivation

Add test filter standardisation for hipdnn.
This works with the companion PR https://github.com/ROCm/rocm-systems/pull/6461

## Technical Details

Switches the `rocprofiler-systems` entry in `fetch_test_configurations.py` from the bespoke `test_rocprofiler_systems.py` to the shared `test_runner.py`, completing the test-filter standardisation handoff for this component (paired with rocm-systems `users/dravindr/tf_rocprofsys`, which now emits `quick` / `standard` / `comprehensive` / `full` tier labels and a matching `<tier>_exclude` label via the pytest CTest hook).

Adds a `COMPONENT_OVERRIDES` entry for `rocprofiler-systems` that mirrors the legacy script's environment:
 - test_dir -> `share/rocprofiler-systems/tests` (where the generated `CTestTestfile.cmake` and the `rocprofsys-tests.pyz` live)
 - PATH prepended with `bin` so `rocprof-sys-*` / `rocprofv3` resolve
 - LD_LIBRARY_PATH prepended with the examples lib dir
 - `ROCPROFSYS_INSTALL_DIR` and `ROCPROFSYS_MAX_THREADS` set explicitly

## Test Plan

Confirm that rocprofiler-systems tests gets triggered through therock CI once https://github.com/ROCm/rocm-systems/pull/6461 is merged.

## Test Result

Successful run: https://github.com/ROCm/TheRock/actions/runs/28819102474/job/85669920329?pr=5522

## Submission Checklist


geomin12 (commented):
tests pass but 10 mins is still too long for quick filter https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md

kcossett-amd:
I see, I was under the impression that quick was < 15 minutes. I'll update the `test_categories.yaml` on our side after discussion with the team.

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -166,6 +183,31 @@
...
+    # tree; ROCPROFSYS_MAX_THREADS caps thread oversubscription on CI hosts.
+    "rocprofiler-systems": {
+        "test_dir": ["share", "rocprofiler-systems", "tests"],
+        "additional_env_paths": {
+            "PATH": [["bin"]],
+            "LD_LIBRARY_PATH": [["share", "rocprofiler-systems", "examples", "lib"]],
+        },
+        "env": {
+            "ROCPROFSYS_INSTALL_DIR": "{rocm_path}",
+            "ROCPROFSYS_MAX_THREADS": "64",
```
geomin12: where is this number coming from??? some machines don't even have 64 threads (cpu cores)

dileepr1: This had been an existing env var, and is adapted from [test_rocprofiler_systems.py](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py).

@kcossett-amd please add if you have any additional info.

kcossett-amd: `ROCPROFSYS_MAX_THREADS` is a compile time var used to set the max number of application threads we hold profiling data for. It is unrelated to the number of CPU cores, so a machine with fewer than 64 cores is fine here. If the program spawns more than this many threads, the extra ones aren't profiled (with the limit at 64, threads 65 and onward won't be collected).

The var is computed based off the number of processors on the machine where rocm-systems was built, which is incompatible with TheRock. So, I hardcoded it to 64 (although it probably should have been 128). Either way, that is a workaround and in the near future (through AIPROFSYST-523), we plan on fixing this to make it able to grow dynamically.

geomin12: ah, this script is not used during compilation. this is only designed and used to be a python wrapper for the test executable scripts

geomin12: we can remove as this is used for compilation, but this file is only a python wrapper around ctest

kcossett-amd: Agreed, you can remove it. ~~It would break the `thread-limit` tests, which depend on this variable. But those tests are disabled anyways because the hardcoding always breaks the `above-limit` variant of the `thread-limit` tests on systems with more/less threads (this was a workaround)~~. 
It truly is a whole debacle which will hopefully be fixed soon :)

kcossett-amd: Agreed. I did this as it was a workaround to get the non `above-limit` `thread-limit` tests working. But I disabled all those tests in the end.

dileepr1: ROCPROFSYS_MAX_THREADS is removed.

geomin12 (approved):

PR #5547, opened 2026-06-01
Add rocprof trace decoder dev component

ApoKalipse-V:
## Motivation

rocprof-trace-decoder was missing it's dev component installation, meaning other projects couldn't find the decoder via cmake.
This is due to the original closed source decoder being a .so only, without dev components.

## Technical Details

Added the dev components such that the cmake target and include headers are part of the build.
Tested locally and it produces the cmake/ and include/ artifact

## JIRA ID

AIPROFSDK-894

## Submission Checklist


venkat1361 (approved):

marbre (approved):
Please update the test sections and confirm that the artifact now contains what you expected after manual inspection. Otherwise drop empty sections as it all folds into the squashed commit msg.

PR #5549, opened 2026-06-01
Extend Dependabot to cover composite actions

marbre:
Dependabot's github-actions ecosystem with directory `/` only scans `.github/workflows/` in practice, missing composite actions defined under `.github/actions/`. As an examplem, PR #5338 updates `configure-aws-credentials` pin in workflows but misses to bump the action in `actions/configure_aws_artifacts_credentials/`.

This adds an explicit entry for the composite action directory so pinned dependencies in those get bumped by dependabot as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,9 @@ updates:
 
   # Check for updates to GitHub Actions
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions/*"
     schedule:
       interval: "weekly"
     groups:


.github/dependabot.yml
```diff
@@ -10,3 +10,21 @@ updates:
...
+          - "*"
+
+  - package-ecosystem: "github-actions"
+    directory: "/.github/actions/configure_aws_artifacts_credentials"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
```
ScottTodd, line 14: Can we group these into a single item by using `directories` instead of `directory`?

* https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/configuring-dependabot-version-updates
* https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files

I also wonder if a recursive search will pick up `.github/actions/` instead of needing to be explicit with each one

ScottTodd (approved):

.github/dependabot.yml
marbre: Oh, seems we can, will update.

marbre: `/.github/actions/*` _should_ work. Will monitor after landing if deps get bumped now.

ScottTodd (approved):

Issue #5374, opened 2026-05-20
RuntimeError: operator torchvision::nms does not exist on gfx900 with ROCm nightly 7.14


JokerGermany:
This is a Follow up to:
https://github.com/ROCm/TheRock/issues/5358

```
venv-linux-amd-3.12) jokergermany@huppyryzen:/media/HDD/programs/AI/ComfyUI$ python3 -m venv venv-linux-amd-3.12
(venv-linux-amd-3.12) jokergermany@huppyryzen:/media/HDD/programs/AI/ComfyUI$ source venv-linux-amd-3.12/bin/activate
(venv-linux-amd-3.12) jokergermany@huppyryzen:/media/HDD/programs/AI/ComfyUI$ TMPDIR=/media/HDD/tmp uv pip install --force-reinstall --no-cache-dir --pre torch torchvision torchaudio --index-url https://rocm.nightlies.amd.com/v2-staging/gfx900/ 
Using Python 3.12.3 environment at: venv-linux-amd-3.12
Resolved 18 packages in 15.61s
      Built rocm==7.14.0a20260520
Prepared 18 packages in 2m 23s
Installed 18 packages in 276ms
 + filelock==3.29.0
 + fsspec==2026.4.0
 + jinja2==3.1.6
 + markupsafe==3.0.3
 + mpmath==1.3.0
 + networkx==3.6.1
 + numpy==2.4.6
 + pillow==12.2.0
 + rocm==7.14.0a20260520
 + rocm-sdk-core==7.14.0a20260520
 + rocm-sdk-libraries-gfx900==7.14.0a20260520
 + setuptools==82.0.1
 + sympy==1.14.0
 + torch==2.10.0+rocm7.14.0a20260520
 + torchaudio==2.11.0+rocm7.14.0a20260520
 + torchvision==0.27.0+rocm7.14.0a20260520
 + triton==3.6.0+rocm7.14.0a20260520
 + typing-extensions==4.15.0
(venv-linux-amd-3.12) jokergermany@huppyryzen:/media/HDD/programs/AI/ComfyUI$ python -c "import torch; import torchvision; print(torch.__version__, torchvision.__version__)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/media/HDD/programs/AI/ComfyUI/venv-linux-amd-3.12/lib/python3.12/site-packages/torchvision/__init__.py", line 8, in <module>
    from torchvision import (
[...]

PR #5554, opened 2026-06-01
Remove stale dependency wheels from S3 during update

lucbruni-amd:
## Motivation

`setuptools` was pinned from `latest` to `81.0.0` as of https://github.com/ROCm/TheRock/pull/4940 because `torch` requires `setuptools<82`. However, `update_dependencies.py` did not seem to be re-run to account for this. If it were to be re-run, the stale `82.0.1` version previously there would be removed via the script.

The symptom was noticed as part of the investigation into https://github.com/ROCm/TheRock/issues/5374:

```
$ pip install --index-url https://rocm.nightlies.amd.com/v2-staging/gfx900/ torch
...
ERROR: Could not find a version that satisfies the requirement setuptools<82 (from torch)
  (from versions: 82.0.1)
```

The index only has `82.0.1`, `torch` needs `<82`, and `--index-url` replaces PyPI entirely so `pip` can't fall back

## Technical Details

- After uploading pinned dependency versions, `update_dependencies.py` now deletes any existing wheels for that package whose version is no longer in the allowed set
- Prevents stale wheels from persisting on the index after a version pin change (e.g. setuptools 82.0.1 remaining after pin changed to 81.0.0 in #4940)
- Respects `--dry-run` and multi-version packages (e.g. `networkx` with `["3.4.2", "latest"]`)

## Test Plan

- [x] Existing 44 unit tests pass
- [ ] Run with `--dry-run` against dev bucket to verify stale detection, which requires S3 creds
- [ ] Run against dev bucket, then regenerate index and verify stale wheels are removed

## Test Result

To observe

### DIFF
--- a/build_tools/third_party/s3_management/update_dependencies.py
+++ b/build_tools/third_party/s3_management/update_dependencies.py
@@ -257,7 +257,7 @@ def upload_missing_whls(
     dry_run: bool = False,
     only_pypi: bool = False,
     target_version: str = "latest",
-) -> None:
+) -> str | None:
     pypi_idx = parse_simple_idx(f"https://pypi.org/simple/{pkg_name}")
     pypi_versions = get_whl_versions(pypi_idx)
 
@@ -274,7 +274,7 @@ def upload_missing_whls(
 
     if not selected_version:
         print(f"No stable versions found for {pkg_name}")
-        return
+        return None
 
     pypi_latest_packages = get_wheels_of_version(pypi_idx, selected_version)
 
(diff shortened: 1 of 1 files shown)


marbre (commented):
Do you think this is needed at all if we have a retention policy in place? If not we can keep the script simpler and do not need to add more complexity.

lucbruni-amd:
I'm on board with keeping the script simple in favour of a retention policy with some sort of version-aware cleanup. If that's possible, we can close this.

lucbruni-amd:
@marbre anything from my side needs to be done for that, or anything I can help with?

marbre:
> @marbre anything from my side needs to be done for that, or anything I can help with?

I don't think so. This is in the backlog but with the refactoring of repo.amd.com we hopefully address this soon.

Issue #5518, opened 2026-05-29
[Feature] non-x86 portability support / upstream ppc64le changes


benrichard-amd:
## Summary

ROCm has been building and running on IBM Power (ppc64le) for some time, maintained in a [branch that tracks TheRock upstream](https://github.com/ROCm/TheRock/tree/benrichard-amd/ppc64le). This issue tracks the effort to upstream those changes into TheRock proper.

## Motivation

TheRock's CMake layer currently encodes several x86-specific assumptions — hardcoded LLVM target, unconditional architecture flags, and features enabled without architecture guards. These assumptions are not wrong on x86, but they make it harder to build on any other host and obscure which features are genuinely portable.

Phase 1 addresses this at the TheRock level, independent of any particular port. Correcting these assumptions makes the build system more honest about what is and is not architecture-specific, and reduces the friction for component maintainers who want to upstream fixes: once the scaffolding is clean, component patches can focus on component-level issues rather than working around build system artifacts.

## Approach

**Phase 1 — TheRock infrastructure changes**

Corrections to x86 assumptions baked into TheRock's CMake layer:
- Use `Native` instead of `X86` when specifying LLVM targets to build — equivalent on x86 hosts, correct everywhere else
- Guard x86-specific compile flags behind architecture checks so they are not passed on non-x86 builds
- Guard x86-only features (RDC, rocprofiler-systems, ESMI) behind explicit architecture checks with clear diagnostics — these features remain enabled on x86 as before
- Refactor kpack's ELF surgery to make it straightforward to add page size and relocation constants for other architectures

**Phase 2 — Component fixes upstream**

ppc64le fixes for individual ROCm components are submitted directly to their respective projects. As they land, TheRock picks them up via normal submodule updates. No patches are carried in TheRock.

Notable work in this phase:
- **Clang**: float16 and bf16 are not currently supported on PowerPC. Patches adding this support in the ppc64le branch need to be upstreamed to LLVM/Clang before ROCm components that depend on them can build cleanly.
- Additional fixes across CLR, RCCL, ROCr, hipBLAS, and others.

## Sequencing

1. Land Phase 1 CMake changes in TheRock
2. Submit component fixes to upstream repos in parallel, with Clang float16/bf16 support as a key dependency
3. TheRock submodule bumps pick up fixes as they merge

## Success Criteria

**Phase 1**: TheRock's CMake layer passes no x86-specific flags or enables no x86-only features on non-x86 hosts, with no regressions on x86.

[...]

PR #5555, opened 2026-06-01
Set LLVM CPU target to Native

benrichard-amd:
## Motivation

Part of #5518 

TheRock does not support cross-compilation (yet). The target machine is the same arch as the build machine. In an effort to make ROCm more portable, use `Native` so that the CPU arch is auto-selected.

## Technical Details

Change CPU target in `LLVM_TARGETS_TO_BUILD` from `X86` to `Native`.

## Test Plan

Build TheRock on an X86 system. Verify it successfully builds and runs on an X86 system.

## Test Result

Builds and runs on X86 system.

## Submission Checklist


ScottTodd (commented):
I suspect this is fine, but I can't safely review this PR until it has a clean CI run showing successful builds and no test regressions. The new AWS runners flaked at https://github.com/ROCm/TheRock/actions/runs/26765701339/job/78895940239?pr=5555 

We should compare the configure and build logs for LLVM before and after this PR, as well as diff the compiler artifacts to check that this doesn't have unintended side effects.

(cc @kailash-khalasi @amd-shiraz )

marbre (commented):
Following #5518, are there plans to somehow hardcode / make it configurable which target we aim to build for or does it solely depend on the runner architecture? If we need to configure it for other parts of the build system I would prefer to make it explicit here as well. This might also allow to build for both if there is desire for cross-compiling.

benrichard-amd:
> Following #5518, are there plans to somehow hardcode / make it configurable which target we aim to build for or does it solely depend on the runner architecture? If we need to configure it for other parts of the build system I would prefer to make it explicit here as well. This might also allow to build for both if there is desire for cross-compiling.

TheRock builds clang and uses it to build ROCm, and also includes it in the SDK. So the runner and target are necessarily the same architecture.


I plan on adding cross-compilation. We will need to re-work some things since it requires building 2 clangs: the cross compiler and the target compiler (to include with SDK), and probably other things.

ScottTodd (approved):

PR #5559, opened 2026-06-01
Remove unused libatomic dependency from libnuma

### LINKED ISSUE #2518: [Issue]: libatomic installation needed on certain distros for rocminfo to work following tarball install for ROCm 7.10

darren-amd:
## Motivation

Fixes #2518.

On clean installs of distros that don't ship `libatomic` (RHEL 8.10/10.0, SLES 15.7, minimal Ubuntu, etc.), `rocminfo` from our tarballs fails to load.

The chain is `rocminfo` -> `libhsa-runtime64.so.1` -> `librocm_sysdeps_numa.so.1`, and bundled `libnuma` is the only thing pulling in libatomic.

## Technical Details

numactl's `configure` unconditionally links `-latomic` (via `AC_SEARCH_LIBS([__atomic_fetch_and_1], [atomic])`), but its atomics lower inline (on gcc and llvm), so the resulting `NEEDED libatomic.so.1` resolves zero symbols and is not a necessary dependency:
```
readelf -d librocm_sysdeps_numa.so.1 | grep NEEDED          # libatomic.so.1 present
nm -D --undefined-only librocm_sysdeps_numa.so.1 | grep -E '__atomic_|LIBATOMIC'   # empty
```

Created a patch that drops the linking before build time: `AC_SEARCH_LIBS([__atomic_fetch_and_1], [atomic])` from configure.ac.

## Test Plan

Run a CI build from this branch and with the built tarball, run `rocminfo` from it in a clean container with no system libatomic (`ubuntu:22.04`).

## Test Result

From **new** build https://github.com/ROCm/TheRock/actions/runs/26832538938:
`librocm_sysdeps_numa.so.1` no longer lists libatomic as NEEDED; `rocminfo` now loads and enumerates `gfx1100` and `libnuma` exports are unchanged.

## Submission Checklist

### DIFF
--- a/third-party/sysdeps/linux/numactl/patch_source.sh
+++ b/third-party/sysdeps/linux/numactl/patch_source.sh
@@ -12,5 +12,22 @@ sed -i 's/libnuma\.la/librocm_sysdeps_numa.la/g' "$NUMACTL_MAKEFILE"
 sed -i 's/libnuma_la_SOURCES/librocm_sysdeps_numa_la_SOURCES/g' "$NUMACTL_MAKEFILE"
 sed -i 's/libnuma_la_LDFLAGS/librocm_sysdeps_numa_la_LDFLAGS/g' "$NUMACTL_MAKEFILE"
 
+# numactl's configure unconditionally links -latomic via
+# AC_SEARCH_LIBS([__atomic_fetch_and_1], [atomic]). On the toolchains we build
+# with, the compiler lowers libnuma's byte-sized atomics inline, so libatomic is
+# never actually used and the probe only leaves an unused libatomic.so.1
+# dependency.
+#
+# NOTE: this is unconditional, so it assumes the target toolchain lowers these
+# atomics inline. On a toolchain that emits them out-of-line (the one case seen
+# is RISC-V + GCC; RISC-V + LLVM lowers inline), libatomic must be linked again.
+_libatomic_probe='^AC_SEARCH_LIBS(\[__atomic_fetch_and_1\],[[:space:]]*\[atomic\])$'
+if ! grep -q "$_libatomic_probe" "$SOURCE_DIR/configure.ac"; then
+  echo "Error: expected libatomic probe not found in numactl configure.ac; the" \
+       "source may have changed and this patch needs revisiting" >&2
+  exit 1
+fi
+sed -i "/$_libatomic_probe/d" "$SOURCE_DIR/configure.ac"
+
 sed -i -E 's|\b(libnuma_)|AMDROCM_SYSDEPS_1.0_\1|' $SOURCE_DIR/versions.ldscript
 sed -i -E 's|@(libnuma_)|@AMDROCM_SYSDEPS_1.0_\1|' $SOURCE_DIR/*.c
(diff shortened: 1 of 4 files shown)


marbre (commented):
Instead of patching the install, why not patching the source (which we already do anyway)?

Adding
```
sed -i '/^AC_SEARCH_LIBS(\[__atomic_fetch_and_1\], \[atomic\])$/d' "$SOURCE_DIR/configure.ac"
```
to `third-party/sysdeps/linux/numactl/patch_source.sh` might already be sufficient. Wdyt?

darren-amd:
Good idea, I was considering that but couldn't think of a way to conditionally do it at build time. It'd be fine for us though since our current toolchain lowers these atomics inline, so I'll add a note in case we change in the future (the only case I saw where this was necessary was RISC-V + GCC, RISC-V + LLVM was fine).

third-party/sysdeps/linux/numactl/patch_source.sh
```diff
@@ -12,5 +12,22 @@ sed -i 's/libnuma\.la/librocm_sysdeps_numa.la/g' "$NUMACTL_MAKEFILE"
...
+#
+# NOTE: this is unconditional, so it assumes the target toolchain lowers these
+# atomics inline. On a toolchain that emits them out-of-line (the one case seen
+# is RISC-V + GCC; RISC-V + LLVM lowers inline), libatomic must be linked again.
+_libatomic_probe='^AC_SEARCH_LIBS(\[__atomic_fetch_and_1\],[[:space:]]*\[atomic\])$'
+if ! grep -q "$_libatomic_probe" "$SOURCE_DIR/configure.ac"; then
+  echo "Error: expected libatomic probe not found in numactl configure.ac; the" \
+       "source may have changed and this patch needs revisiting" >&2
+  exit 1
+fi
```
marbre: If you want to ensure this is applied let's take it at an opportunity to move this into a proper patch file as we have it at least for a few third-party libraries.`patch` will fail if the patchfile cannot be applied.

marbre:
> Good idea, I was considering that but couldn't think of a way to conditionally do it at build time.

Doesn't need to be conditionally does it?

third-party/sysdeps/linux/numactl/patch_source.sh
darren-amd: Sounds good, moved to a patch file

marbre (approved):

PR #5561, opened 2026-06-01
[ci] Adding `determine_rocm_test_dependencies.py` into multi arch CI ecosystem

geomin12:
In order to select which tests to run (use cases: rocm-libraries, rocm-systems, submodules, etc), we need to incorporate `determine_rocm_test_dependencies.py` into the multi arch CI ecosystem.

This is progress for #3343 .

Changes:
- including `determine_rocm_test_dependencies.py` script, where if changed_projects is passed in, it selects the correct tests. Example working here: https://github.com/ROCm/rocm-libraries/actions/runs/26775788121
     - Otherwise, it just defaults to all tests (TheRock CI use-case) and passes `*` to fetch_test_configurations.py
- Added inclusion of both `rocblas` or `projects/rocblas` naming convention

Next:
- with #5449 and this PR, we will introduce this to multi arch CI post-submit

progress on https://github.com/ROCm/TheRock/issues/3491


TorreZuk:
Is this progress on https://github.com/ROCm/TheRock/issues/3491 as well?

geomin12:
> Is this progress on #3491 as well?

yes!

.github/workflows/test_artifacts.yml
```diff
@@ -131,6 +139,15 @@ jobs:
         with:
           python-version: 3.12
 
+      - name: "Determine test dependencies"
+        id: test_deps
+        env:
+          CHANGED_PROJECTS: ${{ inputs.changed_projects }}
+        run: |
+          python ./test_tools/determine_rocm_test_dependencies.py \
```
jayhawk-commits, line 135: Please watch the shell used for this command in Windows, as powershell treats these backslashes differently than bash.

jayhawk-commits (commented):
The `changed_projects` workflow input needs to be added to the top-level `multi_arch_ci.yml` workflow.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -105,18 +110,11 @@ def main():
...
-        help="Changed subproject(s)",
-    )
-    parser.add_argument(
-        "--projects",
-        type=str,
-        nargs="+",
+        nargs="*",
         metavar="PROJECT",
-        help="Alias for --changed",
+        help="Project(s) to test. Accepts 'rocblas' or 'projects/rocblas' format.",
```
erman-gurses: nargs="*" expects space-separated values. However, multi_arch_ci.yml documents changed_projects as comma-separated. Is that intentional? A value like rocblas,hipblas would be parsed as a single project instead of two projects.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -142,15 +145,25 @@ def main():
         print(json.dumps(result, indent=2))
         return
 
-    changed = args.changed or args.projects
+    # Normalize path format (projects/rocblas -> rocblas)
+    changed = args.changed_projects
+    if changed:
+        changed = [Path(p).name for p in changed]
```
erman-gurses: Should we validate that the input is either <project> or projects/<project>? Path(p).name will silently strip arbitrary path components, which may hide malformed input.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -142,15 +145,25 @@ def main():
...
+    # If no projects specified, output "*" for all tests
     if not changed:
-        parser.error(
-            "one of the following arguments is required: --changed, --projects"
-        )
+        if args.gha_output:
+            gha_set_output({"projects_to_test": "*"})
+        else:
+            print("*")
+        return
```
erman-gurses, line 177: Was it intentional that --changed-projects with no values is treated the same as omitting the option entirely? That makes sense for CI, but it's a slightly unusual CLI behavior and may hide accidental invocation errors.

test_tools/tests/determine_rocm_test_dependencies_test.py
```diff
@@ -1,115 +1,55 @@
...
-        self.assertEqual(test_deps["rocwmma"], [])
-
-        # Verify hipRAND has empty TEST_SUBPROJECTS (tests only itself)
-        self.assertIn("hiprand", test_deps)
-        self.assertEqual(test_deps["hiprand"], [])
-
-        # Verify hipFFT has empty TEST_SUBPROJECTS (tests only itself)
-        self.assertIn("hipfft", test_deps)
-        self.assertEqual(test_deps["hipfft"], [])
+        self.assertEqual(test_deps["rocwmma"], [])  # empty TEST_SUBPROJECTS
```
erman-gurses: The new CLI-focused tests are useful, but this change significantly reduces coverage of real dependency mappings (rocsparse, rocprim, rocfft, etc.), leaving only a small subset verified. Could we keep one or two additional representative dependency-resolution assertions so we still catch regressions in the CMake parsing logic?

build_tools/github_actions/detect_external_repo_config.py
```diff
@@ -425,12 +425,20 @@ def main(argv=None):
             .replace("THEROCK_", "")
             .replace("_SOURCE_DIR", "")
         )
+
+        # Extract projects from external_repo JSON if provided
+        projects = ""
+        if args.external_repo_json:
+            external_repo = json.loads(args.external_repo_json)
+            projects = external_repo.get("projects", "")
```
erman-gurses: Should we handle malformed external_repo_json here? This code path will now raise a JSONDecodeError before reaching the existing config generation logic.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -142,15 +145,25 @@ def main():
...
 
     result = get_subprojects_to_test(changed, therock_dir)
+    projects_to_test = ",".join(sorted(result))
```
erman-gurses, line 180: Would it be worth adding a test for the emitted projects_to_test format (comma-separated list)? That would help catch accidental changes that could break workflow consumers.

erman-gurses (commented):
Looks good overall  - please see my comments below:

.github/workflows/test_artifacts.yml
```diff
@@ -74,6 +78,10 @@ on:
       windows_hip_rocr_tests:
         type: string
         default: 'false'
+      changed_projects:
+        description: 'Space-separated list of projects to test (e.g., "rocblas hipblas" or "projects/rocblas").'
+        type: string
+        default: ""
```
tony-davis, line 87: `multi_arch_ci.yml` documents this input as **comma-separated**, but here (and in `multi_arch_ci_linux.yml` / `multi_arch_ci_windows.yml`) it's documented as **space-separated**. The script tolerates both today, but the contradicting contract is what forces the unquoted shell expansion below. Suggest standardizing on comma-separated everywhere so the value can be safely quoted:

```suggestion
      changed_projects:
        description: 'Comma-separated list of projects to test (e.g., "rocblas,hipblas" or "projects/rocblas"). Empty = run all tests.'
        type: string
        default: ""
```

(Apply the same wording to the `workflow_dispatch` block above and to `multi_arch_ci_linux.yml` / `multi_arch_ci_windows.yml`.)

tony-davis (commented):
Thanks for wiring this through — the structure is clean: logic lives in a script rather than inline YAML, reusable inputs use `inputs.*`, `shell: bash` is set for Windows, and the empty-default → run-all fallback keeps every existing caller of `test_artifacts.yml` (including `ci_linux.yml`, `ci_windows.yml`, and the `multi_arch_release_*` workflows, which aren't updated here) backward-compatible. The new subprocess-based tests are real coverage, not padding.

A few non-blocking things I'd suggest tightening — inline. The two I care most about:
1. The `changed_projects` input format is documented inconsistently across the workflow chain (comma vs space), which makes the shell-quoting behavior fragile.
2. There's no guard against an unknown/misspelled project name silently selecting zero tests — the dangerous failure mode for a test-*selection* mechanism (false green).

.github/workflows/test_artifacts.yml
```diff
@@ -131,6 +139,16 @@ jobs:
...
 
+      - name: "Determine test dependencies"
+        id: test_deps
+        shell: bash
+        env:
+          CHANGED_PROJECTS: ${{ inputs.changed_projects }}
+        run: |
+          python ./test_tools/determine_rocm_test_dependencies.py \
+            --changed-projects $CHANGED_PROJECTS \
+            --gha-output
```
tony-davis, line 137: With the format standardized to comma-separated (see the input description), `$CHANGED_PROJECTS` should be quoted (shellcheck SC2086). I traced it through the script: quoting preserves the empty → `*` behavior (empty string flattens to `[]` → wildcard) and the comma list is split inside the script. Note this is only correct once the input is comma-separated — a quoted space-separated value would arrive as one token and not split.

```suggestion
          python ./test_tools/determine_rocm_test_dependencies.py \
            --changed-projects "$CHANGED_PROJECTS" \
            --gha-output
```

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -142,15 +146,33 @@ def main():
...
+    if changed:
+        # Flatten comma-separated values (e.g., "rocblas,hipblas" -> ["rocblas", "hipblas"])
+        flattened = []
+        for item in changed:
+            flattened.extend(p.strip() for p in item.split(",") if p.strip())
+        changed = flattened
+
+    # Normalize path format: "projects/rocblas" -> "rocblas"
+    if changed:
+        changed = [p.removeprefix("projects/") for p in changed]
```
tony-davis, line 160: An unknown or misspelled project (e.g. `rocblass`) flows straight through — `get_subprojects_to_test` seeds the result with the input name itself, so it just matches zero tests downstream and is silently dropped. For a test-selection tool that's a false-green risk. Suggest surfacing it (a warning, so we don't reject a valid project that legitimately has no `TEST_SUBPROJECTS` declaration — switch to `parser.error(...)` if you'd prefer strict):

```suggestion
    # Normalize path format: "projects/rocblas" -> "rocblas"
    if changed:
        changed = [p.removeprefix("projects/") for p in changed]
        # Surface typos so a misspelled project doesn't silently select zero tests
        known = set(parse_cmake_test_subprojects(therock_dir).keys())
        unknown = [p for p in changed if p.lower() not in known]
        if unknown:
            print(
                f"Warning: unrecognized project(s) {unknown}; "
                "no tests will be selected for them",
                file=sys.stderr,
            )
```

(Minor: this re-parses CMake once more; `get_subprojects_to_test` parses again right after. Fine for a CI script, but if you'd rather, compute `test_deps` once and pass it down.)

test_tools/tests/determine_rocm_test_dependencies_test.py
```diff
@@ -1,27 +1,28 @@
...
 
-sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
+THEROCK_DIR = Path(__file__).parent.parent.parent
+SCRIPT = Path(__file__).parent.parent / "determine_rocm_test_dependencies.py"
+
+sys.path.insert(0, str(THEROCK_DIR / "test_tools"))
 
 from determine_rocm_test_dependencies import (
     parse_cmake_test_subprojects,
     get_subprojects_to_test,
```
tony-davis, line 16: The refactor dropped the `list_subprojects` import and its tests, but `--list-subprojects` is still a shipped CLI path — it's now untested. Could we keep a small assertion so it stays covered? Re-add the import:

```suggestion
    get_subprojects_to_test,
    list_subprojects,
```

...and add a test, e.g.:

```python
def test_list_subprojects_with_show_deps(self):
    """list_subprojects returns names, and deps/'empty' when show_deps=True."""
    names = list_subprojects(THEROCK_DIR, show_deps=False)
    self.assertIn("rocblas", names)
    self.assertIn("rocwmma", names)

    deps = list_subprojects(THEROCK_DIR, show_deps=True)
    self.assertEqual(deps["rocblas"], ["hipblas", "rocsolver"])
    self.assertEqual(deps["rocwmma"], "empty")
```

test_tools/determine_rocm_test_dependencies.py
geomin12: added comma separated support

test_tools/tests/determine_rocm_test_dependencies_test.py
geomin12: added more tests to support cmake parsing!

build_tools/github_actions/detect_external_repo_config.py
geomin12: added try catch here for that! good call

test_tools/determine_rocm_test_dependencies.py
geomin12: added a test!

test_tools/determine_rocm_test_dependencies.py
geomin12: i added a check for that case, so if it's malformed (like projects/hello/hello), it'll ignore

test_tools/determine_rocm_test_dependencies.py
geomin12: indeed intentional. for CI, we just want to run everything (opt-in). this is primarily for CI and github actions, so i am going to leave it be. however, if I see later issues of local use for developers, we can update accordingly

.github/workflows/test_artifacts.yml
tony-davis: Resolved in `da1ec47`. `changed_projects` is now documented as comma-separated everywhere — both `test_artifacts.yml` blocks (`workflow_dispatch` + `workflow_call`) plus `multi_arch_ci.yml`, `multi_arch_ci_linux.yml`, and `multi_arch_ci_windows.yml` — so the contract is consistent across the chain. Thanks!

.github/workflows/test_artifacts.yml
tony-davis: Resolved in `da1ec47` — now `--changed-projects "$CHANGED_PROJECTS"`. With the input standardized to comma-separated, the quoted value splits correctly inside the script and the empty → `*` (run-all) behavior is preserved.

test_tools/determine_rocm_test_dependencies.py
tony-davis: Resolved in `da1ec47`. Unrecognized projects now surface a stderr warning (`Warning: unrecognized project(s) ...; no tests will be selected for them`) rather than silently selecting zero tests. Kept it as a warning (not `parser.error`) so a valid project with no `TEST_SUBPROJECTS` declaration isn't rejected, and added `test_unknown_project_warning` to lock it in.

test_tools/tests/determine_rocm_test_dependencies_test.py
tony-davis: Resolved in `da1ec47` — `list_subprojects` coverage is restored via `test_list_subprojects` (asserts names without deps, and the deps dict / `"empty"` sentinel with `show_deps=True`), so the `--list-subprojects` CLI path stays covered.

tony-davis (approved):
Approving — thanks for the thorough follow-up. All four of my comments are addressed in `da1ec47`: comma-separated format standardized across the workflow chain, `$CHANGED_PROJECTS` quoted, unknown-project names now warn instead of silently selecting zero tests, and `list_subprojects` is covered again. The new subprocess-based tests exercise real CLI behavior (empty→`*`, comma parsing, prefix normalization, GHA output, unknown-project warning), and the empty-default keeps the un-updated `test_artifacts.yml` callers backward-compatible. Nice work.

erman-gurses (approved):
LGTM!

jayhawk-commits (approved):
Previous concerns were addressed. LGTM.

PR #5568, opened 2026-06-01
[ci] Setting AWS build runners to 0%

geomin12:
Due to testing in production, we are setting AWS build runners to 0


geomin12:
no longer needed

PR #5580, opened 2026-06-02
dockerfiles: use NINJA_VERSION instead of CMAKE_VERSION for ./install_ninja.sh

scottt:
This trivial patch changes the Ninja version string in `build_manylinux_x86_64.Dockerfile` from the currently misnamed `CMAKE_VERSION` to `NINJA_VERSION`.

@stellaraccident  reused  `CMAKE_VERSION` as the name for the variable passed to `./install_ninja.sh` in 02db71e648 for `build_manylinux_x86_64.Dockerfile`. `build_manylinux_rccl_x86_64.Dockerfile` then inherited that when it was created in cd0270b8be

# Motivation

I create Dockerfiles for more recent Linux distro releases and build TheRock in them to stay one step ahead of breakage from e.g. newer toolchains. Having the CMAKE_VERSION and NINJA_VERSION be distinctly named makes my scripts to sync up the build dependencies easier.

## Submission Checklist

### DIFF
--- a/dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
@@ -37,9 +37,9 @@ RUN ./install_cmake.sh "${CMAKE_VERSION}" && rm -rf /install-cmake
 
 ######## Ninja ########
 WORKDIR /install-ninja
-ENV CMAKE_VERSION="1.12.1"
+ENV NINJA_VERSION="1.12.1"
 COPY install_ninja.sh ./
-RUN ./install_ninja.sh "${CMAKE_VERSION}" && rm -rf /install-ninja
+RUN ./install_ninja.sh "${NINJA_VERSION}" && rm -rf /install-ninja
 
 ######## AWS CLI ######
 WORKDIR /install-awscli
--- a/dockerfiles/build_manylinux_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_x86_64.Dockerfile
@@ -42,9 +42,9 @@ RUN ./install_cmake.sh "${CMAKE_VERSION}" && rm -rf /install-cmake
 
 ######## Ninja ########
 WORKDIR /install-ninja
-ENV CMAKE_VERSION="1.12.1"
+ENV NINJA_VERSION="1.12.1"
 COPY install_ninja.sh ./
-RUN ./install_ninja.sh "${CMAKE_VERSION}" && rm -rf /install-ninja
+RUN ./install_ninja.sh "${NINJA_VERSION}" && rm -rf /install-ninja
 
 ######## AWS CLI ######
 WORKDIR /install-awscli


JeniferC99 (changes requested):
The rename is incomplete: the `RUN ./install_ninja.sh "${CMAKE_VERSION}"` line on the next row still references `${CMAKE_VERSION}`, which is still defined (`"3.27.9"`) from the CMake block above.

dockerfiles/build_manylinux_x86_64.Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN ./install_cmake.sh "${CMAKE_VERSION}" && rm -rf /install-cmake
 
 ######## Ninja ########
 WORKDIR /install-ninja
-ENV CMAKE_VERSION="1.12.1"
+ENV NINJA_VERSION="1.12.1"
 COPY install_ninja.sh ./
 RUN ./install_ninja.sh "${CMAKE_VERSION}" && rm -rf /install-ninja
```
JeniferC99: Needs to be updated to match the renamed `ENV` two lines above:

```diff
-RUN ./install_ninja.sh "${CMAKE_VERSION}" && rm -rf /install-ninja
+RUN ./install_ninja.sh "${NINJA_VERSION}" && rm -rf /install-ninja
```

As-is, `${CMAKE_VERSION}` still expands to `"3.27.9"` from the CMake block above, so `install_ninja.sh` will try to download `ninja v3.27.9` and fail.

dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN ./install_cmake.sh "${CMAKE_VERSION}" && rm -rf /install-cmake
 
 ######## Ninja ########
 WORKDIR /install-ninja
-ENV CMAKE_VERSION="1.12.1"
+ENV NINJA_VERSION="1.12.1"
 COPY install_ninja.sh ./
 RUN ./install_ninja.sh "${CMAKE_VERSION}" && rm -rf /install-ninja
```
JeniferC99: Same fix needed here — `${CMAKE_VERSION}` should be `${NINJA_VERSION}` so the install script gets `"1.12.1"`, not the cmake version inherited from earlier in the file.

scottt:
> The rename is incomplete: the `RUN ./install_ninja.sh "${CMAKE_VERSION}"` line on the next row still references `${CMAKE_VERSION}`, which is still defined (`"3.27.9"`) from the CMake block above.

@JeniferC99 , thanks for catching this! Resolved in the latest version.

JeniferC99 (approved):
LGTM

JeniferC99 (changes requested):
sorry actually, please don't fork it, instead branch it from main so the CI can run naturally.

JeniferC99 (approved):
Looks like I can approve the workflow to run manually. back to +1

scottt:
@JeniferC99 , since all CI check passed. could you press the Github "merge" button for me as I can't do it myself as an external contributor?

PR #5584, opened 2026-06-02
Add a Nix flake defining devShells

slinder1:
This would otherwise live in my own nix configs, but it could be useful generally for anyone already using nix as a package manager. I have not tested on NixOS, so there may be some impurities that would need to be addressed before it worked there.

This assumes the "nix-command" and "flakes" experimental features are enabled in nix/lix, as I didn't factor things out into `default.nix` or `shell.nix`: everything is defined in the flake. This has the side-effect of ensuring inputs are pinned via `flake.lock`, which should mean less surface area for "works on my machine" type impurity bugs.

I hard-coded the supported families and didn't confirm the supported systems is accurate for TheRock. I assume "nightly" for the venv rocm packages, but that could be changed or made configurable too.

## Motivation

Actually developing out of a docker container is clunky, and nix can achieve much of the isolation and reproducibility docker offers. There doesn't need to be a commitment to maintaining the flake, but it would be nice to have as an option.

## Technical Details

See https://github.com/NixOS/nix for the package manager itself.

## Test Plan

I ran the non-family and a couple family configurations through with `--ignore-environment` to try to ensure there are no impurity bugs, but I don't have a grasp on how best to rigorously test the flake automatically. My intention is to support this provisionally (since I am using it myself) and if it benefits others it can be supported more comprehensively.

## Test Result

No issues in the cases I tested. I was able to build a subset of TheRock, the `venv` and `ccache` setup seemed right, and the helper scripts worked.

I also ran `flake-checker` and it doesn't flag any issues:

```
$ nix run nixpkgs#flake-checker
Flake checker results:

The flake checker scanned your flake.lock and didn't identify any issues. All
Nixpkgs inputs:

> Use supported branches
> Are less than 30 days old
> Use upstream Nixpkgs

```

And with the change in `posix_ccache_compiler_check.py` the smoke test in `build_tools/hack/ccache/test_ccache_sanity.sh` shows a cache hit:

```
❄️ nix-shell-env $ build_tools/hack/ccache/test_ccache_sanity.sh 
Cacheable calls:     1 /    1 (100.0%)
  Hits:              1 /    1 (100.0%)
    Direct:          1 /    1 (100.0%)
    Preprocessed:    0 /    1 ( 0.00%)
  Misses:            0 /    1 ( 0.00%)
```

## Submission Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,140 @@
+{
+  description = "TheRock (The HIP Environment and ROCm Kit)";
+
+  inputs = {
+    nixpkgs.url = "https://channels.nixos.org/nixpkgs-unstable/nixexprs.tar.xz";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    flake-root.url = "github:srid/flake-root";
+  };
+
+  outputs =
+    inputs@{ flake-parts, flake-root, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } rec {
+      imports = [
+        flake-root.flakeModule
+      ];
+      systems = [
+        "x86_64-linux"
+        "x86_64-darwin"
+        "aarch64-linux"
+        "aarch64-darwin"
+      ];
+      flake = rec {
+        # TODO: currently we assume "nightly" releases for venv rocm packages,
+        # but we could expose every "channel"
+        families = [
+          "gfx101X-dgpu"
+          "gfx103X-all"
+          "gfx103X-dgpu"
+          "gfx110X-all"
+          "gfx110X-dgpu"
+          "gfx1150"
+          "gfx1151"
+          "gfx1152"
+          "gfx1153"
+          "gfx900"
+          "gfx906"
+          "gfx908"
+          "gfx90a"
+          "gfx120X-all"
+          "gfx90X-dcgpu"
+          "gfx94X-dcgpu"
+          "gfx950-dcgpu"
+        ];
+        forAllFamilies = f: inputs.nixpkgs.lib.genAttrs families (family: f family);
+      };
+      perSystem =
+        { config, pkgs, ... }:
+        let
+          mkFamilyShell =
+            family:
+            let
+              venvPath = if family == "" then ".venv" else ".venv.${family}";
+              ifFamily = s: if family == "" then "" else s;
+            in
+            pkgs.mkShellNoCC {
+              inputsFrom = [ config.flake-root.devShell ];
+              packages = with pkgs; [
+                autoconf
+                automake
+                bison
+                ccache
+                cmake
+                dvc
+                flex
+                gfortran # also includes g++, etc.
+                git
+                libtool
+                ncurses # just for libtinfo, since therock builds its own ncurses
+                ninja
+                pkg-config
+                python3
+                texinfo
+                (stdenv.mkDerivation rec {
+                  pname = "patchelf-rocm";
+                  version = "d0f70eea5397606c486857e0a105e53ec123904a";
+
+                  src = fetchGit {
+                    url = "https://github.com/NixOS/${pname}";
+                    rev = "${version}";
+                  };
+
+                  patchPhase = ''
+                    PATCHELF_GIT_REF="${version}"
+                    SHORT_GIT_REF="''${PATCHELF_GIT_REF:0:12}"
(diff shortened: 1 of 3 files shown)


slinder1:
(Hopefully I'm not spamming anyone; I don't see a `CODEOWNERS` and the `CONTRIBUTING` says to look for recent contributors, so I just added the top-N contributors from the default GitHub view)

slinder1:
I didn't actually check the `ccache` hit rates, it seems like `compiler_check.py` is always failing within the devshell. I will look into it when I have time, but this PR is WIP until then

slinder1:
I think this is ready for review now. The test failures seem unrelated, not sure if I should try rerunning them

ScottTodd:
Not sure if we'll want to accept these Nix-related files here, will need to spend some time reading up and reviewing...

> The test failures seem unrelated, not sure if I should try rerunning them

Fine to ignore. Could even add the `ci:skip` label (see https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request), as these changes are unlikely to affect CI.

slinder1:
> Not sure if we'll want to accept these Nix-related files here, will need to spend some time reading up and reviewing...
> 

Very fair, and I don't want to demand too much time from maintainers. I can continue to use this as part of my own dotfiles if this is too fringe.

> > The test failures seem unrelated, not sure if I should try rerunning them
> 
> Fine to ignore. Could even add the `ci:skip` label (see https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request), as these changes are unlikely to affect CI.

TIL, thanks!

flake.nix
```diff
@@ -0,0 +1,143 @@
...
+                  nativeBuildInputs = [ autoreconfHook ];
+                })
+                (pkgs.writeShellApplication {
+                  name = "therock-update-python";
+                  text = ''
+                    set -x
+                    cd "$FLAKE_ROOT"
+                    pip install --upgrade pip
+                    pip install --upgrade -r requirements.txt
+                    ${ifFamily "pip install --upgrade 'rocm[libraries,devel]' --index-url=https://rocm.nightlies.amd.com/v2/${family}"}
```
ScottTodd, line 99: These per-family Python packages have been replaced with a single multi-arch index, see https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-multi-arch-rocm-python-packages

before;
```
pip install \
  --index-url https://rocm.nightlies.amd.com/v2/gfx1151/ \
  "rocm[libraries,devel]"
```

after:
```
pip install \
  --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
  "rocm[libraries,device-gfx1151]"
```

Instead of a separate index per package, there is one index with install extras that pull in device code packages. Also note that the device code packages are per _target_, not per _family_ (so gfx1100, gfx1101, gfx1102, etc. instead of gfx110X-all)

flake.nix
```diff
@@ -0,0 +1,143 @@
+{
+  description = "TheRock (The HIP Environment and ROCm Kit)";
+
+  inputs = {
+    nixpkgs.url = "https://channels.nixos.org/nixpkgs-unstable/nixexprs.tar.xz";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    flake-root.url = "github:srid/flake-root";
+  };
```
ScottTodd, line 8: There could be value in sharing these between users by contributing to the common repository, but I'd be hesitant to add the files to the top level directory. Would these still work from a subdirectory? We have https://github.com/ROCm/TheRock/tree/main/build_tools/hack for 
> One-off scripts intended to aid in the infrastructure management of TheRock. Once more than a couple users are using these scripts they should graduate out of this directory into their own scripts directory which is more properly maintained.

I could also see a top level directory like what LLVM has for Bazel: https://github.com/llvm/llvm-project/tree/main/utils/bazel for "files that a subset of users care about but are not fully supported or used for all CI/CD"

slinder1: Yes, with some tweaks it doesn't need to be at the root. Honestly it would be simpler to drop `flake-parts` and `flake-root` anyway, and I can instead inline a way to find the root of the repo.

marbre:
Converting to draft as this is still based on single-arch.

PR #5588, opened 2026-06-03
[ci] Fixing artifacts for `sparse` and `solver`

geomin12:
Noticed that both `sparse` and `solver` components write to inconsistent `_*_subproject_names`, so we are now making sure these components write to its respective toml

From [old CI run](https://therock-ci-artifacts.s3.amazonaws.com/26766793639-linux/index.html) compared to a [test CI run with these changes](https://therock-ci-artifacts.s3.amazonaws.com/26842458187-linux/index.html), we see a diff in sizes:
- `blas_lib_generic`: 33 MB vs 18 MB
- `blas_test_generic`: 484 MB vs 74 MB
(obviously the tests are going to their respective components)

With this update, specific components (such as rocblas) only need its necessary components and doesn't install unnecessary components


marbre (commented):
hipBLAS builds and depends on rocSOLVER:
https://github.com/ROCm/TheRock/blob/b14ee1d04c7adfb7a82d77038cce20a514335cbb/math-libs/BLAS/CMakeLists.txt#L497

So I think we're missing something here.

nunnikri:
Superseded by https://github.com/ROCm/TheRock/pull/7607

geomin12:
closing for #7607

Issue #5596, opened 2026-06-03
[Issue] Linux PyTorch gfx950-dcgpu: 8 test failures on torch nightly (2.13)


amd-chiranjeevi:
## Context
- Workflow: Release portable Linux PyTorch Wheels
- Workflow file: `.github/workflows/release_portable_linux_pytorch_wheels.yml`
- Failing run: [↗ View run](https://github.com/ROCm/TheRock/actions/runs/26809392716)
- Platform: `Linux`
- Impacted Arch: `gfx950-dcgpu`
- PyTorch Version: `nightly` (`2.13.0a0+rocm7.14.0a20260602`)
- Python Version: `3.13, 3.11, 3.10, 3.14`

## Failed Tests (8)
```
test_nn.py::TestNNDeviceTypeCUDA::test_CTCLoss_cudnn_cuda
test_nn.py::TestNNDeviceTypeCUDA::test_LSTM_dropout_per_call_randomness_dropout_p_0_5_training_True_cuda
test_nn.py::TestNNDeviceTypeCUDA::test_ctc_loss_cudnn_tensor_cuda_cuda
test_nn.py::TestNNDeviceTypeCUDA::test_upsamplingNearest2d_launch_rocm_cuda
test_cuda.py::TestCuda::test_hip_device_count
test_cuda.py::TestCudaAllocator::test_memory_compile_regions
test_cuda.py::TestMemPool::test_mempool_empty_cache_inactive
test_cuda.py::TestMemPool::test_mempool_limited_memory_with_allocator
```

## Root Cause
The skip-list loader in `run_pytorch_tests.py` detects torch version `2.13` and tries to load `external-builds/pytorch/skip_tests/pytorch_2.13.py`. That file does not exist, so only `generic.py` is loaded. All the tests above are already marked as known failures in the stable-version skip files (`pytorch_2.9.py` – `pytorch_2.12.py`), but those exclusions do not apply to nightly.

The most structurally notable error (two test failures):
```
/__w/TheRock/TheRock/.venv/lib/python3.13/site-packages/torch/include/ATen/hip/Exceptions.h:5:10: fatal error: 'hipblas/hipblas.h' file not found
    5 | #include <hipblas/hipblas.h>
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
FAILED [7.2476s]  test_cuda.py::TestMemPool::test_mempool_limited_memory_with_allocator
```

[...]

Issue #2156, opened 2025-11-14
Stabilize PyTorch release workflows


ScottTodd:
Our nightly release workflows have been unstable recently for several overlapping reasons. This parent issue will help us keep track of known issues.

See the scheduled workflow history at:
* https://github.com/ROCm/TheRock/actions/workflows/release_portable_linux_pytorch_wheels.yml?query=branch%3Amain
* https://github.com/ROCm/TheRock/actions/workflows/release_windows_pytorch_wheels.yml?query=branch%3Amain

> [!IMPORTANT]
>  Due to [Gating releases with PyTorch tests](https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#gating-releases-with-pytorch-tests), ongoing test failures have blocked promotion from "v2-staging" to "v2" in our nightly release index for several weeks now. As such, the [Installing PyTorch Python packages](https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-pytorch-python-packages) release documentation currently only allows for downloading the older packages that passed (or did not run) tests.

jayhawk-commits:
@chiranjeevipattigidi created new issues from pytorch failures late last week, should we add them as sub-issues to this?

ScottTodd:
Enough of the sub-issues have been closed that I think we can call this "fixed" now. There will continue to be ongoing work to keep PyTorch tests stable.

PR #5600, opened 2026-06-03
test: add pytorch_2.13 skip list for known test failures

ethanwee1:
## Summary

Adds a `pytorch_2.13` skip list that mirrors test failures already tracked in the
other per-version skip files (`pytorch_2.9.py`–`pytorch_2.12.py`) and
`generic.py`. Those exclusions are not picked up automatically for the 2.13
wheels, so they are mirrored here to keep full PyTorch test runs against 2.13
from re-flagging already-known failures.

Related: #5596
Parent issue: #2156

## Test plan

- The entries are not newly discovered failures: each is copied from an existing,
  already-tracked skip file (`pytorch_2.9.py`–`pytorch_2.12.py` / `generic.py`),
  with the originating category and reason preserved in inline comments.
- Skip-list selection is version-keyed (`detect_pytorch_version()` → `major.minor`),
  so `pytorch_2.13.py` is picked up automatically for 2.13 wheels when running
  `.github/workflows/test_pytorch_wheels.yml` / the full PyTorch test suite — no
  wiring change is required to activate it.


external-builds/pytorch/skip_tests/pytorch_2.13.py
```diff
@@ -0,0 +1,43 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Known failures on the PyTorch nightly (2.13) wheels. These are already
```
ScottTodd: Torch "nightly" is now up to 2.14. Delete that part of the comment and update the PR title?

ScottTodd (approved):
Fine to add new skip test files like this.

This PR could use a description though:
* Link to an associated issue (https://github.com/ROCm/TheRock/issues/2156 as the parent issue for pytorch release testing, or another [sub]issue)
* Test plan (did you run torch tests locally with this skip list? did you run https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml with it?)

external-builds/pytorch/skip_tests/pytorch_2.13.py
ethanwee1: Done — dropped the stale "nightly" wording from the comment (torch nightly is now 2.14), updated the PR title to "Add pytorch_2.13 skip list for known test failures", and added a PR description linking #5596 / parent #2156 with a test plan.

PR #5602, opened 2026-06-03
Pin rocm/rocm-sdk-* deps with PEP 440 === instead of ==

JeniferC99:
Switches the three rocm-sdk-* version pins in `_dist_info.py`, `external-builds/pytorch/build_prod_wheels.py`, and `build_tools/github_actions/determine_version.py` from PEP 440 `==` to `===` (arbitrary equality / literal-string match) so an index that publishes both bare `X` and local-segmented `X+<local>` silicon-rev variants of the same rocm-sdk wheel cannot satisfy these pins with a different rev than the one that produced the metapackage / torch wheel (`==X` matches both candidates and pip's resolver prefers the local-segmented one, silently cross-stamping the install). Single-rev indices remain unaffected. Currently mirrored as a sed-patch step in [rocm-npi-dev#785](https://github.com/AMD-ROCm-Internal/rocm-npi-dev/pull/785) which can be dropped once this lands and TheRock is bumped there.

### DIFF
--- a/build_tools/github_actions/determine_version.py
+++ b/build_tools/github_actions/determine_version.py
@@ -86,7 +86,11 @@ def main(argv: list[str]):
 
     parsed_version = parse(args.rocm_version)
     version_suffix = derive_version_suffix(args.rocm_version)
-    rocm_sdk_version = f"=={parsed_version}"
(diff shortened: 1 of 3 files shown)


JeniferC99:
Reopening from in-repo branch instead of fork.

PR #5604, opened 2026-06-03
Keep unversioned libhsa-runtime64.so symlink in core wheel to fix rccl dlopen bug

### LINKED ISSUE #5562: [Issue]: Ship unversioned libhsa-runtime64.so alias in _rocm_sdk_core/lib and _rocm_sdk_libraries/lib
### Problem Description

In TheRock multi-arch packaging, the unversioned `libhsa-runtime64.so` alias only ships in `_rocm_sdk_devel/lib/`. The two runtime trees consumers actually ship against: _rocm_sdk_core/lib/ (where the SONAME libhsa-runtime64.so.1 lives) and _rocm_sdk_libraries/lib/, don't carry it.

This breaks librccl for any consumer that doesn't transit `_rocm_sdk_devel/lib/`. librccl ([rocm-systems/projects/rccl/src/misc/rocmwrap.cc:~177](https://github.com/ROCm/rccl/blob/main/src/misc/rocmwrap.cc)) calls `dlopen("libhsa-runtime64.so")` (unversioned). When librccl is loaded from a non-devel tree (or in a `/opt/rocm`-less deployment), the lookup falls through every search path and fails:

`[RCCL] Failed to find ROCm runtime library in (null) (RCCL_ROCR_PATH=(null))`
This bocks any RCCL collective in /opt/rocm-less deployments

The unversioned *.so alias is a Linux convention used as a developer-time link-time hook; the SONAME *.so.<n> is the runtime contract. RCCL's unversioned dlopen is its own bug, but the packaging asymmetry is independently fixable and unblocks /opt/rocm-less deployment without waiting on librccl.

### Operating System

Ubuntu:24.04

### CPU

Not relevant

### GPU

Not relevant

### ROCm Version

7.14.0a20260601

### ROCm Component

rocm-core

### Steps to Reproduce

```
docker run --rm ubuntu:24.04 bash -lc '
  apt-get update -qq && apt-get install -y -qq python3-pip
  python3 -m pip install --break-system-packages --pre \
      --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ \
      "rocm[libraries,devel,device-gfx950]"
  rocm-sdk init

  PURELIB=$(python3 -c "import sysconfig; print(sysconfig.get_paths()[\"purelib\"])")
  for tree in _rocm_sdk_core _rocm_sdk_libraries _rocm_sdk_devel; do
    echo "--- $tree/lib ---"
    out=$(ls "$PURELIB/$tree/lib"/libhsa-runtime64* 2>/dev/null | sed "s|$PURELIB/||")
    [ -z "$out" ] && echo "(empty)" || echo "$out"
  done
'
```

Output (7.14.0a20260601)
```
--- _rocm_sdk_core/lib ---
_rocm_sdk_core/lib/libhsa-runtime64.so.1
--- _rocm_sdk_libraries/lib ---
(empty)
--- _rocm_sdk_devel/lib ---
_rocm_sdk_devel/lib/libhsa-runtime64.so
_rocm_sdk_devel/lib/libhsa-runtime64.so.1
_rocm_sdk_devel/lib/libhsa-runtime64.so.1.21.0
```

_No response_

### Additional Information

_No response_

Muhamed-Husic:
## Motivation

Closes: #5562 

Currently, rocm-sdk-core Python wheel strips out the unversioned libhsa-runtime64.so symlink. However, librccl has a bug where it [explicitly calls `dlopen("libhsa-runtime64.so")`](https://github.com/ROCm/TheRock/issues/5562#:~:text=calls%20dlopen(%22libhsa%2Druntime64.so%22)%20(unversioned).) at runtime instead of using the properly versioned .so.1 SONAME.

As noted by the issue reporter: "[RCCL's unversioned dlopen is its own bug, but the packaging asymmetry is independently fixable and unblocks /opt/rocm-less deployment without waiting on librccl.](https://github.com/ROCm/TheRock/issues/5562#:~:text=RCCL%27s%20unversioned%20dlopen%20is%20its%20own%20bug%2C%20but%20the%20packaging%20asymmetry%20is%20independently%20fixable%20and%20unblocks%20/opt/rocm%2Dless%20deployment%20without%20waiting%20on%20librccl.)"
This PR implements an exception to the packaging rule to resolve this asymmetry, immediately unblocking the dependent community while we wait for the upstream fix in RCCL.

## Technical Details

Added `ensure_core_library_symlinks()` inside `build_tools/build_python_packages.py` and invoked it immediately after `core.populate_runtime_files()`. Because `populate_runtime_files()` has a hardcoded rule to drop unversioned .so aliases from runtime packages. This new function manually adds the libhsa-runtime64.so symlink into core wheel. 
This mirrors the exact pattern already established in the script for profiler libraries (`ensure_profiler_library_symlinks()` function).

## Test Plan

1. Generated wheels using `build_python_packages.py` on existing C++ artifacts from my older local build.
2. In clean virtual env I `pip install` this newly generated `rocm-sdk-core` wheel.
3.  Run reporter's verification code inside the venv.

## Test Result

After I run reporter's verification code
```
  PURELIB=$(python3 -c "import sysconfig; print(sysconfig.get_paths()[\"purelib\"])")
  for tree in _rocm_sdk_core _rocm_sdk_libraries _rocm_sdk_devel; do
    echo "--- $tree/lib ---"
    out=$(ls "$PURELIB/$tree/lib"/libhsa-runtime64* 2>/dev/null | sed "s|$PURELIB/||")
    [ -z "$out" ] && echo "(empty)" || echo "$out"
  done
```
 inside my venv, I get the following result:
```
--- _rocm_sdk_core/lib ---
_rocm_sdk_core/lib/libhsa-runtime64.so
_rocm_sdk_core/lib/libhsa-runtime64.so.1
...
```

## Submission Checklist

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -79,13 +79,18 @@ def ensure_profiler_library_symlinks(profiler: PopulatedDistPackage) -> None:
 
 
 def ensure_core_library_symlinks(core: PopulatedDistPackage) -> None:
-    """Recreate unversioned core library symlinks expected by dlopen()."""
+    """Recreate unversioned core library symlinks expected by dlopen().
+
+    TODO(#5562): Remove this function and its call site ensure_core_library_symlinks(core) in run()
+    once upstream RCCL stops using the unversioned libhsa-runtime64.so name in its dlopen() calls.
+    """
     core_lib_dir = core.platform_dir / "lib"
 
     target = core_lib_dir / "libhsa-runtime64.so.1"
     link = core_lib_dir / "libhsa-runtime64.so"
     if target.exists() and not link.exists():
         link.symlink_to(target.name)
+
 
 def run(args: argparse.Namespace):
     manifest = load_therock_manifest(args.artifact_dir)
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -78,6 +78,15 @@ def ensure_profiler_library_symlinks(profiler: PopulatedDistPackage) -> None:
             link.symlink_to(target.name)
 
 
+def ensure_core_library_symlinks(core: PopulatedDistPackage) -> None:
+    """Recreate unversioned core library symlinks expected by dlopen()."""
+    core_lib_dir = core.platform_dir / "lib"
+
+    target = core_lib_dir / "libhsa-runtime64.so.1"
+    link = core_lib_dir / "libhsa-runtime64.so"
+    if target.exists() and not link.exists():
+        link.symlink_to(target.name)
+
 def run(args: argparse.Namespace):
     manifest = load_therock_manifest(args.artifact_dir)
     kpack_split = manifest.get("flags", {}).get("KPACK_SPLIT_ARTIFACTS", False)
@@ -137,6 +146,7 @@ def run(args: argparse.Namespace):
             ],
         ),
     )
+    ensure_core_library_symlinks(core)
 
     profiler_artifacts = params.filter_artifacts(
         profiler_artifact_filter,


build_tools/build_python_packages.py
```diff
@@ -78,6 +78,15 @@ def ensure_profiler_library_symlinks(profiler: PopulatedDistPackage) -> None:
...
+def ensure_core_library_symlinks(core: PopulatedDistPackage) -> None:
+    """Recreate unversioned core library symlinks expected by dlopen()."""
+    core_lib_dir = core.platform_dir / "lib"
+
+    target = core_lib_dir / "libhsa-runtime64.so.1"
+    link = core_lib_dir / "libhsa-runtime64.so"
+    if target.exists() and not link.exists():
+        link.symlink_to(target.name)
+
 def run(args: argparse.Namespace):
```
JeniferC99, line 95: 1. Pre-commit is currently failing on a missing blank line — there's only one blank line between `ensure_core_library_symlinks` and `def run(...)`, but two are required between top-level functions ([pre-commit run](https://github.com/ROCm/TheRock/actions/runs/26903428582/job/79435028912)).
2. Please drop a TODO referencing the upstream RCCL bug in the docstring so a future maintainer can grep for `#5562` once RCCL is fixed and confidently delete this workaround.
3. A few typos in the PR description, worth a quick pass:
   - **Motivation:** "Curently" → "Currently"; "THis PR" → "This PR"; "reporters verification code" → "reporter's verification code" (appears twice in *Test Plan* / *Test Result*).
   - **Technical Details:** "unversionded" → "unversioned"; "THis mirrors" → "This mirrors"; "funtion" → "function".

Muhamed-Husic: Thanks for the review. I've fixed all PR description typos, addressed the pre-commit blank line requirement, and added the requested TODO referencing upstream RCCL issue (#5562) for future cleanup.

marbre (changes requested):
See my feedback in #5562. We should rather fix this in RCCL.

marbre:
IMHO we should rather go with https://github.com/ROCm/rocm-systems/pull/7127.

Muhamed-Husic:
Makes perfect sense. Since you already opened PR to fix dlopen call in RCCL directly, I will close my PR. Thank you for detailed explanation.

PR #4579, opened 2026-04-15
Enable installation of vendored OpenMPI artifacts via install_rocm_from_artifacts.py

araravik-psd:
This PR adds support for installing TheRock’s vendored OpenMPI as part of artifact-based provisioning.

TheRock builds bundle OpenMPI (mpiexec and dependencies) and publish corresponding artifacts (openmpi_run, openmpi_lib, openmpi_dev). However, the current installer does not fetch these,  which are needed especially for MPI-dependent workflows like RCCL multi-node tests.

**Changes**

- Added new CLI flag:
   --mpi / --no-mpi

- When enabled, installs:
   openmpi_run (includes mpiexec)
   openmpi_lib
   openmpi_dev

- Integrated into existing artifact selection logic without changing default behavior

**Usage**
`python build_tools/install_rocm_from_artifacts.py \
  --run-id <RUN_ID> \
  --artifact-group <ARCH> \
  --output-dir <OUTPUT_DIR> \
  --rccl \
  --mpi \
  --tests`

Issue #1284, opened 2025-08-19
Add Open MPI to TheRock


marbre:
Some RCCL-tests require a Message Passing Interface (MPI) implementation to run in a multi-node set up. These tests get enabled through `USE_MPI` , which is set via `THEROCK_ENABLE_MPI` as soon as #1283 has landed.

We eventually want to vendor Open MPI as an MPI implementation with TheRock.

Open MPI can be build with ROCm-aware support (https://docs.open-mpi.org/en/main/tuning-apps/networking/rocm.html) if UCX gets included in TheRock as well and if the latter is built with ROCm support.

ronlieb:
@marbre  are you planning to use theRock build of flang/clang/clang++ to build OpenMPI ?

marbre:
> [@marbre](https://github.com/marbre) are you planning to use theRock build of flang/clang/clang++ to build OpenMPI ?

@ronlieb there are no concrete plans to build this via clang/flang. The PR bringing in Open MPI is #2599 which does _not_ use clang/flang.

ronlieb:
> > [@marbre](https://github.com/marbre) are you planning to use theRock build of flang/clang/clang++ to build OpenMPI ?
> 
> [@ronlieb](https://github.com/ronlieb) there are no concrete plans to build this via clang/flang. The PR bringing in Open MPI is [#2599](https://github.com/ROCm/TheRock/pull/2599) which does _not_ use clang/flang.

OpenMPI has fortran interfaces and exports "mod" files that are unfortuantely , NOT portable across compiler vendors and for that matter versions of the compiler.   So if you build OpenMPI with gfortran, it will be unusable to any HPC Fortran application built with flang that tries to use the OpenMPI from 2599.

stellaraccident:
@ScottTodd has had a draft pr for a long time that moves third-party/sysdeps to the top level and makes it the only thing that builds with the system compiler. Then regular third-party libraries would build with ROCm's llvm. This would also let them build with sanitizers, etc which have a similar non portability between toolchains.

Currently there is a dep ordering issue that forbids all of third-party from building with the rocm llvm (because the compiler depends on sysdeps).

This could be done with less dramatic surgery easily by just changing the top level add subdirectories to: 

```
third-party/sysdeps 
base
compiler
third-party 
core
...
```

Then third-party libs can build with the rocm llvm toolchain without physically moving anything. 

[...]

Issue #935, opened 2025-06-28
rocm Python wheels are incorrectly capturing extra requirements


ScottTodd:
Looking at recent wheels from https://d2awnip2yjpvqn.cloudfront.net/v2/gfx94X-dcgpu/, the `rocm-*/PKG-INFO` files contain buggy metadata:

For example with https://d2awnip2yjpvqn.cloudfront.net/v2/gfx94X-dcgpu/rocm-7.0.0rc20250627.tar.gz
```
Metadata-Version: 2.1
Name: rocm
Version: 7.0.0rc20250627
Requires-Dist: rocm==7.0.0rc20250627
Requires-Dist: rocm-sdk-core==7.0.0rc20250627
Provides-Extra: libraries
Requires-Dist: rocm-sdk-libraries-dev==7.0.0rc20250627; extra == "libraries"
Provides-Extra: devel
Requires-Dist: rocm-sdk-devel==7.0.0rc20250627; extra == "devel"
```

Note `Requires-Dist: rocm-sdk-libraries-dev==7.0.0rc20250627; extra == "libraries"`. The `rocm-sdk-libraries-dev` package does not exist, it should be `rocm-sdk-libraries-gfx94X-dcgpu`.

This results in package installation issues (need to confirm the exact symptoms, but there is a bug regardless).

ScottTodd:
Looking through logs to see where that `dev` comes from: https://github.com/ROCm/TheRock/actions/runs/15924743167/job/44919376780#step:8:13876 there we see
```
::: Building python package rocm: /opt/python/cp312-cp312/bin/python /therock/output/rocm/setup.py sdist -v --dist-dir /therock/output/dist
Loaded rocm dist_info: version=7.0.0rc20250627, suffix_nonce='', default_target_family='dev', available_target_families=['doc', 'gfx1151', 'run', 'dev'], packages={'meta': rocm, 'core': rocm-sdk-core, 'libraries': rocm-sdk-libraries-{target_family}, 'devel': rocm-sdk-devel}
install_requires=['rocm==7.0.0rc20250627', 'rocm-sdk-core==7.0.0rc20250627']
extras_require={'libraries': ['rocm-sdk-libraries-dev==7.0.0rc20250627'], 'devel': ['rocm-sdk-devel==7.0.0rc20250627']}
```

Compared with these (working as expected!) logs from a local Windows build:
```
[...]

PR #5627, opened 2026-06-04
fix(rocprofiler-sdk): enable rocprofv3 mpi-ranks tests in CI

minseobshin11:
## Motivation

rocprofiler-sdk's `rocprofv3` **mpi-ranks** tests are silently **disabled** unless `find_package(MPI)` succeeds at test-configure time. The shared `no_rocm` test image doesn't include an MPI implementation, so these tests never run in CI. This PR points the rocprofiler-sdk test job at the dedicated OpenMPI test image so the tests find MPI and run.

## Technical Details

- Points the `rocprofiler-sdk` test job at `ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi` via `container_image` in `build_tools/github_actions/fetch_test_configurations.py`.

Depends on #6610, which adds and publishes that image. `test_rocprofiler_sdk.py` configures and builds the tests at test time inside the container, so `find_package(MPI)` resolves the system OpenMPI and `mpiexec` is on `PATH`, enabling the mpi-ranks tests. `rocprofv3` itself does not link MPI, so no build-time MPI is required and OpenMPI stays out of the shipped ROCm artifacts.

The image is referenced by tag pending publish; it will be pinned to an `@sha256` digest once #<image PR> lands and the image is published (see `dockerfiles/README.md`).

## Test Plan

Run the `rocprofiler-sdk` test job in CI on the OpenMPI image and confirm the mpi-ranks tests are selected and executed (not `DISABLED`) and pass.

## Test Result

_CI run on this branch: the rocprofiler-sdk job runs on the OpenMPI test image and the mpi-ranks tests run + pass 

## JIRA ID

JIRA ID : AIROCVAL-45

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -487,6 +487,11 @@ test_matrix = {
         "total_shards_dict": {
             "linux": 1,
         },
+        # rocprofv3 mpi-ranks tests gate on find_package(MPI) and launch under
+        # mpiexec. OpenMPI is not bundled in TheRock artifacts and is provided via
+        # the specialized openmpi image.
+        # TODO: pin to a published @sha256 digest once the image exists.
+        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi:latest",
     },
     # hipDNN tests
     "hipdnn": {
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -490,8 +490,7 @@ test_matrix = {
         # rocprofv3 mpi-ranks tests gate on find_package(MPI) and launch under
         # mpiexec. OpenMPI is not bundled in TheRock artifacts and is provided via
         # the specialized openmpi image.
-        # TODO: pin to a published @sha256 digest once the image exists.
-        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi:latest",
+        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi@sha256:f67d0b02cae8faf0d2f3e4a1de38a01af6bad2eb27f10a5e07bf19748a84d1e6",
     },
     # hipDNN tests
     "hipdnn": {


build_tools/github_actions/test_executable_scripts/test_rocprofiler_sdk.py
```diff
@@ -43,6 +44,55 @@ def setup_env():
...
+
+    install_cmd = [
+        "sudo",
+        "apt-get",
+        "install",
+        "-y",
+        "--no-install-recommends",
+        "libopenmpi-dev",
+        "openmpi-bin",
+    ]
```
ScottTodd: If we want to use a system OpenMPI, how about doing something similar to https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04_rocgdb.Dockerfile here?

You would
* Create a new dockerfile that extends our base test dockerfile, including a build/publish workflow and documentation (see git history there for pointers). Send that as one PR so the new dockerfile is published.
* Once it's published, edit `fetch_test_configurations.py` to use it for `rocprofiler_sdk` like so: https://github.com/ROCm/TheRock/blob/1ca3dfcfe54ebcb295b8fb30b97358d665c6831f/build_tools/github_actions/fetch_test_configurations.py#L131 https://github.com/ROCm/TheRock/blob/1ca3dfcfe54ebcb295b8fb30b97358d665c6831f/build_tools/github_actions/fetch_test_configurations.py#L475-L489

build_tools/github_actions/test_executable_scripts/test_rocprofiler_sdk.py
```diff
@@ -43,6 +44,55 @@ def setup_env():
     )
 
 
+# The rocprofv3 mpi-ranks integration tests gate on find_package(MPI) at CMake
+# configure time and are disabled when no MPI implementation is present. OpenMPI
+# is not bundled in TheRock artifacts and is intentionally kept out of the shared
+# no_rocm test image to keep it minimal, so install it here (the test container
+# runs as a sudo-enabled user) before configuring.
```
ScottTodd: We do actually vendor openmpi now: https://github.com/ROCm/TheRock/tree/main/third-party/openmpi

For context, see:
* https://github.com/ROCm/TheRock/issues/1284
* https://github.com/ROCm/TheRock/pull/2599
* https://github.com/ROCm/TheRock/pull/4579

Relevant code:
* https://github.com/ROCm/TheRock/blob/1ca3dfcfe54ebcb295b8fb30b97358d665c6831f/build_tools/install_rocm_from_artifacts.py#L462-L467
* https://github.com/ROCm/TheRock/blob/1ca3dfcfe54ebcb295b8fb30b97358d665c6831f/build_tools/github_actions/fetch_test_configurations.py#L475-L489
* https://github.com/ROCm/TheRock/blob/1ca3dfcfe54ebcb295b8fb30b97358d665c6831f/build_tools/install_rocm_from_artifacts.py#L489-L493

How about you try adding `--mpi` to `fetch_artifact_args` for `rocprofiler-sdk` in `build_tools/github_actions/fetch_test_configurations.py`? You could also pass `--rocprofiler-sdk`, I'm not sure why that isn't there already...

What I would expect to happen is:
* `build_tools/install_rocm_from_artifacts.py` with some combination of arguments should pull down the mpi artifacts
* those artifacts should (if vendored and used correctly) let these mpi-ranks tests in rocprofiler_sdk be enabled and run correctly

Ideally you can test that locally on a dev machine and then get it working here on the CI, to avoid needing to wait multiple hours for builds to complete. You can also directly trigger https://github.com/ROCm/TheRock/actions/workflows/test_artifacts.yml to test changes to just the CI test scripts if you aren't actually rebuilding any project sources (see some docs at https://github.com/ROCm/TheRock/blob/main/docs/development/ci_overview.md, we can also help with the details there as needed)

minseobshin11: Hi, thanks for letting me know! 

I tried building the vendored OpenMPI locally and tested it end-to-end and it seems like the only change we will need to make is to enable ```THEROCK_ENABLE_MPI=ON``` in the ```multi_arch_build_portable_linux_artifacts.yml```. But since this is a global flag, which also turns on RCCL and rocprofiler-systems MPI, plus the extra build time, I wanted to check with you if that's ok. CMake does have a comment saying "# TODO: Set default to true for Linux branch after necessary changes land
cmake_dependent_option(THEROCK_ENABLE_MPI
  "Enables building components with Message Passing Interface (MPI) support"
  OFF "NOT WIN32" OFF)", so is there any outstanding necessary changes left?
  
  Switching ```fetch_artifact_args``` to select --mpi --rocprofiler-sdk breaks the tests as they need other dependencies, which --tests provides. So I believe enabling the build to produce openmpi is the right move.

ScottTodd: Oh, we don't enable it by default? I see that rocm-systems enables it for RCCL CI: https://github.com/ROCm/rocm-systems/blob/dbfa0b60173c3dc4e1239cf3ce4f61428172c3b8/.github/workflows/therock-ci.yml#L177-L183

I wonder what's keeping us from enabling it. Maybe it should be changed to a sysdep so our packages can use it without conflicting with other system installs? https://github.com/ROCm/TheRock/tree/main/third-party/sysdeps. If it's a test only dependency then we could also do what we did for fftw3, see https://github.com/ROCm/TheRock/blob/30f6e58beeec217db64b10707395de91d06076d1/build_tools/build_tarballs.py#L62

Thoughts @marbre ?

minseobshin11: Hi, here OpenMPI is effectively a test-only dependency. To avoid shipping it/conflicting with a user's system MPI, I added openmpi to DEFAULT_EXCLUDED_ARTIFACTS in build_tarballs.py (same as fftw3) as you mentioned. Happy to go the sysdep route instead if you and @marbre prefer that, let me know!

marbre: I am not sure if it is a pure test-dep as I think it is also picked up by rocSHMEM. Let's please verify how this behaves. If it really is only used in rccl-tests and in rocprofiler-sdk it might be okay to enable it but let's be careful.

minseobshin11: Hi, I verified with a rocSHMEM reconfigure (THEROCK_ENABLE_MPI=ON, OpenMPI present): the configure logs "External MPI detection disabled by user" and no MPI vars land in the cache. I believe that's because TheRock passes -DUSE_EXTERNAL_MPI=OFF unconditionally (comm-libs/CMakeLists.txt:155, independent of THEROCK_ENABLE_MPI), which makes rocSHMEM skip find_package(MPI) (rocshmem/CMakeLists.txt:223-227). So rocSHMEM doesn't pick up the vendored OpenMPI.

ScottTodd: I don't have great answers here, but I think we do need to make a few decisions about how MPI should be built, used, and distributed.

As I understand it, we need to balance two things:
1. A few ROCm subprojects want to use OpenMPI as part of their tests.
2. When we distribute ROCm, we don't want to force our version of OpenMPI onto users, as they may have different versions in use elsewhere on the system.

When we distribute ROCm tests, is it okay if OpenMPI is distributed too? In such cases would we want the tests to use _either_ our provided OpenMPI or a different version installed on the system if available?

Depending on the answers there, I could see a case for converting OpenMPI into a sysdep that has isolated symbols and is used directly by the ROCm subprojects instead of system versions. I could also see us continuing along the lines of what this PR is currently doing, building non-sysdep openmpi artifacts but then excluding them from tarballs (and native packages, and python packages).

Yet another option would be to install openmpi into a test dockerfile at the system level instead of building it ourselves. That would be simplest, if we don't need the vendor it into an artifact and build from source. See https://github.com/ROCm/TheRock/tree/main/dockerfiles#no_rocm_image_dockerfile for docs and these example dockerfiles: https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04_media.Dockerfile and https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04_rocgdb.Dockerfile where some extra system dependencies are installed for running subproject tests. It sounds like before the subproject _tests_ run, the subprojects would still need to be _built_ with some options enabling MPI though? What would the implications there be? If MPI is enabled in the build and you run the tests (or full applications) on a system without MPI would they skip themselves automatically or would there be runtime errors?

ScottTodd (commented):
(marking as 'reviewed' to drop this from my queue pending more discussion on the conversation thread)

build_tools/github_actions/test_executable_scripts/test_rocprofiler_sdk.py
minseobshin11: Hi, I've changed the approach to use a system OpenMPI in a dedicated test image, currently scoped to rocprofiler-sdk. I added a new Docker image that apt installs libopenmpi-dev and openmpi-bin on top of the base no_rocm test image, and pointed only the rocprofiler-sdk test job at it. This keeps OpenMPI in the CI image only (it isn't built into any artifact, tarball, or package so we don't force our OpenMPI onto users). Since rocprofv3 doesn't link MPI and the tests are configured at test time in the container, find_package(MPI) resolves the image's OpenMPI when present; when it's absent the mpi-ranks tests are DISABLED (graceful skip, no runtime errors). This should settle the rocprofiler-sdk case.

minseobshin11:
PR to add a docker image: https://github.com/ROCm/TheRock/pull/6610

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -487,6 +487,10 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
         "total_shards_dict": {
             "linux": 1,
         },
+        # rocprofv3 mpi-ranks tests gate on find_package(MPI) and launch under
+        # mpiexec. OpenMPI is not bundled in TheRock artifacts and is provided via
+        # the specialized openmpi image.
+        "container_image": "ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi@sha256:f67d0b02cae8faf0d2f3e4a1de38a01af6bad2eb27f10a5e07bf19748a84d1e6",
```
ScottTodd, line 493: Checking the CI results:

* Baseline: https://github.com/ROCm/TheRock/actions/runs/29962234421/job/89078588427
    ```
    -- Could NOT find MPI_C (missing: MPI_C_LIB_NAMES MPI_C_HEADER_DIR MPI_C_WORKS) 
    -- Could NOT find MPI_CXX (missing: MPI_CXX_LIB_NAMES MPI_CXX_HEADER_DIR MPI_CXX_WORKS) 
    -- Could NOT find MPI (missing: MPI_C_FOUND MPI_CXX_FOUND) 
    ```
    
    ```
    100% tests passed out of 944
    
    	934 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi.test_mpi_ranks_feature (Disabled)
    	935 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi.test_csv_output_consistency (Disabled)
    	936 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi.test_no_output_for_filtered_ranks (Disabled)
    	937 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi-multiple.test_mpi_ranks_feature (Disabled)
    	938 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi-multiple.test_csv_output_consistency (Disabled)
    	939 - tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi-multiple.test_no_output_for_filtered_ranks (Disabled)
    	943 - tests.integration.execute.rocprofv3-test-mpi-ranks-with-mpi (Disabled)
    	944 - tests.integration.execute.rocprofv3-test-mpi-ranks-with-mpi-multiple (Disabled)
    ```

* This PR: https://github.com/ROCm/TheRock/actions/runs/29961497706/job/89076779199?pr=5627
    ```
    -- Found MPI_C: /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so (found version "3.1")
    -- Found MPI_CXX: /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi_cxx.so (found version "3.1")
    -- Found MPI: TRUE (found version "3.1")
    ```

    ```
     942/1015 Test  #935: tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi.test_csv_output_consistency ...................................................................   Passed    0.25 sec
              Start  939: tests.integration.validate.rocprofv3-test-mpi-ranks-with-mpi-multiple.test_no_output_for_filtered_ranks
    
    100% tests passed out of 952
    ```

ScottTodd (approved):
Looks good, thanks.

PR #5628, opened 2026-06-04
fix: Add missing rocm compiler configs

estewart08:
This adds the main rocm.cfg and alias configs:
  clang-${LLVM_VERSION_MAJOR}.cfg, clang++.cfg, clang.cfg,
  clang-cl.cfg, clang-cpp.cfg, flang.cfg
    
  These seem to have been lost when switching to TheRock.
    
  rocm.cfg contents:
  Enable RUNPATH instead of RPATH:
  -Wl,--enable-new-dtags
    
  Ensures proper path base rocm install directory to
  determine hip version and linking of bitcode libraries:
  --rocm-path='<CFGDIR>/../../..'
    
  Pass the architecture specific clang resource directory.
  This is not passed by default:
  -frtlib-add-rpath

  JIRA ID: LCOMPILER-2285

### DIFF
--- a/base/aux-overlay/CMakeLists.txt
+++ b/base/aux-overlay/CMakeLists.txt
@@ -51,6 +51,28 @@ if(NOT WIN32)
     "${CMAKE_CURRENT_BINARY_DIR}/bin/offload-arch"
     DESTINATION bin
   )
+
+  # Begin creation of compiler configs. We will need LLVM_VERSION_MAJOR.
+  include("${THEROCK_SOURCE_DIR}/compiler/amd-llvm/cmake/Modules/LLVMVersion.cmake")
+  set(rocm_config_name "rocm.cfg")
+  set(rocm_config_dir "${CMAKE_CURRENT_BINARY_DIR}/lib/llvm/bin")
+  set(rocm_config_content "-Wl,--enable-new-dtags\n--rocm-path='<CFGDIR>/../../..'\n-frtlib-add-rpath")
+
+  # Generate rocm.cfg
+  file(MAKE_DIRECTORY "${rocm_config_dir}")
+  message(STATUS "Generating ${rocm_config_dir}/${rocm_config_name}")
+  file(WRITE "${rocm_config_dir}/${rocm_config_name}" ${rocm_config_content})
+  install(FILES "${rocm_config_dir}/${rocm_config_name}"
+    DESTINATION "lib/llvm/bin")
+
+  # Generate additional config aliases
+  set(config_names "clang-${LLVM_VERSION_MAJOR}.cfg;clang++.cfg;clang.cfg;clang-cl.cfg;clang-cpp.cfg;flang.cfg")
+  foreach(name ${config_names})
+    message(STATUS "Generating ${rocm_config_dir}/${name}")
+    file(WRITE "${rocm_config_dir}/${name}" "@rocm.cfg")
+    install(FILES "${rocm_config_dir}/${name}"
+      DESTINATION "lib/llvm/bin")
+  endforeach()
 endif()
 
 install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/overlay/" DESTINATION ".")


base/aux-overlay/CMakeLists.txt
```diff
@@ -51,6 +51,28 @@ if(NOT WIN32)
...
+  set(rocm_config_name "rocm.cfg")
+  set(rocm_config_dir "${CMAKE_CURRENT_BINARY_DIR}/lib/llvm/bin")
+  set(rocm_config_content "--rocm-path='<CFGDIR>/../../..'\n-frtlib-add-rpath")
+
+  # Generate rocm.cfg
+  file(MAKE_DIRECTORY "${rocm_config_dir}")
+  message(STATUS "Generating ${rocm_config_dir}/${rocm_config_name}")
+  file(WRITE "${rocm_config_dir}/${rocm_config_name}" ${rocm_config_content})
+  install(FILES "${rocm_config_dir}/${rocm_config_name}"
+    DESTINATION "lib/llvm/bin")
```
ScottTodd, line 66: > This adds the main rocm.cfg and alias configs:
> clang-${LLVM_VERSION_MAJOR}.cfg, clang++.cfg, clang.cfg, clang-cl.cfg, clang-cpp.cfg, flang.cfg

Can you explain what these files are and how they are used? What does including them enable? Was what producing them before TheRock?

(I'd like to just trust you here, but I'm not seeing enough of an evidence trail explaining what these files are. Someone else could similarly propose to add config files with some different format and I wouldn't know which to accept :D)

base/aux-overlay/CMakeLists.txt
```diff
@@ -51,6 +51,28 @@ if(NOT WIN32)
     "${CMAKE_CURRENT_BINARY_DIR}/bin/offload-arch"
     DESTINATION bin
   )
+
+  # Begin creation of compiler configs. We will need LLVM_VERSION_MAJOR.
+  include("${THEROCK_SOURCE_DIR}/compiler/amd-llvm/cmake/Modules/LLVMVersion.cmake")
+  set(rocm_config_name "rocm.cfg")
+  set(rocm_config_dir "${CMAKE_CURRENT_BINARY_DIR}/lib/llvm/bin")
+  set(rocm_config_content "--rocm-path='<CFGDIR>/../../..'\n-frtlib-add-rpath")
```
ScottTodd: What's `-frtlib-add-rpath`? I see a few loose references in https://github.com/search?q=repo%3AROCm%2Fllvm-project%20frtlib-add-rpath&type=code

I'm specifically wondering if this should be included or omitted on Windows as RPATH (https://en.wikipedia.org/wiki/Rpath) is typically a Unix/Linux/ELF executable concept and not something used on Windows with DLLs.

base/aux-overlay/CMakeLists.txt
```diff
@@ -51,6 +51,28 @@ if(NOT WIN32)
...
+  install(FILES "${rocm_config_dir}/${rocm_config_name}"
+    DESTINATION "lib/llvm/bin")
+
+  # Generate additional config aliases
+  set(config_names "clang-${LLVM_VERSION_MAJOR}.cfg;clang++.cfg;clang.cfg;clang-cl.cfg;clang-cpp.cfg;flang.cfg")
+  foreach(name ${config_names})
+    message(STATUS "Generating ${rocm_config_dir}/${name}")
+    file(WRITE "${rocm_config_dir}/${name}" "@rocm.cfg")
+    install(FILES "${rocm_config_dir}/${name}"
+      DESTINATION "lib/llvm/bin")
```
ScottTodd, line 74: Should these files be produced by the LLVM subproject instead of the TheRock superproject if the files are all under `*/llvm/*` paths? (I think this one could be fine how you have it, but we generally try to limit how much TheRock applies to the build beyond repackaging what subprojects already provide0

estewart08:
I need to close this and reopen another PR with an updated branch name that satisfies the new requirements. We can continue the conversation on the new PR and I will provide more info in the description.

ScottTodd:
The PR review bot is advisory only (it is non-blocking and easy to "override" anyways) and this PR was opened before it was added, fine to keep this PR grandfathered in.

base/aux-overlay/CMakeLists.txt
estewart08: **Can you explain what these files are and how they are used?**
When a user invokes the compiler via clang/flang/clang++/etc, it will automatically pick up this configuration file. The user does not need to define LD_LIBRARY_PATH or a ld.so.config.

**What does including them enable?**
 - Enable **RUNPATH** instead of RPATH:
   `-Wl,--enable-new-dtags`
    
  - Ensures proper path for base rocm install directory to
    determine hip version and linking of bitcode libraries:
    `--rocm-path='<CFGDIR>/../../..'`
    
  - Pass the architecture specific clang resource directory used by compiler runtimes and sanitizer builds.
    This is not passed by default:
    `-frtlib-add-rpath`


**Was what producing them before TheRock?**
This was originally done in build_lightning.sh legacy script for ROCm release before TheRock was around (Pre 7.10).

base/aux-overlay/CMakeLists.txt
estewart08: From my understanding, during the build of the math libraries, in the past, the `-frtlib-add-rpath` option was causing issues with new RHEL-10 RPATH requirements. I decided to put this in `base` so that the config would be present in the packages at install but not used during the build.

base/aux-overlay/CMakeLists.txt
estewart08: **What's -frtlib-add-rpath?**
It adds the architecture specific compiler resource directory. For example, one would be `<path-to-base-rocm>/lib/llvm/lib/clang/23/lib/x86_64-unknown-linux-gnu`. This is also used for the sanitizer libraries.

**I'm specifically wondering if this should be included or omitted on Windows as RPATH**
This code is guarded by `if(NOT WIN32)`.

ScottTodd (approved):
Please re-request review after responding to feedback when you think a PR is ready for another look

Image

(I had ~15 PRs in my queue today and nearly missed this PR since it didn't appear there)

ScottTodd:
Looks like this may have caused https://github.com/ROCm/TheRock/issues/6113

> `rocm-sdk-core-7.14.0a20260624` introduced a new `clang++.cfg` → `rocm.cfg` configuration file that unconditionally injects `-Wl,--enable-new-dtags` into every `clang++` invocation. This breaks PyTorch inductor's precompiled header (PCH) compilation with:
> 
> ```
> clang++: error: cannot specify -o when generating multiple output files
> ```

PR #5637, opened 2026-06-04
Extract therock-tools Python package

stellaraccident:
I'm trying to make our artifact management and general tooling available outside of the project and intend to deploy this wheel to pypi. There will be followons to update our artifacts a bit so that out of tree users can leverage TheRock builds in an ergonomic way.

Changes:
- Add python/therock-tools as an installable package for reusable artifact, fileset, topology, storage, and GitHub helper tooling.
- Add fq therock-* console script entrypoints for packaged CLIs.
- Keep build_tools trampolines and _therock_utils aliases for existing in-tree imports.
- Track therock-tools sources in CMake fileset dependencies and add package-surface tests.


cmake/therock_artifacts.cmake
```diff
@@ -173,6 +173,11 @@ function(therock_provide_artifact slice_name)
 
   # Populate commands.
   set(_fileset_tool "${THEROCK_SOURCE_DIR}/build_tools/fileset_tool.py")
```
stellaraccident: For these deep references, let's just switch them to the canonical new location of the tool and not add the dependency globs. We own all of this code and I'd rather avoid the trampolines if possible in our regular code paths. Here and in other .cmake files.

cmake/therock_subproject.cmake
```diff
@@ -882,6 +882,12 @@ function(therock_cmake_subproject_activate target_name)
 
   set(_fileset_tool "${THEROCK_SOURCE_DIR}/build_tools/fileset_tool.py")
   list(APPEND _fprint_files "${_fileset_tool}")
+  set(_therock_tools_src_dir "${THEROCK_SOURCE_DIR}/python/therock-tools/src/therock_tools")
```
stellaraccident: Ditto - use the canonical location and don't glob.

python/therock-tools/src/therock_tools/data/therock_amdgpu_targets.cmake
```diff
@@ -0,0 +1,418 @@
+# Copyright Advanced Micro Devices, Inc.
```
stellaraccident: We should not move this. Put it back where it was and explain to me why this got pulled into the move closure.

build_tools/fetch_artifacts.py
```diff
@@ -2,396 +2,24 @@
...
-        # Included and not excluded.
-        return True
-
-    return {a for a in artifacts if _should_include(a)}
 
+_THEROCK_TOOLS_SRC = (
+    Path(__file__).resolve().parent.parent / "python" / "therock-tools" / "src"
+)
+if _THEROCK_TOOLS_SRC.exists():
+    sys.path.insert(0, os.fspath(_THEROCK_TOOLS_SRC))
```
stellaraccident: Unconditionally add.

build_tools/fetch_artifacts.py
```diff
@@ -2,396 +2,24 @@
...
-
-    args = parser.parse_args(argv)
-
-    run(args)
-
+_module = importlib.import_module("therock_tools.fetch_artifacts")
 
 if __name__ == "__main__":
-    main(sys.argv[1:])
+    _module.main(sys.argv[1:])
```
stellaraccident: We are only supporting these for direct invocation as CLI tools. Do not hack the imports with sys.modules. If importing, that must be from the canonical locations. Here and all trampolines.

stellaraccident:
Sorry about the high line diff counts. I was unable to get git to recognize a couple of the splices as moves.

stellaraccident:
Note: project has been pushed to pypi in order to reserve the name.

python/therock-tools/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+  "pyzstd>=0.16.0",
+  "tomli>=2.2.1; python_version < '3.11'",
+]
+
+[project.scripts]
+therock-artifact-manager = "therock_tools.artifact_manager:main"
+therock-fetch-artifacts = "therock_tools.fetch_artifacts:main"
+therock-fileset-tool = "therock_tools.fileset_tool:console_main"
+therock-find-artifacts-for-commit = "therock_tools.find_artifacts_for_commit:main"
+therock-find-latest-artifacts = "therock_tools.find_latest_artifacts:main"
```
ScottTodd, line 25: A new similar `build_tools/github_actions/baseline_runs.py` script was recently added in https://github.com/ROCm/TheRock/pull/5435 (cc @jayhawk-commits @erman-gurses @ammallya @geomin12 )

With this as the goal of this new package:

> There will be followons to update our artifacts a bit so that out of tree users can leverage TheRock builds in an ergonomic way.

we should try to consolidate these scripts and clean them up more for external use.

BTW, some more context:
* since writing the "find artifacts for commit" and "find latest artifacts" scripts I've found that Claude Code and Codex are actually proficient enough at just using the `gh` CLI directly, so the scripts are more catering to human developers and CI systems that want deterministic behavior
* those scripts do have a few missing features to properly handle multi-arch workflows (some of that is addressed in the new `baseline_runs.py`)

python/therock-tools/pyproject.toml
ScottTodd, line 1: (can be a follow-up, or as part of this PR while your coding agent has context in memory already) Let's also add a README.md to this folder including
* a link to https://pypi.org/project/therock-tools/
* explaining the split between what belongs in https://github.com/ROCm/TheRock/tree/main/build_tools and what belongs here
* explaining what additional considerations these files have since they are part of published packages, if any (e.g. docstrings, compatibility / API expectations as there will be external users instead of just in-tree users)

python/therock-tools/src/therock_tools/github_api.py
```diff
@@ -0,0 +1,324 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Small GitHub REST helpers used by therock-tools artifact probes.
+
+The TheRock-specific GitHub Actions workflow helpers live under
+``build_tools/github_actions``. This module keeps the installed package limited
+to generic GitHub API operations needed by the artifact lookup tools.
```
ScottTodd, line 8: I don't see changes to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py in this PR. I guess this is just forking some functions and keeping the old ones around, not redirecting any remaining scripts in `build_tools/` to use the new `python/therock-tools/`? Do you want to draw a hard line there or could we have `build_tools/github_actions/github_actions_api.py` extend `python/therock-tools/src/therock_tools/github_api.py`?

python/therock-tools/src/therock_tools/artifacts.py
ScottTodd, line 1: This file (and `archive_util`) is still referenced in a file under `/tests/`: https://github.com/ROCm/TheRock/blob/ee2040338f111a54d5bffa9335ee28114ec2b2f2/tests/test_artifact_structure.py#L38-L42

That caused this test failure: https://github.com/ROCm/TheRock/actions/runs/26989163878/job/79655077100?pr=5637#step:6:27
```
______________ ERROR collecting tests/test_artifact_structure.py _______________
ImportError while importing test module '/home/runner/_work/TheRock/TheRock/tests/test_artifact_structure.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../_tool/Python/3.12.13/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/test_artifact_structure.py:41: in <module>
    from _therock_utils.artifacts import ArtifactName
E   ModuleNotFoundError: No module named '_therock_utils.artifacts'
```

Huh... that _also_ failed "test sanity" here with a test collection error: https://github.com/ROCm/TheRock/actions/runs/26989163878/job/79655144022?pr=5637#step:13:1. Weird since that only ran `python build_tools/github_actions/test_executable_scripts/test_sanity.py` which I'd expect to run that one test directly and not trigger test collection from other files in that directory...

build_tools/fileset_tool.py
```diff
@@ -2,311 +2,19 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""fileset_tool.py
+"""Compatibility CLI trampoline for :mod:`therock_tools.fileset_tool`."""
```
ScottTodd, line 5: > Keep build_tools trampolines and _therock_utils aliases for existing in-tree imports.

Do you want to keep this trampolines in the long term, or should we migrate existing uses (particularly user- and CI-facing code)? I think adding/maintaining/deleting trampolines as scripts are added/changed/deleted will be a chore but this could be useful enough to keep for some transition period.

stellaraccident:
A couple more CI rounds to make sure everything lines up, but if @ScottTodd can provide a quick sanity check, would be appreciated.

python/therock-tools/tests/artifact_backend_test.py
```diff
@@ -15,13 +15,13 @@
...
 
-from _therock_utils.artifact_backend import (
+from therock_tools.artifact_backend import (
     ArtifactBackend,
     LocalDirectoryBackend,
     S3Backend,
     create_backend_from_env,
 )
-from _therock_utils.workflow_outputs import WorkflowOutputRoot
+from therock_tools.workflow_outputs import WorkflowOutputRoot
```
ScottTodd, line 24: Do we still need these `sys.path.insert()` calls in test files now that there is a proper package for `therock_tools`?

ScottTodd (commented):
Cool, I like the idea of organizing more of this code into a real/published package, as that will help enforce some more organization and allow out-of-tree usage without forking or needing to clone the repository.

build_tools/fileset_tool.py
stellaraccident: We can redirect as a follow-up. I was trying to keep this as close to minimal refactor for the first phase as I could.

Issue #4866, opened 2026-04-27
Package rocm-kpack


marbre:
Splitting PyTorch wheels requires [`rocm-kpack`](https://github.com/ROCm/rocm-systems/tree/develop/shared/kpack) which is currently not packaged and therefore needs to be installed from source. Make it available as a package.

BrianHarrisonAMD:
hipDNN is going to start consuming rocm-kpack for rocKE & hip-kernel-provider.

We are taking it on as a direct source dependency to start with, but once this migrates into an installable package we will need to also update those projects. 

Used here for now:
- https://github.com/ROCm/rocm-libraries/pull/9207/changes#diff-549f04c5389c3db70856259c4f612eb5f75b58f97eb45925556f688f82df1b8dR50-R84
- https://github.com/ROCm/TheRock/pull/6427/changes#diff-648e0abe4344fb9579c41807ce6f5e3e6d1f32d71d2ccd767e07c5b072004da9R320

PR #5638, opened 2026-06-05
[rocSHMEM] Package rocshmem4py Python binding in TheRock

drprajap:
## Motivation

Package and publish `rocshmem4py` as a Python wheel through TheRock's existing packaging pipeline.

## Technical Details

rocm-systems [#9471](https://github.com/ROCm/rocm-systems/pull/9471) moved the python binding out of `projects/rocshmem` into a standalone PEP-517 project at `python/rocshmem`, which builds against an *installed* rocSHMEM discovered via `find_package(rocshmem CONFIG)`.

Companion rocm-systems PR: https://github.com/ROCm/rocm-systems/pull/9471 - Merged

This change builds the wheel with `python -m build --wheel` into `PACKAGES_DIR/dist`, so the  existing upload and index steps publish it unchanged.

The wheels are stamped with TheRock's build identity and pinned to an exact   `rocm-sdk-core==<ROCm version>`, e.g. `0.1.0+devrocm10.1.0.dev0.<sha>` for CI and `0.1.0+rocm10.1.0` for a release. rocSHMEM's own version stays available
  as `rocshmem4py.__rocshmem_version__`.

JIRA ID : AIROCSHMEM-224
ISSUE ID : https://github.com/ROCm/rocm-systems/issues/2736

## Test Plan

- `test_rocm_wheels.yml` installs the published wheel into a clean venv, checks its version and `rocm-sdk-core` pin, and runs a GPU smoke test.

Verified workflow commands locally in a fresh environment.

## Test Result

- cp310–cp313 wheels built and finalized with the expected version stamp
- Clean-venv install + gfx942 GPU smoke test pass
- Unit tests cover wheel finalization, version identity, and wheel collection.
- 
## Submission Checklist


comm-libs/CMakeLists.txt
```diff
@@ -160,6 +160,7 @@ if(THEROCK_ENABLE_ROCSHMEM)
       "-DEXPLICIT_ROCM_VERSION=${ROCM_MAJOR_VERSION}.${ROCM_MINOR_VERSION}.${ROCM_PATCH_VERSION}"
       -DBUILD_FUNCTIONAL_TESTS=${THEROCK_BUILD_TESTING}
       -DBUILD_UNIT_TESTS=${THEROCK_BUILD_TESTING}
+      -DBUILD_PYTHON_BINDING=ON
```
ScottTodd: It sounds like from discussion with @stellaraccident we'd prefer for python projects to be built via `python -m build ...` independently and not via CMake integration? Or do we still _also_ need a CMake component?

drprajap: Yes, I am doing rework of this, will mark this as draft. we won't need CMake component after rework (python -m build ...)

drprajap: @ScottTodd revised the changes, removed CMake component as it is not required anymore. Please take a look and let me know if this is how we can package and publish or any other preferred way.

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -157,6 +157,36 @@ jobs:
...
+      # rocm-sdk-* wheels it can't be repackaged from artifacts; it is built
+      # against an installed rocSHMEM, into dist/ for the upload step below.
+      - name: Build rocshmem4py wheel
+        run: |
+          if [ ! -d "${{ env.ARTIFACTS_DIR }}/rocshmem_dev_generic" ]; then
+            echo "No rocshmem artifacts fetched; skipping rocshmem4py wheel"
+            exit 0
+          fi
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
+          git submodule update --init --depth=1 rocm-systems
```
ScottTodd: @marbre expressed some concerns over fetching sources here. I'll let him respond more directly or copy/paste his comments.

Brainstorming / seeing what holes we can poke in this...

These workflows support running with `artifact_run_id` set to a prior workflow run so you can have:
* Run id `30000000000`, commit `abc100`
  * Postsubmit CI (`push` event) builds artifacts, packages, etc. and runs tests
* Run id `30000000001`, commit `def200`
  * Manual trigger of this workflow with `artifact_run_id=30000000000`
    * Build ROCm python packages from the artifacts at commit `abc100`, using the python packaging code ([`build_tools/packaging/python`](https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python)) from commit `def200`
    * (this PR) build the rocshmem4py package using the rocm-systems code at commit `def200`

That doesn't seem so bad, but an alternate approach would be:
* Run id `30000000002`, commit `abc100`
  * Postsubmit CI (`push` event) builds artifacts, packages, etc. and runs tests
  * *Option A: rocshmem python sources are added to an artifact*
  * *Option B: rocshmem python wheel(s) are built and uploaded, perhaps as part of an artifact*
* Run id `30000000003`, commit `def200`
  * Manual trigger of this workflow with `artifact_run_id=30000000002`
    * Build ROCm python packages from the artifacts at commit `abc100`, using the python packaging code ([`build_tools/packaging/python`](https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python)) from commit `def200`
    * *Option A: build rocshmem python wheel(s) from the rocshmem python sources in an artifact*
    * *Option B: extract rocshmem python wheel(s) and upload them together with rocm python packages*

I think how it's currently set up isn't so bad, so long as we're aware of the difference between what gets built where. Where it starts to matter more is when we want to build _multiple_ python projects in addition to the ROCm python packages:
* rocshmem4py
* kpack: https://github.com/ROCm/TheRock/issues/4866
* hipdnn
* amdsmi?

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -157,6 +157,36 @@ jobs:
...
+          # and the compiler are target-neutral, and flattening several
+          # per-family trees into one prefix would collide.
+          python ./build_tools/fileset_tool.py artifact-flatten \
+            "${{ env.ARTIFACTS_DIR }}"/*_generic \
+            -o "$ROCM_PREFIX"
+
+          # Without the device bitcode the build still succeeds, but silently
+          # yields a single-architecture wheel that aborts with
+          # hipErrorInvalidKernelFile (218) on every other GPU.
+          ls "$ROCM_PREFIX"/lib/librocshmem_device_*.bc > /dev/null
```
ScottTodd: Not sure this step is adding much in the workflow. I'd rather have
* The build process itself check for all expected files existing prior to building
* Unit tests for the built packages checking for expected files
* Integration tests using the built packages checking that GPU code runs correctly

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -157,6 +157,36 @@ jobs:
...
+
+          cd rocm-systems/python/rocshmem
+          ROCM_PATH="$ROCM_PREFIX" \
+          CMAKE_PREFIX_PATH="$ROCM_PREFIX" \
+            python -m build --wheel --outdir "${{ env.PACKAGES_DIR }}/dist"
```
ScottTodd: Would you like to add some extra version metadata to the wheel here? Here are our docs on python package versioning: https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#python-package-versions

CI on this PR produced:
* https://github.com/ROCm/TheRock/actions/runs/31060508515/job/92505104234?pr=5638#step:7:438
    ```
    Successfully built rocshmem4py-0.1.0+rocshmem3.6.0-cp312-cp312-linux_x86_64.whl
    ```
* https://therock-ci-artifacts.s3.amazonaws.com/31060508515-linux/python/index.html
    ```
    rocm_sdk_device_gfx942-10.1.0.dev0+2316db9c3173299548d23a4100cf0696729e32f7-py3-none-linux_x86_64.whl
    rocm_sdk_libraries-10.1.0.dev0+2316db9c3173299548d23a4100cf0696729e32f7-py3-none-linux_x86_64.whl
    rocshmem4py-0.1.0+rocshmem3.6.0-cp312-cp312-linux_x86_64.whl
    torch-2.11.0+devrocm10.1.0.dev0.2316db9c3173299548d23a4100cf0696729e32f7-cp312-cp312-linux_x86_64.whl
    torch-2.12.0+devrocm10.1.0.dev0.2316db9c3173299548d23a4100cf0696729e32f7-cp312-cp312-linux_x86_64.whl
    torch-2.13.0+devrocm10.1.0.dev0.2316db9c3173299548d23a4100cf0696729e32f7-cp312-cp312-linux_x86_64.whl
    ```

If the `0.1.0+rocshmem3.6.0` version is used as-is, individual PRs (and release workflow runs) will all produce rocshmem4py packages with the same version. That will make issue reporting more difficult if you can't tell what code went into the package from just the name/version. Is some commit metadata at least stamped into the packages somewhere?

Some ideas to consider for dev builds:
1. Include the rocm-systems commit hash in the `rocshmem3.6.0` version
2. Include the TheRock commit hash as we do for the ROCm packages _in addition to_ the rocshmem version
3. Include the TheRock commit hash as we do for the ROCm packages _instead of_ the rocshmem version

Then similarly for nightly releases and prereleases:
1. Include the ROCm version _in addition to_ the rocshmem version
2. Some other option?

.github/workflows/build_portable_linux_python_packages.yml
drprajap: I think I can go with your suggested alternate approach (B) where it builds rocshmem4py during artifact production against the same pinned rocm-systems checkout and rocSHMEM build store the completed wheel in a separate artifact, and have the Python packaging workflow only collect/publish it

.github/workflows/build_portable_linux_python_packages.yml
drprajap: yes make sense, I will remove this bitcode check from here and make missing architecture bitcode fail in the standalone project, add a built-wheel target inspection test, and add clean-install/import plus minimal GPU integration coverage.

.github/workflows/build_portable_linux_python_packages.yml
drprajap: Thanks for pointing it out, agreed current rocSHMEM-only version is not sufficient.
I think Option 2 for dev builds sounds reasonable - 0.1.0+devrocm10.1.0.dev0-<therock-sha>.rocshmem3.6.0 this can use TheRock’s release/version metadata, retain the linked rocSHMEM version, and ensure dev builds include TheRock commit hash
I have to think more on nightly release/prerelease

.github/workflows/build_portable_linux_python_packages.yml
marbre: My comment in chat:
> The current layering seems not correct as it would indeed build in the step that consumes already built artifacts and then packages those. I think we should move building the wheel to the build phase but as a separate component which allows you to make full use of the build system which then also makes sure you get all artifacts defined as dependences (in your case `rocshmem`  static `.a`, headers and `.b`c ?). There is also no strict need to have this as part of the rocshmem artifact and it can be a separate artifact. Moving to the build phase eliminates your need to pull rocm-systems yourself. Pulling via `git pull` (and not `fetch_sources.py`) is risky as you might not get the same revision used for build (as `fetch_sources.py` can apply patches even though we try to avoid using this mechanism).

So yes, I think option B is what we should go for. Happy to help if further guidance is needed.

.github/workflows/build_portable_linux_python_packages.yml
drprajap: I have addressed this and opted for option 3 - the published wheel uses TheRock’s build/release identity instead of the upstream `+rocshmem3.6.0` local version as that added some complexity when building for nightly, releases ..

For instance  CPython 3.12 on Linux, the filenames would be:

  - CI/PR: `rocshmem4py-0.1.0+devrocm10.1.0.dev0.<40-character-TheRock-SHA>-cp312-cp312-linux_x86_64.whl`
  - Nightly: `rocshmem4py-0.1.0+rocm10.1.0a20260826-cp312-cp312-linux_x86_64.whl`
  - Release candidate: `rocshmem4py-0.1.0+rocm10.1.0rc1-cp312-cp312-linux_x86_64.whl`
  - Final release: `rocshmem4py-0.1.0+rocm10.1.0-cp312-cp312-linux_x86_64.whl`

The linked rocSHMEM version remains available at runtime through `rocshmem4py.__rocshmem_version__`.

.github/workflows/build_portable_linux_python_packages.yml
drprajap: I have revised this bitcode based architecture plumbing - https://github.com/ROCm/rocm-systems/pull/10380 and moved wheel architecture detection in standalone pythong project to use ROCSHMEM_OFFLOAD_TARGETS  from rocshmem package

GPU Integration tests are also added here that checks e2e rocshmem initialization, allocation, synchronization based on built packages

.github/workflows/build_portable_linux_python_packages.yml
drprajap: @marbre  Implemented as per suggestion here, not pulling rocm-systems any more and now it builds rocshmem4py wheel as a separate component and artifact using existing build system. Please review it and see if it aligns with what we discussed.

marbre: CI failed (due to infra) and I re-triggered as I would like to inspect the final package. One thing, which I otherwise need to find out via inspecting the package, is rocshmem4py Python version agnostic? If it is not, we may need a different mechanism here as we only build the wheel once but we would need to build for several Python versions. ROCm is Python agnostic thus this is nothing we provide straight away.

build_tools/tests/finalize_rocshmem4py_wheel_test.py
marbre, line 1: This needs more work to pass in CI:

```
==================================== ERRORS ====================================
__________ ERROR collecting tests/finalize_rocshmem4py_wheel_test.py ___________
ImportError while importing test module '/home/runner/work/TheRock/TheRock/build_tools/tests/finalize_rocshmem4py_wheel_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/_pytest/python.py:507: in importtestmodule
    mod = import_path(
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/_pytest/pathlib.py:587: in import_path
    importlib.import_module(module_name)
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/_pytest/assertion/rewrite.py:197: in exec_module
    exec(co, module.__dict__)
tests/finalize_rocshmem4py_wheel_test.py:15: in <module>
    from finalize_rocshmem4py_wheel import _set_runtime_requirement
packaging/python/finalize_rocshmem4py_wheel.py:31: in <module>
    from change_wheel_version import change_wheel_version
third_party/change_wheel_version/change_wheel_version.py:25: in <module>
    import installer.utils
E   ModuleNotFoundError: No module named 'installer'
```

build_tools/packaging/python/finalize_rocshmem4py_wheel.py
marbre, line 1: Is this really needed and can't it be rather introduced when building the wheel?

drprajap: part of it is no longer needed - _gpu_targets portion is added in python/rocshmem wheel build, so I can remove that

Other changes are not as straightforward to pull during build time because current rocshmem4py wheel build does not accept required TheRock provided inputs with build identity metadata -ROCSHMEM4PY_PACKAGE_VERSION= 0.1.0+devrocm10.1.0.dev0.<TheRock-SHA> -and ROCSHMEM4PY_INSTALL_RPATH , also rocshmem4py cmake doesn't apply RPATH during linking.

What I can do is keep current finalizer and as soon as we have long term fix to override build-provided package-version, metadata and RPATH in rocshmem [setup.py ](https://github.com/ROCm/rocm-systems/blob/develop/python/rocshmem/setup.py) landed in rocm-systems, I can have follow-up PR to remove this finalizer.

build_tools/tests/finalize_rocshmem4py_wheel_test.py
drprajap: yes, going to fix it, repro'd the issue locally

.github/workflows/build_portable_linux_python_packages.yml
drprajap: @marbre yes due to rocshmem4py CPython extension, it is not build with stable abi3 (after we moved to nanobind, I think minimum version supported with stable ABI  doesn't cover python 3.10/3.11, I would need some fact checks - but my initial checks gave minimum version supported Python3.12 with stable abi3 so we would need multiple python version- I have fixed the build which had issue with interpreter list parsing , 

this is the passing build - https://therock-ci-artifacts.s3.amazonaws.com/33651339983-linux/logs/comm-libs/rocshmem4py_build.log
you can inspect the package
`2450.3	  Repacking wheel and rewriting RECORD file... done
2450.3	Finalized rocshmem4py-0.1.0+devrocm10.1.0.dev0.91e8fe8bf49d9ed818bf644580a9fb8711e97d18-cp310-cp310-linux_x86_64.whl
2450.3	Finalized rocshmem4py-0.1.0+devrocm10.1.0.dev0.91e8fe8bf49d9ed818bf644580a9fb8711e97d18-cp311-cp311-linux_x86_64.whl
2450.3	Finalized rocshmem4py-0.1.0+devrocm10.1.0.dev0.91e8fe8bf49d9ed818bf644580a9fb8711e97d18-cp312-cp312-linux_x86_64.whl
2450.3	Finalized rocshmem4py-0.1.0+devrocm10.1.0.dev0.91e8fe8bf49d9ed818bf644580a9fb8711e97d18-cp313-cp313-linux_x86_64.whl`

ScottTodd:
I'll be offline next week and wasn't able to get to this review yet. Hopefully @marbre or someone else can take a look soon (looks like he's provided the most recent comments anyways)

drprajap:
@marbre can you take a look at recent changes, I have addressed review feedback and checked generated artifact/ whl as well. Let me know if it looks okay?

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +175,116 @@ jobs:
...
+
+          for attempt in {1..13}; do
+            if pip --python "${ROCSHMEM4PY_VENV_DIR}" install "${pip_args[@]}"; then
+              break
+            fi
+            if [[ "${attempt}" == 13 ]]; then
+              exit 1
+            fi
+            sleep 15
+          done
```
marbre: This violates our development guidelines and needs to move to a separate script, see https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +175,116 @@ jobs:
...
+          rocshmem4py.rocshmem_init_attr(0, 1, unique_id)
+          try:
+              buffer = rocshmem4py.rocshmem_create_buffer(4096)
+              assert buffer.ptr > 0
+              buffer.free()
+          finally:
+              rocshmem4py.hip_device_synchronize()
+              rocshmem4py.rocshmem_barrier_all()
+              rocshmem4py.rocshmem_finalize()
+          PY
```
marbre: As above, no code inlined into workflow steps please.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +175,116 @@ jobs:
...
+          # Keep this inline: repository/ref may select untrusted checkout code.
+          # Use the provisioned pip to install into a clean, pip-less venv.
+          python - <<'PY'
+          import os
+          import venv
+
+          venv.create(os.environ["ROCSHMEM4PY_VENV_DIR"], with_pip=False, clear=True)
+          PY
+
+          pip_args=(--disable-pip-version-check --no-cache-dir --only-binary=:all: --pre rocshmem4py)
```
marbre: The arguments request unconstrained rocshmem4py in a fresh environment. When an index contains multiple builds, pip selects the newest available binding and its pinned core dependency, regardless of `inputs.rocm_version`. One way would be to pin `rocm-sdk-core==${{ inputs.rocm_version }} in the installation request to than get the correct bindings but cleaner would be to explicitly pin the rocshmem4py version you want to test.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +175,116 @@ jobs:
...
+          if [[ -n "${PACKAGE_INDEX_URL}" ]]; then
+            pip_args+=(--index-url "${PACKAGE_INDEX_URL}")
```
marbre: If installing from the _aggregated_ index, e.g. https://nightly.repo.amd.com/rocm/whl-next/ for nightlies, and not from the core index https://nightly.repo.amd.com/rocm/core/whl-next/, the new package needs to be added to `build_tools/packaging/python/rocm_whl_next_ownership.yaml`. This can be a follow up and works best if the package was uploaded to `core/whl-next` initially but if you already want to test from the an `--index-url` we may need to split up packaging and would need to land this specific test in a follow up.

comm-libs/CMakeLists.txt
```diff
@@ -244,3 +244,51 @@ if(THEROCK_ENABLE_ROCSHMEM)
...
+  ##############################################################################
+  therock_detect_python_versions(
+    _rocshmem4py_python_executables
+    _rocshmem4py_python_versions
+  )
+  # therock_cmake_subproject_declare does not preserve CMake lists in
+  # CMAKE_ARGS, so escape the separators before forwarding the interpreters.
+  list(JOIN _rocshmem4py_python_executables "\;"
+    _rocshmem4py_python_executables_str
+  )
```
marbre, line 261: My agent spotted the following:

---
We should join with ";" here instead of "\;". `therock_cmake_subproject_declare()` uses `cmake_parse_arguments(PARSE_ARGV ...)`, which already preserves semicolons inside an argument. The additional escaping survives into the child configure command.

With two interpreters, the child receives:

`ROCSHMEM4PY_PYTHON_EXECUTABLES=/path/python3.10\;/path/python3.11`

Its `foreach(_python IN LISTS ROCSHMEM4PY_PYTHON_EXECUTABLES`) consequently yields one value containing both paths, so `if(NOT EXISTS "${_python}")` fails. This blocks the manylinux configuration, which supplies four interpreters; a single-interpreter local build would not expose it.

Reproduced this with a minimal CMake/Ninja project using the same argument parsing, target-property storage, and custom-command forwarding. Changing the join separator to ";" makes the child iterate over
the interpreters separately.

BUILD_TOPOLOGY.toml
```diff
@@ -794,6 +794,13 @@ type = "target-neutral"
 artifact_deps = ["core-runtime", "core-hip", "sysdeps"]
 disable_platforms = ["windows"]
 
+[artifacts.rocshmem4py]
+artifact_group = "comm-libs"
+type = "target-neutral"
+artifact_deps = ["rocshmem", "core-runtime", "core-hip", "sysdeps"]
+disable_platforms = ["windows"]
+python_requires = ["installer==0.7.0", "packaging==25.0", "wheel==0.46.2"]
```
marbre: While we have this mechanism we may rather want to add those dependencies to the root requirements.txt.  I think the dependencies specified here are installed in staged CI only, thus a fresh local environment following the README can build the extension but might fail during finalization.

marbre (commented):
Excuse the delay. We're getting closer but there are a few more things to fix.

.github/workflows/test_rocm_wheels.yml
drprajap: yes, makes sense, I will split this as a follow-up after rocshmem4py is published under `core/whl-next`

comm-libs/CMakeLists.txt
drprajap: that is strange, CI did build four CPython wheels from all interpreters with current escaped change, so not sure why agent flagged it, without this (and proposed agent fix) my local parsing yield one executable and rest of the path was cosindered as commands

BUILD_TOPOLOGY.toml
drprajap: will add to requirements.txt as well.

.github/workflows/test_rocm_wheels.yml
drprajap: corrected it now

docs/packaging/python_packaging.md
```diff
@@ -80,6 +80,19 @@ including a commit hash). You can also set an explicit fixed version with the
...
+When rocSHMEM is enabled, TheRock also publishes CPython-specific
+`rocshmem4py` wheels. The upstream public package version is retained, while
+TheRock replaces its local version with the ROCm build identity described in
+the [versioning documentation](versioning.md). For example, a development wheel
+can have version `0.1.0+devrocm10.1.0.dev0.<commit>`, while release wheels use
+versions such as `0.1.0+rocm10.1.0`.
+
+Each wheel declares an exact `rocm-sdk-core==<ROCm version>` runtime
+requirement. This ensures that its compiled extension loads the ROCm libraries
+from the same TheRock build instead of mixing incompatible builds.
```
ScottTodd, line 94: A section down below near the `## rocm-profiler` subsection may be useful too/instead.

These paragraphs are mostly talking about the versions, while it's worth explaining the package itself.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -134,6 +136,22 @@ jobs:
...
           ref: ${{ inputs.ref }}
 
+      # Use the workflow revision instead of the caller-selected repository/ref.
+      - name: Checkout rocshmem4py validation helper
+        if: ${{ runner.os == 'Linux' && inputs.package_find_links_url != '' }}
+        timeout-minutes: 15
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ github.sha }}
```
ScottTodd, line 146: The
```yml
        with:
          ref: ${{ inputs.ref }}
```
code is necessary so these workflows can function when called from other repositories, including:
* https://github.com/ROCm/rockrel (releases)
* https://github.com/ROCm/rocm-systems (component development)

I don't see why a sparse checkout of just these files is needed? The existing checkout step should include the necessary files already.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +192,39 @@ jobs:
...
+          PACKAGE_FIND_LINKS_URL: ${{ inputs.package_find_links_url }}
+          PACKAGE_INDEX_URL: ${{ inputs.package_index_url }}
+          ROCM_VERSION: ${{ inputs.rocm_version }}
+        run: |
+          python "${ROCSHMEM4PY_VALIDATION_SCRIPT}" \
+            install \
+            --venv-dir="${ROCSHMEM4PY_VENV_DIR}" \
+            --rocm-version="${ROCM_VERSION}" \
+            --index-url="${PACKAGE_INDEX_URL}" \
+            --find-links="${PACKAGE_FIND_LINKS_URL}"
```
ScottTodd, line 214: Can this install using `python build_tools/setup_venv.py` like the other package in stalls instead of adding a new validation script with an `install` subcommand?

I think you can follow the same pattern as the "Install rocm[profiler]" step, either into an isolated venv or reusing the same venv.

something like
```yml
run: | 
  python build_tools/setup_venv.py "${ROCSHMEM4PY_VENV_DIR}" \
    --packages "rocshmem4py==${ROCSHMEM4PY_VERSION} rocm-sdk-core==${ROCM_VERSION}"
    --index-url="${PACKAGE_INDEX_URL}" \
    --find-links="${PACKAGE_FIND_LINKS_URL}"
```

build_tools/github_actions/tests/validate_rocshmem4py_install_test.py
ScottTodd, line 1: This is a lot of code testing a test script. I don't think it should be needed - simplify the test/install code enough that there isn't anything left to test ;)

(prefer to reuse the existing setup_venv.py code like the other package installs)

.github/workflows/test_rocm_wheels.yml
```diff
@@ -174,6 +192,39 @@ jobs:
...
+
+      - name: Test installed rocshmem4py wheel
+        if: ${{ !cancelled() && steps.install_rocshmem4py.outcome == 'success' }}
+        timeout-minutes: 5
+        env:
+          ROCSHMEM_HEAP_SIZE: 268435456
```
ScottTodd, line 220: What is this `268435456` number? Note that we run these tests on a variety of GPU systems (datacenter and consumer cards, with different memory limits)

ScottTodd (commented):
This is looking better architecturally I think.

My main concerns at this point are:
* Overlapping responsibilities in `build_tools/github_actions/validate_rocshmem4py_install.py` leading to complicated workflow and script code. Install should use our existing `setup_venv.py` and then there can be a standalone test script that can be run once the package is installed to test it.
* Some complexity in the `build_tools/packaging/python/finalize_rocshmem4py_wheel.py` and `comm-libs/rocshmem4py/CMakeLists.txt` files. We have a few other python projects (amdsmi just came up recently) hoping to be distributed in a similar way and I'd prefer to have a simpler way that they can get included as part of the build->packaging->test systems. As this is now those projects would also need custom code across multiple files. Maybe we could add some shared CMake and script helper functions that projects could use?

PR #5642, opened 2026-06-05
Add PR-aware checkout fetch-depth action and script

cgoea:
## Motivation

For different security scanners we will need to calculate the commit depth of each PR. 

## Technical Details
Computes the minimal actions/checkout fetch-depth for the current event so workflows fetch just enough git history instead of cloning everything: for pull_request events it returns commits + 1.

## Test Plan
Unit-tested via compute_pr_depth_test.py (run by the existing unit_tests.yml job as part of pytest build_tools)

## Submission Checklist

### DIFF
--- a/.github/actions/compute_pr_fetch_depth/action.yml
+++ b/.github/actions/compute_pr_fetch_depth/action.yml
@@ -1,0 +1,29 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+name: "Compute PR-aware fetch-depth"
+
+description: >-
+  Returns the smallest `actions/checkout` fetch-depth that covers all
+  commits this workflow needs to walk.
+
+  * pull_request events: `commits + 1` (PR commits + merge-base).
+  * any other event (push, schedule, workflow_dispatch, ...): `0`
+    (full history). Use this when downstream tooling walks commit
+    history
+
+outputs:
+  value:
+    description: >-
+      String value to pass to actions/checkout fetch-depth. Already
+      coerced to a string so it composes cleanly with checkout's
+      input parser.
+    value: ${{ steps.compute.outputs.value }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Compute fetch-depth
+      id: compute
+      shell: bash
+      run: python build_tools/github_actions/compute_pr_depth.py
--- a/build_tools/github_actions/compute_pr_depth.py
+++ b/build_tools/github_actions/compute_pr_depth.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+"""Compute the smallest `actions/checkout` fetch-depth for the current event."""
+
+import sys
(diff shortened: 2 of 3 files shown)


build_tools/github_actions/compute_pr_depth.py
```diff
@@ -0,0 +1,32 @@
...
+# SPDX-License-Identifier: MIT
+"""Compute the smallest `actions/checkout` fetch-depth for the current event."""
+
+import sys
+from typing import Any
+
+from github_actions_api import gha_load_github_event, gha_set_output
+
+
+def compute_fetch_depth(payload: dict[str, Any]) -> str:
```
HereThereBeDragons, line 12: could/should this be part of github_actions_api ?

.github/actions/compute_pr_fetch_depth/action.yml
```diff
@@ -0,0 +1,29 @@
...
+name: "Compute PR-aware fetch-depth"
+
+description: >-
+  Returns the smallest `actions/checkout` fetch-depth that covers all
+  commits this workflow needs to walk.
+
+  * pull_request events: `commits + 1` (PR commits + merge-base).
+  * any other event (push, schedule, workflow_dispatch, ...): `0`
+    (full history). Use this when downstream tooling walks commit
+    history
```
ScottTodd, line 13: > ## Motivation
> 
> For different security scanners we will need to calculate the commit depth of each PR.
> ## Technical Details
> 
> Computes the minimal actions/checkout fetch-depth for the current event so workflows fetch just enough git history instead of cloning everything: for pull_request events it returns commits + 1.

Can you elaborate on the motivation here? Why is fetching more history a problem? The history in this repository is small, checkout takes < 5 seconds, and I don't think optimizing that is worth +130 LOC of scripts and workflows (plus additional overhead to actually run those scripts/workflows).

cgoea: @ScottTodd  here is not a problem but since the plan is to share this to other repos like rocm-systems/rocm-libraries there fetching the whole history takes a bit of time. thought it makes sense to shorten the fetch history time for PRs by pulling only what is necessary.

build_tools/github_actions/compute_pr_depth.py
cgoea: thought about it but since it's not really a reusable API call but more workflow specific logic so I preferred to keep it separated.

.github/actions/compute_pr_fetch_depth/action.yml
ScottTodd: So, add the complexity where it's needed, and not here?

cgoea: Seemed simpler to have all the scripts/workflows/helper scripts in one place. Spreading them in multiple repos for me makes things more complex.

ScottTodd (approved):

.github/actions/compute_pr_fetch_depth/action.yml
ScottTodd: That might lead to other issues (needing to update one repository to have those changes available in another repository), but okay.

PR #4901, opened 2026-04-28
Revert "[] Adds origami to TheRock build and tests"

davidd-amd:
Reverts ROCm/TheRock#3820

still failing the sdk tests.

I think we need this in the toml file:

```
[components.dbg."math-libs/BLAS/origami/stage"]
exclude = ["lib/python*/**"]
[components.dev."math-libs/BLAS/origami/stage"]
exclude = ["lib/python*/**"]
[components.lib."math-libs/BLAS/origami/stage"]
exclude = ["lib/python*/**"]
[components.test."math-libs/BLAS/origami/stage"]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
include = [
  "bin/origami-tests*",
  "bin/origami/**",
  "lib/python*/site-packages/**",
]     
```

Issue #2042, opened 2025-11-06
[hipSPARSELt] gfx architectures need to be excluded due to incomplete tensilelite support


bstefanuk:
PR #1659 adds hipsparselt to TheRock build system, however hipsparselt is not supported for most architectures
```
[hipSPARSELt configure] CMake Error at cmake/hipsparselt_supported_architectures.cmake:42 (message):
[hipSPARSELt configure]   Unsupported GPU target: gfx1201
[hipSPARSELt configure] 
[hipSPARSELt configure]   Supported targets are:
[hipSPARSELt configure]   gfx942;gfx950;gfx942:xnack+;gfx942:xnack-;gfx950:xnack+;gfx950:xnack-
[hipSPARSELt configure] Call Stack (most recent call first):
[hipSPARSELt configure]   CMakeLists.txt:87 (hipsparselt_validate_gpu_targets)
```
Currently the only supported architectures are gfx942 and gfx950. While these are technically supported by tensilelite, hipsparselt does not have tuned logic files for any other architectures. The hipsparselt library can still be built by setting `HIPSPARSELT_ENABLE_DEVICE=OFF`.

Issue #5024, opened 2026-05-04
[Issue]: Flaky "FAILED ... stage.stamp" Windows build failures during stage/install step


ScottTodd:
Sample failures:

* https://github.com/ROCm/rockrel/actions/runs/25268549810/job/74090141874#step:16:461
    ```
    [58](https://github.com/ROCm/rockrel/actions/runs/25268549810/job/74090141874#step:16:459)
    [05:39:52Z] Mem: 69.1/384.0GB (18%) | CPU: 100% | Jobs: ~17/96 | Disk: 323GB free
    [05:40:23Z] Mem: 65.0/384.0GB (17%) | CPU: 100% | Jobs: ~17/96 | Disk: 323GB free
    [180/263] Building sub-project rocFFT (in background)
    [181/263] Stage installing sub-project rocFFT
    FAILED: math-libs/rocFFT/stamp/stage.stamp B:/build/math-libs/rocFFT/stamp/stage.stamp 
    C:\Windows\system32\cmd.exe /C "cd /D B:\build\math-libs\rocFFT\build && C:\home\runner\_work\_tool\Python\3.12.10\x64\python3.exe C:/home/runner/_work/rockrel/rockrel/build_tools/teatime.py --log-timestamps --label "rocFFT install" --no-interactive B:/build/logs/rocFFT_install.log -- "C:/Program Files/CMake/bin/cmake.exe" --install B:/build/math-libs/rocFFT/build && C:\home\runner\_work\_tool\Python\3.12.10\x64\python3.exe C:/home/runner/_work/rockrel/rockrel/build_tools/fileset_tool.py copy B:/build/math-libs/rocFFT/dist B:/build/math-libs/rocFFT/stage B:/build/third-party/sysdeps/windows/zlib/build/stage B:/build/third-party/sysdeps/windows/zstd/build/stage B:/build/compiler/amd-llvm/stage B:/build/compiler/amd-comgr/stage B:/build/compiler/hipcc/stage B:/build/base/rocm-core/stage B:/build/core/rocm-kpack/stage B:/build/core/clr/stage B:/build/math-libs/rocRAND/stage B:/build/math-libs/hipRAND/stage && "C:\Program File[05:40:53Z] Mem: 65.7/384.0GB (17%) | CPU: 88% | Jobs: ~16/96 | Disk: 323GB free
    [05:41:23Z] Mem: 66.0/384.0GB (17%) | CPU: 100% | Jobs: ~15/96 | Disk: 323GB free
[...]

PR #5650, opened 2026-06-05
feat(math-libs-blas): changes to consume shared blas libraries

davidd-amd:
Re-lands origami into `math-libs/BLAS` and restructures how the BLAS libraries
produce and consume their shared support code: instead of each library statically
embedding a private copy, the shared code is **installed once as a shared library
with a CMake package** and consumed **idiomatically via `find_package` + imported
targets**. This is the TheRock side of the "P3 ratchet" — hipSPARSELt builds its
sparse device libraries against the **installed hipBLASLt package** rather than
rebuilding hipBLASLt inline with `add_subdirectory`.

## Motivation

origami and hipBLASLt's `tensilelite-host` are shared support code consumed by
more than one BLAS library. The default in-project path builds them **static and
embeds a private copy into each consumer**. Multiple consumers each carrying their
own copy means duplicate definitions of the same symbols in a process that loads
both — an **ODR hazard** — plus duplicated code and no single upgrade point.
Install-once / consume-via-`find_package` gives one definition per artifact,
shared by all consumers.

## What this does

**origami — install & consume as a shared library**
- Declared as a `math-libs/BLAS` subproject built from
  `rocm-libraries/shared/origami` with `ORIGAMI_BUILD_SHARED_LIBS=ON`; installs
  `liborigami.so`, headers, `share/doc/origami`, and a CMake package
  (`lib/cmake/origami`, `roc::origami`).
- hipBLASLt consumes it via `find_package(origami CONFIG)` + the `roc::origami`
  imported target (origami is in hipBLASLt's `RUNTIME_DEPS`) instead of a
  statically-embedded per-consumer copy.
- **Python bindings:** disabled (`-DORIGAMI_ENABLE_PYTHON=OFF`); only the C++
  library is built. Bindings and their nanobind/robin-map deps are deferred until
  a nanobind provider lands in TheRock (mirrors the hipDNN direction, #6425).
- Fixes the artifact-accounting gap behind the original revert (#4901): origami's
  `share/doc` licenses are claimed by a dedicated `doc` component so they are no
  longer silently dropped from the SDK. All claims are scoped to origami.

**hipSPARSELt — consume hipBLASLt via find_package (the ratchet)**
- `hipSPARSELt` gains `hipBLAS-common` in `BUILD_DEPS` (`find_package(hipblaslt)`
  loads `hipblaslt-config`, which runs `find_dependency(hipblas-common)`; the
  transitive package must be declared for the super-project dependency provider)
  and `hipBLASLt` in `RUNTIME_DEPS` (co-locates hipBLASLt's shared
  tensilelite-host / origami / rocisa in hipSPARSELt's dist).
- `artifact-blas.toml` claims `lib/hipblaslt/rocisa/**` in the hipBLASLt `lib`
  component so the installed rocisa package (`_rocisa.*.so` + `__init__.py`) is
  captured in artifacts.

**Packaging & CI wiring**
- origami test execution script (`build_tools/github_actions/.../test_origami.py`)
  and its test-configuration entry; `package.json` / `_dist_info.py` updates for
  the origami component.

## rocm-libraries pin

No net advance over main. The origami fixes this depends on — relocatable
installed CTestTestfile (#9047) and Windows rocisa DLL resolution (#9309) — are
already merged to `develop` and present in main's submodule pin, so the carried
downstream patch is **removed** (`patches/amd-mainline/rocm-libraries/` is gone).
The branch pin is an ancestor of main's, so the gitlink merges forward cleanly.

## Ratchet is additive / pin-agnostic

The TheRock wiring here **coexists with the current rocm-libraries pin** and does
not require the rocm-libraries ratchet commit to be present. At this pin
hipSPARSELt still builds hipBLASLt inline via `add_subdirectory` in its own build
tree (a separate CMake invocation from TheRock's standalone hipBLASLt subproject
— no target/configure collision). The ratchet becomes *active* (hipSPARSELt
consuming the installed hipBLASLt via `find_package(hipblaslt)` +
`hipblaslt_create_device_library`) once a later, automated pin bump carries the
rocm-libraries ratchet change. So this PR is **not gated on pin ordering**.

## Verification

- origami re-land: full from-source `gfx90a` superbuild → exit 0; origami builds
  and installs to `dist/rocm`, licenses claimed (no unmatched-artifact failure).
- Ratchet path (at the ratchet rocm-libraries pin, separately): gfx942
  `hipSPARSELt` via `find_package(hipblaslt)` → rc=0 with sparse
  `TensileLibrary_*_gfx942.co/.dat.zlib` generated and rocisa installed with
  hipBLASLt; gfx90a full chain origami → hipBLASLt → hipSPARSELt → rc=0.
- Multi-arch build at the current pin is validated by this PR's CI.

## Notes / follow-ups

- Required by AIHPBLAS-3522.
- hipSPARSELt is excluded on gfx90a (#2042).
- The rocm-libraries ratchet (hipSPARSELt `find_package(hipblaslt)` device-codegen
  + installable rocisa) lands separately in rocm-libraries; its pin bump is
  automated and is what activates the ratchet in TheRock.


math-libs/BLAS/CMakeLists.txt
```diff
@@ -1,6 +1,3 @@
-# Copyright Advanced Micro Devices, Inc.
-# SPDX-License-Identifier: MIT
-
```
ScottTodd, line 3: Add back copyright header

math-libs/BLAS/artifact-blas.toml
```diff
@@ -24,6 +24,32 @@ include = [
...
+# scanner fails ("unmatched"), which is what broke the rocm-sdk build on prior
+# attempts (#3820/#3237/#2813). The doc default glob (**/share/doc/**) claims it.
+[components.doc."math-libs/BLAS/origami/stage"]
+[components.test."math-libs/BLAS/origami/stage"]
+include = [
+  "bin/origami-tests*",
+  "bin/origami/**",
+  # Python extension is fetched here so the test script can set PYTHONPATH
+  # to lib/pythonX.Y/site-packages/ for ctest Python tests.
+  "lib/python*/site-packages/**",
```
ScottTodd: This looks risky. Python `site-packages/` directories can easily include dozens of other packages, so I think we should either
1. Filter to just the expected packages
2. Have the project's build system copy/install the files it wants to distribute out of `lib/python*/site-packages` and to some safer location

build_tools/audit_artifact_accounting.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def discover_descriptors(repo: Path) -> list[Path]:
+    out = []
+    for p in repo.rglob("artifact-*.toml"):
+        s = str(p)
+        if "/build/" in s or "/.claude/" in s or "/_therock_utils/" in s:
+            continue
+        out.append(p)
+    return sorted(out)
```
ScottTodd: Huh? I don't understand this code. Needs comments / docstrings / something.

* Why would we ever hardcode `/.claude/` in such a file?
* Why convert from `Path` to `str` and then search using `/`? Windows uses `\` as the separator

tests/test_artifact_accounting.py
```diff
@@ -0,0 +1,51 @@
...
+            failures[str(toml.relative_to(_REPO))] = undeclared
+
+    assert audited > 0, f"no built stage dirs found under {root}"
+    if failures:
+        lines = [
+            "Unaccounted staged files (claimed by no artifact component; would be",
+            "silently dropped and break the rocm-sdk build). Add the right component",
+            'entry to the descriptor (e.g. [components.doc."<stage>"] for share/doc),',
+            "or allow-list via options.unmatched_exclude:",
+        ]
```
ScottTodd: I'm pretty sure "undeclared" files should be fine, not errors. Not all projects want to distribute all files and I don't see how this "breaks the rocm-sdk build". Did you run this on all existing artifacts? Can you share logs of it running?

This is also pretty similar to https://github.com/ROCm/TheRock/blob/main/tests/test_artifact_structure.py which is wired into CI with https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_artifacts_structure.yml. The PR that added those was https://github.com/ROCm/TheRock/pull/3802. Here's a recent CI run showing that in action: https://github.com/ROCm/TheRock/actions/runs/27147861883/job/80158100996#step:6:27 . We could extend the tests in that file instead of adding a new file that isn't connected to any CI.

math-libs/BLAS/post_hook_origami.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+add_test(origami_python_tests python3 -m pytest "${CMAKE_CURRENT_LIST_DIR}/tests" -v)
+set_tests_properties(origami_python_tests PROPERTIES
+    LABELS "python"
+    SKIP_RETURN_CODE 5
+    WORKING_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}"
+)
+]=])
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/CTestTestfile.cmake.therock" "${_ctest_content}")
```
ScottTodd: Huh? Why add tests for a subproject via a post_hook? This should just go in the project itself. It's at least worth a comment here explaining why that natural option was not chosen.

math-libs/BLAS/CMakeLists.txt
```diff
@@ -353,13 +385,15 @@ if(THEROCK_ENABLE_SPARSE)
...
       BUILD_DEPS
+        origami
         rocm-cmake
         hipSPARSE
         rocSPARSE
         therock-googletest
         therock-msgpack-cxx
       RUNTIME_DEPS
         hip-clr
+        origami
```
ScottTodd, line 400: We typically list each project in _either_ `BUILD_DEPS` _or_ `RUNTIME_DEPS`, not both. Which is origami?

See https://github.com/ROCm/TheRock/blob/38fe1711d1a7385673d53d6a8bdec99c8a544c60/cmake/therock_subproject.cmake#L244-L248

You can consider `RUNTIME_DEPS` to be a superset of `BUILD_DEPS`

math-libs/BLAS/CMakeLists.txt
```diff
@@ -93,6 +90,39 @@ if(_enable_rocRoller)
...
+  BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/origami"
+  BACKGROUND_BUILD
+  CMAKE_ARGS
+    -DHIP_PLATFORM=amd
+    -DORIGAMI_BUILD_SHARED_LIBS=ON
+    -DORIGAMI_BUILD_TESTING=${THEROCK_BUILD_TESTING}
+    -DORIGAMI_ENABLE_PYTHON=ON
+    -DORIGAMI_ENABLE_FETCH=OFF
+  IGNORE_PACKAGES
+    nanobind
```
ScottTodd: What's this `IGNORE_PACKAGES` doing here? Is origami using its own nanobind package? Shouldn't we use nanobind from TheRock?

build_tools/audit_artifact_accounting.py
```diff
@@ -0,0 +1,140 @@
...
+
+def audit_descriptor(toml: Path, root: Path):
+    name = toml.stem
+    if name.startswith("artifact-"):
+        name = name[len("artifact-") :]
+    try:
+        ad = ab.ArtifactDescriptor.load_toml_file(toml, artifact_name=name)
+        scanner = ab.ComponentScanner(root, ad)
+    except Exception as e:
+        return name, [], [], [], f"{type(e).__name__}: {e}"
```
ScottTodd: This code is unreadable. No annotated return type (a tuple of 5 items?) + no docstring. Two letter variable names, no comments.

Please follow https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md

build_tools/audit_artifact_accounting.py
```diff
@@ -0,0 +1,140 @@
...
+        required=True,
+        type=Path,
+        help="Build root the descriptor basedirs are relative to (THEROCK_BINARY_DIR).",
+    )
+    ap.add_argument(
+        "--repo",
+        type=Path,
+        default=_HERE.parent,
+        help="Repo root to discover artifact-*.toml under (default: TheRock root).",
+    )
```
ScottTodd: Will this script ever run with a different `--repo`? Seems unnecessary to add an argument for this.

math-libs/BLAS/artifact-blas.toml
astrelsky: Why not have the origami build create a wheel itself. Then if it wants to be included in TheRock wheels, the origami wheel contents can just be extracted into the directory when created TheRock wheels.

That may not be the clearest explanation, but I think it's the safest way to do it.

I also think the sanest way to test that a python extension going to work when the end user gets it and installs it, is to install it as a wheel and test against that in the ci. But I think that is out of scope for this. (This was mostly a drive-by comment and I'm trying not to get too involved or stick my nose where it doesn't belong.)

tests/test_artifact_accounting.py
davidd-amd: Removed - the thinking at the time was the unaccounted artifacts were landing in the wrong place and breaking the sdk.

build_tools/audit_artifact_accounting.py
davidd-amd: Removed - local test code that shouldn't have been committed. Good catch, thanks.

math-libs/BLAS/artifact-blas.toml
davidd-amd: Filters on just origami now.

math-libs/BLAS/CMakeLists.txt
davidd-amd: Restored

math-libs/BLAS/post_hook_origami.cmake
davidd-amd: So we have to do both because there is a delay in the time it takes for TheRock to pull in the fix from rocm-libraries. When it lands we can remove this file or we block this branch until rocm-libraries bump reaches TheRock.

marbre: I really think this needs to go in through a rocm-libraries submodule as it will otherwise be troublesome to land land the a new r-l bump. Feels rather like a patchfile, which is also discouraged.

math-libs/BLAS/CMakeLists.txt
marbre: We should.

geomin12 (approved):
lgtm from testing aspect, i think the reviews from previous iterations still stand here

0xDELUXA:
It's quite disappointing to see that `rocm-libraries` doesn't seem to actively consider external contributions: https://github.com/ROCm/rocm-libraries/pull/6524.
This PR simply adds `gfx1200` support to `origami`, and I don't see any obvious reason why it couldn't be merged. It would be great to see community contributions being reviewed and integrated more actively.

davidd-amd:
> It's quite disappointing to see that `rocm-libraries` doesn't seem to actively consider external contributions: [ROCm/rocm-libraries#6524](https://github.com/ROCm/rocm-libraries/pull/6524).

Hi @0xDELUXA looks like you are getting traction on your contribution but let me know if not. We highly encourage and value external contributions so thanks for your efforts.

0xDELUXA:
> 
> Hi @0xDELUXA looks like you are getting traction on your contribution but let me know if not. We highly encourage and value external contributions so thanks for your efforts.

Thanks, I appreciate the encouragement. Things seem to be progressing. Happy to contribute!

rocm-libraries
marbre, line 1: A submodule bump should go into a separate PR and is needed to split out before this can be merged.

build_tools/github_actions/test_executable_scripts/test_origami.py
```diff
@@ -0,0 +1,84 @@
...
+# than globbing, since other components (e.g. rocprofiler-sdk) install
+# python3.10/python3.11/python3.12/python3.13 site-packages simultaneously
+# and a glob could resolve to the wrong version.
+python_version = f"python{sys.version_info.major}.{sys.version_info.minor}"
+site_packages_dir = lib_dir / python_version / "site-packages"
+if not (site_packages_dir / "origami").is_dir():
+    raise RuntimeError(
+        f"origami package not found in {site_packages_dir} -- "
+        f"was it built for {python_version}?"
+    )
```
marbre: I am still not convinced that this will work flawlessly. This limits to the availability of origami tests to Python 3.12. Every test runner not using Python 3.12 will hit the runtime error and tests will fail. @geomin12 I don't think we can or rather want to guarantee and pin a specific Python version,

math-libs/BLAS/CMakeLists.txt
```diff
@@ -94,6 +94,40 @@ if(_enable_rocRoller)
...
+  BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/origami"
+  BACKGROUND_BUILD
+  CMAKE_ARGS
+    -DHIP_PLATFORM=amd
+    -DORIGAMI_BUILD_SHARED_LIBS=ON
+    -DORIGAMI_BUILD_TESTING=${THEROCK_BUILD_TESTING}
+    # Python bindings (and their nanobind/tsl-robin-map deps) are built out of
+    # TheRock; land the C++ library only. Re-enable alongside a nanobind
+    # provider once bindings have a home (mirrors the hipDNN direction, #6425).
+    -DORIGAMI_ENABLE_PYTHON=OFF
```
ScottTodd, line 112: Can you update the PR description to reflect the latest code?

> * **Hermetic Python bindings:** nanobind is resolved from TheRock's provided
>       `therock-nanobind` (+ `therock-robin-map`, pulled transitively by
>       `nanobind_add_module()`), replacing the build-time GitHub FetchContent and the
>       `IGNORE_PACKAGES nanobind` / `post_hook_origami.cmake` workarounds. The origami
>       side of this is now upstream in rocm-libraries (#7996), so **no carried patch is
>       needed** — see "rocm-libraries pin" below.

is outdated. Same with

> ## rocm-libraries pin
> 
> Advances the `rocm-libraries` submodule **30 commits past main** to develop-tip `23f6afcf` ("fix(origami): emit a relocatable installed CTestTestfile", #9047). This pin carries both previously-downstream origami fixes — the provided-nanobind change (#7996) and the relocatable installed-CTest fix (#9047) — so the carried downstream patch is **removed entirely** (`patches/amd-mainline/rocm-libraries/` is gone). Main's pin is an ancestor of this one, so the pin moves cleanly forward.

---

I'd recommend writing PR descriptions yourself instead of using AI if this is becoming a pattern, see https://llvm.org/docs/AIToolPolicy.html which we take inspiration from
> To ensure sufficient self review and understanding of the work, it is strongly recommended that contributors write PR descriptions themselves (if needed, using tools for translation or copy-editing).

(verbose PR descriptions that include outdated and irrelevant details make safe and fast review more difficult - the PR description should be a _curated summary_ of the changes to aid in review, not a catalog of the full agent state and history behind a long feature branch)

ScottTodd (approved):
Code changes LGTM.

I retriggered a job on https://github.com/ROCm/TheRock/actions/runs/29832473344?pr=5650 after a Windows build flake (looked like https://github.com/ROCm/TheRock/issues/5024 ?). I'd like to see the Windows python package builds and tests succeeding after this build system change before merge.

PR #5655, opened 2026-06-05
[ci] Add vultr runner for submodule changes 

geomin12:
As we are determining core42 regression issues, we want to make the vultr runner run for submodule updates (to provide stability)

This is temporary until we resolve core42 issues

### DIFF
--- a/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
+++ b/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
@@ -1378,9 +1378,7 @@ class TestMultiLabelRunnerSelection(unittest.TestCase):
         self.assertIsNotNone(builds.linux)
         # Check that the third label was selected
         gfx94x_info = builds.linux.per_family_info[0]
-        self.assertEqual(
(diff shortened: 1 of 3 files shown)


build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1348,7 +1372,9 @@ def test_third_label_selected_when_random_high(self):
 
         # Mock random.random() to return 0.5 (>= 0.369+0.086=0.455)
         with patch("random.random", return_value=0.5):
-            builds = cm.expand_build_configs(targets, ci_inputs, test_type="quick")
+            builds = cm.expand_build_configs(
```
jayhawk-commits, line 1374: The purpose of this `test_third_label_selected_when_random_high` function is unclear to me when prior to this PR, `test-runs-on-labels` only had two entries. Now with this PR, the third label doesn't map to the `0.5` value in this test.

jayhawk-commits (approved):
OK for improving test stability on submodule bump PRs, but a concern on one existing test case's intended behaviour.

PR #5660, opened 2026-06-05
[hipfile] Add a oneTBB dependency for hipFile

akolliasAMD:
## Motivation

hipFile has a dependency on oneTBB (formerly an Intel product, now open-source), which is necessary for thread pools in hipFile's batch API. This PR adds the dependency.

## NOTE

This PR is required for getting hipFile into TheRock. The hipFile --> TheRock PR is here:
https://github.com/ROCm/TheRock/pull/5217

hipFile also has a second dependency on util-linux, which can be found here:
https://github.com/ROCm/TheRock/pull/5337

## Jira ID

AIHIPFILE-170
AIHIPFILE-206

## Technical Details

Adds a third-party dependency via the usual mechanism. oneTBB is pulled as a tarball and configured and built for packaging with TheRock.

The feature is gated on -DTHEROCK_ENABLE_SYSDEPS_ONETBB.

## Test Plan

CI passes, manual inspection of artifacts. The project that needs this dependency is not yet in TheRock.

Everything passes, artifacts are named appropriately, RPATHs are correct.

## Submission Checklist


third-party/sysdeps/linux/onetbb/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    # When included in TheRock, we download sources and set up the sub-project.
+    set(_source_dir "${CMAKE_CURRENT_BINARY_DIR}/source")
+    set(_download_stamp "${_source_dir}/download.stamp")
+
+    therock_subproject_fetch(therock-onetbb-sources
+      SOURCE_DIR "${_source_dir}"
+      # Originally downloaded from: https://github.com/uxlfoundation/oneTBB/archive/refs/tags/v2023.0.0.tar.gz
+      URL "https://github.com/uxlfoundation/oneTBB/archive/refs/tags/v2023.0.0.tar.gz"
```
ScottTodd: We'll need to upload this to our `rocm-third-party-deps` S3 mirror

docs/development/dependencies.md
```diff
@@ -201,6 +201,11 @@ SIMDe (SIMD Everywhere) is a header-only portability library for SIMD intrinsics
 - Import library: `SQLite::SQLite3`
 - Alternatives: none
 
+## oneTBB
```
ScottTodd, line 226: > ## Motivation
> 
> hipFile has another dependency on onetbb. This PR adds the dependency.

Can you elaborate for those of us not already familiar with hipFile or onetbb?

* What is onetbb? https://github.com/uxlfoundation/oneTBB ?
* What is it used for? Is it optional or mandatory?
* How large/complex is it as a dependency?

Looks like it's being used for threadpools and such (e.g. https://github.com/ROCm/rocm-systems/pull/6605)? And it's a _new_ dependency, not one that would be grandfathered in for existing projects? This doesn't seem worth all the complexity and risk of taking a dependency to me... I think we had some discussions about that before on chat (cc @stellaraccident )

third-party/onetbb/patch_install.py
ScottTodd, line 1: Why is onetbb a [Linux] sysdep at all? Isn't it just a few C++ source files that you'd link directly into your library? I don't see why that should be using a shared object with all this file patching, symbol hiding, etc.?

See https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md#dependencies, I would expect a thread pool implementation to be the simpler option there - more similar to flatbuffers/boost/grpc/eigen/etc. and less similar to libdrm/amd-mesa/etc.

ScottTodd (changes requested):

docs/development/dependencies.md
derobins: I have added some explanations above. The discussion concerning the dependency on oneTBB has already been resolved. It is mandatory and a requirement for hipFile to be ready for 7.14. We will not be re-engineering hipFile to use something else.

third-party/onetbb/patch_install.py
derobins: See above. It is far too late in the game for re-engineering discussions.

third-party/sysdeps/linux/onetbb/CMakeLists.txt
derobins: This should not be a problem. I know who to contact about this and we already got the util-linux tarball uploaded for the other PR. This one should go in soon.

derobins: I requested that the tarball be added

third-party/onetbb/patch_install.py
ScottTodd: A sysdep is the _much more complicated_ option for a dependency, with higher costs for complexity, deployment, and runtime overhead. See https://gist.github.com/stellaraccident/866acb0cae35e9f59c0c67d2fe3773c1 (linked from https://github.com/ROCm/TheRock/tree/main/dockerfiles#build_manylinux_dockerfile, and we should still also merge its contents into https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md )

I don't see a world where a threadpool should be a _redistributable_ dependency (https://gist.github.com/stellaraccident/866acb0cae35e9f59c0c67d2fe3773c1#build-only-dependencies-third-party) and not a _build_ dependency (https://gist.github.com/stellaraccident/866acb0cae35e9f59c0c67d2fe3773c1#build-only-dependencies-third-party). You should want to compile the code directly into your project to benefit from LTO and other optimizations, not link against a shared object and go through so many layers of indirection for core functionality. Better yet, just write the code you need for something so foundational, then we won't all have to deal with CVEs, complexity of validating/testing dependency updates (reuploading to the S3 mirror when you want new changes, etc.).

A sysdep is typically a library that you can find via OS package managers that many other applications on the system would want to link against. This is not that (https://github.com/uxlfoundation/oneTBB/blob/master/INSTALL.md), so we don't need this complexity, just link against the library as other projects do for other deps in https://github.com/ROCm/TheRock/tree/main/third-party (not https://github.com/ROCm/TheRock/tree/main/third-party/sysdeps)

I'm drawing a line here. You can escalate to @stellaraccident if you want, but this isn't making sense to me. You're choosing the more complicated option for little benefit and significant cost.

ScottTodd (changes requested):

third-party/sysdeps/linux/onetbb/CMakeLists.txt
ScottTodd: uploaded to https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/oneTBB-2023.0.0.tar.gz , hash I believe should be `F8767B971EC6AEA25DDE58AE0F593E94E7AA75A739A86F67967012F69E2199B1`

derobins:
We can't build as a static library because one of the libraries we need can only be built as a shared library. oneTBB warns that static builds are unwise and not supported anyway.

We have moved it out of sysdeps so it's now just a regular dependency.

BUILD_TOPOLOGY.toml
```diff
@@ -466,6 +466,13 @@ type = "target-neutral"
 artifact_deps = []
 feature_group = "CORE"
 
+[artifacts.onetbb]
+artifact_group = "third-party-libs"
+type = "target-neutral"
+artifact_deps = []
+feature_group = "CORE"
+disable_platforms = ["windows"]
```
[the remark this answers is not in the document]
derobins: Darwin isn't supported

gaoikawa:
FYI - from @dgaliffiAMD :  https://github.com/ROCm/TheRock/pull/4410/

This PR might be a duplicate of that then.

gaoikawa:
FYI - Closing this PR.  Conclusions of the discussions today internally are that hipFile team will pursue other options than TBB.

Issue #5634, opened 2026-06-04
[Multi-arch] Build multi-arch JAX Python packages


ScottTodd:
Similar to https://github.com/ROCm/TheRock/issues/3332 (PyTorch)

marbre:
@ScottTodd, @erman-gurses I think we can close this safely now with what landed to TheRock. Any additional work should use new, updated issues. Feel free to re-open if you disagree.

PR #5633, opened 2026-06-04
Generate and checkout from manifests in PyTorch build workflows

ScottTodd:
## Motivation

Progress on:
* https://github.com/ROCm/TheRock/issues/5110
* https://github.com/ROCm/TheRock/issues/1236

This builds on https://github.com/ROCm/TheRock/pull/5503 to have PyTorch release build workflows either...
1. Generate a manifest from the requested ref (as when using workflow_dispatch of a single configuration directly)
2. Passthrough an already generated manifest

... then the build jobs use the generated/provided manifest to drive source code checkout via the unified `checkout_from_manifest.py` script instead of individual `pytorch_*_repo.py` scripts.

Follow-up work will include:
- [ ] Generate all manifests in [`multi_arch_release_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml) so all jobs are guaranteed to use the same commits
- [ ] Generate all manifests in a new "schedule release" job so those commits are frozen across re-runs

## Technical Details

PyTorch manifest files are JSON like this (https://therock-dev-artifacts.s3.amazonaws.com/26978198370-linux/manifests/pytorch/pytorch_manifest.json):
```jsonc
{
  "schema_version": 1,
  "pytorch": {
    "repo": "https://github.com/pytorch/pytorch",
    "commit": "2d7a24d8c37470f1729ea583bd5d5eed7427f6af",
    "branch": "nightly",
    "version": "2.13.0a0+devrocm7.14.0.dev0-bf2a9c03ea9402047d972ff68334a716ea210f16"
  },
  "pytorch_audio": {
    "repo": "https://github.com/pytorch/audio",
    "commit": "55b1283593bf80e897f04f1ad3b69d6ad2fe0935",
    "branch": "nightly",
    "version": "2.11.0a0+devrocm7.14.0.dev0-bf2a9c03ea9402047d972ff68334a716ea210f16"
  },
}
```

## Test Plan

* Linux, generate manifest from 'nightly' and use it: https://github.com/ROCm/TheRock/actions/runs/26978198370
* Linux, use an already generated release/2.9 manifest: https://github.com/ROCm/TheRock/actions/runs/26978242094
* Watch nightly releases to ensure all data continue to propagate correctly

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5661, opened 2026-06-05
Run PyTorch tests in multi-arch release workflows

### LINKED ISSUE #5496: [Multi-arch] Run PyTorch tests after building packages

ScottTodd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/5496. Now nightly releases of multi-arch PyTorch packages will run the standard set of PyTorch tests in [`test_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml). Note that https://github.com/ROCm/TheRock/pull/4499 will similarly trigger [`test_pytorch_wheels_full.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels_full.yml).

As advertised at https://github.com/ROCm/TheRock#nightly-release-status, see nightly workflow run history at:
* https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_linux_pytorch_wheels.yml
* https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_windows_pytorch_wheels.yml

that's where we'll start to see tests running with these changes.

## Technical Details

The "manifest" TODOs here relate to https://github.com/ROCm/TheRock/issues/5110 and https://github.com/ROCm/TheRock/pull/5633 - test jobs should use the same PyTorch commit as the build job, not just the same floating ref and hope that it resolves to the same commit when the tests run.

The matrix generation script could be shared with JAX for https://github.com/ROCm/TheRock/issues/5634, and maybe even ROCm artifact tests alongside [`build_tools/github_actions/fetch_test_configurations.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/fetch_test_configurations.py). The "test for what was built, or this subset" behavior is generic enough.

Note that unlike [`build_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml) the multi-arch release workflows have no concept of "upload to staging -> test -> promote", see https://github.com/ROCm/TheRock/pull/5107.

## Test Plan

1. Test workflow syntax with workflow_dispatch triggers of these `multi_arch_build_*pytorch_wheels.yml` workflows. Watch for the "configure pytorch tests" step to generate the expected matrix then the requested test runner labels to be used for test jobs that install and test the packages built by the build jobs:
    * Linux gfx94X;gfx950 with 'auto' testing, torch 2.11: https://github.com/ROCm/TheRock/actions/runs/27163214035
    * Windows gfx1151 with 'none' testing, torch 2.12: https://github.com/ROCm/TheRock/actions/runs/27045896457
2. After merge, watch nightly release runs for the full matrix of Python versions, PyTorch versions, AMDGPU targets, and platforms. Test results should be similar to those for the nightly single arch PyTorch releases currently running here in TheRock (e.g. https://github.com/ROCm/TheRock/actions/workflows/release_portable_linux_pytorch_wheels.yml?query=branch%3Amain)
3. If all goes well (or not worse than for single-arch releases), turn down the testing in single-arch release workflows like [`build_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml), see https://github.com/ROCm/TheRock/issues/3340

## Test Result

Linux https://github.com/ROCm/TheRock/actions/runs/27163214035/job/80192318135 configured
```
Requested linux AMDGPU families: ['gfx950-dcgpu', 'gfx94X-dcgpu']
Including gfx950-dcgpu: testing on linux-gfx950-1gpu-ccs-ossci-rocm
Including gfx94X-dcgpu: testing on linux-gfx942-1gpu-core42-ossci-rocm
```

then ran tests on those runners, e.g.
```
= 39637 passed, 1957 skipped, 41 deselected, 48 xfailed, 6 subtests passed in 1002.12s (0:16:42) =
```

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/configure_pytorch_test_matrix.py
+++ b/build_tools/github_actions/configure_pytorch_test_matrix.py
@@ -25,12 +25,8 @@ from github_actions.amdgpu_family_matrix import get_all_families_for_trigger_typ
 from github_actions.github_actions_api import gha_set_output
 
 
-SKIP_TEST_VALUES = {"none", "skip", "false", "off", "0"}
-AUTO_TEST_VALUES = {"", "auto", "built"}
-
-
 def split_families(value: str) -> list[str]:
-    return [item.strip() for item in value.replace(";", " ").split() if item.strip()]
+    return [f.strip() for f in value.split(";") if f.strip()]
 
 
 def find_amdgpu_family(*, requested_family: str, platform: str) -> tuple[str, str, str]:
@@ -60,6 +56,7 @@ def build_test_matrix(
     amdgpu_families: list[str],
     platform: str,
 ) -> dict[str, list[dict[str, str]]]:
+    print(f"Requested {platform} AMDGPU families: {amdgpu_families}")
     include: list[dict[str, str]] = []
     seen_families: set[str] = set()
     for requested_family in amdgpu_families:
@@ -75,34 +72,15 @@ def build_test_matrix(
             print(f"Skipping {matrix_family}: no {platform} test runner is configured")
             continue
 
+        print(f"Including {test_family}: testing on {test_runs_on}")
         include.append({"amdgpu_family": test_family, "test_runs_on": test_runs_on})
 
     return {"include": include}
 
 
-def resolve_requested_test_families(
-    *, build_amdgpu_families: str, test_amdgpu_families: str
-) -> list[str]:
-    """Return the AMDGPU families to test.
-
-    Empty/auto/built test input means "test all built families". Explicit
-    none/skip disables tests.
-    """
-    build_families = split_families(build_amdgpu_families)
-    test_value = test_amdgpu_families.strip()
-    test_value_lower = test_value.lower()
-    if test_value_lower in SKIP_TEST_VALUES:
-        return []
-    if test_value_lower in AUTO_TEST_VALUES:
-        return build_families
-    return split_families(test_amdgpu_families)
-
-
 def emit_outputs(matrix: dict[str, list[dict[str, str]]]) -> None:
-    include = matrix["include"]
     gha_set_output(
         {
-            "enabled": str(bool(include)).lower(),
             "matrix": json.dumps(matrix, separators=(",", ":")),
         }
     )
@@ -113,15 +91,14 @@ def main(argv: list[str]) -> None:
     parser.add_argument(
         "--build-amdgpu-families",
         required=True,
-        help="Semicolon- or space-separated AMDGPU families that were built.",
+        help="Semicolon-separated AMDGPU families that were built.",
     )
     parser.add_argument(
         "--test-amdgpu-families",
         default="",
         help=(
-            "Semicolon- or space-separated AMDGPU families to test. "
-            "Use 'auto' or leave empty to test built families. Use 'none' "
-            "to skip tests."
+            "Semicolon-separated AMDGPU families to test. Use 'auto' or leave "
+            "empty to test built families. Use 'none' to skip tests."
         ),
     )
     parser.add_argument(
@@ -131,10 +108,16 @@ def main(argv: list[str]) -> None:
         help="Test platform (default: current system).",
     )
     args = parser.parse_args(argv)
-    test_amdgpu_families = resolve_requested_test_families(
-        build_amdgpu_families=args.build_amdgpu_families,
-        test_amdgpu_families=args.test_amdgpu_families,
-    )
+
+    built_families = split_families(args.build_amdgpu_families)
+    test_families_arg = args.test_amdgpu_families.strip().lower()
+    if test_families_arg in ("", "auto", "built"):
+        test_amdgpu_families = built_families
+    elif test_families_arg in ("none", "skip"):
+        test_amdgpu_families = []
+    else:
+        test_amdgpu_families = split_families(args.test_amdgpu_families)
+
     matrix = build_test_matrix(
         amdgpu_families=test_amdgpu_families,
         platform=args.platform,
--- a/build_tools/github_actions/configure_pytorch_test_matrix.py
+++ b/build_tools/github_actions/configure_pytorch_test_matrix.py
@@ -81,6 +81,7 @@ def build_test_matrix(
 def emit_outputs(matrix: dict[str, list[dict[str, str]]]) -> None:
     gha_set_output(
         {
+            "enabled": str(bool(matrix["include"])).lower(),
             "matrix": json.dumps(matrix, separators=(",", ":")),
         }
     )
--- a/build_tools/github_actions/configure_pytorch_test_matrix.py
+++ b/build_tools/github_actions/configure_pytorch_test_matrix.py
@@ -2,12 +2,20 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""Configure quick PyTorch wheel test jobs for multi-arch build workflows."""
+"""Configure PyTorch wheel test jobs for multi-arch build workflows.
+
+TODO(#5110): Extract the AMDGPU family -> test runner policy once JAX needs
+the same flow. Standalone workflow_dispatch runs can keep using raw family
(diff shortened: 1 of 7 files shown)


ethanwee1 (approved):
One convergence point worth noting: this PR establishes two nice primitives that #4499 currently works around package_index_url as an output of the publish step, and the exact torch_version build outputs. #4499 today hardcodes the multi-arch index URL and re-resolves the torch version via pip install --dry-run against the live index (which risks picking up a stale wheel if a build cell failed). Once this lands, I'll update #4499 to consume these outputs instead, so both paths share the same index-URL source of truth and test exactly what was built. I'll track that on #4499.

ethanwee1 (approved):
LGTM. Complementary to #4499 with no file overlap, so the two should land independently.

The split is clean: this PR runs the standard suite (test_pytorch_wheels.yml) inline in the multi-arch build workflows per build cell, while #4499 dispatches the full suite (test_pytorch_wheels_full.yml) from the release orchestrator on the nightly cadence.

This PR also gives us two primitives #4499 currently works around: package_index_url as a publish-step output, and the exact torch_version build outputs. Once this merges I'll update #4499 to consume those instead of hardcoding the index URL and re-resolving torch via pip --dry-run, so both paths share one source of truth and test exactly what was built. Tracking on #4499.

One minor nit (non-blocking): in configure_pytorch_test_matrix.py, find_amdgpu_family compares key/family with requested_lower but checks `requested_family in fetch_targets` with the original case, so a gfx target in a different case won't match the fetch-target branch.

build_tools/github_actions/configure_pytorch_test_matrix.py
```diff
@@ -0,0 +1,130 @@
...
+        platform_info = info_for_key.get(platform)
+        if not platform_info:
+            continue
+
+        family = platform_info["family"]
+        fetch_targets = platform_info.get("fetch-gfx-targets", [])
+
+        if requested_lower == key.lower() or requested_lower == family.lower():
+            return family, family, platform_info["test-runs-on"]
+        if requested_family in fetch_targets:
```
erman-gurses: Canonical family matching is case-insensitive, but this alias lookup is case-sensitive. Is that intentional, or should fetch_targets be normalized as well?

build_tools/github_actions/configure_pytorch_test_matrix.py
```diff
@@ -0,0 +1,130 @@
...
+            return requested_family, family, platform_info["test-runs-on"]
```
erman-gurses: Should fetch-target matches return `requested_family` here rather than the canonical family? I assume this is intentional so matrix entries and deduplication are based on the originally requested test target, but I wanted to confirm that behavior.

erman-gurses (approved):
LGTM - I put some minor comments.

build_tools/github_actions/configure_pytorch_test_matrix.py
```diff
@@ -0,0 +1,130 @@
...
+    args = parser.parse_args(argv)
+
+    built_families = split_families(args.build_amdgpu_families)
+    test_families_arg = args.test_amdgpu_families.strip().lower()
+    if test_families_arg in ("", "auto", "built"):
+        test_amdgpu_families = built_families
+    elif test_families_arg in ("none", "skip"):
+        test_amdgpu_families = []
+    else:
+        test_amdgpu_families = split_families(args.test_amdgpu_families)
```
marbre, line 115: From my review agent:

---

Consider:
  - test_families_arg = "auto;gfx94x-dcgpu" → no match → falls to else → split_families("auto;gfx94X-dcgpu") → passes ["auto",  "gfx94X-dcgpu"] to build_test_matrix → find_amdgpu_family is called with "auto" → raises ValueError: No linux AMDGPU family entry found for 'auto'

The simplest fix would be to use the lowercased arg consistently:

```suggestion
    else:
        # Use the original (not lowercased) value since family names are case-sensitive.
        # The sentinel check above already consumed 'auto'/'none'/etc.
        test_amdgpu_families = split_families(args.test_amdgpu_families)
        # Guard against sentinels accidentally appearing in a list.
        sentinels = {"auto", "built", "none", "skip"}
        for f in test_amdgpu_families:
            if f.lower() in sentinels:
                raise ValueError(
                    f"Sentinel value {f!r} cannot be mixed with explicit family names"
                )
```

Or more simply, just normalize inside split_families's caller and check there. It's a minor robustness issue — the current code works fine in the expected usage paths, and the workflow input has a clear description. But if this script gets reused for JAX (as the TODO suggests), the interface becomes more exposed and the silent misbehavior more likely.

build_tools/github_actions/configure_pytorch_test_matrix.py
```diff
@@ -0,0 +1,130 @@
...
+
+from github_actions.amdgpu_family_matrix import get_all_families_for_trigger_types
+from github_actions.github_actions_api import gha_set_output
+
+
+def split_families(value: str) -> list[str]:
+    return [f.strip() for f in value.split(";") if f.strip()]
+
+
+def find_amdgpu_family(*, requested_family: str, platform: str) -> tuple[str, str, str]:
```
marbre: My review agent thinks this should use a dataclass instead of a 3-tuple per the project's style guidelines as it isn't obvious that the third element being a runner label from the tuple position.

marbre (approved):
Just minor points you can, bust must not, address before landing.

build_tools/github_actions/configure_pytorch_test_matrix.py
ScottTodd: For all CI workflow jobs, we currently use the GPU families defined in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py, e.g. `gfx110x` and _not_ individual targets like `gfx1100`. I still want to flatten that all down - jobs can operate on individual targets (or flat lists), without any grouping of multiple targets into "families". We _could_ keep families as a user-facing concept where convenient, but they are adding lots of infrastructure complexity.

ScottTodd: Whoops, this code was trying to be too clever and inspect `fetch_targets`. I'll just remove all that for now.

ScottTodd: Sorry about that, pushed some changes to simplify. Now it just expects the GPU family (same as other workflows) without trying to overly normalize based on "fetch targets" too.

build_tools/github_actions/configure_pytorch_test_matrix.py
ScottTodd: Simplified + added the check for sentinels appearing in the list. There are still sharp edges with workflow_dispatch if you provide targets that don't have a family in amdgpu_family_matrix, as in other workflows. I'd like to clean all that up together later (as part of completing the migration to multi-arch)

build_tools/github_actions/configure_pytorch_test_matrix.py
ScottTodd: Simplified to a more direct family -> test_runs_on lookup, so no need for tuples or a dataclass.

ScottTodd:
Triggered another test workflow run after latest changes: https://github.com/ROCm/TheRock/actions/runs/27236351337. So far so good.

PR #5683, opened 2026-06-08
feat: add LLVM compile parallelism controls to prevent OOM on consumer hardware

GrecAndrei:
Expose `LLVM_PARALLEL_COMPILE_JOBS` and `LLVM_RAM_PER_COMPILE_JOB` at the top-level CMake and forward them into the LLVM sub-build.

Currently only `LLVM_PARALLEL_LINK_JOBS` and `FLANG_PARALLEL_COMPILE_JOBS` are exposed, but LLVM's compile phase is equally memory-intensive. On systems with many cores relative to RAM (e.g. 20 threads / 32 GB), the default unlimited parallelism causes OOM kills because each LLVM compile unit can use 2-6 GB RSS.

Changes:
- `CMakeLists.txt`: add cache variables for `LLVM_PARALLEL_COMPILE_JOBS` and `LLVM_RAM_PER_COMPILE_JOB`
- `compiler/CMakeLists.txt`: forward both through `_extra_llvm_cmake_args` into the LLVM sub-build

Usage:
```bash
cmake -B build -DLLVM_PARALLEL_COMPILE_JOBS=2 -DLLVM_PARALLEL_LINK_JOBS=1
# or auto-compute from RAM:
cmake -B build -DLLVM_RAM_PER_COMPILE_JOB=4096
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,8 @@ set(THEROCK_SANITIZER "" CACHE STRING "Enable project wide sanitizer build ('ASA
 
 # Options to control the concurrency for building LLVM.
 set(FLANG_PARALLEL_COMPILE_JOBS "" CACHE STRING "Limit parallel compile jobs for Flang (reduces memory usage on systems with many cores)")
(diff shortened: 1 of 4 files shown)


GrecAndrei:
Closing this premature PR. Will file an issue for discussion first per CONTRIBUTING.md.

PR #5688, opened 2026-06-08
Disable hipDNN Python bindings test configuration

### LINKED ISSUE #5678: rocm-sdk test fails on Linux: hipDNN frontend Python ext ships in rocm-sdk-devel with a dead CI RPATH and is ABI-locked to cp312
## Description

`rocm-sdk test` fails one check (`testSharedLibrariesLoad`, `ROCmDevelTest`) on Linux because a prebuilt hipDNN frontend Python extension is shipped inside the **`rocm-sdk-devel`** wheel and cannot be loaded:

```
OSError: .../_rocm_sdk_devel/share/hipdnn/python/hipdnn_frontend/hipdnn_frontend_python.cpython-312-x86_64-linux-gnu.so:
        undefined symbol: hipdnnBackendGetSerializedJsonGraph_ext
```

This appears to be a packaging leak rather than a code defect: per #5429, the hipDNN bindings' Python tree was intended to ship *"only via wheel, not the shared-library tarball"* and was excluded from the **lib** component. It is indeed absent from `rocm-sdk-libraries`, but the `share/hipdnn/{python,tests}` tree — including the prebuilt `.so` — still ships in **`rocm-sdk-devel`**, where `rocm-sdk test` discovers and tries to load it.

## Environment

- Linux, Python **3.13**, installed from the nightly index:
  `pip install --pre --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/ "rocm[libraries,devel]"`
- ROCm nightly `7.14.0a20260607` (`rocm-sdk-devel`), gfx110X-all.

## Reproduce

```bash
python3.13 -m venv v && . v/bin/activate
pip install --pre --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/ "rocm[libraries,devel]"
rocm-sdk test     # -> Ran 26 tests ... FAILED (errors=1), the hipdnn_frontend_python load
```

## Root cause — two compounding factors

**(a) Dead CI RPATH.** The shipped `.so` declares `NEEDED: libhipdnn_backend.so` but its RPATH is a GitHub Actions build path that exists only on the runner:

```
$ readelf -d hipdnn_frontend_python.cpython-312-*.so | grep -E 'RPATH|NEEDED'
 (RPATH)   Library rpath: [/home/runner/_work/TheRock/TheRock/output/build/core/clr/dist/lib]
 (NEEDED)  libhipdnn_backend.so
```

So on any non-CI machine the backend lib is unfindable and its symbols (e.g. `hipdnnBackendGetSerializedJsonGraph_ext`) resolve as undefined. Proof it's a *findability* issue, not a missing symbol:

```bash
# bare load: undefined symbol hipdnnBackendGetSerializedJsonGraph_ext
# with backend dir on the path: that symbol resolves
LD_LIBRARY_PATH=.../_rocm_sdk_devel/lib python -c "import ctypes; ctypes.CDLL('.../hipdnn_frontend_python.cpython-312-*.so')"
```

**(b) cp312 ABI lock.** With (a) worked around, the load then fails on `_PyObject_LookupAttr` — a private CPython symbol present in 3.12 and **removed in 3.13**. The extension is a nanobind `.so` tagged `cpython-312` and is the *only* Python version shipped, whereas sibling extensions ship for all supported versions:

```
hipdnn_frontend_python.cpython-312-...so          # cp312 ONLY
libpyroctx.cpython-{310,311,312,313}-...so         # all four
libpyrocpd.cpython-{310,311,312,313}-...so         # all four
```

So even after the RPATH is fixed, the shipped artifact can only load on Python 3.12.

## Suggested direction

The clean fix is the one #5429 already intended: **don't ship `share/hipdnn/python/**` (the prebuilt ext + test suite) in `rocm-sdk-devel`** — extend the lib-component exclusion to the devel component (or exclude this tree from `rocm-sdk test`'s shared-library discovery). That removes both symptoms at once, since the artifact isn't meant to be a user-loadable library. (If the ext *is* meant to ship in devel for some consumer, then it needs a relocatable `$ORIGIN`-relative RPATH and an all-Python build matrix like pyroctx/pyrocpd — but the simpler exclusion seems aligned with #5429.)

## Related

- #5429 (merged) — introduced the bindings + the lib-component exclusion this extends.
- #5652 (open) — Windows DLL-resolution fix for the same wheel test.
- #4853 (open, Windows) — `testSharedLibrariesLoad` failing on ~40 plugin DLLs; same *test*, different platform/component/root-cause.

## Broader context (optional, not the core ask)

The dead `/home/runner` RPATH isn't unique to hipDNN: **40 of 404** shipped `.so` in the `rocm-sdk-*` wheels carry a `/home/runner/...` RPATH (incl. `librocrand.so.1`, `libmiopen_plugin.so`, `libhipblaslt_plugin.so`, the hipDNN test plugins). Most still resolve their deps via co-location, so only the standalone-`dlopen`ed hipDNN frontend actually breaks — but it suggests CI build paths are leaking into shipped binaries instead of `$ORIGIN`-relative RPATHs, which is the same root-cause family as #4853.

tvy-amd:
## Summary

Two related hipDNN Python-bindings CI/packaging changes that unblock CI:

1. **Disable the CI test job (tracks #5689).** Comment out the `hipdnn_python_bindings` entry in the CI test matrix (`build_tools/github_actions/fetch_test_configurations.py`) so the hipDNN Python bindings wheel build + install + pytest job is no longer scheduled on Linux or Windows. The Windows job is broken because loading the bindings needs `AddDllDirectory` / `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` (Linux-only `LD_LIBRARY_PATH` does not apply) — see #5689. All other hipDNN test configurations remain active.
2. **Stop shipping the bindings in `rocm-sdk-devel` (fixes #5678).** Exclude `share/hipdnn/python/**` from the `dev` component in `ml-libs/artifact-hipdnn.toml`. The prebuilt frontend extension was leaking into the devel wheel where `rocm-sdk test` (`testSharedLibrariesLoad`) discovers and `dlopen`s it and fails — the shipped `.so` carries a dead `/home/runner` CI RPATH (so `libhipdnn_backend.so` is unfindable) and is ABI-locked to cp312. This extends the lib-component exclusion from #5429 to the devel component, matching the original intent that the Python tree ship only via wheel, not the shared-library tarball.

## Issue Tracking

- Fixes #5678 — Linux: hipDNN frontend Python ext leaks into `rocm-sdk-devel` (dead CI RPATH + cp312 ABI lock).
- Tracks #5689 — Windows: hipDNN Python bindings test fails (`ModuleNotFoundError` / DLL load); needs `AddDllDirectory`. Re-enable the matrix entry once fixed.
- Extends #5429 — original lib-component exclusion of the hipDNN Python tree.
- Jira ALMIOPEN-1943 — "Add tests for python bindings with hipDNN" (the test work being temporarily disabled in CI).
- Jira ALMIOPEN-1869 — "Build and install python bindings with hipDNN" (the bindings build this packaging fix relates to).

## Decision 

We were thinking of reverting the commit but we have the bindings .so and .dll being packaged in rocm-libraries right now so that would cause Issue https://github.com/ROCm/TheRock/issues/5678 to not be addressed. To revert would require a multiple reverts to land which would delay the fix so we opted for just commenting out the tests and adding to the artifacts.toml file

## Re-enable Checklist (future PR)

- [ ] #5689 resolved (Windows `AddDllDirectory` loader fix) → uncomment `hipdnn_python_bindings` in `fetch_test_configurations.py`.
- [ ] If the prebuilt ext is ever meant to ship in `rocm-sdk-devel`, give it a relocatable `$ORIGIN` RPATH + an all-Python build matrix (like pyroctx/pyrocpd) before removing the `artifact-hipdnn.toml` exclusion.

## Risk Assessment

Low risk. CI-config + packaging-manifest change only; no product code, build wiring, or public API is affected.
- Matrix change removes a single test job; worst case is reduced CI coverage for the hipDNN Python bindings until re-enabled (tracked by #5689).
- TOML change drops a tree (prebuilt extension + Python test suite) that has no valid runtime consumer in `rocm-sdk-devel`; the wheel-distributed bindings are unaffected.

## Testing Summary

- Matrix change comments out one dict entry; no behavior added. Verified the active matrix no longer emits a `hipdnn_python_bindings` job (Linux or Windows), so the broken Windows step is no longer scheduled.
- TOML exclusion verified against the #5678 repro: `share/hipdnn/python/**` no longer present in the `dev` component, so `rocm-sdk test` no longer discovers the unloadable extension. Verified end-to-end via a `multi_arch_ci.yml` workflow_dispatch building gfx94X-dcgpu wheels + running the `rocm-sdk` wheel test.

## Testing Checklist

- [ ] Custom `multi_arch_ci.yml` dispatch (gfx94X-dcgpu) — `rocm-sdk test` `testSharedLibrariesLoad` passes

## Technical Changes

- Comments out the `hipdnn_python_bindings` test-matrix entry in `build_tools/github_actions/fetch_test_configurations.py`, removing the job from CI scheduling while keeping it easy to restore (with a `# !! DISABLED because of #5689` note).
- Adds `exclude = ["share/hipdnn/python/**"]` to `[components.dev."ml-libs/hipDNN/stage"]` in `ml-libs/artifact-hipdnn.toml`, removing the prebuilt frontend extension and Python test suite from the `rocm-sdk-devel` artifact (fixes #5678).

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -484,6 +484,9 @@ test_matrix = {
             "windows": 1,
         },
     },
+    # !! DISABLED because of https://github.com/ROCm/TheRock/issues/5689
+    # !! Windows loading of the python bindings require special LOAD_LIBRARY_SEARCH_DEFAULT_DIRS 
+    # !! We need AddDllDirectory. Commenting out to unblock CI issues.
     # hipDNN Python bindings wheel build + install + pytest
     # "hipdnn_python_bindings": {
     #     "job_name": "hipdnn_python_bindings",
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -485,7 +485,7 @@ test_matrix = {
         },
     },
     # !! DISABLED because of https://github.com/ROCm/TheRock/issues/5689
-    # !! Windows loading of the python bindings require special LOAD_LIBRARY_SEARCH_DEFAULT_DIRS 
+    # !! Windows loading of the python bindings require special LOAD_LIBRARY_SEARCH_DEFAULT_DIRS
     # !! We need AddDllDirectory. Commenting out to unblock CI issues.
     # hipDNN Python bindings wheel build + install + pytest
     # "hipdnn_python_bindings": {
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -485,17 +485,17 @@ test_matrix = {
         },
     },
     # hipDNN Python bindings wheel build + install + pytest
-    "hipdnn_python_bindings": {
-        "job_name": "hipdnn_python_bindings",
-        "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
-        "timeout_minutes": 30,
-        "test_script": f"python {_get_script_path('test_hipdnn_frontend_python.py')}",
-        "platform": ["linux", "windows"],
-        "total_shards_dict": {
-            "linux": 1,
-            "windows": 1,
-        },
-    },
+    # "hipdnn_python_bindings": {
+    #     "job_name": "hipdnn_python_bindings",
+    #     "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
+    #     "timeout_minutes": 30,
+    #     "test_script": f"python {_get_script_path('test_hipdnn_frontend_python.py')}",
+    #     "platform": ["linux", "windows"],
+    #     "total_shards_dict": {
+    #         "linux": 1,
+    #         "windows": 1,
+    #     },
+    # },
     # hipDNN integration tests (unit tests for the integration test harness)
     "hipdnn-integration-tests": {
         "job_name": "hipdnn-integration-tests",
--- a/ml-libs/artifact-hipdnn.toml
+++ b/ml-libs/artifact-hipdnn.toml
@@ -11,6 +11,9 @@ include = [
 ]
 [components.dbg."ml-libs/hipDNN/stage"]
 [components.dev."ml-libs/hipDNN/stage"]
+exclude = [
+  "share/hipdnn/python/**",
+]
 [components.doc."ml-libs/hipDNN/stage"]
 [components.test."ml-libs/hipDNN/stage"]
 include = [


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -485,17 +485,17 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+    #     "job_name": "hipdnn_python_bindings",
+    #     "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
+    #     "timeout_minutes": 30,
+    #     "test_script": f"python {_get_script_path('test_hipdnn_frontend_python.py')}",
+    #     "platform": ["linux", "windows"],
+    #     "total_shards_dict": {
+    #         "linux": 1,
+    #         "windows": 1,
+    #     },
+    # },
```
tvy-amd, line 501: Addresses 
Issues https://github.com/ROCm/TheRock/issues/5689

ml-libs/artifact-hipdnn.toml
```diff
@@ -11,6 +11,9 @@ include = [
 ]
 [components.dbg."ml-libs/hipDNN/stage"]
 [components.dev."ml-libs/hipDNN/stage"]
+exclude = [
+  "share/hipdnn/python/**",
+]
```
tvy-amd, line 16: Addresses:
Issue https://github.com/ROCm/TheRock/issues/5678

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -485,17 +485,17 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
-        "fetch_artifact_args": "--blas --miopen --hipdnn --miopenprovider --tests",
-        "timeout_minutes": 30,
-        "test_script": f"python {_get_script_path('test_hipdnn_frontend_python.py')}",
-        "platform": ["linux", "windows"],
-        "total_shards_dict": {
-            "linux": 1,
-            "windows": 1,
-        },
-    },
+    # "hipdnn_python_bindings": {
```
BrianHarrisonAMD, line 491: Can we add a comment linking to the issue here for tracking?

BrianHarrisonAMD (commented):
PR summary should be updated to also include details about the artifact-hipdnn.toml change for resolving #5678.

BrianHarrisonAMD (approved):
LGTM for unbreaking CI runs, but let's add the extra issue tracking details before merging.

We should verify using a custom workflow trigger to build the wheels and verify that the `rocm-sdk test` no longer has breaks. 

The testing workaround can be verified from regular CI not having the broken testing step for Windows.

build_tools/github_actions/fetch_test_configurations.py
tvy-amd: done

geomin12 (commented):
should we revert instead of commenting out?

tvy-amd:
We have the bindings .so and .dll being packaged in rocm-libraries right now so that would cause Issue https://github.com/ROCm/TheRock/issues/5678 to not be addressed.

We could do a multi-PR revert but that might take too long to address the issues with the bumps and PRs.

geomin12 (approved):

ml-libs/artifact-hipdnn.toml
ScottTodd, line 16: Could add a comment here (what the files are, why they shouldn't be included in this component, link to the issue for even more context as needed, etc.)

ScottTodd (approved):

ml-libs/artifact-hipdnn.toml
tvy-amd: I can do it in another PR - I want to get this as soon as possible

PR #5692, opened 2026-06-08
[ci] updating `test_filtering.md` to have clearer execution time expectations

geomin12:
Received some confusion for test filtering document. Updating execution time expectation to be clearer and more upfront


dileepr1 (approved):
Approved

PR #5352, opened 2026-05-20
Support testing multi-arch ROCm wheels using PEP 503 `--index-url` inputs

ScottTodd:
## Motivation

Expanding our ability to test for https://github.com/ROCm/TheRock/issues/5347.

The `.github/workflows/test_rocm_wheels.yml` workflow was not compatible with `--index-url https://rocm.nightlies.amd.com/whl-multi-arch/`, only CI uploads like `--find-links https://therock-ci-artifacts.s3.amazonaws.com/26115600835-linux/python/index.html`. Now it supports both modes.

## Technical Details

The main fix here is to omit `--index-subdir=${{ inputs.amdgpu_family }}` when `${{ inputs.kpack_split == 'true' }}`, similar to https://github.com/ROCm/TheRock/blob/48e5305a977d3b598af3113ff259f1f088e82248/.github/workflows/test_pytorch_wheels.yml#L161-L175

## Test Plan

* Manually triggered https://github.com/ROCm/TheRock/actions/runs/26132646536/job/76860876521
    ```
      kpack_split: true
      package_index_url_base: https://rocm.nightlies.amd.com/whl-multi-arch/
      package_find_links_url:
      amdgpu_family: gfx950-dcgpu
      rocm_version: 7.14.0a20260519
      test_runs_on: linux-gfx950-1gpu-ccs-ossci-rocm
    ```
* Watch CI on this PR to check that `--find-links` path still works
  * Linux gfx942: https://github.com/ROCm/TheRock/actions/runs/26133134983/job/76877270781?pr=5352 success,
      ```
      Run pip freeze
      rocm==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
      rocm-sdk-core==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
      rocm-sdk-device-gfx942==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
      rocm-sdk-libraries==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
      ```
  * Windows gfx110X: https://github.com/ROCm/TheRock/actions/runs/26133134983/job/76879169824?pr=5352 known timeout,
      ```
      Run pip freeze
      rocm==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
      rocm-sdk-core==7.14.0.dev0+03fb5824e759640bf4f8e9b4e3c842561d1befd7
[...]

PR #5693, opened 2026-06-08
ci: support installing ROCm from a PEP 503 index for PyTorch wheel builds

ethanwee1:
## What

Add an optional `rocm_package_index_url` input to the multi-arch PyTorch wheel
build and release workflows (and the CI variant), letting builds install ROCm
from a PEP 503 simple index via pip `--index-url`, as an alternative to the
existing flat find-links page (`rocm_package_find_links_url`).

Both inputs are now optional; exactly one must be supplied. When the index URL
is set it takes precedence; otherwise the find-links page is used; if neither
is set the build fails fast with a clear error.

## Why

The published ROCm indexes (e.g. `https://rocm.devreleases.amd.com/whl-multi-arch/`
and the nightlies equivalent) are PEP 503 simple indexes, which pip can only
consume with `--index-url`. The build workflows previously hardcoded
`--find-links`, which only works against a flat `index.html` that links wheel
files directly. Pointing `--find-links` at a simple index silently finds no
packages and falls back to PyPI, producing confusing
`Could not find a version that satisfies the requirement rocm==...` failures.
This change lets the published indexes be used directly.

## Files changed

- `multi_arch_build_portable_linux_pytorch_wheels.yml`
- `multi_arch_release_linux_pytorch_wheels.yml`
- `multi_arch_build_portable_linux_pytorch_wheels_ci.yml` (CI variant)

ISSUE ID : #3291


.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
```diff
@@ -242,10 +250,23 @@ jobs:
...
+          # TheRock CI artifacts index.html) via --find-links.
+          rocm_source_flag=()
+          if [ -n "${{ inputs.rocm_package_index_url }}" ]; then
+            rocm_source_flag=(--index-url "${{ inputs.rocm_package_index_url }}")
+          elif [ -n "${{ inputs.rocm_package_find_links_url }}" ]; then
+            rocm_source_flag=(--find-links "${{ inputs.rocm_package_find_links_url }}")
+          else
+            echo "ERROR: set rocm_package_index_url or rocm_package_find_links_url" >&2
+            exit 1
+          fi
```
ScottTodd: No need for inline bash here (https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash), the script already supports both arguments and if one is omitted that should be fine. See https://github.com/ROCm/TheRock/blob/fc6a6fbc7d2f5f623739977af33c15bb045f6b0c/external-builds/pytorch/build_prod_wheels.py#L1272-L1276

Should look more like https://github.com/ROCm/TheRock/pull/5352 with
```diff
            python build_tools/setup_venv.py ${VENV_DIR} \
              --packages "${pkg}==${{ inputs.rocm_version }}" \
+             --index-url=${{ inputs.package_index_url_base }} \
+             --find-links=${{ inputs.package_find_links_url }} \
```

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ScottTodd, line 1: also update the Windows workflow files

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
```diff
@@ -43,9 +43,13 @@ on:
...
       rocm_package_find_links_url:
-        description: URL for pip --find-links to install ROCm packages
+        description: URL for pip --find-links to install ROCm packages (flat page; use this OR rocm_package_index_url)
         type: string
-        required: true
+        default: ""
+      rocm_package_index_url:
+        description: URL for pip --index-url to install ROCm packages (PEP 503 index, e.g. repo.amd.com; use this OR rocm_package_find_links_url)
+        type: string
+        default: ""
```
ScottTodd, line 60: Note that we also want to move away from `--find-links` altogether, but for that we'd need to change how the index pages are generated for our artifacts (and possibly also provide a `--extra-index-url` with dependencies in each bucket)

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ethanwee1: Done — removed the inline bash. Both `--index-url` and `--find-links` are now passed directly to build_prod_wheels.py (which ignores an empty value), as in #5352.

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ethanwee1: Done — added `rocm_package_index_url` to multi_arch_build_windows_pytorch_wheels.yml and multi_arch_release_windows_pytorch_wheels.yml (build call + passthrough).

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ethanwee1: Acknowledged — leaving the full move away from `--find-links` (index-page generation + per-bucket `--extra-index-url`) as a follow-up; this PR just adds the index-url option alongside it.

ScottTodd (approved):
I think this will be safe to merge without a corresponding change in https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml and https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_windows_pytorch_wheels.yml, because the workflow code that dispatches to these files omits the new optional input: https://github.com/ROCm/TheRock/blob/8484067b044134dc2a4b1f3fe45c415b2a631a17/.github/workflows/multi_arch_release_linux.yml#L238-L245

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels_ci.yml
ScottTodd, line 1: Missed one workflow: https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_pytorch_wheels_ci.yml

(should keep them in sync)

ethanwee1: Good catch — done in 419efebd. multi_arch_build_windows_pytorch_wheels_ci.yml now has the same rocm_package_index_url input and passes both flags, in sync with the other three.

Issue #4102, opened 2026-03-23
[Issue]: c10d


xalteropsx:
### Problem Description

hasattr(torch._C, "_c10d_init")   returning false 

Name: torch
Version: 2.12.0a0+rocm7.13.0a20260313

window

### Operating System

window

### CPU

ryzen 7

### GPU

rx 7900xtx

### ROCm Version

7.1

### Additional Information

torch.distributed failing all

astrelsky:
> Hi @xalteropsx,
> 
> Thanks for reporting this issue, we currently do not have support for `torch.distributed` on Windows, but this is on our roadmap, please see https://github.com/ROCm/TheRock/issues/1179 for updates.

It is practically trivial for pytorch 2.11, you just need libuv and remove the environment flag disabling it from the script used to build torch. For older pytorch you need to update gloo to what 2.11 has.

#3284


I have been manually building pytorch with it enabled for a while now to get around this problem.

tcgu-amd:
Hi @xalteropsx that's expected for now. Please see https://github.com/ROCm/TheRock/issues/1314. Thanks!

astrelsky:
> Hi @xalteropsx that's expected for now. Please see https://github.com/ROCm/TheRock/issues/1314. Thanks! 

@tcgu-amd may I get a reason for the apparent refusal for the 2 minute build of libuv and enabling the flag to build gloo on windows? It's a really simple fix locally (ci plumbing is more complicated and not something I'm good at or I'd do it myself). It seems to be problematic enough and a commonly reported issue that you would expect it to be fixed relatively quickly when there is a known working solution.

xalteropsx:
amd lazy developer i hope they replace with active one since it was an year issue not month issue

astrelsky:
> amd lazy developer i hope they replace with active one since it was an year issue not month issue

I know it's hard, but try to be nice about it. Making the developers more frustrated will make it worse not better.

xalteropsx:
if developer take an year to resolve an small issue you say we make frustrated they are been paid by amd but not work as hard as other asian developer who do thing in few days or month not an year , i been regret but amd lack of skilled and lazy developer its a fact not lie many know about it already

astrelsky:
> if developer take an year to resolve an small issue you say we make frustrated they are been paid by amd but not work as hard as other asian developer who do thing in few days or month not an year , i been regret but amd lack of skilled and lazy developer its a fact not lie many know about it already 

[...]

Issue #2008, opened 2025-11-04
[Issue]: unsloth compatiblity issue with torch therock windows


lihaofd:
### Problem Description

780M iGPU on windows
python -m pip install   --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/   "rocm[libraries,devel]"
python -m pip install   --index-url https://rocm.nightlies.amd.com/v2/gfx110X-all/   --pre torch torchaudio torchvision

https://docs.unsloth.ai/get-started/install-and-update/amd
pip install --no-deps unsloth unsloth-zoo
pip install --no-deps git+https://github.com/unslothai/unsloth-zoo.git
pip install "unsloth[amd] @ git+https://github.com/unslothai/unsloth"

https://colab.research.google.com/github/unslothai/notebooks/blob/main/nb/GPT_OSS_MXFP4_(20B)-Inference.ipynb

from unsloth import FastLanguageModel 
always shows 
AttributeError: module 'torch.distributed' has no attribute 'is_initialized'

### Operating System

windows 11

### CPU

7840U

### GPU

780M

### ROCm Version

2.10.0a0+rocm7.10.0a20251031

astrelsky:
Same problem, or similar, trying to use amd quark.

andyluo7:
Can you try it on WSL?

lihaofd:
wsl have issues reported before - https://github.com/ROCm/TheRock/issues/690#issuecomment-2911051236
I can also reproduce 
WSL environment detected.
hsa api call failure at: /therock/src/rocm-systems/projects/rocminfo/rocminfo.cc:1324

adityas-amd:
Yes, this is the same underlying known limitation on Windows ROCm limitation as #[1179](https://github.com/orgs/ROCm/projects/99/views/3?sliceBy%5Bvalue%5D=adityas-amd&pane=issue&itemId=129652720&issue=ROCm%7CTheRock%7C1179). The current Windows ROCm (“TheRock”) PyTorch builds do not enable torch.distributed (Gloo is Linux-only in these builds, and RCCL is not supported on Windows yet), so common APIs like dist.is_initialized and dist.init_process_group are missing. 
If you want to test Unsloth with AMD acceleration right now, please use Linux + ROCm where Unsloth is tested and torch.distributed works.
On Windows, Unsloth support is limited. Please wait until RCCL (and/or a Windows-compatible distributed backend) is supported in the TheRock builds.

[...]

Issue #1179, opened 2025-08-01
[Issue]: AttributeError: module 'torch.distributed' has no attribute 'init_process_group'


AznamirWoW:
### Problem Description

installed the latest [torch-2.9.0a0+rocm7.0.0rc20250728-cp311-cp311-win_amd64.whl](https://d25kgig7rdsyks.cloudfront.net/v2/gfx110X-dgpu/torch-2.9.0a0%2Brocm7.0.0rc20250728-cp311-cp311-win_amd64.whl)

ran a training script that had

`    dist.init_process_group(
        backend="gloo" if sys.platform == "win32" or device.type != "cuda" else "nccl",
        init_method="env://",
        world_size=n_gpus if device.type == "cuda" else 1,
        rank=rank if device.type == "cuda" else 0,
    )`

### Operating System

Windows 11

### CPU

7600x

### GPU

7900 GRE

### ROCm Version

7.0.0

ScottTodd:
I would not expect distributed APIs to work until we somehow enable GLOO (the collective communications library used by PyTorch: https://github.com/pytorch/gloo) on Windows: https://github.com/ROCm/TheRock/blob/7615bf9fe01815bf0bc3591f9eb859c169f5e9e8/external-builds/pytorch/build_prod_wheels.py#L371-L373

For that, we would need RCCL (the ROCm Communication Collectives Library), which is currently unsupported on Windows: https://github.com/ROCm/TheRock/blob/main/docs/development/windows_support.md#supported-subprojects

Maybe we can improve the error message there and note this as a known feature gap until then.

subunder:
@AznamirWoW Could you try it again with the `2.10.0a0+rocm7.0.0rc20250917` torch wheel?

I just closed my [issue](https://github.com/ROCm/TheRock/issues/1314) because `import torch.distributed` started working as it should.

Idk about the backends, so whether it’s gloo, nccl, or any other one…

I have this small script:
```
#!/usr/bin/env python3
import torch.distributed as dist

def main():
    print("Available distributed backends:")
    for backend in ["gloo", "nccl", "mpi", "ucc", "hccl", "rccl"]:
        try:
            if dist.is_backend_available(backend):
                print(f"  - {backend}")
        except Exception:
            # Some backends may not even be recognized; skip them
            pass

if __name__ == "__main__":
    main()

```
…but it doesn’t print anything.

But for example:
`print(dist.Backend)` 
prints:
[...]

PR #969, opened 2025-07-03
Enable GLOO in build_prod_wheels.py

erman-gurses:

PR #5694, opened 2026-06-08
Enable GLOO backend for torch.distributed on Windows

### LINKED ISSUE #3284: [Feature]: Add libuv for gloo and torch.distributed on windows
### Suggestion Description

Add libuv to build pytorch with distributed support on windows. It takes seconds to build and there is even a vcpkg you can use. It is MIT licensed. This has been a known issue for months where the fix is simply adding an MIT licensed third party library. #1179 #2008 

https://github.com/libuv/libuv

### Operating System

Windows

### GPU

_No response_

### ROCm Component

_No response_

harkgill-amd:
## Motivation

Resolves https://github.com/ROCm/TheRock/issues/3284

Our Windows PyTorch wheels have been shipping without torch.distributed support causing third party apps to fail with errors such as 
```
File "D:\languages\python\Lib\site-packages\torch\distributed\distributed_c10d.py", line 23, in
from torch._C._distributed_c10d import (
ModuleNotFoundError: No module named 'torch._C._distributed_c10d'; 'torch._C' is not a package
```
Examples

- https://github.com/ROCm/TheRock/issues/1179
- https://github.com/ROCm/TheRock/issues/4102
- https://github.com/ROCm/TheRock/issues/2008

These issues are seen on single GPU systems where there's no actual distributed work being done. We've been mitigating these issues by patching the upstream libraries to have fallback support when torch.distributed isn't available but the longterm solution is to enable the GLOO backend.

## Technical Details

In `build_windows_pytorch_wheels.yml`
- Add vcpkg libuv install step and set libuv_ROOT so PyTorch's CMake finds it

In `external-builds/pytorch/build_prod_wheels.py`
- Set USE_GLOO=ON unconditionally (was Linux-only)
- Add copy_libuv_to_torch_lib() to bundle uv.dll into torch/lib/, following the same pattern as copy_msvc_libomp_to_torch_lib()

## Test Plan

- Run `Build Windows Pytorch Wheels` workflow to generate wheels
- Test that both the GLOO backend and torch.distributed are available.

## Test Result
Using the wheels built from https://github.com/ROCm/TheRock/actions/runs/27149444602 on gfx1151,

```
python -c "import torch.distributed; print('torch.distributed available:', torch.distributed.is_available()); print('GLOO available:', torch.distributed.is_gloo_available())"
torch.distributed available: True
GLOO available: True
```

## Submission Checklist

### DIFF
--- a/.github/workflows/build_windows_pytorch_wheels.yml
+++ b/.github/workflows/build_windows_pytorch_wheels.yml
@@ -187,6 +187,11 @@ jobs:
           choco install --no-progress -y awscli
           echo "$PATH;C:\Program Files\Amazon\AWSCLIV2" >> $GITHUB_PATH
 
+      - name: Install libuv for GLOO
+        run: |
+          vcpkg install libuv:x64-windows
+          echo "libuv_ROOT=C:/vcpkg/installed/x64-windows" >> $GITHUB_ENV
+
       - name: Install sccache
         if: ${{ inputs.cache_type == 'sccache' }}
         run: |
--- a/.github/workflows/build_windows_pytorch_wheels.yml
+++ b/.github/workflows/build_windows_pytorch_wheels.yml
@@ -187,10 +187,13 @@ jobs:
           choco install --no-progress -y awscli
           echo "$PATH;C:\Program Files\Amazon\AWSCLIV2" >> $GITHUB_PATH
 
-      - name: Install libuv for GLOO
+      - name: Build libuv from source for GLOO
         run: |
-          vcpkg install libuv:x64-windows
-          echo "libuv_ROOT=C:/vcpkg/installed/x64-windows" >> $GITHUB_ENV
+          git clone https://github.com/libuv/libuv.git -b v1.47.0 --depth 1
+          cmake -S libuv -B libuv/build -DBUILD_TESTING=OFF
+          cmake --build libuv/build --config Release
+          cmake --install libuv/build --prefix "${{ github.workspace }}/libuv-install"
+          echo "libuv_ROOT=${{ github.workspace }}/libuv-install" >> $GITHUB_ENV
 
       - name: Install sccache
         if: ${{ inputs.cache_type == 'sccache' }}
--- a/.github/workflows/build_windows_pytorch_wheels.yml
+++ b/.github/workflows/build_windows_pytorch_wheels.yml
@@ -187,6 +187,9 @@ jobs:
           choco install --no-progress -y awscli
           echo "$PATH;C:\Program Files\Amazon\AWSCLIV2" >> $GITHUB_PATH
 
+      # Build a fixed version of libuv from source, no matter the pytorch ref.
+      # gloo requires >= 1.26; v1.47.0 matches upstream's source build.
+      # https://github.com/pytorch/pytorch/blob/main/.ci/pytorch/windows/arm64/bootstrap_libuv.bat
       - name: Build libuv from source for GLOO
         run: |
           git clone https://github.com/libuv/libuv.git -b v1.47.0 --depth 1
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -475,9 +475,7 @@ def _setup_common_build_env(
         "USE_KINETO": os.environ.get("USE_KINETO", "ON" if not is_windows else "OFF"),
     }
 
-    # GLOO enabled for only Linux
-    if not is_windows:
-        env["USE_GLOO"] = "ON"
+    env["USE_GLOO"] = "ON"
 
     # At checkout, we compute some additional env vars that influence the way that
     # the wheel is named/versioned.
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -925,6 +925,18 @@ def copy_msvc_libomp_to_torch_lib(pytorch_dir: Path):
     shutil.copy2(omp_path, target_lib)
 
 
+def copy_libuv_to_torch_lib(pytorch_dir: Path):
+    libuv_root = os.environ.get("libuv_ROOT", "")
+    if not libuv_root:
+        return
+    uv_dll = Path(libuv_root) / "bin" / "uv.dll"
+    if not uv_dll.exists():
+        raise RuntimeError(f"Did not find uv.dll at '{uv_dll}'")
+    target_lib = pytorch_dir / "torch" / "lib"
+    print(f"Copying libuv from '{uv_dll}' to '{target_lib}'")
+    shutil.copy2(uv_dll, target_lib)
+
+
 def do_build_pytorch(
     args: argparse.Namespace,
     pytorch_dir: Path,
@@ -1054,6 +1066,7 @@ def do_build_pytorch(
     # Windows-specific settings.
     if is_windows:
         copy_msvc_libomp_to_torch_lib(pytorch_dir)
+        copy_libuv_to_torch_lib(pytorch_dir)
 
         use_flash_attention = (
             "1"


astrelsky:
You're going to hit a link error with pytorch nightly that it can't find `gloo_hip.lib` even though gloo was built successfully and is at `build/lib/gloo.lib`. I never tracked down the change that caused it and instead took the lazy approach of copying it to `build/gloo_hip.lib`.

harkgill-amd:
> You're going to hit a link error with pytorch nightly that it can't find `gloo_hip.lib` even though gloo was built successfully and is at `build/lib/gloo.lib`. I never tracked down the change that caused it and instead took the lazy approach of copying it to `build/gloo_hip.lib`.

Thanks for the heads up. I've kicked off https://github.com/ROCm/TheRock/actions/runs/27155403450 which'll test this change against the nightly branch - let's see if it hits the same `gloo_hip.lib` error and go from there.

EDIT: Ya, ended up hitting the error you mentioned. Looks like there was a change in how the GLOO setup is done between `release/2.11` and `nightly` so we end up hitting the append here, https://github.com/pytorch/pytorch/blob/nightly/cmake/Dependencies.cmake#L1339. A simple guard to skip this for Windows should be sufficient as we're ok with CPU only gloo.lib.

harkgill-amd:
https://github.com/pytorch/pytorch/pull/186650 will resolve the build failure with nightly branch. 

Updated Build Windows PyTorch Wheels workflow to allow testing against my pytorch fork with the above change - build is passing now that we're guarding against gloo_hip on Windows. https://github.com/ROCm/TheRock/actions/runs/27163904722/job/80186050020

astrelsky:
> [pytorch/pytorch#186650](https://github.com/pytorch/pytorch/pull/186650) will resolve the build failure with nightly branch.
> 
> Updated Build Windows PyTorch Wheels workflow to allow testing against my pytorch fork with the above change - build is passing now that we're guarding against gloo_hip on Windows. https://github.com/ROCm/TheRock/actions/runs/27163904722/job/80186050020

Huh, that doesn't seem right but on closer inspection it appears there is a gloo and gloo_hip? Any idea what the difference is? I'm wondering why my builds didn't fail due to symbol collisions or other weird stuff if they're supposed to be different.

harkgill-amd:
Ok so with `release/2.11`, we were actually getting a working gloo_hip.lib so even though gloo.lib is enough to resolve the `torch.distributed` errors, there shouldn't be any reason why we can't get the former working with the `nightly`.

The difference is that older releases (2.11.etc) rely on `FindHIP.cmake` which works correctly with how the gloo submodule is currently setup to locate HIP.

`nightly` strips away `FindHIP.cmake` in favour of native CMake HIP language support but the gloo submodule is lagging behind at [3135b0b](https://github.com/pytorch/gloo/tree/3135b0b41b67dde590eef0938a0bf3d6238df5f7) (still FindHIP.cmake dependent). [bcd1672](https://github.com/pytorch/gloo/commit/bcd1672ee07538123ea8f4fac76832efc58fb8ef) updates gloo's CMake to be aligned with what nightly expects - bumping the upstream submodule to include this commit should be all we need.

.github/workflows/build_windows_pytorch_wheels.yml
```diff
@@ -187,6 +187,11 @@ jobs:
           choco install --no-progress -y awscli
           echo "$PATH;C:\Program Files\Amazon\AWSCLIV2" >> $GITHUB_PATH
 
+      - name: Install libuv for GLOO
+        run: |
+          vcpkg install libuv:x64-windows
+          echo "libuv_ROOT=C:/vcpkg/installed/x64-windows" >> $GITHUB_ENV
+
```
astrelsky: libuv only takes a few seconds to build from source. Additionally, if I remember correctly, there is already a libuv submodule in either the pytorch third_part directory or in one of those submodules. I think the cleanest way to ensure it also works when building locally is to have it libuv built from source.

ScottTodd: Agreed. I'd prefer to reuse what sources already exist and not add a dependency on a separate tool install (vcpkg) for developers that want to reproduce what our CI does. This also risks installing a different version of libuv from machine to machine, workflow run to workflow run.

Ideally pytorch itself would do the source build like with its other dependencies/submodules. Is it not doing that? What do upstream torch releases do?

harkgill-amd: So upstream pytorch goes about installing libuv in a couple different ways,

- Prebuilt libuv tarball for CI builds https://github.com/pytorch/pytorch/blob/main/.ci/pytorch/windows/setup_build.bat#L24-L28
- Build from source for ARM https://github.com/pytorch/pytorch/blob/main/.ci/pytorch/windows/arm64/bootstrap_libuv.bat
- `conda-forge` package suggested for developers missing `libuv` during build https://github.com/pytorch/pytorch/blob/main/CMakeLists.txt#L467-L485

Build from source makes sense. Testing out the [build from source](https://github.com/ROCm/TheRock/commit/ec5e3f0a8a1bf16dada4181860e42236b18ae4a3) step [here](https://github.com/ROCm/TheRock/actions/runs/27283909159/job/80585707241) - if it passes, I'll add it into this PR.

harkgill-amd: Build passed which also confirms that gloo submodule bump in torch fixed the nightly branch. Switched to build from source.

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -478,9 +475,7 @@ def _setup_common_build_env(
         "USE_KINETO": os.environ.get("USE_KINETO", "ON" if not is_windows else "OFF"),
     }
 
-    # GLOO enabled for only Linux
-    if not is_windows:
-        env["USE_GLOO"] = "ON"
+    env["USE_GLOO"] = "ON"
```
ScottTodd, line 478: Should we make this conditional on the `libuv_ROOT` environment variable being set too? What happens if I try to build locally without first building/installing libuv?

Could be worth a note in https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#prerequisites-and-setup too

If making it conditional, I'd
* Pull the environment variable and DLL check below up into a `is_libuv_available()` style function that can be called here
* Add some logging here for both cases: enabled since the environment variable is set, disabled since the environment variable is not set

.github/workflows/build_windows_pytorch_wheels.yml
```diff
@@ -187,6 +187,14 @@ jobs:
           choco install --no-progress -y awscli
           echo "$PATH;C:\Program Files\Amazon\AWSCLIV2" >> $GITHUB_PATH
 
+      - name: Build libuv from source for GLOO
+        run: |
+          git clone https://github.com/libuv/libuv.git -b v1.47.0 --depth 1
+          cmake -S libuv -B libuv/build -DBUILD_TESTING=OFF
+          cmake --build libuv/build --config Release
+          cmake --install libuv/build --prefix "${{ github.workspace }}/libuv-install"
+          echo "libuv_ROOT=${{ github.workspace }}/libuv-install" >> $GITHUB_ENV
```
ScottTodd, line 199: Can you add a comment here explaining why this clones and builds from a fixed source across _all_ pytorch versions? I could see a case for wanting to use libuv version associated with each pytorch version (e.g. using a git submodule instead of this separately tracked `v1.47.0` ref.

```suggestion
      # Build a fixed version of libuv from source, no matter the pytorch ref.
      # This similar to how upstream pytorch installs libuv v1.40.0 in their CI, see
      # https://github.com/pytorch/pytorch/blob/main/.ci/pytorch/windows/setup_build.bat
      - name: Build libuv from source for GLOO
        run: |
          git clone https://github.com/libuv/libuv.git -b v1.47.0 --depth 1
          cmake -S libuv -B libuv/build -DBUILD_TESTING=OFF
          cmake --build libuv/build --config Release
          cmake --install libuv/build --prefix "${{ github.workspace }}/libuv-install"
          echo "libuv_ROOT=${{ github.workspace }}/libuv-install" >> $GITHUB_ENV
```

ScottTodd (approved):

external-builds/pytorch/build_prod_wheels.py
harkgill-amd: PyTorch's CMake already handles the case where libuv isn't installed by disabling both `USE_DISTRIBUTED` and `USE_GLOO`. https://github.com/pytorch/pytorch/blob/main/CMakeLists.txt#L474-L480
```
    if(NOT libuv_tmp_LIBRARY)
      set(USE_DISTRIBUTED OFF)
      set(USE_GLOO OFF)
      message(
        WARNING
          "Libuv is not installed in current conda env. Set USE_DISTRIBUTED to OFF. "
          "Please run command 'conda install -c conda-forge libuv=1.51' to install libuv."
      )
```
We previously ran into this same warning which is what prompted disabling `USE_GLOO` on Windows in the first place https://github.com/ROCm/TheRock/pull/969#issuecomment-3062530251. The current implementation makes sense to me - always set to ON and let PyTorch disable it if dependency isn't present but can add in the conditional if the extra logging here would help.

ScottTodd: Ah, okay sure. Fine to keep as-is with the simpler code to maintain here then. SGTM.

harkgill-amd:
One last piece before this can go in, `release/2.9` and `release/2.10` have gloo pinned to a commit that is one prior to the commit that introduced Windows ROCm support ([1b4337a](https://github.com/pytorch/gloo/commit/1b4337a6fc264f5d4b61ff68bb4b00d22d30d6f3)) - this is currently causing [failures](https://github.com/ROCm/TheRock/actions/runs/27299990263). The possible options are to,

- Bump both PyTorch branches gloo commit to 1b4337a (just one commit up)
- Disable GLOO for 2.9 and 2.10 with a guard in `build_prod_wheels.py`

I'm testing out the former to see if builds pass with just the submodule bump and will check in the PyTorch team.

ScottTodd:
SGTM. I'd prefer to add support uniformly across all supported versions (so yes, cherry-pick what is needed onto our release branches if otherwise low risk). Short of that, we can add the workaround to our build script as needed.

harkgill-amd:
Builds with GLOO enabled are now passing on all branches

- nightly https://github.com/ROCm/TheRock/actions/runs/27643154900
- release/2.12 https://github.com/ROCm/TheRock/actions/runs/27299841358
- release/2.11 https://github.com/ROCm/TheRock/actions/runs/27149444602
- release/2.10 https://github.com/ROCm/TheRock/actions/runs/27622038723
- release/2.9 https://github.com/ROCm/TheRock/actions/runs/27632682115

.github/workflows/build_windows_pytorch_wheels.yml
```diff
@@ -187,6 +187,17 @@ jobs:
...
+      # Build a fixed version of libuv from source, no matter the pytorch ref.
+      # gloo requires >= 1.26; v1.47.0 matches upstream's source build.
+      # https://github.com/pytorch/pytorch/blob/main/.ci/pytorch/windows/arm64/bootstrap_libuv.bat
+      - name: Build libuv from source for GLOO
+        run: |
+          git clone https://github.com/libuv/libuv.git -b v1.47.0 --depth 1
+          cmake -S libuv -B libuv/build -DBUILD_TESTING=OFF
+          cmake --build libuv/build --config Release
+          cmake --install libuv/build --prefix "${{ github.workspace }}/libuv-install"
+          echo "libuv_ROOT=${{ github.workspace }}/libuv-install" >> $GITHUB_ENV
```
ScottTodd, line 199: Oops, this workflow is deprecated and is not currently used anywhere. I just sent https://github.com/ROCm/TheRock/pull/5892 to delete it. This code needs to go in these files instead:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_windows_pytorch_wheels_ci.yml (which I'm renaming to `multi_arch_build_windows_pytorch_wheels_ci.yml`)

harkgill-amd: Ah I see, adding them in through https://github.com/ROCm/TheRock/pull/5912

PR #5695, opened 2026-06-08
[CI] Skip flaky pytorch tests

zjliu-amd:
PR for bringing up pytorch 2.11 release branch on nightly ROCm.

### DIFF
--- a/external-builds/pytorch/skip_tests/generic.py
+++ b/external-builds/pytorch/skip_tests/generic.py
@@ -23,6 +23,9 @@ skip_tests = {
             #   AssertionError: <torch.cuda.Stream ...> != <torch.cuda.Stream cuda_stream=0x0>
             # Seems to fails on Linux and Windows across torch versions and all tested GPUs.
             "test_side_stream_backward_overlap",
+            # Flaky test that starts a CPU computation and a GPU one, and tests based on the assumption
+            # that the CPU one finishes first, but there's no mechanism ensures that.
+            "test_reentrant_parent_error_on_cpu_cuda",
         ],
         "cuda": [
             # HIP_VISIBLE_DEVICES and CUDA_VISIBLE_DEVICES not working


external-builds/pytorch/skip_tests/generic.py
```diff
@@ -23,6 +23,9 @@
             #   AssertionError: <torch.cuda.Stream ...> != <torch.cuda.Stream cuda_stream=0x0>
             # Seems to fails on Linux and Windows across torch versions and all tested GPUs.
             "test_side_stream_backward_overlap",
+            # Flaky test that starts a CPU computation and a GPU one, and tests based on the assumption
+            # that the CPU one finishes first, but there's no mechanism ensures that.
+            "test_reentrant_parent_error_on_cpu_cuda",
```
ScottTodd, line 28: Please add a description to this PR.

You could include or link to logs showing how the test failed, for example.

Source code for this test:
https://github.com/pytorch/pytorch/blob/e7841ce43c05f7f9e3c54e8ab7e0d5fd65ca5512/test/test_autograd.py#L13257-L13319

Is there an upstream issue tracking de-flaking this test? We generally don't want to carry code downstream if we can avoid it, especially for extended periods of time.

zjliu-amd: Thanks for the review. I believe this is more of a rock CI issue, therefore I decided to put it here. Example run: https://github.com/ROCm/TheRock/actions/runs/27056253036/job/79866543236
This test is flaky by nature because it assumes a CPU computation finishes before a GPU one, and all the following assertions are based on that. But the only thing the test does to ensure that assumption is to make GPU do more computation and hope it finishes later.
However, in the upstream CI, this is not that much a problem because upstream CI has multiple level of reruns (pytest level and process level), which stabilizes the test. Also, the upstream CI presumably has more CPU capacity (usually upstream CI can use more than 100 cores, depending on the specific hardware), and therefore there's a higher probability the CPU computation finishes before GPU. In my local reproduction, I can only reproduce the exact same error with high probability when limiting CPU usage to 1 core. So I think this is more of a rock CI environment issue and put it here.

ScottTodd: Huh? If the test is flaky, the test is flaky. That's a test design issue, not something to patch over with retries or by choosing to run the test on different hardware.

PR #5700, opened 2026-06-08
[ci] Setting AWS build runners to 0%

kailash-khalasi:
Due to an upgrade in production, we are setting AWS build runners to 0

Reopened from #5568 (rebased on main).

### DIFF
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -51,8 +51,8 @@ def select_weighted_label(labels_config: list[dict], context_name: str) -> str:
 BUILD_RUNNER_LABELS = {
     "linux": {
         "default": [
-            {"label": "azure-linux-scale-rocm", "weight": 0.8},
-            {"label": "aws-linux-scale-rocm-prod", "weight": 0.2},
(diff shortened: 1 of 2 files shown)


amd-shiraz (approved):

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -51,8 +51,8 @@ def select_weighted_label(labels_config: list[dict], context_name: str) -> str:
 BUILD_RUNNER_LABELS = {
     "linux": {
         "default": [
-            {"label": "azure-linux-scale-rocm", "weight": 0.8},
-            {"label": "aws-linux-scale-rocm-prod", "weight": 0.2},
+            {"label": "azure-linux-scale-rocm", "weight": 1.0},
+            {"label": "aws-linux-scale-rocm-prod", "weight": 0.0},
```
ScottTodd, line 55: Looks like https://docs.python.org/3/library/random.html#random.random is `0.0 <= X < 1.0` so this is _probably_ safe, but we could also remove the `aws-linux-scale-rocm-prod` line entirely or leave it commented out

ScottTodd (approved):

PR #5702, opened 2026-06-08
[ci] Adopting `test_filtering.md` correctly in all test scripts and github actions scripts

geomin12:
As test standarization work is set in and to match behavior in https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md, we are converting the default to standard for all scripts / workflows

Previously, the default was full which is inaccurate


JonathanLichtnerAMD (approved):
LGTM

PR #5708, opened 2026-06-09
[ROCM-20223] CV Integration - RPP inclusion (3/3)

kiritigowda:
## Motivation

Integrate RPP (ROCm Performance Primitives) into TheRock under a new cv-libs group

Track: Jira ID - ROCM-20223

## Technical Details

Integrates [RPP](https://github.com/ROCm/rocm-libraries/tree/develop/projects/rpp) into TheRock under a new top-level `cv-libs/` (computer vision) group.

- **New `cv-libs/` component**: `CMakeLists.txt`, `artifact-rpp.toml`, and `README.md`. RPP builds from `rocm-libraries/projects/rpp` with the HIP backend (`-DBACKEND=HIP`, `-DRPP_AUDIO_SUPPORT=OFF`).

- **Minimal, audited dependency set** (verified against upstream RPP CMake): `RUNTIME_DEPS hip-clr`, `BUILD_DEPS rocm-half`, `COMPILER_TOOLCHAIN amd-hip`, and `therock_explicit_finders.cmake` for RPP's `find_path(half/half.hpp)`. 

- **Build topology**: new `cv-libs` build stage, artifact group, and `rpp` artifact (`feature_group = "CV_LIBS"`, `artifact_deps = ["core-runtime", "core-hip", "base", "sysdeps"]`).

- **Feature flags**: `THEROCK_ENABLE_CV_LIBS` (group) and `THEROCK_ENABLE_RPP`. On by default on Linux; experimental and **off by default on Windows** (opt-in via `-DTHEROCK_ENABLE_RPP=ON`).

- **CI**: new `cv-libs` stage in the Linux portable build workflow and an `rpp` CTest job (builds the installed test tree and runs `ctest`, 60 min timeout for RPP's long QA suites).

- **Packaging / install / dist**: `amdrocm-rpp`, `amdrocm-rpp-devel`, `amdrocm-rpp-test` Debian/RPM packages wired into the core metapackages; `--rpp` flag in `install_rocm_from_artifacts.py`; `librpp.so*` entry in `_dist_info.py` under the `libraries` package (no Windows DLL pattern, since RPP is not built on Windows in CI); `rpp` in the libraries wheel filter in `build_python_packages.py`, plus a `lib/llvm/lib` rpath on that wheel so `librpp.so` resolves libomp from core; coverage in `analyze_build_times.py` and the manylinux/container configure paths.

- **Docs**: README feature-flag tables, build_system, artifacts, installing_artifacts, ci_overview, ci_behavior_manipulation, and windows_support updated.

## Notes

- Windows support for RPP is experimental and not built in CI; upstream blockers (e.g. unconditional `stdc++fs` link) are tracked in [ROCm/rpp#729](https://github.com/ROCm/rpp/issues/729).

## Test Plan

- [x] `cv-libs` CI stage builds RPP green on Linux for the target families

- [x] `rpp` CTest job passes (installed test tree configures + `ctest` runs)

- [x] `python build_tools/topology_to_cmake.py --validate-only` passes (verified locally)

- [x] `pytest build_tools/tests/build_topology_test.py build_tools/tests/fetch_sources_mirror_test.py` passes (verified locally: 42 passed)

- [x] `install_rocm_from_artifacts.py --rpp --tests` fetches the expected artifacts

- [x] Packages build and the `amdrocm-rpp*` deps resolve

## Test Result

RPP artifacts built with TheRock

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -726,6 +726,20 @@ test_matrix = {
             "linux": 1,
         },
     },
+    "rpp": {
+        "job_name": "rpp",
+        "fetch_artifact_args": "--rpp --tests",
+        # Sized for comprehensive/full, which runs the perf suites serially and
+        # upstream allows 4000s each. quick and standard are far under this.
+        # TODO(ROCm/rocm-libraries#10187): lower once perf tests are split out
+        # of the correctness suite.
+        "timeout_minutes": 60,
+        "test_script": f"python {_get_script_path('test_rpp.py')}",
+        "platform": ["linux"],
+        "total_shards_dict": {
+            "linux": 1,
+        },
+    },
     # aqlprofile tests
     "aqlprofile": {
         "job_name": "aqlprofile",
--- a/build_tools/github_actions/test_executable_scripts/test_rpp.py
+++ b/build_tools/github_actions/test_executable_scripts/test_rpp.py
@@ -1,0 +1,115 @@
+import logging
+import os
+import re
+import shlex
+import subprocess
+from pathlib import Path
+import sys
+import platform
+
+logging.basicConfig(level=logging.INFO)
+THEROCK_BIN_DIR_STR = os.getenv("THEROCK_BIN_DIR")
+if THEROCK_BIN_DIR_STR is None:
+    logging.info(
+        "++ Error: env(THEROCK_BIN_DIR) is not set. Please set it before executing tests."
+    )
+    sys.exit(1)
+THEROCK_BIN_DIR = Path(THEROCK_BIN_DIR_STR)
+SCRIPT_DIR = Path(__file__).resolve().parent
+THEROCK_DIR = SCRIPT_DIR.parent.parent.parent
+THEROCK_TEST_DIR = Path(THEROCK_DIR) / "build"
+
+RPP_TEST_PATH = str(Path(THEROCK_BIN_DIR).resolve().parent / "share" / "rpp" / "test")
+if not os.path.isdir(RPP_TEST_PATH):
+    logging.info(f"++ Error: rpp tests not found in {RPP_TEST_PATH}")
+    sys.exit(1)
+else:
+    logging.info(f"++ INFO: rpp tests found in {RPP_TEST_PATH}")
+env = os.environ.copy()
+
+# GitHub Actions passes an empty string when a workflow input is left blank.
+TEST_TYPE = (os.getenv("TEST_TYPE") or "standard").lower()
+
+
+def test_filter_args():
+    """CTest filter for the requested category, per docs/development/test_filtering.md.
+
+    The two `test_type_1` entries are RPP's performance suites and take ~58% of
+    the total runtime, so they are reserved for comprehensive/full.
+    """
+    if TEST_TYPE == "quick":
+        return ["-R", "rpp_sanity_test"]
+    if TEST_TYPE == "standard":
+        return ["-E", "test_type_1"]
+    return []
+
+
+# set env variables required for tests
+def setup_env(env):
+    ROCM_PATH = Path(THEROCK_BIN_DIR).resolve().parent
+    env["ROCM_PATH"] = str(ROCM_PATH)
+    logging.info(f"++ rpp setting ROCM_PATH={ROCM_PATH}")
+    if platform.system() == "Linux":
+        HIP_LIB_PATH = Path(THEROCK_BIN_DIR).resolve().parent / "lib"
+        logging.info(f"++ rpp setting LD_LIBRARY_PATH={HIP_LIB_PATH}")
+        if "LD_LIBRARY_PATH" in env:
+            env["LD_LIBRARY_PATH"] = f"{HIP_LIB_PATH}:{env['LD_LIBRARY_PATH']}"
+        else:
+            env["LD_LIBRARY_PATH"] = str(HIP_LIB_PATH)
+    else:
+        logging.info("++ rpp tests only supported on Linux")
+        sys.exit(0)
+
+
+def execute_tests(env):
+    RPP_TEST_DIR = Path(THEROCK_TEST_DIR) / "rpp-test"
+    RPP_TEST_DIR.mkdir(parents=True, exist_ok=True)
+
+    # rpp ships its tests as CMake source, built here against the installed
+    # headers/libs because some test deps are only available from the system.
+    # TODO(ROCm/rocm-libraries#10187): drop once prebuilt test binaries ship.
+    cmd = [
+        "cmake",
+        "-GNinja",
+        RPP_TEST_PATH,
+    ]
+    logging.info(f"++ Exec [{RPP_TEST_DIR}]$ {shlex.join(cmd)}")
+    subprocess.run(cmd, cwd=RPP_TEST_DIR, check=True, env=env)
+
+    filter_args = test_filter_args()
+    logging.info(f"++ rpp test category TEST_TYPE={TEST_TYPE}")
+
+    cmd = [
+        "ctest",
+        "-N",
+    ] + filter_args
+    logging.info(f"++ Exec [{RPP_TEST_DIR}]$ {shlex.join(cmd)}")
+    ctest_list = subprocess.run(
+        cmd,
+        cwd=RPP_TEST_DIR,
+        check=True,
+        env=env,
+        capture_output=True,
+        text=True,
+    )
+    logging.info(ctest_list.stdout)
+    match = re.search(r"Total Tests:\s*(\d+)", ctest_list.stdout)
+    if match is None:
+        raise RuntimeError(
+            "Failed to determine CTest test count from `ctest -N` output"
+        )
+    if int(match.group(1)) == 0:
+        raise RuntimeError(f"CTest discovered zero rpp tests for TEST_TYPE={TEST_TYPE}")
+
+    cmd = [
+        "ctest",
+        "--extra-verbose",
+        "--output-on-failure",
+    ] + filter_args
+    logging.info(f"++ Exec [{RPP_TEST_DIR}]$ {shlex.join(cmd)}")
+    subprocess.run(cmd, cwd=RPP_TEST_DIR, check=True, env=env)
+
+
+if __name__ == "__main__":
+    setup_env(env)
+    execute_tests(env)
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -562,13 +562,16 @@ def retrieve_artifacts_by_run_id(args):
             extra_artifacts.append("rpp")
             # test_rpp.py compiles the test suite against the installed tree,
             # so the _lib expansion below is not sufficient:
-            #   rpp_dev  - lib/cmake/rpp for find_package(rpp), plus headers.
-            #   base_dev - include/half/half.hpp, which api/rppdefs.h includes
-            #              to define Rpp16f.
-            # OpenMP needs nothing extra here: libomp.so and omp.h both ship
-            # in amd-llvm_lib, already fetched via base_artifact_patterns.
+            #   rpp_dev      - lib/cmake/rpp for find_package(rpp), plus headers.
+            #   base_dev     - include/half/half.hpp, which api/rppdefs.h
+            #                  includes to define Rpp16f.
+            #   amd-llvm_dev - lib/llvm/lib/cmake/AMDDeviceLibs. rpp-config.cmake
+            #                  calls find_dependency(HIP), and hip-config-amd.cmake
+            #                  in turn resolves AMDDeviceLibs through the
+            #                  lib/cmake/AMDDeviceLibs shim in base_lib.
             argv.append("rpp_dev")
             argv.append("base_dev")
+            argv.append("amd-llvm_dev")
         if args.libhipcxx:
             extra_artifacts.append("libhipcxx")
             argv.append("amd-llvm_dev")


kiritigowda:
requires PR #5705

cv-libs/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+if(THEROCK_ENABLE_RPP)
+  ##############################################################################
+  # rpp (ROCm Performance Primitives) library
+  ##############################################################################
+  # RPP is a comprehensive, high-performance computer vision library for AMD
+  # processors with a HIP backend. Its sources live in the rocm-libraries
+  # monorepo (projects/rpp). Unlike the media-libs (rocDecode/rocJPEG), RPP does
```
LakshmiKumar23: Please remove mention on media-libs here.

cv-libs/README.md
```diff
@@ -0,0 +1,34 @@
+# ROCm Computer Vision Libraries
+
+This directory contains computer vision libraries for AMD GPUs.
+
+- **RPP** (ROCm Performance Primitives) -- a comprehensive, high-performance
+  computer vision library for AMD processors with a HIP backend.
+
+## Dependencies
+
+Unlike the media libraries (rocDecode / rocJPEG), RPP does not require the
```
LakshmiKumar23: remove reference to rocDecode/rocJPEG

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -645,6 +645,16 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
             "linux": 1,
         },
     },
+    "rpp": {
+        "job_name": "rpp",
+        "fetch_artifact_args": "--rpp --tests",
+        "timeout_minutes": 30,
```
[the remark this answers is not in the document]
rrawther: Are we going to not run those longer tests?

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -263,6 +263,7 @@ def determine_target_family() -> str:
 LibraryEntry("rocm_smi64", "core", "librocm_smi64.so*", "")
 LibraryEntry("rocdecode", "core", "librocdecode.so*", "")
 LibraryEntry("rocjpeg", "core", "librocjpeg.so*", "")
+LibraryEntry("rpp", "core", "librpp.so*", "rpp*.dll")
```
rrawther: I think the second entry should be "libraries"  instead of core

rrawther (changes requested):
added some review comments

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
kiritigowda: changes to libs

spolifroni-amd (dismissed):
If you want to generate install documentation yourself, generate doc similar to that for projects/rocprim, ensuring that the doc generated references the right packages.

kiritigowda:
Requires - https://github.com/ROCm/rocm-libraries/pull/8295 to land in rocm-libraries before this can be merged.

cv-libs/README.md
```diff
@@ -0,0 +1,33 @@
+# ROCm Computer Vision Libraries
+
+This directory contains computer vision libraries for AMD GPUs.
+
+- **RPP** (ROCm Performance Primitives) -- a comprehensive, high-performance
+  computer vision library for AMD processors with a HIP backend.
```
ScottTodd: Is there an RFC somewhere for inclusion of RPP and "Computer Vision Libraries" into ROCm core (TheRock's continuous builds+releases)? I'm wondering about the increased scope here and what it may mean for CI/CD. Could you get someone from product or @stellaraccident to give this the green light before we review? (We should add more of an intake process with some written down policies to help guide there)

Once the subproject has the go-ahead for inclusion, we'll probably want to split this into several smaller PRs instead of jumping straight to builds + tests + edits across the whole project.

patches/amd-mainline/rocm-libraries/0001-rpp-make-public-include-directories-relocatable.patch
```diff
@@ -0,0 +1,54 @@
+From 0000000000000000000000000000000000000001 Mon Sep 17 00:00:00 2001
+From: therockbot <therockbot@amd.com>
+Date: Wed, 10 Jun 2026 00:00:00 +0000
+Subject: [PATCH] rpp: make public include directories relocatable
```
ScottTodd: 1. Patches should generally be authored by humans, not bot accounts.
2. We will not accept a patch to rocm-libraries for a new component like this when rocm-libraries currently (after significant effort) has 0 patches. Land the changes in rocm-libraries. See https://github.com/ROCm/TheRock/tree/main/patches#rules-for-creating-patches

LakshmiKumar23 (approved):

cv-libs/README.md
rrawther: We got greenlight from PM (Saad/Marylin). Please talk to him.

rrawther (approved):

BUILD_TOPOLOGY.toml
```diff
@@ -124,6 +124,10 @@ description = "ROCm media submodules (Mesa for VA-API)"
 submodules = ["amd-mesa"]
 disable_platforms = ["windows"]
 
+[source_sets.cv-libs]
+description = "ROCm computer vision libraries submodules (RPP lives in rocm-libraries)"
+submodules = []
```
marbre:
```suggestion
submodules = []
disable_platforms = ["windows"]
```
?

patches/amd-mainline/rocm-libraries/0002-rpp-guard-test-suite-OpenCV-only-compare-helper.patch
marbre, line 1: As Scott pointed out in https://github.com/ROCm/TheRock/pull/5708/changes#r3399770819, patches will not be accepted.

build_tools/bump_submodules.py
marbre, line 1: There shouldn't be the option to bump cv-libs separately as it will always result in bumping rocm-libraries. I suggest to drop changes to this script.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
marbre, line 1: Was it aligned with BU that rpp should go into the `rocm-sdk-*` Python packages or should it rather go to separated packages. @saadrahim?

marbre (commented):
As Scott pointed out, patches need to be upstreamed and cannot go into TheRock. Futhermore, BU needs to confirm the Python packaging approach here.

We discussed this before already and Scott raised it again but this should go into smaller PRs:
* Build
* Packaging
* Testing

This will make it easier to review and to land as there probably isn't much to discuss about the build part. Packaging however will need folks working on native Linux packages, whereas others need to review Python packaging (after BU approved). Testing is yet another party.

patches/amd-mainline/rocm-libraries/0001-rpp-make-public-include-directories-relocatable.patch
kiritigowda: This patch was a stopgap, they are removed as rocm-libraries is updated

LakshmiKumar23 (approved):

kiritigowda:
Requires rocm-libraries - f513fcc4772a8c54ff852f887df4ee51f9731ead +

stellaraccident (commented):
I want to align internally before landing this into rocm core.. Generally frameworks are customers of rocm core, not a part of it.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -263,6 +263,7 @@ def determine_target_family() -> str:
 LibraryEntry("rocm_smi64", "core", "librocm_smi64.so*", "")
 LibraryEntry("rocdecode", "core", "librocdecode.so*", "")
 LibraryEntry("rocjpeg", "core", "librocjpeg.so*", "")
+LibraryEntry("rpp", "core", "librpp.so*", "")
```
rrawther: why this is part of core, not libs?

rrawther (commented):
please address review comments

patches/amd-mainline/rocm-libraries/0002-rpp-guard-test-suite-OpenCV-only-compare-helper.patch
kiritigowda: Patches were temp, they have landed into rocm-libs

saadrahim:
> I want to align internally before landing this into rocm core.. Generally frameworks are customers of rocm core, not a part of it.

RPP is a low level library operating on the VCN engines and HIP runtime. Frameworks will be built on the capabilities exposed by RPP.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
stellaraccident: Should be libraries

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
stellaraccident: Yes, we are aligned to add it to the libraries package. It is ~23MiB (generic for all arches).

cv-libs/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+# SPDX-License-Identifier: MIT
+
+if(THEROCK_ENABLE_RPP)
+  ##############################################################################
+  # rpp (ROCm Performance Primitives) library
+  ##############################################################################
+  # RPP is a comprehensive, high-performance computer vision library for AMD
+  # processors with a HIP backend. Its sources live in the rocm-libraries
+  # monorepo (projects/rpp). It only needs the HIP runtime, the half-precision
+  # headers and OpenMP.
```
stellaraccident: For my own edification, does the library itself require OMP? Or is that just for tests?

cv-libs/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+      # Audio augmentations require FFTS, which is not yet bundled in TheRock, so
+      # this is a known functional gap vs. an upstream RPP build.
+      # TODO: bundle FFTS as a sysdep and flip this ON to enable audio support.
+      -DRPP_AUDIO_SUPPORT=OFF
+    CMAKE_INCLUDES
+      therock_explicit_finders.cmake
+    # RPP's bundled test suite (utilities/test_suite) calls find_package(rpp)
+    # on itself. Ignore it so the dep provider falls through to the system
+    # resolver instead of erroring on an undeclared super-project package.
+    IGNORE_PACKAGES
```
stellaraccident, line 32: Can you please create an issue to fix this upstream? Ideally one of two things happens:

1. You update your bundled test suite to not find_package() itself when built in-tree.
2. You build the rpp project itself without its test project and declare a second rpp-tests project that builds the integration tests atop rpp.

I prefer #2 for ROCm overall, but all of the existing projects were hardcoded for #1 and we keep cargo culting that.

cv-libs/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+      rpp
+    INTERFACE_LINK_DIRS
+      "lib"
+    INSTALL_RPATH_DIRS
```
stellaraccident: I am surprised that this and the INTERFACE_INSTALL_RPATH_DIRS directives are required. Can you double check and let me know what errors you get if excluded?

cv-libs/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+      "lib;lib/rocm_sysdeps/lib"
+  )
+  therock_cmake_subproject_glob_c_sources(rpp
+    SUBDIRS
+    .
+  )
+  therock_cmake_subproject_provide_package(rpp rpp lib/cmake/rpp)
+  therock_cmake_subproject_activate(rpp)
+  therock_test_validate_shared_lib(
+    PATH ${CMAKE_CURRENT_BINARY_DIR}/rpp/dist/lib
```
stellaraccident: Quote non-trivial CMake expansions please.

BUILD_TOPOLOGY.toml
stellaraccident: I think that Marius' suggestion is correct.

BUILD_TOPOLOGY.toml
```diff
@@ -322,6 +333,14 @@ type = "generic"
 artifact_group_deps = ["profiler-core", "compiler", "core-amdsmi"]
 source_sets = ["rocm-systems"]
 
+[artifact_groups.cv-libs]
+description = "Computer Vision Libraries"
+type = "generic"
+artifact_group_deps = ["third-party-sysdeps", "base", "core-runtime", "hip-runtime"]
+# RPP sources live in the rocm-libraries monorepo.
+# TODO: rocm-systems included for projects/hip/VERSION (see CMakeLists.txt)
+source_sets = ["rocm-libraries", "rocm-systems"]
```
stellaraccident, line 334: You declare the cv-libs source set above but then only refer to rocm-libraries. I suspect you can drop the source-set. The reason not to would be that we may want to extend this tooling later to do sparse checkout of projects like this (so that they don't need to clone the entire rocm-libraries tree in a stage that only builds this one thing). But that feature is not yet implemented and having a stub above doesn't get us very far to it while just adding confusion.

stellaraccident (approved):
This looks good to me in principle. Please get Marius' stamp on specifics (he made some comments originally that look actually unresolved) before landing. And coordinate with DevOps on actual commit so that they are aware and can handle any unanticipated fallout.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
kiritigowda: Done, it's libraries now

cv-libs/CMakeLists.txt
kiritigowda: RPP does utilize OMP; It's host-only: OpenMP parallelizes the CPU tensor kernels (RPP has a HOST
backend alongside HIP). There's no -fopenmp-targets or device offload, so it adds no
per-architecture dimension to the build.

cv-libs/CMakeLists.txt
kiritigowda: Filed ROCm/rocm-libraries#10163. It describes both options and calls out #2 (a separate
rpp-tests project built against installed RPP) as the preferred shape, per your note.

cv-libs/CMakeLists.txt
kiritigowda: Done all quoted.

BUILD_TOPOLOGY.toml
kiritigowda: Done — dropped [source_sets.cv-libs] entirely.

build_tools/github_actions/test_executable_scripts/test_rpp.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            env["LD_LIBRARY_PATH"] = str(HIP_LIB_PATH)
+    else:
+        logging.info("++ rpp tests only supported on Linux")
+        sys.exit(0)
+
+
+def execute_tests(env):
+    RPP_TEST_DIR = Path(THEROCK_TEST_DIR) / "rpp-test"
+    RPP_TEST_DIR.mkdir(parents=True, exist_ok=True)
```
ScottTodd, line 66: Per my earlier comment

> Once the subproject has the go-ahead for inclusion, we'll probably want to split this into several smaller PRs instead of jumping straight to builds + tests + edits across the whole project.

Including tests in this PR itself raises a few extra questions. Please prefer sequences of smaller PRs so they can be individually reviewed by area experts in a timely manner. Bundling changes to build, packaging, and testing together increases the surface area that reviewers need to consider.

Now, the specific comment here: please add support for test filter levels to this script using the `TEST_TYPE` environment variable following https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md. I see from the test run on https://github.com/ROCm/TheRock/actions/runs/30561055293/job/90957557429?pr=5708 that this is using 43 minutes to test. If the `TEST_TYPE` environment variable is set to `quick`, the budget for _all_ subprojects is 5 minutes, with a few currently taking 20 minutes (see https://github.com/ROCm/TheRock/actions/runs/30523867591/usage for example). This would be the _slowest test job_ as currently implemented.

cc @dileepr1

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -284,6 +284,7 @@ def determine_target_family() -> str:
 LibraryEntry("miopen", "libraries", "libMIOpen.so*", "MIOpen*.dll")
 LibraryEntry("origami", "libraries", "liborigami.so*", "origami*.dll")
 LibraryEntry("hipdnn", "libraries", "libhipdnn_backend.so*", "hipdnn_backend*.dll")
+LibraryEntry("rpp", "libraries", "librpp.so*", "")
```
ScottTodd: Want to sort `rpp` next to `rocdecode` and `rocjpeg`? Or next to some other library instead? These are roughly sorted by dependency ordering so adding all the way at the end of the list like this could suggest that rpp has dependencies on the math/blas/ml libraries

build_tools/build_python_packages.py
```diff
@@ -324,6 +324,8 @@ def _run_kpack_split(
     lib.rpath_dep(core, "lib")
     lib.rpath_dep(core, "lib/rocm_sysdeps/lib")
     lib.rpath_dep(core, "lib/host-math/lib")
+    # rpp needs libomp, which ships in core under lib/llvm/lib.
+    lib.rpath_dep(core, "lib/llvm/lib")
```
ScottTodd, line 328: Here are other files in `lib/llvm/lib/`, FYI:
```
libLLVM.so.23.0git
libLLVMOffload.so.23.0git
libMLIR.so.23.0git
libMLIRExecutionEngineShared.so.23.0git
libarcher.so
libc++.modules.json
libclang-cpp.so.23.0git
libclang.so.23.0git
libmlir_apfloat_wrappers.so.23.0git
libmlir_arm_runner_utils.so.23.0git
libmlir_arm_sme_abi_stubs.so.23.0git
libmlir_async_runtime.so.23.0git
libmlir_c_runner_utils.so.23.0git
libmlir_float16_utils.so.23.0git
libmlir_runner_utils.so.23.0git
libomp.so
libomptarget.so.23.0git
libunwind.so.1
```

@marbre any concerns with this rpath dep also including libMLIR, libclang, libc++, etc.?

ScottTodd (commented):
Generally LGTM given Stella's approval for inclusion.

My main concerns right now would be
1. The test time as currently implemented (40 minutes compared to other "quick" subproject tests that take closer to 5-20 minutes)
2. We can add the `ci:run-all-archs` label to this PR to build and test for all available GPU targets so there aren't any surprises about build or test results post-merge once our nightly releases, submodule updates, etc. start building this for a wider set of targets than are normally tested on PRs.

Edit: We could also add the `ci:asan` label to check that this component builds with ASan too

cv-libs/CMakeLists.txt
kiritigowda: Removal has no errors, both removed.

All three directories already arrive transitively via RUNTIME_DEPS (hip-clr gives lib and lib/rocm_sysdeps/lib, amd-llvm gives lib/llvm/lib), and therock_subproject.cmake dedups the declared and transitive lists. I built the cv-libs stage in the manylinux CI image both ways: identical binaries apart from RPATH ordering, no unresolved DT_NEEDED either way.

I'd copied the pair from media-libs, which has the same redundant declarations, I can clean that too.

Thanks!

build_tools/github_actions/test_executable_scripts/test_rpp.py
kiritigowda: Added. The suite is 8 ctest entries

quick → -R rpp_sanity_test → 2 tests, 9.6s
standard → -E test_type_1 → 6 tests, ~16 min
comprehensive/full → all 8, ~38 min

kiritigowda: @ScottTodd & @marbre -- Agreed. Splitting into three, landing in sequence since each depends on the previous:

1. Build — BUILD_TOPOLOGY.toml, cv-libs/ (CMake, artifact descriptor, README), feature flag
2. Packaging — build_python_packages.py, _dist_info.py, install_rocm_from_artifacts.py, docs
3. CI tests — test_rpp.py, fetch_test_configurations.py

I'll keep this PR open and peel 1 and 2 off into separate PRs. As those merge, this PR's diff will shrink accordingly, leaving it scoped to just the CI test changes.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
kiritigowda: done

ScottTodd (approved):
Thanks for splitting this into smaller PRs and for adding the test filter levels here. LGTM now.

kiritigowda:
> Thanks for splitting this into smaller PRs and for adding the test filter levels here. LGTM now.

@ScottTodd Thank you for taking the time to review and approve!

PR #4803, opened 2026-04-24
packaging: add Windows MSI installer for ROCm Artifacts

idass1990:
## Motivation

ROCm on Windows currently has no standardized installer. Users must manually copy DLLs and configure PATH, which is error-prone and not suitable for enterprise deployment. This PR introduces per-machine MSI installers built with WiX v4, enabling silent scripted installation, automatic upgrades, clean uninstall, and automatic PATH configuration.

> **Note:** This is phase 1 of the Windows Installer project, focusing on runtime redistributable packages. See tracking issue #1987.

### Generator script (`generate_msi_wxs.py`)

- Uses **`ArtifactCatalog`** from `_therock_utils.artifacts` as the source of truth for file selection — reads `artifact_manifest.txt` embedded in each artifact archive directly, with no TOML parsing or glob re-evaluation at generation time
- `--artifacts-url`: download and extract `.tar.zst` artifacts from remote storage (S3/CI) without a local build; automatically fetches legacy DLLs (`amdhip64_6.dll`, `amd_comgr_2.dll`) from DVC (`rocm-systems/shared/amdgpu-windows-interop/legacy/`)
- `--package` selector: single script generates multiple named MSI packages
- Flat install layout: files install to `bin/`, `lib/`, `share/` directly under `InstallDir`
- **System PATH**: adds `InstallDir\bin` to the machine `PATH` (appended, removed cleanly on uninstall) so ROCm binaries and DLLs are discoverable without manual configuration
- **Long path support**: `ENABLE_LONG_PATHS` sets the `LongPathsEnabled` registry key (enabled by default; disable with `ENABLE_LONG_PATHS=0`). A pre-existing value set by the user is preserved on uninstall — the installer only reverts the key if it was the one that set it
- `per_artifact_includes`: per-artifact file scoping via `ArtifactCatalog` includes, for cases where only a subset of an artifact's files are needed in a package
- `INSTALLFOLDER` property: users can override the install directory at `msiexec` invocation time
- `LEGACY_INSTALL` property: installs key DLLs into `C:\Windows\System32` for applications that load ROCm DLLs by bare name; enabled by default, disable with `LEGACY_INSTALL=0`
- Deterministic WiX element IDs via SHA-256
- Unit test suite (`tests/generate_msi_wxs_test.py`, 43 tests) wired into the `Unit Tests` CI workflow (runs on Linux and Windows)

### MSI packages defined

| Package | `--package` | Artifacts | Install path |
|---|---|---|---|
[...]

Issue #1987, opened 2025-10-31
[Windows] Add Windows package installers


rahulc-gh:
Needed for 7.10 ????

ScottTodd:
TBD what we want for this and how we would implement it. See the diagram at https://github.com/ROCm/TheRock/blob/main/docs/development/assets/therock_build_architecture.excalidraw.svg. The "msi" and "inf" native package installers would be for Windows.

ScottTodd:
@amd-aakash @idass1990 please add context to the issue description here. I want to dup a few other issues against this one.

TaiXeflar:
Maybe can start from CPack? 
Adding `include(CPack)` to TheRock somewhere `CMakeLists.txt` and specify the build dist (like `TheRock/build/dist/*`) as the packaging rootdir. Then specify Package Generators like InnoSetup/WIX to pack them up to a executable installer, or 7z setup as a decompress redistributables. (Note: But we probably need to check if those `*.cmake` files will change or not via manual or CI/CD cpack build.)

Not recommends NSIS now. So far it has 2GB limit now...

idass1990:
## Windows MSI Packaging — Implementation Milestones

Posting this as requested in https://github.com/ROCm/TheRock/pull/4803#pullrequestreview-4547966190 to track sequencing and outstanding dependencies.

### Milestone 1 — Generator script (in progress, PR #4803)

`generate_msi_wxs.py` producing MSI packages for `runtime` (HIP + comgr) and `core` (HIP + OpenCL + AMD SMI + ROCR-Runtime) from CI artifact tarballs. Includes:
- ArtifactCatalog-driven file selection (no hand-maintained file lists)
- `--artifacts-url` CI mode with automatic DVC fetch of legacy DLLs from `rocm-systems`
- LegacyInstall feature (System32 DLL placement, on by default, disable with `LEGACY_INSTALL=0`)
- Unit test suite (38 tests)

> **Note:** The `runtime` package is functional and tested against nightly CI artifacts. The `core` package definition is in place but is currently in development — `core-amdsmi` and `core-runtime` Windows artifacts are not yet published in nightly runs, so those components are not yet included in the built MSI.

### Milestone 2 — CI workflow: build MSI packages

Add `multi_arch_build_native_windows_packages.yml` that:
- Accepts an artifact run ID or URL as input
- Calls `generate_msi_wxs.py --artifacts-url` for each package
- Runs `wix build -arch x64` to produce `.msi` files
[...]

PR #5712, opened 2026-06-09
[PoC] Added MSI package building process

gkrivor:
## Motivation

Adds a proof-of-concept Windows MSI packaging pipeline to TheRock, including WiX source definitions for multiple ROCm component MSIs, build scripts to generate them, and workflow automation to produce MSIs from CI artifacts. It also includes RFC docs capturing package file mappings and MSI-vs-DEB comparisons.

## Technical Details

Introduce WiX v4 packaging definitions (.wxs/.wxi) and per-package build scripts for multiple ROCm MSI packages.
Add a GitHub Actions job to fetch artifacts and build/upload MSI outputs on Windows runners.
Add RFC0012 documentation artifacts (brief manifests + full MSI/DEB comparisons) for several packages.

## Submission Checklist


andrei-kochin:
@ScottTodd  @idass1990  while we are struggling with CI you can take a look at draft MSI artifacts here:
[Release Windows packages (dev, gfx110X) · ROCm/TheRock@0ff4ac9](https://github.com/ROCm/TheRock/actions/runs/27362289023)

Thank you @gkrivor !

andrei-kochin:

idass1990 (commented):
# Code Review: PR #5712 — [PoC] Added MSI package building process (Updated)

**Author:** gkrivor | **Base:** main | **Files:** 108 changed | **+33,276 / −0**
**Reviewed against PR #4803 commit `f5339ec` (June 12, 2026)**

---

## What Changed in PR #4803 Since the Original Review

PR #4803 received two significant commits on June 12 that are directly relevant:

**Commit `8fe9dd6` — "packaging: rewrite MSI generator with artifact TOML support"**
- Replaces the static allowlist (`amdrocm-runtimes-artifacts.txt`) with TOML descriptor-driven generation
- Stage-dir scoping: globs run against per-component stage dirs (`build/<basedir>/stage/`) preventing cross-artifact bleed
- `--artifacts-url` flag to pull `.tar.zst` artifacts from S3 without a local build
- Deterministic WiX element IDs via SHA-256
- Full unit test suite (`generate_msi_wxs_test.py`, 509 lines)
- Flat install layout (`bin/`, `lib/`, `share/` directly under `InstallDir`)
- `INSTALLFOLDER` user-overridable at `msiexec` invocation time

**Commit `f5339ec` — "packaging: remove generated wxs, artifacts list, and wixpdb"**
- Removes `amdrocm-runtimes.wxs`, `amdrocm-runtimes-artifacts.txt`, `amdrocm-runtimes.wixpdb` from the repo
- Addresses ScottTodd's binary-file and drift concerns directly

---

## Updated ScottTodd Concern Tracking

| ScottTodd concern (PR #4803) | Status in PR #4803 now | Status in PR #5712 |
|---|---|---|
| Land RFC #3973 first | Unresolved — RFC still open | **Not addressed** |
| System32 DLL deletion needs plan | **Removed** from scope (commit 2026-05-13) | Not applicable |
| Why only `bin/` and `lib/`? | **Partially addressed** — TOML descriptors now define scope per artifact | **Persists** — blas runtime still only ships `bin/` |
| Document prebuilt artifact usage | Addressed in rewritten `msi-generator-usage.md` (`--artifacts-url`) | Partially addressed via CI, no dev docs |
| Long-paths registry key on uninstall | Still open (promised but not yet committed) | Not applicable |
| `.wixpdb` in `.gitignore` | `.wixpdb` removed from commit; **no `.gitignore` update yet** | **Not addressed** |
| Artifact list drift vs Linux packages | **Addressed** — TOML descriptors are shared source of truth with build system | **Regressed** — 108 hand-maintained files |
| Hardcoded versions → use generator | **Addressed** — `_read_rocm_version()` extracts version from build tree | **Regressed** — `1.0.0.0` / `7.14` hardcoded |
| Generate component lists from BOM | **Addressed** — `generate_msi_wxs.py` fully replaces manual lists | **Regressed** — abandoned generator entirely |
| Don't hardcode `build/dist` paths | **Addressed** — paths are relative to stage dirs or `dist_root` arg | **Critical regression** — `B:\build` hardcoded |
| Version-specific registry key | Still open (not committed yet) | Not applicable |

**The approaches in PR #4803 and PR #5712 have fundamentally diverged.** #4803 chose a generator-script approach that addresses nearly all of ScottTodd's concerns. #5712 chose a hand-authored static approach that regresses on all of them.

---

## New Findings from Updated Review

### CRITICAL — WiX `Directory/@Name` backslash bug (confirmed by Copilot, all 11 packages affected)

Every `.wxs` file in PR #5712 has this pattern:
```xml
<Directory Id="INSTALLFOLDER" Name="AMD\$(var.ProductName)\$(var.ProductVersion)">
```

In WiX v4, `Directory/@Name` must be a **single directory segment** — backslashes are not path separators, they are literal characters in the directory name. WiX will either error or create a directory literally named `AMD\ROCm\7.14` as a single folder rather than a three-level hierarchy. This is confirmed by Copilot in 11 separate inline comments. All 11 packages are affected. The correct form is:

```xml
<StandardDirectory Id="ProgramFiles64Folder">
  <Directory Id="dir_amd" Name="AMD">
    <Directory Id="dir_rocm" Name="$(var.ProductName)">
      <Directory Id="INSTALLFOLDER" Name="$(var.ProductVersion)">
```

This is a WiX correctness bug that will either fail at compile time or produce broken installs. **This alone blocks the PR.**

### CRITICAL — Flattened artifact layout vs. deep build-tree paths (Copilot confirmed)

The CI workflow in `multi_arch_release_windows.yml` calls:
```bash
python build_tools/artifact_manager.py fetch --flatten --output-dir="${{ env.DIST_DIR }}"
```
But `common.wxi` defines:
```xml
<?define BuildRoot = "B:\build" ?>
```
And component files reference deep build-tree paths like:
```xml
Source="$(var.BuildRoot)/math-libs/BLAS/hipBLAS/dist/bin/hipblas.dll"
```

With `--flatten`, artifacts land flat in `DIST_DIR` — there are no `math-libs/BLAS/hipBLAS/dist/` subdirectories. The source paths will not exist. **Every MSI build in the CI job will fail** even if `BuildRoot` were corrected. The WiX source files would need to be rewritten to reference flat paths, or the artifact fetch would need to not use `--flatten`.

### CRITICAL — Hardcoded `BuildRoot = "B:\build"` — fatal portability defect

**`common.wxi`:**
```xml
<?define BuildRoot = "B:\build" ?>
```

This hardcodes an absolute path to a specific build machine's drive letter. It will fail on every other machine. ScottTodd raised this exact issue on PR #4803. The fix is to pass `BuildRoot` as a `-d BuildRoot=...` define argument to `wix build`, sourced from the calling `.bat` or CI workflow, not hardcoded in the shared include.

### CRITICAL — Hardcoded version `1.0.0.0` and `ProductVersion = "7.14"` throughout

Every `.wxs` has `Version="1.0.0.0"` and `common.wxi` has `ProductVersion = "7.14"`. These break upgrade detection and silent installs. The version must come from the build system. PR #4803 now has `_read_rocm_version()` that extracts the version from the build tree — the same mechanism should be used here.

### HIGH — `contents: read` missing from CI job permissions (Copilot confirmed)

```yaml
build_msi_packages:
  permissions:
    id-token: write    # only permission set
```

When a job sets any permission explicitly, all others default to `none`. `actions/checkout` requires `contents: read`. Without it, the checkout step will fail on repositories with restricted default permissions. Fix:
```yaml
permissions:
  id-token: write
  contents: read
```

### HIGH — `build-all.bat` silently swallows failures (Copilot confirmed)

```bat
for /f %%I IN ('dir /b "%~dp0*-build.bat"') do (
  call "%~dp0%%I" /no_pause
  if !ERRORLEVEL! neq 0 (
    rem popd
    rem exit /b 1   ← COMMENTED OUT
  )
)
```

Individual package build failures are silently ignored. The CI step will report success even if every MSI fails to build. The `exit /b 1` must be uncommented, or at minimum the loop must accumulate failures and exit non-zero at the end.

### HIGH — MSI S3 upload logic silently lost in merge conflict resolution

Copilot resolved the merge conflict on June 13 by accepting `main`'s deletion of `release_windows_packages.yml` (removed in main PR #5790). The S3 upload sections that were added to that file in PR #5712 are now **silently gone**. If S3 upload of MSIs is required for the release pipeline, that logic needs to be added to whichever workflow now handles releases.

### MEDIUM — No inter-package dependency declarations (`dep:Requires`)

Each `.wxs` declares `dep:` namespace but never uses `<dep:Requires>`. For example, `amdrocm-blas-dev` has no declared dependency on `amdrocm-blas`. Users can install dev headers without the runtime DLLs and get a broken environment.

### MEDIUM — WiX extension versions not locked

```yaml
dotnet tool install --version 6.0.0 --global wix
wix extension add -g WixToolset.UI.wixext/6.0.0
wix extension add -g WixToolset.Dependency.wixext/6.0.0
```

The WiX tool itself is pinned, but extensions are fetched from NuGet at install time. Use a `dotnet-tools.json` manifest or checked-in `wix.json` to lock extension versions.

### LOW — `build-all.bat` script discovery is fragile

The script discovers build scripts by globbing `*-build.bat` with no defined ordering. Any temporary file matching that pattern will be invoked. An explicit ordered list is safer.

### LOW — Missing `.gitignore` for WiX build artifacts

No `.gitignore` entry for `*.msi`, `*.wixpdb`, or `*.wixobj`. ScottTodd called this out on #4803. PR #4803 already removed a committed `.wixpdb`; without `.gitignore` the problem will recur.

### LOW — PR description is entirely blank

The PR template sections (Motivation, Technical Details, Test Plan, Test Result) are all unfilled. This should be filled before requesting review.

---

## Comparison: PR #4803 Approach vs. PR #5712 Approach

| Dimension | PR #4803 (`generate_msi_wxs.py`) | PR #5712 (static `.wxs`) |
|---|---|---|
| File list source | TOML artifact descriptors (build system) | Hand-authored per-file in `.wxi` includes |
| Versioning | `_read_rocm_version()` from build tree | Hardcoded `1.0.0.0` / `7.14` |
| Path portability | Relative to stage dir / `dist_root` arg | Hardcoded `B:\build` |
| Maintenance burden | Low — add a file to TOML, regenerate | High — manually add component to `.wxi` |
| Test coverage | 509-line unit test suite | None |
| Build output committed | No (removed in latest commit) | N/A |
| `.gitignore` | Not yet updated | Not updated |
| Inter-package deps | Not implemented | Not implemented |
| WiX `Directory` correctness | Correct (flat layout, no backslash in Name) | **Incorrect** — backslash in `Directory/@Name` on all 11 packages |
| CI artifact path alignment | Designed to work with `--artifacts-url` | **Incompatible** with `--flatten` CI fetch |

---

## Recommendation

**PR #5712 should not be merged.** Beyond the original blockers, there are now two additional correctness bugs confirmed by Copilot's own review — invalid `Directory/@Name` syntax affecting all 11 packages, and fundamental incompatibility between the CI artifact fetch mode (`--flatten`) and the deep build-tree source paths in the WiX files. The S3 upload work was silently lost in the merge conflict resolution.

**The more viable path for this work is PR #4803**, which now has:
- A complete generator script addressing all of ScottTodd's architectural concerns
- Unit test coverage (509 lines)
- Correct portability (no hardcoded paths or versions)
- TOML-driven artifact sourcing aligned with the build system

What PR #4803 still needs before it can merge:
1. RFC #3973 landing (or explicit team waiver)
2. `.gitignore` update for `*.msi`, `*.wixpdb`, `*.wixobj`
3. Long-paths registry key uninstall behavior clarified and implemented
4. `dep:Requires` inter-package dependency declarations
5. CI integration (workflow jobs could be adapted from PR #5712's approach, using the generator script)

idass1990 (commented):
see my review breakdown in comments

build_tools/packaging/windows/build-all.bat
ScottTodd, line 1: This overlaps with work that @idass1990 has been doing on https://github.com/ROCm/TheRock/pull/4803. Please coordinate on
* https://github.com/ROCm/TheRock/issues/1987
* https://github.com/ROCm/TheRock/pull/3973

build_tools/packaging/windows/amdrocm-dnn-dev-components.wxi
```diff
@@ -0,0 +1,242 @@
...
+    </Component>
+    <Component Id="c_amdrocm_dnn_dev_ihfhf_G.hpp" Guid="*">
+      <File Id="f_amdrocm_dnn_dev_ihfhf_G.hpp" Source="$(var.BuildRoot)/ml-libs/hipDNN/dist/include/hipdnn/frontend/hipdnn_frontend/Graph.hpp" KeyPath="yes" />
+    </Component>
+    <Component Id="c_amdrocm_dnn_dev_ihfhf_H.hpp" Guid="*">
+      <File Id="f_amdrocm_dnn_dev_ihfhf_H.hpp" Source="$(var.BuildRoot)/ml-libs/hipDNN/dist/include/hipdnn/frontend/hipdnn_frontend/Handle.hpp" KeyPath="yes" />
+    </Component>
+    <Component Id="c_amdrocm_dnn_dev_ihfhf_L.hpp" Guid="*">
+      <File Id="f_amdrocm_dnn_dev_ihfhf_L.hpp" Source="$(var.BuildRoot)/ml-libs/hipDNN/dist/include/hipdnn/frontend/hipdnn_frontend/Logging.hpp" KeyPath="yes" />
+    </Component>
```
ScottTodd, line 65: These files need to be generated dynamically based on build system outputs, as the already-in-review https://github.com/ROCm/TheRock/pull/4803 does. We're not going to check in 30k LOC of generated code that needs to be continually updated whenever someone adds a new `Logging.hpp` to any one of our repositories.

.github/workflows/multi_arch_release_windows.yml
```diff
@@ -122,6 +122,66 @@ jobs:
...
+        shell: cmd
+        run: .\build_tools\packaging\windows\build-all.bat "${{ env.DIST_DIR }}" /no_pause
+
+      - name: Upload MSI Packages as artifacts
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: ${{ !cancelled() }}
+        with:
+          name: msi-packages-${{ matrix.family_info.amdgpu_family }}
+          path: ${{ env.DIST_DIR }}\*.msi
+          if-no-files-found: warn
```
ScottTodd, line 183: We don't use github artifacts on this project. We use AWS S3 buckets. See

.github/workflows/multi_arch_release_windows.yml
```diff
@@ -122,6 +122,66 @@ jobs:
             --release-type="${{ inputs.release_type }}" \
             --kpack-split="${{ needs.build_python_packages.outputs.kpack_split }}"
 
+  build_msi_packages:
```
ScottTodd, line 125: Start with the ability to build packages locally in one PR, _then_ connect to release workflows in another PR. This is too risky to review and land in a single PR, we need to be able to test it locally.

ScottTodd (changes requested):
This PR has no description or documentation for adding +33k LOC. Please read https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md (in particular https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#new-feature-development - large PRs without prior communication are unlikely to be accepted, especially if another PR is already in review for the same feature)

build_tools/packaging/windows/amdrocm-dnn-dev-components.wxi
gkrivor: You have a wrong interpretation. Original expectations were: I just take dist/* data and pack it into a package using **/*.* mask. But current build process produces about 200k+ files where only about 12k+ are unique. Such parts highlight a component/build system issue when it produces a wrong structure.

But for now it is enough to build a package without unused files.

.github/workflows/multi_arch_release_windows.yml
gkrivor: GitHub artifacts used because AWS S3 wasn't working correct in Release Windows action.

Would be great if you could cover this part or add someone who could.

.github/workflows/multi_arch_release_windows.yml
gkrivor: Good to know.

But I'm not sure it is good for this case. When you separate the logic, you may get an infinity loop with merge conflicts/fixes in a second part. This PR may/will also include tests which will prevent any unexpected changes in Windows build/packaging process.

So, it depends on an actual state of repositories and how flexible/robust they will be.

gkrivor:
@ScottTodd it is a "prior communication" process right now :) I just shared an object of the discussion, not an actual code review request :)

gkrivor:
@idass1990, @andrei-kochin please, do not request Copilot at this stage. It does a wrong interpretation of the code. Especially regarding misunderstanding of CI-based painful development process :)

.github/workflows/multi_arch_release_windows.yml
ScottTodd: > Especially regarding misunderstanding of CI-based painful development process :)

You can do MUCH better than a "CI-based painful development process" for this. See https://github.com/ROCm/TheRock/blob/main/docs/development/development_guide.md#overall-build-architecture and https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#working-effectively-from-forks

1. Put the package construction code in a script that operates on artifacts
2. Build packages locally, do most of your iteration there
3. Add a standalone workflow that downloads artifacts, produces packages, then uploads packages. most of these are even runnable from personal fork repositories on github-hosted runners and complete in < 10 minutes.:
    * https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_python_packages.yml
    * https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_native_linux_packages.yml
    * https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_tarballs.yml
4. Call that workflow from release workflows

Adding to the release workflow is the final step after all other testing is complete. You shouldn't have to do much iteration at all on that point, so 4hour+ runs like https://github.com/ROCm/TheRock/actions/runs/27404516888 are just an inefficient use of CI and developer time. The risk of "an infinity loop with merge conflicts/fixes in a second part" is low since the code is all simple plumbing: scripts -> call scripts from standalone workflow -> call standalone workflow from release workflow. See also https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash

andrei-kochin: > > Especially regarding misunderstanding of CI-based painful development process :)
> 
> https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#working-effectively-from-forks

Thanks @ScottTodd for suggestions! Let me try to play around on my fork and generate artifacts there and see if it simplify things

PR #5722, opened 2026-06-09
feat(build): forward top-level includes and fix packaging perms

### LINKED ISSUE #6396: Support building from a pre-built build tree when container build user differs from run user
## Problem

When TheRock is configured, built, and installed inside a container (e.g. Docker) where the file/directory owner differs from the user running a later stage — but all files/dirs are group/other-writable (`o+w`) — two things break:

1. **External CMake super-project hooks cannot reach subprojects.** `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` set on the super-project is not forwarded to subprojects, so an external hook used to tweak TheRock's CMake behavior never runs for subproject configures.

2. **`py_packaging` materialization fails on cross-owner files.** `py_packaging.py` does a `shutil.copy2` before modifying a file, but `copy2` preserves the source mode bits. When the source file owner differs from the destination user, the owner-write bit is not guaranteed on the copy, and the subsequent `patchelf` pass cannot open the file for writing. Some upstream files (e.g. LLVM's OMPD gdb module) are additionally installed read-only, which triggers the same failure even for same-owner builds.

## Scenario

- Build tree / install tree produced by one user (container image build user).
- A later configure/build/install stage run by a different user against the existing pre-built tree.
- All files/dirs have write permission for all users (`o+w`), so the operation *should* succeed on permissions alone.

## Desired behavior

- Forward `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` from the super-project to all subprojects so external hooks apply uniformly.
- Ensure materialized files are owner-writable before `patchelf` runs, regardless of source ownership or read-only source mode bits.

PeterCDMcLean:
Closes #6396

## Motivation

Configure, build and install from an existing source / build / install in a container where the file / dir owner is different, but all files / dirs have write permissions for all users.

## Technical Details

1. To enable external cmake super-project hooks to reach subprojects, propagate `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` to all subprojects.
2. `py_packaging.py` performs a `copy2` before modifying. However, if the source file owner is not the same as the destination some permission bits are not set properly, and the subsequent `patchelf` pass cannot open the file for writing. The owner-write bit is now restored after materialization.

## Test Plan

- Built locally with other owner and `o+w` on all files.
- Added unit test `MaterializeReadOnlySourceTest` in `build_tools/tests/py_packaging_test.py` verifying a read-only source is materialized owner-writable.

## Test Result

Proper cross-owner build with no errors (using an external hook to modify TheRock cmake behavior). New unit test passes.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -13,6 +13,7 @@ import os
 from pathlib import Path
 import platform
 import shlex
+import stat
 import subprocess
 import shutil
 import sys
@@ -490,6 +491,10 @@ class PopulatedDistPackage:
         # We have to patch many files, so we do not hard-link: always copy.
         log(f"  MATERIALIZE: {relpath} (from {src_path})", vlog=2)
         shutil.copy2(src_path, dest_path)
+        # copy2 preserves source mode bits. Some upstream files (e.g. LLVM's
+        # OMPD gdb module) are installed read-only, and patchelf below opens the
+        # file for writing, so guarantee the owner-write bit first.
+        os.chmod(dest_path, os.stat(dest_path).st_mode | stat.S_IWUSR)
         if self.files.has(relpath):
             log(f"WARNING: Path already materialized: {relpath}")
         else:
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -714,6 +714,14 @@ function(therock_cmake_subproject_activate target_name)
 
   # Compose CMAKE_PROJECT_TOP_LEVEL_INCLUDES for the subproject: its generated
   # init file first, then any files listed in THEROCK_SUBPROJECT_CMAKE_INCLUDES.
+  foreach(_hook IN LISTS THEROCK_SUBPROJECT_CMAKE_INCLUDES)
+    if(NOT IS_ABSOLUTE "${_hook}")
+      message(FATAL_ERROR "THEROCK_SUBPROJECT_CMAKE_INCLUDES: '${_hook}' must be an absolute path")
+    endif()
+    if(NOT EXISTS "${_hook}")
+      message(FATAL_ERROR "THEROCK_SUBPROJECT_CMAKE_INCLUDES: '${_hook}' does not exist")
+    endif()
+  endforeach()
   set(_subproject_top_level_includes "${_cmake_project_init_file}" ${THEROCK_SUBPROJECT_CMAKE_INCLUDES})
 
   # Handle compiler toolchain.
@@ -723,6 +731,7 @@ function(therock_cmake_subproject_activate target_name)
     "${_compiler_toolchain}" "${_cmake_project_toolchain_file}"
     "${_subproject_top_level_includes}")
   list(APPEND _fprint_files "${_cmake_project_toolchain_file}")
+  list(APPEND _fprint_files ${THEROCK_SUBPROJECT_CMAKE_INCLUDES})
 
   # Customize any other super-project CMake variables that are captured by
   # _init.cmake.
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -712,11 +712,16 @@ function(therock_cmake_subproject_activate target_name)
   # TODO: split into 'build' and 'configure'? Keeping them in sync seems useful.
   _therock_cmake_subproject_build_env_pairs(_build_env_pairs)
 
+  # Compose CMAKE_PROJECT_TOP_LEVEL_INCLUDES for the subproject: its generated
+  # init file first, then any files listed in THEROCK_SUBPROJECT_CMAKE_INCLUDES.
+  set(_subproject_top_level_includes "${_cmake_project_init_file}" ${THEROCK_SUBPROJECT_CMAKE_INCLUDES})
+
   # Handle compiler toolchain.
   set(_compiler_toolchain_addl_depends)
   set(_compiler_toolchain_init_contents)
   _therock_cmake_subproject_setup_toolchain("${target_name}"
-    "${_compiler_toolchain}" "${_cmake_project_toolchain_file}")
+    "${_compiler_toolchain}" "${_cmake_project_toolchain_file}"
+    "${_subproject_top_level_includes}")
   list(APPEND _fprint_files "${_cmake_project_toolchain_file}")
 
   # Customize any other super-project CMake variables that are captured by
@@ -971,19 +976,6 @@ function(therock_cmake_subproject_activate target_name)
       "STAGE_DESTINATION_DIR=${_rel_stage_destination_dir}"
     )
 
-    # Forward the superproject's CMAKE_PROJECT_TOP_LEVEL_INCLUDES (if any) to the
-    # subproject, alongside its generated init file, so a top-level
-    # -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=... reaches every subproject configure.
-    set(_subproject_top_level_includes "${_cmake_project_init_file}")
-    if(CMAKE_PROJECT_TOP_LEVEL_INCLUDES)
-      list(APPEND _subproject_top_level_includes ${CMAKE_PROJECT_TOP_LEVEL_INCLUDES})
-    endif()
-    # Escape the ';' list separator so the multi-file value reaches cmake as a
-    # single argument. Unescaped, the ';' is emitted verbatim into the generator's
-    # '/bin/sh -c' rule as a shell command separator, which then tries to execute
-    # the trailing include file as a program.
-    string(REPLACE ";" "\\;" _subproject_top_level_includes "${_subproject_top_level_includes}")
-
     # Configure command.
     add_custom_command(
       OUTPUT "${_configure_stamp_file}"
@@ -998,7 +990,6 @@ function(therock_cmake_subproject_activate target_name)
         "-DCMAKE_INSTALL_PREFIX=${_stage_destination_dir}"
         "-DTHEROCK_STAGE_INSTALL_ROOT=${_stage_dir}"
         "-DCMAKE_TOOLCHAIN_FILE=${_cmake_project_toolchain_file}"
-        "-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=${_subproject_top_level_includes}"
         ${_cmake_args}
       # CMake doesn't always generate a compile_commands.json so touch one to keep
       # the build graph sane.
@@ -1642,7 +1633,7 @@ endfunction()
 #   * amd-hip: Extends the amd-llvm toolchain to also depend on HIP, making
 #     it ready to use to compile HIP code.
 function(_therock_cmake_subproject_setup_toolchain
-    target_name compiler_toolchain toolchain_file)
+    target_name compiler_toolchain toolchain_file top_level_includes)
   string(APPEND CMAKE_MESSAGE_INDENT "  ")
   set(_build_env_pairs "${_build_env_pairs}")
   set(_toolchain_contents)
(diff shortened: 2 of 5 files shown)


cmake/therock_subproject.cmake
```diff
@@ -957,6 +957,19 @@ function(therock_cmake_subproject_activate target_name)
...
+    # -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=... reaches every subproject configure.
+    set(_subproject_top_level_includes "${_cmake_project_init_file}")
+    if(CMAKE_PROJECT_TOP_LEVEL_INCLUDES)
+      list(APPEND _subproject_top_level_includes ${CMAKE_PROJECT_TOP_LEVEL_INCLUDES})
+    endif()
+    # Escape the ';' list separator so the multi-file value reaches cmake as a
+    # single argument. Unescaped, the ';' is emitted verbatim into the generator's
+    # '/bin/sh -c' rule as a shell command separator, which then tries to execute
+    # the trailing include file as a program.
+    string(REPLACE ";" "\\;" _subproject_top_level_includes "${_subproject_top_level_includes}")
```
marbre:
```suggestion
    list(JOIN _subproject_top_level_includes "$<SEMICOLON>" _subproject_top_level_includes)
```

I think this can be simplified by using CMake's `$<SEMICOLON>"`.

build_tools/_therock_utils/py_packaging.py
marbre, line 1: This change lgtm. It is a tiny one but to get it landed faster I'd propose splitting it into a separate PR next time.

cmake/therock_subproject.cmake
marbre, line 1: So far we silently dropped forwarding `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` with this change we silently pass it forward to every single project, without proper documenting this behavior change. I see the need for this, but maybe let's introduce a new, dedicated `THEROCK_SUBPROJECT_CMAKE_INCLUDES` variables with explicit semantics and clear document that "this propagates to all subprojects". Thus avoids any confusion with what a top level include is.

Depending on what exactly you're trying, the toolchain file mechanism might also already cover your use case and might be an appropriate alternative.

PeterCDMcLean: My thought on it at the time was that CMAKE_PROJECT_TOP_LEVEL_INCLUDES is a documented CMake variable to force include some cmake in a project. TheRock as a superproject would just behave a little more like a single CMake project by automatically propagating it.

Adding a new variable like `THEROCK_SUBPROJECT_CMAKE_INCLUDES`, we would still need to use `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` with the same files to monkey path top-level TheRock cmake. I would prefer the variable just be automatically forwarded. I don't see the value in adding a dedicated TheRock version (unless it was scoped for a specific subproject like ex: hipBLASLt_CMAKE_INCLUDES similar to how hipBLASLt_CMAKE_ARGS works)

Some further details on why this is needed:
`configure_file` cmake native function fails when the owner of the target file is not the same as the user running a build (despite having +w permission). This is common in docker containers build under one UID and run in another. `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` lets us inject a macro that fixes this without needing to add something like this directly to TheRock.
There is other monkey patching needed for some subproject activation when there are prebuilt stage, bin dirs that cmake install absolutely refuses to overwrite another UID's files even when the running UID has +w.

I'd prefer to avoid cluttering up TheRock cmake with this specific container / uid compatibility hence the monkey patching.

As noted, will update documentation.

cmake/therock_subproject.cmake
PeterCDMcLean: Will test this

PeterCDMcLean: `$<SEMICOLON>` gets evaluated to `;` before being passed to the shell and is then misinterpreted.

cmake/therock_subproject.cmake
marbre: @PeterCDMcLean an alternative worth considering: rather than passing `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` as a `-D` argument (which requires the semicolon-escaping workaround), bake it directly into the generated toolchain file.
Since the toolchain is already generated per-subproject via `file(CONFIGURE OUTPUT ... @ONLY),` you can append something like:

```cmake
set(CMAKE_PROJECT_TOP_LEVEL_INCLUDES "@_cmake_project_init_file@" @THEROCK_SUBPROJECT_CMAKE_INCLUDES@)
```

and remove the `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=` from the `add_custom_command`. The toolchain is processed before `project()`, so `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` is already set when CMake consults it. This eliminates the shell-separator escaping problem and keeps the logic in one place.

For the variable name: if the goal is to expose this as a user-facing TheRock option, `THEROCK_SUBPROJECT_CMAKE_INCLUDES` gives it an explicit "propagates to all subprojects" contract without repurposing the CMake built-in but I am not too strongly opinionated when looking through the CMake docs and if you'd rather just transparently forward the CMake variable, that works too. The main change would just be moving the mechanism into the toolchain file generation instead of the command line.

Let me know what you think.

cmake/therock_subproject.cmake
PeterCDMcLean: I moved the CMAKE_PROJECT_TOP_LEVEL_INCLUDES to the toolchain file. no more escaping. good call

cmake/therock_subproject.cmake
PeterCDMcLean: I adopted your idea to use `THEROCK_SUBPROJECT_CMAKE_INCLUDES`. I realized that there is a remote possibility that other users may already be using `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` and overloading it / propagating may break their code.

cmake/therock_subproject.cmake
```diff
@@ -712,11 +712,16 @@ function(therock_cmake_subproject_activate target_name)
   # TODO: split into 'build' and 'configure'? Keeping them in sync seems useful.
   _therock_cmake_subproject_build_env_pairs(_build_env_pairs)
 
+  # Compose CMAKE_PROJECT_TOP_LEVEL_INCLUDES for the subproject: its generated
+  # init file first, then any files listed in THEROCK_SUBPROJECT_CMAKE_INCLUDES.
+  set(_subproject_top_level_includes "${_cmake_project_init_file}" ${THEROCK_SUBPROJECT_CMAKE_INCLUDES})
```
marbre, line 725: As this `THEROCK_SUBPROJECT_CMAKE_INCLUDES` must be an absolute path should we check it and further if the file exists?

```cmake
  foreach(_hook IN LISTS THEROCK_SUBPROJECT_CMAKE_INCLUDES)
    if(NOT IS_ABSOLUTE "${_hook}")
      message(FATAL_ERROR "THEROCK_SUBPROJECT_CMAKE_INCLUDES: '${_hook}' must be an absolute path")
    endif()
    if(NOT EXISTS "${_hook}")
      message(FATAL_ERROR "THEROCK_SUBPROJECT_CMAKE_INCLUDES: '${_hook}' does not exist")
    endif()
  endforeach()
```

cmake/therock_subproject.cmake
```diff
@@ -712,11 +712,16 @@ function(therock_cmake_subproject_activate target_name)
...
+
   # Handle compiler toolchain.
   set(_compiler_toolchain_addl_depends)
   set(_compiler_toolchain_init_contents)
   _therock_cmake_subproject_setup_toolchain("${target_name}"
-    "${_compiler_toolchain}" "${_cmake_project_toolchain_file}")
+    "${_compiler_toolchain}" "${_cmake_project_toolchain_file}"
+    "${_subproject_top_level_includes}")
   list(APPEND _fprint_files "${_cmake_project_toolchain_file}")
```
marbre, line 735:
```suggestion
  list(APPEND _fprint_files ${THEROCK_SUBPROJECT_CMAKE_INCLUDES})

```
I think the hook file contents also need to be fingerprinted. If a user edits `edits /abs/path/my_hook.cmake` the toolchain file would be still byte-to-byute identical and subproject configure would be skipped even though includes have been altered.

marbre (changes requested):
Looks mostly good to me, just two things remaining to make this more robust.

cmake/therock_subproject.cmake
PeterCDMcLean: Added the sanity check

cmake/therock_subproject.cmake
PeterCDMcLean: Added the hooks to the fingerprint
I didn't add `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` here as it should not affect the subprojects. Of course it _could_ but we cannot cover every case  of back-door CMake 
:)

cmake/therock_subproject.cmake
```diff
@@ -853,7 +853,15 @@ function(therock_cmake_subproject_activate target_name)
     string(APPEND _init_contents "set(THEROCK_USER_POST_HOOK \"@_post_hook_path@\")\n")
   endif()
   set(_global_post_include "${THEROCK_SOURCE_DIR}/cmake/therock_global_post_subproject.cmake")
-  string(APPEND _init_contents "cmake_language(DEFER CALL include \"@_global_post_include@\")\n")
+  # The subproject toolchain sets CMAKE_PROJECT_TOP_LEVEL_INCLUDES, which try_compile
```
PeterCDMcLean: @marbre 
Moving the include hook from `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` to the `_cmake_project_init_file` file caused some `try_compile` to fail. The  `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` does not affect `try_compile` so after migrating to the init file some `try_compile` started to fail. This is one of the ways to fix this issue without backing out the project init file change. Do you forsee any issues with doing this? I'm not as familiar with what the post_include does but I can't imagine it being required for try_compile.

marbre: I wasn't fully sure either and had a longer discussion with 2 different agents. It turned out that we should rather guard the deferred global post hook with `IN_TRY_COMPILE` instead of `CMAKE_PROJECT_NAME STREQUAL "CMAKE_TRY_COMPILE"`. The current check fixes compiler ABI detection, but whole-project `try_compile(...)` still replays `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` under the probe project’s own name, so the hook still fires there. `IN_TRY_COMPILE` covers both forms.

See my suggestion below.

cmake/therock_subproject.cmake
```diff
@@ -839,7 +853,15 @@ function(therock_cmake_subproject_activate target_name)
...
+  # re-reads during compiler detection, so this init file also runs inside the scratch
+  # project. Do not schedule the global post hook there: it references targets that only
+  # exist in the real subproject (aborting the compiler check) and its rpath/debug-split
+  # fixups are meaningless for a throwaway project. The scratch project is named
+  # CMAKE_TRY_COMPILE.
+  string(APPEND _init_contents "if(NOT CMAKE_PROJECT_NAME STREQUAL \"CMAKE_TRY_COMPILE\")\n")
+  string(APPEND _init_contents "  cmake_language(DEFER CALL include \"@_global_post_include@\")\n")
```
marbre:
```suggestion
  # The subproject toolchain sets CMAKE_PROJECT_TOP_LEVEL_INCLUDES, so the
  # generated init file is also replayed inside nested try_compile()/try_run()
  # scratch-project configures, not just the real subproject configure.
  #
  # Do not schedule the global post hook there: it may reference targets that
  # only exist in the real subproject, and its rpath/debug-split/stamp-touch
  # side effects are meaningless for throwaway probe builds.
  string(APPEND _init_contents "get_property(_therock_in_try_compile GLOBAL PROPERTY IN_TRY_COMPILE)\n")
  string(APPEND _init_contents "if(NOT _therock_in_try_compile)\n")
  string(APPEND _init_contents "  cmake_language(DEFER CALL include \"@_global_post_include@\")\n")
  string(APPEND _init_contents "endif()\n")
```

cmake/therock_subproject.cmake
marbre, line 1029:
```suggestion
        "${_global_post_include}"
        ${THEROCK_SUBPROJECT_CMAKE_INCLUDES}
```
Eventually we want to add `THEROCK_SUBPROJECT_CMAKE_INCLUDES` here but I am not fully sure and would need to take a closer look. Can also be as part of a follow up but we should file an issue if we decide to not address on this PR.

marbre: Suggestion doesn't seem to apply well here via GH UI, should be a separate commit if at all.

marbre:
Please address the merge conflicts or move packaging related changes to a separate PR.

cmake/therock_subproject.cmake
PeterCDMcLean: I applied this change. A small note, and outside the scope of this PR, but the project init file / toolchain file should be created by a target and that target dependency should be what is under `DEPENDS`. That will create a proper dependency chain so that this would not have been required (since  THEROCK_SUBPROJECT_CMAKE_INCLUDES is only an input into the toolchain file)

cmake/therock_subproject.cmake
PeterCDMcLean: Applied suggestion

PR #5730, opened 2026-06-09
build: add rust 1.95 to the build images

amd-arosa:
## Motivation

Mirage needs Rust to build the UX for our emulators.

## Technical Details

- Add `dockerfiles/install_rust.sh`: installs the Rust toolchain via `rustup` (currently `1.95.0`).

## Test Plan

Ran both scripts inside the actual pinned manylinux build image and verified:
1. Both toolchains install and the binaries resolve on `PATH` from a fresh login shell.
2. The installs do not make the builds non-portable.

## Test Result

- Toolchains install cleanly: `rustc 1.95.0` / `cargo 1.95.0
- Portability preserved:
  - No new files in `/usr/lib`, `/usr/lib64`, `/usr/local/lib(64)`, or
    `/etc/ld.so.conf.d`.
  - `ld.so.cache` unchanged (no new libs registered with the dynamic linker).
  - No Rust`pkg-config` `.pc` files added.
  - `/usr/local/bin` contains only binary symlinks, no `.so` files.
  - All toolchain shared libraries are confined to `RUSTUP_HOME`/`CARGO_HOME`, which are not on the default linker path; `LIBRARY_PATH` is unset
    and `LD_LIBRARY_PATH` references only the gcc-toolset.

## Submission Checklist

### DIFF
--- a/dockerfiles/build_manylinux_x86_64.Dockerfile
+++ b/dockerfiles/build_manylinux_x86_64.Dockerfile
@@ -57,6 +57,14 @@ ENV GOOGLE_TEST_VERSION="1.16.0"
 COPY install_googletest.sh ./
 RUN ./install_googletest.sh "${GOOGLE_TEST_VERSION}" && rm -rf /install-googletest
 
+######## Rust (rustup) ########
+ENV RUSTUP_HOME="/usr/local/rustup"
+ENV CARGO_HOME="/usr/local/cargo"
+WORKDIR /install-rust
+ENV RUST_VERSION="1.95.0"
+COPY install_rust.sh ./
+RUN ./install_rust.sh "${RUST_VERSION}" && rm -rf /install-rust
+
 ######## Yum Packages #######
 # We are pinning to gcc-toolset-12 until it is safe to upgrade. The latest
 # manylinux containers use gcc-toolset-14 or later, which is not yet compatible
--- a/dockerfiles/install_rust.sh
+++ b/dockerfiles/install_rust.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Install the Rust toolchain via rustup.
+#
+# Usage: ./install_rust.sh <TOOLCHAIN_VERSION>
+# Example: ./install_rust.sh "1.88.0"
+#          ./install_rust.sh "stable"
+#
+# Installs rustup/cargo to a shared location (RUSTUP_HOME/CARGO_HOME) and
+# symlinks the toolchain binaries into /usr/local/bin so they are on PATH
+# for all users.
+
+set -euo pipefail
+
+RUST_VERSION="${1:-stable}"
+
+export RUSTUP_HOME="/usr/local/rustup"
+export CARGO_HOME="/usr/local/cargo"
+
+echo "Installing Rust toolchain '${RUST_VERSION}' via rustup..."
+curl --silent --fail --show-error --location \
+    "https://sh.rustup.rs" \
+    --output rustup-init.sh
+
+sh rustup-init.sh -y \
+    --no-modify-path \
+    --profile minimal \
+    --default-toolchain "${RUST_VERSION}"
+
+rm -f rustup-init.sh
+
+# Make the toolchain available system-wide.
+chmod -R a+rwX "${RUSTUP_HOME}" "${CARGO_HOME}"
+for bin in "${CARGO_HOME}/bin/"*; do
+    ln -sf "${bin}" "/usr/local/bin/$(basename "${bin}")"
+done
+
+echo "rust installed successfully:"
+rustc --version
+cargo --version


amd-arosa:
image build
https://github.com/ROCm/TheRock/actions/runs/27231082955
https://github.com/ROCm/TheRock/actions/runs/27231091511

dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
ScottTodd, line 1: Uhhh I don't think this RCCL dockerfile is actually in use. I think we can remove it. You can at least leave it out of your changes in this PR

I see no references across the org: https://github.com/search?q=org%3AROCm%20therock_build_manylinux_rccl_x86_64&type=code compared to many references for the non-rccl dockerfile: https://github.com/search?q=org%3AROCm+therock_build_manylinux_x86_64&type=code

Added in https://github.com/ROCm/TheRock/pull/1713 but never used? https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_rccl_x86_64 has a total of 81 downloads *total* compared to 404k (yes, thousand) for https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_x86_64

cc @araravik-psd

dockerfiles/build_manylinux_x86_64.Dockerfile
```diff
@@ -57,6 +57,21 @@ ENV GOOGLE_TEST_VERSION="1.16.0"
...
+ENV RUST_VERSION="1.95.0"
+COPY install_rust.sh ./
+RUN ./install_rust.sh "${RUST_VERSION}" && rm -rf /install-rust
+
+######## Node.js (nvm) ########
+ENV NVM_DIR="/usr/local/nvm"
+WORKDIR /install-node
+ENV NODE_VERSION="22"
+COPY install_node.sh ./
+RUN ./install_node.sh "${NODE_VERSION}" && rm -rf /install-node
```
ScottTodd: I checked the binary size for this by summing the manifest layers from
1. https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_x86_64/917780540?tag=main (baseline)
2. https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_x86_64/930506497?tag=users-arosa-rust (this PR)

Looks like about 1380835760 bytes before, 1687952224 bytes after (1.38Gb --> 1.69Gb, so a 22% increase). I think that's tolerable, but we could consider splitting to a separate dockerfile for Mirage to use. I know that @stellaraccident and a few others may also want Bazel added to a build image. Do we want to have a single image for more tools, or separate images with subsets?

dockerfiles/build_manylinux_x86_64.Dockerfile
```diff
@@ -57,6 +57,21 @@ ENV GOOGLE_TEST_VERSION="1.16.0"
...
 
+######## Rust (rustup) ########
+ENV RUSTUP_HOME="/usr/local/rustup"
+ENV CARGO_HOME="/usr/local/cargo"
+WORKDIR /install-rust
+ENV RUST_VERSION="1.95.0"
+COPY install_rust.sh ./
+RUN ./install_rust.sh "${RUST_VERSION}" && rm -rf /install-rust
+
+######## Node.js (nvm) ########
```
ScottTodd: > Mirage needs Rust + Node to build the UX for our emulators.

Checking my understanding of https://github.com/ROCm/rocm-systems/blob/develop/emulation/mirage/README.md, the "UX" here is a CLI and a web interface / dashboard? Not a native GUI application? For a native GUI application it would be tricky to produce portable Linux binaries even by building under manylinux. If just CLI + web then that should be fine, easy enough to build and this dockerfile could indeed be appropriate to use.

dockerfiles/install_node.sh
```diff
@@ -0,0 +1,45 @@
...
+NODE_VERSION="${1:-22}"
+NVM_VERSION="${2:-v0.40.3}"
+
+export NVM_DIR="/usr/local/nvm"
+mkdir -p "${NVM_DIR}"
+
+echo "Installing nvm ${NVM_VERSION}..."
+curl --silent --fail --show-error --location \
+    "https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh" \
+    | bash
```
ScottTodd: This looks correct - it is what https://github.com/nvm-sh/nvm#install--update-script suggests

(we review install steps like this carefully to watch for security issues)

dockerfiles/install_rust.sh
```diff
@@ -0,0 +1,42 @@
...
+
+RUST_VERSION="${1:-stable}"
+
+export RUSTUP_HOME="/usr/local/rustup"
+export CARGO_HOME="/usr/local/cargo"
+
+echo "Installing Rust toolchain '${RUST_VERSION}' via rustup..."
+curl --silent --fail --show-error --location \
+    "https://sh.rustup.rs" \
+    --output rustup-init.sh
```
ScottTodd, line 25: This also looks correct, per https://rustup.rs/ and https://rust-lang.org/tools/install/

dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
ScottTodd: Sent https://github.com/ROCm/TheRock/pull/5761 to delete it

dockerfiles/build_manylinux_x86_64.Dockerfile
amd-arosa: it is just cli + webui
they are portable.

dockerfiles/build_manylinux_rccl_x86_64.Dockerfile
amd-arosa: ill remove it when #5761 gets merged (or #5761 will delete it after this gets merged)

dockerfiles/build_manylinux_x86_64.Dockerfile
stellaraccident: Rust will be getting more use within rocm. It is just a price we will have to pay. Since the standard runners aren't doing d-in-d anymore, my expectation is that these will ~always get cached quickly on all build nodes, no?

dockerfiles/install_node.sh
```diff
@@ -0,0 +1,45 @@
...
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# Install Node.js via nvm (Node Version Manager).
+#
+# Usage: ./install_node.sh <NODE_VERSION> [NVM_VERSION]
+# Example: ./install_node.sh "22"
+#
+# Installs nvm to a shared location (NVM_DIR) and symlinks the active
+# node/npm/npx binaries into /usr/local/bin so they are on PATH for all users.
```
stellaraccident: I want to be very-very careful about introducing node dependencies into the build via npm. Can we coordinate offline as to whether this is the right thing to do? Once we add it to the base docker, it will get used, and the supply chain for NPM is far from secure.

If you'd like to land the rust changes now, I have no issue with those and we don't need to talk about them more.

stellaraccident (approved):
Approved for adding the rust deps. I would like to discuss the npm situation further. You can land this if you back out the npm changes for a followup.

dockerfiles/build_manylinux_x86_64.Dockerfile
ScottTodd: I think they'll get cached and loaded quickly. @amd-shiraz could you confirm? (The risk here would be if we add 300MB to the dockerfile and it takes ... say 5 seconds more to download/init across 1000s of workflow runs per week that would add up)

dockerfiles/install_node.sh
amd-arosa: removed node, and disabling the mirage webui.

dockerfiles/build_manylinux_x86_64.Dockerfile
amd-arosa: removed node so it's like 100MB smaller.

Issue #1944, opened 2025-10-27
[rocWMMA] GFX Targets not supported


bsyrowik:
The following GFX architectures are not currently supported by the rocWMMA library:

- gfx906  (incompatible)
- gfx101X (incompatible)
- gfx103X (incompatible)
- gfx1103 (added in https://github.com/ROCm/rocm-libraries/pull/2301; availability expected in a future ROCm release)
- gfx1152 (added in https://github.com/ROCm/rocm-libraries/pull/2850; availability expected in ROCm 7.11)
- gfx1153 (added in https://github.com/ROCm/rocm-libraries/pull/2850; availability expected in ROCm 7.11)

The gfx906 (MI50), gfx101X, and gfx103X devices are fundamentally incompatible with rocWMMA, and as such will not be supported.    Support for gfx1103 may be added some time in the future.

oscarbg:
curious, I assume gfx115x and gfx120x are already supported right?

bsyrowik:
> curious, I assume gfx115x and gfx120x are already supported right?

Yes, that is correct.

Djip007:
gfx1103 was added a few day ago: https://github.com/ROCm/TheRock/blob/main/RELEASES.md#rocm-for-gfx110x-all

(for now there is a probleme with pytorch distribution)

[edit]: if most rocm lib look to work , rocWMMA still not usable with gfx1103 (at least not with llama.cpp)...)

it1984-tea:
Hi @bsyrowik 

Firstly, nice work with the team on therock, it's shaping up nicely.

Perhaps you are already aware, a massive number of gfx906 GPUs ended up with homelabbers. I'm guessing some datacenter in china offloaded their cards or something and now many people have them. They make excellent inference cards.

MI50s are mentioned almost weekly on local LLM communities online.

I understand what I said above might be irrelevant to AMD, but look at this:
https://www.reddit.com/r/LocalLLaMA/comments/1ns2fbl/for_llamacppggml_amd_mi50s_are_now_universally/

Because these cards are everywhere, users contribute optimizations in projects that indirectly improve performance for other AMD cards. By extending the usefulness/unlocking the full performance of these cards, you make them more relevant = more code contributions. I hope this will nudge the decision in the right direction. Please add rocWMMA support for gfx906.

bsyrowik:
> Hi [@bsyrowik](https://github.com/bsyrowik)
> 
> Firstly, nice work with the team on therock, it's shaping up nicely.
[...]

PR #5744, opened 2026-06-10
Add gfx1250 (CDNA5) build support

rahulc-gh:
## Motivation

   This PR adds initial build support for **gfx1250**, a **CDNA5** architecture GPU (GFX IP version 12.5.0). This is a build-only bringup — no test hardware is available yet.

   ## Changes

   ### `cmake/therock_amdgpu_targets.cmake`

   Registered `gfx1250` as a new CMake AMDGPU target — the single source of truth from which all other build system logic derives. The target is classified as a data center GPU
   (`dcgpu-all`) and placed in the `gfx125X-all` and `gfx125X-dcgpu` families, following the same pattern as `gfx94X-dcgpu` (MI300) and `gfx950-dcgpu` (MI350). `rocWMMA` is
   excluded via `EXCLUDE_TARGET_PROJECTS` ([LCOMPILER-2316](https://amd-hub.atlassian.net/browse/LCOMPILER-2316)).

   ### `build_tools/hack/env_check/AMDGPU_LLVM_TARGET.py`

   Added the CDNA5/gfx125X product name mapping used by environment check tooling to map LLVM target names to human-readable product names.

   ### `build_tools/github_actions/amdgpu_family_matrix.py`

   Added a `gfx125x` entry to the **nightly** CI matrix. Configured as:

   - Linux build-only (`test-runs-on: ""`, `fetch-gfx-targets: []`)
   - No Windows entry — no hardware or demand for Windows builds at this stage

   Placing gfx1250 in the nightly matrix (rather than presubmit) means ROCm stack builds run on the **nightly schedule** and via **`workflow_dispatch`** only — no per-PR or
   per-push CI cost while the target is in early bringup.

   ### `build_tools/github_actions/new_amdgpu_family_matrix.py`

   Two changes to keep this file in sync with `amdgpu_family_matrix.py`:

   - Added `gfx125X-dcgpu` to the `amdgpu_nightly` predefined group
   - Added a `gfx125X` / `dcgpu` matrix entry: Linux build-only, `run_tests: False`, empty `runs_on` and `fetch-gfx-targets`, `bypass_tests_for_releases: True`, `push_on_success:
   False`

   ### `build_tools/github_actions/configure_pytorch_release_matrix.py`

   This is now the single place that controls which GPU families are included in PyTorch builds for each PyTorch ref. Each entry in `PYTORCH_REFS_LINUX` / `PYTORCH_REFS_WINDOWS`
   carries an `exclude_amdgpu_families` key. A shared `_PYTORCH_EXCLUDED_FAMILIES` set holds families whose upstream PyTorch support is not yet merged (currently `gfx125x`). The
   script filters the incoming `--amdgpu-families` list per ref and emits `amdgpu_families` as a new column in every matrix row; build jobs consume `matrix.amdgpu_families` instead
    of `inputs.amdgpu_families`.

   ### `multi_arch_release_linux_pytorch_wheels.yml`, `multi_arch_release_windows_pytorch_wheels.yml`

   Both workflows now pass `--amdgpu-families` to `configure_pytorch_release_matrix.py` and use `matrix.amdgpu_families` in the build job. The ROCm stack build is unaffected and
   continues to build all families including gfx1250.

   ## What is intentionally excluded

   - **Per-PR / per-push builds** — gfx1250 is in the nightly matrix only. Use `workflow_dispatch` on `multi_arch_ci.yml` with `linux_amdgpu_families: gfx125x` to trigger manually.
   - **PyTorch builds** — excluded via `_PYTORCH_EXCLUDED_FAMILIES` in `configure_pytorch_release_matrix.py` until upstream PyTorch support lands. See #5833.
   - **Windows build** — omitted from both CI matrix files until there is a concrete need.
   - **Test runner labels** — left empty everywhere (`test-runs-on: ""`); to be filled in once gfx1250 hardware is available.
   - **Sanitizer xnack+ support** — not added to `cmake/therock_sanitizers.cmake`.

   ## Follow-up tasks (not in this PR)

   - [ ] Enable test runner once gfx1250 hardware is provisioned
   - [ ] Remove `gfx125x` from `_PYTORCH_EXCLUDED_FAMILIES` in `configure_pytorch_release_matrix.py` once upstream PyTorch support lands (#5833)
   - [ ] Add Windows build support if/when needed

[LCOMPILER-2316]: https://amd-hub.atlassian.net/browse/LCOMPILER-2316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -231,11 +231,13 @@ jobs:
       actions: write
     steps:
       - name: Trigger PyTorch Build
+        # pytorch_amdgpu_families excludes families whose upstream PyTorch support
+        # is not yet merged (e.g. gfx125X). See https://github.com/ROCm/TheRock/issues/5833.
         uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
         with:
           workflow: multi_arch_release_linux_pytorch_wheels.yml
           inputs: |
-            { "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
+            { "amdgpu_families": "${{ fromJSON(inputs.build_config).pytorch_amdgpu_families }}",
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -231,13 +231,11 @@ jobs:
       actions: write
     steps:
       - name: Trigger PyTorch Build
-        # pytorch_amdgpu_families excludes families whose upstream PyTorch support
-        # is not yet merged (e.g. gfx125X). See https://github.com/ROCm/TheRock/issues/5833.
         uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
         with:
           workflow: multi_arch_release_linux_pytorch_wheels.yml
           inputs: |
-            { "amdgpu_families": "${{ fromJSON(inputs.build_config).pytorch_amdgpu_families }}",
+            { "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
               "release_type": "${{ inputs.release_type }}",
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -406,13 +406,13 @@ amdgpu_family_info_matrix_nightly = {
     "gfx125x": {
         "linux": {
             # No hardware available for testing yet; build-only.
-            # PyTorch builds are skipped until gfx1250 support is upstreamed
-            # to pytorch/pytorch. See https://github.com/ROCm/TheRock/issues/5833.
+            # PyTorch builds are included — workflow_dispatch can be used
+            # to trigger manually; nightly schedule runs both ROCm stack
+            # and PyTorch builds.
             "test-runs-on": "",
             "family": "gfx125X-dcgpu",
             "fetch-gfx-targets": [],
             "build_variants": ["release"],
-            "skip_pytorch": True,
         },
     },
 }
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -406,13 +406,13 @@ amdgpu_family_info_matrix_nightly = {
     "gfx125x": {
         "linux": {
             # No hardware available for testing yet; build-only.
-            # PyTorch builds are included — workflow_dispatch can be used
-            # to trigger manually; nightly schedule runs both ROCm stack
-            # and PyTorch builds.
+            # PyTorch builds are skipped until gfx1250 support is upstreamed
+            # to pytorch/pytorch. See https://github.com/ROCm/TheRock/issues/5833.
             "test-runs-on": "",
             "family": "gfx125X-dcgpu",
             "fetch-gfx-targets": [],
             "build_variants": ["release"],
+            "skip_pytorch": True,
         },
     },
 }
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -245,6 +245,16 @@ amdgpu_family_info_matrix_presubmit = {
             "nightly_check_only_for_family": True,
         },
     },
+    "gfx125x": {
+        "linux": {
+            # No hardware available for testing yet; build-only nightly.
+            "test-runs-on": "",
+            "family": "gfx125X-dcgpu",
+            "fetch-gfx-targets": [],
+            "build_variants": ["release"],
+            "nightly_check_only_for_family": True,
+        },
+    },
 }
 
 # The 'postsubmit' matrix runs on 'push' triggers (for every commit to the default branch).
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -245,16 +245,8 @@ amdgpu_family_info_matrix_presubmit = {
             "nightly_check_only_for_family": True,
         },
     },
-    "gfx125x": {
-        "linux": {
-            # No hardware available for testing yet; build-only.
-            "test-runs-on": "",
-            "family": "gfx125X-dcgpu",
-            "fetch-gfx-targets": [],
-            "build_variants": ["release"],
-        },
-    },
 }
+
 
 # The 'postsubmit' matrix runs on 'push' triggers (for every commit to the default branch).
 amdgpu_family_info_matrix_postsubmit = {
@@ -407,6 +399,18 @@ amdgpu_family_info_matrix_nightly = {
         "windows": {
             "test-runs-on": "",
             "family": "gfx1153",
+            "fetch-gfx-targets": [],
+            "build_variants": ["release"],
+        },
+    },
+    "gfx125x": {
+        "linux": {
+            # No hardware available for testing yet; build-only.
+            # PyTorch builds are included — workflow_dispatch can be used
+            # to trigger manually; nightly schedule runs both ROCm stack
+            # and PyTorch builds.
+            "test-runs-on": "",
+            "family": "gfx125X-dcgpu",
             "fetch-gfx-targets": [],
             "build_variants": ["release"],
         },
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -247,12 +247,11 @@ amdgpu_family_info_matrix_presubmit = {
     },
     "gfx125x": {
         "linux": {
-            # No hardware available for testing yet; build-only nightly.
+            # No hardware available for testing yet; build-only.
             "test-runs-on": "",
             "family": "gfx125X-dcgpu",
             "fetch-gfx-targets": [],
             "build_variants": ["release"],
-            "nightly_check_only_for_family": True,
         },
     },
 }
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -421,6 +421,10 @@ class BuildConfig:
 
     per_family_info: list[dict]  # Per-family metadata for test/artifact jobs
     dist_amdgpu_families: str  # Semicolon-separated
+    # Subset of dist_amdgpu_families for PyTorch builds; excludes families
+    # whose upstream PyTorch support is not yet merged. See:
+    # https://github.com/ROCm/TheRock/issues/5833
+    pytorch_amdgpu_families: str  # Semicolon-separated
     artifact_group: str
     build_variant_label: str
     build_variant_suffix: str
@@ -848,6 +852,7 @@ def _expand_build_config_for_platform(
             break
 
     per_family_info: list[dict] = []
+    pytorch_family_names: list[str] = []
     for family_name in families:
         # select_targets already validates family names and filters by
         # platform availability. Family name uniqueness is validated by
@@ -946,6 +951,8 @@ def _expand_build_config_for_platform(
                 ),
             }
         )
+        if not platform_info.get("skip_pytorch", False):
+            pytorch_family_names.append(platform_info["family"])
 
     if not per_family_info:
         return None
@@ -961,6 +968,7 @@ def _expand_build_config_for_platform(
     return BuildConfig(
         per_family_info=per_family_info,
         dist_amdgpu_families=";".join(family_names),
+        pytorch_amdgpu_families=";".join(pytorch_family_names),
         artifact_group=f"multi-arch-{suffix or 'release'}",
         build_variant_label=variant_config["build_variant_label"],
         build_variant_suffix=suffix,
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -421,10 +421,6 @@ class BuildConfig:
 
     per_family_info: list[dict]  # Per-family metadata for test/artifact jobs
     dist_amdgpu_families: str  # Semicolon-separated
-    # Subset of dist_amdgpu_families for PyTorch builds; excludes families
-    # whose upstream PyTorch support is not yet merged. See:
-    # https://github.com/ROCm/TheRock/issues/5833
-    pytorch_amdgpu_families: str  # Semicolon-separated
     artifact_group: str
     build_variant_label: str
     build_variant_suffix: str
@@ -852,7 +848,6 @@ def _expand_build_config_for_platform(
             break
 
     per_family_info: list[dict] = []
-    pytorch_family_names: list[str] = []
     for family_name in families:
         # select_targets already validates family names and filters by
         # platform availability. Family name uniqueness is validated by
@@ -951,8 +946,6 @@ def _expand_build_config_for_platform(
                 ),
             }
         )
-        if not platform_info.get("skip_pytorch", False):
-            pytorch_family_names.append(platform_info["family"])
 
     if not per_family_info:
         return None
@@ -968,7 +961,6 @@ def _expand_build_config_for_platform(
     return BuildConfig(
         per_family_info=per_family_info,
         dist_amdgpu_families=";".join(family_names),
-        pytorch_amdgpu_families=";".join(pytorch_family_names),
         artifact_group=f"multi-arch-{suffix or 'release'}",
         build_variant_label=variant_config["build_variant_label"],
         build_variant_suffix=suffix,
--- a/build_tools/github_actions/configure_pytorch_release_matrix.py
+++ b/build_tools/github_actions/configure_pytorch_release_matrix.py
@@ -17,33 +17,107 @@ from github_actions.github_actions_api import gha_set_output
 
 PYTHON_VERSIONS = ["3.10", "3.11", "3.12", "3.13", "3.14"]
 
-PYTORCH_REFS_LINUX = [
-    "release/2.9",
-    "release/2.10",
-    "release/2.11",
-    "release/2.12",
-    "nightly",
+# PyTorch git refs per platform.
+# Each ref may optionally carry an `exclude_amdgpu_families` set: GPU families
+# that should be omitted for this ref (e.g. because upstream support is not yet
+# merged). When `exclude_amdgpu_families` is absent the ref inherits whatever
+# families are passed in via --amdgpu-families.
+#
+# Format: list of dicts with keys:
+#   pytorch_git_ref        – git branch/tag passed to the build job
+#   exclude_amdgpu_families – (optional) set[str] of family names to drop
+# GPU families not yet upstreamed to pytorch/pytorch. Excluded from all PyTorch
+# ref builds until the respective upstream PRs merge.
+# See https://github.com/ROCm/TheRock/issues/5833.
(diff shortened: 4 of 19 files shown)


rahulc-gh:
runtime compilation fix for gfx1250 - https://github.com/ROCm/rocm-systems/pull/7024 (followed by promotion is needed)

geomin12 (commented):
pre-commit and unit tests failing

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -245,6 +245,15 @@ def select_build_runner(platform: str, build_variant: str) -> str:
...
     },
+    "gfx125x": {
+        "linux": {
+            # No hardware available for testing yet; build-only.
+            "test-runs-on": "",
+            "family": "gfx125X-dcgpu",
+            "fetch-gfx-targets": [],
+            "build_variants": ["release"],
+        },
+    },
```
ScottTodd: Is "presubmit" the matrix that you want to add this to? Each target [family] that we add to presubmit adds build time to _all_ workflow runs. We could add to postsubmit or nightly to decrease CI load and add an opt-in `gfx1250` / `gfx125X` label for use with https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request.

(I think this will be important enough to make a case for presubmit, but we may want to load balance by limiting other jobs to make room for it)

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -245,6 +245,15 @@ def select_build_runner(platform: str, build_variant: str) -> str:
             "nightly_check_only_for_family": True,
         },
     },
+    "gfx125x": {
+        "linux": {
+            # No hardware available for testing yet; build-only.
+            "test-runs-on": "",
+            "family": "gfx125X-dcgpu",
```
ScottTodd: Adding the build to the CI like this (and not just the build system with the `therock_amdgpu_targets.cmake` changes) will start building rocm artifacts and pytorch packages as part of our nightly releases. If this fails to build in pytorch for some reason, we'll want a way to exclude from the pytorch build.

Can you trigger a dev release from this branch with at least `gfx94X,gfx125X` in the Linux target families list? See https://github.com/ROCm/TheRock/actions/workflows/multi_arch_release.yml, and check "Build PyTorch wheels"

Inputs should be roughly:

Image

cmake/therock_amdgpu_targets.cmake
```diff
@@ -242,6 +242,12 @@ therock_add_amdgpu_target(gfx1201 "AMD RX 9070 / XT" FAMILY dgpu-all gfx120X-all
     rocprofiler-compute # https://github.com/ROCm/TheRock/issues/2892
 )
 
+# gfx125X family
+therock_add_amdgpu_target(gfx1250 "AMD Instinct MI450/MI450X/MI455X CDNA" FAMILY dcgpu-all gfx125X-all gfx125X-dcgpu
+  EXCLUDE_TARGET_PROJECTS
+    rocWMMA # https://github.com/ROCm/TheRock/issues/1944
```
ScottTodd: https://github.com/ROCm/TheRock/issues/1944 is marked "fixed". Want to re-open that issue or file a new one for gfx125X?

build_tools/github_actions/amdgpu_family_matrix.py
ScottTodd: > * **PyTorch / JAX wheel workflows** — the 8 `workflow_dispatch` YAML files with
>       `amdgpu_family` choice lists are not updated. Framework support for gfx1250
>       must be confirmed upstream before adding it to wheel build pipelines.

This is not accurate. The scheduled multi-arch release workflows use a dynamic list of _all_ registered target families. You are probably looking at old single-arch build/release workflows and manual workflow_dispatch triggers.

ScottTodd: Code:
* https://github.com/ROCm/TheRock/blob/a11c58d1becde1b2d96e4a1fa1d566d88870d6c5/.github/workflows/multi_arch_release_linux.yml#L233-L238
* https://github.com/ROCm/rockrel/blob/1b0b03a5ef34151adce4e375ca327b380bd33cce/.github/workflows/multi_arch_release.yml#L26-L29
    ```yml
          linux_amdgpu_families:
            description: 'Linux GPU families. "all" = all, "none" = skip Linux.'
            type: string
            default: "all"
    ```
* https://github.com/ROCm/TheRock/blob/a8879cb0491170195119eec4211414dec6af080a/build_tools/github_actions/configure_multi_arch_ci.py#L734-L773

build_tools/github_actions/amdgpu_family_matrix.py
rahulc-gh: We will need it in presubmit , so no regression comes in . Eg : last rocm-system promotion did not validate the gfx1250 at pre-submit and the build were broken.

The fix came rocm-system [PR 7024](https://github.com/ROCm/rocm-systems/pull/7024) . Even if it slows down we need this check at pre-commit @suchitravemuri1  and @amd-aakash  - can confirm as well

cmake/therock_amdgpu_targets.cmake
rahulc-gh: I will update the compiler ticket here which broke rocwmma

build_tools/github_actions/amdgpu_family_matrix.py
rahulc-gh: @ScottTodd  is it possible to enable only rocm artifacts (need of the hour) . Pytorch can come in later, as we are not sure if gfx1250 code is upstreamed or not

ScottTodd: Try it, see what happens. We can add a carve-out to the build system and release workflows as needed.

build_tools/github_actions/amdgpu_family_matrix.py
rahulc-gh: @ScottTodd - I get your point , we can bring in the chagne of enabling precommit later on. Made changes and moved this to nightly matrix.

Build was success with - super set PR 5799 (which is PR 5744 + rocm-systems promotion)
https://github.com/ROCm/TheRock/actions/runs/27381232863/job/80938564729?pr=5789

pytorch failure is under debug and need some more env settign for gfx1250, as per @jammm we need below env settings

USE_FLASH_ATTENTION=OFF
USE_MEM_EFF_ATTENTION=OFF
USE_ROCM_CK_SDPA=OFF
USE_ROCM_CK_GEMM=OFF

ScottTodd: FYI, we can turn off those settings for _the entire pytorch build_ here: https://github.com/ROCm/TheRock/blob/3f200b237339a63c3c2452ea83f1262404caa233/external-builds/pytorch/build_prod_wheels.py#L923-L1026

Any finer-grained filtering needs to go in triton/aotriton/pytorch directly.

For now we could unblock getting gfx1250 support in TheRock by filtering out gfx1250 from the target list in that script or some of the layers above it in workflows / CI configuration. There are a few ways to do it and I'm not sure if I have a preference right now...

rahulc-gh: Thanks for the pointers , I was planning to bring the pytorch fixes is subsequent PR as this PR needs rocm-systems bump as well . If pytorch build fails in the ci-nightly for gfx1250 for couple of days it will not impact much correct ??

So order would be Bump PR followed by PR5744, followed by PR 5825 . @ScottTodd Let me know if this works for you.

EG- https://github.com/ROCm/TheRock/pull/5825 
Running one manual dispatch as suggested below - https://github.com/ROCm/TheRock/actions/runs/27445149206

cmake/therock_amdgpu_targets.cmake
rahulc-gh: this is been updated and a copmiler promotion already in prgress, after that thsi exclude can be removed https://github.com/ROCm/theRock/pull/5806

build_tools/github_actions/amdgpu_family_matrix.py
ScottTodd: > If pytorch build fails in the ci-nightly for gfx1250 for couple of days it will not impact much correct ??

We shouldn't introduce known build failures in the multi-arch PyTorch releases, and those releases currently build rocm + torch for all target [families] that are enumerated in the family matrix file. We've gone several days without a successful nightly release for a number of reasons and we shouldn't add another reason to the pile.

https://therock-hud-dev.amd.com/#multi-arch-release
image

rahulc-gh:
> pre-commit and unit tests failing

The change does not impact any test as its enabling build compilation of new target. After the R-S bump the gfx1250 is compiling fine. and test failures should not be related.

rahulc-gh:
Some pointers here the pytorch failure seen is due to multiarch support and pytorch gfx1250 not being upstremed.

Single Arch gfx1250 Job (manual dispatch from PR branch (Which has ENV vars suggested by Aaryaman TheROCK issue https://github.com/ROCm/TheRock/issues/5833 PR - https://github.com/ROCm/TheRock/pull/5825 ) - GH action workflow [Multi-Arch Release (dev) | Linux: gfx125x | Windows: none · ROCm/TheRock@7f7be90](https://github.com/ROCm/TheRock/actions/runs/27481648249)

Rocm-artifacts: https://therock-dev-artifacts.s3.amazonaws.com/27481648249-linux/index.html
PyTorch-artifacts: https://therock-dev-artifacts.s3.amazonaws.com/27481648249-linux/python/index.html

rahulc-gh:
Test job triggered for the skip-pytorch changes - https://github.com/ROCm/TheRock/actions/runs/27566709384

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -231,11 +231,13 @@ jobs:
...
     steps:
       - name: Trigger PyTorch Build
+        # pytorch_amdgpu_families excludes families whose upstream PyTorch support
+        # is not yet merged (e.g. gfx125X). See https://github.com/ROCm/TheRock/issues/5833.
         uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
         with:
           workflow: multi_arch_release_linux_pytorch_wheels.yml
           inputs: |
-            { "amdgpu_families": "${{ fromJSON(inputs.build_config).dist_amdgpu_families }}",
+            { "amdgpu_families": "${{ fromJSON(inputs.build_config).pytorch_amdgpu_families }}",
```
ScottTodd: As gfx1250 support becomes available across pytorch release branches (including upstream `main`), I don't think this mapping with be flexible enough, as it applies to all pytorch versions: https://github.com/ROCm/TheRock/blob/67d17e221bb9d4148e12dc9c056b423a6222df47/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml#L120-L133

A few alternatives to consider:

1. Filter in https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/build_prod_wheels.py
    * positive: One spot to apply filtering, similar to what we do for `aotriton_unsupported_archs_for_version`
    * negative: Hidden behavior - workflows would say they are building for one list but then actually filter just before building
2. Filter in PyTorch itself (`PYTORCH_ROCM_ARCH` env var)
    * positive: TheRock doesn't need to do anything, different PyTorch branches can choose what they support without needing any changes here
    * negative: if we provide an explicit list of targets to build, we expect that list to be what is used (as we then want to run tests on the packages, publish them, etc.)
3. We added this [`build_tools/github_actions/configure_pytorch_release_matrix.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_pytorch_release_matrix.py) recently. That looks like a good place to decide across pytorch versions. We could emit `amdgpu_targets` to the matrix rows in addition to `python_version` and `pytorch_git_ref`
    * positive: small script focused on just choosing what to build/test, looks like the natural place for this. If we later decide to vary GPU target support across PyTorch versions we can add that to that file too.

In each case I'm wondering how we want to handle "yes, I know it won't build but I want to try anyways". I think we'll also want similar filtering of supported GPU targets per framework version for JAX too? (cc @erman-gurses )

Maybe @jithunnair-amd or @HereThereBeDragons have recommendations/opinions? I'm leaning towards (3).

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -422,6 +422,10 @@ class BuildConfig:
 
     per_family_info: list[dict]  # Per-family metadata for test/artifact jobs
     dist_amdgpu_families: str  # Semicolon-separated
+    # Subset of dist_amdgpu_families for PyTorch builds; excludes families
+    # whose upstream PyTorch support is not yet merged. See:
+    # https://github.com/ROCm/TheRock/issues/5833
+    pytorch_amdgpu_families: str  # Semicolon-separated
```
geomin12: since gfx1250 is a temporary case, could we just ignore it for pytorch via if checks in python code? thay way, we don't make any adjustments to yml files and once it is ready, we can just remove a temporary check

geomin12: should we just add a `if: ${{ inputs.amdgpu_families == gfx1250 }}` ignore to yml files instead? i don't know if we need all these changes...

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -951,6 +956,8 @@ def _expand_build_config_for_platform(
                 ),
             }
         )
+        if not platform_info.get("skip_pytorch", False):
```
geomin12: would prefer to keep `dist_amdgpu_families` as is, but use the `skip_pytorch` somehow with if checks in the pytorch workflows?

geomin12 (commented):
unit tests fail

rahulc-gh:
re-runnign the test after changes suggested in option 3 - https://github.com/ROCm/TheRock/actions/runs/27595426463

build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -17,33 +17,107 @@
...
+    },
+    {
+        "pytorch_git_ref": "release/2.12",
+        "exclude_amdgpu_families": _PYTORCH_EXCLUDED_FAMILIES,
+    },
+    {
+        "pytorch_git_ref": "nightly",
+        "exclude_amdgpu_families": _PYTORCH_EXCLUDED_FAMILIES,
+    },
 ]
```
ScottTodd: We could just inline `gfx125x` here, rather than use a `_PYTORCH_EXCLUDED_FAMILIES` constant. At the point where we start enabling gfx125x for versions the generic constant name won't be as applicable.

Then we could also drop it from windows (where we don't try to build gfx125x anyways)

build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -64,6 +138,12 @@ def main(argv: list[str] | None = None) -> int:
         choices=["linux", "windows"],
         help="Platform to generate matrix for (default: current system)",
     )
+    parser.add_argument(
+        "--amdgpu-families",
+        type=str,
+        default="",
+        help="Semicolon-separated AMD GPU families to build PyTorch for",
+    )
```
ScottTodd, line 137: Could document in the help string that families that are not supported for release versions will be filtered out of this list.

build_tools/github_actions/configure_pytorch_release_matrix.py
ScottTodd, line 1: Hmm... I think this will work for our release workflows (and it is what I prefer, so thank you), but the "CI" workflows are still just passing through `inputs.amdgpu_families` here: https://github.com/ROCm/TheRock/blob/bbd306dc169ea699b714f04e36e1ce4a24876a23/.github/workflows/build_portable_linux_pytorch_wheels_ci.yml#L201-L228

So if someone sends a PR with a gfx125X/run-all-archs label to opt-in to building on a PR, or it runs on postsubmit, the pytorch CI builds will still try to include gfx1250 and may fail.

I guess we should update those CI workflows to use this new script too, @geomin12 (I also want to update them to use the new pytorch manifest generation / checkout code I've been working on, so this could all line up together...)

ScottTodd: I'm taking a look at using this script from our CI workflows too. Not much time for development today though.

rahulc-gh: Thanks appreciate your time , please update the PR if you can.  gfx1250 support will be enabled for release/2.11 first followed by 2.12 and upstream mainline. 2.9 and 2.10 will not have the support.

rahulc-gh: I am doing some clean and address your new comments as well

build_tools/github_actions/configure_pytorch_release_matrix.py
rahulc-gh: Done the clean up.

ScottTodd:
Please get the "pre-commit" and "unit tests" passing to continue the review here. (you should be running those locally, before committing/pushing)

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -402,6 +403,18 @@ def select_build_runner(platform: str, build_variant: str) -> str:
             "build_variants": ["release"],
         },
     },
+    "gfx125x": {
```
ScottTodd, line 406: from the unit test failure:
https://github.com/ROCm/TheRock/actions/runs/27642913439/job/81747324777?pr=5744#step:5:1186
`FAILED github_actions/tests/fetch_package_targets_test.py::FetchPackageTargetsTest::test_amdgpu_families_accepted_by_choice_workflows - AssertionError: build_portable_linux_pytorch_wheels.yml is missing amdgpu_family options that fetch_package_targets can produce: ['gfx125X-dcgpu']`

I'll send a PR deleting https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_windows_pytorch_wheels.yml

* We're using https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml now
* We don't need that explicit list here: https://github.com/ROCm/TheRock/blob/ec7bdfd5f3e3704e1e52dfe2a60bc6eed8f462a5/.github/workflows/build_windows_pytorch_wheels.yml#L54-L72

ScottTodd: Sent https://github.com/ROCm/TheRock/pull/5892 to delete some old workflows

rahulc-gh: to get the test passing I added it to missing files but without your PR5892, the gate would fail here 

Fix: Added gfx125X-dcgpu to the amdgpu_family choice options in:                                                                                           
  - .github/workflows/build_portable_linux_pytorch_wheels.yml                                                                                                
  - .github/workflows/build_windows_pytorch_wheels.yml                                                                                                       
  - .github/workflows/copy_release.yml                                                                                                                       
  - .github/workflows/test_linux_jax_wheels.yml

ScottTodd: Yeah, it's fine to add the files that currently require it here.

ScottTodd (approved):
This is probably going to break postsubmit CI (https://github.com/ROCm/TheRock/actions/workflows/multi_arch_ci.yml?query=branch%3Amain) by trying to build torch release/2.10 for gfx1250. If we're comfortable with that this looks okay to merge.

(I'm working on changes to get CI and releases in alignment with what they build/test but that's not going to be a fast and low risk change)

ScottTodd:
Actually if it's just `amdgpu_family_info_matrix_nightly` that should be fine and not break postsubmit CI, until someone adds `ci:run-all-archs` or a `gfx120X` label to a PR.

rahulc-gh:
Manual trigger for nightly looks good - https://github.com/ROCm/TheRock/actions/runs/27650468161/job/81804470180 
Building for gfx125x after compiler and R-S bump.

Also downstream pytroch seems to have filtered our the gfx125x for all pytorch release branch except 2.11
https://github.com/ROCm/TheRock/actions/runs/27660781694/job/81804537934

Going ahead with merge. Thanks @ScottTodd  for all feedback

build_tools/github_actions/configure_pytorch_release_matrix.py
zjliu-amd: Hi, do you think the later clean up commit missed the `exclude_amdgpu_families` entry for pytorch 2.11? The latest nightly build of 2.11 failed: https://github.com/ROCm/rockrel/actions/runs/27674732763/job/81847294581

Issue #5737, opened 2026-06-09
[Issue] PyTorch "Configure AWS Credentials for sccache" steps failing on PRs from forks


ScottTodd:
On https://github.com/ROCm/TheRock/pull/5729,
* https://github.com/ROCm/TheRock/actions/runs/27226628166/job/80428068877?pr=5729#step:11:1
* https://github.com/ROCm/TheRock/actions/runs/27226628166/job/80428068813?pr=5729#step:11:1
* https://github.com/ROCm/TheRock/actions/runs/27226628166/job/80428069036?pr=5729#step:11:1

```
It looks like you might be trying to authenticate with OIDC. Did you mean to set the `id-token` permission? If you are not trying to authenticate with OIDC and the action is working successfully, you can ignore this message.
Assuming role with user credentials
Retry AssumeRole: attempt 1 of 12 failed: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci. Retrying after 43ms.
Assuming role with user credentials
Retry AssumeRole: attempt 2 of 12 failed: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci. Retrying after 79ms.
Assuming role with user credentials
Retry AssumeRole: attempt 3 of 12 failed: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci. Retrying after 110ms.
Assuming role with user credentials
Retry AssumeRole: attempt 4 of 12 failed: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci. Retrying after 10ms.
Assuming role with user credentials
Retry AssumeRole: attempt 5 of 12 failed: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci. Retrying after 270ms.

```

[...]

PR #5729, opened 2026-06-09
Expand Linux PyTorch presubmit matrix to include release/2.11 and release/2.12

ScottTodd:
## Motivation

We recently had a build break that only affected torch versions release/2.11+:
* https://github.com/ROCm/TheRock/issues/5714

Expanding build coverage on presubmit to catch these issues earlier is part of:

* https://github.com/ROCm/TheRock/issues/3291
* https://github.com/ROCm/TheRock/issues/3177

## Technical Details

> [!IMPORTANT]
> This is expected to add around 40-60 minutes of build runner usage _per pytorch version_ to CI workflow runs. These builds may run in parallel to GPU test jobs and the existing pytorch build job, so total time to signal is not expected to be impacted (other than increasing the load on our CPU runner pools).
>
> Migrating the build runners to AWS (higher core count, closer to the AWS sccache) will also improve build times.

We can choose to conditionally enable these additional jobs to limit how many workflow runs include them. References:
* https://github.com/ROCm/TheRock/pull/5326
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci.py
* https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md

I'm limiting to just Linux for now. We could also expand coverage on Windows (though those builds are slower).

## Test Plan

Watch CI on this PR
* Index: https://therock-ci-artifacts-external.s3.amazonaws.com/ROCm-TheRock/27493147226-linux/python/index.html
    ```
    torch-2.10.0+devrocm7.14.0.dev0.14f3a948a07bc0bf89c0879f2e8d154342d12d59-cp312-cp312-linux_x86_64.whl	171 MB	2026-06-14 11:57 UTC
    torch-2.11.0+devrocm7.14.0.dev0.14f3a948a07bc0bf89c0879f2e8d154342d12d59-cp312-cp312-linux_x86_64.whl	174 MB	2026-06-14 12:04 UTC
    torch-2.12.0+devrocm7.14.0.dev0.14f3a948a07bc0bf89c0879f2e8d154342d12d59-cp312-cp312-linux_x86_64.whl	178 MB	2026-06-14 11:58 UTC
    ```
* 2.10: https://github.com/ROCm/TheRock/actions/runs/27493147226/job/81271740334?pr=5729
* 2.11: https://github.com/ROCm/TheRock/actions/runs/27493147226/job/81271740345?pr=5729
[...]

Issue #5714, opened 2026-06-09
[Issue] Linux PyTorch : build fails — ncclDevComm undeclared in nccl_reduce_scatter_offset.cu (NCCL_DEVICE_HAS_REDUCE_COPY not guarded for USE_ROCM)


TriveniTadapaneni:
## Context
- Workflow: Release portable Linux PyTorch Wheels
- Workflow file: `.github/workflows/release_portable_linux_pytorch_wheels.yml`
- Failing run: [↗ View run](https://github.com/ROCm/TheRock/actions/runs/27192289742)
- Platform: `Linux`
- Impacted Arch: `gfx1152`,`gfx1150`,`gfx1153`,`gfx908`,`gfx1151`,`gfx90a`,`gfx950-dcgpu`,`gfx94X-dcgpu`,
- PyTorch Versions: `nightly`, `release/2.12`, `release/2.11`
- Python Versions: all (3.10, 3.11, 3.12, 3.13, 3.14)
- rocm_version: `7.14.0a20260609`
- RCCL version bundled in SDK: `2.29.7`

## Summary

All 15 build jobs for gfx1152 fail at the PyTorch HIP compile stage. The failure is caused by `NCCL_DEVICE_HAS_REDUCE_COPY` being defined (RCCL 2.29.7 ≥ 2.29.7) without the `!defined(USE_ROCM)` guard that was already applied to the related `NCCL_HAS_SYMMEM_DEVICE_SUPPORT` macro. This causes the compiler to enter a `#ifdef NCCL_DEVICE_HAS_REDUCE_COPY` block that uses NVIDIA-internal device-side NCCL types (`ncclDevComm`, `ncclCoopCta`, `ncclDevCommRequirements`, etc.) that RCCL does not expose.

## Error

**nightly + release/2.12** — fails in `nccl_reduce_scatter_offset.cu`:

```text
FAILED: [code=1] caffe2/CMakeFiles/torch_hip.dir/__/torch/csrc/distributed/c10d/symm_mem/ops/nccl_reduce_scatter_offset.cu.o
/__w/TheRock/TheRock/external-builds/pytorch/pytorch/torch/csrc/distributed/c10d/symm_mem/ops/nccl_reduce_scatter_offset.cu:78:5: error: unknown type name 'ncclDevComm'; did you mean 'ncclComm'?
   78 |     ncclDevComm devComm) {
[...]

PR #5745, opened 2026-06-10
 [ci]: Auto-apply `ci:run-all-archs` label to submodule bump PRs

amd-chiranjeevi:
## Motivation

 Submodule bump PRs (rocm-systems, rocm-libraries) currently only build the default presubmit families (gfx94x, gfx110x, gfx1151, gfx120x), missing build coverage for older and nightly-only families. Instead of permanently adding families to the presubmit matrix (which adds CI load to every PR), automatically apply `ci:run-all-archs` when `bump_automation.py` creates a bump PR. This gives bump PRs:
  - **Builds** across all known families (gfx900, gfx901x, gfx103x, gfx950, etc.)

  - Families with `nightly_check_only_for_family` or no hardware remain build-only

## Test Plan
  
  - Verify the next bump PR created by `bump_automation.py` has the
    `ci:run-all-archs` label applied automatically.
  - Check the CI run on that bump PR to confirm all families appear
    in the job matrix.
    
  ## Test Result

 Triggered bump automation via workflow run [#27602090547](https://github.com/ROCm/TheRock/actions/runs/27602090547). Resulting bump [PR #5878](https://github.com/ROCm/TheRock/pull/5878) was automatically labeled `ci:run-all-archs`. CI confirmed builds across all expected families: `gfx900/906/908, gfx90a, gfx94X, gfx950, gfx101X, gfx103X, gfx110X, gfx1150/51/52/53, gfx120X`. 4 test failures observed (`hipsparselt` on gfx94X,` hipsparse/rocsparse/libhipcxx_hipcc` on Windows gfx110X) — pre-existing, unrelated to this change.  

  ## Submission Checklist
  

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -273,11 +273,15 @@ jobs:
     needs: [build_python_packages]
     name: Build PyTorch (fat + split)
     if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_pytorch == true }}
+    strategy:
+      fail-fast: false
+      matrix:
+        pytorch_git_ref: ["release/2.10", "release/2.11", "release/2.12"]
```
ScottTodd: Keep PRs focused on one topic at a time please. I'm already making this change in https://github.com/ROCm/TheRock/pull/5729 (as announced on https://github.com/ROCm/TheRock/issues/5714#issuecomment-4662532344)

amd-chiranjeevi: Thanks Scott for the heads up. I will revert pytorch changes in this PR.

ScottTodd: No problem, I mainly want us to avoid duplicating work. Each expansion of the CI matrix has the potential to add considerable load to our build and test systems, as well as increase the risk of flakes, so I want us to roll out these changes carefully.

In the case of the GPU targets/families, we probably want to have submodule update PRs run more jobs than regular PRs. We can currently achieve that with the `ci:run-all-archs` label https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md or we can add some dedicated code to the configure script for that. We could even have https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/bump_automation.py automatically apply that label to the bump PRs that it creates 🤔 

BTW since I created the https://github.com/ROCm/TheRock/pull/5729 from a fork, it's affected by https://github.com/ROCm/TheRock/issues/5737 and the workflow jobs failed. I could recreate from the shared repository but we need to get the CI system back to working on all kinds of PRs. I also decided on that PR to start with just expanding the _Linux_ matrix, leaving Windows alone, as I think/hope that should get us sufficient coverage with minimal extra configuration. Like with the GPU targets/families though, what we should really do is dynamically choose which jobs to run (more on submodule update PRs, less on routine build system changes, opt-in when a developer wants extra testing and sees something the CI system doesn't know to look for on its own yet, etc.)

amd-chiranjeevi: >  We could even have https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/bump_automation.py automatically apply that label to the bump PRs that it creates 🤔

Thanks Scott, this approach looks perfect, I'll proceed with this.

amd-chiranjeevi: now I am looking for what `ci:run-all-archs` label enable. as per current infra availability, only Linux gf94x, windows gfx110x tests are being tested on the prs.

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -256,6 +256,8 @@ def select_build_runner(platform: str, build_variant: str) -> str:
             "family": "gfx950-dcgpu",
             "fetch-gfx-targets": ["gfx950"],
             "build_variants": ["release", "asan", "tsan"],
+            # Build-only on PRs (including ci:run-all-archs); tests run on push and nightly
+            "postsubmit_check_only_for_family": True,
```
geomin12: is this the right change?

presubmit = PR
postsubmit = on push

wouldn't the label `ci:run-all-archs` cover this and we don't need to add this scenario?

build_tools/github_actions/amdgpu_family_matrix.py
geomin12, line 1: my biggest worry here, is for submodule bumps, we need to make sure specific archs only runs sanity checks (and not full tests)

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -941,6 +941,16 @@ def _expand_build_config_for_platform(
...
+        # If postsubmit_check_only_for_family is set, skip tests on PRs only
+        if platform_info.get("postsubmit_check_only_for_family", False) and not (
+            is_push or is_schedule
+        ):
+            test_runs_on = ""
+            print(
+                f"  {family_name}: postsubmit_check_only_for_family flag set, "
+                f"disabling test runner for pull request runs"
+            )
+
```
geomin12: i think we would still want gfx950 to run all tests since we have enough machines, but the machines that we do not have enough of, we should only run sanity

amd-chiranjeevi: okay.

build_tools/github_actions/amdgpu_family_matrix.py
amd-chiranjeevi: we would still want gfx950 to run all tests since we have enough machines, but the machines that we do
  not have enough of, we should only run sanity.

build_tools/github_actions/bump_automation.py
```diff
@@ -238,12 +238,20 @@ def create_therock_bump(submodule, token):
...
             token,
             f"repos/{THEROCK_REPO}/pulls",
             method="POST",
             data={"title": title, "head": branch_name, "base": "main", "body": body},
         )
+
+        # Add ci:run-all-archs label to the PR
+        gh_api(
+            token,
+            f"repos/{THEROCK_REPO}/issues/{pr['number']}/labels",
```
geomin12: is this the right route? shouldn't it be 


```suggestion
            f"repos/{THEROCK_REPO}/pull/{pr['number']}/labels",
```

amd-chiranjeevi: Pull requests are a type of issue, so you use the Issues API to manage labels for pull requests. We cannot assign labels directly when creating a pull request; instead, we can add them after the PR is created. 

Labels on PRs are managed exclusively through the Issues API (`/issues/{number}/labels`); there is no `/pull/{number}/labels` endpoint. Using `/pull/` would result in a 404. doc here: https://docs.github.com/en/rest/issues/labels?apiVersion=2026-03-10

geomin12 (approved):

arjun-raj-kuppala (approved):
LGTM

Issue #5749, opened 2026-06-10
[Issue] Release portable Linux, Windows packages (nightly): `Trigger building PyTorch wheels" fails — 'nightly' not in allowed values)`


TriveniTadapaneni:
## Context
- **Workflow:** Release portable Linux packages
- **Workflow file:** `.github/workflows/release_portable_linux_packages.yml`,`Release Windows packages` (`.github/workflows/release_windows_packages.yml`)
- **Failing run:** [↗ View run](https://github.com/ROCm/TheRock/actions/runs/27250778384), [↗ View run](https://github.com/ROCm/TheRock/actions/runs/27250800497)
- **Trigger:** Scheduled nightly (`schedule` event)
- **Platform:** Linux, Windows

---

## PyTorch/JAX/native package dispatch rejected (10 GPU families)

**Affected arches:** `gfx950-dcgpu, gfx94X-dcgpu, gfx110X-all, gfx120X-all, gfx1151, gfx1150, gfx1153, gfx90a, gfx908, gfx1152`

**Failing step:** `Trigger building PyTorch wheels`

**Exact error:**
```text
🔎 Found workflow, id: 174743032, name: Release portable Linux PyTorch Wheels, path: .github/workflows/release_portable_linux_pytorch_wheels.yml
🚀 Calling GitHub API to dispatch workflow...
##[error]Provided value 'nightly' for input 'release_type' not in the list of allowed values - https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event
```

**Root cause:** Commit [a7415d438b3c](https://github.com/ROCm/TheRock/commit/a7415d438b3c5925f15e287eed468b3129812f29) changed `release_type` in the `workflow_dispatch` trigger of `release_portable_linux_pytorch_wheels.yml`, `release_portable_linux_jax_wheels.yml`, and `build_native_linux_packages.yml` from `type: string` to `type: choice` with only `- dev`. The nightly scheduler passes `"nightly"`, which the GitHub API now rejects against the choice allowlist.

**Fix:** Add `nightly` and `prerelease` to the `options` list for `release_type` in the `workflow_dispatch` trigger of all three affected workflow files.

amd-chiranjeevi:
Hi @ScottTodd 
https://github.com/ROCm/TheRock/pull/5750 - created the revert PR to fix immediately.
but if you have other plans/fix in place, please ignore and close this.

ScottTodd:
[...]

PR #5481, opened 2026-05-27
Make CI release type explicit and limit workflow_dispatch inputs to [ci, dev]

ScottTodd:
## Motivation

Follow-up to:
* https://github.com/ROCm/TheRock/pull/4408
* https://github.com/ROCm/TheRock/pull/5452#discussion_r3313698727

This changes the release type used throughout workflows and associated scripts for CI runs (starting from [`multi_arch_ci.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml) or equivalent) from using `""` (empty string) to using `"ci"`.

With that in place, it then replaces all uses of the `release_type` input for `workflow_dispatch` events with a `choice` input type that only allows `[dev]` or `[ci, dev]`, preventing developers from triggering `nightly` or `prerelease` builds from this repository. All releases with those types should be triggered in https://github.com/ROCm/rockrel instead (where permissions will be more limited).

## Test Plan

* I spot checked a few workflows in the UI
  * [`multi_arch_build_tarballs.yml`](https://github.com/ROCm/TheRock/actions/workflows/multi_arch_build_tarballs.yml) --> https://github.com/ROCm/TheRock/actions/runs/26544216544 --> https://therock-ci-artifacts.s3.amazonaws.com/26544216544-linux/tarballs/index.html
  * [`release_portable_linux_packages.yml`](https://github.com/ROCm/TheRock/actions/workflows/release_portable_linux_packages.yml)
    image
* Watch CI and nightly release results for any unexpected changes to bucket selection

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

Generated with Codex

PR #5750, opened 2026-06-10
Revert "Make CI release type explicit and limit workflow_dispatch inputs to [ci, dev]"

amd-chiranjeevi:
Reverts ROCm/TheRock#5481

`release_type` in the `workflow_dispatch` trigger of `release_portable_linux_pytorch_wheels.yml`, `release_portable_linux_jax_wheels.yml`, and `build_native_linux_packages.yml` from `type: string` to `type: choice` with only `- dev`.

This is causing `workflow_dispatch`  failures in `nightly`

```
🔎 Found workflow, id: 174743032, name: Release portable Linux PyTorch Wheels, path: .github/workflows/release_portable_linux_pytorch_wheels.yml
🚀 Calling GitHub API to dispatch workflow...
##[error]Provided value 'nightly' for input 'release_type' not in the list of allowed values - https://docs.github.com/rest/actions/workflows#create-a-workflow-dispatch-event
```

related issue: https://github.com/ROCm/TheRock/issues/5749

### DIFF
--- a/.github/actions/configure_aws_artifacts_credentials/action.yml
+++ b/.github/actions/configure_aws_artifacts_credentials/action.yml
@@ -12,8 +12,8 @@ description: >-
 
 inputs:
   release_type:
-    description: 'Release type: "ci", "dev", "nightly", or "prerelease"'
-    default: ci
+    description: 'Release type: "" for CI, or "dev", "nightly", "prerelease".'
+    default: ""
 
 runs:
   using: "composite"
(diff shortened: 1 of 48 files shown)


ScottTodd (commented):
I'll prepare a partial revert of just the single-arch release files instead.

ScottTodd:
Partial revert: https://github.com/ROCm/TheRock/pull/5757

ScottTodd:
Landed the partial revert instead. If that isn't enough then will do a full revert.

Issue #2025, opened 2025-11-05
[Announcement] Deleting published therock_pytorch_dev docker packages on 2025-11-10


ScottTodd:
### Overview

We plan on deleting old docker packages from https://github.com/orgs/ROCm/packages?repo_name=TheRock:

* https://github.com/ROCm/TheRock/pkgs/container/therock_pytorch_dev_ubuntu_24_04_gfx942
* https://github.com/ROCm/TheRock/pkgs/container/therock_pytorch_dev_ubuntu_24_04_gfx1151
* https://github.com/ROCm/TheRock/pkgs/container/therock_pytorch_dev_ubuntu_24_04_gfx1201
* https://github.com/ROCm/TheRock/pkgs/container/therock_pytorch_dev_ubuntu_24_04_gfx1100
* https://github.com/ROCm/TheRock/pkgs/container/therock_pytorch_dev_ubuntu_24_04_gfx950

### Proposed timeline

* 2025-11-05: announcement
* 2025-11-10: mark packages as private
* 2025-11-17: delete packages

> [!IMPORTANT]
> If you have concerns about this, please voice them before those dates!

### Background

These docker packages built pytorch from source before we had nightly rocm and torch releases running. We stopped publishing them 4 months ago once that changed.

New instructions for installing torch are at https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-pytorch-python-packages and instructions for building torch from source are at https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#build-pytorch-with-rocm-support.

### Alternative docker images

Users wishing to have a dockerfile with torch and rocm should create their own containers and follow those instructions.
* In this project we run rocm and pytorch tests in https://github.com/ROCm/TheRock/pkgs/container/no_rocm_image_ubuntu24_04, see https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04.Dockerfile and https://github.com/ROCm/TheRock/blob/2ba822350b3e8c953f1bb5f8f5e5d1d1729d4bd2/.github/workflows/test_pytorch_wheels.yml#L69
[...]

PR #5761, opened 2026-06-10
Delete build_manylinux_rccl_x86_64.Dockerfile

ScottTodd:
## Motivation

Following up on https://github.com/ROCm/TheRock/pull/5730#discussion_r3391242304, I don't think this Dockerfile is currently in use. In fact, I don't think it was ever used since it was added back in https://github.com/ROCm/TheRock/pull/1713?

## Technical Details

* I see no references across the org: https://github.com/search?q=org%3AROCm%20therock_build_manylinux_rccl_x86_64&type=code compared to many references for the non-rccl dockerfile: https://github.com/search?q=org%3AROCm+therock_build_manylinux_x86_64&type=code
* https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_rccl_x86_64 has a total of 81 downloads *total* compared to 404k (yes, thousand) for https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_x86_64

## Test Plan

Should have no impact on any CI here. If there are still references they can continue to use the published images until if/when we delete https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_rccl_x86_64. I'm not planning that at the moment, but we could delete the images like we did for:
* https://github.com/ROCm/TheRock/issues/2025

## Submission Checklist


marbre (commented):
If we drop the Dockerfile we should also drop `dockerfiles/install_openmpi.sh`.

ScottTodd:
Deleted that too. Ping @araravik-psd ?

araravik-psd (approved):
The particular docker image is not used anymore as openmpi is now vendored with therock so there is no need for a different docker image. Approving this PR

PR #5778, opened 2026-06-11
Link per-ISA device files into rocm-sdk-devel

marbre:
## Motivation
In kpack-split mode rocm-sdk-devel is a single generic wheel
(target_family=None), so _devel_artifact_filter dropped every per-ISA
artifact. The per-ISA device files (.kpack archives, .co/.dat/.hsaco
kernels, MIOpen DBs, per-arch .so) were therefore never linked into the
_rocm_sdk_devel tree, while host .so libraries (from generic artifacts)
were - an asymmetry that left devel incomplete for development against a
specific GPU.

## Technical Details
Each rocm-sdk-device-<arch> wheel now carries its own devel-link
manifest and devel expansion materializes it:

- populate_device_files emits .devel_links/<target_family>.json mapping
  each device file to its relative hardlink target into the
  rocm-sdk-libraries overlay. The device setup.py ships it unchanged
  (its package-data glob already covers the platform dir).
- _reconcile_device_links, run by get_devel_root on every call (so on
  rocm-sdk init / path), enumerates installed rocm-sdk-device-* wheels,
  hardlinks each manifest entry from the libraries overlay into
  _rocm_sdk_devel, and records the new path in that wheel's RECORD so
  pip uninstall prunes it. It is idempotent and version-locked.

Wheels have no install hook, so a newly installed device wheel is
linked on the next devel-root access; removal is automatic via pip.
Per-file hardlinks are used because a directory cannot be hardlinked
and a directory symlink would reintroduce the Windows privilege problem
that the symlink-to-hardlink conversion exists to avoid.

Reconcile is hardened against partial/duplicate state:

- _devel_link_ok rejects symlink destinations so they are always
  replaced with true hardlinks.
- The fast path short-circuits only when every link is a correct
  hardlink and cleanly recorded in its wheel's RECORD, repairing an
  interrupted prior run (hardlink created but RECORD write lost) that
  would otherwise leave a link pip uninstall cannot prune.
- _record_has_entries detects duplicate RECORD rows and
  _ensure_record_entries rewrites to drop them.

## Test Plan
- build_tools/tests: py_packaging_test.py (manifest emission) and new
  devel_reconcile_test.py (link, idempotent, add-arch, version-skip,
  prune, RECORD repair, duplicate-row removal, symlink replacement,
  stale-file overwrite).
- E2E: built nightly wheels (rockrel run 27109338630, gfx942+gfx950)
  and installed into a fresh venv.

## Test Result
- build_tools/tests: 537 passed, 4 skipped; pre-commit clean.
- E2E: device wheels ship the manifest (1631 gfx942 / 992 gfx950
  links); rocm-sdk init hardlinks all into _rocm_sdk_devel and records
  them; re-init idempotent; adding gfx950 links it without touching
  gfx942; pip uninstall removes exactly the arch's links; no
  device-reconcile dangling links.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -660,6 +660,16 @@ Devel contents expanded to '.venv/lib/python3.12/site-packages/_rocm_sdk_devel'
 These contents are useful for using the package outside of Python and _lazily_ expanded on the
 first use when used from Python.
 
+> [!NOTE]
+> The devel tree is expanded (and its device files linked) only once - on the
+> first `rocm-sdk init` or the first use of a devel tool such as `hipcc`. If you
+> install or remove a `rocm-sdk-device-*` wheel (for example, adding a second GPU
+> target) **after** that first expansion, re-run `rocm-sdk init` to link the new
+> device files into the devel tree. The compiler tools do not re-scan on their
+> own, so a device wheel added later is not picked up until you run
+> `rocm-sdk init` (or any `rocm-sdk path`) again. Uninstalling a `rocm-sdk-device-*`
+> wheel removes its devel files automatically via `pip`.
+
 Once you have verified your installation, you can continue to use it for
 standard ROCm development or install PyTorch, JAX, or another supported Python ML
 framework.
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -180,6 +180,22 @@ After installing, verify your installation:
 rocm-sdk test
 ```
 
+The `rocm-sdk-devel` development files (headers, CMake config, and the device
+`.kpack`/kernel files from your `rocm-sdk-device-*` wheels) are expanded on first
+use. To expand them eagerly, run `rocm-sdk init`.
+
+> [!NOTE]
+> The devel tree is expanded - and its device files linked from the installed
+> `rocm-sdk-device-*` wheels - only once: on the first `rocm-sdk init` /
+> `rocm-sdk test`, or the first use of a devel tool such as `hipcc`. If you
+> install or remove a `rocm-sdk-device-*` wheel (for example, adding a second GPU
+> target) **after** that first expansion, re-run `rocm-sdk init` or `rocm-sdk test`
+> to link the new device files. The compiler tools do not re-scan on their own,
+> so a device wheel added later is not picked up until you run one of those again.
+> Uninstalling a `rocm-sdk-device-*` wheel removes its devel files automatically
+> via `pip`. If the devel tree ever ends up in a bad state, recreate the virtual
+> environment.
+
 #### Supported Python `[device-*]` install extras
 
 For packages which include device-specific code (such as `rocm`, `torch`, and
@@ -659,16 +675,6 @@ Devel contents expanded to '.venv/lib/python3.12/site-packages/_rocm_sdk_devel'
 
 These contents are useful for using the package outside of Python and _lazily_ expanded on the
 first use when used from Python.
-
-> [!NOTE]
-> The devel tree is expanded (and its device files linked) only once - on the
-> first `rocm-sdk init` or the first use of a devel tool such as `hipcc`. If you
-> install or remove a `rocm-sdk-device-*` wheel (for example, adding a second GPU
-> target) **after** that first expansion, re-run `rocm-sdk init` to link the new
-> device files into the devel tree. The compiler tools do not re-scan on their
-> own, so a device wheel added later is not picked up until you run
-> `rocm-sdk init` (or any `rocm-sdk path`) again. Uninstalling a `rocm-sdk-device-*`
-> wheel removes its devel files automatically via `pip`.
 
 Once you have verified your installation, you can continue to use it for
 standard ROCm development or install PyTorch, JAX, or another supported Python ML
--- a/build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
+++ b/build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
@@ -79,9 +79,20 @@ def _get_module_path(expand_devel: bool) -> Path:
          These tools should pass `expand_devel=False`.
       B) Other tools that benefit from the extra files in the 'devel' package
          will expand expand it by passing `expand_devel=True`.
+
+    NOTE: the "already expanded" check is one-shot. Once the devel tree exists
+    it is returned directly, WITHOUT re-running the device-link reconcile in
+    `rocm_sdk._devel._reconcile_device_links` (only `_expand_devel_module()` /
+    `rocm-sdk init` does that). So a `rocm-sdk-device-*` wheel installed or
+    removed after the first expansion is NOT picked up by these trampolines
+    (e.g. hipcc); refresh it with an explicit `rocm-sdk init` / `rocm-sdk path`.
+    This is intentional: reconciling on every compiler invocation would add a
+    subprocess + metadata scan to a build hot path.
     """
     if _has_devel_module():
         if _is_devel_module_expanded():
+            # One-shot: returns the existing tree without re-reconciling device
+            # links (see NOTE above).
             return _get_devel_module_path()
         elif expand_devel:
             _expand_devel_module()
--- a/build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
+++ b/build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
@@ -85,7 +85,8 @@ def _get_module_path(expand_devel: bool) -> Path:
     `rocm_sdk._devel._reconcile_device_links` (only `_expand_devel_module()` /
     `rocm-sdk init` does that). So a `rocm-sdk-device-*` wheel installed or
     removed after the first expansion is NOT picked up by these trampolines
-    (e.g. hipcc); refresh it with an explicit `rocm-sdk init` / `rocm-sdk path`.
+    (e.g. hipcc); refresh it with an explicit
+    `rocm-sdk init` / `rocm-sdk path` / `rocm-sdk test`.
     This is intentional: reconciling on every compiler invocation would add a
     subprocess + metadata scan to a build hot path.
     """
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
@@ -313,6 +313,11 @@ def _reconcile_device_links(
     Idempotent and safe to call on every `get_devel_root()`: links already in
     place are left untouched and add nothing to RECORD. Returns the number of
     links created during this call.
+
+    Note: the core CLI trampolines (hipcc etc., see rocm_sdk_core._cli) only
+    reach `get_devel_root()` on the FIRST devel expansion, so a device wheel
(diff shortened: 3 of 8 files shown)


nunnikri:
@marbre Thanks for this PR. Its much better approach than https://github.com/ROCm/TheRock/pull/5768. I will close that in favor of this one. 
Only concern I have is changes in build_tools/_therock_utils/py_packaging.py will create the json files needed . But it needs  `rocm-sdk init` or `rocm-sdk path` to actually create the paths. So we need to do both, 
pip install ...
rocm-sdk init
But our documentation is not asking to do only `pip install `

marbre:
> @marbre Thanks for this PR. Its much better approach than #5768. I will close that in favor of this one. Only concern I have is changes in build_tools/_therock_utils/py_packaging.py will create the json files needed for . But how the expand only `rocm-sdk init` or `rocm-sdk path` is needed to actually create the paths. So we need to do both, pip install ... rocm-sdk init But our documentation is not asking to do only `pip install `

Fair point. I will push a note to our `RELEASES.md` to make this clear to the users _after_ CI went through and produced wheels to test :) I don't want to reset CI right now for just a doc change.

ScottTodd:
release docs: https://github.com/ROCm/TheRock/blob/604398267d866d104f4425b8235c4d642df3ea0c/RELEASES.md?plain=1#L652-L665

(we should copy that up to the multi-arch install section too)

nunnikri (approved):
LGTM

nunnikri:
The test passed for rocHPL with the ci build. 
Steps:
python3 -m venv .venv
source .venv/bin/activate
 pip install rocm[libraries,devel,device-all] --pre     --find-links=https://therock-ci-artifacts.s3.amazonaws.com/27355791104-linux/python/index.html
rocm-sdk init

export ROCM_RT=$(dirname $ROCM)/_rocm_sdk_core
export ROCM_LB=$(dirname $ROCM)/_rocm_sdk_libraries
export ROCM_DEV=$(dirname $ROCM)/_rocm_sdk_devel
export LD_LIBRARY_PATH=$ROCM_RT/lib:$ROCM_LB/lib:$ROCM_RT/lib/rocm_sysdeps/lib/:$LD_LIBRARY_PATH

git clone https://github.com/ROCm/rocHPL.git
cd rocHPL
./install.sh --with-rocm=$ROCM --with-rocblas=$ROCM_LB
./build/mpirun_rochpl -P 4 -Q 2 -N 256000 --NB 512

RELEASES.md
```diff
@@ -660,6 +660,16 @@ Devel contents expanded to '.venv/lib/python3.12/site-packages/_rocm_sdk_devel'
...
 
+> [!NOTE]
+> The devel tree is expanded (and its device files linked) only once - on the
+> first `rocm-sdk init` or the first use of a devel tool such as `hipcc`. If you
+> install or remove a `rocm-sdk-device-*` wheel (for example, adding a second GPU
+> target) **after** that first expansion, re-run `rocm-sdk init` to link the new
+> device files into the devel tree. The compiler tools do not re-scan on their
+> own, so a device wheel added later is not picked up until you run
+> `rocm-sdk init` (or any `rocm-sdk path`) again. Uninstalling a `rocm-sdk-device-*`
+> wheel removes its devel files automatically via `pip`.
```
ScottTodd: Let's move this up into the multi-arch releases section (and copy surrounding text too).

This is currently nested under `## Per-family releases` (line 425), when it should be under `## Multi-arch releases` (line 51). Per-family releases don't have `rocm-sdk-device` wheels.

docs/packaging/python_packaging.md
```diff
@@ -14,15 +14,18 @@ We generate the following types of packages:
   evaluated at the point of install, allowing it to perform some selection logic
   to detect dependencies and needs. Most other packages are installed by
   asking to install extras of this one (i.e. `rocm[libraries]`, etc).
+
```
ScottTodd: Did you intend to add these newlines between list items? That adds spacing between them: https://github.com/ROCm/TheRock/blob/users/marbre/devel-device-file-links/docs/packaging/python_packaging.md#general-design

Image

Where there previously was no spacing: https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#general-design

Image

I'm fine with it either way, but we could also try out some other text layout options to highlight the different package types, like a table.

RELEASES.md
```diff
@@ -660,6 +660,16 @@ Devel contents expanded to '.venv/lib/python3.12/site-packages/_rocm_sdk_devel'
 These contents are useful for using the package outside of Python and _lazily_ expanded on the
 first use when used from Python.
 
+> [!NOTE]
+> The devel tree is expanded (and its device files linked) only once - on the
+> first `rocm-sdk init` or the first use of a devel tool such as `hipcc`. If you
```
ScottTodd: We could also suggest `rocm-sdk test` for this, as it will actually report issues, while `rocm-sdk init` will just try to initialize without detailed reporting.

build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
```diff
@@ -79,9 +79,20 @@ def _get_module_path(expand_devel: bool) -> Path:
...
          will expand expand it by passing `expand_devel=True`.
+
+    NOTE: the "already expanded" check is one-shot. Once the devel tree exists
+    it is returned directly, WITHOUT re-running the device-link reconcile in
+    `rocm_sdk._devel._reconcile_device_links` (only `_expand_devel_module()` /
+    `rocm-sdk init` does that). So a `rocm-sdk-device-*` wheel installed or
+    removed after the first expansion is NOT picked up by these trampolines
+    (e.g. hipcc); refresh it with an explicit `rocm-sdk init` / `rocm-sdk path`.
+    This is intentional: reconciling on every compiler invocation would add a
+    subprocess + metadata scan to a build hot path.
```
ScottTodd: Could also recommend `rocm-sdk test` here

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
```diff
@@ -177,6 +185,199 @@ def _find_tarfile(rocm_sdk_devel_path: Path):
...
+                for link in links
+            ],
+        )
+        for record_path, links in plans
+    ):
+        return 0
+
+    lock_path = devel_py_pkg_path / ".devel_reconcile.lock"
+    with open(lock_path, "a") as lock_file:
+        file_lock = FileLock(lock_file)
```
ScottTodd, line 350: My review agent caught a race condition/bug here. This `FileLock` class we wrote has different behavior on Linux and Windows, see https://github.com/ROCm/TheRock/blob/d85d05bb30f29fe872936f2760dbb1a5d87a8df7/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py#L247-L270 (from https://github.com/ROCm/TheRock/commit/7c0c931af3eeaba3f943b4cb4774145228af08d2)

Specifically, it seems like a file lock on an empty file is not blocking, while locking a byte range in a non-empty file is.

I (plus help from Codex) think that this may be sufficient in the FileLock implementation to fix:
```python
  self.lock_size = max(os.path.getsize(file.name), 1)
  ...
  msvcrt.locking(self.file.fileno(), msvcrt.LK_NBLCK, self.lock_size)
  ...
  msvcrt.locking(self.file.fileno(), msvcrt.LK_UNLCK, self.lock_size)
```
Seems like "On Windows, locking 1 byte starting at offset 0 is valid even when the file is empty. It locks a byte range; the file does not need to already contain that byte."

Otherwise we could write at least one byte to the lock file here.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
```diff
@@ -177,6 +185,199 @@ def _find_tarfile(rocm_sdk_devel_path: Path):
...
+                    hardlink_target = dest_path.parent / link["target"]
+                    if not hardlink_target.is_file():
+                        # The target ships in the same wheel as this manifest, so
+                        # it should exist; skip defensively if it somehow does not.
+                        continue
+                    if dest_path.exists() or dest_path.is_symlink():
+                        dest_path.unlink()
+                    dest_path.hardlink_to(hardlink_target)
+                    created += 1
+                _ensure_record_entries(record_path, recorded_names)
```
ScottTodd, line 375: I (plus help from Codex) think there could be edge cases here with the "skip defensively" and partially constructed/cleaned up folders in the site-packages directory. I don't think there's actually much we can reliably do about that at low enough cost other than to focus our documentation on well-supported python package install commands and recommend creating fresh virtual environments if something goes wrong.

ScottTodd (approved):

docs/packaging/python_packaging.md
marbre: Restored what we had before.

RELEASES.md
marbre: Add `rocm-sdk test` as another option.

build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
marbre: Added.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
marbre: Thanks for catching this and the suggestion. Implemented.

ScottTodd (approved):

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_devel.py
```diff
@@ -67,13 +69,19 @@ def get_devel_root() -> Path:
     # we haven't expanded yet or expansion failed and we need to retry
     tarfile_path, _ = _find_tarfile(rocm_sdk_devel_path)
     if (devel_py_pkg_path / "__init__.py").exists() and not tarfile_path:
+        # The generic devel content is expanded one-shot, but per-ISA device
+        # files are owned by independently-installed rocm-sdk-device-* wheels.
+        # Reconcile their links on every call so a device wheel installed after
+        # the first expansion is picked up on the next `rocm-sdk init`.
+        _reconcile_device_links(site_lib_path, devel_py_pkg_path, di.__version__)
         return devel_py_pkg_path
```
ScottTodd, line 77: Looks like this fails on Python 3.10: https://github.com/ROCm/TheRock/issues/5814

https://github.com/ROCm/rockrel/actions/runs/27422263294/job/81050704163#step:14:2746
```
  File "/opt/python/cp310-cp310/lib/python3.10/site-packages/rocm_sdk/_devel.py", line 76, in get_devel_root
    _reconcile_device_links(site_lib_path, devel_py_pkg_path, di.__version__)
  File "/opt/python/cp310-cp310/lib/python3.10/site-packages/rocm_sdk/_devel.py", line 326, in _reconcile_device_links
    plans = _discover_device_link_plans(site_lib_path, expected_version)
  File "/opt/python/cp310-cp310/lib/python3.10/site-packages/rocm_sdk/_devel.py", line 232, in _discover_device_link_plans
    md.MetadataPathFinder.invalidate_caches()
TypeError: MetadataPathFinder.invalidate_caches() missing 1 required positional argument: 'cls'
```

Issue #5777, opened 2026-06-11
[CI] Per-ref build status for full PyTorch test dispatch (test exactly what was built)


ethanwee1:
Follow-up from #4499 (review feedback from @WBobby).

## Problem

The full PyTorch test dispatch (`dispatch_pytorch_wheels_full_test` in `.github/workflows/multi_arch_release_linux_pytorch_wheels.yml`) gates on the **aggregate** `needs.build_pytorch_wheels.result`. The build is a matrix over `python_version` × `pytorch_git_ref` (`release/2.9`, `release/2.10`, `release/2.11`, ..., `nightly`), but the dispatch job cannot tell which refs individually succeeded.

This forces a bad either/or:

- `needs.build_pytorch_wheels.result == 'success'` (current, after #4499) — **underreports**: a single failed leg (even an unrelated ref/python) skips the entire full-test dispatch for that night.
- `always() && result != 'cancelled'` — **misreports**: dispatch runs even when the leg we test failed, and the resolver picks up a stale wheel from the live index, reporting a misleading green.

Both are stuck on the same root cause: only aggregate status is visible.

## Proposed fix

- Surface **per-ref build status** (e.g. per-matrix-cell job outputs or a status artifact) so the dispatch can decide independently for each tested ref.
- Test the **exact version built this run**, consuming the `torch_version` / `package_index_url` build-job outputs added in #5661 instead of re-resolving via `pip install --dry-run` against the live index. This also removes the index-propagation-lag staleness window.

## References

- Originating review: #4499 (WBobby's interim approval)
- Reproducibility / manifest context: #5110, #5633
- Build-output primitives to consume: #5661

ethanwee1:
Reopening — work started in draft PR #5780. Approach: consume per-ref manifests from #5633 (depends on it merging).

ethanwee1:
Resolved by #5780, which moved the full PyTorch test dispatch into the per-ref build workflow so each build's success gates its own dispatch — exactly the per-ref build status this issue asked for. Closing as completed.

PR #5780, opened 2026-06-11
[CI] Dispatch full PyTorch tests per-ref via build manifests (follow-up to #4499)

ethanwee1:
## Summary
Dispatches the full PyTorch test suite from the **per-ref build workflow** (`multi_arch_build_portable_linux_pytorch_wheels.yml`), next to the existing `test_pytorch_wheels` job, using that build's own `torch_version` output.

Refs #5777. Tracked under #5110.

## Problem
The previous dispatch lived in the orchestrator and:
- Gated on the **aggregate** matrix result, so a single failed build leg skipped the dispatch for **all** refs.
- Re-resolved the version with a loose `torch==X.Y.*` `pip --dry-run`, which can pick up a **stale** wheel from a prior run.

GitHub merges matrix-job outputs across cells (last-cell-wins), so per-ref status/version could not be read from the orchestrator's `build_pytorch_wheels.outputs.*` directly — hence the workarounds above.

## Approach
Per @ScottTodd's review, move the dispatch into the reusable build workflow, which builds a **single `(python_version, pytorch_git_ref)`**:
- New `dispatch_pytorch_wheels_full_test` job runs after `build_pytorch_wheels`, reading `needs.build_pytorch_wheels.outputs.torch_version` / `package_index_url` directly — **no re-resolution**.
- Because the job depends on a single-config build, it only runs when **that** build succeeds, so a failed ref skips **only its own** dispatch.
- Scoped to `run_full_pytorch_tests` + `release_type == 'nightly'` + `python_version == '3.12'` + `gfx94X-dcgpu` built, keeping the daily (release branches) / Sunday (nightly) cadence.
- The orchestrator just forwards `run_full_pytorch_tests` down to the build workflow.

This removes the orchestrator-side aggregate setup job and `collect_pytorch_manifest_versions.py`.

## TODO / follow-ups (tracked in #5857)
- [ ] Teach `configure_pytorch_test_matrix.py` about a **test-type axis (short vs full)** and emit matrix entries along it, instead of the hardcoded `gfx94X` / `py3.12` `if` conditions. This generalizes as full tests expand to e.g. Windows gfx1151 and opt-in Linux gfx950 (@ScottTodd suggestion).
- [ ] Source `test_runs_on` / `test_runs_on_multi_gpu` dynamically from `configure_pytorch_test_matrix.py` (`${{ matrix.test_runs_on }}`) instead of hardcoding `linux-gfx942-*-ossci-rocm`, so the dispatch doesn't go stale if runners migrate away from those labels (@ScottTodd).

## Test plan
- `workflow_dispatch_inputs_test.py`
- Both workflow YAMLs parse.

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
@@ -77,6 +77,12 @@ on:
         description: "Compiler cache type ('none', 'sccache', or 'ccache')"
         type: string
         default: "sccache"
+      run_full_pytorch_tests:
+        description: >-
+          Dispatch the full PyTorch test suite after a successful build.
+          Only takes effect for nightly gfx94X-dcgpu Python 3.12 builds.
+        type: boolean
+        default: false
   workflow_dispatch:
     inputs:
       amdgpu_families:
@@ -139,6 +145,12 @@ on:
           - sccache
           - ccache
         default: sccache
+      run_full_pytorch_tests:
+        description: >-
+          Dispatch the full PyTorch test suite after a successful build.
+          Only takes effect for nightly gfx94X-dcgpu Python 3.12 builds.
+        type: boolean
+        default: false
 
 run-name: "Build Multi-Arch Linux PyTorch Wheels (${{ inputs.release_type }}, ${{ inputs.python_version }}, ${{ inputs.pytorch_git_ref || 'manifest URL' }})"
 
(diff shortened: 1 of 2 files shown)


ethanwee1:
Follow-up to #4499 (per WBobby's review). Tracked under #5110.

Scope of this PR (the **consumer** side): make `dispatch_pytorch_wheels_full_test` in `multi_arch_release_linux_pytorch_wheels.yml` decide **per-ref** and pass the **exact built version**, by reading the per-ref manifest (`pytorch.version`) instead of re-resolving via `pip install --dry-run`. This removes both the aggregate `build_pytorch_wheels.result == 'success'` gate (one failed leg currently skips all dispatch) and the risk of dispatching a stale wheel.

Depends on #5633 (manifest generation + `checkout_from_manifest.py`). Implementation will land here once #5633 merges. @ScottTodd flagging to avoid overlap with your listed follow-up "Generate all manifests in `multi_arch_release_linux_pytorch_wheels.yml`" — happy to consume whatever manifest path/output you settle on there.

.github/workflows/multi_arch_release_linux_pytorch_wheels.yml
```diff
@@ -146,18 +146,66 @@ jobs:
...
     if: >-
       ${{
+        always() &&
         inputs.run_full_pytorch_tests &&
         inputs.release_type == 'nightly' &&
-        needs.build_pytorch_wheels.result == 'success'
+        needs.setup_full_test_manifests.result == 'success'
       }}
-    needs: [build_pytorch_wheels]
+    needs: [build_pytorch_wheels, setup_full_test_manifests]
```
ScottTodd: This currently waits until _all_ pytorch wheel builds succeed, just to then fan out again to release versions for testing. I think this would be simpler if you moved the "dispatch pytorch wheels full test" code to be next to the "test_pytorch_wheels" job here: https://github.com/ROCm/TheRock/blob/55d8c4a80a61d989cb7193a5cb763fa5f737d2bb/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L410-L455

That already has access to the torch version as a job output. Could add a new job there that dispatches if gfx94X was built and the current workflow input python version is 3.12

ethanwee1: Good call — done. Moved the dispatch into `multi_arch_build_portable_linux_pytorch_wheels.yml` as a new `dispatch_pytorch_wheels_full_test` job right after `test_pytorch_wheels`. Since that workflow builds a single `(python_version, pytorch_git_ref)`, it dispatches off `needs.build_pytorch_wheels.outputs.torch_version` directly (no re-resolution), only when that build succeeds, gated on `run_full_pytorch_tests` + `release_type == nightly` + `python_version == 3.12` + `gfx94X-dcgpu` built. A failed ref now only skips its own dispatch instead of blocking all of them. The orchestrator just forwards `run_full_pytorch_tests`, and the old aggregate setup job + `collect_pytorch_manifest_versions.py` are gone.

ScottTodd: The other thing we could do there would be teach `configure_pytorch_test_matrix.py` to be aware of the test type (short/full) and generate matrix entries along that axis too. I think as we get into running full tests across e.g. windows gfx1151, or linux gfx950 opt-in that will help

ethanwee1: Agreed — that test-type (short/full) axis in `configure_pytorch_test_matrix.py` is the right long-term home for this, especially for Windows gfx1151 / opt-in Linux gfx950. Tracked as a follow-up in #5857 and noted as a TODO in the PR description; keeping this PR scoped to the hardcoded gfx94X/py3.12 conditions for now.

ethanwee1:
Validation build (dev, gfx94X-dcgpu, py3.12) on this branch: https://github.com/ROCm/TheRock/actions/runs/27631865773

Note: the first attempt hit a `startup_failure` because the new `dispatch_pytorch_wheels_full_test` job requests `actions: write`, but a reusable workflow job cannot request more permissions than its caller grants. Fixed in `2b446e1a` by granting `actions: write` to the `build_pytorch_wheels` job in `multi_arch_release_linux_pytorch_wheels.yml`. The re-dispatched run above passes graph validation and is building.

(dev release_type exercises the build path; the full-test dispatch is gated on `release_type == nightly` and was separately validated end-to-end via a smoke test of the cadence + dispatch step.)

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
```diff
@@ -415,3 +465,56 @@ jobs:
...
+
+      - name: Dispatch full PyTorch test workflow
+        if: ${{ steps.cadence.outputs.dispatch == 'true' }}
+        uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
+        with:
+          workflow: test_pytorch_wheels_full.yml
+          inputs: |
+            { "amdgpu_family": "gfx94X-dcgpu",
+              "test_runs_on": "linux-gfx942-1gpu-ossci-rocm",
+              "test_runs_on_multi_gpu": "linux-gfx942-8gpu-ossci-rocm",
```
ScottTodd, line 509: (Fine for a follow-up, this intersects with other work I have on my plate)

We can also pull these dynamically using https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_pytorch_test_matrix.py as above with `test_runs_on: ${{ matrix.test_runs_on }}`. There's a bit of risk here that if we migrate runners from the `linux-gfx942-1gpu-ossci-rocm` label to another provider then this code could go stale if we miss updating it.

ScottTodd (approved):

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ethanwee1: Thanks for the review! Agreed — tracked this together with the test-type axis in #5857 (sourcing `test_runs_on`/`test_runs_on_multi_gpu` dynamically from `configure_pytorch_test_matrix.py` so the labels do not go stale). Keeping this PR scoped to the hardcoded values for now.

PR #5786, opened 2026-06-11
[ml-libs] Exclude hipDNN python from test artifact

tvy-amd:
## Summary
Nightly single-arch is broken because it uses the test artifacts for sharedlib testing

Replace the per-extension python includes in the hipDNN `test` component with a single `exclude = ["share/hipdnn/python/**"]`. The previous include list pulled python artifacts into the test component, which broke the single-arch nightly; excluding the python tree mirrors the `dev` and `lib` components and unblocks the nightly.

## Risk Assessment

Low risk. Artifact-manifest-only change scoped to the hipDNN test component; no product code, build logic, or public API affected.

## Testing Summary

- Manifest change validated by inspection against the existing `dev`/`lib` component exclude patterns.

## Testing Checklist

- [x] Single-arch nightly - Status: Pending
- [x] PR CI - GitHub PR checks - Status: Pending

## Technical Changes

- Removes the `share/hipdnn/python/**/*.{py,so,pyd}` and `share/hipdnn/tests/python/**/*.py` includes from `[components.test."ml-libs/hipDNN/stage"]`.
- Adds `exclude = ["share/hipdnn/python/**"]` to the test component, matching the `dev` and `lib` components.

### DIFF
--- a/ml-libs/artifact-hipdnn.toml
+++ b/ml-libs/artifact-hipdnn.toml
@@ -20,8 +20,7 @@ include = [
   "bin/*hipdnn_*_test*",
   "bin/hipdnn/CTestTestfile.cmake",
   "lib/test_plugins/**",
-  "share/hipdnn/python/**/*.py",
-  "share/hipdnn/python/**/*.so",
-  "share/hipdnn/python/**/*.pyd",
-  "share/hipdnn/tests/python/**/*.py",
+]
+exclude = [
+  "share/hipdnn/python/**",
 ]
--- a/ml-libs/artifact-hipdnn.toml
+++ b/ml-libs/artifact-hipdnn.toml
@@ -23,6 +23,8 @@ include = [
 ]
 # Exclude python frontend from the test artifact; ships with a dead CI RPATH
 # and is ABI-locked to cp312. See https://github.com/ROCm/TheRock/issues/5678
+# Needed because single-arch builds don't exclude test artifacts (multi-arch
+# does); remove this once single-arch is removed.
 exclude = [
   "share/hipdnn/python/**",
 ]
--- a/ml-libs/artifact-hipdnn.toml
+++ b/ml-libs/artifact-hipdnn.toml
@@ -21,6 +21,8 @@ include = [
   "bin/hipdnn/CTestTestfile.cmake",
   "lib/test_plugins/**",
 ]
+# Exclude python frontend from the test artifact; ships with a dead CI RPATH
+# and is ABI-locked to cp312. See https://github.com/ROCm/TheRock/issues/5678
 exclude = [
   "share/hipdnn/python/**",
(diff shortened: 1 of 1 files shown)


BrianHarrisonAMD (approved):
LGTM, but we should add a comment linking to the issue.

Before merging, we need to verify by triggering a manual single-arch build flow and hceck that `rocm-sdk test` doesn't fail.

ml-libs/artifact-hipdnn.toml
```diff
@@ -20,8 +20,7 @@ include = [
...
   "bin/hipdnn/CTestTestfile.cmake",
   "lib/test_plugins/**",
-  "share/hipdnn/python/**/*.py",
-  "share/hipdnn/python/**/*.so",
-  "share/hipdnn/python/**/*.pyd",
-  "share/hipdnn/tests/python/**/*.py",
+]
+exclude = [
+  "share/hipdnn/python/**",
 ]
```
BrianHarrisonAMD, line 30: Comment to track this back to the issue.

tvy-amd: Added a comment in the toml referencing #5678.

tvy-amd:
Note: Single-arch is being removed but still landing because we can land this PR quickly
https://github.com/ROCm/TheRock/actions/runs/27368721998/job/80921529141 <- Here is it passing
```
(rocm_sdk.tests.devel_test.ROCmDevelTest.testSharedLibrariesLoad) ... ++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/python -m rocm_sdk path --root
ok
```

ml-libs/artifact-hipdnn.toml
```diff
@@ -20,8 +20,11 @@ include = [
...
-  "share/hipdnn/python/**/*.so",
-  "share/hipdnn/python/**/*.pyd",
-  "share/hipdnn/tests/python/**/*.py",
+]
+# Exclude python frontend from the test artifact; ships with a dead CI RPATH
+# and is ABI-locked to cp312. See https://github.com/ROCm/TheRock/issues/5678
+# Needed because single-arch builds don't exclude test artifacts (multi-arch
+# does); remove this once single-arch is removed.
+exclude = [
+  "share/hipdnn/python/**",
```
ScottTodd, line 29: Filtering the artifacts to contain the files you want is just general good practice. The fact that single-arch tests caught an issue there is incidental. I think you'll still want this change for multi-arch builds and CI/CD.

BrianHarrisonAMD: Ideally we want these artifacts for tests, but due to the knock on impacts we are removing them for now.
Eventually this should be included, and tests re-added once it's safe to land.

PR #5792, opened 2026-06-11
[ci] Add hipFFT and rocFFT single-proc multi-GPU test jobs in CI

dileepr1:
## Summary

Adds dedicated multi-GPU CI jobs for **hipFFT** and **rocFFT** that run only the `*multi_gpu*` subset of each component's tests on 8-GPU runners, keeping them off the single-GPU runners (where those tests were previously skipped).

- `fetch_test_configurations.py`: add `hipfft-multi-gpu` and `rocfft-multi-gpu` components. Each is gated with `run_for_test_types: [comprehensive, full]` (new generic scheduling gate) so a scarce 8-GPU runner is only claimed for tiers that actually define `multigpu_<tier>` labels.
- `test_runner.py`: for `MULTI_GPU_COMPONENTS`, query the namespaced `multigpu_<category>` ctest label (and namespaced exclude labels) so the job selects only the multi-GPU tests defined in the component's `test_categories.yaml`. Jobs reuse the single-GPU `rocfft`/`hipfft` test binaries/artifacts via `COMPONENT_DIR_MAPPING`.
- Unit tests added for both the scheduling gate and the label rewriting.

The rocm-libraries submodule pinned on `main` already ships the `multigpu_tests` category (`multigpu_comprehensive`/`multigpu_full`) for both components, so **no temporary submodule pin is needed** (the previous `tf_rocfft_fix` pin has been dropped).

This is a part of adding test filter standardization in FFT components

JIRA ID : ROCM-20736

## Test plan

- [x] `python3 -m unittest fetch_test_configurations_test unit_test_runner` (63 tests pass)
- [ ] CI: confirm `hipfft-multi-gpu` and `rocfft-multi-gpu` jobs schedule on 8-GPU runners for comprehensive/full and are absent for quick/standard
- [ ] CI: confirm single-GPU `hipfft`/`rocfft` jobs no longer attempt multi-GPU tests


malcolmroberts:
This should be renamed to be single-proc multi-gpu; we have multi-proc, multi-gpu coming as well, so it would be good to differentiate.

dileepr1:
> This should be renamed to be single-proc multi-gpu; we have multi-proc, multi-gpu coming as well, so it would be good to differentiate.

Done!

dileepr1:
I need a PR bot waiver. 
The PR meets the requirements, but the bot times out.

geomin12 (commented):
thanks! this is great, one thing:

can we add a label for multi-gpu in https://github.com/ROCm/TheRock/blob/1f1f06bf4cac2a28635aa563a411ede7625746d4/build_tools/github_actions/configure_multi_arch_ci.py#L967 ? so if a dev requests for multi-gpu, it runs all multi-gpu tests

example: rccl update, we want to run all multi-gpu tests

dileepr1:
> thanks! this is great, one thing:
> 
> can we add a label for multi-gpu in
> 
> https://github.com/ROCm/TheRock/blob/1f1f06bf4cac2a28635aa563a411ede7625746d4/build_tools/github_actions/configure_multi_arch_ci.py#L967
> 
> ? so if a dev requests for multi-gpu, it runs all multi-gpu tests
> example: rccl update, we want to run all multi-gpu tests

I'll handle the label in the new PR https://github.com/ROCm/TheRock/pull/7816 since it will support multiple components.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -566,6 +566,48 @@ def _family_matches(
...
+        "job_name": "rocfft-multi-gpu",
+        "base_project": "rocfft",
+        "fetch_artifact_args": "--fft --rand --tests",
+        "timeout_minutes": 30,
+        "test_script": f"python {_get_script_path('test_runner.py')}",
+        "platform": ["linux"],
+        "total_shards_dict": {
+            "linux": 1,
+        },
+        "run_for_test_types": ["comprehensive", "full"],
```
geomin12, line 608: what about standard?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -566,6 +566,48 @@ def _family_matches(
...
+        "timeout_minutes": 30,
+        "test_script": f"python {_get_script_path('test_runner.py')}",
+        "platform": ["linux"],
+        "total_shards_dict": {
+            "linux": 1,
+        },
+        # Only schedule on tiers that define multigpu_<tier> labels in hipfft's
+        # test_categories.yaml (multigpu_comprehensive/full). This keeps the scarce
+        # 8-GPU runner from being claimed by "quick"/"standard" runs that have no
+        # multi-GPU tests.
```
geomin12, line 589: tbh we can remove this. the field name is enough context

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -566,6 +566,48 @@ def _family_matches(
...
+        "run_for_test_types": ["comprehensive", "full"],
+        # Architectures that we have multi GPU setup for testing.
```
geomin12, line 591: remove this, the docs at top of file cover thsi

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -566,6 +566,48 @@ def _family_matches(
...
+        "test_script": f"python {_get_script_path('test_runner.py')}",
+        "platform": ["linux"],
+        "total_shards_dict": {
+            "linux": 1,
+        },
+        # Only schedule on tiers that define multigpu_<tier> labels in hipfft's
+        # test_categories.yaml (multigpu_comprehensive/full). This keeps the scarce
+        # 8-GPU runner from being claimed by "quick"/"standard" runs that have no
+        # multi-GPU tests.
+        "run_for_test_types": ["comprehensive", "full"],
```
geomin12, line 590: let's add standard. i believe for submodule bumps, those run standard and we want this coverage

build_tools/github_actions/fetch_test_configurations.py
dileepr1: We have multigpu tests for comprehensive & full categories only (multigpu_comprehensive and multigpu_full in test_categories.yaml).
Adding 'standard' to run_for_test_types will match zero tests, but will still request for an 8gpu runner and run no tests.

build_tools/github_actions/fetch_test_configurations.py
dileepr1: done!

build_tools/github_actions/fetch_test_configurations.py
dileepr1: done!

geomin12 (approved):
lgtm thanks, this is great. 

FYI, seeing failure: https://github.com/ROCm/TheRock/actions/runs/34611571399/job/103335520015?pr=5792, we should exclude if failing

PR #5698, opened 2026-06-08
Revert "[ci] Improving multi arch CI for external repos "

ScottTodd:
Reverts ROCm/TheRock#5450 to fix https://github.com/ROCm/TheRock/issues/5654

This broke CI on PRs from forks, e.g. https://github.com/ROCm/TheRock/actions/runs/26984086713?pr=5531:
```
Error: Could not assume role with user credentials: User: arn:aws:sts::324352301041:assumed-role/therock-runners-prod-runner-s3-external/aws-sdk-js-session-1780614257367 is not authorized to perform: sts:TagSession on resource: arn:aws:iam::692859939525:role/therock-ci-external
```

PR #5799, opened 2026-06-11
[ci] Improving multi arch CI for external repos (re-land)

geomin12:
Relanding #5698

As the multi arch CI fails in rocm-libraries from other workflows (such as validate package, python builds, etc), this PR addresses those errors. Also in order for `Copy prebuilt stage artifacts` to work for external, we require IAM role ([errors from sample run](https://github.com/ROCm/rocm-libraries/actions/runs/26469059871/job/77937255846#step:6:597))

external repo tested here: https://github.com/ROCm/rocm-libraries/actions/runs/27375104784
forked test here: https://github.com/ROCm/TheRock/actions/runs/27375409471?pr=5796
normal run working here: https://github.com/ROCm/TheRock/actions/runs/27382744190?pr=5799

This is another iteration as we continue to find and fix errors

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -148,7 +148,7 @@ jobs:
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
     with:
       dist_amdgpu_families: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
-      rocm_version: ${{ inputs.rocm_deb_package_version }}
+      rocm_version: ${{ inputs.rocm_deb_package_version || inputs.rocm_package_version }}
       native_package_type: deb
     permissions:
       contents: read
@@ -161,7 +161,7 @@ jobs:
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
     with:
       dist_amdgpu_families: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
-      rocm_version: ${{ inputs.rocm_rpm_package_version }}
+      rocm_version: ${{ inputs.rocm_rpm_package_version || inputs.rocm_package_version }}
       native_package_type: rpm
     permissions:
       contents: read
@@ -284,6 +284,8 @@ jobs:
       kpack_split: ${{ needs.build_python_packages.outputs.kpack_split }}
       rocm_package_find_links_url: ${{ needs.build_python_packages.outputs.package_find_links_url }}
       rocm_version: ${{ inputs.rocm_package_version }}
+      repository: ${{ inputs.repository }}
+      ref: ${{ inputs.ref }}
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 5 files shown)


geomin12:
i had to cancel to finish https://github.com/ROCm/TheRock/actions/runs/27375409471?pr=5796. i'll re-run once https://github.com/ROCm/TheRock/actions/runs/27375409471?pr=5796 finishes

tony-davis (approved):
Approving. The risk to the working internal CI is low here, and that's the
key reason this is safe to land and iterate on:

- The one job with a shared-behavior change — `copy_prebuilt_stages`
  (containerized + `AWS_SHARED_CREDENTIALS_FILE`) — is gated on
  `prebuilt_stages != '' && baseline_run_id != ''`, so it's skipped on the
  normal internal lanes (it shows `skipping` in this run). The credential/
  container change only exercises the external prebuilt-stage path.
- The `repository`/`ref` inputs default to empty, so internal callers
  resolve to `github.repository` and the triggering ref — no behavior change
  for non-external runs.
- The `rocm_deb/rpm_package_version || rocm_package_version` fallback only
  changes behavior when the version was previously empty, which is strictly
  safer.

Net: this can't take down `main` the way the earlier revert (#5698) did,
because the fork-credential path it touches is skipped on internal CI. That
makes fast iteration on the external path the right call.

The red checks on this run (CodeQL/gitleaks/component tests) look
pre-existing/environmental rather than introduced — there's no path from a
checkout-`repository`/`ref` change or a skipped job to those failures.

Non-blocking nits for whenever convenient:
- `test_rocm_wheels.yml` adds `repository`/`ref` without the `default: ""`
  the other two workflows use — align for consistency.
- A one-line comment on why `copy_prebuilt_stages` runs in a container
  would save the next reader a git-blame.

One thing to keep in mind as this effort continues: this stays low-risk only
while the changes are confined to the gated/external path. If a later
iteration touches the ungated jobs that run on every internal multi-arch
build, that one warrants a closer look.

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -161,8 +169,10 @@ jobs:
     uses: ./.github/workflows/multi_arch_build_native_linux_packages.yml
     with:
       dist_amdgpu_families: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
-      rocm_version: ${{ inputs.rocm_rpm_package_version }}
+      rocm_version: ${{ inputs.rocm_rpm_package_version || inputs.rocm_package_version }}
```
ScottTodd, line 172: Can't just mix package versions like this. They have different syntax/constraints: https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#constraints-and-design-guidelines

This caused https://github.com/ROCm/TheRock/issues/7161

ScottTodd: > * The `rocm_deb/rpm_package_version || rocm_package_version` fallback only
> changes behavior when the version was previously empty, which is strictly
> safer.

It is *not* safer. It is plainly wrong. Guys please involve reviewers who understand the packaging or read the docs before making changes like this.

PR #5801, opened 2026-06-11
Run JAX  tests in multi-arch release workflows

### LINKED ISSUE #3878: [CI] Build and test JAX Python packages as part of ci.yml

erman-gurses:
## Motivation

The JAX release workflow produces release artifacts but does not currently validate those artifacts by executing the JAX wheel test suite as part of the release process. This creates a gap between artifact generation and functional validation.

## Solution

Introduced `test_linux_jax_wheels_partial.yml` workflow into the JAX release workflow so that release artifacts are automatically validated after they are built.

The workflow now follows a build → test pattern, where JAX release artifacts are produced first and then consumed by the JAX wheel test workflow running on the `gfx94X-dcgpu` family.

## Workflow Relationship

```mermaid
flowchart TD
    A["multi_arch_release_linux<br/>.yml"]
    B["multi_arch_release_linux<br/>_jax_wheels.yml"]
    C["multi_arch_build_linux<br/>_jax_wheels.yml"]
    D["test_linux_jax_wheels<br/>_partial.yml"]
    E["Target gfx942 from<br/>gfx94X-dcgpu family"]
    F["pytest"]

    A --> B
    B --> C
    C --> D
    D --> E
    E --> F
```
## Changes

- Integrate `test_linux_jax_wheels_partial.yml` into the JAX release workflow.
- Execute JAX wheel validation as part of the release process and surface results in the release workflow.
- Introduce representative-family JAX testing using `gfx94X-dcgpu`.
- Add `generate_target_to_run` integration in `multi_arch_build_linux_jax_wheels.yml`.
- Wire `amdgpu_family` through:
  - `multi_arch_release_linux_jax_wheels.yml`
  - `multi_arch_build_linux_jax_wheels.yml`
  - `test_linux_jax_wheels_partial.yml`
- Pass JAX build artifacts and metadata from the release workflow to the test workflow.
- Install ROCm for validation directly from the generated multi-arch tarball instead of rediscovering artifacts via `install_rocm_from_artifacts.py`.
- Considers only `rocm-jaxlib-v0.9.1` for now.

## Validation

- Verified JAX release artifacts are successfully consumed by the test workflow.
- Verified JAX wheel tests execute against release-produced artifacts.
- Verified successful execution on the `gfx942 ` target.
- Confirmed test results are reported as part of release validation.

## Test Runs

- Verified JAX release workflow integration:
 -https://github.com/ROCm/TheRock/actions/runs/27723263544/job/82045827564

- Verified JAX wheel validation workflow on representative `gfx94X-dcgpu` (`gfx942`) test target:
  -https://github.com/ROCm/TheRock/actions/runs/27733673536

## Follow-ups

- [ ] Add  rocm-jaxlib-v0.10.0 support


.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -107,16 +166,80 @@ jobs:
...
+          python build_tools/github_actions/publish_jax_to_release_bucket.py \
+            --source-dir="${{ env.PACKAGE_DIST_DIR }}" \
+            --release-type="${{ inputs.release_type }}"
+
+  generate_target_to_run:
+    name: Generate target_to_run
+    runs-on: ubuntu-24.04
+    outputs:
+      test_runs_on: ${{ steps.configure.outputs.test-runs-on }}
+      bypass_tests_for_releases: ${{ steps.configure.outputs.bypass_tests_for_releases }}
```
ScottTodd: If this output is unused, I think we can omit it here.

For multi-arch PyTorch packages we immediately publish to https://rocm.nightlies.amd.com/whl-multi-arch/ without the concept of "upload to staging --> test --> promote out of staging", since there isn't a good way to partially promote the multi-arch packages.
* https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#gating-releases-with-pytorch-tests
* https://github.com/ROCm/TheRock/issues/3332#issuecomment-4361527784

JAX doesn't have any split device wheels of its own so I guess we would be deciding:
* (for example) we test on gfx942 and gfx1100
* gfx942 tests complete in 1h and pass
* gfx1100 tests complete in 3h and fail
* what do we do? wait for all tests to succeed before publishing?

I like the simpler approach matching PyTorch: always publish nightly releases that pass the build, build towards pre-merge test coverage to catch regressions early, add good monitoring and status reporting to e.g. https://therock-hud-dev.amd.com/ so project contributors and users can see nightly build/test status for what slips through.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -1,94 +1,77 @@
...
-        description: Branch, tag, or SHA to checkout.
+        description: "Branch, tag, or SHA to checkout."
         type: string
         default: ""
   workflow_dispatch:
     inputs:
+      amdgpu_family:
+        description: "Representative GPU family to use for JAX tests."
+        type: string
+        default: "gfx94X-dcgpu"
```
ScottTodd, line 44: Can you also prepare a PR in rockrel to add this input to https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_jax_wheels.yml?

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -257,6 +257,7 @@ jobs:
           workflow: multi_arch_release_linux_jax_wheels.yml
           inputs: |
             {
+              "amdgpu_family": "gfx94X-dcgpu",
```
ScottTodd: I'm not thrilled with this "representative GPU family to use for tests". Someone could trigger a run of these workflows with Linux building for just gfx1151 and this would continue anyways trying to run tests that will fail. That's especially true since the workflow doesn't currently get _any_ list of GPU targets so it doesn't know what the provided `tar_url` / `rocm_version` were built with support for.

* If keeping this, rename to something more specific like `test_amdgpu_family`
* Did we run tests on more GPUs with the per-family JAX releases? What happens if we just try to test on all machines that we have runners for, similar to PyTorch?
* What about a script like https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_pytorch_test_matrix.py from https://github.com/ROCm/TheRock/pull/5661 ? We could do whatever filtering we want in there, like only emit a matrix row for gfx94X _if gfx94X is in the inputs_
    * The other thing I'd like to do with these matrix scripts is start limiting testing to only a few Python versions, so we have the option of running `10 minutes of tests x 4 $FRAMEWORK versions x 1 python version` instead of `10 minutes of tests x 4 $FRAMEWORK versions x 4 python versions` (particularly on pull_request / push events)
* If you don't go the full script route yet, I think we should still skip testing if not built for gfx94X... could filter up here in this file - pass empty string to `test_amdgpu_family` if `fromJSON(inputs.build_config).dist_amdgpu_families` does not contain gfx94X and then skip `generate_target_to_run` and `test_jax_wheels` in `.github/workflows/multi_arch_build_linux_jax_wheels.yml` if the input is empty

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -181,3 +192,77 @@ jobs:
...
+          TARBALL="${RUNNER_TEMP}/therock.tar.gz"
+          S3_URI="$(printf '%s' "${THEROCK_TAR_URL}" | sed -E 's#^https://([^.]+)\.s3\.amazonaws\.com/#s3://\1/#')"
+
+          rm -rf "${DEST}"
+          mkdir -p "${DEST}"
+
+          aws s3 cp "${S3_URI}" "${TARBALL}"
+          tar -xzf "${TARBALL}" -C "${DEST}"
+
+          ln -sfn "${DEST}" /opt/rocm
```
ScottTodd: Please don't use the `aws` CLI. I want to remove all of our dependencies on it, not add new ones. Use boto3 from Python.

Here I think you can just `wget` or `curl` the original URL instead of converting from https to S3 URI and using `aws s3 cp`

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -181,3 +192,77 @@ jobs:
       - name: Install base JAX test requirements
         run: |
           pip install -r external-builds/jax/requirements-jax.txt
+
+      - name: Configure AWS Credentials
+        uses: ./.github/actions/configure_aws_artifacts_credentials
+        with:
+          release_type: ${{ inputs.release_type }}
```
ScottTodd: Shouldn't need this step if all you're doing is downloading from public buckets.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -107,16 +166,80 @@ jobs:
...
+
+      - name: Extract JAX versions from built wheels
+        id: write_jax_versions
+        run: |
+          python3 ./build_tools/github_actions/write_jax_versions.py \
+            --dist-dir "${{ env.PACKAGE_DIST_DIR }}"
+
+      - name: Install boto3
+        run: |
+          python -m pip install boto3
```
ScottTodd: nit: could move this up towards the top of the job, after "Set up Python" and before " Build JAX Wheels", like here: https://github.com/ROCm/TheRock/blob/67d17e221bb9d4148e12dc9c056b423a6222df47/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml#L169-L176

(otherwise we can easily end up with multiple `pip install` steps throughout a workflow... easier to just install all python deps upfront)

.github/workflows/test_linux_jax_wheels_partial.yml
erman-gurses: used wget

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Renamed as `test_amdgpu_family` for now -  will follow up with matrix one

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: I raised a PR for that

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Good point — bypass_tests_for_releases is currently unused, so I can remove it here. For this PR I am keeping the simpler PyTorch-like flow: publish once the build passes, and use JAX tests as coverage/early regression signal rather than a staging-to-promotion gate. The longer-term matrix/script approach can come in a follow-up.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -107,16 +174,101 @@ jobs:
...
+
+          filename = Path(urlparse(private_tar_url).path).name
+          filename = quote(filename, safe="")
+
+          public_tarball_base = package_index_url.replace(
+              "/whl-multi-arch/",
+              "/tarball-multi-arch/",
+          )
+          print(f"public_tarball_url={public_tarball_base}{filename}")
+          PY
```
ScottTodd: Let's escape the URL before it gets to `"tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",`, here: https://github.com/ROCm/TheRock/blob/e0a36227c819755d06e5b5bd2c8a87e6a795f0db/build_tools/github_actions/upload_tarballs.py#L92-L114

Use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote, as in https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/generate_local_index.py and other files

then drop this `derive_public_tarball_url` step

---

Please no inline scripts like this in workflows (coding agents keep trying to do this and I'll reject the changes every time). See https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash (we can clarify there that inline python is just as bad, if not worse, than inline bash)

`PRIVATE_TAR_URL` isn't an accurate name... on this example run: https://github.com/ROCm/TheRock/actions/runs/27701602688/job/81939242051#step:13:34 you have
```
    PRIVATE_TAR_URL: https://therock-dev-artifacts.s3.amazonaws.com/27137785103-linux/tarballs/therock-dist-linux-multiarch-7.14.0.dev0+08645fc9c96eadf22bcd6a709c5113803271eb72.tar.gz
    PACKAGE_INDEX_URL: https://rocm.devreleases.amd.com/whl-multi-arch/
```

That URL isn't "private", it isn't valid. The URL should be escaped like https://therock-dev-artifacts.s3.amazonaws.com/27137785103-linux/tarballs/therock-dist-linux-multiarch-7.14.0.dev0%2b08645fc9c96eadf22bcd6a709c5113803271eb72.tar.gz (replace `+` with `%2b`)

You could also
* let `test_jax_wheels_partial.yml` escape as needed
* have the `publish_jax_to_release_bucket.py` script (which this PR adds) write `public_tarball_url` to `GITHUB_OUTPUT` itself using this function: https://github.com/ROCm/TheRock/blob/e0a36227c819755d06e5b5bd2c8a87e6a795f0db/build_tools/github_actions/github_actions_api.py#L342

ScottTodd (commented):
Generally LGTM. Need to escape the tarball URL at the right place though, to cut out the inline python code in the workflow file

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Thanks for the feedback

amd-chiranjeevi:
Jax nightly failed due to mismatch input param name , which was modified by pr https://github.com/ROCm/rockrel/pull/61
```
[Invalid workflow file: .github/workflows/multi_arch_release_linux_jax_wheels.yml#L45](https://github.com/ROCm/rockrel/actions/runs/27738669318/workflow)
The workflow is not valid. .github/workflows/multi_arch_release_linux_jax_wheels.yml (Line: 45, Col: 27): Invalid input, test_amdgpu_family is not defined in the referenced workflow.
```

I can see this PR has the fix, aligning name in all the workflows.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Addressed -  I removed the inline Python URL derivation from multi_arch_build_linux_jax_wheels.yml and moved the escaping to build_tools/github_actions/upload_tarballs.py as suggested.

upload_tarballs.py now URL-encodes the tarball filename before emitting tarball_urls.multiarch, so the URL is already valid before it reaches:

${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}

I verified this against the failing example from the review. The unescaped URL containing + returned 404, while the escaped URL (%2b) downloaded successfully and the JAX test workflow completed successfully using the escaped tarball URL.

Updated test results with latest tests below:
https://github.com/ROCm/TheRock/actions/runs/27723263544/job/82045827564
https://github.com/ROCm/TheRock/actions/runs/27733673536

ScottTodd (approved):
Nice, thanks for addressing all the feedback :)

PR #5828, opened 2026-06-12
Override CK GfxArch behavior in kpack mode

arvindcheru:
## Motivation

This pull request updates `packaging_utils.py` to handle a special case for the `amdrocm-ck` package when kpack mode is enabled. The change ensures that, when `--enable-kpack` is used, CK packages are treated as generic artifacts rather than being tied to a specific Gfxarch.

## Technical Details

Behavioral change for CK packages in kpack mode:
* Updated for the `amdrocm-ck` package when kpack mode is enabled, ensuring CK is treated as providing generic artifacts.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -230,6 +230,12 @@ def is_gfxarch_package(pkg_info, enable_kpack=False):
         if pkgname in ["amdrocm-rccl", "amdrocm-rccl-test"]:
             return True
 
+        # Override CK Gfxarch behavior in kpack mode
+        # When --enable-kpack is used, CK should look for generic artifacts
+        # Root Cause: CK provides generic artifacts only
+        if pkgname in ["amdrocm-ck"]:
+            return False
+
     return is_key_defined(pkg_info, "Gfxarch")
 
 


raramakr (approved):

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -230,6 +230,12 @@ def is_gfxarch_package(pkg_info, enable_kpack=False):
         if pkgname in ["amdrocm-rccl", "amdrocm-rccl-test"]:
             return True
 
+        # Override CK Gfxarch behavior in kpack mode
+        # When --enable-kpack is used, CK should look for generic artifacts
+        # Root Cause: CK provides generic artifacts only
+        if pkgname in ["amdrocm-ck"]:
```
nunnikri, line 236: This will work for now. But if ck starts to produce gfx arch specific artifacts do we need more code changes?
cc: @raramakr

raramakr: Once CK begins producing GFX architecture–specific artifacts, we can simply revert the PR.

nunnikri: Can we think of making it more future ready. Unless its too complex, we should support ck for both gfx arch and generic arch packages. 
Ideal flow should be ck should be gfx arch by default, but if the actual artifact has only generic artifact, should go with that. I want to go with current change only if handling both cases is too complex

arvindcheru: Alternate Approach discussed is here - https://github.com/ROCm/TheRock/pull/5874

arvindcheru:
As per review, alternate approach finalized and changes are now merged as part of https://github.com/ROCm/TheRock/pull/5874. Hence closing this PR.

PR #1068, opened 2025-07-17
[ci] Remove custom "platform" variable.

ScottTodd:
See https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#detecting-the-operating-system

PR #5832, opened 2026-06-12
[ci] Use ubuntu24.04 for configure_test_matrix job

geomin12:
As configure test matrix step does not require a GPU, we are using our CPU builders to determine this

linux configure test matrix working: https://github.com/ROCm/TheRock/actions/runs/27442979100/job/81121623123#step:1:2

windows configure test matrix working: https://github.com/ROCm/TheRock/actions/runs/27442979100/job/81121781166#step:1:2

### DIFF
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -98,31 +98,14 @@ run-name: Test Artifacts (${{ inputs.amdgpu_families }}, ${{ inputs.release_type
 jobs:
   configure_test_matrix:
     name: "Configure test matrix (${{ inputs.amdgpu_families }})"
-    # if there is a test machine available
     if: ${{ inputs.test_runs_on != '' }}
-    runs-on: ${{ inputs.test_runs_on }}
+    runs-on: ubuntu-24.04
     outputs:
       sanity_component: ${{ steps.configure.outputs.sanity_component }}
       components: ${{ steps.configure.outputs.components }}
       platform: ${{ steps.configure.outputs.platform }}
       shard_arr: ${{ steps.configure.outputs.shard_arr }}
     steps:
-      - name: "Fetch 'build_tools' from repository"
-        if: ${{ runner.os == 'Windows' }}
-        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
-        with:
-          repository: ${{ inputs.repository || github.repository }}
-          ref: ${{ inputs.ref }}
-          sparse-checkout: build_tools
-          path: "prejob"
-
-      # Checkout failure is possible on Windows, as it's the first job on a GPU test runner.
-      # Post-job cleanup isn't necessary since no executables are launched in this job.
-      - name: Pre-job cleanup processes on Windows
-        if: ${{ runner.os == 'Windows' }}
-        shell: powershell
-        run: . '${{ github.workspace }}\prejob\build_tools\github_actions\cleanup_processes.ps1'
-
       - name: Checking out repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/test_artifacts.yml
```diff
@@ -98,9 +106,8 @@ run-name: Test Artifacts (${{ inputs.amdgpu_families }}, ${{ inputs.release_type
 jobs:
   configure_test_matrix:
     name: "Configure test matrix (${{ inputs.amdgpu_families }})"
-    # if there is a test machine available
     if: ${{ inputs.test_runs_on != '' }}
-    runs-on: ${{ inputs.test_runs_on }}
+    runs-on: ${{ inputs.build_runs_on || inputs.test_runs_on }}
```
ScottTodd: What if we reverted more of https://github.com/ROCm/TheRock/pull/1068 instead? This should be runnable on `ubuntu-24.04`, I don't think we need a self-hosted runner here. Should be similar to https://github.com/ROCm/TheRock/blob/5c841238e4a58eb4a856011dddf288b54d78cf2b/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L372-L393

Mostly need the `--platform linux` / `--platform windows` input

.github/workflows/test_artifacts.yml
```diff
@@ -18,6 +18,10 @@ on:
         default: ""
       test_runs_on:
         type: string
+      build_runs_on:
+        description: "Runner for configure_test_matrix job (no GPU needed). Falls back to test_runs_on if not specified."
+        type: string
+        default: ""
```
ScottTodd: Reminder that any additional `workflow_dispatch` inputs will also need to be added to the stub workflow at https://github.com/ROCm/rockrel/blob/main/.github/workflows/test_artifacts.yml for nightly releases

.github/workflows/test_artifacts.yml
geomin12: oh i thought we wanted to avoid that! but this makes sense to revert (good use of free GH runners :) )

.github/workflows/test_artifacts.yml
geomin12: i'm reverting change :)

.github/workflows/test_artifacts.yml
ScottTodd: Yeah my opinion has changed now that we're handling so many jobs on these runners and queue times are long. The idea there was "if we have a runner, we might as well ask that runner itself to give input on how tests should be configured". Really all we need though is to expand some inputs (platform, projects to test, test type, gpu families, etc.) to an output matrix (list of jobs with arguments, timeouts, test scripts, etc.). That doesn't need an actual GPU.

I'd even like to move more of that upfront into configure_multi_arch_ci.py so we know the jobs before the _builds_ even start, but the current code lets test_artifacts.yml be run directly, which is useful too. (sorta equivalent to a run through multi_arch_ci.yml with all prebuilt artifacts and just "test rocm" jobs... if we wanted to define that)

.github/workflows/test_artifacts.yml
```diff
@@ -143,7 +126,9 @@ jobs:
           TEST_LABELS: ${{ inputs.test_labels }}
           RUN_EXTENDED_TESTS: ${{ inputs.run_extended_tests }}
           WINDOWS_HIP_ROCR_TESTS: ${{ inputs.windows_hip_rocr_tests }}
-        run: python ./build_tools/github_actions/fetch_test_configurations.py
+        run: |
+          python ./build_tools/github_actions/fetch_test_configurations.py \
+            --platform=${{ contains(inputs.test_runs_on, 'windows') && 'windows' || 'linux' }}
```
ScottTodd, line 131: This string check on the runner label is _probably_ safe, but an explicit input would be safer. I like how simple this is at least 😅. LGTM

geomin12: eh i'm not a fan... PR coming soon

ScottTodd: Oh, the other argument for using the runner itself was that it couldn't get out of sync with the platform. If we have two inputs then they need to match.

For PyTorch builds we have separate workflows for Linux and Windows so we don't need a "platform" input.

We do actually check the `test_runs_on` string already there: https://github.com/ROCm/TheRock/blob/563854f95f05ae63ca0975731506dd40837a4a9e/.github/workflows/test_pytorch_wheels.yml#L120

ScottTodd: Same here: https://github.com/ROCm/TheRock/blob/563854f95f05ae63ca0975731506dd40837a4a9e/.github/workflows/test_rocm_wheels.yml#L118

geomin12: actually i see rockrel and that's pretty minimal impact hahaha, i'm testing then opening PR

PR #5835, opened 2026-06-12
[ci] Use `ubuntu24.04` for `configure_test_matrix` job

geomin12:
As configure test matrix step does not require a GPU, we are using ubuntu24 to determine this

linux configure test matrix working: https://github.com/ROCm/TheRock/actions/runs/27445401758/job/81129474031

windows configure test matrix working: https://github.com/ROCm/TheRock/actions/runs/27445401758/job/81130272475

tried to reopen #5832 but it was force-pushed :(

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -15,13 +15,15 @@ Required environment variables:
   - RUNNER_OS (https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#detecting-the-operating-system)
 """
 
+import argparse
 import ast
 import json
 import logging
 import os
+import platform as platform_module
 import sys
-from pathlib import Path
 from copy import deepcopy
+from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -676,7 +677,15 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
 
 
 def run():
-    platform = os.getenv("RUNNER_OS").lower()
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--platform",
+        type=str,
+        default=os.getenv("RUNNER_OS", "linux").lower(),
```
ScottTodd: We could also use

```python
import platform as platform_module

default=platform_module.system().lower()
```

here, like we do in other scripts. That would allow Windows developers to try out the script without needing to also set the `RUNNER_OS` CI environment variable

ScottTodd (approved):

build_tools/github_actions/fetch_test_configurations.py
geomin12: oooo i like that

PR #5816, opened 2026-06-12
Disable PyTorch sccache on fork PRs via a computed cache_type

subodh-dubey-amd:
## Summary

Fixes #5737 (supersedes the Draft #5738). PyTorch CI jobs fail on PRs from forks at the sccache steps:

```
User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform:
sts:TagSession on resource: arn:aws:iam::324352301041:role/therock-ci
... AccessDenied ... s3:GetObject on therock-pytorch-sccache-ci/.../.sccache_check
```

### Root cause

sccache uses an S3 bucket reached via an OIDC-assumed IAM role (`therock-ci`). Fork PRs don't get OIDC tokens, so the role can't be assumed. The gate `github.repository_owner == 'ROCm'` is **always true** on fork PRs (it's the base repo), and it only guarded the credentials step — `Verify sccache` and the build still passed `--use-sccache`, so sccache started and failed on S3 access.

### Fix

Compute the effective cache type **once**, in [build_tools/github_actions/compute_pytorch_cache_type.py](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/compute_pytorch_cache_type.py) (reusing `_is_current_run_pr_from_fork()` from `s3_buckets.py`), downgrading `sccache` -> `none` on fork PRs / non-ROCm repos. Every sccache step in the two `*_pytorch_wheels_ci.yml` workflows now keys off that single `steps.cache.outputs.cache_type` instead of repeating fork checks. `ccache`/`none` pass through unchanged; in-org runs keep sccache and its hard-fail semantics.

This follows the review direction on #5738 ("compute earlier… this all needs to go through scripts, not yml code") and limits changes to the `_ci.yml` workflows (the release / `multi_arch_build_*` workflows are never fork-triggered).

## Files

- `build_tools/github_actions/compute_pytorch_cache_type.py` (new)
- `build_tools/github_actions/tests/compute_pytorch_cache_type_test.py` (new, 8 cases)
- `build_portable_linux_pytorch_wheels_ci.yml`, `build_windows_pytorch_wheels_ci.yml` — add "Determine cache type" step; gate all sccache steps on its output.

## Test plan

- [x] Unit tests (fork / in-org / non-PR / non-ROCm / none / ccache).
- [x] Real-script test against fork-shaped event payloads: fork+sccache->none, in-org+sccache->sccache, fork+ccache->ccache.
[...]

PR #5848, opened 2026-06-14
Add public is_current_run_pr_from_fork() wrapper for cache-type gating

subodh-dubey-amd:
Follow-up to #5816, addressing the review nit at https://github.com/ROCm/TheRock/pull/5816#discussion_r3405042032 (importing the private `_is_current_run_pr_from_fork` across files).

Adds a public `is_current_run_pr_from_fork()` wrapper in `s3_buckets.py` and repoints `compute_pytorch_cache_type.py` (and its test mocks) to it. No behavior change.

## Test plan
- [x] `compute_pytorch_cache_type_test.py` (8 cases) pass
- [x] Real fork-event smoke test: fork+sccache -> none

## Submission Checklist
 Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -170,6 +170,15 @@ def _is_current_run_pr_from_fork() -> bool:
     )
 
 
+def is_current_run_pr_from_fork() -> bool:
+    """Whether the current workflow run is a pull request from a fork.
+
+    Public wrapper around the module-internal implementation, for reuse by
+    other build scripts.
+    """
+    return _is_current_run_pr_from_fork()
+
+
 def get_artifacts_bucket_config_for_workflow_run(
     github_repository: str,
     release_type: str | None = None,
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -142,43 +142,6 @@ def get_release_bucket_config(
     return _BUCKET_CONFIGS_BY_NAME[bucket_name]
 
 
-def _is_current_run_pr_from_fork() -> bool:
-    """Check if the current workflow run is a pull request from a fork.
-
-    Reads the GitHub event payload to check the .fork property on the
-    head repo, matching the behavior of the GitHub Actions expression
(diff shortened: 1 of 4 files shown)


build_tools/_therock_utils/s3_buckets.py
```diff
@@ -170,6 +170,15 @@ def _is_current_run_pr_from_fork() -> bool:
     )
 
 
+def is_current_run_pr_from_fork() -> bool:
+    """Whether the current workflow run is a pull request from a fork.
+
+    Public wrapper around the module-internal implementation, for reuse by
+    other build scripts.
+    """
+    return _is_current_run_pr_from_fork()
```
ScottTodd: Huh? What? No. This doesn't make sense.

1. The function does not make sense to make public on `s3_buckets.py`. It would fit better in https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/github_actions_api.py
2. Don't just add a "public wrapper to a module-internal implementation". That's adding boilerplate for the sake of boilerplate.

ScottTodd (changes requested):

build_tools/_therock_utils/s3_buckets.py
subodh-dubey-amd: To avoid two copies, I'd keep a single implementation in `github_actions_api.py` and have `s3_buckets.py` import it. I checked usages first — that private function has exactly one caller (inside `s3_buckets.py` itself) and is imported nowhere else, so consolidating is safe.

| Option | `s3_buckets.py` | Result |
| --- | --- | --- |
| **A — consolidate (preferred)** | drop its private fn, import from `github_actions_api` | one implementation; touches only its single internal caller |
| **B — minimal** | leave `_is_current_run_pr_from_fork()` as-is | smaller diff, but the same logic lives in two modules |

subodh-dubey-amd: Going with **A** unless you'd rather I keep `s3_buckets.py` untouched.

ScottTodd (approved):

PR #5866, opened 2026-06-15
fix(ci): Add necessary environment settings for pytorch full suite workflow

zjliu-amd:
Closing https://github.com/ROCm/frameworks-internal/issues/16941
Adding required env vars and other environment things to fix full suite testing environment issues.

The `ldconfig` related changes are needed in addition to `LD_LIBRARY_PATH` because [a test](https://github.com/ROCm/pytorch/blob/release/2.11/test/export/test_export_opinfo.py#L216-L219) spawns a subprocess with a clean environment, and therefore LD_LIBRARY_PATH is wiped in that subprocess' environment, causing it fail to find dynamic libs.


ethanwee1 (approved):
LGTM but please include the tests that require lines 322-330 in the description because I do not understand the comments in the changes.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -301,15 +303,32 @@ jobs:
           echo "CPLUS_INCLUDE_PATH=${ROCM_SYSDEPS_INCLUDE}:${CPLUS_INCLUDE_PATH:-}" >> "${GITHUB_ENV}"
           echo "C_INCLUDE_PATH=${ROCM_SYSDEPS_INCLUDE}:${C_INCLUDE_PATH:-}" >> "${GITHUB_ENV}"
           echo "PKG_CONFIG_PATH=${ROCM_SYSDEPS_PKGCONFIG}:${PKG_CONFIG_PATH:-}" >> "${GITHUB_ENV}"
-          echo "LD_LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LD_LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
-          echo "LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LD_LIBRARY_PATH=${ROCM_HOME}/lib:${ROCM_CORE}/lib:${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LD_LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LIBRARY_PATH=${ROCM_HOME}/lib:${ROCM_CORE}/lib:${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
```
ScottTodd, line 307: We really shouldn't need to set `LD_LIBRARY_PATH` to use python packages, that should be handled internally by the projects/packages. I guess this is fine for now, but if tests need these workarounds then users likely need them too, and we should fix that.

.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -301,15 +303,32 @@ jobs:
...
           echo "CXX=${ROCM_HOME}/lib/llvm/bin/clang++" >> "${GITHUB_ENV}"
 
+      - name: Install C++ build tools and register test library paths
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends pkg-config
+          PY_BIN="$(readlink -f "$(command -v python)")"
+          PY_LIBDIR="$(ldd "${PY_BIN}" 2>/dev/null | awk '/libpython/ {print $3}' | head -n1 | xargs -r dirname)"
+          # Register lib dirs in the loader cache, not just LD_LIBRARY_PATH: test_export_opinfo
+          # spawns a subprocess with a wiped env that still imports torch (needs ROCm libs + libpython).
```
ScottTodd, line 322: That sounds like `test_export_opinfo` should be fixed somehow ;) I'm not really comfortable setting too many environment variables process-wide that could affect other tests too.

Why does `import torch` fail when the environment is "wiped"? 

https://github.com/ROCm/pytorch/blob/release/2.11/test/export/test_export_opinfo.py#L216-L219 already has a skip
```python
    @unittest.skipIf(
        IS_WINDOWS,
        'Subprocess with CUDA_VISIBLE_DEVICES="" imports op_db which triggers '
        "get_device_capability(); 0 devices raises Invalid device id on Windows.",
    )
```
that suggests some other code paths are not tolerant of empty device lists. We could expand that for Linux ROCm as well, or adjust the torch loading path to tolerate having zero visible devices and load without crashing (if that's the issue)

ScottTodd (commented):
I'll defer to you guys working more directly on pytorch support here, but I'm sort of worried that we'll miss some real issues if we relax the test environment this much. Looking at https://github.com/ROCm/frameworks-internal/issues/16941 (a private issue, so hard to reference from public PRs like this...), there is at least one case
> `test_utils.py::TestStandaloneCPPJIT::test_load_standalone` → `libroctracer64.so.4: cannot open shared object file` — [28071362663 / 83106472307](https://github.com/ROCm/TheRock/actions/runs/28071362663/job/83106472307) (2/6)

that really feels like the code itself should be changed, not the test environment

PR #5874, opened 2026-06-16
detect generic-only Gfxarch artifacts in kpack

arvindcheru:
## Motivation

This pull request refines how the build and packaging tools determine whether a package is considered "gfxarch" (graphics architecture-specific), especially in multi-architecture (kpack) scenarios. The changes ensure that classification as a gfxarch package now requires both the appropriate metadata and the presence of architecture-specific artifacts, improving accuracy and robustness in package handling.

These changes collectively make the packaging system more robust and accurate when handling complex multi-architecture scenarios, especially for packages that blur the line between generic and arch-specific content.

## Technical Details

Key changes include:
**Refinement of gfxarch package detection:**

- The `is_gfxarch_package` function now requires an explicit `artifacts_dir` parameter in kpack mode, and only returns `True` if architecture-specific artifact directories actually exist for the package. This is enforced via a new helper function `_has_arch_specific_artifacts`. [[1]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L209-R254) [[2]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L227-R269)

**Consistent API updates across the codebase:**

- All calls to `is_gfxarch_package` throughout the packaging utilities and build logic are updated to pass the `artifacts_dir` parameter, ensuring consistent and correct behavior when determining package type. [[1]](diffhunk://#diff-8d3460579f955a151f676758348a06c2a320b53c796971d9d3c80a8d9202a14bL136-R136) [[2]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L388-R433) [[3]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L586-R633) [[4]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L608-R650) [[5]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L836-R886) [[6]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L940-R986) [[7]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L991-R1037) [[8]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L1073-R1119) [[9]](diffhunk://#diff-b3621e1bb54f31ceebf1eb353540dbd531facd59d6a9e84d7036a469d7994229L1118-R1166)

## Test Result

https://github.com/ROCm/TheRock/actions/runs/27591257894
https://github.com/ROCm/TheRock/actions/runs/27837045021

## Submission Checklist

JIRA ID : ROCM-25383

### DIFF
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -133,7 +133,10 @@ def build_package_variants(pkg_name, config: PackageConfig) -> list:
     pkg_info = get_package_info(pkg_name)  # Raises ValueError if not found
 
     if config.enable_kpack:
-        if is_gfxarch_package(pkg_info, config.enable_kpack):
+        _probe = kpack_gfx_probe_kwargs(
+            config.enable_kpack, config.artifacts_dir, config.gfxarch_list
+        )
+        if is_gfxarch_package(pkg_info, config.enable_kpack, **_probe):
             # GfxArch=True: host + devices + meta + non-versioned
             return build_gfxarch_package_variants(pkg_name, config)
         else:
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -133,7 +133,7 @@ def build_package_variants(pkg_name, config: PackageConfig) -> list:
     pkg_info = get_package_info(pkg_name)  # Raises ValueError if not found
 
     if config.enable_kpack:
-        if is_gfxarch_package(pkg_info, config.enable_kpack):
+        if is_gfxarch_package(pkg_info, config.enable_kpack, config.artifacts_dir):
             # GfxArch=True: host + devices + meta + non-versioned
             return build_gfxarch_package_variants(pkg_name, config)
         else:
@@ -547,7 +547,6 @@ def create_package_config(args: argparse.Namespace) -> PackageConfig:
 
     # Auto-detect kpack from manifest if not explicitly requested via --enable-kpack
     artifacts_dir = Path(args.artifacts_dir).resolve()
-    set_artifacts_dir(artifacts_dir)
     if not args.enable_kpack:
         args.enable_kpack = load_kpack_from_manifest(artifacts_dir)
         if args.enable_kpack:
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -133,10 +133,7 @@ def build_package_variants(pkg_name, config: PackageConfig) -> list:
     pkg_info = get_package_info(pkg_name)  # Raises ValueError if not found
 
     if config.enable_kpack:
-        _probe = kpack_gfx_probe_kwargs(
-            config.enable_kpack, config.artifacts_dir, config.gfxarch_list
-        )
-        if is_gfxarch_package(pkg_info, config.enable_kpack, **_probe):
+        if is_gfxarch_package(pkg_info, config.enable_kpack):
             # GfxArch=True: host + devices + meta + non-versioned
             return build_gfxarch_package_variants(pkg_name, config)
         else:
@@ -550,6 +547,7 @@ def create_package_config(args: argparse.Namespace) -> PackageConfig:
 
     # Auto-detect kpack from manifest if not explicitly requested via --enable-kpack
     artifacts_dir = Path(args.artifacts_dir).resolve()
+    set_artifacts_dir(artifacts_dir)
     if not args.enable_kpack:
         args.enable_kpack = load_kpack_from_manifest(artifacts_dir)
         if args.enable_kpack:
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -881,7 +881,9 @@ def filter_components_fromartifactory(
             dir_suffix = "generic"
     else:
         dir_suffix = (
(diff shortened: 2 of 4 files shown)


raramakr (approved):
LGTM , Pls verify single arch build as well before merging

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -224,11 +262,11 @@ def is_gfxarch_package(pkg_info, enable_kpack=False):
         if pkgname.endswith("-devel") and not is_meta_package(pkg_info):
             return False
 
-        # Override RCCL Gfxarch behavior in kpack mode
```
nunnikri, line 227: 1. If enable_kpack=True but artifacts_dir is None (e.g., set_artifacts_dir wasn't called, or called with None), the function falls through silently with no return — returning None implicitly. That's falsy, meaning RCCL would get generic artifacts instead of arch-specific ones. This is a regression from the old hardcoded return True.
2. There's no explicit return False — the implicit None return is a bug waiting to surface.

Required action: Add return False after the if enable_kpack: block, and document what the caller should expect when artifacts_dir is unset.

build_tools/packaging/linux/build_package.py
```diff
@@ -547,6 +547,7 @@ def create_package_config(args: argparse.Namespace) -> PackageConfig:
 
     # Auto-detect kpack from manifest if not explicitly requested via --enable-kpack
     artifacts_dir = Path(args.artifacts_dir).resolve()
+    set_artifacts_dir(artifacts_dir)
```
nunnikri: set_artifacts_dir() is called in build_package.py as a side effect before load_kpack_from_manifest(). This means _has_arch_specific_artifacts silently depends on set_artifacts_dir having been called first — if call order changes, the behavior breaks with no error.

Better: pass artifacts_dir as a parameter directly to _has_arch_specific_artifacts instead of threading it through a global. The caller already has it.

nunnikri: globals().get() works but implies _artifacts_dir might not exist as a name — it should always exist if initialized at module level. This is a code smell that will confuse readers.

nunnikri (commented):
Comments from agent

build_tools/packaging/linux/packaging_utils.py
arvindcheru: Thanks for the comments. Fixed in latest version

build_tools/packaging/linux/build_package.py
arvindcheru: Thanks for the comments, removed usage of global, instead added parameter artifact dir in latest version

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -206,16 +206,52 @@ def is_packaging_disabled(pkg_info):
...
+
+    Returns:
+    bool: True if any arch-specific artifact directory exists, False otherwise.
+    """
+    artifactory = pkg_info.get("Artifactory")
+    if not artifactory:
+        return False
+
+    for artifact in artifactory:
+        artifact_prefix = artifact["Artifact"]
```
nunnikri:
```suggestion
        artifact_prefix = artifact.get("Artifact")
        if not artifact_prefix:
            continue

```

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -206,16 +206,52 @@ def is_packaging_disabled(pkg_info):
     return is_key_defined(pkg_info, "Disablepackaging")
 
 
-def is_gfxarch_package(pkg_info, enable_kpack=False):
+def _has_arch_specific_artifacts(pkg_info, artifacts_dir):
+    """Check if arch-specific artifact directories exist for a package.
```
nunnikri: The doc string as per coding guidelines should be like this.
```
def _has_arch_specific_artifacts(pkg_info, artifacts_dir):
    """Check if arch-specific artifact directories exist for a package.

    Args:
        pkg_info (dict):  ....
        artifacts_dir (Path): ....

       """
Key differences:

Args: not Parameters:
Each arg indented 4 spaces, continuation lines indented 8
Blank line after the one-line summary before Args:

nunnikri (approved):
Minor suggestions given.

build_tools/packaging/linux/packaging_utils.py
arvindcheru: Thanks, updated review comments in the latest version

build_tools/packaging/linux/packaging_utils.py
arvindcheru: Thanks, updated review comments in the latest version

arvindcheru:
Override reason: New pre-review flag was introduced after this PR was reviewed & approved. (Flag applied 
during rebase doesn't reflect actual review state.)

PR #5896, opened 2026-06-17
Move aux-overlay compiler symlinks and cmake configs to amd-llvm_dev

raramakr:
The aux-overlay stage contains symlinks and cmake redirect files that
  are part of the LLVM toolchain. Move these from base_lib to amd-llvm_dev:
  - bin/ (amdclang, amdflang, amdlld, offload-arch symlinks)
  - lib/cmake/ (AMDDeviceLibs, clang, lld, llvm config files)
  - llvm and amdgcn symlinks

  Only share/therock/ (dist_info.json, therock_manifest.json) remains in base_lib

### DIFF
--- a/compiler/artifact-amd-llvm.toml
+++ b/compiler/artifact-amd-llvm.toml
@@ -58,3 +58,13 @@ exclude = [
 [components.dbg."compiler/hipcc/stage"]
 [components.dev."compiler/hipcc/stage"]
 [components.doc."compiler/hipcc/stage"]
+
+# aux-overlay symlinks and cmake redirects for LLVM toolchain
+[components.dev."base/aux-overlay/stage"]
+default_patterns = false
+include = [
+  "bin/**",
+  "lib/cmake/**",
+  "llvm",
+  "amdgcn",
+]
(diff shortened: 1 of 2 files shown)


compiler/artifact-amd-llvm.toml
```diff
@@ -58,3 +58,13 @@ exclude = [
 [components.dbg."compiler/hipcc/stage"]
 [components.dev."compiler/hipcc/stage"]
 [components.doc."compiler/hipcc/stage"]
+
+# aux-overlay symlinks and cmake redirects for LLVM toolchain
+[components.dev."base/aux-overlay/stage"]
+default_patterns = false
+include = [
+  "bin/**",
```
nunnikri, line 66: Adding whole bin in dev doesnt look right. We add the specific ones here like bin/amd*.

compiler/artifact-amd-llvm.toml
```diff
@@ -58,3 +58,13 @@ exclude = [
...
+  "lib/cmake/**",
+  "llvm",
+  "amdgcn",
```
nunnikri, line 69: amdgcn should be in runtime not dev.

nunnikri (commented):
Can you also check the artifacts and confirm there is no difference in runtime package and dev package apart from additional amd* softlinks?

PR #5900, opened 2026-06-17
feat: extend security scanners with zizmor

cgoea:
## Motivation

 Workflow files are a common attack surface (script injection, unpinned actions, over-broad permissions, etc.). This PR adds [zizmor](https://docs.zizmor.sh/) as a workflow/action security auditor. 
JIRA ID : ROCM-28051
 
## Technical Details

- **Reusable workflow** `.github/workflows/zizmor.yml` — `workflow_call` entry
  point with `scan_mode` (`changed`/`all`), `report_formats`, `scan_path`,
  `severity_threshold`, and `persona` inputs. 
- **Post-merge workflow** `.github/workflows/zizmor_main.yml` — runs on push to
  `main` and `workflow_dispatch` with `scan_mode: all`, uploading SARIF to the
  code-scanning Security tab (filterable by `Tool: zizmor`).
- **PR gate** — added a `zizmor` job to `.github/workflows/pre_commit_security.yml`.
- **Config** `zizmor.yml` — repo-root config with a central `rules:` block for
  reviewable, centralized suppressions.

## Test Plan

- Run the new unit suite: `python3 -m unittest build_tools/scan_tools/github_actions/tests/zizmor_test.py -v`

## Test Result

- Unit tests: **21 passed** (`Ran 21 tests ... OK`).

## Submission Checklist


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,147 @@
...
+        description: >-
+          PEM-encoded private key paired with 'gh_app_id'. See
+          'gh_app_id' for usage details.
+        required: false
+
+permissions:
+  contents: read
+
+jobs:
+  scan:
```
haribabug, line 70: Include timeout-minutes: 20 or equivalent value

haribabug (approved):

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,147 @@
...
+          GitHub App ID used to mint an installation access token for
+          zizmor's online audits (e.g. the 'unpinned-uses' audit
+          resolves action SHAs against the GitHub API). Optional:
+          when both 'gh_app_id' and 'gh_app_private_key' are passed
+          the reusable workflow generates a short-lived App token
+          and exports it as 'GH_TOKEN' to zizmor; otherwise it falls
+          back to the calling workflow's default GITHUB_TOKEN, which
+          has only the permissions granted to this reusable workflow
+          (minimum: contents: read). Fork PRs that can't read the
+          base repo's secrets land on the fallback automatically.
```
marbre: My agent is flagging that neither gh_app_id nor gh_app_private_key are passed to pre_commit_security.zml or zizmor_main.zml (or vice versa). Seems both paths always fall back to github.token. Docs here + the step to mint an app token seem therefore to be dead. Needs a caller or just drop it?

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,645 @@
...
+        "Zizmor fail threshold: severity >= %s (lower-severity findings "
+        "still appear in reports)",
+        args.severity_threshold.upper(),
+    )
+
+    sarif_target = next((t for t in targets if t.fmt == "sarif"), None)
+    non_sarif = [t for t in targets if t.fmt != "sarif"]
+
+    if files is not None and not files:
+        gha_set_output({"sarif_path": "", "non_sarif_paths": ""})
```
marbre, line 634: From my agent:

gha_set_output({sarif_path, ...}) is called before _run_zizmor. The upload steps use if: always() && steps.scan.outputs.sarif_path != ''. If zizmor errors out (return 2) before the SARIF file is written, upload-sarif will still fire with a non-empty path pointing at a missing file, failing that step. Consider only emitting sarif_path after the file is confirmed written, or guard the upload on file existence

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,645 @@
...
+def _is_audited_path(relpath: str) -> bool:
+    norm = relpath.replace(os.sep, "/")
+    for pattern in _AUDITED_PATTERNS:
+        if fnmatch.fnmatchcase(norm, pattern):
+            return True
+        if pattern.startswith("**/") and fnmatch.fnmatchcase(
+            norm, pattern[len("**/") :]
+        ):
+            return True
+    return False
```
marbre: From my agent:

_is_audited_path uses fnmatch.fnmatchcase with patterns like .github/workflows/*.yml. Unlike shell globs, fnmatch's * matches /, so .github/workflows/sub/nested.yml would match. Harmless here (nested workflow dirs aren't a thing), but the pattern doesn't mean what it looks like. A comment or a stricter check would prevent surprise.

.github/workflows/zizmor.yml
cgoea: leftover from the sandbox. removed

build_tools/scan_tools/github_actions/zizmor.py
cgoea: fixed now.

.github/workflows/pre_commit_security.yml
```diff
@@ -24,3 +24,10 @@ jobs:
     uses: ./.github/workflows/gitleaks.yml
     with:
       report_formats: csv
+
+  # Audit workflow/action/dependabot files introduced by this PR
+  # (`scan_mode: changed`, the reusable workflow's default).
```
marbre: The `files introduced by this PR` seems off. Drop the entire comment?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,118 @@
...
+        id: depth
+        uses: ./.github/actions/compute_pr_fetch_depth
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: ${{ steps.depth.outputs.value }}
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
```
marbre:
```suggestion
        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1# v6.3.0
```

v5 isn't a specific version and rather a floating tag. Let's rather pin down to a specific version and let's pick a not already outdated one.

.github/workflows/zizmor_main.yml
```diff
@@ -0,0 +1,31 @@
...
+    uses: ./.github/workflows/zizmor.yml
+    with:
+      scan_mode: all
+      report_formats: sarif
+    permissions:
+      contents: read
+      # Required so the reusable workflow can call
+      # github/codeql-action/upload-sarif and have the findings
+      # appear under Security -> Code scanning
+      security-events: write
```
marbre, line 35: This only passes the permissions to the reusable workflow but we lack defining those at the top level. Move (?) or declare at top level.

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,662 @@
...
+    user_json = next((t for t in user_targets if t.fmt == "json"), None)
+    if user_json is not None:
+        tally_path = user_json.path
+        runs: list[_ReportTarget] = list(user_targets)
+    else:
+        tally_path = Path(_INTERNAL_TALLY_PATH)
+        runs = [*user_targets, _ReportTarget(fmt="json", path=tally_path)]
+
+    for tgt in runs:
+        cmd = [*base_args, "--format", tgt.fmt]
```
marbre: zizmor is run N+1 times here. Okay for a single changed file but heavy for a full repo-scan. Any way to run it just once? E.g. does zizmor support multiple formats in a single-shot or can we convert later?

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,662 @@
...
+            }
+        )
+        counts = _tally_findings_by_severity(tally_path)
+    except RuntimeError as exc:
+        log.error("%s", exc)
+        _emit_non_sarif_reports(non_sarif)
+        return 2
+    finally:
+        if internal_tally_used:
+            Path(_INTERNAL_TALLY_PATH).unlink(missing_ok=True)
```
marbre: From my agent:

This runs even when `_run_zizmor` raises, before the tally file is written, so `unlink(missing_ok=True)` is safe. But if in a future refactor the tally is written earlier, the finally block would delete it before the error is surfaced. Low risk as written, but the pattern is slightly fragile.

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,662 @@
...
+    if summary_chunks:
+        gha_append_step_summary("\n\n".join(summary_chunks))
+
+
+def _ensure_zizmor() -> Path:
+    spec = f"zizmor=={_ZIZMOR_VERSION}"
+    log.info("Installing %s", spec)
+    try:
+        subprocess.run(
+            [sys.executable, "-m", "pip", "install", "--quiet", "--upgrade", spec],
```
marbre:
```suggestion
            [sys.executable, "-m", "pip", "install", "--quiet", spec],
```
zizmor is already pinned and I don't think `--upgrade` is useful here.

If we want to further harden we could go beyond and pass `--require-hashes` + `--hash=...` . That introduces a further maintenance burden but might be worth to consider here to mitigate CDN-level tampering attacks between PyPI and the runner. @haribabug wdyt?

marbre (commented):
Looks almost good. Please address the remaining pieces and I guess than it is good to go.

.github/workflows/zizmor_main.yml
cgoea: Yes, by design, top level minimal permissions, each workflow enhances permissions on need basis.

cgoea: LE: you meant workflow top level, not the top level workflow. got it.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,118 @@
...
+        id: depth
+        uses: ./.github/actions/compute_pr_fetch_depth
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: ${{ steps.depth.outputs.value }}
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1# v6.3.0
```
marbre, line 91:
```suggestion
        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
```

.github/workflows/zizmor.yml
marbre, line 1: Have we considered using https://github.com/zizmorcore/zizmor-action?

build_tools/scan_tools/github_actions/zizmor.py
```diff
@@ -0,0 +1,693 @@
...
+
+
+def _ensure_zizmor() -> Path:
+    spec = f"zizmor=={_ZIZMOR_VERSION}"
+    log.info("Installing %s", spec)
+    try:
+        subprocess.run(
+            [sys.executable, "-m", "pip", "install", "--quiet", spec],
+            check=True,
+        )
```
marbre, line 119: We may still want to hash the install alongside with pinning the version? At least file a follow up issue?

.github/workflows/zizmor.yml
cgoea: Yes, it needs a license. Overcomplicates a simple scan

.github/workflows/zizmor.yml
marbre: This is still unaddressed in the most recent version but needs to be fixed.

build_tools/scan_tools/github_actions/zizmor.py
marbre: @cgoea can we file a follow up issue and link it here?

marbre (changes requested):
Ping. Latest review comments still unaddressed and checkout will fail. Let's get this ready to land @cgoea.

cgoea:
Moving to rocm-security-gh repo. Closing this PR because of that.

PR #5903, opened 2026-06-17
[ci] Add TEST_SUBPROJECTS to debug-tools components

lumachad:
## Motivation

Declare test-selection mappings so determine_rocm_test_dependencies.py can pick the right tests when a debug-tools subproject changes:
  - amd-dbgapi       -> rocgdb, rocr-debug-agent-tests (both depend on it)
  - rocr-debug-agent -> rocr-debug-agent-tests
  - rocgdb           -> empty (tests only itself)

rocr-debug-agent-tests gets no TEST_SUBPROJECTS entry: its sources live inside the rocr-debug-agent tree (EXTERNAL_SOURCE_DIR .../rocr-debug-agent/test), so the manifest diff never reports it as a changed project and the key would never be looked up. The tests are still selected via the rocr-debug-agent and amd-dbgapi mappings above.

Empty TEST_SUBPROJECTS entries are placed last in the declare so the parser's greedy dependency-list match does not consume a later keyword.

Requires the hyphen-aware parser fix to be effective for these hyphenated subproject names.

## Test Plan

Validated during multi-arch CI execution.

## Test Result

To be confirmed once multi-arch CI cross-triggers work.

## Submission Checklist

### DIFF
--- a/debug-tools/CMakeLists.txt
+++ b/debug-tools/CMakeLists.txt
@@ -89,6 +89,9 @@ if(THEROCK_ENABLE_AMD_DBGAPI)
       "lib"
     INTERFACE_PKG_CONFIG_DIRS
       "share/pkgconfig"
+    TEST_SUBPROJECTS
+      rocgdb
+      rocr-debug-agent-tests
   )
 
   therock_cmake_subproject_glob_c_sources(amd-dbgapi SUBDIRS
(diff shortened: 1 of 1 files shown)


geomin12 (approved):

lumachad:
@geomin12 Thanks. I'll wait for https://github.com/ROCm/TheRock/pull/5902 to be merged before merging this one.

Issue #2616, opened 2025-12-18
[ci] Fixing TheRock test filters


geomin12:
Currently, these tests are having timeouts and incredibly slow tests comparable to its rocm-ci and Azure CI counterparts

- hipblas (takes 5 mins on rocm-ci, around 1h in TheRock CI - determine machine issue)
- hipsparse (takes 1 hour to run tests, with constant timeouts. need to reduce test set)
- rccl (tests are hanging, as we lack 4gpu setup. currently in progress as we got a 4gpu setup)
- rocsparse (tests take 6 hours to run, using the `quick` filter)
    - opened internal ticket SWDEV-572770 with discussion
- hipsparselt (smoke tests take 30 min+, quick tests take 6-8 hours, full tests probably take even longer) - something strange has happened with hipsparselt and is now taking an incredibly amount of time.

We want to investigate why these tests take so long and how to reduce the test times (whether working with component team or resolving machine issue)

geomin12:
rccl (tests are hanging, as we lack 4gpu setup. currently in progress as we got a 4gpu setup) - resolved with #2831

geomin12:
- hipsparselt (smoke tests take 30 min+, quick tests take 6-8 hours, full tests probably take even longer) - ticket open with SWDEV-574424 / SWDEV-572768 resolved

PR #5918, opened 2026-06-17
Update fetch_test_configurations.py for hipblas

TorreZuk:
## Motivation

This pull request increases the test coverage and reliability for the `hipblas` job by extending its timeout and enabling test sharding for both Linux and Windows platforms.  The TODO issue was closed so clearing that out.   Comprehensive hits time limits for tests so 2 shards and 1 hour required.

Test configuration improvements for `hipblas`:

* Increased the `timeout_minutes` for the `hipblas` job from 30 to 60 to allow more time for tests to complete.
* Updated `total_shards_dict` for `hipblas` to run tests in 2 shards on both Linux and Windows, improving parallelism and coverage.

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -209,13 +209,12 @@ test_matrix = {
     "hipblas": {
         "job_name": "hipblas",
         "fetch_artifact_args": "--blas --tests",
-        "timeout_minutes": 30,
+        "timeout_minutes": 60,
         "test_script": f"python {_get_script_path('test_runner.py')}",
         "platform": ["linux", "windows"],
-        # TODO(#2616): Enable full tests once known machine issues are resolved
         "total_shards_dict": {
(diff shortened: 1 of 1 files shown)


geomin12 (approved):
lgtm for comprehensive. we'll start implementing timeouts on a per level basis: https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md

marbre:
Removing myself as Geo is more into this and has already approved. Therefore should be good to land.

tony-davis (approved):
Approving. This is a low-risk, CI-config-only change and it's doing the right thing: it re-enables fuller hipblas testing rather than trimming it, and the TODO(#2616) it removes points at an issue that's genuinely closed. The consumer unit tests in `fetch_test_configurations_test.py` don't hardcode hipblas's timeout or shard values, so nothing there breaks. CI is still running, so worth a glance at the rollup before merge, but the change itself looks good. Thanks Torre!

Issue #5919, opened 2026-06-17
[Issue]: [ASAN]hipcc_check test fails: ASan runtime library ordering error


raramakr:
### Problem Description

The test_rocm_sanity test fails when executing the compiled hipcc_check binary. The test successfully compiles the HIP test program using hipcc with
  --offload-arch=gfx942, but execution fails with an AddressSanitizer (ASan) runtime error:

  ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD

CI job: https://github.com/ROCm/TheRock/actions/runs/27521358174/job/81374619372
**Error:**
tests/test_rocm_sanity.py::TestROCmSanity::test_hip_printf 
-------------------------------- live log call ---------------------------------
INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [None]$ /__w/TheRock/TheRock/build/lib/llvm/bin/offload-arch
INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [/__w/TheRock/TheRock/build/bin]$ /__w/TheRock/TheRock/build/bin/hipcc /__w/TheRock/TheRock/tests/hipcc_check.cpp -Xlinker -rpath=/__w/TheRock/TheRock/build/bin/../lib/ --offload-arch=gfx942 -o hipcc_check
INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [/__w/TheRock/TheRock/build/bin]$ ./hipcc_check
ERROR    test_rocm_sanity:test_rocm_sanity.py:38 Command failed!
ERROR    test_rocm_sanity:test_rocm_sanity.py:39 command stdout:
ERROR    test_rocm_sanity:test_rocm_sanity.py:42 command stderr:
ERROR    test_rocm_sanity:test_rocm_sanity.py:44 ==1113==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
FAILED  

### Operating System

Ubuntu 24.04

### CPU

NA

### GPU

NA

### ROCm Version

7.14

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

<details>
<summary>rocminfo --support output</summary>

```
Paste output here
```

</details>

sa-faizal:
### New occurrence, different component

[...]

PR #5920, opened 2026-06-17
fix: Set LD_PRELOAD for ASan runtime in test workflows

raramakr:
The ASan runtime library must be loaded first in the library list,
  otherwise tests fail with:
  "ASan runtime does not come first in initial library list"

  Use clang++ --print-file-name to dynamically resolve the ASan library
  path and set LD_PRELOAD before test execution.

Jira ID: https://github.com/ROCm/TheRock/issues/5919

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -153,8 +153,6 @@ jobs:
           echo "ASAN_OPTIONS=detect_odr_violation=0:quarantine_size_mb=600" >> $GITHUB_ENV
           echo "HSA_XNACK=1" >> $GITHUB_ENV
           echo "ASAN_SYMBOLIZER_PATH=${{ env.OUTPUT_ARTIFACTS_DIR }}/llvm/bin/llvm-symbolizer" >> $GITHUB_ENV
-          ASAN_LIB=$(${{ env.OUTPUT_ARTIFACTS_DIR }}/lib/llvm/bin/clang++ --print-file-name=libclang_rt.asan-x86_64.so)
-          echo "LD_PRELOAD=${ASAN_LIB}" >> $GITHUB_ENV
 
       - name: Test
         id: test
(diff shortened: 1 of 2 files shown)


.github/workflows/test_component.yml
```diff
@@ -153,6 +153,8 @@ jobs:
           echo "ASAN_OPTIONS=detect_odr_violation=0:quarantine_size_mb=600" >> $GITHUB_ENV
           echo "HSA_XNACK=1" >> $GITHUB_ENV
           echo "ASAN_SYMBOLIZER_PATH=${{ env.OUTPUT_ARTIFACTS_DIR }}/llvm/bin/llvm-symbolizer" >> $GITHUB_ENV
+          ASAN_LIB=$(${{ env.OUTPUT_ARTIFACTS_DIR }}/lib/llvm/bin/clang++ --print-file-name=libclang_rt.asan-x86_64.so)
```
geomin12: are we sure EACH component test needs this? we already added it to tests that need it: https://github.com/ROCm/TheRock/blob/2d417fa49bc4e657bcbcf0ae549238784175f0fe/build_tools/github_actions/test_executable_scripts/test_hiptests.py#L146 from here https://github.com/search?q=repo%3AROCm%2FTheRock+LD_PRELOAD+language%3APython&type=code&l=Python

raramakr: The above change was merged a few months back. The issue was also observed in the June 14th build: https://github.com/ROCm/TheRock/actions/runs/27521358174/job/81374619372.
Rather than adding it in individual test scripts, it is better to set it from the top-level script.

It may not be required for all test cases, but many of them might require this setting.
rocprofiler sdk was also settting the same

raramakr: Will move LD_PRELOAD from env  and use it where ever its required. Setting it as env variable will result in false positive leaks from python and bash

PR #5922, opened 2026-06-17
ci(compiler): Build a dedicated static LLVM for amd-comgr

chinmaydd:
## Motivation

amd-comgr is built with `COMGR_STATIC_LLVM=ON` to embed a single, isolated, symbol-hidden copy of LLVM (so it can be dlopen'd into arbitrary processes without interposing with their LLVM). 

But the shipped amd-llvm build uses `LLVM_LINK_LLVM_DYLIB=ON`, whose exported clang/lld static libraries carry an INTERFACE dependency on libLLVM.so:

```
  # amd-llvm dist ClangTargets.cmake
  set_target_properties(clangBasic PROPERTIES
    INTERFACE_LINK_LIBRARIES "LLVM")
```

Linking those while `COMGR_STATIC_LLVM` also puts the LLVM component static archives on comgr's link line pulls TWO copies of LLVM into libamd_comgr.so, whose independent global state (`cl::opt` registry, ManagedStatic, Target/Pass registries) corrupts comgr's in-process compiler at runtime.

## Technical Details

Add a second, build-only LLVM/Clang/LLD subproject (amd-llvm-static) built from the same sources with `LLVM_LINK_LLVM_DYLIB=OFF`, and link amd-comgr against it instead of the dylib amd-llvm. 

Now, comgr ends up with exactly one copy. This keeps the shipped toolchain dylib-linked (small clang/lld/flang binaries); the static build is not shipped, and since dist trees are merged along RUNTIME_DEPS only, it stays out of comgr's dist.

---

Trade-off: this adds a second LLVM compile to the build, but shipped artifacts are unchanged. The alternative (flip the whole toolchain to LLVM_LINK_LLVM_DYLIB=OFF) avoids the extra build but bloats every clang/lld/flang binary.

---

JIRA ID : ROCM-25975

## Test Plan

CI for all related components should pass.

## Submission Checklist


mrezavand:
@lamb-j @Jonathan03ant  I believe the issue here will benefit from the same solution: https://github.com/ROCm/ROCm/issues/6511

PR #5929, opened 2026-06-17
[ci] Enable OIDC for external repos, skip for fork PRs

### LINKED ISSUE #5654: [Issue] PR runs from forks are failing to upload to S3 with `Could not assume role with user credentials`
See https://github.com/ROCm/TheRock/pull/5531 and https://github.com/ROCm/TheRock/actions/runs/26984086713?pr=5531

Linux AWS: `Error: Could not assume role with user credentials: User: arn:aws:sts::324352301041:assumed-role/therock-runners-prod-runner-s3-external/aws-sdk-js-session-1780614257367 is not authorized to perform: sts:TagSession on resource: arn:aws:iam::692859939525:role/therock-ci-external`

Windows Azure: `Error: Could not assume role with user credentials: User: arn:aws:iam::692859939525:user/therock-external-upload is not authorized to perform: sts:TagSession on resource: arn:aws:iam::692859939525:role/therock-ci-external`

Traced back to https://github.com/ROCm/TheRock/pull/5450#discussion_r3306986534. I'd suggest either reverting that or implementing the suggested changes from that review thread.

geomin12:
Add iam_role="therock-ci-external" back to therock-ci-artifacts-external bucket config to enable OIDC authentication for external repos like rocm-libraries.

An error was occuring for WSL ROCDXG as this is a github hosted runner and had no way to provide credentials for external repos

For fork PRs, the iam_role is set to None in
get_artifacts_bucket_config_for_workflow_run() so that OIDC is skipped and runner base credentials are used instead. This fixes issue #5654 where fork PRs failed with "Could not assume role" because they don't have the required OIDC trust relationship.

Summary:
- External repos (non-fork): Use OIDC with therock-ci-external role
- Fork PRs: Skip OIDC, use runner base credentials (Linux only)

Working in test fork PR: https://github.com/ROCm/TheRock/actions/runs/27710372755/job/81975007386?pr=5923
Working in external repo: https://github.com/ROCm/rocm-libraries/actions/runs/27709741844/job/81971020614

Truly resolves #5654 with proper external role plugged in now

Resolves errors found in https://github.com/ROCm/rocm-libraries/pull/8297

### DIFF
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -46,9 +46,9 @@ class S3BucketConfig:
 
 
 s3_bucket_configs = [
-    # CI (self-hosted runners include credentials for therock-ci-artifacts-external)
+    # CI (external repos use OIDC with therock-ci-external; fork PRs use runner base credentials)
     S3BucketConfig("therock-ci-artifacts", iam_role="therock-ci"),
-    S3BucketConfig("therock-ci-artifacts-external", iam_role=None),
+    S3BucketConfig("therock-ci-artifacts-external", iam_role="therock-ci-external"),
     # Release type "dev"
     S3BucketConfig("therock-dev-artifacts", iam_role="therock-dev"),
     S3BucketConfig("therock-dev-packages", iam_role="therock-dev"),
@@ -216,4 +216,18 @@ def get_artifacts_bucket_config_for_workflow_run(
         is_pr_from_fork=is_pr_from_fork,
     )
     _log(f"  bucket: {config.name}")
+
+    # For fork PRs, skip OIDC and use runner base credentials instead.
+    # Fork PRs cannot assume IAM roles via OIDC because they don't have
+    # the required trust relationship. Return a config without an IAM role
+    # so the configure-aws-credentials step is skipped.
+    if is_pr_from_fork and config.iam_role is not None:
+        _log("  Fork PR detected, skipping OIDC (using runner base credentials)")
+        config = S3BucketConfig(
+            name=config.name,
+            region=config.region,
+            iam_account=config.iam_account,
+            iam_role=None,
+        )
+
     return config
--- a/build_tools/tests/s3_buckets_test.py
+++ b/build_tools/tests/s3_buckets_test.py
@@ -39,12 +39,16 @@ class TestGetArtifactsBucketConfig(unittest.TestCase):
             release_type="ci", repository="ROCm/TheRock", is_pr_from_fork=True
         )
         self.assertEqual(config.name, "therock-ci-artifacts-external")
+        # The raw lookup returns the external role for forks; the OIDC skip for
+        # forks happens in get_artifacts_bucket_config_for_workflow_run, not here.
+        self.assertEqual(config.iam_role, "therock-ci-external")
 
     def test_ci_external_repo(self):
         config = get_artifacts_bucket_config(
             release_type="ci", repository="ROCm/rocm-libraries", is_pr_from_fork=False
         )
         self.assertEqual(config.name, "therock-ci-artifacts-external")
+        self.assertEqual(config.iam_role, "therock-ci-external")
 
     def test_release_type_dev(self):
         config = get_artifacts_bucket_config(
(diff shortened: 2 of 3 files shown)


tony-davis (changes requested):
Thanks for tracking this down and re-enabling OIDC the right way, Geo. The logic looks correct to me: both production callers (`write_artifacts_bucket_info.py` and `workflow_outputs.py`) go through `get_artifacts_bucket_config_for_workflow_run`, so the fork-PR role strip applies everywhere it's consumed, and external repos keep the `therock-ci-external` role. One blocking item before I approve.

**Needs a regression test (blocking).** This is a defect fix (#5654) in product code, but the new behavior is currently uncovered. The existing `test_workflow_run_from_fork` only asserts `config.name`, so it passes both before and after this change — the fork-PR OIDC skip and the re-added external role would not be caught if they regressed again. The good news is `build_tools/tests/s3_buckets_test.py` already has all the scaffolding, so this is a ~15-minute add. I verified the tests below fail on `main` and pass on this branch.

In `TestGetArtifactsBucketConfig`, tighten the two existing cases:

```python
def test_ci_fork_pr(self):
    config = get_artifacts_bucket_config(
        release_type="ci", repository="ROCm/TheRock", is_pr_from_fork=True
    )
    self.assertEqual(config.name, "therock-ci-artifacts-external")
    # The raw lookup returns the external role for forks; the OIDC skip for
    # forks happens in get_artifacts_bucket_config_for_workflow_run, not here.
    self.assertEqual(config.iam_role, "therock-ci-external")

def test_ci_external_repo(self):
    config = get_artifacts_bucket_config(
        release_type="ci", repository="ROCm/rocm-libraries", is_pr_from_fork=False
    )
    self.assertEqual(config.name, "therock-ci-artifacts-external")
    self.assertEqual(config.iam_role, "therock-ci-external")
```

In `TestGetArtifactsBucketConfigForWorkflowRun`, extend the fork case and add two more (this is the coverage that actually locks in the fix):

```python
def test_workflow_run_from_fork(self):
    fake_run = {
        "id": 12345,
        "head_repository": {"full_name": "SomeUser/TheRock"},
    }
    config = get_artifacts_bucket_config_for_workflow_run(
        github_repository="ROCm/TheRock", workflow_run=fake_run
    )
    self.assertEqual(config.name, "therock-ci-artifacts-external")
    # Fork PRs cannot assume an IAM role via OIDC (no trust relationship),
    # so the wrapper must strip the role and fall back to runner base
    # credentials. Regression coverage for #5654.
    self.assertIsNone(config.iam_role)
    self.assertIsNone(config.write_access_iam_role)

def test_workflow_run_external_repo_uses_oidc(self):
    # An external (non-fork) repo such as rocm-libraries keeps the
    # therock-ci-external role so it can authenticate via OIDC.
    fake_run = {
        "id": 12345,
        "head_repository": {"full_name": "ROCm/rocm-libraries"},
    }
    config = get_artifacts_bucket_config_for_workflow_run(
        github_repository="ROCm/rocm-libraries", workflow_run=fake_run
    )
    self.assertEqual(config.name, "therock-ci-artifacts-external")
    self.assertEqual(config.iam_role, "therock-ci-external")

def test_workflow_run_same_repo_keeps_internal_role(self):
    # A same-repo (non-fork) ROCm/TheRock PR is unaffected by the fork skip
    # and keeps the internal therock-ci role.
    fake_run = {
        "id": 12345,
        "head_repository": {"full_name": "ROCm/TheRock"},
    }
    config = get_artifacts_bucket_config_for_workflow_run(
        github_repository="ROCm/TheRock", workflow_run=fake_run
    )
    self.assertEqual(config.name, "therock-ci-artifacts")
    self.assertEqual(config.iam_role, "therock-ci")
```

**One non-blocking question:** #5654 traced the original failure to a `sts:TagSession` denial on `therock-ci-external` itself, which is why it was set to `None`. Can you confirm in the description that the role's trust policy/permissions have since been fixed (the linked green external run suggests so)? That makes it clear we're not re-introducing the same denial for the external-repo path.

The three red checks (`CI Summary`, hipBLAS gfx94X, Windows sanity) don't look attributable to this change — for a non-fork `ROCm/TheRock` PR the code path is unchanged — but worth a quick confirm they're known-flaky/green on `main` before merge.

tony-davis (approved):
Approving — the blocking item from my last pass is resolved. `f18dfd0` adds substantive regression coverage in `build_tools/tests/s3_buckets_test.py`, and I confirmed each new/tightened test pins a specific behavior that would fail if the fix regressed:

- `test_workflow_run_from_fork` now asserts `iam_role` and `write_access_iam_role` are both `None` — directly locks in the fork-PR OIDC skip (#5654).
- `test_workflow_run_external_repo_uses_oidc` asserts `rocm-libraries` keeps `therock-ci-external` — guards the re-added external role.
- `test_workflow_run_same_repo_keeps_internal_role` asserts same-repo `ROCm/TheRock` keeps `therock-ci` — guards against an over-broad strip.
- The tightened `test_ci_fork_pr` / `test_ci_external_repo` now assert the role from the raw lookup, not just the bucket name.

Thanks also for the `docs/development/s3_buckets.md` update documenting the external-repo vs fork-PR credential behavior — that wasn't required and it's a nice touch.

One non-blocking follow-up (doesn't need another review cycle): could you add a line to the PR description confirming the `therock-ci-external` trust policy / `sts:TagSession` permission has been fixed since #5654? The green external run you linked is good evidence it's resolved — just want it stated so it's clear we're not re-introducing the original denial on the external path.

PR #4603, opened 2026-04-16
[ci] Adding host-asan for postsubmit

geomin12:
Re-landing https://github.com/ROCm/TheRock/pull/4489 but with fixed unit tests!

PR #4500, opened 2026-04-13
[ci] Adjust expect_failure and expect_pytorch_failure logic

ScottTodd:
## Motivation

Handling a few related issues that will complicate multi-arch release pipeline work (https://github.com/ROCm/TheRock/issues/3334):

### 1. `expect_failure` plumbing

`expect_failure` was optionally set on a build _variant_ (e.g. release/asan/tsan). As of https://github.com/ROCm/TheRock/commit/3c528c064bc45175f39e5c0748c6505d8bc0d32c, it is _never set now_.

We were passing `expect_failure` all the way through to `multi_arch_build_portable_linux.yml` where stages are built, but it was ignored there. That workflow doesn't have a single job that it could mark as `continue-on-error` as the job that it replaces: https://github.com/ROCm/TheRock/blob/740c338f2e4d94c3aa930f4fc7ed73ee9822e291/.github/workflows/build_portable_linux_artifacts.yml#L61-L67

Solution: stop handling `expect_failure` altogether. We can add it back later as needed (somehow).

### 2. `expect_pytorch_failure` plumbing

`expect_pytorch_failure` was being read from the _build variant_ when it actually exists on a GPU family: https://github.com/ROCm/TheRock/blob/96c9642ef3cb601f7e67850a5c78e53322e41575/build_tools/github_actions/amdgpu_family_matrix.py#L178-L184 https://github.com/ROCm/TheRock/blob/96c9642ef3cb601f7e67850a5c78e53322e41575/build_tools/github_actions/configure_multi_arch_ci.py#L861

In multi-arch CI we currently build for all GPU families in a single workflow, with a matrix in `build_pytorch_wheels_per_family`. We could do two layers of filtering here: one for "should this workflow run build pytorch at all?" and another for "does the pytorch build for this GPU family work?"

Solution: correct the configuration plumbing, add TODO to handle in the matrix somehow

## Technical Details

I included a few other loosely related / incidental changes here (could move to a separate PR on request):

* Reordered arguments to `expand_build_configs`
* Folded `TestDualLabelRunnerSelection` tests into `TestExpandBuildConfigs` where they belong (and can reuse helper functions)

## Test Plan

* Updated unit tests
* Watch CI runs, including with a target like gfx906 that fails to build pytorch (would fail before, should still fail)

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5931, opened 2026-06-17
Remove defunct expect_[pytorch_]failure properties

ScottTodd:
## Motivation

Cleaning up properties that were used for per-family CI and did not get fully integrated into multi-arch CI as part of https://github.com/ROCm/TheRock/issues/3340.

> [!IMPORTANT]
> I'm not fully confident in these changes as I'm not as closely involved with ASan/TSan/etc. CI. I mainly want to pay down some of the tech debt here as other work is touching these files and apparently dead code is costly to work around.

See also:
* https://github.com/ROCm/TheRock/pull/4500 (an earlier attempt at this)
* https://github.com/ROCm/TheRock/pull/3982 (removed `expect_failure` for "tsan")
* https://github.com/ROCm/TheRock/pull/4603 (added `expect_failure` for "host-asan")
* https://github.com/ROCm/TheRock/pull/5920 (fixes test failures for "asan")
  * https://github.com/ROCm/TheRock/issues/5919

## Technical Details

* Conditions like
    ```yml
    # If we are expecting a build failure, do not run tests to save machine capacity.
    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).expect_failure == false }}
    ```    
    Don't make sense for multi-arch CI. If the build fails then the `failure()` check will skip the jobs
* We _may_ want to skip building rocm python packages, building pytorch packages, building rocm native linux packages, etc. for sanitizer variants. We have a different mechanism for that right now though: `build_native_linux=(suffix != "asan")`.
  * We could extend that to "host-asan" and "tsan" as needed
  * If we want to move that into `all_build_variants`, we could pick a different name like `build_python_packages: False` to express the intent more clearly.

## Test Plan

* CI on this PR will use the "release" variant
* We can also trigger "asan" and/or "host-asan" builds if we want to test them

## Submission Checklist


nunnikri (approved):
Looks fine to me. Can merge if ci runs fine. But would let @geomin12 to give the final approval

geomin12 (approved):

PR #5934, opened 2026-06-17
Scope PEP 503 index regeneration to the affected package

### LINKED ISSUE #5455: [Issue] Python package tests can fail if server-side index generation is too slow
## Problem description

Sample failure: https://github.com/ROCm/TheRock/actions/runs/26467183538/job/77959495159#step:5:48

```
Tue, 26 May 2026 21:11:24 GMT ++ Exec [B:\runner\_work\TheRock\TheRock]$ 'B:\runner\_work\TheRock\TheRock\.venv\Scripts\python.exe' -m pip install --find-links=https://therock-ci-artifacts.s3.amazonaws.com/26467183538-windows/python/index.html 'rocm[libraries,device-gfx1100,device-gfx1101,device-gfx1102,device-gfx1103]==7.14.0.dev0+3b16cb2c51ddf2aca9d84aefb1ac164b51708cd7'
Tue, 26 May 2026 21:11:25 GMT Looking in links: https://therock-ci-artifacts.s3.amazonaws.com/26467183538-windows/python/index.html
Tue, 26 May 2026 21:11:25 GMT ERROR: Could not find a version that satisfies the requirement rocm==7.14.0.dev0+3b16cb2c51ddf2aca9d84aefb1ac164b51708cd7 (from versions: 0.1.0)
Tue, 26 May 2026 21:11:25 GMT ERROR: No matching distribution found for rocm==7.14.0.dev0+3b16cb2c51ddf2aca9d84aefb1ac164b51708cd7
```

That package does exist on https://therock-ci-artifacts.s3.amazonaws.com/26467183538-windows/python/index.html **now**, but the build/upload job only ran less than a minute earlier: https://github.com/ROCm/TheRock/actions/runs/26467183538/job/77957529122
```
Tue, 26 May 2026 21:10:38 GMT [INFO] Uploading 11 total files to s3://therock-ci-artifacts/26467183538-windows/python:
Tue, 26 May 2026 21:10:38 GMT   rocm-7.14.0.dev0+3b16cb2c51ddf2aca9d84aefb1ac164b51708cd7.tar.gz
Tue, 26 May 2026 21:10:38 GMT   rocm_sdk_core-7.14.0.dev0+3b16cb2c51ddf2aca9d84aefb1ac164b51708cd7-py3-none-win_amd64.whl
```

If the test job gets scheduled and picks up a runner before the server-side lambda function finishes generating that index.html page, the tests will fail.

See also:
* https://github.com/ROCm/TheRock/pull/4900#discussion_r3156918282
* Various other similar test failures on and off since moving the index generation server side

## Potential solutions

I'd rather not introduce a fixed artificial delay in workflows. Consider a few options and their tradeoffs:

1. In build jobs: upload packages in a build job then wait
    * Test jobs can trust the package index to already exist
    * Test jobs can't start running/queuing until that wait completes
2. In test jobs: attempt to install, retry/wait if failed or index does not yet exist
    * Each test jobs needs to know that the index might not yet exist
    * Test jobs can start running/queuing immediately

marbre:
The Lambda that regenerates the PEP 503 index was sweeping the entire `v4/whl/` prefix on every S3 PUT, taking  up to 10 minutes. Because the flat multi-arch layout maps every wheel to the same prefix, a single upload triggered a full scan and regeneration of all package index pages.

To mitigate this, this patch introduces optional `package_name` and `update_root_index` parameters to the index generation in script. When `package_name` is set, the S3 listing is narrowed to `{prefix}/{package_name}-`,  restricting the index regeneration to only that specific wheel. The hyphen delimiter is on purpose as `torch` without it would also match e.g. `torchaudio`. When `update_root_index=False`, the root index write is skipped for routine version uploads where the package listing is unchanged.

Passing both `package_name` and `update_root_index` set to `True` raises a ValueError: the root index must be built from a full prefix listing, not a package-scoped one, as the scoped object list would clobber the root index with only that package's entries. The Lambda wrapper (updated in a companion PR) handles policy: existing packages get a scoped update; brand-new packages fall back to a full sweep so the root index correctly lists them.

All existing behavior is preserved: calling `update_pep503_index` with no new arguments performs the same full sweep as before. The Lambda wrapper is updated in a separate PR to pass the targeted parameters.

The revised `manage.py` and Lambda wrapper deployed this should resolve #5455.

### DIFF
--- a/build_tools/third_party/s3_management/manage.py
+++ b/build_tools/third_party/s3_management/manage.py
@@ -397,19 +397,20 @@ class S3Index:
         out.append(f'<!--TIMESTAMP {int(time.time())}-->')
         return '\n'.join(out)
 
-    def upload_pep503_htmls(self) -> None:
+    def upload_pep503_htmls(self, update_root_index: bool = True) -> None:
         for subdir in self.subdirs:
-            index_html = self.to_simple_packages_html(subdir=subdir)
-            for bucket in INDEX_BUCKETS:
-                print(f"INFO Uploading {subdir}/index.html to {bucket.name}")
-                bucket.Object(
-                    key=f"{subdir}/index.html"
-                ).put(
-                    # ACL='public-read',
-                    CacheControl='no-cache,no-store,must-revalidate',
-                    ContentType='text/html',
-                    Body=index_html
-                )
+            if update_root_index:
+                index_html = self.to_simple_packages_html(subdir=subdir)
+                for bucket in INDEX_BUCKETS:
+                    print(f"INFO Uploading {subdir}/index.html to {bucket.name}")
+                    bucket.Object(
+                        key=f"{subdir}/index.html"
+                    ).put(
+                        # ACL='public-read',
+                        CacheControl='no-cache,no-store,must-revalidate',
+                        ContentType='text/html',
+                        Body=index_html
+                    )
             for pkg_name in self.get_package_names(subdir=subdir):
                 compat_pkg_name = pkg_name.lower().replace("_", "-")
                 index_html = self.to_simple_package_html(subdir=subdir, package_name=pkg_name)
@@ -424,12 +425,13 @@ class S3Index:
                         Body=index_html
                     )
 
-    def save_pep503_htmls(self) -> None:
+    def save_pep503_htmls(self, update_root_index: bool = True) -> None:
         for subdir in self.subdirs:
-            print(f"INFO Saving {subdir}/index.html")
-            makedirs(subdir, exist_ok=True)
-            with open(path.join(subdir, "index.html"), mode="w", encoding="utf-8") as f:
-                f.write(self.to_simple_packages_html(subdir=subdir))
+            if update_root_index:
+                print(f"INFO Saving {subdir}/index.html")
+                makedirs(subdir, exist_ok=True)
+                with open(path.join(subdir, "index.html"), mode="w", encoding="utf-8") as f:
+                    f.write(self.to_simple_packages_html(subdir=subdir))
             for pkg_name in self.get_package_names(subdir=subdir):
                 makedirs(path.join(subdir, pkg_name), exist_ok=True)
                 with open(path.join(subdir, pkg_name, "index.html"), mode="w", encoding="utf-8") as f:
@@ -535,7 +537,7 @@ class S3Index:
             rc.fetch_pep658()
         return rc
 
-def update_pep503_index(prefix: str, package_name: Optional[str] = None, compute_sha256: bool = False, upload: bool = True):
+def update_pep503_index(prefix: str, package_name: Optional[str] = None, update_root_index: bool = True, compute_sha256: bool = False, upload: bool = True):
     """
     Regenerates the PEP 503 simple index for a given S3 prefix.
     Fetches valid artifacts, applies allow-list filtering, optionally updates
@@ -550,6 +552,11 @@ def update_pep503_index(prefix: str, package_name: Optional[str] = None, compute
             this package only (e.g. "torch"). The S3 listing is narrowed to
             "{prefix}/{package_name}-*" for efficiency. When None, all packages
             under the prefix are processed (full sweep).
+        update_root_index: If True (default), regenerate the root index page
+            at "{prefix}/index.html". Must be False when package_name is set:
+            the root index must be built from a full listing, not a
+            package-scoped one — passing both would clobber the root index
+            with only that package's entries.
 
     IMPORTANT:
         `initialize_bucket()` must be called prior to invoking this function.
@@ -564,6 +571,13 @@ def update_pep503_index(prefix: str, package_name: Optional[str] = None, compute
     Returns:
         int: The number of S3 objects included in the index after filtering.
     """
+    if package_name and update_root_index:
+        raise ValueError(
+            "package_name and update_root_index=True cannot be used together: "
+            "the root index must be built from a full prefix listing, not a "
+            "package-scoped one. Use a separate full-sweep call to update the root index."
+        )
+
     print(f"Processing prefix: {prefix}" + (f", package: {package_name}" if package_name else ""))
 
     # Record start time to measure S3 fetch duration
@@ -580,9 +594,9 @@ def update_pep503_index(prefix: str, package_name: Optional[str] = None, compute
     if compute_sha256:
         idx.compute_sha256()
     elif not upload:
-        idx.save_pep503_htmls()
+        idx.save_pep503_htmls(update_root_index=update_root_index)
     else:
-        idx.upload_pep503_htmls()
+        idx.upload_pep503_htmls(update_root_index=update_root_index)
 
     return len(idx.objects)
 
--- a/build_tools/third_party/s3_management/manage.py
+++ b/build_tools/third_party/s3_management/manage.py
@@ -65,6 +65,21 @@ def initialize_bucket(bucket_name: str | None) -> None:
     INDEX_BUCKETS = {BUCKET}
 
 ACCEPTED_FILE_EXTENSIONS = ("whl", "zip", "tar.gz")
+
+
+def _make_list_prefix(prefix: str, package_name: Optional[str]) -> str:
+    """Return the S3 listing prefix to use for a given package scope.
+
+    When package_name is provided, narrows the prefix to only that package's
+    objects (e.g. "v4/whl/torch-"). The hyphen delimiter in wheel filenames
+    makes this unambiguous: "torch-" cannot match "torchaudio-" wheels.
+    When package_name is None, returns the bare prefix for a full sweep.
+    """
+    if package_name:
+        return f"{prefix}/{package_name}-"
+    return prefix
+
+
 PREFIXES = [
     # Note: v2-staging first, in case issues are observed while the script runs
     # and the developer wants to more safely cancel the script.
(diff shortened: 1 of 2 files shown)


build_tools/third_party/s3_management/manage.py
```diff
@@ -231,7 +246,7 @@ def __init__(self: S3IndexType, objects: List[S3Object], prefix: str) -> None:
         # should dynamically grab subdirectories like whl/test/cu101
         # so we don't need to add them manually anymore
         self.subdirs = {
-            path.dirname(obj.key) for obj in objects if path.dirname != prefix
+            path.dirname(obj.key) for obj in objects if path.dirname(obj.key) != prefix
```
araravik-psd: My agent brought this up when I was refactoring last time but changes were breaking lambda and indexing, wanted to add it here.

build_tools/third_party/s3_management/manage.py
```diff
@@ -65,6 +65,21 @@ def initialize_bucket(bucket_name: str | None) -> None:
...
+
+def _make_list_prefix(prefix: str, package_name: Optional[str]) -> str:
+    """Return the S3 listing prefix to use for a given package scope.
+
+    When package_name is provided, narrows the prefix to only that package's
+    objects (e.g. "v4/whl/torch-"). The hyphen delimiter in wheel filenames
+    makes this unambiguous: "torch-" cannot match "torchaudio-" wheels.
+    When package_name is None, returns the bare prefix for a full sweep.
+    """
+    if package_name:
```
araravik-psd, line 78: An empty string "" would fall through to the full-sweep path. This is probably fine in practice since the Lambda would never send "" in our case, but since the type hint is Optional[str], being explicit with if package_name is not None would be more correct and clearer?

build_tools/third_party/s3_management/tests/manage_test.py
```diff
@@ -0,0 +1,78 @@
...
+        ("v4/whl", "torchaudio", "v4/whl/torchaudio-"),
+        ("v4/whl", None, "v4/whl"),
+        ("v2/gfx94X-dcgpu", None, "v2/gfx94X-dcgpu"),
+    ],
+)
+def test_make_list_prefix(prefix: str, package_name: str | None, expected: str) -> None:
+    assert _make_list_prefix(prefix, package_name) == expected
+
+
+def test_make_list_prefix_torch_does_not_match_torchaudio() -> None:
```
araravik-psd, line 41: I dont see a test for fetch_object_names or from_S3 with a mocked S3 client verifying that the correct Prefix is passed to the paginator. I believe this would add value as the core behavior change that prevents the 10 minute full sweep

ScottTodd:
Want to tag https://github.com/ROCm/TheRock/issues/5455 in the PR description? This should help mitigate that, right?

build_tools/third_party/s3_management/manage.py
marbre: This is already addressed by the second commit part of this PR.

build_tools/third_party/s3_management/tests/manage_test.py
marbre: I decided to follow the convention in `update_dependencies_test.py`, which only tests pure functions and leaves S3-touching code untested rather than mocking boto3. The core behavior change, computing the narrowed listing prefix, is covered by `test_make_list_prefix`, which tests the pure `_make_list_prefix` helper that `fetch_object_names` delegates to. The boto3 paginator call itself is trivial plumbing with no logic to verify beyond what the unit tests already cover.

build_tools/third_party/s3_management/manage.py
marbre: `is not None` alone is actually incomplete here. An empty string would pass the `is not None check` and produce `v4/whl/-` as the S3 listing prefix, silently narrowing to nothing. The intended check is indeed `if package_name`, which correctly handles both `None` and `""` as "full sweep", which is the intended behaviour. Updated the docstring to make this explicit, and added an empty string test case to pin it.

build_tools/third_party/s3_management/manage.py
```diff
@@ -384,19 +397,20 @@ def to_simple_packages_html(
...
-                print(f"INFO Uploading {subdir}/index.html to {bucket.name}")
-                bucket.Object(
-                    key=f"{subdir}/index.html"
-                ).put(
-                    # ACL='public-read',
-                    CacheControl='no-cache,no-store,must-revalidate',
-                    ContentType='text/html',
-                    Body=index_html
-                )
+            if update_root_index:
```
araravik-psd: When `package_name` is provided, `fetch_object_names()` scopes the S3 listing to `{prefix}/{package_name}-`, so `self.objects` contains only artifacts for that package.

This works for an existing package because `update_root_index=False` and only the package-specific index is regenerated. However, for the first artifact of a new package, the Lambda sets `update_root_index=True`. In that case, `to_simple_packages_html()` builds the root index from the package-scoped `self.objects`, which could overwrite `{prefix}/index.html` with only the new package.

Could we use a separate full-prefix listing for the root-index update only when `update_root_index=True`, while keeping the package-specific index generation scoped to `package_name`?

araravik-psd (changes requested):
See this as an issue when a new package which did not exist in the bucket is added for indexing. Looks like this could mess up the root index, can you please take a look?

build_tools/third_party/s3_management/manage.py
marbre: Great catch @araravik-psd! I refactored this and got rid of `update_root_index`. To handle this case, the Lambda now extracts the package name from the S3 key and checks whether a per-package index page already exists. If it does, only that package's index is regenerated using the narrowed prefix; if not (brand-new package), a full sweep is performed so the root index correctly lists it (by not passing a package name).

PTAL.

araravik-psd: Thanks for addressing the new-package case. The full sweep for a brand-new package looks correct.

However, for an existing package, `package_name` narrows `self.objects` to that package, but `upload_pep503_htmls()` still unconditionally regenerates and uploads `{subdir}/index.html`. Wouldn’t this overwrite the root index with only the focused package?

I think the package-scoped path still needs to skip the root-index write, while the full-sweep path should update both the root and package indexes.

marbre: You're right and I missed this. I now re-introduced `update_root_index`. PTAL.

araravik-psd: Thanks for the update. I verified the flow with the corresponding Lambda changes:

* Existing packages use a package-scoped listing with `update_root_index=False`.
* Brand-new packages use a full-prefix sweep so the root index is rebuilt from all packages.
* The added validation preventing `package_name` from being combined with `update_root_index=True` also protects against accidentally updating the root index.

araravik-psd (approved):

build_tools/third_party/s3_management/manage.py
ScottTodd, line 1: I had my review agent generate some diagrams to explain this change. Is this all accurate?

### Before

```mermaid
  flowchart TD
    A[CI uploads 59 wheel files] --> B[Each S3 PUT triggers Lambda]
    B --> C["Lambda calls update_pep503_index(prefix='v4/whl')"]
    C --> D[List all objects under v4/whl]
    D --> E[Fetch metadata for all matching wheels]
    E --> F[Regenerate v4/whl/index.html]
    F --> G[Regenerate every package index page]
```

### After

```mermaid
  flowchart TD
    A[CI uploads wheel] --> B[S3 PUT triggers Lambda]
    B --> C[Lambda extracts package name from wheel key]
    C --> D{Package index already exists?}
    D -- yes --> E1[Call update_pep503_index prefix='v4/whl', package_name='amd_torch_device_gfx942',
    update_root_index=false]
    E1 --> F1[List only v4/whl/amd_torch_device_gfx942-*]
    F1 --> G1[Fetch metadata only for that package's wheels]
    G1 --> H1[Skip v4/whl/index.html]
    H1 --> I1[Regenerate only v4/whl/amd-torch-device-gfx942/index.html]

    D -- no --> E2[Call update_pep503_index prefix='v4/whl' with no package_name]
    E2 --> F2[Full sweep]
    F2 --> G2[Regenerate root v4/whl/index.html]
    G2 --> H2[Regenerate package indexes]
```

build_tools/third_party/s3_management/manage.py
```diff
@@ -434,12 +449,13 @@ def compute_sha256(self) -> None:
...
-    def fetch_object_names(cls: Type[S3IndexType], prefix: str) -> List[str]:
+    def fetch_object_names(cls: Type[S3IndexType], prefix: str, package_name: Optional[str] = None) -> List[str]:
         obj_names = []
         paginator = CLIENT.get_paginator('list_objects_v2')
-        page_iterator = paginator.paginate(Bucket=BUCKET_NAME, Prefix=prefix)
+        list_prefix = _make_list_prefix(prefix, package_name)
+        page_iterator = paginator.paginate(Bucket=BUCKET_NAME, Prefix=list_prefix)
         for page in page_iterator:
-            for obj in page['Contents']:
+            for obj in page.get('Contents', []):
```
ScottTodd, line 458: If possible, how do we want to handle the case where there are no objects found when searching for `package_name` ? Just log? Raise an exception?

ScottTodd (approved):

build_tools/third_party/s3_management/manage.py
marbre: It is :)

build_tools/third_party/s3_management/manage.py
marbre: As of today you can't pass it from cli and the lambda wrapper should guard this. So I think for now this is safe to keep as is.

Issue #36, opened 2025-01-30
Windows platform support bringup


ScottTodd:
We're aiming to support as much of the toolkit as possible on Windows.

* Our build target is "native" Windows (not WSL 1 or WSL 2), compiling using standard build tools (e.g. MSVC).
* We should support local builds, have regular CI builds (GitHub Actions), and eventually be able to distribute artifacts.
* We should also have any appropriate documentation, developer guides, etc. needed.

### Setup notes

* I'm on Windows 11 version 23H2, OS build 22631.4751
* I enabled Windows developer mode and granted my user account the ability to create symlinks
* I created a Dev Drive by following https://learn.microsoft.com/en-us/windows/dev-drive/ (500GB, shared with other projects)
* I installed standard software development tools and configured developer options:
  * git from https://git-scm.com/downloads (latest version `2.45.1`) then enabled symlinks with
        
    ```
    git config --global core.symlinks true
    ```
  * Build Tools for Visual Studio 2022 from https://visualstudio.microsoft.com/downloads/ (latest [version `17.12.4`](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes#17.12.4))
  * CMake from https://cmake.org/download/ (latest, [version `3.31.5`](https://cmake.org/cmake/help/latest/release/3.31.html))
  * Ninja as a CMake generator from https://ninja-build.org/ (latest, [version `1.12.1`](https://github.com/ninja-build/ninja/releases/tag/v1.12.1))
  * ccache from https://ccache.dev/ (latest, [version `4.10.2`](https://ccache.dev/releasenotes.html#_ccache_4_10_2))

### Fetching sources

After cloning TheRock, I tried to checkout sources using `python ./build_tools/fetch_sources.py`. That needed the repo tool (https://gerrit.googlesource.com/git-repo/+/HEAD/README.md, docs at https://source.android.com/docs/setup/reference/repo), which is not distributed on Windows, so I downloaded the script manually. I did then have to edit [`build_tools/fetch_sources.py`](https://github.com/nod-ai/TheRock/blob/main/build_tools/fetch_sources.py) to exec `python D:/path/to/repo` instead of just `repo`. We can teach that script how to download the file on its own and run it in a portable way.

### Configure and build

After fetching sources, I tried configuring and building with CMake (under VSCode):
[...]

PR #5941, opened 2026-06-17
fix: Fix RUNPATH for libdyninstAPI_RT.so

ajanicijamd:
## Motivation

`libdyninstAPI_RT.so` in `dist/rocm/lib/rocprofiler-systems/` was not having its RUNPATH
patched by the post hook introduced in #5287, leaving it with an incorrect `$ORIGIN`-relative
RUNPATH compared to the other Dyninst libraries in that directory.

## Technical Details

Unlike the other Dyninst libraries, `libdyninstAPI_RT.so` is **not** built by Dyninst directly.
rocprofiler-systems sets `DYNINST_BUILD_RTLIB=OFF` and instead builds its own runtime library
(`rocprofiler-systems-rt-library`), which is installed as `librocprof-sys-rt.so` in `lib/`.
A compatibility symlink `lib/rocprofiler-systems/libdyninstAPI_RT.so → ../librocprof-sys-rt.so.X.Y.Z`
is then installed.

This means:
- The `dyninstAPI_RT` CMake target never exists, so the `if(TARGET "dyninstAPI_RT")` check in
  the post hook silently skips it.
- The real file lives in `lib/`, so the `THEROCK_INSTALL_RPATH_ORIGIN` mechanism cannot be used:
  setting the origin to `lib/rocprofiler-systems` on a file whose `$ORIGIN` resolves to `lib/`
  would generate wrong `$ORIGIN/..` paths.

The fix removes `dyninstAPI_RT` from `_rocprofsys_lib_targets` and adds an `install(CODE)` block
that:
1. Detects the installed symlink at `lib/rocprofiler-systems/libdyninstAPI_RT.so`.
2. Replaces it with a real copy of the underlying library.
3. Uses `patchelf` to set the RUNPATH on the copy to
   `$ORIGIN/..:$ORIGIN:$ORIGIN/../llvm/lib:$ORIGIN/../rocm_sysdeps/lib`, matching the pattern
   used by all other libraries in `lib/rocprofiler-systems/`.

The `patchelf` binary is located via the `PATCHELF` environment variable (consistent with other
TheRock sysdeps scripts) or via `find_program`.

## JIRA ID

ROCM-25537

## Test Plan

- Build TheRock.
- In the directory `/therock/output/build/dist/rocm/lib/rocprofiler-systems` in the manylinux
  container, verify the RUNPATH of `libdyninstAPI_RT.so`:

    patchelf --print-rpath libdyninstAPI_RT.so

## Test Result

- The output of `patchelf` above is:

    $ORIGIN/..:$ORIGIN:$ORIGIN/../llvm/lib:$ORIGIN/../rocm_sysdeps/lib

- `libdyninstAPI_RT.so` is now a regular file (not a symlink) with the correct RUNPATH,
consistent with the other Dyninst libraries in `lib/rocprofiler-systems/`.

## Submission Checklist


Xch666137:
Adding a root-cause analysis and a verified workaround for Windows, since we hit the exact same failure (`MIOpenNeuron.cpp` / hiprtc: `'type_traits' file not found`, HIPRTC_ERROR_COMPILATION) on gfx1201 (RX 9070 XT, Windows 11, no Visual Studio installed) with TheRock pip wheels (torch 2.13.0+rocm10.0.0, rocm-sdk-core 10.0.0).

**Root cause chain (all verified locally):**

1. MIOpen JIT-compiles RNN/reduction kernels at runtime via hiprtc, and the kernel sources `#include <type_traits>` etc. — they need a C++ standard library.   On Windows, the built-in bsdtar (`C:\Windows\System32\tar.exe`) extracts these subtrees directly, e.g.: tar -xf therock-dist-linux-gfx120X-all-10.0.0.tar.gz --wildcards "lib/llvm/include/c++/*" "lib/llvm/include/amdgcn-amd-amdhsa/*" (Git Bash's GNU tar may fail on the same wildcards.)
2. TheRock Windows artifacts ship **no C++ stdlib headers at all**:
   - the full file list of the `rocm-sdk-core`/`rocm-sdk-devel` 10.0.0 pip wheels (7,344 entries) and of `therock-dist-windows-gfx120X-all-10.0.0.tar.gz` (10,544 entries) contain no `include/c++/v1`;
   - `compiler/pre_hook_amd-llvm.cmake` sets `LLVM_ENABLE_LIBCXX OFF` with `TODO(#36): Enable libunwind, libcxx, and libcxxabi on Windows?` — so Windows amd-llvm never builds libc++ (the Linux tarball does ship `lib/llvm/include/c++/v1`, 1,778 files). This matches #3588, which only helped Linux.
3. On Linux, JIT compilation works because clang probes a GCC/libstdc++ installation; the official CI additionally injects `CPLUS_INCLUDE_PATH=${ROCM_HOME}/lib/rocm_sysdeps/include` (see `.github/workflows/test_pytorch_wheels_full.yml`). Windows has neither, so every fused-RNN/MIOpen-JIT kernel fails at the first stdlib include.

**Verified workaround** (fresh-JIT compile of nn.GRU/LSTM/RNN/Conv2d/BatchNorm2d all pass after clearing the MIOpen cache):

1. Extract two subtrees from the **Linux** tarball (`therock-dist-linux-gfx120X-all-10.0.0.tar.gz`): `lib/llvm/include/c++/*` (libc++ headers) and `lib/llvm/include/amdgcn-amd-amdhsa/*` (llvm-libc amdgcn C headers incl. the generated `__config_site`).
2. Add two **empty shim headers** named `cmath` and `limits` in a directory that takes precedence — device code gets math builtins without them, and MIOpen's own `miopen_limits.hpp` (JIT mode declares `std::numeric_limits` specializations for plain `namespace std`) otherwise conflicts with libc++'s `std::__1` (`reference to 'numeric_limits' is ambiguous`).
3. Export `CPLUS_INCLUDE_PATH` with, in order: shim dir → libc++ `c++/v1` → amdgcn `c++/v1` → amdgcn C-header dir → the wheel's clang resource dir (`.../site-packages/_rocm_sdk_core/lib/llvm/lib/clang/23/include`).

Happy to test a Windows wheel that ships these headers, if that's useful.

PR #5950, opened 2026-06-18
[ml-libs] ALMIOPEN-1972 Bundle golden_reference_data in hipdnn-integration-tests test artifact

adickin-amd:
## Summary

The integration-tests CMakeLists installs `golden_reference_data/` to `lib/golden_reference_data/` in the stage, and the test binary resolves it at `<exe_dir>/../lib/golden_reference_data` (hardcoded in `FileUtilities.hpp`). The artifact descriptor was not capturing that path, so the tensor data was silently dropped from the packaged artifact. CI pulls test artifacts to machines without DVC access, so the data must be present in the artifact at build time.

## Risk Assessment

Low risk. Single-line addition to an artifact descriptor glob pattern with no effect on build logic, test code, or GPU dispatch.

## ASIC Coverage

ASIC-independent. This is a packaging change to an artifact descriptor only — no kernel selection, dispatch logic, or test behavior is altered. Passing PR CI is sufficient.

## Testing Summary

- No local testing applicable; this change only affects artifact packaging output.

## Testing Checklist

## Technical Changes

- Adds `"lib/golden_reference_data/**"` to the `test` component includes in `ml-libs/artifact-hipdnn-integration-tests.toml` so the golden reference tensor data installed by the subproject is captured in the test artifact.

### DIFF
--- a/ml-libs/artifact-hipdnn-integration-tests.toml
+++ b/ml-libs/artifact-hipdnn-integration-tests.toml
@@ -3,6 +3,7 @@
 include = [
   "bin/*hipdnn_*_test*",
   "bin/hipdnn_integration_tests_ctest/CTestTestfile.cmake",
+  "lib/integration_test_bundles/**",
 ]
 [components.run."ml-libs/hipdnn_integration_tests/stage"]
 include = [


adickin-amd:
I let the build run overnight and confirmed that the artifacts were archived properly at the old golden_ref folder name.  I have recently renamed that folder to indicate its general integration bundles rather than just golden data in this PR https://github.com/ROCm/rocm-libraries/pull/8584.  That PR isnt in TheRock yet but we can still update the toml file now and merge it in because its a noop if the folder doesnt exist.  

The reason Im doing it early is because bumps can take extra time and this ensures I get it in rocm-libraries as soon as possible

mousdahl-amd (approved):

adickin-amd:
merging, this is just an archive change and wont have affected any of the tests that fail.

Issue #5937, opened 2026-06-17
**Pipeline integration** — Create a ASAN release build into the ROCm release pipeline similar to multi_arch_release.yml with weekly schedule


nunnikri:
## Parent
ROCm ASAN Packaging for TheRock #5914 

## Goal
Create a new CI workflow file for ASAN builds, modelled on [`multi_arch_release.yml`](https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release.yml), running on a weekly schedule.

## Tasks
- [ ] Create a new workflow file (e.g., `asan_release.yml`) in the release pipeline repo
- [ ] Enable basic testing
- [ ] Set trigger to weekly schedule (e.g., `cron: '0 2 * * 1'`) plus manual `workflow_dispatch`
- [ ] Reuse existing multi-arch-ci-asan to the build step
- [ ] Record build time and artifact size for audit and ci cost analysis

## Acceptance Criteria
- Workflow triggers automatically on the weekly schedule and via manual dispatch with basic testing enabled
- ASAN artifacts are uploaded to the correct s3 area
- Non-ASAN (standard) release pipeline is not affected

PR #5968, opened 2026-06-18
feat(ci): Adding `multi_arch_release_asan.yml`

geomin12:
Adding a replicate of `multi_arch_release.yml` for ASAN! For now, we will land this and test with a dev bucket (via workflow dispatch)

Progress on #5937
ISSUE ID #5937

Ran a quick CI check with push: https://github.com/ROCm/TheRock/actions/runs/27791869153/job/82242548547

### DIFF
--- a/.github/workflows/multi_arch_build_tarballs.yml
+++ b/.github/workflows/multi_arch_build_tarballs.yml
@@ -34,13 +34,6 @@ on:
           - ci
           - dev
         default: ci
-      build_variant:
-        description: "Build variant (release, asan). Affects tarball naming."
-        type: choice
-        options:
-          - release
-          - asan
-        default: release
   workflow_call:
     inputs:
       artifact_run_id:
@@ -61,10 +54,6 @@ on:
       release_type:
         type: string
         default: ci
-      build_variant:
-        description: "Build variant (release, asan). Affects tarball naming."
-        type: string
-        default: "release"
       repository:
         description: "Repository to checkout. Defaults to github.repository."
         type: string
@@ -118,7 +107,6 @@ jobs:
             --dist-amdgpu-families="${{ inputs.dist_amdgpu_families }}" \
             --platform="${{ inputs.platform }}" \
             --package-version="${{ inputs.package_version }}" \
-            --build-variant="${{ inputs.build_variant }}" \
             --output-dir="${{ github.workspace }}/tarballs"
 
       - name: Configure AWS Credentials
--- a/.github/workflows/multi_arch_build_tarballs.yml
+++ b/.github/workflows/multi_arch_build_tarballs.yml
@@ -34,6 +34,13 @@ on:
           - ci
           - dev
         default: ci
+      build_variant:
+        description: "Build variant (release, asan). Affects tarball naming."
+        type: choice
+        options:
+          - release
+          - asan
+        default: release
   workflow_call:
     inputs:
       artifact_run_id:
@@ -54,6 +61,10 @@ on:
       release_type:
         type: string
         default: ci
+      build_variant:
+        description: "Build variant (release, asan). Affects tarball naming."
+        type: string
+        default: "release"
       repository:
         description: "Repository to checkout. Defaults to github.repository."
         type: string
@@ -107,6 +118,7 @@ jobs:
             --dist-amdgpu-families="${{ inputs.dist_amdgpu_families }}" \
             --platform="${{ inputs.platform }}" \
             --package-version="${{ inputs.package_version }}" \
+            --build-variant="${{ inputs.build_variant }}" \
             --output-dir="${{ github.workspace }}/tarballs"
 
       - name: Configure AWS Credentials
--- a/.github/workflows/multi_arch_release_asan.yml
+++ b/.github/workflows/multi_arch_release_asan.yml
@@ -9,10 +9,6 @@
 name: Multi-Arch Release ASAN
 
 on:
-  # TODO: Remove push trigger before merging
-  push:
-    branches:
-      - users/geomin12/asan-packaging
   schedule:
     # Runs every Saturday at 12:00 PM UTC
     - cron: "0 12 * * 6"
--- a/.github/workflows/multi_arch_release_asan.yml
+++ b/.github/workflows/multi_arch_release_asan.yml
@@ -1,0 +1,190 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# This workflow is the entry point for weekly ASAN releases.
+#
+# It builds ASAN-instrumented ROCm packages on a weekly schedule and
+# publishes them to release buckets. This is modeled on multi_arch_release.yml
+# but uses the ASAN build variant.
+#
+# For manual dev releases, trigger via workflow_dispatch on TheRock.
+
+name: Multi-Arch Release ASAN
+
+on:
+  schedule:
+    # Runs every Saturday at 02:00 AM UTC
+    - cron: "0 2 * * 6"
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Release type; developer-triggered jobs should use "dev".'
+        type: choice
+        options:
+          - dev
+          - nightly
+        default: dev
+      linux_amdgpu_families:
+        description: 'Comma-separated list of Linux GPU families. "all" = all, "none" = skip platform.'
+        type: string
+        default: "gfx94x"
+      linux_test_labels:
+        type: string
+        description: "If enabled, reduce test set on Linux to the list of labels prefixed with 'test:'. ex: test:rocprim, test:hipcub"
+        default: ""
+      repository:
+        description: "Repository to checkout. Defaults to github.repository."
+        type: string
+        default: ""
+      ref:
+        description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
+        type: string
+        default: ""
+
+permissions:
+  contents: read
+
+run-name: >-
+  Multi-Arch Release ASAN (${{ inputs.release_type || 'nightly' }})
+  | Linux: ${{ inputs.linux_amdgpu_families || 'gfx94x,gfx110x,gfx120x' }}
+
+jobs:
+  setup:
+    uses: ./.github/workflows/setup_multi_arch.yml
+    with:
+      build_variant: "asan"
+      release_type: ${{ inputs.release_type || 'nightly' }}
+      linux_amdgpu_families: ${{ inputs.linux_amdgpu_families || 'gfx94x,gfx110x,gfx120x' }}
+      linux_test_labels: ${{ inputs.linux_test_labels || '' }}
+      repository: ${{ inputs.repository || github.repository }}
+      ref: ${{ inputs.ref }}
+
+  linux_build_and_test:
+    name: Linux::${{ fromJSON(needs.setup.outputs.linux_build_config || '{}').build_variant_label || 'skip' }}
+    needs: setup
+    if: ${{ needs.setup.outputs.linux_build_config != '' }}
+    uses: ./.github/workflows/multi_arch_ci_linux.yml
+    secrets: inherit
+    with:
+      build_config: ${{ needs.setup.outputs.linux_build_config }}
+      test_labels: ${{ needs.setup.outputs.linux_test_labels }}
+      rocm_package_version: ${{ needs.setup.outputs.rocm_package_version }}
+      rocm_deb_package_version: ${{ needs.setup.outputs.rocm_deb_package_version }}
+      rocm_rpm_package_version: ${{ needs.setup.outputs.rocm_rpm_package_version }}
+      test_type: ${{ needs.setup.outputs.test_type }}
+    permissions:
+      contents: read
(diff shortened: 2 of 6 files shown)


.github/workflows/multi_arch_release_asan.yml
```diff
@@ -0,0 +1,151 @@
...
+  Multi-Arch Release ASAN (${{ inputs.release_type || 'nightly' }})
+  | Linux: ${{ inputs.linux_amdgpu_families || 'all' }}
+  | Windows: ${{ inputs.windows_amdgpu_families || 'none' }}
+
+jobs:
+  setup:
+    uses: ./.github/workflows/setup_multi_arch.yml
+    with:
+      build_variant: "asan"
+      release_type: ${{ inputs.release_type || 'nightly' }}
```
ScottTodd: Nightly releases **must** go in https://github.com/ROCm/rockrel now.

.github/workflows/multi_arch_release_asan.yml
```diff
@@ -0,0 +1,151 @@
...
+        default: ""
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Release type; developer-triggered jobs should use "dev".'
+        type: choice
+        options:
+          - dev
+          - nightly
+        default: dev
```
ScottTodd, line 53: The only release type allowed in this repository is now `dev`. This will soon be enforced via IAM roles.

Wrapper workflows over in rockrel like https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release.yml can allow other types.

.github/workflows/multi_arch_release_asan.yml
```diff
@@ -0,0 +1,151 @@
...
+
+  linux_release:
+    name: Linux::${{ fromJSON(needs.setup.outputs.linux_build_config || '{}').build_variant_label || 'skip' }}
+    needs: setup
+    if: ${{ needs.setup.outputs.linux_build_config != '' }}
+    uses: ./.github/workflows/multi_arch_release_linux.yml
+    secrets: inherit
+    with:
+      build_config: ${{ needs.setup.outputs.linux_build_config }}
+      rocm_package_version: ${{ needs.setup.outputs.rocm_package_version }}
```
ScottTodd, line 100: What files is this job going to upload? I don't trust our scripts to currently separate where non-asan and asan packages are published to different subfolders. We can't mix them in the same index (at least without new naming).

Need to see what happens for tarballs, python packages, native linux packages, etc.

ScottTodd (changes requested):
Need to test this with dev releases before we go straight to publishing weekly. This is very high risk.

.github/workflows/multi_arch_release_asan.yml
nunnikri: Do we need changes in build_tools/_therock_utils/storage_location.py and or else will it conflict with currently nightly build artifacts?
cc: @ScottTodd

.github/workflows/multi_arch_release_asan.yml
ScottTodd: https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/publish_rocm_to_release_buckets.py is where we upload tarballs to `therock-{release_type}-tarball/v4/tarball/`, python packages to `therock-{release_type}-python/v4/whl/`, etc. does what do the packaging RFCs say about where asan packages should be uploaded to?

.github/workflows/multi_arch_release_asan.yml
ScottTodd: I'm less concerned with https://github.com/ROCm/TheRock/blob/main/build_tools/_therock_utils/workflow_outputs.py and https://github.com/ROCm/TheRock/blob/main/build_tools/_therock_utils/storage_location.py. Those are scoped per-workflow run already, so unless we try to build asan packages and non-asan packages in a single workflow run (which this is not doing), files will still be scoped like https://therock-ci-artifacts.s3.amazonaws.com/27709933411-linux/index.html from https://github.com/ROCm/TheRock/actions/runs/27709933411

The bigger concern is where in the release buckets we copy files from the artifact bucket. That's using https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/publish_rocm_to_release_buckets.py

(and https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/publish_pytorch_to_release_bucket.py, https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/publish_jax_to_release_bucket.py)

nunnikri: Yes you are right. The final release bucket is of main concern. 
As per RFC we dont need wheel for asan. Need only tar and native packaging. 
And tar ball is in same location therock-{release_type}-tarball/v4/tarball/ but with a different name with a -asan suffix. 
Native packages currently its uploaded to therock-{release_type}-packages/v4/packages/deb and therock-{release_type}-packages/v4/packages/rpm. 
For asan it needs to be in a different location therock-{release_type}-packages/v4/packages-asan/deb and therock-{release_type}-packages/v4/packages-asan/rpm.

geomin12: ack, i'll land a PR there once this lands

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -43,6 +43,18 @@
...
+
+    For ASAN builds (--build-variant asan), python packages are skipped and
+    native packages are published to separate paths:
+
+    s3://therock-dev-artifacts/12345-linux/packages/deb/
+      -> s3://therock-dev-packages/v4/packages-asan/deb/20250101-12345/
+    s3://therock-dev-artifacts/12345-linux/packages/rpm/
+      -> s3://therock-dev-packages/v4/packages-asan/rpm/20250101-12345/
+
+    Tarballs already include -asan suffix in filenames from the build system.
```
nunnikri: Publish changes looks ok to me. But for getting -asan suffix might need to update build_tools/build_tarballs.py to take --build-variant as an argument.

.github/workflows/multi_arch_release_asan.yml
```diff
@@ -0,0 +1,147 @@
...
+        default: ""
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Release type; developer-triggered jobs should use "dev".'
+        type: choice
+        options:
+          - dev
+        default: dev
+      prerelease_version:
```
nunnikri: prerelease part can be removed as we restrict to 'dev' in dispatch.

nunnikri (approved):
LGTM. Now need to do the rockrel part

.github/workflows/multi_arch_build_tarballs.yml
```diff
@@ -34,6 +34,13 @@ on:
           - ci
           - dev
         default: ci
+      build_variant:
+        description: "Build variant (release, asan). Affects tarball naming."
+        type: choice
+        options:
+          - release
+          - asan
+        default: release
```
ScottTodd: nit: I think having a "build variant" named "release" is confusing when it appears right next to "release_type", but I can't think of a better name for "no sanitizers enabled" at the moment ("default"? "base"?).

.github/workflows/multi_arch_release_asan.yml
```diff
@@ -0,0 +1,143 @@
...
+      test_type: ${{ needs.setup.outputs.test_type }}
+      release_type: ${{ inputs.release_type || 'dev' }}
+      repository: ${{ inputs.repository || github.repository }}
+      ref: ${{ needs.setup.outputs.ref }}
+      build_pytorch: ${{ inputs.build_pytorch }}
+      python_version: ${{ inputs.python_version }}
+    permissions:
+      contents: read
+      actions: write
+      id-token: write
```
ScottTodd: Let's remove `windows_release` from this workflow if we aren't planning on producing Windows ASan packages any time in the near future, or at least pass through `build_variant: "asan"` there too. Could also remove the Windows workflow inputs?

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -167,12 +186,15 @@ def publish_native_linux_packages(
...
 
     for pkg_type in ["deb", "rpm"]:
         source = artifacts_root.native_linux_packages(pkg_type)
 
         if release_type == "prerelease":
             dest_prefix = f"v4/packages/{pkg_type}"
+        elif is_asan:
+            dest_prefix = f"v4/packages-asan/{pkg_type}/{today}-{artifacts_root.run_id}"
         else:
             dest_prefix = f"v4/{pkg_type}/{today}-{artifacts_root.run_id}"
```
ScottTodd: Careful here. Looks like this is missing a branch for `is_asan and release_type == "prerelease"`. Otherwise, a "prerelease" release type with "asan" build variant would overwrite non-asan packages in `v4/packages/`.

How about splitting into two steps:
```python

if is_asan:
  dest_part_1 = "v4/packages-asan"
else:
  dest_part_1 = "v4/packages"

if release_type == "prerelease":
  dest_part_2 = f"{pkg_type}"
else:
  dest_part_2 = f"{pkg_type}/{today}-{artifacts_root.run_id}"

dest_prefix = f"{dest_part_1}/{dest_part_2}"
```

build_tools/build_tarballs.py
```diff
@@ -193,9 +204,7 @@ def main(argv=None):
             run_github_repo=args.run_github_repo,
         )
         family_dirs.append(flatten_dir)
-        tarball_name = (
-            f"therock-dist-{args.platform}-{family}-{args.package_version}.tar.gz"
-        )
+        tarball_name = f"therock-dist-{args.platform}-{family}-{args.package_version}{variant_suffix}.tar.gz"
```
ScottTodd: @LiamfBerry / @saadrahim do you have an opinion from the RFC on https://github.com/ROCm/TheRock/pull/4438 about where in the file name this `-asan` variant should go in the file names?

We'll have
```
rocm-sdk-gfx<target>-X.Y.Z-linux.tar.gz            # x86_64 host (implicit)
rocm-sdk-gfx<target>-X.Y.Z-linux_ppc64le.tar.gz    # non-x86 host (explicit)
```

should `-asan` go at the end, or somewhere in the middle?

```
rocm-sdk-asan-gfx<target>-X.Y.Z-linux.tar.gz  # start (after package name, before GPU target)
rocm-sdk-gfx<target>-asan-X.Y.Z-linux.tar.gz  # middle (between GPU target/family and version)
rocm-sdk-gfx<target>-X.Y.Z-linux-asan.tar.gz  # end, "suffix" like on this PR
```

build_tools/github_actions/publish_rocm_to_release_buckets.py
nunnikri: Thats a good catch. Thanks @ScottTodd

ScottTodd (approved):
LGTM once comments are addressed. The higher risk parts of this (jumping straight to nightly releases) were resolved already.

build_tools/build_tarballs.py
marbre: Citing https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0012-Repo-Structure.md:

        - **Variant separation.** Non-standard build variants — `asan`
          today, and reserved future `rpath` / `debug` variants — must
          be separable so that a package manager pointed only at the
          standard repo never accidentally resolves an ASAN or debug
          build (e.g. `dnf install rocm` must not pull an ASAN build).
          *How* that separation is realized — sibling folders, a
          distinct index, or a disabled-by-default repo stanza such as
          the `amdrocm-stable-rpath` stanza inside `amdrocm-repo-stable`
          — is left to the implementation.


So for native Linux packages, this should go to a different subfolder and it will not be encoded in the filename. I would have assumed the same for ASAN tarballs.

.github/workflows/multi_arch_build_tarballs.yml
saadrahim: | Variant | Typical meaning |
|---|---|
| **Debug** | No/low optimization, debug symbols, assertions/logging enabled |
| **ASan Debug** | Debug build plus AddressSanitizer for memory bug detection |
| **Release** | Optimized build, usually no sanitizer, fewer debug checks |
| **RelWithDebInfo** | Release optimizations plus debug symbols |
| **MinSizeRel** | Release optimized for binary size |

saadrahim: Release is good

geomin12: we can keep it as `release` for now as other build / download flows call it release. however, if confusing for developers in the future, we can change the naming

build_tools/github_actions/publish_rocm_to_release_buckets.py
geomin12: good idea and good catch! very clean too, with options for more (tsan, lsan, etc)

build_tools/build_tarballs.py
geomin12: i actually prefer different subfolders in this scenario, as it matches the other cases (consistency). i'm going to land this, but if we feel differently, can update / revert

PR #4490, opened 2026-04-13
[ci] Add rocsparse and hipsparse to test_runner

jsandham:
## Motivation

Support for ctest -L label-based test filtering (quick/standard/comprehensive/full), which is required for proper sharding and test categorization. Additionally, several artifact and configuration issues prevented the tests from running correctly end-to-end.

## Technical Details

In `fetch_test_configurations.py` switched from `test_hipsparse.py`/`test_rocsparse.py` to using `test_runner.py` and then added both rocsparse and hipsparse to `COMPONENT_DIR_MAPPING` in test_runner.py. Need to merge corresponding rocsparse and hipsparse PRs: https://github.com/ROCm/rocm-libraries/pull/5886 and https://github.com/ROCm/rocm-libraries/pull/6040 before merging this.

## Test Plan

Verify CI passes

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5983, opened 2026-06-19
ci: mitigate gfx110X Windows sparse test failures via legacy scripts

kliegeois:
## Motivation

The `Test rocsparse`/`Test hipsparse` jobs on `gfx110X-all` (Windows, AMD Radeon PRO V710 MxGPU, ~25 GB partition) are failing with mass `hipErrorOutOfMemory` (e.g. 5,739 failures in [run 27729923416](https://github.com/ROCm/TheRock/actions/runs/27729923416/job/82121441467)). Tracking issue: ROCm/rocm-libraries#8592.

This is **not a kernel regression**. It is a test-scope expansion: before #4490, Windows sparse ran the legacy per-component scripts (`test_rocsparse.py` / `test_hipsparse.py`), which were the last green basis. #4490 switched sparse to `test_runner.py`, and Multi-Arch CI's default `TEST_TYPE=standard` now runs the `pre_checkin` configs that failed with OOM.

The underlying `pre_checkin` OOM is still being investigated (see [#8592](https://github.com/ROCm/rocm-libraries/issues/8592) and PR #5960). This PR restores green CI to unblock bumps in the meantime.

## Technical Details

- Add an optional per-platform `test_script_by_platform` override in `fetch_test_configurations.py` (same shape/spirit as the existing per-platform `total_shards_dict` / `exclude_family`).
- Use it to route **Windows** `rocsparse`/`hipsparse` back to the pre-#4490 legacy scripts. **Linux stays on `test_runner.py`** for full `standard` coverage.
- The legacy scripts already exist and already carry the `gfx110X-all` Windows ignore list (`test_hipsparse.py`), so this is the exact last-green behavior.
- The artifact-include fixes and `COMPONENT_DIR_MAPPING` entries added by #4490 are intentionally **left untouched** (they are inert on the legacy path), so this is a minimal, single-file change rather than a full revert.

This is a **temporary mitigation**: once the `pre_checkin` OOM is fixed in rocm-libraries, revert by deleting the two `test_script_by_platform` blocks (Windows falls back to `test_runner.py`).

## Test Plan

- Multi-Arch CI: confirm `gfx110X-all` Windows `rocsparse`/`hipsparse` go green.
- Confirm Linux sparse lanes still run via `test_runner.py` (`standard`) and are unaffected.

ISSUE ID : https://github.com/ROCm/rocm-libraries/issues/8592

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -328,15 +328,14 @@ test_matrix = {
         "job_name": "hipsparse",
         "fetch_artifact_args": "--blas --tests",
         "timeout_minutes": 30,
-        "test_script": f"python {_get_script_path('test_runner.py')}",
         # Temporary mitigation for ROCm/rocm-libraries#8592: the gfx110X
         # Windows V710 MxGPU partition OOMs on the pre_checkin sparse configs
-        # that the test_runner.py (standard) path now runs. Keep Windows on
-        # the pre-#4490 legacy script (last green) until the OOM is fixed;
-        # Linux stays on test_runner.py for full coverage.
-        "test_script_by_platform": {
-            "windows": f"python {_get_script_path('test_hipsparse.py')}",
-        },
+        # that the test_runner.py (standard) path runs, cascading into mass
+        # hipErrorOutOfMemory failures. Route sparse back to the pre-#4490
+        # legacy script (the last green basis, which already carries the
+        # gfx110X ignore list) until the underlying OOM is fixed. Restore
+        # test_runner.py afterwards.
+        "test_script": f"python {_get_script_path('test_hipsparse.py')}",
         "platform": ["linux", "windows"],
         "total_shards_dict": {
             "linux": 1,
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -340,6 +348,12 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
         "fetch_artifact_args": "--blas --tests",
         "timeout_minutes": 30,
         "test_script": f"python {_get_script_path('test_runner.py')}",
```
geomin12: you can just revert back to rocsparse.py here, and add a comment! i would prefer not to add another entry in the dictionary to compliate even more

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -329,6 +329,14 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+        # Temporary mitigation for ROCm/rocm-libraries#8592: the gfx110X
+        # Windows V710 MxGPU partition OOMs on the pre_checkin sparse configs
+        # that the test_runner.py (standard) path now runs. Keep Windows on
+        # the pre-#4490 legacy script (last green) until the OOM is fixed;
+        # Linux stays on test_runner.py for full coverage.
```
geomin12: you can just revert back to hipsparse.py here, and add a comment! i would prefer not to add another entry in the dictionary to compliate even more

build_tools/github_actions/fetch_test_configurations.py
kliegeois: Done in 3994336 — reverted directly to the legacy script (both platforms) with a comment, and dropped the dict entry / loop logic. Thanks!

build_tools/github_actions/fetch_test_configurations.py
kliegeois: Done in 3994336 — reverted directly to the legacy script (both platforms) with a comment, and dropped the dict entry / loop logic. Thanks!

geomin12 (approved):
lgtm, thanks for comments

PR #5984, opened 2026-06-19
Fix kpack libraries pure package name

### LINKED ISSUE #5048: Multi-arch: Rename internal structure of rocm_sdk_libraries wheel
Right now when unpacking rocm_sdk_libraries multiarch wheel you get the following structure:

```
_rocm_sdk_libraries/
rocm_sdk_libraries-7.13.0a20260501.dist-info/
rocm_sdk_libraries_None/
```

To stay in the convention and remove the special treatement in the promotion script, rename `rocm_sdk_libraries_None/` to `rocm_sdk_libraries/`. `None` most likely appears because in single arch it was called `rocm_sdk_libraries_gfx110X-all /` but now in multi arch rocm_sdk_libraries does not have any arch attached to it.

marbre:
The kpack-split libraries package has `THIS_TARGET_FAMILY` set to `None`. The setup script was interpolating that value into the pure Python package name, emitting `rocm_sdk_libraries_None` in wheels and installs instead of `rocm_sdk_libraries`.

With this patch, the base pure Python package name is used for target-neutral libraries while keeping target-specific names for legacy (rocm-sdk-libraries-gfx*) wheels.

Fixes #5048.

Assisted-by: Codex <codex@openai.com>

### DIFF
--- a/build_tools/packaging/python/templates/rocm-sdk-libraries/setup.py
+++ b/build_tools/packaging/python/templates/rocm-sdk-libraries/setup.py
@@ -31,7 +31,12 @@ def import_dist_info():
 dist_info = import_dist_info()
 my_package = dist_info.ALL_PACKAGES["libraries"]
 print(f"Loaded dist_info package: {my_package}")
-pure_py_package = f"rocm_sdk_libraries_{dist_info.THIS_TARGET_FAMILY}"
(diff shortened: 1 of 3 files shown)


ScottTodd (approved):

marbre:
Test failures are unrelated.

PR #5986, opened 2026-06-19
feat(quartz): notify_quartz reusable workflow + dispatcher

HereThereBeDragons:
# Summary

First step of porting the Quartz CI-reporting scripts from ROCm/Quartz-Tester
into TheRock. This adds the self-report bridge that lets a TheRock workflow
push its run status to Quartz for ingestion.

- **`.github/workflows/notify_quartz.yml`** (new): reusable `workflow_call`
  workflow. Mints a Quartz Hauly GitHub App token and invokes the dispatcher.
  `continue-on-error: true` so reporting never fails the calling workflow.
- **`build_tools/github_actions/notify_quartz.py`** (new): stdlib-only
  dispatcher. Builds a payload for the current run (resolved via
  `GITHUB_RUN_ID` + the Actions API) and dispatches it to a Quartz ingest
  workflow via `workflow_dispatch`, at run start (`run_phase=started`) and
  end (`run_phase=completed`). Derives the run conclusion from a
  `toJSON(needs)` blob when no explicit conclusion is passed, and strips the
  inline `jobs` array (setting `fetch_jobs`) when the payload would exceed
  GitHub's 65,535-char dispatch-inputs cap. Targets `ROCm/Quartz` by default
  (override with `--quartz-repo` / `QUARTZ_REPO`).
- **`build_tools/github_actions/tests/notify_quartz_test.py`** (new): stdlib
  `unittest` coverage for the dispatcher's pure logic: reporting-path
  normalization (`_normalize_reporting_path`) and payload path overrides
  (`_build_payload`). Runs standalone, no network or GitHub API needed.

Copied from Quartz-Tester `main` except the pinned action SHAs, which were
bumped to match TheRock's current pins (`actions/checkout` v6.0.3,
`actions/create-github-app-token` v3.2.0), and the default ingest target,
which was switched from ROCm/Quartz-Tester to ROCm/Quartz.

## References
- https://github.com/ROCm/Quartz/issues/17
- RFC: [RFC0011-Quartz-CICD-Datahub.md](https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md),
  background on what Quartz is and why we report CI/CD data to it.
- [Quartz Implementation Plan](https://github.com/orgs/ROCm/projects/156), the issue tracker. Note: it's not fully up to
  date. 

## Notes

- These files are **standalone**; they only become active once
  `notify_quartz` jobs are actually added to the workflows (a follow-up).
- Requires repo secrets `GH_APP_HAULY_ID` / `GH_APP_HAULY_PRIVATE_KEY`
  (Quartz Hauly GitHub App), already installed in TheRock.

---

Co-authored-by: Ciprian Goea <ciprian.goea@amd.com>


HereThereBeDragons:
@cgoea lets see what feedback we get and who of us has time to apply them

marbre (commented):
Cosmetics:
* The RFC calls the TheRock-side script `send_to_quartz.py`; this PR ships `notify_quartz.py`. Worth aligning the RFC or the filename so the two docs don't drift?
* Default `receive_therock_data.yml` (underscores) vs RFC's `receive-therock-data.yml` (hyphens). It's configurable, but a mismatch with the actual file in the target repo produces a silent 404 on dispatch (masked by continue-on-error). Please confirm the real filename in `ROCm/quartz`.

The most significant seems to be the dispatch granularity. The RFC says to call `send_to_quartz.py` on each build/test job at start and finish (`if: always()`), with `version = run_attempt*2 + signal_type`. This PR instead dispatches per workflow-run: one started (no jobs) and one completed signal carrying the whole jobs array fetched from the API. Different model than the documented schema/ingest path, why? Is the per-run model the intended direction? No need to update the RFC but it should be argued in the PR description.

Furthermore, the RFC has the sender set version. This script emits neither version nor `signal_type`, as already mentioned above. Does the receiver now computes it?

HereThereBeDragons:
Thanks for the review:

> * The RFC calls the TheRock-side script `send_to_quartz.py`; this PR ships `notify_quartz.py`. Worth aligning the RFC or the filename so the two docs don't drift?

true, i would rather update the RFC now to notify_quartz as this is a bit shorter but still says the same.

> * Default `receive_therock_data.yml` (underscores) vs RFC's `receive-therock-data.yml` (hyphens). It's configurable, but a mismatch with the actual file in the target repo produces a silent 404 on dispatch (masked by continue-on-error). Please confirm the real filename in `ROCm/quartz`.

Yes, we noticed that. Also missed to update the RFC. We switched from hyphens to underscores due to all the other workflows adhering to this standard.

> The most significant seems to be the dispatch granularity. The RFC says to call `send_to_quartz.py` on each build/test job at start and finish (`if: always()`), with `version = run_attempt*2 + signal_type`. This PR instead dispatches per workflow-run: one started (no jobs) and one completed signal carrying the whole jobs array fetched from the API. Different model than the documented schema/ingest path, why? Is the per-run model the intended direction? No need to update the RFC but it should be argued in the PR description.

This PR does not contain all the changes needed in theRock to enable quartz. This is just the dedicated workflow that we will run at the beginning and at the end of every workflow file called within the release flow.

Here is a short example workflow: https://github.com/ROCm/Quartz-Tester/blob/main/.github/workflows/test_pytorch_wheels.yml

This granularity of capturing metrics at the beginning and at the end of a workflow I think is a good trade off between not bloating the workflows too much but also getting enough info we need. The workflows offer a good level of granularity by design (setup, rocm build, rocm tests, pytorch build, pytorch tests, windows vs linux, jax? etc.)

If more fine-grained notification is wanted we can implement this at a later point without much refactoring, but already with the current configuration we should get around 60 - 80 commits per nightly run - and I am not sure we gain so much extra info having a more fine-grained reporting.

[side note: auto-triggering a workflow to capture the info was sadly not possible because you cannot capture workflow inputs or step outputs]

> Furthermore, the RFC has the sender set version. This script emits neither version nor `signal_type`, as already mentioned above. Does the receiver now computes it?

This is no longer needed as the notify_quartz job within the workflow says in which run_phase it is (started or completed) - and we get the run attempt for free

build_tools/github_actions/notify_quartz.py
ScottTodd, line 1: Can you add a link to the PR description to any tracking issues for Quartz work, or at least the RFC file: https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md . That should help others seeing this PR and the eventual merged commit to understand what Quartz is.

.github/workflows/notify_quartz.yml
```diff
@@ -0,0 +1,83 @@
...
+      quartz_workflow_file:
+        description: "Filename of the Quartz ingest workflow to dispatch to"
+        required: false
+        type: string
+        default: "receive_therock_data.yml"
+      quartz_workflow_ref:
+        description: "Branch in Quartz repo where the ingest workflow lives"
+        required: false
+        type: string
+        default: "main"
```
ScottTodd, line 67: Do we need the flexibility to dispatch any workflow file name from any ref in any repository, or will these be fixed for all uses of Quartz in TheRock? If we can limit this to a fixed value then we wouldn't need to be as careful about untrusted inputs being able to trigger workflows other than those we design for.

build_tools/github_actions/notify_quartz.py
```diff
@@ -0,0 +1,706 @@
...
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%(levelname)-8s %(message)s",
+    )
+    args = build_parser().parse_args(argv)
+
+    token = args.token
+    if not token:
+        log.error("No GitHub token provided (--token or QUARTZ_HAULY_TOKEN)")
+        return 1
```
ScottTodd: nit: can use `parser.error()` instead of `log.error()` and `return` here, or just mark `token` as `required=True`. I think `required` may still work with
```python
    p.add_argument(
        "--token",
        default=os.environ.get("QUARTZ_HAULY_TOKEN", ""),
```

see the suggest code patterns in https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#use-argparse-for-cli-flags

build_tools/github_actions/notify_quartz.py
```diff
@@ -0,0 +1,706 @@
...
+
+      * any `failure`   -> `failure`
+      * any `cancelled` -> `cancelled`
+      * any `success`   -> `success` (skipped jobs alongside successes
+        don't drag the verdict down; matches the inline
+        `contains(needs.*.result, 'failure') || ...` ternary callers
+        used to write)
+      * all `skipped`   -> `skipped` (every dependency was skipped, so
+        the run did effectively nothing — don't lie about success;
+        receiver maps `skipped` to a cancelled-style row state)
```
ScottTodd, line 636: We may want to redefine "success" and "skipped" a bit given how we have checks like "TheRock CI" set up here in TheRock, rocm-libraries, and rocm-systems. 

The job graph looks like:
```mermaid
graph TD
  setup --> linux_ci
  setup --> windows_ci
  linux_ci --> summary
  windows_ci --> summary
```

The "setup" and "summary" jobs always run, while the linux_ci and windows_ci jobs may be skipped depending on file edits, PR labels, etc. We might want that to show up here as "skipped"

I don't think we have any _release_ workflows with that level of dynamic filtering, so this probably won't factor in initially.

build_tools/github_actions/notify_quartz.py
ScottTodd, line 1: Depending on how much code we carry here to integrate with Quartz we may want a separate subfolder with other scripts, unit tests, documentation, etc.

I think both https://github.com/ROCm/TheRock/tree/main/build_tools and https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions could use some more organization (though not quite at the granularity of https://github.com/ROCm/TheRock/tree/main/build_tools/scan_tools/github_actions for a single file)

ScottTodd (approved):
This looks okay to land and iterate.

Do you have a list of milestones tracked somewhere where we could see how this script and workflow will get sequenced with other changes to implement https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md#implementation-phases ? Are we already at https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0011-Quartz-CICD-Datahub.md#d-therock-repository-changes now?

HereThereBeDragons:
sorry for the delays @scott, @marbre 
we had more changes needed (mainly adding the reporting_workflow addition so we know which exact workflow is responding) and i wanted to wait for that before replying to the comments.

build_tools/github_actions/notify_quartz.py
HereThereBeDragons: https://github.com/orgs/ROCm/projects/156 we have this. but those are not really update. these files here are standalone. they only become active when we add notify_quartz jobs to the workflows.
i can extend the pr description with it

.github/workflows/notify_quartz.yml
HereThereBeDragons: for therock it will always be branch `main` and `receive_therock_data.yml`.
however we want it to be flexible for development purpose (and having the development parts as close to prod all possible)

the workflow only offers workflow call and not workflow dispatch - so a normal user would not be able to trigger it via the web ui. not sure if making it a "choice" would make it any safer?

build_tools/github_actions/notify_quartz.py
HereThereBeDragons: i dont think i fully get your ask?
if in a pr we run setup and summary but skip linux_ci and windows_ci, the pr ci result will say the workflow ran successfully - that would also be the conclusion here. 
what would you like to have different? that if both linux_ci and windows_ci is skipped that everything is marked as skipped? that would be different then what the github api will report.

i would argue that kind of treatement you probably want quartz/database and not when transferring the raw metrics:
- quartz for status.json - but is not applicable for release/prerelease
- database for all other workflows we will capture in the future for therock (store the raw but get those changed conclusion during readout)

build_tools/github_actions/notify_quartz.py
HereThereBeDragons: i dont think this is needed we will only have those 3 files part of this pr + doc. all other changes are in the normal workflow files directly. 

for the doc it would probably the best:
- link people to the quartz repo for all the doc about how to use quartz as a downstream project
- have some doc about how to add quartz to workflows 

any preference where to put it? especially the 2nd part.

build_tools/github_actions/notify_quartz.py
ScottTodd: > if in a pr we run setup and summary but skip linux_ci and windows_ci, the pr ci result will say the workflow ran successfully - that would also be the conclusion here.

We may want it to say "workflow was skipped" instead of "workflow was successful" if a commit only changes markdown/text files and CI was skipped. I'd like to be able to query for "recent successful runs" and get back a list of only runs that actually ran builds, excluding runs that did nothing.

Fine to proceed without that fine detail and explore a few different ways to filter/implement it.

build_tools/github_actions/notify_quartz.py
ScottTodd: For adding quartz to workflows, these docs are currently most relevant:
* https://github.com/ROCm/TheRock/blob/main/docs/development/ci_overview.md
* https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md
* https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md

Fine to choose a spot in one of those or a new file. I want to reorganize those docs soon anyways. For example, I want to recategorize https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#testing-pytorch-release-workflows as more of a "how to use the CI system" guide and write similar content for each workflow that has a workflow_dispatch trigger, merging with https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#workflow-dispatch perhaps.

.github/workflows/notify_quartz.yml
```diff
@@ -0,0 +1,94 @@
...
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.GH_APP_HAULY_ID }}
+          private-key: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
+          owner: "ROCm"
```
ScottTodd: This generates a token scoped to all repositories that the app is installed for. While the install is limited to Quartz-relevant repositories, should we scope to a subset here in TheRock for some extra security? Maybe just `${{ inputs.quartz_repo }}`? (may need to strip the organization owner prefix for that input though)

https://github.com/actions/create-github-app-token#create-a-token-for-multiple-repositories-in-the-current-owners-installation

ScottTodd (approved):

build_tools/github_actions/notify_quartz.py
HereThereBeDragons: we are capturing all workflow inputs and all job/step outputs. my idea would be to dump this all raw into the db. so if we can get this from there (i guess setup.yml will provide this) we will be able to draw this conclusion

.github/workflows/notify_quartz.yml
```diff
@@ -0,0 +1,94 @@
...
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.GH_APP_HAULY_ID }}
+          private-key: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
+          repositories: ${{ inputs.quartz_repo }}
```
marbre, line 82: My agent flagged that the `repositories: owner/repo` form will break silently if `--quartz-repo` is overridden to something ROCm-org and therefore suggest to split to pass `owner: <org-part>` and `repositories: <repo-name-part>` separately. I don't think this doesn't matters here much, but as the script provides this option, we may want to make it more robust here.

build_tools/github_actions/notify_quartz.py
```diff
@@ -0,0 +1,761 @@
...
+        original_size = len(payload_json)
+        payload.get("workflow_run", {}).pop("jobs", None)
+        payload_json = json.dumps(payload)
+        fetch_jobs = True
+        inputs = {"payload_json": payload_json, "fetch_jobs": str(fetch_jobs).lower()}
+        log.warning(
+            "Inputs too large (%d encoded chars), stripped jobs (%d -> %d chars); "
+            "receiving side will re-fetch",
+            original_size,
+            original_size,
```
marbre: `"tripped jobs (%d -> %d chars)` passes `original_size` twice for the before value, so both before and after are in terms of payload_json size. That part is at least internally consistent, though the "before" is redundant with the first sentence.  Was the intend to capture `len(json.dumps(inputs))` before stripping and use that as the first arg?

marbre (approved):
LGTM

.github/workflows/notify_quartz.yml
HereThereBeDragons: from https://github.com/actions/create-github-app-token#repositories

> Repository entries may include an owner, for example owner/repo1. The owner portion must match the owner input, or the current repository owner if owner is unset.

so i dont think this is any issue.

build_tools/github_actions/notify_quartz.py
HereThereBeDragons: thanks for catching this

.github/workflows/notify_quartz.yml
marbre: The problem here is rather when this is cross-orgs. In that case the owner portion does not match and it breaks, thus what you copied from the docs rather underlines that this might be a potential problem @HereThereBeDragons.

HereThereBeDragons: the token creation is solely for the communication therock -> quartz.  as such the gh app is only installed in therock/rockrel/quartz and then in the quartz-sandboxes.
i cannot see a reason why cross-org we would want to allow a successful token generation.
maybe i am missing something?

PR #6000, opened 2026-06-19
[NOTFORMERGE] Enable HotSwap + Static Comgr

chinmaydd:


chinmaydd:
This should be superseded by 6002 / 6003.

Issue #5453, opened 2026-05-26
Use CMake presets more consistently in CI/CD workflows


ScottTodd:
Following up on https://github.com/ROCm/TheRock/pull/5153#discussion_r3220301745.

We have https://github.com/ROCm/TheRock/blob/main/CMakePresets.json but not all CI/CD workflows are using presets to set their CMake options (including flag settings). For these presets to work as expected, we should use them consistently in our workflows and make it clear when each preset should be used.

See also:
* https://github.com/ROCm/TheRock/issues/1781

### Background: how presets are used today

For multi-arch CI _**and**_ releases, this is the flow today:

* CMake presets are attached to "build variants" here: https://github.com/ROCm/TheRock/blob/587268c23ad01bed66314596ddee39412f9d040e/build_tools/github_actions/amdgpu_family_matrix.py#L89-L117
* The "build variant" is selected during CI configuration: https://github.com/ROCm/TheRock/blob/587268c23ad01bed66314596ddee39412f9d040e/build_tools/github_actions/configure_multi_arch_ci.py#L179-L183
* The cmake preset is extracted from the variant: https://github.com/ROCm/TheRock/blob/587268c23ad01bed66314596ddee39412f9d040e/build_tools/github_actions/configure_multi_arch_ci.py#L911
* The cmake preset is passed from the CI configuration through to the "build artifacts" workflow: https://github.com/ROCm/TheRock/blob/587268c23ad01bed66314596ddee39412f9d040e/.github/workflows/multi_arch_release_linux.yml#L46

### Specific issues with the current approach

https://github.com/ROCm/TheRock/pull/5153 intended to set `THEROCK_FLAG_STAMP_LIBRARY_GIT_VERSIONS` off for developers and CI builds and enable in releases, but due to how CMake presets are not used throughout all such workflows it actually enabled it for multi-arch Windows CI and did not enable it for Linux releases:

Configuration | Intended flag value | Actual flag value
-- | -- | -- 
Linux local development | off | ✅ off 
Linux CI | off | ✅ off
Linux releases | on | ❌ off
Windows local development | off | ✅ off
Windows CI | off | ❌ on
Windows releases | on | ✅ on

### Proposed fixes

I had drafted two different ways to approach this:

* https://github.com/ROCm/TheRock/pull/5191
* https://github.com/ROCm/TheRock/pull/5192

agent review feedback considering each:

[...]

PR #6034, opened 2026-06-22
docs(rfc): RFC Draft - Debug Symbol Generation & Packaging

amd-nicknick:
## Motivation

ISSUE ID : https://github.com/ROCm/TheRock/issues/5004

This RFC aims to address the existing gap of ROCm software field debugging / profiling difficulties by providing a unified mechanism for generating and packaging host-side debug symbols.

## Technical Details

This RFC introduces new CMake flags: `THEROCK_GENERATE_DEBUG_INFO`, `THEROCK_DEBUG_INFO_LEVEL`, `THEROCK_SPLIT_DEBUG_INFO` as the unified flags to control debug symbol generation across OS/configurations.
It also proposes removal of LLVM static object from artifacts to reduce size.

## Test Plan

N/A, RFC

## Test Result

N/A, RFC

## Submission Checklist


lamikr:
Instead of introducing THEROCK_GENERATE_DEBUG_INFO, THEROCK_DEBUG_INFO_LEVEL and THEROCK_SPLIT_DEBUG_INFO, would is be possible to use the existing

CMAKE_BUILD_TYPE=Release
CMAKE_BUILD_TYPE=Debug
CMAKE_BUILD_TYPE=RelWithDebInfo
CMAKE_BUILD_TYPE=MinSizeRel

flags in therock config phase and then pass those instead to subprojects and also check during our artifact managing phase separation of debug symbols to own files from main ones if needed. Not sure would cmake also allow us to have some own definitions for more fine grained instructions there if needed. (In style remove host side code debug info but keep gpu-side debug traces in place)

Separation of debug info may be needed to be done in the end of each package build, otherwise the binaries of some of the produced libs may grow so big that the linking of the other packages requiring them slows down and eventually fails.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+system. Its central requirement is that **Release builds remain Release builds**: debug
+information is added without downgrading optimization to `RelWithDebInfo`.
+
+Our goals are to:
+
+1. **Generate host-side debug symbols for all supported operating systems** in their
+   native formats (DWARF/`.debug` on Linux, PDB on Windows).
+1. **Cover every TheRock-compiled host binary and library built through CMake**,
+   including `rocm-systems`, `rocm-libraries`, CMake-built system dependencies, and the
+   bootstrapped compiler.
```
ScottTodd: _Every_ binary? Even LLVM? LLVM with debug symbols was slowwwww when last I checked. There may be other projects like that too. Could relax this to _most_ binaries/libraries?

---

For more context, we have https://github.com/ROCm/TheRock/blob/main/docs/development/development_guide.md#per-project-cmake-flags `-D{project}_BUILD_TYPE` to allow this flexibility. See https://github.com/ROCm/TheRock/pull/691 which added that:
> Adds a `{project_name}_BUILD_TYPE` override, allowing fine grained setting of build type per project. Used in our packaging preset to have the compiler build in Release vs RelWithDebInfo (which saves us some 600MB of debug info for the common case where we trust the compiler or debug it with reproducers vs stack traces).

and the `linux-release-package` CMake preset: https://github.com/ROCm/TheRock/blob/3ddfc3b69a79adc37dede9414369ad9431539f46/CMakePresets.json#L15-L18

...which we should still actually use, see 
* https://github.com/ROCm/TheRock/issues/5453
* https://github.com/ROCm/TheRock/issues/1781

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+| Build-id split pipeline      | `cmake/therock_global_post_subproject.cmake`, `cmake/therock_install_linux_build_id_files.cmake` | Forces `-Wl,--build-id`, runs `objcopy --only-keep-debug` to emit `.build-id/XX/YYY.debug`.                                                    |
+| `dbg` artifact component     | `build_tools/_therock_utils/artifact_builder.py`                                                 | Default include pattern `.build-id/**/*.debug` (Linux only).                                                                                   |
+| MSVC embedded debug          | `cmake/therock_compiler_config.cmake`                                                            | Sets `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded` (i.e. `/Z7`, not `/Zi`) for `Debug,RelWithDebInfo` — required for ccache/sccache.          |
+| Release presets              | `CMakePresets.json`                                                                              | `linux-release-package` uses `CMAKE_BUILD_TYPE=RelWithDebInfo` (`-O2`) to obtain symbols; `windows-release` uses `Release` with no debug info. |
```
ScottTodd: (Also see my other comment)

Note that the `linux-release-package` cmake preset is not actually used right now: https://github.com/ROCm/TheRock/issues/1781

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+  packaging concern — it only pays off when producing shippable `dbg` artifacts and adds
+  needless install-time overhead to local/dev builds. Embedding (`OFF`) is the local
+  developer mode on Linux; on Windows splitting is unconditional because PDBs cannot be
+  embedded in the image.
+
+The options inject flags into the **active configuration's** flag set rather than into
+a single build type. Concretely, the toolchain emitted by `therock_subproject.cmake`
+appends to the flag variables for the configuration being built — crucially including
+`CMAKE_<LANG>_FLAGS_RELEASE` — so that `Release` keeps `-O3 -DNDEBUG` and merely gains
+symbols.
```
ScottTodd: I originally agreed with @lamikr here... I'd prefer to use CMake's built in options if they work for our purposes.

I guess here we do actually want custom behavior though, keeping `-O3 -DNDEBUG` but adding symbol generation. Does that actually work though? Isn't that the point of `RelWithDebInfo`?
* https://stackoverflow.com/questions/48754619/what-are-cmake-build-type-debug-release-relwithdebinfo-and-minsizerel
* https://gist.github.com/MangaD/475b8b413aff7682b803fb007083fb5c
* https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

I think it would be easier to talk about this with data on hand:
* Binary size in each build mode
* Time to compile in each build mode
* Binary size of debug symbols
* Performance of some benchmarks in each build mode
* etc.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+This RFC therefore proposes to **remove the LLVM/Clang/Flang component static archives —
+along with the LLVM/Clang `include/` and CMake config content that only served them —
+from the shipped `amd-llvm` `dev` package.** The clang resource-directory archives
+(`lib/llvm/lib/clang/**`, used by the shipped `clang` to link user programs) and the
+runtime dylibs are retained. Because the archives are no longer shipped, no debug-info
+handling is required for them.
+
+> **Discussion** This is a user-visible packaging reduction: any downstream
+> consumer that statically links LLVM/Clang or uses `find_package(LLVM/Clang)` against
+> the shipped SDK would lose that capability.
```
ScottTodd: Can this be moved out of this RFC? I'm not sure it's in scope here, and there may be some projects that do actually depend on this (despite warnings to not)

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+
+### 6. Packaging
+
+Debug symbols are surfaced as **separate deliverables** in every channel, leaving
+runtime and development packages stripped:
+
+- **Tarballs:** the `dbg` artifact component is archived and published as a distinct
+  debug tarball per build, parallel to the runtime/devel tarballs.
+- **Python / `rocm_sdk`:** native libraries shipped by `rocm_sdk` are accompanied by a
+  separate debug wheel (or archive) carrying their `.debug`/`.pdb` files.
```
ScottTodd: This sounds tricky. High risk here of mismatched binaries and their symbols. Is there precedent for python packages just containing debug symbols for other python packages?

We could also consider:
* a remote symbol server 
* a separate pip index with packages that include debug symbols (maybe...?)

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+
+### 7. CI integration
+
+These changes become the default for nightly and release builds:
+
+- **Linux preset** (`CMakePresets.json` `linux-release-package`): switch
+  `CMAKE_BUILD_TYPE` from `RelWithDebInfo` back to `Release` and set
+  `THEROCK_GENERATE_DEBUG_INFO=ON`, `THEROCK_DEBUG_INFO_LEVEL=full`,
+  `THEROCK_SPLIT_DEBUG_INFO=ON`. The per-subproject `*_BUILD_TYPE=Release` overrides
+  for `amd-llvm`, host-blas, and SuiteSparse remain.
```
ScottTodd: This is what I was expecting, but the paragraphs higher up in this RFC conflict (saying that _all_ libraries ship symbols, including LLVM)

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Phase 5: CI default enablement
+
+- Update presets and workflows; publish debug artifacts to the debug bucket.
+- Enable by default for nightly/release.
+
+## Risks and Considerations
+
+- **Package size / storage.** Full debug symbols are large. Mitigated by always
+  splitting, compressing (`-gz`), shipping debug separately.
```
ScottTodd: Could really use data here for how large they are (and how much including them affects build time). Hard to judge this without it.

I'd add an implementation plan phase that is just collecting data and performing experiments, before going too deep.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: IMO from a developer's perspective, I'd want `RelWithDebInfo` to do exactly as what it normally does:
* `-O2`
* `-g`
* (Of course for Windows we'd add `/Z7` / proper pdb splitting).

It is clear that when I choose to use `RelWithDebInfo`, I want binaries that is "close to release, but still source step-able". This is the expected behavior whenever I come across a CMake project and I'd like to upheld this expectation.

The entire reason I'd like to split up to separate CMake flags is to support `Release` preset, but with symbol generation:
* `-O3`
* `-g`

The symbols can still tell you a lot of info (globals, types / structs, function name,...) but source line stepping becomes unreliable. We'd normally only expect assembly level stepping & debugging.

TLDR:
If we switch `RelWithDebInfo` to `-O3`, we break the local debuggability expectation. If we use `RelWithDebInfo` directly, we lose `-O3` optimization.

I think we should strive to do `-O3` + `-g` for releases (anything coming out of CI by default), as this is the config that does not sacrifice performance for debuggability.

I agree with the data collection part, build time / binary size (it shouldn't) should not be impacted by this. That's why I'm building a PoC first and we could observe impact and discuss based on data :)

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: > _Every_ binary? Even LLVM? LLVM with debug symbols was slowwwww when last I checked. There may be other projects like that too. Could relax this to _most_ binaries/libraries?

We should build LLVM with `-O3` + `-g`, it won't be slow (in runtime, build time needs experiment) since it would be identical to `Release`. I have encountered in-field LLVM issues and symbol files are crucial to debug, as these often appears only in specific condition / has very low failure rate.

> For more context, we have https://github.com/ROCm/TheRock/blob/main/docs/development/development_guide.md#per-project-cmake-flags `-D{project}_BUILD_TYPE` to allow this flexibility.

Yes, we're not getting rid of that, the new flags `THEROCK_GENERATE_DEBUG_INFO`, `THEROCK_DEBUG_INFO_LEVEL`, `THEROCK_SPLIT_DEBUG_INFO` works globally in all subprojects. For example:
* `THEROCK_GENERATE_DEBUG_INFO=1`
* `THEROCK_DEBUG_INFO_LEVEL=full`
* `THEROCK_SPLIT_DEBUG_INFO=1`
* `amd-llvm_BUILD_TYPE=Release`

Will build `amd-llvm` with `-O3` +`NDEBUG` + `-g`/`Z7`+`/PDB:`, and split the debug symbol at the end of build on Linux.
Another example:
* `THEROCK_GENERATE_DEBUG_INFO=1`
* `THEROCK_DEBUG_INFO_LEVEL=full`
* `THEROCK_SPLIT_DEBUG_INFO=1`
* `amd-llvm_BUILD_TYPE=RelWithDebInfo`

Will build `amd-llvm` with `-O2` +`NDEBUG` + `-g`/`Z7`+`/PDB:`, and split the debug symbol at the end of build on Linux.


**However, you're right that this design does NOT support individual target selecting different debug symbol policy.**
Eg. I cannot have `amd-llvm` to NOT generate debug info while the rest generates.
For this case, I think we could base on the experiment & observe the size/time impact to decide if we need mechanism. @ScottTodd, thoughts?

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: See my other comment, this paragraph is badly worded, it should be:

> The per-subproject `*_BUILD_TYPE=Release` overrides for build types remains, but symbol generation logic will follow the new flags `THEROCK_GENERATE_DEBUG_INFO`, `THEROCK_DEBUG_INFO_LEVEL`, `THEROCK_SPLIT_DEBUG_INFO` regardless of the default CMake behavior.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: Cooking PoC branch, I think we could just continue the discussion alongside poc & not bloat up the RFC as it is more of a spec?

If you think it's valuable to add PoC details for documentation purpose I'm happy to add :)

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: Python side it's kind of a mess... I researched a bit and it looks like most projects expect you to rebuild Debug yourself if you ever need symbols.

Pytorch used to generates & distribute as separate zip, but they stopped a while ago: https://github.com/pytorch/pytorch/blob/5b88630d363bf79e0dc9263588cf5f8054af54cf/.ci/manywheel/build_common.sh#L397-420

Remote symbol server would be the best option here, but I think we could leave that as 2nd phase?
There are more problems with database, structure, ingest / egress, splitting by nightlies/pre-submit/packaging...

I'm starting to consider if we should just drop Python native libs from this RFC's scope & pick up in the future, since matching the symbol with wheel would cause some confusions without symbol server.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
amd-nicknick: Without this part, it will cause the artifact size grow problem back when I added ccache for Windows.
https://github.com/ROCm/TheRock/pull/4419#discussion_r3262592363

There's 2 paths forward from my pov:
1. Keep this in RFC to remove.
2. Redesign partially, add support for individual project debug symbol toggle (prev comment bold text https://github.com/ROCm/TheRock/pull/6034#discussion_r3480056135).
    **Even if we add it, we still lose the ability to debug llvm.**

amd-nicknick:
@lamikr
> Instead of introducing THEROCK_GENERATE_DEBUG_INFO, THEROCK_DEBUG_INFO_LEVEL and THEROCK_SPLIT_DEBUG_INFO, would is be possible to use the existing
> 
> CMAKE_BUILD_TYPE=Release CMAKE_BUILD_TYPE=Debug CMAKE_BUILD_TYPE=RelWithDebInfo CMAKE_BUILD_TYPE=MinSizeRel

I'd hope to use them as well, but CMake has undesired defaults, see: https://github.com/ROCm/TheRock/pull/6034#discussion_r3478117982

> flags in therock config phase and then pass those instead to subprojects and also check during our artifact managing phase separation of debug symbols to own files from main ones if needed. Not sure would cmake also allow us to have some own definitions for more fine grained instructions there if needed. (In style remove host side code debug info but keep gpu-side debug traces in place)

There's https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG_INIT.html which we could override the default flags :)

> Separation of debug info may be needed to be done in the end of each package build, otherwise the binaries of some of the produced libs may grow so big that the linking of the other packages requiring them slows down and eventually fails.

Yes agreed. If the output is a shared object, it should be split. The problem comes from LLVM, where we also packaged static object files used for static linking. See: https://github.com/ROCm/TheRock/pull/6034#discussion_r3480211012

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
ScottTodd: Yeah let's get the data. I think we may want the option to choose which subprojects generate debug info as we can already change build type, sanitizer setting, etc. Some subprojects may generate too much debug info, may not be compatible with debug info settings (e.g. if they override the options that we try to set), or might be too slow somehow when it is enabled.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
ScottTodd: I'd want to run any changes to the LLVM/Clang/Flang files that we distribute by some of the LLVM maintainers/contributors like @lamb-j and @ronlieb . I'm not sure how much downstream would be affected by such changes.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
ScottTodd: > I'm starting to consider if we should just drop Python native libs from this RFC's scope & pick up in the future, since matching the symbol with wheel would cause some confusions without symbol server.

I'm fine with leaving python packaging details to a follow-up. Let's make progress where it's easier first.

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
ronlieb: > I'd want to run any changes to the LLVM/Clang/Flang files that we distribute by some of the LLVM maintainers/contributors like @lamb-j and @ronlieb . I'm not sure how much downstream would be affected by such changes.

@kzhuravl  and @slinder1  ^^

docs/rfcs/RFC00XX-Debug-Symbol-Generation-Packaging.md
```diff
@@ -0,0 +1,326 @@
...
+
+The existing pipeline is retained and made Release-clean:
+
+1. Targets are compiled with `-g{level} -gz` injected into the active config and linked
+   with `-Wl,--build-id` (already forced in `therock_global_post_subproject.cmake`).
+1. At install, `objcopy --only-keep-debug` extracts symbols from the unstripped
+   build-tree binary into `.build-id/XX/YYY.debug`
+   (`therock_install_linux_build_id_files.cmake`); the staged binary is stripped via
+   `cmake --install --strip`.
+1. The `dbg` artifact component collects `.build-id/**/*.debug`.
```
raramakr, line 193: @amd-nicknick Are we planning to create dbg artifact components for the ASAN build as well?

raramakr: cc @ScottTodd

PR #6039, opened 2026-06-22
fix(ci): Fix fetch_artifacts to match xnack-suffixed artifact variants

### LINKED ISSUE #6038: install_rocm_from_artifacts.py / fetch_artifacts.py drop xnack-suffixed artifacts when only base arch is requested (breaks ASAN test runs)
### Summary
When `--amdgpu-targets gfx942` is passed, the artifact fetch matches only the literal `gfx942` and silently drops xnack-suffixed variants such as `gfx942:xnack+`. ASAN builds emit their test artifacts with the `:xnack+` suffix, so those test artifacts are never downloaded, and downstream test jobs fail with missing artifacts.

### Reproduction
```
python ./build_tools/install_rocm_from_artifacts.py --amdgpu-targets gfx942
```
pulls only base-arch artifacts (`blas_lib_gfx942`, `fft_lib_gfx942`, ...), whereas
```
python ./build_tools/install_rocm_from_artifacts.py --amdgpu-targets gfx942,gfx942:xnack+
```
additionally pulls `blas_lib_gfx942:xnack+`, `blas_test_gfx942:xnack+`, `rccl_test_gfx942:xnack+`, etc.

Example artifact that is **not** fetched with base arch alone:
`https://therock-ci-artifacts.s3.amazonaws.com/27891627402-linux/rccl_test_gfx942%3Axnack%2B.tar.zst`

### Root cause
- `build_tools/fetch_artifacts.py` builds `targets_to_match = {"generic"} ∪ amdgpu_targets` and includes an artifact only if `ArtifactName.target_family in targets_to_match` (exact equality).
- `build_tools/_therock_utils/artifacts.py` (`ArtifactName.from_filename`) parses `rccl_test_gfx942:xnack+.tar.zst` into `target_family = "gfx942:xnack+"`.
- `"gfx942:xnack+" != "gfx942"`, so the artifact is excluded. The xnack suffix is never normalized or treated as a variant of the base arch.

### Context
- Nightly builds currently do **not** produce separate xnack artifacts (e.g. https://therock-nightly-artifacts.s3.amazonaws.com/27922350966-linux/index.html ), so this is invisible on the nightly path.
- **ASAN** CI builds **do** emit `:xnack+`-suffixed artifacts (e.g. `rccl_test_gfx950:xnack+.tar.zst` at https://therock-ci-artifacts.s3.amazonaws.com/27891627402-linux/index.html ), and this has not been plumbed through the build/CI/CD scripts.
- Believed to be the root cause of ROCM-25535.

### Proposed direction (to discuss)
Either (a) match base arch inclusively so a requested `gfx942` also pulls its `gfx942:xnack+` variants, or (b) have the test/CI workflow explicitly request the xnack variants. Preference seems to be pulling all variants of a requested base arch. Noted separately that this script is overly static and slated for replacement.

geomin12:
When requesting a base arch (e.g., gfx942), also match variants with xnack suffixes (e.g., gfx942:xnack+, gfx942:xnack-). This fixes ASAN test runs which emit artifacts with :xnack+ suffixes that were being silently dropped.

Fixes #6038
ISSUE ID https://github.com/ROCm/TheRock/issues/6038

test logs locally:
```
$ python3 build_tools/fetch_artifacts.py --run-id 27891627402 --amdgpu-targets gfx942  --dry-run
Retrieving bucket info for workflow run...
  github_repository: ROCm/TheRock
  release_type: ci
  workflow_run_id: 27891627402
  head_github_repository: ROCm/TheRock
  is_pr_from_fork: False
  bucket: therock-ci-artifacts
Retrieving artifacts from 's3://therock-ci-artifacts/27891627402-linux'
Matching artifact target families: ['generic', 'gfx942']

Filtered artifacts to download:
  s3://therock-ci-artifacts/27891627402-linux/amd-dbgapi_dbg_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/amd-dbgapi_dev_generic.tar.zst
...
  s3://therock-ci-artifacts/27891627402-linux/rccl_doc_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rccl_lib_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rccl_lib_gfx942.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rccl_run_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rccl_test_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rccl_test_gfx942:xnack+.tar.zst <----- this one is here now!!
  s3://therock-ci-artifacts/27891627402-linux/rdc_dbg_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rdc_dev_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rdc_doc_generic.tar.zst
  s3://therock-ci-artifacts/27891627402-linux/rdc_lib_generic.tar.zst
...
  s3://therock-ci-artifacts/27891627402-linux/wsl-rocdxg_lib_generic.tar.zst

Skipping downloads since --dry-run was set
```

### DIFF
--- a/build_tools/fetch_artifacts.py
+++ b/build_tools/fetch_artifacts.py
@@ -59,6 +59,31 @@ def log(*args, **kwargs):
     sys.stdout.flush()
 
 
+def _get_base_arch(target: str) -> str:
+    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'."""
+    return target.split(":")[0]
+
+
+def _matches_target(artifact_target: str, requested_targets: set[str]) -> bool:
+    """Check if artifact target matches requested targets (base arch matches variants)."""
+    # Direct match
+    if artifact_target in requested_targets:
+        return True
+
+    # Check if the artifact's base arch matches any requested target
+    artifact_base = _get_base_arch(artifact_target)
+    if artifact_base in requested_targets:
+        return True
+
+    # Check if any requested target's base arch matches the artifact's base arch
+    # (handles case where user requests "gfx942:xnack+" explicitly)
+    for requested in requested_targets:
+        if _get_base_arch(requested) == artifact_base:
+            return True
+
+    return False
+
+
 def list_artifacts_for_group(
     backend: ArtifactBackend,
     artifact_group: str | None,
@@ -69,6 +94,9 @@ def list_artifacts_for_group(
     Inclusive matching: accepts both family-named archives (mono-arch pipeline)
     and individual-target archives (split/kpack pipeline). Whichever naming
     convention is present in the bucket will be matched.
+
+    Base architecture matching: requesting a base arch (e.g., "gfx942") will
+    also match variants with suffixes (e.g., "gfx942:xnack+", "gfx942:xnack-").
 
     Args:
         backend: ArtifactBackend instance configured for the target run
@@ -103,7 +131,7 @@ def list_artifacts_for_group(
     data = set()
     for filename in all_artifacts:
         an = ArtifactName.from_filename(filename)
-        if an and an.target_family in targets_to_match:
+        if an and _matches_target(an.target_family, targets_to_match):
             data.add(filename)
 
     if not data:
--- a/build_tools/fetch_artifacts.py
+++ b/build_tools/fetch_artifacts.py
@@ -61,27 +61,18 @@ def log(*args, **kwargs):
 
 def _get_base_arch(target: str) -> str:
     """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'."""
-    return target.split(":")[0]
+    if not target:
+        return ""
+    base = target.split(":")[0]
+    return base if base else target
 
 
 def _matches_target(artifact_target: str, requested_targets: set[str]) -> bool:
-    """Check if artifact target matches requested targets (base arch matches variants)."""
-    # Direct match
-    if artifact_target in requested_targets:
-        return True
-
-    # Check if the artifact's base arch matches any requested target
-    artifact_base = _get_base_arch(artifact_target)
-    if artifact_base in requested_targets:
-        return True
-
-    # Check if any requested target's base arch matches the artifact's base arch
-    # (handles case where user requests "gfx942:xnack+" explicitly)
-    for requested in requested_targets:
-        if _get_base_arch(requested) == artifact_base:
-            return True
-
-    return False
+    """Match if the artifact's base arch equals any requested target's base arch."""
+    if not artifact_target:
+        return False
+    requested_bases = {_get_base_arch(t) for t in requested_targets}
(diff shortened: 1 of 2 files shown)


build_tools/fetch_artifacts.py
```diff
@@ -59,6 +59,31 @@ def log(*args, **kwargs):
     sys.stdout.flush()
 
 
+def _get_base_arch(target: str) -> str:
+    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'."""
+    return target.split(":")[0]
```
davidd-amd: I was going to ask about cases like `gfx123:foo:xnack+` but this looks like it will cover it. That said do we need to guard against `len(target.split(":")[0]) == 0` and what if we get back `garbage-*&%^%$-gfx942` will that survive?

build_tools/fetch_artifacts.py
```diff
@@ -59,6 +59,31 @@ def log(*args, **kwargs):
...
+    if artifact_base in requested_targets:
+        return True
+
+    # Check if any requested target's base arch matches the artifact's base arch
+    # (handles case where user requests "gfx942:xnack+" explicitly)
+    for requested in requested_targets:
+        if _get_base_arch(requested) == artifact_base:
+            return True
+
+    return False
```
marbre: My agent suggest to simplify to

```suggestion
def _matches_target(artifact_target: str, requested_targets: set[str]) -> bool:
    """Match if the artifact's base arch equals any requested target's base arch."""
    requested_bases = {_get_base_arch(t) for t in requested_targets}
    return _get_base_arch(artifact_target) in requested_bases
```

* Branch 1 (direct match) is subsumed: equal strings have equal bases.
* Branch 2 (artifact_base in requested_targets) and branch 3 (loop) both just test base-arch overlap.

build_tools/fetch_artifacts.py
```diff
@@ -70,6 +95,9 @@ def list_artifacts_for_group(
     and individual-target archives (split/kpack pipeline). Whichever naming
     convention is present in the bucket will be matched.
 
+    Base architecture matching: requesting a base arch (e.g., "gfx942") will
+    also match variants with suffixes (e.g., "gfx942:xnack+", "gfx942:xnack-").
```
marbre, line 90: Might there be a case where one really only want to fetch exactly one? I guess not, so this should be fine. Just to confirm.

build_tools/fetch_artifacts.py
marbre: I think we need to to guard as str.split will always return a non-empty list and even with an empty base no-matches are harmless. Same for garbage.

marbre (approved):

PR #6045, opened 2026-06-22
ci: make compiler-runtime build tests (Comgr) blocking

lamb-j:
Splits the compiler-runtime "Run build tests" step so that **only the Comgr suite gates CI**, while the remaining build tests keep running and reporting without blocking.

- **Run build tests (Comgr, blocking)** — runs `build-test-amd-comgr` (lit + ctest + test-unit) with no `continue-on-error`, so Comgr failures fail the job.
- **Run build tests (other, non-blocking)** — runs the full `therock-build-tests` aggregate with `continue-on-error: true`, so rocjitsu/mirage keep reporting but do not block.

This step is shared via `multi_arch_ci_linux.yml`, so Comgr now gates both the primary Multi-Arch CI and the Multi-Arch CI ASAN runs.

rocjitsu is being moved out of this aggregate separately; once that lands the non-blocking step can be dropped.

ISSUE ID : #6112


lamb-j:
@kasaurov I think we need some fixes from the next compiler promotion before the libhipcxx test will pass. Mostly filing this as a WIP to get these blocking (which I think they need to be)

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -175,7 +175,6 @@ jobs:
 
       - name: Run build tests
         if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
```
kasaurov, line 178: It looks like `[rocjitsu build-test]` is now a part of `Run Build tests` in `Stage - Compiler Runtime` and it is failing, so we need to narrow down the condition for  `continue-on-error:`.

kasaurov:
I can see two ways to have both blocking and non-blocking unit tests:
1. Split the step into two steps in the workflow. Simple, but has a downside - component-specific target name in the YAML, something like: 
```
- name: Run build tests (blocking)
  if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
  run: |
    cmake --build "${BUILD_DIR}" --target build-test-amd-comgr -- -k 0
- name: Run build tests (non-blocking)
  if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
  continue-on-error: true
  run: |
    cmake --build "${BUILD_DIR}" --target therock-build-tests -- -k 0
```
2. Add a BLOCKING option to `therock_cmake_subproject_build_test()` and a second aggregate target. A little bit more complex, but keeps the YAML fully generic and allows each component to declare its own blocking/non-blocking status in CMake.

@ScottTodd which one seems more reasonable and aligned with TheRock?

ScottTodd:
For simple and deterministic unit tests, there's no better way than making the tests pass and keeping them passing. I'd do the simple thing in the short term to get the tests that do pass marked as blocking... which is probably your option 1.

lamb-j:
We shouldn't need to split anything up. I talked to the rocjitsu team, and they're planning to move those tests to runtime tests, as some of them use GPUs @kasaurov

So we just need to wait for that to happen, and for the libhipcxx fix to come in, then we can land this

amd-eochoalo:
Hello, 👋, I'm the person who is moving the rocjitsu tests. I'm kind of new to TheRock's code base. I just want to verify that moving them to runtime-tests is the way to go. I'm exploring the code base and I'd like to understand the difference between placing the test execution on runtime-tests and on the artifact test job (or test_component.yml?). I was trying to model the move similar to what rocPRIM is doing where tests are built and installed in one stage (for the case of rocjitsu in compiler-runtime) and then executed in a later stage. But rocPRIM executes the tests on (I think?) test_artifact, test_component,

ScottTodd:
> I'm exploring the code base and I'd like to understand the difference between placing the test execution on runtime-tests and on the artifact test job (or test_component.yml?)

See https://github.com/ROCm/TheRock/blob/main/docs/development/adding_tests.md (maybe we should add a section for `therock_cmake_subproject_build_test` there?)

What most subprojects in TheRock do right now is
1. build a test artifact (https://github.com/ROCm/TheRock/blob/main/docs/development/artifacts.md) that includes their installable test binaries/libraries/data
2. On CI, after the full source build of all artifacts completes, https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_artifacts.yml runs and it uses https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_component.yml for each subproject/component (e.g. hipfft, rocblas, hipblaslt, etc.)
3. That test_component.yml runs the project's installable tests on a GPU runner. Sometimes these GPU runners have multi-hour queues to get scheduled and sometimes the GPU runners have driver/hardware/network issues.

What comgr has been adding here has been unit tests that run on the CPU build machines as part of the build process. The tests happen to make use of the build directory itself, rather than package files into an installable test artifact (that would mean _lots_ of files for LLVM).

amd-eochoalo:
> What comgr has been adding here has been unit tests that run on the CPU build machines as part of the build process. The tests happen to make use of the build directory itself, rather than package files into an installable test artifact (that would mean lots of files for LLVM).

Thanks! I think this clarifies some things. Originally I was told that the runtime-tests stage had runners with GPUs available. I don't believe that is the case and only running via  test_component.yml actually runs on machines with access to GPUs. 

Also, I don't believe we are moving the comgr tests which use LLVM, I'm only thinking of moving the rocjitsu tests which don't require LLVM. (Well, one does use llvm-mc, but it is conditionally executed and I believe can be skipped in this case without causing too much trouble). I think that having an "INSTALL_TESTS" option for rocjitsu that builds the test artifact during compiler-runtime stage and then using test_component.yml seems to be the path forward. Thanks!

ScottTodd:
> Thanks! I think this clarifies some things. Originally I was told that the runtime-tests stage had runners with GPUs available. I don't believe that is the case and only running via test_component.yml actually runs on machines with access to GPUs.

Correct, the "runtime-tests" _build stage_ just builds `hip-tests`, `rocrtst`, etc. and produces test artifacts for those components. The tests are not run on GPU machines until test_artifact.yml/test_component.yml.
* https://github.com/ROCm/TheRock/blob/c677e87f485518632db39d7e06013b34a58b97e6/.github/workflows/multi_arch_build_portable_linux.yml#L92-L96
* https://github.com/ROCm/TheRock/blob/c677e87f485518632db39d7e06013b34a58b97e6/BUILD_TOPOLOGY.toml#L267-L271

ScottTodd (commented):
Marking as "reviewed" to drop from my review queue until tests pass. Happy to start enforcing these tests once ready.

skganesan008 (approved):

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -174,7 +174,17 @@ jobs:
...
+        env:
+          TEATIME_FORCE_INTERACTIVE: 1
+        run: |
+          cmake --build "${BUILD_DIR}" --target build-test-amd-comgr -- -k 0
+
+      # Remaining build tests (e.g. rocjitsu, mirage) still run and report, but
+      # are not yet gating. Comgr is split out above so its failures block CI.
+      # TODO(#6112): rocjitsu tests are being moved out; drop this step once done.
+      - name: Run build tests (other, non-blocking)
         if: ${{ !cancelled() && inputs.stage_name == 'compiler-runtime' }}
```
ScottTodd, line 188: Oh, the logic here still only runs for the compiler-runtime stage? We have other stages that want to start running build tests too (https://github.com/ROCm/TheRock/pull/7376 in the profiler-apps stage)

Was there something stopping this from just using `if: ${{ !cancelled()`?

ScottTodd: Oh actually that PR is building the new subproject in the compiler-runtime stage so this is still sufficient. We should try to generalize fully soon though.

Issue #6048, opened 2026-06-23
RFC: Custom Python bindings layering


BrianHarrisonAMD:
This issue tracks discussion for a proposed RFC defining the ownership and packaging model for custom ROCm library Python bindings.

The RFC uses hipDNN as the motivating example and proposes:

- keeping native library ownership in `rocm-libraries`;
- keeping TheRock / `rocm_sdk` focused on normal native ROCm SDK artifacts;
- relocating `hip-python` into the shared bindings repository as the generated C API binding area;
- placing custom hipDNN binding projects under `rocm-bindings/hipdnn/` with `hipdnn-nanobind` producing extension artifacts and `hipdnn-frontend` producing the final wheel;
- defining Python ABI, runtime/device package, manifest, and validation responsibilities for custom binding wheels.

A draft PR will add the RFC document and link back to this issue.

BrianHarrisonAMD:
RFC PR opened: https://github.com/ROCm/TheRock/pull/6050

Issue #5701, opened 2026-06-08
rocm-sdk: import-test shipped Python extensions + resolve abi3 vs Python <3.12 support


davidd-amd:
## Decision (recorded)

**We will not contort the product to accommodate a test-only dependency like `rocisa`. It remains a known sharp edge.**

- `rocisa` (`_rocisa.abi3.so`, nanobind `STABLE_ABI` → CPython **3.12+ only**) ships in `rocm-sdk-libraries` but is **test/codegen-only**; it is not a supported runtime import for SDK consumers.
- Python **3.10/3.11 are a hard requirement** for the product, and abi3 cannot cover them. Rather than per-version builds or packaging changes for a test-only dep, we accept that `rocisa` is **not importable on 3.10/3.11** and leave it as-is.
- The #5529 change — `testSharedLibrariesLoad` skipping `.abi3.so` / `.cpython-*` extensions — is the **accepted accommodation**. It stays.

## Why this is acceptable here (and the boundary)

The skip + sharp-edge is tolerable **only because `rocisa` is test-only**. The leniency must NOT be read as license to ship unimportable extensions generally:

- A **runtime-required** Python extension shipped to SDK consumers must be importable on **all** supported Pythons (3.10–3.14). For nanobind that means **per-version builds** (abi3 alone is 3.12+).
- A **test/build-only** extension may either remain a documented sharp edge (rocisa) or be excluded from the version-agnostic runtime wheel (the pattern origami uses).

## Remaining (small) work

1. **Document the sharp edge**: note that `rocisa` is test/codegen-only and not importable on Python < 3.12; consumers must not depend on importing it at runtime on 3.10/3.11.
2. **Record the policy** above (runtime extensions → per-version / import-testable on all supported Pythons; test-only → sharp edge or excluded) somewhere durable (packaging docs).
3. The `ctypes.CDLL` skip in `libraries_test.py` stands; do **not** convert it into an `importlib` import for `rocisa` (that would force the declined product changes).

## Context

- Surfaced during the origami reland (#5650), which keeps its (non-abi3) extension test-only / excluded from the SDK — the correct pattern given 3.10/3.11 are required.
- Related: #5529 (the accepted `ctypes.CDLL` skip + rocisa stable-ABI test), rocisa `ROCISA_USE_STABLE_ABI`.

ScottTodd:
Yes, Python 3.10 and 3.11 support is still required for ROCm, and PyTorch.

[...]

PR #5609, opened 2026-06-03
docs/rfcs: Add RFC for HIP Python integration into TheRock

domcharrier:
Proposes integrating HIP Python (official Pythonic ROCm bindings) into TheRock's super-build and CI: single-branch generated bindings, wheels assembled by staging ROCm artifacts (no second compile), and the open decision of the code-generation-in-CI model (automatic-per-build vs developer-triggered with an API freeze).
## Motivation
HIP Python is the official Pythonic ROCm bindings (named in the ROCm programming guide, on PyPI since 2023) and is already a core dependency of RAPIDS ports (hipDF, hipMM) and Numba HIP. Its newest version has been restructured into a multi-package layout that deliberately mirrors TheRock's component split, which makes TheRock the natural home for building and shipping the bindings.
This RFC exists to:
- Get agreement on **landing HIP Python in TheRock's super-build and CI** on a single branch carrying the generated bindings.
- Resolve the key open decision: **how code generation runs in CI** — fully automatic per build vs developer-triggered at release prep (with an API freeze).
- Surface the related decisions that block the work: **generator licensing** (open-source vs keep private inside AMD-AIOSS, which itself tilts the codegen-model choice), final **package naming/contents**, the **Python version matrix**, and refreshing the **outdated public docs**.
It captures a public record of the proposed approach and the decisions to be made, consistent with how the other RFCs in `docs/rfcs/` are used.
## Technical Details
This is a **documentation-only** change. It adds one new RFC and links it from the index:
- New: `docs/rfcs/RFC0012-HIP-Python-TheRock-Integration.md`
- Modified: `docs/rfcs/README.md` (adds the RFC0012 entry to the index)
The RFC follows the existing RFC conventions (YAML metadata block with `author(s)`/`created`/`modified`/`status: draft`/`discussion`) and covers:
- **Architecture** — multi-package split (with a Mermaid diagram of the packages and their current contents), the code-generation pipeline (hardened by a ~190-test "parkour"), the two-layer `sccache`-friendly super-build (compile once, assemble wheels by staging artifacts — no second compile), the flexible shared-library loader, host/device **buffer-protocol interoperability** (PEP 3118: `array.array`/`bytearray`/NumPy; device: CUDA Array Interface for CuPy/Numba HIP), and the CUDA Python interop layer.
- **Mapping to TheRock sources**, the **proposed integration** (single-branch generated bindings, build/packaging in CI, docs from `*.pyi` stubs).
[...]

PR #6050, opened 2026-06-23
docs(rfcs): add custom Python bindings layering RFC

BrianHarrisonAMD:
ISSUE ID : #6048

## Summary

Adds an RFC for how custom ROCm library Python bindings should be layered, using hipDNN as the example. Separates native library ownership, generated C API bindings, custom extension artifacts, and final wheel packaging. Discussion issue: [#6048](https://github.com/ROCm/TheRock/issues/6048).

Filed as `RFC00XX` on purpose. `RFC0013` is claimed by both this PR and [#6904](https://github.com/ROCm/TheRock/pull/6904), and there are several RFCs in flight against a contended numbering space, so the number gets assigned at merge. Same approach as [#6034](https://github.com/ROCm/TheRock/pull/6034) and [#6118](https://github.com/ROCm/TheRock/pull/6118).

## Risk Assessment

Minimal. Docs only, no product code, build config, packaging behavior, APIs, or tests affected.

## Review Feedback Addressed

- **Background section** relating the three Python surfaces: the `rocm-sdk-*` wheels, generated `hip-python` bindings, and custom component bindings.
- **Layout** changed to `rocm-bindings/<component>/python/[backend,frontend]`, grouped by component then language, with `nanobind` dropped from the directory names.
- **Release channel** section built on [RFC0012](https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0012-Repo-Structure.md): binding wheels publish from `extras/rocm-bindings/` into the central `whl/` and `whl-next/` indices, on a cadence the bindings repo owns. Remaining decisions are listed as open items.
- **Cross-repo compatibility** section: public API coverage belongs in `rocm-libraries`, not in the bindings repo used as a safety net.
- **Mixing generated and custom bindings** section covering the hipFile case, with a preference order and the packaging consequence.
- **Related documents** are hyperlinks now.

## Doc Drift Resolved

- **TheRock no longer builds hipDNN Python bindings.** [#6425](https://github.com/ROCm/TheRock/pull/6425) removed the packer, pytest script, CI matrix entry, `share/hipdnn/python/**` artifact rules, and the `nanobind`/`robin-map` subprojects. The "transitional packer to be retired" text is updated, and the attempt is written up as a rejected alternative.
- **Current hipDNN state** rewritten against the real tree. The scikit-build and parent-CMake modes described before are gone; it's now a standalone CMake extension project plus a separate staging project, which makes the relocation closer to a move than a rewrite.
- **Python ABI strategy** reconciled with [#5701](https://github.com/ROCm/TheRock/issues/5701). hipDNN's 3.12 floor is currently implicit, and the RFC asks reviewers to make it explicit.
- **Validation** gains clean-environment install and RPATH audit checks, both from [#5678](https://github.com/ROCm/TheRock/issues/5678).
- **New risks**: the release channel not materializing (`ROCm/rocm-bindings` doesn't exist, RFC0012 names no bindings area, TheRock path already removed), and two binding packages disagreeing on how to find ROCm in a mixed install.

## Testing Summary

- Intra-doc anchors and repo-relative links validated, all resolve.
- Markdown fence balance checked.
- `pre-commit` run on both changed files.

## Testing Checklist

- [x] Anchor/link validation - 9 anchors resolve to headings, 9 repo links exist - Status: Passed
- [x] Markdown structure - fence balance - Status: Passed
- [x] pre-commit - `pre-commit run --files docs/rfcs/RFC00XX-Custom-Python-Bindings-Layering.md docs/rfcs/README.md` - Status: Passed

## Technical Changes

- Renames `RFC0013-Custom-Python-Bindings-Layering.md` to `RFC00XX-Custom-Python-Bindings-Layering.md` and expands it.
- Updates `docs/rfcs/README.md` to index it.

The `RFC0012` index line is also kept. That file is on `main` but was never added to the index, so this fixes an existing gap. This RFC links to it for the release channel. Happy to split that line into its own PR if preferred.


docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+author: Brian Harrison (bharriso)
+created: 2026-06-23
+modified: 2026-06-23
+status: draft
+discussion: https://github.com/ROCm/TheRock/issues/6048
+---
+
+# Custom Python Bindings Layering
+
+This RFC proposes the ownership and packaging model for ROCm library Python bindings that are not generated C API wrappers. hipDNN is the motivating example, but the model is intended to apply to any ROCm library that needs a custom Python binding layer over component-specific C, C++, or higher-level APIs.
```
ScottTodd: (commenting here before I read further, some of this may be answered already)

How about we expand on a sort of "background" section here to help clarify how different python packages and bindings projects fit together?

In particular:
* We have https://github.com/ROCm/TheRock/pull/5609 in discussion for https://github.com/ROCm/hip-python, I suppose with "generated C API wrappers" for projects like hipfft, hiprand, hipsolver, hipsparse
* We have https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md for the `rocm`, `rocm-sdk-core`, `rocm-sdk-libraries`, `rocm-sdk-device-*`, `rocm-sdk-devel` packages that include native binaries/libraries and minimal Python code to make those accessible to frameworks like PyTorch
* This RFC is for ROCm projects that want their own Python bindings (and packages?)

Looking at the layering from a few angles, some high level questions:
* Do we want these python bindings to be available via new python packages, or existing python packages? If existing packages, which (hip-python? the `rocm-` packages?)
* What development model should these python bindings have? Should individual subprojects build their own bindings and be in full control over packaging, or is this something that a superproject/packaging/release/CI systems (e.g. TheRock) or dedicated bindings project (e.g. hip-python) should be directing?
* What dependencies will exist within and between these python bindings? Could hipDNN python bindings depend on hip-python, or just the native hipDNN binaries?
  * If we did something like `pip install hipdnn-python` for the bindings, would that be self-contained, would it depend on shared objects from `rocm-sdk-libraries`, etc.?

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+
+## Summary
+
+ROCm should separate native library ownership, generated C API bindings, custom Python binding extension builds, and final Python wheel packaging.
+
+The proposed target state is:
+
+1. **Native component repositories remain native-first.** `rocm-libraries/projects/<component>` owns native libraries, headers, CMake package exports, plugins, and native tests.
+1. **TheRock / `rocm_sdk` publishes normal ROCm SDK native artifacts.** TheRock does not own custom Python binding version policy, nanobind extension matrices, or final binding wheel release metadata.
+1. **Custom binding projects live in `rocm-bindings`.** For hipDNN, `rocm-bindings/hipdnn/hipdnn-nanobind` owns the native Python extension build, and `rocm-bindings/hipdnn/hipdnn-frontend` owns the final wheel.
```
ScottTodd: https://github.com/ROCm/rocm-bindings as a new repository, or a subfolder in an existing repository?

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+| TheRock / `rocm_sdk`                             | Native ROCm build orchestration, native artifact split, ROCm SDK runtime/devel/device packages                                                    | Custom Python binding version matrix, nanobind extension artifacts, final binding wheel metadata |
+| `rocm-bindings/<component>/<component>-nanobind` | C++/nanobind extension sources, extension build, ABI-specific extension artifacts, artifact manifest, producer-side tests                         | Final user-facing wheel metadata and installed-wheel tests                                       |
+| `rocm-bindings/<component>/<component>-frontend` | Final Python wheel, import package, runtime ROCm initialization, package metadata, wheel tags, installed-wheel tests, docs, release/upload policy | Native library implementation, generated C API bindings, TheRock artifact split                  |
```
ScottTodd: Some other directory structures to consider, since "nanobind" is unique to Python and "frontend" is not, and we may also want rust and other language bindings in this repository:

* `rocm-bindings/python/<component>/[backend,frontend]`
* `rocm-bindings/<component>/python/[backend,frontend]`

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+  -> normal ROCm SDK runtime/devel artifacts containing native hipDNN outputs
+
+rocm-bindings/hipdnn/hipdnn-nanobind
+  -> Python-native extension artifacts built with find_package() against installed ROCm/hipDNN devel artifacts
+
+rocm-bindings/hipdnn/hipdnn-frontend
+  -> installable wheels that stage the matching nanobind artifact
+```
+
+TheRock does not need to know the `hipdnn-frontend` Python version matrix. It only needs to publish the normal native hipDNN runtime/devel artifacts through `rocm_sdk` or equivalent native install artifacts.
```
ScottTodd: So do you foresee TheRock adding rocm-bindings as a git submodule and building these python bindings for e.g. hipDNN as part of nightly releases, or would these python bindings have a separate release channel?

We could treat them as extensions on top of rocm, similar to rocm-examples, RDC, and several frameworks.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+
+```text
+rocm-libraries/projects/hipdnn
+  -> native hipDNN libraries, headers, CMake configs
+
+TheRock / rocm_sdk
+  -> normal ROCm SDK runtime/devel artifacts containing native hipDNN outputs
+
+rocm-bindings/hipdnn/hipdnn-nanobind
+  -> Python-native extension artifacts built with find_package() against installed ROCm/hipDNN devel artifacts
```
ScottTodd: How do you want to test that changes to the native hipDNN libraries don't break the hipDNN python bindings once they are split across repositories? Will there be enforced forward/backwards compatibility requirements and checks? Or synchronization points as part of regular releases?

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
```diff
@@ -0,0 +1,450 @@
...
+Either way, the developer document should link back to this RFC and avoid restating the full alternatives discussion.
+
+## Related documents
+
+- `docs/rfcs/README.md` — RFC process and metadata expectations.
+- `docs/rfcs/RFC0003-Build-Tree-Normalization.md` — build tree/source organization background, including forward compatibility with language binding layers.
+- `docs/rfcs/RFC0010-Test-Scripts-Migration.md` — precedent for moving ownership to the repository where code and tests co-evolve.
+- `docs/development/build_system.md` — build/runtime dependency distinction and CMake `find_package()` dependency resolution.
+- `docs/packaging/python_packaging.md` — ROCm Python packaging model, `rocm_sdk`, and framework build/runtime initialization guidance.
+- `docs/development/artifacts.md` — TheRock artifact components and runtime/devel split.
```
ScottTodd: nit: can you turn these into hyperlinks?

e.g.

```diff
- - `docs/rfcs/README.md` — RFC process and metadata expectations.
+ - [`docs/rfcs/README.md`](/docs/rfcs/README.md) — RFC process and metadata expectations.
```

(I've seen coding agents like to just reference file paths but for humans having them be clickable helps a lot :P)

ScottTodd (commented):
I had a few comments while reading but they were all addressed later in the document in some form. This looks pretty comprehensive.

Are there other projects beyond hipDNN that may want to use this proposed `rocm-bindings` repository? If so, could collect feedback from their maintainers too.

gaoikawa:
Hey @BrianHarrisonAMD,

@ScottTodd said:
> Are there other projects beyond hipDNN that may want to use this proposed rocm-bindings repository? If so, could collect feedback from their maintainers too.

I believe hipFile team is very interested in this RFC.  Please loop in @riley-dixon.

Thanks,
Gary

riley-dixon:
`hipFile` currently has experimental bindings in `hip-python`, with a PoC high-level Python API that exists as an extension within `hip-python` on top of the generated bindings. One question I have is, is it possible, or alternatively forbidden, to have a "front-end" Python API that uses both custom `nanobind` bindings and the generated `hip-python` bindings?

To briefly explain the context around the question, 1 or 2 functions of the `hipFile` API needs special error handling in the C-binding layer prior to returning to Python-land (grabbing `errno()` or `hipPeekAtLastError()`). `hip-python` does not currently provide that behaviour, and _if_ we cannot extend `hip-python` to provide that error handling, can we use `nanobind` targeted towards those specific functions while still being able to use the generated bindings for the other functions? At a high-level, there be one user-facing Python API with a dependency on the generated bindings, and a dependency on the custom bindings. How all of that gets packaged is not immediately clear to me.

Or is the answer no and we would need to provide a complete custom `nanobind` implementation?

One clarifying point is that all of this would exist in the theoretical `rocm-bindings` repo.

Right now, this is contingency planning should `hipFile` be unable to make the above addition into `hip-python`.

BrianHarrisonAMD:
> `hipFile` currently has experimental bindings in `hip-python`, with a PoC high-level Python API that exists as an extension within `hip-python` on top of the generated bindings. One question I have is, is it possible, or alternatively forbidden, to have a "front-end" Python API that uses both custom `nanobind` bindings and the generated `hip-python` bindings?
> 
> To briefly explain the context around the question, 1 or 2 functions of the `hipFile` API needs special error handling in the C-binding layer prior to returning to Python-land (grabbing `errno()` or `hipPeekAtLastError()`). `hip-python` does not currently provide that behaviour, and _if_ we cannot extend `hip-python` to provide that error handling, can we use `nanobind` targeted towards those specific functions while still being able to use the generated bindings for the other functions? At a high-level, there be one user-facing Python API with a dependency on the generated bindings, and a dependency on the custom bindings. How all of that gets packaged is not immediately clear to me.
> 
> Or is the answer no and we would need to provide a complete custom `nanobind` implementation?
> 
> One clarifying point is that all of this would exist in the theoretical `rocm-bindings` repo.
> 
> Right now, this is contingency planning should `hipFile` be unable to make the above addition into `hip-python`.

It makes sense to me that we say hip-python is the project that owns the c-api layer bindings (mostly generated), and there might be cases where it needs to be extended slightly in order to properly serve the c-api bindings that it owns. For the specifics of how that works with packaging, and project layout, I am not 100% sure. It would be best to sync with the component owner for hip-python, and to review once the details of rocm-bindings are laid out to come up with a design for how this fits.

@stellaraccident, or @amd-shiraz might be able to provide better guidance here as well. I believe there will be a follow-up describing how rocm-bindings packaging, and releases should work.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: My understanding was that rocm-binding would be a separate repository that would own the packaging, and release for any of the bindings built off the base packages from TheRock.

I think the specifics of how that would work is TBD from an upcoming RFC though.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: `rocm-bindings/<component>/python/[backend,frontend]` makes sense to me. 
I imagine you would want to keep it all grouped by component.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: My understanding of the proposed future direction was that rocm-bindings would own a separate release channel, and it would be an extension on top of TheRock artifacts.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Good question!
This came up in an earlier draft too, so it's worth addressing directly.

Once split, the dependency flow is rocm-libraries → TheRock → rocm-bindings. rocm-bindings is fundamentally a downstream consumer of the public API that rocm-libraries produces, and it builds against the release artifacts coming out of TheRock rather than against source.

Because of that, anything we'd catch at the bindings layer is really one of two things: a breaking change to the public API that slipped through, or a critical defect. In either case, finding it here points to a gap in the lower-level component testing. The bindings are too far downstream to be the right place to first catch a regression in the native libraries.

So instead of leaning on the bindings repo as the safety net, the goal is to push coverage down to the source:
rocm-libraries should carry enough public-API test coverage that a binding update can't be silently broken. The public API is the contract, so that's the surface we'd want forward/backward-compatibility discipline on, with sync happening at the regular release boundaries. Early manual testing is always an option to de-risk a specific change, but it's a backstop rather than the mechanism we'd rely on.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
domcharrier: Both make sense. HIP Python is not just a set of bindings but an interoperable ecosystem.

Food for thought: 

The HIP Python module build is CMake driven, the packaging is done by a custom Python script (to enable a superbuild).

The compilation and packaging of non-Cython bindings code, handwritten Python modules etc. could be integrated into the build process and bundled with the existing HIP Python wheels.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Added a Background section with a table separating the three surfaces (SDK wheels, generated `hip-python` bindings, custom component bindings), plus answers to your four questions:

- **New packages.** A custom frontend isn't an SDK wheel or a generated binding. Putting it in `rocm-sdk-*` recreates #5678; putting it in `hip-python` puts hand-written code inside a generated package.
- **The component team drives it**, in `rocm-bindings`. TheRock supplies native artifacts, not the Python version matrix or ABI strategy.
- **Frontend wheels depend on the ROCm SDK runtime wheels.** `hip-python` is optional, not required and not forbidden. See the new *Mixing generated and custom bindings* section.
- **SDK-dependent**, via `Requires-Dist`. No vendored ROCm shared libraries.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: New repo. I also added a Release channel section so the packaging side isn't left TBD.

RFC0012 defines the publish surface (central `whl/` and `whl-next/` indices per stream) and doesn't name a bindings area. Proposal: publish from `rocm/extras/rocm-bindings/`, feeding the central indices. `extras/` matches RFC0012's own definition of independently-released projects, and `omnistat/` is the precedent for a wheel-only entry.

Four things left open in the doc: one folder or one per component, whether RFC0012 needs an amendment, which streams, and what happens to `hip-python`'s existing PyPI channel.

Worth noting `ROCm/rocm-bindings` still 404s, so the RFC now carries a risk entry for the channel not existing yet.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Went with `rocm-bindings/<component>/python/[backend,frontend]`.

Also dropped `nanobind` from the directory names since it's an implementation detail that would show up in every component's path. `backend`/`frontend` describe the role (extension producer vs published wheel), so pybind11 or Cython components fit without renaming anything. Language-first grouping is recorded as a rejected alternative: a component's bindings across languages tend to move together more than all components' Python bindings do.

One wart flagged in the doc: hipDNN already has native libs called `hipdnn_backend` and `hipdnn_frontend`. In the layout, `python/backend` and `python/frontend` mean binding roles, not those libraries. They line up for hipDNN by coincidence.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Separate channel. TheRock does not add `rocm-bindings` as a submodule and does not build binding wheels.

This is settled rather than proposed at this point. #5429 wired hipDNN binding builds into TheRock, and #6425 removed all of it in July: packer, CI matrix entry, `share/hipdnn/python/**` artifact rules, and the `nanobind`/`robin-map` subprojects. That's now written up as a rejected alternative with the reasons it didn't hold.

Placement maps onto RFC0012 as `extras/rocm-bindings/`. Cadence is owned by the bindings repo, with the supported ROCm range in `Requires-Dist` and the exact built-against version in the artifact manifest.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Added a *Compatibility across the repository split* section.

Flow is `rocm-libraries` -> TheRock -> `rocm-bindings`, so the bindings repo consumes released artifacts. Anything it catches is either a public API break that slipped through or a native defect, and both mean component-level testing missed something. It's too far downstream to be where a native regression gets caught first.

So coverage pushes down: `rocm-libraries` owns public API test coverage, sync happens at release boundaries, and every layer declares minimum versions of its direct inputs so skew fails the build instead of failing at import. Building against a local ROCm build stays available as a backstop, not as the mechanism.

docs/rfcs/RFC0013-Custom-Python-Bindings-Layering.md
BrianHarrisonAMD: Done, plus links for RFC0008, RFC0012, #5609, #6425, #5678, and #5701.

Fair hit :P

BrianHarrisonAMD:
Pushed an update addressing the open threads. Two things for this sub-thread:

@riley-dixon yes, that works, and it's in the RFC now instead of buried in this thread. New section: *Mixing generated and custom bindings*, using hipFile as the example. A frontend can depend on both a generated binding package and a custom extension. Making you hand-write a full binding just to special-case two functions that need `errno` / `hipPeekAtLastError()` would be a bad outcome.

Preference order:

1. **Extend the generated binding.** Error capture on a fallible call is general enough that it probably belongs in `hip-python`'s hand-coded override mechanism, which already exists for this. Needs `hip-python` owner buy-in.
2. **Small custom extension next to the generated binding.** If the override path doesn't work out, add an extension under `rocm-bindings/<component>/python/backend` covering only the affected functions, with a frontend wheel that re-exports the generated bindings for everything else. The frontend wheel is the single user-facing import.
3. **Full custom binding.** Only when the user-facing API isn't really a C API wrapper. That's hipDNN.

On packaging for case 2, which you said wasn't clear: the frontend wheel gets a `Requires-Dist` on the `rocm-bindings-*` distribution alongside the ROCm runtime wheels, and its manifest records the generated-binding version it built against next to the native ROCm versions. Two versioned inputs, two declared ranges. Picking any of the three shapes doesn't change where you sit in the layering table.

@gaoikawa thanks for the pointer, hipFile is a named case in the RFC now so it should be reviewable from your side directly.

One thing to flag for planning: `ROCm/rocm-bindings` doesn't exist yet, and RFC0012 doesn't name a bindings area, but #6425 already removed the TheRock-side path. The RFC proposes a destination (`extras/rocm-bindings/`) and carries an explicit risk entry, since right now binding work is between two homes with the old one already gone.

domcharrier:
> Extend the generated binding. Error capture on a fallible call is general enough that it probably belongs in hip-python's hand-coded override mechanism, which already exists for this. Needs hip-python owner buy-in.

Small update:

* hipFILE code generation has been adjusted according to @riley-dixon's requests (details: underlying issue "clobbering of libc error status" was confirmed and could be attributed to re-acquiring of GIL after `with nogil` section).

---

IMO we should distinguish between (a) where the sources are located and (b) how we do the packaging. There are multiple options:

* The hip-python build recipes could fetch and package the component bindings for ROCm, it's a CMake process.
* But that's just one option, each component could also create their own package and we pull them into the main
   "hip-python" or "rocm-bindings" (whatever it will be called) package as required (or key-enabled optional) dependency.
* The component package's could also ship as their own packages independent of hip-python or expressing themselves a dependency to hip-python's bindings.
* ...

Related to this:

On hip-python's side, I've ensured carefully that I do not put any `__init__.py` into `site-packages/rocm`
and `site-packages/rocm/bindings` so that these installation folders are proper namespace packages according to <https://peps.python.org/pep-0420/>.

Other ROCm Python wheels can thus add their own sub-packages and sub-modules to `rocm` and `rocm/bindings` (and remove them from there) as they please. (Of course, they would have to be careful as well to not put any `__init__.py` files at the top of these directories.)

PR #6052, opened 2026-06-23
fix(ci): add container spec to AWS kube-mode runner jobs, revert JAX to Azure

kailash-khalasi:
## Summary

ISSUE ID https://github.com/ROCm/TheRock/issues/6061

Fixes a regression introduced by #6037 where jobs migrated to \`aws-linux-scale-rocm-prod\` failed because AWS runners operate in Kubernetes mode and enforce that every job has a \`container:\` block.

### Root cause

Azure self-hosted runners are VMs — jobs run directly on the host with a full shell and a GitHub tool cache available. AWS runners are Kubernetes-mode: each job runs inside an ephemeral pod and a \`container:\` image is required.

Two separate issues had to be fixed:

**1. Missing \`container:\` block**
Jobs without a container spec fail immediately with:
> \`Jobs without a job container are forbidden on this runner\`

Fixed by adding \`container: python:3.12-slim\` to the affected jobs.

**2. \`actions/setup-python\` incompatible inside a container**
After adding the container spec, \`actions/setup-python\` failed with:
> \`The version '3.12' with architecture 'x64' was not found for this operating system.\`

On Azure VM runners, \`actions/setup-python\` downloads Python into the host's tool cache (\`$RUNNER_TOOL_CACHE\`). Inside a kube-mode container, that tool cache is not mounted — so the action has nothing to work with. Since \`python:3.12-slim\` already ships Python 3.12, the \`actions/setup-python\` step is both redundant and broken inside the container. Removed it from both affected jobs.

**3. \`build_jax_wheels\` cannot run on kube-mode runners**
This job calls \`docker buildx build\` + \`docker run\` internally via \`ci_build dist_wheels\`. Kube-mode runners have no Docker daemon, so this cannot be fixed with a \`container:\` block. Reverted to \`azure-linux-scale-rocm\` until a Docker-in-Docker or native build solution is in place.

### Changes

- **\`test_artifacts_structure.yml\`** — add \`container: python:3.12-slim\`, remove \`actions/setup-python\`
- **\`multi_arch_build_tarballs.yml\`** — add \`container: python:3.12-slim\`, remove \`actions/setup-python\`
- **\`multi_arch_build_linux_jax_wheels.yml\`** — revert \`build_jax_wheels\` back to \`azure-linux-scale-rocm\`

## Test plan

- [ ] Verify \`test_artifact_structure\` (Linux + Windows) passes on AWS runner
- [ ] Verify \`build_tarballs\` passes on AWS runner
[...]

Issue #3591, opened 2026-02-24
[Feature]: Decreasing release workflow volume for JAX builds


JeniferC99:
### Suggestion Description

Concerns from  https://github.com/ROCm/TheRock/pull/3567

Raw notes are at https://github.com/ScottTodd/claude-rocm-workspace/blob/main/reviews/pr_3567.md

Each build takes 50-70 minutes, with a breakdown like this:

Phase	Time	% of Total
Docker LLVM 18 build from source	9-13 min	~18%
Bazel: jax-rocm-pjrt (cub_sort_kernel_rocm.cu.cc dominates)	21-31 min	~42%
Bazel: jaxlib	12-17 min	~23%
Bazel: jax-rocm-plugin	3-4 min	~6%
Docker setup, auditwheel, S3 upload	~5 min	~9%
We already build LLVM as part of ROCm... can rocm-jax use the same LLVM? Or install from a package / prebuilt location? Or at least build once and share across each python version build, since the compiler is independent of the python version?

The single largest bottleneck is cub_sort_kernel_rocm.cu.cc which takes 10-14 minutes alone (compiles CUB sort kernels for 9 GPU targets). I see this in the logs too:

20:56:50,335 - DEBUG - ROCm tookit path: /opt/rocm-7.12.0a20260217
20:56:50,340 - DEBUG - ROCm AMD GPU targets: gfx908,gfx90a,gfx942,gfx950,gfx1030,gfx1100,gfx1101,gfx1200,gfx1201
Is there an existing issue to properly plumb through the list of targets to build for?

Logs of log spam (to be fair, PyTorch has way more I think...)

Some concerning warnings during build/test, including some deprecations we should get ahead of

charleshofer:
> We already build LLVM as part of ROCm... can rocm-jax use the same LLVM? Or install from a package / prebuilt location? Or at least build once and share across each python version build, since the compiler is independent of the python version?

We cannot use the LLVM that ships with ROCm. However, the JAX team has been building manylinux Docker images with Python, LLVM, and ROCm as part of a separate workflow that only runs when the Dockerfile is changed. Something we could do is build a manylinux image with just Python and LLVM, and then let the ROCm image.

charleshofer:
> 20:56:50,335 - DEBUG - ROCm tookit path: /opt/rocm-7.12.0a20260217
[...]

PR #3567, opened 2026-02-23
Add support of python 3.11,3.13,3.14 for jax

JeniferC99:
## Motivation

https://amd-hub.atlassian.net/browse/AIDEVOPS-67 - Framework support for 7.12 Release
A request came to add python 3.11,3.13,3.14 support for jax

## Technical Details

Add support of python 3.11,3.13,3.14 for jax

## Test Plan

Add python 3.11,3.13,3.14 in the workflow's matrix and run a build. 
https://github.com/ROCm/TheRock/actions/runs/22324002971

## Test Result

Jax built successfully for python 3.11,3.13,3.14

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6054, opened 2026-06-23
feat: Add JAX release support v0.10.0

### LINKED ISSUE #3878: [CI] Build and test JAX Python packages as part of ci.yml

erman-gurses:
## Summary
ISSUE ID: https://github.com/ROCm/TheRock/issues/3878
Related PR: https://github.com/ROCm/rockrel/pull/65
Extend the JAX multi-arch release workflows to support both `rocm-jaxlib-v0.9.1` and `rocm-jaxlib-v0.10.0` release lines.

### Release workflow

- Add matrix support for:
  - `rocm-jaxlib-v0.9.1`
  - `rocm-jaxlib-v0.10.0`
- Add release-line specific metadata:
  - `jax_repository`
  - `build_mode`
  - `gfx_arch`

#### rocm-jaxlib-v0.9.1

- Continue using the existing native build flow.
- Build from `ROCm/rocm-jax`.
- Continue using `build/ci_build`.

#### rocm-jaxlib-v0.10.0

- Build from `ROCm/jax`.
- Build a manylinux image using the release branch Dockerfile from `ROCm/rocm-jax`.
- Build wheels inside the manylinux container using `build/build.py`.

### Testing

On rockrel: 

https://github.com/ROCm/rockrel/actions/runs/28217952215/job/83613621212
https://github.com/ROCm/rockrel/actions/runs/28224689859

- Preserve existing GPU-family based runner selection through `configure_target_run.py`.
- Add a JAX device visibility sanity check before running tests:

```bash
python -c "import jax; print(jax.local_devices())"

- Verified matrix expansion across Python 3.11–3.14.
- Verified native build path for rocm-jaxlib-v0.9.1.
- Verified manylinux build path for rocm-jaxlib-v0.10.0.
- Verified release artifact publication.
- Verified JAX device visibility check executes before tests.

TODO: 
Run JAX builds inside a top-level GitHub Actions job container, similar to existing PyTorch workflows.

### CODE REVIEW

File: .github/workflows/multi_arch_build_linux_jax_wheels.yml [hunk 1]
```diff
@@ -181,9 +199,77 @@ jobs:
...
             --therock-path="${TAR_URL}" \
-            ${SOURCE_ARG} \
             dist_wheels
 
+      - name: Build manylinux image
+        if: ${{ inputs.build_mode == 'manylinux' }}
+        working-directory: jax-source
+        env:
+          BRANCH: ${{ inputs.jax_ref }}
+          THEROCK_INDEX_URL: https://rocm.nightlies.amd.com/whl-multi-arch/
```
ScottTodd: This environment variable doesn't make sense to set together with `TAR_URL`

We should _either_ build from tarballs or from python packages. **No mixing different release package types as part of a single build process.**

When last I checked, JAX did not yet appear to follow https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#building-python-based-projects, so it could use tarballs. Looking at https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock, that is doing a system-wide pip install then symlinking the devel folder to `/opt/rocm`. That's functionally equivalent to using tarballs (well, it should be, but this also uncovered that the devel package was missing some device files, which was fixed by https://github.com/ROCm/TheRock/pull/5778)

These are steps in the right direction:
* https://github.com/ROCm/rocm-jax/pull/458
* https://github.com/jax-ml/jax/pull/38276

maybe even fully removing the dependency on tarballs and `/opt/rocm` and making it so `pip install rocm jax --index-url=...` will work?

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -181,9 +199,77 @@ jobs:
...
+          curl -fsSL \
+            -o Dockerfile.jax-manylinux_2_28-therock \
+            "https://raw.githubusercontent.com/ROCm/rocm-jax/${BRANCH}/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock"
+
+          docker build \
+            -t "${MANYLINUX_IMAGE_TAG}" \
+            --file=Dockerfile.jax-manylinux_2_28-therock \
+            --build-arg=THEROCK_INDEX_URL="${THEROCK_INDEX_URL}" \
+            --build-arg=THEROCK_VERSION="${THEROCK_VERSION}" \
+            --build-arg=GFX_ARCH="${GFX_ARCH}" \
```
ScottTodd, line 231: This `GFX_ARCH` variable is brittle. I see that this workflow is passing "all" here, but the underlying dockerfile is using this for a pip install extra without supporting a list of devices (except via an awkward `gfx1100,device-gfx1101,device-gfx1102,device-gfx1103` style with just the first omitting `device-`):

```dockerfile
RUN --mount=type=cache,mode=0755,target=/root/.cache/pip \
    /opt/python/cp312-cp312/bin/python3 -m pip install \
        --pre --index-url "${THEROCK_INDEX_URL}" \
        "rocm[libraries,devel,device-gfx950]${THEROCK_VERSION:+==}${THEROCK_VERSION}" && \
        "rocm[libraries,devel,device-${GFX_ARCH}]${THEROCK_VERSION:+==}${THEROCK_VERSION}" && 
    ln -sf /opt/python/cp312-cp312/bin/rocm-sdk /usr/local/bin/rocm-sdk && \
    rocm-sdk init
```

I commented about `GFX_ARCH` at https://github.com/ROCm/rocm-jax/pull/470.

If JAX actually needs the list of architectures (did I hear somewhere that it doesn't build its own device code?), we could use something like what PyTorch has with `PYTORCH_ROCM_ARCH`, passing the full list of gfx archs that are included in the current built:
* https://github.com/ROCm/TheRock/blob/8ae25c5e342043a76d7d7338e96385cc32065cdc/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L264-L270
* https://github.com/ROCm/TheRock/blob/8ae25c5e342043a76d7d7338e96385cc32065cdc/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L288
* https://github.com/ROCm/TheRock/blob/8ae25c5e342043a76d7d7338e96385cc32065cdc/external-builds/pytorch/build_prod_wheels.py#L1422-L1428

Note that for pytorch with gfx1250 bringup we started been filtering out GPU targets for different release branches using https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_pytorch_release_matrix.py. Might want to support the same thing for JAX at some point as needed.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
ScottTodd: This workflow is not limited to producing nightly releases. We also need to support dev and prerelease indexes where applicable.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -181,9 +199,77 @@ jobs:
...
+        working-directory: jax-source
+        env:
+          BRANCH: ${{ inputs.jax_ref }}
+          THEROCK_INDEX_URL: https://rocm.nightlies.amd.com/whl-multi-arch/
+          THEROCK_VERSION: ""
+          GFX_ARCH: ${{ inputs.gfx_arch }}
+        run: |
+          curl -fsSL \
+            -o Dockerfile.jax-manylinux_2_28-therock \
+            "https://raw.githubusercontent.com/ROCm/rocm-jax/${BRANCH}/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock"
```
ScottTodd: Isn't this workflow already using `actions/checkout` to get JAX source code, including build scripts? We shouldn't need to `curl ... https://raw.githubusercontent.com/...` to get a source file, much less a dockerfile.

I'm seeing anti-patterns with the build architecture here in a few places:
* Overuse/misuse of Docker
  * Builds outside of Docker should be supported
  * Scripts are more portable and flexible, especially for interactive use, see https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#quickstart for example
  * Looking at the logs on https://github.com/ROCm/rockrel/actions/runs/28031195089/job/82972087846 this builds an image then builds within that image? Why multiple steps in a single workflow? That's a lot of overhead
  * If Windows support is ever considered, we are unlikely to use Docker there
* Downloading files from disconnected and unpinned sources
  * https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock installs LLVM 18 (by building from source), see https://github.com/ROCm/TheRock/pull/3567#discussion_r2848983085 and https://github.com/ROCm/TheRock/issues/3591 . We may be able to reuse LLVM from the ROCm packages.
  * Installing unversioned https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip (supply chain risk if every CI/CD build will be invoking this dockerfile instead of building within an already produced dockerfile that is updated at low frequency)
  * The dockerfile base itself is similarly unversioned `FROM quay.io/pypa/manylinux_2_28_x86_64`, compared to https://github.com/ROCm/TheRock/blob/598b0c6131a612dcf5eff93549190c5e4459b7ca/dockerfiles/build_manylinux_x86_64.Dockerfile#L9-L12

We don't need to solve those and related architectural issues here, but I'd rather use `actions/checkout` instead of `curl` like this.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -181,9 +199,77 @@ jobs:
...
+          ROCM_VERSION_BUILD="${ROCM_VERSION#*+}"
+
+          if [ "${RELEASE_TYPE}" = "nightly" ]; then
+            echo "ML_WHEEL_VERSION_SUFFIX=+rocm${ROCM_VERSION_BASE}a$(date -u +%Y%m%d)" >> "$GITHUB_ENV"
+          else
+            echo "ML_WHEEL_VERSION_SUFFIX=+rocm${ROCM_VERSION_BASE}.${ROCM_VERSION_BUILD}" >> "$GITHUB_ENV"
+          fi
+        env:
+          ROCM_VERSION: ${{ inputs.rocm_version }}
+          RELEASE_TYPE: ${{ inputs.release_type }}
```
ScottTodd: I'd prefer if this followed the docs here:
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#python-package-versions
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#external-python-package-versions
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/versioning.md#working-with-python-package-versions

Specifically:
* We've been using a `+devrocm` suffix for dev releases instead of `+rocm`. That preserves the version sorting of `stable > nightly > dev` even in the local version identifier
* We have scripts (with unit tests) for such version construction, complex logic does not belong in workflow files (https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash)

You can follow what pytorch does, maybe just using the `version_suffix` output from `determine_version.py`?
* https://github.com/ROCm/TheRock/blob/440b69e17c44a3d675df01abf07defc8335c6573/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L240-L245
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/determine_version.py, note the TODO: https://github.com/ROCm/TheRock/blob/598b0c6131a612dcf5eff93549190c5e4459b7ca/build_tools/github_actions/determine_version.py#L25

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -181,9 +199,77 @@ jobs:
...
+            --volume "${GITHUB_WORKSPACE}:/workspace" \
+            --workdir /workspace/jax-source \
+            "${MANYLINUX_IMAGE_TAG}" \
+            bash -lc '
+              python build/build.py build \
+                --wheels=jax-rocm-plugin,jax-rocm-pjrt \
+                --python_version="${PYTHON_VERSION}" \
+                --bazel_startup_options=--bazelrc=build/rocm/rocm.bazelrc \
+                --bazel_options=--config=rocm_release_wheel \
+                --bazel_options=--repo_env=ROCM_PATH=$(rocm-sdk path --root) \
```
ScottTodd: Relating to my comment above, this is using _both_
* tarballs via `TAR_URL`
* python packages via `--repo_env=ROCM_PATH=$(rocm-sdk path --root)`

Do not mix different package types in a single build! That's just asking for subtle issues where files from both packages may be used at unexpected times.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
gulsumgudukbay: > This environment variable doesn't make sense to set together with `TAR_URL`
> 
> We should _either_ build from tarballs or from python packages. **No mixing different release package types as part of a single build process.**
> 
> When last I checked, JAX did not yet appear to follow https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#building-python-based-projects, so it could use tarballs. Looking at https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock, that is doing a system-wide pip install then symlinking the devel folder to `/opt/rocm`. That's functionally equivalent to using tarballs (well, it should be, but this also uncovered that the devel package was missing some device files, which was fixed by #5778)
> 
> These are steps in the right direction:
> 
> * [Strip all env vars from TheRock dockerfiles rocm-jax#458](https://github.com/ROCm/rocm-jax/pull/458)
> * [[ROCm] Bake TheRock-aware RUNPATH into wheels and fix /opt/rocm-less CI jax-ml/jax#38276](https://github.com/jax-ml/jax/pull/38276)
> 
> maybe even fully removing the dependency on tarballs and `/opt/rocm` and making it so `pip install rocm jax --index-url=...` will work?

Yes, in fact I did all the tests of https://github.com/jax-ml/jax/pull/38276 with pip installing theRock multi arch wheels and no environment variables

.github/workflows/multi_arch_build_linux_jax_wheels.yml
gulsumgudukbay: Can we fully make the JAX workflows using TheRock wheels instead of tarballs for runtime (unit testing etc)? I do not think tarballs represent a common JAX user scenario. People would typically want to use TheRock wheels with JAX just like they do for CUDA

.github/workflows/multi_arch_build_linux_jax_wheels.yml
gulsumgudukbay: And adding to this, I strongly believe wheel testing is better because it easily catches packaging issues like you mentioned @ScottTodd . The reason we caught those issues (like devel bridging or libhsa so version issues) is that we tested with wheels and no environment variable (which matches a good user experience)

gulsumgudukbay (changes requested):
It would be great if we could move to wheels instead of tarballs and environment variable testing. Unit testing with theRock wheels without any environment variables also allows us to surface packaging issues if there are any. That is how we caught previous packaging issues. This way theRock CI not only tests libraries and systems bugs, it will also test the actual packaging

.github/workflows/multi_arch_build_linux_jax_wheels.yml
tsrw2048: The formatting for the manylinux docker argument (GFX_ARCH) has been changed to to take `device-<ARCH>` to make it less brittle. I think we are still using `device-all` here but for future changes, this should be handled.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: @tsrw2048 - can you please take a look at this?

gulsumgudukbay (approved):

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Addressed this one

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: I am using `determine_version.py` currently

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Removed `TAR_URL`

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: ROCM_PACKAGE_FIND_LINKS_URL: ${{ inputs.rocm_package_find_links_url }} I use this now like PyTorch does

erman-gurses: This was addressed by separating the build paths instead of using both package sources in the same build.

1) rocm-jaxlib-v0.9.1 continues to use the legacy build/ci_build flow with --therock-path, so it remains tarball-based for compatibility with the existing rocm-jax build infrastructure.

2) rocm-jaxlib-v0.10.0 uses the newer manylinux flow, where ROCm is installed from Python packages via the configured package index and discovered through rocm-sdk path --root.

As a result, tarballs and Python packages are no longer used together within the same build. The two mechanisms are isolated behind separate build_mode paths (native vs manylinux), so each build uses a single package source.

This also moves the newer path closer to the long-term goal of validating the actual ROCm packaging and wheel installation flow rather than relying on extracted tarballs.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
ScottTodd: Our new AWS build runners also don't support this style of Docker usage at the moment, see failures from yesterday: https://github.com/ROCm/rockrel/actions/runs/28049076051/job/83034722127#step:9:23 `ERROR: failed to connect to the docker API at unix:///var/run/docker.sock; check if the path is correct and if the daemon is running: dial unix /var/run/docker.sock: connect: no such file or directory`

We switched the JAX builds back to the older Azure runners in https://github.com/ROCm/TheRock/pull/6052 and it sounds like @kailash-khalasi and others are working to add back support for this style of Docker to the newer runners.

We'd prefer for the build workflow to use a top level container and then just run steps (scripts etc.) all within that one container. That's more compatible with other scripts which interface with build inputs/outputs as well as local interactive development.

ScottTodd: (See https://github.com/ROCm/TheRock/issues/6090 for the `Docker daemon unavailable` error)

tsrw2048: I'm not sure if it is doable to switch the Docker based workflow here since JAX wheels are built in a specific environment (the custom manylinux image) to ensure that the resulting wheels are compatible in many different settings. This is the same environment that we use for testing functionality and performance on our side so changing the environment runs the risk of a divergence between what we test and what gets distributed through TheRock.

But it seems like the runners were switched back to the ones that supported Docker temporarily and there is ongoing work for the new ones? Am I correct in understanding that this should resolve the issue for now?

ScottTodd: > I'm not sure if it is doable to switch the Docker based workflow here since JAX wheels are built in a specific environment (the custom manylinux image) to ensure that the resulting wheels are compatible in many different settings.

We do the same for all of our packages. JAX is the only component right now that builds its docker image _during the workflow itself_. 
* https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md#manylinux-x86-64
* https://github.com/ROCm/TheRock/tree/main/dockerfiles#build_manylinux_dockerfile
* ROCm example: https://github.com/ROCm/TheRock/blob/649a622dfd30917f80306aa2d8fc0af454f23113/.github/workflows/build_portable_linux_python_packages.yml#L99-L107 https://github.com/ROCm/TheRock/blob/649a622dfd30917f80306aa2d8fc0af454f23113/.github/workflows/build_portable_linux_python_packages.yml#L153-L160
* PyTorch example: https://github.com/ROCm/TheRock/blob/649a622dfd30917f80306aa2d8fc0af454f23113/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L146-L154 https://github.com/ROCm/TheRock/blob/649a622dfd30917f80306aa2d8fc0af454f23113/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L272-L289

> But it seems like the runners were switched back to the ones that supported Docker temporarily and there is ongoing work for the new ones? Am I correct in understanding that this should resolve the issue for now?

Yes it's resolved for now.

erman-gurses: So then we have a new TODO item as _Run JAX builds inside a top-level GitHub Actions job container, similar to existing PyTorch workflow._

erman-gurses:
@ScottTodd , Is there any blocker for now from your side? If not. good to merge?

ScottTodd:
I can do another review today. 15 other PRs in the queue.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -103,6 +116,19 @@ on:
...
         type: string
         default: ""
+      jax_repository:
+        description: "Repository containing the JAX release branch."
+        type: string
+        required: true
+      build_mode:
+        description: "Build mode: native or manylinux."
+        type: string
+        required: true
```
ScottTodd: nit: If there are only two expected string values, for workflow_dispatch you could use a "choice" input type to make this easier to use correctly.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -103,6 +116,19 @@ on:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
         default: ""
+      jax_repository:
+        description: "Repository containing the JAX release branch."
+        type: string
+        required: true
```
ScottTodd: Can you add a default here, or at least mention which repositories are expected in the description? What syntax is this expecting, `ROCm/jax` or `https://github.com/ROCm/jax`?

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -103,6 +116,19 @@ on:
...
+      build_mode:
+        description: "Build mode: native or manylinux."
+        type: string
+        required: true
+      gfx_arch:
+        description: "GFX architecture used for the manylinux image build."
+        type: string
+        required: false
+        default: ""
```
ScottTodd: Add an example to the description? We have a few syntax styles (`gfx94X`, `gfx942`, `gfx942-dcgpu`, etc.) and that helps a lot when working across multiple workflows.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -103,6 +116,19 @@ on:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
         default: ""
```
ScottTodd, line 134: These `repository` and `ref` inputs aren't commonly provided when triggered by developers, so I'd prefer to keep them as the last inputs in the list. Looks like a few other workflows have similarly started adding inputs beyond them though.

I would move `jax_repository`, `build_mode`, and `gfx_arch` near the top of the inputs list since they will be set frequently.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -181,9 +207,63 @@ jobs:
...
+            --rocm-version "${{ inputs.rocm_version }}"
+
+      - name: Build JAX Wheels in manylinux container
+        if: ${{ inputs.build_mode == 'manylinux' }}
+        working-directory: jax-source
+        env:
+          ROCM_VERSION: ${{ inputs.rocm_version }}
+          PYTHON_VERSION: ${{ inputs.python_version }}
+        run: |
+          docker run --rm \
```
ScottTodd, line 247: Can you also update the docs at https://github.com/ROCm/TheRock/tree/main/external-builds/jax#build-instructions ? (The general progression for development work and feature enablement should be local builds --> dev builds --> release builds, this jumps straight to release builds in github actions workflows which is the hardest to debug)

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -24,6 +23,10 @@ on:
         description: "ROCm package version to build against."
         type: string
         required: true
+      rocm_package_find_links_url:
+        description: "ROCm package index / find-links URL for the JAX manylinux build."
+        type: string
+        default: "https://rocm.devreleases.amd.com/whl-multi-arch/"
```
ScottTodd: This URL is not compatible with `--find-links`, it is only compatible with `--index-url` or `--extra-index-url`. I'd leave this default off, as we have in the pytorch workflows: https://github.com/ROCm/TheRock/blob/1371136d458b954b766618014c09d5cd72b27e40/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L115-L118

A `--find-links` URL would be more like https://therock-dev-artifacts.s3.amazonaws.com/27944292460-linux/python/index.html

.github/workflows/multi_arch_release_linux_jax_wheels.yml
```diff
@@ -52,6 +55,10 @@ on:
...
```
ScottTodd: Remove default that won't work here too

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -220,6 +207,18 @@ jobs:
               --jax-requirements rocm-jax/build/requirements.txt
           fi
 
+      - name: Install ROCm packages
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --pre \
+            --index-url "${ROCM_PACKAGE_FIND_LINKS_URL}" \
```
ScottTodd: Wait. Which URL format do you want? They are different. The variable name here says "find links url" but it's installing with `--index-url`?

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -220,6 +207,18 @@ jobs:
...
+            rocm \
+            rocm-sdk-devel \
+            rocm-sdk-device-gfx942
```
ScottTodd: `test_amdgpu_family` is an input but this assumes that gfx942 is all that needs to be installed?

Use this code instead:
* ROCm:
  * https://github.com/ROCm/TheRock/blob/1371136d458b954b766618014c09d5cd72b27e40/.github/workflows/test_rocm_wheels.yml#L147-L176
* torch:
  * https://github.com/ROCm/TheRock/blob/1371136d458b954b766618014c09d5cd72b27e40/.github/workflows/test_pytorch_wheels.yml#L148-L157
  * https://github.com/ROCm/TheRock/blob/1371136d458b954b766618014c09d5cd72b27e40/.github/workflows/test_pytorch_wheels.yml#L198-L212

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -220,6 +207,18 @@ jobs:
               --jax-requirements rocm-jax/build/requirements.txt
           fi
 
+      - name: Install ROCm packages
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --pre \
+            --index-url "${ROCM_PACKAGE_FIND_LINKS_URL}" \
+            rocm \
+            rocm-sdk-devel \
```
ScottTodd: Do JAX tests need `rocm-sdk-devel`? They shouldn't. Devel packages are for _building_, and most users of frameworks are not expected to install them.

See my other comment, this should just install `rocm[libraries,device-gfxXXXX]`, or maybe something like `jax[device-gfxXXXX]` in the future if we want to go the same route as we did for torch extras.

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -220,6 +207,18 @@ jobs:
...
+            rocm-sdk-device-gfx942
+          rocm-sdk init
+          echo "ROCM_ROOT=$(rocm-sdk path --root)" >> "$GITHUB_ENV"
+          ln -sfn "$(rocm-sdk path --root)" /opt/rocm
```
ScottTodd: Please don't symlink rocm python packages to `/opt/rocm/`.
* Framework should not be dependent on the devel package at runtime (e.g. in tests)
* Python packages exist in python environments, they should not leak outside to the full system like this

external-builds/jax/README.md
gulsumgudukbay: why do we keep something that is not used?

external-builds/jax/README.md
```diff
@@ -132,9 +161,12 @@ provide it via **tarballs** with arbitrary install locations.
    rocm-jax/jax_rocm_plugin/wheelhouse/
```
gulsumgudukbay, line 163: this is the case only for 0.9.1

external-builds/jax/README.md
```diff
@@ -168,24 +200,25 @@ For more detailed build options (including building `jax_rocm7_plugin` and
...
    ```bash
-   # Install jaxlib, jax_rocm7_plugin, and jax_rocm7_pjrt from the GPU-family index
-   pip install --index-url "https://rocm.nightlies.amd.com/v2/<amdgpu_family>/" \
-     jaxlib jax_rocm7_plugin jax_rocm7_pjrt
+   pip install \
+   --index-url <package_index_url> \
+   jax_rocm7_plugin \
+   jax_rocm7_pjrt
+
+   pip install \
```
gulsumgudukbay: Remove lines 219 to 221. 
the `pip install jax==<JAX_VERSION> ` will pull jaxlib as well as a dependency anyways.

gulsumgudukbay (changes requested):

external-builds/jax/README.md
erman-gurses:
```
1. Locate built wheels:

   After a successful build, wheels will be available in:


   rocm-jax/jax_rocm_plugin/wheelhouse/

```

 Here is the full part -  do you want me delete all of it?

external-builds/jax/README.md
erman-gurses: Is this obsolete? Let me know what do I need to add?

.github/workflows/test_linux_jax_wheels_partial.yml
erman-gurses: Removed

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Updated the doc -  reviewed by @gulsumgudukbay -  finalizing that

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Solved

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Added

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Added

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Done

.github/workflows/test_linux_jax_wheels_partial.yml
erman-gurses: I updated the test workflow to install only the ROCm runtime packages (rocm[libraries] plus the required device extras) instead of rocm-sdk-devel. The workflow no longer needs the development package or a system-wide /opt/rocm symlink.

.github/workflows/test_linux_jax_wheels_partial.yml
erman-gurses: Addressed. The workflow now expands test_amdgpu_family into the appropriate ROCm device extras using expand_amdgpu_families.py and installs rocm[libraries,<device-extras>]. This removes the previous hardcoded gfx942 dependency and supports the selected test GPU family.

.github/workflows/multi_arch_release_linux_jax_wheels.yml
erman-gurses: Addressed. The default has been removed. The ROCm package index is now provided by the caller.

.github/workflows/test_linux_jax_wheels_partial.yml
erman-gurses: Right — JAX uses --index-url here, so I renamed the input to `rocm_package_index_url` and removed the default. I also added `package_index_url` to `upload_python_packages.py` output plumbing so JAX can consume the index URL directly while keeping `package_find_links_url` for existing consumers.

external-builds/jax/README.md
gulsumgudukbay: PYTHON_VERSION=<python versions, comma separated>
ROCM_VERSION=<rocm_version>

python3 build/ci_build --therock-path "<path_to_tarball_or_rocm_dir>" \
  --python-versions="$PYTHON_VERSION" \
  --rocm-version="$ROCM_VERSION" \
  dist_wheels

external-builds/jax/README.md
gulsumgudukbay: > ```
> 1. Locate built wheels:
> 
>    After a successful build, wheels will be available in:
> 
> 
>    rocm-jax/jax_rocm_plugin/wheelhouse/
> ```
> 
> Here is the full part - do you want me delete all of it?

the wheels will be in    rocm-jax/jax_rocm_plugin/wheelhouse/ for the 0.9.1 case. For 0.10.0 case, since the repo root is not rocm-jax anymore, it will be in a different location (jax/dist is the default, where jax is the repo root)

erman-gurses: True -  updated - better?

external-builds/jax/README.md
```diff
@@ -61,60 +61,79 @@ This repository builds the following ROCm-enabled JAX artifacts:
...
+- **ROCm**:
+  - JAX 0.9.1: TheRock tarball
+  - JAX 0.10.0: ROCm Python packages from the TheRock multi-arch package index
 
 ### Steps
 
-1. Checkout rocm-jax and jax:
+1. Checkout the source repository for your JAX version.
+
+   **JAX 0.9.0**
```
gulsumgudukbay:
```suggestion
   **JAX 0.9.1**
```

external-builds/jax/README.md
```diff
@@ -168,24 +210,21 @@ For more detailed build options (including building `jax_rocm7_plugin` and
    pip install -r external-builds/jax/requirements-jax.txt
```
gulsumgudukbay: where does this path come from (i.e. external-builds is not coming from our side, jax repo)?

erman-gurses: It existed before

gulsumgudukbay: no it never existed in JAX. IDK if it ever existed in DevOps side. 
Should be as follows - this is what QA uses:
```
cd jax
pip install -r build/test-requirements.txt
pip install pytest-html pytest-csv uv pytest-json-report

```

erman-gurses: It existed in TheRock.

So used it in here

https://github.com/ROCm/TheRock/blob/9c8b5bf0cd71c7f8dd8cc0610609db5c12512562/.github/workflows/multi_arch_build_linux_jax_wheels.yml#L187

```
- name: Install python deps for CI
  run: |
    pip install -r external-builds/jax/requirements-jax.txt
```

erman-gurses: Updated

gulsumgudukbay (approved):

tsrw2048 (approved):

erman-gurses:
Latest Test: https://github.com/ROCm/rockrel/actions/runs/28196469890 (In Progress)

ScottTodd:
> Latest Test: https://github.com/ROCm/rockrel/actions/runs/28196469890 (In Progress)

You can also test dev releases here in TheRock: https://github.com/ROCm/TheRock/actions/workflows/multi_arch_release.yml . We'd prefer that - access to rockrel is intentionally limited as workflow runs there have the potential to push to nightly/prerelease buckets.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -155,21 +184,26 @@ jobs:
...
+          pip install -r external-builds/jax/requirements-jax.txt
 
-      - name: Determine build arguments
-        env:
-          BUILD_JAXLIB: ${{ inputs.build_jaxlib }}
+      - name: Debug Index URL Propagation
+        run: |
+          echo "=== DEBUG: rocm_package_index_url Propagation ==="
+          echo "inputs.rocm_package_index_url = '${{ inputs.rocm_package_index_url }}'"
+          echo "=================================================="
```
ScottTodd: Remove debug code? inputs are already visible in logs

https://github.com/ROCm/rockrel/actions/runs/28078871973/job/83128967499#step:1:21

Image

ScottTodd (approved):
Thanks, looking good now :)

.github/workflows/multi_arch_build_linux_jax_wheels.yml
erman-gurses: Just removed

erman-gurses:
Latest Tests are green:
https://github.com/ROCm/rockrel/actions/runs/28217952215/job/83613621212
https://github.com/ROCm/rockrel/actions/runs/28224689859

marbre (commented):
I think your changes can go in @erman-gurses but make sure to land this coordinated with your PR to rockrel (whcih still pins your own branch!).

erman-gurses:
> I think your changes can go in @erman-gurses but make sure to land this coordinated with your PR to rockrel (whcih still pins your own branch!).

I updated already @marbre, please see the PR: https://github.com/ROCm/rockrel/pull/65

PR #6067, opened 2026-06-23
fix(ci): Increase OPENBLAS_NUM_THREADS and OMP_NUM_THREADS to 14

geomin12:
Update thread limit env vars to 14


tony-davis (approved):
Based on meeting over CI contention. Thanks!

PR #5375, opened 2026-05-20
Disable iree / fusilli build

AaronStGeorge:
## Motivation

As fusilli will no longer be maintained, at a minimum the build should be disabled. Maybe it should be fully removed? I haven't got the green light yet but I'd vote yes.

## Technical Details

Disables fusilli build in monolithic and multi-arch pipelines

## Test Plan

CI (both multi-arch and not)

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6068, opened 2026-06-23
Add flag for enabling Hip Kernel Provider and default off

adickin-amd:
## Motivation

We dont want to have hip kernel provider in the upcoming release.

## Risk
 Minimal- this change should just disable the engine in the release.

### DIFF
--- a/FLAGS.cmake
+++ b/FLAGS.cmake
@@ -32,6 +32,16 @@ therock_declare_flag(
 )
 
 therock_declare_flag(
+  NAME THEROCK_ENABLE_HIPKERNELPROVIDER
+  DEFAULT_VALUE OFF
+  DESCRIPTION "Enable the Hip Kernel Provider engine"
+  CMAKE_VARS
+    THEROCK_ENABLE_HIPKERNELPROVIDER=OFF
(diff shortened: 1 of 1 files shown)


FLAGS.cmake
```diff
@@ -31,6 +31,16 @@ therock_declare_flag(
...
+therock_declare_flag(
+  NAME THEROCK_ENABLE_HIPKERNELPROVIDER
+  DEFAULT_VALUE OFF
+  DESCRIPTION "Enable the Hip Kernel Provider engine"
+  CMAKE_VARS
+    THEROCK_ENABLE_HIPKERNELPROVIDER=OFF
+  SUB_PROJECTS
+    hipDNN
+)
+
```
BrianHarrisonAMD, line 43: I think since this is a project added by TheRock feature, we can disable it like this instead:

https://github.com/ROCm/TheRock/pull/5375/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL251-R251

The other flags look to be related to project settings, and this is enabling a project in TheRock.

PR #5141, opened 2026-05-08
Improve clang ccache hit rates on Windows

ScottTodd:
## Motivation

Tentatively fixes https://github.com/ROCm/TheRock/issues/4519.

This makes ccache functional for `clang` compile commands on our Windows CI runners that use `B:\` volume mounts. Previously even with a "warm" cache we would get cache misses for all clang commands resulting in 4+ hour build times. Now we can get ~95% cache hits for builds under 1 hour.

## Technical Details

Detailed analysis: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/windows-ccache-hitrate.md.

* https://github.com/ROCm/TheRock/pull/4419 enabled reproducible build options, producing byte-identical amd-llvm compiler binaries
* Our Windows CI runners use `B:\` volume mounts that resolve to GUID-based workspace paths like `C:\B109CE3D-03D2-40EB-AD46-20E30992D028\build\`
* Setting ccache's `base_dir` was not sufficient to normalize this
* The GUID paths are embedded in CMake compile commands like `-DHIP_COMPILER_FLAGS` and by clang for resource directory headers

Two changes were needed to fully get cache hits (without editing the runner configuration itself):

1. ~~Stabilize filesystem paths used for the source checkout and entry point commands using [`subst`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/subst) to "associate a path with a drive letter"~~ (reverted to keep workflows simpler strictly required)
2. Explicitly set clang's `-resource-dir` to the stable (unresolved) path to avoid LLVM's call to [`GetModuleFileNameW`](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew) in [`llvm/lib/Support/Windows/Path.inc`](https://github.com/llvm/llvm-project/blob/7377bac59b4aea64da09873b44df1430571e93c3/llvm/lib/Support/Windows/Path.inc#L207-L212))

## Test Plan

Multiple rounds of Linux and Windows CI on this PR watching ccache statistics.

## Test Result

Pre-testing experiment results: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/windows-ccache-hitrate.md#experiment-results

Baseline (https://github.com/ROCm/TheRock/actions/runs/25588131640):

Run description | Build time | Cache hitrate | Logs link 
-- | -- | -- | --
[...]

PR #6071, opened 2026-06-23
Fix print/log function usage in setup_ccache.py

ScottTodd:
## Motivation

This should fix crashes that occur on these error handling paths since the `_log` helper function does not accept a `file` argument.

## Technical Details

Back on https://github.com/ROCm/TheRock/pull/5141, I added additional logging as part of debugging https://github.com/ROCm/TheRock/issues/4519. For the logging to not interfere with the suggested `eval "$(./build_tools/setup_ccache.py)"` usage of this script it was sent to `stderr` instead of `stdout` via this helper function.

## Test Plan

* CI should still configure and use ccache
  * ... on Windows. On Linux we're using new runners on AWS that don't have access to the ccache bucket
* Error paths are not exercised right now, either on CI or via unit tests, but I think this is safe enough to trust without that

## Submission Checklist

### DIFF
--- a/build_tools/setup_ccache.py
+++ b/build_tools/setup_ccache.py
@@ -165,15 +165,13 @@ def run(args: argparse.Namespace):
         if config_file.read_text() != config_contents:
             _log(
                 f"NOTE: {config_file} does not match expected. Run with --init to regenerate",
-                file=sys.stderr,
             )
         if not IS_WINDOWS and (
             not compiler_check_file.exists()
             or compiler_check_file.read_text() != POSIX_COMPILER_CHECK_SCRIPT
         ):
-            print(
+            _log(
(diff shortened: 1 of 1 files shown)


build_tools/setup_ccache.py
amd-nicknick, line 180: Minor cosmetic: This can be simplified to `subprcess.run(check=True`

amd-nicknick (approved):
Fix LGTM. CI build pass 🎉

build_tools/setup_ccache.py
ScottTodd: Outside the scope of this PR IMO.

amd-nicknick: Yep, no concern :)

PR #6076, opened 2026-06-23
feat: Add tarballs excluding tests

marbre:
## Motivation

This teaches the stage-aware artifact fetch path to exclude selected
artifact components and whole artifacts. This is used to generate
tarballs which do not contain test artifacts and fftw3. Tarballs with
tests are generated in addition to those.

JIRA ID: ROCM-26486

## Technical Details

The script now creates test-less tarballs by default, whereas the reusable
workflow enables test tarballs by default in release automation.

Right now this excludes `_test` artifacts but some test files seem to
come in via other artifacts which needs to be addressed in a follow
up.

The naming explicitly not follow the draft RFC #4438, which was not
landed yet, after an internal discussion.

## Test Plan

* https://github.com/ROCm/TheRock/actions/runs/28102279885

## Test Result

* https://therock-dev-artifacts.s3.amazonaws.com/28102279885-linux/tarballs/index.html

## Submission Checklist

### DIFF
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -136,26 +136,67 @@ def parse_target_families(args: argparse.Namespace) -> List[str]:
 
 
 def find_available_artifacts(
-    artifact_names: Set[str],
-    target_families: List[str],
-    available: Set[str],
-) -> List[str]:
+    artifact_names: set[str],
+    target_families: list[str],
+    available: set[str],
+    excluded_components: set[str] | None = None,
+) -> list[str]:
     """Find which artifacts exist in the available set.
 
     Iterates artifact_names × target_families × components × extensions,
     returning filenames that are present in `available`. Prefers .tar.zst
     over .tar.xz when both exist.
     """
+    excluded_components = excluded_components or set()
     matched = []
     for artifact_name in sorted(artifact_names):
         for tf in target_families:
             for comp in ARTIFACT_COMPONENTS:
+                if comp in excluded_components:
+                    continue
                 for ext in ARTIFACT_EXTENSIONS:
                     filename = f"{artifact_name}_{comp}_{tf}{ext}"
                     if filename in available:
                         matched.append(filename)
                         break  # Found this artifact, don't check other extensions
     return matched
+
+
+def parse_excluded_components(raw_components: str) -> set[str]:
+    """Parse and validate component names passed to --exclude-components."""
+    components = {
+        comp.strip()
+        for comp in raw_components.replace(";", ",").split(",")
+        if comp.strip()
+    }
+    invalid_components = components - set(ARTIFACT_COMPONENTS)
+    if invalid_components:
+        raise ValueError(
+            "Invalid artifact component(s): "
+            f"{', '.join(sorted(invalid_components))}. "
+            f"Valid components are: {', '.join(ARTIFACT_COMPONENTS)}"
+        )
+    return components
+
+
+def parse_excluded_artifacts(
+    raw_artifacts: str,
+    valid_artifacts: set[str],
+) -> set[str]:
+    """Parse and validate artifact names passed to --exclude-artifacts."""
+    artifacts = {
+        artifact.strip()
+        for artifact in raw_artifacts.replace(";", ",").split(",")
+        if artifact.strip()
+    }
+    invalid_artifacts = artifacts - valid_artifacts
+    if invalid_artifacts:
+        raise ValueError(
+            "Invalid artifact name(s): "
+            f"{', '.join(sorted(invalid_artifacts))}. "
+            f"Valid artifacts are: {', '.join(sorted(valid_artifacts))}"
+        )
+    return artifacts
 
 
 # =============================================================================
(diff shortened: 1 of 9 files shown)


build_tools/install_rocm_from_artifacts.py
```diff
@@ -166,7 +166,10 @@ def extract_version_from_asset_name(
     prefix = f"therock-dist-{platform_str}-{artifact_group}-"
     suffix = ".tar.gz"
     if asset_name.startswith(prefix) and asset_name.endswith(suffix):
-        return asset_name[len(prefix) : -len(suffix)]
+        version = asset_name[len(prefix) : -len(suffix)]
+        if version.startswith("tests-"):
+            return None
+        return version
```
ScottTodd, line 172: This could use a comment showing an example "version" that starts with "tests-" and why that should return `None` instead of the `version` (should it just be parsed differently?)

build_tools/install_rocm_from_artifacts.py
```diff
@@ -182,6 +185,8 @@ def list_available_nightly_gpu_families(platform_str: str = PLATFORM) -> set[str
 
     for page in paginator.paginate(Bucket=NIGHTLY_BUCKET_NAME, Prefix=prefix):
         for obj in page.get("Contents", []):
+            if "-tests-" in obj["Key"]:
+                continue
```
ScottTodd, line 189: Why always skip objects with `-tests` in their names? I wonder if we'd want to preserve the existing behavior here where tarballs include tests and let developers choose which tarball variant they want, instead of just deciding for them here.

(Related to this, we outgrew the `install_rocm_from_artifacts.py` script a loooong time ago IMO. It's serving both CI _and_ various workflows and user scenarios when it doesn't have the feature set or robustness needed to effectively serve either)

build_tools/build_tarballs.py
```diff
@@ -55,12 +59,15 @@
 import sys
 from pathlib import Path
 
+DEFAULT_EXCLUDED_ARTIFACTS: list[str] = ["fftw3"]
+DEFAULT_EXCLUDED_COMPONENTS: list[str] = ["test"]
```
ScottTodd, line 63: Can you add a comment here on why fftw3 is an excluded artifact? It's only used by tests?

Then, not necessarily blocking here, but I think we could structure this more clearly by listing what tarballs we want to produce and having the code map directly to those:
* no tests (unqualified default file name)
* including tests
* potential others? lib only? doc only? one with just "core runtime"?

For example:
```python
STANDARD_TARBALL_ARTIFACTS = {
  "include": ["*"],
  "exclude": ["fftw3"],
}
STANDARD_TARBALL_COMPONENTS = {
  "include": ["*"],
  "exclude": ["test"],
}

WITH_TESTS_TARBALL_ARTIFACTS = {
  "include": ["*"],
  "exclude": [],
}
WITH_TESTS_TARBALL_COMPONENTS = {
  "include": ["*"],
  "exclude": [],
}
```

build_tools/artifact_manager.py
```diff
@@ -334,6 +375,16 @@ def do_fetch(args: argparse.Namespace):
         )
 
     target_families = parse_target_families(args)
+    excluded_artifacts = parse_excluded_artifacts(
+        args.exclude_artifacts,
+        set(topology.artifacts.keys()),
+    )
```
ScottTodd, line 381: nit: this would be more clear with named args
```suggestion
    excluded_artifacts = parse_excluded_artifacts(
        raw_artifacts=args.exclude_artifacts,
        valid_artifacts=set(topology.artifacts.keys()),
    )
```

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#use-named-arguments-for-complicated-function-signatures

(Swapping the argument ordering would be an easy mistake to make as-is)

build_tools/tests/install_rocm_from_artifacts_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+class TestReleaseDiscovery(unittest.TestCase):
+    def test_extract_version_ignores_test_tarball(self) -> None:
+        self.assertIsNone(
+            mod.extract_version_from_asset_name(
+                "therock-dist-linux-gfx94X-dcgpu-tests-7.13.0.tar.gz",
+                "gfx94X-dcgpu",
+                "linux",
+            )
+        )
```
ScottTodd, line 27: I see that this is just adding unit tests for the changed behavior. That's fine, but we could also add a `TODO: backfill test coverage for other behavior` comment to this file to highlight that the coverage here is very sparse.

build_tools/build_tarballs.py
ScottTodd, line 246: There's been a bug in here that we could fix as part of this PR or another PR.

If we build for a single family (e.g. as part of a dev release), this `len(families) > 1` check skips building the "multiarch" tarball. I think we should still build it so workflows like JAX that depend on there being that "includes everything that was built" tarball is still produced.

ScottTodd (approved):

marbre:
We talked offline and agreed to address in follow ups.

build_tools/build_tarballs.py
ScottTodd: I'm testing a fix for this, fyi: https://github.com/ROCm/TheRock/tree/users/scotttodd/build-tarball-single-family

ScottTodd:
The build tarballs workflow job in nightly releases now takes ~1 hour instead of ~30 minutes:
* before: https://github.com/ROCm/rockrel/actions/runs/28066115163/job/83104468190
* after: https://github.com/ROCm/rockrel/actions/runs/28411570730/job/84198547877

That's expected but still rather slow, especially for just network, file IO, and decompression/compression. Would be worth optimizing or sharding across multiple runners if we're close to full utilization on the runners already (need to verify).

PR #6079, opened 2026-06-23
fix(ci): Add BUILD_VARIANT env var for ASAN test skipping

geomin12:
The is_asan() function was checking ARTIFACT_GROUP which doesn't contain "asan" for ASAN builds. Add BUILD_VARIANT input/env var through the workflow chain and update is_asan() to check it.

Fixes test skipping for ASAN builds (hipcc, rocminfo tests).

### DIFF
--- a/build_tools/github_actions/github_actions_api.py
+++ b/build_tools/github_actions/github_actions_api.py
@@ -773,6 +773,6 @@ def get_first_gpu_architecture(env=None, therock_bin_dir: str | None = None) -> 
 
 
 def is_asan():
-    """Using artifact_group, determines if this is an asan build"""
-    ARTIFACT_GROUP = os.getenv("ARTIFACT_GROUP", "")
-    return "asan" in ARTIFACT_GROUP
+    """Determines if this is an ASAN build using BUILD_VARIANT env var."""
+    BUILD_VARIANT = os.getenv("BUILD_VARIANT", "")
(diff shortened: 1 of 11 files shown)


nunnikri (approved):
LGTM

build_tools/github_actions/github_actions_api.py
```diff
@@ -773,6 +773,6 @@ def get_first_gpu_architecture(env=None, therock_bin_dir: str | None = None) ->
 
 
 def is_asan():
-    """Using artifact_group, determines if this is an asan build"""
-    ARTIFACT_GROUP = os.getenv("ARTIFACT_GROUP", "")
-    return "asan" in ARTIFACT_GROUP
+    """Determines if this is an ASAN build using BUILD_VARIANT env var."""
+    BUILD_VARIANT = os.getenv("BUILD_VARIANT", "")
+    return BUILD_VARIANT == "asan"
```
ScottTodd: This doesn't belong in this file, as this environment variable is not something provided by GitHub Actions but rather something custom for our own workflows/scripts. Fine to make this fix though.

(code like this is why I added myself to CODEOWNERS here)

ScottTodd (approved):

build_tools/github_actions/github_actions_api.py
geomin12: oh fair, i can move it elsewhere

PR #6086, opened 2026-06-24
fix(ci): correct multi-arch index URL in PyTorch test and upload summaries

ethanwee1:
## Summary

Fixes the PyTorch test workflow step summary so multi-arch installs show the correct flat `whl-multi-arch/` index URL instead of incorrectly appending an AMDGPU family subdir (e.g. `gfx94X-dcgpu`).

This was a **summary/repro-link bug only**. The actual pip install path in `test_pytorch_wheels.yml` was already correct when `multi_arch=true`; the step summary and copy/paste repro instructions were misleading people into using the wrong index URL.

## What was going wrong

Multi-arch wheels are published to a **flat** index (no family subdir):

- Dev builds: `https://rocm.devreleases.amd.com/whl-multi-arch/`
- Nightly: `https://rocm.nightlies.amd.com/whl-multi-arch/`

GPU selection uses pip **device extras**, not a path suffix:

```bash
pip install --index-url=https://rocm.devreleases.amd.com/whl-multi-arch/ \
  'torch[device-gfx942]==2.11.0+rocm7.14.0a20260623'
```

But `summarize_test_pytorch_workflow.py` always appended `--index-subdir` (e.g. `gfx94X-dcgpu`) to the index URL, and `test_pytorch_wheels.yml` always passed that subdir — even in multi-arch mode. The step summary then showed:

```bash
# Wrong repro link shown in summary
pip install \
  --index-url=https://rocm.devreleases.amd.com/whl-multi-arch/gfx94X-dcgpu/ \
  "torch[device-gfx942]==2.11.0+rocm7.14.0a20260623"
```

That URL does not exist for multi-arch packages. The correct install (what pip actually ran in multi-arch jobs) was:

```bash
# Correct multi-arch install
pip install \
  --index-url=https://rocm.devreleases.amd.com/whl-multi-arch/ \
  'torch[device-gfx942]==2.11.0+rocm7.14.0a20260623'
```

Separately, manually dispatched **Test PyTorch Full** runs were also failing when dispatched with the default `multi_arch=false`, which caused pip to use per-family index layout on a multi-arch URL:

```bash
# Wrong manual dispatch (multi_arch=false, wrong index bucket)
pip install --index-url=https://rocm.nightlies.amd.com/whl-multi-arch/gfx94X-dcgpu \
  torch==2.11.0+rocm7.14.0a20260622
```

That is a dispatch-input issue, not fixed by this PR — but the misleading summary made it harder to spot.

## What changed

1. **`summarize_test_pytorch_workflow.py`**: only append `index_subdir` when `device_extras` is not set (per-family mode).
2. **`test_pytorch_wheels.yml`**: pass `--index-subdir` only when `multi_arch != true`; pass `--device-extras` in multi-arch mode instead.

## Related runs (for context)

- Multi-arch **build** (dev, `release/2.11_gfx1250`, wheels published correctly): https://github.com/ROCm/TheRock/actions/runs/28056964518
- **Smoke test** from that build (multi_arch install worked; summary link was wrong): see "Summarize workflow inputs" step in the build run above
- **Failed manual Test PyTorch Full** (wrong dispatch inputs: `multi_arch=false`, nightlies index + gfx94X subdir): https://github.com/ROCm/TheRock/actions/runs/28049321998
- **Corrected Test PyTorch Full dispatch** (unblocks testing with proper inputs): https://github.com/ROCm/TheRock/actions/runs/28071296482

## Correct manual dispatch for dev multi-arch wheels

When re-running Test PyTorch Full against a dev build:

- `multi_arch`: `true`
- `package_index_url`: `https://rocm.devreleases.amd.com/whl-multi-arch/`
- `torch_version`: e.g. `2.11.0+rocm7.14.0a20260623`
- `pytorch_git_ref`: e.g. `release/2.11_gfx1250`
- `amdgpu_family`: `gfx94X-dcgpu` (selects device extras only; not an index subdir)

---
ISSUE ID : #3340


build_tools/github_actions/summarize_test_pytorch_workflow.py
ScottTodd, line 1: We should also fix https://github.com/ROCm/TheRock/blob/d5927980ea0a94885c3971f492c47f8b52de61ef/build_tools/github_actions/upload_python_packages.py#L164-L214

See an example run: https://github.com/ROCm/TheRock/actions/runs/28063143966

Image

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -159,12 +159,17 @@ jobs:
...
-            --pytorch-git-ref=${{ inputs.pytorch_git_ref }} \
-            --index-url=${{ inputs.package_index_url }} \
-            --index-subdir=${{ inputs.amdgpu_family }} \
-            --device-extras="${{ steps.expand.outputs.device_extras }}"
+          summarize_args=(
+            --torch-version=${{ inputs.torch_version }}
+            --pytorch-git-ref=${{ inputs.pytorch_git_ref }}
+            --index-url=${{ inputs.package_index_url }}
+          )
+          if [ "${{ inputs.multi_arch }}" != "true" ]; then
```
ScottTodd: I'd like to start removing some of these branches in workflows soon (maybe now, or at least not introduce new ones)

See https://github.com/ROCm/TheRock/issues/3340

build_tools/github_actions/summarize_test_pytorch_workflow.py
ScottTodd: Relevant workflow code:
* https://github.com/ROCm/TheRock/blob/d5927980ea0a94885c3971f492c47f8b52de61ef/.github/workflows/build_portable_linux_python_packages.yml#L170-L177
* https://github.com/ROCm/TheRock/blob/d5927980ea0a94885c3971f492c47f8b52de61ef/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels_ci.yml#L309-L315

ethanwee1: Fixed in 07b519f2. `upload_python_packages.py` now ignores dot-directories (e.g. `.kpack_staging`) when detecting the GPU-family layout via a new `detect_family_subdirs()` helper, so multi-arch PyTorch uploads no longer get misclassified as per-family and instead emit the flat kpack-split summary. This covers both call sites you linked (build_portable_linux_python_packages.yml and multi_arch_build_portable_linux_pytorch_wheels_ci.yml), since both run upload_python_packages.py --multiarch. Added a unit test for the dot-dir case.

.github/workflows/test_pytorch_wheels.yml
ethanwee1: Done — dropped the new branch. Since the summarize script now ignores `--index-subdir` when `--device-extras` is set, the workflow can pass both unconditionally, so test_pytorch_wheels.yml is back to its original (no `if multi_arch`) form and no new branch is introduced. (07b519f2)

build_tools/github_actions/summarize_test_pytorch_workflow.py
ethanwee1: Both of these call upload_python_packages.py with --multiarch, so the dot-dir fix in 07b519f2 corrects the summary for both without per-workflow changes.

build_tools/github_actions/upload_python_packages.py
```diff
@@ -260,7 +272,9 @@ def run(args: argparse.Namespace):
     if not args.output_dir:
         if args.multiarch:
             # Detect flat (kpack-split) vs legacy per-family layout from dist/.
-            family_subdirs = sorted(d.name for d in dist_dir.iterdir() if d.is_dir())
+            # Dot-directories (e.g. .kpack_staging) are staging artifacts, not
+            # GPU families, and must not trigger the per-family layout.
+            family_subdirs = detect_family_subdirs(dist_dir)
```
ScottTodd, line 277: This may not be a reliable check either in the future. We may upload packages into subdirectories instead of flat, e.g.

```
today:
  torch.whl
  rocm.whl
  jax.whl

later:
  torch/
    torch.whl
  rocm/
    rocm.whl
  jax/
    jax.whl
```

Currently the flat structure results in pages like https://rocm.nightlies.amd.com/whl-multi-arch/torch/ having this source:
```html
<html>
  <body>
    <h1>Links for torch</h1>
    <a href="[../torch-2.10.0%2Brocm7.14.0a20260520-cp310-cp310-linux_x86_64.whl](view-source:https://rocm.nightlies.amd.com/whl-multi-arch/torch-2.10.0%2Brocm7.14.0a20260520-cp310-cp310-linux_x86_64.whl)">torch-2.10.0+rocm7.14.0a20260520-cp310-cp310-linux_x86_64.whl</a><br/>
    <a href="[../torch-2.10.0%2Brocm7.14.0a20260520-cp311-cp311-linux_x86_64.whl](view-source:https://rocm.nightlies.amd.com/whl-multi-arch/torch-2.10.0%2Brocm7.14.0a20260520-cp311-cp311-linux_x86_64.whl)">torch-2.10.0+rocm7.14.0a20260520-cp311-cp311-linux_x86_64.whl</a><br/>
```
(note the `../` at the start of each link)

if we move the packages into subfolders then those links won't need the `../` indirection.

(cc @marbre )

---

By the time we do that, we should be able to also remove the `multiarch` branch from this code though.

ScottTodd (approved):

build_tools/github_actions/upload_python_packages.py
ethanwee1: Good point — agreed this flat-layout check won't hold once packages move into per-package subdirectories, and the `multiarch` branch can be dropped at that time. Filed #6251 to track both. Keeping the current check as-is for this PR since it matches today's flat layout.

marbre: Yes, to implement RFC0012, we will move away from a flat index in near future.

PR #6089, opened 2026-06-24
fix(packaging): preserve Gfxarch behavior for metapackages in kpack mode

raramakr:
Metapackages (like amdrocm-core-sdk) have no artifacts of their own,
  so artifact-based gfxarch detection returns False for them. This
  causes metapackage dependency expansion to fail - instead of depending
  on gfx-specific variants (amdrocm-core-sdk7.14-gfx1100, etc.), the
  metapackage only depends on the raw DEBDepends list.

  Fix by checking is_meta_package() before artifact verification and
  using Gfxarch metadata directly for metapackages.

Jira ID: https://github.com/ROCm/TheRock/issues/6093

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -265,6 +265,9 @@ def is_gfxarch_package(
 
     # In kpack mode, verify arch-specific artifacts exist
     if enable_kpack:
+        # Metapackages don't have artifacts but should still respect Gfxarch metadata
+        if is_meta_package(pkg_info):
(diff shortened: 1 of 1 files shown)


nunnikri (approved):

raramakr:
> 🚫 **Please fix the failed policies before requesting reviews.**
> 
> The following policy checks failed:
> 
> * ❌ Unit Test
> 
> The **`Not ready to Review`** label has been added to this PR. Once all policies pass, the label will be removed automatically.

For Linux native packaging, there are currently no unit tests available. I will be uploading a PR soon to cover all basic functionality with appropriate unit test cases.

PR #6098, opened 2026-06-24
fix: Add xnack variant support to Linux native packaging

raramakr:
ROCm builds for certain GPUs (e.g., gfx942) produce separate artifact
  directories for each xnack variant (:xnack+, :xnack-). Previously,
  packaging would only collect from one directory, missing variant artifacts.

  Unify all xnack variants into a single package per GPU target by:
  - Stripping xnack suffixes from target names for package naming
    (gfx942:xnack+ -> gfx942)
  - Globbing all variant directories during artifact collection

  This gives users one package that works regardless of xnack configuration

  Changes:
  - build_package.py: Strip xnack suffixes from discovered target families
  - packaging_utils.py: Strip xnack suffixes in target list normalization
  - packaging_utils.py: Glob for xnack variant artifact directories (:xnack+, :xnack-)

Jira ID: https://github.com/ROCm/TheRock/issues/6099

### DIFF
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -95,7 +95,9 @@ def get_all_target_families(artifact_dir):
 
     # Use ArtifactCatalog from _therock_utils to get all target families
     catalog = ArtifactCatalog(artifact_dir)
-    return sorted(catalog.all_target_families)
+    # Strip xnack suffixes (e.g., gfx942:xnack+ -> gfx942)
+    targets = {t.split(":", 1)[0] for t in catalog.all_target_families}
+    return sorted(targets)
 
 
 ################### Package Variant Builders #######################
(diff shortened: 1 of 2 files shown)


nunnikri (approved):
These changes are good. But might need similar change in python packaging as well.
The wheel package name is derived from target_family (e.g. rocm-sdk-device-gfx942). If gfx942:xnack+ appears as a target family, the wheel package name would become rocm-sdk-device-gfx942:xnack+ — a colon in a wheel name is invalid in PEP 508. Wheel packaging needs to strip xnack suffixes before using target families in package names. Can be put in this PR or different PR

raramakr:
> These changes are good. But might need similar change in python packaging as well. The wheel package name is derived from target_family (e.g. rocm-sdk-device-gfx942). If gfx942:xnack+ appears as a target family, the wheel package name would become rocm-sdk-device-gfx942:xnack+ — a colon in a wheel name is invalid in PEP 508. Wheel packaging needs to strip xnack suffixes before using target families in package names. Can be put in this PR or different PR

Will check the mentioned use case for wheel and will address in a different PR

PR #6102, opened 2026-06-24
Enable LLVM Assertions

kasaurov:
## Motivation

- Enable LLVM Assertion in PSDB and other testing based on `compiler/amd-staging` branch.
- Fix pytorch/gloo multi-arch assert by introducing a patch and a mechanism to apply it

## Technical Details

- Set `LLVM_ENABLE_ASSERTIONS=ON`. This enables LLVM Assertions in PSDB and Nightly builds.
- external-builds/pytorch: carry gloo native-HIP patch to fix multi-arch assert
   - PyTorch's bundled gloo pins (release/2.10-2.12) use the legacy FindHIP `hip_add_library` wrapper, which preprocesses device code with `clang -x hip -E -P` for every `--offload-arch`. With LLVM assertions enabled and multiple offload archs, this trips the new-driver assertion `HostOffloadingInputs.size() == 1` and breaks TheRock's multi-arch PyTorch build (ROCm/TheRock#6102).
   - Carry a patch that replaces gloo's `cmake/Hip.cmake` with the upstream native-HIP version (`enable_language(HIP)` + per-target `HIP_ARCHITECTURES`), which does not preprocess, keeps all requested archs, and mirrors gloo tip.
   - Deliver it via a patch-apply mechanism modeled on external-builds/uccl and build_tools/fetch_sources.py:
      - Port `apply_repo_patches`/`apply_all_patches` (git am --whitespace=nowarn --committer-date-is-author-date --no-gpg-sign) plus save_repo_patches into external-builds/pytorch/repo_management.py, applied to the checkout and its submodules, split into base (pre-hipify) and hipified (post-hipify) sets.
      - Wire base/hipified application into do_checkout, getattr-guarded so the audio/vision/apex/triton checkouts remain no-ops.
      - Add --patch/--no-patch and --patchset flags and a save-patches subcommand to pytorch_torch_repo.py.
      - Store patches under external-builds/pytorch/patches/<version>/<submodule>/base/. The patchset is selected from the PyTorch ref (release/2.12 -> 2.12); refs with no matching directory (e.g. nightly, whose gloo is already native) are a clean no-op.

## Test Plan

CI Build

## Submission Checklist


external-builds/pytorch/pytorch_torch_repo.py
```diff
@@ -57,6 +65,20 @@ def add_common(command_parser: argparse.ArgumentParser):
...
+            help="Base directory of TheRock-carried patches. Patches are laid out "
+            "as '<patch-dir>/<patchset>/<repo-or-submodule-relpath>/{base,hipified}/"
+            "*.patch' and applied with `git am` (mirrors external-builds/uccl).",
+        )
+        command_parser.add_argument(
+            "--patchset",
+            default=None,
+            help="Patchset directory name (defaults to the --repo-hashtag mangled "
+            "to a version, e.g. 'release/2.12' -> '2.12').",
+        )
```
ScottTodd, line 81: We haven't accepted patch files for pytorch for over 6 months (since https://github.com/ROCm/TheRock/pull/2795) and that is not going to change now.

See:
* https://github.com/ROCm/TheRock/blob/main/patches/README.md
* https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/README.md

Per the docs, changes must land upstream in pytorch and they *may* be cherry-picked back to older release branches if absolutely essential.

compiler/CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ if(THEROCK_ENABLE_COMPILER)
       # Features
       -DLLVM_INCLUDE_TESTS=${THEROCK_ENABLE_LLVM_TESTS}
       -DTHEROCK_BUILD_COMGR_TESTS=${THEROCK_BUILD_COMGR_TESTS}
+      -DLLVM_ENABLE_ASSERTIONS=ON
```
ScottTodd: What is the performance impact of this change? Do you want this in *all* builds, or just CI/nightly but not production releases?

We might want to move this setting to a CMake preset in https://github.com/ROCm/TheRock/blob/main/CMakePresets.json, see also:
* https://github.com/ROCm/TheRock/issues/5453

ScottTodd: composable_kernel is failing to build for gfx1200 with this change, on both Linux and Windows: https://github.com/ROCm/TheRock/actions/runs/31365742506/job/93387415191?pr=6102#step:18:19355

```
[composable_kernel] LLVM ERROR: Found 1 machine code errors.
[composable_kernel] PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and instructions to reproduce the bug.
[composable_kernel] Stack dump:
[composable_kernel] 0.	Program arguments: C:\\scratch-b\\build\\core\\clr\\dist\\lib\\llvm\\bin\\ld.lld --no-undefined -shared -plugin-opt=mcpu=gfx1200 -plugin-opt=O3 --lto-CGO3 --no-undefined C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\device_grouped_conv3d_bwd_data_wmma_v3_bilinear_ndhwgc_gkzyxc_ndhwgk_bf16_instance-a81f89-amdgcn-amd-amdhsa-gfx1200-eff2aa.o -o C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\device_grouped_conv3d_bwd_data_wmma_v3_bilinear_ndhwgc_gkzyxc_ndhwgk_bf16_instance-d57ab9.hipfb.amdgcn.gfx1200-579c67.img
[composable_kernel] 1.	Running pass 'CallGraph Pass Manager' on module 'ld-temp.o'.
[composable_kernel] 2.	Running pass 'Machine Instruction Scheduler' on function '@_ZN2ck16tensor_operation6device12_GLOBAL__N_145kernel_grouped_conv_bwd_data_wmma_cshuffle_v3INS_29GridwiseGemm_wmma_cshuffle_v3INS_13tensor_layout11convolution6NDHWGKENS6_6GKZYXCENS_5TupleIJNS6_6NDHWGCEEEESA_NS9_IJDF16bEEESC_fDF16bSC_DF16bNS0_12element_wise11PassThroughESE_NSD_8BilinearELNS1_18GemmSpecializationE7ELi256ELi128ELi128ELi32ELi8ELi8ELi16ELi16ELi8ELi1ENS_8SequenceIJLi4ELi64ELi1EEEENSH_IJLi1ELi0ELi2EEEESJ_Li2ELi8ELi8ELb0ELi1ESI_NSH_IJLi0ELi2ELi1EEEESK_Li1ELi1ELi8ELb0ELi1ELi1ELi1ENSH_IJLi1ELi16ELi1ELi16EEEENSH_IJLi8ELi8ELi8EEEELNS_26BlockGemmPipelineSchedulerE0ELNS_24BlockGemmPipelineVersionE0EDF16bDF16bLb0ELb0ELb0ELb1ELb0EEENS_16TensorDescriptorINS9_IJNS_5EmbedINS9_IJiiiiiEEESS_Lb0EEENS_11PassThroughIiEENS_3PadIiNS_17integral_constantIiLi0EEESY_Lb0EEESZ_SZ_SV_SV_NSR_INS9_IJiiEEENS9_IJiNSX_IiLi1EEEEEELb0EEES13_S13_SV_SV_NS_5SliceIiSY_iEENS14_IiiiEES15_S16_S15_S16_SV_NS_23Merge_v2_magic_divisionINS9_IJiiiiEEEEES19_NS_8RightPadIiiLb0EEES1B_NS_7UnMergeIS10_Lb0EEESV_EEENS9_IJNSH_IJLi0EEEENSH_IJLi1EEEENSH_IJLi2EEEENSH_IJLi3EEEENSH_IJLi4EEEENSH_IJLi5EEEENSH_IJLi6EEEENSH_IJLi7EEEENSH_IJLi8EEEENSH_IJLi9EEEENSH_IJLi10EEEENSH_IJLi11EEEENSH_IJLi12EEEENSH_IJLi13EEEENSH_IJLi14EEEENSH_IJLi15EEEENSH_IJLi16EEEENSH_IJLi17EEEENSH_IJLi18EEEENSH_IJLi20ELi22ELi24ELi26EEEENSH_IJLi19ELi21ELi23ELi25EEEENSH_IJLi27EEEENSH_IJLi28EEEENSH_IJLi29EEEENSH_IJLi30EEEEEEENS9_IJNSH_IJLi1ELi2ELi3ELi4ELi5EEEES1L_S1M_S1N_S1O_S1P_S1Q_NSH_IJLi12ELi13EEEENSH_IJLi14ELi15EEEENSH_IJLi16ELi17EEEES1X_NSH_IJLi19EEEENSH_IJLi20EEEENSH_IJLi21EEEENSH_IJLi22EEEENSH_IJLi23EEEENSH_IJLi24EEEENSH_IJLi25EEEENSH_IJLi26EEEES20_S21_S22_S23_NSH_IJLi31ELi32EEEENSH_IJLi33EEEEEEENSH_IJLi31ELi33ELi32EEEExEENSQ_INS9_IJNS1C_ISS_Lb0EEESV_S13_S13_S13_SV_SV_S15_S15_S15_NS_6FreezeIiEES2O_S2O_SV_S19_SV_S1B_S1B_S1D_SV_EEENS9_IJS1F_S1G_S1H_S1I_S1J_S1K_S1L_S1M_S1O_S1Q_S1N_S1P_S1R_S1S_NSH_IJLi15ELi16ELi17ELi14EEEES1X_S29_S2A_S2B_S2C_EEENS9_IJS25_S1L_NSH_IJLi7ELi8EEEENSH_IJLi9ELi10EEEENSH_IJLi11ELi12EEEES1S_S1T_S1U_S1V_S1W_NSH_IJEEES2V_S2V_S1X_S29_S2A_S2B_S2C_NSH_IJLi23ELi24EEEES2F_EEENSH_IJLi23ELi25ELi24EEEExEENS9_IJNSQ_INS9_IJST_SV_NSW_IiiiLb0EEES30_S30_SV_SV_NSR_IS10_S10_Lb0EEES31_S31_SV_SV_S2O_S16_S2O_S16_S2O_S16_SV_S19_SV_S1B_S1B_NS1C_INS9_IJiNSX_IiLi128EEEEEELb0EEES34_EEENS9_IJS1F_S1G_S1H_S1I_S1J_S1K_S1L_S1M_S1N_S1O_S1P_S1Q_S1R_S1S_S1T_S1U_S1V_S1W_S1X_NSH_IJLi19ELi20ELi21ELi22EEEES2D_S2E_S2F_S2G_S20_EEENS9_IJS25_S1L_S1M_S1N_S1O_S1P_S1Q_S26_S27_S28_S1X_S29_S2V_S2A_S2V_S2B_S2V_S2C_S2D_S2E_S2F_S2G_S20_NSH_IJLi28ELi29EEEENSH_IJLi30ELi31EEEEEEENSH_IJLi28ELi29ELi30ELi31EEEExEEEEES3C_Li32ENS1_49DeviceGroupedConvBwdDataMultipleD_Wmma_CShuffleV3ILi3ES7_S8_SB_SA_DF16bDF16bfDF16bSC_DF16bSE_SE_SF_LNS1_37ConvolutionBackwardDataSpecializationE0ELb1ELb1ELi256ELi128ELi128ELi32ELi8ELi8ELi16ELi16ELi8ELi1ESI_SJ_SJ_Li2ELi8ELi8ELi1ESI_SK_SK_Li1ELi1ELi8ELi1ELi1ELi1ESL_SM_LSN_0ELSO_0EDF16bDF16bLi1ELi1EE8GemmArgsENS1_30ComputePtrOffsetOfStridedBatchILi1ELi1ELi1EvEENS3I_ILi1ELi1ELi0EvEELNS_25InMemoryDataOperationEnumE0ELb0ELb0ELb0ELi1ELNS_10TailNumberE10EEEvNT_8ArgumentESt5arrayIT5_XT4_EEiT6_T7_i'
[composable_kernel] Exception Code: 0x80000003
[composable_kernel]  #0 0x00007ff65cbb27b5 (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0xe27b5)
[composable_kernel]  #1 0x00007ffadcc3bb04 (C:\Windows\System32\ucrtbase.dll+0x7bb04)
[composable_kernel]  #2 0x00007ffadcc3cad1 (C:\Windows\System32\ucrtbase.dll+0x7cad1)
[composable_kernel]  #3 0x00007ff65cb9726e (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0xc726e)
[composable_kernel]  #4 0x00007ff65e4cd628 (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0x19fd628)
[composable_kernel]  #5 0x00007ff65e4cd497 (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0x19fd497)
[composable_kernel]  #6 0x00007ff65e4d460c (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0x1a0460c)
[composable_kernel]  #7 0x00007ff65e8577d7 (C:\scratch-b\build\core\clr\dist\lib\llvm\bin\ld.lld.exe+0x1d877d7)
```

https://therock-ci-artifacts.s3.amazonaws.com/31365742506-linux/logs/math-libs/gfx120X-all/composable_kernel_build.log

Prior to any merge of this change we should add the `ci:run-all-archs` label to get complete coverage of all gfx targets, see https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

external-builds/pytorch/patches/2.11/pytorch/base/0001-native-hip-cpp-tests.patch
```diff
@@ -0,0 +1,131 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: TheRock <dev@therock>
+Date: Fri, 24 Jul 2026 12:00:00 -0400
+Subject: [PATCH] Use native CMake HIP language for HIP library and C++ tests
```
ScottTodd, line 4: *IF* we land patches, they should be authored by a real developer and reviewed as upstream PRs/commits. These patches just have `From: TheRock <dev@therock>`

ScottTodd (changes requested):
No new patch files.

compiler/CMakeLists.txt
kasaurov: This issue is unrelated to activities for switching ON LLVM assertions. 
Specifically to confirm that I have https://github.com/ROCm/TheRock/pull/6368/ which runs clean `compiler/amd-staging` branch on the same compiler submodule pin and experiences the same issue. This is a regression in upstream LLVM merged to `compiler/amd-staging`. It is worth investigation and fix before bumping compiler submodule in `main`, but cannot block turning Assertion On in `compiler/amd-staging` branch.
The regression entered ROCm/llvm-project's amd-staging through one of the six "merge main into amd-staging" commits that landed after the known-good 049cba8d and up to the known-bad bba90d22 (206 upstream commits, 137 of them in 66a577272 alone), and it reached TheRock on Aug 8 when the nightly 63b7b392d "Compiler submodule updates" bumped compiler/amd-llvm from 049cba8d to bba90d22.

kasaurov: Good questions, so:
- **Performance**. Measured on a controlled pair: this PR and #7228 carry identical patches and resolve to the same compiler pin c836d199, differing only in this flag. Assertions cost roughly 20% more build time overall (Linux +20.9%, Windows +19.8%) and 15–35% on stages that compile device code, worst case Math Libs gfx110X-all at +35%. The critical path grows by about 40 minutes. There is no runtime cost — assertions change what the compiler checks while compiling, not the code it emits.
- **Scope**. This PR targets **compiler/amd-staging** only, and nothing releases from that branch. The flow is one-way, with **main** merged into **amd-staging**, so production releases are unaffected by this change.
- **Placement.** By https://github.com/ROCm/TheRock/pull/6102/changes/31348cb6cc624d0b2c20d94395c4ecd03aef3a7b I've replaced the hardcoded **-DLLVM_ENABLE_ASSERTIONS=ON** with a declared flag in **FLAGS.cmake** — **LLVM_ENABLE_ASSERTIONS, DEFAULT_VALUE OFF**, with **CMAKE_VARS** scoped to the **amd-llvm** subproject — turned on for this branch via **BRANCH_CONFIG.json**. It is therefore off by default everywhere else, including **main** if this is ever ported there, and overridable per build with **-DTHEROCK_FLAG_LLVM_ENABLE_ASSERTIONS=...** On presets specifically: I don't think they can express "CI but not releases" today, since Linux CI passes an empty preset (the TODO referencing #1781) while Windows CI uses windows-release, the same preset releases use — a preset entry would do nothing on Linux CI and would enable assertions in Windows releases. If we later want this on **main**, the flag can be delivered more selectively without touching the declaration: a **ci:assertions** label mirroring **ci:asan**, emission from **configure_stage.py**, or a dedicated build variant.

external-builds/pytorch/patches/2.11/pytorch/base/0001-native-hip-cpp-tests.patch
kasaurov: Good catch, just in case I fixed that by https://github.com/ROCm/TheRock/pull/6102/changes/60ca8d421e4263b44307efc1b63409f411b61195

compiler/CMakeLists.txt
ScottTodd: Cool, building LLVM with asserts on the `compiler/amd-staging` branch and without asserts on the `main` branch by using `BRANCH_CONFIG.json` would address most of my concerns.

external-builds/pytorch/pytorch_torch_repo.py
kasaurov: A few things:
1. the PR is targeting **compiler/amd-staging** branch not **main**
2. I don't like idea of the patches either and it is more a temporary solution to unblock the PR rather than a permanent solution
3. The PyTorch version 2.13 doesn't require and patches to work, support for v 2.10 was dropped recently. Maybe we can let v 2.11 & 2.12 related targets just fail for now (remove all the patches and try the slow process of integrating the changes to PyTorch) -- @kzhuravl , @skganesan008 - thoughts?

ScottTodd: We could also limit CI testing on the `compiler/amd-staging` branch to just building the pytorch `release/2.13` branch with a change here: https://github.com/ROCm/TheRock/blob/a3e3f450db81b55f53a05581ebf52a21583c1126/build_tools/github_actions/configure_pytorch_release_matrix.py#L50-L53

In general I'd like for most of our energy to be focused on "live at head" development (latest code across all projects), with less and less effort spent on version N-1, N-2, and so on. Here we'd be making a tradeoff where there would be a visibility gap for "LLVM without assertions on pytorch 2.11 and 2.12 on the compiler/amd-staging branch" could break if we switch to testing LLVM _with assertions_ on that branch and disable those pytorch branch builds.

WDYT?

PR #6109, opened 2026-06-24
fix: Split WSL DXG library into separate amdrocm-wsl package

raramakr:
The librocdxg.so library requires GLIBC_2.38 and GLIBCXX_3.4.30+
  which are not available on RHEL/CentOS 9 (glibc 2.34). This was
  causing amdrocm-runtime installation to fail on EL9 systems with:
    nothing provides libc.so.6(GLIBC_2.38)(64bit)
    nothing provides libstdc++.so.6(GLIBCXX_3.4.30)(64bit)

  Since librocdxg.so is only needed for Windows Subsystem for Linux
  (WSL) environments, move it to separate optional packages so that
  native Linux users are not affected by its build dependencies.

  Changes:
  - Remove wsl-rocdxg artifact from amdrocm-runtime package
  - Remove wsl-rocdxg artifact from amdrocm-runtime-devel package
  - Create new amdrocm-wsl package containing wsl-rocdxg lib and doc
  - Create new amdrocm-wsl-devel package containing wsl-rocdxg dev
  - Update RELEASES.md with WSL package installation instructions

  Fixes: ROCm runtime installation failure on CentOS Stream 9

JIRA ID: https://amd-hub.atlassian.net/browse/ROCM-27077

### DIFF
--- a/build_tools/packaging/linux/package.json
+++ b/build_tools/packaging/linux/package.json
@@ -291,6 +291,41 @@
     "Gfxarch": "False"
   },
   {
+    "Package": "amdrocm-wsl-devel",
+    "Version": "",
+    "Architecture": "amd64",
+    "BuildArch": "x86_64",
+    "DEBDepends": [
+      "amdrocm-wsl"
+    ],
+    "RPMRequires": [
+      "amdrocm-wsl"
+    ],
+    "Maintainer": "AMD HSA Support <dl.HSA-Runtime-Support@amd.com>",
+    "Description_Short": "AMD ROCm WSL support library development files",
+    "Description_Long": "Development files for AMD ROCm DXG support library for Windows Subsystem for Linux (WSL)",
+    "Section": "devel",
+    "Priority": "optional",
+    "Group": "unknown",
+    "License": "MIT",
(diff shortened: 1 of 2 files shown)


nunnikri (approved):

build_tools/packaging/linux/package.json
idass1990: Should this also be pulled out of amdrocm-runtime-devel and added to a new amdrocm-wsl-devel package?

idass1990: **From Claude:**

if the lib/doc components were WSL-only and moved out of amdrocm-runtime, the dev headers for that library should also be WSL-only. Developers compiling WSL-specific code against wsl-rocdxg headers would still pull in this artifact via amdrocm-runtime-devel, which may transitively reintroduce the glibc version dependency on EL9 build environments. This should be either moved to amdrocm-wsl (or a new amdrocm-wsl-devel) or explicitly justified as intentional.

raramakr: Ack , will update the PR

raramakr: Done

idass1990 (approved):
LGTM

Issue #6077, opened 2026-06-23
[Issue] ROCDXG skips uploads on fork PRs leading to rocm-sdk test `no match for pattern 'librocdxg*.so*'` failures


ScottTodd:
Seeing this on a few of my PRs, all created from forks...? I haven't seen any failures on the default branch or other PRs yet so I'm not sure if this is unique to the specific runners or some files that are missing due to fork PRs having different access 🤔 

Example: https://github.com/ROCm/TheRock/actions/runs/28051028398/job/83064423105?pr=6071

```
Run rocm-sdk test
testCLI (rocm_sdk.tests.base_test.ROCmBaseTest.testCLI) ... ++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/python -m rocm_sdk --help
ok
testTargets (rocm_sdk.tests.base_test.ROCmBaseTest.testTargets) ... ++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/python -m rocm_sdk targets
ok
testVersion (rocm_sdk.tests.base_test.ROCmBaseTest.testVersion) ... ++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/python -m rocm_sdk version
ok
test_initialize_process_check_version (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_check_version) ... ok
test_initialize_process_check_version_asterisk (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_check_version_asterisk) ... ok
test_initialize_process_check_version_mismatch (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_check_version_mismatch) ... ok
test_initialize_process_check_version_mismatch_warning (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_check_version_mismatch_warning) ... ok
test_initialize_process_check_version_pattern (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_check_version_pattern) ... ok
test_initialize_process_env_preload_1 (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_env_preload_1) ... ok
test_initialize_process_env_preload_2_comma (rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_env_preload_2_comma) ... ok
[...]

PR #5283, opened 2026-05-15
Package WSL ROCDXG with ROCm runtime

jayhawk-commits:
## Summary

- Adds the `wsl-rocdxg` artifact to ROCm core runtime packaging.
- Includes `librocdxg` in the ROCm Python core package metadata.
- Includes ROCDXG runtime/doc files in `amdrocm-runtime` and development files in `amdrocm-runtime-devel`.
- Adds native package maintainer scripts that enable ROCR DXG detection for future shells when installed inside WSL.
- Documents WSL runtime usage in `RELEASES.md`.

## Motivation

ROCDXG is now built as a TheRock WSL artifact. This change makes that artifact available through the same ROCm runtime package surfaces users already install for core runtime support.

This addresses the packaging follow-up tracked in #5256 .

For WSL usage, ROCR needs `HSA_ENABLE_DXG_DETECTION=1` so it can route through DXG. Native packages can set this for future shells via `/etc/profile.d/rocm-wsl-rocdxg.sh`; Python packages and tarballs document the manual environment setup.

## Validation

- `pre-commit run --files RELEASES.md build_tools/packaging/linux/package.json build_tools/packaging/linux/template/scripts/amdrocm-runtime-postinst.j2 build_tools/packaging/linux/template/scripts/amdrocm-runtime-postrm.j2 build_tools/build_python_packages.py build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py`
- `python -m json.tool build_tools/packaging/linux/package.json`
- `git diff --check`

## Notes

CI should validate the rendered native packages and confirm the Python core package tests tolerate `rocdxg` being included in the core runtime library set.

PR #6111, opened 2026-06-24
fix: make rocdxg an optional library 

idass1990:
## Summary

   - Adds an `optional` flag to `LibraryEntry` and marks `rocdxg` with `optional=True`.
   - `find_libraries` skips an optional library when no file matches its glob, instead of raising `FileNotFoundError`.
   - `preload_libraries` resolves each shortname individually rather than zipping `shortnames` against a possibly-shorter `paths` list.

 ## Motivation

The `rocdxg` library only builds on a WSL host, so it is absent from fork PRs (which don't upload the `wsl-rocdxg` artifact) and from local single-command Linux builds (which can't produce it without a WSL host).

Because `rocdxg` was declared as an unconditional `LibraryEntry` in the always-installed `core` package, `testPreloadLibraries` — which iterates every installed library — failed with `no match for pattern 'librocdxg*.so*'`.

This implements the fix suggested in the #5283 review thread: making the library entry optional. It is data-driven, so future conditional libraries only need to pass `optional=True`.

This addresses
ISSUE ID #6077.
   
## Technical Details

- **`_dist_info.py`** — `LibraryEntry.__init__` gains `optional: bool = False`, stored as `self.optional`. `rocdxg` is the only entry that sets it. No special-casing elsewhere — any future conditional library opts in the same way.
- **`find_libraries` (`__init__.py`)** — when `relpath.glob(entry_pattern)` returns no matches and `lib_entry.optional` is set, the loop `continue`s instead of raising. This mirrors the existing skip for Windows-absent libraries (`is_windows and not lib_entry.dll_pattern`) a few lines above. Required libraries are unaffected and still raise `FileNotFoundError`.
- **`preload_libraries` (`__init__.py`)** — previously did `paths = find_libraries(*shortnames)` then `zip(shortnames, paths)`. Once `find_libraries` can skip an entry, that `paths` list is shorter than `shortnames` and the `zip` silently mispairs names with the wrong libraries (and drops the tail). This was already latent for Windows-skipped libraries. The fix resolves one shortname at a time (`find_libraries(shortname)`), skipping when the result is empty, so each name maps to its own path. The `_ALL_CDLLS` dedup cache and `rtld_global` behavior are unchanged.

## Test Result

- `python -m py_compile build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py build_tools/packaging/python/templates/rocm/src/rocm_sdk/__init__.py`
- Verified by inspection: `testPreloadLibraries` skips `rocdxg` cleanly when the artifact is absent; required-library-missing still raises `FileNotFoundError` and missing-package still raises `ModuleNotFoundError`.

## Notes

CI should confirm the rocm-sdk core tests pass on a fork PR (where the `rocdxg` artifact is absent).

## Submission Checklist

### DIFF
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/__init__.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/__init__.py
@@ -61,6 +61,10 @@ def find_libraries(*shortnames: str) -> list[Path]:
             entry_pattern = lib_entry.so_pattern
         matching_paths = sorted(relpath.glob(entry_pattern))
         if len(matching_paths) == 0:
+            if lib_entry.optional:
+                # Optional libraries (e.g. WSL-only rocdxg) may be absent from
+                # the distribution. Skip rather than failing.
+                continue
             raise FileNotFoundError(
                 f"Could not find rocm library '{shortname}' at path '{relpath},' no match for pattern '{entry_pattern}'"
             )
@@ -97,12 +101,17 @@ def preload_libraries(*shortnames: str, rtld_global: bool = True):
     """
     import ctypes
 
-    paths = find_libraries(*shortnames)
     mode = ctypes.RTLD_GLOBAL if rtld_global is True else ctypes.RTLD_LOCAL
-    for shortname, path in zip(shortnames, paths):
+    for shortname in shortnames:
         if shortname in _ALL_CDLLS:
             continue
-        cdll = ctypes.CDLL(str(path), mode=mode)
+        # Resolve per shortname: find_libraries may skip entries (libraries
+        # absent on the current platform or optional ones not present in the
+        # distribution), so a positional zip against shortnames would mispair.
+        paths = find_libraries(shortname)
+        if not paths:
+            continue
+        cdll = ctypes.CDLL(str(paths[0]), mode=mode)
         _ALL_CDLLS[shortname] = cdll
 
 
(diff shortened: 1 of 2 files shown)


build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/find_libraries_test.py
ScottTodd, line 1: I don't think this unit test belongs here. Files under `python/templates/rocm/src` are **distributed to users**. The existing tests in https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests are for `rocm-sdk test`

We have https://github.com/ROCm/TheRock/blob/main/build_tools/tests/py_packaging_test.py for the packaging itself, but I think the real test for the libraries is just https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/libraries_test.py. That should pass before and after these changes in each configuration, and this PR is already fixing the existing test failures in there (https://github.com/ROCm/TheRock/issues/6077). This doesn't seem worth writing some new +130 LOC test file for.

ScottTodd (commented):
Code changes LGTM. Unit tests don't look appropriate here.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/find_libraries_test.py
idass1990: makes sense , ill drop the unit test. thanks!

ScottTodd (approved):

Issue #5998, opened 2026-06-19
onboard therock-PR-Bot to therock repo


amd-shiraz:
onboard therock-PR-Bot to therock repo. 

tested and demoed to team at infra repo https://github.com/ROCm/TheRock-Infra/pull/339

PR #5994, opened 2026-06-19
ci: TheRock PR bot initial commit

samathew121:
## Motivation

therock-pr-bot tool to add policy based checks on PRs. 
right now we are adding label "not ready for review" for PRs that does not have issue/jira linked to it. and PRs that has zero unittest to it. we have other checks at place around forbidden files, pre-commit failure, branch name convention. and more checks marked as soon to enabled.  

## Technical Details

policy.yaml based PR BOT. this is not a llm/AI based bot. it is using code to deterministically provide checks. 

## Test Plan

tested in therock-infra repo and on this PR

## Test Result
Screenshot 2026-06-19 140311

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

ISSUE  ID  #5998

PR #6114, opened 2026-06-24
Delete "conventional commits" policy requirement from PR bot

ScottTodd:
## Motivation

Related to https://github.com/ROCm/TheRock/issues/5998. Follow-up to https://github.com/ROCm/TheRock/pull/5994#discussion_r3453865409, removing the Conventional Commits (https://www.conventionalcommits.org/en/v1.0.0/) requirement.

This sort of policy change needs to start with a discussion and inclusion in https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

I'm being picky here because
* I'm seeing lots of wasted energy being spent across the project to satisfy this check when it has not been sufficiently reviewed and implemented.
* Conventional commits will not work in https://github.com/ROCm/rocm-libraries and https://github.com/ROCm/rocm-systems where there are already established subproject tag styles used for commits with `[component-name]` syntax. We could aim to standardize across related ROCm projects, so we could formalize and adopt _that_ style here.
* See recent arguments against conventional commits at https://sumnerevans.com/posts/software-engineering/stop-using-conventional-commits/ and https://news.ycombinator.com/item?id=48414027

## Test Plan

Watch the PR bot run on this PR. It doesn't have unit tests yet so I can't test these changes effectively otherwise.

## Submission Checklist

### DIFF
--- a/skills/therock_pr_bot/policy.yml
+++ b/skills/therock_pr_bot/policy.yml
@@ -21,12 +21,8 @@ pr:
     # GitHub auto-revert branches, e.g. revert-5217-users/derobins/add_hipfile_support
     - '^revert-[0-9]+-[a-z0-9][a-z0-9\-_\/]*$'
 
-  # PR title rules (Conventional Commits style):
-  #   type(optional-scope): short description
-  # Allowed types: feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert
+  # PR title rules
   title:
-    pattern:
-      - '^(feat|fix|docs|style|refactor|perf|test|build|ci|chore|revert)(\([a-z0-9\-]+\))?(!)?: .+'
     title_min_length: 10
     title_max_length: 80
 
--- a/skills/therock_pr_bot/FAQ.md
+++ b/skills/therock_pr_bot/FAQ.md
@@ -37,31 +37,8 @@ ______________________________________________________________________
 ## 📝 PR Title
 
 **What does it check?**
-The PR title must follow **Conventional Commits** style so the changelog and release notes can be generated automatically.
 
 > **Note:** In the results table, the title and description checks are reported together as a single **PR Title/Description** row. Any title *or* description failure shows up there.
-
-**Required format**
-
-```
-type(optional-scope): short description
-```
-
-**Allowed types**
-
-| Type       | When to use                              |
-| ---------- | ---------------------------------------- |
-| `feat`     | A new feature                            |
-| `fix`      | A bug fix                                |
-| `docs`     | Documentation only changes               |
-| `style`    | Formatting, whitespace — no logic change |
-| `refactor` | Code restructure — no feature or fix     |
-| `perf`     | Performance improvement                  |
-| `test`     | Adding or fixing tests                   |
-| `build`    | Build system or dependency changes       |
-| `ci`       | CI / workflow changes                    |
-| `chore`    | Maintenance tasks                        |
-| `revert`   | Reverting a previous commit              |
 
 **Length rules**
 
@@ -74,8 +51,8 @@ type(optional-scope): short description
 Edit the PR title on GitHub (top of the PR page → pencil icon) to match the format, e.g.:
 
 ```
-feat(auth): add token refresh support
-fix(ci): correct codeql workflow trigger
+Add token refresh support
+Correct codeql workflow trigger
 ```
 
 ______________________________________________________________________
--- a/skills/therock_pr_bot/policy_check.py
+++ b/skills/therock_pr_bot/policy_check.py
@@ -295,11 +295,7 @@ def ensure_pr_title(policy: Policy, title: str, errors: List[str]) -> None:
     if policy.title_patterns and not any(
         p.search(title) for p in policy.title_patterns
     ):
-        errors.append(
-            "**Error:** Title does not follow Conventional Commits style.\n"
-            "**Expected:** start with a valid type (feat, fix, docs, …).\n"
-            f"{fmt}"
-        )
+        errors.append("**Error:** Title does not follow policy patterns.\n" f"{fmt}")
 
     if policy.forbidden_title_patterns:
         matched = [


kuhar (commented):
I'm fully in support, conventional commits don't bring value to projects like rocjitsu

kuhar:
Another example, the conventional commit check complains about PRs opened by the dependabot: https://github.com/ROCm/rocm-systems/pull/7886

ScottTodd:
> Another example, the conventional commit check complains about PRs opened by the dependabot: [ROCm/rocm-systems#7886](https://github.com/ROCm/rocm-systems/pull/7886)

For that, see also my review in https://github.com/ROCm/TheRock-Infra/pull/349#discussion_r3470139934

```yml
  # Special case: automated dependency "bump" PRs.
  # If the PR author is one of these bot accounts, ALL policy checks are
  # auto-approved (every row in the results table is marked Pass). The '[bot]'
  # suffix on the GitHub login is matched automatically.
  # Examples: "Bumps ROCm/rocm-systems from a0952b2 to 971dc69".
  bump_bot_authors:
    - assistant-librarian
    - systems-assistant
```

> My preference would be to hold bot PRs to the same standard as human-authored PRs, especially when we control what code and PR descriptions the bots generate. We can structure our policies such that our automated PRs meet all requirements. If we don't, then human-authored PRs with the exact same changes will be held to a different standard.
> 
> That being said, we could allow-list GitHub's dependabot and other bot authors as we don't control the branch names they use or the PR descriptions they write.

amd-shiraz:
@kuhar  we are tracking the dependabot use case and we will address then to be handled in subsequent improvements PRs. to add to this, we have discussed this review feedback from @ScottTodd  "to hold bot PRs to the same standard as human-authored PRs, especially when we control what code and PR descriptions the bots generate. We can structure our policies such that our automated PRs meet all requirements." , this will need changes also from a separation automation outside of this and will take bit more time. for now what we want to do is immediately whitelist bump PRs  coming from systems and libraries and enabled forked PRs with https://github.com/ROCm/TheRock/pull/6149 , while we get to the fix suggested here soon.

amd-shiraz:
with latest changes in , Bump PR whitelisted is working https://github.com/ROCm/TheRock/pull/6172 cc @kuhar , future bump PRs will be whitelisted for PR bot while address the standardization aspect next. cc @samathew121

skills/therock_pr_bot/policy.yml
```diff
@@ -21,12 +21,8 @@ pr:
     # GitHub auto-revert branches, e.g. revert-5217-users/derobins/add_hipfile_support
     - '^revert-[0-9]+-[a-z0-9][a-z0-9\-_\/]*$'
 
-  # PR title rules (Conventional Commits style):
-  #   type(optional-scope): short description
-  # Allowed types: feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert
+  # PR title rules
   title:
-    pattern:
-      - '^(feat|fix|docs|style|refactor|perf|test|build|ci|chore|revert)(\([a-z0-9\-]+\))?(!)?: .+'
```
ScottTodd, line 29: How about we allow any PR title that includes at least one `[some text]` or `(some text)` prefix? Like these:

```
[feat]: some cool feature
[hipblas][rocblas] add new unit tests
(ci) add new workflow
```

Could use a regex like `'^(?:\[[A-Za-z0-9][A-Za-z0-9._-]*\]|\([A-Za-z0-9][A-Za-z0-9._-]*\))+(?::)?\s+\S.*$'`

Per Codex (this would also be easier to iterate on with a unit test suite of PR titles we want to allow/reject):

> - `(?: ... )+` requires at least one leading scope.
> - `\[...\]|\(...\)` allows either [scope] or (scope).
> - `[A-Za-z0-9][A-Za-z0-9._-]*` prevents empty scopes and allows names like hipblas, rocm-libraries, gfx110X.
> - `(?::)?` allows an optional colon after the final scope.
> - `\s+\S.*` requires whitespace and a non-empty title after the scope prefix.
> 
> If you want to forbid the colon form, drop (?::)?. If you want lowercase-only scopes, change both A-Za-z ranges to a-z.

ScottTodd:
Looks like github rulesets can also enforce conventional commits: https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#enforce-conventional-commits (I still don't think we should adopt that, but where possible I'd prefer to use things built into github vs rolling our own)

image

amd-shiraz:
> Looks like github rulesets can also enforce conventional commits: https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#enforce-conventional-commits (I still don't think we should adopt that, but where possible I'd prefer to use things built into github vs rolling our own)
> 
> image

does this also apply to PR title/description ? My understanding is this is only for commit message and PR related resources are separate.

ScottTodd:
Yeah I think that's just the commit message not the PR title. So we could require that if someone pushes to a branch here every commit on that branch would need to follow the policy, no individual "typo fix" commits. If we merged with rebase&merge or merge commits that would ensure that commits on main all follow the convention at least.

(all that to say... probably not a good fit for what we want solved, but good to know about at least)

ScottTodd:
Still waiting on a review here @amd-shiraz .

skills/therock_pr_bot/policy.yml
amd-shiraz: should be fine. [hipblas] mentioning in the pr title sort of doesn't add much, as that component info is coming from the file anyway. If if the bigger consensus is to include all with least enforcement and restrictions i guess that is fine as well.

amd-shiraz (approved):
approving if the consensus from repo policy council. I suggest pls update this doc https://amd.atlassian.net/wiki/spaces/MLSE/pages/1798202764/TheRock+PR+Review+BOT+Policies+enforced accordingly to reflect this change in PR Title, and we can adjust automation from our side.

kuhar (approved):

ScottTodd:
I can also revise this PR to enforce `[]` or `()` or `:` prefixes in some form, rather than fully delete the current requirement in one step. I have some changes to CONTRIBUTING.md planned and could batch this together with that. If we decide on what we want then I'm happy to get all the code and docs aligned.

amd-shiraz:
@samathew121  once this gets merged, pls make the change accordingly from our side automation wise. thnx

samathew121:
> @samathew121 once this gets merged, pls make the change accordingly from our side automation wise. thnx

Sure

skills/therock_pr_bot/policy.yml
ScottTodd: See commit history in https://github.com/ROCm/rocm-libraries for historical conventions there

It's kind a mix, with many PRs/commits using a list of project prefixes like `[hipblaslt]` and `[tensilelist]`, and that is pretty useful (LLVM does it also)
image

Not all PRs have a single project that would fit in a short prefix.

More recent history shows people using the conventional commits format that the policy bot has been flagging: 
image

The "fix", "feat", "build", "refactor" isn't ... that bad, in my opinion, and I do like how it enforces some more structure in the commits, as people are picking up on putting the project in the parenthesis.

amd-shiraz:
@ScottTodd  may be for now lets delete it and keep it as is. i feel what goes in [ ]  as enforcement might need some more discussions we can take up in next iteration.

ScottTodd:
That works for me, thanks :)

kuhar:
@ScottTodd @amd-shiraz why does the bot still complain here? https://github.com/ROCm/rocm-systems/pull/8784

ScottTodd:
rocm-systems has its own copy of the policy bot script + config: https://github.com/ROCm/rocm-systems/tree/develop/tools/systems_pr_bot. We're iterating here in TheRock and then rolling out changes to the other repositories.

Eventually there may also be:
* A central/shared implementation workflow (and scripts used by it), with per-repository policy.yml files
* Other ways to adjust behavior within each repository such as adding directory exclusions, custom rules per project, etc.

PR #5794, opened 2026-06-11
Delete ci.yml and related files in favor of multi_arch_ci.yml

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/3340. These per-family / single-arch CI workflows have been replaced with multi-arch CI workflows:

Repository | Workflows
-- | --
TheRock | <ul><li>https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml</li><li>https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci_asan.yml</li><li>https://github.com/ROCm/TheRock/pull/5282</li></ul>
rocm-libraries | <ul><li>https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci.yml</li><li>https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci-nightly.yml</li><li>https://github.com/ROCm/rocm-libraries/pull/7906</li></ul>
rocm-systems | <ul><li>https://github.com/ROCm/rocm-systems/pull/6169</li><li>https://github.com/ROCm/rocm-systems/pull/6610</li></ul>

> [!TIP]
> See https://github.com/ROCm/TheRock/blob/main/docs/development/ci_overview.md and https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md for documentation about multi-arch CI.
>
> These documents will continue to be updated in future PRs.

> [!IMPORTANT]
> After this change there will be **no new "ci nightly" workflow runs** and **no new "ci tsan" workflow runs**:
> * https://github.com/ROCm/TheRock/actions/workflows/ci_nightly.yml
> * https://github.com/ROCm/TheRock/actions/workflows/ci_tsan.yml
>
> The current equivalent for "ci nightly" (which built/tested each GPU target family individually) is nightly multi-arch releases for all supported GPUs (https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release.yml) which trigger runs of https://github.com/ROCm/rockrel/actions/workflows/test_artifacts.yml using the "comprehensive" test filter for all available test runners.

## Technical Details

[...]

PR #4652, opened 2026-04-17
Flip `KPACK_SPLIT_ARTIFACTS` to `ON`

marbre:
This flips the `KPACK_SPLIT_ARTIFACTS` flag to enable kpack artifact splitting by default.

* Fixes https://github.com/ROCm/TheRock/issues/3339
* Fixes https://github.com/ROCm/TheRock/issues/3338

## Impact on multi-arch CI/CD

Libraries will now be processed by kpack, which will produce individual files for each GPU _architecture_ (not GPU _family_!), stored within `.tar.zst` compressed artifacts.

## Impact on non-multi-arch CI/CD

Non-multi-arch CI/CD is not set up to support the flag, see failures at https://github.com/ROCm/TheRock/actions/runs/24579570384?pr=4652. To keep non-multi-arch CI/CD across TheRock/rocm-libraries/rocm-systems/etc., we decided to opt-out of the flag in `build_tools/github_actions/build_configure.py`.

PR #4549, opened 2026-04-14
kpack split enable on integration branch 20260415

marbre:

PR #4257, opened 2026-03-31
Enable KPACK_SPLIT_ARTIFACTS flag on integration branch

stellaraccident:
Flip the kpack artifact splitting flag to ON via BRANCH_FLAGS.cmake
so that CI builds on this branch produce split (generic + arch-specific)
artifacts. The kpack runtime in CLR is already enabled unconditionally;
this enables the build-side splitting that produces the archives CLR
will load.

Changes:
- Add BRANCH_FLAGS.cmake with therock_override_flag_default(KPACK_SPLIT_ARTIFACTS ON)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #6118, opened 2026-06-24
docs: add feature flag lifecycle and canary promotion RFC

BrianHarrisonAMD:
JIRA ID : ALMIOPEN-1948

## Summary

Draft RFC proposing a disciplined feature-flag lifecycle for TheRock and the ROCm libraries it builds:

- **Flag taxonomy + decision rule** — runtime flags preferred; build-time only for artifact/ABI/topology changes (litmus questions included).
- **Generic runtime-flag contract** — a shipped `share/therock/feature_flags.json`, `dladdr`-relative discovery, `ROCM_FEATURE_<NAME>` precedence, and an example `rocm_feature_flags.h` reference header projects copy in (no shipped/linked/auto-included dependency; standalone builds fall back to compile-time defaults).
- **Flag metadata & hygiene** — owner/created/expires/stage/issue on every flag, surfaced in the configure report and manifest; expiry enforcement and a per-cycle flag-debt audit.
- **Canary-to-mainline promotion train** — a fixed, automated soak-and-promote cadence in the reserved `ci_weekly.yml` slot, release-manager-gated.
- **Team-owned both-state CI**, **backout/kill-switch policy**, **maintainer playbooks**, and **worked examples** (build-time exemplar `KPACK_SPLIT_ARTIFACTS`; runtime exemplars).

Extends the existing `FLAGS.cmake` registry rather than replacing it; consumes RFC0011 (Quartz) signals; references RFC0008 (Multi-Arch Packaging) as the build-time exemplar.

## Numbering

Filed with an **RFC00XX placeholder**. `0012` was taken by `RFC0012-Repo-Structure` and `0013` is contested by `users/nicknick/rfc0013-debug-symbol`. Final number will be determined once it's ready for merge.

## Status

- `mdformat` (gfm) and the other pre-commit hooks were run locally and pass.


tony-davis:
@BrianHarrisonAMD , this is really great work. The whole RFC is thoughtful, but the both-state CI section especially is exactly what teams need, and I think it is the highest-value part for day-to-day work. Honestly I would love to see this adopted right away; it would make flag use so much more frictionless for the devs. It is good enough that I actually want to see it expanded, so I opened a small amendment PR against this branch to do that, and added you as a reviewer: #6903.

The point of my amendment PR is to expand the reach. As written, both-state CI is specified against TheRock's own orchestrator, but the developers who lean on it live in rocm-libraries and rocm-systems, and those repos run their own CI (`.github/scripts/therock_configure_ci.py` + `therock_matrix.py`). Nothing in TheRock's CI steers a rocm-libraries PR, so the RFC as-is would give us the label on TheRock PRs and then still need separate (and probably subtly different) follow-ups in each consumer repo. My amendment closes that gap so we do not get the drift the RFC is otherwise trying to kill.

The good news is the fix is small, because the consumers already have both halves of the pattern: they already parse steering labels, and they already inject `-DTHEROCK_FLAG_*=ON` (the rocKE line). So the amendment makes both-state CI a repo-agnostic contract with one shared helper in TheRock (`flag_both_state.py`) that all three orchestrators call, and updates Goals, Implementation hooks, a new Multi-repo adoption subsection, P3, Open Question 3, and References to match.

One tidy-up while I was in there: the RFC referenced `configure_ci.py`, which was deleted in favor of `multi_arch_ci.yml` (#5794). I pointed those at `configure_multi_arch_ci.py` and described the long-lived-branch coverage via the workflow `on.push.branches` list, since `long_lived_full_match` is not in the tree anymore.

Please give the amendment a read. If you like it, all you have to do is merge it (it targets your branch). Or you can comment on it there and we can change it before adoption. I tried to make it easy either way. My motto here would be: define the contract once, implement it once, call it everywhere.

tony-davis:
Really like where this landed, Brian. One small thing I want to raise on the both-state trigger, because I think it'll bite whoever implements it.

The `flag:<NAME>:both` label bakes the flag name into the label text. GitHub only lets you apply a label that already exists, so a brand-new flag's label won't exist until someone creates it, per repo. That's a per-flag maintenance step the RFC doesn't currently account for (someone has to `gh label create flag:<NAME>:both` before a dev can even click it).

Suggestion: use one generic, permanent label as the switch, and declare the single flag to toggle in the PR description. The label exists once and never needs maintenance; the body carries the parameter. To match the existing convention (`ci:skip`, `ci:run-all-archs`, `ci:docs-only`, and friends), I'd name it `ci:flag-both-state`. I'd keep it to exactly one flag per PR on purpose (multiple gets hard to reason about, and expensive fast), and validate the name against the flag registry so a typo fails loud instead of silently doing nothing.

The flag name goes in the PR body on its own line, something like:

```
Flag: MY_FEATURE
```

Nice side effect on permissions: applying a label needs triage/write access, so `ci:flag-both-state` stays the "run it" gate, while the description just supplies the "which flag" detail. The helper is already name-agnostic, so this only changes where it reads the flag name from, not the leg-expansion logic.

I'd be happy to send a small PR against your branch if you'd like it.

tony-davis:
Thanks Brian! Per your note I pushed this straight to your branch rather than opening another PR (commit `e8acdcabe`), and added myself to the author list as you suggested.

What's in it: the per-flag `flag:<NAME>:both` label is replaced by a single permanent `ci:flag-both-state` label as the switch, with the one flag named in the PR description as `Flag: <NAME>` and validated against the flag registry. I kept it to exactly one flag per PR on purpose, and left single-state pinning to flip branches. The helper sketch picks up `parse_both_state_request(pr_labels, pr_body, known_flags)`, and `expand_flag_both_state` now takes the validated `flag_name`. There's a changelog entry spelling it all out.

One heads up on the diff: it reads bigger than it is. Only 4 table rows actually changed content; the other 26 are mdformat recomputing column padding after one cell got longer. The "hide whitespace changes" toggle takes it from 263 lines down to 191. The Python sketch is the part actually worth your eyes, and even there a chunk of the churn is just one nesting level disappearing when the multi-flag loop went away.

Revert or rework anything that doesn't fit your intent. It's your doc.

tony-davis:
Opened [#7112](https://github.com/ROCm/TheRock/pull/7112) against this branch, aligning the RFC with @stellaraccident's typed build flag provider protocol in #6984. Kept it as a separate PR so the diff is reviewable on its own (+164/-52) instead of more churn here.

@ScottTodd, this is the "where flags are declared" tension you tagged @BrianHarrisonAMD on. I resolved it in #6984's favor.

Briefly, what's in it:

- **Origination.** "All flags start in TheRock as the single source of truth" becomes "TheRock owns the mechanism and the aggregated inventory, not the birthplace." Stella said on #6984 that she wants flags to originate in either place "so we aren't so strictly interlocked," and I think she's right. The old rule blocks a team from guarding code until a declaration lands in TheRock *and* that version reaches their branch.
- **Flags are typed.** `BOOL`/`INTEGER` with `VALID_VALUES`, so the glossary's "named boolean toggle" and the declaration example both needed updating. Runtime flags stay `BOOL`-only for now.
- **Distribution boundary.** #6984 says never install a build-time input. We ship an installed `feature_flags.json`, so I added a subsection on why runtime state is on the other side of that line by construction rather than in violation of it.
- **Fail-closed on names.** A missing file still falls back silently (standalone builds have none), but a misspelled flag name has to be detectable now.
- Plus a P0 phase for #6984 itself, two resolved decisions, and two new open questions (how the inventory aggregates flags declared outside TheRock, and whether runtime flags should be typed too).

Please leave comments on [#7112](https://github.com/ROCm/TheRock/pull/7112) rather than here, so the discussion stays attached to the diff. I'm happy to fold it in, rework it, or drop any piece of it.

Thanks @BrianHarrisonAMD for the RFC and for taking the last amendment, @stellaraccident for #6984 (the typed protocol is a better foundation than what I had sketched), and @ScottTodd for catching the conflict between the two in the first place.

stellaraccident:
> Opened [#7112](https://github.com/ROCm/TheRock/pull/7112) against this branch, aligning the RFC with @stellaraccident's typed build flag provider protocol in #6984. Kept it as a separate PR so the diff is reviewable on its own (+164/-52) instead of more churn here.
> 
> @ScottTodd, this is the "where flags are declared" tension you tagged @BrianHarrisonAMD on. I resolved it in #6984's favor.
> 
> Briefly, what's in it:
> 
> * **Origination.** "All flags start in TheRock as the single source of truth" becomes "TheRock owns the mechanism and the aggregated inventory, not the birthplace." Stella said on [feat(build): add typed ROCm build flag provider protocol #6984](https://github.com/ROCm/TheRock/pull/6984) that she wants flags to originate in either place "so we aren't so strictly interlocked," and I think she's right. The old rule blocks a team from guarding code until a declaration lands in TheRock _and_ that version reaches their branch.
> * **Flags are typed.** `BOOL`/`INTEGER` with `VALID_VALUES`, so the glossary's "named boolean toggle" and the declaration example both needed updating. Runtime flags stay `BOOL`-only for now.
> * **Distribution boundary.** [feat(build): add typed ROCm build flag provider protocol #6984](https://github.com/ROCm/TheRock/pull/6984) says never install a build-time input. We ship an installed `feature_flags.json`, so I added a subsection on why runtime state is on the other side of that line by construction rather than in violation of it.
> * **Fail-closed on names.** A missing file still falls back silently (standalone builds have none), but a misspelled flag name has to be detectable now.
> * Plus a P0 phase for [feat(build): add typed ROCm build flag provider protocol #6984](https://github.com/ROCm/TheRock/pull/6984) itself, two resolved decisions, and two new open questions (how the inventory aggregates flags declared outside TheRock, and whether runtime flags should be typed too).
> 
> Please leave comments on [#7112](https://github.com/ROCm/TheRock/pull/7112) rather than here, so the discussion stays attached to the diff. I'm happy to fold it in, rework it, or drop any piece of it.
> 
> Thanks @BrianHarrisonAMD for the RFC and for taking the last amendment, @stellaraccident for #6984 (the typed protocol is a better foundation than what I had sketched), and @ScottTodd for catching the conflict between the two in the first place.

Ugh, sorry I forgot to land that. I just rebased and kicked the CI. Will land shortly. Then you all can take it where you would like.

kliegeois:
I have a question about flag retirement timing, and a concrete proposal I'd like to suggest for the RFC.

Once a feature has shipped in a release and been validated by QA, when should its flag be removed? The RFC already says flags are ephemeral and should be retired after default-on settles for roughly one cycle (Overview, Hygiene rule 3, and the per-cycle flag-debt audit), but it doesn't spell out a release-relative rule that teams can apply day to day.

## Proposed retirement timeline

Our team has been considering using this model, and @tony-davis confirmed it aligns with his intent for when a flag should be retired:

| Window | Keep flags?| Rationale |
|--------|---------------|-----------|
| **Current release** | Yes, all flags whose features shipped in this release | QA validation and kill-switch backout still matter for the release customers are on |
| **Active development** | Yes, all flags for work in flight (`in-development`, `canary`) | Both-state CI and promotion still depend on the toggle |
| **Second-most-recent release and older** | No, retire flags whose defaults were promoted ON and validated | One full release cycle of soak + QA is enough; keeping flags longer is debt |

In other words: keep flags for the **last release** and **current work**; start removing flags from the **release before that** once the feature is default-on and QA-signed-off.

## How this maps to the RFC today

- A flag promoted to `default-on` in release *N* would move to `deprecated` during release *N+1* (Hygiene rule 3's ~one-cycle kill-switch window).
- Removal PRs would target release *N+2* (when release *N* is the second-most-recent).
- The per-cycle flag-debt audit (Hygiene rule 5) could explicitly call out flags whose promoting release is now two releases behind.

## Suggestion for the RFC

Would it make sense to add a short subsection under **Hygiene rules** or **Stages** that states this release-relative rule explicitly? Something like:

> *Retirement cadence:* a flag whose default was promoted ON in release *R* enters `deprecated` in *R+1* and must be removed (flag, dead OFF path, and declaration) by the end of *R+2*, unless marked `long-lived`. The per-cycle flag-debt audit treats any `default-on` or `deprecated` flag from release *R−1* or earlier as overdue.

That would give teams a simple checklist without changing the underlying lifecycle mechanics.

BrianHarrisonAMD:
> I have a question about flag retirement timing, and a concrete proposal I'd like to suggest for the RFC.
> 
> Once a feature has shipped in a release and been validated by QA, when should its flag be removed? The RFC already says flags are ephemeral and should be retired after default-on settles for roughly one cycle (Overview, Hygiene rule 3, and the per-cycle flag-debt audit), but it doesn't spell out a release-relative rule that teams can apply day to day.
> 
> ## Proposed retirement timeline
> Our team has been considering using this model, and @tony-davis confirmed it aligns with his intent for when a flag should be retired:
> 
> Window	Keep flags?	Rationale
> **Current release**	Yes, all flags whose features shipped in this release	QA validation and kill-switch backout still matter for the release customers are on
> **Active development**	Yes, all flags for work in flight (`in-development`, `canary`)	Both-state CI and promotion still depend on the toggle
> **Second-most-recent release and older**	No, retire flags whose defaults were promoted ON and validated	One full release cycle of soak + QA is enough; keeping flags longer is debt
> In other words: keep flags for the **last release** and **current work**; start removing flags from the **release before that** once the feature is default-on and QA-signed-off.
> 
> ## How this maps to the RFC today
> * A flag promoted to `default-on` in release _N_ would move to `deprecated` during release _N+1_ (Hygiene rule 3's ~one-cycle kill-switch window).
> * Removal PRs would target release _N+2_ (when release _N_ is the second-most-recent).
> * The per-cycle flag-debt audit (Hygiene rule 5) could explicitly call out flags whose promoting release is now two releases behind.
> 
> ## Suggestion for the RFC
> Would it make sense to add a short subsection under **Hygiene rules** or **Stages** that states this release-relative rule explicitly? Something like:
> 
> > _Retirement cadence:_ a flag whose default was promoted ON in release _R_ enters `deprecated` in _R+1_ and must be removed (flag, dead OFF path, and declaration) by the end of _R+2_, unless marked `long-lived`. The per-cycle flag-debt audit treats any `default-on` or `deprecated` flag from release _R−1_ or earlier as overdue.
> 
> That would give teams a simple checklist without changing the underlying lifecycle mechanics.

Yea sounds good to me. Ill update with this feedback in the next update pass.

kliegeois:
> > I have a question about flag retirement timing, and a concrete proposal I'd like to suggest for the RFC.
> > Once a feature has shipped in a release and been validated by QA, when should its flag be removed? The RFC already says flags are ephemeral and should be retired after default-on settles for roughly one cycle (Overview, Hygiene rule 3, and the per-cycle flag-debt audit), but it doesn't spell out a release-relative rule that teams can apply day to day.
> > ## Proposed retirement timeline
> > Our team has been considering using this model, and @tony-davis confirmed it aligns with his intent for when a flag should be retired:
> > Window	Keep flags?	Rationale
> > **Current release**	Yes, all flags whose features shipped in this release	QA validation and kill-switch backout still matter for the release customers are on
> > **Active development**	Yes, all flags for work in flight (`in-development`, `canary`)	Both-state CI and promotion still depend on the toggle
> > **Second-most-recent release and older**	No, retire flags whose defaults were promoted ON and validated	One full release cycle of soak + QA is enough; keeping flags longer is debt
> > In other words: keep flags for the **last release** and **current work**; start removing flags from the **release before that** once the feature is default-on and QA-signed-off.
> > ## How this maps to the RFC today
> > 
> > * A flag promoted to `default-on` in release _N_ would move to `deprecated` during release _N+1_ (Hygiene rule 3's ~one-cycle kill-switch window).
> > * Removal PRs would target release _N+2_ (when release _N_ is the second-most-recent).
> > * The per-cycle flag-debt audit (Hygiene rule 5) could explicitly call out flags whose promoting release is now two releases behind.
> > 
> > ## Suggestion for the RFC
> > Would it make sense to add a short subsection under **Hygiene rules** or **Stages** that states this release-relative rule explicitly? Something like:
> > > _Retirement cadence:_ a flag whose default was promoted ON in release _R_ enters `deprecated` in _R+1_ and must be removed (flag, dead OFF path, and declaration) by the end of _R+2_, unless marked `long-lived`. The per-cycle flag-debt audit treats any `default-on` or `deprecated` flag from release _R−1_ or earlier as overdue.
> > 
> > 
> > That would give teams a simple checklist without changing the underlying lifecycle mechanics.
> 
> Yea sounds good to me. Ill update with this feedback in the next update pass.

Thanks @BrianHarrisonAMD !

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -0,0 +1,1317 @@
...
+| There is no dedicated `ci_nightly.yml` in the tree today (removed in the multi-arch migration). The design constraint still holds: GitHub `schedule` crons run only on the default branch (`main`), so a scheduled nightly cannot run on canary.                                                                                                                                                                                                                                                                                                                                                                                         | `.github/workflows/ci_nightly.yml`                                                             |
```
ScottTodd, line 123: Scheduled nightly releases, including full tests, run over in https://github.com/ROCm/rockrel now. The "ci nightly" that we had before overlapped with nightly releases as it was building the same code and running the same tests, so we retired it.

We can either add back a "ci nightly" in TheRock with feature flags or a "canary release" in rockrel.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -0,0 +1,1317 @@
...
+| ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `feature/*` | Immature flags (`in-development`), OFF everywhere.                                                                                                 | Presubmit only (EXISTING).                                                                                                     |
+| `canary`    | Soak the to-be-promoted flag default or defaults for one cycle. Only the current promotion batch is flipped ON here; everything else matches main. | CI configured to run on `canary` (add `canary` to `on.push.branches` in `.github/workflows/multi_arch_ci.yml` for postsubmit). |
+| `main`      | Mainline and trunk; the promotion target.                                                                                                          | Presubmit and postsubmit (EXISTING).                                                                                           |
```
ScottTodd: I'm not sure we need a git branch for canary. This sounds like a lot of extra infrastructure for what is essentially a single file (`BRANCH_CONFIG.json`). Could we just add a `FLAGS_CANARY.json` file to the default branch and have that run regular validation?

We maintained several integration branches for the kpack split flag flip:
* https://github.com/ROCm/TheRock/pull/4257
* https://github.com/ROCm/TheRock/pull/4397
* https://github.com/ROCm/TheRock/pull/4652
* https://github.com/ROCm/TheRock/pull/4549

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -0,0 +1,1317 @@
...
+  tsan) pattern in `amdgpu_family_matrix.py`, gated by the label so that it is opt-in per PR and
+  never the default across all families.
+- **Runtime:** the helper attaches a second test-env entry; the test workflow runs the target once
+  per entry. No extra build is required.
+- **Wiring (all three orchestrators call the same helper).** TheRock's `configure_multi_arch_ci.py`,
+  rocm-libraries' `.github/scripts/therock_configure_ci.py`, and rocm-systems'
+  `.github/scripts/therock_configure_ci.py` each call `expand_flag_both_state` on the legs they
+  build. Both consumer repos already check out TheRock and invoke its `build_tools/` scripts in CI
+  setup, so the helper is reachable without vendoring. Nothing toggles a flag from CI today, so the
+  helper plus these three call sites are the new work (see Multi-repo adoption).
```
ScottTodd, line 738: To build in multiple configurations I think we'll want a new top level workflow like how we have:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci_asan.yml
* `multi_arch_ci_flags.yml` ?

That would give us
* a unique workflow run id for CI runs with flag flips (the workflow run id is used as a namespace for artifacts uploaded to cloud storage buckets, see https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md)
* workflow runs for "standard" and "flags" would appear as separate checks on pull requests and could be independently cancelled/retried
* code like https://github.com/ROCm/TheRock/blob/main/build_tools/find_latest_artifacts.py that searches workflow run history can limit to `--workflow` to get only "default ci" workflow runs / artifacts / etc.
* dashboards like https://therock-hud.amd.com/ can also filter by workflow

We would add the same `multi_arch_ci_flags.yml` to rocm-libraries and rocm-systems.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -0,0 +1,1317 @@
...
+
+### Multi-repo adoption (consumer CI orchestrators)
+
+Both-state CI must behave identically wherever a flag-guarded change is developed. There are three
+independent CI orchestrators, and a change in a consumer repo is steered only by that repo's
+orchestrator, never by TheRock's:
```
ScottTodd, line 744: Shouldn't need any CI system changes in these repositories outside of the entry point workflow that calls into multi-arch CI, except maybe some dependency injection of which flags to use.

ScottTodd (commented):
The overall process sgtm. Some of the details about multi-arch CI integration could use some refinement. I think we could start by writing down what requirements we want fulfilled at the program level and then derive the implementation from there.

Implementation goals from the CI/CD side:
* Minimal infrastructure code outside of TheRock. No forked scripts/workflows. Code just `uses:` the github actions from TheRock and passes some configuration inputs
* Minimal branch maintenance. Prefer only a single long lived development branch where all work happens, including tracking which flags are being considered for promotion.

tony-davis:
@ScottTodd , thanks for digging into this one. I think you're right across the board here, and this is honestly the review I was hoping for. You know this CI plumbing way better than I do, and the branch and script pieces were always the part I was least sure about.

On canary as a branch: the kpack integration-branch history you linked makes the case better than I could have. Four PRs just to flip one flag is a lot of overhead for something a JSON file on main could do instead. I'd like to rework that section around a file.

On the dedicated multi_arch_ci_flags.yml workflow: that's a better shape than what I had, especially for the artifact-namespacing and dashboard-filtering reasons. Makes sense to me.

On multi-repo adoption: I think you're right that consumer repos shouldn't need to import Python at all. If the reusable workflow does the leg expansion and repos just pass in which flag to use, that's a cleaner boundary than what I sketched in #6903.

I'd like to take a pass at fleshing this out as another amendment PR against Brian's branch, same pattern as the last few, so you can see it concretely and tell me where it's still wrong. Give me a bit to put it together.

tony-davis:
Status check since this has a few threads going now. Two amendment PRs are open against this branch:

- ~~#7855: Brian's retirement-cadence cleanup (release-relative language, the `PROMOTED_IN` field). Looks ready to me; just waiting on kliegeois's review.~~ **Merged.**
- #7913: my rework of the canary and CI sections based on Scott's feedback above (canary as a committed `FLAGS_CANARY.json` file instead of a branch, a dedicated `multi_arch_ci_flags.yml` workflow instead of the shared-helper-import design, and a note that we don't need to revive `ci_nightly.yml` since rockrel already covers nightlies). @ScottTodd, tagged you as reviewer there so you can see it concretely.

The two touch different parts of the doc and shouldn't conflict, so either can merge independently once approved. Once both land on this branch, this PR will pick up the changes automatically.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
BrianHarrisonAMD: I believe resolved with Tony's updates in #7913.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
BrianHarrisonAMD: I believe resolved with Tony's updates in #7913.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
BrianHarrisonAMD: I believe resolved with Tony's updates in #7913.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
BrianHarrisonAMD: I think this, and the other items raised in #7913 are still outstanding, but I believe @tony-davis you are planning a follow-up to address these once you can sync on this?

(That's my understanding from [this comment anyways](https://github.com/ROCm/TheRock/pull/7913#issuecomment-5591476995))

PR #6121, opened 2026-06-24
Add precompile test option for rocfft/hipfft tests

malcolmroberts:
## Objective

Reduce test execution time for rocfft and hipfft.

## Technical Details

The RTC kernel generation can be more efficiently parallelized when as a preliminary batch job.  This commit adds that command-line argument to the test.  Typically, we have about a 10% time reduction.

## Test Plan

Covered by normal tests.

## Test Result

Tests run.

## Submission Checklist

- [ x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_hipfft.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hipfft.py
@@ -29,10 +29,13 @@ if test_type == "quick":
 else:
     # Standard/comprehensive: "--test_prob" is the probability that a given test will run.
     # Due to the large number of tests for hipFFT, we only run a subset.
+    # The option --precompile=precompiled.db reduces test time by more
+    # efficiently parallelizing runtime compilation.
     test_filter = [
         "--gtest_filter=-*multi_gpu*",
         "--test_prob",
         "0.01",
+        "--precompile=precompiled.db",
     ]
 
 cmd = [f"{THEROCK_BIN_DIR}/hipfft-test"] + test_filter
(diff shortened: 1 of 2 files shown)


geomin12 (commented):
looks like hipfft passed! 

however, rocfft failed https://github.com/ROCm/TheRock/actions/runs/28133376305/job/83336281594?pr=6121

github is weird where it doesn't show jobs running after adding a label, but the jobs did run...

malcolmroberts:
> looks like hipfft passed!
> 
> however, rocfft failed https://github.com/ROCm/TheRock/actions/runs/28133376305/job/83336281594?pr=6121
> 
> github is weird where it doesn't show jobs running after adding a label, but the jobs did run...

The point of this PR is to run the tests with the precompile option, and that is not what happened in the link provided.  I can't seem to re-run this PR (gh run rerun returned that I don't have admin rights), so I'm worried that the files that I modified aren't actually being used in the test infrastructure at all.  Or perhaps the test runs with the old test infrastructure, which is an issue with how PRs are tested in the rock.  Any idea what's up here, @geomin12 ?

malcolmroberts:
Ok, it looks like this entire file might be unused; it seems that https://github.com/ROCm/rocm-libraries/blob/develop/projects/hipfft/clients/tests/test_categories.yaml controls the actual code execution.

malcolmroberts:
The file that this modifying is actually deprecated, but has not been removed.

PR #6125, opened 2026-06-25
fix(ci): Fix upload_tarballs.py for single-family builds like ASAN

geomin12:
The script was failing for ASAN builds because the tarball naming logic expected either a multiarch tarball or a family name without hyphens. ASAN builds produce single-family tarballs like gfx94X-dcgpu which contain a hyphen in the family name.

Changes:
- Parse family name by finding where the version starts (first digit)
- Build a dict of all family tarball URLs
- Fall back to the first family tarball as "multiarch" for single-family builds to maintain backwards compatibility

working here: https://github.com/ROCm/TheRock/actions/runs/28115137163/job/83315737020

broken here: https://github.com/ROCm/TheRock/actions/runs/28049441058/job/83112141566

### DIFF
--- a/build_tools/github_actions/upload_tarballs.py
+++ b/build_tools/github_actions/upload_tarballs.py
@@ -60,16 +60,30 @@ def _select_shared_tarball_url(
     output_root: WorkflowOutputRoot,
     platform: str,
 ) -> str | None:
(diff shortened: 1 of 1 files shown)


geomin12:
PR #6076 fixes this

PR #6128, opened 2026-06-25
[NFC] Refactor multi-arch build config inputs

ScottTodd:
## Motivation

These argument lists were getting long, and I want to add a few more fields without needing to plumb across each function. Now the (frozen) dataclasses are passed through and each function can pick and choose what it wants to read.

## Test Plan

* Unit tests
* CI on this PR should generate the same set of jobs as before

## Submission Checklist


geomin12 (approved):
thanks, this is much much cleaner

PR #6078, opened 2026-06-23
Add gha_update_pr_comment foundation (POST/PATCH in github_actions_api)

amd-hsivasun:
## Motivation

When a TheRock submodule bump merges, upstream PRs in `rocm-systems` and `rocm-libraries` have no signal that their changes are now part of TheRock. Post-merge breadcrumb comments on those PRs need a shared way to create or update idempotent PR comments from CI.

This PR lands that foundation only: `gha_update_pr_comment()` in `github_actions_api.py`. Wiring into `bump_submodules.yml` (breadcrumbs) is a follow-on PR. The helper is also intended as a reusable alternative to ad-hoc upsert logic elsewhere (e.g. `therock_pr_bot/policy_check.py`).

## Technical Details
This PR lands the foundation only: extend github_actions_api with write support and add gha_update_pr_comment(). 
Follow-on PRs (not in this PR):
- Bump breadcrumbs: `post_bump_breadcrumbs.py` + step in `bump_submodules.yml`
- Manifest-diff report discoverability: step summary link (does not use this helper)

- Extend  gha_send_request with keyword-only method and body for POST/PATCH (REST and gh api paths).
- Add gha_update_pr_comment() to create or update a sticky PR comment identified by an HTML marker (paginated GET → PATCH or POST).
- Add unit tests for POST/PATCH plumbing, upsert paths, pagination, and GET empty-body backward compatibility.
- Document auth expectations in the helper docstring: callers set GITHUB_TOKEN on the step (GITHUB_TOKEN: ${{ github.token }} for in-repo; App token override for cross-repo). Posting comments requires pull-requests: write in job permissions.

## Test Plan

- pytest build_tools/github_actions/tests/github_actions_api_test.py -v

## Test Result

- 47 passed locally in build_tools/github_actions/tests/github_actions_api_test.py

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6147, opened 2026-06-25
feat(manifest-diff): Add manifest-diff link to multi-arch CI setup step summary

amd-hsivasun:
## Motivation

The manifest-diff HTML report is generated and uploaded to S3 on multi-arch CI runs, but it isn't currently easy to find. This PR adds the report link to the **setup** job step summary **Build outputs** table — the same place reviewers already use for Logs, Artifacts, and Manifests.

This is the discoverability follow-on called out in #6078 (separate from the bump-breadcrumb PR comment work).

## Technical Details

In `configure_multi_arch_ci_summary.py`, `_append_build_rocm()` now adds a **Manifest diff** row when building the linux output URLs:

- Uses the existing `WorkflowOutputRoot` pattern: `{run_id}-linux/logs/manifest-diff/index.html`
- Reuses the linux `output_root` already created in the platform loop (manifest-diff uploads under the linux run tree)
- Adds a short static note that:
  - the report may show no changes when no submodule pointers moved, and
  - manifest-diff is not run on ASAN workflows yet, so the link may be unavailable there

No workflow, gating, or `generate_manifest_diff_report.py` changes.

## Test Plan

- [x] Open a Multi-Arch CI PR run and confirm the setup step summary **Build outputs** table shows the **Manifest diff** link
- [x] Confirm the S3 URL resolves after the `manifest_diff` job completes

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci_summary.py
+++ b/build_tools/github_actions/configure_multi_arch_ci_summary.py
@@ -195,27 +195,29 @@ def _append_build_rocm(
             "-- | -- | -- | --",
         ]
     )
+    linux_output_root = None
     for platform_name in ["linux", "windows"]:
         output_root = WorkflowOutputRoot.from_workflow_run(
             run_id=ci_inputs.run_id, platform=platform_name
         )
+        if platform_name == "linux":
+            linux_output_root = output_root
         log_url = output_root.log_root_index().https_url
         artifact_url = output_root.artifact_index().https_url
         manifest_url = output_root.manifests_index().https_url
         lines.append(
             f"{platform_name.capitalize()} | {log_url} | {artifact_url} | {manifest_url}"
         )
-    output_root = WorkflowOutputRoot.from_workflow_run(
-        run_id=ci_inputs.run_id, platform="linux"
-    )
-    manifest_diff_url = output_root.log_file("manifest-diff", "index.html").https_url
-    lines.append(f"Manifest diff | {manifest_diff_url} | — | —")
+    manifest_diff_url = linux_output_root.log_file(
+        "manifest-diff", "index.html"
+    ).https_url
+    lines.append(f"Manifest diff *(if produced)* | {manifest_diff_url} | — | —")
     lines.append("")
     lines.append(
         "> The manifest-diff report compares submodules across the CI commit range. "
         "Expect no changes when this run does not advance any submodule pointers. "
-        "Manifest-diff is not run on ASAN workflows yet — the report link may be "
-        "unavailable on those runs."
+        "The report is generated after the setup job completes and the link may be "
+        "unavailable until then; it is not produced on ASAN workflows."
     )
 
 
--- a/build_tools/github_actions/configure_multi_arch_ci_summary.py
+++ b/build_tools/github_actions/configure_multi_arch_ci_summary.py
@@ -205,6 +205,18 @@ def _append_build_rocm(
         lines.append(
             f"{platform_name.capitalize()} | {log_url} | {artifact_url} | {manifest_url}"
         )
+    output_root = WorkflowOutputRoot.from_workflow_run(
+        run_id=ci_inputs.run_id, platform="linux"
+    )
+    manifest_diff_url = output_root.log_file("manifest-diff", "index.html").https_url
+    lines.append(f"Manifest diff | {manifest_diff_url} | — | —")
+    lines.append("")
+    lines.append(
+        "> The manifest-diff report compares submodules across the CI commit range. "
+        "Expect no changes when this run does not advance any submodule pointers. "
+        "Manifest-diff is not run on ASAN workflows yet — the report link may be "
+        "unavailable on those runs."
+    )
 
 
 def _append_build_pytorch(lines: list[str], outputs: CIOutputs) -> None:
--- a/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
+++ b/build_tools/github_actions/tests/configure_multi_arch_ci_test.py
@@ -1363,6 +1363,7 @@ class TestFormatSummary(unittest.TestCase):
             build_variant_cmake_preset="",
             build_native_linux=True,
             build_pytorch=False,
+            build_jax=False,
         )
         jobs = cm.JobDecisions(
             build_rocm=cm.BuildRocmDecision(action=cm.JobAction.RUN),
@@ -1370,6 +1371,7 @@ class TestFormatSummary(unittest.TestCase):
             build_rocm_python=cm.JobGroupDecision(action=cm.JobAction.RUN),
             build_pytorch=cm.JobGroupDecision(action=cm.JobAction.RUN),
             test_pytorch=cm.JobGroupDecision(action=cm.JobAction.RUN),
+            build_jax=cm.JobGroupDecision(action=cm.JobAction.SKIP),
         )
         outputs = cm.CIOutputs(
             is_ci_enabled=True,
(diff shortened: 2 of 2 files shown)


erman-gurses (commented):
Sorry for late review - initially, can you please point out the link?

amd-hsivasun:
> Sorry for late review - initially, can you please point out the link?

The link can be found in the step summary.
image

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -197,6 +197,18 @@ def _append_build_rocm(
         lines.append(
             f"{platform_name.capitalize()} | {log_url} | {artifact_url} | {manifest_url}"
         )
+    output_root = WorkflowOutputRoot.from_workflow_run(
+        run_id=ci_inputs.run_id, platform="linux"
+    )
+    manifest_diff_url = output_root.log_file("manifest-diff", "index.html").https_url
+    lines.append(f"Manifest diff | {manifest_diff_url} | — | —")
```
erman-gurses: This adds the `manifest-diff` row unconditionally. Since the PR note says the report is not run on ASAN workflows, this can surface a broken link when the artifact is not produced. Could we gate this row on whether the manifest-diff report exists, or render a clear unavailable state instead?

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -197,6 +197,18 @@ def _append_build_rocm(
         lines.append(
             f"{platform_name.capitalize()} | {log_url} | {artifact_url} | {manifest_url}"
         )
+    output_root = WorkflowOutputRoot.from_workflow_run(
+        run_id=ci_inputs.run_id, platform="linux"
+    )
```
erman-gurses: `platform="linux"` is hardcoded here. If the summary is meant to reflect the active build configuration, this should come from the selected build/output context instead of assuming Linux. Otherwise the link construction becomes brittle if the artifact location ever changes.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1135,6 +1135,36 @@ def test_skipped_ci_write_outputs_summary(self):
...
+            is_ci_enabled=True,
+            builds=cm.BuildConfigs(linux=linux),
+            jobs=jobs,
+        )
+        result = format_summary(self._inputs(), outputs)
+        self.assertIn("## Build outputs", result)
+        self.assertIn("Linux |", result)
+        self.assertIn("Windows |", result)
+        self.assertIn("Manifest diff |", result)
+        self.assertIn("12345-linux/logs/manifest-diff/index.html", result)
```
erman-gurses: The test only checks for substring presence. It would be stronger to assert the exact rendered table row, so we know the link is inserted in the intended column order and not just appearing somewhere in the summary text.

erman-gurses (commented):
Please see my comments below

build_tools/github_actions/configure_multi_arch_ci_summary.py
amd-hsivasun: The issue is that the manifest-diff runs on a separate workflow job. It was run this way to make sure the setup job runs smooth without waiting on the manifest-diff job. Thus the info of whether or not the manifest-diff was generated wont be present in the setup job as it will finish before the manifest-diff. Any suggestions for a workaround if this is a needed gate?

I was initially considering adding logic to check if the manifest-diff will run, but I decided it would be adding too much code for a simple gate.

erman-gurses: Thanks for the clarification - that makes sense. I agree that gating this based on artifact existence isn't practical since the setup job completes before the `manifest-diff` job runs.

My concern is less about the implementation and more about the UX. This line:

`lines.append(f"Manifest diff | {manifest_diff_url} | — | —")`

always presents the link as if the report is available, even though we know there are cases (e.g. ASAN) where it won't exist.

Would it make sense to either:

adjust the label to indicate the report is generated later (or only if produced), or
have the manifest-diff job append/update the summary once the artifact has been uploaded?

Either approach would avoid users interpreting a broken link as a CI failure while keeping the setup job simple.

build_tools/github_actions/configure_multi_arch_ci_summary.py
amd-hsivasun: The platform="linux" here is intentional rather than an assumption — the manifest-diff job always runs on a Linux runner (aws-linux-scale-rocm-prod), and upload_test_report_script.py derives the upload path from platform.system().lower(), which will always resolve to "linux" on that runner. So the report is always stored under the linux output root by construction. Happy to add a comment in the code if you'd like, but wanted to confirm first that you agree with the reasoning.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
amd-hsivasun: Now checks to see if the link is in the specific row

build_tools/github_actions/configure_multi_arch_ci_summary.py
erman-gurses: I see - I think we can keep it as it is.

erman-gurses (approved):
LGTM! Please make sure about the unit tests. Looks like they are failing.

amd-hsivasun:
> LGTM! Please make sure about the unit tests. Looks like they are failing.

Unit Tests have been fixed merging

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1353,6 +1353,44 @@ def test_skipped_ci_write_outputs_summary(self):
...
+            jobs=jobs,
+        )
+        result = format_summary(self._inputs(), outputs)
+        self.assertIn("## Build outputs", result)
+        self.assertIn("Linux |", result)
+        self.assertIn("Windows |", result)
+        manifest_diff_url = (
+            "https://therock-ci-artifacts.s3.amazonaws.com"
+            "/12345-linux/logs/manifest-diff/index.html"
+        )
```
ScottTodd, line 1388: This fails on PRs from forks. Can't depend on a specific bucket URL like this, see https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md#ci-buckets

https://github.com/ROCm/TheRock/actions/runs/29369295281/job/87208493407?pr=6278#step:5:1173

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
ScottTodd, line 1350: The comments here were there for a reason. Please read before making changes like this. Let's revert these new tests.

build_tools/github_actions/configure_multi_arch_ci_summary.py
```diff
@@ -195,16 +195,30 @@ def _append_build_rocm(
             "-- | -- | -- | --",
         ]
     )
+    linux_output_root = None
     for platform_name in ["linux", "windows"]:
         output_root = WorkflowOutputRoot.from_workflow_run(
             run_id=ci_inputs.run_id, platform=platform_name
         )
+        if platform_name == "linux":
+            linux_output_root = output_root
```
ScottTodd, line 204: Why is this file stored only under the "linux" output? What about Windows?

ScottTodd (commented):
Please author/review changes more carefully. Too many basic issues here.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
ScottTodd: Reverted the test in https://github.com/ROCm/TheRock/pull/6546

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
ScottTodd: We might want to override the default bucket URL under [CI] unit tests in https://github.com/ROCm/TheRock/blob/main/.github/workflows/unit_tests.yml to catch issues like this where PRs from forks and those not from forks have different behavior.

PR #6158, opened 2026-06-25
Promote JAX packages in multi-arch releases

JeniferC99:
## Motivation

As multi-arch becomes the standard release flow, JAX should be promoted in `PACKAGES_TO_PROMOTE_MULTI_ARCH`.

## Technical Details

Adds `jax_rocm7_pjrt` and `jax_rocm7_plugin` to `PACKAGES_TO_PROMOTE_MULTI_ARCH`. `jaxlib` is intentionally excluded: it is not produced as a multi-arch artifact — only the plugin/pjrt wheels exist under `whl-multi-arch/`, while `jaxlib` is published solely in the per-family `whl/<gfx>/` indexes — so it stays in the single-arch `PACKAGES_TO_PROMOTE` set only.

## Submission Checklist


build_tools/packaging/download_python_packages.py
```diff
@@ -157,6 +157,9 @@
     "amd_torch_device",
     "amd_torchvision_device",
     "apex",
+    "jax_rocm7_pjrt",
+    "jax_rocm7_plugin",
+    "jaxlib",
```
araravik-psd: do we have a jaxlib file getting created for multi-arch builds? I dont see it in s3, see only jax_rocm_prjt and jax_rocm7_plugin for multi-arch artifacts. Can you please check this?

araravik-psd: Thanks @JeniferC99 for addressing this

araravik-psd (approved):

build_tools/packaging/download_python_packages.py
JeniferC99: Confirmed — we don't build a `jaxlib` wheel for multi-arch. Both build modes produce only the plugin + pjrt wheels: see [multi_arch_build_linux_jax_wheels.yml#L256-L269](https://github.com/ROCm/TheRock/blob/99f48ca9ecc4f3c92dadc2baef9e560e5e7d18b0/.github/workflows/multi_arch_build_linux_jax_wheels.yml#L256-L269) (`--wheels=jax-rocm-plugin,jax-rocm-pjrt`); the `jaxlib_git_hash` bazel arg only stamps metadata. This matches S3 (`jax_rocm7_plugin`/`jax_rocm7_pjrt` present, no `jaxlib`). I've dropped `jaxlib` from `PACKAGES_TO_PROMOTE_MULTI_ARCH` in 768b9e6 and updated the PR description; `jaxlib` stays in the single-arch set since it's published under the per-family `whl/<gfx>/` indexes.

erman-gurses: I confirmed this back in time while I was working on JAX release. After 0.8.2, we do not build and release `jaxlib` wheel

PR #6160, opened 2026-06-25
fix: Disable RPM stripping for all packages

raramakr:
Hardcode disable_rpm_strip to True to prevent binary stripping
during RPM package generation, preserving debug symbols.

Jira ID: https://amd-hub.atlassian.net/browse/ROCM-26016

### DIFF
--- a/build_tools/packaging/linux/rpm_package.py
+++ b/build_tools/packaging/linux/rpm_package.py
@@ -179,7 +179,7 @@ def generate_spec_file(pkg_name, specfile, config: PackageConfig):
         "conflicts": conflicts,
         "rpmrecommends": rpmrecommends,
         "rpmsuggests": rpmsuggests,
-        "disable_rpm_strip": is_rpm_stripping_disabled(pkg_info),
+        "disable_rpm_strip": True,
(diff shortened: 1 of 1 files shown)


nunnikri (approved):

marbre (commented):
@raramakr @nunnikri is or was it ever needed to stripping at the packaging stage at all? We don't build with debug symbols as far as I am aware of and if we would, we would rather split those out in a step prior to packaging to then allow pushing the split out debug symbols to their own dedicated packages.

raramakr:
> @raramakr @nunnikri is or was it ever needed to stripping at the packaging stage at all? We don't build with debug symbols as far as I am aware of and if we would, we would rather split those out in a step prior to packaging to then allow pushing the split out debug symbols to their own dedicated packages.

@marbre Stripping debug symbols and generating debuginfo/dbgsym packages are typically handled by the packaging tools (e.g., rpmbuild, dpkg-buildpackage) by default. They provide options to control or override this behavior if needed.
Depending on how we want to handle debug information, we can adjust the configuration accordingly—for example, by disabling stripping during packaging or by explicitly splitting debug symbols in a prior step and packaging them separately.

PR #4518, opened 2026-04-14
Update `build_python_packages.py` to package `rocprofiler-compute` and `rocprofiler-systems` into profiler wheel: Phase 3

erman-gurses:
Progress on #3573
## Motivation

This PR introduces a dedicated `rocm-profiler` wheel for `rocprof-sys` and `rocprof-compute` by packaging directly from profiler artifacts instead of relying on `base`.

The goal is to:
- Package `rocprofiler-compute` and `rocprofiler-systems` into a dedicated profiler wheel
- Ensure CLI tools (`rocprof-compute`, `rocprof-sys-*`) are available and functional from the installed wheel
- Align packaging with TheRock’s artifact-driven model
- Resolve runtime dependency issues (sysdeps and libomp) so profiler tools execute correctly

---

### Artifact Selection
- Switched profiler artifact selection to:
  - `rocprofiler-compute` (`lib`, `run`)
  - `rocprofiler-systems` (`lib`, `run`)
- Extended selection to include:
  - `sysdeps` (`lib`) for runtime dependency closure

### Runtime Payload Updates
- Added required `rocm_sysdeps` runtime libraries:
  - `librocm_sysdeps_dw.so*`
  - `librocm_sysdeps_elf.so*`
  - `librocm_sysdeps_z.so*`
  - `librocm_sysdeps_zstd.so*`
  - `librocm_sysdeps_liblzma.so*`
  - `librocm_sysdeps_bz2.so*`

### RPATH Fixes
- Added additional RPATH dependency:
  - `core/lib`
  - `core/lib/llvm/lib`
- This enables resolution of `libomp.so` from `rocm-sdk-core`

### CLI Runtime Fix
- Updated `_cli.py` to extend `LD_LIBRARY_PATH` at runtime:
  - Points to `_rocm_sdk_core/lib/rocm_sysdeps/lib`
- Ensures packaged sysdeps are visible to profiler binaries

### Other Packaging Updates
- Recreated missing symlink:
  - `bin/rocprof-compute -> ../libexec/rocprofiler-compute/rocprof-compute`
- Updated packaging templates to include `_rocm_profiler` platform files
- Ensured artifact-driven packaging instead of relying on `base`

---

### Artifact Preparation
- Synced CI artifacts:
  ```
  s3://therock-ci-artifacts/24104028483-linux/
  ```
- Extracted:
  - `base_{lib,run}_generic`
  - `rocprofiler-compute_{lib,run}_generic`
[...]

Issue #3573, opened 2026-02-23
Create Separate Profiler Wheel for rocprof-sys and rocprof-compute


erman-gurses:
## Context

Currently, ROCm Python packaging creates these wheels:
- `rocm` (meta package, sdist only)
- `rocm-sdk-core` (runtime: HIP, compilers, rocprofiler-sdk with rocprofv3)
- `rocm-sdk-libraries-{target_family}` (math/ML libraries)
- `rocm-sdk-devel` (development files in tarball)

**Problem:** rocprofiler-systems (rocprof-sys) and rocprofiler-compute (rocprof-compute) are defined as artifacts but NOT included in any Python wheel, even though rocprofv3 (from rocprofiler-sdk) IS included.

**Requirement:** Create a new separate wheel `rocm-profiler` that ships both rocprofiler-systems and rocprofiler-compute binaries.

## Artifacts Analysis

From BUILD_TOPOLOGY.toml:

1. **rocprofiler-sdk** (already in core package)
   - artifact_group: profiler-core
   - Contains: rocprofv3, rocprof-attach, rocprofiler SDK libraries
   - Status: ✅ Already included in rocm-sdk-core

2. **rocprofiler-compute**
   - artifact_group: profiler-core
   - Contains: rocprof-compute binaries and tools
   - Status: ❌ NOT included in any wheel

3. **rocprofiler-systems**
   - artifact_group: profiler-apps
   - Contains: rocprof-sys-* binaries (rocprof-sys-avail, rocprof-sys-instrument, rocprof-sys-run, rocprof-sys-sample, rocprof-sys-causal)
   - Status: ❌ NOT included in any wheel

## Implementation Plan

- [x] Phase 1: Define New Package Entry
- [x] Phase 2: Create Package Template
- [x] Phase 3: Update Build Script
- [x] Phase 4: Update Meta Package (rocm)
- [x] Phase 5: Keep rocprofiler-sdk in Core Package
- [x] Phase 6: Follow Meta Package Convention
- [x] Phase 7: Documentation Updates

## Phase 1: Define New Package Entry

**File:** `/root/TheRock/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py`

**Changes:**
1. Add new PackageEntry for profiler package (after line ~205):
```python
PackageEntry(
    "profiler",
    "rocm-profiler",
    pure_py_package_name="rocm_profiler",
    template_directory="rocm-profiler",
    required=False,
)
```

[...]

PR #6179, opened 2026-06-26
fix: Add headers present in tarball and deb but missing in pip wheel

jonatluu:
### Problem
Issue ID: https://github.com/ROCm/TheRock/issues/6893

When installing the multi-arch ROCm pip SDK six expected public header trees are absent from the pip prefix although they are present in both the deb and tarball channels of the same build. To rectify this missing headers from _rocm_sdk_devel will be added in the pip wheel so that the pip wheel matches the deb and tarball. 

### Solution

So that pip wheel is no longer missing headers when compared to deb and tarball, adding the following headers:
``` 
composable-kernel: ck, ck_tile
libhipcxx: libhipcxx, cuda, nv
rocprofiler-systems: rocprofiler-systems
```
The behavior of py_packaging.py and build_python_packages.py will be modified so that these headers will be included in pip wheel

### Test Plan

1) Verify that headers (composable-kernel, libhipcxx, rocprofiler-systems) below are in deb and tarball 
2) Verify that the headers are not in pip wheel
3) After code change verify that the headers are in the pip wheel

### Test

PASS: Headers are now present in pip wheel after code change. Previously the directories (ck, ck_tile, cuda, nv, rocprofiler-systems) and the expected headers were missing.
```
(.venv) root@CTR-SVDT-L004:/home/runner/_work/TheRock/TheRock/packages/rocm-sdk-devel/build/lib/rocm_sdk_devel/_rocm_sdk_devel/include# ls **ck**
BUILD_TIME_OPTIMIZATION.md  config.h        host_utility       stream_config.hpp   tensor_operation  version.h.in
README.md                   config.h.in     library            tensor              utility           wrapper
ck.hpp                      filesystem.hpp  problem_transform  tensor_description  version.h

(.venv) root@CTR-SVDT-L004:/home/runner/_work/TheRock/TheRock/packages/rocm-sdk-devel/build/lib/rocm_sdk_devel/_rocm_sdk_devel/include# ls **ck_tile/**
README.md  core  core.hpp  host  host.hpp  ops  ref  remod.py  utility

(.venv) root@CTR-SVDT-L004:/home/runner/_work/TheRock/TheRock/packages/rocm-sdk-devel/build/lib/rocm_sdk_devel/_rocm_sdk_devel/include# ls **cuda**
__atomic       __functional    __memory_resource  __warp   discard_memory   numeric     type_traits
__barrier      __fwd           __numeric          atomic   functional       ptx         version
__bit          __latch         __ptx              barrier  latch            semaphore   warp
__cccl_config  __mdspan        __semaphore        bit      mdspan           std         work_stealing
__cmath        __memcpy_async  __type_traits      cmath    memory_resource  stream_ref

(.venv) root@CTR-SVDT-L004:/home/runner/_work/TheRock/TheRock/packages/rocm-sdk-devel/build/lib/rocm_sdk_devel/_rocm_sdk_devel/include# ls **nv/**
detail  target

(.venv) root@CTR-SVDT-L004:/home/runner/_work/TheRock/TheRock/packages/rocm-sdk-devel/build/lib/rocm_sdk_devel/_rocm_sdk_devel/include# ls **rocprofiler-systems**
categories.h  causal.h  dl  dyninst  types.h  user.h
```
JIRA ID: ROCM-26128


build_tools/_therock_utils/py_packaging.py
```diff
@@ -583,7 +583,6 @@ def populate_devel_files(
         devel_artifact_names.difference_update(
             {
                 "rocprofiler-compute",
-                "rocprofiler-systems",
             }
         )
```
ScottTodd, line 590: The code and comments now disagree here.

ScottTodd (commented):
I'll defer to the reviewers from https://github.com/ROCm/TheRock/pull/4518 or @erman-gurses from https://github.com/ROCm/TheRock/issues/3573 here. This introduces overlapping files between two python packages which _might_ be okay, need to hear from what other rocprofiler-systems contributors/maintainers expect.

ScottTodd:
(also please write a more complete PR description, "add missing headers" doesn't explain the context in enough detail)

erman-gurses:
Can you reproduce the issue so I can clearly see what was the problem. Also please add a test plan to explain what are we testing against, etc. Some more details please.

build_tools/_therock_utils/py_packaging.py
```diff
@@ -577,13 +577,15 @@ def populate_devel_files(
...
+        # Explicitly removing rocprofiler-compute here ensures correct package
+        # ownership. Note: rocprofiler-systems is now included in the devel package
+        #
+        # Package ownership:
+        #   - rocm-profiler → owns profiler runtime (rocprofiler-compute, rocprofiler-systems)
+        #   - rocm-sdk-devel → includes rocprofiler-systems, excludes rocprofiler-compute
         devel_artifact_names.difference_update(
             {
                 "rocprofiler-compute",
-                "rocprofiler-systems",
```
erman-gurses, line 586: Maybe we should  keep rocprofiler-systems excluded here and add a component- or path-scoped way to include only its headers.

profiler.populate_runtime_files() adds rocprofiler-systems to runtime_artifact_names. Once it is no longer removed here, _devel_artifact_filter() accepts every rocprofiler-systems component except those listed in exclude_components. In the kpack-split path, only the test component is excluded.

That means this can include more than include/rocprofiler-systems/**; it may also select its lib and run artifacts. Files already materialized in rocm-profiler would then be added to the devel package as links into _rocm_profiler, even though rocm-sdk-devel does not require that package.

Could we preserve the runtime exclusion and explicitly include only the rocprofiler-systems development/header artifact or the required header path?
@dgaliffiAMD any idea?

build_tools/build_python_packages.py
```diff
@@ -263,6 +263,8 @@ def _run_kpack_split(
             "nlohmann-json",
             "rocshmem",
             "rocjitsu",
+            "composable-kernel",
+            "libhipcxx",
```
erman-gurses, line 267: Could you add an regression test for this packaging behavior in build_tools/tests/py_packaging_test.py?

The test should create minimal dev, lib, and run artifacts and verify that:

include/ck/** and include/ck_tile/** from composable-kernel are included.
include/libhipcxx/**, include/cuda/**, and include/nv/** from libhipcxx are included.
include/rocprofiler-systems/** is included.
Composable Kernel non-development files from its lib or run components are not unintentionally copied into rocm-sdk-devel.
rocprofiler-systems runtime paths such as bin/rocprof-sys-*, lib/librocprof-sys*, and libexec/rocprofiler-systems/** are not included in the devel package.
The devel package does not create links into _rocm_profiler.

Since addl_artifact_names filters by artifact name rather than component, adding composable-kernel may select its dev, lib, run, and doc components, not only the ck and ck_tile headers. If its non-development files are intentionally required in rocm-sdk-devel, could the test document that expected ownership instead?

This would verify both the missing-header fix and the package-ownership boundaries affected by the new artifact inclusions.

erman-gurses (commented):
Please see my comments:

build_tools/build_python_packages.py
marbre, line 267: This is clearly wrong. libhipcxx was fixed with commit 708b3f2 already. This touches _single family-wheels_ which are deprecated and no longer produced, thus it touches the wrong location. Furthermore, it was confirmed internally that CK is excluded on purpose.

marbre (changes requested):
This address an issue that is not an issue. While 708b3f2 fixed libhipcxx already, this touches the wrong code path and tries to add CK which is excluded on purpose.

raramakr:
Closing the PR

PR #6183, opened 2026-06-26
ci(bump-automation): switch rocm-systems TheRock ref updates to ci-env action

araravik-psd:
## Summary

Updates `build_tools/github_actions/bump_automation.py` to bump the TheRock ref in rocm-systems via `.github/actions/ci-env/action.yml`instead of individual workflow files.

## Reason for change

rocm-systems introduced a composite action at `.github/actions/ci-env/action.yml` as the single source of truth for the TheRock git ref, mirroring the pattern already used in rocm-libraries (ROCm/rocm-systems#7887).

The previous approach used `update_ref_in_file()` to overwrite `ref:` lines across 7 individual workflow files. This no longer works because the SHA now lives in `action.yml` under a `value:` field, which requires `update_ci_env_file()` to update correctly.

## Changes

`build_tools/github_actions/bump_automation.py`:

- `ROCM_SYSTEMS_FILES`: replaced 7 individual workflow file paths with the single `.github/actions/ci-env/action.yml`
- `SUBMODULE_CONFIG["rocm-systems"]["updater"]`: changed from `"ref"` to `"ci-env"`

## Related

- rocm-systems PR introducing the composite action: ROCm/rocm-systems#7887

### DIFF
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -144,11 +144,20 @@ def update_ref_in_file(file_path, new_sha):
 def update_ci_env_file(file_path, new_sha):
     """Update the therock-ref value in a ci-env composite action file.
 
-    Matches:
-      therock-ref:
-        description: ...
-        value: "<old_sha>" # <date> commit
+    Handles two formats:
+
+    1. rocm-libraries style — SHA as a static output value:
+         therock-ref:
+           description: ...
+           value: "<old_sha>" # <date> commit
+
+    2. rocm-systems style — SHA as a fallback inside a bash run script:
+         THEROCK_REF="<old_sha>" # <date>
     """
+    import re
+
+    date = datetime.now(timezone.utc).strftime("%Y-%m-%d")
+
     with open(file_path, "r") as f:
         lines = f.readlines()
 
@@ -156,6 +165,8 @@ def update_ci_env_file(file_path, new_sha):
     in_therock_ref = False
     for line in lines:
         stripped = line.strip()
+
+        # Format 1: static value: field under therock-ref: key
         if stripped == "therock-ref:":
             in_therock_ref = True
             updated_lines.append(line)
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/bump_automation.py
```diff
@@ -29,7 +23,7 @@
     "rocm-systems": {
         "repo": "ROCm/rocm-systems",
         "files": ROCM_SYSTEMS_FILES,
-        "updater": "ref",
+        "updater": "ci-env",
```
ScottTodd, line 26: > The previous approach used `update_ref_in_file()` to overwrite `ref:` lines across 7 individual workflow files. This no longer works because the SHA now lives in `action.yml` under a `value:` field, which requires `update_ci_env_file()` to update correctly.

https://github.com/ROCm/rocm-systems/pull/7887 currently does _not_ have the sha under a `value:` field like
```yml
outputs:
  therock-ref:
    description: "TheRock commit hash to use"
    value: "f1e3fc2311fde45bf3e17b0e3349200be9caa4b4" # 2026-06-23
```
it has it here:
```yml
      run: |
        # Use base_ref for PRs; ref_name for push/dispatch events.
        BASE_BRANCH="${{ github.base_ref || github.ref_name }}"
        if [[ "${BASE_BRANCH}" =~ ^release/therock-.+$ ]]; then
          # PR targets a release branch — check out the same branch in TheRock.
          THEROCK_REF="${BASE_BRANCH}"
        else
          # Default: use the pinned commit SHA (updated by bump automation).
          THEROCK_REF="f1e3fc2311fde45bf3e17b0e3349200be9caa4b4" # 2026-06-23
        fi
```

geomin12 (commented):
lgtm after https://github.com/ROCm/rocm-systems/pull/7887 is merged

marbre:
We agreed to migrate to multi-arch first. Removing myself as a reviewer as long as this is under development.

PR #6203, opened 2026-06-29
ci(debug-tools): add amd-llvm_dev to rocgdb artifact set and deduplicate

### LINKED ISSUE #6204: ci(debug-tools): amd-llvm_dev missing from ROCgdb artifact set breaks OpenMP offload tests
## Problem

ROCgdb's OpenMP GPU offload tests ([ROCm/ROCgdb#134](https://github.com/ROCm/ROCgdb/pull/134)) fail in CI with:

```
gdb compile failed, ld.lld: error: unable to find library -lompdevice
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)
/__w/ROCgdb/ROCgdb/build/lib/llvm/bin/clang-linker-wrapper: error: 'clang' failed
flang-23: error: linker command failed with exit code 1 (use -v to see invocation)
```

The linker cannot find `-lompdevice` because `amd-llvm_dev` is not included in the `--debug-tools` artifact set fetched by `install_rocm_from_artifacts.py`. The `ompdevice` static library ships inside `amd-llvm_dev` and is required at link time when building OpenMP GPU offload programs with `-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa`.

## Fix

Add `amd-llvm_dev` to the `--debug-tools` artifact set in `build_tools/install_rocm_from_artifacts.py`, and deduplicate the scattered `amd-llvm_dev` appends across `--rocdecode`, `--rocjpeg`, and `--libhipcxx` into a single `needs_amd_llvm_dev` flag.

lumachad:
ROCgdb's OpenMP GPU offload tests ([ROCm/ROCgdb#134](https://github.com/ROCm/ROCgdb/pull/134))
fail in CI because `amd-llvm_dev` is missing from the `--debug-tools` artifact set:

```
gdb compile failed, ld.lld: error: unable to find library -lompdevice
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)
```

`-lompdevice` ships inside `amd-llvm_dev` and is required when building OpenMP GPU
offload programs with `-fopenmp -fopenmp-targets=amdgcn-amd-amdhsa`.

Closes #6204

### DIFF
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -421,6 +421,8 @@ def retrieve_artifacts_by_run_id(args):
             extra_artifacts.append("mpfr")
             extra_artifacts.append("expat")
             extra_artifacts.append("ncurses")
+            # rocgdb tests require amd-llvm_dev for compiler headers/tools.
+            argv.append("amd-llvm_dev")
         if args.fft:
             extra_artifacts.append("fft")
             extra_artifacts.append("fftw3")
--- a/build_tools/tests/install_rocm_from_artifacts_test.py
+++ b/build_tools/tests/install_rocm_from_artifacts_test.py
@@ -4,6 +4,7 @@
 
 """Unit tests for install_rocm_from_artifacts.py."""
 
+import argparse
 from datetime import datetime
 from pathlib import Path
 import os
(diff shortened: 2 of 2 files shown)


lumachad:
ROCgdb is adding OMP tests and those require the amd-llvm_dev artifact that contains ompdevice.a: https://github.com/ROCm/ROCgdb/pull/134

marbre (commented):
While this PR aims to not add duplicate flags, aren't we doing this anyway for things like `base_dev`? I see the reason not to add redundancy but we should either refactor for all or keep it simple and maybe just add `amd-llvm_dev` where needed until we get to a more complete refactoring (or until we rewrite the entire script which we probably should do anyway).

lumachad:
> While this PR aims to not add duplicate flags, aren't we doing this anyway for things like `base_dev`? I see the reason not to add redundancy but we should either refactor for all or keep it simple and maybe just add `amd-llvm_dev` where needed until we get to a more complete refactoring (or until we rewrite the entire script which we probably should do anyway).

Reasonable point. Ideally we'd fetch all of these automatically based on the artifact relationships. In that sense, I think refactoring to remove other duplicated bits may only be a temporary step until automatic artifact fetching happens.

lumachad:
Simplified code and unit tests (only rocgdb amd_llvm_dev validation).

marbre (approved):
LGTM!

lumachad:
Thanks. Merging.

PR #6117, opened 2026-06-24
Add optional JAX release build toggle

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/5634 and https://github.com/ROCm/TheRock/issues/6218.

Follow-up to these PRs:
* https://github.com/ROCm/TheRock/pull/4366
* https://github.com/ROCm/TheRock/pull/5727

This brings PyTorch and JAX back into alignment with optional triggering, a script to define the job matrix, and the ability to trigger builds for a single python version at a time to save CI resources.

## Technical Details

> [!IMPORTANT]
> The `python_version` input will need to be added to https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release_linux_jax_wheels.yml, like with https://github.com/ROCm/rockrel/pull/54 and https://github.com/ROCm/rockrel/pull/55.
>
> I've prepared this PR for that: https://github.com/ROCm/rockrel/pull/70.

## Test Plan

* New unit tests
* Trigger a dev release with `multi_arch_release_linux.yml`, setting `build_jax` to `false`
  * https://github.com/ROCm/TheRock/actions/runs/28469583076
* Trigger `multi_arch_release_linux_jax_wheels.yml` for a single `python_version`
  * https://github.com/ROCm/TheRock/actions/runs/28469829633

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6082, opened 2026-06-24
Use configure_pytorch_release_matrix.py to drive the CI job matrix

ScottTodd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/6030. On https://github.com/ROCm/TheRock/pull/5744 we added gfx1250 support to ROCm but support is not yet there in PyTorch across all supported release branches, so we filtered it from the GPU targets list _for release workflows_. We did not have a way to filter _for CI workflows_, so PRs that used the `ci:run-all-archs` label like bump PRs (example: https://github.com/ROCm/TheRock/pull/6020) were failing to build pytorch.

Now we'll get the same filtering across CI and releases.

## Technical Details

This is like https://github.com/ROCm/TheRock/pull/5942 which added `build_tools/github_actions/configure_rocm_python_test_matrix.py`. Dynamically generating a job matrix using a script will give us greater flexibility (finer-grained inclusions/exclusions, opt-in to expand the matrix, etc.), at the cost of some code complexity and indirection.

> [!NOTE]
> CI workflows use `workflow_call` to run pytorch jobs while release workflows use `workflow_dispatch`. This generates a new matrix for CI but _continues to use the same workflow_dispatch inputs for releases_, to keep the workflows easy to dispatch directly (otherwise you'd need to type out a JSON matrix to trigger). There is potential there for the generated/used matrix to be inconsistent across that boundary.

## Test Plan

* updated unit tests
* `multi_arch_ci.yml` run using prebuilt artifacts: https://github.com/ROCm/TheRock/actions/runs/28060884840/job/83076620803
* `multi_arch_ci_asan.yml` run using prebuilt artifacts: https://github.com/ROCm/TheRock/actions/runs/28202142368  image
* `multi_arch_release.yml` run: https://github.com/ROCm/TheRock/actions/runs/28187486907 --> triggered https://github.com/ROCm/TheRock/actions/runs/28196069145  image
* `multi_arch_ci.yml` on this PR with the opt-in gfx125x label: https://github.com/ROCm/TheRock/actions/runs/28200714865?pr=6082  image

## Test Result

* `gfx125X-dcgpu` / `gfx125x` is filtered from pytorch builds except for the release/2.11 branch
* a "build-pytorch" section is added to the "Multi-Arch CI Configuration" summary, listing which matrix jobs are going to be added to the build graph

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6213, opened 2026-06-29
fix(jax): skip JAX wheel builds/tests on ASan builds

### LINKED ISSUE #6210: [Issue] Jax test fails: `ASan runtime not first in library list `
## Context
- **Workflow:** `multi_arch_release_linux_jax_wheels.yml` (ROCm/rockrel)
- **Failing run:** [Run 28355352077 (2026-06-29)](https://github.com/ROCm/rockrel/actions/runs/28355352077)
- **Platform:** Linux
- **Impacted Python versions:** 3.11, 3.12, 3.13, 3.14
- **JAX refs:** rocm-jaxlib-v0.9.1, rocm-jaxlib-v0.10.0
- **Suspected Root cause commit:** [TheRock `66e7f19`](https://github.com/ROCm/TheRock/commit/66e7f19298af519333a32133788325562e46e787)

## Error

All 8 test matrix jobs fail at the first sanity check before any pytest runs:

```text
+ python -c 'import jax; print(jax.local_devices())'
==747==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
##[error]Process completed with exit code 1.
```
## Suspected Root Cause

`trigger_release_jax_wheels` in [multi_arch_release_linux.yml lines 252–272](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux.yml#L252-L272) is missing a `build_variant != 'asan'` guard. Introduced by commit [66e7f19298](https://github.com/ROCm/TheRock/commit/66e7f19298)

All 8 `Test JAX Wheels` jobs fail at `python -c "import jax; print(jax.local_devices())"`. The JAX plugin wheels were built against an ASan-instrumented ROCm tarball produced by the ASAN nightly, so the plugin `.so` files are linked against `libclang_rt.asan.so`. At test time, ASan requires its runtime to be the first library loaded; it is not, causing an immediate abort.

The proximate cause: `trigger_release_jax_wheels` in [multi_arch_release_linux.yml](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux.yml#L252) has no `if: ${{ inputs.build_variant != 'asan' }}` guard, so the ASAN nightly dispatches JAX wheel builds unconditionally.

## AI  Suggested Fix

Add `if: ${{ inputs.build_variant != 'asan' }}` to the `trigger_release_jax_wheels` job in [multi_arch_release_linux.yml](https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux.yml).

erman-gurses:
## Description
Fixes: https://github.com/ROCm/TheRock/issues/6210
JAX test jobs are failing with:

> ASan runtime does not come first in initial library list

This happens because ASan-instrumented ROCm tarballs are being used to build JAX wheels, but the test environment cannot load the ASan runtime correctly.

### Changes
- Added `if: ${{ inputs.build_variant != 'asan' }}` guard to the `trigger_release_jax_wheels` job in `multi_arch_release_linux.yml`.

This prevents dispatching JAX builds for ASan nightlies while keeping normal builds unaffected.

### Testing
- Verified that normal (non-ASan) builds continue to work.
- ASan builds will no longer trigger JAX wheel jobs (as expected).

---

### DIFF
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -253,6 +253,7 @@ jobs:
   trigger_release_jax_wheels:
     needs: [build_artifacts, build_tarballs, build_python_packages]
     name: Trigger Release JAX Wheels
+    if: ${{ inputs.build_variant != 'asan' }}
     runs-on: ubuntu-24.04
     permissions:
       actions: write


erman-gurses:
@geomin12 , can you please let me know how to test this from ASAN side?

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -253,6 +253,7 @@ jobs:
   trigger_release_jax_wheels:
     needs: [build_artifacts, build_tarballs, build_python_packages]
     name: Trigger Release JAX Wheels
+    if: ${{ inputs.build_variant != 'asan' }}
```
ScottTodd, line 256: I have some open PRs that will support a different solution here:

* https://github.com/ROCm/TheRock/pull/6082
* https://github.com/ROCm/TheRock/pull/6117

`multi_arch_release_asan.yml` will set `build_pytorch: false` and `build_jax: false` which will propagate through all the workflows.

We shouldn't have something different here for jax than we have for pytorch (see line 234 just above)

ScottTodd (commented):
The asan release workflows have deeper issues than just this. Let's disable the nightly/scheduled release jobs before trying to patch like this.

erman-gurses:
> Let's disable the nightly/scheduled release jobs before trying to

Sure @ScottTodd , This will drop then since we will solve it with toggle.

erman-gurses:
Closing this since the other solutions are in progress.

PR #6215, opened 2026-06-29
test: Add pytest test filter standardization for tensilelite

dileepr1:
## Motivation

Extends Test Filter Standardization (RFC0010) to the **tensilelite** test job.
Unlike the gtest/ctest components that go through `test_runner.py`, tensilelite
tests are Python/pytest, so this adds a pytest analog that drives the same
`quick`/`standard`/`comprehensive`/`full` tiers from a `test_categories.yaml`.

Paired rocm-libraries PR: ROCm/rocm-libraries#8616 (branch
`users/dravindr/tf_tensilelite`) — ships
`projects/hipblaslt/tensilelite/test_categories.yaml` and installs it to
`share/hipblaslt/tensilelite/`.

This PR is part of the activity [ROCM-20735](https://amd-hub.atlassian.net/browse/ROCM-20735).

## Changes

- **`build_tools/github_actions/test_executable_scripts/pytest_runner.py`** (new) —
  generic pytest analog of `test_runner.py`. Reads `test_categories.yaml` from
  the installed component tree and for the requested category:
  - collects across the category's `test_paths` (for tensilelite: `rocisa_tests`
    + `Tensile/Tests/unit`),
  - applies the pytest marker include/exclude expression plus `skip-gfxNNNN`
    GPU-arch filtering (markers registered in `tensilelite/pytest.ini`),
  - shards with GTest-style modulo arithmetic (`SHARD_INDEX`/`TOTAL_SHARDS`),
  - runs pytest from `share/hipblaslt/tensilelite/`, replicating the env setup
    (PYTHONPATH/ROCM_PATH/LD_LIBRARY_PATH/PATH) from the legacy
    `test_tensilelite.py`. `--timeout`/`-n` are added only when
    `pytest-timeout`/`pytest-xdist` are importable.
- **`build_tools/github_actions/fetch_test_configurations.py`** — tensilelite
  `test_script` -> `pytest_runner.py` with per-category sharding
  (quick/standard/comprehensive/full = 1/2/4/4), plus generic support for
  dict-shaped `total_shards_dict[platform][test_type]`.

The category mapping is derived from the legacy `test_tensilelite.py`
(comprehensive/full reproduce its rocisa + Tensile-unit set; quick/standard run
the same minus GPU-hardware tests). The legacy `test_tensilelite.py` is kept as
a fallback for one release.

## Test Plan

- Validated locally: `py_compile`, `black`, and an end-to-end run of
  `pytest_runner.py` against a synthetic install tree (path resolution, marker
  include/exclude, hyphenated `skip-gfxNNNN` filtering, sharding, exit-code
  propagation).
- CI on the paired rocm-libraries PR #8616 exercises this runner against the
  packaged tensilelite test artifacts for each category.

Test Run: https://github.com/ROCm/TheRock/actions/runs/30459622504/job/90623144758?pr=6215

[ROCM-20735]: https://amd-hub.atlassian.net/browse/ROCM-20735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,289 @@
...
+    if num_workers and num_workers > 1 and find_spec("xdist") is not None:
+        cmd.append(f"--numprocesses={num_workers}")
+    elif num_workers and num_workers > 1:
+        logging.warning("pytest-xdist not installed; running serially")
+
+    logging.info(
+        f"Running pytest with {len(test_ids)} tests "
+        f"(timeout={timeout}s, workers={num_workers}) in {cwd}"
+    )
+    return subprocess.run(cmd, cwd=str(cwd), env=env, check=False).returncode
```
reboss, line 177: I'm not sure this supports sharding the way you think.  Right now, it seems to only support sharding across multiple processes on the same system

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -217,11 +217,16 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+        "timeout_minutes": 480,  # 8 hours for full category
+        "test_script": f"python {_get_script_path('pytest_runner.py')}",
         "platform": ["linux"],
         "total_shards_dict": {
-            "linux": 1,
+            "linux": {
+                "quick": 1,
+                "standard": 2,
+                "comprehensive": 4,
+                "full": 4,
```
reboss: Pytests are usually fairly quick to run, is there really a need for sharding them?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -866,13 +871,27 @@ def run():
...
+
             job_config_data["shard_arr"] = [i + 1 for i in range(total_shards)]
             job_config_data["total_shards"] = total_shards
 
             # If the test type is quick tests, we only need one shard for the test job
             # Note: Benchmarks always use test_type="full" but have total_shards=1 anyway
-            if test_type == "quick":
+            # Skip this override when the component already declares per-category
+            # shard counts (it has set quick explicitly above).
+            if test_type == "quick" and not isinstance(shards_config, dict):
```
reboss: I'd revert the changes to this file, and then revert the changes to the `total_shards_dict` lines above so that tensilelite is consistent with other projects.  i.e.
```python
        "total_shards_dict": {
            "linux": 1,
         }
```

instead of:
```python
        "total_shards_dict": {
            "linux": {
                "quick": 1,
                "standard": 2,
                "comprehensive": 4,
                "full": 4,
            }
        }
```

reboss (approved):

reboss (changes requested):
Sorry I meant to request changes, not approve

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Removed the sharding entirely. pytest_runner.py now invokes pytest directly on the category's test_paths with the marker expression; no more collect-only + modulo distribution or --numprocesses shard handling.

build_tools/github_actions/fetch_test_configurations.py
dileepr1: Agreed - dropped sharding for tensilelite. It now runs a single shard like the other projects.

build_tools/github_actions/fetch_test_configurations.py
dileepr1: Done. Reverted run() back to the original int-only shard logic and set tensilelite total_shards_dict to {"linux": 1} so it's consistent with other projects.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -217,8 +217,8 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
     "tensilelite": {
         "job_name": "tensilelite",
         "fetch_artifact_args": "--blas --tests",
-        "timeout_minutes": 15,
-        "test_script": f"python {_get_script_path('test_tensilelite.py')}",
+        "timeout_minutes": 480,  # 8 hours for full category
```
archana-ramalingam: Have we run the tests to validate the 8 hrs?

dileepr1: Thanks for pointing this out. I see tensilelite finishes within 10m, reverting it back.

archana-ramalingam (approved):

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,276 @@
...
+        for arg in (category_config.get("pytest_args", []) or [])
+    ]
+
+    exec_settings = config.get("execution_settings", {})
+    timeout = exec_settings.get("category_timeouts", {}).get(TEST_TYPE)
+    # parallel_workers may be overridden per category (e.g. GPU GEMM tests want
+    # more xdist workers than the default), falling back to the global setting.
+    num_workers = category_config.get(
+        "parallel_workers", exec_settings.get("parallel_workers", 1)
+    )
```
archana-ramalingam: Can we add env var overrides for num_workers and timeout in `pytest_runner.py`? 

Currently the only way to change these is modifying `test_categories.yaml` and rebuilding.
```
timeout = int(os.getenv("PYTEST_TEST_TIMEOUT", 0)) or exec_settings.get("category_timeouts", {}).get(TEST_TYPE)
num_workers = int(os.getenv("PYTEST_NUM_WORKERS", 0)) or category_config.get(
        "parallel_workers", exec_settings.get("parallel_workers", 1)
    )
```

dileepr1: I've added these changes.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+"""
+
+import logging
+import os
+import re
+import subprocess
+import sys
+from importlib.util import find_spec
+from pathlib import Path
+
```
geomin12: nit: no empty space

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+# Map job name -> install location, relative to the ROCm prefix (THEROCK_BIN_DIR
+# parent). Components live under share/ rather than bin/ because they are Python
+# packages + pytest modules, not native test executables.
+INSTALLED_COMPONENTS = {
+    "tensilelite": "share/hipblaslt/tensilelite",
+}
+
+
+def get_env_value(name, default=None):
+    return os.getenv(name, default)
```
geomin12: could we just use os.getenv? no need for this function as it just wraps function

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+    if not raw:
+        return 0
+    try:
+        value = int(raw)
+    except ValueError:
+        logging.warning(f"Ignoring non-integer {name}={raw!r}")
+        return 0
+    if value < 0:
+        logging.warning(f"Ignoring negative {name}={raw!r}")
+        return 0
```
geomin12, line 76: this is tricky, because in python, int() will just result in shorterning `1.52423` -> `1`

and negative is still valid i believe? i would double check this to make sure it works proper

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def load_test_categories_yaml(yaml_path):
+    try:
+        with open(yaml_path, "r") as f:
+            config = yaml.safe_load(f)
+        logging.info(f"Loaded test categories from {yaml_path}")
+        return config
+    except FileNotFoundError:
+        logging.error(f"test_categories.yaml not found at {yaml_path}")
```
geomin12: instead of sys.exit(1), could we just raise Exception? 

i feel this is more error-prone, as you could add an error log but forget to exit. let's make it just one clean op!

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+        sys.exit(1)
+    except yaml.YAMLError as e:
+        logging.error(f"Invalid YAML syntax in {yaml_path}: {e}")
+        sys.exit(1)
+
+
+def extract_gpu_arch(amdgpu_families):
```
geomin12, line 103: oh i think `test_runner.py` has some shared functions. can we use those?

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    Build a pytest -m expression from a category's marker config plus the
+    GPU-arch skip markers registered in pytest.ini (skip-gfxXXXX).
+    """
+    include = category_config.get("pytest_markers", []) or []
+    exclude = category_config.get("exclude_markers", []) or []
+
+    terms = []
+    if include:
+        terms.append("(" + " or ".join(include) + ")")
```
geomin12, line 126: i think join is 
```suggestion
        terms.append("(" + " or "".join(include) + ")")
```
?

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+        cmd.append(f"--numprocesses={num_workers}")
+    elif num_workers and num_workers > 1:
+        logging.warning("pytest-xdist not installed; running serially")
+
+    logging.info(
+        f"Running pytest on {existing} with markers: {marker_expr or 'none'}, "
+        f"extra args: {extra_args or 'none'} "
+        f"(timeout={timeout}s, workers={num_workers}) in {cwd}"
+    )
+    return subprocess.run(cmd, cwd=str(cwd), env=env, check=False).returncode
```
geomin12, line 177: let's just do subprocess.check_call here. i don't see a point in doing sys.exit at the end

build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -0,0 +1,303 @@
...
+                existing_path,
+            ],
+        )
+    )
+    return env
+
+
+if __name__ == "__main__":
+    TEST_COMPONENT_NAME = get_env_value("TEST_COMPONENT")
+    TEST_TYPE = get_env_value("TEST_TYPE", "quick")
```
geomin12: let's do standard instead of quick

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Done - removed the wrapper and call os.getenv directly.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Done.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Verified it's actually safe here, since the input values are strings. 
Floats will get truncated as you mentioned above, but that is not the case here.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Agreed. Added a _fail(message) helper.
Kept it as sys.exit rather than a bare raise so we exit cleanly without dumping a traceback for expected config errors.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Test_runner.py had the function added to the shared utils script, but was later removed upon request. 
So as it stands, pulling it directly from test_runner.py won't be import safe.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Thanks, I confirmed this works.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: subprocess.run(check=False).returncode + sys.exit(code) propagates pytest's exact exit code (1 = failures, 5 = no tests collected, etc.) so CI reports the real result, and it avoids printing a Python traceback for an ordinary test failure.
I've kept this and added a short comment explaining the intent.
Please let me know if a change is needed.

build_tools/github_actions/test_executable_scripts/pytest_runner.py
dileepr1: Kept it as quick to maintain parity with test_runner.py, however we always expect TEST_TYPE to have a value.
Please let me know if a change is needed.

madkasul (approved):

PR #6260, opened 2026-06-30
[release/7.14] bump submodule rocm-systems 06-30-2026

araravik-psd:
Creating this bump PR to cherry pick the below fixes into the release branch

https://github.com/ROCm/rocm-systems/pull/7976 - RCCL change
https://github.com/ROCm/rocm-systems/pull/7937 - Kpack change
https://github.com/ROCm/rocm-systems/pull/7867 - workflow related pinning commit for release


ScottTodd (approved):
Approving these cherrypicks for devops. We could wait for CI here or just merge to get the RC1 build started earlier. We have pretty high confidence in these changes and want a release candidate build queued up for QA soon.

* If we wait and CI fails, we won't have a candidate to consider
* If we wait and CI succceeds, the candidate will be delayed
* If we merge early and CI fails, the candidate can be rejected and replaced with a newer candidate
* If we merge early and CI succeeds, the candidate will be ready faster

These changes are just in RCCL and kpack, so they'll only affect the comm-libs stage and artifact splitting (math-libs stage, comm-libs stage). We should expect build signal for the comm-libs stage and math-libs stage as well as _some_ test signal for math-libs. The most coverage we'll get will actually be from pytorch builds and tests, most of which is in release workflows and not CI workflows.

I'd vote to merge after some of the build stages complete without waiting for CI to fully finish.

astyrrian1 (approved):

PR #6263, opened 2026-06-30
fix(ci): allow workflow_dispatch to test nightly_check_only architectures

geomin12:
Previously, architectures with nightly_check_only_for_family flag had their test runners disabled for all non-scheduled triggers. This change allows workflow_dispatch triggers to also enable test runners for these architectures, enabling manual testing of nightly-only archs.

works here: https://github.com/ROCm/TheRock/actions/runs/28465742475 with setting here: https://github.com/ROCm/therock-ci-config/blob/89a7b32a40268208bd7fa38d16775417f5b67932/runner-config.json#L71

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -935,15 +935,15 @@ def _expand_build_config_for_platform(
                 f"disabling tests for quick test run"
             )
 
-        # If nightly_check_only_for_family is set for schedule runs only
-        if (
-            platform_info.get("nightly_check_only_for_family", False)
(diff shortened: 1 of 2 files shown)


rahulc-gh (approved):
LGTM. Add more details in the PR as well. Which is the temp workflow disapatch trigger you plan to use and the issue you faced as motivation of having this PR for testing specifc arch

geomin12:
> LGTM. Add more details in the PR as well. Which is the temp workflow disapatch trigger you plan to use and the issue you faced as motivation of having this PR for testing specifc arch

i'll these notes in the necessary workflows in rocm-libraries and rocm-systems!

PR #4813, opened 2026-04-24
Pin commit SHA in setup so all release jobs use the same commit

ScottTodd:
## Motivation

As reported at https://github.com/ROCm/rockrel/pull/35#discussion_r3135281106, the new https://github.com/ROCm/rockrel/blob/main/.github/workflows/multi_arch_release.yml workflow (in rockrel, not TheRock) uses a floating ref when triggered by `workflow_dispatch` or `schedule` events. This has the expected behavior for the first job triggered, but sub-jobs _also_ use a floating ref, and if they run hours later this is likely to be a different commit (across artifact build stages, across packaging jobs, etc.).

Progress on:
* https://github.com/ROCm/TheRock/issues/1236
* https://github.com/ROCm/TheRock/issues/3334

## Technical Details

For https://github.com/ROCm/TheRock/issues/1236, we can extend the setup job to also choose commits for PyTorch release branches, JAX release branches, etc.

## Test Plan

Test run in my fork: https://github.com/ScottTodd/TheRock/actions/runs/24906712749
1. Setup:
    ```
    Run actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
      with:
        repository: ScottTodd/TheRock
    ```
2. Build job:
    ```
    Run actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
      with:
        repository: ScottTodd/TheRock
        ref: 260481ba880ed2dd8135410e5c8b31365a6475ee
    ```

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6264, opened 2026-06-30
feat: Adding Artifact reuse and stage skip functionality to multi-arch CI

ammallya:
## Motivation
Issue ID [Issue 3399](https://github.com/ROCm/theROCK/issues/3399)

TheROCK already has the building blocks for stage-aware prebuilt-artifact reuse:
- `stage_impact.py` computes which build stages a changed-file set affects 
- `baseline_runs.py` selects a healthy, commit-compatible baseline workflow run and verifies it contains required artifacts
- the `prebuilt_stages` skip mechanism is plumbed through the multi-arch workflows  — each stage already guards on `!contains(inputs.prebuilt_stages, 'stage-name')`

This PR adds skip capability with a **report-only default** so we can watch the decisions across real runs before any stage is actually skipped, and flip a single switch (`STAGE_REUSE_MODE=skip-stage`) to turn skipping on.

## Technical Details

New module `build_tools/github_actions/stage_reuse_decision.py` that reuses the existing tooling (no new impact or baseline logic) and applies **two gates** before a stage is reported reusable:

1. **Impact gate** (`stage_impact.analyze_stage_impact`) — the change does not affect the stage → it is a *candidate* for reuse.
2. **Availability gate** (`baseline_runs.select_baseline_run`) — a single healthy, commit-compatible baseline run actually contains the artifacts that stage would produce. This runs **even in dry-run**, so the report reflects what could really be skipped, not just what is logically unaffected.

Two-way `STAGE_REUSE_MODE` switch:
- **`dry-run`** (default) — runs both gates and prints, per stage, `[STAGE-REUSE] stage 'X' ... WOULD be skipped (dry-run)` to the console and the step summary (including the baseline run id it would copy from), but returns **no** stages, so `prebuilt_stages` is unchanged and every stage still builds.
- **`skip-stage`** — merges only the *confirmed-available* stages into the per-stage decisions so the orchestrator copies their artifacts and skips the build.

Safety:

- The manual `prebuilt_stages` / `baseline_run_id` workflow_dispatch inputs are never touched — they keep working unchanged in every mode.
- Conservative fallbacks reuse nothing: full-CI triggers (e.g. `build_tools/`, workflow, or topology changes), runs without a changed-file list (schedule / dispatch), no baseline found, or any baseline-lookup error.
- Reporting and baseline lookup failures never fail CI configuration.

## Test Plan

- New unit tests in `build_tools/github_actions/tests/stage_reuse_decision_test.py` using an injected fake topology and an injected baseline selector:
  - mode parsing + default
  - dry-run verifies artifacts present; unaffected-but-artifacts-missing → rebuild; partial-family availability → rebuild
  - no baseline found → rebuild; baseline lookup error is safe
  - enforce applies only available stages
  - no-diff is conservative; step summary shows baseline + availability
- Existing `stage_impact_test.py`, `baseline_runs_test.py`, and `configure_multi_arch_ci_test.py` continue to pass

## Test Result
```
 python -m pytest build_tools/github_actions/tests/stage_reuse_decision_test.py -v
======================================================================================================================== test session starts =========================================================================================================================
platform linux -- Python 3.12.3, pytest-8.4.2, pluggy-1.6.0 -- /home/ergurses/TheRock/.venv/bin/python
cachedir: .pytest_cache
rootdir: /home/ergurses/TheRock/build_tools
configfile: pyproject.toml
plugins: cov-7.1.0, check-2.8.0
collected 23 items                                                                                                                                                                                                                                                   

build_tools/github_actions/tests/stage_reuse_decision_test.py::ModeParsingTest::test_default_is_dry_run PASSED                                                                                                                                                 [  4%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::ModeParsingTest::test_explicit_modes PASSED                                                                                                                                                     [  8%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_baseline_lookup_error_is_safe PASSED                                                                                                                                 [ 13%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_dry_run_unaffected_but_artifacts_missing_rebuilds PASSED                                                                                                             [ 17%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_dry_run_verifies_artifacts_present PASSED                                                                                                                            [ 21%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_enforce_applies_only_available_stages PASSED                                                                                                                         [ 26%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_no_baseline_found_rebuilds_candidates PASSED                                                                                                                         [ 30%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::AvailabilityGateTest::test_partial_family_availability_rebuilds PASSED                                                                                                                          [ 34%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::GuardrailTest::test_full_ci_trigger_skips_baseline_query PASSED                                                                                                                                 [ 39%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::GuardrailTest::test_no_diff_is_conservative PASSED                                                                                                                                              [ 43%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::GuardrailTest::test_step_summary_shows_baseline_and_availability PASSED                                                                                                                         [ 47%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::DefaultBaselineSelectorTest::test_empty_history_disables_commit_rule PASSED                                                                                                                     [ 52%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::DefaultBaselineSelectorTest::test_history_fetch_error_disables_commit_rule PASSED                                                                                                               [ 56%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::DefaultBaselineSelectorTest::test_history_is_fetched_and_threaded PASSED                                                                                                                        [ 60%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::DefaultBaselineSelectorTest::test_no_sha_means_no_history_fetch PASSED                                                                                                                          [ 65%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlatformAwareAvailabilityTest::test_no_platforms_selected_disables_auto_reuse PASSED                                                                                                            [ 69%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlatformAwareAvailabilityTest::test_single_platform_default_is_linux PASSED                                                                                                                     [ 73%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlatformAwareAvailabilityTest::test_stage_reused_only_when_available_on_all_platforms PASSED                                                                                                    [ 78%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlatformAwareAvailabilityTest::test_stage_reused_when_present_on_both_platforms PASSED                                                                                                          [ 82%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlanStageReuseTest::test_plan_none_changed_files_is_full_rebuild PASSED                                                                                                                         [ 86%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::PlanStageReuseTest::test_plan_returns_candidates_without_baseline PASSED                                                                                                                        [ 91%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::TargetFamiliesTest::test_always_includes_generic PASSED                                                                                                                                         [ 95%]
build_tools/github_actions/tests/stage_reuse_decision_test.py::TargetFamiliesTest::test_dedupes_and_appends_generic PASSED                                                                                                                                     [100%]

========================================================================================================================= 23 passed in 0.06s =========================================================================================================================
=== Deciding job configuration ===
[STAGE-REUSE] mode=dry-run
[STAGE-REUSE] conservative full rebuild: no stages eligible for reuse.
[STAGE-REUSE]   reason: 'build_tools/github_actions/configure_multi_arch_ci.py' matches a conservative full-CI trigger
[STAGE-REUSE]   reason: 'build_tools/github_actions/stage_reuse_decision.py' matches a conservative full-CI trigger
[STAGE-REUSE]   reason: 'build_tools/github_actions/tests/stage_reuse_decision_test.py' matches a conservative full-CI trigger
```
Writing job summary:
```
### Stage reuse analysis

- mode: `dry-run`
- full rebuild required: `True`
- baseline run checked: _none_
- unaffected candidates: _none_
- available in baseline: _none_
- applied: _none_
- reasons:
  - 'build_tools/github_actions/configure_multi_arch_ci.py' matches a conservative full-CI trigger
  - 'build_tools/github_actions/stage_reuse_decision.py' matches a conservative full-CI trigger
  - 'build_tools/github_actions/tests/stage_reuse_decision_test.py' matches a conservative full-CI trigger
```
## Submission Checklist

- [] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,308 @@
...
+            github_repository=github_repository,
+            workflow_name=workflow_name,
+            branch=branch,
+            platform=platform or "linux",
+            current_commit_sha=current_commit_sha,
+            ordered_commit_shas=None if current_commit_sha is None else [],
+            max_age_hours=max_age_hours,
+        )
+
+    return _select
```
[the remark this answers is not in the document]
ammallya: Resolved with latest commit

.github/workflows/setup_multi_arch.yml
```diff
@@ -154,6 +158,10 @@ jobs:
           WINDOWS_TEST_LABELS: ${{ inputs.windows_test_labels }}
           PREBUILT_STAGES: ${{ inputs.prebuilt_stages }}
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
+          STAGE_REUSE_MODE: ${{ inputs.stage_reuse_mode }}
+          STAGE_REUSE_CURRENT_SHA: ${{ steps.checkout.outputs.commit }}
+          STAGE_REUSE_MAX_AGE_HOURS: "72"
+          STAGE_REUSE_COMMIT_HISTORY: "50"
```
erman-gurses: Should these become workflow inputs (or shared configuration) instead of hardcoded values? I could see wanting to tune the recency window or commit history depth without modifying the workflow.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -635,14 +635,38 @@ def decide_jobs(
...
+    # which stages WOULD be skipped and apply nothing; in "skip-stage" the
+    # eligible stages are merged into stage_decisions so the orchestrator
+    # skips their builds and copies artifacts instead.
+    from stage_reuse_decision import (
+        StageReuseMode,
+        compute_auto_stage_reuse,
+        render_step_summary,
+    )
+    target_families = _stage_reuse_target_families(ci_inputs)
+    auto = compute_auto_stage_reuse(
```
erman-gurses: `compute_auto_stage_reuse()` now performs `stage-impact analysis`, `baseline selection`, `artifact verification`, and `reporting`. Have we considered separating the planning portion from the reporting/execution plumbing to keep the decision logic independently reusable?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -689,6 +713,12 @@ def decide_jobs(
 # Step 4: Select Targets
 # ---------------------------------------------------------------------------
 
+def _stage_reuse_target_families(ci_inputs: "CIInputs") -> list[str]:
```
erman-gurses: Should this helper live closer to the stage-reuse implementation? It appears to exist solely to prepare inputs for `compute_auto_stage_reuse()`.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -635,14 +635,38 @@ def decide_jobs(
...
+
+    # Behind STAGE_REUSE_MODE: in dry-run we only PRINT
+    # which stages WOULD be skipped and apply nothing; in "skip-stage" the
+    # eligible stages are merged into stage_decisions so the orchestrator
+    # skips their builds and copies artifacts instead.
+    from stage_reuse_decision import (
+        StageReuseMode,
+        compute_auto_stage_reuse,
+        render_step_summary,
+    )
```
erman-gurses: Is there a reason these imports are local rather than module-level? If it's to avoid import cycles, it might be worth adding a short comment.

.github/workflows/setup_multi_arch.yml
erman-gurses: These are becoming part of the stage-reuse interface - will make sense to document this environment contract in the module docstring.

erman-gurses (approved):
Looks good to me overall -  please see my comments.

.github/workflows/multi_arch_ci.yml
```diff
@@ -42,6 +42,10 @@ on:
         type: string
         default: ""
         description: "Workflow run ID to copy prebuilt stage artifacts from; required when prebuilt_stages is set"
+      stage_reuse_mode:
+        type: string
+        default: "dry-run"
+        description: "dry-run (report only) | skip-stage (actually skip unaffected stages)"
```
geomin12: is it supposed to be this? we already have `prebuilt_stages` which handles this... do we mean `dry-run` vs `reuse-stage`?

.github/workflows/setup_multi_arch.yml
geomin12: do we ever see a scenario where it will be different numbers? will this be set by users or by devops code?

.github/workflows/setup_multi_arch.yml
```diff
@@ -154,6 +158,10 @@ jobs:
           WINDOWS_TEST_LABELS: ${{ inputs.windows_test_labels }}
           PREBUILT_STAGES: ${{ inputs.prebuilt_stages }}
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
+          STAGE_REUSE_MODE: ${{ inputs.stage_reuse_mode }}
+          STAGE_REUSE_CURRENT_SHA: ${{ steps.checkout.outputs.commit }}
```
geomin12, line 180: you can actaully get this from ${{ github.sha }} (current sha)

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,351 @@
...
+artifacts (reuse) instead of being rebuilt.
+
+---------------------------------
+This module is wired into CI behind a two-way mode switch so it can be
+observed before it changes anything:
+* ``dry-run``  - DEFAULT. Compute the analysis and PRINT, for each stage that
+                 *would* be skipped, a line to the console + step summary, but
+                 return NO auto stages, so ``prebuilt_stages`` is unchanged and
+                 every stage still builds exactly.
+* ``skip-stage``  - Compute the analysis and actually return the reuse stages so
```
geomin12: mentioned before, but we should rename this to `reuse-stage`. skip-stage is a bit unclear and confusing, assuming it would just skip and not reuse

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,351 @@
...
+    baseline_run_id: Optional[str]
+    baseline_html_url: Optional[str]
+    available_stages: tuple[str, ...]
+    unavailable_stages: tuple[str, ...]
+    applied_reuse_stages: tuple[str, ...]
+    reasons: tuple[str, ...]
+    report_lines: tuple[str, ...] = field(default_factory=tuple)
+
+
+LOG_PREFIX = "[STAGE-REUSE]"
```
geomin12: we could use a python logger to handle thiz

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,351 @@
...
+    # history fetch fails (or returns empty), disable the commit rule (pass both
+    # as None) rather than enabling it with an empty window -- recency and
+    # artifact availability still gate the selection.
+    ordered_commit_shas = None
+    effective_commit_sha = current_commit_sha
+    if current_commit_sha is not None:
+        try:
+            from github_actions_api import gha_query_recent_branch_commits
+
+            ordered_commit_shas = gha_query_recent_branch_commits(
```
geomin12: should we add checks here though to make sure it is a passing build? some artifacts might be doc changes and may not have any artifacts at all

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,351 @@
...
+                ordered_commit_shas = None
+        except Exception as exc:  
+            print(
+                f"{LOG_PREFIX} could not fetch branch history "
+                f"({exc}); skipping commit-compatibility rule."
+            )
+            effective_commit_sha = None
+            ordered_commit_shas = None
+
+    def _select(required):
```
geomin12: redundant function? seems to just be a wrapper

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,351 @@
...
+
+    from stage_impact import analyze_stage_impact
+
+    impact = analyze_stage_impact(
+        changed_inputs=list(changed_files),
+        topology=topology,
+        platform=platform,
+    )
+    candidates = tuple(impact.copy_stages)
+    rebuild = tuple(impact.rebuild_stages)
```
geomin12: is there an ability to add build flags for this too? will be needed for superrepos. not required here but will be needed later

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,18 +681,62 @@ def _determine_test_type(
...
+    # modules (stage_reuse plumbing conceptually depends on this pipeline).
+
+    from stage_reuse_decision import (
+        StageReuseMode,
+        compute_auto_stage_reuse,
+        render_step_summary,
+        log_report,
+        render_step_summary,
+        stage_reuse_target_families,
+    )
```
erman-gurses: `render_step_summary` is listed twice here.

erman-gurses (commented):
Looks all of the comments were addressed or added TODOs - Please see my comment below:

build_tools/github_actions/stage_reuse_decision.py
ammallya: Added a note to add it in future PR's

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -706,19 +777,41 @@ def decide_jobs(
...
+    is known we use it directly; otherwise we fall back to the requested family
+    inputs, defaulting to Linux only.
+    """
+    if targets is not None:
+        platforms = []
+        if targets.linux_families:
+            platforms.append("linux")
+        if targets.windows_families:
+            platforms.append("windows")
+        return platforms or ["linux"]
```
geomin12: i think we should return default nothing

if there's nothing to reuse, we should just rebuild instead of assuming linux is available

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -706,19 +777,41 @@ def decide_jobs(
...
+
+    platforms = []
+    if ci_inputs.linux_amdgpu_families:
+        platforms.append("linux")
+    if ci_inputs.windows_amdgpu_families:
+        platforms.append("windows")
+    return platforms or ["linux"]
```
geomin12: nit but could combine into an `if (targets and targets.linux_families) or ci_inputs.linux_amdgpu_families` ?

also, we should use a set here too. what if targets + ci_inputs are available? it'll return ["linux","linux"]

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -4,50 +4,107 @@
...
+                   and every stage still builds exactly.
+* ``reuse-stage`` - Compute the analysis and actually return the reuse stages so
+                     the orchestrator copies their artifacts and skips the build.
+
+Note: this ``STAGE_REUSE_MODE`` switch drives the *automatic* detection layer.
+It is orthogonal to the ``prebuilt_stages`` workflow input, which is the
+explicit, manual list of stages to reuse and is always honored regardless of
+mode.
+
+Environment contract
```
geomin12: Environment variables?

geomin12 (approved):
overall, lgtm. please address comments before merging

lumachad:
Would you mind please documenting how this is supposed to be used by component CI's?

erman-gurses:
> Would you mind please documenting how this is supposed to be used by component CI's?

Sure, we will add documentation for that.
CC: @ammallya

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: Refactored this to first choose either targets or ci_inputs as the source of truth, then derive the platform list from that single source. That removes the duplicated logic and also eliminates the possibility of duplicate platform entries, so a set is no longer needed.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,18 +681,61 @@ def _determine_test_type(
...
+        target_families=target_families,
+    )
+    log_report(auto)
+
+    try:
+        gha_append_step_summary(render_step_summary(auto))
+    except Exception as exc:  # never fail CI config on a reporting error
+        print(f"[STAGE-REUSE] could not write step summary: {exc}")
+    # Only reuse-stage mode returns non-empty applied_reuse_stages.
+    for stage in auto.applied_reuse_stages:
```
jayhawk-commits: In `reuse-stage` mode, this still keeps `baseline_run_id` set to the manual input even when `auto.applied_reuse_stages` came from a verified baseline. That means the generated build config can contain auto `prebuilt_stages` with an empty `baseline_run_id`. The Linux and Windows copy jobs are gated on both fields being non-empty, so they will not copy artifacts, while the stage build jobs are skipped because the stage is marked prebuilt. Please propagate the selected auto baseline run id, and add a configure-level test for that output shape, when auto reuse applies.

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,625 @@
...
+                selector = baseline_selector
+            elif baseline_selector_factory is not None:
+                selector = baseline_selector_factory(plat)
+            else:
+                selector = _default_baseline_selector(platform=plat)
+            baseline = selector(required)
+        except Exception as exc:
+            baseline_error = str(exc)
+
+        if reported_baseline_run_id is None and baseline is not None:
```
jayhawk-commits: For multi-platform reuse, this currently keeps only the first platform's baseline run ID, while stage availability is computed independently per platform. That means Linux could be verified against run A and Windows against run B, but the workflow can only copy from one `baseline_run_id`. Please either require all reused platforms to resolve to the same baseline run before applying reuse, or carry separate per-platform source run IDs through the build config and copy jobs.

erman-gurses: Thanks - I am taking the conservative route here and requiring all reused platforms to resolve to the same verified baseline run before automatic reuse is applied. I have also added a test to lock in the multi platform mismatch behavior.

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: This is a good catch - I updated the auto-reuse path to propagate the verified `baseline_run_id` selected during stage reuse into the generated build config, while preserving the existing manual `baseline_run_id` behavior for explicit `prebuilt_stages`.
I also added a configure-level test that verifies the propagated `baseline_run_id` and `prebuilt_stages` are emitted together when automatic reuse is applied.

lumachad:
Have we considered the case where we have ASAN build stages and we want to reuse those as well?

marbre:
@ammallya, I only glanced at this but I know @ScottTodd worked on a prototype (https://github.com/ROCm/TheRock/issues/3399#issuecomment-4000020246) and thus he has opinion for sure and it would be valuable to wait for his review before this is merged.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,69 @@ def _determine_test_type(
...
+
+    from stage_reuse_decision import (
+        StageReuseMode,
+        compute_auto_stage_reuse,
+        render_step_summary,
+        log_report,
+        stage_reuse_target_families,
+    )
+
+    build_platforms = _stage_reuse_build_platforms(ci_inputs, targets)
```
jayhawk-commits: This is still running before `select_targets()`, so on normal PR/push/scheduled runs the workflow family inputs are empty. `_stage_reuse_build_platforms(ci_inputs, None)` returns `[]`, which disables automatic reuse, even though target selection later chooses Linux and Windows families. Also, `target_families` is built from those same pre-selection inputs, so simply passing platforms would still verify only `generic` artifacts for normal CI. Can we run the reuse decision after target selection, or pass the selected targets here and derive both platforms and target families from that result?

ammallya: Resolved with latest patch

.github/workflows/multi_arch_ci.yml
```diff
@@ -42,6 +42,22 @@ on:
...
         default: ""
         description: "Workflow run ID to copy prebuilt stage artifacts from; required when prebuilt_stages is set"
+      stage_reuse_mode:
+        type: string
+        default: "dry-run"
+        # This drives the AUTOMATIC stage-reuse detection layer, which is
+        # separate from the manual `prebuilt_stages` input above. dry-run only
+        # reports which unaffected stages could be reused; reuse-stage actually
+        # reuses them (skips the build and copies artifacts from a baseline run).
+        description: "dry-run (report only) | reuse-stage (auto-reuse unaffected stages)"
```
ScottTodd, line 55: Let's use a `choice` input type here instead of `string`, to make it easier to switch without the risk of typos. The code in `setup_multi_arch.yml` uses `workflow_call` instead of `workflow_dispatch` and can keep using `string`.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs

```suggestion
      stage_reuse_mode:
        type: choice
        default: "dry-run"
        options:
          - "dry-run"
          - "reuse-stage"
        # This drives the AUTOMATIC stage-reuse detection layer, which is
        # separate from the manual `prebuilt_stages` input above. dry-run only
        # reports which unaffected stages could be reused; reuse-stage actually
        # reuses them (skips the build and copies artifacts from a baseline run).
        description: "dry-run (report only) | reuse-stage (auto-reuse unaffected stages)"
```

.github/workflows/setup_multi_arch.yml
ScottTodd: It depends on what commit ref you want.

`steps.checkout.outputs.commit` gets the TheRock commit even when running in rockrel (and maybe rocm-libraries / rocm-systems?), see the workflow output:
* https://github.com/ROCm/TheRock/pull/4813
* https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/.github/workflows/setup_multi_arch.yml#L124

For runs in rocm-libraries and rocm-systems, we may want to use baseline runs from those repositories instead of TheRock though.

Ultimately I think the best artifact reuse we'll be able to get will be some query against a database where we say "here are the artifact fingerprints that my current source code produces, fetch me artifacts for any with matching fingerprints from the database", which we'd want to abstract away from the github actions runs API entirely. We wouldn't need to scan commit history or workflow runs with that architecture.

docs/development/stage_reuse.md
```diff
@@ -0,0 +1,119 @@
...
+
+Both feed the same underlying `prebuilt_stages` decision consumed by the build
+jobs; automatic reuse simply *adds* to whatever you set manually.
+
+## How it works
+
+`setup_multi_arch.yml` runs `build_tools/github_actions/configure_multi_arch_ci.py`,
+which calls into `build_tools/github_actions/stage_reuse_decision.py`. For every
+stage the change leaves unaffected, two gates must both pass before the stage is
+reused:
```
ScottTodd: nit: could turn these file names into hyperlinks so they're clickable by humans :)

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Nope - follow the style guide. These imports must be moved to the top. https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

No import cycles here. If there are cycles we should break them instead of use local imports.

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+    """
+    families = list(dict.fromkeys([*linux_amdgpu_families, *windows_amdgpu_families]))
+    if "generic" not in families:
+        families.append("generic")
+    return families
+
+
+def required_artifacts_for_stages(topology, stage_names, target_families):
+    """Artifact/family pairs the given stages produce."""
+    from baseline_runs import RequiredArtifact
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+        return set()
+    return set(baseline.artifact_availability.matched_filenames)
+
+
+def _stage_artifacts_available(
+    topology, stage_name, target_families, available_filenames
+):
+    """True when every artifact this stage produces has an archive present."""
+    from artifact_manager import ARTIFACT_COMPONENTS
+    from _therock_utils.artifact_backend import ARTIFACT_EXTENSIONS
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+    if changed_files is None:
+        return StageReusePlan(
+            candidate_stages=(),
+            rebuild_stages=(),
+            full_rebuild_required=True,
+            reasons=("no changed-file list available",),
+        )
+
+    if topology is None:
+        from _therock_utils.build_topology import get_topology
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+
+        topology = get_topology()
+
+    from stage_impact import analyze_stage_impact
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+                f"{LOG_PREFIX} no build platforms selected; automatic stage reuse disabled.",
+            ),
+        )
+
+    # Resolve the platform set. ``platforms`` wins; fall back to the legacy
+    # single ``platform`` arg; default to linux only when neither is given.
+    resolved_platforms = _resolve_platforms(platforms, platform)
+
+    if topology is None and changed_files is not None:
+        from _therock_utils.build_topology import get_topology
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

(share this between functions, maybe pass `topology` as an argument instead of calling `get_topology()` within multiple function bodies in the same file)

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+
+def _default_baseline_selector(*, platform: str | None) -> BaselineSelector:
+    """Build a selector bound to baseline_runs.select_baseline_run.
+    ``select_baseline_run`` already requires each candidate run to have
+    healthy build jobs (``required_successful_job_name_substrings=("Build",)``)
+    AND to contain all requested artifacts. A run with no artifacts (e.g. a
+    docs-only change) therefore fails the availability gate and is never
+    selected, so no extra "passing build" check is needed here.
+    """
+    from baseline_runs import select_baseline_run
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+    # window every candidate resolves to `unknown` and is rejected, so reuse
+    # never activates. Fetch the real history here. If the SHA is set but the
+    # history fetch fails (or returns empty), disable the commit rule (pass both
+    # as None) rather than enabling it with an empty window -- recency and
+    # artifact availability still gate the selection.
+    ordered_commit_shas = None
+    effective_commit_sha = current_commit_sha
+    if current_commit_sha is not None:
+        try:
+            from github_actions_api import gha_query_recent_branch_commits
```
ScottTodd: Move imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+
+
+def render_step_summary(result: AutoStageReuse) -> str:
+    """Render a GitHub step-summary markdown block for the analysis."""
+    baseline = f"`{result.baseline_run_id}`" if result.baseline_run_id else "_none_"
+    candidates = _format_stage_list(result.candidate_stages)
+    available = _format_stage_list(result.available_stages)
+    applied = _format_stage_list(result.applied_reuse_stages)
+
+    out = ["### Stage reuse analysis", ""]
```
ScottTodd, line 653: Let's change the timing of when this runs. See on https://github.com/ROCm/TheRock/actions/runs/28963586877?pr=6264 how the reuse analysis appears _first_, when it should be either
1. mixed into the existing output, where the "Building all stages from source." text already is (this could have an expandable/collapsible) section for "reuse analysis"
2. appended to the bottom of the current summary

Image

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1253,14 +1343,18 @@ def configure(ci_inputs: CIInputs, git_context: GitContext) -> CIOutputs:
...
     print("\n=== Selecting GPU target families ===")
     targets = select_targets(ci_inputs)
     targets.log()
 
+    print("\n=== Deciding job configuration ===")
+    # Target selection runs first so automatic stage reuse scopes its
+    # platform/family verification to what is actually being built. On normal
+    # PR/push/schedule runs the raw workflow inputs are empty; the resolved
+    # `targets` are what determine the real platforms and families.
+    jobs = decide_jobs(ci_inputs=ci_inputs, git_context=git_context, targets=targets)
```
ScottTodd: I think this is trying too hard to explain the behavior change. That context can go in the commit message or PR title, it doesn't need to be in a code comment here.

It's fine to decide which jobs to run after selecting targets. We don't need to qualify with "normal PR/push/schedule runs" as a "normal" run is not well defined

```suggestion
    # Target selection runs first so automatic stage reuse can scope its
    # prebuilt artifact checks to those targets.
    jobs = decide_jobs(ci_inputs=ci_inputs, git_context=git_context, targets=targets)
```

Watch out for this though:
* https://github.com/ROCm/TheRock/issues/4687

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1253,14 +1343,18 @@ def configure(ci_inputs: CIInputs, git_context: GitContext) -> CIOutputs:
...
-    print("\n=== Deciding job configuration ===")
-    jobs = decide_jobs(ci_inputs=ci_inputs, git_context=git_context)
-    jobs.log()
-
     print("\n=== Selecting GPU target families ===")
     targets = select_targets(ci_inputs)
     targets.log()
 
+    print("\n=== Deciding job configuration ===")
+    # Target selection runs first so automatic stage reuse scopes its
```
ScottTodd: If changing the order, please update the comments and function declaration ordering in the file:

* https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L11-L12
* https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L591-L593
* https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L745-L747

(that would be easier to make as a standalone PR ahead of this one)

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,79 @@ def _determine_test_type(
...
+    if targets is not None:
+        target_families = stage_reuse_target_families(
+            targets.linux_families,
+            targets.windows_families,
+        )
+    else:
+        target_families = stage_reuse_target_families(
+            ci_inputs.linux_amdgpu_families,
+            ci_inputs.windows_amdgpu_families,
+        )
```
ScottTodd: I don't understand the branch here. Isn't the `targets` list of families sufficient? That should be populated _from the inputs_ in `select_targets()`.

We shouldn't be qualifying with "normal CI" and "real gfx families", all the code here should be designed to be flexible across multiple different ways that the workflows can be used.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,79 @@ def _determine_test_type(
...
+        mode=StageReuseMode.from_environ(),
+        platforms=build_platforms,
+        target_families=target_families,
+    )
+    log_report(auto)
+
+    try:
+        gha_append_step_summary(render_step_summary(auto))
+    except Exception as exc:  # never fail CI config on a reporting error
+        print(f"[STAGE-REUSE] could not write step summary: {exc}")
```
ScottTodd: No need for this exception handling. The summary step is part of the CI configuration and it should succeed. See also here where we don't catch an exception: https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L1259-L1268

(there is a legitimate use of a lazy import)

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,79 @@ def _determine_test_type(
...
+    auto = compute_auto_stage_reuse(
+        changed_files=git_context.changed_files,
+        mode=StageReuseMode.from_environ(),
+        platforms=build_platforms,
+        target_families=target_families,
+    )
+    log_report(auto)
+
+    try:
+        gha_append_step_summary(render_step_summary(auto))
```
ScottTodd: Per my other comment about ordering, I think this should be part of the `write_outputs()` function so it appears _after_ or _inline_, not _before_ the existing summary output: https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L1237-L1268

Either append to the end or fold the code into https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci_summary.py (by passing around some data/structs)

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -742,6 +799,39 @@ def decide_jobs(
...
+    linux_families = (
+        targets.linux_families
+        if targets is not None
+        else ci_inputs.linux_amdgpu_families
+    )
+    windows_families = (
+        targets.windows_families
+        if targets is not None
+        else ci_inputs.windows_amdgpu_families
+    )
```
ScottTodd: Is there ever a case where "target selection is not known"? I don't think there should be? Don't create problems just to work around them later with fallbacks, solve them the right way with the code structure.

Given that, I don't see the point of this function. It's 30 lines including confusing comments just to determine if `targets.linux_families` and `targets.windows_families` are non-empty.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -742,6 +799,39 @@ def decide_jobs(
 # ---------------------------------------------------------------------------
 
 
+def _stage_reuse_build_platforms(
+    ci_inputs: "CIInputs",
+    targets: "TargetSelection | None",
```
ScottTodd: These type annotations are using strings instead of the actual types. Why? In the current file you can just use the real types, as in https://github.com/ROCm/TheRock/blob/73a658d545d8b8aaf0aa3d08c0c80bb37667878a/build_tools/github_actions/configure_multi_arch_ci.py#L1151-L1156

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,79 @@ def _determine_test_type(
...
+        target_families = stage_reuse_target_families(
+            targets.linux_families,
+            targets.windows_families,
+        )
+    else:
+        target_families = stage_reuse_target_families(
+            ci_inputs.linux_amdgpu_families,
+            ci_inputs.windows_amdgpu_families,
+        )
+    auto = compute_auto_stage_reuse(
```
ScottTodd: Please use a more descriptive name than just `auto` for this variable (that is a keyword in C++ so I'd also avoid it here in Python)

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -681,22 +681,79 @@ def _determine_test_type(
...
+        changed_files=git_context.changed_files,
+        mode=StageReuseMode.from_environ(),
+        platforms=build_platforms,
+        target_families=target_families,
+    )
+    log_report(auto)
```
ScottTodd: I don't see that much use in importing and calling a `log_report()` function at the only use of `compute_auto_stage_reuse()`. Why not just have `compute_auto_stage_reuse()` do the logging itself?

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+import functools
+from dataclasses import dataclass, field
+from typing import Callable, Optional, Sequence
+
+logger = logging.getLogger(__name__)
+
+
+class StageReuseMode(enum.Enum):
+    DRY_RUN = "dry-run"
+    ENFORCE = "reuse-stage"
```
ScottTodd: Why "ENFORCE" when the value is "reuse-stage"? Weird name to choose. How about one of these instead?
* `ENABLED = "enabled"`
* `REUSE_STAGE = "reuse-stage"`

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+
+
+# Accepted spellings for each mode. ``skip-stage`` is the legacy alias that
+# predates the ``reuse-stage`` rename and is kept for backwards compatibility
+# with any callers still passing the old value.
+_STAGE_REUSE_MODE_ALIASES = {
+    "dry-run": StageReuseMode.DRY_RUN,
+    "reuse-stage": StageReuseMode.ENFORCE,
+    "skip-stage": StageReuseMode.ENFORCE,
+}
```
ScottTodd: No "backwards compatibility aliases". This is all code that is only used in-process from one repository right now, there are no external callers "passing the old value". Do the actual rename if you want it.

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+) -> list[str]:
+    """Family list whose artifacts must be verified for stage reuse.
+    Lives here (next to the reuse implementation) rather than in the caller,
+    since it only exists to prepare inputs for ``compute_auto_stage_reuse``.
+    Always includes ``generic`` because every stage produces a generic archive.
+    """
+    families = list(dict.fromkeys([*linux_amdgpu_families, *windows_amdgpu_families]))
+    if "generic" not in families:
+        families.append("generic")
+    return families
```
ScottTodd: This function looks like it should just be merged into `compute_auto_stage_reuse`, and `generic` can be dropped from it. "generic" is not a family, so we shouldn't be loose with typing like this.

* configure_multi_arch_ci.py can just call `compute_auto_stage_reuse()` and pass `linux_amdgpu_families` and `windows_amdgpu_families` as arguments
* `compute_auto_stage_reuse()` can do what it needs with generic stages / artifacts without needing to ask "did the families list include "generic"

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+    baseline_selector_factory: Callable[[str], BaselineSelector] | None = None,
+) -> AutoStageReuse:
+    """Compute auto stage-reuse decisions, verified against a baseline run.
+    A stage is only reusable when it is unaffected by the change AND its
+    artifacts are present in a healthy baseline run for *every* platform in
+    ``platforms``. This guards against the case flagged in review where the
+    resulting ``prebuilt_stages`` are applied to both Linux and Windows build
+    configs: a stage available only in the Linux baseline must not be skipped
+    on Windows. ``platform`` remains accepted as a single-platform shorthand
+    for backwards compatibility.
```
ScottTodd: No "backwards compatibility" in new code! Please check your AI code output, this is a common anti-pattern that agents slip in.

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+        try:
+            if baseline_selector is not None:
+                selector = baseline_selector
+            elif baseline_selector_factory is not None:
+                selector = baseline_selector_factory(plat)
+            else:
+                selector = _default_baseline_selector(platform=plat)
+            baseline = selector(required)
+        except Exception as exc:
+            baseline_error = str(exc)
```
ScottTodd, line 355: What sorts of exceptions is this catching and reporting? Could use some comments here. Transient network (rate-limit etc.) and API errors are may be okay to catch and log, but configuration errors should be raised and reported ~immediately. Just catching the generic `Exception` base class is dangerous.

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#distinguish-between-different-error-conditions

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+    platforms: Sequence[str] | None, platform: str | None
+) -> tuple[str, ...]:
+    """Normalize the platform inputs to a de-duplicated, ordered tuple."""
+    if platforms:
+        resolved = tuple(dict.fromkeys(platforms))
+    elif platform:
+        resolved = (platform,)
+    else:
+        resolved = ("linux",)
+    return resolved
```
ScottTodd: Remove. This helper function appears to have just been added for "backwards compatibility"

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -0,0 +1,652 @@
...
+                                   (default ``multi_arch_ci.yml``).
+* ``STAGE_REUSE_CURRENT_SHA``    - current commit SHA; enables the
+                                   commit-compatibility rule when set.
+* ``STAGE_REUSE_MAX_AGE_HOURS``  - recency window in hours; disables the recency
+                                   rule when unset.
+* ``STAGE_REUSE_COMMIT_HISTORY`` - number of branch commits to fetch for
+                                   ancestry (default ``50``).
+"""
+
+from __future__ import annotations
```
ScottTodd: This import is explicitly banned by the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

ScottTodd (changes requested):
The general approach taken here and plumbing through the scripts is reasonable. I spotted quite a few issues with the specific changes though. Please watch for these coding errors more carefully (local review iterations via coding agents can help there)

Do you have a plan for which sorts of PR runs to first get working with prebuilt artifacts? If you are aiming to get rocm-systems and rocm-libraries using them, a faster path would be to directly set which workflow runs and prebuilt artifacts to use in the workflows whenever the TheRock commit is updated, as that can use the existing manual inputs and none of this extra code:
* rocm-libraries can use a prebuilt compiler-runtime stage
* rocm-systems with changes to hip/CLR unfortunately can't reuse artifacts this way since the compiler and runtimes are together in a stage, we'd need finer-grained artifact fprint-based reuse for that
* rocm-systems with changes to RCCL can reuse all but the comm-libs stage
* rocm-systems with changes to the "storage libs" / "profiler apps" / etc. stages can reuse other stages

lumachad:
> The general approach taken here and plumbing through the scripts is reasonable. I spotted quite a few issues with the specific changes though. Please watch for these coding errors more carefully (local review iterations via coding agents can help there)
> 
> Do you have a plan for which sorts of PR runs to first get working with prebuilt artifacts? If you are aiming to get rocm-systems and rocm-libraries using them, a faster path would be to directly set which workflow runs and prebuilt artifacts to use in the workflows whenever the TheRock commit is updated, as that can use the existing manual inputs and none of this extra code:

Shameless plug, but debug-tools is fairly simple to get working with prebuilt artifacts. It only depends on compiler-runtime. I could help test that setup (I've been using some local scripting to do that).

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -470,11 +481,27 @@ def test_plan_none_changed_files_is_full_rebuild(self):
...
+    def test_linux_only(self):
+        self.assertEqual(srd._build_platforms(["gfx94x"], ()), ("linux",))
+
+    def test_windows_only(self):
+        self.assertEqual(srd._build_platforms((), ["gfx110x"]), ("windows",))
+
+    def test_both_platforms(self):
+        self.assertEqual(
+            srd._build_platforms(["gfx94x"], ["gfx110x"]), ("linux", "windows")
+        )
```
ScottTodd, line 504: What are these tests checking?

The test cases are pretty hard to read with the function call and checks inlined and positional arguments being used instead of named keyword arguments.

ScottTodd (approved):
Thanks for the changes. No more major concerns from me. Having this default to dry-run will let us evaluate it and then choose when to enable. Notably we won't want to reuse artifacts for release builds (the release workflows already don't use that data).

ScottTodd:
Something in here nearly doubled the time it takes to run project unit tests, see https://github.com/ROCm/TheRock/issues/6752

locally I see this with `pytest build_tools/github_actions/tests/ -vv --durations=0`:
`240.15s call     github_actions/tests/configure_multi_arch_ci_test.py::TestFamilyTestFilters::test_submodule_bump_tests_only_enables_tests_with_submodule_changes`

Issue #6218, opened 2026-06-29
[ASAN]: Skip building python packages and frameworks in asan pipelines until supported


ScottTodd:
## Background

Python packages are currently out of scope for ASan releases, yet workflow runs like https://github.com/ROCm/rockrel/actions/runs/28341408029

1. built rocm python packages then did not promote them: https://github.com/ROCm/TheRock/blob/714878997782eefb0a7ba3b8764bc4bf027bf907/build_tools/github_actions/publish_rocm_to_release_buckets.py#L270-L278
2. triggered pytorch workflows (which failed to build)
3. triggered jax workflows (which built and then uploaded, **overwriting non-asan packages**)

---

See also:
* https://github.com/ROCm/rockrel/pull/64
* https://github.com/ROCm/rockrel/pull/69
* https://github.com/ROCm/TheRock/issues/6210
* https://github.com/ROCm/TheRock/pull/6213

## Ideas to fix

I have some open PRs that will help:

* https://github.com/ROCm/TheRock/pull/6082
* https://github.com/ROCm/TheRock/pull/6117

`multi_arch_release_asan.yml` will set `build_pytorch: false` and `build_jax: false` which will propagate through all the workflows. Could do the same for `build_rocm_python`?

ScottTodd:
* The PyTorch trigger should now be skipped in asan workflows
* JAX will be skipped with https://github.com/ROCm/TheRock/pull/6270
* ROCm Python packages will still build but will not be uploaded. We can also add a `build_rocm_python` option to cover for that case too.

I think we'll also want `build_rocm_native_linux` and `build_rocm_native_windows` soon, and then we'll be getting up there with the number of workflow inputs (still some headroom before the hard limit though). Once we have all those options in place we can also start having CI jobs choose which jobs to run based on files modified, like skip pytorch builds if only jax sources are modified.

PR #6270, opened 2026-06-30
Thread JAX build matrix through setup

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/5634 and https://github.com/ROCm/TheRock/issues/6218.

This builds on https://github.com/ROCm/TheRock/pull/6117 to provide the JAX equivalent to https://github.com/ROCm/TheRock/pull/6082. Now the dynamic matrix of JAX builds to run will be generated at the _start_ of CI and release pipelines during "setup" / "configure CI" and a table will be included in the summary: 
image

## Technical Details

The new `build_jax` option now has these settings:

Workflow | `build_jax` value | notes
-- | -- | --
`multi_arch_ci.yml` | `false` | Matching previous behavior<br>can later be enabled (opt-in, automatic based on files edited, etc.)
`multi_arch_ci_asan.yml` | `false` | New behavior due to https://github.com/ROCm/TheRock/issues/6218
`multi_arch_release.yml` | `inputs.build_jax`<br>(default `true`) | Matching previous behavior
`multi_arch_release_asan.yml` | `false` | New behavior due to https://github.com/ROCm/TheRock/issues/6218

## Test Plan

* Dev release triggered to observe CI configuration: https://github.com/ROCm/TheRock/actions/runs/28470132517

## Submission Checklist


erman-gurses (approved):
LGTM - should we also test it for the `true` case just in case?

ScottTodd:
(Note that the base branch for this PR is https://github.com/ROCm/TheRock/pull/6117, which is also up for review still)

ScottTodd:
> should we also test it for the `true` case just in case?

Where specifically? I'm testing both this and https://github.com/ROCm/TheRock/pull/6117 at around the same time and I don't want to trigger too many full release builds. I think the main risk would be some of the default values not getting set or used as expected, which could cause jax builds to be skipped or enabled where we don't want them to be. I hope that the test workflow runs I've triggered already have exercised the rest of the plumbing sufficiently.

erman-gurses:
> > should we also test it for the `true` case just in case?
> 
> Where specifically? I'm testing both this and #6117 at around the same time and I don't want to trigger too many full release builds. I think the main risk would be some of the default values not getting set or used as expected, which could cause jax builds to be skipped or enabled where we don't want them to be. I hope that the test workflow runs I've triggered already have exercised the rest of the plumbing sufficiently.

I forgot about #6117 for a second - reviewing it.

PR #6287, opened 2026-07-01
build(rocgdb): add -fstack-clash-protection to C and CXX flags

lumachad:
## Summary

Appends `-fstack-clash-protection` to `CUSTOM_C_FLAGS` and `CUSTOM_CXX_FLAGS`
in the rocgdb CMake wrapper. This flag instructs the compiler to insert probes
when allocating large stack frames so that a stack-clash attack cannot silently
skip the guard page and corrupt adjacent memory. The flag is guarded behind a
`Linux` platform check as it is not supported on all platforms.

ISSUE ID: ROCM-26825

## Testing Summary

- Build/CI: the flag is appended before the autotools configure step, so a
  successful rocgdb build confirms it is accepted by the compiler.

### DIFF
--- a/debug-tools/rocgdb/CMakeLists.txt
+++ b/debug-tools/rocgdb/CMakeLists.txt
@@ -127,9 +127,11 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
   string(APPEND CUSTOM_CXX_FLAGS " -O0 -g3")
 endif()
 
-# Harden against stack clash attacks.
-string(APPEND CUSTOM_C_FLAGS " -fstack-clash-protection")
-string(APPEND CUSTOM_CXX_FLAGS " -fstack-clash-protection")
+# Harden against stack clash attacks (not supported on all platforms).
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  string(APPEND CUSTOM_C_FLAGS " -fstack-clash-protection")
+  string(APPEND CUSTOM_CXX_FLAGS " -fstack-clash-protection")
+endif()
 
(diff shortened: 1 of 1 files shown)


lancesix (approved):
lgtm, thanks.

debug-tools/rocgdb/CMakeLists.txt
```diff
@@ -127,6 +127,10 @@ if("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
   string(APPEND CUSTOM_CXX_FLAGS " -O0 -g3")
 endif()
 
+# Harden against stack clash attacks.
+string(APPEND CUSTOM_C_FLAGS " -fstack-clash-protection")
+string(APPEND CUSTOM_CXX_FLAGS " -fstack-clash-protection")
```
marbre:
```suggestion
# Harden against stack clash attacks (not supported on all platforms).
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
  string(APPEND CUSTOM_C_FLAGS " -fstack-clash-protection")
  string(APPEND CUSTOM_CXX_FLAGS " -fstack-clash-protection")
endif()
```

I think this is Linux specific. Assuming you want to build rocgdb for Windows or macOS the flag would be either ignored or result in an error.

lumachad: Yeah, good catch. Let me patch that up.

marbre (approved):
LGTM, assuming that CI passes.

lumachad:
CI looks clean, with only a Windows unrelated failure.

PR #6290, opened 2026-07-01
feat(ci): add rocgdb submodule autobump

lumachad:
Adds automated submodule bumping for \`debug-tools/rocgdb/source\`, matching how \`rocm-systems\` and \`rocm-libraries\` already work: the schedule job polls ROCm/rocgdb every 12 hours and opens a fresh PR for each newer commit. rocgdb tracks the \`amd-staging-rocgdb-16\` branch rather than the default branch.

Five commits:

1. **refactor** — clean up \`bump_automation.py\` (constants, helpers, type hints); no behavior change.
2. **feat** — add rocgdb to the existing bump path, with branch-aware polling and a push handler that skips the upstream ref-update step it doesn't need.
3. **test** — unit tests for \`bump_automation.py\`.
4. **fix** — reuse the rocm-systems token for rocgdb until a dedicated GitHub App is provisioned.
5. **fix** — skip bump PR creation when one already exists for the same commit: \`create_therock_bump\` now checks for an open PR on the exact bump branch before cloning TheRock, preventing duplicate PRs when the scheduler fires multiple times between merges.

JIRA: N/A

### DIFF
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -302,6 +302,11 @@ def create_therock_bump(submodule: str, token: str) -> None:
 
         current_sha = get_submodule_sha("HEAD", submodule)
 
+        if current_sha == latest:
+            print(f"[INFO] {submodule} is already at {latest[:7]}, nothing to bump")
+            os.chdir(original_cwd)
+            return
+
         # Fetch the exact target commit in the submodule. A plain depth-1 fetch
         # only retrieves the default branch tip, which misses commits that live
         # on a non-default branch (e.g. rocgdb's amd-staging-rocgdb-16).
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -49,7 +49,8 @@ SUBMODULE_CONFIG = {
         "repo": "ROCm/rocgdb",
         "files": [],
         "updater": "submodule-only",
-        "token_key": "rocgdb",
+        # We will reuse the rocm-systems token for now.
+        "token_key": "systems",
         "branch": "amd-staging-rocgdb-16",
     },
 }
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -267,6 +267,22 @@ def create_therock_bump(submodule: str, token: str) -> None:
     # Get latest SHA from upstream submodule repo
     latest = latest_commit(repo, token, branch)
 
+    # The submodule path may contain slashes (e.g. debug-tools/rocgdb/source);
+    # flatten it so the bump branch name is a single ref component.
+    branch_name = f"bump-{submodule.replace('/', '-')}-{latest[:7]}"
+
+    # Skip if a PR for this exact commit is already open.
+    open_prs = gh_api(
+        token,
+        f"repos/{THEROCK_REPO}/pulls?state=open&head=ROCm:{branch_name}",
+    )
+    if open_prs:
+        print(
+            f"[INFO] Bump PR for {branch_name} already open"
+            f" (#{open_prs[0]['number']}), skipping"
+        )
+        return
+
     # Use a temp directory for safe cloning
     with tempfile.TemporaryDirectory() as tmpdir:
         clone_dir = os.path.join(tmpdir, "TheRock")
@@ -276,9 +292,6 @@ def create_therock_bump(submodule: str, token: str) -> None:
         )
         os.chdir(clone_dir)
 
-        # The submodule path may contain slashes (e.g. debug-tools/rocgdb/source);
-        # flatten it so the bump branch name is a single ref component.
-        branch_name = f"bump-{submodule.replace('/', '-')}-{latest[:7]}"
         run(["git", "checkout", "-b", branch_name])
 
         # Initialize the submodule if needed
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -45,6 +45,13 @@ SUBMODULE_CONFIG = {
         "updater": "ci-env",
         "token_key": "libraries",
     },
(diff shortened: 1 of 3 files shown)


lumachad:
Manual test run here: https://github.com/ROCm/TheRock/actions/runs/28510195621 (disabled rocm-systems/rocm-libraries bumps so we don't create noise).

lumachad:
@amd-chiranjeevi Since you've touched this workflow before, could you please also take a look at it?

lumachad:
dry-run test run here: https://github.com/ROCm/TheRock/actions/runs/29236368424/job/86771874361.

lumachad:
@ammallya I've made the rocgdb auto-bump match rocm-systems and rocm-libraries with the multiple PR's and closing stale PR's on push.

build_tools/github_actions/bump_automation.py
```diff
@@ -197,81 +226,109 @@ def close_stale_prs(submodule, old_sha, systems_token):
...
     """Create a bump PR for the given submodule in TheRock."""
-    repo = SUBMODULE_CONFIG[submodule]["repo"]
+    config = SUBMODULE_CONFIG[submodule]
+    repo = config["repo"]
+    branch = config.get("branch")
 
     original_cwd = os.getcwd()
     # Get latest SHA from upstream submodule repo
-    latest = latest_commit(repo, token)
+    latest = latest_commit(repo, token, branch)
```
erman-gurses, line 268: Could we add a small rocgdb-specific test for the configuration wiring?

The existing `LatestCommitTest` verifies that `latest_commit()` correctly handles an explicitly provided branch, but it hard-codes both `ROCm/rocgdb` and `amd-staging-rocgdb-16`. It does not verify that `create_therock_bump("debug-tools/rocgdb/source", token)` actually reads those values from `SUBMODULE_CONFIG` and passes them to `latest_commit()`.

A focused test asserting that call would help catch accidental changes to the `rocgdb` repository or branch configuration - it is a minor coverage suggestion rather than a blocker.

build_tools/github_actions/bump_automation.py
```diff
@@ -280,49 +337,56 @@ def create_therock_bump(submodule, token):
...
+def handle_schedule(tokens: dict[str, str], submodule: str = "all") -> None:
+    """Create bump PRs for the specified submodule(s)."""
     if submodule in ("all", "rocm-systems"):
-        create_therock_bump("rocm-systems", systems_token)
+        create_therock_bump("rocm-systems", tokens["systems"])
     if submodule in ("all", "rocm-libraries"):
-        create_therock_bump("rocm-libraries", libraries_token)
+        create_therock_bump("rocm-libraries", tokens["libraries"])
+    if submodule in ("all", "rocgdb"):
+        create_therock_bump("debug-tools/rocgdb/source", tokens["rocgdb"])
```
erman-gurses, line 360: Could we add a `handle_schedule()` test for the new `rocgdb` option?

The existing tests cover the branch-aware `latest_commit()` helper, but they do not verify that the workflow-facing value `rocgdb` is mapped to `debug-tools/rocgdb/source` and uses `tokens["rocgdb"]`.

A small test around `handle_schedule(tokens, "rocgdb")` would cover the new scheduling integration and ensure the other submodules are not invoked.

build_tools/github_actions/tests/bump_automation_test.py
```diff
@@ -0,0 +1,360 @@
...
+                "bump_automation.get_submodule_sha", return_value="oldsha1234567"
+            ):
+                with patch("bump_automation.close_stale_prs") as mock_close:
+                    with patch(
+                        "bump_automation.tempfile.TemporaryDirectory"
+                    ) as mock_tmp:
+                        handle_push(
+                            "before",
+                            "after",
+                            {"systems": "tok", "libraries": "tok", "rocgdb": "tok"},
```
erman-gurses: We might use distinct token values here and assert the exact `close_stale_prs()` call?

Since the `rocgdb` configuration intentionally reuses the systems token for now, passing the same `tok` value for every key means this test would still pass if the wrong token were selected.

For example, use separate `systems-token` and `rocgdb-token` values, then assert that `close_stale_prs()` receives `systems-token`.

erman-gurses (approved):
LGTM overall - please see my comments below:

build_tools/github_actions/bump_automation.py
lumachad: Done. Added `CreateTheRockBumpRocgdbConfigTest.test_reads_repo_and_branch_from_submodule_config`, which calls `create_therock_bump("debug-tools/rocgdb/source", "token")` and asserts that `latest_commit` is invoked with `"ROCm/rocgdb"` and `"amd-staging-rocgdb-16"` as read from `SUBMODULE_CONFIG`.

build_tools/github_actions/bump_automation.py
lumachad: Done. Added `HandleScheduleTest` verifying that `handle_schedule(tokens, "rocgdb")` calls `create_therock_bump` exactly once with `"debug-tools/rocgdb/source"` and `tokens["rocgdb"]`. A second test confirms neither `"rocm-systems"` nor `"rocm-libraries"` is invoked.

build_tools/github_actions/tests/bump_automation_test.py
lumachad: Done. Updated the test to use distinct `"systems-token"`, `"libraries-token"` and `"rocgdb-token"` values and added an assertion that `close_stale_prs` receives `"systems-token"` (since rocgdb's `token_key` is `"systems"`).

lumachad:
I ran a manual check with this one and it works for the case where we have nothing to bump. I'll wait until we have something to bump to and run another check to be sure.

lumachad:
Tested rocgdb bump creation by running the workflow manually and restricted to rocgdb: https://github.com/ROCm/TheRock/pull/6725

PR #6291, opened 2026-07-01
build(pytorch) [Release 7.14] Remove pytorch 2.9 + nightly

HereThereBeDragons:
Adjust pytorch build matrix matching 7.14 release matrix.
Drops building pytorch 2.9 and nightly

### DIFF
--- a/build_tools/github_actions/configure_pytorch_release_matrix.py
+++ b/build_tools/github_actions/configure_pytorch_release_matrix.py
@@ -27,12 +27,8 @@ PYTHON_VERSIONS = ["3.10", "3.11", "3.12", "3.13", "3.14"]
 #   pytorch_git_ref         – git branch/tag passed to the build job
 #   exclude_amdgpu_families – (optional) set[str] of family names to drop
 
+# adjusted for release/therock-7.14
 PYTORCH_REFS_LINUX: list[dict] = [
-    {
-        "pytorch_git_ref": "release/2.9",
-        # gfx125x not supported for PyTorch 2.9.
-        "exclude_amdgpu_families": {"gfx125x"},
-    },
     {
         "pytorch_git_ref": "release/2.10",
         # gfx125x not supported for PyTorch 2.10.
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -49,21 +45,14 @@
         # See https://github.com/ROCm/TheRock/issues/5833.
         "exclude_amdgpu_families": {"gfx125x"},
     },
-    {
```
ScottTodd, line 52: With the more recent code on the `main` branch we could choose to drop `nightly` if the `release_type` is `prerelease`, see https://github.com/ROCm/TheRock/blob/a13cf65fa2b08bc1b1527c0f21c21511f5eb0f96/build_tools/github_actions/configure_pytorch_release_matrix.py#L29-L51

build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -27,12 +27,8 @@
 #   pytorch_git_ref         – git branch/tag passed to the build job
 #   exclude_amdgpu_families – (optional) set[str] of family names to drop
 
+# adjusted for release/therock-7.14
 PYTORCH_REFS_LINUX: list[dict] = [
-    {
-        "pytorch_git_ref": "release/2.9",
-        # gfx125x not supported for PyTorch 2.9.
-        "exclude_amdgpu_families": {"gfx125x"},
-    },
```
ScottTodd, line 35: If we're dropping support for the pytorch `release/2.9` branch, shouldn't we also do that on TheRock's `main` branch for nightly releases too? Why just for TheRock's `release/7.14` branch?

ScottTodd (commented):
Do we actually need this? I'd prefer to keep changes on the release branches minimal. Without this change we'll build more packages than we publish and I think our package promotion scripts already handle that?

build_tools/github_actions/configure_pytorch_release_matrix.py
HereThereBeDragons: agree that would be a good change on main.

build_tools/github_actions/configure_pytorch_release_matrix.py
HereThereBeDragons: yes we should. i havent heard any resistant against it since i proposed this to lockstep this with the release.

HereThereBeDragons:
> Do we actually need this? I'd prefer to keep changes on the release branches minimal. Without this change we'll build more packages than we publish and I think our package promotion scripts already handle that?

i would argue the less people need to think about what we are triggering in the release, the better. if we manage to have this more automated and dont need so many changes on the release branch thats of course preferred but currently we are not there yet.

for the download script yes you can have include and exclude globs, for promotion just include glob.
however also for accountability in the future and reducing confusion i think it would be better to just build what we want and not build everything and then select.

one more thought:
maybe in the future we immediately drop the old versions we no longer want to support on main before the branch cut and then just have the exception "if release skip nightly"

ScottTodd (approved):
> > Do we actually need this? I'd prefer to keep changes on the release branches minimal. Without this change we'll build more packages than we publish and I think our package promotion scripts already handle that?
> 
> i would argue the less people need to think about what we are triggering in the release, the better. if we manage to have this more automated and dont need so many changes on the release branch thats of course preferred but currently we are not there yet.

Agreed. Alright, I'm fine with this change then.

astyrrian1 (approved):

Issue #6301, opened 2026-07-01
JAX 0.10.2 Tests are not completed or takes longer time than expected


erman-gurses:
0.10.2 support came with the PR https://github.com/ROCm/TheRock/pull/6195, however some of the tests take longer time to finish or never finishes.  Please see below:

https://github.com/ROCm/rockrel/actions/runs/28410737372
https://github.com/ROCm/rockrel/actions/runs/28417997096
https://github.com/ROCm/rockrel/actions/runs/28491238833
https://github.com/ROCm/rockrel/actions/runs/28497809718

Latest manual run finishes the test in expected time: 
https://github.com/ROCm/rockrel/actions/runs/28491006880

magaonka-amd:
@erman-gurses  thanks for reporting 
https://amd-hub.atlassian.net/browse/ROCM-27290?focusedCommentId=502651&sourceType=mention this bug is the reason for that slowdown and hang.

magaonka-amd:
Can you tell me from where these dockers are built/pulled from ? we will have to bake that env variable to make pytest work consistently

https://github.com/ROCm/rocm-jax/commit/0ffca82c19502def750542d95ca22d25557ab546 I have baked it here from rocm-jax repo rocm-jaxlib-v0.10.2 branch. is there another place from where docker is built/pulled from? @erman-gurses

erman-gurses:
@magaonka-amd , in the JAX workflows, here are the docker related sections:
https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_linux_jax_wheels.yml#L215-L288
https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_linux_jax_wheels_partial.yml#L129-L145

ScottTodd:
Let's add a job or step level timeout of 30 minutes, 1 hour, or 2 hours. The 6 hour default timeout is an inefficient use of expensive GPU runners.

Let's also add the JAX version to the test job names so these are easier to navigate: 

<img width="336" height="566" alt="Image" src="https://github.com/user-attachments/assets/445bf3d5-ee40-434b-bbef-e486b08d226d" />

magaonka-amd:
I'm going to push the docker fix out in next hour or so.

erman-gurses:
Raised a timeout PR: https://github.com/ROCm/TheRock/pull/6306

erman-gurses:
_Let's also add the JAX version to the test job names so these are easier to navigate:_

[...]

PR #6309, opened 2026-07-01
ci(jax): set ROCm runtime env for JAX 0.10.2 tests

magaonka-amd:
#6301

## Summary
- JAX 0.10.2 pytest jobs slow down or hang on the GPU test runners (#6301).
- Set `ROCPROFILER_QUEUE_INTERPOSITION=0`, `DEBUG_HIP_DYNAMIC_QUEUES=0`, `HSA_NO_SCRATCH_RECLAIM=1` for the JAX test step in `test_linux_jax_wheels_partial.yml` (the 0.10.2 release test path: `multi_arch_release_*` -> `multi_arch_build_*` -> this workflow).
- Gated with `if: inputs.jax_ref == 'rocm-jaxlib-v0.10.2'` and written via `$GITHUB_ENV`, so the vars are **never defined** for other JAX releases -- no behavior change for them.

## Notes
- Same three env vars were baked into the rocm-jax base image in ROCm/rocm-jax@0ffca82; this mirrors that fix for the TheRock-published test container path, which the JAX tests actually run in.

## Test plan
- [ ] Trigger the 0.10.2 JAX wheel test workflow and confirm the previously hanging tests complete within the step timeout.
- [ ] Confirm a non-0.10.2 JAX run skips the "Set ROCm runtime env" step (vars unset).


magaonka-amd:
I dont like setting env variables like this , but this is temp. fixes should land from respective rocm teams soon.

erman-gurses (approved):
Thanks - LGTM!
Test: https://github.com/ROCm/TheRock/actions/runs/28540777711/job/84613734332

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -262,6 +262,12 @@ jobs:
           # other releases keep the platform allocator.
           XLA_PYTHON_CLIENT_ALLOCATOR: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && 'default' || 'platform' }}
 
+          # JAX 0.10.2 only: ROCm/HIP runtime tuning to avoid a pytest
+          # slowdown/hang; other releases keep the runtime defaults.
+          ROCPROFILER_QUEUE_INTERPOSITION: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '0' || '' }}
+          DEBUG_HIP_DYNAMIC_QUEUES: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '0' || '' }}
+          HSA_NO_SCRATCH_RECLAIM: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '1' || '' }}
```
ScottTodd, line 270: > * Same three env vars were baked into the rocm-jax base image in [ROCm/rocm-jax@0ffca82](https://github.com/ROCm/rocm-jax/commit/0ffca82); this mirrors that fix for the TheRock-published test container path, which the JAX tests actually run in.

This is okay for now, but for these packages to be called mature and release ready, the default behavior when someone runs `pip install jax` from our release index and runs unit tests or full programs should not suffer from hangs, timeouts, etc. We shouldn't require users to set different environment variables based on the version that they install. Some users may use dockerfiles we publish but not all, and packages need to work on their own.

Test workflows like this one should also match user scenarios as closely as possible, so I'd prefer to not use behavior-changing environment variables like this.

erman-gurses: @magaonka-amd, can you please add TODO for that since it is a temp solution.

magaonka-amd: then fix needs to come from corresponding systems team not from us. I'm only trying to help you folks unblock CI here. I share same concern as you folks regarding using env variables. but without these we simply can't make JAX CI work. It is concerning that JAX framework tests are catching these systems bug but not the lower level tests.

If you dont prefer to have this change , I understand it. but CI will continue to have hang or slow execution issue.

ScottTodd: This is fine for now, I just don't want to set too much precedent for version-specific environment variables that are required for correct behavior.

magaonka-amd: thanks, added TODO , lets hope for next jax release all these systems bug are fixed and merged.

ScottTodd (approved):

PR #6313, opened 2026-07-01
ci: Log artifact archive file sizes and sha256 hashes during upload

ScottTodd:
## Motivation

This extra logging will help debug issues like https://github.com/ROCm/TheRock/issues/4202 where build jobs can silently overwrite artifacts with different contents. Projects _should_ be producing identical artifacts across all build jobs, but that isn't actually enforced in the build system currently.

Previously we'd have logs like
```
Processing 81 artifacts (81 to compress, 0 pre-compressed)...
  ++ Compressing support_dev_generic
  ++ Compressing prim_test_generic
  ++ Compressing miopenprovider_test_generic
  ++ Compressing rocalution_dev_generic
  ++ Compressing prim_dev_generic
  ++ Uploading support_dev_generic.tar.zst
  ++ Compressing composable-kernel_dev_generic
  ++ Compressing rocwmma_dbg_generic
  ++ Uploading composable-kernel_doc_generic.tar.zst
  ++ Uploading hipblasltprovider_dbg_generic.tar.zst
```

Now we'll have logs like
```
Processing 44 artifacts (44 to compress, 0 pre-compressed)...
  ++ Compressing amd-llvm_dbg_generic
  ++ Compressing amd-llvm_dev_generic
  ++ Compressing amd-llvm_doc_generic
  ...
  ++ Compressing fftw3_run_generic
  ++ Compressed base_dev_generic (files=33, uncompressed_size=302119 bytes) -> base_dev_generic.tar.zst (compressed_size=62696 bytes, sha256=8b25469d31e063db66a09f50a42d694bcd680e1d79930ccde6ea67d24ddb4c85)
  ++ Compressing flatbuffers_dev_generic
  ++ Uploading base_dev_generic.tar.zst (compressed_size=62696 bytes, sha256=8b25469d31e063db66a09f50a42d694bcd680e1d79930ccde6ea67d24ddb4c85)
  ++ Uploading base_dev_generic.tar.zst.sha256sum (artifact_sha256=8b25469d31e063db66a09f50a42d694bcd680e1d79930ccde6ea67d24ddb4c85, size=66 bytes)
  ++ Compressed core-hip_doc_generic (files=1, uncompressed_size=16 bytes) -> core-hip_doc_generic.tar.zst (compressed_size=171 bytes, sha256=d1a2b7c5229e66f33b15626a9654862c166ae78a18e65ae8d46b5a635e791ab4)
  ++ Compressing flatbuffers_run_generic
  ++ Uploading core-hip_doc_generic.tar.zst (compressed_size=171 bytes, sha256=d1a2b7c5229e66f33b15626a9654862c166ae78a18e65ae8d46b5a635e791ab4)
  ++ Uploading core-hip_doc_generic.tar.zst.sha256sum (artifact_sha256=d1a2b7c5229e66f33b15626a9654862c166ae78a18e65ae8d46b5a635e791ab4, size=66 bytes)
```

We'll be able to inspect the sha256 values along with file counts and uncompressed/compressed sizes to spot if different build stages are uploading different versions of artifact archives.

## Test Plan

As we don't yet have https://github.com/ROCm/rocm-libraries/commit/c534dd117e1197658ed8c137e6cbe5133b86bfd5, I would expect gfx90c and gfx950 to produce different generic miopen artifacts and we should now see that in the logs.

## Submission Checklist

### DIFF
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -549,6 +549,44 @@ class UploadRequest:
     backend: ArtifactBackend
 
 
+def _format_bytes(size: Optional[int]) -> str:
+    return "unknown" if size is None else f"{size} bytes"
+
+
+def _directory_file_count_and_size_sum(path: Path) -> Optional[tuple[int, int]]:
+    try:
+        count = 0
+        total = 0
+        for child in path.rglob("*"):
+            if child.is_file():
+                count += 1
+                total += child.stat().st_size
+        return count, total
+    except OSError:
+        return None
+
+
+def _hash_file_path(archive_path: Path) -> Path:
+    return Path(f"{archive_path}.sha256sum")
+
+
+def _read_sha256sum_value(hash_path: Path) -> str:
+    """Read the digest from a hash sidecar.
+
+    Existing files in tests and logs use both plain digest and
+    ``sha256sum``-style ``digest filename`` formats. The digest is always the
+    first whitespace-delimited token.
+    """
+    content = hash_path.read_text().strip()
+    if not content:
+        raise ValueError(f"Empty sha256sum file: {hash_path}")
+    return content.split()[0]
+
+
+def _artifact_sha256(source_path: Path) -> str:
+    return _read_sha256sum_value(_hash_file_path(source_path))
+
+
 def compress_artifact(request: CompressRequest) -> Optional[Path]:
     """Compress a single artifact directory using fileset_tool.py artifact-archive."""
     try:
@@ -575,7 +613,7 @@ def compress_artifact(request: CompressRequest) -> Optional[Path]:
         cmd.extend(
             [
                 "--hash-file",
-                str(request.archive_path) + ".sha256sum",
+                str(_hash_file_path(request.archive_path)),
                 str(request.source_dir),
             ]
         )
@@ -586,6 +624,21 @@ def compress_artifact(request: CompressRequest) -> Optional[Path]:
                 f"fileset_tool.py artifact-archive failed (returncode={result.returncode}): {result.stderr}"
             )
 
+        sha256 = _artifact_sha256(request.archive_path)
+        file_count_and_size = _directory_file_count_and_size_sum(request.source_dir)
+        if file_count_and_size is None:
+            file_count = "unknown"
+            uncompressed_size = None
+        else:
+            file_count, uncompressed_size = file_count_and_size
+        log(
+            f"  ++ Compressed {request.source_dir.name} "
+            f"(files={file_count}, "
+            f"uncompressed_size={_format_bytes(uncompressed_size)}) "
+            f"-> {request.archive_path.name} "
+            f"(compressed_size={_format_bytes(request.archive_path.stat().st_size)}, "
+            f"sha256={sha256})"
+        )
         return request.archive_path
     except Exception as e:
         log(f"  !! Failed to compress {request.source_dir.name}: {e}")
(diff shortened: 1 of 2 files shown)


build_tools/artifact_manager.py
```diff
@@ -586,6 +622,21 @@ def compress_artifact(request: CompressRequest) -> Optional[Path]:
...
+        else:
+            file_count, uncompressed_size = file_count_and_size
+        log(
+            f"  ++ Compressed {request.source_dir.name} "
+            f"(files={file_count}, "
+            f"uncompressed_size={_format_bytes(uncompressed_size)}) "
+            f"-> {request.archive_path.name} "
+            f"(compressed_size={_format_bytes(request.archive_path.stat().st_size)}, "
+            f"sha256={sha256})"
+        )
```
ScottTodd, line 639: These logs are providing good data on https://github.com/ROCm/TheRock/actions/runs/28549197565?pr=6313

Agent analysis over gfx900,gfx906,gfx1151,gfx1153:

>  Key results:
>
>  - 300 compressed artifact records total.
>  - 69 repeated archive names across jobs, all _generic.
>  - All 69 repeated generic archives had different SHA values across the targets that produced them.
>  - Upload metadata matched compression metadata: 0 upload size/hash mismatches.
>
>  miopen_lib_generic does support your theory:
>
>  ```
>  gfx1151  files=4  uncompressed=1429560045  compressed=15930113  sha=ffb3c96c24e9...
>  gfx1153  files=4  uncompressed=1429560045  compressed=15932764  sha=8e9639240810...
>  gfx900   files=4  uncompressed=1429305069  compressed=15932732  sha=9e25bb518eaf...
>  gfx906   files=4  uncompressed=1429305069  compressed=15941213  sha=ad2f0aa928f0...
>  ```
>
>  So gfx1151/gfx1153 are 254,976 bytes larger uncompressed than gfx900/gfx906, with all four archive SHAs different.
>
>  Other repeated generic artifacts with uncompressed-size differences, which are the strongest evidence of real payload
>  differences:
>
>  ```
>  rand_test_generic                         delta=1,232,896
>  composable-kernel_dev_generic             delta=960,796
>  miopen_lib_generic                        delta=254,976
>  miopen_test_generic                       delta=171,776
>  prim_test_generic                         delta=114,688
>  fft_test_generic                          delta=65,536
>  rocwmma_test_generic                      delta=12,288
>  rocalution_lib_generic                    delta=12,288
>  blas_test_generic                         delta=4,112
>  hipkernelprovider_test_generic            delta=3,912
>  hipkernelprovider_lib_generic             delta=3,904
>  ```
>
>  There are also likely-content-different but less conclusive cases where uncompressed size is identical but compressed
>  size changes a lot:
>
>  ```
>  hipdnn-integration-tests_test_generic      compressed delta=2,228,470
>  hipdnn_test_generic                        compressed delta=251,304
>  hipdnn-samples_test_generic                compressed delta=164,745
>  hiptensor_test_generic                     compressed delta=113,575
>  ```
>
>  The noisy part: SHA alone is not enough here. 31/69 repeated generic artifacts had identical uncompressed size and
>  compressed-size deltas of <=10 bytes, and 16 repeated groups were zero-uncompressed artifacts. Those are plausibly
>  archive metadata/timestamp/order churn rather than meaningful content differences.

So some high level takeaways:
* Test artifacts may not be generic (as we expected/feared)
  * Are we actually seeing test failures as a result of this? Why/why not?
  * We do omit test artifacts from rocm python packages for this (suspected) reason and we now ship tarballs that omit tests too (in addition to the "include everything" tarballs)
* Some artifacts have the same uncompressed size and file count yet different sha256 hashes, maybe nondeterminism or embedded data like timestamps (harmless?)
  * This will make tests like https://github.com/ROCm/TheRock/pull/4203 harder, as we can't just rely on the sha256 values needing to match (currently)
* miopen_lib_generic does have a few distinct uncompressed sizes: one for gfx900/gfx906 (CK disabled) and another for gfx1151/gfx1153 (CK enabled), hopefully those are the same after https://github.com/ROCm/rocm-libraries/pull/8988
* A few other artifacts like rocalution_lib_generic, composable-kernel_dev_generic, and hipkernelprovider_lib_generic differ across shards

Quite a few things to follow up on there, but at least with this PR we'll have extra logs to help debug at a distance, and we can check the final uploaded sha256 against the logs to see which shard uploaded the "winning" artifact. As for the root causes for these differences, that will be easier to debug by building from source locally.

build_tools/artifact_manager.py
```diff
@@ -599,14 +650,30 @@ def upload_artifact(request: UploadRequest) -> bool:
...
+            # Upload the artifact's sha256sum file.
             if sha_path.exists():
+                sha256sum_sha256 = _read_sha256sum_value(sha_path)
+                if sha256sum_sha256 != sha256:
+                    log(
+                        f"  !! WARNING: sha256 mismatch for {request.artifact_key}: "
+                        f"computed_sha256={sha256}, sha256sum_sha256={sha256sum_sha256}"
+                    )
+                log(
+                    f"  ++ Uploading {request.artifact_key}.sha256sum "
```
marbre, line 673: Do we just want to warn or rather let the upload fail here?

build_tools/tests/artifact_manager_tool_test.py
marbre, line 1: Looks good to me but my review agent pointed out the following:

> No test covers the mismatch warning path (where sha256sum_sha256 != sha256). That path is hard to hit in normal operation but is the most interesting new behavior introduced. A test that injects a bad sidecar and asserts the warning is logged would strengthen coverage.

marbre (approved):

build_tools/artifact_manager.py
ScottTodd: I'd rather start with just informational logging and then switch to errors once we better understand where the differences are coming from.

build_tools/tests/artifact_manager_tool_test.py
ScottTodd: I don't like the concept of adding unit tests exclusively for logging, as logging is not part of the API contract. If we add control flow changes (treat as errors) then yeah we can add tests for those code paths too.

PR #6335, opened 2026-07-03
fix(emulation): add bundled libdrm dependency for rocjitsu

### LINKED ISSUE #6229: [Issue]: rocjitsu fails to build: `fatal error: libdrm/drm.h: No such file or directory`
### Problem Description

Building ROCm with TheRock without libdrm development headers installed causes an error when building rocjitsu.

```
/home/danielsu/workspace/TheRock/rocm-systems/emulation/rocjitsu/lib/rocjitsu/external_headers/hsa_headers/linux/uapi/kfd_ioctl.h:26:10: fatal error: libdrm/drm.h: No such file or directory
   26 | #include <libdrm/drm.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

The rocjitsu build continues without error after adding the below snippet to https://github.com/ROCm/TheRock/blob/main/emulation/CMakeLists.txt#L22:
```
RUNTIME_DEPS
        ${THEROCK_BUNDLED_LIBDRM}
```

I don't have write perms for this repo yet, can help make a PR for this after I get access.

### Operating System

Ubuntu 24.04.4 LTS

### CPU

Threadripper 1900X

### GPU

Vega 56/64

### ROCm Version

TheRock@a512f42

### ROCm Component

_No response_

### Steps to Reproduce

Ensure `libdrm-dev` is not installed on the system.
Follow the initial setup steps as from the README.
```
cmake -B build -GNinja -DTHEROCK_AMDGPU_FAMILIES=gfx900-dgpu
ninja -C build
```

_No response_

### Additional Information

_No response_

dysu-amd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/6229

## Technical Details

Adds TheRock bundled libdrm as a rocjitsu dependency.

## Test Plan

Verify that local build passes without libdrm system headers installed.
Verify that CI passes.

## Test Result

Local build passes.
~CI TBD~ Looks like the emulation components aren't being built in CI yet?

## Submission Checklist


marbre (commented):
Drive-by, I think this needs an artifact dep on `third-party-sysdeps` in `BUILD_TOPOLOGY.toml`.

ScottTodd (approved):
> ~CI TBD~ Looks like the emulation components aren't being built in CI yet?

rocjitsu is built as part of the compiler-runtime stage, e.g.
* https://github.com/ROCm/TheRock/actions/runs/29254609447?pr=6335
* https://therock-ci-artifacts.s3.amazonaws.com/29254609447-linux/logs/compiler-runtime/index.html

I see on https://therock-ci-artifacts.s3.amazonaws.com/29254609447-linux/logs/compiler-runtime/rocjitsu_configure.log compared to a baseline run of https://therock-ci-artifacts.s3.amazonaws.com/29271407614-linux/logs/compiler-runtime/rocjitsu_configure.log that this line is new:
```
0.0	-- Sub-project pkg-config prefixes: /__w/TheRock/TheRock/build/third-party/sysdeps/linux/libdrm/build/stage/lib/rocm_sysdeps
```

but I don't see any other CMake configure logs specific to libdrm. Is this working as expected there?

dysu-amd:
> but I don't see any other CMake configure logs specific to libdrm. Is this working as expected there?

I'm not sure about that. Looked into it more and it seems that rocjitsu actually vendors its own drm and hsa headers: https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/CMakeLists.txt#L61-L71

It's the vendored hsa headers that are causing the error in https://github.com/ROCm/TheRock/issues/6229, because it still looks for system drm `#include <libdrm/drm.h>`: https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/lib/rocjitsu/external_headers/hsa_headers/linux/uapi/kfd_ioctl.h#L26

Is this PR an OK fix in that context? Or should the fix be on the rocjitsu side?

ScottTodd:
> I'm not sure about that. Looked into it more and it seems that rocjitsu actually vendors its own drm and hsa headers: https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/CMakeLists.txt#L61-L71

I would expect projects in rocm-systems to reference each other directly, not copy files like that. Header files that are part of a stable API are somewhat fuzzy there though.

> It's the vendored hsa headers that are causing the error in #6229, because it still looks for system drm `#include <libdrm/drm.h>`: https://github.com/ROCm/rocm-systems/blob/develop/emulation/rocjitsu/lib/rocjitsu/external_headers/hsa_headers/linux/uapi/kfd_ioctl.h#L26

I see some other projects in rocm-systems using `#include "drm.h"` instead of `#include <libdrm/drm.h>`. Generally we should depend on the common version of each dependency that we bundle/vendor via the superproject build system, instead of system files that may differ depending on the build container/environment/host/etc. I thought CMake could still do that even with the bracket include method though.

> Is this PR an OK fix in that context? Or should the fix be on the rocjitsu side?

I think it's probably both. Add the dep as you have it here (I think it's harmless if unused), then also adjust rocjitsu to use it instead of a version from outside the source build system?

dysu-amd:
Overriding and merging, the pytorch failure is due to DNS (https://github.com/ROCm/TheRock/issues/6534), Windows libhipcxx_hipcc gfx110X error is known (https://github.com/ROCm/TheRock/issues/4617). Both seem unrelated.

PR #6341, opened 2026-07-04
fix(ci): Resolve undefined pthread symbols in ASAN gtest links

amd-hsong:
## Motivation

ASAN builds fail to link gtest-based test executables with undefined pthread symbols. This PR fixes the issue.

## Technical Details

See Issue https://github.com/ROCm/rocm-libraries/issues/8889.

Two link flags in the sanitizer stanza reach the build through different CMake channels in cmake/therock_sanitizers.cmake:

  - `-fsanitize=address` is added via `CMAKE_C/CXX_FLAGS_INIT`.
  - `-shared-libsan` is added via `add_link_options()`.

  CMake's `FindThreads` detection runs a `try_compile` probe that inherits `*_FLAGS_INIT` but **not** `add_link_options()`. So the probe links with `-fsanitize=address` but without `-shared-libsan`, using clang's default **static** sanitizer runtime — which auto-links `-lpthread`. The probe therefore succeeds without explicit threads, CMake sets `CMAKE_HAVE_LIBC_PTHREAD=1`, and
  `Threads::Threads` becomes an empty target.

  Real executables, however, link with `-shared-libsan` (shared runtime, no auto `-lpthread`). gtest propagates the now-empty `Threads::Threads`, so nothing puts `-pthread` on the link line, and the pthread TLS symbols go undefined on toolchains where they live only in `libpthread` for glibc < 2.34 (the manylinux CI container is based on glibc 2.28).

  This surfaced when rocThrust switched its default SQLite3 to the system (shared) package: the previously in-tree SQLite compilation had been incidentally pulling `libpthread` onto the test link lines and masking this problem.

## Test Plan

Verified locally in the `manylinux_2_28` build container: ASAN rocThrust test binaries linked successfully.

## Submission Checklist

### DIFF
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -57,6 +57,16 @@ function(therock_sanitizer_configure
     # https://github.com/ROCm/TheRock/issues/1782
     string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-fsanitize=${_sanitizer_string}>\n")
     string(APPEND _stanza "  $<$<AND:$<LINK_LANGUAGE:C,CXX>,$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>>:-shared-libsan>)\n")
+    # CMake's FindThreads probe (a try_compile) inherits CMAKE_C/CXX_FLAGS_INIT
+    # above but NOT the add_link_options(), so it never sees -shared-libsan and
+    # uses clang's default static sanitizer runtime. That static runtime auto-links
+    # -lpthread, so the probe passes and sets CMAKE_HAVE_LIBC_PTHREAD=1 and
+    # Threads::Threads becomes empty. Real links, however, use shared linkage
+    # (-shared-libsan above, no auto -lpthread), so the empty Threads::Threads
+    # puts nothing on the link and pthread_* go undefined where they live only in
+    # libpthread for glibc < 2.34. Force -pthread here.
+    # https://github.com/ROCm/rocm-libraries/issues/8889
+    string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-pthread>)\n")
 
     # Note: autotools-based subprojects (e.g. rocgdb) ignore add_link_options() because
     # they pass CMAKE_*_LINKER_FLAGS directly to their configure script. Those subprojects
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -57,16 +57,19 @@ function(therock_sanitizer_configure
     # https://github.com/ROCm/TheRock/issues/1782
     string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-fsanitize=${_sanitizer_string}>\n")
     string(APPEND _stanza "  $<$<AND:$<LINK_LANGUAGE:C,CXX>,$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>>:-shared-libsan>)\n")
-    # CMake's FindThreads probe (a try_compile) inherits CMAKE_C/CXX_FLAGS_INIT
-    # above but NOT the add_link_options(), so it never sees -shared-libsan and
-    # uses clang's default static sanitizer runtime. That static runtime auto-links
-    # -lpthread, so the probe passes and sets CMAKE_HAVE_LIBC_PTHREAD=1 and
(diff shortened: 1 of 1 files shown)


amd-hsong:
@lumachad @stellaraccident Could you please take a look at this PR? Or feel free to suggest someone to review. Thanks.

lumachad:
I'm not sure about unconditionally adding -pthread to all linking steps, even for cases where we don't need it.

lumachad:
Would something like this work? If so, it would be best since it doesn't pass -pthread around:

```
      # Fix FindThreads detection under ASAN. The threads probe is a try_compile
      # that inherits CMAKE_C/CXX_FLAGS_INIT (-fsanitize) but NOT the directory
      # add_link_options() above, so it never sees -shared-libsan and links clang's
      # default STATIC sanitizer runtime, which auto-links -lpthread. The probe then
      # passes trivially, CMAKE_HAVE_LIBC_PTHREAD=1, and Threads::Threads is empty.
      # Feed the same shared-runtime flag to the probe via CMAKE_REQUIRED_LINK_OPTIONS
      # so it matches real link behavior and detects the real pthread dependency.
      # try_compile does NOT evaluate generator expressions, so guard on compiler id
      # at configure time (keeps -shared-libsan off gcc/gfortran, cf. issue #1782).
      # https://github.com/ROCm/rocm-libraries/issues/8889
      string(APPEND _stanza "if(CMAKE_CXX_COMPILER_ID MATCHES \"Clang\")\n")
      string(APPEND _stanza "  list(APPEND CMAKE_REQUIRED_LINK_OPTIONS -shared-libsan)\n")
      string(APPEND _stanza "endif()\n")
```

You'll need to manually run the ASAN builds to make sure it really does the right thing across the board, as we have issues with gfortran use.

amd-hsong:
@lumachad Thanks for the suggestion. I tried it and it did not work. It seems that the condition `if(CMAKE_CXX_COMPILER_ID MATCHES \"Clang\")\n")` did not trigger - looks like `CMAKE_CXX_COMPILER_ID` was populated too late.

But changing the condition to `if(CMAKE_CXX_COMPILER MATCHES \"clang\")\n")` worked. This is how I tested it: 
1. under TheRock, change the rocm-libraries submodule to point to commit https://github.com/ROCm/rocm-libraries/commit/83e1488a809de54fe31ca1f9acf43b2a74205def
2. `docker buildx build --file dockerfiles/build_manylinux_x86_64.Dockerfile --tag therock-manylinux-20260703 dockerfiles/`
3. `docker run --name therock-20260703 --rm -it -v $(pwd):/therock -w /therock --device /dev/kfd --device /dev/dri therock-manylinux-20260703 bash`
4. in the container, do:
```
  cmake --preset linux-release-asan -B build-asan -GNinja \
    -DTHEROCK_ENABLE_ALL=OFF \
    -DTHEROCK_ENABLE_PRIM=ON \
    -DTHEROCK_AMDGPU_FAMILIES=gfx942

  ninja -C build-asan rocThrust
  ```
  Let me know if this is sufficient enough. Thanks.

lumachad:
Added @marbre for a second opinion on this.

amd-hsong:
@marbre @lumachad Any further comments on this PR?

lumachad:
> @marbre @lumachad Any further comments on this PR?

Not from me. I'd still like to hear from @marbre. This is a broad change and we need to be extra careful on how it changes things.

marbre (commented):
This looks correct and as it only affects sanitized builds this seems not to broad. Before I approve and we merge, can we run an ASAN build in CI?

amd-hsong:
Stared Multi-Arch CI ASAN on this branch (`amd-hsong:therock_fix_asan_ci_link_error`): https://github.com/amd-hsong/TheRock/actions/runs/29348624186

lumachad:
Built with ASAN locally on my end. Seemed fine.

amd-hsong:
@marbre @lumachad Looks like the Multi-Arch CI ASAN runs I triggered above (https://github.com/amd-hsong/TheRock/actions/workflows/multi_arch_ci_asan.yml) will NOT work because they are from my personal fork of TheRock. I couldn't push to TheRock repo directly (due to permission issues) and had to create a fork in order to do this PR.

lumachad:
> @marbre @lumachad Looks like the Multi-Arch CI ASAN runs I triggered above (https://github.com/amd-hsong/TheRock/actions/workflows/multi_arch_ci_asan.yml) will NOT work because they are from my personal fork of TheRock. I couldn't push to TheRock repo directly (due to permission issues) and had to create a fork in order to do this PR.

Right. You shouldn't use a fork but rather a user branch.

amd-hsong:
Closing in favor of https://github.com/ROCm/TheRock/pull/6607 so I can manually run the **Multi-Arch CI ASAN** workflow there.

PR #5613, opened 2026-06-03
RFC: Optional HPC meta package in the ROCm Core SDK

saadrahim:
## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6363, opened 2026-07-06
fix: Ship hipTensor and rocALUTION as opt-in HPC tarball

evedovelli:
## Motivation

hipTensor and rocALUTION are treated as an opt-in HPC SDK expansion: in the native `.deb`/`.rpm` packages they live in the `amdrocm-hpc` metapackage, not the default `amdrocm` metapackage. The tarball distribution path did not follow this model — `build_tarballs.py` flattened all artifacts into a single tree, so the HPC libraries were dumped into the default ROCm tarball's `bin/`/`lib/`.

This PR ships the HPC SDK for tarballs per [RFC0014: AMD HPC SDK Release Model](https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0014-HPC-SDK-Release-Model.md). Per the RFC, the HPC SDK tarball is a **self-contained superset** of the ROCm Core tarball (full Core **plus** the HPC libraries), published as an **alternative** to the Core tarball — not an overlay/add-on. The default Core tarball no longer contains the HPC libraries.

## Technical Details

**Naming (per RFC0014).** The `core+hpc` tarball carries a `core+hpc` discriminator right after `therock-dist-`, naming the archive after what it contains:

```
therock-dist-core+hpc-{platform}-{family}-{version}.tar.gz   # full Core + HPC
therock-dist-{platform}-{family}-{version}.tar.gz            # Core only
```

The two are **alternatives, not companions** — a user downloads one or the other. They carry the same version and are produced from the same build, so the Core content in the `core+hpc` tarball is bit-identical to the standalone Core tarball.

**Tarball production (`build_tools/build_tarballs.py`):**
- Added `HPC_ARTIFACTS = ["hiptensor", "rocalution"]`; the default Core tarball excludes them (via `DEFAULT_EXCLUDED_ARTIFACTS`, which is `HPC_EXCLUDED_ARTIFACTS + HPC_ARTIFACTS`).
- Produces the self-contained `core+hpc` tarball by fetching the full artifact set with the HPC libraries **kept** (dropping only `fftw3`), guarded so it is only emitted when the build actually contains HPC artifacts (detected via a cheap HPC-only probe fetch). The `core+hpc` fetch reuses the shared download cache, so Core artifacts are not re-downloaded — only the HPC artifacts are new.
- Emits `-tests` variants of both the Core and `core+hpc` tarballs under `--include-test-tarballs`.
- Centralized filename construction in a `tarball_name(...)` helper (`core+hpc-` prefix, `-tests-` suffix) and the `core+hpc` fetch/emit logic in `add_hpc_tarball_tasks(...)`, shared by the per-family and multiarch (KPACK-split) paths.

**Artifact fetching (`build_tools/artifact_manager.py`):**
- Added an `--include-artifacts` allowlist flag (mirrors `--exclude-artifacts`, applied before exclusions), used to drive the HPC-presence probe.

**Install script (`build_tools/install_rocm_from_artifacts.py`):**
- `--hpc` (for `--release`/`--latest-release`) now downloads the single self-contained `core+hpc` tarball **instead of** the Core tarball — one download, no overlay. If no `core+hpc` tarball exists for the release/family (e.g. an older release predating the HPC SDK), it logs a clear warning and falls back to the Core tarball so the install still succeeds. (The `--run-id` path already installs HPC via `--hiptensor`/`--rocalution`.)
- `_release_asset_name(..., hpc=True)` builds the `core+hpc`-prefixed name; `is_non_default_tarball` / release discovery skip `core+hpc` tarballs so they are never mistaken for a default release.

**Upload (`build_tools/github_actions/upload_tarballs.py`):**
- `_select_multiarch_tarball_url` excludes `core+hpc` tarballs (alongside the existing `-tests-` exclusion) when picking the canonical multiarch URL.

## Test Plan

- `python -m pytest build_tools/tests/build_tarballs_test.py build_tools/tests/artifact_manager_tool_test.py build_tools/tests/install_rocm_from_artifacts_test.py build_tools/github_actions/tests/upload_tarballs_test.py` (all passing).
- Manual dry-run of the install script:
  - `install_rocm_from_artifacts.py --release <ver> --amdgpu-family <fam> --hpc --dry-run` → downloads the single `therock-dist-core+hpc-...` tarball.
  - Without `--hpc` → downloads the `therock-dist-...` Core tarball.
- End-to-end against a recent multi-arch CI run for a supported family (e.g. gfx942):
  - Default tarball contains no `libhiptensor.so` / `librocalution.so`.
  - A `core+hpc` tarball is produced containing the full Core tree **plus** those libs.

## Submission Checklist

ISSUE ID : ROCM-27538


schundu007:
## ❌ PR Gate — Gate failing

**Classification:** `mixed` · **Labels:** _none_

| | Category | Status | Matched check |
|---|---|---|---|
| ✅ | **lint** | success | `pre-commit` |
| ❌ | **tests** | failure | `Windows::release / Test Python (gfx110X-all, windows-gfx110X-gpu-rocm, 3.12, native) / Test ROCm wheels | gfx110X-all | py3.12 | native` +62 |
| ⚪ | **docs** | missing | _no matching check_ |
| ⚪ | **license** | missing | _no matching check_ |

**Runner recommendation:** `gpu` 🖥️ GPU — Mixed changes — use GPU runner to be safe

<sub>Updated by GitPulse · 2026-07-07 04:32 UTC</sub>

saadrahim (changes requested):
THIs PR has hard coded values. I don't think we can merge as is.

evedovelli:
> THIs PR has hard coded values. I don't think we can merge as is.

@saadrahim the PR is still a draft and need to be properly tested before we consider to merge it. Regarding the hard coded values, unless I missed something, they only exist in tests and mimic the existing tests for the core tarball.

build_tools/install_rocm_from_artifacts.py
```diff
@@ -712,6 +754,15 @@ def main(argv):
         help="Show what would be downloaded/copied without actually doing it",
     )
 
+    parser.add_argument(
+        "--hpc",
+        default=False,
+        action=argparse.BooleanOptionalAction,
+        help="Also install the opt-in HPC SDK expansion tarball (hipTensor, "
+        "rocALUTION) on top of the default release tarball. Only applies to "
```
evedovelli: Remove the mention to hipTensor and rocALUTION from this help.

marbre:
We discussed this offline and there a few things to clarify which ideally go into a spec or an RFC. Thus we will hold back on this PR until the items are clarified (e.g. should an hpc tarball be self-contained and include ROCm core or rather be an overlay  to be unpacked into the same destination as a ROCm core tarball).

evedovelli:
> We discussed this offline and there a few things to clarify which ideally go into a spec or an RFC. Thus we will hold back on this PR until the items are clarified (e.g. should an hpc tarball be self-contained and include ROCm core or rather be an overlay to be unpacked into the same destination as a ROCm core tarball).

@marbre I've made the changes to this PR according to the new definitions in https://github.com/ROCm/TheRock/pull/5613

marbre:
> > We discussed this offline and there a few things to clarify which ideally go into a spec or an RFC. Thus we will hold back on this PR until the items are clarified (e.g. should an hpc tarball be self-contained and include ROCm core or rather be an overlay to be unpacked into the same destination as a ROCm core tarball).
> 
> @marbre I've made the changes to this PR according to the new definitions in #5613

Let's land the RFC first and not pick up implementing it before it wasn't finalized. I wasted myself too much time on implementing things that than changes as the RFC wasn't landed yet. I think it is okay to take the time and wait until it is fully fleshed out.

Also see my latest response to #5613, we should avoid refactoring the refactoring at a short cadance.

PR #6377, opened 2026-07-07
fix: Declare rocprofiler-sdk's optional roc{decode,jpeg} deps

### LINKED ISSUE #6376: cmake failure: rocprofiler-sdk implicitly depends on rocdecode
After
```
cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES=gfx1151
ninja -C build
```
I see
```
[rocprofiler-sdk configure] CMake Error at /scratch/mgehre/TheRock/build/media-libs/rocdecode/build/rocdecode-config.cmake:16 (include):
[rocprofiler-sdk configure]   include could not find requested file:
[rocprofiler-sdk configure] 
[rocprofiler-sdk configure]     /scratch/mgehre/TheRock/build/media-libs/rocdecode/build/rocdecode-targets.cmake
[rocprofiler-sdk configure] Call Stack (most recent call first):
[rocprofiler-sdk configure]   /scratch/mgehre/TheRock/cmake/therock_subproject_dep_provider.cmake:73 (find_package)
[rocprofiler-sdk configure]   cmake/Modules/Findrocdecode.cmake:30 (find_package)
[rocprofiler-sdk configure]   /scratch/mgehre/TheRock/cmake/therock_subproject_dep_provider.cmake:73 (find_package)
[rocprofiler-sdk configure]   cmake/rocprofiler_config_interfaces.cmake:368 (find_package)
[rocprofiler-sdk configure]   CMakeLists.txt:109 (include)
```

This might be because rocprofiler-sdk tries to find rocdecode (via rocm-systems/projects/rocprofiler-sdk/cmake/Modules/Findrocdecode.cmake), but the dependency on rocdecode is not represented in https://github.com/ROCm/TheRock/blob/176bfea06c646b573a6b902d8cd1b333d4b88008/profiler/CMakeLists.txt#L92.

It like like after this trace is triggered once, rocprofiler-sdk's CMakeCache.txt will set
`rocdecode_DIR=/scratch/mgehre/TheRock/build/media-libs/rocdecode/build`
instead of the correct `/scratch/mgehre/TheRock/build/media-libs/rocdecode/dist/lib/cmake/rocdecode`
which will make all subsequent builds also fail.

Please fix this instance of the problem and potentially provide some detection to avoid undeclared dependencies becoming races.

marbre:
## Motivation

rocprofiler-sdk performs an unconditional but optional `find_package(rocdecode)`/`find_package(rocjpeg)` to detect its rocDecode and rocJPEG trace features. These packages were never declared as dependencies, so TheRock's dependency provider fell through to the system resolver, which latched onto rocdecode's build tree and tried to include a `rocdecode-targets.cmake` that only exists after install, failing the rocprofiler-sdk configure.

## Technical Details

This patch declares both components as optional runtime deps to resolve them from the super-project dist. Deps are guarded so that these dependencies are still optional, as they are currently disabled in CI and the build topology would need a significant re-ordering.

Fixes #6376

## Submission Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -496,6 +496,10 @@ add_subdirectory(compiler)
 add_subdirectory(third-party)
 add_subdirectory(core)
 add_subdirectory(debug-tools)
+# media-libs (rocdecode/rocjpeg) must be added before profiler: rocprofiler-sdk
+# performs optional find_package(rocdecode)/find_package(rocjpeg) for trace
+# feature detection.
+add_subdirectory(media-libs)
 # Note that rocprofiler-register is in base and is what higher level clients
 # depend on. The profiler itself is independent.
 add_subdirectory(profiler)
@@ -507,7 +511,6 @@ add_subdirectory(comm-libs)
 add_subdirectory(storage-libs)
 add_subdirectory(math-libs)
 add_subdirectory(ml-libs)
-add_subdirectory(media-libs)
 
 if(THEROCK_BUILD_TESTING)
   add_subdirectory(examples)
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -6,6 +6,18 @@ if(THEROCK_ENABLE_MPI)
   set(_openmpi_optional_dep therock-openmpi)
 endif()
 
+# rocprofiler-sdk performs optional find_package(rocdecode)/find_package(rocjpeg)
+# to detect the rocDecode/rocJPEG trace features. Declare them as dependencies
+# when available so they resolve from the super-project dist rather than
+# escaping to a stray build tree.
+set(_rocprofiler_sdk_optional_deps)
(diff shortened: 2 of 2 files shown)


marbre:
@dgaliffiAMD, @jbonnell-amd, this fixes the race but in CI rocprofiler-system will still always be build without those two optional deps. If there is a desire to change that, we should file a follow up issue.

mgehre-amd (commented):
Thanks for the quick fix!

mgehre-amd (approved):

dgaliffiAMD (approved):
Thank you. You're change looks good, but I'd like someone from the rocprofiler-sdk team to weigh-in. I will add them to the review.

dan-garvey (approved):

dan-garvey:
failures seem unrelated

venkat1361 (approved):

PR #6378, opened 2026-07-07
feat(rocgdb): Add rocgdb-corefile test component with dedicated runner

lumachad:
Add a rocgdb-corefile entry to the test matrix that runs the five GPU corefile-specific tests explicitly via --tests, routed to the linux-gfx942-gpu-rocm-mathlib runner regardless of the AMDGPU family under test.

Guard the per-component runner selection loop to skip components that already have test_runner pre-pinned, and document the runner assignment contract above test_matrix.

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -133,6 +133,24 @@ _rocgdb_common = {
     "container_options": ["--cap-add=SYS_PTRACE"],
 }
 
+
+# Runner assignment for test components
+# =====================================
+# Most components have their runner selected at runtime by the per-component loop
+# below, which draws from the AMDGPU-family runner pool configured in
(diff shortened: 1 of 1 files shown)


geomin12 (approved):
cool it works! however, do we have enough capacity for this job to run as frequent?

nice though! this'll work great, i see it runs fairly fast too

lumachad:
> cool it works! however, do we have enough capacity for this job to run as frequent?
> 
> nice though! this'll work great, i see it runs fairly fast too

They are just a handful of tests and should run fairly quickly. So I expect the capacity will be OK for now. But I should monitor it closely if/when we merge it.

I'm also investigating with @ndhillon07 adding a bit more capacity to cover these jobs.

lumachad:
Blocked on https://github.com/ROCm/ROCgdb/pull/249.

lumachad:
This should now run clean after the last ROCgdb bump.

Issue #6173, opened 2026-06-26
[Issue] Linux  PyTorch release/2.12: test_mempool_empty_cache_inactive fails — fatal error: 'hipblas/hipblas.h' file not found


amd-chiranjeevi:
## Context

- **Workflow:** `multi_arch_release_linux_pytorch_wheels.yml` (ROCm/rockrel)
- **Failing run:** [Run 28213360075 (2026-06-26)](https://github.com/ROCm/rockrel/actions/runs/28213360075)
- **Failing job:** [Test PyTorch | gfx110X-all (job 83589518189)](https://github.com/ROCm/rockrel/actions/runs/28213360075/job/83589518189)
- **Platform:** Linux
- **Affected arch:** `gfx110X-all, gfx1151, gfx94x, gfx950, gfx120x`
- **PyTorch version:** `2.12 , nightly`
- **Python version:** `3.10`
- **ROCm version:** `7.15.0a20260626`
- **TheRock ref used:** `e9a508103a54e64cd2f85c052909bd56d0daf834`

## Summary

9 tests fail in the PyTorch test suite. The primary structural failure is `test_cuda.py::TestMemPool::test_mempool_empty_cache_inactive`, which calls `torch.utils.cpp_extension.load_inline` to JIT-compile a `dummy_allocator` C++ extension at test runtime. Compilation fails because `ATen/hip/Exceptions.h` (line 5, shipped inside the installed wheel) unconditionally includes `<hipblas/hipblas.h>`, but the hipBLAS development headers are not available in the test container (`ghcr.io/rocm/no_rocm_image_ubuntu24_04`).

Five additional test failures (`test_memory_compile_regions`, `test_memory_plots`, `test_memory_plots_free_segment_stack`, `test_memory_snapshot`, `test_memory_snapshot_script`, `test_memory_snapshot_with_cpp`) likely share the same root cause (they all attempt C++ extension JIT compilation that hits the same missing header).

`test_no_triton_on_import` and `test_layer_norm_backwards_eps` are separate failures (see below).

## Error Log

```
In file included from /github/home/.cache/torch_extensions/py310_rocm71460850/dummy_allocator/main_hip.cpp:5:
/__w/rockrel/rockrel/.venv/lib/python3.10/site-packages/torch/include/ATen/hip/Exceptions.h:5:10: fatal error: 'hipblas/hipblas.h' file not found
    5 | #include <hipblas/hipblas.h>
[...]

PR #4936, opened 2026-04-29
Install rocm[devel] pkg during pytorch tests

pragupta:
## Motivation

This is required  for some pytorch UTs to pass. 
```
external-builds/pytorch/pytorch/test/test_cuda.py::TestMemPool::test_mempool_expandable [1/2] c++ ...
FAILED: [code=1] main_hip.o
In file included from /github/home/.cache/torch_extensions/py314_cpu/dummy_allocator/main_hip.cpp:5:
/__w/TheRock/TheRock/.venv/lib/python3.14/site-packages/torch/include/ATen/hip/Exceptions.h:5:10: fatal error: 'hipblas/hipblas.h' file not found
    5 | #include <hipblas/hipblas.h>
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
...
subprocess.CalledProcessError: Command '['ninja', '-v']' returned non-zero exit status 1.
RuntimeError: Error building extension 'dummy_allocator'
```

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6379, opened 2026-07-07
fix(ci): install rocm[devel] in test_pytorch_wheels.yml

amd-maradosa:
Issue #6173 

## Motivation

`test_pytorch_wheels.yml` installs only torch[device-gfxNNN], which brings the ROCm runtime but not the development headers. Tests that JIT-compile a C++ extension (e.g. test_cuda.py::TestMemPool::test_mempool_empty_cache_inactive) pull in `ATen/hip/Exceptions.h`, which includes `<hipblas/hipblas.h>`. Since that header ships only in `rocm[devel]`, the compile fails with `fatal error: 'hipblas/hipblas.h' file not found

## Technical Details

Add two steps after "Set up virtual environment", exactly mirroring the already working `test_pytorch_wheels_full.yml`:
  - Install ROCm devel packages — pip install "rocm[devel,...]", which provides the hipBLAS/hipSPARSE/etc. headers.
  - Initialize ROCm SDK and configure environment — rocm-sdk init and export ROCM_PATH/ROCM_HOME (plus include/lib/cmake vars and CC/CXX) so torch.utils.cpp_extension resolves the ROCm include dir instead of guessing the empty venv root.

## Test Plan

Run test_pytorch_wheels.yml from this branch (gfx94X-dcgpu, torch 2.12.0+rocm7.15.0a20260626, py3.10, release/2.12). [CI Run](https://github.com/ROCm/TheRock/actions/runs/28861385152)
## Test Result

dummy_allocator now compiles against rocm_sdk_devel/include and test_mempool_empty_cache_inactive passes. Note: the other failures listed in #6173 (test_memory_*, torch.cuda init SIGABRT) have unrelated root causes and are out of scope for this PR.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -211,6 +211,58 @@ jobs:
               --activate-in-future-github-actions-steps
           fi
 
+      - name: Install ROCm devel packages
+        timeout-minutes: 15
+        run: |
+          if [ "${{ inputs.multi_arch }}" = "true" ]; then
+            pip install "rocm[devel,${{ steps.expand.outputs.device_extras }}]" \
+              --index-url=${{ inputs.package_index_url }}
+          else
+            pip install "rocm[devel]" \
+              --index-url=${{ inputs.package_index_url }}/${{ inputs.amdgpu_family }}
+          fi
+
+      - name: Initialize ROCm SDK and configure environment
+        run: |
+          rocm-sdk init
+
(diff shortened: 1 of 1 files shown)


marbre (changes requested):
See prior discussion on https://github.com/ROCm/TheRock/pull/4936 and https://github.com/ROCm/TheRock/pull/3814.

amd-maradosa:
Okay, I see this is not the right way to move forward since end users don't need the `rocm[devel]` package and this might mask out the real-user runtime bugs. Thanks for letting me know

PR #6383, opened 2026-07-07
refactor: Default local package indexes to flat multi-arch layout

### LINKED ISSUE #6382: Cleanup generate_local_index.py and drop single-family support
`generate_local_index.py` still treats the older single-directory indexing path as its default CLI behavior, even though flat kpack-split multi-arch packages are now the preferred package layout. Switch the default to multi-arch kpack enabled behavior and flag guard the kpack-disabled path. The pure single-family path can be dropped.

marbre:
## Motivation

Updates `generate_local_index.py` so its default CLI mode writes a top-level `index.html` for flat kpack-split package directories. Keeps the kpack-disabled per-family layout available through an explicit `--per-family-indexes` option.

## Technical Details

This preserves the existing `generate_multiarch_indexes()` helper for callers such as `upload_python_packages.py`, and adds CLI coverage for the new default mode, custom output paths, pattern filtering, etc.

Closes #6382.

## Test Plan

Added unit tests for the new behavior.

## Test Result

Unit tests pass.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/python/generate_local_index.py
+++ b/build_tools/packaging/python/generate_local_index.py
@@ -5,9 +5,11 @@
 """Generate simple HTML index files for pip --find-links from local files.
 
 This module provides utilities to generate index.html files that pip can use
-with the --find-links option. It supports both simple single-directory indexes
-and multi-arch builds where each GPU family needs its own index that includes
-both family-specific packages and generic packages from the parent directory.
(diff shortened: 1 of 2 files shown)


araravik-psd (approved):
LGTM

ScottTodd (approved):

Issue #6095, opened 2026-06-24
[Issue]: PyTorch CI doesn't pin numpy version


zjliu-amd:
### Problem Description

The latest numpy release (2.5.0, released Jun 21) changed the behaviour of `threshold` function, causing pytorch UT to fail: https://github.com/ROCm/rockrel/actions/runs/28071131699/job/83114331597

This isn't a problem in upstream CI jobs because pytorch pins numpy version, but the Rock CI doesn't. Relevant files:
https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/requirements-test.txt doesn't pin numpy version

https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml#L214-L217 and https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels_full.yml#L333-L337 install dependencies

### Operating System

Ubuntu

### CPU

N/A

### GPU

N/A

### ROCm Version

nightly

### Steps to Reproduce

```
python -m pytest -vv -s \
  test/test_unary_ufuncs.py::TestUnaryUfuncsCUDA::test_reference_numerics_normal__refs_nn_functional_threshold_cuda_uint8
```

Run this test with numpy 2.5.0 installed, this will fail. Changing numpy version to 2.4.6 or any version before 2.5, it passes

PR #6384, opened 2026-07-07
fix(ci): Install pytorch pinned requirements in core UT workflow

zjliu-amd:
## Motivation

PyTorch Core UT workflow only installs the requirements specified in TheRock, but not the more detailed pytorch requirements, which cause numpy version not pinned and therefore leading to [UT failures](https://github.com/ROCm/TheRock/issues/6095). Besides, this is also causing a lot of UTs missing dependencies such as scipy. This requirements.txt file IS installed in [test_pytorch_wheels_full](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels_full.yml#L335-L336), but not in this workflow.

### DIFF
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -208,6 +208,7 @@ jobs:
       - name: Install test requirements
         run: |
           python -m pip install -r external-builds/pytorch/requirements-test.txt
+          python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
           pip freeze
 
       - name: Run rocm-sdk sanity tests
--- a/external-builds/pytorch/README.md
+++ b/external-builds/pytorch/README.md
@@ -243,6 +243,9 @@ capabilities tailored for AMD ROCm GPUs. See the script for detailed
 instructions. Here are a few examples:
 
 ```bash
+# Install test dependencies
+python -m pip install -r pytorch/.ci/docker/requirements-ci.txt
+
 # Basic usage (auto-detect everything, no extra args):
 python run_pytorch_tests.py
 


HereThereBeDragons:
similar is done in https://github.com/ROCm/TheRock/pull/6400

lucbruni-amd:
@HereThereBeDragons I'm in favour of this PR since it pulls the other missing test deps (not just `numpy`) thus resolving other UT issues. It also fixes the dependency drift issue and the need to pin deps. Should we close https://github.com/ROCm/TheRock/pull/6400 and https://github.com/ROCm/TheRock/pull/6423 and proceed with this one?

ethanwee1:
I am in favor of this move.

HereThereBeDragons:
yeah i think this seems more elegant. but we probably then want some comment in requirements-test.txt that for the ci testing we also use the pytorch requirements.txt and some versions may be overwritten?

zjliu-amd:
This change, however, could lead to a dependency installation failure on high python versions (such as 3.14). An example is https://github.com/ROCm/TheRock/actions/runs/28899244667/job/85731743101
Some dependencies don't provide prebuilt wheels for python 3.14 in the pinned versions, which result in pip building them from source. But the compiler in the CI can't find some headers they need. Could be the same problem as https://github.com/actions/setup-python/issues/489
We can possibly fix this by either pinning some dependencies to a higher version which provides python 3.14 binary, or by exporting an env var as suggested in https://github.com/actions/setup-python/issues/489. There are 4 dependencies need to be built for pytorch 2.11, possibly less on 2.12 and more on 2.10 (haven't confirmed).

Update: Fixed by bumping dependency version in pytorch

ethanwee1:
Rather than pin an old numpy in TheRock long-term, we're going to fix this at the source and let TheRock inherit upstream's pins. Two upstream approaches are staged:

First, cap numpy in upstream test CI: [Cap numpy <2.5 in requirements-ci.txt to avoid strict-cast regression by ethanwee1 · Pull Request #1…](https://github.com/pytorch/pytorch/pull/189566) adds a <2.5 upper bound to the numpy entries in .ci/docker/requirements-ci.txt (keeping the existing floors), so CI resolves to the 2.4 line (currently 2.4.6), the last release before numpy tightened negative-int to uint8 scalar casting. Least invasive, but since the numpy change was deliberate it's only a stopgap.

Second, fix the reference itself: [Fix nn.functional.threshold OpInfo reference for strict NumPy uint8 casting by ethanwee1 · Pull Requ…](https://github.com/pytorch/pytorch/pull/189273) makes the nn.functional.threshold OpInfo test dtype-safe (non-negative fill for uint8, -9 kept for signed/float), so it survives any numpy. An upstream reviewer (rgommers) already suggested this here: [nn.functional.threshold OpInfo reference raises OverflowError on unsigned dtypes with NumPy >= 2.5 ·…](https://github.com/pytorch/pytorch/issues/189267#issuecomment-4917213219)

Once the reference fix (option 2) is validated upstream, we'll drop TheRock's external-builds/pytorch/requirements-test.txt numpy pins entirely so we inherit upstream's requirements-ci.txt versions.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -214,6 +208,7 @@ jobs:
       - name: Install test requirements
         run: |
           python -m pip install -r external-builds/pytorch/requirements-test.txt
+          python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
```
ScottTodd, line 211: Note that this `external-builds/pytorch/pytorch` directory is the default but it is _not_ the path we use when _building_ on Windows. Looks like we do use it here for _testing_ consistently on both platforms though.

ScottTodd, line 211: Looks like this will bump the requirements install time from ~5 seconds (example: https://github.com/ROCm/TheRock/actions/runs/29119539622/job/86451121264#step:11:1) to ~1m07s (example: https://github.com/ROCm/TheRock/actions/runs/29264970035/job/86867806066#step:11:1)

If pytorch needs all of the extra packages, that's fine.

To optimize, we can look into tuning a persistent pip cache on our GPU test runners and/or switching from `pip` to `uv`.

ScottTodd (approved):
Agreed that this is nicer than https://github.com/ROCm/TheRock/pull/6400

.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 211: We could also document the test requirements here: https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#runningtesting-pytorch

Note that we already suggest `pip install -r pytorch/.ci/docker/requirements-ci.txt` in the summary that is posted on test runs: https://github.com/ROCm/TheRock/blob/2505830f5e2c1ef2280309d6fae79aa2a3861ded/build_tools/github_actions/summarize_test_pytorch_workflow.py#L85-L97

See where that markdown is used here for example: https://github.com/ROCm/TheRock/actions/runs/29119539622

Image

zjliu-amd: This looks fine to me, as long as we get more stable and comprehensive dependency installation

zjliu-amd:
@ScottTodd There's only one thing I want to ask for your opinion: currently some of the pinned dependencies don't have prebuilt wheels for python 3.14, which is in our test matrix. If we run `pip install -r requirements-ci.txt` in python 3.14 workflows, it will try to build those dependencies from source. However, the build will fail because the rock CI's environment lacks some env vars or packages.
We already have a workaround in place, which is to bump dependencies version in `requirements-ci.txt` to a version that ships prebuilt 3.14 wheels for python 3.14 only, but this could potentially cause subtle bugs if the dependencies change their behaviour in new versions, and that would also require us to maintain `requirements-ci.txt` for each release branch and python version separately, instead of inheriting upstream's pin.
I'm also testing if we can update the workflow so we can build dependencies from source, here are my changes: https://github.com/ROCm/TheRock/compare/main...users/zjliu-amd/pip-build-from-src. This should enable the pip to build from source successfully ([example](https://github.com/ROCm/TheRock/actions/runs/29342829395/job/87118917406)).
Which way do you think is a better solution, or do you have other insights? I'd like to know your opinion :)

ScottTodd:
We need to be very careful with installing system-wide `-dev` packages. The rocm (and torch) packages that we build and distribute are intended to be portable across multiple linux distributions and self-contained (i.e. no/minimal dependencies on libraries that must be installed separately). Installing `-dev` packages includes the risk of packages pulling in _other_ `-dev` packages and the resulting binaries referencing system libraries. For more details see
* https://github.com/ROCm/TheRock/tree/main/dockerfiles#build_manylinux_dockerfile
* https://github.com/ROCm/TheRock/blob/main/dockerfiles/build_manylinux_x86_64.Dockerfile
* https://gist.github.com/stellaraccident/866acb0cae35e9f59c0c67d2fe3773c1

I'd prefer to use requirements that include prebuilt wheels, and we can even host such wheels ourselves in our python package mirror repositories like https://rocm.nightlies.amd.com/whl-multi-arch/

zjliu-amd:
That makes sense. I'll update the requirements file in pytorch, then we can merge this PR

PR #6391, opened 2026-07-07
 fix: Add dedicated devel package dependency handling in kpack mode

raramakr:
Devel packages require special dependency resolution in multi-arch (kpack)
  mode - they should depend on the full meta package (host + all devices)
  rather than just the host package.

  Key changes:
  - Add process_devel_dependencies_kpack() for devel-specific dependency logic
  - Use GFX_META for artifact filtering since devel packages are arch-agnostic
  - All devel packages now return GfxArch=False in kpack mode (no host/device split)

  Dependency resolution examples (kpack mode):
  - amdrocm-blas-devel -> amdrocm-blas8.2 (full meta, not host)
  - amdrocm-core-devel -> amdrocm-core8.2 + devel deps (with version suffix)
  
  JIRA ID: https://github.com/ROCm/TheRock/issues/6406
  
**Test Output:**
  [devel_dependency_output.txt](https://github.com/user-attachments/files/29776962/devel_dependency_output.txt)

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -164,6 +164,20 @@ def is_meta_package(pkg_info):
     return is_key_defined(pkg_info, "Metapackage")
 
 
+def is_devel_package(pkg_info):
+    """
+    Verifies whether this is a development package (ends with -devel).
+
+    Parameters:
+    pkg_info (dict): A dictionary containing package details.
+
+    Returns:
+    bool: True if package name ends with -devel, False otherwise.
+    """
+    pkg_name = pkg_info.get("Package", "")
+    return pkg_name.endswith("-devel")
+
+
 def is_rpm_stripping_disabled(pkg_info):
     """
     Verifies whether Disable_RPM_STRIP key is enabled for a package.
@@ -257,10 +271,9 @@ def is_gfxarch_package(
            cannot classify the package as gfx-arch-specific without an artifact path.
     """
     if enable_kpack:
-        pkgname = pkg_info.get("Package", "")
-        # Only non-metapackage -devel should be non-gfxarch
-        # Metapackages like amdrocm-core-devel should create arch-specific variants
-        if pkgname.endswith("-devel") and not is_meta_package(pkg_info):
+        # All devel packages (meta and non-meta) are GfxArch=False in kpack mode
+        # Devel packages don't get host/device split - they depend on full runtime
+        if is_devel_package(pkg_info):
             return False
 
     # In kpack mode, verify arch-specific artifacts exist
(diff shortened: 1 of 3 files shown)


build_tools/packaging/linux/packaging_utils.py
arvindcheru, line 276: DEB Build Failing in CI - amdrocm-dnn-test10.1 depended on malformed names: amdrocm-blas10.1-, amdrocm-rand10.1- amdrocm-dnn-test is built as a simple (no-arch) package
JSON still has "Gfxarch": "True", but kpack routing ignores that if there are no gfx-specific artifact dirs. CI logged Building versioned variant for amdrocm-dnn-test. is_gfxarch_package() uses artifact dirs, not only the JSON flag

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -663,7 +725,80 @@ def process_main_dependencies_kpack(
...
+    config: Configuration object containing package metadata
+
+    Returns: Comma-separated string of versioned dependencies
+    """
+    is_meta = is_meta_package(pkg_info)
+    pkg_name = pkg_info.get("Package")
+    dep_list = list(pkg_info.get(field_key, []) or [])
+
+    # Filter deps without artifacts
+    dep_list = filter_dependencies_by_artifacts(
```
arvindcheru, line 757: Filter deps without artifacts

```
+   Devel packages are built without a gfx_arch, but their runtime
+   dependencies target the full kpack meta package. Use GFX_META explicitly
+   so the generic empty-arch path does not retain gfx dependencies that
+   cannot be named without an architecture.
     dep_list = filter_dependencies_by_artifacts(
-    dep_list, config.artifacts_dir, config.gfx_arch
+   dep_list, config.artifacts_dir, GFX_META
     )

```
We may need to update this .

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -1056,6 +1187,12 @@ def has_artifact_for_arch(pkg_name, artifacts_dir, gfx_arch):
     if is_meta_package(pkg_info):
         return True
 
+    # For GFX_META or empty gfx_arch, check if ANY arch artifacts exist
```
arvindcheru, line 1190: We may need enhance here as below

```
`-    For GFX_META or empty gfx_arch, check if ANY arch artifacts exist
-    (regular GfxArch packages have gfx942, gfx1100, etc. but not gfx_meta)
-    Empty gfx_arch is used for devel packages - they just need deps to exist
-    if gfx_arch in (GFX_META, ""):
+    For GFX_META, check if ANY arch artifacts exist (regular GfxArch packages
+    have gfx942, gfx1100, etc. but not gfx_meta). An empty gfx_arch is not a
+    valid target for a gfx package: retaining it would later generate a
+    malformed package name with a trailing hyphen.
+    if gfx_arch == GFX_META:
`
```

arvindcheru (dismissed):
Request to check the latest CI failure on DEB Build:

- Devel packages are built without a gfx_arch, but their runtime dependencies target the full kpack meta package. Use GFX_META explicitly so the generic empty-arch path does not retain gfx dependencies that cannot be named without an architecture.
- RPM simulate is rpm -Uvh --test --nodeps, so bad Requires are ignored. DEB simulate is apt install --simulate, which resolves Depends. So this will be visible on DEB only.
- Please go through the comments, and the sample code 
- Update the unit test for the new support also.

raramakr:
To keep things simple, I moved the refactoring changes to a separate PR: https://github.com/ROCm/TheRock/pull/8155
 This PR contains only the development package dependency logic. cc @arvindcheru

raramakr:
> Request to check the latest CI failure on DEB Build:
> 
> * Devel packages are built without a gfx_arch, but their runtime dependencies target the full kpack meta package. Use GFX_META explicitly so the generic empty-arch path does not retain gfx dependencies that cannot be named without an architecture.
> * RPM simulate is rpm -Uvh --test --nodeps, so bad Requires are ignored. DEB simulate is apt install --simulate, which resolves Depends. So this will be visible on DEB only.
> * Please go through the comments, and the sample code
> * Update the unit test for the new support also.

Fixed the installation issue  . Used GFX_META explicitly for devel packages
Also added unit tests

build_tools/packaging/linux/packaging_utils.py
raramakr: Done

build_tools/packaging/linux/packaging_utils.py
raramakr: Done

build_tools/packaging/linux/packaging_utils.py
raramakr: The malformed dependency name is resolved.
But the depenedency of amdrocm-dnn-test  need to be verified and should be addressed in separate PR.
 Not related to this PR

arvindcheru, line 1125: Stale comment in resolve_versioned_dependency_list still mentions -devel metapackages; devel never enters that function now. Please check

PR #6394, opened 2026-07-07
  fix: add devel package dependency tree documentation

raramakr:
Document the dependency resolution rules for development packages in
  kpack mode. Adds Section 4 covering four use cases:
  - Devel non-meta with GfxArch=False runtime dependency
  - Devel non-meta with GfxArch=True runtime dependency
  - Devel meta with GfxArch=False devel dependencies
  - Devel meta with GfxArch=True devel dependencies

  Key design decisions documented:
  - All devel packages are GfxArch=False (no host-device split)
  - GfxArch=True runtime deps resolve to full meta (host + all devices)
  - Devel meta packages depend on runtime meta for complete SDK
  
  JIRA ID: https://github.com/ROCm/TheRock/issues/6406

### DIFF
--- a/docs/packaging/nativepackage_dependency_tree.md
+++ b/docs/packaging/nativepackage_dependency_tree.md
@@ -1,10 +1,11 @@
 # ROCm Package Dependency Trees
 
-This document shows the dependency tree structure for the three types of packages in the ROCm packaging system:
+This document shows the dependency tree structure for the package types in the ROCm packaging system:
 
 1. **Non-GfxArch Package** - Architecture-independent packages
 1. **GfxArch Package** - Architecture-specific packages (host-device split)
 1. **Meta Package** - Aggregator packages with no content
+1. **Devel Package** - Development packages (headers, cmake configs, pkgconfig)
 
 ______________________________________________________________________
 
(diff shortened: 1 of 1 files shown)


docs/packaging/nativepackage_dependency_tree.md
```diff
@@ -1,10 +1,11 @@
 # ROCm Package Dependency Trees
 
-This document shows the dependency tree structure for the three types of packages in the ROCm packaging system:
+This document shows the dependency tree structure for the package types in the ROCm packaging system:
 
 1. **Non-GfxArch Package** - Architecture-independent packages
 1. **GfxArch Package** - Architecture-specific packages (host-device split)
 1. **Meta Package** - Aggregator packages with no content
+1. **Devel Package** - Development packages (headers, cmake configs, pkgconfig)
```
nunnikri, line 8: Can we make this index with softlinks to the corresponding sections?

docs/packaging/nativepackage_dependency_tree.md
```diff
@@ -491,7 +492,456 @@ Dependencies = [all deps from package.json, versioned]
...
+
+#### Package Configuration
+
+```json
+{
+  "Package": "amdrocm-developer-tools-devel",
+  "Metapackage": "True",
+  "Gfxarch": "False",
+  "DEBDepends": [
+    "amdrocm-llvm-devel",
```
nunnikri, line 718: Hope this is given as only illustration and not the actual package.

docs/packaging/nativepackage_dependency_tree.md
```diff
@@ -491,7 +492,456 @@ Dependencies = [all deps from package.json, versioned]
...
+______________________________________________________________________
+
+### 4.3 Devel Meta with GfxArch=False Devel Dependencies
+
+**Example**: `amdrocm-developer-tools-devel` depends on devel packages for GfxArch=False components
+
+**Characteristics**:
+
+- Meta package aggregating devel packages
+- All child devel packages have GfxArch=False runtime dependencies
```
nunnikri, line 707: This language is bit confusing. 
I think what you meant is devel package has GfxArch=False, and the dependencies of these devel packages also have GfxArch=False. Or we have to define the terms in the beginning. What is child means.

docs/packaging/nativepackage_dependency_tree.md
```diff
@@ -491,7 +492,456 @@ Dependencies = [all deps from package.json, versioned]
...
+______________________________________________________________________
+
+### 4.2 Devel Non-Meta with GfxArch=True Runtime Dependency
+
+**Example**: `amdrocm-blas-devel` depends on `amdrocm-blas` (GfxArch=True)
+
+**Characteristics**:
+
+- Devel package for a gfxarch runtime package
+- Runtime dependency resolves to **full meta** (host + all device packages)
```
nunnikri, line 595: Instead of runtime dependency more accurate description would be dependency with device specific packages or dependency which has GfxArch=True

docs/packaging/nativepackage_dependency_tree.md
```diff
@@ -491,7 +492,456 @@ Dependencies = [all deps from package.json, versioned]
...
+______________________________________________________________________
+
+### 4.4 Devel Meta with GfxArch=True Devel Dependencies
+
+**Example**: `amdrocm-core-devel` depends on devel packages for GfxArch=True components
+
+**Characteristics**:
+
+- Meta package aggregating devel packages
+- Child devel packages have GfxArch=True runtime dependencies
```
nunnikri, line 796: We telling the runtime dependency would be the second level of dependency right? If we can make this bit more clear or mentioned the first level dependencies first and second level dependencies later

nunnikri (commented):
Overall looks good. Thanks for getting the documentation complete. The images looks good. The written language needs to be more consistent and less ambiguous.

PR #6395, opened 2026-07-07
artifact_manager.py: add --find-matching-commit

mgehre-amd:
Resolving a --run-id from a checked-out commit today requires a separate manual step with `find_artifacts_for_commit.py`. This wires that lookup directly into `fetch`, so a developer can go from a commit to a bootstrapped build tree in less commands without knowing the CI run ID ahead of time.

Now we can go from zero to building hipBLASLt in just
```
STAGE_NAME=math-libs
AMDGPU_FAMILY=gfx1151
BUILD_DIR=build
THEROCK_ROOT="/scratch/mgehre/TheRock"

python3 ./build_tools/artifact_manager.py fetch \
  --find-matching-commit \
  --stage "$STAGE_NAME" \
  --amdgpu-families "$AMDGPU_FAMILY" \
  --output-dir "$BUILD_DIR" \
  --bootstrap

cmake -B "$BUILD_DIR" -S . -GNinja \
  -DTHEROCK_AMDGPU_FAMILIES="$AMDGPU_FAMILY" \
  -DTHEROCK_DIST_AMDGPU_FAMILIES="$AMDGPU_FAMILY" \
  -DTHEROCK_ENABLE_ALL=OFF \
  -DTHEROCK_ENABLE_BLAS=ON

ninja -C $BUILD_DIR hipBLASLt+dist
```

### DIFF
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -1171,10 +1171,17 @@ def _add_target_args(parser: argparse.ArgumentParser):
     )
 
 
-def _add_backend_args(parser: argparse.ArgumentParser):
-    """Add common backend-related arguments to a subparser."""
+def _add_backend_args(
+    parser: argparse.ArgumentParser,
+) -> argparse._MutuallyExclusiveGroup:
+    """Add common backend-related arguments to a subparser.
+
+    Returns the mutually exclusive group that owns --run-id so a subcommand can
+    add alternative run-id sources (e.g. fetch's --find-matching-commit) to it.
+    """
     _add_common_args(parser)
-    parser.add_argument(
+    run_id_group = parser.add_mutually_exclusive_group()
+    run_id_group.add_argument(
         "--run-id",
         type=str,
         default=os.getenv("THEROCK_RUN_ID", os.getenv("GITHUB_RUN_ID", "local")),
@@ -1199,6 +1206,7 @@ def _add_backend_args(parser: argparse.ArgumentParser):
         default=os.getenv("THEROCK_LOCAL_STAGING_DIR"),
         help="Local staging directory (sets THEROCK_LOCAL_STAGING_DIR)",
     )
+    return run_id_group
 
 
 def main(argv: Optional[List[str]] = None):
@@ -1213,7 +1221,7 @@ def main(argv: Optional[List[str]] = None):
     fetch_parser = subparsers.add_parser(
         "fetch", help="Fetch inbound artifacts for a stage"
     )
-    _add_backend_args(fetch_parser)
+    fetch_run_id_group = _add_backend_args(fetch_parser)
     fetch_parser.add_argument(
         "--stage",
         type=str,
@@ -1221,7 +1229,9 @@ def main(argv: Optional[List[str]] = None):
         help="Build stage name (default: 'all' fetches all artifacts)",
     )
     _add_target_args(fetch_parser)
-    fetch_parser.add_argument(
+    # Mutually exclusive with --run-id (same group): --find-matching-commit
+    # resolves the run ID itself, so passing both on the command line is an error.
+    fetch_run_id_group.add_argument(
         "--find-matching-commit",
         action="store_true",
         help="Resolve --run-id automatically by finding the CI workflow run "
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -48,6 +48,7 @@ from dataclasses import dataclass
 import os
 import platform as platform_module
 import shutil
+import subprocess
 import sys
 import threading
 import time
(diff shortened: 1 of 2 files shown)


build_tools/artifact_manager.py
```diff
@@ -1101,6 +1154,13 @@ def main(argv: Optional[List[str]] = None):
         help="Build stage name (default: 'all' fetches all artifacts)",
     )
     _add_target_args(fetch_parser)
+    fetch_parser.add_argument(
+        "--find-matching-commit",
+        action="store_true",
+        help="Resolve --run-id automatically by finding the CI workflow run "
+        "that produced artifacts for the current HEAD commit. "
+        "Uses the same lookup as find_artifacts_for_commit.py.",
+    )
```
ScottTodd: I think adding a new argument like this is okay, but I'd like to be careful about putting too much functionality into individual scripts. We should aim to follow the unix tools philosophy of having each tool have a well-defined role and allowing tools to work together (e.g. by piping output from one to another).

This is a bit of a slippery slope in that we may also want a `--find-baseline-commit` that does some more searching while `--find-matching-commit` just checks the current commit itself.

build_tools/artifact_manager.py
```diff
@@ -1101,6 +1154,13 @@ def main(argv: Optional[List[str]] = None):
         help="Build stage name (default: 'all' fetches all artifacts)",
     )
     _add_target_args(fetch_parser)
+    fetch_parser.add_argument(
+        "--find-matching-commit",
+        action="store_true",
+        help="Resolve --run-id automatically by finding the CI workflow run "
```
ScottTodd: Could this be moved into `_add_backend_args()` where `--run-id` is defined? Probably as a mutually exclusive argument with `--run-id` itself so you can't specify both at the same time.

We may only want this for the fetch command though, not `push` or `copy`, in which case yeah it would need to live here... maybe that's more reason to keep commit lookup outside of this script per my other comment.

If we keep this here (on just the fetch subcommand), we could raise `parser.error()` if `--run-id` and `--find-matching-commit` are both set at the same time.

---

Here's another idea to solve your problem: you could use https://github.com/ROCm/TheRock/blob/main/build_tools/fetch_artifacts.py instead, and _that_ could gain the `--find-matching-commit` argument, mutually exclusive with `--run-id` there: https://github.com/ROCm/TheRock/blob/a20ab791d74c94354704b791a85343babb83409b/build_tools/fetch_artifacts.py#L359-L369

The fetch_artifacts.py script has include/exclude regex patterns for artifacts while this script has `--stage`, so if you specifically want to only download bootstrapped artifacts for certain stages then this script is a better fit today.

build_tools/artifact_manager.py
mgehre-amd: I generally support the unix philosophy; here I'm looking for a single command to get my build dir populated (so my agent can get to work).
If you intend that there is another layer of scripts on top of this and user-facing, that would also work for me.

build_tools/artifact_manager.py
mgehre-amd: I'd like to keep this in artifact_manager.py because only it has the --bootstrap and --stage functionality.


I added the check for being mutally exclusive

build_tools/artifact_manager.py
```diff
@@ -106,6 +108,53 @@ def get_topology(topology_path: Optional[Path] = None) -> BuildTopology:
...
+def _resolve_run_id_for_commit(
+    args: argparse.Namespace, target_families: List[str]
+) -> str:
+    """Resolves --run-id by finding the CI workflow run for the current HEAD commit.
+
+    Reuses find_artifacts_for_commit(), the same lookup find_artifacts_for_commit.py's
+    own CLI uses, so a developer with a commit checked out doesn't need to run that
+    script separately to discover a run ID.
+    """
+    commit = _get_current_commit(get_default_topology_path().parent)
```
ScottTodd, line 132: `get_default_topology_path().parent` is a roundabout way to get the repository root and it assumes that the topology file is located at the root. If we ever move that file it would be difficult to notice that we should also update this code.

Let's use a code pattern like this to get the root dir instead: https://github.com/ROCm/TheRock/blob/3c7e76106232d5f91fda12e72dfa6cd1be691c6f/build_tools/bump_submodules.py#L34-L35

Also note that if we want to use this code from repositories like rocm-libraries and rocm-systems, there will be multiple git repositories checked out and workflow runs will be scoped to each repository, so this look up code may not be general enough.

build_tools/artifact_manager.py
```diff
@@ -106,6 +108,53 @@ def get_topology(topology_path: Optional[Path] = None) -> BuildTopology:
...
+    log(f"Looking up CI workflow run for commit {commit}...")
+    try:
+        results = find_artifacts_for_commit(
+            commit=commit,
+            artifact_groups=target_families,
+            platform=args.platform,
+        )
```
ScottTodd, line 139: Per my other comment, this won't be able to handle workflow runs in rocm-libraries/rocm-systems unless we also pass `--repo` and `--workflow` or detect when they should be used. See https://github.com/ROCm/TheRock/blob/ec7bc4e63f6ab8da5c6c0a14cbc7930d87fe35ff/build_tools/find_artifacts_for_commit.py#L213-L224

Details like that are why I'd rather we have specialized scripts responsible for details like that and not scope creep existing scripts to handle all possible cases.

ScottTodd (commented):
Still not sure about this, especially considering how we want to handle workflow run lookups across multiple repositories.

Issue #6266, opened 2026-06-30
Promotion script: Follow up tasks for multi-arch


HereThereBeDragons:
Promotion script in https://github.com/ROCm/TheRock/pull/5109 needs the following follow-up tasks:

- [ ] Rewrite the test suite to use multi arch and offer and e2e testing for live promoted packages (test currently runs on 7.9 single arch packages that are failing as they dont contain the therock_manifest.json)
- [x] Remove the `"rocm_sdk_libraries_None"` as #5984 fixed this.

- [x] double check that new jax wheels are promoted correctly
- [x] security issue: tighten extracting targz. most likely we want `targz.extractall(tmp_path, filter="tar")`  but should also try out if it works with `targz.extractall(tmp_path, filter="data")`  https://peps.python.org/pep-0706/#filters

marbre:
Assigning @JeniferC99. Please reach out if anything is unclear and if you need further information.

CC @amd-aakash, @rahulc-gh

JeniferC99:
https://github.com/ROCm/TheRock/pull/6409 - promotion script fix - merged
https://github.com/ROCm/TheRock/pull/6606/changes - e2e testing script - review process

PR #6399, opened 2026-07-07
Rewrite promote_packages test for multi-arch releases

JeniferC99:
## Motivation

The promotion test (`build_tools/packaging/tests/promote_packages_test.py`) still points at retired 7.9 single-arch prerelease packages and single-arch index paths, so it no longer runs end to end against what we actually ship. This is the first follow-up item in ROCm/TheRock#6266 (rewrite the promotion test suite for multi-arch + live end-to-end coverage). It stays an on-demand script (run by hand before a release), not per-PR CI, since it pulls large artifacts and builds venvs.

## Technical Details

Retargets the test at the live multi-arch prerelease index (`whl-multi-arch/`) and makes it self-updating instead of hardcoded:

- Discovers the latest `X.Y.ZrcN` from the index (overridable with `--version`).
- Dedups each package to a single upstream version (the index publishes e.g. `torch 2.10.0` and `2.14.0a0` for the same RC) and intersects the `cpXX` tag across `torch` and the per-gfx device wheels so the set is coherent/installable.
- Derives expected promoted filenames by stripping the RC suffix, replacing the brittle per-arch hardcoded lists.
- Adds a multi-arch keep-list scenario (`multi_arch_targets=[keep_arch]`) that asserts non-kept-arch device wheels (`amd_torch_device_gfx*`, `amd_torchvision_device_gfx*`, `rocm_sdk_device_gfx*`) are pruned while the RC suffix is still stripped everywhere.
- Tolerates torch-family local versions that carry a git hash (e.g. `+git43422b04.rocm7.14.0`).

Open questions carried into this draft (from #6266):

- Version baseline: keep auto-discovering the latest RC (current default) or pin a known RC per release branch for deterministic runs?
- Should the run assert a fixed arch set, or accept whatever the index currently publishes for the chosen gfx targets?
- Include the `therock-dist-*` multi-arch tarball once `tarball-multi-arch/` exposes a stable, enumerable listing (it is JS-rendered today).

Still on-demand/manual by design — confirming that's the intended long-term shape rather than wiring into CI. Selection/discovery logic validated live against `7.14.0rc1`; the full download + promote + venv-install run is pending.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/tests/promote_packages_test.py
+++ b/build_tools/packaging/tests/promote_packages_test.py
@@ -2,63 +2,77 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""Test suite for the promote_packages package promotion script.
+"""Manual / on-demand test suite for the promote_packages promotion script.
 
-This test suite validates the functionality of promote_packages.py, which
-promotes release candidate (RC) packages to final releases by removing RC suffixes
-from version strings (e.g., 7.9.0rc1 → 7.9.0).
+This exercises promote_packages.py end-to-end against **live multi-arch**
+release-candidate (RC) artifacts published to
+https://rocm.prereleases.amd.com/. Promotion strips the RC suffix from version
+strings (e.g. 7.14.0rc1 -> 7.14.0) and, for multi-arch indices, prunes the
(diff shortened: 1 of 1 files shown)


JeniferC99:

PR #6402, opened 2026-07-07
Add JAX wheel promotion coverage to promote_packages test

JeniferC99:
## Motivation

Second follow-up from ROCm/TheRock#6266: double-check that JAX wheel promotion is correct and give it explicit test coverage. This is stacked on #6399 (the multi-arch test rewrite) and complements #6158, which removed `jaxlib` from the multi-arch promote list because `jaxlib` is not built multi-arch — only `jax_rocm7_plugin` and `jax_rocm7_pjrt` are.

## Technical Details

Adds a `checkPromoteJax` scenario to `promote_packages_test.py`:

- Covers `jax_rocm7_plugin` and `jax_rocm7_pjrt`, which promote via the local-version path (the RC suffix is stripped from the `+rocm<ver>` segment; the promote script early-returns from the internal file rewrites for these packages).
- Handles JAX-specific packaging the other wheels don't have: `manylinux_*` platform tags (not `linux_x86_64`) and the two upstream versions the index publishes per RC on different manylinux baselines (e.g. `0.10.0` on manylinux_2_27, `0.9.1` on manylinux_2_28) — dedups to the newest upstream.
- Installs `--no-deps` (jax/jaxlib live on PyPI and aren't part of the promoted set), so it validates the promoted wheels are structurally installable without pulling the world.
- Docstring notes that `jaxlib` is intentionally not promoted multi-arch (#6158).

Selection/discovery validated live against `7.14.0rc1` (picks plugin+pjrt at the newest upstream, correct cp tag, RC suffix stripped). Full download + promote + install run still pending, same as #6399. Stacked on `users/jechrist/promote-test-multi-arch`; kept draft.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/tests/promote_packages_test.py
+++ b/build_tools/packaging/tests/promote_packages_test.py
@@ -23,9 +23,13 @@ WHAT IS VALIDATED
      per-gfx device wheels (amd_torch_device_gfx*, amd_torchvision_device_gfx*,
      rocm_sdk_device_gfx*) and deletes the rest, while still stripping the RC
      suffix everywhere.
-  3. Partial promotions (only ROCm, or only torch) do NOT produce a coherent,
+  3. JAX wheels (jax_rocm7_plugin, jax_rocm7_pjrt) promote correctly: the RC
+     suffix is stripped from the `+rocm<ver>` local segment and the wheels stay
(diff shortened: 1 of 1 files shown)


JeniferC99:

Issue #6065, opened 2026-06-23
[Feature] Expand component test coverage to more OS container images


ScottTodd:
For Linux, we currently run "component tests" (rocblas, rocfft, hipblaslt, miopen, etc.) only on Ubuntu. We should expand to other distros like RHEL 8.10 to ensure that there aren't any test failures unique to these other distros.

## Step 1: choose which containers to cover, create test images

We can create a prioritized list from the supported distros list: https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/system-requirements.html#supported-distributions . As of 2026-06-23 that list is:

Operating System | Kernel | Glibc
-- | -- | --
Ubuntu 24.04.4 | 6.8 [GA], 6.17 [HWE] | 2.39
Ubuntu 22.04.5 | 5.15 [GA], 6.8 [HWE] | 2.35
RHEL 10.1 | 6.12.0-124 | 2.39
RHEL 10.0 | 6.12.0-55 | 2.39
RHEL 9.7 | 5.14.0-611 | 2.34
RHEL 9.6 | 5.14.0-570 | 2.34
RHEL 9.4 | 5.14.0-427 | 2.34
RHEL 8.10 | 4.18.0-553 | 2.28
SLES 15 SP7 | 6.4.0-150700.51 | 2.38
Debian 13 | 6.12 | 2.35
Debian 12 | 6.1.0 | 2.36
Rocky Linux 9 | 5.14.0-570 | 2.34
Oracle Linux 10 | 6.12.0 (UEK) | 2.39
Oracle Linux 9 | 5.15.0 (UEK) | 2.34
Oracle Linux 8 | 5.15.0 (UEK) | 2.28

In many cases, building on manylinux should hopefully be sufficient for distributing safely across these distros. If/when there are test failures we should root cause why that did not hold.

For the images we maintain and use here in TheRock, see the docs: https://github.com/ROCm/TheRock/tree/main/dockerfiles#no_rocm_image_dockerfile (note: deps needed for testing and _no preinstalled rocm_).

Code references:

* https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubi10.Dockerfile
  * https://github.com/ROCm/TheRock/blob/main/.github/workflows/publish_no_rocm_image_ubi10.yml
* https://github.com/ROCm/TheRock/blob/main/dockerfiles/no_rocm_image_ubuntu24_04.Dockerfile
[...]

PR #6405, opened 2026-07-07
feat(ci): add RHEL 8 native package install test

arvindcheru:
## Motivation

Add RHEL 8 (`rhel8`) to the native Linux package **install smoke test** pipeline, alongside existing profiles such as `ubuntu2404`, `rhel10`, and `sles16`. This validates that ROCm RPMs can be installed from CI-built repositories on an EL8 userspace (UBI 8.10).

## Technical Details

Key changes include:

**In scope — install smoke test (this PR):**
- Repo setup, `dnf install` of ROCm metapackages, and basic install verification inside a UBI 8 container
- Same pattern as the existing Docker-based native install jobs for other OS profiles

**Out of scope — functional / kernel-level validation:**
- RHEL 8.10 / kernel 4.18 kpack issues cannot be exercised in Docker because containers use the **host kernel**
- That class of testing belongs on the complementary KVM/QEMU path ([#6318](https://github.com/ROCm/TheRock/pull/6318), automation in [#6413](https://github.com/ROCm/TheRock/pull/6413)), where a full VM (e.g. Rocky Linux 8.10) runs the target kernel and a GPU can be exposed via PCI passthrough

**CI workflow and OS profile support**
- Add `rhel8` to `multi_arch_ci_linux.yml` (new `test_native_packages_rhel8` job)
- Add `rhel8` to the install-test matrix in `multi_arch_release_linux.yml`
- Add `rhel8` to `test_native_linux_packages_install.yml` (workflow_call + workflow_dispatch)

**Package install test script**
- Add `rhel8`-specific handling in `native_linux_package_install_test.py` for EL8 install behavior
- Extend unit tests in `native_linux_package_install_ut_test.py` for rhel8 paths
- Update CLI examples/docs for rhel8 RPM install invocations

ISSUE ID: #6065

## Test Plan

- [ ] `workflow_dispatch` of `test_native_linux_packages_install.yml` with `os_profile=ubuntu2404` (regression — no change in behavior)
- [ ] `workflow_dispatch` of `test_native_linux_packages_install.yml` with `os_profile=rhel8`
- [ ] Multi-arch CI path: `test_native_packages_rhel8` job passes on a PR CI run
- [ ] Local/unit tests:
  - `bash build_tools/packaging/linux/tests/setup_python_cmd_test.sh`
  - `python -m pytest build_tools/packaging/linux/tests/get_url_repo_params_test.py -k rhel8`
  - `python -m pytest build_tools/packaging/linux/tests/native_linux_package_install_ut_test.py -k rhel8`

## Test Result

1. rhel8 sanity - https://github.com/ROCm/TheRock/actions/runs/28956659590
2. 

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -200,6 +200,23 @@ jobs:
       contents: read
       id-token: write
 
+  test_native_packages_rhel8:
+    needs: [build_native_rpm_packages]
+    name: Test RPM Install - rhel8
+    if: ${{ !failure() && !cancelled() && fromJSON(inputs.build_config).build_native_linux == true }}
+    uses: ./.github/workflows/test_native_linux_packages_install.yml
+    with:
+      package_install_url: ${{ needs.build_native_rpm_packages.outputs.package_repository_url }}
+      release_type: ci
+      os_profile: rhel8
+      artifact_group: ${{ fromJSON(inputs.build_config).dist_amdgpu_families }}
+      test_type: sanity
+      repository: ${{ inputs.repository }}
+      ref: ${{ inputs.ref }}
+    permissions:
+      contents: read
+      id-token: write
+
   test_native_packages_rhel10:
     needs: [build_native_rpm_packages]
     name: Test RPM Install - rhel10
--- a/build_tools/packaging/linux/native_linux_package_install_test.py
+++ b/build_tools/packaging/linux/native_linux_package_install_test.py
@@ -184,7 +184,6 @@ APT_UPDATE_TIMEOUT_SEC = 120
 ZYPP_CLEAN_TIMEOUT_SEC = 60
 ZYPP_REFRESH_TIMEOUT_SEC = 120
 DNF_CLEAN_TIMEOUT_SEC = 60
-DNF_MAKECACHE_TIMEOUT_SEC = 120
 INSTALL_TIMEOUT_SEC = 1800  # 30 minutes
 ROCMINFO_TIMEOUT_SEC = 30
 RDHC_TIMEOUT_SEC = 30
@@ -293,11 +292,6 @@ def _run_streaming(cmd: list[str], timeout_sec: int) -> int:
         raise
 
 
-def _is_rhel8_profile(os_profile: str) -> bool:
-    """True only for the ``rhel8`` OS profile (not other EL8/RHEL-like names)."""
-    return os_profile.lower() == "rhel8"
-
-
 class NativeLinuxPackageInstallTest:
     """Runner for the native Linux package install test (repo setup, install, verification)."""
 
@@ -358,10 +352,6 @@ class NativeLinuxPackageInstallTest:
         True if SLES, False otherwise
         """
         return self.os_profile.lower().startswith("sles")
-
-    def _is_rhel8(self) -> bool:
-        """Check if the OS profile is ``rhel8``."""
-        return _is_rhel8_profile(self.os_profile)
 
     def __init__(
         self,
@@ -717,25 +707,8 @@ gpgcheck=0
         except subprocess.TimeoutExpired:
             print("[WARN] dnf clean timed out (may not be critical)")
 
-        # Refresh metadata (required on EL8 before first ``dnf install`` from a new .repo)
-        print(f"\nRefreshing repository metadata (dnf makecache --repo={repo_name})...")
-        try:
-            makecache_cmd = ["dnf", "makecache", f"--repo={repo_name}"]
-            return_code = _run_streaming(makecache_cmd, DNF_MAKECACHE_TIMEOUT_SEC)
-            if return_code == 0:
-                print("\n[PASS] DNF repository metadata refreshed")
-                return True
-            print(
-                f"\n[FAIL] dnf makecache failed for repo '{repo_name}' "
-                f"(exit code: {return_code})"
-            )
-            return False
-        except subprocess.TimeoutExpired:
-            print("\n[FAIL] dnf makecache timed out")
-            return False
-        except OSError as e:
-            print(f"\n[FAIL] Error refreshing DNF repository metadata: {e}")
-            return False
+        print("\n[PASS] DNF repository setup complete")
+        return True
 
     def setup_rpm_repository(self) -> bool:
         """Setup RPM repository on the system.
@@ -832,11 +805,7 @@ gpgcheck=0
                 ] + self.package_names
             print("[INFO] Using zypper for SLES package installation")
         else:
-            dnf_args = ["dnf", "install", "-y"]
-            if self._is_rhel8():
-                # rhel8 only: EL8 repos occasionally need --nobest for ROCm deps.
-                dnf_args.append("--nobest")
-            cmd = dnf_args + self.package_names
+            cmd = ["dnf", "install", "-y"] + self.package_names
         print(f"\nRunning: {' '.join(cmd)}")
         print("=" * 80)
         print("Installation progress (streaming output):\n")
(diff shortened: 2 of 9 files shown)


.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -200,6 +200,23 @@ jobs:
       contents: read
       id-token: write
 
+  test_native_packages_rhel8:
```
nunnikri, line 203: Isnt it better to make this also matrix like in release workflow

lamikr:
This pr may be overlapping with the approach on https://github.com/ROCm/TheRock/pull/6318

It is not possible to test the rhel 8.10 kernel 4.18 kpack problem by running the test at least by using a docker based testing because docker uses the host machines kernel for running the tests. (I.e host machine should then need to have old kernel like 4.18 installed)

Alternative way that we try to build for this is to generate image that can be run with the kvm/qemu based virtualization and then use the pci-passthrough mechanism to enable the gpu from the host-machine. Image generation part for that
is implemented in PR https://github.com/ROCm/TheRock/pull/6318 but at the moment it will still needs to be hooked to the test machine. Note that we are using here the rockylinux 8.10 which is 1:1 binary compatible with the rhel 8.10 but allows image generation without license keys. Pipeline has been tested at the moment manually by launching the generated image with pci-passthrough parameters and then installing therock release, but now that part needs to be automatized.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -707,8 +717,25 @@ def _setup_dnf_repository(self) -> bool:
         except subprocess.TimeoutExpired:
             print("[WARN] dnf clean timed out (may not be critical)")
 
-        print("\n[PASS] DNF repository setup complete")
-        return True
+        # Refresh metadata (required on EL8 before first ``dnf install`` from a new .repo)
+        print(f"\nRefreshing repository metadata (dnf makecache --repo={repo_name})...")
+        try:
+            makecache_cmd = ["dnf", "makecache", f"--repo={repo_name}"]
```
nunnikri: We shouldnt add this step. None of our documentation have this step mentioned. dnf install ideally should do this automatically internally.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -805,7 +832,11 @@ def install_rpm_packages(self) -> bool:
                 ] + self.package_names
             print("[INFO] Using zypper for SLES package installation")
         else:
-            cmd = ["dnf", "install", "-y"] + self.package_names
+            dnf_args = ["dnf", "install", "-y"]
+            if self._is_rhel8():
+                # rhel8 only: EL8 repos occasionally need --nobest for ROCm deps.
+                dnf_args.append("--nobest")
```
nunnikri: Why are we using --nobest option? We should always install the latest version.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -1057,6 +1088,11 @@ def test_rdhc(self) -> bool:
  --repo-url https://rocm.nightlies.amd.com/rpm/20260204-21658678136/rhel8/x86_64/ \\
  --gfx-arch gfx94x --release-type nightly --install-prefix /opt/rocm/core
 
+ # --test-type full on RHEL 8 (rdhc needs pciutils/kmod on the host — install before running)
```
nunnikri, line 1060: do we have to add pciutils/kmod in the pre requisite step of container setup?

arvindcheru: > do we have to add pciutils/kmod in the pre requisite step of container setup?
For Full Mode only, we are adding to prerequisite

build_tools/packaging/linux/native_linux_package_install_test.py
arvindcheru: Thanks for the comments, it was added for RHEL8 only, will check and update this.

.github/workflows/multi_arch_ci_linux.yml
arvindcheru: Thanks for the comment, Will check and update this

arvindcheru:
> This pr may be overlapping with the approach on #6318
> 
> It is not possible to test the rhel 8.10 kernel 4.18 kpack problem by running the test at least by using a docker based testing because docker uses the host machines kernel for running the tests. (I.e host machine should then need to have old kernel like 4.18 installed)
> 
> Alternative way that we try to build for this is to generate image that can be run with the kvm/qemu based virtualization and then use the pci-passthrough mechanism to enable the gpu from the host-machine. Image generation part for that is implemented in PR #6318 but at the moment it will still needs to be hooked to the test machine. Note that we are using here the rockylinux 8.10 which is 1:1 binary compatible with the rhel 8.10 but allows image generation without license keys. Pipeline has been tested at the moment manually by launching the generated image with pci-passthrough parameters and then installing therock release, but now that part needs to be automatized.

Thanks for the context on [#6318](https://github.com/ROCm/TheRock/pull/6318) — I agree these address different problems and are complementary rather than overlapping.

This PR (#6405) adds RHEL 8 to the existing native package install CI path: repo setup, dnf install of ROCm metapackages, and basic install verification inside a UBI 8 container (same pattern as ubuntu2404 / rhel10). It is intentionally scoped as an install smoke test, not a functional or kernel-level validation.

As you noted, Docker-based testing cannot exercise the RHEL 8.10 / kernel 4.18 kpack issue because containers use the host kernel. That class of problem belongs on the KVM/QEMU + PCI passthrough track ([#6318](https://github.com/ROCm/TheRock/pull/6318), and automation in [#6413](https://github.com/ROCm/TheRock/pull/6413)), where a full VM runs the target kernel and a real GPU can be exposed.

Thank you again for bringing this up, I’ll also clarify the scope in the PR description.

build_tools/packaging/linux/native_linux_package_install_test.py
arvindcheru: Thanks for the comments, this is addressed in the latest version changes

build_tools/packaging/linux/native_linux_package_install_test.py
arvindcheru: Usage of nobest removed, is addressed in the latest version changes

.github/workflows/multi_arch_ci_linux.yml
arvindcheru: Request to take/move the matrix like usage changes along with next upcoming changes/PR.  Thanks

nunnikri (approved):

nunnikri:
> This pr may be overlapping with the approach on #6318
> 
> It is not possible to test the rhel 8.10 kernel 4.18 kpack problem by running the test at least by using a docker based testing because docker uses the host machines kernel for running the tests. (I.e host machine should then need to have old kernel like 4.18 installed)
> 
> Alternative way that we try to build for this is to generate image that can be run with the kvm/qemu based virtualization and then use the pci-passthrough mechanism to enable the gpu from the host-machine. Image generation part for that is implemented in PR #6318 but at the moment it will still needs to be hooked to the test machine. Note that we are using here the rockylinux 8.10 which is 1:1 binary compatible with the rhel 8.10 but allows image generation without license keys. Pipeline has been tested at the moment manually by launching the generated image with pci-passthrough parameters and then installing therock release, but now that part needs to be automatized.

@lamikr your changes is critical for validation of ROCm features in RHEL8 and old kernel env. This PR is only focusing on verifying  the rpm packages created are installable or not. Not really any functional verification. I believe both are complementary. If you feel this a duplicate effort, lets discuss offline and confirm

lamikr:
Thanks for the explanation, I think both of these are are then good as they serve different purposes. As the qemu/kvm based test will require a dedicated gpu with pci-passthrough configured., the pool of those machines will probably be smaller and so tests runs there needs also run faster just to verify known issues we have detected. 

For example for now we plan to only run there couple of test-cases from rocfft-tests to verify the fix for issues like
https://github.com/ROCm/rocm-systems/pull/7681. For that it's enough to run only a couple of test-cases from "rocfft-test" with " --gtest_filter=bitwise" parameter instead of running a full set of tests.

PR #6407, opened 2026-07-07
fix(ci): Handle null family_overrides in external repo config

geomin12:
When external repos call setup_multi_arch.yml without specifying family_overrides in their external_repo JSON, the workflow expression `toJSON(fromJSON(inputs.external_repo).family_overrides)` produces the string "null". This caused json.loads("null") to return None, and then None.items() raised an AttributeError.

Add a check for null/empty external_overrides before iterating.

Working as expected: https://github.com/ROCm/rocm-libraries/actions/runs/28896388653/job/85722082854


tony-davis (approved):
Looks good.

PR #6409, opened 2026-07-07
refactor(packaging): harden the package promotion script

JeniferC99:
## Motivation

Follow-ups from #6266 for the promotion script itself: multi-arch keep-lists raised on Linux-only archs whose `requires.txt` device sections carry an environment marker, the `rocm_sdk_libraries_None` workaround is dead code since #5984 (merged 2026-06-22), and the sdist extraction was unhardened. The `promote_packages_test.py` rewrite that originally rode along in this PR has been split into #6606 so this one carries only the `promote_packages.py` changes.

## Technical Details

- `_apply_keep_list_to_requires_txt` strips a section's environment marker (e.g. `[device-gfx942:sys_platform == "linux"]`) before extracting the arch, so keep-lists for Linux-only archs (gfx942/gfx950/gfx1250) find the real dep body under the marked section instead of raising `ValueError`.
- Remove the dead `rocm_sdk_libraries_None` package-name override (the multi-arch `rocm_sdk_libraries` wheel now ships `rocm_sdk_libraries/_dist_info.py`, so the derived name resolves directly).
- Harden the sdist extraction with PEP 706 `filter="data"` (refuses absolute-path / `..` members; requires Python 3.12+, per the repo baseline).
- Refresh the module docstring: document the `therock-dist-*.tar.gz` rename path, fix the stale `--delete` flag name, and point TESTING at the `--input-dir` invocation.

## Submission Checklist


araravik-psd (approved):
LGTM, the hardening changes also look good. I have tested the test package on promote_package.py, I see it is getting promoted properly

`(.venv) arravikum@aravind-azure-testVM:~/TheRock_promote_packages_test/TheRock$ python ./build_tools/packaging/promote_packages.py   --input-dir=./build_tools/packaging/promotion/download/tarball   --src-version-type=rc   --delete-old-on-success
Parsing arguments ...done
Looking for .whl and .tar.gz in build_tools/packaging/promotion/download/tarball/*

Promoting tarball from rc to final: therock-dist-windows-multiarch-tests-7.14.0rc3.tar.gz
  Detected version: 7.14.0rc3
  New version: 7.14.0
  Rename therock-dist-windows-multiarch-tests-7.14.0rc3.tar.gz to therock-dist-windows-multiarch-tests-7.14.0.tar.gzRepacked therock-dist-windows-multiarch-tests-7.14.0rc3.tar.gz as release build_tools/packaging/promotion/download/tarball/therock-dist-windows-multiarch-tests-7.14.0.tar.gz`

build_tools/packaging/promote_packages.py
```diff
@@ -965,7 +974,9 @@ def promote_targz_sdist(
         tmp_path = pathlib.Path(tmp_dir)
 
         targz = tarfile.open(filename)
-        targz.extractall(tmp_path)
+        # PEP 706: refuse members with absolute paths / `..` traversal and strip
+        # unsafe metadata when extracting the downloaded sdist.
+        targz.extractall(tmp_path, filter="data")
```
HereThereBeDragons, line 980: did you double check that file ownership did not change? uid/gid?
because that would be one of the obvious differences between "tar" and "data"

build_tools/packaging/promote_packages.py
```diff
@@ -552,8 +558,14 @@ def _apply_keep_list_to_requires_txt(path: pathlib.Path, keep_archs: list[str])
             section_match = section_re.match(stripped)
             if section_match:
                 section_name = section_match.group(1)
-                if section_name.startswith("device-") and section_name != "device-all":
-                    arch_token = section_name[len("device-") :]
+                # A section can carry an environment marker, e.g.
+                # `[device-gfx942:sys_platform == "linux"]` for a linux-only arch
+                # (its plain `[device-gfx942]` section is empty and the real dep
```
HereThereBeDragons, line 563: thanks for finding this.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+download_python_packages.py, or copied by hand) and it will run every
+promotion scenario over that set.
+
+Promotion strips the RC suffix from version strings (e.g. 7.14.0rc1 -> 7.14.0)
+and, for multi-arch inputs, prunes the per-gfx device wheels down to a subset of
+architectures.
+
+It is intentionally a standalone, on-demand script (run by a human before a
+release), NOT a pytest/CI target: it installs the wheels into throwaway
+virtualenvs, which is too heavy for per-PR CI. Keep it runnable by hand.
```
HereThereBeDragons: `which is too heavy for per-PR CI.`
this is not really needed here

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+"""Manual / on-demand test suite for the promote_packages promotion script.
+
+This exercises promote_packages.py end-to-end against a **local directory of
+already-downloaded** release-candidate (RC) artifacts. It does NOT fetch
+anything over the network: point it at a directory that already contains the
+wheels / sdist / therock-dist tarballs for one RC (e.g. produced by
+download_python_packages.py, or copied by hand) and it will run every
+promotion scenario over that set.
+
+Promotion strips the RC suffix from version strings (e.g. 7.14.0rc1 -> 7.14.0)
```
HereThereBeDragons: promotion does not only strip RC suffix.
the script can do
rc -> a
a -> rc
a -> release
rc -> release
and if you just strip the archs in multiarch than it does not even change the version of the package

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+     its filename rc->final rename (promote_targz_tarball never opens the
+     archive, so this is exercised with cheap stand-ins).
+  6. Promoted filenames drop the RC suffix and every wheel reports the final
+     version.
+
+The RC version, the target platform, and the set of gfx architectures are all
+DERIVED from the files in the input directory (the RC token is read from the
+rocm sdist / rocm_sdk_core wheel; the final version is that with the prerelease
+segment stripped). Expected promoted filenames are derived by applying the same
+rc->final rename to whatever was found, so there are no brittle hardcoded lists.
```
HereThereBeDragons: Expected promoted filenames are derived by applying the same
rc->final rename to whatever was found, so there are no brittle hardcoded lists. 

what does this mean?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+JAX_WHEEL_PREFIXES = ("jax_rocm7_plugin-", "jax_rocm7_pjrt-")
 
-    if len(files) != len(expected_promoted_pkgs):
-        return (
-            False,
-            f"Files found and expected promoted packages are not the same amount ({len(files)} vs {len(expected_promoted_pkgs)})",
+# Per-gfx device wheel marker, e.g. `rocm_sdk_device_gfx942-...` /
+# `amd_torch_device_gfx1201-...`. The trailing `-` anchors the arch token so
+# `gfx11` can't match `gfx1153`.
+_DEVICE_ARCH_RE = re.compile(r"device_(gfx[0-9a-z]+)-")
```
HereThereBeDragons: please reuse as much from promote_packages.py as possible

e.g. this regex already defined in promote_packages.py

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+
+# RC version token as it appears on the rocm sdist / rocm_sdk_core wheel.
```
HereThereBeDragons: why is this needed?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        return name.endswith(".whl") or name.endswith(".tar.gz")
+    return False
+
+
+def _device_arch(name: str) -> str | None:
+    m = _DEVICE_ARCH_RE.search(name)
+    return m.group(1) if m else None
+
+
+def _torch_triton_pin(wheel_path: Path) -> str | None:
```
HereThereBeDragons: this is also true for torch, torchaudio, torchvision. why this specificly?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+    torch's own metadata rather than assuming newest.
+    """
+    with zipfile.ZipFile(wheel_path) as z:
+        meta = next(n for n in z.namelist() if n.endswith(".dist-info/METADATA"))
+        text = z.read(meta).decode("utf-8", errors="replace")
+    m = re.search(r"^Requires-Dist:\s*triton==([^\s;]+)", text, re.MULTILINE)
+    return m.group(1) if m else None
+
+
+def detect_rc_version(names: set[str]) -> str:
```
HereThereBeDragons: the promote_packages.py will work on any wheel. you dont need a full set of rocm wheels. you can also just promote a single torch wheel.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+    """Read the RC version token from the rocm sdist or rocm_sdk_core wheel."""
+    for regex in (_SDIST_VER_RE, _CORE_VER_RE):
```
HereThereBeDragons: why not use `pkginfo.Wheel(wheelpath).version`? thats more reliable and then you can also check if the file name is correct

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+# `gfx11` can't match `gfx1153`.
+_DEVICE_ARCH_RE = re.compile(r"device_(gfx[0-9a-z]+)-")
+
+# RC version token as it appears on the rocm sdist / rocm_sdk_core wheel.
+_SDIST_VER_RE = re.compile(r"^rocm-(\d+\.\d+\.\d+(?:rc\d+|a\d+)?)\.tar\.gz$")
+_CORE_VER_RE = re.compile(r"^rocm_sdk_core-(\d+\.\d+\.\d+(?:rc\d+|a\d+)?)-")
+
+
+def _upstream_version(name: str) -> Version:
+    """Extract the upstream version field from a wheel/sdist filename.
```
HereThereBeDragons: upstream version is a bad naming. its more like the base package version.
also better to use `Version(Wheel(path).version).public`

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+            if m:
+                return m.group(1)
+    raise RuntimeError(
+        "Could not determine the RC version from the input directory: expected a "
+        "rocm-<ver>.tar.gz sdist or a rocm_sdk_core-<ver>-*.whl to be present."
+    )
+
+
+class InputSet:
+    """Classification of an input directory into promotion groups."""
```
HereThereBeDragons: for so much code too little description

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+
+    def __init__(self, input_dir: Path, platform: str | None) -> None:
+        names = {p.name for p in input_dir.iterdir() if p.is_file()}
+        if not names:
+            raise RuntimeError(f"No files found in input directory {input_dir}")
+
+        self.input_dir = input_dir
+        self.rc_str = detect_rc_version(names)
```
HereThereBeDragons: you could have multiple packages with different versions in the same directory

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        self.rc_version = Version(self.rc_str)
+        self.final_version = Version(self.rc_version.base_version)
+        self.final_str = str(self.final_version)
+        self.src_version_type = "rc" if "rc" in self.rc_str else "a"
```
HereThereBeDragons: no you cant assume this. how should this even work if you only regex for "rc"?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        self.platform = platform or (
+            "windows" if any("win_amd64" in n for n in names) else "linux"
```
HereThereBeDragons: you will ahve both linux and windows packages in a single directory. this does not work

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
         )
 
-    for file in files:
-        if not file in expected_promoted_pkgs:
-            return False, f"{file} not matching any of the expected package names"
+        self.device_files = {n for n in names if _device_arch(n)}
```
HereThereBeDragons: these are platform dependent

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        self.jax_files = {n for n in names if n.startswith(JAX_WHEEL_PREFIXES)}
```
HereThereBeDragons: just grep for jax. no need to match the wheel prefixes specifically

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        "Could not determine the RC version from the input directory: expected a "
+        "rocm-<ver>.tar.gz sdist or a rocm_sdk_core-<ver>-*.whl to be present."
+    )
+
+
+class InputSet:
+    """Classification of an input directory into promotion groups."""
+
+    def __init__(self, input_dir: Path, platform: str | None) -> None:
+        names = {p.name for p in input_dir.iterdir() if p.is_file()}
```
HereThereBeDragons: this does not recurse into subdirs but what you get from download_python_packages.py is:

```
<output-dir>/
  wheels/                       # ALL wheels flat here, no per-arch subdirs
    rocm_sdk_core-<version>-...whl
    rocm_sdk_devel-<version>-...whl
    rocm_sdk_libraries-<version>-...whl
    amd_torch_device-gfx942-...whl      # device pkgs carry gfx in filename
    amd_torch_device-gfx950-...whl
    torch-<ver>+rocm<version>-...whl
    torchvision-...whl
    triton-...whl
    rocm-..tar.gz
    ...
  tarball-multi-arch/           # only if --include-tarballs
    therock-dist-linux-<arch>-<version>.tar.gz
    therock-dist-windows-<arch>-<version>.tar.gz
    ...
```
so tarballs and .whl should not be mixed

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        self.platform = platform or (
+            "windows" if any("win_amd64" in n for n in names) else "linux"
         )
 
-    for file in files:
-        if not file in expected_promoted_pkgs:
-            return False, f"{file} not matching any of the expected package names"
+        self.device_files = {n for n in names if _device_arch(n)}
+        self.jax_files = {n for n in names if n.startswith(JAX_WHEEL_PREFIXES)}
+        self.tarball_files = {
```
HereThereBeDragons: there should be no mix of files  of wheels and tarballs.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+        self.present_arches = sorted(
+            {a for n in self.device_files if (a := _device_arch(n))}
+        )
+        # Auto-detect: keep the first present arch, everything else is pruned.
+        self.keep_arch = self.present_arches[0] if self.present_arches else None
+        self.dropped_device_files = {
+            n for n in self.device_files if _device_arch(n) != self.keep_arch
+        }
+
+    def _non_pinned_triton(self) -> set[str]:
```
HereThereBeDragons: why only look for triton? and why only check the highest version? could check all the torch versions.
but you could argue that the installation into venv should also find this issue

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+
+
+def checkPromoteTarball(
+    dir_path: Path,
+    tarball_names: set[str],
+    version: str,
+    final_version: str,
+    src_version_type: str = "rc",
+) -> bool:
+    """Every therock-dist tarball present (per-gfx AND multiarch) is promoted by
```
HereThereBeDragons: i thought the packages were already downloaded?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
 
 
-def checkPromoteOnlyTorch(
-    dir_path: Path, expected_version: Version, platform: str
-) -> bool:  # should fail
-    print("")
-    print(
-        "================================================================================="
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
HereThereBeDragons: this is too long. please move into a separate parsing function

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    success = False
-    with tempfile.TemporaryDirectory(
-        prefix="PromoteRcToFinalTest-PromoteOnlyTorch-"
-    ) as tmp:
-        tmp_dir = Path(tmp)
-        # make a copy in a separate dir to not pollute it
-        for file in dir_path.glob("*"):
-            shutil.copy2(file, tmp_dir)
+    parser.add_argument(
+        "--platform",
```
HereThereBeDragons: packages will be mixed in the directory

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    print("")
     print(
-        "================================================================================="
+        f"Testing promotion {inputs.rc_version} -> {inputs.final_version} on "
+        f"{inputs.platform} from {inputs.input_dir}"
     )
     print(
-        "TEST DONE: Testing promotion of only PyTorch packages. Result:"
-        + (" SUCCESS" if success else " FAILURE")
+        f"  arches present: {inputs.present_arches or '(none)'}; kept: {inputs.keep_arch}"
```
HereThereBeDragons: arches -> archs

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        "torchvision-0.24.0+rocm7.9.0rc1-cp312-cp312-win_amd64.whl",
-    ]
-
-    url_and_packages = [(URL, package) for package in packages]
-    url_and_packages.append(
-        (
-            "https://rocm.prereleases.amd.com/tarball/",
-            "therock-dist-windows-gfx1151-7.9.0rc1.tar.gz",
-        )
+    multi_arch_input = inputs.aggregator_files | inputs.device_files
```
HereThereBeDragons: probably need a comment that it only is designed for multi arch packages?

HereThereBeDragons (commented):
this is just a partial review. just so you can already work a bit on it.
some comments might be wrong as i didnt get the full picture yet. sorry in advance for that

build_tools/packaging/promote_packages.py
JeniferC99: The file owner (uid/gid) does not change. Inside the source tarball the files say root, but that never survives promotion either way: we extract as a normal user (which can't set files to root) and then re-pack, which stamps everything as the current user. That's true with both the old and new filter, so ownership is identical.

The only thing data changes is that it drops the setgid bit on folders (2755 → 775). But it makes no difference for a Python package (pip doesn't care about that bit).

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +309,354 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+            print(f"\n[ERROR] {func_name} failed:\n{res[1]}")
+            return False
+        if not expect_success and res[0]:
+            print(
+                f"\n[ERROR] {func_name} succeeded but the promotion should NOT have been coherent."
+            )
+            proc = subprocess.run(
+                ["ls", tmp_dir], capture_output=True, encoding="utf-8"
+            )
+            print(proc.stdout)
```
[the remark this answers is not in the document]
JeniferC99: Addressed. Replaced the `ls` shell-out with a cross-platform `Path.glob` dump, so the debug listing works on Windows and no longer risks a `FileNotFoundError` masking the real failure.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,249 @@
...
+
+    def _non_pinned_triton(self) -> set[str]:
+        torch_wheels = [n for n in self.aggregator_files if n.startswith("torch-")]
+        triton_wheels = {n for n in self.aggregator_files if n.startswith("triton-")}
+        if not torch_wheels or not triton_wheels:
+            return set()
+        torch = max(torch_wheels, key=_upstream_version)
+        pin = _torch_triton_pin(self.input_dir / torch)
+        if not pin:
+            return set()
```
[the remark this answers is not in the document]
JeniferC99: Addressed. It now raises a `RuntimeError` when torch wheels are present but none carry an exact `triton==` pin (and when a pin has no matching triton wheel), instead of silently returning an empty set.

build_tools/packaging/promote_packages.py
```diff
@@ -965,7 +974,9 @@ def promote_targz_sdist(
         tmp_path = pathlib.Path(tmp_dir)
 
         targz = tarfile.open(filename)
-        targz.extractall(tmp_path)
+        # PEP 706: refuse members with absolute paths / `..` traversal and strip
+        # unsafe metadata when extracting the downloaded sdist.
+        targz.extractall(tmp_path, filter="data")
         targz.close()
```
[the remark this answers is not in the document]
JeniferC99: Added a comment noting `extractall(filter=...)` requires Python 3.12+, matching the repo's Python baseline.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Removed — the docstring no longer includes that rationale.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Fixed the docstring: promotion rewrites the prerelease segment generally (strip for rc/a → release, or replace for a → rc / rc → a), and a multi-arch run can prune only the per-gfx device wheels with the version unchanged. The scenarios here drive the common rc/a → release path.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: It means the test never hardcodes expected output names — it reads the source version from the package (`rocm_sdk_core`/`rocm` sdist), drops the prerelease to get the final version, and for each filename swaps source→final (the same rename promotion does), e.g. `rocm_sdk_core-7.9.0rc20260501-…whl` → `rocm_sdk_core-7.9.0-…whl`. So expectations adapt to whatever's in the dir — no stale lists.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Done — `_DEVICE_ARCH_RE` now builds on `promote_packages._GFX_ARCH` instead of redefining the gfx token, so they can't drift. Also dropped the filename version regexes in favor of pkginfo reads.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Not needed anymore — removed. Those regexes parsed the RC version from the filename; the version now comes from package metadata via pkginfo.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: torch/torchvision/torchaudio multi-versioning is handled generically by `_newest_per_package`. triton is special-cased only because torch pins it exactly (`triton==<build>`) and a download can carry several triton builds, so 'newest' isn't necessarily the one torch wants — we read torch's metadata and pick that build.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Agreed — the test relies on that. Promotion is exercised over whatever subset is present; only the install validation narrows to one coherent stack because a venv can't hold mutually-exclusive versions. Promoting a single torch wheel is effectively what the partial-promotion scenarios do.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Done. We now read the version from the package's own metadata via pkginfo (`Wheel(path).version` / `SDist(path).version`) instead of the filename. As you noted, that also lets us confirm the filename matches the metadata.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Removed the `_upstream_version` helper and the bad name. Versions now come from pkginfo metadata (`Version(Wheel(path).version)`) rather than the filename. I kept the full version (incl. the `+rocm…` local segment) for the newest-per-package comparison since the local part only acts as a tiebreaker; happy to switch to `.public` if you'd prefer.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Expanded the descriptions — the module docstring now covers each scenario, the input layout, and the coherent-install-set rationale, the `InputSet` docstring enumerates every derived group, and the non-trivial helpers each have a docstring.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Handled. A download does ship multiple framework versions; promotion runs over all of them. The only place a single coherent set is needed is the install validation, where `_newest_per_package` keeps the newest of each distribution and triton is pinned to the chosen torch.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: No longer assumed. The source version is read from package metadata (pkginfo), and the rc-vs-alpha type comes from the parsed PEP 440 prerelease (`Version.pre`), so an alpha like `7.13.0a…` classifies as `a` — no "rc" substring assumption.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Fixed. `wheels/` does carry both; they can't co-install. The test picks one platform per run (default linux, or `--platform windows`), keeping that platform's wheels plus platform-agnostic files, before classifying or installing. The other platform is covered by re-running with `--platform windows`.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Right — device wheels are platform-dependent. `names` is filtered to a single platform before any classification, so `device_files` and the other groups only ever hold one platform's wheels.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Done — now just `"jax" in n`, no prefix or extension matching.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Fixed. Reads `<root>/wheels/` (flat `iterdir`, no per-arch subdirs to recurse) and `<root>/tarball-multi-arch/` separately, so `.whl` and `.tar.gz` are never mixed.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Wheels + the rocm sdist are read from `<root>/wheels/` and tarballs from `<root>/tarball-multi-arch/`, each classified from its own subdir — matching `download_python_packages.py`'s layout.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Only triton because it's the one exact `==` cross-pin — a download can carry several triton builds and torch names one. Not just the highest: `_non_pinned_triton` consults every torch wheel present. And yes, the venv install would also surface an incoherent triton — this pre-filter just avoids staging a knowingly-incoherent set.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: You're right. The placeholder tarballs were something I added during testing to avoid copying the real multi-GB archives, but that's no longer needed — the scenario now stages the actual tarballs from `tarball-multi-arch/` and promotes them directly. Removed the placeholder helper.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Done — pulled the argument parsing out into a separate `parse_args` function.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: The input is filtered to one platform before staging, so linux/windows wheels are never mixed in a venv, and wheels vs tarballs are read from separate subdirs.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Done — renamed throughout.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: Added — `checkPromoteMultiArch` now documents it's only meaningful for multi-arch packages and needs ≥2 gfx archs with device wheels (single-arch inputs skip it).

JeniferC99:
moved e2e test script to https://github.com/ROCm/TheRock/pull/6606

PR #6424, opened 2026-07-08
fix(build-tools): honor per-output remote key in fetch_dvc_artifacts

AnaghaRaoAMD:
## Summary

`build_tools/fetch_dvc_artifacts.py` (TheRock's in-tree `dvc pull` replacement) read **only** the `[core]` default remote and ignored each pointer output's `remote:` key. Blobs routed to a non-default DVC remote were fetched from the *default* remote's prefix and 404'd, breaking the `math-libs` "Fetch sources" step on every Linux and Windows GPU target.

This teaches the fetcher to honor DVC's per-output remote routing:

- Parse the optional per-output `remote:` key from each `outs:` entry.
- Parse **all** `['remote "<name>"']` sections in `.dvc/config`, not just the `[core]` default.
- Resolve each output to its named remote (falling back to the default) and fetch its blob from that remote's bucket/prefix. `.dir` manifest children inherit the parent output's remote.
- Build one S3 client per configured remote (each has its own `allow_anonymous_login` flag), shared read-only across worker threads.

Backward compatible: pointers with no `remote:` key and single-remote configs behave exactly as before.

**Fixes:** [ROCm/rocm-libraries#9183](https://github.com/ROCm/rocm-libraries/issues/9183)
**Root cause:** [ROCm/rocm-libraries#9157](https://github.com/ROCm/rocm-libraries/pull/9157) added a dedicated `golden-data` DVC remote + per-output `remote:` keys; it was reverted by [#9182](https://github.com/ROCm/rocm-libraries/pull/9182) to unblock CI and will re-land once this merges.

## Risk Assessment

Low. Scoped to a single build-tool script and its unit tests. No behavior change for existing pointer files (no `remote:` key) or single-remote configs; the default-remote path is preserved via a thin wrapper. New code activates only when a pointer carries a `remote:` key naming a non-default remote.

## Testing Summary

- Unit tests: 49/49 pass (40 pre-existing + 9 new covering `remote:` parsing, multi-remote config parsing/resolution, and end-to-end per-output routing including the exact dual-remote #9183 scenario).
- Real-S3 end-to-end (anonymous, as CI) using the actual `golden-data` pointers from rocm-libraries #9157, including the object named in the issue (`e1aab7b90300368b26ea8442792606e0`):
  - **Before (main):** `2 file(s) failed to fetch` — `404 HeadObject` on both golden-data pointers (reproduces #9183).
  - **After (this PR):** `fetched=10, exit 0` — all `.bin` files materialized and md5-verified.

## Testing Checklist

- [x] Unit tests — `python -m unittest fetch_dvc_artifacts_test` (from `build_tools/tests/`) — 49/49 passed
- [x] Real-S3 repro: original fetcher 404s, patched fetcher pulls golden-data blobs (anonymous)
- [x] Patch applies cleanly to current `main`; only the two fetcher files changed
- [ ] PR CI — GitHub PR checks — Pending

## Technical Changes

- `_Out` gains an optional `remote: str | None` field; `_to_out` parses the `remote:` key.
- `_Remote` gains a `name` field (used as the per-remote client-cache key).
- New `_RemoteConfig` (all remotes + default name) with a `resolve()` helper for per-output lookup.
- New `_parse_dvc_remotes()` parses every remote section eagerly (URL validated, anonymous flag read); `_parse_dvc_config()` retained as a thin wrapper returning the `[core]` default remote for backward compatibility.
- `_process_pointer()` resolves each output's remote and selects the matching S3 client.
- `pull()` pre-builds one S3 client per remote (dict is read-only during the parallel phase, so no locking) and passes a `client_for` provider down.

### DIFF
--- a/build_tools/fetch_dvc_artifacts.py
+++ b/build_tools/fetch_dvc_artifacts.py
@@ -239,7 +239,7 @@ def _parse_dvc_remotes(path: Path) -> _RemoteConfig:
             continue
         url = cp[section].get("url")
         if not url:
-            raise FetchError(f'{path}: [remote "{name}"] missing \'url\'')
+            raise FetchError(f"{path}: [remote \"{name}\"] missing 'url'")
         parsed = urlparse(url)
         if parsed.scheme != "s3":
             raise FetchError(f"{path}: only s3:// remotes are supported, got {url!r}")
--- a/build_tools/fetch_dvc_artifacts.py
+++ b/build_tools/fetch_dvc_artifacts.py
@@ -13,6 +13,7 @@ Pointer file schema (the only one we handle):
     - md5: <32-char-hex>      # may end with `.dir` for directory hashes
       size: <bytes>
       hash: md5
+      remote: <name>         # optional: per-output remote override (see below)
       path: <relpath-from-pointer-file-dir>
 
 Remote config (.dvc/config, INI):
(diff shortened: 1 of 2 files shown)


JeniferC99 (approved):
lgtm

PR #6425, opened 2026-07-08
revert(hipdnn): remove Python bindings from TheRock

tvy-amd:
## Summary

We thought hipDNN bindings were going to land in TheRock artifacts, but the new artitecture is that we are going to have a different repo for all bindings.
So, this PR reverts TheRock-side hipDNN Python binding build, packaging, and wheel-validation CI wiring added by [PR #5429](https://github.com/ROCm/TheRock/pull/5429), so binding-specific CI can move out of TheRock per ALMIOPEN-2285. JIRA ID : ALMIOPEN-2285

## Risk Assessment

Medium risk. The user risk hint was low, but this changes hipDNN build, package, and test infrastructure by removing TheRock Python binding paths; local test-matrix unit coverage and formatting checks passed, with PR CI still needed for build and artifact validation.

## ASIC Coverage

Standard PR CI is sufficient. This is ASIC-independent build, CI, and artifact packaging cleanup that removes hipDNN Python binding wiring from TheRock; it does not change kernels, provider support surface, dispatch behavior, or GPU defaults, so no dedicated multi-arch sweep is required.

## Testing Summary

- Local Python unit coverage validates CI test-matrix generation after removing the hipDNN Python bindings job entry.
- Local formatting validation confirmed the changed Python files are Black-compliant.
- Reference search confirmed no remaining `nanobind`, `robin-map`, `therock-nanobind`, or `therock-robin-map` references in touched TheRock build/test areas.
- PR CI is pending and must validate the CMake, artifact packaging, and repository pre-commit effects.

## Testing Checklist

- [x] Test matrix unit tests - `python -m unittest build_tools.github_actions.tests.fetch_test_configurations_test` - Status: Passed
- [x] Python formatting - `python -m black --check build_tools/github_actions/tests/fetch_test_configurations_test.py` - Status: Passed
- [x] Dependency reference check - Search for `nanobind`, `robin-map`, `therock-nanobind`, and `therock-robin-map` references - Status: Passed

## Technical Changes

- Removes the disabled hipDNN Python bindings CI matrix entry and deletes the wheel-packaging and frontend pytest scripts from TheRock.
- Stops configuring hipDNN with Python bindings enabled and removes TheRock build dependencies only needed by that bindings path.
- Removes `share/hipdnn/python/**` artifact include/exclude rules and the shared loader-environment helper used only by the deleted Python binding test.
- Removes the now-unused `third-party/nanobind` and `third-party/robin-map` CMake subprojects from the third-party build list.
- Keeps the required companion test-file change non-functional; no new test behavior is added.


third-party/nanobind/CMakeLists.txt
tvy-amd, line 1: I don't think we want this in TheRock anymore because we are moving the bindings into their own repo. 

hipDNN were the only ones using this from what I know, and since we are not going to publish bindings in TheRock artifacts anymore, I believe we can safely remove

third-party/nanobind/CMakeLists.txt
```diff
@@ -1,27 +0,0 @@
-# Copyright Advanced Micro Devices, Inc.
-# SPDX-License-Identifier: MIT
-
-therock_subproject_fetch(therock-nanobind-sources
-  CMAKE_PROJECT
-  # Originally mirrored from: https://github.com/wjakob/nanobind/archive/refs/tags/v2.12.0.tar.gz
-  URL https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/nanobind-2.12.0.tar.gz
```
tvy-amd, line 7: I'll ask to remove these artifacts after this PR is in

build_tools/github_actions/test_executable_scripts/libhipcxx_utils.py
```diff
@@ -17,22 +16,6 @@ def prepend_env_path(env: dict, var_name: str, new_path: str):
         env[var_name] = new_path
 
 
-def build_rocm_loader_env(artifacts_path: Path) -> dict:
```
tvy-amd, line 20: This was just added for testing the hipdnn python bindings, since we getting rid of them, we might as well remove this too

build_tools/github_actions/test_executable_scripts/hipdnn/pack_frontend_wheel.py
tvy-amd, line 1: Moving all these testing files into rocm-libraries instead for now until rocm-bindings repo is up

build_tools/github_actions/tests/fetch_test_configurations_test.py
```diff
@@ -1,6 +1,7 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
+
```
BrianHarrisonAMD, line 4: Random newline hehe

third-party/nanobind/CMakeLists.txt
BrianHarrisonAMD: Did we double check no one else is using these since they were added?

BrianHarrisonAMD (approved):
Looks good, but we need to double check no one else started depending on the 3rd party deps, or any of the common scripts we added.

If they are confirmed fully dead code, then we should be good to remove.

third-party/nanobind/CMakeLists.txt
tvy-amd: Double checked in TheRock and doesn't seem like anyone is using this

geomin12 (commented):
i see build failures. don't seem unrelated, so worth a retry

i also agree Brian, we should make sure this builds all properly and double check that no components use this. a search should solve that though

ml-libs/CMakeLists.txt
```diff
@@ -102,7 +102,6 @@ if(THEROCK_ENABLE_HIPDNN)
       "-DHIP_DNN_SKIP_TESTS=$<NOT:$<BOOL:${THEROCK_BUILD_TESTING}>>"
       -DHIP_DNN_BUILD_PLUGINS=OFF
       -DHIP_DNN_GENERATE_SDK_HEADERS=OFF
-      -DHIPDNN_BUILD_PYTHON_BINDINGS=ON
```
marbre, line 105: Should this be rather set explicitly to `OFF`?

third-party/nanobind/CMakeLists.txt
marbre: Right now it is unused but even if we move the hipDNN bindings to a separate repo (and eventually we have more than one?), do you intend to build nanobind _there_ or should TheRock provide these deps?

BrianHarrisonAMD: I think it makes sense to keep the artifacts uploaded, and use those until it's clear where the new final home is.

ml-libs/CMakeLists.txt
tvy-amd: I'm actually removing this option in rocm-libraries. It is defaulted off so it's fine to remove here.

third-party/nanobind/CMakeLists.txt
tvy-amd: I believe TheRock shouldn't provide the dependencies; since in the future it will be rocm-libraries ->TheRock->rocm-bindings. If it did, it would cause a cycle dependencies.

So mostly likely rocm-bindings will build nanobind.

tvy-amd: I will keep the artifacts around in s3 then

ScottTodd (approved):

PR #6426, opened 2026-07-08
feat(ci): Add submodule_bump_tests_only flag for gfx950

geomin12:
Add a new flag `submodule_bump_tests_only` that limits test execution to submodule bump pushes and workflow_dispatch triggers. This prevents gfx950 tests from running on every postsubmit push while still running builds.

Also refactors submodule change detection into a reusable `GitContext.has_submodule_changes` property, simplifying three locations that previously duplicated this logic.

Behavior:
- gfx950 builds: Always run on all postsubmit pushes
- gfx950 tests: Only run when submodule changes are detected or on manual workflow_dispatch triggers

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -299,6 +299,18 @@ class GitContext:
         """
         return GitContext()
 
+    @property
+    def has_submodule_changes(self) -> bool | None:
+        """Check if any submodules were modified in the changed files.
+
+        Returns:
+            True if submodule changes detected, False if no submodule changes,
+            None if git context is unavailable (changed_files or submodule_paths missing).
+        """
+        if self.changed_files is None or self.submodule_paths is None:
+            return None
+        return bool(set(self.submodule_paths) & set(self.changed_files))
+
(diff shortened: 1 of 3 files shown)


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -299,6 +299,13 @@ def empty() -> "GitContext":
         """
         return GitContext()
 
+    @property
+    def has_submodule_changes(self) -> bool:
+        """Check if any submodules were modified in the changed files."""
+        if self.changed_files is None or self.submodule_paths is None:
+            return False
+        return bool(set(self.submodule_paths) & set(self.changed_files))
+
```
erman-gurses, line 313: This helper now returns False both when there are no submodule changes and when changed_files/submodule_paths are unavailable. Is this intentional?

Previously, the callers explicitly checked for None before deciding to skip work, so "unknown" and "no changes" were treated differently. With this refactor, missing metadata will also be interpreted as "no submodule changes", which could cause ASAN/tests to be skipped if the git context couldn't be populated.

erman-gurses (approved):
LGTM overall - one concern mentioned in my comment.

PR #6429, opened 2026-07-08
[TEST ONLY] Deprecate HIPCC.  Replace calls with amdclang++

david-salinas:
## Motivation

Draft for testing this PR: https://github.com/ROCm/libhipcxx/pull/20
When this runs clean we can land https://github.com/ROCm/libhipcxx/pull/20, and then do a submodule bump with the changes in build_tools and math_libs in TheRock repo.

The HIPCC Tool will be deprecated, and eventually removed in an upcoming release. Users will be directed to use amdclang++ instead and will be given a migration guide.

closes
Jira: https://amd-hub.atlassian.net/browse/ROCM-8370

Replace the hipcc fallback defaults for HOST_COMPILER and CUDA_COMPILER with amdclang++. hipcc is deprecated in favor of calling amdclang++ directly with explicit --hip-path and --hip-device-lib-path flags.

Callers that set $CXX or $CUDACXX are unaffected.

## Technical Details

  Replacing all calls to hipcc with calls to amdclang++ with needed clang options like -x hip and --hip-path and hip-device-lib-path, etc ...

## Test Plan
Full CI cycle

## Submission Checklist


ScottTodd:
> [TEST ONLY]

Are you looking for a review here? If not, please mark as a draft PR. Note that you can also trigger `multi_arch_ci.yml` via workflow_dispatch for testing if you don't want to actually create a pull request or want more precise control over which build/test workflow stages are included, see https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md.

PR #4812, opened 2026-04-24
Add test filter standardisation to rocrtst

dileepr1:
## Motivation

This PR adds test filter standardisation for rocrtst.
It works together with the rocm-systems companion PR [5485](https://github.com/ROCm/rocm-systems/pull/5485)

## Technical Details

Rocrtst uses ctest based testing and the CTest test file was already being added to therock bin directory.
This PR changes the runner from test_rocrtst.py to the unified test_runner.py.

## Test Plan

Confirm that rocrtst is getting tested properly with the changes in this PR.

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6436, opened 2026-07-09
feat(ci): add mirage/rocjitsu emulation tests

amd-arosa:
## Motivation

rocrtst and most of our other runtime tests need a real GPU, which ties them to
our (limited) pool of GPU runners. rocjitsu is a CPU-based GPU emulator — it
executes AMD GPU code objects on the host — and mirage is the CLI that drives
it. Running tests through mirage means we can exercise the runtime on plain CPU
machines, and eventually on architectures we have no silicon for yet.

This is the first emulation test wired into the matrix. The other goal was to
set it up so adding the next one doesn't require rethinking any of the CI
plumbing.

https://github.com/ROCm/rocm-systems/issues/8384

## Technical Details

Five files, and all of the emulation logic is in one of them:
[`build_tools/github_actions/emulation.py`](build_tools/github_actions/emulation.py).
`fetch_test_configurations.py` stays a list of components and calls into it.

A component opts in with one field, and the generator derives a second matrix
entry from the fully expanded hardware one:

```python
"rocrtst": {
    ...
    "emulate": "rocjitsu",
    "emulate_test_type": "quick",
    "emulate_env": {"ROCRTST_PLATFORM_OVERRIDE": "EMULATOR"},
},
```

The derived `rocrtst (emulated mi350x)` job runs on the CPU cluster
(`linux_cpu_runner`, no GPU devices in the container), additionally fetches the
`mirage` and `rocjitsu` artifacts, runs unsharded, and has its whole
`test_script` wrapped in `mirage run ... -- <script>`. Everything that excludes
the hardware job — `exclude_family`, test labels, project selection — excludes
the emulated variant too.

**The emulated job runs the same entry point as the hardware job.** rocrtst is
already on the standardized `test_runner.py` / `test_categories.yaml` flow from
#4812 + ROCm/rocm-systems#5485, and the emulated variant just selects a
different category of it. `test_rocrtst.py` is untouched.

**This PR needs no companion change in rocm-systems.** It pins `quick`, a
category rocrtst already declares at the pinned submodule commit, and
introduces no new category name. `emulate_test_type` is a **pin, not a
default**, so a nightly asking for `comprehensive` cannot drag the emulated job
into a category the emulator can't finish.

`emulate_env` is the one concession to not changing the component. rocrtst
already ships ~50 `platforms.EMULATOR.blocked_tests` entries in
`share/rocrtst/platform_config.yaml`, but it only applies them when it believes
it is emulated — which it decides from
`/sys/module/amdgpu/parameters/emu_mode`, a file rocjitsu does not provide. Two
blocked tests are in `quick` and fail outright without the override. So the
generator compiles the dict into `--env NAME=VALUE` on the mirage command,
validated as quote-free/space-free literals since the command crosses two
layers of shell quoting.

The better home for that variable is rocrtst's own `test_categories.yaml` under
`env_variables`, which `parse_test_categories.py` compiles into the CTest
entry's `ENVIRONMENT` property — scoped to the test binary rather than the whole
mirage session, and reproducing for anyone running `ctest` by hand. Both the
code comment and `docs/development/adding_tests.md` say so; `emulate_env` is the
escape hatch for when the component cannot be changed, and moving it later is a
one-line deletion here.

Two smaller pieces:

- `test_emulation.py` checks that the emulated GPU comes up and reports the
  target the job was built for. `TEST_TYPE` picks the depth:
  `rocm_agent_enumerator` at `quick`, plus `rocminfo` at `standard` and above.
  The expected target comes from `AMDGPU_TARGETS` — the job's own input —
  rather than a table of what each mirage profile emulates. It's the cheapest
  end-to-end check of mirage → rocjitsu → emulated KFD/HSA → ROCr, so when it
  fails every other emulated job is expected to fail too.
- The emulated timeout is `10 x timeout_minutes` **capped at 60 min**. Uncapped,
  rocrtst's 15 min hardware budget became 150 min — hours of a CPU slot for a
  wedged emulator. Deliberately no floor at the hardware budget: the emulated
  variant runs a cheaper category, so a component that hits the cap needs an
  `emulate_test_type`, not more time.

Linux only, since the emulation artifacts are Linux only.

### The one change outside emulation.py

`test_component.yml` gets 4 lines, 3 of them comment:

- `TEST_COMPONENT` falls back to `job_name`, so a component that decorates its
  job name (`rocrtst (emulated mi350x)`) can set `test_component` to the
  undecorated name. Without it `test_runner.py` resolves a test directory that
  does not exist. This is the one piece of the emulated job that cannot live in
  `emulation.py`.

The CPU-runner plumbing this PR previously carried — the `linux_cpu_runner`
guard on "Driver / GPU sanity check" and the base/GPU split of
`container_options` — has since landed on `main` on its own, so those hunks are
gone and the emulated job just inherits the behaviour.

### Changed since the last review

Rebased onto `main` (60 commits). Two things fell out of the diff because they
landed on `main` independently: the `linux_cpu_runner` guard on the
"Driver / GPU sanity check" step, and the base/GPU split of `container_options`.
`test_component.yml` is down from 8 lines to 4, and the five review-response
commits are squashed into one.

Per @geomin12's review:

- All emulation logic moved out of `fetch_test_configurations.py` into the new
  `emulation.py`, matrix side and run-time side together. The generator keeps
  three thin calls.
- `test_artifacts.yml`, `reproduce_test_failure.py`, `test_runner.py` and
  `fetch_test_configurations_test.py` are back to `main` — zero diff.
- `test_emulation_smoke.py` → `test_emulation.py`, depth from `TEST_TYPE`,
  expected target from the job's inputs, `check=True` instead of inspecting
  `returncode`.
- `-e KUBE_CPU_REQUEST` is back on the GPU-only container options, and the
  CPU-only runner-label change is gone.
- The `emulation_test.py` unit tests are dropped.
- Comments trimmed.

Two deliberate deviations, both flagged in the review thread:

- I kept the `TEST_COMPONENT` fallback in `test_component.yml` — without it the
  emulated job looks for a test directory named `rocrtst (emulated mi350x)`.
- The empty-ctest-selection guard is deferred rather than made unconditional:
  `ctest` exits 0 when a label selection matches nothing, so making that a hard
  failure could turn currently-green hardware jobs red. Its own PR.

### Known follow-ups

- Move `ROCRTST_PLATFORM_OVERRIDE` into rocrtst's `test_categories.yaml`, and
  consider an `ffm-quick` tier there covering the 62 tests that pass under
  rocjitsu rather than the 13 `quick` gives. Both are rocm-systems changes and
  neither needs to block this.
- The component-repo copies of `therock-test-component.yml` (`rocm-systems`,
  `rocm-libraries`) lack the `test_component` fallback added here. rocrtst lives
  in `rocm-systems`, so their pinned TheRock ref must not move past this commit
  until they carry it.
- `gfx125X` (`mi450x`) generates emulated entries correctly but has
  `test-runs-on: ""`, so its whole test lane is skipped. Enabling it is a
  follow-up PR, together with turning on its builds.
- **The emulated job does not run in this PR's own presubmit.** `gfx950-dcgpu`
  carries `submodule_bump_tests_only: True`, so `configure_multi_arch_ci.py`
  blanks its `test-runs-on` on any PR without a submodule bump, and the whole
  Linux gfx950 test lane skips. `gfx94X`/`mi300x` is deliberately not in
  `EMULATED_PROFILES`, and `gfx125X`/`mi450x` is dark. Net: presubmit here can
  go green without exercising emulation at all. It runs on nightlies; a
  `workflow_dispatch` of `test_artifacts.yml` is the manual escape hatch.
- Emulated jobs are placed on CPU runners, but the *lane* is still gated on GPU
  availability twice over: `configure_test_matrix` is `if: test_runs_on != ''`,
  and `test_components` needs `test_sanity_check`, which runs on a GPU runner.
  So a family with no GPU pool runs no emulated jobs either — which is the real
  reason gfx125X above is dark. Decoupling that is what makes emulation useful
  for architectures we have no silicon for, and it is scheduling policy that
  belongs in the follow-up rather than here.
- `emulation` is filtered out by `test:` labels, so a PR labelled
  `test:rocrtst` runs the emulated rocrtst job without the check that would
  explain a failure. Exempting it like `sanity` is a scheduling-policy call I
  left alone.

## Test Plan

Ran the **exact command the generator emits**, against real gfx950-dcgpu
artifacts (run 31148543107) on a CPU-only path:

```
$THEROCK_BIN_DIR/mirage run --profile mi350x --emulator rocjitsu \
  --env TEST_EMULATOR=rocjitsu --env TEST_EMULATOR_PROFILE=mi350x \
  --env ROCRTST_PLATFORM_OVERRIDE=EMULATOR \
  ${THEROCK_BIN_DIR:+--env THEROCK_BIN_DIR=$THEROCK_BIN_DIR} ... \
  -- python build_tools/github_actions/test_executable_scripts/test_runner.py
```

against the `CTestTestfile.cmake` the artifacts actually ship, not a
hand-written one. Measured `quick` both with and without the override, and
`standard` as the alternative. Ran `test_emulation.py` the same way at both
depths, plus its mismatch and not-emulated paths. Plus
`pytest build_tools/github_actions` and `pre-commit`.

## Test Result

Emulated rocrtst, gfx950-dcgpu / mirage `mi350x` / rocjitsu:

| Category               | Tests run                        | Wall time  | Result                  |
| ---------------------- | -------------------------------- | ---------- | ----------------------- |
| `quick` (pinned)       | **13** (+6 skipped by the filter) | **9.6 s**  | **100% passed**         |
| `quick`, no override   | 16                               | 12 s       | 14 passed, **2 failed** |
| `standard`             | 65                               | 9 min 45 s | 62 passed, **3 failed** |

Without `ROCRTST_PLATFORM_OVERRIDE=EMULATOR`, `IPC` and
`Deallocation_Notifier_Test` fail, and `IPC` leaves a child spinning at 100% CPU
after the suite has already reported — rocrtst ships both in its EMULATOR block
list but never applies it, because it cannot tell it is emulated.

`standard` is affordable on time but not on correctness. Its three extra
failures are emulator gaps, not ROCr regressions, and excluding them needs a
rocrtst-side change — hence the follow-up above rather than a wider pin here:

- `MemoryAllocateContiguousTest` — `hsa_amd_interop_map_buffer` returns 4096
  (dmabuf interop unimplemented)
- `SvmMemory_AccessedBy_All_Devices_Test` — SVM attribute query returns 512,
  expected 514
- `FP_Exception_Shutdown` — needs exception delivery rocjitsu doesn't model

`test_emulation.py`: `quick` and `standard` both report a `gfx950` agent under
the `mi350x` profile in ~3 s. Exits 1 on an `AMDGPU_TARGETS` mismatch and when
run outside a mirage session.

Unit tests: `build_tools/github_actions` fully green, 776 passed / 0 failed
(the two `stage_reuse_decision` failures noted earlier are fixed on `main`).
`pre-commit` clean.

Matrix parity: generated the Linux matrix from this branch and from `origin/main`
for gfx950-dcgpu, gfx94X-dcgpu, gfx110X-all, gfx1151, gfx120X-all and
gfx125X-all. The only difference on any family is the two emulated entries added
on gfx950-dcgpu and gfx125X-all — no existing job is added, dropped or altered.
(Generating `main` twice produces its own variation in `rocgdb-corefile`,
`hiptensor` and `rocfft` from the weighted runner draw; that is excluded from the
comparison as pre-existing non-determinism.)

## Submission Checklist


.github/workflows/test_artifacts.yml
```diff
@@ -217,13 +217,19 @@ jobs:
...
+      #      only occupies scarce hardware. That matters most for emulated
+      #      components, whose timeouts are 10x their hardware counterparts.
+      #   1. For workflow_dispatch, honor the explicit test_runs_on input if provided
+      #   2. Benchmark components use the dedicated benchmark runner
+      #   3. Multi-GPU components use the multi-GPU runner
       #   4. Per-component runner (set by fetch_test_configurations.py)
       test_runs_on: >-
         ${{
+          (matrix.components.linux_cpu_runner == true &&
+            'azure-linux-scale-rocm') ||
```
geomin12: oh, we also use aws-linux-scale-rocm-prod now. this is an oudated label

build_tools/github_actions/reproduce_test_failure.py
```diff
@@ -152,19 +171,26 @@ def run_linux(args: argparse.Namespace) -> int:
...
-        "--rm",
-        "-it",
-        "--ipc",
-        "host",
-        "--group-add",
-        "video",
-        "--device",
-        "/dev/kfd",
-        "--device",
-        "/dev/dri",
```
geomin12, line 167: should be fine to keep for cpu tests, as it won't make an impact. let's keep this

build_tools/github_actions/test_executable_scripts/test_rocjitsu_corpus.py
geomin12, line 1: these are some massive changes and a huge PR. can we break it up into smaller PRs?

geomin12 (commented):
before we continue adding features, you should collaborate with @dileepr1 about test filter standardization. we can incorporate these features / scripts there

for every change we make to these test scripts, it's more debt as we will eventually move entirely to test filter work

build_tools/github_actions/test_executable_scripts/test_rocjitsu_corpus.py
amd-arosa: split it in half. can't think of good way to split it further.

.github/workflows/test_artifacts.yml
```diff
@@ -231,15 +231,24 @@ jobs:
...
       #      Bot-triggered runs (e.g., nightly from rockrel) skip this and use rules below.
-      #   2. Multi-GPU components use the multi-GPU runner (required for correct execution)
-      #   3. Benchmark components use the dedicated benchmark runner
-      #   4. CPU-only components use standard GitHub runners (aws-linux-scale-rocm-prod)
+      #   3. Multi-GPU components use the multi-GPU runner (required for correct execution)
+      #   4. Benchmark components use the dedicated benchmark runner
       #   5. Per-component runner (set by fetch_test_configurations.py)
       test_runs_on: >-
         ${{
+          (matrix.components.linux_cpu_runner == true &&
```
geomin12: rebase pls. i made some fixes here

.github/workflows/test_artifacts.yml
```diff
@@ -231,15 +231,24 @@ jobs:
...
+      #      override, which is an `inputs.` value the matrix cannot opt out of.
+      #   2. For user-triggered workflow_dispatch, honor test_runs_on input if provided.
       #      Only applies to workflow_dispatch events (not PR/workflow_call chains).
+      #      User-triggered means github.actor does NOT contain 'github-actions'.
       #      Bot-triggered runs (e.g., nightly from rockrel) skip this and use rules below.
-      #   2. Multi-GPU components use the multi-GPU runner (required for correct execution)
-      #   3. Benchmark components use the dedicated benchmark runner
-      #   4. CPU-only components use standard GitHub runners (aws-linux-scale-rocm-prod)
+      #   3. Multi-GPU components use the multi-GPU runner (required for correct execution)
+      #   4. Benchmark components use the dedicated benchmark runner
```
geomin12: i made a fix. pls rebase and keep original order

.github/workflows/test_component.yml
```diff
@@ -197,7 +200,25 @@ jobs:
...
+          # builtin profile pinning the emulated GPU (e.g. "mi350x"). Both are
+          # empty for jobs running against real hardware.
+          #
+          # These describe the job to anything reading its environment; they are
+          # not how the test script learns it is emulated. `mirage host` clears
+          # the environment, so the values the script actually sees come from
+          # the `--env TEST_EMULATOR=...` literals that _wrap_in_mirage_run()
+          # bakes into test_script.
+          TEST_EMULATOR: ${{ fromJSON(inputs.component).emulator }}
+          TEST_EMULATOR_PROFILE: ${{ fromJSON(inputs.component).emulator_profile }}
```
geomin12: we should keep these environment variables agnostic of component. 

we shouldn't have to add extra "emulator profile" or "emulator" env variables, or emulator specific comments.  this should be contained in emtulation_test file or in some setup process

.github/workflows/test_component.yml
```diff
@@ -226,7 +247,8 @@ jobs:
             --repository "${{ github.repository }}" \
             --amdgpu-family "${{ inputs.amdgpu_families }}" \
             --amdgpu-targets "${{ inputs.amdgpu_targets }}" \
-            --test-script "${{ fromJSON(inputs.component).test_script }}" \
+            --test-script '${{ fromJSON(inputs.component).test_script }}' \
+            --test-component "${{ fromJSON(inputs.component).test_component || fromJSON(inputs.component).job_name }}" \
```
geomin12: seems irrelevant to mirate/rocjitsu test. can be split up into new PR if it's CI script updates

build_tools/github_actions/test_executable_scripts/test_emulation_smoke.py
geomin12, line 1: we should call this just test_emulation.py, then use https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md and test filter env variable to determine whether this is "smoke" or "standard" or "comprehensive"

build_tools/github_actions/test_executable_scripts/test_emulation_smoke.py
```diff
@@ -0,0 +1,156 @@
...
+    if not gfx_names:
+        problems.append("rocminfo reported no gfx agent name")
+
+    expected = EXPECTED_GFX_BY_PROFILE.get(profile)
+    if expected and not any(name.startswith(expected) for name in gfx_names):
+        problems.append(
+            f"expected a {expected} agent for mirage profile '{profile}', "
+            f"but rocminfo reported {sorted(gfx_names) or 'none'}"
+        )
+    return problems
```
geomin12: seems hacky. why can't we just use the inputs given to this script and use that? why have to parse rocminfo?

build_tools/github_actions/test_executable_scripts/test_emulation_smoke.py
```diff
@@ -0,0 +1,156 @@
...
+EXPECTED_GFX_BY_PROFILE = {
+    "mi300x": "gfx942",
+    "mi350x": "gfx950",
+    "mi450x": "gfx1250",
+}
+
+# rocminfo has to come up, enumerate the emulated agent and exit. Anything
+# beyond this means the stack is wedged, and a hang that runs out the step
+# timeout would otherwise print nothing at all.
+ROCMINFO_TIMEOUT_SECONDS = 15 * 60
```
geomin12: we run gpu_sanity_check script before this runs. that'll handle there. if emulator specifically requires rocminfo to run, please add a comment noting why

build_tools/github_actions/test_executable_scripts/test_emulation_smoke.py
```diff
@@ -0,0 +1,156 @@
...
+        print(f"ERROR: rocminfo not found at {rocminfo}", file=sys.stderr)
+        return 1
+
+    # Already inside the mirage session, so this is a plain invocation.
+    cmd = [str(rocminfo)]
+    logging.info(f"++ Exec $ {' '.join(cmd)}")
+    try:
+        result = subprocess.run(
+            cmd,
+            check=False,
```
geomin12: check MUST BE TRUE

otherwise, this will never get caught and any error will be considered pass. please update to check_call or check=True

build_tools/github_actions/test_executable_scripts/test_emulation_smoke.py
```diff
@@ -0,0 +1,156 @@
...
+    if result.stderr:
+        print(result.stderr, file=sys.stderr)
+
+    if result.returncode != 0:
+        print(
+            f"ERROR: rocminfo exited {result.returncode} under mirage profile "
+            f"'{profile}'.",
+            file=sys.stderr,
+        )
+        return result.returncode
```
geomin12: we can just do check_call or check=True to check if tests passed. can be removed

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -590,6 +593,56 @@ def build_ctest_command(
...
+        list_cmd.append(arg)
+    list_cmd.append("-N")
+    try:
+        result = subprocess.run(
+            list_cmd,
+            cwd=THEROCK_DIR,
+            env=environ_vars,
+            capture_output=True,
+            text=True,
+            timeout=CTEST_LIST_TIMEOUT_SECONDS,
```
geomin12: can be added to another PR. seems irrelevant to mirate/rocjitsu specific tests

build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -638,6 +691,30 @@ def main():
...
+    # only to run the category their matrix entry names, so a missing category
+    # (an artifact built before the component declared it, a typo, a category
+    # with no <name>_<gfx> entry) means the job tested nothing at all.
+    if selection_is_empty(cmd):
+        print(
+            f"WARNING: no tests match category '{category}'"
+            + (f" on {gpu_arch}" if gpu_arch else "")
+            + f" in {TEST_DIR}.",
+            file=sys.stderr,
+        )
```
geomin12: we should return code 1 for this if no tests were found

build_tools/github_actions/tests/emulation_test.py
geomin12, line 1: this is confusing. doesn't test_emulation_smoke.py just handle this during test run time?

geomin12: we really don't need unit tests for test_executable_scripts. the intent of `test_executable_scripts` is a simple python wrapper around the script.  we don't need unit tests for a wrapper and it'll be tested on CI run with actual execution against GPU

build_tools/github_actions/tests/fetch_test_configurations_test.py
```diff
@@ -546,6 +560,711 @@ def fake_get_all_families(_):
         hipblas = next(j for j in components if j["job_name"] == "hipblas")
         self.assertEqual(hipblas["test_runner"], "linux-gfx942-default")
 
+    # -----------------------
```
geomin12, line 649: fetch_test_configuration is only meant to fetch test jobs. we should make it component agnostic

this can be removed

geomin12: this is also an awful amount tests for just a dictionary entry

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -56,12 +57,18 @@ def _get_script_path(script_name: str) -> str:
 # --ulimit memlock=-1:-1 - Prevents memory allocation issues with ROCm inside container
 # --ulimit nofile=1048576:1048576 - Increase open file limit for RCCL
 # --security-opt seccomp=unconfined - enables memory mapping, and is recommended for containers running in HPC environments
+# -e KUBE_CPU_REQUEST - Pass the pod's CPU budget in; test_component.yml's
+#   "Set thread limits" step derives OMP_NUM_THREADS / OPENBLAS_NUM_THREADS
+#   from it. Not GPU-specific, and previously only reached GPU jobs: CPU-only
+#   jobs (hipfile, rocgdb-cpu, every emulated component) are the ones most
+#   likely to oversubscribe without it.
```
geomin12: irrelevant to scope of this PR. can be opened in separate PR

build_tools/github_actions/reproduce_test_failure.py
geomin12, line 1: updates in here are irrelevant to scope of PR. please remove and add in a separate PR

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -1051,6 +1423,14 @@ def run():
...
                 )
                 continue
+        elif component.get("linux_cpu_runner", False):
+            # CPU-only components (every emulated one, plus hipfile / rocgdb-cpu)
+            # are placed on the CPU cluster by test_artifacts.yml. Don't draw a
+            # GPU runner label for them: it would never be used, it consumes a
+            # draw from the weighted pool that a GPU job should get, and it
+            # leaves a GPU label on the component that debugging tools and the
+            # workflow_dispatch override would otherwise take at face value.
+            logging.info(f"  {job_name}: CPU-only component, no GPU runner assigned")
```
geomin12: updates in here are irrelevant to scope of PR. please remove and add in a separate PR

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -80,10 +87,267 @@ def _get_script_path(script_name: str) -> str:
...
+# emulator, so the family under test only decides which GPU mirage pretends to
+# be -- the job itself runs on a CPU runner with no GPU attached.
+#
+# Keys are matched as prefixes of the AMDGPU_FAMILIES value ("gfx950-dcgpu",
+# "gfx94X-dcgpu", "gfx125X-dcgpu", ...), longest key first. Each key also
+# covers the bare targets under it ("gfx94" matches "gfx942", "gfx125" matches
+# "gfx1250"), so no key is a prefix of another today; matching is still
+# longest-first so a future split (e.g. a "gfx951" that is not an MI350X) can
+# be added as a longer, more specific key.
+_MIRAGE_PROFILE_BY_FAMILY_PREFIX = {
```
geomin12: thought this was handled in test_emulation_smoke.py ?

this script is only meant to fetch the dictionary items below and report to CI which tests to run with what container options etc

perhaps we add these helpers to emulation.py so this python script can be component agnostic?

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -80,10 +87,267 @@ def _get_script_path(script_name: str) -> str:
...
+    "gfx94": "mi300x",  # gfx94X family label (gfx940/gfx941/gfx942)
+    "gfx950": "mi350x",
+    "gfx125": "mi450x",  # gfx125X family label, and gfx1250
```
geomin12: repeat in test_emulation_smoke.py. let's avoid de-dup

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -80,10 +87,267 @@ def _get_script_path(script_name: str) -> str:
...
+    # that test_artifacts.yml's `test_runs_on` chain checks *ahead of*
+    # `linux_cpu_runner` -- otherwise the emulated variant is routed to GPU
+    # hardware it cannot use. That chain is, in order: multi_gpu_runner,
+    # the workflow_dispatch override, is_benchmark, then linux_cpu_runner.
+    emulated.pop("multi_gpu", None)
+    emulated.pop("multi_gpu_runner", None)
+    emulated.pop("is_benchmark", None)
+    for key in _EMULATION_MATRIX_KEYS:
+        emulated.pop(key, None)
+    return emulated
```
geomin12: this python script is designed to just get the component test wrapper data and feed it into CI. not sure why we are including "build_emulated_jobs" functions in the CI test retrieval script. 

let's remove this and add it to somewhere more relevant. i want to keep this file component agnostic. all the code here seems very specific to emulation/rocjitsu/mirate, so let's create a new file relevant to this pls

if a developer wanted to see what CI tests we run, they look here. they shouldn't need to look at "emulation build" or "get mirage profile" work

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -815,6 +1079,57 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+        # that it already ships. Two of those, IPC and
+        # Deallocation_Notifier_Test, are in `quick` and fail outright under
+        # rocjitsu; IPC also leaves a child spinning at 100% CPU after the suite
+        # has reported. Measured both ways: with the override 13/13 pass, without
+        # it 14 pass and 2 fail.
+        #
+        # This belongs in rocrtst's own test_categories.yaml, which compiles
+        # `env_variables` into the CTest entry's ENVIRONMENT property and would
+        # scope it to the test binary and reproduce outside CI. Keeping it here
+        # is the cost of not requiring a rocm-systems change to land first.
```
geomin12: comments are good, but can we keep it concise and scoped? let's assume a developer reading these comments have no context.

I don't even know what Deallocation_Notifier_Test, do i really need to know this?

geomin12 (changes requested):
let's keep this PR scoped to the title. when I see `add mirage/rocjitsu emulation tests`, i expect just some new script additions + dictionary entries in `fetch_test_configurations.py`

a lot of these updates are adding emulation feature that could be in a separate PR

I also am not understanding the context of emulation. i'm on the only reviewer, but let's get some rocjitsu / mirage / emulation team members to review and see how this all fits in

pls update title to reflect the PR correctly. this PR can definitely be chunked up and i would recommend to do so. 

there's a lot of code bloat too , the comments are helpful of course but almost paragraph size. let's keep it concise and remember, anyone can read and it must be understandable with no context

dileepr1:
Looking from the Test Filter Standardization point of view, the core design is right: the emulated rocrtst variant just wraps the unchanged test_runner.py invocation in a mirage run session and pins an existing category (emulate_test_type: quick) — no new category, no per-component filtering in fetch_test_configurations.py, no fork of the test definition. 
That's exactly the direction we want, and keeping all the logic in emulation.py is clean.

A suggestion: emulate_env is a second mechanism for injecting test environment that overlaps with test_categories.yaml's env_variables. 
You can move the ROCRTST_PLATFORM_OVERRIDE=EMULATOR to test_categories.yaml (test_categories.yaml support env variables which can be passed directly to ctest - either with global scope or category wise - like that in [rocblas](https://github.com/ROCm/rocm-libraries/blob/develop/projects/rocblas/clients/gtest/test_categories.yaml) or [rccl](https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/test/test_categories.yaml)).

amd-arosa:
> A suggestion: emulate_env is a second mechanism for injecting test environment that overlaps with test_categories.yaml's env_variables. You can move the ROCRTST_PLATFORM_OVERRIDE=EMULATOR to test_categories.yaml (test_categories.yaml support env variables which can be passed directly to ctest - either with global scope or category wise - like that in [rocblas](https://github.com/ROCm/rocm-libraries/blob/develop/projects/rocblas/clients/gtest/test_categories.yaml) or [rccl](https://github.com/ROCm/rocm-systems/blob/develop/projects/rccl/test/test_categories.yaml)).

The override has to be scoped to emulated runs, and test_categories.yaml env is global or per-category. The emulated job runs the same quick category hardware does, so putting ROCRTST_PLATFORM_OVERRIDE=EMULATOR there would set it on real GPUs too and skip ~50 tests that pass. Scoping it would need an emulator-only category.

eagarvey-amd (approved):
LGTM. Emulation correctness validation feels best running right alongside the real thing. I think this is ready for landing and iteration

amd-eochoalo (approved):
@amd-arosa I think this looks ok from the rocjitsu side. I am not too familiar with TheRock to comment on that fully, but it seems reasonable. My only comment is about the failed 2 or 3 failed tests in the quick and standard run. Are there tickets for these already in rocm-systems?

Also, with so many tests (and my lack of familiarity with TheRock) can you provide a link to the GH that runs one of the mirage emulation tests?

amd-arosa:
> Also, with so many tests (and my lack of familiarity with TheRock) can you provide a link to the GH that runs one of the mirage emulation tests?

https://github.com/ROCm/TheRock/actions/runs/32404138069

eagarvey-amd:
> @amd-arosa I think this looks ok from the rocjitsu side. I am not too familiar with TheRock to comment on that fully, but it seems reasonable. My only comment is about the failed 2 or 3 failed tests in the quick and standard run. Are there tickets for these already in rocm-systems?
> 
> Also, with so many tests (and my lack of familiarity with TheRock) can you provide a link to the GH that runs one of the mirage emulation tests?

The CI failures appear to be timeouts and/or consistent with the base commit.

Specifically, compared against the exact base SHA CI run 32291068602 for 691329a, the timeout failures that were not seen
  on base are:

  - Windows gfx110X-all / rocfft
      - rocfft-test_quick_suite
      - Timed out after ~600s.
      - Base Windows rocfft passed.

  - Linux gfx94X-dcgpu / rocprofiler-systems
      - transpose-sampling
      - Log shows Timeout after 120s.
      - Base rocprofiler-systems passed.

  - Linux gfx94X-dcgpu / rocprofiler-compute
      - Whole Test step timed out after 60 minutes.
      - Base rocprofiler-compute passed.

  - Windows gfx110X-all / libhipcxx_amdclang
      - std/utilities/template.bitset/bitset.members/left_shift.pass.cpp
      - std/utilities/template.bitset/bitset.members/reset_one.pass.cpp
      - std/utilities/template.bitset/bitset.members/right_shift.pass.cpp
      - Then the whole Test step timed out after 30 minutes.
      - Base libhipcxx_amdclang failed too, but I did not see these timeout signatures in the base log.
      
      
  The other failures/timeouts are consistent with the base SHA CI. What's to be done about these timeouts? Is it worth rerunning the CI if we have high confidence failures are due to flakiness?

docs/development/adding_tests.md
```diff
@@ -66,8 +66,238 @@ In [`fetch_test_configurations.py`](../../build_tools/github_actions/fetch_test_
...
+[rocjitsu](https://github.com/ROCm/rocm-systems/tree/develop/emulation/rocjitsu) software GPU emulator instead
+of talking to a real device.
+
+Opting a component in is a single field:
+
+```
+"rocrtst": {
+    "job_name": "rocrtst",
+    ...
+    "emulate": "rocjitsu",
```
geomin12, line 97: can we make "emulate": "rocjitsu" bold? to highlight that this is the field being changed

docs/development/adding_tests.md
```diff
@@ -66,8 +66,238 @@ In [`fetch_test_configurations.py`](../../build_tools/github_actions/fetch_test_
...
+}
+```
+
+That produces a second matrix entry alongside the hardware one, named
+`rocrtst (emulated mi350x)`, which:
+
+- runs on the CPU cluster (`linux_cpu_runner`), with no GPU devices mapped into
```
geomin12:
```suggestion
- runs on the Linux CPU builder  with no GPU devices mapped into
```

docs/development/adding_tests.md
```diff
@@ -66,8 +66,238 @@ In [`fetch_test_configurations.py`](../../build_tools/github_actions/fetch_test_
...
+  the container — rocjitsu emulates the GPU in software, so an emulated job on
+  GPU hardware would hold a scarce runner for its whole timeout and use none
+  of it,
+- additionally fetches the `mirage` and `rocjitsu` artifacts,
+- gets 10x the component's `timeout_minutes`, capped at one hour — the budget
+  an emulated component has to fit in to be worth scheduling at all, and so the
```
geomin12: fix wording. it's confusing

docs/development/adding_tests.md
```diff
@@ -66,8 +66,238 @@ In [`fetch_test_configurations.py`](../../build_tools/github_actions/fetch_test_
...
+(see [Choosing what an emulated job runs](#choosing-what-an-emulated-job-runs)),
+but through the same mechanism.
+
+Everything that excludes the hardware job — `exclude_family`, test labels,
+project selection — excludes the emulated variant too. The one exception is the
+multi-GPU availability check: rocjitsu emulates the device in software, so a
+component whose family has no multi-GPU pool still gets its emulated variant.
+
+> [!IMPORTANT]
+> Emulated jobs must never be routed to GPU hardware — they would hold a scarce
```
geomin12: if this is true, do we have logic where "emulate_only" is mapped to a CPU builder? seems it still routes to GPU

docs/development/adding_tests.md
geomin12, line 1: i would reach through the changes here again. a lot of it is written, but it is hard to read, with lots of random commas and punctuations that don't match well

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -855,6 +856,25 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+        "job_name": "emulation",
+        "fetch_artifact_args": "--base-only",
+        "timeout_minutes": 3,
+        "test_script": f"python {_get_script_path('test_emulation.py')}",
+        "platform": ["linux"],
+        "total_shards_dict": {
+            "linux": 1,
+        },
+        "emulate": "rocjitsu",
+        "emulate_only": True,
```
geomin12, line 920: this doesn't map to CPU job yet

docs/development/adding_tests.md
monorimet: The docs now explain that emulate_only keeps the base entry out of the matrix, and the emitted job is the derived build_emulated_job() result with linux_cpu_runner: true.

build_tools/github_actions/fetch_test_configurations.py
monorimet: Added the linux_cpu_runner flag on this for now.

lumachad:
Could you please point out in the jobs ran in this PR where exactly is the emulated job? We should at least differentiate it with a proper name so we know what to look for.

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -1154,6 +1201,9 @@ def run():
                     f"Excluding job {job_name}: multi-GPU required but no multi-GPU runner configured"
                 )
                 continue
+        elif component.get("linux_cpu_runner", False):
+            # CPU jobs route through linux_cpu_runner, not a family runner.
+            component.pop("test_runner", None)
```
geomin12: no need. our logic already handles this. can remove

docs/development/adding_tests.md
```diff
@@ -99,49 +101,30 @@ Opting a component in is a single field:
...
-> per-component runner. So the emulated variant must drop *every* key checked
-> ahead of it — `build_emulated_job` pops `multi_gpu`, `multi_gpu_runner`, and
-> `is_benchmark`. Anything added ahead of `linux_cpu_runner` in that chain must
-> be popped there too. A `workflow_dispatch` run that sets `test_runs_on`
-> explicitly still overrides this, as it does for every other CPU-only
-> component.
+> `emulate_only` keeps the base entry out of the matrix. The emitted job is the
+> one returned by `build_emulated_job()`, which sets `linux_cpu_runner: true` and
+> removes `multi_gpu`, `multi_gpu_runner`, and `is_benchmark` so
+> `test_artifacts.yml` routes it to the Linux CPU builder.
```
geomin12: is handled here: https://github.com/ROCm/TheRock/blob/c3621af90a101adb38758ecff0a27b9dddface6d/.github/workflows/test_artifacts.yml#L247-L258

can remove

geomin12 (approved):
thanks. few nits!

lumachad:
Do we have a way to handle additional time/timeout requirements due to emulation taking longer for some tests?

monorimet:
> Do we have a way to handle additional time/timeout requirements due to emulation taking longer for some tests?

Not yet. For now, the supported mechanism is global 10x scaling plus emulate_test_type to choose a cheaper category. 
 If we need component-specific extra time, adding an explicit optional field like emulate_timeout_minutes can be added in a follow-up PR either separately or bundled with whichever jobs need the exception.

lumachad:
> > Do we have a way to handle additional time/timeout requirements due to emulation taking longer for some tests?
> 
> Not yet. For now, the supported mechanism is global 10x scaling plus emulate_test_type to choose a cheaper category. If we need component-specific extra time, adding an explicit optional field like emulate_timeout_minutes can be added in a follow-up PR either separately or bundled with whichever jobs need the exception.

A general 10x timeout might not be a good choice for some components like rocgdb, as tests may timeout sporadically due to concurrency of the GPU runners and those should timeout as fast as possible. Making it 10x means it will take much longer to do so.

Maybe we should have overriding per-component. And then each component can decide what to do when it knows it is testing against emulation.

Issue #6388, opened 2026-07-07
Replace per-run Dockerfile build with frozen CI image and optimize trigger policy


erman-gurses:
The TODO will be tracked here from #3878.

Related PR: #6439

## Summary

The JAX CI wheel workflow currently builds the rocm-jax manylinux Dockerfile as part of every workflow run:

```bash
docker build \
  -t "${MANYLINUX_IMAGE_TAG}" \
  --file=Dockerfile.jax-manylinux_2_28-therock \
  --build-arg=THEROCK_INDEX_URL="${ROCM_PACKAGE_FIND_LINKS_URL}" \
  --build-arg=THEROCK_VERSION="${THEROCK_VERSION}" \
  --build-arg=GFX_ARCH="${GFX_ARCH}" \
  --progress=plain \
  .
```

This should eventually be replaced with a frozen/prebuilt TheRock CI image that is updated infrequently instead of rebuilt during each CI run.

For the initial JAX CI rollout, the extra runtime is acceptable, but this issue tracks follow-up work to improve both:

- how long the JAX CI build path takes
- when JAX CI should run

## Problem

The current rocm-jax manylinux Dockerfile build has several issues:

- It builds LLVM from source, even though LLVM is already built and distributed as part of ROCm.
- It installs the AWS CLI from a domain we do not control.
- It likely has poor cache behavior; Docker layer caching is not a good substitute for ccache/sccache.
- Building the container during every workflow run increases CI runtime and makes failures harder to isolate.
- Dockerfile setups like this are awkward for local development, debugging, and reproducing CI behavior.
- The JAX CI workflow should not need to build a custom container dynamically just to run the wheel build.

There is also a trigger-scope issue: JAX CI should not run for unrelated changes. For example, PRs that only modify PyTorch-specific code should not need to trigger JAX wheel builds.

## Direction

Instead of building the rocm-jax manylinux image during every workflow run, the JAX CI wheel workflow should use one of:

- the existing TheRock manylinux build image
- a small frozen/prebuilt variant of that image that includes Bazel and any other JAX build dependencies

The workflow should then:

1. Run in the frozen/prebuilt TheRock CI image.
2. Install ROCm packages into a venv from the CI `--find-links` URL.
3. Run the JAX `python build/build.py` wheel build directly.
4. Upload the resulting JAX wheels to the run-scoped CI Python artifacts.

## Follow-up items

- Replace the per-run rocm-jax Dockerfile build with a frozen/prebuilt CI image.
[...]

PR #6037, opened 2026-06-22
ci: update default Linux build runner from Azure to AWS

kailash-khalasi:
## Summary

Migrates all Linux workflow jobs from \`azure-linux-scale-rocm\` to \`aws-linux-scale-rocm-prod\`, following the [amdgpu_family_matrix.py](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/amdgpu_family_matrix.py) update that routes 100% of regular Linux builds to AWS.

### Changes

**Input defaults** (stale fallbacks that would route ad-hoc dispatches to the wrong pool):
- \`.github/workflows/multi_arch_build_portable_linux.yml\` — \`build_runs_on\` default
- \`.github/workflows/multi_arch_build_portable_linux_artifacts.yml\` — \`build_runs_on\` default

**Hardcoded / conditional \`runs-on\` fields:**
- \`multi_arch_ci_linux.yml\` — \`copy_prebuilt_stages\` job (was hardcoded)
- \`multi_arch_build_native_linux_packages.yml\`
- \`multi_arch_build_tarballs.yml\`
- \`manifest-diff.yml\`
- \`build_portable_linux_python_packages.yml\`
- \`multi_arch_build_portable_linux_pytorch_wheels.yml\`
- \`multi_arch_build_portable_linux_pytorch_wheels_ci.yml\`
- \`test_artifacts_structure.yml\`
- \`multi_arch_build_linux_jax_wheels.yml\`

### Not changed

- \`azure-linux-scale-rocm-heavy-ramdisk\` — still used for sanitizer (asan/tsan) builds; no AWS ramdisk equivalent yet
- \`azure-windows-scale-rocm\` — Windows builds remain on Azure

## Test plan

- [ ] Verify standard release pipeline is unaffected (configure script already selects \`aws-linux-scale-rocm-prod\` at runtime)
- [ ] Verify direct \`workflow_dispatch\` on \`multi_arch_build_portable_linux.yml\` without \`build_runs_on\` now routes to \`aws-linux-scale-rocm-prod\`
- [ ] Verify \`copy_prebuilt_stages\` job in \`multi_arch_ci_linux.yml\` routes to AWS

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6439, opened 2026-07-09
feat: (jax) Add Build JAX wheels in Multi-Arch CI

erman-gurses:
Progress #3878.
 ## Summary

Adds JAX wheel build coverage to Multi-Arch CI.

This replaces the placeholder JAX CI workflow with a real manylinux JAX
wheel build flow, wires it into Linux Multi-Arch CI, and enables JAX
wheel builds by default alongside PyTorch.

## Changes

- Replaces the placeholder JAX CI reusable workflow with a real build flow:
  - checks out TheRock, rocm-jax, and JAX
  - builds the JAX manylinux image
  - builds `jax-rocm-plugin` and `jax-rocm-pjrt` wheels
  - extracts JAX wheel versions
  - uploads built wheels to run-scoped CI Python artifacts

- Adds `build_jax` to `multi_arch_ci.yml`:
  - default: `true`
  - push / pull_request: JAX builds run by default
  - workflow_dispatch: JAX builds can be disabled manually

- Wires `build_jax_wheels` into `multi_arch_ci_linux.yml`:
  - depends on `build_python_packages`
  - consumes `build_python_packages.outputs.package_find_links_url`
  - uses `jax_build_matrix` from the generated Linux build config

- Adds a CI-specific JAX matrix helper:
  - CI uses only manylinux JAX refs
  - release workflows keep the full JAX release matrix

## CI vs release matrix behavior

The release JAX matrix still includes:

- `rocm-jaxlib-v0.9.1`
- `rocm-jaxlib-v0.10.0`
- `rocm-jaxlib-v0.10.2`

The CI JAX matrix filters out the legacy/native `rocm-jaxlib-v0.9.1`
entry because the new CI workflow is manylinux-only.

Default CI currently generates:

- `rocm-jaxlib-v0.10.0`
- `Python 3.12`

This produces 1 JAX CI build cell.

## Testing

Ran local Python checks:

```bash
python -m compileall build_tools/github_actions/configure_multi_arch_ci.py \
  build_tools/github_actions/configure_jax_release_matrix.py
```  
 CI Test:  https://github.com/ROCm/TheRock/actions/runs/29396032944/job/87311830950?pr=6439
 
 ## Follow up

- Wire JAX wheel tests into Multi-Arch CI after the build lane is stable.
- https://github.com/ROCm/TheRock/pull/6439#discussion_r3581639396

 ## TODO
https://github.com/ROCm/TheRock/issues/6388


marbre (changes requested):
Please see compare with PyTorch where we limit to only build for Python 3.12. We shouldn't run the full Python matrix on every CI build.

erman-gurses:
> Please see compare with PyTorch where we limit to only build for Python 3.12. We shouldn't run the full Python matrix on every CI build.

Thanks  - that makes sense  - just updated the code

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
```diff
@@ -109,40 +112,142 @@ on:
...
+            Dockerfile.jax-manylinux_2_28-therock
+
+          docker build \
+            -t "${MANYLINUX_IMAGE_TAG}" \
+            --file=Dockerfile.jax-manylinux_2_28-therock \
+            --build-arg=THEROCK_INDEX_URL="${ROCM_PACKAGE_FIND_LINKS_URL}" \
+            --build-arg=THEROCK_VERSION="${THEROCK_VERSION}" \
+            --build-arg=GFX_ARCH="${GFX_ARCH}" \
+            --progress=plain \
+            .
```
marbre, line 197: As this workflow is intended to run on every CI run we should avoid building a docker container every time. This overhead might be acceptable as an intermediate solution for release builds but I think it is not for CI builds.

marbre (commented):
Is it worth to build `rocm-jaxlib-v0.10.0` and `rocm-jaxlib-v0.10.2`?

Furthermore, the CI description needs to be updated with regards to the Python versions:
```
Default CI currently generates:

* rocm-jaxlib-v0.10.0
* rocm-jaxlib-v0.10.2
* Python 3.11, 3.12, 3.13, and 3.14
```

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
erman-gurses: I see the point but I am not controlling that part - If you want I can keep only `rocm-jaxlib-v0.10.0` until other solution comes from JAX Team or should we wait?

marbre: I am aware you're not in control of it but I would rather delay this PR until it was fixed upstream. Would switching to 0.10.0 let us avoid re-building the Docker container every time? Normally I would prefer to stick with the latest patch version which in our case seems to be 0.10.2.

Anyway, let's wait for @ScottTodd  to weight in. I would rather push back to merge this as is.

erman-gurses: 0.10.0 is the most stable one - 0.10.2 has some issues from the upstream. If we chose one, I think 0.10.0 is the best candidate.
`Would switching to 0.10.0 let us avoid re-building the Docker container every time?` Just an idea, It will reduce the pressure on the runners

Sounds good to me to wait for Scott's input.

marbre: Well if it still requires Docker my push back remains. As 0.10.2 is a patch release I would also rather go with that version instead of 0.10.0. Let's wait for Scott to weight in.

erman-gurses: @ScottTodd , If I gate this for the `rocm-systems` and `rocm-libraries` bumps only, would it be ok? Since my main purpose is to detect how JAX is affected by those bump PRs.

ScottTodd: I'll take a more detailed look tomorrow (only got through ~50% of my PR review queue today). Some quick thoughts:

* At least this depends on `build_python_packages` and not tarball builds. If it depended on tarball builds that would bloat the storage size needed for each CI run.
* We should compare the job time between JAX and PyTorch and look for opportunities to optimize both. I see ~1h of build time for both with opportunities (ccache/sccache and other tricks) to get that down to ~20 minutes.
* As for what actually happens during the build, I see that JAX is still building LLVM from source (we already build a copy in ROCm, ideally we wouldn't need another copy) and I see that we're still building a dockerfile every run (which pulls in uncontrolled dependencies).
* If we build for releases already, we should at least have the ability to build on CI, even if only opt-in via a label or workflow_dispatch argument. I do like the idea of running these new jobs on at least bump PRs in TheRock, so we can catch if/when changes to rocm-systems and rocm-libraries break the JAX build (we've seen changes to RCCL break the PyTorch build).

So I think we can find a reasonable middle-ground for the CI system that gets us this much desired coverage, but there are still a few things to continue development on for the final shape that we'll all be happy with. Let's make sure that enough of that is tracked (and prioritized/assigned) so we don't let it slip for too long.

(will continue reviewing tomorrow)

erman-gurses: Thanks, I opened an issue to track follow ups https://github.com/ROCm/TheRock/issues/6388, I will modify and fill out as they are determined.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1065,7 +1068,13 @@ def _expand_build_config_for_platform(
     jax_build_matrix: list[dict[str, str]] = []
     build_jax = jobs.build_jax.action == JobAction.RUN and platform == "linux"
     if build_jax:
-        jax_build_matrix = generate_jax_matrix(ci_inputs.python_versions or None)
+        requested_python_versions = ci_inputs.python_versions or None
+
+        if ci_inputs.release_type == "ci":
+            jax_build_matrix = generate_jax_matrix_for_ci(requested_python_versions)
+        else:
+            jax_build_matrix = generate_jax_matrix(requested_python_versions)
```
ScottTodd: Let's match the style used for pytorch and add a `generate_jax_matrix_for_release_type()` function instead of having explicit functions for just `generate_jax_matrix_for_ci()` and `generate_jax_matrix()`, then pass `ci_inputs.release_type` as an argument. Then that function can also decide if it wants to generate a different matrix for dev/nightly/prerelease releases too (for PyTorch we'll want to include the "nightly" for nightly release runs but exclude it from prerelease runs).

See
* https://github.com/ROCm/TheRock/blob/59d3a3d9018b9605c976a9c6bb50c0a7e156c7c3/build_tools/github_actions/configure_pytorch_release_matrix.py#L29-L35
* https://github.com/ROCm/TheRock/blob/59d3a3d9018b9605c976a9c6bb50c0a7e156c7c3/build_tools/github_actions/configure_pytorch_release_matrix.py#L118-L136

build_tools/github_actions/configure_jax_release_matrix.py
```diff
@@ -65,6 +68,46 @@ def generate_jax_matrix(
...
+
+    matrix: list[dict[str, str]] = []
+    for cell in generate_jax_matrix(versions):
+        build_mode = str(cell["build_mode"])
+        jax_ref = str(cell["jax_ref"])
+
+        if build_mode != "manylinux":
+            continue
+        if jax_ref not in CI_JAX_REFS:
+            continue
```
ScottTodd: This dynamic matrix construction that first generates the release matrix then filters out cells is _okay_, but how about just generating the matrix for only the requested refs in the first place?

Could

1. Add `jax_refs` as an argument to `generate_jax_matrix(jax_refs, python_versions)`
2. Pass either `CI_JAX_REFS` or `RELEASE_JAX_REFS` to that function based on a `release_type` argument, have the function read the associated `jax_ref`, `jax_repository`, `build_mode`, and `gfx_arch` values from `JAX_REFS` (could rename to `JAX_REF_CONFIGS`

some rough pseudocode;

```python
JAX_REF_CONFIGS = [
    {
        "jax_ref": "rocm-jaxlib-v0.9.1",
        "jax_repository": "ROCm/rocm-jax",
        "build_mode": "native",
        "gfx_arch": "",
    },
    ...
]

generate_jax_matrix_for_release_type(jax_refs, python_versions, release_type):
  if release_type == "ci":
    return generate_jax_matrix(jax_refs, python_versions, CI_JAX_REFS)
  else:
    return generate_jax_matrix(jax_refs, python_versions, RELEASE_JAX_REFS) 
```

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
```diff
@@ -109,40 +112,142 @@ on:
...
+        run: |
+          pip install -r external-builds/jax/requirements-jax.txt
+
+      - name: Validate CI build mode
         run: |
           if [ "${{ inputs.build_mode }}" != "manylinux" ]; then
-            echo "Only build_mode=manylinux is supported by the planned JAX CI workflow."
+            echo "Only build_mode=manylinux is supported by this CI workflow."
             exit 1
           fi
```
ScottTodd, line 176: We could also add a unit test to https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/configure_jax_release_matrix_test.py that ensures that the CI matrix which is generated does not include unsupported build modes. That will catch errors quickly, while this check will wait until after all the other build jobs complete and the jax workflow finally runs.

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
```diff
@@ -109,40 +112,142 @@ on:
...
 
-      - name: Emit placeholder outputs
-        id: placeholder_outputs
+      - name: Set package dist dir
+        run: |
+          echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
```
ScottTodd: This doesn't need a workflow step to write to an environment variable dynamically right? Can't you just set `PACKAGE_DIST_DIR` as a `env:` variable on the step(s) that need it?

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
ScottTodd: Please add a TODO here with a link to an issue covering removing this dockerfile build. Our build containers should be frozen and updated infrequently, not build as part of every workflow run. This container in particular (https://github.com/ROCm/rocm-jax/blob/rocm-jax-infra/docker/manylinux/Dockerfile.jax-manylinux_2_28-therock) has multiple issues:
* It builds LLVM from source (when we already build and distribute LLVM as part of ROCm)
* It installs the AWS CLI from a domain we do not control
* It likely has poor interactions with caches (docker layer caches are a poor substitute for ccache)
* Dockerfiles setups like this are awkward for local development (including debugging, development, and reproducing what CI does)

We shouldn't need this at all, all we should need is to run in our existing https://github.com/ROCm/TheRock/blob/main/dockerfiles/build_manylinux_x86_64.Dockerfile image or a slightly modified image that also includes Bazel and any other build dependencies (for the entire job), install rocm into a venv, and then run the `python build/build.py` script.

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
```diff
@@ -109,40 +112,142 @@ on:
...
-            echo "jax_version="
-            echo "jaxlib_version="
-            echo "jax_plugin_version="
-            echo "jax_pjrt_version="
-          } >> "${GITHUB_OUTPUT}"
+          python build_tools/github_actions/upload_python_packages.py \
+            --input-packages-dir="${{ env.PACKAGE_DIST_DIR }}" \
+            --artifact-group="${{ inputs.artifact_group }}" \
+            --run-id="${{ github.run_id }}" \
+            --multiarch
```
ScottTodd, line 252: (Fine for a follow-up PR)

This script is fine to reuse here, but we should update the summary that it generates for both jax and torch: https://github.com/ROCm/TheRock/blob/59d3a3d9018b9605c976a9c6bb50c0a7e156c7c3/build_tools/github_actions/upload_python_packages.py#L176-L226

See on https://github.com/ROCm/TheRock/actions/runs/29052192290?pr=6439 how the pytorch and jax steps just include instructions for `pip install rocm` and mention kpack, when they should really say more like `pip install jax` and `pip install torch`

Image

ScottTodd (commented):
I'm okay with the extra time (~1h6m) that this will add to CI runs here in TheRock, but yes let's follow up later on some of the specifics to optimize both how long that takes and when it triggers (e.g. exclude jax builds for PRs that only modify pytorch code)

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
erman-gurses: Will address it at the follow-up PR -  noted in the PR description.

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
erman-gurses: Removed

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
erman-gurses: Added additional unit tests

build_tools/github_actions/configure_jax_release_matrix.py
erman-gurses: Refactored

build_tools/github_actions/configure_multi_arch_ci.py
erman-gurses: Refactored

ScottTodd (approved):

erman-gurses:
@marbre, please let me know if the TODO https://github.com/ROCm/TheRock/issues/6388, OK for you.

marbre (approved):
> @marbre, please let me know if the TODO #6388, OK for you.

Fine for me as it has an assignee and Scott is okay addressing in a follow up.

.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
```diff
@@ -109,40 +112,141 @@ on:
...
 permissions:
   contents: read
+  id-token: write
 
 jobs:
   build_jax_wheels:
-    name: Placeholder | py ${{ inputs.python_version }} | jax ${{ inputs.jax_ref }}
-    runs-on: ubuntu-24.04
+    name: Build JAX | ${{ inputs.artifact_group }} | jax ${{ inputs.jax_ref }} | py ${{ inputs.python_version }}
+    runs-on: ${{ github.repository_owner == 'ROCm' && 'azure-linux-scale-rocm' || 'ubuntu-24.04' }}
```
ScottTodd, line 123: Ah, missed during review.

Can we switch from `azure-linux-scale-rocm` to `aws-linux-scale-rocm-prod`? Similar to https://github.com/ROCm/TheRock/pull/6037

(uploading to S3 failed on this run on PR from a fork: https://github.com/ROCm/TheRock/actions/runs/29513125898/job/87707018784?pr=6608, and the credentials are different on Azure and AWS runners)

erman-gurses: Sure, will raise a quick PR for that.

erman-gurses: Ohh, I remember that last time `aws` was not supporting Docker. That was the reason we kept `azure-linux-scale-rocm`.  Is `aws-linux-scale-rocm-prod` supporting Docker now?
https://github.com/ROCm/TheRock/pull/6052

PR #6451, opened 2026-07-09
feat(ci): Add --projects flag for selective builds

geomin12:
## Summary

Add the ability to build specific projects without needing to build an entire stage. This enables external repositories (like rocm-systems) to trigger builds that only compile the components they changed.

**This is Part 1 of the selective build feature.** Part 2 (workflow integration) will follow in a separate PR after this lands.

### New Features

- **`--projects` flag**: Enable specific projects for building
  ```bash
  # Build only rocBLAS and MIOpen
  python build_tools/configure_stage.py --projects rocblas miopen --oneline
  # Output: -DTHEROCK_ENABLE_ALL=OFF -DTHEROCK_ENABLE_BLAS=ON -DTHEROCK_ENABLE_MIOPEN=ON
  ```

- **`--list-projects` flag**: Show available projects and their artifact mappings
  ```bash
  python build_tools/configure_stage.py --list-projects
  ```

- **`--skip-stages` flag**: Compute which stages can be skipped based on projects
  ```bash
  python build_tools/configure_stage.py --projects hip --skip-stages
  # Output: dctools-core,debug-tools,math-libs,media-libs,...
  ```

- **rocm-systems support**: Handle paths like `projects/hip` from external repos via `normalize_project_name()` and `rocm_systems_projects.json`

### Technical Details

- `artifact_subprojects.json` manifest auto-generated by CMake maps artifacts to subprojects
- `build_topology.py` extended with project resolution logic (`resolve_projects_to_features`, `get_stages_for_projects`)
- `test_artifacts` field added to `core-runtime` and `core-hip` in BUILD_TOPOLOGY.toml for proper test inclusion
- hipsparselt moved from blas to sparse `split_databases` (fixes #5449)

## Test plan

- [x] All 34 unit tests pass (`pytest build_tools/tests/configure_stage_test.py`)
- [x] `--list-projects` shows correct mappings
- [x] `--projects hip` resolves correctly to `THEROCK_ENABLE_HIP_RUNTIME`
- [x] `--projects projects/hip` (path format) normalizes correctly
- [x] `--skip-stages` computes correct stages
- [ ] CI workflows pass

## Next steps
 - .github/workflows/* changes to pass projects input to stages
  - Stage-specific project routing (changes go to correct stage)


build_tools/_therock_utils/build_topology.py
```diff
@@ -1098,3 +1103,163 @@ def get_python_requires_for_stage(self, build_stage: str) -> List[str]:
...
+            if build_manifest.exists():
+                manifest = self.load_subproject_manifest(build_manifest)
+        if manifest is None:
+            manifest = self.load_subproject_manifest()
+
+        for artifact in self.artifacts.values():
+            alias_map[artifact.name.lower()] = artifact.name
+
+            if manifest and artifact.name in manifest:
+                for alias in manifest[artifact.name]:
```
jayhawk-commits, line 1201: This uses the artifact packaging manifest as the project-to-feature source, but those mappings do not always match the CMake feature gates. For example, `artifact_subprojects.json` puts `hipSPARSE`, `hipSOLVER`, and `rocSOLVER` under `blas`, so `--projects hipsparse/hipsolver/rocsolver` emits only `-DTHEROCK_ENABLE_BLAS=ON`. Those projects are gated by `THEROCK_ENABLE_SPARSE` / `THEROCK_ENABLE_SOLVER`, so the selected project would not be built. Can we make the project map feature-oriented, or add overrides/tests for these cases instead of deriving feature selection directly from artifact packaging slices?

rocm_systems_projects.json
```diff
@@ -0,0 +1,25 @@
...
+  "rccl-tests": "rccl",
+  "rdc": "rdc",
+  "rocdbgapi": "amd-dbgapi",
+  "rocdecode": "rocdecode",
+  "rocjpeg": "rocjpeg",
+  "rocm-core": "base",
+  "rocm-smi-lib": "base",
+  "rocminfo": "core-runtime",
+  "rocprofiler": "rocprofiler-sdk",
+  "rocprofiler-compute": "rocprofiler-systems",
```
jayhawk-commits: This override points `rocprofiler-compute` at `rocprofiler-systems`, even though the generated artifact/subproject manifest already maps the `rocprofiler-compute` subproject to the `rocprofiler-compute` artifact. Because these overrides are applied after the manifest, `--projects rocprofiler-compute --platform linux` currently emits `-DTHEROCK_ENABLE_ROCPROFSYS=ON` instead of the rocprofiler-compute feature. Can this entry map to `rocprofiler-compute` or avoid overriding an existing manifest alias?

jayhawk-commits (dismissed):

build_tools/generate_subproject_manifest.py
```diff
@@ -0,0 +1,67 @@
...
+                old_manifest = json.load(f)
+            if new_manifest == old_manifest:
+                print("Manifest is up-to-date.")
+            else:
+                print(
+                    "Manifest is stale. Run: python build_tools/generate_subproject_manifest.py"
+                )
+                sys.exit(1)
+        else:
+            shutil.copy(generated, MANIFEST_PATH)
```
erman-gurses, line 62: Since `artifact_subprojects.json` is included as a generated, tracked file in this PR, should CI run python `build_tools/generate_subproject_manifest.py --verify` to ensure it stays synchronized with the CMake-generated manifest?

build_tools/tests/configure_stage_test.py
```diff
@@ -0,0 +1,312 @@
...
+        cls.topology = get_topology()
+
+    def test_alias_map_includes_artifact_names(self):
+        """Test that alias map includes artifact names."""
+        alias_map = self.topology.get_alias_to_artifact_map()
+        self.assertIn("blas", alias_map)
+        self.assertIn("fft", alias_map)
+        self.assertIn("miopen", alias_map)
+
+    def test_alias_map_includes_split_databases(self):
```
erman-gurses: We have good coverage for alias resolution, but I don't see a test that validates the generated artifact_subprojects.json matches what CMake emits. Since this manifest is now a source of truth for project resolution, should we add a test (or CI verification) to catch accidental drift?

rocm_systems_projects.json
erman-gurses: This appears to overwrite the canonical `rocprofiler-compute -> rocprofiler-compute` mapping created from `BUILD_TOPOLOGY.toml`. Because the ROCm Systems mappings are applied unconditionally afterward, --projects rocprofiler-compute will resolve to the rocprofiler-systems artifact rather than the dedicated rocprofiler-compute artifact. should this map to "rocprofiler-compute`, or should the alias-map merge avoid overwriting canonical artifact names?

erman-gurses (commented):
Added some comments below:

build_tools/generate_subproject_manifest.py
erman-gurses: @geomin12  - I think we should add this verification before merging. Since `artifact_subprojects.json` is generated from CMake but checked into the repository and consumed by project-selection logic, the existing unit tests can still pass against a stale file. Could we run python `build_tools/generate_subproject_manifest.py --verify` in CI or pre-commit so changes to the CMake definitions cannot leave the tracked manifest out of sync?

build_tools/tests/configure_stage_test.py
```diff
@@ -0,0 +1,406 @@
...
+        valid_features = {
+            topology.get_artifact_feature_name(a) for a in topology.artifacts.values()
+        }
+
+        for subproject, feature in features.items():
+            self.assertIn(
+                feature,
+                valid_features,
+                f"Invalid feature '{feature}' for subproject '{subproject}'",
+            )
```
erman-gurses: Thanks for adding the manifest validation. However, this test validates `subproject_features.json;` it does not verify that the checked-in `artifact_subprojects.json` matches what CMake generates.

Could we also run `python build_tools/generate_subproject_manifest.py --verify` in CI or pre-commit? That would address the original drift concern by regenerating the CMake manifest and comparing it with the tracked file.

erman-gurses (approved):
LGTM -  Thanks!

jayhawk-commits (approved):
Previous concerns with the sparse and solver projects are addressed. LGTM

project_mappings.json
ScottTodd, line 1: Can this be moved into a subfolder?

(I push back on most file additions to the repository root)

artifact_subprojects.json
ScottTodd, line 1: Can this be moved into a subfolder?

(I push back on most file additions to the repository root)

This also looks like it could be generated from `BUILD_TOPOLOGY` and the artifact manifests. I don't think it should be checked in like this if so.

PR #6452, opened 2026-07-09
fix: [pytorch] Resolve ROCm deps via RUNPATHs

venkat1361:
## Motivation

Commit https://github.com/ROCm/TheRock/commit/6a55012d931359f125b540ea6f609c44629bc60c added rocprofiler-sdk to PyTorch’s Linux startup preload list to fix nightly smoke tests after PyTorch/Kineto gained a runtime dependency on `librocprofiler-sdk.so.1`.

When PyTorch is launched under an external `rocprofv3`, the profiler may already have loaded and initialized one rocprofiler-sdk provider. PyTorch startup can then explicitly load the wheel-bundled rocprofiler-sdk, creating a duplicate-provider path in the same process. In that scenario, the rocprofv3 tool configuration path can be invoked after rocprofiler-sdk has already closed its configuration window, leading to:
```
ROCPROFILER_STATUS_ERROR_CONFIGURATION_LOCKED
Configuration request occurred outside of valid rocprofiler configuration period
```
This PR treats the original issue as a packaging/runtime dependency issue instead of an import-time preload issue. PyTorch’s ROCm-linked shared libraries should resolve their `DT_NEEDED` dependencies through relocatable wheel RUNPATHs.

JIRA ID: AIPROFSDK-969
## Technical Details

This PR removes rocprofiler-sdk from PyTorch’s generated _rocm_init.py preload list.

After pytorch build step unpacks the generated torch wheel, finds ROCm-linked torch shared objects, removes non-relocatable absolute RUNPATH entries, appends $ORIGIN-relative RUNPATH entries pointing to the installed ROCm Python package library directories, and repacks the wheel before copying.

This lets dependencies such as `librocprofiler-sdk.so.1` resolve through the dynamic loader instead of through explicit `ctypes/rocm_sdk.initialize_process()` preloading. If an external `rocprofv3` has already loaded a compatible SDK provider with the same SONAME, the loader can use the already-loaded provider instead of PyTorch forcing another absolute SDK load.

This PR also updates package promotion logic so torch ELF RUNPATHs stay valid when ROCm package versions are promoted. Text files such as `_rocm_init.py`, `version.py`, and metadata were already updated during promotion, but RUNPATH strings embedded in ELF files also need their ROCm version fragments updated when promoting from nightly/RC versions to release versions.

## Test Plan

- Added test in `build_tools/packaging/tests/promote_keep_list_test.py`

## Test Result

- https://github.com/ROCm/TheRock/actions/runs/29043018157
## Submission Checklist

### DIFF
--- a/build_tools/packaging/promote_packages.py
+++ b/build_tools/packaging/promote_packages.py
@@ -75,6 +75,7 @@ import importlib.util
 import pathlib
 import re
 import shutil
+import subprocess
 import sys
 import tarfile
 import tempfile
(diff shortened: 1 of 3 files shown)


geomin12 (approved):
lgtm as it is passing

ScottTodd (commented):
Woah woah woah wait please. This needs my review or someone else familiar with ROCm and PyTorch packaging. CI is not testing this.

ScottTodd:
Reverted: https://github.com/ROCm/TheRock/pull/6623. Please file a github issue explaining the problem you want to solve. This (RUNPATH wheel mutation) is not a viable approach as implemented.

zjliu-amd:
I was looking for solutions other than adding env vars for PR https://github.com/ROCm/TheRock/pull/5866 and found this, I can provide a justification from pytorch test's standpoint.
The main problem comes from the libraries' location. In the upstream, all rocm libraries are bundled in pytorch's wheel, inside `torch/lib`. But in the rock build, rocm dependencies are installed as individual packages, and by default they are not added to any system configurations, env vars, nor the runpath in binaries is set, causing a lot of pytorch tests wouldn't be able to find them.

ScottTodd:
> But in the rock build, rocm dependencies are installed as individual packages, and by default they are not added to any system configurations, env vars, nor the runpath in binaries is set, causing a lot of pytorch tests wouldn't be able to find them.

Documentation for library init is at https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#using-packages-from-frameworks. The details matter a lot here, and "a lot of pytorch tests fail" is not specific enough, we need to look case by case to see what packages are installed, what code paths are used, what happens, and what should happen instead.

venkat1361:
Hi @ScottTodd filed issue https://github.com/ROCm/TheRock/issues/6695

Issue #6454, opened 2026-07-09
[Issue]:  RPM packaging extremely slow for ASAN artifacts


raramakr:
### Problem Description

ASAN RPM package builds are significantly slower than DEB builds for the same artifacts.

https://github.com/ROCm/rockrel/actions/runs/28908556503/
Build DEB Packages : 1h 20m  
Build RPM Packages : 4h 57m 

Root Cause

  RPM uses gzip level 9 compression by default for package payloads. For large ASAN artifacts (~175GB with debug symbols), gzip-9 is extremely slow due to:
  - Single-threaded compression algorithm
  - High CPU overhead at maximum compression level

Proposed Solution
  
  Add %define _binary_payload w1.zstdio to the RPM spec template to use zstd level 1 compression.

  Benefits:
  - ~4x faster builds
  - ~1% smaller packages (zstd achieves slightly better compression)
  - Native multi-threading support

### Operating System

Ubuntu 24

### CPU

NA

### GPU

NA

### ROCm Version

ROCm 7.14

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

<details>
<summary>rocminfo --support output</summary>

```
Paste output here
```

</details>

raramakr:
Now the rpm package build time is around 45minutes with zstd compression

PR #6455, opened 2026-07-09
fix: Use zstd compression for faster RPM builds

raramakr:
Switch from gzip-9 to zstd-9 for RPM payload compression and better build time

 ```
Used **amdrocm-llvm** ASAN package for comparison
 ------------------------------------------------------------------------------------------
  Compression     Package Size    Build Time   Size vs Baseline       Time vs Baseline
  ------------------------------------------------------------------------------------------
  GZIP-9          7.87 GiB                  Around1 hr         —                      —
  ZSTD-1          8.22 GiB                  3m 24s       +360 MiB (+4.5%)       57 min faster
  ZSTD-6         7.14 GiB                   6m 44s       -745 MiB (-9.3%)         53 min faster
  ZSTD-9          6.82 GiB                  8m 9s        -1071 MiB (-13.3%)       52 min faster
  ------------------------------------------------------------------------------------------
```

JIRA ID: https://github.com/ROCm/TheRock/issues/6454

### DIFF
--- a/build_tools/packaging/linux/template/rpm_specfile.j2
+++ b/build_tools/packaging/linux/template/rpm_specfile.j2
@@ -16,6 +16,7 @@
 {%- if exclude_libpython_requires %}
 %define __requires_exclude libpython*.*
 {%- endif %}
+%define _binary_payload w1.zstdio
 
 Name: {{ pkg_name }}
 Version: {{ version }}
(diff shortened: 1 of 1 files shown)


nunnikri (approved):
Changes looks fine. Please wait for ci to complete the build. 
Also we might have to do a quick install test in all the supported OS with rpm flavor here. https://rocm.docs.amd.com/en/latest/compatibility/compatibility-matrix.html

Issue #3939, opened 2026-03-12
[Feature]: Run rocm[devel]-dependent PyTorch tests


darren-amd:
### Suggestion Description

Follows up on #2152. Waiting until current tests are stabilized before doing this split.

Several PyTorch tests compile C++ extensions at runtime (e.g. `dummy_allocator`) which require development headers like `hipblas/hipblas.h` from `rocm[devel]`. These tests are currently skipped because only `rocm[libraries]` is installed at test time.

Installing `rocm[devel]` for the entire test run is undesirable because it would mask regressions/issues in the torch runtime package configuration that end users have.

## Proposed Plan

Split the test [workflow](https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_pytorch_wheels.yml) into two phases:

1. **Phase 1 - Core tests (without `rocm[devel]`):**
   Run the full test suite as-is with only `torch` installed (the current behavior). Tests that require devel packages remain skipped. This ensures we catch regressions in the runtime configuration.

2. **Phase 2 — Devel tests (with `rocm[devel]`):**
   Install `rocm[devel]` (version derived from the installed `rocm` package via `importlib.metadata`), set `ROCM_PATH` via `hipconfig --rocmpath`, and run only the tests that require `rocm[devel]`. This would re-enable the [currently skipped tests](https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/skip_tests/generic.py) that require these packages.

amd-maradosa:
Hi @darren-amd, test_mempool_empty_cache_inactive (one of the tests that compile C++ extensions at runtime) is currently failing in the release pipeline ([link](https://github.com/ROCm/rockrel/actions/runs/28213360075/job/83589518214)) rather than being skipped. Is the "waiting until stabilized" condition met now, or is there still a blocker on the two-phase split?

darren-amd:
Hi @amd-maradosa,

There are still quite a few torch test failures that are being worked through but I'll bring this up during the triage call. For the linked failure though, it looks like a bug where the "cuda" skip list is being written over: https://github.com/ROCm/TheRock/blob/da03bd9956d80ca18e3170608020d7c31c74b56f/external-builds/pytorch/skip_tests/generic.py#L12

amd-maradosa:
[...]

PR #6464, opened 2026-07-10
fix(pytorch-tests): merge duplicate common.cuda skip block in generic.py

amd-maradosa:
## Summary

The `"common"` dict in `external-builds/pytorch/skip_tests/generic.py` defined the `"cuda"` key **twice**. Python keeps only the last literal for a duplicate key, so the first `cuda` block was silently discarded — dropping three tests from the skip list:

- `test_mempool_empty_cache_inactive`
- `test_memory_snapshot`
- `test_memory_plots`

Because these were dropped from `generic.py`, they only stayed skipped on torch versions whose `pytorch_<version>.py` happened to re-list them. `release/2.12` does **not** re-list `test_mempool_empty_cache_inactive`, so it ran and failed with `RuntimeError: Error building extension 'dummy_allocator'` (the `hipblas.h`-at-runtime error that requires `rocm[devel]`, which isn't installed at test time).

Observed failure: rockrel full-wheels run, cell py3.10 / torch `release/2.12` / gfx1151.

The duplicate `cuda` key was introduced in #3929 (commit `71a0072`), which added a new `common.cuda` block above the pre-existing one instead of merging into it.

## Fix

Fold the three tests into the single surviving `"cuda"` block so they are skipped across all versions from one place, and remove the duplicate key.

Related: #3939 and https://github.com/ROCm/TheRock/issues/6173

## Test plan

- [x] `python external-builds/pytorch/skip_tests/test_skip_tests.py` → "All skip-test sanity checks passed."
- [x] AST check confirms no duplicate dict keys remain and all three tests are present in `common.cuda`.

🤖 Generated with Claude Code

### DIFF
--- a/external-builds/pytorch/skip_tests/generic.py
+++ b/external-builds/pytorch/skip_tests/generic.py
@@ -9,6 +9,12 @@ skip_tests = {
         }
     },
     "common": {
+        "autograd": [
+            # Stream comparison mismatch on ROCm (non-default stream vs default stream)
+            #   AssertionError: <torch.cuda.Stream ...> != <torch.cuda.Stream cuda_stream=0x0>
+            # Seems to fails on Linux and Windows across torch versions and all tested GPUs.
+            "test_side_stream_backward_overlap",
+        ],
         "cuda": [
             # RuntimeError: Error building extension 'dummy_allocator'
             # Skipped across all PyTorch versions; the hipblas.h include error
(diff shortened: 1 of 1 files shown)


amd-maradosa:
CI doesn't run the `test_mempool_empty_cache_inactive` test: [CI link](https://github.com/ROCm/TheRock/actions/runs/29092518448/job/86361034991) which is as expected

amd-maradosa:
Hi @ScottTodd, hope you are the right person for review. This wasn't a major change and I don't think extra test files are necessary so I'm not sure how to remove this missing unit test problem the bot is reporting.

darren-amd (approved):
Thanks, this looks good to me

PR #5901, opened 2026-06-17
[CI] Enable sccache for multi-arch release builds (Linux + Windows)

amd-chiranjeevi:
## Motivation

Enables sccache for the multi-arch release build pipeline (TheRock core — not PyTorch),

  workflow call chain:

  multi_arch_release.yml
    → multi_arch_release_{linux,windows}.yml
      → multi_arch_build_{portable_linux,windows}.yml
        → multi_arch_build_{portable_linux,windows}_artifacts.yml

## What changed

  | | Value |
  |---|---|
  | Bucket | `therock-sccache-{dev,nightly,prerelease}` |
  | Region | `us-east-1` |
  | Account | - |
  | IAM role | - |
  | Key prefix (Linux) | `linux/multiarch/v1/` |
  | Key prefix (Windows) | `windows/multiarch/v1/` |
 | `cache_type` default | `sccache` |
 
 

## Test Plan

**to be updated**

## Test Result

**to be updated**

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5471, opened 2026-05-27
Migrate PyTorch sccache to AWS prod account (324352301041, us-east-1)

subodh-dubey-amd:
> [!CAUTION]
> Depends on  [ROCm/TheRock-Infra#240](https://github.com/ROCm/TheRock-Infra/pull/240)

## Summary

The sccache buckets in `692859939525` (us-east-2) wired up in #3532 are replaced with buckes in `324352301041` (us-east-1) per [ROCm/TheRock-Infra#240](https://github.com/ROCm/TheRock-Infra/pull/240). Re-point the 6 PyTorch wheel workflows and flip `cache_type` default from `none` to `sccache`.

| | Old | New |
|---|---|---|
| Bucket (release) | `therock-${release_type}-pytorch-sccache` | `therock-pytorch-sccache-${release_type}` |
| Bucket (CI) | `therock-ci-pytorch-sccache` | `therock-pytorch-sccache-ci` |
| Region | `us-east-2` | `us-east-1` |
| Account | `692859939525` | `324352301041` |

Role names (`therock-{ci,dev,nightly,prerelease}`) and OIDC trust (`ROCm/TheRock` on dev/nightly/prerelease) are preserved.

## Files (7)

- 6 workflows: `build_{portable_linux,windows}_pytorch_wheels{,_ci}.yml`, `multi_arch_release_{linux,windows}_pytorch_wheels.yml` — env block + sccache AWS-credentials step + `cache_type` default (both `workflow_call` and `workflow_dispatch`).
- `docs/development/s3_buckets.md` — updated bucket names + note about the account/region split.

## Validation - measurable speedup against main baseline

| Branch | `cache_type` | Bucket state | Run | `Build PyTorch wheels` |
| --- | --- | --- | --- | --- |
| `main` | `none` (current default) | n/a | [26583018702](https://github.com/ROCm/TheRock/actions/runs/26583018702) (May 28) | **58m 58s** |
| this PR | `sccache` | bucket populated by prior run | [26632339907](https://github.com/ROCm/TheRock/actions/runs/26632339907) (May 29, commit `a67cd81`) | **14m 00s** |

## Validation

- [x] All resource names cross-checked against the infra PR.
- [x] After infra ships   [ROCm/TheRock-Infra#240](https://github.com/ROCm/TheRock-Infra/pull/240): `workflow_dispatch` `build_portable_linux_pytorch_wheels_ci.yml` with `cache_type=sccache`; verify credentials step succeeds.

[...]

Issue #748, opened 2025-05-30
[Issue]: Linker Error: Duplicate symbol `__hip_cuid_...` in `rocSOLVER` when building with `ccache` on Windows


7shi:
### Problem Description

I encountered `duplicate symbol` linker errors when building `rocSOLVER` on Windows with `ccache` enabled.

```text
[155/176] Building sub-project rocSOLVER (in background)
FAILED: math-libs/BLAS/rocSOLVER/stamp/build.stamp E:/work/TheRock/build/math-libs/BLAS/rocSOLVER/stamp/build.stamp
C:\WINDOWS\system32\cmd.exe /C "cd /D E:\work\TheRock\build\math-libs\BLAS\rocSOLVER\build && E:\work\TheRock\.venv\Scripts\python.exe E:/work/TheRock/build_tools/teatime.py --log-timestamps --label rocSOLVER --interactive E:/work/TheRock/build/logs/rocSOLVER_build.log -- E:/work/TheRock/.venv/Lib/site-packages/cmake/data/bin/cmake.exe -E env --unset=ROCM_PATH --unset=ROCM_DIR --unset=HIP_PATH --unset=HIP_DIR -- E:/work/TheRock/.venv/Lib/site-packages/cmake/data/bin/cmake.exe --build E:/work/TheRock/build/math-libs/BLAS/rocSOLVER/build && E:\work\TheRock\.venv\Lib\site-packages\cmake\data\bin\cmake.exe -E touch E:/work/TheRock/build/math-libs/BLAS/rocSOLVER/stamp/build.stamp"
[rocSOLVER] [1/247] Building CXX object library/src/CMakeFiles/rocsolver.dir/common/rocsparse.cpp.obj
...(snip)...
[rocSOLVER] [247/247] Linking CXX shared library staging\rocsolver.dll
[rocSOLVER] FAILED: staging/rocsolver.dll library/src/rocsolver.lib
[...]

PR #6495, opened 2026-07-12
ci: enable sccache for multi-arch Linux builds (per-TU -cuid)

subodh-dubey-amd:
## Motivation
Enable sccache for the multi-arch **Linux** release build pipeline (TheRock core, not PyTorch) — the same pattern used for PyTorch wheels in #5471, consolidating the Linux enablement from #5901 (@amd-chiranjeevi).

Turning caching on for HIP builds surfaced two correctness blockers that left math-libs either **uncacheable** or **mis-linked**; this PR fixes them so the cache actually works. Related to #748 (the `__hip_cuid_` duplicate-symbol class) — #748 itself is the Windows + ccache + rocSOLVER case (fixed separately), so this does not close it.

## What changed
**Enablement** — a `cache_type` input (default `sccache`) threaded through the Linux call chain:
```
multi_arch_release.yml
  → multi_arch_release_linux.yml
    → multi_arch_build_portable_linux.yml
      → multi_arch_build_portable_linux_artifacts.yml
```

| | Value |
|---|---|
| Bucket | `therock-sccache-<release_type>` — i.e. `therock-sccache-{ci,dev,nightly,prerelease}` |
| Region | `us-east-1` |
| AWS account | `324352301041` |
| Key prefix | `linux/multiarch/v2/` |
| `cache_type` default | `sccache` |

Also sets `-DCMAKE_UNITY_BUILD=OFF -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON` (unity/PCH bundle multiple sources into one invocation, which sccache can't cache).

**Correctness fixes** — what makes HIP RDC compiles actually cacheable:
- joined `-resource-dir=<dir>` — otherwise every compile is rejected as "multiple input files"
- `cuid_launcher.py` — deterministic per-TU `-cuid` so cached RDC objects don't collide on `__hip_cuid_` (distinct key per TU; warm hits preserved)
- exclude launcher-incompatible subprojects (`rocprofiler-systems-examples`, `hip-tests`) from the CMake launcher
- fail-loud (#6504) — an unusable sccache emits a GHA `::warning::` and builds uncached, instead of silently degrading
- Tensile per-arch + rocWMMA `-Xclang` patches (from #5901; + a per-arch unit-test fix)

## Validation — cold → warm (`gfx94X-dcgpu`, sccache)
Back-to-back runs on this branch into a fresh `v2` namespace: [cold 29201944080](https://github.com/ROCm/TheRock/actions/runs/29201944080) · [warm 29205432183](https://github.com/ROCm/TheRock/actions/runs/29205432183).

**Cache hit-rate (math-libs stage):**

| | Cold | Warm |
|---|---|---|
| rocWMMA build | ✅ | ✅ |
| `__hip_cuid_` duplicate-symbol errors | 0 | 0 |
| HIP cache hit-rate | 0.00% (0 hit / 4212 miss) | **99.12%** (4175 hit / 37 miss) |
| C/C++ cache hit-rate | 25.27% | 100.00% |
| Non-cacheable compilations | 0 | 0 |

**Build time saved (stage wall-clock):**

| Stage | Cold | Warm | Saved | Warm hit-rate |
|---|---|---|---|---|
| compiler-runtime | 21m 06s | 11m 04s | ~10m (48%) | 99.97% |
| math-libs (gfx94X) | 84m 24s | 61m 08s | ~23m (28%) | 99.12% HIP / 100% C/C++ |
| **critical path (both)** | **~105m** | **~72m** | **~33m (~31%)** | |

Cold is 0% HIP by design (the `-cuid` makes every HIP key unique, so `v2` starts empty and repopulates); the warm run reuses it at **99%+**, saving **~33 min per family, per build**. Residual warm time is uncached work: configure, RDC device linking, staging/dist, packaging.

## Scope / notes
- **Linux only.** Windows sccache is deferred — unrelated `sccache: failed to persist temporary file` from `composable_kernel`'s ~263-char object paths; Windows keeps its current cache path.
- Key prefix `v1`→`v2`: the per-TU `-cuid` changes every HIP key, so this is a fresh cache generation.
- The 4 rocm-libraries patches were verified against the current submodule pin.

## Submission checklist
- [x] Reviewed the contributing guidelines.


patches/amd-mainline/rocm-libraries/0001-rocWMMA-Remove-dead-Xclang-fallow-half-arguments-and.patch
ScottTodd, line 1: No new patches to the rocm-libraries submodule except for extreme circumstances, which this PR is not. These changes must be sent to https://github.com/ROCm/rocm-libraries and come through normal submodule updates.

See https://github.com/ROCm/TheRock/blob/main/patches/README.md

.github/workflows/multi_arch_release.yml
```diff
@@ -97,6 +101,14 @@ on:
...
         type: string
         default: ""
+      cache_type:
+        description: "Compiler cache type (sccache, ccache, or none)"
+        type: choice
+        options:
+          - sccache
+          - ccache
+          - none
+        default: sccache
```
ScottTodd, line 111: Please limit the use of new workflow inputs for experiments like this. We still need to complete the experiments and remove the inputs from https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_linux_pytorch_wheels.yml. Each workflow input adds complexity, especially for workflows which span multiple repositories (releases are here and in https://github.com/ROCm/rockrel, CI is here and in rocm-libraries, rocm-systems, etc.).

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -154,12 +170,35 @@ jobs:
...
+            # (ROCm/TheRock#748). The wrapper finds sccache via THEROCK_SCCACHE
+            # (written to $GITHUB_ENV by setup_sccache_rocm.py --gha).
+            CUID_LAUNCHER="${PWD}/build_tools/cuid_launcher.py"
+            COMPILER_LAUNCHER_ARGS="-DCMAKE_C_COMPILER_LAUNCHER=${CUID_LAUNCHER} -DCMAKE_CXX_COMPILER_LAUNCHER=${CUID_LAUNCHER}"
+          elif [ "${{ inputs.cache_type }}" = "ccache" ]; then
+            COMPILER_LAUNCHER_ARGS="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
+          fi
+          # Disable unity/jumbo builds so sccache can cache individual compilations.
+          # Unity builds pass multiple source files per compiler invocation, which
+          # sccache cannot cache (reported as "Non-cacheable: multiple input files").
```
ScottTodd, line 196: This code is too complex for a workflow file. Please follow https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md

* Move code into scripts that developers can also use locally. This workflow is already running `python build_tools/setup_sccache_rocm.py --gha`, so why is this newly setting additional environment variables 60 lines down in the workflow? Can set them all in the script by writing to `GITHUB_ENV`.
* Design for the local experience first, then update workflow files. See how https://github.com/ROCm/TheRock/blob/main/build_tools/setup_ccache.py is designed to set some environment variables on its own.
* Limit branching. Let's choose one cache implementation and stick to that. Run experiments on branches, keep the code on `main` focused.

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -154,12 +170,35 @@ jobs:
         if: ${{ steps.stage_config.outputs.pip_install_cmd }}
         run: pip install ${{ steps.stage_config.outputs.pip_install_cmd }}
 
+      - name: Configure AWS Credentials for sccache
+        if: ${{ inputs.cache_type == 'sccache' && github.repository_owner == 'ROCm' }}
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # v6.2.0
```
ScottTodd, line 175: We need a plan for the cache to work on PRs from forks. Not going to invest in anything that doesn't include that.

The standard way to do that is for unauthenticated uses to have readonly access and authenticated uses to have read+write access.

build_tools/cuid_launcher.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Compiler launcher that injects a stable, per-translation-unit ``-cuid`` before
+handing off to sccache.
+
+Why this exists (ROCm/TheRock#748)
```
ScottTodd, line 8: I don't necessarily think a custom compiler launcher is a good long term fix for https://github.com/ROCm/TheRock/issues/748 . I'd rather fix the project source code and/or ccache/sccache. There are other problems we'd like to use compiler launches to solve (e.g. build job server to load balance jobs across cmake subproject builds), so we'd need both launches to compose together. Better to just not need a custom launcher at all.

build_tools/setup_sccache_rocm.py
```diff
@@ -99,6 +100,52 @@ def sccache_build_env(sccache_path: Path, hip_launcher: bool = True) -> dict[str
...
+    """
+
+    def _unusable(message: str) -> None:
+        if gha:
+            print(
+                f"::warning::sccache unusable: {message}. "
+                "Building WITHOUT compiler cache."
+            )
+            return None
+        raise RuntimeError(message)
```
ScottTodd, line 126: > * fail-loud ([ci: warn (not silently skip) when sccache is unusable #6504](https://github.com/ROCm/TheRock/pull/6504)) — an unusable sccache emits a GHA `::warning::` and builds uncached, instead of silently degrading

Huh? A warning log is *not* "loud". If the github actions workflows, which we fully control, do not configure a usable cache, that should be a hard error that fails the job. No degrading to building with no cache.

What _could_ degrade more gracefully on CI would be transient errors like if the remote cache server itself is offline or otherwise inaccessible. Not being able to find the `sccache` binary or set its configuration options is an irrecoverable error.

ScottTodd (changes requested):
This PR is too big to review. Need to land sccache support patches in relevant subprojects and then step this forward one smaller PR at a time.

PR #6505, opened 2026-07-13
Revert "fix(ci): correct multi-arch index URL in PyTorch test and upload summaries"

amd-chiranjeevi:
Reverts ROCm/TheRock#6086


marbre (changes requested):
It is unclear what this revert should fix. If it is intended to fix the math-libs related failures, this is not the correct fix as this is unrelated.

marbre:
As discussed in our standup, this is unrelated.

PR #6517, opened 2026-07-13
fix(emulation): package rocjitsu hooks and CLI test payload

newling:
## Motivation

The rocjitsu build installs its runtime library, instrumentation hook, CLI, configs, and schemas, but the artifact descriptor did not package all of that test tooling.

This change makes the installed rocjitsu CLI available to CI and development workflows that fetch rocjitsu test artifacts.

## Summary

- include `librocjitsu_hooks.so` alongside `librocjitsu.so` in the `lib` artifact
- validate `librocjitsu_hooks.so` as an installed shared library
- add a `test` component containing `bin/rocjitsu`
- keep configs and schemas in the `run` artifact through `share/rocjitsu/**`

The CLI is available when rocjitsu test artifacts are fetched, for example with:

```shell
install_rocm_from_artifacts.py --rocjitsu --tests
```

## Local validation

- Ran the artifact descriptor overlap unit test: `python3 -m unittest build_tools.tests.artifact_descriptor_overlap_test`
- Ran `git diff --check` on the changed descriptor.
- Ran `build_tools/fileset_tool.py artifact` against a temporary synthetic rocjitsu install tree and confirmed the descriptor routes:
  - `lib/librocjitsu.so` and `lib/librocjitsu_hooks.so` to the `lib` artifact
  - `bin/rocjitsu` to the `test` artifact
  - `share/rocjitsu/**` to the `run` artifact
  - `include/rocjitsu/**` to the `dev` artifact

JIRA ID : AIHPBLAS-3605
https://amd-hub.atlassian.net/browse/AIHPBLAS-3605

### DIFF
--- a/emulation/artifact-rocjitsu.toml
+++ b/emulation/artifact-rocjitsu.toml
@@ -1,10 +1,5 @@
 # rocjitsu
 [components.lib."emulation/rocjitsu/stage"]
-default_patterns = false
-include = [
-  "lib/librocjitsu.so",
-  "lib/librocjitsu_hooks.so",
-]
 [components.dbg."emulation/rocjitsu/stage"]
 [components.dev."emulation/rocjitsu/stage"]
 [components.run."emulation/rocjitsu/stage"]
--- a/emulation/artifact-rocjitsu.toml
+++ b/emulation/artifact-rocjitsu.toml
@@ -4,6 +4,5 @@
 [components.dev."emulation/rocjitsu/stage"]
 [components.run."emulation/rocjitsu/stage"]
 include = [
-  "bin/rocjitsu",
   "share/rocjitsu/**",
 ]
--- a/emulation/artifact-rocjitsu.toml
+++ b/emulation/artifact-rocjitsu.toml
@@ -6,3 +6,7 @@
 include = [
   "share/rocjitsu/**",
 ]
+[components.test."emulation/rocjitsu/stage"]
+include = [
+  "bin/rocjitsu",
+]
--- a/emulation/artifact-rocjitsu.toml
+++ b/emulation/artifact-rocjitsu.toml
@@ -4,7 +4,6 @@ default_patterns = false
 include = [
   "lib/librocjitsu.so",
   "lib/librocjitsu_hooks.so",
-  "lib/librocjitsu_kmd.so",
 ]
 [components.dbg."emulation/rocjitsu/stage"]
 [components.dev."emulation/rocjitsu/stage"]
--- a/emulation/artifact-rocjitsu.toml
+++ b/emulation/artifact-rocjitsu.toml
@@ -3,10 +3,13 @@
 default_patterns = false
 include = [
   "lib/librocjitsu.so",
+  "lib/librocjitsu_hooks.so",
+  "lib/librocjitsu_kmd.so",
 ]
 [components.dbg."emulation/rocjitsu/stage"]
 [components.dev."emulation/rocjitsu/stage"]
 [components.run."emulation/rocjitsu/stage"]
 include = [
+  "bin/rocjitsu",
   "share/rocjitsu/**",
 ]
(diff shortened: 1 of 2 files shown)


amd-eochoalo (approved):

geomin12 (approved):
lgtm! errors are flaky infra issues, can retry

emulation/artifact-rocjitsu.toml
```diff
@@ -3,10 +3,12 @@
 default_patterns = false
 include = [
   "lib/librocjitsu.so",
+  "lib/librocjitsu_hooks.so",
 ]
 [components.dbg."emulation/rocjitsu/stage"]
 [components.dev."emulation/rocjitsu/stage"]
 [components.run."emulation/rocjitsu/stage"]
 include = [
+  "bin/rocjitsu",
```
amd-arosa: we already ship mirage. mirage is built to be the public rocjitsu cli.

also just `alias rocjitsu="mirage run"`

amd-arosa (changes requested):
don't ship the rocjitsu cli.
we should ship the hook if it's ready.
please confirm the  hook doesn't containe all rocjitsu symbols.
also add it to the symbol check in the cmake

emulation/artifact-rocjitsu.toml
```diff
@@ -3,10 +3,12 @@
 default_patterns = false
```
amd-arosa, line 3: remove `default_patterns = false` and the entire include part

emulation/artifact-rocjitsu.toml
newling: I have an end-to-end CI workflow (race detection in rocm-libraries) working in this PR: https://github.com/ROCm/rocm-libraries/pull/9363 that uses rocjitsu cli. That PR builds rocjitsu in the CI job though, which I'd prefer to avoid. That is why I wanted to ship rocjitsu (i.e. this PR) -- then I wouldn't need to build rocjitsu in CI.  

I have tried getting the same end-to-end working with Mirage (instead of rocjitsu cli) in this PR: https://github.com/ROCm/rocm-libraries/pull/9613 but it looks like I need to install cargo. In https://github.com/ROCm/rocm-libraries/actions/runs/29791059379/job/88513099846 : `Could not find MIRAGE_CARGO_FOUND using the following names: cargo` in. 

Using cargo is not a blocker, but it adds a new dimension of complexity. I want this to be as easy as possible for developers to reproduce. The 'best' for rocm-libraries developers would've been just to install rocjitsu. 

Anyway I will push try cargo (i.e. I'll keep iterating on  https://github.com/ROCm/rocm-libraries/pull/9613), see if I can get that over the line.

amd-arosa: just updated https://ghcr.io/rocm/therock_build_manylinux_x86_64

emulation/artifact-rocjitsu.toml
```diff
@@ -1,12 +1,12 @@
...
-include = [
-  "lib/librocjitsu.so",
-]
 [components.dbg."emulation/rocjitsu/stage"]
 [components.dev."emulation/rocjitsu/stage"]
 [components.run."emulation/rocjitsu/stage"]
 include = [
   "share/rocjitsu/**",
 ]
+[components.test."emulation/rocjitsu/stage"]
```
newling, line 14: @amd-arosa adding components not intended for the public but needed for testing appears to be done like this elsewhere in TheRock

amd-arosa (approved):

emulation/artifact-rocjitsu.toml
amd-arosa: sure looks good for me

emulation/artifact-rocjitsu.toml
```diff
@@ -3,10 +3,15 @@
 default_patterns = false
 include = [
   "lib/librocjitsu.so",
+  "lib/librocjitsu_hooks.so",
```
Groverkss, line 6: Why are we shipping these hooks? Who is the consumer of these?

PR #6533, opened 2026-07-14
fix: pass ROCM_PATH to rocprofiler-systems sub-build

ajanicijamd:
## Motivation

* TheRock builds `rocprofiler-systems` without AI NIC profiling support
  (`ROCPROFSYS_BUILD_AINIC=OFF`), even on systems where the TheRock-bundled
  AMD SMI library satisfies the minimum version requirement (≥ 26.3).
* `rocprofiler-systems` detects AI NIC support by reading the AMD SMI version
  directly from `${ROCM_PATH}/include/amd_smi/amdsmi.h`. TheRock deliberately
  unsets `ROCM_PATH` from the build environment for all sub-projects
  (`cmake/therock_subproject.cmake`) to prevent accidental use of
  system-installed ROCm. Without `ROCM_PATH` set, the version check fails
  silently and all NIC collector code is excluded at compile time via
  `#if defined(ROCPROFSYS_BUILD_AINIC)` guards.
* `rocprofiler-sdk` already correctly works around this by passing
  `-DROCM_PATH="${THEROCK_BINARY_DIR}/dist/rocm"` in its `CMAKE_ARGS`.
  The `rocprofiler-systems` declaration was missing the same argument.

JIRA ID: ROCM-28005

## Technical Details

* One line added to the `CMAKE_ARGS` of the `rocprofiler-systems` declaration
  in `profiler/CMakeLists.txt`:
  `-DROCM_PATH="${THEROCK_BINARY_DIR}/dist/rocm"`.
* This points `rocprofiler-systems`' AMD SMI version detection at the
  TheRock-bundled headers (AMD SMI 26.5 in current builds), which satisfies
  the ≥ 26.3 requirement and causes `ROCPROFSYS_BUILD_AINIC` to be set to `ON`.

## Test Plan

* Build TheRock from source.
* Verify AI NIC support is compiled in:
  `grep ROCPROFSYS_BUILD_AINIC profiler/rocprofiler-systems/build/CMakeCache.txt`
  Expected: ROCPROFSYS_BUILD_AINIC:INTERNAL=ON
  `grep -i ainic logs/rocprofiler-systems_configure.log`
  Expected: -- AINIC support enabled (AMD SMI >= 26.3)
* Verify NIC sampling implementation strings are present in the built library:
  `strings dist/rocm/lib/librocprof-sys.so | grep -E "NicCollector|Discovered.*AI NIC|enabled for AI NIC PMC"`
  

## Test Result

Built successfully. All three checks pass with the fix applied.

## Submission Checklist


profiler/CMakeLists.txt
```diff
@@ -334,6 +334,7 @@ if(THEROCK_ENABLE_ROCPROFSYS)
       -DROCPROFSYS_PYTHON_VERSIONS:STRING="${_ROCPROFSYS_PYTHON_VERSIONS_STR}"
       -DROCPROFSYS_PYTHON_ROOT_DIRS:STRING="${_ROCPROFSYS_PYTHON_ROOT_DIRS_STR}"
       "-DROCPROFSYS_USE_MPI=${THEROCK_ENABLE_MPI}"
+      -DROCM_PATH="${THEROCK_BINARY_DIR}/dist/rocm"
```
ScottTodd, line 337: This is the wrong fix.

* rocprofiler-systems should be changed to not depend on `ROCM_PATH` here: https://github.com/ROCm/rocm-systems/blob/453799a7f50e144ccb7324cdb6bf24303b818278/projects/rocprofiler-systems/cmake/Packages.cmake#L266-L267 . We should be _removing_ uses of `ROCM_PATH` from the build system, not adding them, see https://github.com/ROCm/TheRock/issues/3976.
* The `${THEROCK_BINARY_DIR}/dist/rocm` directory is one possible final build output, it is **not** guaranteed to exist and be populated during subproject builds and is **not** something that subprojects should reference. See https://github.com/ROCm/TheRock/blob/main/docs/development/build_system.md#build-directory-layout

If just reading a source file, this doesn't necessarily need to read from a build/binary directory. rocprofiler-systems is already taking a `RUNTIME_DEPS` on `amdsmi` though, so it _could_ depend on that directory... could check what other subprojects are doing when they have similar requirements.

ajanicijamd: Thanks for the review. I will close this PR and create a new one with the correct fix in rocprofiler-systems.

PR #6536, opened 2026-07-14
ci: run quick PyTorch wheel test subset via run_test.py

ethanwee1:
ISSUE ID: #3291

## What

Switch the quick/per-PR **Test PyTorch Wheels** workflow (`test_pytorch_wheels.yml`) to run its unit tests through PyTorch's `run_test.py` — via the full suite's runner `run_pytorch_tests_full.py` — instead of invoking pytest directly through `run_pytorch_tests.py`.

The **same subset** of unit-test modules is preserved via `--include`: `test_nn`, `test_torch`, `test_cuda`, `test_unary_ufuncs`, `test_binary_ufuncs`, `test_autograd` (the list previously hard-coded in `run_pytorch_tests.py`).

This reuses the full suite's harness so the quick and full suites behave consistently (skip-list integration, serial execution, GPU selection, the `os._exit` hang-workaround).

## How

Brought over the supporting setup from the full workflow so `run_test.py` behaves the same:
- `--cap-add=SYS_PTRACE` on the container
- ROCm test env vars (`CI`, `PYTORCH_TEST_WITH_ROCM`, `PYTORCH_TESTING_DEVICE_ONLY_FOR`, `OPENBLAS_NUM_THREADS`, `PYTORCH_NUM_PYTEST_RERUNS`, `TEST_CONFIG=default`)
- `git safe.directory`
- `requirements-ci.txt` install
- test-report artifact upload

Per review feedback, the quick subset relies on **only the ROCm runtime** — pulled in transitively by the torch wheel (`PYTORCH_EXTRA_INSTALL_REQUIREMENTS` pins `rocm[libraries]`) — so it no longer installs `rocm[devel]`, and the `rocm-sdk init` / `ROCM_PATH` / `ROCM_HOME` / `LD_LIBRARY_PATH` setup step was dropped (packaged wheels must run via a plain `pip install torch`).

## Test

Ran the quick test workflow across **all supported PyTorch versions** (`release/2.9`, `release/2.10`, `release/2.11`, `release/2.12`, and upstream `nightly`) on gfx94X-dcgpu, py3.12, multi-arch index `https://rocm.nightlies.amd.com/whl-multi-arch/` with `device-gfx942` extras, torch `<ver>+rocm7.15.0a20260712`.

Both mechanisms dispatched for a side-by-side comparison:
- **New** = this branch (`run_test.py` via `run_pytorch_tests_full.py`)
- **Traditional** = `main` (direct pytest via `run_pytorch_tests.py`)

| PyTorch ref | torch version | New (`run_test.py`) | Traditional (`run_pytorch_tests.py`) |
|---|---|---|---|
| `release/2.9`  | 2.9.1   | https://github.com/ROCm/TheRock/actions/runs/29350894518 | https://github.com/ROCm/TheRock/actions/runs/29350919125 |
| `release/2.10` | 2.10.0  | https://github.com/ROCm/TheRock/actions/runs/29350896145 | https://github.com/ROCm/TheRock/actions/runs/29350920680 |
| `release/2.11` | 2.11.0  | https://github.com/ROCm/TheRock/actions/runs/29350897768 | https://github.com/ROCm/TheRock/actions/runs/29350923282 |
| `release/2.12` | 2.12.0  | https://github.com/ROCm/TheRock/actions/runs/29350899337 | https://github.com/ROCm/TheRock/actions/runs/29350924561 |
| `nightly`      | 2.14.0a0 | https://github.com/ROCm/TheRock/actions/runs/29350901050 | https://github.com/ROCm/TheRock/actions/runs/29350926185 |

_Earlier single-version run on `release/2.9`: https://github.com/ROCm/TheRock/actions/runs/29349160266 (surfaced that `--exclude-quantization-tests` is not recognized by `run_test.py` on older release branches)._

### Post-review validation (runtime-only, no devel / no env vars)

After dropping `rocm[devel]` and the env-var setup, re-validated on the latest nightlies index (torch `2.14.0a0+rocm7.15.0a20260721`, gfx94X-dcgpu, py3.12, `pytorch_git_ref: nightly`):

- https://github.com/ROCm/TheRock/actions/runs/30390330251


.github/workflows/test_pytorch_wheels.yml
```diff
@@ -200,20 +216,62 @@ jobs:
...
+          echo "C_INCLUDE_PATH=${ROCM_SYSDEPS_INCLUDE}:${C_INCLUDE_PATH:-}" >> "${GITHUB_ENV}"
+          echo "PKG_CONFIG_PATH=${ROCM_SYSDEPS_PKGCONFIG}:${PKG_CONFIG_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LD_LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LD_LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
+          echo "USE_MSLK=0" >> "${GITHUB_ENV}"
+
+          # Use ROCm's clang for JIT C++ compilation (torch.utils.cpp_extension)
+          # instead of requiring gcc/build-essential.
+          echo "CC=${ROCM_HOME}/lib/llvm/bin/clang" >> "${GITHUB_ENV}"
+          echo "CXX=${ROCM_HOME}/lib/llvm/bin/clang++" >> "${GITHUB_ENV}"
```
jithunnair-amd: If this is now going to be common between the two yml files, it might be best to do this via a helper bash script? Also, is this the pared-down list of env vars as in https://github.com/pytorch/pytorch/pull/188429?

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -182,10 +194,33 @@ jobs:
         timeout-minutes: 15
         run: |
           python build_tools/setup_venv.py ${VENV_DIR} \
-            --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }}" \
+            --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }} rocm[libraries,${{ steps.expand.outputs.device_extras }}]" \
```
ScottTodd: The torch packages that we build depend on the rocm packages, so this changed line should not be needed.

https://github.com/ROCm/TheRock/blob/ccaba7c998c248af55bfc803bd8e48eacc0333a6/external-builds/pytorch/build_prod_wheels.py#L1040-L1043

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -182,10 +194,33 @@ jobs:
...
+
+          echo "ROCM_HOME=${ROCM_HOME}"
+
+          # Add ROCm bin to PATH for all subsequent steps
+          echo "${ROCM_BIN}" >> "${GITHUB_PATH}"
+
+          # Persist the runtime discovery/link paths across workflow steps
+          echo "ROCM_PATH=${ROCM_HOME}" >> "${GITHUB_ENV}"
+          echo "ROCM_HOME=${ROCM_HOME}" >> "${GITHUB_ENV}"
+          echo "LD_LIBRARY_PATH=${ROCM_HOME}/lib/host-math/lib:${ROCM_SYSDEPS}/lib:${LD_LIBRARY_PATH:-}" >> "${GITHUB_ENV}"
```
ScottTodd: This step should not be needed. What tests specifically are failing without these environment variables set?

The packaging requirement is that users can `pip install torch` / `pip install rocm` / etc. and run their applications. No additional environment variables should be needed. No `ROCM_PATH`, no `ROCM_HOME`, no `LD_LIBRARY_PATH`. The code that we package and distribute has to be portable and not rely on global system state like these environment variables.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -204,13 +239,34 @@ jobs:
...
+      # through PyTorch's run_test.py (via run_pytorch_tests_full.py) instead of
+      # invoking pytest directly. This reuses the full test workflow's runner so
+      # behavior (skip-list integration, serial execution, GPU selection) stays
+      # consistent between the quick and full suites. The --include list is the
+      # subset previously hard-coded in run_pytorch_tests.py.
       - name: Run PyTorch tests
         timeout-minutes: 120
         run: |
-          python ./external-builds/pytorch/run_pytorch_tests.py \
+          python ./external-builds/pytorch/run_pytorch_tests_full.py \
```
ScottTodd: Comments like "this workflow has always run" are dated and don't really have a place in the code itself. They belongs in the PR description, review comment threads, etc.

How about rewording like so?
> Rather than calling pytest or PyTorch's run_test.py directly, we use a
> test runner script that invokes run_test.py after first handling our custom
> skip-list integration, serialized execution, GPU selection, and more.
> The list of included test suites is limited here compared to in test_pytorch_wheels_full.yml.

.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 1: > Switch the quick/per-PR **Test PyTorch Wheels** workflow (`test_pytorch_wheels.yml`) to run its unit tests through PyTorch's `run_test.py`

Note that we don't currently run this workflow as part of CI workflows (see https://github.com/ROCm/TheRock/issues/3291 which you tagged on this PR)

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -182,10 +194,33 @@ jobs:
...
+      # needs the ROCm runtime libraries (rocm[libraries], not rocm[devel]).
+      # We keep just the runtime paths PyTorch needs to discover and load the
+      # wheel-provided ROCm SDK; the compiler/header/cmake env vars from the
+      # full workflow are intentionally omitted (see pytorch/pytorch#188429).
+      - name: Initialize ROCm SDK and configure environment
+        run: |
+          rocm-sdk init
+
+          ROCM_HOME="$(rocm-sdk path --root)"
+          ROCM_BIN="$(rocm-sdk path --bin)"
```
ScottTodd: These `rocm-sdk path` commands only work if the `rocm[devel]` package is installed, and we should not be requiring it for python package tests.

The linked test runs from the PR description are also using the devel package like https://github.com/ROCm/TheRock/actions/runs/29350897768/job/87146548689

.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 230: Can we rename `run_pytorch_tests_full.py` if it will now also be used for "non-full" testing too? Then delete the old https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/run_pytorch_tests.py

That could be a follow-up change, but I don't want to leave orphaned files and stale documentation in the repository for long. See also the docs: https://github.com/ROCm/TheRock/blob/ccaba7c998c248af55bfc803bd8e48eacc0333a6/external-builds/pytorch/README.md?plain=1#L239-L281

ScottTodd (commented):
Happy to see more the migration to reusing PyTorch's run_test.py instead of our fully separate custom test runner. Do you also want to tag https://github.com/ROCm/TheRock/issues/2406 on this PR maybe? Or check if some of the details there are already solved?

.github/workflows/test_pytorch_wheels.yml
jithunnair-amd: Can't agree with this sentiment more. Let's work to get rid of all these env vars, otherwise it's a usability nightmare.

ScottTodd: Separately I'm also going to dig into why so many of the ROCm components themselves are requiring variables like these in their local tests (see scripts in https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions/test_executable_scripts). I'm going to write down more of the requirements in files like a TESTING.md (see https://github.com/ROCm/TheRock/issues/6711) that we can point developers at.

ethanwee1:
Follow-ups stacked on this PR: #6935 consolidates the duplicate test runners by renaming `run_pytorch_tests_full.py` to `run_pytorch_tests.py` and deleting the old script, and #6937 adds a quick-test summary job to `test_pytorch_wheels.yml`.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -204,13 +219,33 @@ jobs:
...
+            --continue-on-collection-errors
+
+      - name: Upload test reports
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: test-reports-${{ inputs.amdgpu_family }}
+          path: external-builds/pytorch/pytorch/test/test-reports/
+          if-no-files-found: ignore
+          retention-days: 30
```
ScottTodd, line 251: While these test reports look great, the test summary at the tail end of the logs seems like it could be more directly useful?

Compare a baseline run: https://github.com/ROCm/TheRock/actions/runs/30565987573/job/90950498017

Image

to the run testing this PR: https://github.com/ROCm/TheRock/actions/runs/30390330251/job/90379992249

Image

We might want to adjust how we monitor pytorch test failures after this change (cc @TriveniTadapaneni @amd-chiranjeevi )

The test reports have e.g.
```
========== 859 passed, 39 skipped, 138 deselected in 89.43s (0:01:29) ==========
The following tests failed consistently: ['test/test_torch.py::TestTorch::test_sobol_invalid_inputs']
```
at the end. Maybe we could add a workflow step that pulls out the relevant details from the reports or `cat`s them to the github logs for easier viewing without downloading?

ScottTodd (approved):

.github/workflows/test_pytorch_wheels.yml
jithunnair-amd: @ethanwee1 I think we have discussed the idea of updating the workflow run summary to have all the failures listed in it. @ScottTodd Should we also make that part of this PR?

jithunnair-amd: @ScottTodd For more context, here is the [main workhorse script](https://github.com/ROCm/pytorch/blob/develop/.automation_scripts/pytorch-unit-test-scripts/summarize_xml_testreports.py) for parsing the xmls/logs that we use, and this is the [main summarization script](https://github.com/ROCm/pytorch/blob/develop/.automation_scripts/pytorch-unit-test-scripts/generate_summary.py). This is [an example](https://github.com/ROCm/pytorch/actions/runs/30444136839) of the kind of output it generates.

ScottTodd: This might be covered by this draft PR:
* https://github.com/ROCm/TheRock/pull/6937

ethanwee1: Yep, scott pointed out the followup test summary PR.

PR #6545, opened 2026-07-14
feat(ci): allow schedule triggers to respect explicit family inputs

geomin12:
Modify select_targets() to check for explicit linux_amdgpu_families and windows_amdgpu_families inputs on schedule triggers before defaulting to all families.

This enables external repositories (like rocm-libraries) to call setup_multi_arch.yml on a schedule trigger while limiting builds to specific GPU families, rather than being forced to run all.

Noticing that in rockrel: https://github.com/ROCm/rockrel/actions/runs/29295395434/job/87093082810#step:4:23, `all` is not honored: https://github.com/ROCm/rockrel/actions/runs/29295395434/job/87093082810#step:4:32 as the plumbing is only in workflow_dispatch

Now, the flag will be properly honored

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -832,10 +832,25 @@ def select_targets(ci_inputs: CIInputs) -> TargetSelection:
         linux_names = list(defaults)
         windows_names = list(defaults)
     elif ci_inputs.is_schedule:
-        # Full nightly coverage: every known family, including targets
-        # that are too slow or expensive for per-push CI.
-        linux_names = list(all_families.keys())
-        windows_names = list(all_families.keys())
+        # Schedule trigger: use explicit inputs if provided, else all families.
+        # "all" or empty = all known families. "none" = skip platform.
+        linux_names = list(ci_inputs.linux_amdgpu_families)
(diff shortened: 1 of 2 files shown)


tony-davis (approved):
Nice fix, Geo! This cleans up a real gap; scheduled runs were quietly ignoring the caller's family inputs and always building everything, so external callers like rockrel couldn't scope a nightly. The plumbing lines up nicely (the env vars flow through `setup_multi_arch.yml` on every trigger already), and `test_schedule_respects_explicit_inputs` is a proper before/after regression, so I'm happy with the coverage. Approving.

A few optional thoughts, none blocking:

- The schedule branch is now almost a twin of the `workflow_dispatch` branch (the only real difference is empty meaning "all" here vs "skip" there). Might be worth pulling out a small helper (something like `_resolve_explicit_families(names, all_families, empty_means_all)`) so the two paths don't drift apart later. Totally fine to leave as-is if you'd rather not.
- The new tests cover explicit families and `none`, but not the `all` case on schedule (which is what the description leads with). A quick parametrized case there would lock that branch too.
- Tiny nit on the description: I think before this change schedule ignored inputs entirely and always ran all families, so `all` was kind of working by accident. The actual win is honoring explicit families and `none`. Probably worth a one-line tweak so future readers aren't confused.

Thanks for chasing this down!

PR #6546, opened 2026-07-14
Revert broken manifest diff test

ScottTodd:
## Motivation

1. These new tests fail on PRs from forks: https://github.com/ROCm/TheRock/pull/6147#pullrequestreview-4698665652 as the bucket URL is dynamic unless properly mocked out
2. The "test_normal_summary" test case above already explicitly says that it is **not** checking the generated markdown as that is not part of the API and we don't want change detector tests.

## Submission Checklist


ScottTodd:
Post-merge review please. We can also fully revert https://github.com/ROCm/TheRock/pull/6147.

PR #6550, opened 2026-07-14
Drop per-family release support from setup_venv.py

ScottTodd:
## Motivation

* Progress on https://github.com/ROCm/TheRock/issues/3340.
  * Per-family python package releases have been discontinued in favor of multi-arch releases which use a single index like https://rocm.nightlies.amd.com/whl-multi-arch/ instead of per-family indexes like https://rocm.nightlies.amd.com/v2/gfx94X-dcgpu/. We no longer need this script to support index subdirectories.
* Follow-up to https://github.com/ROCm/TheRock/issues/5285
  * DNS issues on `linux-gfx942-1gpu-ossci` runners were resolved, so we no longer need to fallback to `--find-links` S3 URLs instead of the CDN URLs. We may want to deliberately use S3 directly instead of the CDN to save on cloud billing charges, but that should be a direct decision outside of this script, not a fallback path.

## Test Plan

* Unit test coverage
* Locally ran `python build_tools/setup_venv.py test.venv --packages rocm[libraries,devel,device-gfx1100] --index-name nightly` on Windows and confirmed that it ran `pip install --index-url=https://rocm.nightlies.amd.com/whl-multi-arch/ 'rocm[libraries,devel,device-gfx1100]` in the venv it created which successfully installed packages
* Python package tests on the CI runs of this PR will use the script.
* Nightly release PyTorch tests will exercise the path where the DNS fallback was needed.

## Submission Checklist

### DIFF
--- a/build_tools/setup_venv.py
+++ b/build_tools/setup_venv.py
@@ -47,7 +47,6 @@ import shlex
 import shutil
 import subprocess
 import sys
-import re
 import time
 
 from github_actions.github_actions_api import *
--- a/build_tools/setup_venv.py
+++ b/build_tools/setup_venv.py
@@ -50,12 +50,10 @@ from pathlib import Path
 import platform
 import shlex
 import shutil
-import socket
 import subprocess
 import sys
 import re
 import time
-from urllib.parse import urlparse
 
 from github_actions.github_actions_api import *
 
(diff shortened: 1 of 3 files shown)


build_tools/setup_venv.py
```diff
@@ -520,36 +373,24 @@ def main(argv: list[str]):
...
     install_options.add_argument(
         "--find-links",
         type=str,
         help="Package location URL for pip --find-links (compatible with --index-url)",
     )
 
-    # Scrape available subdirs for --index-subdir choices.
-    available_subdirs = _scrape_rocm_index_subdirs()
-    install_options.add_argument(
-        "--index-subdir",
```
ScottTodd, line 540: Oh right, this still has uses like https://github.com/ROCm/TheRock/blob/710812f49673c52777894c0a6fed0ce9d526ac12/.github/workflows/test_rocm_wheels.yml#L163-L176

Need to prune those first.

ScottTodd: Handling those in:
* https://github.com/ROCm/TheRock/pull/6551
* https://github.com/ROCm/TheRock/pull/6552

geomin12 (approved):

Issue #4769, opened 2026-04-23
[Multi-arch] Opt-in to THEROCK_KPACK_SPLIT_ARTIFACTS in non-multi-arch CI


ScottTodd:
We flipped in `THEROCK_KPACK_SPLIT_ARTIFACTS` for multi-arch CI in https://github.com/ROCm/TheRock/pull/4652 but were not able to similarly flip it for non-multi-arch CI. To flip the flag for CI in repositories like rocm-libraries and rocm-systems we can either migrate to multi-arch CI (https://github.com/ROCm/TheRock/issues/3343) or make it possible to flip the flag in the context of the non-multi-arch workflows.

geomin12:
per https://github.com/ROCm/TheRock/pull/6694 and https://github.com/ROCm/rocm-systems/pull/8860 and https://github.com/ROCm/rocm-libraries/pull/9596 landing, this is now closed and completed

PR #6551, opened 2026-07-14
Limit test_pytorch_wheels to only kpack (multi-arch) packages

ScottTodd:
## Motivation

* Progress on https://github.com/ROCm/TheRock/issues/3340
* Partially unblocks https://github.com/ROCm/TheRock/pull/6550

We only publish multi-arch packages now, and per-family CI/CD is being removed. This trims pytorch test workflow code to only use the kpack (multi-arch) path.

## Technical Details

> [!IMPORTANT]
> We will also need to remove the `multi_arch` input from the wrapper workflow in https://github.com/ROCm/rockrel/blob/main/.github/workflows/test_pytorch_wheels_full.yml
> * https://github.com/ROCm/rockrel/pull/76

Note that we can't quite delete the `KPACK_SPLIT_ARTIFACTS` flag altogether here https://github.com/ROCm/TheRock/blob/710812f49673c52777894c0a6fed0ce9d526ac12/FLAGS.cmake#L18-L22 due to https://github.com/ROCm/TheRock/issues/4769, and this code that is being used in rocm-libraries and rocm-systems: https://github.com/ROCm/TheRock/blob/710812f49673c52777894c0a6fed0ce9d526ac12/build_tools/github_actions/build_configure.py#L111-L117

## Test Plan

Manually triggered run at https://github.com/ROCm/TheRock/actions/runs/29376216394/job/87230121159 installed packages as expected.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -17,13 +17,10 @@ on:
         type: string
         default: "linux-gfx942-1gpu-ccs-csp-ossci-rocm"
       package_index_url:
-        description: >-
-          Base Python package index URL to test.
-          Per-family: nightly/dev URL with a "v2" or "v2-staging" subdir (without a GPU family subdir).
-          Multi-arch: unified index URL (e.g. "https://rocm.nightlies.amd.com/whl-multi-arch/").
+        description: Python package index URL to test, typically nightly/dev URL with a "whl-multi-arch" subdir
         required: true
         type: string
-        default: "https://rocm.nightlies.amd.com/v2"
+        default: "https://rocm.nightlies.amd.com/whl-multi-arch/"
       python_version:
         required: true
         type: string
@@ -55,12 +52,6 @@ on:
           - "single"
           - "all"
         default: "single"
-      multi_arch:
-        description: >-
-          Install from a multi-arch index using device extras
-          (e.g. torch[device-gfx942]) instead of per-family index subdirs.
-        type: boolean
-        default: false
 
   workflow_call:
     inputs:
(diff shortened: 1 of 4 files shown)


.github/workflows/test_pytorch_wheels.yml
```diff
@@ -198,18 +182,10 @@ jobs:
...
-          if [ "${{ inputs.multi_arch }}" = "true" ]; then
-            python build_tools/setup_venv.py ${VENV_DIR} \
-              --packages "torch[${{ steps.expand.outputs.device_extras }}]==${{ inputs.torch_version }}" \
-              --index-url=${{ inputs.package_index_url }} \
-              --activate-in-future-github-actions-steps
-          else
-            python build_tools/setup_venv.py ${VENV_DIR} \
-              --packages torch==${{ inputs.torch_version }} \
-              --index-url=${{ inputs.package_index_url }} \
-              --index-subdir=${{ inputs.amdgpu_family }} \
```
marbre, line 210: Seems we may be able to drop `--index-subdir` from `setup_venv.py`?

marbre (approved):

.github/workflows/test_pytorch_wheels.yml
marbre: Just saw the linked issue :D

ethanwee1 (approved):
LGTM

.github/workflows/test_pytorch_wheels.yml
ScottTodd: Yeah that's what I started on first, then I remembered these uses of it were still there :P

ScottTodd:
Thanks for the reviews, sent https://github.com/ROCm/rockrel/pull/76 for rockrel which will need to be merged at the same time.

.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 151: There is also a reference to `--index-subdir` here. I'll remove that in a follow-up PR

https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/summarize_test_pytorch_workflow.py

PR #6558, opened 2026-07-15
feat(ci): Run full JAX test suite

charleshofer:
## Motivation

Run the whole JAX test suite instead of a handful of tests, through ROCm/jax's `ci/run_pytest_rocm.sh`, which sets up pytest and excludes the multi-GPU tests. Coverage goes from a few hundred tests to roughly 28k-29k per job, which is what surfaced the ROCm 10 plugin-naming break (fixed separately, already on `main`) and the sticky library faults below.

## Technical Details

**The job is a script, not a workflow step.** The environment, the deselected tests and the retry logic all lived in the YAML, so reproducing a failure locally meant transcribing it by hand. `external-builds/jax/run_jax_tests.py` now owns all of it, and the workflow calls it, so this reproduces a CI job:

```bash
python external-builds/jax/run_jax_tests.py \
  --jax-dir jax_tests --jax-version 0.10.2 --amdgpu-family gfx94X-dcgpu
```

`--dry-run`, `--debug` (run only what the skip lists skip), `--no-retries` and `--cpus N` cover the rest.

**The suite script stays the source of truth.** The runner adds only what is ours and reads the allocator and XLA flags back out of the script's environment section, so the retry pass matches the run it checks and a version changing those values needs no change here. If that section cannot be read, the job fails with an error naming what to update rather than retrying under a mismatched environment.

**Installs.** ROCm via `setup_venv.py`, wheels via `install_jax_wheels.py`, test deps via `install_jax_test_requirements.py` from JAX's own pinned lock file, all with retries, since one PyPI timeout has cost a job a whole run. The ROCm packages are `rocm[libraries]` alone, since that is what RELEASES.md tells users to install; verified on MI355X that matmul, cholesky, FFT and the MIOpen convolutions all pass without `[devel]`.

**Skip lists.** Deselected tests move to `external-builds/jax/skip_tests/`, following the PyTorch layout, so an entry carries the reason it exists and what would let it be removed. MIOpen selects convolution kernels nondeterministically on gfx94X, so the float32 conv tests flake on `v0.10.1` and `v0.10.2`, along with pooling, Toeplitz and Hankel, which build on the same kernels, and `polymul`, which convolves fixed inputs.

**Sticky library faults need a new process.** In-process reruns clear worker aborts and one-off runtime errors. `rocblas_cgemm failed` and `Failed to load HSACO` are different: rocBLAS caches a failed lazy Tensile load for a problem shape, so that worker fails the shape for the rest of its life while its siblings on the same GPU keep passing, and no in-process rerun can clear it. When the suite fails, `test_tools/list_pytest_failed_tests.py` pulls the failed node ids out of the JSON reports and the runner re-runs them serially in a fresh process. That pass decides the job, but only when every report ends in a finished session, so a run pytest stopped early cannot go green on the failures it happened to record. The list is capped at 40, so a badly broken job still fails.

**CPU sizing.** A container shares the host's kernel, so a job holding 25 cores of a shared box reads 192 and sizes every thread pool from that: over a thousand threads per worker against 16 workers. The run is pinned to `KUBE_CPU_REQUEST`, which takes a worker to about a hundred. Where the variable is not injected yet (OSSCI-2631), it warns and takes one thread per worker, as the PyTorch jobs do.

**Also:** test step timeout raised to 120 minutes, since the suite runs 35-70 minutes per version and 60 killed a job that had reached every test; test logs uploaded as an artifact; driver and GPU info printed, whose device serials identify a recurring bad device across runs; the workflow renamed to match what it now runs; unit tests added for the runner, the install scripts, the skip lists and the failed-test lister.

## Test Plan

- ROCm 7, [run 30527360755](https://github.com/ROCm/TheRock/actions/runs/30527360755): `v0.10.0` and `v0.10.2` pass on all four Python versions; `v0.10.1` has 4 failures out of 28396 passed, and `v0.11.0` has 4 xdist serialization errors out of 29187 passed.
- ROCm 10 before the plugin-name fix, [run 30561401221](https://github.com/ROCm/TheRock/actions/runs/30561401221): all 15 jobs failed with 1468 failures each, essentially all `No FFI handler registered for hipsolver_*` or `'NoneType' object has no attribute ...`.
- ROCm 10 after it, [run 30701879601](https://github.com/ROCm/TheRock/actions/runs/30701879601): 15 failures across 7 of 15 jobs, 33-56 minutes per job: 8 rocBLAS complex gemm, 2 HSACO `InvalidImage`, 2 hipSOLVER, 3 `testPolyMul7`.
- Retry pass, [run 30729289444](https://github.com/ROCm/TheRock/actions/runs/30729289444): the two jobs that faulted, `testVecmat7` on `rocblas_cgemm` and `testPinv4` on HSACO, both passed the fresh-process retry in seconds on the same GPU after failing every in-process attempt.
- Worker-count probes: 8 workers, [run 30717048816](https://github.com/ROCm/TheRock/actions/runs/30717048816), still left 19 failures across 6 of 15 jobs; 4 ran clean but cost 45% more wall time. The retry pass buys the same result without the wall time.
- Full matrix on this branch, [run 30781739300](https://github.com/ROCm/TheRock/actions/runs/30781739300): all 15 test jobs green.

Ruled out: memory creep with and without `HSA_NO_SCRATCH_RECLAIM`, GPU memory exhaustion (which gives a clean `RESOURCE_EXHAUSTED`, never `INTERNAL`), and co-tenancy on sibling GPUs, since neighbours were present through the fault-free stretches too.

## Open items

- rocBLAS and HIP bug reports for the sticky complex-gemm and code-object load failures.
- The gfx94X convolution tests and `polymul` are deselected rather than fixed; the MIOpen kernel-selection root cause is still open.
- `KUBE_CPU_REQUEST` is not injected on every runner pool yet (OSSCI-2631).
- Test-size filtering for PR versus nightly is handled separately in #6753.

JIRA ID: AIJAX-210

### DIFF
--- a/.github/workflows/test_multi_arch_linux_jax_wheels.yml
+++ b/.github/workflows/test_multi_arch_linux_jax_wheels.yml
@@ -142,6 +142,8 @@ jobs:
         --group-add video
         --user root
         --env-file /etc/podinfo/gha-gpu-isolation-settings
+        -e KUBE_CPU_REQUEST
+        -e KUBE_MEMORY_REQUEST
         --dns 8.8.8.8
         --dns 8.8.4.4
 
@@ -227,42 +229,20 @@ jobs:
             --amdgpu-families "${{ inputs.test_amdgpu_family }}" \
             --output-mode=device-extras
 
+      # rocm[libraries] alone, since that is what RELEASES.md tells users to
+      # install: anything the wheels need belongs there, not in [devel].
       - name: Install ROCm packages
+        env:
+          ROCM_PACKAGES: rocm[libraries${{ steps.expand.outputs.device_extras && format(',{0}', steps.expand.outputs.device_extras) || '' }}]==${{ inputs.rocm_version }}
         run: |
-          python -m pip install --upgrade pip
+          python build_tools/setup_venv.py "${VIRTUAL_ENV}" \
+            --packages "${ROCM_PACKAGES}" \
+            --find-links "${ROCM_PACKAGE_FIND_LINKS_URL}"
 
-          pkg="rocm[libraries]"
-          if [ -n "${{ steps.expand.outputs.device_extras }}" ]; then
-            pkg="${pkg%]},${{ steps.expand.outputs.device_extras }}]"
-          fi
-
-          python -m pip install --pre \
-            --find-links "${ROCM_PACKAGE_FIND_LINKS_URL}" \
-            "${pkg}==${{ inputs.rocm_version }}"
-
-          echo "ROCM_ROOT=$(rocm-sdk path --root)" >> "$GITHUB_ENV"
-
+      # Package names, versions and the index arrive as env vars set above.
       - name: Install JAX wheels from package index
         run: |
-          echo "Installing from:"
-          echo "  ${WHEEL_INDEX_URL}"
-
-          pip install \
-            --index-url "${WHEEL_INDEX_URL}" \
-            "${JAX_PLUGIN_PACKAGE}==${JAX_PLUGIN_VERSION}" \
-            "${JAX_PJRT_PACKAGE}==${JAX_PJRT_VERSION}"
-
-          if [ -n "${JAXLIB_VERSION}" ]; then
-            pip install \
-              --index-url "${WHEEL_INDEX_URL}" \
-              "jaxlib==${JAXLIB_VERSION}"
-
-            pip install "jax==${JAX_VERSION}"
-          else
-            pip install \
-              "jax==${JAX_VERSION}" \
-              "jaxlib==${JAX_VERSION}"
-          fi
+          python external-builds/jax/install_jax_wheels.py
 
       # jaxlib comes from upstream PyPI and only knows the ROCm majors that
       # existed when it was released, so it has to be taught about newer ones.
@@ -271,41 +251,38 @@ jobs:
           python external-builds/jax/patch_installed_jax_rocm_plugin_names.py \
             --plugin-package "${JAX_PLUGIN_PACKAGE}"
 
-      - name: Run JAX tests
-        timeout-minutes: 60
-        env:
-          # Avoid large upfront GPU memory reservation.
-          XLA_PYTHON_CLIENT_PREALLOCATE: "false"
-
-          # Set XLA allocator. JAX 0.10.2 uses the default allocator; all
-          # other releases keep the platform allocator.
-          XLA_PYTHON_CLIENT_ALLOCATOR: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && 'default' || 'platform' }}
-
-          # JAX 0.10.2 only: ROCm/HIP runtime tuning to avoid a pytest
-          # slowdown/hang; other releases keep the runtime defaults.
-          # TODO:(magaonka-amd) remove these as soon as fixes from corresponding system teams lands.
-          ROCPROFILER_QUEUE_INTERPOSITION: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '0' }}
-          DEBUG_HIP_DYNAMIC_QUEUES: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '0' }}
-          HSA_NO_SCRATCH_RECLAIM: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '1' }}
-
-          # Reduce compilation/autotune memory pressure.
-          XLA_FLAGS: >-
-            --xla_gpu_force_compilation_parallelism=1
-            --xla_gpu_enable_nccl_comm_splitting=false
-            --xla_gpu_enable_command_buffer=
-
+      # The device serial numbers it prints identify a recurring bad device
+      # across runs, since the runner and the device index change every time.
+      - name: Driver / GPU sanity check
+        timeout-minutes: 3
         run: |
-          set -euxo pipefail
+          python build_tools/print_driver_gpu_info.py
 
-          unset ROCM_ROOT
-          unset HIP_DEVICE_LIB_PATH
-          unset JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH
-          unset JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH
+      - name: Install JAX test requirements
+        run: |
+          python external-builds/jax/install_jax_test_requirements.py \
+            --jax-dir jax \
+            --python-version "${PYTHON_VERSION}"
 
-          python -c "import jax; print(jax.local_devices())"
+      # The runner owns the test environment, the skip lists, the pod's CPU
+      # allocation and both layers of retry, so that the same command
+      # reproduces this job locally. See the comments in the script.
+      - name: Run JAX tests
+        # The suite runs 35-70 min across versions, so 60 left no headroom: one
+        # job reached 100% of the tests and was still killed by the timeout.
+        timeout-minutes: 120
+        run: |
+          python external-builds/jax/run_jax_tests.py \
+            --jax-dir jax \
+            --jax-ref "${{ inputs.jax_ref }}" \
+            --amdgpu-family "${{ inputs.test_amdgpu_family }}"
 
-          echo "=== Starting full test suite ==="
-          pytest jax/tests/multi_device_test.py -q --log-cli-level=INFO
-          pytest jax/tests/core_test.py -q --log-cli-level=INFO
-          pytest jax/tests/util_test.py -q --log-cli-level=INFO
-          pytest jax/tests/scipy_stats_test.py -q --log-cli-level=INFO
+      - name: Upload test logs
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: jax-test-logs-${{ inputs.jax_ref }}-py${{ inputs.python_version }}-${{ inputs.test_amdgpu_family }}
+          path: |
+            jax/logs/
+            jax/test-artifacts/
+          if-no-files-found: ignore
--- a/external-builds/jax/run_jax_tests.py
+++ b/external-builds/jax/run_jax_tests.py
@@ -1,0 +1,654 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Runs the JAX test suite against installed JAX ROCm wheels.
+
+The suite script in the ROCm/jax checkout (ci/run_pytest_rocm.sh) stays the source
+of truth for how the tests run. This adds only what is ours:
+
+  * the ROCm runtime workarounds this repository needs;
+  * the known-bad tests from skip_tests/, as a pytest -k expression;
+  * two layers of retry, described under "Retries" below;
+  * holding the run to the CPUs the pod was given, described under "Resources".
+
+Nothing the suite script decides is repeated here. Its environment is read back
+out of the script, so a version that changes its allocator or its XLA flags,
+including behind a conditional, needs no change in this file. Reading it is
+required rather than best-effort: the retry pass decides the result, so running
+it under a different environment than the suite would make that verdict
+meaningless. A script whose section markers have been renamed fails the run with
+an error naming them.
+
+Retries
+-------
+
+--in-process-reruns is pytest-rerunfailures: it repeats a failed test in the same
+worker, recovering a crashed worker or a one-off runtime error.
+
+--fresh-process-retry re-runs the reported failures in a new process, and decides
+the result. It is not interchangeable with the above: a sticky failure leaves the
+process failing every later test the same way, which no rerun inside it can
+clear. It only speaks for the suite when the suite's own pytest sessions ran to
+the end, which their reported exit codes are what say.
+
+--no-retries turns off both, running the suite exactly as the script does.
+
+Resources
+---------
+
+A container shares the host's kernel, so nothing a process can query describes
+the container: on the gfx942 runners a job holding 25 cores of an 8-way shared
+box reads 192 cores and 2268 GiB, and sizes every thread pool from that. What
+the pod actually got arrives in KUBE_CPU_REQUEST, and --cpus says the same thing
+on a machine that has no CI runner to inject it.
+
+The run is pinned to that many CPUs, which the suite and the retry pass inherit.
+Measured on a gfx942 runner, one worker holds 1021 threads unpinned and 107
+pinned to 25 cores with OpenBLAS capped, against 16 workers at a time.
+
+Example usage:
+
+    # As CI runs it.
+    python run_jax_tests.py --jax-dir jax --jax-ref rocm-jaxlib-v0.11.0 \\
+        --amdgpu-family gfx94X-dcgpu
+
+    # The suite with both retry layers off.
+    python run_jax_tests.py --jax-dir jax --no-retries
+
+    # Print the environment and commands without running anything.
+    python run_jax_tests.py --jax-dir jax --jax-version 0.10.2 --dry-run
+
+    # Run only the tests the skip list would have skipped.
+    python run_jax_tests.py --jax-dir jax --jax-version 0.10.2 --debug
+"""
+
+import argparse
+import dataclasses
+import json
+import os
+import platform
+import shlex
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+THIS_SCRIPT_DIR = Path(__file__).resolve().parent
+THEROCK_DIR = THIS_SCRIPT_DIR.parent.parent
+
+sys.path.insert(0, os.fspath(THIS_SCRIPT_DIR))
+sys.path.insert(0, os.fspath(THEROCK_DIR / "test_tools"))
+
+from list_pytest_failed_tests import failed_in_report
+from skip_tests.create_skip_tests import keyword_expression
+
+JAX_REF_PREFIX = "rocm-jaxlib-v"
+
+# Relative to the ROCm/jax checkout --jax-dir names, which is also the working
+# directory the suite runs in, rather than to wherever this script was called
+# from.
+RELATIVE_SUITE_SCRIPT = Path("ci") / "run_pytest_rocm.sh"
+
+# Sourced by the suite script for the JAXCI_* defaults the section below reads,
+# and safe to source here too: it only assigns variables.
+RELATIVE_SUITE_ENV_FILE = Path("ci") / "envs" / "default.env"
+
+# One report per pytest invocation the suite script makes, matched by pattern so
+# that a version splitting the suite differently needs no change here.
+REPORT_GLOB = "logs/pytest_results*.json"
+
+# The pytest exit codes of a session that ran to the end. Above them the session
+# stopped early instead: 2 interrupted, 3 internal error, 4 usage error, 5
+# nothing collected. Such a report lists only the tests that got that far.
+PYTEST_EXIT_ALL_PASSED = 0
+PYTEST_EXIT_TESTS_FAILED = 1
+
+# The section of the suite script that only computes the environment: exports,
+# echoes and device queries, with the installs above it and the test run below it.
+# Evaluating it is how the per-version environment reaches the retry pass, which
+# has to match the run it is checking. Sourcing the script itself would run the
+# tests, and copying its values here would be a second source of truth.
+ENV_SECTION_START = "# Set up the generic test environment variables"
+ENV_SECTION_END = "# Run tests"
+
+# The section exports its own scratch variables, lowercase by convention, next to
+# the configuration worth keeping. The xdist worker count is for the parallel run
+# and would contradict the serial retry.
+SUITE_ENV_EXCLUDES = ["JAX_ENABLE_ROCM_XDIST"]
+
+# ROCm/HIP runtime tuning that avoids a pytest slowdown and hang. Ours rather than
+# the suite's, so it belongs here.
+# TODO:(magaonka-amd) remove once the system teams' fixes land.
+THEROCK_ENV = {
+    "ROCPROFILER_QUEUE_INTERPOSITION": "0",
+    "DEBUG_HIP_DYNAMIC_QUEUES": "0",
+    "HSA_NO_SCRATCH_RECLAIM": "1",
+}
+
+# A test runner is often a single-device slice of a shared host, where these hide
+# the other devices but not other tenants. Logged next to the device serials from
+# print_driver_gpu_info.py, they tell a noisy neighbour apart from a fault.
+GPU_VISIBILITY_ENV_VARS = [
+    "ROCR_VISIBLE_DEVICES",
+    "HIP_VISIBLE_DEVICES",
+    "GPU_DEVICE_ORDINAL",
+]
+
+# What the pod was given, injected by the CI runners. A container shares the
+# host's kernel, so nothing the process can query is about the container: the
+# gfx942 runners report 192 cores and 2268 GiB to a job holding 25 cores of an
+# 8-way shared box, and the cgroups are quiet because CI sets CPU requests and
+# no limits. These variables are the only account of the real allocation.
+CPU_REQUEST_VAR = "KUBE_CPU_REQUEST"
+MEMORY_REQUEST_VAR = "KUBE_MEMORY_REQUEST"
+
+# The most pytest workers the suite starts in parallel, see the num_processes
+# cap in ci/run_pytest_rocm.sh. Each one loads its own copy of every library.
+SUITE_MAX_WORKERS = 16
+
+# Set by the plugin build. Installed wheels carry their own bitcode and linker, so
+# a leftover value points them at paths that do not exist here.
+UNSET_VARS = [
+    "ROCM_ROOT",
+    "HIP_DEVICE_LIB_PATH",
+    "JAX_ROCM_PLUGIN_INTERNAL_BITCODE_PATH",
+    "JAX_ROCM_PLUGIN_INTERNAL_LLD_PATH",
+]
+
+
+def jax_version_from_ref(jax_ref: str) -> str:
+    """The version a rocm-jax ref names, e.g. rocm-jaxlib-v0.11.0 -> 0.11.0."""
+    return jax_ref.removeprefix(JAX_REF_PREFIX) if jax_ref else ""
+
+
+def pytest_addopts(
+    jax_version: str,
+    amdgpu_family: str,
+    in_process_reruns: int,
+    debug: bool = False,
+) -> str:
+    """The PYTEST_ADDOPTS value for one configuration."""
+    opts = []
+    if in_process_reruns:
+        # Crashed workers are retried regardless of --only-rerun. Everything else
+        # is left to the fresh-process pass.
+        opts += [
+            "--reruns",
+            str(in_process_reruns),
+            "--reruns-delay",
+            "5",
+            "--only-rerun",
+            "INTERNAL",
+        ]
+
+    expression = keyword_expression(jax_version, amdgpu_family, debug)
+    if expression:
+        opts += ["-k", expression]
+
+    return shlex.join(opts)
+
+
+class SuiteEnvironmentError(Exception):
+    """The environment the suite runs under could not be read."""
+
+
+def env_section(script: str) -> str:
+    """The part of the suite script that only computes the environment."""
+    lines = script.splitlines()
+    starts = [i for i, line in enumerate(lines) if line.strip() == ENV_SECTION_START]
+    ends = [i for i, line in enumerate(lines) if line.strip() == ENV_SECTION_END]
+    if not starts or not ends or ends[-1] <= starts[0]:
+        return ""
+    return "\n".join(lines[starts[0] + 1 : ends[-1]])
+
+
+def _read_env_dump(path: Path) -> dict[str, str]:
+    entries = path.read_bytes().decode(errors="replace").split("\0")
+    pairs = (entry.partition("=") for entry in entries)
+    return {name: value for name, separator, value in pairs if separator}
+
+
+def suite_environment(jax_dir: Path, env: dict[str, str]) -> dict[str, str]:
+    """What the suite script's environment section exports.
+
+    Evaluated in a shell that dumps its environment on either side of the section,
+    so the result is what the section itself changed rather than a guess at which
+    names look interesting.
+
+    Raises:
+        SuiteEnvironmentError: if the section cannot be found or evaluated. This
+            fails the run by design, because the retry pass decides the result and
+            a wrong environment there would decide it wrongly.
+    """
+    section = env_section((jax_dir / RELATIVE_SUITE_SCRIPT).read_text())
+    if not section:
+        raise SuiteEnvironmentError(
+            f"{RELATIVE_SUITE_SCRIPT} has no section between '{ENV_SECTION_START}' and"
+            f" '{ENV_SECTION_END}'. The script most likely renamed them, so update"
+            " ENV_SECTION_START and ENV_SECTION_END to match it."
+        )
+    if not (jax_dir / RELATIVE_SUITE_ENV_FILE).exists():
+        raise SuiteEnvironmentError(
+            f"{RELATIVE_SUITE_ENV_FILE} is missing, and the section reads its defaults."
+            f" Is {jax_dir} a complete ROCm/jax checkout?"
+        )
+
+    with tempfile.TemporaryDirectory() as tmp:
+        before = Path(tmp) / "before"
+        after = Path(tmp) / "after"
+        dump = "env -0 > {}".format
+        # The section's own status, kept across the dump that follows it, which
+        # would otherwise be the status of the shell and always zero.
+        script = "\n".join(
+            [
+                f"source {shlex.quote(os.fspath(RELATIVE_SUITE_ENV_FILE))}",
+                dump(shlex.quote(os.fspath(before))),
+                section,
+                "section_status=$?",
+                dump(shlex.quote(os.fspath(after))),
+                "exit ${section_status}",
+            ]
+        )
+        proc = subprocess.run(
+            ["bash", "-c", script],
+            cwd=jax_dir,
+            env=env,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
+        if not (before.exists() and after.exists()):
+            raise SuiteEnvironmentError(
+                f"Evaluating the section of {RELATIVE_SUITE_SCRIPT} produced no environment:"
+                f" {proc.stderr.strip()}"
+            )
+        if proc.returncode != 0:
+            # It exports and queries devices, so a failure means some of what it
+            # exported was computed from a command that did not work.
+            raise SuiteEnvironmentError(
+                f"The section of {RELATIVE_SUITE_SCRIPT} exited"
+                f" {proc.returncode}, so the environment it exported is only"
+                f" partly what the suite will run under: {proc.stderr.strip()}"
+            )
+        baseline, exported = _read_env_dump(before), _read_env_dump(after)
+
+    return {
+        name: value
+        for name, value in exported.items()
+        if baseline.get(name) != value
+        and name.isupper()
+        and name not in SUITE_ENV_EXCLUDES
+    }
+
+
+def available_cpus() -> int:
+    """The CPUs this process may run on, which is the node's on a CI runner."""
+    if hasattr(os, "sched_getaffinity"):
+        return len(os.sched_getaffinity(0))
+    return os.cpu_count() or 1
+
+
+def requested_cpus(env: dict[str, str], available: int) -> int | None:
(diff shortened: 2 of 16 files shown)


marbre (commented):
As the workflow is named `test_linux_jax_wheels_partial.yml` we may want to rename to reflect that it now runs the full tests? Furthermore, what does this add to CI time wise or do we still need to restrict to a subset? Please link to a successful test run.

ScottTodd:
We will also likely want to have a few different set sizes so we can run tests on PRs as well as nightly releases.

Aside: this is another case where conventional commits are not adding any value to the project. This is not a "fix(ci)" change, this is sizable behavior change. Still waiting on reviews on https://github.com/ROCm/TheRock/pull/6114 to remove that from the PR bot.

erman-gurses (commented):
Can you please address the comments above?

gulsumgudukbay:
> We will also likely want to have a few different set sizes so we can run tests on PRs as well as nightly releases.
> 
> Aside: this is another case where conventional commits are not adding any value to the project. This is not a "fix(ci)" change, this is sizable behavior change. Still waiting on reviews on #6114 to remove that from the PR bot.

Regarding different set sizes comment, I opened this PR: https://github.com/ROCm/TheRock/pull/6753

erman-gurses:
Latest test: [Release Linux JAX Wheels (dev, 7.15.0a20260724, Multiarch) · ROCm/TheRock@75c3cd7](https://github.com/ROCm/TheRock/actions/runs/30115597223)

erman-gurses:
Next time, please format your branch name as described in here - will be easy to track it.
https://github.com/ROCm/TheRock/blob/610521238aeec9332b7bfa2f58a2f27d14c7846d/skills/therock_pr_bot/FAQ.md?plain=1#L26-L38

ScottTodd:
> Next time, please format your branch name

This is now documented in CONTRIBUTING.md at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#branch-naming-policy and enforced via a repository branch protection ruleset. Branches that do not match approved patterns will be rejected on push.

erman-gurses:
Do we have more current tests than this one? https://github.com/ROCm/TheRock/actions/runs/30186122185

gulsumgudukbay:
> Do we have more current tests than this one? https://github.com/ROCm/TheRock/actions/runs/30186122185

https://github.com/ROCm/TheRock/actions/runs/30395137799 this is the one from yesterday.

erman-gurses:
> > Do we have more current tests than this one? https://github.com/ROCm/TheRock/actions/runs/30186122185
> 
> you started a CI job yesterday.

Ok - but it is failed - did we fix those failures?

gulsumgudukbay:
> > > Do we have more current tests than this one? https://github.com/ROCm/TheRock/actions/runs/30186122185
> > 
> > 
> > you started a CI job yesterday.
> 
> I pasted that one in my message - that is failed - so did we fix those failures?

0.9.1 should be ignored. it is deprecated. for 0.10.2, I am checking with @tsrw2048

tsrw2048:
> > > > Do we have more current tests than this one? https://github.com/ROCm/TheRock/actions/runs/30186122185
> > > 
> > > 
> > > you started a CI job yesterday.
> > 
> > 
> > I pasted that one in my message - that is failed - so did we fix those failures?
> 
> 0.9.1 should be ignored. it is deprecated. for 0.10.2, I am checking with @tsrw2048

Looking into it now. The 0.10.2 branch shouldn't have failures so something in the configuration likely did not get moved over.

erman-gurses:
Latest Test: https://github.com/ROCm/TheRock/actions/runs/30497479491

erman-gurses (dismissed):
LGTM -  Assuming the Release and CI workflows are passing successfully. 
Please do merge commit and re-test (Release and CI workflows) after land this PR: https://github.com/ROCm/TheRock/pull/6708

erman-gurses:
> LGTM - Assuming the Release and CI workflows are passing successfully. Please do merge commit and re-test (Release and CI workflows) after land this PR: #6708

Running Release Build Test after merge commit:
https://github.com/ROCm/TheRock/actions/runs/30527360755

erman-gurses:
Another Test: https://github.com/ROCm/TheRock/actions/runs/30561401221

gulsumgudukbay:
https://github.com/ROCm/TheRock/actions/runs/30781739300 latest release run all green.

erman-gurses:
> https://github.com/ROCm/TheRock/actions/runs/30781739300 latest release run all green.

Do we have a full matrix test from this PR's branch?

gulsumgudukbay:
> > https://github.com/ROCm/TheRock/actions/runs/30781739300 latest release run all green.
> 
> Do we have a full matrix test from this PR's branch?

https://github.com/ROCm/TheRock/actions/runs/30825822253 - gfx950 run
I will start a gfx94X run once the PR check is done (since it is using gfx94X)

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -295,10 +336,52 @@ jobs:
...
+            logs/pytest_results_single.json logs/pytest_results_multi.json)"
+
+          if [[ -z "${retries}" ]]; then
+            echo "Nothing to retry."
+            exit 1
+          fi
+
+          # Word splitting is what turns the list into pytest arguments.
+          # shellcheck disable=SC2086
+          JAX_SKIP_SLOW_TESTS=true "python${PYTHON_VERSION}" -m pytest -n 0 --tb=short ${retries}
```
erman-gurses: Could this retry use the same environment setup as `run_pytest_rocm.sh`?

That script exports version-specific settings such as `XLA_PYTHON_CLIENT_ALLOCATOR` and, for some versions, additional XLA_FLAGS. Because the script is executed rather than sourced, those values do not persist for this direct pytest invocation.

The retry can therefore pass under a different configuration from the original failing run. Please share the environment setup between both executions or route the retry through the same script.

erman-gurses: Can you pass the failed node IDs as an array instead of expanding ${retries} unquoted?

list_failed_tests.py emits raw newline-separated pytest node IDs, and this expansion subjects them to shell word splitting and pathname expansion. Parametrized node IDs can contain spaces and glob characters, so a single node ID may be split or expanded before pytest receives it.

Since this fresh-process retry determines whether the job passes, that could cause an original failure not to be rerun or cause unintended tests to run. Please preserve every node ID as exactly one argument, for example by reading the helper output into a Bash array and invoking pytest with "${retry_nodeids[@]}".

erman-gurses (commented):
Please see my comments - I also have one thing to point out -  can we move the logic in the workflow file to a python script as a separated file then we can call it from the `test_linux_jax_wheels_partial.yml`

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: Fixed in ad527837d.

It was not hypothetical. Scanning the JSON reports from the last full matrix: 744 of the 64,692 node IDs in the 0.11.0/py3.13 job and 675 of the 55,346 in the 0.10.0/py3.12 job contain whitespace, for example

    tests/pallas/ops_test.py::OpsTest::test_array_reduce_add_bfloat16_(0, 1)
    tests/lax_control_flow_test.py::LaxControlFlowTest::testScanVmap_jit_scan=False_jit_f=False_in_axes=(0, 0)_impl=unroll1

If one of those failed, word splitting handed pytest two node IDs that do not exist, so the retry errored out instead of rerunning the real failure. It now reads the list into an array with `mapfile` and passes `"${retries[@]}"`.

I kept the command substitution rather than `mapfile < <(...)` on purpose, so `set -e` still sees a non-zero exit from `list_failed_tests.py` when its 40-failure cap is exceeded.

On pathname expansion specifically: 128 IDs per job contain `*`, from the shape-poly and export tests, but with neither `nullglob` nor `failglob` set an unmatched pattern is passed through literally, so those were never affected. The array covers them regardless.

gulsumgudukbay: Fixed in ad527837d.

The difference was real. What I verified against rocm/jax:

- `XLA_PYTHON_CLIENT_ALLOCATOR` is `platform` on `rocm-jaxlib-v0.10.0`, `default` on `v0.10.1` and `v0.10.2`, and `address` on `v0.11.0`. The retry set none, so it ran under JAX's default allocator, which differs from the suite on 0.10.0 and 0.11.0.
- On `v0.10.1` and `v0.10.2` the script appends `--xla_gpu_enable_cublaslt=false` when `rocm-sdk` is on PATH, which it always is in these images. The step-level `XLA_FLAGS` did not have it, so the retry would have taken the hipBLASLt gemm path while the suite ran without it. Since the failures being retried are rocBLAS gemm faults, that is exactly where a retry could have passed for the wrong reason.

Neither remedy you suggested is available, so I mirrored both values in the step `env`, keyed on `jax_ref`. Routing the retry through `run_pytest_rocm.sh` is not possible: it takes no arguments and hardcodes `tests` along with its marker filter, its deselects and `-n $num_processes`, so calling it again re-runs all ~28k tests instead of the handful that failed. Sourcing it for the environment is also out, since it sources `ci/utilities/install_wheels_locally.sh` and ends in `exit`, so sourcing would reinstall the wheels and then terminate the step. The script exports its own copies of both values, so the step env only takes effect for the retry and the suite pass is unchanged.

gulsumgudukbay:
> Please see my comments - I also have one thing to point out - can we move the logic in the workflow file to a python script as a separated file then we can call it from the `test_linux_jax_wheels_partial.yml`

@erman-gurses  The parts that parse or decide anything are already out of YAML: report parsing lives in `list_failed_tests.py`, and the ROCm-major plugin renaming lives in `patch_installed_jax_rocm_plugin_names.py`. What remains in the step is orchestration, roughly twenty lines: install the test deps with retries, run the suite script, and on failure re-run the failed node IDs in a fresh process.

A Python driver would not remove that work, it would wrap it. We would get subprocess calls around the same commands, environment plumbing that `env:` and the per-version expressions give us for free today, and one more layer between the step log and the command that actually failed, which matters when the thing being debugged is a library fault inside pytest.

For consistency, this step is 47 lines of shell, in line with the rest of the repo: `multi_arch_build_windows_pytorch_wheels.yml` has 59, `test_native_linux_packages_install.yml` 42, and `test_pytorch_wheels_full.yml` 39.

So I would keep the boundary where it is, Python for anything that parses or decides and shell for launching processes.

gulsumgudukbay:
@erman-gurses  after addressing your PR comments, these are the new runs:
https://github.com/ROCm/TheRock/actions/runs/30831999820 - gfx94X - All jobs succeeded.

erman-gurses:
> @erman-gurses after addressing your PR comments, these are the new runs: https://github.com/ROCm/TheRock/actions/runs/30831999820 - gfx94X - All jobs succeeded.

I think PR can be separated as:
PR 1 - Test fixes / stabilization
PR 2 - Test extension / full-suite enablement
CC: @ScottTodd /  @marbre

RELEASES.md
```diff
@@ -455,6 +455,22 @@ print(jax.devices())
...
+> ([jax-ml/jax#39634](https://github.com/jax-ml/jax/pull/39634)); until a `jaxlib`
+> carrying it ships, teach the installed copy about the new major:
+>
+> ```bash
+> python build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py \
+>     --plugin-package "jax_rocm${ROCM_MAJOR}_plugin"
+> ```
+>
+> The script edits the installed files in place, is idempotent, and is a no-op on ROCm 7
+> or once a fixed `jaxlib` is installed.
```
ScottTodd, line 472: I'm not thrilled by this, but it seems like an acceptable workaround while we wait for upstream. In the longer term I'd prefer if we weren't dependent on upstream to make the packages we distribute functional. That may require building packages ourselves or at least mirroring them into our index with a patching/repackaging step.

build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py
ScottTodd, line 1: I'd rather this file go in https://github.com/ROCm/TheRock/tree/main/external-builds/jax , similar to https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/windows_patch_fat_wheel.py. The `build_tools/github_actions/` folder is for code that we only expect to run on CI, but this file is referenced by `RELEASES.md` that users will see.

Then you could document it a bit in https://github.com/ROCm/TheRock/blob/main/external-builds/jax/README.md, similar to https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/README.md#bundling-pytorch-and-rocm-together-into-a-fat-wheel

build_tools/github_actions/list_failed_tests.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Lists the failed tests in a pytest JSON report, for a fresh-process retry.
+
+When a pytest-xdist worker's first complex-gemm or code-object load fails,
+the library caches the failure and every later call in that process fails the
+same way, so no in-process rerun can clear it. A new process can, and whether
+a failure survives that pass is what decides the job.
```
ScottTodd: "When a pytest-xdist worker's first complex-gemm or code-object load fails,"
huh? That is _very_ specific for a generic utility. Reword the comment? I'd say keep it generic to cover when _any_ type of test failure is sticky and affects future test cases, then maybe give these as specific examples that have been observed.

test_tools/list_pytest_failed_tests.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Lists the failed tests in a pytest JSON report, for a fresh-process retry.
```
ScottTodd, line 5: https://github.com/ROCm/TheRock/tree/main/test_tools may be a better spot for this file, and we could name it `list_pytest_failed_tests.py` as we also run tests via ctest.

@geomin12 maybe you have an opinion? This will currently just be used for JAX but I could see it being used for other components/frameworks if we find it a good home

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -295,10 +343,52 @@ jobs:
...
+          retry() { for _ in 1 2 3; do "$@" && return 0; sleep 20; done; return 1; }
+
+          retry python -m pip install uv~=0.11.2
+          retry python -m uv pip install -r jax/build/test-requirements.txt
+
+          # Use Python requirements lock file for correct version dependencies
+          lock="jax/build/requirements_lock_${PYTHON_VERSION/./_}.txt"
+          retry python -m uv pip install --upgrade -c "$lock" scipy pytest
+
+          retry python -m pip install pytest-json-report
```
ScottTodd: This is far too much code in this workflow file, please follow the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash

We can add a test runner script like https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/run_pytorch_tests.py or https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/run_pytorch_tests_full.py to include multiple steps.

Aim for workflow code that just calls scripts, like https://github.com/ROCm/TheRock/blob/58d2f9771438624c0e7b26e2dc09c66a84424352/.github/workflows/test_pytorch_wheels.yml#L193-L247

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -295,10 +343,52 @@ jobs:
...
+            logs/pytest_results_single.json logs/pytest_results_multi.json)"
+
+          if [[ -z "${failed}" ]]; then
+            echo "Nothing to retry."
+            exit 1
+          fi
+
+          # Parametrized nodeids contain spaces, so one line is one argument.
+          mapfile -t retries <<< "${failed}"
+          JAX_SKIP_SLOW_TESTS=true "python${PYTHON_VERSION}" -m pytest -n 0 --tb=short "${retries[@]}"
```
ScottTodd: This test retrying should ideally be part of the test runner itself and not be a part of the github actions workflow file. You can check https://github.com/pytorch/pytorch/blob/main/test/run_test.py for some ideas or chat with @ethanwee1 / @jithunnair-amd about their experiences with it.

See where we just set the `PYTORCH_NUM_PYTEST_RERUNS` env var in another workflow for example: https://github.com/ROCm/TheRock/blob/b070c6f38860cec6e28f37e1c4bb66c6002146e1/.github/workflows/test_pytorch_wheels_full.yml#L198

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -264,28 +264,76 @@ jobs:
...
+          # in-process rerun can stand in for.
+          #
+          # JAX 0.10.1 and 0.10.2 only: MIOpen selects convolution kernels
+          # nondeterministically on gfx94X, so the float32 conv tests flake.
+          # Deselect them. Keeps convert/conversion tests, which also
+          # match "conv". polymul convolves too, on fixed inputs, and returned
+          # a different wrong answer in each of the five jobs it has failed.
+          PYTEST_ADDOPTS: >-
+            --reruns 2 --reruns-delay 5 --only-rerun INTERNAL
+            ${{ contains(fromJSON('["rocm-jaxlib-v0.10.1", "rocm-jaxlib-v0.10.2"]'), inputs.jax_ref) && '-k "((not conv) or convert or conversion) and not sumpool and not minmaxpool and not toeplitz and not hankel and not polymul"' || '' }}
```
ScottTodd: Please let's move this into real code (not github actions workflows) like https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch/skip_tests. We need the ability to run tests locally with the same behavior as on CI, so all of this "if the workflow has these inputs set these environment variables" and "if this workflow step fails, run these three other commands piping outputs from command to script to tool" steps aren't feasible to run locally.

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -264,28 +264,76 @@ jobs:
...
+        run: |
+          echo "host=$(hostname)"
+          echo "ROCR_VISIBLE_DEVICES=${ROCR_VISIBLE_DEVICES:-unset}"
+          echo "GPU_DEVICE_ORDINAL=${GPU_DEVICE_ORDINAL:-unset}"
+          echo "HIP_VISIBLE_DEVICES=${HIP_VISIBLE_DEVICES:-unset}"
+
+          # The serial number is what identifies a recurring bad device across
+          # runs, since the runner and the device index change every time.
+          "${ROCM_ROOT}/bin/amd-smi" static --asic --board
+          "${ROCM_ROOT}/bin/amd-smi" process
```
ScottTodd: I'd rather we reuse https://github.com/ROCm/TheRock/blob/main/build_tools/print_driver_gpu_info.py for this. We can report those environment variables as part of that script too.

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -231,7 +231,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
 
-          pkg="rocm[libraries]"
+          pkg="rocm[libraries,devel]"
           if [ -n "${{ steps.expand.outputs.device_extras }}" ]; then
             pkg="${pkg%]},${{ steps.expand.outputs.device_extras }}]"
           fi
```
ScottTodd: Packages must work without `rocm[devel]` installed. That's what https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-multi-arch-jax-python-packages documents.

We've had the same discussion with pytorch. If files from a particular ROCm subproject are *essential* to a framework operating then they must be packaged as part of `rocm[libraries]`, not `[devel]`. Some non-essential tests (e.g. JIT compiling with devel header files) may depend on the `[devel]` package and should be filterable from test runs.

ScottTodd (changes requested):
This PR is doing several things all at once, it would be easier to review if it were split into smaller more focused PRs.

.github/workflows/test_linux_jax_wheels_partial.yml
ScottTodd: For the package install steps, https://github.com/ROCm/TheRock/blob/main/build_tools/setup_venv.py has retry support built in, may want to use that here as we do in other workflows

gulsumgudukbay:
> This PR is doing several things all at once, it would be easier to review if it were split into smaller more focused PRs.

Hi @ScottTodd  the PR is already split into three as per our discussion in the thread. 
three of the eleven commits stand alone, but the rest are consequences of the full suite itself, so they can't land ahead of it.

    main
     ├─ PR 1  ROCm 10 plugin naming + MIOpen Winograd   (943e06e, 70a42e1, 7e1de3f)
     │        fixes #7067
     │   └─ #6558 (rebased)  full JAX test suite + the flake handling it needs
     │        (0cc5cc6, e041e16, 0d19e7d, 2e32c72, 857cf95, 36d1141, ad52783)
     │          └─ #6753  test-size filtering (small/medium/large)
     └─ PR 3  release job naming  (f2a3b83) : independent, touches other files
 [ci(jax): teach the installed JAX stack about new ROCm major versions by gulsumgudukbay · Pull Reque…](https://github.com/ROCm/TheRock/pull/7084) this is the first one with ROCm 10 accommodation and Winograd environment variable. once this is merged, those 3 commits will be removed from [feat(ci): Run full JAX test suite by charleshofer · Pull Request #6558 · ROCm/TheRock](https://github.com/ROCm/TheRock/pull/6558) and this will be rebased. 
 
also [ci(jax): group the JAX release jobs by release instead of calling them all builds by gulsumgudukbay…](https://github.com/ROCm/TheRock/pull/7087) is the third PR I was talking about, which is just a cosmetic change for the release workflow.

build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py
gulsumgudukbay: The file is now external-builds/jax/patch_installed_jax_rocm_plugin_names.py and it is documented in external-builds/jax/README.md. It landed in #7084, so it is no longer part of this PR.

build_tools/github_actions/list_failed_tests.py
gulsumgudukbay: The comment is generic now. It describes any failure that is sticky and breaks the tests after it, and gives complex-gemm and code-object load failures as two examples we have seen.

test_tools/list_pytest_failed_tests.py
gulsumgudukbay: The file is now test_tools/list_pytest_failed_tests.py, with that name. It only reads pytest JSON reports, so it is not JAX specific.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: i added external-builds/jax/run_jax_tests.py as the test runner. The workflow step is now just a call to it with --jax-dir, --jax-ref and --amdgpu-family. The install steps also became scripts: install_jax_wheels.py and install_jax_test_requirements.py.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: I put both retries in the runner now. But please note that they are two different things:

- in-process reruns: pytest-rerunfailures, set by --in-process-reruns (env JAX_NUM_PYTEST_RERUNS), same idea as PYTORCH_NUM_PYTEST_RERUNS. It repeats a failed test in the same worker, which helps with a crashed worker or a one-off error.
- fresh-process retry: a new pytest process that runs only the failed tests. Some ROCm faults are sticky. After the first one, every later test in that process fails the same way, so a rerun inside it cannot help. Only a new process can tell a real failure from a broken process, so this pass decides the job result. --no-retries turns both off and runs the suite the way the script does.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: now i made the skip lists data files in external-builds/jax/skip_tests/, one per JAX version, like external-builds/pytorch/skip_tests. The runner builds the pytest -k filter from them, so there is no "if the workflow has this input, set that variable" logic left.

The same command runs locally and on CI:

  python external-builds/jax/run_jax_tests.py --jax-dir jax --jax-version 0.10.2 --amdgpu-family gfx94X-dcgpu

--dry-run prints the environment and the commands without running them, and
--debug runs only the tests the skip lists would skip.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: The workflow now calls build_tools/print_driver_gpu_info.py instead of the inline amd-smi commands.
I did not add the environment variables to that script, because I did not want to change a shared script from this PR. The runner prints ROCR_VISIBLE_DEVICES, HIP_VISIBLE_DEVICES and GPU_DEVICE_ORDINAL itself, next to the driver output. If you prefer them inside print_driver_gpu_info.py, please let me know

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: the job installs rocm[libraries] plus the device extras only, no [devel].
I also checked it on MI355X with ROCm 10.1.0.dev0: after removing rocm-sdk-devel from the venv, matmul, cholesky, FFT and the MIOpen convolution tests all still pass.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: > For the package install steps, https://github.com/ROCm/TheRock/blob/main/build_tools/setup_venv.py has retry support built in, may want to use that here as we do in other workflows

The ROCm packages are installed with build_tools/setup_venv.py, so they get its retry support. The JAX wheels and the test requirements go through install_jax_wheels.py and install_jax_test_requirements.py, which also retry.

gulsumgudukbay:
https://github.com/ROCm/TheRock/actions/runs/31042212819 full green release matrix

external-builds/jax/run_jax_tests.py
```diff
@@ -0,0 +1,442 @@
...
+            f"::error::{len(failed)} failures is more than --max-retry-tests"
+            f" {args.max_retry_tests}, so nothing will be retried"
+        )
+        return returncode
+
+    # This pass decides the result: a sticky fault only clears in a new process.
+    # It runs under the suite's own environment, read above, so it takes the same
+    # path as the run it is checking.
+    log(f"=== Retrying {len(failed)} failure(s) in a fresh process")
+    return run_command(retry_command(failed), jax_dir, env)
```
erman-gurses, line 650: Could we check the JSON reports’ top-level pytest `exitcode` before allowing the fresh-process retry to replace the original suite result?

`collect_failed()` currently preserves only individual failed/error node IDs. It discards whether the corresponding pytest session completed normally. A report may contain completed test results - and potentially failed node IDs - while the overall session later exits because of an interruption or internal error.

In that case, the fresh retry could pass the recorded node IDs and this function would return 0, even though the original pytest session may have stopped before running all selected tests.

Can you only allow the retry to decide the result when all newly generated reports have `exitcode` 0 or 1, with at least one report having `exitcode == 1` for ordinary test failures? An exit code greater than 1, a malformed report, or no newly generated reports should preserve the original suite failure. Existing reports should also be removed before starting the suite so a local rerun cannot consume stale results.

erman-gurses (commented):
Please see my comment:

external-builds/jax/run_jax_tests.py
gulsumgudukbay: The fresh-process retry now reads each JSON report's top-level pytest exit code and only decides the result when every report ended 0 or 1 with at least one 1; a higher code, a missing or unparseable one, or no reports leaves the suite failure standing. Reports are cleared before the suite starts so a rerun can't consume stale ones. Checked against pytest-json-report: an ordinary failure and a crashed xdist worker both report 1, so the sticky ROCm faults the retry exists for are still covered, while an interrupted session reports 2 and is now red.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
```diff
@@ -227,42 +229,21 @@ jobs:
...
-          fi
-
-          python -m pip install --pre \
+          python build_tools/setup_venv.py "${VIRTUAL_ENV}" \
+            --packages "${ROCM_PACKAGES}" \
             --find-links "${ROCM_PACKAGE_FIND_LINKS_URL}" \
-            "${pkg}==${{ inputs.rocm_version }}"
-
-          echo "ROCM_ROOT=$(rocm-sdk path --root)" >> "$GITHUB_ENV"
+            --pre
```
ScottTodd: nit: I don't think `--pre` should be needed if you match exactly `==${{ inputs.rocm_version }}`. Fine to keep it though.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
```diff
@@ -271,41 +252,89 @@ jobs:
...
-            --xla_gpu_enable_nccl_comm_splitting=false
-            --xla_gpu_enable_command_buffer=
-
+      # The device serial numbers it prints identify a recurring bad device
+      # across runs, since the runner and the device index change every time.
+      - name: Driver / GPU sanity check
+        continue-on-error: true
+        timeout-minutes: 3
+        run: |
+          python build_tools/print_driver_gpu_info.py
```
ScottTodd, line 259: Fine to start like this, but we should not use `continue-on-error` for sanity checks. If a runner is so unhealthy that sanity checks fail, it isn't worth trying to run real tests.

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -271,41 +252,89 @@ jobs:
...
+          grep Cpus_allowed_list /proc/self/status
+          for f in /sys/fs/cgroup/cpu.max /sys/fs/cgroup/memory.max \
+            /sys/fs/cgroup/cpu/cpu.cfs_quota_us \
+            /sys/fs/cgroup/memory/memory.limit_in_bytes; do
+            if [ -r "${f}" ]; then echo "${f}=$(cat "${f}")"; fi
+          done
+
+      # numpy and scipy bundle OpenBLAS, the one threaded library the suite
+      # pulls in. It starts a thread per core it can see, in every worker.
+      - name: Cap OpenBLAS threads
```
ScottTodd: Fine to start with this in the workflow, but this should aim for either:
* A script that users can run on their own machines/nodes, e.g. as part of the test runner itself
* Something built into the test runner infrastructure itself so individual workflows don't need to stabilize the environment themselves

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -271,41 +252,89 @@ jobs:
...
-          pytest jax/tests/scipy_stats_test.py -q --log-cli-level=INFO
+      # The runner owns the test environment, the skip lists and both layers of
+      # retry, so that the same command reproduces this job locally. See
+      # external-builds/jax/run_jax_tests.py.
+      - name: Run JAX tests
+        # The suite runs 35-70 min across versions, so 60 left no headroom: one
+        # job reached 100% of the tests and was still killed by the timeout.
+        timeout-minutes: 120
+        run: |
+          python external-builds/jax/run_jax_tests.py \
```
ScottTodd: nit: no need to mention `external-builds/jax/run_jax_tests.py` in the comment if that's exactly the script that the step is running :P you could say "see **the comments in** [script name]"

external-builds/jax/run_jax_tests.py
```diff
@@ -0,0 +1,508 @@
...
+from list_pytest_failed_tests import failed_in_report
+from skip_tests.create_skip_tests import keyword_expression
+
+JAX_REF_PREFIX = "rocm-jaxlib-v"
+
+SUITE_SCRIPT = pathlib.Path("ci") / "run_pytest_rocm.sh"
+
+# Sourced by the suite script for the JAXCI_* defaults the section below reads,
+# and safe to source here too: it only assigns variables.
+SUITE_ENV_FILE = pathlib.Path("ci") / "envs" / "default.env"
```
ScottTodd: On the surface, these do not look portable across different working directories: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#dont-make-assumptions-about-the-current-working-directory

Code below runs `jax_dir / SUITE_SCRIPT`, so how about adding a comment here explaining that these paths are _relative to the jax directory_, and/or renaming the variables to `RELATIVE_SUITE_SCRIPT` and `RELATIVE_SUITE_ENV_FILE`?

external-builds/jax/run_jax_tests.py
```diff
@@ -0,0 +1,508 @@
...
+
+    # Run only the tests the skip list would have skipped.
+    python run_jax_tests.py --jax-dir jax --jax-version 0.10.2 --debug
+"""
+
+import argparse
+import dataclasses
+import json
+import os
+import pathlib
```
ScottTodd: nit: as all uses in this file are `pathlib.Path`, you can do `from pathlib import Path` here and trim characters throughout the file. This is also the canonical recommendation from https://docs.python.org/3/library/pathlib.html#basic-use

Image

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#use-pathlib-for-filesystem-paths

external-builds/jax/run_jax_tests.py
```diff
@@ -0,0 +1,508 @@
...
+            env=env,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
+        if not (before.exists() and after.exists()):
+            raise SuiteEnvironmentError(
+                f"Evaluating the section of {SUITE_SCRIPT} produced no environment:"
+                f" {proc.stderr.strip()}"
+            )
```
ScottTodd, line 265: Do you want `check=True` here also? What if the command produces the before/after sections but returns a failure error code?

.github/workflows/test_multi_arch_linux_jax_wheels.yml
ScottTodd, line 1: Is `_partial` still a good name for this file, if it runs the "full JAX test suite"?

ScottTodd (approved):
A few style comments that would be good to address, as well as a few areas for later refactoring as CI evolves, but overall LGTM to get this running.

erman-gurses (approved):
Thanks for addressing comments - Please address Scott's comments as well then share a clean run results (assuming we have all green)

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Deleted, you're right that it's redundant, since the spec pins an exact .dev0 version and the rocm metapackage pins its children to that same version, so pip already allows prereleases across the whole graph.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Dropped, so the driver check now fails the job rather than letting 120 minutes of tests run on a runner that already failed its sanity check

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: Moved into run_jax_tests.py, which now reports the pod's allocation and pins the run to it, so the suite script and the retry pass both inherit it and --cpus N reproduces the same behavior on any machine.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: fixed it, the comment now says "see the comments in the script" instead of repeating the path the step already runs.

external-builds/jax/run_jax_tests.py
gulsumgudukbay: Renamed to RELATIVE_SUITE_SCRIPT and RELATIVE_SUITE_ENV_FILE, with a comment noting they are relative to the JAX checkout.

external-builds/jax/run_jax_tests.py
gulsumgudukbay: Done, and `pathlib.` no longer appears anywhere in the file

external-builds/jax/run_jax_tests.py
gulsumgudukbay: thanks, however check=True alone would not have caught it btw. the env dump runs after the section, so the shell's status is always the dump's, so the section's own status is now saved and re-raised as the script's exit, and a non-zero one aborts with SuiteEnvironmentError rather than letting the retry pass run under a half-built environment.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Renamed to test_multi_arch_linux_jax_wheels.yml to match the workflow's own name:, which also frees _partial for the actual partial suite coming in #6753.

gulsumgudukbay:
https://github.com/ROCm/TheRock/actions/runs/31151324282 all green release workflow

Issue #6303, opened 2026-07-01
[ASAN] Document how to install and use ASan packages


ScottTodd:
- [ ] Add install instructions to https://github.com/ROCm/TheRock/blob/main/RELEASES.md
    - [ ] How to tell what a given install is (will the versions or directory names be different?)
    - [ ] How to switch between asan and non-asan packages (e.g. run production workload under non-asan then switch to asan to debug issues)
    - [ ] How to preload libraries etc. as needed
    - [ ] What is/isn't supported (native supported, python not?)
    - [ ] Information about what ASan is and what to expect (instrumented host code? instrumented device code? expected performance overhead? expected binary size differences?)
- [ ] Add status badge to https://github.com/ROCm/TheRock/blob/main/README.md#nightly-release-status
- [ ] Update https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md as needed

PR #6559, opened 2026-07-15
docs: Add ASAN-instrumented library install and usage instructions

nunnikri:
Add ASAN-instrumented library install and usage instructions to RELEASES.md

- Add ASAN build status badge to Multi-arch release status table
- Add Installing ASAN-instrumented libraries section covering:
  - Native DEB/RPM package install via packages-asan repo
  - Tarball install with GPU-family-specific and multi-arch options
- Add Using ASAN-instrumented libraries section covering:
  - HSA_XNACK=1 for device-side instrumentation
  - Locating the ASAN runtime directory via amdclang --print-file-name
  - LD_LIBRARY_PATH (recommended) and LD_PRELOAD options

JIRA ID: https://amd-hub.atlassian.net/browse/ROCM-24584
Github Issue : https://github.com/ROCm/TheRock/issues/6303


RELEASES.md
```diff
@@ -86,6 +88,7 @@ Key differences from
 | PyTorch packages - Windows build/test  | [![Multi-Arch Release Windows PyTorch Wheels](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_windows_pytorch_wheels.yml/badge.svg)](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_windows_pytorch_wheels.yml) |
 | PyTorch packages - full tests          | [![Test PyTorch Wheels (Full Suite)](https://github.com/ROCm/rockrel/actions/workflows/test_pytorch_wheels_full.yml/badge.svg)](https://github.com/ROCm/rockrel/actions/workflows/test_pytorch_wheels_full.yml)                                            |
 | JAX packages - Linux build/test        | [![Multi-Arch Release Linux JAX Wheels](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_linux_jax_wheels.yml/badge.svg)](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_linux_jax_wheels.yml)                   |
+| ASAN instrumented build                | [![Multi-Arch Release ASAN](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_asan.yml/badge.svg)](https://github.com/ROCm/rockrel/actions/workflows/multi_arch_release_asan.yml)                                                       |
```
ScottTodd: We could move this up to just under "Build ROCm artifacts/tarballs/packages". The ASan release workflow also triggers https://github.com/ROCm/rockrel/actions/workflows/test_artifacts.yml and other workflows, so it's fine to put it before "Test ROCm artifacts"

ScottTodd: Let's keep this table in sync with the one in README.md: https://github.com/ROCm/TheRock/blob/a20ab791d74c94354704b791a85343babb83409b/README.md?plain=1#L28-L40

(I sorta like having it in both places but we could consolidate...)

RELEASES.md
```diff
@@ -40,6 +40,8 @@ Table of contents:
       <!-- - [Using ROCm from WSL](#using-rocm-from-wsl) -->
     <!-- TODO: native Windows packages -->
   - [Installing multi-arch tarballs](#installing-multi-arch-tarballs)
+  - [Installing ASAN-instrumented libraries](#installing-asan-instrumented-libraries)
+  - [Using ASAN-instrumented libraries](#using-asan-instrumented-libraries)
```
ScottTodd: We could nest "Using ASan" under "Installing ASan", similar to how "Using multi-arch ROCm Python packages" is nested under "Installing multi-arch ROCm Python packages". I see that there are other subsections though, so maybe not. Thoughts?

```suggestion
  - [Installing ASAN-instrumented libraries](#installing-asan-instrumented-libraries)
    - [Using ASAN-instrumented libraries](#using-asan-instrumented-libraries)
```

I guess we could also re-categorize as
* Python packages
  * Install
  * Usage
* Native packages
  * Linux
    * Install
    * Usage

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
...
+sanitizer runtime to service the shadow-memory accesses that instrumentation
+generates on the device.
+
+#### Available ASAN packages
+
+ASAN packages are currently published for the following GPU families:
+
+| GPU Family | Example targets                          |
+| ---------- | ---------------------------------------- |
+| gfx94X     | gfx942 (MI300X / MI300A), gfx941, gfx940 |
```
ScottTodd: Do we have gfx941 and gfx940? I thought only gfx942?

https://github.com/ROCm/TheRock/blob/a20ab791d74c94354704b791a85343babb83409b/cmake/therock_amdgpu_targets.cmake#L92-L96

https://llvm.org/docs/AMDGPUUsage.html

Image

```suggestion
| gfx94X     | gfx942 (MI300X / MI300A) |
```

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
...
+libraries.
+
+Device-side (GPU) ASAN additionally relies on the **XNACK**
+(retry-on-page-fault) capability of supported AMD GPUs, which allows the
+sanitizer runtime to service the shadow-memory accesses that instrumentation
+generates on the device.
+
+#### Available ASAN packages
+
+ASAN packages are currently published for the following GPU families:
```
ScottTodd: Let's directly clarify "Linux only, not Windows" somewhere in this section.

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
 > See also [this issue](https://github.com/ROCm/TheRock/issues/1658) discussing
 > relevant environment variables.
 
+### Installing ASAN-instrumented libraries
+
+ROCm ships a set of libraries built with
+[AddressSanitizer (ASAN)](https://clang.llvm.org/docs/AddressSanitizer.html)
```
ScottTodd: Technically the capitalization from that page is `ASan`, not `ASAN`. Should we be consistent with that in our documentation here too? (I think we should)

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
...
+#### Prerequisites
+
+- **Linux kernel 5.6 or higher** with Heterogeneous Memory Management (HMM)
+  support for device-side ASAN. Check with `uname -r`.
+- An **XNACK-capable GPU** for device-side ASAN. Host-side ASAN works without
+  XNACK, but device-side instrumentation requires it.
+- The ROCm compiler toolchain (`amdclang` / `amdclang++`) from your ROCm
+  installation.
+- Sufficient free disk space — the instrumented libraries are larger than their
+  uninstrumented counterparts.
```
ScottTodd, line 664: Can we quantify this a bit? ballpark sizes? 10GB? 100GB?

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
 > See also [this issue](https://github.com/ROCm/TheRock/issues/1658) discussing
 > relevant environment variables.
 
+### Installing ASAN-instrumented libraries
+
+ROCm ships a set of libraries built with
+[AddressSanitizer (ASAN)](https://clang.llvm.org/docs/AddressSanitizer.html)
+instrumentation. These instrumented libraries let you run your HIP or ROCm
```
ScottTodd: How about linking to https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md here somewhere?

Similar to how we link to https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md from the ROCm Python package release docs: https://github.com/ROCm/TheRock/blob/501323f456f3ed6fbde5395b7574d522739b5352/RELEASES.md?plain=1#L123-L133

I also see https://rocm.docs.amd.com/en/docs-6.0.0/conceptual/using-gpu-sanitizer.html when I search "rocm asan", but that page doesn't exist for newer releases? If there is a recent page on rocm.docs.amd.com we should link it from here (and sanitizers.md) too.

RELEASES.md
```diff
@@ -619,6 +622,240 @@ ls install/.kpack/
...
+>     └── ...   # additional instrumented ROCm libraries
+> ```
+
+### Using ASAN-instrumented libraries
+
+Once installed (via native packages or tarball), set `ROCM_ASAN_PATH` to the
+root of your ASAN install tree, then configure the remaining environment
+variables in the shell from which you launch your application. Your production
+ROCm installation is left untouched; unset the variables to return to normal
+operation.
```
ScottTodd: This is a lot of text for how to use ASan. Could be moved to a separate page like https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md to keep the RELEASES.md page focused more on how to install the different release types.

I'd say it's borderline... the whole doc is pretty large at this point.

ScottTodd (approved):
Overall LGTM, thanks. A few comments I'd like addressed then okay.

RELEASES.md
raramakr: Usage moved to sanitizers.md

RELEASES.md
raramakr: Done

raramakr: Done

RELEASES.md
raramakr: Removed gfx941, gfx940

RELEASES.md
raramakr: This is not addressed

RELEASES.md
raramakr: Done

RELEASES.md
raramakr: Done

raramakr:
@ScottTodd, most of the review comments have been addressed, except for two or three. Hopefully, we can merge this now.

RELEASES.md
ScottTodd: looking briefly at https://rocm.nightlies.amd.com/, asan tarballs are ~10x as large as non-asan tarballs?
2.8GB vs 28GB
Image

For later, I'd like to understand more what is contributing to that size increase... is it debug symbols? If it's debug symbols, we have an RFC for debug symbol generation and packaging: https://github.com/ROCm/TheRock/pull/6034. We could ship smaller binaries and provide a debug symbol server, or at least separate out the symbols if the packages are functional without them...

ScottTodd (approved):

docs/development/sanitizers.md
```diff
@@ -103,6 +103,82 @@ Handling sanitizer builds can be very complicated because of the need to fit all
...
+Once you have ASan-instrumented ROCm libraries installed (via
+[native packages or tarball](../../RELEASES.md#installing-asan-instrumented-libraries)),
+set `ROCM_ASAN_PATH` to the root of your ASan install tree, then configure the
+remaining environment variables in the shell from which you launch your
+application. Your production ROCm installation is left untouched; unset the
+variables to return to normal operation.
+
+```bash
+# DEB / RPM install:
+export ROCM_ASAN_PATH=/opt/rocm/core
```
geomin12, line 117: if we want to avoid production installation, should we route this to another example route?

docs/development/sanitizers.md
```diff
@@ -103,6 +103,82 @@ Handling sanitizer builds can be very complicated because of the need to fit all
...
+
+Without this, only host-side (CPU) errors will be detected.
+
+### Step 2: Locate the ASan runtime directory
+
+The host ASan runtime is used here only to resolve the directory where all ASan
+runtime libraries are installed:
+
+```bash
+ASAN_LIB_PATH=$(amdclang --print-file-name=libclang_rt.asan-x86_64.so)
```
geomin12, line 138: should we ensure that amdclang is the one from ROCM_ASAN_PATH? what if this gets pulled from the system

docs/development/sanitizers.md
```diff
@@ -103,6 +103,82 @@ Handling sanitizer builds can be very complicated because of the need to fit all
...
+## Using ASan-Instrumented Libraries
+
+Once you have ASan-instrumented ROCm libraries installed (via
+[native packages or tarball](../../RELEASES.md#installing-asan-instrumented-libraries)),
+set `ROCM_ASAN_PATH` to the root of your ASan install tree, then configure the
+remaining environment variables in the shell from which you launch your
+application. Your production ROCm installation is left untouched; unset the
+variables to return to normal operation.
+
+```bash
```
geomin12, line 115: maybe docker container step addition? to avoid overlap? it always gets messy

RELEASES.md
nunnikri: Yes @ScottTodd debug symbols are the main reason for size increase. sanitize option would would inject some additional code and will need much higher runtime memory requirements. But the actual disk space increase will be much less than the debug symbols. If we decide to move all debug symbols to a different file and host a symbol server that will be the best.

docs/development/sanitizers.md
nunnikri: Ya thats a product decision to be made. @saadrahim do we have to revisit this for next ROCm release?

docs/development/sanitizers.md
nunnikri: System provided clang/llvm wont have amd prefix. Only ROCm released clang has amdprefix. But if there are multiple ROCm installation available it might be selecting the last installed one or the one user has set as default.

PR #2851, opened 2026-01-12
Enable Linux FBGEMM_GENAI by default for all pytorch versions 

araravik-psd:
With latest changes to upstream we see that pytorch builds are passing with FBGEMM_GENAI enabled irrespective of pytorch version.

This PR aims at enabling FBGEMM_GENAI by default on Linux. Disabling FBGEMM_GENAI can also be achieved by setting enable_pytorch_fbgemm_genai_linux to False which would disable building Pytorch with FBGEMM_GENAI.

Related PRs:

https://github.com/pytorch/pytorch/pull/168377
https://github.com/jagadish-amd/pytorch/commit/bfba6b4ea6e98b924ce18de2dc527bf514bd68da
https://github.com/jagadish-amd/pytorch/commit/0356369f4e4db3f25aa0cbe3d5afda6e4783ed65

Test with passing builds enabling FBGEMM_GENAI by default:

https://github.com/ROCm/TheRock/actions/runs/20903783506

PR #6577, opened 2026-07-15
build(pytorch): Remove 2.9/fbgemm_genai flag, no torch nightly for prerelease

HereThereBeDragons:
## Summary

Removes PyTorch `release/2.9` from build matrices, stops building PyTorch nightly for prereleases, and removes the custom `fbgemm_genai` flag in favor of PyTorch's own defaults.

### 1. Remove building of PyTorch `release/2.9`

Drops `release/2.9` from the CI and release build matrices and deletes the associated 2.9 skip-test file. Also changes the PyTorch full test suite to run on Windows for **all** PyTorch versions instead of gating on 2.9 only.

### 2. Skip PyTorch nightly for `release_type == prerelease`

For prerelease builds we no longer build the PyTorch `nightly` ref. This reduces the extra churn needed on release branches to disable building torch nightly.

### 3. Remove the `fbgemm_genai` flag — take the default from PyTorch

We no longer need this logic; we can trust PyTorch to set the right default. Currently `fbgemm_genai` (renamed to MSLK) is only supported on gfx942 and gfx950, both Linux-only, and PyTorch [auto-enables it based on the listed architectures](https://github.com/pytorch/pytorch/blob/ffc572d6aba76129928e0caaa96a93bd0585f903/CMakeLists.txt#L966-L968).

Our own flag had gotten inconsistent:
- Built `fbgemm_genai` for 2.9 and >= 2.11.
- 2.10 was excluded because we explicitly disabled the flag.
- 2.11 was "accidentally" included: `fbgemm_genai` was renamed to MSLK upstream, so our flag no longer applied and PyTorch's default took over.

Removing our flag makes behavior consistent and correct across versions.

Supersedes #2851.


HereThereBeDragons:
changes related to the full test suite and mslk/fbgemm_genai were aligned with @ethanwee1 and @jithunnair-amd

ethanwee1 (approved):
LGTM!

HereThereBeDragons:
test for the config matrix:

prerelease
```
python build_tools/github_actions/configure_pytorch_release_matrix.py \
  --release-type prerelease --platform linux --amdgpu-families "gfx94X-dcgpu" 2>/dev/null \
  | sed -n '/^{/,/^}/p' | jq -r '.pytorch_matrix | fromjson | .[].pytorch_git_ref' | sort -u
```
output:

```
release/2.10
release/2.11
release/2.12
```


nightly
```
python build_tools/github_actions/configure_pytorch_release_matrix.py  \
 --release-type nightly --platform linux --amdgpu-families "gfx94X-dcgpu" 2>/dev/null \
   | sed -n '/^{/,/^}/p' | jq -r '.pytorch_matrix | fromjson | .[].pytorch_git_ref' | sort -u
```

output:

```
nightly
release/2.10
release/2.11
release/2.12
```

build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -35,14 +35,12 @@
 #      built for nightly but not published to stable.
```
ScottTodd: Can you update the comment here now that this code was added below?
```python
    if release_type == "prerelease":
        return [ref for ref in RELEASE_PYTORCH_REFS[platform] if ref != "nightly"]
```
I was thinking we could have
```python
RELEASE_PYTORCH_REFS = { ... } 
RELEASE_STABLE_PYTORCH_REFS = { ... }
```
to make it very explicit (no needing to parse through the code to know when there are changes), but this is okay too

build_tools/github_actions/configure_pytorch_release_matrix.py
```diff
@@ -35,14 +35,12 @@
...
 RELEASE_PYTORCH_REFS = {
     "linux": [
-        "release/2.9",
         "release/2.10",
         "release/2.11",
         "release/2.12",
         "nightly",
```
ScottTodd: If we keep this for all release types but filter down below, how about at least adding a comment?

```suggestion
        "nightly",  # filtered out for `release_type == "prerelease"`
```

build_tools/github_actions/configure_pytorch_release_matrix.py
ScottTodd: Same here
```suggestion
        "nightly",  # filtered out for `release_type == "prerelease"`
```

ScottTodd (approved):

build_tools/github_actions/configure_pytorch_release_matrix.py
jithunnair-amd, line 69: @pragupta We should be able to remove this exclusion for nightly and release/2.12 now? Maybe in a separate PR?
cc @ethanwee1 @HereThereBeDragons

jithunnair-amd (approved):

build_tools/github_actions/configure_pytorch_release_matrix.py
HereThereBeDragons: agree might be easier to read. please have another look

ScottTodd (approved):

PR #6589, opened 2026-07-16
Include libhipcxx dev artifact in rocm-sdk-devel package

### LINKED ISSUE #6588: libhipcxx headers not included in Python packages
libhipcxx is a header-only library and the header files should be included in the rocm-sdk-devel wheel.

marbre:
## Motivation

libhipcxx is a header-only library and produces no runtime artifact, so it must be explicitly listed in `addl_artifact_names` to appear in the kpack-split devel package.

## Technical Details

This patch does not touch the legacy packages which already diverged from the kpack-split enabled multi-arch packages.

Closes #6588.

## Submission Checklist

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -384,11 +384,16 @@ def _run_kpack_split(
     devel = PopulatedDistPackage(params, logical_name="devel", target_family=None)
     devel.populate_devel_files(
         addl_artifact_names=[
+            # Header-only libraries not included in runtime packages.
             "prim",
             "rocwmma",
+            "libhipcxx",
+            # Third party dependencies needed by hipDNN consumers.
             "flatbuffers",
             "nlohmann-json",
+            # rocshmem only provides a static library.
             "rocshmem",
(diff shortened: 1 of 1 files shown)


marbre:
Overriding policy here as unit tests don't seem applicable here.

obersteiner (approved):
LGTM thanks!

marbre:
Confirmed that the headers are now included in `rocm-sdk-devel`. CI failures are unrelated.

build_tools/build_python_packages.py
```diff
@@ -384,11 +384,16 @@ def _run_kpack_split(
     devel = PopulatedDistPackage(params, logical_name="devel", target_family=None)
     devel.populate_devel_files(
         addl_artifact_names=[
+            # Header-only libraries not included in runtime packages.
             "prim",
             "rocwmma",
+            "libhipcxx",
```
saadrahim, line 390: Are we sure about libhipcxx? I thought this library had a .so artifact? @RobsonRLemos

Issue #6598, opened 2026-07-16
ROCm Python Package Indexes


marbre:
## Overview

The new stream-subdomain repository structure needs pip-compatible Python
package indexes that aggregate packages owned by separate product areas such as
`core`, `pytorch`, `jax`, wheel-producing extras, and future product groups.

The public central Python index names are:

```text
/rocm/whl/
/rocm/whl-next/
```

Release streams are selected by hostname, not by path components. Examples:

```text
https://dev.repo.amd.com/rocm/whl/
https://nightly.repo.amd.com/rocm/whl/
https://nightly.repo.amd.com/rocm/whl-next/
https://weekly.repo.amd.com/rocm/whl/
https://rc.repo.amd.com/rocm/whl/
https://stable.repo.amd.com/rocm/whl/
```

The bare-domain `https://repo.amd.com/rocm/` path is the legacy singleton
folder and is unrelated to the new stream-subdomain `/rocm/` tree.

## Goals

- Provide stable pip-compatible aggregate indexes:

  ```text
  https://nightly.repo.amd.com/rocm/whl/
  https://nightly.repo.amd.com/rocm/whl-next/
  ```

- Ensure direct project URLs work:

  ```text
  .../rocm/whl/torch/
  .../rocm/whl/rocm-sdk-core/
  .../rocm/whl-next/torch/
  ```

- Keep package ownership separated by product area.
- Avoid teaching `manage.py` about CloudFront routing.
- Avoid duplicating per-package index pages under aggregate roots.
- Keep legacy bare-domain `/rocm/...` indexes unchanged.

## Proposed Design

Product publishers store packages in structured package directories. The
central index name is mirrored under the product-local path:

```text
/rocm/core/whl/rocm-sdk-core/
  index.html
  rocm_sdk_core-...whl

/rocm/pytorch/whl/torch/
  index.html
  torch-...whl

/rocm/pytorch/whl-next/torch/
  index.html
  torch-...whl
```

`manage.py` or rather a new `manage_structured.py` generates product-local package indexes only. For the new
structured layout, it needs a package-directory mode that discovers files under
`<product>/<index>/<normalized-package>/` and writes
`<product>/<index>/<normalized-package>/index.html`.

A separate aggregate generator uploads only the root project-list pages:

```text
/rocm/whl/index.html
/rocm/whl-next/index.html
```

Each root page lists canonical project links:

```html
<a href="rocm-sdk-core/">rocm-sdk-core</a><br/>
[...]

PR #6601, opened 2026-07-16
Add structured product-local PEP 503 index generator

### LINKED ISSUE #6600: Add structured layout support to (forked) manage.py
Add a generator that produces pip-compatible PEP 503 indexes for the new stream-subdomain structured layout:

```
<product>/<index>/<normalized-package>/index.html
<product>/<index>/index.html            (optional product root listing)
```

`<index>` is `whl` or `whl-next`. The generator must not know anything about CloudFront routing, aggregate `/rocm/whl/` roots, S3 bucket -> URL mapping, or release streams (streams are selected by hostname, not path).

### Approach: new self-contained file, not an extension of flat `manage.py`

Flat mode (`manage.py`) is to be deprecated and later dropped. We do not extend or refactor it as extending would either force a refactor of flat's guts or bolt a parallel branch into a file we intend to delete. Instead add a new
self-contained file and delete the old `manage.py` (+ its tests) once the structured tree is live.

New file: `build_tools/third_party/s3_management/manage_structured.py` (+ `tests/manage_structured_test.py`).

marbre:
## Motivation
The new stream-subdomain repository structure requires to publish Python
packages under product-local package directories:

    <product>/<index>/<normalized-package>/<filename>

where <index> is whl or whl-next and the release stream is selected by
the bucket, not by path. The legacy generator `manage.py` only supports
to index a flat directory and assumes a single whl-mutli-arch or rather
per GPU family: package identity comes from the wheel filename, package
pages link with ../filename, discovery accepts only objects directly in
the prefix, and an allow-list plus version threshold gate which packages
appear. None of those assumptions fit to new the structured layout.

* Progress on #6598.
* Closes #6600.

## Technical Details
Add a new, self-contained generator (`manage_structured.py`) rather than
extending flat `manage.py`, which is deprecated and will be dropped once
the structured tree is live.

Key design decisions:

* Directory name is authoritative for package identity (PEP 503-
  normalized). Package name extraction uses `packaging.utils`
  (`parse_wheel_filename` / `parse_sdist_filename`) rather than a
  first-hyphen heuristic, which is invalid for sdists whose project
  names contain hyphens (e.g. `llnl-hatchet-2024.1.tar.gz`).

* Each contained file's parsed package name must match its directory
  name; a mismatch raises immediately (catches misfiled artifacts).

* Package pages use same-directory artifact links (no ../, no path
  separators) so pages do not depend on public CloudFront path mapping.

* S3 keys are stored raw. `+` is percent-encoded (`urllib.parse.quote`)
  only when rendered into an href, and the display text is HTML-escaped.
  Escaping at render rather than at list time keeps package-name parsing
  correct for PEP 440 local-version wheels such as
  `torch-2.10.0+rocm7.14.0-cp310-cp310-linux_x86_64.whl`, which
  `packaging`'s parser rejects if the `+` is already percent-encoded.

* No allow-list or version-threshold filtering: the directory contents
  are trusted. Version cleanup is a publisher/lifecycle concern.

* Checksums are omitted by default (`skip_checksum=True`). Pass
  `--include-checksums` to include `#sha256` fragments in artifact
  links. The checksum-skip default replaces an upstream nightly-path
  heuristic (`whl/nightly` prefix check) that was never active in
  TheRock or rockrel.

* Uploading is opt-in: the CLI writes index pages to local disk by
  default and only uploads to S3 when `--upload` is passed, so a bare
  invocation is a safe, side-effect-free preview. The importable
  `generate_structured_index(..., upload=True)` keeps uploading as its
  default for library/Lambda callers.

* S3 multipart uploads produce composite checksums (`base64==-N`) that
  are not valid SHA256 digests. `fetch_metadata` guards against these
  and falls back to object metadata (`checksum-sha256`,
  `x-amz-meta-checksum-sha256`); it issues a HEAD when either size or
  checksum is missing, since one HEAD populates both. `fetch_pep658`
  applies the same multipart guard to `.metadata` sidecar checksums.

* Single-package scoping (`--package`) regenerates one package
  directory and skips the root index, mirroring the flat Lambda scoping
  added in #5934 but without the filename-prefix hack since each package
  is already its own directory.

* As this is novel code and not a fork, it is placed in
  `build_tools/packaging/python/` alongside the other Python packaging
  index generators rather than under `third_party/`.

## Test Plan
New unit tests added, executed in CI. To execute manually:
```
  cd build_tools/packaging/python
  python -m pytest tests/manage_structured_test.py -v
```

## Test Result
43 tests pass in `manage_structured_test.py`. The rest of the
`s3_management` suite (legacy `manage.py`) is unchanged and still
passes.


build_tools/packaging/python/manage_structured.py
ScottTodd, line 1: If this is a new implementation and not a fork of third party code, it could go in https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python instead of https://github.com/ROCm/TheRock/tree/main/build_tools/third_party/s3_management (which I also notice has lingering references to "v2-staging" and other non-multi-arch packages/indexes)

build_tools/third_party/s3_management/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+        help="Include #sha256 fragments in artifact links (default: omitted).",
+    )
+    return parser
+
+
+def main() -> None:
+    args = create_parser().parse_args()
+    bucket_name = args.bucket or getenv("S3_BUCKET_PY")
+    if not bucket_name:
+        raise RuntimeError("Bucket must be provided via --bucket or S3_BUCKET_PY")
```
ScottTodd: 1. How about setting the default value for the argument to `getenv("S3_BUCKET_PY"`) instead of doing this? I think an explicit `--bucket` argument should take precedence over an environment variable (global state) and that code pattern has that behavior while this does not.
2. If keeping this, could use `parser.error()` instead of `raise RuntimeError()`

build_tools/packaging/python/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+        help=(
+            "Regenerate only this package directory and skip the product root "
+            "index (used for targeted single-package regeneration)."
+        ),
+    )
+    parser.add_argument(
+        "--do-not-upload",
+        action="store_true",
+        help="Write index pages to local disk instead of uploading to S3.",
+    )
```
ScottTodd, line 538: This is okay, just some ideas...

* We could also add a  `--dry-run` argument and/or flip the default to "don't upload" and require an explicit `--upload` to take the action that has side effects.
* We could have an "output mode" that chooses between local and remote using `choices=[...]` in argparse: https://docs.python.org/3/library/argparse.html#choices

build_tools/packaging/python/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+# SPDX-License-Identifier: MIT
+
+"""Structured product-local PEP 503 index generator.
+
+Generates pip-compatible PEP 503 simple indexes for the stream-subdomain
+structured layout, where each package lives in its own directory:
+
+    <product>/<index>/<normalized-package>/<filename>
+    <product>/<index>/<normalized-package>/index.html
+    <product>/<index>/index.html            (product-local root listing)
```
ScottTodd, line 12: Can you add an example usage comment to this docstring? I looked at the arguments in `create_parser()` but it isn't totally clear to me how this operates.

Then we could also document in https://github.com/ROCm/TheRock/blob/main/build_tools/third_party/s3_management/README.md or https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md once this is in use.

build_tools/third_party/s3_management/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+def fetch_pep658(
+    client: botocore.client.BaseClient,
+    bucket_name: str,
+    files: list[PackageFile],
+) -> None:
+    """Populate the PEP 658 .metadata sidecar checksum for each file.
+
+    Adapted from the legacy flat generator's S3Index.fetch_pep658.
```
ScottTodd: fwiw I don't think there is enough unique code in here to warrant referencing "legacy generators" via docstrings, all the function really does is:
1. call the boto3 API to fetch each metadata file
2. decode the metadata and store into a list

(if this is what is anchoring this file to third_party/ then I think the new code is novel enough to go outside)

build_tools/third_party/s3_management/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+def fetch_metadata(
+    client: botocore.client.BaseClient,
+    bucket_name: str,
+    files: list[PackageFile],
+) -> None:
+    """Populate checksum and size for each file via S3 HEAD requests.
+
+    Adapted from the legacy flat generator's S3Index.fetch_metadata. Falls back
+    to checksum-sha256 object metadata for older files without a native
+    checksum.
```
ScottTodd: Similar comment here. The code is obvious enough to be produced independently from third_party reference code

build_tools/packaging/python/manage_structured.py
```diff
@@ -0,0 +1,551 @@
...
+            f' data-core-metadata="{pep658_sha}"'
+        )
+    # networkx version/python constraints (pytorch/pytorch#152191):
+    # 3.4.2 for Python 3.10, 3.5+ for Python 3.11+.
+    if file.filename == "networkx-3.4.2-py3-none-any.whl":
+        attributes += ' data-requires-python="&gt;=3.10"'
+    elif file.filename.startswith("networkx-") and file.filename.endswith(
+        "-py3-none-any.whl"
+    ):
+        attributes += ' data-requires-python="&gt;=3.11"'
```
ScottTodd, line 251: This *does* look similar enough to third_party code, but just a few lines are small enough that this could be treated as a reference to that issue and not a forking of code.

Do we need it in this function itself though? We do still support Python 3.10 https://github.com/ROCm/TheRock/blob/5b6302a0c3134ff84dbb3051bdae084dcc9d4459/build_tools/github_actions/configure_pytorch_release_matrix.py#L23

so I guess yes?

ScottTodd (approved):

build_tools/packaging/python/manage_structured.py
marbre: True. I'd kind of like to keep the script aligned and upload_dependencies won't be a fork and rather is refactored. But I guess it might make sense to move it over.

build_tools/packaging/python/manage_structured.py
marbre: Yeah, basically this matches `manage.py`'s behavior and I agree that going for another argument might be useful here indeed. Will take a look at the other scripts we use for indexing before landing this.

build_tools/packaging/python/manage_structured.py
marbre: I could have started with #6622 which is required to test :D I add a bit more to the docstring. I add more docs as soon as I validated in a test deployment.

build_tools/packaging/python/manage_structured.py
marbre: Yeah, it was flagged during an internal review already and is fragile but I think it is still needed as long as we support 3.10.

build_tools/packaging/python/manage_structured.py
marbre: Will move to `build_tools/packaging/python`.

build_tools/third_party/s3_management/manage_structured.py
marbre: Done.

build_tools/packaging/python/manage_structured.py
marbre: I went with `--upload` which is now explicitly needed in CLI.

build_tools/packaging/python/manage_structured.py
marbre: Done.

build_tools/third_party/s3_management/manage_structured.py
marbre: Agreed. Moved it out.

PR #6605, opened 2026-07-16
feat: hipTensor build decoupled from CK

### LINKED ISSUE #6268: [Feature]: hipTensor checks _ck_enable to build forcing it to mimic the platform availability from CK
### Suggestion Description

Since hipTensor decides to build or not according to `_ck_enabled`, and `_ck_enabled` is ON or OFF depending of the platform availability for CK. If CK is enable for a platform family and hipTensor is set as excluded from that family, hipTensor's build will get triggered with an empty GPU_TARGETS and the build will fail.

hipTensor should build for an empty GPU_TARGETS without errors and should be decoupled from `_ck_enabled`.

### Operating System

_No response_

### GPU

_No response_

### ROCm Component

hipTensor

evedovelli:
## Motivation

Decouples hipTensor's build from _ck_enabled flag, allowing CK to be enable for a platform family while hipTensor is excluded from the same family.

## Technical Details

hipTensor's build was modified so it won't break in case CK isn't available for a specific target. In that case, a thin implementation layer just returns "not implemented" for any API call that would require CK's backend available. With this change in place, we can now decouple hipTensor's build from CK availability.

## Test Plan

Build TheRock and hipTensor for targets where CK isn't available (e.g. gfx103X, gfx906), and build and test hipTensor for targets where CK is available (e.g. gfx942, gfx950).

## Test Result

Builds and tests passing accordingly.

## Submission Checklist

ISSUE ID : https://github.com/ROCm/TheRock/issues/6268


geomin12 (commented):
lgtm, but odd... i don't see the multi arch CI triggered at all..... could we try re-triggering the build so we can see it pass?

math-libs/CMakeLists.txt
```diff
@@ -648,11 +648,17 @@ if(_ck_enabled)
...
-if(_ck_enabled AND THEROCK_ENABLE_HIPTENSOR)
+if(THEROCK_ENABLE_HIPTENSOR)
   ##############################################################################
   # hipTensor
   ##############################################################################
 
+  # Set composable_kernel as a build dependency only when it's available.
+  set(hiptensor_build_deps)
+  if(_ck_enabled)
+    list(APPEND hiptensor_build_deps composable_kernel)
```
marbre:
```suggestion
    list(APPEND _hiptensor_build_deps composable_kernel)
```

math-libs/CMakeLists.txt
```diff
@@ -671,7 +677,7 @@ if(_ck_enabled AND THEROCK_ENABLE_HIPTENSOR)
     BUILD_DEPS
       rocm-cmake
       therock-googletest
-      composable_kernel
+      ${hiptensor_build_deps}
```
marbre:
```suggestion
      ${_hiptensor_build_deps}
```

marbre (commented):
LGTM, just a minor NIT. Let's see what CI says.

Ryker0627 (approved):
LGTM

marbre (approved):
LGTM, assuming that CI passes.

PR #6606, opened 2026-07-16
test(packaging): rewrite promote_packages multi-arch/JAX/tarball test

JeniferC99:
## Motivation

Split out of #6409 so that PR carries only the `promote_packages.py` hardening while the on-demand promotion **test** rewrite is reviewed on its own. The prior test targeted retired 7.9 single-arch packages and no longer ran end to end; JAX promotion had no coverage; and per-gfx/multi-arch `therock-dist-*.tar.gz` tarball promotion was untested. Kept as an on-demand script (not per-PR CI) since it builds venvs against large real artifacts.

## Technical Details

- Runs against a local directory of already-downloaded RC packages (`--input-dir`); wheels + the `rocm` sdist are read from `wheels/` and tarballs from `tarball-multi-arch/`.
- Auto-detects the source RC and final versions (from package metadata via `pkginfo`), the platform, and the gfx arches from the files present; derives expected promoted names by the same rc/a->final rename instead of hardcoded lists.
- Multi-arch scenario asserts non-kept-arch device wheels are pruned; triton is selected by torch's exact `Requires-Dist: triton==` pin; JAX (`jax_rocm7_plugin` / `jax_rocm7_pjrt`) installs `--no-deps` (jax/jaxlib live on PyPI, `jaxlib` stays single-arch per #6158).
- Install checks are narrowed to the running interpreter's CPython tag and to one coherent stack per family (newest per distribution) so a multi-CPython, multi-version download stays co-installable.
- Validated end to end on downloaded `7.14.0rc3` (all promote+install scenarios pass in a fresh venv). 

## Submission Checklist

### DIFF
--- a/build_tools/packaging/tests/promote_packages_test.py
+++ b/build_tools/packaging/tests/promote_packages_test.py
@@ -2,63 +2,102 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
-"""Test suite for the promote_packages package promotion script.
+"""Manual / on-demand test suite for the promote_packages promotion script.
 
-This test suite validates the functionality of promote_packages.py, which
-promotes release candidate (RC) packages to final releases by removing RC suffixes
-from version strings (e.g., 7.9.0rc1 → 7.9.0).
+This exercises promote_packages.py end-to-end against a **local directory of
+already-downloaded** pre-release artifacts. It does not download RC packages
+itself — point --input-dir at a download_python_packages.py output root: wheels
++ the rocm sdist are read from `<root>/wheels/` and distribution tarballs from
+`<root>/tarball-multi-arch/` (either subdir may be absent). Wheels and tarballs
+are never read from the same directory. Aggregator install checks call
+`pip install` without `--no-index`, so PyPI may still be reached for transitive
+deps; JAX wheels install with `--no-deps` only.
 
-The test suite downloads RC packages from https://rocm.prereleases.amd.com/ and runs
-comprehensive validation tests to ensure that:
-  1. Complete promotion of all ROCm SDK and PyTorch packages works correctly
-  2. Partial promotions (only ROCm or only PyTorch) fail as expected
-  3. Promoted packages have correct filenames without RC suffixes
-  4. All promoted wheels have consistent version strings
-  5. Promoted packages can be successfully installed in a virtual environment
+A real download mixes both OSes and several package versions in `wheels/`; the
+input is filtered to one --platform (default linux) and, for the install checks,
+to one coherent stack per family (see below), since e.g. multiple torch versions
+can't be installed together.
 
-NOTES:
-  - Tests run in isolated temporary directories (including venv) to avoid polluting the workspace
-  - Original RC packages are downloaded fresh or loaded from cache for each test run
-   (set with --cache-dir, directory is created if it doesn't exist)
-  - Tests use real package files to ensure end-to-end validation
-  - Platform is auto-detected based on the system but can be overridden with the --platform argument.
+Promotion rewrites the prerelease segment of the version rather than only
+stripping "rc": promote_packages.py can strip it (rc/a -> release, e.g.
+7.14.0rc1 -> 7.14.0) or replace it (a -> rc, rc -> a). For a multi-arch input it
+additionally prunes the per-gfx device wheels down to a keep-list of
+architectures; when only the arch set is trimmed (no prerelease-type change) the
+version string is left unchanged. These tests drive the common rc/a -> release
+path.
 
-TEST SCENARIOS:
-  - checkPromoteEverything: Tests promotion of all packages. This test should SUCCEED.
+It is intentionally a standalone, on-demand script (run by a human before a
+release), NOT a pytest/CI target: it installs the wheels into throwaway
+virtualenvs. Keep it runnable by hand.
 
-  - checkPromoteOnlyRocm: Tests promotion of only ROCm SDK packages while leaving
-    PyTorch packages as RC. This test should FAIL.
+WHAT IS VALIDATED
+  1. Full promotion of the arch-agnostic aggregator packages
+     (rocm, rocm_sdk_core, rocm_sdk_devel, rocm_sdk_libraries, torch,
+     torchvision, torchaudio, triton) succeeds and installs.
+  2. Multi-arch promotion with a keep-list keeps only ONE arch's per-gfx device
+     wheels (amd_torch_device_gfx*, amd_torchvision_device_gfx*,
+     rocm_sdk_device_gfx*) and prunes the rest, while still stripping the RC
+     suffix everywhere. The kept arch and the pruned arch(es) are auto-detected
+     from whatever device wheels are present.
+  3. JAX wheels (jax_rocm7_* / jax_rocm10_* plugin and pjrt) promote correctly: the RC
+     suffix is stripped from the `+rocm<ver>` local segment and the wheels stay
+     structurally installable (installed --no-deps, since jax/jaxlib live on
+     PyPI). Note jaxlib itself is NOT published multi-arch (see ROCm/TheRock#6158).
+  4. Partial promotions (only ROCm, or only torch) do NOT produce a coherent,
+     installable, single-version set (expected to FAIL).
+  5. Every therock-dist tarball present (per-gfx AND multiarch) is promoted by
+     its filename rc->final rename (promote_targz_tarball renames by filename,
+     not by repacking).
+  6. Promoted filenames drop the RC suffix and every wheel reports the final
+     version.
 
-  - checkPromoteOnlyTorch: Tests promotion of only PyTorch-related packages while
-    leaving ROCm SDK packages as RC. This test should FAIL.
+The source version, the target platform, and the set of gfx architectures are
+all DERIVED from the files in the input directory. Nothing is hard-coded to a
+specific release. How the expected output is worked out:
 
-PACKAGE TYPES TESTED:
-  - ROCm SDK packages: rocm, rocm_sdk_core, rocm_sdk_devel, rocm_sdk_libraries_*
-  - PyTorch packages: torch, torchaudio, torchvision, triton
-  - Distribution tarballs: therock-dist-{platform}-gfx{arch}-{version}.tar.gz
+  1. Read the source version from the package itself -- the rocm_sdk_core wheel
+     (or the rocm sdist) via pkginfo, e.g. "7.9.0rc20260501".
+  2. Drop the prerelease tail to get the final version, e.g. "7.9.0".
+  3. For every discovered filename, swap the source version for the final
+     version -- the same rename promotion does -- and assert the promoted
+     directory contains exactly that set. For example:
+       rocm_sdk_core-7.9.0rc20260501-...whl -> rocm_sdk_core-7.9.0-...whl
+       torch-2.7.1+rocm7.9.0rc20260501-...whl -> torch-2.7.1+rocm7.9.0-...whl
 
-PREREQUISITES:
-  - pip install -r ./build_tools/packaging/requirements.txt
+Coherent install stack: promote_packages works on any subset of wheels, so where
+`wheels/` carries several versions of a distribution (torch 2.9..2.14, jax
+0.9/0.10) the install checks use only the newest of each, with triton pinned to
+whatever the chosen torch requires (`Requires-Dist: triton==<ver>`). A download
+also ships every supported CPython (cp310..cp314) while the install checks run
+under just one interpreter, so version-specific wheels are further narrowed to
+that interpreter's tag (Python-agnostic files -- the rocm sdist, `py3-none-*` SDK
+wheels -- always qualify). Promotion is Python-agnostic and behaves identically
+for the versions/CPythons left out.
 
-USAGE:
-  # Test on current platform (auto-detected):
-  python ./build_tools/packaging/tests/promote_packages_test.py
+PREREQUISITES
+  pip install -r ./build_tools/packaging/requirements.txt
 
-  # Use cached packages to speed up repeated test runs:
-  python ./build_tools/packaging/tests/promote_packages_test.py --cache-dir=/tmp/package_cache
+USAGE
+  # Promote + verify a download_python_packages.py output tree:
+  python ./build_tools/packaging/tests/promote_packages_test.py --input-dir ./download
+
+  # Force a platform (otherwise defaults to linux when both are present):
+  python ./build_tools/packaging/tests/promote_packages_test.py \
+      --input-dir ./download --platform linux
 """
 
 import argparse
-import shutil
-import sys
 import os
-from pathlib import Path
-import tempfile
-from packaging.version import Version
-from pkginfo import Wheel
+import re
+import shutil
 import subprocess
-import urllib
-import platform as platform_module
+import sys
+import tempfile
+import zipfile
+from pathlib import Path
+
+from packaging.version import Version
+from pkginfo import SDist, Wheel
 
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
 import promote_packages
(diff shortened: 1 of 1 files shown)


build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
-USAGE:
-  # Test on current platform (auto-detected):
-  python ./build_tools/packaging/tests/promote_packages_test.py
-
-  # Use cached packages to speed up repeated test runs:
-  python ./build_tools/packaging/tests/promote_packages_test.py --cache-dir=/tmp/package_cache
+"""Manual / on-demand test suite for the promote_packages promotion script.
+
+This exercises promote_packages.py end-to-end against a **local directory of
+already-downloaded** pre-release artifacts. It does not download RC packages
```
HereThereBeDragons:
```suggestion
already-downloaded** pre-release artifacts. It does not download packages
```

either say something like "to-be-promoted" packages are introduce before that you only do a partial test of the promotion script: "rc" -> release and do not care about rc -> a, a -> rc, a -> release

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+itself — point --input-dir at a download_python_packages.py output root: wheels
++ the rocm sdist are read from `<root>/wheels/` and distribution tarballs from
+`<root>/tarball-multi-arch/` (either subdir may be absent). Wheels and tarballs
+are never read from the same directory. Aggregator install checks call
+`pip install` without `--no-index`, so PyPI may still be reached for transitive
+deps; JAX wheels install with `--no-deps` only.
+
+A real download mixes both OSes and several package versions in `wheels/`; the
```
HereThereBeDragons:
```suggestion
A real download mixes both OSes (Linux, Windows) and several package versions in `wheels/`; the
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+stripping "rc": promote_packages.py can strip it (rc/a -> release, e.g.
+7.14.0rc1 -> 7.14.0) or replace it (a -> rc, rc -> a). For a multi-arch input it
+additionally prunes the per-gfx device wheels down to a keep-list of
+architectures; when only the arch set is trimmed (no prerelease-type change) the
+version string is left unchanged. These tests drive the common rc/a -> release
+path.
+
+It is intentionally a standalone, on-demand script (run by a human before a
+release), NOT a pytest/CI target: it installs the wheels into throwaway
+virtualenvs. Keep it runnable by hand.
```
HereThereBeDragons:
```suggestion
virtualenvs.
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+
+WHAT IS VALIDATED
+  1. Full promotion of the arch-agnostic aggregator packages
```
HereThereBeDragons: we never used arch-agnostic before. generally what i heard more often is talking about "generic" and "device" packages. you might want to explain this first in a section what type of packages exist and how they look like. then in the "what is validated" name the function and just a short sentence what it does.

something like:
(renaming checkPromoteEverything -> checkPromoteGeneric
and checkPromoteMultiArch -> checkPromoteArchKeepList to better reflect what they do)

```
WHAT IS VALIDATED  (function -> what it guarantees)
  checkPromoteGeneric         generic (arch-agnostic) packages (rocm*, torch,
                              torchvision, torchaudio, triton) promote + install;
                              no device packages
  checkPromoteArchKeepList    keep-list keeps one arch's device packages, prunes the rest
                              (this is where device packages are promoted + installed)
  checkPromoteJax             JAX wheels promote + install (--no-deps: structural
                              check only, no PyPI, no runtime)
  checkPartialPromotion       partial promotion FAILS as expected: rocm-only /
                              torch-only is incoherent  (run twice)
  checkPromoteTarball         every therock-dist tarball renamed rc->final

  Assertions used inside the above (not standalone scenarios):
  checkPromotedFileNames    promoted filenames drop the RC suffix
  checkAllWheelsSameVersion every wheel reports the final version

```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+
+The source version, the target platform, and the set of gfx architectures are
+all DERIVED from the files in the input directory. Nothing is hard-coded to a
+specific release. How the expected output is worked out:
+
+  1. Read the source version from the package itself -- the rocm_sdk_core wheel
+     (or the rocm sdist) via pkginfo, e.g. "7.9.0rc20260501".
+  2. Drop the prerelease tail to get the final version, e.g. "7.9.0".
+  3. For every discovered filename, swap the source version for the final
+     version -- the same rename promotion does -- and assert the promoted
```
HereThereBeDragons:
```suggestion
     version and assert the promoted
```
not the same. the code is different. you use here string replace, promote_packages uses regex replacement.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+     PyPI). Note jaxlib itself is NOT published multi-arch (see ROCm/TheRock#6158).
+  4. Partial promotions (only ROCm, or only torch) do NOT produce a coherent,
+     installable, single-version set (expected to FAIL).
+  5. Every therock-dist tarball present (per-gfx AND multiarch) is promoted by
+     its filename rc->final rename (promote_targz_tarball renames by filename,
+     not by repacking).
+  6. Promoted filenames drop the RC suffix and every wheel reports the final
+     version.
+
+The source version, the target platform, and the set of gfx architectures are
```
HereThereBeDragons: overall would be nice to have a better overview how this file works. here you coudl e.g. also refer to the inputset

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+all DERIVED from the files in the input directory. Nothing is hard-coded to a
+specific release. How the expected output is worked out:
+
+  1. Read the source version from the package itself -- the rocm_sdk_core wheel
+     (or the rocm sdist) via pkginfo, e.g. "7.9.0rc20260501".
```
HereThereBeDragons: lets update to a newer version that has multiarch? 7.13 introduced it but also can go for 7.14

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+  2. Drop the prerelease tail to get the final version, e.g. "7.9.0".
```
HereThereBeDragons: this might create misunderstandings. as you are not just dropping the tail you are using "Version" to correctly parse the version and retrieving the release version of it

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+       torch-2.7.1+rocm7.9.0rc20260501-...whl -> torch-2.7.1+rocm7.9.0-...whl
+
+Coherent install stack: promote_packages works on any subset of wheels, so where
+`wheels/` carries several versions of a distribution (torch 2.9..2.14, jax
+0.9/0.10) the install checks use only the newest of each, with triton pinned to
+whatever the chosen torch requires (`Requires-Dist: triton==<ver>`). A download
+also ships every supported CPython (cp310..cp314) while the install checks run
+under just one interpreter, so version-specific wheels are further narrowed to
+that interpreter's tag (Python-agnostic files -- the rocm sdist, `py3-none-*` SDK
+wheels -- always qualify). Promotion is Python-agnostic and behaves identically
```
HereThereBeDragons: considering how often jax changes how its installation and packages look like in each version i would not put that statement here.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+  1. Read the source version from the package itself -- the rocm_sdk_core wheel
+     (or the rocm sdist) via pkginfo, e.g. "7.9.0rc20260501".
+  2. Drop the prerelease tail to get the final version, e.g. "7.9.0".
+  3. For every discovered filename, swap the source version for the final
+     version -- the same rename promotion does -- and assert the promoted
+     directory contains exactly that set. For example:
+       rocm_sdk_core-7.9.0rc20260501-...whl -> rocm_sdk_core-7.9.0-...whl
+       torch-2.7.1+rocm7.9.0rc20260501-...whl -> torch-2.7.1+rocm7.9.0-...whl
+
+Coherent install stack: promote_packages works on any subset of wheels, so where
```
HereThereBeDragons: what is a coherent install stack for you? please explain more so that readers have the same mental model as you. if you think this is duplication and its enough to refer to the inputset docstring, just do that. just maybe then add more file description of what inputset will provide

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+that interpreter's tag (Python-agnostic files -- the rocm sdist, `py3-none-*` SDK
+wheels -- always qualify). Promotion is Python-agnostic and behaves identically
+for the versions/CPythons left out.
+
+PREREQUISITES
+  pip install -r ./build_tools/packaging/requirements.txt
+
+USAGE
+  # Promote + verify a download_python_packages.py output tree:
+  python ./build_tools/packaging/tests/promote_packages_test.py --input-dir ./download
```
HereThereBeDragons, line 160: if you have both linux and windows: will it test both platforms?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
 
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
 import promote_packages
 
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent.parent))
 import setup_venv
 
+PLATFORMS = ("linux", "windows")
+
+# Arch-agnostic aggregator wheels: always promoted, never pruned by the
```
HereThereBeDragons: why call them aggregator wheels? they are not collecting anything. these are generic packages.

the 2nd statement is also not true. some of those packages are touched when you use the keeplist to reduce the number of available archs.

see from promote_packages.py:
```
# Wheels whose metadata can carry per-gfx references. Per-arch wheels (e.g.
# `rocm_sdk_device_gfx1010`) are skipped earlier in the main loop based on
# filename, so they never reach this pass; this list is the set of *aggregators*
# whose Provides-Extra/Requires-Dist or AVAILABLE_TARGET_FAMILIES entries we
# may need to trim.
GFX_AWARE_WHEEL_PREFIXES = (
    "rocm-",  # tar.gz sdist
    "rocm_sdk_",  # core, libraries, devel
    "rocm_profiler-",
    "torch-",
    "torchvision-",
)
```

HereThereBeDragons: now i also understand where your confusion from the aggregator wheels come from:
in promote_packages.py aggregators are a subset of generic (non-device) wheels: those wheels that have all the archs aggregated.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+)
+
+# Per-gfx device wheel marker, e.g. `rocm_sdk_device_gfx942-...` /
+# `amd_torch_device_gfx1201-...`. Reuses promote_packages._GFX_ARCH (the same
+# arch-token shape the promotion script keys off) so the two never drift; the
+# trailing `-` anchors the token so `gfx11` can't match `gfx1153`.
+_DEVICE_ARCH_RE = re.compile(rf"device_({promote_packages._GFX_ARCH})-")
+
+
+def _is_aggregator(name: str) -> bool:
```
HereThereBeDragons: again misnomer what an aggregator is

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+    "torchvision-",
+    "torchaudio-",
+    "triton-",
+)
+
+# Per-gfx device wheel marker, e.g. `rocm_sdk_device_gfx942-...` /
+# `amd_torch_device_gfx1201-...`. Reuses promote_packages._GFX_ARCH (the same
+# arch-token shape the promotion script keys off) so the two never drift; the
+# trailing `-` anchors the token so `gfx11` can't match `gfx1153`.
+_DEVICE_ARCH_RE = re.compile(rf"device_({promote_packages._GFX_ARCH})-")
```
HereThereBeDragons: lets move it closer to where it is used

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+def _device_arch(name: str) -> str | None:
+    m = _DEVICE_ARCH_RE.search(name)
+    return m.group(1) if m else None
+
+
+def _wheel_platform(name: str) -> str | None:
+    """The platform a wheel targets, from its filename tag, or None if it is
+    platform-agnostic (the rocm sdist, any `-none-any` wheel). A download's
+    `wheels/` dir carries both OSes, so this is used to pick one."""
+    if "win_amd64" in name:
```
HereThereBeDragons:
```suggestion
    if "-win_" in name:
```

who knows for what other cpu archs we will provide rocm eventually.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+    """Whether `name` can be installed under the interpreter running the tests.
+
+    A real download ships every supported CPython (cp310..cp314); the install
+    checks run under exactly one, so version-specific wheels for other CPythons
+    must be dropped -- pip rejects them with "not a supported wheel on this
+    platform". Python-agnostic files (rocm sdist, `py3-none-*` SDK wheels) always
+    qualify. Promotion itself is Python-agnostic, so this only narrows the
+    install/expectation sets, never what promotion is exercised against."""
+    tag = _wheel_pytag(name)
+    return tag is None or tag == _RUNNER_PYTAG
```
HereThereBeDragons: why not reduce this to a single function like:

```
if "-cp3" not in name:  # no CPython tag -> python-agnostic
        return True
runner = f"cp{sys.version_info.major}{sys.version_info.minor}"
return f"-{runner}-" in name
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+
+
+def _list_files(directory: Path) -> set[str]:
+    """Filenames in `directory`, or an empty set if it doesn't exist."""
+    if not directory.is_dir():
+        return set()
+    return {p.name for p in directory.iterdir() if p.is_file()}
+
+
+def _torch_triton_pin(wheel_path: Path) -> str | None:
```
HereThereBeDragons, line 304: i think at some other level, maybe at the top docstring of the file you need to explain why triton needs this special treatment but not torchvision or similar.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
-    if len(files) != len(expected_promoted_pkgs):
+    def __init__(self, input_dir: Path, platform: str | None) -> None:
+        # download_python_packages.py emits wheels + the rocm sdist under
+        # `wheels/` and distribution tarballs under `tarball-multi-arch/` -- never
+        # mixed. Read each from its own subdir; either may be absent.
+        self.wheels_dir = input_dir / "wheels"
+        self.tarball_dir = input_dir / "tarball-multi-arch"
+
+        names = _list_files(self.wheels_dir)
+        if not names:
```
HereThereBeDragons: you need to document that if wheels no wheels are existing (evne if tarballs would) this is a hard failure. maybe we dont want this and just run then the tarball testS?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+            raise RuntimeError(f"No wheels found in {self.wheels_dir}")
+
+        self.source_str = detect_source_version(self.wheels_dir, names)
```
HereThereBeDragons: this should be a class function. also with this you could clean up the __init__ by moving more stuff in helper functions

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
 
-    # get files and strip path from them
-    files = dir_path.glob("*")
-    files = [file.name for file in files]
+def _triton_wheel_matches_pin(name: str, pin: str) -> bool:
+    # Prefix-match the upstream triton version so pin 3.7.1 cannot match 3.7.10.
+    return name.startswith(f"triton-{pin}-") or name.startswith(f"triton-{pin}+")
+
+
+def detect_source_version(input_dir: Path, names: set[str]) -> str:
```
HereThereBeDragons: this needs a check that there are no mixed version packages in the same dir.
then each test should also check that the packages it needs are available before it does anything

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        # rc vs a comes from the parsed prerelease kind, not a substring test
+        # (an alpha input like 7.13.0a20260501 must classify as "a").
+        pre = self.source_version.pre
+        self.src_version_type = pre[0] if pre else "rc"
+
+        # `wheels/` holds both linux and windows wheels, and they can't install
+        # together -- pick one platform and keep only its wheels plus the
+        # platform-agnostic files (rocm sdist, `-none-any`). Default to linux
+        # unless only windows is present.
+        present_platforms = {pl for n in names if (pl := _wheel_platform(n))}
```
HereThereBeDragons: why not run the test again for all the platforms found? 
for this you could pull this platform detection logic outside the input set and into main (or something similar)

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        self.platform = platform or (
+            "linux"
+            if "linux" in present_platforms
+            else next(iter(present_platforms), "linux")
+        )
+        names = {n for n in names if _wheel_platform(n) in (self.platform, None)}
+
+        # Device wheels feed the multi-arch scenario, which both promotes (prunes
+        # non-kept archs) and installs the kept arch. They are narrowed to the
```
HereThereBeDragons: i know i used prune in the beginning but there is a reason why it is called keep-list. you as the user declare which archs to keep, not which to prune. inclusion vs exclusion. keep list to not make errors and have wrong assumptions of what will be available

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        # newest version per package: a download ships several torch/torchvision
+        # versions per arch (e.g. amd_torch_device_gfx1010 2.10/2.11/2.12) whose
+        # device wheels cannot co-install, and the newest ones line up with the
+        # newest torch/torchvision picked for the aggregator install stack.
+        # Pruning is still fully exercised across every arch within that CPython.
+        self.device_files = self._newest_per_package(
+            {n for n in names if _device_arch(n) and _py_installable(n)}
+        )
+        self.jax_files = {n for n in names if "jax" in n}
+        self.aggregator_files = {
```
HereThereBeDragons: naming

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+            "linux"
+            if "linux" in present_platforms
+            else next(iter(present_platforms), "linux")
+        )
+        names = {n for n in names if _wheel_platform(n) in (self.platform, None)}
+
+        # Device wheels feed the multi-arch scenario, which both promotes (prunes
+        # non-kept archs) and installs the kept arch. They are narrowed to the
+        # runner's CPython (like the aggregator install set) and collapsed to the
+        # newest version per package: a download ships several torch/torchvision
```
HereThereBeDragons: maybe have a helper function that sets all the files

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+            if n.startswith("therock-dist")
+            and n.endswith(".tar.gz")
+            and self.source_str in n
+        }
+
+        self.present_archs = sorted(
+            {a for n in self.device_files if (a := _device_arch(n))}
+        )
+        # Auto-detect: keep the first present arch, everything else is pruned.
+        self.keep_arch = self.present_archs[0] if self.present_archs else None
```
HereThereBeDragons: i think it would be better to keep multiple archs and not just take the first one but one random or something like this, the middle one, or anything which isnt the default one (doesnt need to be the first). you also should check that it is not one of those "compute kernel" aotriton family packages like gfx11 or gfx12_ but that it is a proper arch

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+
+        # A download ships several mutually-exclusive torch/jax versions that
+        # can't co-install. promote_packages works on any subset, so the install
+        # scenarios use one coherent stack per family (newest of each package,
+        # with triton pinned to the chosen torch); promotion behaves identically
+        # for the versions left out.
+        self.install_set = self._coherent_install_set()
+        self.jax_install_set = self._newest_per_package(self.jax_files)
+
+    def _non_pinned_triton(self) -> set[str]:
```
HereThereBeDragons: why do we need this? are we running on all torhc files? if not it might be better to just check for the specific triton file when we want to install a specific torch version

to be more concrete:
why would we even have triton files of a given version not matching a pytorch version? triton will be available for any py/rocm/torch combo we are building for. there will be no extra triton not matching the torch version available for a given rocm version. there must be a match.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        """Triton wheels not pinned by any torch wheel in the set.
+
+        Every torch wheel is consulted (a directory can hold more than one), and
+        a triton wheel is kept if *any* torch pins it. If torch wheels are present
+        but none carry an exact `triton==` pin we raise rather than silently
+        staging every triton build (which would make installs incoherent).
+        """
```
HereThereBeDragons, line 511: not reviewing this function as i think we should drop it

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+            if any(_triton_wheel_matches_pin(n, pin) for pin in pins)
+        }
+        if not matched:
+            raise RuntimeError(
+                f"torch pins triton=={sorted(pins)} but no matching triton wheel "
+                f"is present in {self.wheels_dir} (found: {sorted(triton_wheels)})"
+            )
+        return triton_wheels - matched
+
+    def _dist_version(self, name: str) -> Version:
```
HereThereBeDragons: reuse this in detect_source_version()?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        path = self.wheels_dir / name
+        meta = SDist(path) if name.endswith(".tar.gz") else Wheel(path)
+        return Version(meta.version)
+
+    def _newest_per_package(self, names: set[str]) -> set[str]:
```
HereThereBeDragons: _newest_per_package -> _newest_version_per_package?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+                continue
+            pkg = n.split("-", 1)[0]  # wheel/sdist distribution name (no hyphens)
+            if pkg not in newest or self._dist_version(n) > self._dist_version(
+                newest[pkg]
+            ):
+                newest[pkg] = n
+        return set(newest.values())
+
+    def _coherent_install_set(self) -> set[str]:
+        """One installable stack from the aggregator wheels: the newest of each
```
HereThereBeDragons: rename. see earlier comments

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+        meta = SDist(path) if name.endswith(".tar.gz") else Wheel(path)
+        return Version(meta.version)
+
+    def _newest_per_package(self, names: set[str]) -> set[str]:
+        """Keep only the highest version of each distribution, restricted to
+        wheels installable under the running interpreter. A download can ship
+        several torch / jax versions (that cannot be installed together) across
+        several CPythons (only one of which matches this interpreter)."""
+        newest: dict[str, str] = {}
+        for n in names:
```
HereThereBeDragons: more efficient for loop so that you dont always have to reopen the package and reread _dist_version

```
def _newest_per_package(self, names: set[str]) -> set[str]:
    versions = {n: self._dist_version(n) for n in names}  # one read each
    newest: dict[str, str] = {}
    for n in names:
        pkg = n.split("-", 1)[0]
        if pkg not in newest or versions[n] > versions[newest[pkg]]:
            newest[pkg] = n
    return set(newest.values())
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
+
+    def promoted(self, names: set[str]) -> set[str]:
+        return {promoted_name(n, self.source_str, self.final_str) for n in names}
+
+
+def promoted_name(rc_name: str, version: str, final_version: str) -> str:
+    return rc_name.replace(version, final_version)
+
+
+def checkPromotedFileNames(dir_path: Path, expected: set[str]) -> tuple[bool, str]:
```
HereThereBeDragons: needs explanation what it does.
and that it is a helper function

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -116,43 +425,51 @@ def checkAllWheelsSameVersion(
         wheel = Wheel(file)
         version = Version(wheel.version)
 
-        if (
-            str(version) == str(expected_version) and version.local == None
-        ):  # rocm packages
-            continue
-        elif str(version.local) == "rocm" + str(expected_version):  # torch packages
+        local_tag = "rocm" + str(expected_version)
```
HereThereBeDragons: put this outside the loop. also would rename it to local_version

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -2,110 +2,419 @@
...
             False,
-            f"Files found and expected promoted packages are not the same amount ({len(files)} vs {len(expected_promoted_pkgs)})",
+            f"Promoted file set mismatch.\n  missing: {missing}\n  unexpected: {unexpected}",
         )
-
-    for file in files:
-        if not file in expected_promoted_pkgs:
-            return False, f"{file} not matching any of the expected package names"
-
     return True, ""
```
HereThereBeDragons, line 646: need docstring. make clear that this only compares that the rocm version part is matching. 
maybe want to rename to reflect that?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -116,43 +425,51 @@ def checkAllWheelsSameVersion(
         wheel = Wheel(file)
         version = Version(wheel.version)
 
-        if (
-            str(version) == str(expected_version) and version.local == None
-        ):  # rocm packages
-            continue
-        elif str(version.local) == "rocm" + str(expected_version):  # torch packages
+        local_tag = "rocm" + str(expected_version)
+        if str(version) == str(expected_version) and version.local is None:
```
HereThereBeDragons:
```suggestion
        if version == expected_version and version.local is None:
```

build_tools/packaging/tests/promote_packages_test.py
HereThereBeDragons: also clarify that .tar.gz sdist is not compared here

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -116,43 +425,51 @@ def checkAllWheelsSameVersion(
...
+                f"{file.name} has version {version}, but expected version is {expected_version}",
             )
 
     return True, ""
 
 
-def checkInstallation(dir_path: Path) -> tuple[bool, str]:
-    """
-    Note: dir_path must be a TemporaryDirectory, otherwise you must clean up the .venv created here yourself.
+def checkInstallation(dir_path: Path, no_deps: bool = False) -> tuple[bool, str]:
```
HereThereBeDragons: note frome claude:
Not offline — reaches PyPI (468). No --no-index. For the non-JAX scenarios (no --no-deps), torch's Requires-Dist pulls sympy, networkx, jinja2, fsspec, filelock, typing-extensions — none staged — so pip downloads them. Unless setup_venv.create_venv pins an index (I did not see it), the "Everything" scenario needs the internet to pass. Decide the intent:

accept network -> fix any docstring that claims hermetic, and note the test is not runnable in an air-gapped CI;
structural-only -> add --no-deps here too (like JAX), and you are just asserting the wheels unpack/install;
fully offline -> --no-index --find-links <dir> and stage transitive deps (heavy).

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -116,43 +425,51 @@ def checkAllWheelsSameVersion(
...
+        python_exe = setup_venv.find_venv_python_exe(dir_path / ".venv")
         if python_exe is None:
             return (
                 False,
                 "Problem when installing temporary venv: Python executable not found",
             )
 
-        # only install rocm wheels/sdist and pytorch wheels, not therock-dist tarball or .venv
+        # Install wheels/sdists, not the therock-dist tarball or the venv itself.
         packages = [
```
HereThereBeDragons: add a check that packages is not empty

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -116,43 +425,51 @@ def checkAllWheelsSameVersion(
...
             p
             for p in dir_path.glob("*")
```
HereThereBeDragons: i wonder if the package list should not be given by the calling function. would reduce assumptions and allow to guide more explicitly which packages to install.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
     return True, ""
 
 
+def _stage_inputs(src: Path, dst: Path, names: set[str]) -> None:
```
HereThereBeDragons: description and maybe better call it copy than staging?
and do we need somewhere to verify that we copied the correct files and didnt miss something?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+            )
+            # Cross-platform directory dump (the old `ls` shell-out broke on
+            # Windows and masked the real failure).
+            for entry in sorted(p.name for p in tmp_dir.glob("*")):
+                print(f"    {entry}")
+            return False
+    return True
+
+
+def _banner(msg: str) -> None:
```
HereThereBeDragons: have a look at build_tools/hack/env_check/device.py section_bar() for nicer formatting of such bars and centering content

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+def _stage_inputs(src: Path, dst: Path, names: set[str]) -> None:
+    for file in src.glob("*"):
+        if file.is_file() and file.name in names:
+            shutil.copy2(file, dst)
+
+
+def _run_checks(
+    tmp_dir: Path, checks: list[tuple[str, tuple[bool, str]]], expect_success: bool
+) -> bool:
+    """Run named checks. Returns whether the overall expectation held."""
```
HereThereBeDragons: wrong description or wrong implementation. did you want to run the checks here? would you like to stop the checks if one of them failed or run all of them first and then evaluate?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
 
 
+def _stage_inputs(src: Path, dst: Path, names: set[str]) -> None:
+    for file in src.glob("*"):
+        if file.is_file() and file.name in names:
+            shutil.copy2(file, dst)
+
+
+def _run_checks(
+    tmp_dir: Path, checks: list[tuple[str, tuple[bool, str]]], expect_success: bool
```
HereThereBeDragons: describe what checks should be as a content

build_tools/packaging/tests/promote_packages_test.py
HereThereBeDragons: claude suggests something like this:

```
_run_checks(tmp_dir, [
    ("checkPromotedFileNames", lambda: checkPromotedFileNames(tmp_dir, expected_files)),
    ("checkAllWheelsSameVersion", lambda: checkAllWheelsSameVersion(tmp_dir, expected_version)),
    ("checkInstallation", lambda: checkInstallation(tmp_dir)),
    ("partial set must not install", _expect_failure(lambda: checkInstallation(tmp_dir))),
])
```

and


```
for name, check in checks:
        ok, msg = check()          # lambda runs here, one at a time
        if not ok:
            print(f"\n[ERROR] {name} failed:\n{msg}")
            for entry in sorted(p.name for p in tmp_dir.glob("*")):
                print(f"    {entry}")
            return False           # short-circuit: later checks never run
    return True
```

and

```
def _expect_failure(
    check: Callable[[], tuple[bool, str]],
) -> Callable[[], tuple[bool, str]]:
    """Wrap a check so that its FAILURE is the passing outcome."""
    def wrapped() -> tuple[bool, str]:
        ok, msg = check()
        if ok:
            return False, "installed but should have been incoherent"
        return True, ""            # the check failed -> that is what we wanted
    return wrapped
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+) -> bool:
+    """Run named checks. Returns whether the overall expectation held."""
+    for func_name, res in checks:
+        if expect_success and not res[0]:
+            print(f"\n[ERROR] {func_name} failed:\n{res[1]}")
+            return False
+        if not expect_success and res[0]:
+            print(
+                f"\n[ERROR] {func_name} succeeded but the promotion should NOT have been coherent."
```
HereThereBeDragons: this might be a big assumption what exactly is failing

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+
+def _run_checks(
+    tmp_dir: Path, checks: list[tuple[str, tuple[bool, str]]], expect_success: bool
+) -> bool:
+    """Run named checks. Returns whether the overall expectation held."""
+    for func_name, res in checks:
+        if expect_success and not res[0]:
+            print(f"\n[ERROR] {func_name} failed:\n{res[1]}")
+            return False
+        if not expect_success and res[0]:
```
HereThereBeDragons: if expected == false, then ALL tests need to fail for it to be successful. but if we expect false, we only need one test to be failing to be considered true.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
 
-def checkPromoteOnlyTorch(
-    dir_path: Path, expected_version: Version, platform: str
-) -> bool:  # should fail
-    print("")
-    print(
-        "================================================================================="
+def parse_args(argv: list[str]) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="On-demand test of pre-release->final package promotion over a local directory."
```
HereThereBeDragons: add more how to call it. like the docstring at the very top of the file

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    ) as tmp:
-        tmp_dir = Path(tmp)
-        # make a copy in a separate dir to not pollute it
-        for file in dir_path.glob("*"):
-            shutil.copy2(file, tmp_dir)
+    parser.add_argument(
+        "--platform",
+        choices=PLATFORMS,
+        default=None,
+        help="Target platform (default: linux when both are present in wheels/).",
```
HereThereBeDragons: i thought linux is the default?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    args = parser.parse_args(argv)
+    if not args.input_dir.is_dir():
+        parser.error(f"--input-dir {args.input_dir} is not a directory")
+    return args
 
-        promote_packages.main(tmp_dir, match_files="*torch*", delete=True)
 
-        success = True
+if __name__ == "__main__":
+    p = parse_args(sys.argv[1:])
```
HereThereBeDragons: please put this entire main in a separate function. it is way too long and our code style wants this

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    print("")
     print(
-        "================================================================================="
+        f"Testing promotion {inputs.source_version} -> {inputs.final_version} on "
+        f"{inputs.platform} from {inputs.wheels_dir}"
     )
     print(
-        "TEST DONE: Testing promotion of only PyTorch packages. Result:"
-        + (" SUCCESS" if success else " FAILURE")
+        f"  archs present: {inputs.present_archs or '(none)'}; kept: {inputs.keep_arch}"
```
HereThereBeDragons: please use uniform capitalization. before you used it, now not

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        "torchvision-0.24.0+rocm7.9.0rc1-cp312-cp312-win_amd64.whl",
-    ]
-
-    url_and_packages = [(URL, package) for package in packages]
-    url_and_packages.append(
-        (
-            "https://rocm.prereleases.amd.com/tarball/",
-            "therock-dist-windows-gfx1151-7.9.0rc1.tar.gz",
-        )
+    multi_arch_input = inputs.install_set | inputs.device_files
```
HereThereBeDragons: you already have a InputSet. why not give this to the check function? why assign it here again.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+            inputs.keep_arch,
+            multi_arch_input,
+            dropped_promoted_names,
+            expected_multi_arch,
+            src,
+        )
+    else:
+        _banner(
+            "TEST: multi-arch promotion — SKIPPED "
+            "(need >=2 gfx archs with device wheels to exercise pruning)"
```
HereThereBeDragons: keep-list

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+        )
+        res_multi = True
+
+    res_only_rocm = checkPartialPromotion(
```
HereThereBeDragons: these functions and results could also be all a list like you have at the very bottom

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        )
-        print("checkPromoteEverything: " + ("SUCCESS" if res_everything else "FAILURE"))
-        print("checkPromoteOnlyRocm: " + ("SUCCESS" if res_rocm else "FAILURE"))
-        print("checkPromoteOnlyTorch: " + ("SUCCESS" if res_torch else "FAILURE"))
-        print(
-            "================================================================================="
-        )
+    _banner("SUMMARY")
+    print(f"checkPromoteEverything:  {'SUCCESS' if res_everything else 'FAILURE'}")
+    print(f"checkPromoteMultiArch:   {'SUCCESS' if res_multi else 'FAILURE'}")
```
HereThereBeDragons: "skipped" from above is being lost here.

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+            return False
+    return True
+
+
+def _banner(msg: str) -> None:
+    line = "=" * 81
+    print(f"\n{line}\n{msg}\n{line}")
+
+
 def checkPromoteEverything(
```
HereThereBeDragons: naming + missing description

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    )
-    print("TEST: Testing promotion of all packages")
-    print(
-        "================================================================================="
+    dir_path: Path,
+    expected_version: Version,
+    input_files: set[str],
+    expected_files: set[str],
+    src_version_type: str = "rc",
+) -> bool:
```
HereThereBeDragons: guards for the inputs taht they are not empty and are sane?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    _banner("TEST: promotion of all aggregator packages (should SUCCEED)")
+    with tempfile.TemporaryDirectory(prefix="PromoteTest-Everything-") as tmp:
+        tmp_dir = Path(tmp)
+        _stage_inputs(dir_path, tmp_dir, input_files)
+        promote_packages.main(tmp_dir, delete=True, src_version_type=src_version_type)
+        ok = _run_checks(
+            tmp_dir,
+            [
+                (
+                    "checkPromotedFileNames",
```
HereThereBeDragons: lets stop the tests as soon one of them fails? makes it cheaper and might safe a lot of time

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        prefix="PromoteRcToFinalTest-PromoteEverything-"
-    ) as tmp:
+    return ok
+
+
+def checkPromoteMultiArch(
+    dir_path: Path,
+    expected_version: Version,
+    keep_arch: str,
+    input_files: set[str],
```
HereThereBeDragons: missing guards

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    dropped_promoted_names: set[str],
+    expected_files: set[str],
+    src_version_type: str = "rc",
+) -> bool:
+    """Only meaningful for MULTI-ARCH packages: it exercises the keep-list pass
+    (`--multi-arch-targets`), which keeps `keep_arch`'s per-gfx device wheels and
+    prunes the rest. It needs >=2 gfx archs with device wheels present to have
```
HereThereBeDragons: again the misnomer about pruning vs keeping

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
             ),
             ("checkInstallation", checkInstallation(tmp_dir)),
-        ]:
-            if not res[0]:
-                print("")
+        ]
+        ok = _run_checks(tmp_dir, checks, expect_success=True)
+
+        # Explicitly confirm the non-kept arch device wheels were pruned.
+        if ok:
```
HereThereBeDragons: this is not good enough. you also need to look on the inside of the aggregaotr wheels that the dependencies are really cleaned up

build_tools/packaging/tests/promote_packages_test.py
HereThereBeDragons: and maybe a better description

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    return ok
+
+
+def checkPromoteJax(
+    dir_path: Path,
+    expected_version: Version,
+    input_files: set[str],
+    expected_files: set[str],
+    src_version_type: str = "rc",
+) -> bool:
```
HereThereBeDragons: description

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    _banner("TEST: promotion of JAX wheels (should SUCCEED)")
+    if not input_files:
+        print("[SKIP] no JAX wheels found for this version; skipping.")
+        _banner("TEST DONE: promote JAX. Result: SKIPPED")
+        return True
```
HereThereBeDragons: shouldnt it return something for skipped? could be None

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-    print(
-        "================================================================================="
+    return ok
+
+
+def checkPromoteJax(
+    dir_path: Path,
+    expected_version: Version,
+    input_files: set[str],
+    expected_files: set[str],
```
HereThereBeDragons: shouldnt be the expected jax files be calculated here?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-            ("checkInstallation", checkInstallation(tmp_dir)),
-        ]:
-            if res[0]:
-                success = False
-                print("")
+        _stage_inputs(dir_path, tmp_dir, tarball_names)
+        promote_packages.main(tmp_dir, delete=True, src_version_type=src_version_type)
+        produced = {p.name for p in tmp_dir.glob("*")}
+        for rc_name in sorted(tarball_names):
+            fin_name = promoted_name(rc_name, version, final_version)
```
HereThereBeDragons: fin?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        "================================================================================="
-    )
-    print(
-        "TEST DONE: Testing promotion of only rocm packages. Result:"
-        + (" SUCCESS" if success else " FAILURE")
-    )
-    print(
-        "================================================================================="
+                ok = False
+            elif rc_name in produced:
```
HereThereBeDragons:
```suggestion
            if rc_name in produced:
```

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
+    the discovered name with the rc->final rename applied.
+    """
+    _banner(
+        "TEST: promotion of therock-dist tarballs (per-gfx + multiarch, should SUCCEED)"
     )
-    return success
+    if not tarball_names:
+        print("[SKIP] no therock-dist tarballs found in tarball-multi-arch/; skipping.")
+        _banner("TEST DONE: promote tarball. Result: SKIPPED")
+        return True
```
HereThereBeDragons: skip retunr value?

build_tools/packaging/tests/promote_packages_test.py
```diff
@@ -162,310 +479,364 @@ def checkInstallation(dir_path: Path) -> tuple[bool, str]:
...
-        "================================================================================="
+                ok = False
+            elif rc_name in produced:
+                print(f"\n[ERROR] rc tarball {rc_name} survived promotion")
+                ok = False
+    _banner("TEST DONE: promote tarball. Result: " + ("SUCCESS" if ok else "FAILURE"))
+    return ok
+
+
+def checkPartialPromotion(
```
HereThereBeDragons: description

HereThereBeDragons (commented):
thanks for the pr and sorry for the many comments. while thinking about it came up with some follow up tasks :) 

Promote-packages follow-up: two standalone tools

Both are **separate PRs**; `promote_packages.py` stays untouched and dumb.
Coherence is a cross-file property, so it lives in a post-pass that reads the
produced files as a set, never in the per-file loop.

### Tool 1 — version-location scout (read-only report)
- Unzips a promoted wheel (or glob) and finds every place the version token appears.
- Reports only hits **outside** the script's known rewrite locations
  (METADATA `Requires-Dist`/`Summary`, `requires.txt`, `_dist_info.py`).
- Catches packages like JAX that scatter the version into random files
  (`_version.py`, plugin `__init__.py`).
- Text members only — do not scan binaries (cannot rewrite a `.so`/`.pyd`, and
  pip resolves on METADATA, not baked strings).
- Never gates: exit 0, it is a report a human uses to extend the script.

### Tool 2 — post-promotion install smoke test (detector, not a gate)
- Standalone script pointed at a promoted dir; runs **after** all promotion +
  delete completes.
- Input is a matrix of **partial** seed specs, e.g. `{torch: 2.10, py: 3.11}`,
  `{jax: 0.9.1, py: 3.12}`, `{gfx: gfx942}`; the resolver fills the rest from
  files actually present.
  - torch given → newest matching torch for that py tag + its exact
    `Requires-Dist: triton==` pin + matching torchvision/torchaudio + rocm aggregators.
  - jax given → that jax/jaxlib set, installed `--no-deps` (runtime deps live on PyPI).
  - gfx given → that arch's device wheels join; else aggregator-only.
- Per entry: resolve → venv → `pip install` in place → if GPU present AND gfx
  given, run minimal ROCm sanity; else SKIP.
- Detector, not a gate: RC sources are already deleted inline, so it cannot roll
  back — FAIL means re-download + re-promote.
- Exit non-zero on any entry FAIL.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: applied.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: applied

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: removed --platform, so no more windows option, ignores win_amd64 wheels in a mixed wheels/ tree, and hard-errors on a Windows-only dir.

build_tools/packaging/tests/promote_packages_test.py
JeniferC99: updated.

PR #6607, opened 2026-07-16
fix(ci): Resolve undefined pthread symbols in ASAN gtest links

amd-hsong:
## Motivation

ASAN builds fail to link gtest-based test executables with undefined pthread symbols. This PR fixes the issue.

## Technical Details

See Issue https://github.com/ROCm/rocm-libraries/issues/8889.

Two link flags in the sanitizer stanza reach the build through different CMake channels in cmake/therock_sanitizers.cmake:

  - `-fsanitize=address` is added via `CMAKE_C/CXX_FLAGS_INIT`.
  - `-shared-libsan` is added via `add_link_options()`.

  CMake's `FindThreads` detection runs a `try_compile` probe that inherits `*_FLAGS_INIT` but **not** `add_link_options()`. So the probe links with `-fsanitize=address` but without `-shared-libsan`, using clang's default **static** sanitizer runtime — which auto-links `-lpthread`. The probe therefore succeeds without explicit threads, CMake sets `CMAKE_HAVE_LIBC_PTHREAD=1`, and
  `Threads::Threads` becomes an empty target.

Real executables, however, link with `-shared-libsan` (shared runtime, no auto `-lpthread`). gtest propagates the now-empty `Threads::Threads`, so nothing puts `-pthread` on the link line, and the pthread TLS symbols go undefined on toolchains where they live only in `libpthread` for glibc < 2.34 (the manylinux CI container is based on glibc 2.28).

  This surfaced when rocThrust switched its default SQLite3 to the system (shared) package: the previously in-tree SQLite compilation had been incidentally pulling `libpthread` onto the test link lines and masking this problem.

## Test Plan

Verify both locally and by the Multi-Arch CI ASAN that the build now passes.

## Test Result

The build passes.

## Submission Checklist

### DIFF
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -57,6 +57,19 @@ function(therock_sanitizer_configure
     # https://github.com/ROCm/TheRock/issues/1782
     string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-fsanitize=${_sanitizer_string}>\n")
     string(APPEND _stanza "  $<$<AND:$<LINK_LANGUAGE:C,CXX>,$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>>:-shared-libsan>)\n")
+    # Fix FindThreads detection under ASAN. The threads probe is a try_compile
+    # that inherits CMAKE_C/CXX_FLAGS_INIT (-fsanitize) but NOT the directory
+    # add_link_options() above, so it never sees -shared-libsan and links clang's
+    # default STATIC sanitizer runtime, which auto-links -lpthread. The probe then
+    # passes trivially, CMAKE_HAVE_LIBC_PTHREAD=1, and Threads::Threads is empty.
+    # Feed the same shared-runtime flag to the probe via CMAKE_REQUIRED_LINK_OPTIONS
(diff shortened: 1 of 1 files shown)


amd-hsong:
Started the **Multi-Arch CI ASAN**: https://github.com/ROCm/TheRock/actions/runs/29543280824

amd-hsong:
This PR replaces https://github.com/ROCm/TheRock/pull/6341 - see the previous review comments there.

lumachad (approved):
Leaving previous approval.

amd-hsong:
@marbre Can you please take a look at the CI ASAN run: https://github.com/ROCm/TheRock/actions/runs/29543280824? Lots of tests were canceled (they waited too long or something?) and errors from tests that did run are unrelated.

lumachad:
> @marbre Can you please take a look at the CI ASAN run: https://github.com/ROCm/TheRock/actions/runs/29543280824? Lots of tests were canceled (they waited too long or something?) and errors from tests that did run are unrelated.

@marbre can correct me if I'm wrong here, but we don't currently expect the ASAN tests (and those that actually run) to be clean. The build itself needs to be clean. I'm not sure about the python wheels ones. Maybe those need to be clean.

I'd re-trigger those. The Linux build looks OK as far as I can see.

marbre (approved):
Looking at the latest nightly ASAN build and triggered test results I don't think we can expect tests to pass.

PR #6609, opened 2026-07-16
refactor(debug-tools): unify rocr-debug-agent into a single subproject

lumachad:
Merge the separate `rocr-debug-agent-tests` subproject into
`rocr-debug-agent`, mirroring how `rocgdb` handles tests as part of a
single component.

Previously the tests were a separate subproject requiring its own
`amd-hip` toolchain declaration. Since `amd-hip` is a strict superset of
`amd-llvm` for non-HIP code, switching the unified subproject to
`amd-hip` is transparent to the library build while enabling the upstream
`ENABLE_TESTS` option to pull in the test subdirectory.

Changes:
- Switch `rocr-debug-agent` to `COMPILER_TOOLCHAIN amd-hip`,
  `USE_TEST_AMDGPU_TARGETS`, and `-DENABLE_TESTS=${THEROCK_BUILD_TESTING}`
- Add `hip-clr` as a runtime dep and `INSTALL_RPATH_EXECUTABLE_DIR`
- Add `test` component to `artifact-rocr-debug-agent.toml`
- Remove the separate `rocr-debug-agent-tests` subproject, artifact
  descriptor, topology entry, and all downstream references
  (packaging, install script, unit test, README, CMakeLists comment)

ISSUE ID : N/A


lumachad:
@nunnikri @raramakr Could you please sanity check this change from the perspective of packaging? I don't expect a problem, but we are dropping a rocr-debug-agent-tests component and moving it to the existing rocr-debug-agent.

The artifact of rocr-debug-agent will now also include tests.

lumachad:
> @nunnikri @raramakr Could you please sanity check this change from the perspective of packaging? I don't expect a problem, but we are dropping a rocr-debug-agent-tests component and moving it to the existing rocr-debug-agent.
> 
> The artifact of rocr-debug-agent will now also include tests.

ping @raramakr @nunnikri

nunnikri:
@lumachad the ci is already running deb/rpm package creation and basic test. But may not cover debug tools test. You can verify the installation using the ci build with the following commands. 

Debian package:
echo "deb [trusted=yes] https://therock-ci-artifacts.s3.amazonaws.com/30056483033-linux/packages/deb  stable main" | sudo tee /etc/apt/sources.list.d/rocm-test.list
sudo apt update
sudo apt install  amdrocm-debugger-test

rpm package: 
sudo tee /etc/yum.repos.d/rocm-test.repo <<EOF
[rocm-test]
name=Native Linux Package Test Repository
baseurl=https://therock-ci-artifacts.s3.amazonaws.com/30056483033-linux/packages/rpm/x86_64
enabled=1
gpgcheck=0
EOF
sudo dnf clean all
dnf install amdrocm-debugger-test

And then run your test.

lumachad:
> @lumachad the ci is already running deb/rpm package creation and basic test. But may not cover debug tools test. You can verify the installation using the ci build with the following commands.
> 
> Debian package: echo "deb [trusted=yes] https://therock-ci-artifacts.s3.amazonaws.com/30056483033-linux/packages/deb stable main" | sudo tee /etc/apt/sources.list.d/rocm-test.list sudo apt update sudo apt install amdrocm-debugger-test
> 
> rpm package: sudo tee /etc/yum.repos.d/rocm-test.repo <<EOF [rocm-test] name=Native Linux Package Test Repository baseurl=https://therock-ci-artifacts.s3.amazonaws.com/30056483033-linux/packages/rpm/x86_64 enabled=1 gpgcheck=0 EOF sudo dnf clean all dnf install amdrocm-debugger-test
> 
> And then run your test.

Thanks. I actually just want to make sure this unification of two components sharing the same sources won't break something in packaging. rocr-debug-agent testing should be fine as the artifacts will be the combined artifacts of both components.

marbre (commented):
For some other components we're actually considering splitting tests out to their own projects :) However, if here the motivation seems to follow the style ROCgdb follows and that is most likely dictated by upstream or is it a part we controll?

Having tests separated allows to move them out to separate build stages in case they depend on something normally build in parallel without needing to move everything. I think this most likely won't ever apply for rocr-debug-agent?

lumachad:
> For some other components we're actually considering splitting tests out to their own projects :) However, if here the motivation seems to follow the style ROCgdb follows and that is most likely dictated by upstream or is it a part we controll?
> 
> Having tests separated allows to move them out to separate build stages in case they depend on something normally build in parallel without needing to move everything. I think this most likely won't ever apply for rocr-debug-agent?

That's a good data point actually. My motivation to unify the tests with the library is an upcoming round of unit tests we want to add, and the tests would require some library sources to be visible. So unifying would satisfy that. But if we're thinking of going the "separate tests" direction, I can rework our bits with that in mind.

debug agent is not upstream, so we don't have restrictions. Even rocgdb could potentially be split into a separate -tests component as we just copy the testsuite files over.

geomin12 (commented):
lgtm but going to let build/debug folks review

lumachad:
Still considering whether this makes sense based on what @marbre mentioned.

lumachad:
Dropping.

PR #6610, opened 2026-07-16
ci: add no_rocm_image_ubuntu24_04_openmpi test image

minseobshin11:
## Motivation

Several ROCm subproject tests gate on `find_package(MPI)` and launch under `mpiexec`, but the shared `no_rocm` test image doesn't include an MPI implementation — so those tests are silently `DISABLED` in CI. OpenMPI is intentionally not bundled in ROCm artifacts (we don't want to force our MPI onto users, who may run their own), so this PR provides it at the system level in a dedicated test image. The first consumer is rocprofiler-sdk's `rocprofv3` **mpi-ranks** tests; the test-job wiring lands in a follow-up (#5627).

## Technical Details

- Adds `dockerfiles/no_rocm_image_ubuntu24_04_openmpi.Dockerfile`, extending the base `no_rocm_image_ubuntu24_04` image with a system OpenMPI install (`apt install libopenmpi-dev openmpi-bin` — Ubuntu 24.04's OpenMPI 4.1.x).
- Registers the publish job `publish_no_rocm_image_ubuntu24_04_openmpi` in `.github/workflows/publish_no_rocm_image_ubuntu24_04.yml` (mirrors the `_media`/`_rocgdb` extension jobs so it rebuilds when the base image changes).
- Documents the image in `dockerfiles/README.md`.

Published image: `ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi`.

## Test Plan

- Trigger `publish_no_rocm_image_ubuntu24_04.yml` (push to `main`/`stage/docker/**`, or `workflow_dispatch`) and confirm the image builds and publishes.
- Verify the image provides OpenMPI — `mpiexec` on `PATH` and `find_package(MPI)` succeeds.

## Test Result

_Image builds and publishes via `publish_no_rocm_image_ubuntu24_04.yml` 

## JIRA ID

JIRA ID : AIROCVAL-45

## Submission Checklist

### DIFF
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -105,10 +105,13 @@ by ocltst
 | [`no_rocm_image_ubuntu24_04_openmpi.Dockerfile`](no_rocm_image_ubuntu24_04_openmpi.Dockerfile) | https://github.com/ROCm/TheRock/pkgs/container/no_rocm_image_ubuntu24_04_openmpi |
 
 Extended version of no_rocm_image_ubuntu24_04.Dockerfile, containing a system
-OpenMPI install (`libopenmpi-dev`, `openmpi-bin`) required to build and run the
-rocprofv3 `mpi-ranks` integration tests. OpenMPI is intentionally not bundled in
-TheRock artifacts (to avoid forcing a specific MPI onto users), so it is provided
-at the system level here for the test environment.
+OpenMPI install (`libopenmpi-dev`, `openmpi-bin`) for the rocprofv3 `mpi-ranks`
+integration tests. Like the other `no_rocm_image_*` images this is a test-only
+image (not a ROCm build image): the rocprofiler-sdk tests are configured and
+compiled in-container at test time, so `find_package(MPI)` needs the OpenMPI
+headers and `mpiexec`. OpenMPI is intentionally not bundled in TheRock artifacts
+(to avoid forcing a specific MPI onto users), so it is provided at the system
+level here for the test environment.
 
 ### `rocm_runtime.Dockerfile`
 
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -100,6 +100,16 @@ This includes dejagnu, make, gcc, g++ and gfortran.
 Used in ocltst execution. It installs OCL ICD package required
 by ocltst
 
+| Source .Dockerfile                                                                             | Published package                                                                |
+| ---------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


ScottTodd (approved):
This approach makes sense to me, though I haven't interfaced as directly with consumers of the libraries that use MPI to judge from that angle.

@marbre could you review too?

dockerfiles/no_rocm_image_ubuntu24_04_openmpi.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+# no_rocm_image_ubuntu24_04_openmpi:
+# Extend the base no_rocm_image_ubuntu24_04 image with a system OpenMPI install
+# required to build and run the rocprofv3 mpi-ranks integration tests. OpenMPI is
+# intentionally not bundled in TheRock artifacts (to avoid forcing a specific MPI
+# onto users), so it is provided at the system level for the test environment. The
+# corresponding published image is:
+#   ghcr.io/rocm/no_rocm_image_ubuntu24_04_openmpi:latest
+RUN sudo apt-get install -y --no-install-recommends \
+    libopenmpi-dev \
```
marbre, line 13: This pulls in quite some dev dependencies oh which some are already bundled. Probably fine for a test only image **_not intended to build_**.

dockerfiles/README.md
```diff
@@ -100,6 +100,16 @@ This includes dejagnu, make, gcc, g++ and gfortran.
 Used in ocltst execution. It installs OCL ICD package required
 by ocltst
 
+| Source .Dockerfile                                                                             | Published package                                                                |
+| ---------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
+| [`no_rocm_image_ubuntu24_04_openmpi.Dockerfile`](no_rocm_image_ubuntu24_04_openmpi.Dockerfile) | https://github.com/ROCm/TheRock/pkgs/container/no_rocm_image_ubuntu24_04_openmpi |
+
+Extended version of no_rocm_image_ubuntu24_04.Dockerfile, containing a system
+OpenMPI install (`libopenmpi-dev`, `openmpi-bin`) required to build and run the
```
marbre: The PR description says this is intended to test but the docs here say

> required to build

let's clarify this first as it shouldn't be considered as build image. The approach to use a test image dedicated for a specific purpose might be fine though.

minseobshin11: hi I was referring to building the rocprofiler tests and running them, but you are right to flag it. I have reworded it to clarify.

dockerfiles/no_rocm_image_ubuntu24_04_openmpi.Dockerfile
minseobshin11: should be fine as it is a test only image

PR #6611, opened 2026-07-16
feat(quartz): Add notify_quartz to all release workflows

HereThereBeDragons:
## Summary

Wire Quartz CI notifications into the build/release/test workflows so run status flows into Quartz for ingestion. Only touches workflows that are reachable by the release workflows.
Notifications are emitted via the `notify_quartz` composite action (hosted in `ROCm/Quartz`), invoked at the start and end of each workflow, and gated to run only where `vars.NOTIFY_QUARTZ_ENABLED == 'true'`.

Issue: https://github.com/ROCm/Quartz/issues/19

Note: depends on https://github.com/ROCm/Quartz/pull/53, which adds the `notify_quartz` action in ROCm/Quartz

## Changes

- Call the `notify_quartz` composite action from Quartz for `started` and `completed` phases
- remove notify_quartz files that were moved to ROCm/Quartz (`notify_quartz.yml`, `notify_quartz.py`, and its tests).
- Gate every notify step/job on `NOTIFY_QUARTZ_ENABLED`, with `continue-on-error: true` so a Quartz outage never fails a build.
- Forward the Hauly secrets (`GH_APP_HAULY_ID` / `GH_APP_HAULY_PRIVATE_KEY`) to the action, adding `secrets: inherit` on every edge into a notify-bearing workflow. (Job/step will not fail when the secrets are empty)


HereThereBeDragons:
pre-commit is failing due to disabling quartz notify with `if: false`

ScottTodd:
> pre-commit is failing due to disabling quartz notify with `if: false`

https://github.com/rhysd/actionlint/blob/v1.7.12/docs/checks.md#if-cond-constant

you can opt out of actionlint checks as needed here: https://github.com/ROCm/TheRock/blob/5b6302a0c3134ff84dbb3051bdae084dcc9d4459/.pre-commit-config.yaml#L47-L51

HereThereBeDragons:
thanks scott. at the same time i was asking claude and it provided an easier solution:

```
notify:
    if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}  # TEMP: disabled, set repo var NOTIFY_QUARTZ_ENABLED=true or revert to re-enable
    runs-on: ubuntu-24.04
```
> Since NOTIFY_QUARTZ_ENABLED isn't defined as a repo variable, the expression is always false and the job skips, exactly like if: false, but actionlint sees an expression instead of a constant.

ScottTodd:
That's fine too, but someone may see it and treat it as a decision to add the variable.

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -96,6 +96,16 @@ permissions:
...
 jobs:
+  notify_quartz_start:
+    if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}
+    uses: ./.github/workflows/notify_quartz.yml
+    with:
+      run_phase: started
+      reporting_workflow: build_portable_linux_python_packages.yml
+      workflow_inputs: ${{ toJSON(inputs) }}
+      workflow_captured_outputs: ${{ toJSON(needs) }}
+    secrets: inherit
```
cgoea:
```suggestion
    secrets:
      GH_APP_HAULY_ID: ${{ secrets.GH_APP_HAULY_ID }}
      GH_APP_HAULY_PRIVATE_KEY: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
```
limit scope to only these.

HereThereBeDragons:
> vars.NOTIFY_QUARTZ_ENABLED == 'true'

we redecided with marius that we will take this as a global variable in both repos: therock and rockrel to enable/disable quartz. as right now we only want to capture workflows triggering releases we only need to declare it in rockrel.

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -96,6 +96,18 @@ permissions:
 run-name: Build portable Linux Python Packages (${{ inputs.artifact_group }}, ${{ inputs.package_version }})
 
 jobs:
+  notify_quartz_start:
+    if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}
+    uses: ./.github/workflows/notify_quartz.yml
+    with:
+      run_phase: started
+      reporting_workflow: build_portable_linux_python_packages.yml
```
ScottTodd: I don't suppose any variables from https://docs.github.com/en/actions/reference/workflows-and-actions/variables would work here instead of spelling this `reporting_workflow: build_portable_linux_python_packages.yml` out directly?
* `GITHUB_ACTION` 
* `GITHUB_ACTION_PATH`
* `GITHUB_WORKFLOW`

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -96,6 +96,18 @@ permissions:
...
+      workflow_inputs: ${{ toJSON(inputs) }}
+      workflow_captured_outputs: ${{ toJSON(needs) }}
+    secrets:
+      GH_APP_HAULY_ID: ${{ secrets.GH_APP_HAULY_ID }}
+      GH_APP_HAULY_PRIVATE_KEY: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
```
ScottTodd: These secrets won't be available when the workflows run on PRs from forks. How is that handled? Will those workflow runs just not report into Quartz? I think that would be fine, priority is something like:
1. notify for workflow runs from release jobs
2. notify for push events
3. notify for pull_request events (non-fork PR)
4. notify for pull_request events (fork PR)
5. notify for workflow_dispatch (maybe? e.g. dev releases)

.github/workflows/notify_quartz.yml
ScottTodd: > Add `notify_quartz_start` and `notify_quartz_completed` jobs to the build/release/test workflows (29 files). Both call the `notify_quartz.yml` reusable workflow and are gated with `if: github.repository == 'ROCm/rockrel'`, so they are only run for release-related workflows.

Where is that gate? I'm not seeing it. I do see `if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}` at all the call sites though.

.github/workflows/setup_multi_arch.yml
```diff
@@ -126,6 +126,18 @@ permissions:
...
+    if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}
+    uses: ./.github/workflows/notify_quartz.yml
+    with:
+      run_phase: started
+      reporting_workflow: setup_multi_arch.yml
+      workflow_inputs: ${{ toJSON(inputs) }}
+      workflow_captured_outputs: ${{ toJSON(needs) }}
+    secrets:
+      GH_APP_HAULY_ID: ${{ secrets.GH_APP_HAULY_ID }}
+      GH_APP_HAULY_PRIVATE_KEY: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
```
ScottTodd: These jobs are adding a lot of noise to the workflow job graph like on https://github.com/ROCm/TheRock/actions/runs/30363212967/job/90293860606?pr=6611 :/

Some alternate implementation ideas:
* Use a composite action (https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) like https://github.com/ROCm/TheRock/tree/main/.github/actions/setup_test_environment
* Move these start/stop steps to pre- and post- job hooks on the runners themselves: https://docs.github.com/en/actions/how-tos/manage-runners/self-hosted-runners/run-scripts (we might need export the data differently for that to work)

---

Screenshots

Image

Before:

Image

After: 

Image

After, zoomed in: 

Image

.github/workflows/build_portable_linux_python_packages.yml
HereThereBeDragons: right now we are only interested in release workflows. if we are capturing everything we can remove the guard completely

.github/workflows/notify_quartz.yml
HereThereBeDragons: thanks, forgot to update the pr description

.github/workflows/setup_multi_arch.yml
HereThereBeDragons: i agree it creates a lot of noise - and i am open to find another solution that is maintainable

i also had a discussion with @marbre  as i had the same concern. he also suggested to look into composite actions - especially for downstream projects then at a later point but in the first place it would be more important to get quartz started in the first place.

we looked also into:
- having a dedicated workflow triggered on starting and completion of workflows - did not work because you cannot extract the job outputs (and maybe also not the workflow inputs?? dont fully remember)

to you suggestions:
- i dont think post/pre scripts is viable considering how many switching runners we have, that they are taken care of by multiple different groups. see e.g. how long the prolense onboarding takes. you would have to have it as infra as a code and you would have to maintain 2 entry points: one for linux and one for windows.
- will composite actions create less noise? i never created one. we want to have it really at the beginning and at the end of the workflow, independent of the results of the jobs in between. and the collection at the end must require all the other jobs so it can collect all the outputs of the jobs.

.github/workflows/build_portable_linux_python_packages.yml
HereThereBeDragons: i am not sure - the description is also too little ot know. did you try @cgoea?
i can only tell you the workflow path returned by the rest api is the triggering workflow and not the workflow being called via `workflow call`/`uses:`. thats why we introduced it in the first place.

.github/workflows/notify_quartz.yml
HereThereBeDragons: updated it

.github/workflows/setup_multi_arch.yml
ScottTodd: I think a composite action would be less noise, yes. The action code is inserted into the current job, rather than as a separate job.

For example:
* (various workflows) -> `configure_aws_artifacts_credentials`:
  * https://github.com/ROCm/TheRock/actions/runs/30375097819/job/90328783631#step:20:1
  * https://github.com/ROCm/TheRock/blob/aaa447d0f3d8bc5cb8fb7a303b603cf35e53756e/.github/workflows/multi_arch_build_portable_linux_artifacts.yml#L209-L213
  * https://github.com/ROCm/TheRock/blob/main/.github/actions/configure_aws_artifacts_credentials/action.yml
* `.github/workflows/test_component.yml` --> `setup_test_environment`:
  * https://github.com/ROCm/TheRock/actions/runs/30379188423/job/90346368827#step:7:1
  * https://github.com/ROCm/TheRock/blob/aaa447d0f3d8bc5cb8fb7a303b603cf35e53756e/.github/workflows/test_component.yml#L117-L131
  * https://github.com/ROCm/TheRock/blob/main/.github/actions/setup_test_environment/action.yml

Going one step further, we could also have the action provide its own "post" step so we wouldn't need to have each usage location call both `notify_quartz_start` _and_ `notify_quartz_completed`. See https://dev.to/abiwinanda/github-action-adding-post-steps-in-composite-actions-5ak3 . That's a bit awkward with composite actions, so another option would be to publish the Quartz actions on their own (similar to actions/checkout, actions/setup-python, etc.) instead of including all the code here. I haven't done that myself but see the docs at https://docs.github.com/en/actions/how-tos/create-and-publish-actions/manage-custom-actions

cgoea (approved):

.github/workflows/setup_multi_arch.yml
HereThereBeDragons: had a chat with marius. will follow up as soon as i have some presentable results

ScottTodd: Thanks! I don't want to stall this work for too long, but that big job graph with so many extra elements in the UI would be tough to live with 😅

HereThereBeDragons:
ci currently failing because rocm/quartz is still private repo.
example workflow from the test repo: https://github.com/ROCm/Quartz-Tester-rockrel/actions/runs/31016237134

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -107,13 +107,27 @@ jobs:
...
       ARTIFACT_RUN_ID: "${{ inputs.artifact_run_id != '' && inputs.artifact_run_id || github.run_id }}"
       ARTIFACTS_DIR: "${{ github.workspace }}/artifacts"
       PACKAGES_DIR: "${{ github.workspace }}/packages"
       RELEASE_TYPE: ${{ inputs.release_type }}
 
     steps:
+      - name: "Quartz - started ${{ env.QUARTZ_WORKFLOW_POSTFIX }}"
+        continue-on-error: true
+        if: ${{ vars.NOTIFY_QUARTZ_ENABLED == 'true' }}
+        uses: ROCm/Quartz/.github/actions/notify_quartz@main
```
ScottTodd: Can we pin to a commit SHA here instead of using a floating `main`, at least prior to merge? https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#pin-action-uses-versions-to-commit-shas

ScottTodd (commented):
Structurally LGTM. Please let me know when the prerequisite tasks are complete (merging https://github.com/ROCm/Quartz/pull/53 and making that repository visible) so we can review the CI logs on this PR and I can issue a final approval.

HereThereBeDragons:
@ScottTodd it is now in state for the final review.

ScottTodd (approved):

HereThereBeDragons:
ci failures are unrelated infra flakes

PR #6615, opened 2026-07-16
feat(ci): Selecting `runner_labels` from `therock-ci-config` and overlaying on amdgpu_family_matrix.py

geomin12:
Currently, all the amdgpu family data is in `therock-ci-config`. Introducing a new change requires extra work in two repos. However, TheRock should be the source of truth, with dynamic label updates from `therock-ci-config`. 

Now, `therock-ci-config` passes runner label and overlays on top of amdgpu_family_matrix.py (assuming available). This way, we can continue to add more architectures and continue to pull dynamic labels.

Working here: https://github.com/ROCm/TheRock/actions/runs/29756824877/job/88401206226?pr=6615#step:4:69 and https://github.com/ROCm/TheRock/actions/runs/29756824877/job/88418069399?pr=6615#step:6:22

### DIFF
--- a/build_tools/github_actions/amdgpu_family_matrix.py
+++ b/build_tools/github_actions/amdgpu_family_matrix.py
@@ -7,9 +7,9 @@ AMD GPU Family Matrix and runner selection utilities for GitHub workflows.
 Architecture:
 - SUPPORTED BUILD ARCHITECTURES are defined locally in this file (TheRock repo).
   This is the source of truth for what GPU families can be built.
-- CI RUNNER CONFIGURATION is loaded from external config (ROCm/therock-ci-config)
-  when available. This only provides runner labels (test-runs-on, etc.) for where
-  tests can actually be executed.
+- CI RUNNER CONFIGURATION is loaded from external config when available
+  (https://github.com/ROCm/therock-ci-config). This only provides runner labels
+  (test-runs-on, etc.) for where tests can actually be executed.
 
 The external config overlays runner labels onto the local architecture definitions.
 A family can exist for building even without CI runners configured for testing.
@@ -27,6 +27,7 @@ TODO(#2200): clarify AMD GPU family selection
 # NOTE: when doing changes here, also check that they are done in new_amdgpu_family_matrix.py
 #############################################################################################
 
+import copy
 import os
 import random
 import sys
@@ -66,10 +67,6 @@ def load_external_runner_config() -> dict | None:
     config = load_runner_config(config_path)
     _log(f"Loaded external runner config from {ci_config_path}")
     return config
-
-
-# Keep old name as alias for backwards compatibility
-load_external_config = load_external_runner_config
 
 
 def is_asan():
@@ -520,7 +517,18 @@ def _extract_runner_labels_from_v1(external_config: dict) -> dict:
 
 
 def _overlay_runner_config(families: dict, external_config: dict) -> dict:
-    """Overlay external runner configuration onto local family definitions."""
+    """Overlay external runner configuration onto local family definitions.
+
+    Merges runner labels from external config onto local build definitions.
+    Only runner-related keys (test-runs-on, build-runs-on) are overlaid;
+    build architecture support (amdgpu_targets, build_variants) stays local.
+
+    Example transformation for gfx94X:
+        Before (local):
+            "gfx94X": {"linux": {"test-runs-on": "", "amdgpu_targets": "gfx942"}}
+        After (with external config):
+            "gfx94X": {"linux": {"test-runs-on": "linux-mi300-gpu", "amdgpu_targets": "gfx942"}}
+    """
     # V2 format: runner_labels directly available
     # V1 format: extract from gpu_families
     runner_labels = external_config.get("runner_labels", {})
@@ -529,8 +537,6 @@ def _overlay_runner_config(families: dict, external_config: dict) -> dict:
 
     if not runner_labels:
         return families
-
-    import copy
 
     result = copy.deepcopy(families)
     _log("Overlaying runner labels from external config:")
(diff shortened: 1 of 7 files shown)


build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -490,13 +487,101 @@ def get_all_families_for_trigger_types(trigger_types):
...
+    return runner_labels
+
+
+def _overlay_runner_config(families: dict, external_config: dict) -> dict:
+    """Overlay external runner configuration onto local family definitions."""
+    # V2 format: runner_labels directly available
+    # V1 format: extract from gpu_families
+    runner_labels = external_config.get("runner_labels", {})
+    if not runner_labels:
+        runner_labels = _extract_runner_labels_from_v1(external_config)
```
ScottTodd, line 539: I think this design should work for the cases we currently care about.

Looking forward, some ideas:
* We could drop the "runner labels" list here in TheRock down to empty lists or a single value to emphasize that the up to date list of labels is expected to be sourced from the remote configuration file. Or at least add some code comments where there are local values that overrides are expected.
* I could still see us wanting to dynamically set which GPU targets/families are tested by presubmit in different repositories.
  * Let's say we add enough gfx1200 runners to handle presubmit jobs in rocm-libraries so we turn the build on and start using the test runners. Then some time later, we lose those runners and want to switch over to gfx1151. If we only want to build for what we plan on testing, keeping them coupled could make sense.
  * In the prior example, we could also build for both gfx1200 and gfx1151 (assuming good cache hit rates, build runner optimization, and fast source builds from solid coding practices in the math libraries), then just let the runner configuration choose which of those has enough test runner capacity to run the tests.

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -490,13 +487,101 @@ def get_all_families_for_trigger_types(trigger_types):
...
+
+    if not runner_labels:
+        return families
+
+    import copy
```
ScottTodd: Let's move this import to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

* It's not heavy enough to warrant a lazy import
* There is no import cycle to work around

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -490,13 +487,101 @@ def get_all_families_for_trigger_types(trigger_types):
...
+            if platform not in external_family:
+                continue
+
+            # Overlay all keys from runner_labels onto local definitions
+            external_platform = external_family[platform]
+            for key, value in external_platform.items():
+                old_value = family_config[platform].get(key)
+                family_config[platform][key] = value
+                if old_value != value:
+                    _log(f"  {family_name}/{platform}: {key} = {value}")
```
ScottTodd, line 564: I was a little worried about how easy this function would be to understand. It's merging some data and it isn't obvious when scanning the code why we want to merge the data and what specific fields are being changed.

A few things help:
* This logging
* More code comments? You could give an example of the object before and after, in a style like this: https://github.com/ROCm/TheRock/blob/44164a5a330ebd8ae581c1d91f8f4a7251808666/build_tools/github_actions/configure_rocm_python_test_matrix.py#L85-L95

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -4,9 +4,15 @@
 """
 AMD GPU Family Matrix and runner selection utilities for GitHub workflows.
 
-NOTE: The primary source of truth for GPU families and runner labels is
-ROCm/therock-ci-config (runner-config.json). The definitions in this file
-serve as a fallback when external config is not available.
+Architecture:
+- SUPPORTED BUILD ARCHITECTURES are defined locally in this file (TheRock repo).
+  This is the source of truth for what GPU families can be built.
+- CI RUNNER CONFIGURATION is loaded from external config (ROCm/therock-ci-config)
```
ScottTodd: Can you turn this into a full URL to make it easier to follow for developers?

https://github.com/ROCm/therock-ci-config

I've seen this style in comments from coding agents where they put the relevant _text_, but not relevant _links_, and humans really benefit from being able to actually click through to the specific issue, PR, repository, etc. without needing to take the extra step on substituting "github" or doing a web search

build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -54,10 +64,14 @@ def load_external_config() -> dict | None:
         _log(f"CI config not found at {ci_config_path}, using local fallback")
         return None
     config = load_runner_config(config_path)
-    _log(f"Using external CI config from {ci_config_path}")
+    _log(f"Loaded external runner config from {ci_config_path}")
     return config
 
 
+# Keep old name as alias for backwards compatibility
+load_external_config = load_external_runner_config
```
ScottTodd: No backwards compatibility needed within the file itself ;)

(careful with agents doing this)

build_tools/github_actions/amdgpu_family_matrix.py
ScottTodd: Another idea: the CI system could check the dynamic "what runners do we have labels for" data and then determine which build jobs to enable. That way we wouldn't need to also put the "what runs on presubmit", "what runs on postsubmit", etc. decisions in the external config.

geomin12: this is neat! I remember joseph bringing up ideas about "how bad queues are for each label and selecting based on that too" particularly for windows. could be neat and combined with idea there

i'll land this first, then we can work on a better solution here. i think the separation of labels will help a lot in terms of selection and dynamic actions

ScottTodd (approved):

PR #6616, opened 2026-07-16
[CI] Capture and expose GITHUB_STEP_SUMMARY per job

HereThereBeDragons:
## Summary

Adds a mechanism to forward a job's accumulated step summary to a downstream job
via `toJSON(needs)`.

A reusable workflow can only read another job's data through `needs`, which carries
job **outputs**, not the rendered `$GITHUB_STEP_SUMMARY` (GitHub clears that
between steps). To bridge this, every summary append is mirrored to a persistent
per-job file under `$RUNNER_TEMP`, then published as a step output at the end of
the job.

Will be used to report the `$GITHUB_STEP_SUMMARY` to Quartz. 
Part of https://github.com/ROCm/Quartz/issues/4

## Changes

- **New `capture_job_summary.py`**: CLI that publishes the accumulated job summary
  as a step output (default output name `summary`).
- **`github_actions_api.py`**:
  - `gha_job_summary_mirror_path()`: resolves the mirror file under `$RUNNER_TEMP`
    (returns `None` when unset so callers skip mirroring instead of failing).
  - `gha_append_step_summary()`: gains `mirror_to_job_file=True`; each append is
    now also written to the mirror so it can be read back later in the job.
  - `gha_set_job_summary_output()`: reads the mirror and publishes it via
    `gha_set_output`. Empty/missing mirror publishes nothing.
  - `gha_set_output()` heredoc delimiter changed `EOF` to `EOF_mag1c` to avoid
    collisions with summary content that contains `EOF`.
- **Tests**: new `JobSummaryTest` class covering mirror-path resolution, append +
  mirror behavior, `mirror_to_job_file=False`, missing-env handling, and
  round-tripping single/multi-line summaries through the output heredoc.

## Behavior change

`gha_append_step_summary` now mirrors to `$RUNNER_TEMP/job_summary.md` **by
default**, so all existing callers begin writing that file. `$RUNNER_TEMP` is
emptied at job start/end, so the mirror does not leak between jobs. Pass
`mirror_to_job_file=False` to opt out.

## Test plan

- `pytest build_tools/github_actions/tests/github_actions_api_test.py` — 48 passed
- Functionality tested and verified in Quartz devel stub workflows

### DIFF
--- a/build_tools/github_actions/github_actions_api.py
+++ b/build_tools/github_actions/github_actions_api.py
@@ -364,16 +364,40 @@ def gha_set_output(vars: Mapping[str, str | Path]):
         for k, v in vars.items():
             value = "" if v is None else str(v)
             if "\n" in value:
-                f.write(f"{k}<<EOF\n{value}\nEOF\n")
+                # EOF_mag1c delimiter must NOT occure in value
+                f.write(f"{k}<<EOF_mag1c\n{value}\nEOF_mag1c\n")
             else:
                 print(f"OUTPUT {k}={value}")
                 f.write(f"{k}={value}\n")
 
 
-def gha_append_step_summary(summary: str):
+def gha_job_summary_mirror_path() -> Path | None:
+    """Return the per-job summary mirror file under $RUNNER_TEMP, or None.
+
+    GitHub clears $GITHUB_STEP_SUMMARY between steps, so the rendered job
+    summary cannot be read back at the end of a job. We mirror every append
+    into $RUNNER_TEMP/job_summary.md, which persists for the whole job, so a
+    later step can publish it as a job output (see
+    gha_set_job_summary_output). $RUNNER_TEMP is emptied at the start and end
+    of each job, so the mirror does not leak between jobs.
+
+    Returns None when $RUNNER_TEMP is not set (the caller should skip
+    mirroring rather than fail). Local tests point $RUNNER_TEMP at a temp dir.
+    """
+    runner_temp = os.getenv("RUNNER_TEMP")
+    if not runner_temp:
+        return None
+    return Path(runner_temp) / "job_summary.md"
+
+
+def gha_append_step_summary(summary: str, mirror_to_job_file: bool = True):
     """Appends a string to the GitHub Actions job summary.
 
-    This appends to the file located at the $GITHUB_STEP_SUMMARY environment variable.
+    This appends to the file located at the $GITHUB_STEP_SUMMARY environment
+    variable. When mirror_to_job_file is set (the default), the same text
+    is also appended to the persistent mirror at $RUNNER_TEMP/job_summary.md
+    (see gha_job_summary_mirror_path) so the accumulated summary can be read
+    back later in the job and published as an output.
 
     See
       * https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-job-summary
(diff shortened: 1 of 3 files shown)


build_tools/github_actions/github_actions_api.py
```diff
@@ -364,16 +364,40 @@ def gha_set_output(vars: Mapping[str, str | Path]):
...
-def gha_append_step_summary(summary: str):
+def gha_job_summary_mirror_path() -> Path | None:
+    """Return the per-job summary mirror file under $RUNNER_TEMP, or None.
+
+    GitHub clears $GITHUB_STEP_SUMMARY between steps, so the rendered job
+    summary cannot be read back at the end of a job. We mirror every append
+    into $RUNNER_TEMP/job_summary.md, which persists for the whole job, so a
+    later step can publish it as a job output (see
+    gha_set_job_summary_output). $RUNNER_TEMP is emptied at the start and end
+    of each job, so the mirror does not leak between jobs.
```
ScottTodd, line 382: Interesting approach. This will work so long as all of our summary code goes through `gha_append_step_summary`, which should hopefully be the case for TheRock's own scripts. That won't be the case in other projects though, or third party github actions that we call.

How much do we care about those cases? Are there workflows that will run in projects other than TheRock that will want to report the job summary to Quartz too?

There are a few other options mentioned at https://github.com/orgs/community/discussions/27649 that we could consider too:
* https://github.com/ipdxco/job-summary-url-action/ compute the URL to the summary so it could be scraped (since there isn't / wasn't an API for that)
* https://github.com/austenstone/job-summary get the job summary as text and other formats

The job-summary action looks like a pretty good fit, if we're okay with using a third party action for this (we could also fork it)

ScottTodd (commented):
The motivation here LGTM. See my comment - let's make sure we handle all the cases we care about, we may want to use a different implementation approach to catch output not sent via our python function.

build_tools/github_actions/github_actions_api.py
HereThereBeDragons: to be honest we are mainly interested to get the cloudfront/s3 urls we write in the summary in the setup step but do not expose it as outputs.

one would need to check if those suggestions above still work.

there was a time where it was just <url to workflows run id>/summary_raw to access the summary markdown. but this is not the case anymore.
i checked it, and i hope i did it throoughly enough. but more or less what i remember: the job id for the summary is not exposed to be part of the jobs list - or any other obvious connection one can do via the rest api. the only way to get the job id for the summary was to look on the html side of the workflow overview.

so we dont really care about third parties output - we only care about the summary if we define in the summary metrics we do not expose as outputs.

edit: 

another thing if forgot:
considering all the use cases for quartz i would prefer to be as independent as possible from small github actions we dont have control over. 

overall if we have any better solution than this i would welcome it. (like why does github not expose the summary by api?)

HereThereBeDragons: and yes you are right everyone needs then to use `gha_append_step_summary`. the only missing one is the pytorch full test suite.

ScottTodd (approved):

build_tools/github_actions/github_actions_api.py
ScottTodd: This approach is okay given what data you want to collect and how narrowly scoped the changes are. Let's keep an eye on if requirements change and consider the other approaches that scale to all summary text as needed.

PR #6620, opened 2026-07-16
Add structured product-local dependency mirroring

### LINKED ISSUE #6613: Expand for tooling for structured product-local dependency mirroring
## Goal

Currently `update_dependencies.py` mirrors dependencies wheels from PyPI into a flat structure. The goal of this task is to refactor `update_dependencies.py` to mirror dependencies into the core product-local package directory:

```
core/<index>/<normalized-package>/<filename>
```

Examples (from current `PACKAGES_PER_PROJECT`):

```
core/whl/numpy/numpy-2.0.0-cp312-cp312-linux_x86_64.whl
core/whl/networkx/networkx-3.4.2-py3-none-any.whl
core/whl/ml-dtypes/ml_dtypes-0.5.0-cp312-cp312-linux_x86_64.whl
core/whl/setuptools/setuptools-81.0.0-py3-none-any.whl
core/whl/rocm-bootstrap/rocm_bootstrap-...whl
```

### Single pip index, multiple S3 origins

CloudFront makes the central `rocm/whl/` and `rocm/whl-next/` the **single pip `--index-url`**. Every product-local directory feeds that central index via CloudFront routing. From pip's perspective there is only one index. With this, all packages (first-party wheels, dependencies, core, pytorch) can be resolved in one resolution pass. The product-local separation is invisible to pip when targeting the aggregated index.

### All deps go to core

All mirrored deps are placed under `core/<index>/<package>/` regardless of which `PACKAGES_PER_PROJECT["project"]` value they carry. Since every product depends on ROCm core, the core bucket is always reachable for any pip install against the central index. This resolves the shared-dep ownership problem: `numpy` lives in `core/whl/numpy/`, both torch and JAX consume it through CloudFront routing from the same place. No per-project bucket selection needed.

Bucket selection is **unchanged** and we can continue using a single `--bucket` / `S3_BUCKET_PY`, exactly as in flat mode. The `project` field in `PACKAGES_PER_PROJECT` is a **filter only** (controls which subset of packages to process); it never drove bucket selection and still doesn't.

marbre:
## Motivation

The stream-subdomain Python index requires mirrored PyPI dependencies to live in product-local package directories (`core/<index>/<package>/`) so the structured index generator can discover per-package directories, rather than in the flat prefixes used today.

Closes #6613.

## Technical Details

Add an opt-in structured mode to `update_dependencies.py`. This matches a pattern that will be also introduced to
`publish_rocm_to_release_buckets.py` in a separate PR.

* `structured_dependency_key(index, pkg_name, filename)`: pure helper that composes `core/<index>/<normalized-package>/<filename>`. The package directory is derived from the package name (canonical dashed form), not the wheel filename, which keeps its PEP 427 underscore escaping.
* `get_selected_packages(package, dependency_names)`: extract package selection from `run_update_dependencies`. Supports `all` to span every project and keeps the per-project and per-dependency filtering with fail-fast on unknown names.
* `upload_missing_whls` gains `structured`/`index` params; the destination key branches on the mode. `is_wheel_allowed` is retained in both modes: it is a platform/Python-tag filter (`linux_x86_64`/`manylinux`/`py3`/`cp310`-`cp314`), not a package allow-list, so skipping it would mirror macOS, aarch64, musllinux, and PyPy wheels into the index.
* `main()` adds `--structured` and `--index {whl,whl-next}`; `--package` now accepts `all` and defaults to `all` in structured mode, `torch` in flat mode. `--prefix`, `--auto-detect-prefixes`, and `--base-prefix` are rejected under `--structured` via `parser.error`.
* `rocm-bootstrap` moves to project `rocm`; it is a dependency of the `rocm` metapackage, not of `torch`.

## Test Plan

Tests run automatically in CI. Run them manually via:
```
python -m pytest \
  build_tools/third_party/s3_management/tests/update_dependencies_test.py
```

## Test Result

60 passed. Full `s3_management` + `python_package_paths` suite: 136 passed.

## Submission Checklist

### DIFF
--- a/build_tools/third_party/s3_management/tests/update_dependencies_test.py
+++ b/build_tools/third_party/s3_management/tests/update_dependencies_test.py
@@ -210,7 +210,9 @@ def test_resolve_target_prefixes_auto_detect_requires_base_prefix() -> None:
 
 def test_structured_dependency_key_composition() -> None:
     key = structured_dependency_key(
-        "whl", "numpy", "numpy-2.0.0-cp312-cp312-linux_x86_64.whl"
+        index="whl",
+        pkg_name="numpy",
+        filename="numpy-2.0.0-cp312-cp312-linux_x86_64.whl",
     )
     assert key == "core/whl/numpy/numpy-2.0.0-cp312-cp312-linux_x86_64.whl"
 
--- a/build_tools/third_party/s3_management/tests/update_dependencies_test.py
+++ b/build_tools/third_party/s3_management/tests/update_dependencies_test.py
@@ -12,6 +12,7 @@ sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
 import update_dependencies
 from update_dependencies import (
     PACKAGES_PER_PROJECT,
+    _run_structured,
     get_dependency_package_names,
     get_project_paths,
     get_selected_packages,
@@ -286,6 +287,24 @@ def test_get_selected_packages_unknown_dependency_raises() -> None:
     with pytest.raises(ValueError, match="Unknown --dependency-package"):
         get_selected_packages(
             package="torch", dependency_names=frozenset({"not-a-dep"})
+        )
+
+
+# ---------------------------------------------------------------------------
+# _run_structured early validation
+# ---------------------------------------------------------------------------
+
+
+def test_run_structured_rejects_bad_index() -> None:
+    # Guards programmatic callers that bypass argparse choices: fail fast
+    # before any network/upload work rather than mid-run.
+    with pytest.raises(ValueError, match="index="):
+        _run_structured(
+            bucket=FakeBucket(),
+            selected_packages={"numpy": PACKAGES_PER_PROJECT["numpy"]},
+            index="wheels",
+            dry_run=True,
+            only_pypi=False,
         )
 
 
--- a/build_tools/third_party/s3_management/tests/update_dependencies_test.py
+++ b/build_tools/third_party/s3_management/tests/update_dependencies_test.py
@@ -9,12 +9,17 @@ import pytest
 
 sys.path.insert(0, os.fspath(Path(__file__).parent.parent))
 
+import update_dependencies
 from update_dependencies import (
+    PACKAGES_PER_PROJECT,
     get_dependency_package_names,
     get_project_paths,
+    get_selected_packages,
     is_wheel_allowed,
+    main,
     normalize_package_name,
     resolve_target_prefixes,
+    structured_dependency_key,
 )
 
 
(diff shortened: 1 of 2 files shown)


build_tools/third_party/s3_management/tests/update_dependencies_test.py
```diff
@@ -195,3 +201,161 @@ def test_resolve_target_prefixes_auto_detect_requires_base_prefix() -> None:
...
+# ---------------------------------------------------------------------------
+# structured_dependency_key
+# ---------------------------------------------------------------------------
+
+
+def test_structured_dependency_key_composition() -> None:
+    key = structured_dependency_key(
+        "whl", "numpy", "numpy-2.0.0-cp312-cp312-linux_x86_64.whl"
+    )
+    assert key == "core/whl/numpy/numpy-2.0.0-cp312-cp312-linux_x86_64.whl"
```
ScottTodd, line 217: nit: could use keyword args for at least one of these unit tests to make it more clear what the "whl" and "numpy" arguments are used for and why they should appear in that order in the resulting key.

build_tools/third_party/s3_management/update_dependencies.py
ScottTodd, line 1: Do any of the CI workflows use `build_tools/third_party/s3_management/` now that execution has moved server-side?

We could exclude these directories from running CI with a change to https://github.com/ROCm/TheRock/blob/5b6302a0c3134ff84dbb3051bdae084dcc9d4459/build_tools/github_actions/configure_ci_path_filters.py#L207-L230

ScottTodd (approved):

build_tools/third_party/s3_management/update_dependencies.py
marbre: CI workflows currently don't use any of these but I am looking for a workflow that runs `upload_dependencies.py` on schedule instead of running it server side. Anyway, I think we at least want to run unit tests for these scripts and not exclude those but multi-arch CI can be skipped, yes.

build_tools/third_party/s3_management/tests/update_dependencies_test.py
marbre: Done :)

build_tools/third_party/s3_management/update_dependencies.py
ScottTodd: Unit tests always run on every PR, regardless of files modified.

build_tools/third_party/s3_management/update_dependencies.py
```diff
@@ -82,16 +82,88 @@
...
 def normalize_package_name(name: str) -> str:
     """Normalize a Python distribution name for comparison."""
     return re.sub(r"[-_.]+", "-", name).lower()
 
 
+def structured_dependency_key(index: str, pkg_name: str, filename: str) -> str:
+    """Return the structured product-local S3 key for a dependency wheel.
+
+    The package directory is derived from ``pkg_name`` (normalized to the
+    canonical dashed form); ``filename`` is the wheel/sdist basename and keeps
```
araravik-psd: Nit: the docstring says filename is "the wheel/sdist basename," but by the time a filename reaches structured_dependency_key(), it's already passed through is_wheel_allowed(), which
  rejects anything not ending in .whl So sdists never actually flow through this path "wheel basename" would be more accurate?

araravik-psd (approved):
LGTM

build_tools/third_party/s3_management/update_dependencies.py
marbre: ` structured_dependency_key` is itself a pure, extension-agnostic key composer. It doesn't inspect the extension; it'd compose a correct key for an sdist too. So the docstring is describing the function's capability (general) vs. the caller's reality (wheels only) but here only a wheel would reach that path. Will go with a fix.

PR #6622, opened 2026-07-16
Add structured product-local publishing for ROCm core packages

### LINKED ISSUE #6621: Publish ROCm core wheels into structured product-local directories
TheRock's release publish scripts currently place Python wheels in a flat, per-architecture-family S3 layout. The new aggregate stream-subdomain Python indexes (`/rocm/whl/`, `/rocm/whl-next/`) require a structured product-local layout instead:

```
  /rocm/<product>/<index>/<normalized-package>/<filename>
```

The index generator part will be addressed with #6600 and then discovers packages under `<product>/<index>/ roots`, but nothing on the producer side writes wheels into that shape. Without a producer that emits the expected layout, the generator has nothing to index and the aggregate `/rocm/whl/` indexes can't be populated. Therefore we need to teach the ROCm core release publisher scripts to push Python packages into the structured product-local package directories so the layout in S3 matches what `manage_structured.py` expects to discover.

## Details
* Wheels land at `core/<index>/<normalized-package>/<filename>` (e.g. `core/whl/rocm-sdk-core/rocm_sdk_core-...whl`).
* Package directory names are PEP 503-normalized (lowercase; runs of `-`, `_,` `.` collapsed to `-`).
* `<index>` selects `whl` vs `whl-next`.
* Release stream (`dev/nightly/stable/...`) is not encoded in the path - it's selected by S3 bucket selection.

## Scope

* In scope:
  * rocm/core producer side only, behind an opt-in `--structured` flag (default off). Existing flat / legacy per-family and bare-domain `/rocm/...` publishing paths remain unchanged.
* Out of scope:
  * PyTorch/JAX publishers (separate issues).
  * Any workflow/release-script wiring.
  * Deduplication on `whl` and `whl-next`: As of today anything would go to `whl-next` whereas the intended design `whl` should provide a backward compatible index but shouldn't hold copies of anything not needed for backward compatibility (in that case it should be linked to `whl-next`). For ROCm core only a new `rocm` package that pulls in all device packages would qualify to go into `whl`. To be addresses in a follow-up however.

marbre:
## Motivation

The new stream-subdomain repository structure serves Python packages from product-local package directories:

```
    <product>/<index>/<normalized-package>/<filename>
```

where `<index>` is `whl` or `whl-next`. The structured index generator (`manage_structured.py`) discovers per-package directories, but the release publishers still write wheels flat under `v4/whl`. This teaches the ROCm core publisher to place packages into the structured layout so the generator can index them.

Closes #6621.
Depends on #6600.

## Technical Details

Add `build_tools/_therock_utils/python_package_paths.py`, a pure path/key helper shared by the release publishers:

* `pep503_normalize` and `package_name_from_filename` using spec-aware parsing (`packaging.utils` `parse_wheel_filename` / `parse_sdist_filename`) so that sdist/zip filenames with hyphens in the project name are handled correctly. Wheel dist names escape to underscores per PEP 427 so split on first "-" would have worked for wheels, but spec-aware parsing is used throughout for consistency and correctness.
* `structured_key` computing `<product>/<index>/<package>/<filename>`
* `plan_local_uploads` / `plan_key_copies` returning `PlannedUpload` / `PlannedCopy` for the upload (local dir) and copy (S3-to-S3) publisher flows respectively.

Furthermore, this wires an opt-in `--structured` flag (plus `--index {whl,whl-next}`) into `publish_rocm_to_release_buckets.py`. When set, `publish_python_packages` delegates to a structured path that lists the flat source prefix and copies each artifact into
`core/<index>/<package>/<filename>`. Structured publishing is multi-arch only (validated with `parser.error`); the flat and legacy per-family paths are unchanged. The product segment is `"core"` per RFC0012.

PyTorch and JAX publishers use the local-dir upload flow and publish to a different path; they are handled in a follow-up.

## Test Plan

Tests executed in CI. Run locally via
```
python -m pytest \
  build_tools/tests/python_package_paths_test.py \
  build_tools/github_actions/tests/publish_rocm_to_release_buckets_test.py
```

## Test Result

38 targeted tests pass, including a round-trip asserting generated keys are consumable by `manage_structured.discover_packages`.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/python_package_paths.py
+++ b/build_tools/_therock_utils/python_package_paths.py
@@ -1,0 +1,177 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Path/key computation for structured product-local Python package publishing.
+
+The stream-subdomain repository layout places Python artifacts under
+product-local package directories:
+
+    <product>/<index>/<normalized-package>/<filename>
+
+where <index> is ``whl`` or ``whl-next``. Release stream (dev/nightly/
+prerelease) is selected by the target bucket, never encoded in the path. This
+module computes the per-file destination keys the release publishers use when
+run with ``--structured``; the generator in
+``third_party/s3_management/manage_structured.py`` later discovers and indexes
+those directories.
+
+pep503_normalize + package-name extraction here intentionally mirror
+manage_structured.py so producer output round-trips through its
+discover_packages(). They are duplicated (a few lines) rather than shared to
+avoid a dependency from _therock_utils onto third_party/s3_management.
+"""
+
+import dataclasses
+from pathlib import Path
+from re import sub
+
+from _therock_utils.storage_location import StorageLocation
+from packaging.utils import (
+    InvalidSdistFilename,
+    InvalidWheelFilename,
+    parse_sdist_filename,
+    parse_wheel_filename,
+)
+
+
+# Distribution artifacts placed into package directories.
+ACCEPTED_FILE_EXTENSIONS = (".whl", ".tar.gz", ".zip")
+
+# Valid aggregate index names (the second path segment).
+INDEX_NAMES = ("whl", "whl-next")
+
+
+def pep503_normalize(name: str) -> str:
+    """Normalize a package name per PEP 503.
+
+    Lowercase and collapse runs of "-", "_", and "." to a single "-".
+    """
+    return sub(r"[-_.]+", "-", name.lower())
+
+
+def package_name_from_filename(filename: str) -> str:
+    """Return the PEP 503-normalized package name for a distribution artifact.
+
+    Uses spec-aware parsing (packaging.utils) so that sdist/zip filenames
+    with hyphens in the project name are handled correctly (e.g.
+    ``llnl-hatchet-2024.1.tar.gz`` -> ``llnl-hatchet``).
+
+    Raises:
+        ValueError: if the filename cannot be parsed as a wheel or sdist.
+    """
+    if filename.endswith(".whl"):
+        try:
+            name, _, _, _ = parse_wheel_filename(filename)
+            return name
+        except InvalidWheelFilename as e:
+            raise ValueError(f"Cannot parse wheel filename {filename!r}: {e}") from e
+    try:
+        name, _ = parse_sdist_filename(filename)
+        return name
+    except InvalidSdistFilename as e:
+        raise ValueError(f"Cannot parse sdist filename {filename!r}: {e}") from e
+
+
+def is_accepted_artifact(filename: str) -> bool:
+    return filename.endswith(ACCEPTED_FILE_EXTENSIONS)
+
+
+def structured_key(product: str, index: str, filename: str) -> str:
+    """Compute the structured S3 key for an artifact.
+
+    Returns ``<product>/<index>/<normalized-package>/<filename>``.
+
+    Raises:
+        ValueError: if ``index`` is not a valid aggregate index name.
+    """
+    if index not in INDEX_NAMES:
+        raise ValueError(f"index={index!r} is invalid, must be one of {INDEX_NAMES}")
+    package = package_name_from_filename(filename)
+    return f"{product}/{index}/{package}/{filename}"
+
+
+@dataclasses.dataclass
+class PlannedUpload:
+    """A local artifact to upload into a structured package directory.
+
+    Attributes:
+        source: Local path to the artifact.
+        dest: Destination location in the structured layout.
+    """
+
+    source: Path
+    dest: StorageLocation
+
+
+@dataclasses.dataclass
+class PlannedCopy:
+    """An S3 artifact to copy into a structured package directory.
+
+    Attributes:
+        source: Source location (an existing S3 object).
+        dest: Destination location in the structured layout.
+    """
+
+    source: StorageLocation
+    dest: StorageLocation
+
+
+def plan_local_uploads(
+    source_dir: Path,
+    dest_bucket: str,
+    product: str,
+    index: str,
+) -> list[PlannedUpload]:
+    """Plan structured uploads for accepted artifacts in a local directory.
+
+    Enumerates top-level artifacts in ``source_dir`` (not recursive: publish
+    sources are flat directories of wheels/sdists) and computes their structured
+    destinations. Sorted by filename for stable, reviewable output.
+    """
+    return [
+        PlannedUpload(
+            source=path,
+            dest=StorageLocation(
+                dest_bucket, structured_key(product, index, path.name)
+            ),
+        )
+        for path in sorted(source_dir.iterdir())
+        if path.is_file() and is_accepted_artifact(path.name)
+    ]
+
+
+def plan_key_copies(
+    source_keys: list[str],
+    source_bucket: str,
+    dest_bucket: str,
+    product: str,
+    index: str,
+) -> list[PlannedCopy]:
+    """Plan structured copies for accepted artifacts listed from S3.
+
+    Args:
+        source_keys: Full S3 keys returned by a listing (in ``source_bucket``).
+        source_bucket: Bucket the source keys live in.
+        dest_bucket: Destination bucket name.
+        product: Product segment (e.g. ``core``).
+        index: Index segment (``whl`` or ``whl-next``).
+
+    Returns:
+        Planned copies for accepted artifacts, sorted by source key. Only the
+        basename of each source key is used to place the artifact (the
+        structured layout is flat within each package directory).
+    """
+    plans: list[PlannedCopy] = []
+    for key in sorted(source_keys):
+        filename = key.rsplit("/", 1)[-1]
+        if not is_accepted_artifact(filename):
+            continue
+        plans.append(
+            PlannedCopy(
+                source=StorageLocation(source_bucket, key),
+                dest=StorageLocation(
+                    dest_bucket, structured_key(product, index, filename)
+                ),
+            )
+        )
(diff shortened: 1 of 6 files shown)


build_tools/github_actions/publish_rocm_to_release_buckets.py
ScottTodd, line 164: >  Structured publishing is multi-arch only (validated with parser.error); the flat and legacy per-family paths are unchanged.

In a separate PR we can remove the old code paths from this script, see https://github.com/ROCm/TheRock/issues/3340

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -134,9 +137,25 @@ def publish_python_packages(
         kpack split enabled (flat):
         s3://therock-dev-artifacts/12345-linux/python/*.whl
           -> s3://therock-dev-python/v4/whl/*.whl
+
+    When ``structured`` is set (multi-arch only), packages are copied into
+    product-local package directories instead of the flat v4/whl prefix:
+
+        s3://therock-dev-artifacts/12345-linux/python/rocm_sdk_core-...whl
+          -> s3://therock-dev-python/core/<index>/rocm-sdk-core/rocm_sdk_core-...whl
```
ScottTodd: Would a versioned prefix before `core/` still be useful? What if we change the schema later?

How about `v5/core/`? We can still redirect via cloudfront so what gets served to https://rocm.nightlies.amd.com/ (or repo.amd.com with the new domain) uses that directory

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -243,6 +291,20 @@ def main(argv: list[str]) -> None:
         default="false",
         help='Whether kpack split is enabled ("true" or "false")',
     )
+    parser.add_argument(
+        "--structured",
+        action="store_true",
+        help="Publish python packages into product-local package directories "
+        "(core/<index>/<package>/) instead of the flat v4/whl prefix. "
+        "Multi-arch (--kpack-split true) only.",
+    )
```
ScottTodd, line 300: Are we also going to have "structured" directories for other package types? Native Linux packages, tarballs, etc.? If so, let's update the docstrings and add some placeholder code paths to implement that later?

build_tools/github_actions/publish_rocm_to_release_buckets.py
```diff
@@ -243,6 +291,20 @@ def main(argv: list[str]) -> None:
...
+    parser.add_argument(
+        "--index",
+        default="whl",
+        choices=["whl", "whl-next"],
+        help="Product-local index name for structured publishing (default: "
+        "whl). Selects the core/<index>/ path segment.",
+    )
```
ScottTodd, line 307: Let's qualify this as `python-index` or `whl-index` given that this script publishes other package types as well.

Then we may want to add argparse groups for each package type

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -0,0 +1,177 @@
...
+# Valid aggregate index names (the second path segment).
+INDEX_NAMES = ("whl", "whl-next")
+
+
+def pep503_normalize(name: str) -> str:
+    """Normalize a package name per PEP 503.
+
+    Lowercase and collapse runs of "-", "_", and "." to a single "-".
+    """
+    return sub(r"[-_.]+", "-", name.lower())
```
ScottTodd, line 49: This function already exists here: https://github.com/ROCm/TheRock/blob/16a74bea899c6d783eba8114c44dec14c9768aa5/build_tools/packaging/python/manage_structured.py#L144-L149

Let's have `manage_structured.py` import from this utils file?

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -0,0 +1,177 @@
...
+The stream-subdomain repository layout places Python artifacts under
+product-local package directories:
+
+    <product>/<index>/<normalized-package>/<filename>
+
+where <index> is ``whl`` or ``whl-next``. Release stream (dev/nightly/
+prerelease) is selected by the target bucket, never encoded in the path. This
+module computes the per-file destination keys the release publishers use when
+run with ``--structured``; the generator in
+``third_party/s3_management/manage_structured.py`` later discovers and indexes
```
ScottTodd: Update path to https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/manage_structured.py

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -0,0 +1,177 @@
...
+those directories.
+
+pep503_normalize + package-name extraction here intentionally mirror
+manage_structured.py so producer output round-trips through its
+discover_packages(). They are duplicated (a few lines) rather than shared to
+avoid a dependency from _therock_utils onto third_party/s3_management.
```
ScottTodd: Relating to my other comment, let's just add the dep since the code isn't in third_party now? Comments like this can be brittle :P

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -0,0 +1,177 @@
...
+def package_name_from_filename(filename: str) -> str:
+    """Return the PEP 503-normalized package name for a distribution artifact.
+
+    Uses spec-aware parsing (packaging.utils) so that sdist/zip filenames
+    with hyphens in the project name are handled correctly (e.g.
+    ``llnl-hatchet-2024.1.tar.gz`` -> ``llnl-hatchet``).
+
+    Raises:
+        ValueError: if the filename cannot be parsed as a wheel or sdist.
+    """
```
ScottTodd, line 61: Same here, duplicated function: https://github.com/ROCm/TheRock/blob/16a74bea899c6d783eba8114c44dec14c9768aa5/build_tools/packaging/python/manage_structured.py#L74-L95

Utils files are made to be importable by other scripts, let's have this be the one version and import from there.

https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#no-duplicate-code (agents sure like to just copy things instead of reuse, gotta be careful)

build_tools/github_actions/tests/publish_rocm_to_release_buckets_test.py
```diff
@@ -184,6 +185,156 @@ def test_asan_skips_python_packages(self, mock_copy):
...
+                "--structured",
+                "--skip-native-packages",
+                "--dry-run",
+            ]
+        )
+
+        # copy_directory used only for tarballs (not python) under structured.
+        self.assertEqual(mock_copy_dir.call_count, 1)
+        # One copy_file per accepted artifact; index.html is filtered out.
+        self.assertEqual(mock_copy_file.call_count, 3)
```
ScottTodd: These test cases checking exact call counts for implementation detail functions are pretty brittle. Are they actually needed?

We can fill in more details in https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#testing-standards to help agents write better tests, as they tend to overfit to the actual implementation unless you guide them or use test-driven development (write the tests _first_, then the implementation)

ScottTodd (approved):
A few style issues that I'd like addressed, but otherwise lgtm

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: Yeah, this is an oversight be the agent and the current PR doesn't reflect that still want versioned prefixes. I thought about addressing this together with switching to the new buckets in a follow up PR but it is valid to already this here.

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: This is only needed for Python packages :) For tarballs we won't have an aggregated index (or if we would the mechanism would be different) and for native Linux packages it would break the layout assumed by package managers. For the latter the aggregated indexes therefore use a different technique. What need to be changes is the publishing location though but I'd like to address this in a follow up.

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: Fair!

build_tools/_therock_utils/python_package_paths.py
marbre: `managed_structured.py` will be used server-side. If we import form `python_package_paths.py` we need to deploy that as well. As this is just a single line, duplicating seemed the better compromise to me.

build_tools/_therock_utils/python_package_paths.py
marbre: Good catch!

build_tools/_therock_utils/python_package_paths.py
marbre: Yeah, this code was drafted before we moved it to the new home. I would rather not add the dep though.

build_tools/_therock_utils/python_package_paths.py
marbre: This was not the agents decision and rather mine as it makes deployment easier. I will land as is and but file an issue to re-evaluate sharing the code.

build_tools/github_actions/tests/publish_rocm_to_release_buckets_test.py
marbre: Funnily this was TDD driven. I will check before landing and eventually drop.

build_tools/github_actions/publish_rocm_to_release_buckets.py
marbre: I think this needs a quick offline discussion :)

PR #6646, opened 2026-07-17
build: wire TEST_SUBPROJECTS for math-libs and ml-libs components

endurthiabhilash:
As part of enabling multi-arch on rocm-libraries #3343, adding TEST_SUBPROJECTS entries to therock_cmake_subproject_declare blocks so downstream consumers are tracked for testing:

- hipBLASLt: hipblasltprovider, MIOpen, rocBLAS
- rocBLAS: hipblasltprovider, hipSPARSE, MIOpen, rocSOLVER, rocSPARSE, rocWMMA
- rocSPARSE: rocPRIM, rocBLAS
- hipBLAS: MIOpen, rocWMMA, hipSOLVER, hipSPARSE, rocSPARSE
- rocPRIM: rocSOLVER
- libhipcxx: rocThrust
- composable_kernel: MIOpen
- MIOpen: hipDNN_samples, hipdnn_integration_tests
- miopenprovider: hipDNN_samples

## Submission Checklist

### DIFF
--- a/math-libs/BLAS/CMakeLists.txt
+++ b/math-libs/BLAS/CMakeLists.txt
@@ -160,7 +160,9 @@ therock_cmake_subproject_declare(hipBLASLt
     ${optional_profiler_deps}
   TEST_SUBPROJECTS
     hipSPARSELt
+    hipblasltprovider
     MIOpen
+    rocBLAS
 )
 therock_cmake_subproject_glob_c_sources(hipBLASLt
   SUBDIRS
@@ -238,7 +240,12 @@ therock_cmake_subproject_declare(rocBLAS
     ${optional_profiler_deps}
   TEST_SUBPROJECTS
     hipBLAS
+    hipblasltprovider
+    hipSPARSE
+    MIOpen
     rocSOLVER
+    rocSPARSE
+    rocWMMA
 )
 therock_cmake_subproject_glob_c_sources(rocBLAS
 SUBDIRS
@@ -283,6 +290,8 @@ if(THEROCK_ENABLE_SPARSE)
       hipSPARSE
       rocSOLVER
       hipSOLVER
+      rocPRIM
+      rocBLAS
   )
   therock_cmake_subproject_glob_c_sources(rocSPARSE
     SUBDIRS
@@ -515,6 +524,11 @@ therock_cmake_subproject_declare(hipBLAS
     rocBLAS
     ${hipBLAS_optional_deps}
   TEST_SUBPROJECTS
+    MIOpen
+    rocWMMA
+    hipSOLVER
+    hipSPARSE
+    rocSPARSE
 )
 therock_cmake_subproject_glob_c_sources(hipBLAS
 SUBDIRS
--- a/math-libs/CMakeLists.txt
+++ b/math-libs/CMakeLists.txt
@@ -124,6 +124,7 @@ if(THEROCK_ENABLE_PRIM)
       hipCUB
       rocThrust
       rocSPARSE
+      rocSOLVER
   )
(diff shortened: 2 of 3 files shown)


geomin12 (commented):
lgtm, but let's get component teams to review and approve as well pls

ScottTodd:
Wait why do we need a dedicated variable for this? Can't we get this information from the build graph?

If A depends on B, we should build/test A when B changes. We don't need B to be aware that A depends on it.

geomin12:
> Wait why do we need a dedicated variable for this? Can't we get this information from the build graph?
> 
> If A depends on B, we should build/test A when B changes. We don't need B to be aware that A depends on it.

purpose of the variable is to get granular component and their dependencies. I wanted to use build_topology.toml, but it was too broad. 

previously in rocm-libraries impl, for example, any blas (rocblas, hipblaslt, hipblas) update, we would build all of BLAS and test all of BLAS for all updates. however, ROCm devs let me know this is not efficient (ex: for rocblas updates, we don't need to test hipblaslt)! i looked around and couldn't find the granularity i wanted, so ended with this impl.

@endurthiabhilash and I are happy to place it somewhere else, or if there is a way to pull this data more dynamically, let me know. it's fairly hard-coded and minimal impact

ScottTodd:
Yeah, the build topology is not the right tool to use here.

In the CMake next to where `TEST_SUBPROJECTS` was added there are `BUILD_DEPS` and `RUNTIME_DEPS` lists. I think the inverse of this should work:

https://github.com/ROCm/TheRock/blob/44164a5a330ebd8ae581c1d91f8f4a7251808666/cmake/therock_subproject.cmake#L1482-L1527

that says "for a given project, what are all the transitive _dependencies_" when here you want "for a given project, what are all the transitive _consumers_". May need a subproject registry if one doesn't already exist to be able to compute that.

The explicit `TEST_SUBPROJECTS` list _would_ let you do things like "even though rocBLAS depends on hipBLASLt, don't test it for changes to hipBLASLt", but IDK if we want that?

geomin12:
> Yeah, the build topology is not the right tool to use here.
> 
> In the CMake next to where `TEST_SUBPROJECTS` was added there are `BUILD_DEPS` and `RUNTIME_DEPS` lists. I think the inverse of this should work:
> 
> https://github.com/ROCm/TheRock/blob/44164a5a330ebd8ae581c1d91f8f4a7251808666/cmake/therock_subproject.cmake#L1482-L1527
> 
> that says "for a given project, what are all the transitive _dependencies_" when here you want "for a given project, what are all the transitive _consumers_". May need a subproject registry if one doesn't already exist to be able to compute that.
> 
> The explicit `TEST_SUBPROJECTS` list _would_ let you do things like "even though rocBLAS depends on hipBLASLt, don't test it for changes to hipBLASLt", but IDK if we want that?

ack! this hard-coded item is matching the `therock-matrix.py` (but with more guidance from ROCm component teams). 

I'm planning to swap CI systems over for superrepos (so we get it done). this will work for now, then we can work on swapping this out this portion to dynamically select and make it more sophisticated

@endurthiabhilash and I will create GH issue

geomin12 (approved):
as mentioned here (https://github.com/ROCm/TheRock/pull/6646#issuecomment-5023815335), this is not an ideal solution. However, this is a replicate of `therock-matrix.py` and this will apply to multi-arch.

After multi-arch migration happens, @endurthiabhilash and I are going to work to improve this retrieval to be dynamic. Especially if https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0003-Build-Tree-Normalization.md occurs, we want our solution to be workable for both CMakeLists scenarios

PR #6359, opened 2026-07-06
ci(wsl-rocdxg): build inside manylinux container via Docker in WSL

dysu-amd:
### Summary

Reworks the multi_arch_build_wsl_rocdxg_artifacts.yml workflow so librocdxg.so builds inside TheRock's manylinux Docker image via Docker in WSL, instead of directly in WSL.

Requires the updated wkmi artifact from https://github.com/ROCm/rocm-systems/pull/8323

Resolves https://github.com/ROCm/TheRock/issues/6245 and https://github.com/ROCm/TheRock/issues/6509

### Motivation

Running the build directly in the WSL environment meant building against glibc as provided by Ubuntu 24, which is newer than the glibc provided in TheRock's manylinux image. This could cause errors when linking rocdxg against libraries built with manylinux, which is most (if not all) of them. Building rocdxg inside the manylinux image would resolve this.

### Changes

- Containerized configure & build stages. The Configure and Build stage steps now `docker run` the pinned manylinux image (MANYLINUX_IMAGE, referenced by digest), bind-mounting the workspace and build dir with src==dst so the in-container cmake invocation matches a native WSL one. Inside the container it installs requirements.txt (+ any staged pip deps), then runs the same cmake configure/build.
- Docker-in-WSL bring-up. New Start Docker in WSL step launches dockerd detached (no init system under WSL2), polls docker info until the socket is ready, and pulls the image. WSL additional-packages trimmed to docker.io python3 python3-pip python3-psutil python3-venv — the C/C++ toolchain now lives in the container.
- Windows SDK mounting. The SDK shared-include dir is bind-mounted read-only at a clean, space-free path (THEROCK_WSL_WIN_SDK_MOUNT=/opt/winsdk-shared) and passed via -DTHEROCK_WSL_WIN_SDK, replacing the previous host-side symlink workaround for the (x86) path segment. WSLENV extended to propagate MANYLINUX_IMAGE and the mount path.

### Test plan

- [x] Ensure the rocdxg WSL CI passes. (https://github.com/ROCm/TheRock/actions/runs/29254598695/job/86843570590?pr=6359)
- [x] Ensure other relevant CI passes.

PR #6658, opened 2026-07-17
ci(wsl-rocdxg): run builds in WSL-native directory

### LINKED ISSUE #6657: [Issue]: WSL ROCDXG CMake configure stage is 9x slower (5 -> 45 mins)
### Description:
WSL ROCDXG stage is seeing a ~9x increase in CMake configure times after https://github.com/ROCm/TheRock/pull/6359 was merged in.

Before (~5 mins):
https://github.com/ROCm/rockrel/actions/runs/29132229530/job/86491811468#step:19:287

After (~45 mins):
https://github.com/ROCm/rockrel/actions/runs/29378454574/job/87239551560#step:20:1008

### Cause:

Possibly due to interactions between Windows/WSL/Docker filesystems, accessing Windows-native files from within WSL is known to be very slow, plus we are also mounting it in a Docker container ([ref](https://stackoverflow.com/questions/68972448/why-is-wsl-extremely-slow-when-compared-with-native-windows-npm-yarn-processing))

dysu-amd:
### Summary

Moves the rocdxg build driectory from `/mnt/c` (Windows-native) into a native WSL directory, mounting that into Docker instead. This addresses the ~9x increase in rocdxg CMake configure times introduced in https://github.com/ROCm/TheRock/pull/6359

Resolves https://github.com/ROCm/TheRock/issues/6657

### Details

- Keep git checkout on Windows side (via `actions/checkout`) to avoid needing to inject an auth token in WSL
- TheRock git diffs every subproject, which introduces a lot of overhead when run inside WSL on Windows-native files
- After checkout, sequentially copy source files from Windows into WSL with rsync
- The build now runs in WSL on WSL-native files, making it faster than even the previous baseline (from 15 to 9 mins)

### Test Plan

Previous baseline: https://github.com/ROCm/rockrel/actions/runs/29132229530/job/86491811468
New results: https://github.com/ROCm/TheRock/actions/runs/29767265737/job/88660682653

- [x] See if CMake configure times improve
- [x] Make sure the same three artifacts are still being uploaded

### DIFF
--- a/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
+++ b/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
@@ -198,9 +198,6 @@ jobs:
           THEROCK_WIN_SDK_SHARED: C:\Program Files (x86)\Windows Kits\10\Include\${{ steps.detect_sdk.outputs.sdk_version }}\shared
           WSLENV: THEROCK_HOST_WORKSPACE/p:THEROCK_WIN_SDK_SHARED/p
         run: |
-          # The source and build trees live on the WSL-native ext4 filesystem
-          # (see the "Sync source tree to WSL" step for why), so they are not
-          # bridged /mnt/c paths.
           wsl_build_dir="${HOME}/therock-wsl-rocdxg/build"
           if [ ! -d "${THEROCK_WIN_SDK_SHARED}" ]; then
             echo "Windows SDK include path not found: ${THEROCK_WIN_SDK_SHARED}"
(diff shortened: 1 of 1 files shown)


dysu-amd:
CI failures are all known and being tracked
https://github.com/ROCm/TheRock/issues/6720
https://github.com/ROCm/rocm-libraries/issues/9545
https://github.com/ROCm/therock/issues/4617

jayhawk-commits (approved):
Thanks for reducing the configure build time with these changes. No blocking concerns on my end.

PR #6698, opened 2026-07-20
Add structured product-local publishing for PyTorch wheels

marbre:
## Motivation

Add an opt-in `--structured` mode to the PyTorch publisher that uploads wheels into product-local package directories
(`v5/rocm/pytorch/<index>/<package>/`) via `plan_local_uploads`, mirroring the ROCm core publisher.

Closes #6691.

## Technical Details

 `--python-index` selects `whl` or `whl-next`. The default (flat `v4/whl`) path and the `package_index_url` output are unchanged; the index-URL evolution is deferred to the workflow wiring task.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6699, opened 2026-07-20
Add structured product-local publishing for JAX wheels

### LINKED ISSUE #6692: Publish JAX wheels into structured product-local directories
Similar as for ROCm wheel, #6621, JAX wheels need be published to

```
  /rocm/jax/<index>/<normalized-package>/<filename>
```
as the new aggregate stream-subdomain Python indexes (`/rocm/whl/`, `/rocm/whl-next/`) require a structured product-local layout.

The index generator part was addressed with b92ee0a (#6621) and #6686.  The scripts used to publish JAX need to be expanded to push Python packages into the structured product-local package directories so the layout in S3 matches what `manage_structured.py` expects to discover.

## Details
* Wheels land at `/v5/rocm/jax/<index>/<normalized-package>/<filename>` (accessible via CF as e.g. `/rocm/jax/whl/jax-rocm7-pjrt/jax_rocm7_pjrt...whl`).
* Package directory names are PEP 503-normalized (lowercase; runs of `-`, `_,` `.` collapsed to `-`).
* `<index>` selects `whl` vs `whl-next`.
* Release stream (`dev/nightly/stable/...`) is not encoded in the path - it's selected by S3 bucket selection.
* To be gated behind an opt-in `--structured` flag (default off). Existing flat / legacy per-family and bare-domain publishing paths remain unchanged.

marbre:
## Motivation

Add an opt-in `--structured` mode to the JAX publisher that uploads wheels into product-local package directories
(`v5/rocm/jax/<index>/<package>/`) via `plan_local_uploads`, mirroring the ROCm core and PyTorch publishers. 

Closes #6692.

## Technical Details

`--python-index` selects `whl` or `whl-next`. The default (flat `v4/whl`) path and the `package_index_url` output are unchanged; the index-URL evolution is deferred to the workflow wiring task.

## Test Plan

Also add a publisher test module (the JAX publisher previously had none), covering both the flat and structured paths.

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/tests/publish_jax_to_release_bucket_test.py
+++ b/build_tools/github_actions/tests/publish_jax_to_release_bucket_test.py
@@ -1,0 +1,241 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Unit tests for publish_jax_to_release_bucket.py."""
+
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from unittest import mock
+
+sys.path.insert(0, os.fspath(Path(__file__).parent.parent.parent))
+
+from github_actions.publish_jax_to_release_bucket import main
+
+
+class TestPublishJaxToReleaseBucket(unittest.TestCase):
+    """Tests for the main() CLI entry point."""
+
+    def setUp(self):
+        # Real directory so the script's existence check passes; the
+        # upload itself is mocked so no S3 contact happens.
+        self._tmp = tempfile.TemporaryDirectory()
+        self.source_dir = Path(self._tmp.name)
+
+    def tearDown(self):
+        self._tmp.cleanup()
+
+    def _touch(self, name: str) -> None:
+        (self.source_dir / name).write_bytes(b"")
+
+    # -----------------------------------------------------------------------
+    # Flat publishing (default)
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/tests/publish_jax_to_release_bucket_test.py
```diff
@@ -0,0 +1,241 @@
...
+                "--source-dir",
+                os.fspath(self.source_dir),
+                "--release-type",
+                "dev",
+                "--dry-run",
+            ]
+        )
+        self.assertEqual(mock_upload.call_count, 1)
+        _source, dest = mock_upload.call_args.args
+        self.assertEqual(dest.relative_path, "v4/whl")
```
ScottTodd: Redundant test case with `test_dev_uploads_to_v4_whl_in_dev_python`, as on https://github.com/ROCm/TheRock/pull/6698#discussion_r3616460697

ScottTodd (approved):
We may want to merge the jax and torch scripts so we aren't maintaining all this code for both frameworks.

erman-gurses (approved):

build_tools/github_actions/tests/publish_jax_to_release_bucket_test.py
marbre: Dropped it.

PR #6708, opened 2026-07-20
ci(jax): release v0.10.0-v0.11.0, build single ref in PR CI, drop v0.9.1/native

gulsumgudukbay:
Issue: https://github.com/ROCm/TheRock/issues/6710
## Motivation

Bring the TheRock JAX CI/release version set up to date. The **release** workflow should build/test `rocm-jaxlib-v0.10.0`, `v0.10.1`, `v0.10.2`, and `v0.11.0`, and drop `v0.9.1` (the last version using the legacy `native` `ROCm/rocm-jax` build
flow). Everything `0.10.0+` uses the manylinux flow from `ROCm/jax`. **PR CI** builds a single, most-stable ref to keep runner load low.

## JAX versions: CI vs Release

| Workflow | `release_type` | JAX refs built |
| --- | --- | --- |
| **Release** (dev / nightly / prerelease) | `dev`, `nightly`, `prerelease` | `0.10.0`, `0.10.1`, `0.10.2`, `0.11.0` |
| **PR CI** | `ci` | `0.11.0` only (default / most stable) |

Following the PyTorch CI convention, the base CI config favors breadth across GPU targets/platforms/frameworks, so PR CI builds a single ref (additional refs can be opted in as needed). `0.11.0` is the default.

## Changes

**Version matrix (`build_tools/github_actions/configure_jax_release_matrix.py`)**
- `RELEASE_JAX_REFS`: `0.10.0, 0.10.1, 0.10.2, 0.11.0` (added `v0.10.1` and `v0.11.0`; removed `v0.9.1`).
- `CI_JAX_REFS`: single ref `v0.11.0`.
- Python version support is now **declarative per ref** via `exclude_python_versions` in `JAX_REF_CONFIGS` (JAX dropped Python 3.11 in `0.11.0`), replacing the previous regex-on-ref-name logic.

**Remove the legacy `native` build flow (only `0.9.1` used it)**
- Drop the `build_mode` key from the matrix generator, the JAX build/release/CI workflows, the CI summary table, and the unit tests.
- Delete the dead native build step and the CI build-mode validation step; simplify the now-unconditional manylinux steps.
- Remove the orphaned `tar_url` input from the JAX build/release chain. The manylinux flow installs ROCm from the pip `--find-links` index, not a TheRock tarball.
- `trigger_release_jax_wheels` no longer `needs: build_tarballs` (it only consumes `build_python_packages`), so the JAX release no longer waits on / is skipped by the tarball job.

**Test runner (`build_tools/github_actions/configure_target_run.py`)**
- JAX tests now run on the family's multi-GPU runner (`test-runs-on-multi-gpu`), i.e. `linux-gfx942-8gpu-ossci-rocm` for `gfx94X-dcgpu`, falling back to the single-GPU runner where a family has none.

**Refresh stale defaults**
- Update stale `jax_ref` `workflow_dispatch` defaults (`0.8.2`/`0.9.1`) to `rocm-jaxlib-v0.11.0` in `test_linux_jax_wheels_partial.yml`, `test_linux_jax_wheels.yml`, and `test_jax_dockerfile.yml`.

**Docs**
- `RELEASES.md`: supported JAX versions `0.10.0–0.11.0`, default `JAX_VERSION=0.11.0`.
- `external-builds/jax/README.md`: manylinux-only flow; dropped `0.9.1`/native content; refreshed the supported-version table; clarified upstream (`jax-ml/jax`) vs downstream (`ROCm/jax`) vs infra (`ROCm/rocm-jax`); added the real `build/build.py` build command and the `ci/run_pytest_rocm.sh` test command.

## Test Plan

- `pytest build_tools/github_actions/tests/configure_jax_release_matrix_test.py` - 8 passed.
- `pytest build_tools/github_actions/tests/configure_target_run_test.py` - 4 passed.
- Full CI-config/summary suite (`-k "summary or multi_arch_ci or jax"`) - passed.
- `yaml.safe_load` on all edited workflows - all parse.

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_linux_jax_wheels.yml
+++ b/.github/workflows/multi_arch_build_linux_jax_wheels.yml
@@ -29,10 +29,6 @@ on:
         description: "Repository containing the JAX release branch."
         type: string
         required: true
-      build_mode:
-        description: "Build mode: native or manylinux."
-        type: string
-        required: true
       gfx_arch:
         description: "ROCm device package selector used for the manylinux build (e.g. device-gfx942, device-gfx950, or device-all)."
         type: string
@@ -50,10 +46,6 @@ on:
         description: "URL for pip --find-links to install ROCm packages for the JAX manylinux build."
         type: string
         default: ""
-      tar_url:
-        description: "URL to the TheRock tarball used for the build."
-        type: string
-        required: true
       release_type:
         description: "Release type."
         type: string
@@ -92,13 +84,6 @@ on:
         description: "GitHub repository in owner/repo format (e.g. ROCm/jax or ROCm/rocm-jax)."
         type: string
         default: "ROCm/jax"
-      build_mode:
-        description: "Build mode."
-        type: choice
-        options:
-          - native
-          - manylinux
-        required: true
       gfx_arch:
         description: "ROCm device package selector used for the manylinux build (e.g. device-gfx942, device-gfx950, or device-all)."
         type: string
@@ -116,10 +101,6 @@ on:
         description: "URL for pip --find-links to install ROCm packages for the JAX manylinux build."
         type: string
         default: ""
-      tar_url:
-        description: "URL to the TheRock tarball used for the build."
-        type: string
-        required: true
       release_type:
         description: "Release type."
         type: string
@@ -166,7 +147,6 @@ jobs:
           ref: ${{ inputs.jax_ref }}
 
       - name: Checkout JAX
-        if: ${{ inputs.build_mode == 'manylinux' }}
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           path: jax-source
@@ -190,30 +170,11 @@ jobs:
 
       - name: Set package dist dir
         run: |
-          if [ "${{ inputs.build_mode }}" = "manylinux" ]; then
-            echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
-          else
-            echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/rocm-jax/jax_rocm_plugin/wheelhouse" >> "$GITHUB_ENV"
-          fi
+          echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
 
-      - name: Build JAX Wheels (native, v0.9.1)
-        if: ${{ inputs.build_mode == 'native' }}
-        working-directory: rocm-jax
-        env:
-          ROCM_VERSION: ${{ inputs.rocm_version }}
-          PYTHON_VERSION: ${{ inputs.python_version }}
-          TAR_URL: ${{ inputs.tar_url }}
-        run: |
-          python3 build/ci_build \
-            --compiler=clang \
-            --python-versions="${PYTHON_VERSION}" \
-            --rocm-version="${ROCM_VERSION}" \
-            --therock-path="${TAR_URL}" \
-            dist_wheels
       # TODO: THEROCK_INDEX_URL need to be renamed in the future to
       # THEROCK_FIND_LINKS_URL to avoid confusion with pip --index-url.
       - name: Build manylinux image
-        if: ${{ inputs.build_mode == 'manylinux' }}
         working-directory: jax-source
         env:
           THEROCK_VERSION: ""
@@ -238,7 +199,6 @@ jobs:
             --rocm-version "${{ inputs.rocm_version }}"
 
       - name: Build JAX Wheels in manylinux container
-        if: ${{ inputs.build_mode == 'manylinux' }}
         working-directory: jax-source
         env:
           ROCM_VERSION: ${{ inputs.rocm_version }}
--- a/.github/workflows/multi_arch_build_linux_jax_wheels.yml
+++ b/.github/workflows/multi_arch_build_linux_jax_wheels.yml
@@ -131,6 +131,7 @@ jobs:
       jax_pjrt_version: ${{ steps.write_jax_versions.outputs.jax_pjrt_version }}
     env:
       MANYLINUX_IMAGE_TAG: therock-jax-manylinux:${{ inputs.jax_ref }}-${{ inputs.python_version }}
+      PACKAGE_DIST_DIR: ${{ github.workspace }}/jax-source/dist
 
     steps:
       - name: Checkout TheRock
@@ -167,10 +168,6 @@ jobs:
       - name: Install python deps for CI
         run: |
           pip install -r external-builds/jax/requirements-jax.txt
-
-      - name: Set package dist dir
-        run: |
-          echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
 
       # TODO: THEROCK_INDEX_URL need to be renamed in the future to
       # THEROCK_FIND_LINKS_URL to avoid confusion with pip --index-url.
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -256,7 +256,9 @@ jobs:
             }
 
   trigger_release_jax_wheels:
-    needs: [build_artifacts, build_tarballs, build_python_packages]
+    # The manylinux JAX build only consumes build_python_packages output
+    # (find-links index); it no longer needs the tarball job.
+    needs: [build_artifacts, build_python_packages]
     name: Trigger Release JAX Wheels
     if: ${{ inputs.build_jax != false && fromJSON(inputs.build_config).build_jax == true }}
     runs-on: ubuntu-24.04
@@ -273,7 +275,6 @@ jobs:
               "release_type": "${{ inputs.release_type }}",
               "rocm_version": "${{ inputs.rocm_package_version }}",
               "rocm_package_find_links_url": "${{ needs.build_python_packages.outputs.package_find_links_url }}",
-              "tar_url": "${{ fromJSON(needs.build_tarballs.outputs.tarball_urls).multiarch }}",
               "repository": "${{ inputs.repository || github.repository }}",
               "ref": "${{ inputs.ref || '' }}",
               "python_version": "${{ inputs.python_version }}"
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -256,8 +256,6 @@ jobs:
             }
 
   trigger_release_jax_wheels:
-    # The manylinux JAX build only consumes build_python_packages output
-    # (find-links index); it no longer needs the tarball job.
     needs: [build_artifacts, build_python_packages]
     name: Trigger Release JAX Wheels
     if: ${{ inputs.build_jax != false && fromJSON(inputs.build_config).build_jax == true }}
--- a/build_tools/github_actions/configure_jax_release_matrix.py
+++ b/build_tools/github_actions/configure_jax_release_matrix.py
@@ -24,23 +24,27 @@ CI_PYTHON_VERSIONS = {
 }
 
 JAX_REF_CONFIGS = {
-    "rocm-jaxlib-v0.9.1": {
-        "jax_ref": "rocm-jaxlib-v0.9.1",
-        "jax_repository": "ROCm/rocm-jax",
-        "build_mode": "native",
-        "gfx_arch": "",
-    },
     "rocm-jaxlib-v0.10.0": {
         "jax_ref": "rocm-jaxlib-v0.10.0",
         "jax_repository": "ROCm/jax",
-        "build_mode": "manylinux",
+        "gfx_arch": "device-all",
+    },
+    "rocm-jaxlib-v0.10.1": {
+        "jax_ref": "rocm-jaxlib-v0.10.1",
+        "jax_repository": "ROCm/jax",
         "gfx_arch": "device-all",
     },
     "rocm-jaxlib-v0.10.2": {
         "jax_ref": "rocm-jaxlib-v0.10.2",
         "jax_repository": "ROCm/jax",
-        "build_mode": "manylinux",
         "gfx_arch": "device-all",
+    },
+    "rocm-jaxlib-v0.11.0": {
+        "jax_ref": "rocm-jaxlib-v0.11.0",
+        "jax_repository": "ROCm/jax",
+        "gfx_arch": "device-all",
+        # JAX dropped Python 3.11 support in 0.11.0.
+        "exclude_python_versions": ["3.11"],
     },
 }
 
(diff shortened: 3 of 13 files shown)


erman-gurses:
Testing the release workflow in here: https://github.com/ROCm/TheRock/actions/runs/29781831449/job/88484790728

erman-gurses:
`py 3.11- 0.11.0` was failed -  will be disabled
https://github.com/ROCm/TheRock/actions/runs/29853852504/job/88742021045

.github/workflows/multi_arch_release_linux.yml
erman-gurses: Since this PR removes tar_url and the native JAX build path, does trigger_release_jax_wheels still need to depend on build_tarballs?

The manylinux JAX build now only consumes build_python_packages.outputs.package_find_links_url, so keeping build_tarballs in `needs` makes the JAX workflow wait unnecessarily for the tarball job. It also can cause skipping the JAX release workflow when a tarball job fails.

erman-gurses (commented):
Looks good overall -  please see my comments

We also need to update these files:
https://github.com/ROCm/TheRock/blob/main/RELEASES.md
https://github.com/ROCm/TheRock/blob/main/external-builds/jax/README.md

build_tools/github_actions/configure_jax_release_matrix.py
```diff
@@ -95,6 +102,10 @@ def generate_jax_matrix(
     matrix: list[dict[str, str]] = []
     for py in python_versions:
         for ref in jax_refs:
+            # JAX dropped Python 3.11 in 0.11.0 and later.
+            ver = re.search(r"v(\d+)\.(\d+)", ref)
+            if py == "3.11" and ver and (int(ver[1]), int(ver[2])) >= (0, 11):
+                continue
```
erman-gurses: This looks little brittle  -  we could make the supported Python versions declarative in JAX_REF_CONFIGS instead of deriving them from the ref name.
CC: @ScottTodd

build_tools/github_actions/configure_jax_release_matrix.py
```diff
@@ -51,17 +54,21 @@
...
-# CI uses the manylinux package path only. Exclude rocm-jaxlib-v0.9.1 because
-# it uses the legacy native/tarball flow.
+# CI builds the same JAX refs as release. All supported refs use the manylinux
+# package path.
 CI_JAX_REFS = {
     "linux": [
         "rocm-jaxlib-v0.10.0",
+        "rocm-jaxlib-v0.10.1",
+        "rocm-jaxlib-v0.10.2",
+        "rocm-jaxlib-v0.11.0",
```
erman-gurses: Do we have to add all of the JAX versions for CI Workflow?

ScottTodd: Please keep this limited to only a single ref for CI builds. We can add a way to opt-in to building additional refs as needed, but the base CI configuration currently favors breadth of coverage across GPU targets, platforms, and frameworks. Using the full matrix of release versions here would put significant extra load on the CI runner queues.

Can follow what pytorch does for CI workflows. Generally I'd like for CI to build the "most stable" release to reduce the chance of flaky tests unrelated to ROCm changes appearing in CI workflow runs, but we could also just choose release N (as opposed to release N-1, or 'latest' nightly/main)

build_tools/github_actions/configure_jax_release_matrix.py
ScottTodd: Agreed. Make it explicit. Someone should be able to read the top of the file and see what versions are built. They shouldn't need to look down 40 lines in the file and parse through this regex code.

How about something like this?
```diff
 JAX_REF_CONFIGS = {
     "rocm-jaxlib-v0.11.0": {
         "jax_ref": "rocm-jaxlib-v0.11.0",
         "jax_repository": "ROCm/jax",
         "gfx_arch": "device-all",
+        "exclude_python_versions": ["3.11"],  # JAX dropped Python 3.11 in 0.11.0+
    },
```

Or this?
```diff
 JAX_REF_CONFIGS = {
     "rocm-jaxlib-v0.10.1": {
         "jax_ref": "rocm-jaxlib-v0.10.1",
         "jax_repository": "ROCm/jax",
         "gfx_arch": "device-all",
+        "python_versions": ["3.12", "3.13", "3.14"],
    },
```

One thing to watch out for is that we have a `python_version` input to the release workflows used to make dev releases limit what they build, and if framework support is uneven across Python versions that could be confusing for developers using the workflows (well, and also end users installing the frameworks): https://github.com/ROCm/TheRock/blob/4c68147d30f33ed3a7cfa505508d16c41d60894e/.github/workflows/multi_arch_release.yml#L96-L99

build_tools/github_actions/tests/configure_jax_release_matrix_test.py
```diff
@@ -113,6 +96,20 @@ def test_generated_rows_cover_workflow_matrix_inputs(self):
...
+            platform="linux",
+        )
+        combos = {(row["python_version"], row["jax_ref"]) for row in matrix}
+
+        # 3.11 was dropped in JAX 0.11.0, so it must not be paired with 0.11.0.
+        self.assertNotIn(("3.11", "rocm-jaxlib-v0.11.0"), combos)
+        # 3.11 is still valid for the 0.10.x refs.
+        self.assertIn(("3.11", "rocm-jaxlib-v0.10.2"), combos)
+        # Other Python versions are still paired with 0.11.0.
+        self.assertIn(("3.12", "rocm-jaxlib-v0.11.0"), combos)
```
ScottTodd: This test case is fine, but if the logic were more direct and not regex-based then there wouldn't be as much of a need to test it at all. Basically if the code just says "0.11.0 is built for python 3.12+", we don't _also_ need a test that checks "is 0.11.0 built for python 3.12+?". When the code says "based on the regex pattern matching of the ref name, remove 3.11 from the list" then yeah, it's worth testing.

For some ideas, see the tests we have for pytorch matrix code:
* https://github.com/ROCm/TheRock/blob/4c68147d30f33ed3a7cfa505508d16c41d60894e/build_tools/github_actions/tests/configure_pytorch_release_matrix_test.py#L94-L106
* https://github.com/ROCm/TheRock/blob/4c68147d30f33ed3a7cfa505508d16c41d60894e/build_tools/github_actions/tests/configure_pytorch_test_matrix_test.py#L53-L61

.github/workflows/multi_arch_release_linux.yml
gulsumgudukbay: Done, I updated both RELEASES.md (versions + default 0.11.0) and external-builds/jax/README.md (removed the 0.9.1/native tarball flow, refreshed the supported-version table, and replaced the build/test steps with the current build/build.py and ci/run_pytest_rocm.sh commands).

gulsumgudukbay: I removed the tarball dependencies. The manylinux JAX build only consumes build_python_packages.outputs.package_find_links_url, so trigger_release_jax_wheels now only needs: [build_artifacts, build_python_packages]. No more waiting on / being skipped by the tarball job.

build_tools/github_actions/configure_jax_release_matrix.py
gulsumgudukbay: Made it declarative per your suggestionand dropped the regex and added exclude_python_versions: ["3.11"] right in the rocm-jaxlib-v0.11.0 config, so supported versions are readable at the top of the file. The matrix loop just honors that list. This also keeps the python_version release input sane: passing 3.11 with the 0.11.0 ref yields no rows instead of an invalid build.

build_tools/github_actions/configure_jax_release_matrix.py
gulsumgudukbay: Trimmed CI to a single stable ref (rocm-jaxlib-v0.11.0), matching the PyTorch CI approach, with a comment that tells additional refs can be opted in when needed. The full set (0.10.0/0.10.1/0.10.2/0.11.0) is retained only in the release matrix, with 0.11.0 as the default.

build_tools/github_actions/tests/configure_jax_release_matrix_test.py
gulsumgudukbay: since the logic is now declarative rather than regex-based, I kept one small test that the exclude_python_versions config is honored and added test_ci_builds_single_jax_ref to lock the single-ref CI requirement.

erman-gurses:
FYI @gulsumgudukbay 
Release Build tests: https://github.com/ROCm/TheRock/actions/runs/30032481350 (Some tests are failed)
CI Build test: https://github.com/ROCm/TheRock/actions/runs/30031704446/job/89309959352?pr=6708 (Passed)

erman-gurses:
@gulsumgudukbay , please update your PR Title/Description based on CI vs Release Workflow JAX versions.

gulsumgudukbay:
> @gulsumgudukbay , please update your PR Title/Description based on CI vs Release Workflow JAX versions.

Updated

erman-gurses:
Release Test: https://github.com/ROCm/TheRock/actions/runs/30105524675 (one build failed -  retriggered)
CI Test: https://github.com/ROCm/TheRock/actions/runs/30066515970/job/89409795408?pr=6708 (green)
CC: @gulsumgudukbay

build_tools/github_actions/configure_target_run.py
```diff
@@ -41,8 +46,12 @@ def get_runner_label(target: str, platform: str) -> str:
...
             continue
 
-        # If there is a test machine available for this target, run on it.
-        test_runs_on_machine = platform_for_key.get("test-runs-on")
+        # JAX tests need multiple GPUs, so prefer the family's multi-GPU runner
+        # and fall back to the single-GPU runner when there is no dedicated
+        # multi-GPU machine for this family.
+        test_runs_on_machine = platform_for_key.get(
+            "test-runs-on-multi-gpu"
+        ) or platform_for_key.get("test-runs-on")
```
erman-gurses: This looks good to me - @ScottTodd is the code owner for this file so better to get his opinion.

build_tools/github_actions/configure_jax_release_matrix.py
ScottTodd: Oh FYI, we _originally_ built just a single pytorch release ref on CI but expanded to build three pytorch release branches based on build breaks that only appeared on some torch versions - see https://github.com/ROCm/TheRock/pull/5729. The general principle still applies - we'd like to build just a representative list of configurations on CI most of the time, and only expand that matrix where we know that the coverage is worth the cost. So far the JAX builds seem more stable across versions than the pytorch builds due to which parts of ROCm JAX depends on compared to torch.

If the builds were optimized to 10-20 minutes instead of 1h+ then I'd be more comfortable enabling them.

build_tools/github_actions/configure_target_run.py
ScottTodd: Can we separate the changes in this file to another pull request? Switching tests from single GPU to multi-GPU seems unrelated to changing which JAX release branches are built?

erman-gurses: CC: @gulsumgudukbay

gulsumgudukbay: sure, working on a new PR @ScottTodd

gulsumgudukbay: @ScottTodd  the new PR is here: https://github.com/ROCm/TheRock/pull/6853

build_tools/github_actions/configure_jax_release_matrix.py
gulsumgudukbay: @ScottTodd Okay, I am leaving CI on the single stable ref for now, with an opt-in for more refs if we hit version-specific breaks. Will keep the JAX build time in mind as a factor for expanding coverage later.

erman-gurses:
Should we land https://github.com/ROCm/TheRock/pull/6853 first then this current PR https://github.com/ROCm/TheRock/pull/6708?

gulsumgudukbay:
> Should we land #6853 first then this current PR #6708?

I think we should land this first, as it is starting to block ROCm version bump PR s as well since the PRs are seeing 0.9.1 version errors right now and we need nightly builds for the 4 versions for consumption.

erman-gurses:
> > Should we land #6853 first then this current PR #6708?
> 
> I think we should land this first, as it is starting to block ROCm version bump PR s as well since the PRs are seeing 0.9.1 version errors right now and we need nightly builds for the 4 versions for consumption.

Running sanity for release workflow: https://github.com/ROCm/TheRock/actions/runs/30466683657

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -190,30 +170,11 @@ jobs:
 
       - name: Set package dist dir
         run: |
-          if [ "${{ inputs.build_mode }}" = "manylinux" ]; then
-            echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
-          else
-            echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/rocm-jax/jax_rocm_plugin/wheelhouse" >> "$GITHUB_ENV"
-          fi
+          echo "PACKAGE_DIST_DIR=${GITHUB_WORKSPACE}/jax-source/dist" >> "$GITHUB_ENV"
```
ScottTodd: nit: as `PACKAGE_DIST_DIR` is always `${GITHUB_WORKSPACE}/jax-source/dist` now, it can just be a `env: ` variable applied to the whole job or individual steps that need it, without a dedicated step like this that just runs `echo "VARIABLE=VALUE" >> "$GITHUB_ENV"`

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -256,7 +256,9 @@ jobs:
             }
 
   trigger_release_jax_wheels:
-    needs: [build_artifacts, build_tarballs, build_python_packages]
+    # The manylinux JAX build only consumes build_python_packages output
+    # (find-links index); it no longer needs the tarball job.
+    needs: [build_artifacts, build_python_packages]
```
ScottTodd, line 259: nit: This comment saying "this doesn't depend on another job" isn't useful in the source. Leave it to the git history.

build_tools/github_actions/tests/configure_jax_release_matrix_test.py
```diff
@@ -113,6 +96,28 @@ def test_generated_rows_cover_workflow_matrix_inputs(self):
...
+        self.assertIn(("3.11", "rocm-jaxlib-v0.10.2"), combos)
+        # Non-excluded Python versions are still paired with 0.11.0.
+        self.assertIn(("3.12", "rocm-jaxlib-v0.11.0"), combos)
+
+    def test_ci_builds_single_jax_ref(self):
+        matrix = m.generate_jax_matrix_for_release_type(
+            release_type="ci",
+            platform="linux",
+        )
+        self.assertEqual({row["jax_ref"] for row in matrix}, {"rocm-jaxlib-v0.11.0"})
```
ScottTodd, line 122: This test is checking that `rocm-jaxlib-v0.11.0` exists in the matrix, not that the matrix has exactly or only one ref in it.

Do you want to keep this test?

ScottTodd (approved):
Generally lgtm. Please check that the CI and dev release builds succeed. I left a few small comments that would be nice to address too.

erman-gurses (approved):

.github/workflows/multi_arch_build_linux_jax_wheels.yml
gulsumgudukbay: deleted the Set package dist dir step and moved the value to a job-level env: on build_jax_wheels. This matches what multi_arch_build_linux_jax_wheels_ci.yml already does, so the two workflows are now consistent.

.github/workflows/multi_arch_release_linux.yml
gulsumgudukbay: removed the two-line explanation

build_tools/github_actions/tests/configure_jax_release_matrix_test.py
gulsumgudukbay: Keeping it, the assertion is set equality, not membership: assertEqual({row["jax_ref"] for row in matrix}, {"rocm-jaxlib-v0.11.0"}) fails if any second ref appears. The confusing part is that the matrix has one row per Python version while the set collapses to a single entry, so I added a comment saying that explicitly.

erman-gurses:
Release Workflow Test: https://github.com/ROCm/TheRock/actions/runs/30502570523

erman-gurses (approved):
LGTM - Assuming Release and CI Tests are passing

gulsumgudukbay:
The release and PR JAX tests are passing.

PR #6709, opened 2026-07-20
feat(ci): selective build for external repos

geomin12:
Enables external repositories (rocm-systems, rocm-libraries) to trigger selective builds in TheRock, only building and testing the stages/projects that are affected by their changes.

  Key Changes:

  1. Stage Selection (affected_stages) - External repos can specify which build stages to run (e.g.,
  math-libs, compiler-runtime), skipping unneeded stages entirely
  2. Project Selection (projects) - Pass specific projects to build via --projects flag in
  configure_stage.py, enabling only touched components
  3. TEST_SUBPROJECTS Expansion - Automatically includes test dependencies (e.g., if rocprim changes,
  also build/test hipcub which depends on it)
  4. Skip Irrelevant Jobs for External Repos (as not all components are built):
    - Python packages build
    - Native packages (DEB/RPM)
    - PyTorch/JAX wheel builds
    - Python package tests

Works for [rocm-systems](https://github.com/ROCm/rocm-systems/actions/runs/29605298577/job/88406756020?pr=8292) and [rocm-libraries](https://github.com/ROCm/rocm-libraries/actions/runs/29605284244/job/88412267729?pr=9524)

ISSUE ID: #3343 

Next:
- implementing this to superrepos
- collaborating and including "Last Known Good Run" (LKGR) from #6264  to further optimize


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -92,6 +92,33 @@ def _get_all_build_stages() -> list[str]:
...
+    topology = get_topology()
+    affected = topology.get_stages_for_projects(projects)
+
+    if not affected:
+        # No stages found - fall back to full build
+        return "all"
+
+    return ",".join(sorted(affected))
+
+
```
tony-davis: This helper isn't called anywhere (I checked; only the definition exists), and it duplicates the logic that actually ships in `compute_affected_stages.py`. I think the #6660 split orphaned it. Suggest removing it so we don't leave a third "projects to stages" implementation lying around.

```suggestion
```

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -143,12 +147,17 @@ jobs:
       - name: Get stage configuration
         id: stage_config
         run: |
+          PROJECTS_ARG=""
+          if [[ -n "${{ inputs.projects }}" ]]; then
+            PROJECTS_ARG="--projects ${{ inputs.projects }}"
+          fi
```
tony-davis, line 155: `${{ inputs.projects }}` expanded straight into the shell is the Actions injection footgun. Passing it through `env:` keeps it as a plain variable. The unquoted `${PROJECTS}` on the args line still word-splits into multiple `--projects` values, so behavior is unchanged.

```suggestion
        env:
          PROJECTS: ${{ inputs.projects }}
        run: |
          PROJECTS_ARG=""
          if [[ -n "${PROJECTS}" ]]; then
            PROJECTS_ARG="--projects ${PROJECTS}"
          fi
```

.github/workflows/multi_arch_build_windows_artifacts.yml
```diff
@@ -169,11 +173,16 @@ jobs:
       - name: Get stage configuration
         id: stage_config
         run: |
+          PROJECTS_ARG=""
+          if [[ -n "${{ inputs.projects }}" ]]; then
+            PROJECTS_ARG="--projects ${{ inputs.projects }}"
+          fi
```
tony-davis, line 181: Same injection fix here.

```suggestion
        env:
          PROJECTS: ${{ inputs.projects }}
        run: |
          PROJECTS_ARG=""
          if [[ -n "${PROJECTS}" ]]; then
            PROJECTS_ARG="--projects ${PROJECTS}"
          fi
```

.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
```diff
@@ -180,11 +184,16 @@ jobs:
         # where GitHub step outputs are native. The explicit --platform=linux
         # still makes configure_stage.py emit Linux/WSL feature selections.
         run: |
+          PROJECTS_ARG=""
+          if [[ -n "${{ inputs.projects }}" ]]; then
+            PROJECTS_ARG="--projects ${{ inputs.projects }}"
+          fi
```
tony-davis, line 192: And the same one here.

```suggestion
        env:
          PROJECTS: ${{ inputs.projects }}
        run: |
          PROJECTS_ARG=""
          if [[ -n "${PROJECTS}" ]]; then
            PROJECTS_ARG="--projects ${PROJECTS}"
          fi
```

build_tools/github_actions/compute_affected_stages.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Add test_tools to path for TEST_SUBPROJECTS parsing
+REPO_ROOT = SCRIPT_DIR.parents[1]
+sys.path.insert(0, str(REPO_ROOT / "test_tools"))
+from determine_rocm_test_dependencies import get_subprojects_to_test
+
+from github_actions_api import gha_set_output
+
+
+def compute_affected(changed_projects: str) -> tuple[str, str]:
```
tony-davis, line 24: This is the heart of the selective-build logic (normalization + TEST_SUBPROJECTS expansion + stage mapping) and there's no direct unit test for it. A small `compute_affected_stages_test.py` covering the empty input, a `projects/x` normalization case, and one expansion case (e.g. rocprim pulling in its TEST_SUBPROJECTS) would lock it down and match the pattern the bot expects.

tony-davis (changes requested):
Thanks for pushing on selective builds, Geo! This is going to save a ton of CI time for the sub-repos, and the fact that it's a clean no-op for TheRock's own builds (everything defaults to `all` / empty) makes it low-risk to land. I've got a couple of things I'd like to see before merge, plus a few smaller notes.

The blocking one is small: `_get_affected_stages()` in `configure_multi_arch_ci.py` isn't called anywhere. I think it got orphaned by the split with #6660 (that PR is where changed-project detection actually lives now), and the real stage computation is in the new `compute_affected_stages.py`. So it's a third copy of "projects to stages" that nobody runs. I left a suggestion to delete it.

The other thing I care about is the `${{ inputs.projects }}` interpolation going straight into the `run:` blocks in the three artifact workflows. That's the classic Actions script-injection shape, and the value ultimately comes from an external repo's `changed_projects`. Interestingly your own #6660 does this the safe way already (passes SHAs as `gh api` argv, no shell splicing), so this is just bringing these three in line. I left suggestions to move it into `env:`.

Testing is my one real gap concern. The `configure_stage_test.py` additions are good and they'd actually fail on a regression (nice, not padding). But the new entry point that carries the selective-build logic, `compute_affected_stages.py`, has no direct test, including the path-normalization and TEST_SUBPROJECTS expansion. That's the piece most likely to drift silently in the sub-repos' CI. Could we add a small unit test for it? (Worth noting the same missing-test issue has #6660 hard-blocked by the bot right now.)

Non-blocking, but while we're here:

- The stage gating uses `contains(inputs.affected_stages, 'math-libs')`, which is a substring match on the joined string. No collision today, but it's brittle if a future stage name is a substring of another. A safe idiom is `contains(format(',{0},', inputs.affected_stages), ',math-libs,')`. Happy to push suggestions for all of them if you want to tighten it now.
- There are now three different project-name normalizers across #6660's `find_matched_subtrees`, this PR's `compute_affected_stages.py` (`split('/')[-1]`), and `configure_stage.normalize_project_name`. They agree on the normal `category/name` case, so this is fine for now, but a shared helper would be nice before a fourth one shows up.
- Small heads-up: a few Windows jobs are red on the latest run (hipblaslt, miopen, rocthrust, libhipcxx, Validate Artifact Structure). I'm fairly sure they're pre-existing/flaky since this change can't affect internal builds, but it'd be good to confirm they're not new before merge.

tony-davis (approved):
Thanks, Geo; all the blocking items are handled. The dead `_get_affected_stages` is gone, the three artifact workflows moved `projects` into `env:` (no more shell interpolation), and `compute_affected_stages.py` has real test coverage now for normalization, TEST_SUBPROJECTS expansion, and the stage mapping. Green across the bot, unit tests, and pre-commit.

I left the two `contains()`/normalizer notes as optional; they're fine as-is, just worth a shared helper if a third caller shows up. Not blocking. Approving.

build_tools/configure_stage.py
```diff
@@ -2,39 +2,7 @@
...
-    python configure_stage.py --stage math-libs --print
-
-The script generates flags like:
-    -DTHEROCK_AMDGPU_FAMILIES=gfx94X-dcgpu
-    -DTHEROCK_ENABLE_ALL=OFF
-    -DTHEROCK_ENABLE_BLAS=ON
-    -DTHEROCK_ENABLE_FFT=ON
-    ...
-"""
+"""Generate CMake configuration for building a specific stage or projects."""
```
ScottTodd, line 5: These usage comments were useful. Why are you deleting them?

build_tools/configure_stage.py
```diff
@@ -56,17 +24,9 @@ def log(msg: str):
...
 def normalize_project_name(name: str) -> str:
-    """Normalize a project name, handling paths like 'projects/hip' -> 'hip'.
-
-    The changed_projects input from external repos may include paths like:
-    - 'projects/hip' -> 'hip'
-    - 'projects/rocblas' -> 'rocblas'
-    - 'hip' -> 'hip' (already normalized)
-    """
-    # Strip 'projects/' prefix if present
+    """Normalize 'projects/hip' -> 'hip'."""
```
ScottTodd, line 27: Why delete comments?

build_tools/configure_stage.py
```diff
@@ -56,17 +24,9 @@ def log(msg: str):
...
     if name.startswith("projects/"):
         name = name[len("projects/") :]
-    # Strip any trailing path components (e.g., 'hip/src/foo.cpp' -> 'hip')
     if "/" in name:
         name = name.split("/")[0]
```
ScottTodd, line 31: Why delete comments?

.github/workflows/setup_multi_arch.yml
```diff
@@ -161,6 +173,8 @@ jobs:
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
           # Skip path filtering for external repos (git sha won't exist in TheRock)
           SKIP_PATH_FILTERS: ${{ inputs.external_repo != '' && 'true' || '' }}
+          # Disable native packages and PyTorch for external repos (incomplete builds)
+          IS_EXTERNAL_REPO: ${{ inputs.external_repo != '' && 'true' || '' }}
```
ScottTodd, line 177: Why? We need the ability to build native packages and pytorch for changes to rocm-libraries and rocm-systems. That's one of the main reasons to migrate those projects to use "multi-arch CI".

A blunt "is external repo -> disable this job and that job and those other jobs" switch is not appropriate here. We already have filtering support in the CI configuration for individual jobs.

.github/workflows/multi_arch_ci_windows.yml
```diff
@@ -171,7 +182,8 @@ jobs:
   test_python_packages_per_family:
     needs: [build_python_packages]
     name: Test Python
-    if: ${{ !failure() && !cancelled() && toJSON(fromJSON(inputs.build_config).test_python_packages_matrix) != '[]' }}
+    # Skip for external repos (build_python_packages is skipped)
+    if: ${{ !failure() && !cancelled() && inputs.external_repo_config == '' && toJSON(fromJSON(inputs.build_config).test_python_packages_matrix) != '[]' }}
```
ScottTodd, line 186: No, we need to be able to build/test packages for changes to external repositories.

ScottTodd (changes requested):
The design here is too hacky. Not seeing the full vision. Have you written down the migration plan somewhere we can discuss? I see PRs attached to https://github.com/ROCm/TheRock/issues/3343 but https://github.com/ROCm/TheRock/issues/3343 only has a small _edit to the original issue_ saying "skip packaging workflows".

If I can't trust people to review changes to these workflows and scripts adequately I'll add myself to CODEOWNERS and block all changes.

geomin12:
> The design here is too hacky. Not seeing the full vision. Have you written down the migration plan somewhere we can discuss? I see PRs attached to #3343 but #3343 only has a small _edit to the original issue_ saying "skip packaging workflows".
> 
> If I can't trust people to review changes to these workflows and scripts adequately I'll add myself to CODEOWNERS and block all changes.

i need to expand on those! let me do that - it was all in my head and forgot to update the issue

.github/workflows/setup_multi_arch.yml
geomin12: i'll highlight my thought process in the gh issue

geomin12:
> The design here is too hacky. Not seeing the full vision. Have you written down the migration plan somewhere we can discuss? I see PRs attached to #3343 but #3343 only has a small _edit to the original issue_ saying "skip packaging workflows".
> 
> If I can't trust people to review changes to these workflows and scripts adequately I'll add myself to CODEOWNERS and block all changes.

I dropped more context in #3343. let me know, seems we are not aligned :( so i want to make sure we are!

build_tools/github_actions/compute_affected_stages.py
```diff
@@ -0,0 +1,73 @@
...
+    if not affected:
+        print("No stages found for projects, building all stages")
+        return "all", ""
+
+    stages = ",".join(sorted(affected))
+    # Space-separated for --projects arg compatibility
+    projects_str = " ".join(expanded_list)
+    print(f"Affected stages: {stages}")
+    print(f"Expanded projects output: {projects_str}")
+    return stages, projects_str
```
erman-gurses, line 58: Currently one global selection is passed to both Linux and Windows. For example, rocprofiler-sdk or amdsmi selects compiler-runtime, but those artifacts are disabled on Windows. filter_projects_for_stage() still returns the project because it does not consider disable_platforms; the later Windows feature resolution then drops it, leaving THEROCK_ENABLE_ALL=OFF with no selected project feature while the Windows stage job still runs.

We might either compute separate Linux/Windows selections or add platform_name filtering to the topology APIs and skip a platform when no applicable projects remain? A regression test using a Windows-disabled compiler-runtime project such as rocprofiler-sdk would cover this path.

.github/workflows/setup_multi_arch.yml
```diff
@@ -123,6 +133,8 @@ jobs:
       rocm_rpm_package_version: ${{ steps.rocm_package_version.outputs.rocm_rpm_package_version }}
       ref: ${{ steps.checkout.outputs.commit }}
       external_repo_config: ${{ steps.external_repo_config.outputs.config_json }}
+      affected_stages: ${{ steps.affected_stages.outputs.affected_stages }}
+      expanded_projects: ${{ steps.affected_stages.outputs.expanded_projects }}
```
erman-gurses, line 137: Could we preserve the advertised "all" default in the setup job output?

When external_repo == '', the affected_stages step is skipped, so` steps.affected_stages.outputs.affected_stages` is empty rather than "all". Any caller that explicitly forwards this output will bypass the called workflow’s input default and can cause every stage condition to evaluate false.

build_tools/github_actions/compute_affected_stages.py
```diff
@@ -0,0 +1,73 @@
...
+        print("No changed_projects specified, building all stages")
+        return "all", ""
+
+    raw_projects = [p.strip() for p in changed_projects.split(",") if p.strip()]
+    if not raw_projects:
+        print("Empty projects list after parsing, building all stages")
+        return "all", ""
+
+    # Normalize "projects/hip" -> "hip"
+    projects = [p.split("/")[-1] if "/" in p else p for p in raw_projects]
```
erman-gurses, line 36: We can reuse the existing `normalize_project_name()` here instead of taking the final path component.

Please see the link: https://github.com/ROCm/TheRock/blob/main/build_tools/configure_stage.py#L58-L72

For example, projects/hip correctly becomes hip, but a changed path such as `projects/hip/src/foo.cpp` would become foo.cpp. That project would not map to a stage, causing this selective-build path to fall back to "all".

Can we also add a regression test for this nested-path case?

erman-gurses (commented):
Please see my comments below:

geomin12:
closing to do some rework on this.

PR #6717, opened 2026-07-21
fix(packaging): register ROCm libraries for Linux native packages

### LINKED ISSUE #6716: Linux native packages do not register /opt/rocm/lib with the dynamic loader
## Problem Description

TheRock Linux native packages create the intended compatibility path:

```text
/opt/rocm/lib -> /etc/alternatives/rocm-lib -> /opt/rocm/core-X.Y/lib
```

but do not install an entry under `/etc/ld.so.conf.d` for that path. As a result, an external native program can link against ROCm successfully and then fail at startup because its `DT_NEEDED` libraries are absent from the loader cache.

This reproduces in the official `rocm/dev-ubuntu-24.04:7.14.0-full` image (digest `sha256:439edaa8f0c4be4a3728e528f87b8a2ea1f051f34cf10b27caa4bd94f562eda7`) without GPU access:

```bash
cat >/tmp/probe.cpp <<'EOF'
#include <hip/hip_runtime_api.h>
int main() {
    int version = 0;
    return hipRuntimeGetVersion(&version) == hipSuccess ? 0 : 1;
}
EOF

hipcc /tmp/probe.cpp -o /tmp/probe
env -u LD_LIBRARY_PATH /tmp/probe
```

Result:

```text
error while loading shared libraries: libamdhip64.so.7: cannot open shared object file
exit code: 127
```

`ldd` also reports `libamdhip64.so.7 => not found`. Bare lookup similarly fails for `libhipblas.so.3` and `librocblas.so.5`.

Adding `/opt/rocm/lib` to `/etc/ld.so.conf.d` and running `ldconfig` changes the same binary to exit 0. It also fixes the user-reported case in ggml-org/llama.cpp#25807: an unmodified llama.cpp build that otherwise exits 127 on `libhipblas.so.3`.

The same failure was reproduced with the stable ROCm 7.11 package path from the commit that introduced package-based runtime images, so this is not specific to llama.cpp or newly introduced by 7.14.

## Expected Behavior

Programs linked against a system-level ROCm DEB/RPM installation should resolve the active ROCm libraries without requiring `LD_LIBRARY_PATH` or application-specific absolute RUNPATHs.

The stable loader entry should follow the active `update-alternatives` selection and refresh correctly across install, upgrade, version switching, removal, and purge/erase.

## Scope

This report is limited to Linux native DEB/RPM packages. It does not propose global loader configuration for Python virtual environments, tarballs, arbitrary prefixes, or Windows DLL packages.

## Environment

- Operating system: Ubuntu 24.04 container
- ROCm version: 7.14
- Installation method: TheRock Linux native packages
- GPU: not required to reproduce

liminfei-amd:
## Motivation

Linux native packages create `/opt/rocm/lib` as an `update-alternatives` link to the active versioned library directory, but the path is not registered with the system dynamic loader. External native consumers therefore link successfully and then exit 127 on ROCm SONAMEs such as `libamdhip64.so.7` and `libhipblas.so.3` unless users set `LD_LIBRARY_PATH` or add application-specific RUNPATHs.

The user-reported downstream case is ggml-org/llama.cpp#25807. The same defect reproduces independently with a minimal no-GPU HIP consumer.

Fixes #6716.

This change is intentionally scoped to Linux DEB/RPM packages. It does not alter Python wheel initialization, tarball activation, arbitrary-prefix installs, or Windows DLL discovery.

## Technical Details

- Add package metadata for dynamic-loader paths and make the non-gfx `amdrocm-runtime` package own a versioned loader drop-in containing `/opt/rocm/lib`.
- Install the file as a DEB conffile and RPM `%config(noreplace)` file.
- Refresh `ldconfig` after runtime configuration changes and after `amdrocm-core` changes the existing alternatives.
- Keep RPM upgrades safe by preventing the old package's `%preun` from removing alternatives installed by the new package.
- Extend native package sanity testing with a no-GPU HIP consumer that runs with `LD_LIBRARY_PATH` removed.

The existing versioned layout and active-version contract remain unchanged:

```text
/opt/rocm/lib -> /etc/alternatives/rocm-lib -> /opt/rocm/core-X.Y/lib
```

## Test Plan

- Generate and build real DEB and RPM packages.
- Install, upgrade, remove, and purge/erase both package formats.
- Switch between two versioned alternatives and verify loader-cache selection.
- Run a no-GPU HIP consumer and an unmodified build reproducing ggml-org/llama.cpp#25807 before and after installing the generated loader payload.

## Test Result

- Package-generation tests: 25/25 passed.
- Loader-focused tests: 7/7 passed.
- Real DEB and RPM build/install/upgrade/remove lifecycle tests passed.
- Multi-version alternatives switching passed.
- Official ROCm 7.14 image, same binaries and environment:

| Phase | Minimal HIP consumer | llama.cpp #25807 reproducer |
|---|---:|---:|
| Before package | 127 (`libamdhip64.so.7`) | 127 (`libhipblas.so.3`) |
| After package install | 0 | 0 |
| After DEB remove | 0 (conffile retained) | 0 |
| After DEB purge | 127 | 127 |

The no-device message emitted by llama.cpp occurs after successful dynamic loading and does not affect this loader result.

## Submission Checklist

- [x] Targeted `main`.
- [x] Added regression coverage.
- [x] Validated DEB and RPM behavior, including upgrades and multi-version selection.
- [x] Linked the tracking issue.

### DIFF
--- a/build_tools/packaging/linux/deb_package.py
+++ b/build_tools/packaging/linux/deb_package.py
@@ -129,8 +129,16 @@ def create_versioned_deb_package(pkg_name, config: PackageConfig):
         if build_config.enable_rpath:
             convert_runpath_to_rpath(package_dir)
 
+        has_ldconfig_file = generate_ldconfig_file(pkg_info, package_dir, build_config)
+
         # Generate install file after copying, so we can check for hidden files
-        generate_install_file(pkg_info, deb_dir, build_config, dest_dir)
+        generate_install_file(
+            pkg_info,
+            deb_dir,
+            build_config,
+            dest_dir,
+            has_ldconfig_file=has_ldconfig_file,
+        )
 
     package_with_dpkg_build(package_dir)
 
@@ -191,7 +199,14 @@ def generate_changelog_file(pkg_info, deb_dir, config: PackageConfig):
         f.write(template.render(context))
 
 
-def generate_install_file(pkg_info, deb_dir, config: PackageConfig, dest_dir=None):
+def generate_install_file(
+    pkg_info,
+    deb_dir,
+    config: PackageConfig,
+    dest_dir=None,
+    *,
+    has_ldconfig_file=False,
+):
     """Generate a Debian install entry in `debian/install`.
 
     Parameters:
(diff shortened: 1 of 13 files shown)


liminfei-amd:
Superseded by #6718. GitHub automatically closed this PR when the source branch was renamed to satisfy the repository branch-name policy; the commit is unchanged.

PR #5529, opened 2026-05-29
[tensilelite] Re-apply test component config and fix PyTorch SDK test failure

archana-ramalingam:
## Summary

Re-apply tensilelite test component configuration (reverted in #5520) and fix the `testSharedLibrariesLoad` failure on Python 3.10/3.11.

## Technical details

- Re-applies #4991: tensilelite test component for artifact-based CI testing.
- Fixes #5516: 
   - `_rocisa.abi3.so` uses `PyType_FromMetaclass` (Python 3.12+ symbol). 
   - The `testSharedLibrariesLoad` test loads all `.so` files via `ctypes.CDLL`, which fails for Python C extensions under older interpreters. 
   - Added a skip for `.abi3.so` and `.cpython-*` extensions in `libraries_test.py` — these must be loaded via `importlib`, not `ctypes.CDLL`. Same pattern as the existing `amd_smi`/`goamdsmi` skip.
- `-DROCISA_USE_STABLE_ABI=ON` is preserved — the stable ABI build is correct; the issue was only in how the PyTorch SDK test loads the `.so`.

## Test plan

- [x] `testSharedLibrariesLoad` passes on Python 3.10/3.11 (no longer attempts to ctypes-load `_rocisa.abi3.so`)
   - gfx950: https://github.com/ROCm/TheRock/actions/runs/26786234522
   - gfx1151: https://github.com/ROCm/TheRock/actions/runs/26823115936
- [x] tensilelite tests pass via `test_tensilelite.py`:
   - rocisa: 48 passed
   -  tensilelite: 709 passed, 200 skipped, 1 xfailed

Fixes #5516

AIGECORE-1997

Issue #3674, opened 2026-02-27
[ci] Investigate using a cache for artifacts on test machines


ScottTodd:
## Background

Self-hosted GPU test runners have long queue times. While runner capacity expansion (on-prem + cloud), test filtering (smoketests vs full), and job filtering (change-based test selection) are being addressed separately, we can also reduce pressure by optimizing the test jobs themselves. Some jobs show 1-3 minutes of overhead from installing software, downloading artifacts, and unpacking artifacts.

For example, https://github.com/ROCm/TheRock/actions/runs/22457240043/job/65080326403 for hipsolver tests spent 18 seconds running (smoke) tests out of 1m37s total:
step | time
-- | --
fetch `build_tools` | 9s
checkout | 12s
setup test env | 39s
health checks | 4s
tests | 18s
post checkout | 7s

About 30s is spent there downloading and flattening artifacts. Another run for miopen https://github.com/ROCm/TheRock/actions/runs/22457240043/job/65080326481 spent almost 1 minute downloading and flattening artifacts.

## Ideas

Some design notes: https://github.com/ScottTodd/claude-rocm-workspace/blob/main/tasks/active/test-workflow-overhead.md

I think we should do a quantitative analysis of overhead like this and then considering implementing a cache for downloading, unzipped, and flattened artifacts.

Brainstorming potential building blocks / solutions:

* Set a `THEROCK_RUNNER_CACHE_ROOT` env var on test machines that jobs can write to and read from
* Use the cache for _downloads_ (`.tar.xz`)
* Use the cache for _extracted downloads_ (folders, one per artifact hash, possible to reuse across workflow runs)
* Use the cache for _flattened directories_ (merged folders from multiple artifacts, possible to reuse for multiple jobs within a workflow run)

Could build something into these scripts:
* [`build_tools/install_rocm_from_artifacts.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/install_rocm_from_artifacts.py) (used by test workflows)
* [`build_tools/fetch_artifacts.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/fetch_artifacts.py) (used by `build_tools/install_rocm_from_artifacts.py`)
* [`build_tools/artifact_manager.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/artifact_manager.py) (used by build workflows, not as relevant here today)

PR #6730, opened 2026-07-21
Drop hip-python from test requirements

marbre:
## Motivation

The `requirements-test.txt` pulls in `hip-python==7.2.2.562.43` which is not desired.

## Technical Details

PR #5529 added `hip-python` as a test dependencies, but following https://github.com/ROCm/TheRock/pull/5529#discussion_r3599390876 it can be dropped.

## Submission Checklist

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -20,7 +20,6 @@ joblib>=1.4.0
 numpy>=1.26.0
 filelock>=3.0.0
 syrupy==5.3.2
-hip-python; sys_platform != "win32"
 pytest-xdist>=1.32.0
 
 # libhipcxx test requirement


requirements-test.txt
```diff
@@ -20,7 +20,6 @@ joblib>=1.4.0
 numpy>=1.26.0
 filelock>=3.0.0
 syrupy==5.3.2
-hip-python; sys_platform != "win32"
```
ScottTodd, line 23: > PR #5529 added `hip-python` as a test dependencies, but following [#5529 (comment)](https://github.com/ROCm/TheRock/pull/5529#discussion_r3599390876) it can be dropped.

I'm not sure about that comment, see https://github.com/ROCm/TheRock/pull/5529#discussion_r3624190394

archana-ramalingam: Yes, @marbre since the submodule bumps (rocm-libraries) started running the gfx950 tests recently, removing `hip-python` will silently skip 12 GPU tests.
Example tests that will be skipped but required on gfx950, passing now: https://github.com/ROCm/TheRock/actions/runs/29815512767/job/88684173821#step:14:4809
Do you have guidance on the right mechanism to add `hip-python` in TheRock workflow?

marbre: No, I have not. There isn't a `hip-python` release for the current ROCm 7.14 production release. Maybe @domcharrier or @psamfass can give guidance.

domcharrier: @marbre  @archana-ramalingam 

What APIs are called? If this subset is stable between ROCm 7.2.2 and ROCm LATEST then it is very likely safe to use `hip-python~=7.2.2`. All libraries and function symbols are loaded lazily. There would only be an issue if you use newly added HIP APIs or a signature had changed (type/order/number of args/return values).

archana-ramalingam: The TensileLite gfx950 GPU subtile tests use hip-python for HIP runtime calls to assemble and execute generated subtile kernels — `hipModuleLoad`, `hipModuleLaunchKernel`, `hipMalloc`, `hipMemcpy`, `hipDeviceSynchronize`, etc. 

`hip-python 7.2.2` uses `dlopen("libamdhip64.so")` at runtime — so it calls into whichever ROCm is installed (10.x in TheRock). It works today because the APIs we use haven't changed, but newer ROCm versions could add/change struct fields that the 7.2 wrapper might not know about.

marbre: Thanks for elaborating @domcharrier and @archana-ramalingam. My assumption was therefore wrong and I will close this PR. Sorry for the noise.

ScottTodd: _I think we should move this requirement to be local to only the tests that need it._

The unconditional download of 64MB hip-python packages is actually contributing pretty significantly to test time on some gfx942 runners with slow network, and pip/uv caching isn't enabled (? see also https://github.com/ROCm/TheRock/issues/3674).

For example: https://github.com/ROCm/TheRock/actions/runs/32165005832/job/96055002747?pr=7449#step:8:160

```diff
 Tue, 18 Aug 2026 18:57:10 GMT   Resolved 49 packages in 45.50s
 Tue, 18 Aug 2026 18:57:10 GMT      Building perfetto==0.7.0
 Tue, 18 Aug 2026 18:57:10 GMT         Built perfetto==0.7.0
 Tue, 18 Aug 2026 18:57:11 GMT   Downloading pygments (1.2MiB)
 Tue, 18 Aug 2026 18:57:11 GMT   Downloading setuptools (1.0MiB)
 Tue, 18 Aug 2026 18:57:11 GMT   Downloading numpy (15.9MiB)
 Tue, 18 Aug 2026 18:57:11 GMT   Downloading sccache (14.0MiB)
 Tue, 18 Aug 2026 18:57:11 GMT   Downloading botocore (13.3MiB)
+Tue, 18 Aug 2026 18:57:11 GMT   Downloading hip-python (63.8MiB)
 Tue, 18 Aug 2026 18:57:53 GMT    Downloaded setuptools
 Tue, 18 Aug 2026 18:57:56 GMT    Downloaded pygments
 Tue, 18 Aug 2026 19:00:28 GMT    Downloaded botocore
 Tue, 18 Aug 2026 19:00:35 GMT    Downloaded sccache
 Tue, 18 Aug 2026 19:00:51 GMT    Downloaded numpy
+Tue, 18 Aug 2026 19:06:07 GMT    Downloaded hip-python
+Tue, 18 Aug 2026 19:06:07 GMT   Prepared 49 packages in 8m 56s
```

Aggregate statistics from that run show all 51 test jobs downloaded the 64MB wheel but **none of them used it** as TensileLite is the only user with relevant test imports and the "quick" category ran with `not gpu and not skip-gfx94X`. Timing was pretty variable due to differing network speeds:

Metric                 |    Duration
-- | --
Aggregate runner time  |   50m 39.8s
Mean                   |       59.6s
Median                 |       16.5s
p90                    |    2m 12.9s
p95                    |    7m 55.6s
Maximum                |    8m 55.8s
Under 30s              |  44/51 jobs
Over 2 minutes         |   6/51 jobs

---

Out of curiosity, why is hip-python so large, at 64MB compressed? Looking at https://pypi.org/project/hip-python/7.2.2.562.43/#files, I see:

file name | uncompressed size
-- | --
`hipblas.cpython-314-x86_64-linux-gnu.so ` | 101MB
`hip.cpython-314-x86_64-linux-gnu.so ` | 81MB
`hipsparse.cpython-314-x86_64-linux-gnu.so ` | 47MB
`hipsolver.cpython-314-x86_64-linux-gnu.so ` | 41MB
`rccl.cpython-314-x86_64-linux-gnu.so ` | 6MB
`chipblas.cpython-314-x86_64-linux-gnu.so ` | 5MB
`hiprand.cpython-314-x86_64-linux-gnu.so` | 3MB

ScottTodd: > **none of them used it** as TensileLite is the only user with relevant test imports and the "quick" category ran with `not gpu and not skip-gfx94X`.

correction: it might be used at "quick", but only for tensilelite and no other components

ScottTodd: While we're at it, we can move at least sccache and numpy to component-specific test requirements too: https://github.com/ROCm/TheRock/blob/04a247437155a66af4efc9e7b05ccbd5eec57263/requirements-test.txt#L16-L29

botocore from boto3 is what we use to download artifacts, so I don't think we can easily slice that out

ScottTodd: For " # TODO: Keep in sync with rocm-libraries/projects/hipblaslt/tensilelite/requirements.txt ", what if we copy that requirements.txt into the test artifact for tensilelite?

ScottTodd: Filed https://github.com/ROCm/TheRock/issues/7491 to follow-up.

PR #6733, opened 2026-07-21
ci(rocprofiler-systems): Support --repeat until-pass:N CTest flag

### LINKED ISSUE #6729: [Issue] test_runner.py has no option to rerun CTests on failure
As the title states, `test_runner.py` is missing support for `--repeat until-pass:N` flag. This was used by `rocprofiler-systems` in `test_rocprofiler_systems.py` as some of our tests can fail once and then pass in subsequent runs depending on the load on the CI machine.

Furthermore, with parallelism, dyninst could fail an invariant if two basic blocks were discovered at the same time. Allowing tests to run multiple times to account for these small factors would resolve our problems.

kcossett-amd:
## Motivation

Fixes #6729

In the previous test implementation (where each component had a per component test file), we used `--repeat until-pass:3` as there was a possibility a test could fail for various reasons (current load on test machine, OpenMP discovering two basic blocks at the same time). Re-running the test would likely make it pass.

This behavior would match all our other CI workflows.

## Technical Details

 - Add support for `--repeat until-pass:N`, and then have `rocprofiler-systems` use `--repeat until-pass:3`

## Test Plan

Local. 
**Cannot be tested on CI**, as that would require an already failing `rocprofiler-systems` test.

## Test Result

Locally, when I force our component to only run a failing test (that only fails on my computer unless I run with `taskset 0-15`), I see the expected output:
```
Test project /home/rocm/therock-branches/therock-new-main/output-linux-portable/build/dist/rocm/share/rocprofiler-systems/tests
    Start   1: rocprofiler-systems-pytest-config
1/3 Test   #1: rocprofiler-systems-pytest-config ...   Passed    0.43 sec
    Start 217: mpi-binary-rewrite
2/3 Test #217: mpi-binary-rewrite ..................***Failed    2.50 sec
    Start 217: mpi-binary-rewrite
    Test #217: mpi-binary-rewrite ..................***Failed    3.17 sec
    Start 217: mpi-binary-rewrite
    Test #217: mpi-binary-rewrite ..................***Failed    2.06 sec
    Start 618: rocprofiler-systems-test-cleanup
3/3 Test #618: rocprofiler-systems-test-cleanup ....   Passed    0.17 sec

67% tests passed, 1 tests failed out of 3

Label Time Summary:
all               =   2.06 sec*proc (1 test)
binary_rewrite    =   2.06 sec*proc (1 test)
cleanup           =   0.17 sec*proc (1 test)
full              =   2.06 sec*proc (1 test)
global            =   0.60 sec*proc (2 tests)
mpi               =   2.06 sec*proc (1 test)
prerequisite      =   0.43 sec*proc (1 test)

Total Test time (real) =   8.36 sec

The following tests FAILED:
        217 - mpi-binary-rewrite (Failed)
```

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_runner.py
+++ b/build_tools/github_actions/test_executable_scripts/test_runner.py
@@ -171,6 +171,10 @@ environ_vars["ROCM_PATH"] = str(ROCM_PATH)
 #   instrumented-function listings). "--output-on-failure" is always passed, so
 #   failing tests still show their full output; only passing-test noise is
 #   suppressed.
+#
+# - ctest_repeat_until_pass: Int (default unset/off). When set to N, adds
+#   "--repeat until-pass:N" so ctest re-runs a failing test up to N times and
+#   only reports failure if it never passes.
 COMPONENT_OVERRIDES = {
     # ctest fragments live under libexec, not bin.
     # ctest_parallel pinned to 1: tests are pytest runs that parallelize
@@ -219,6 +223,9 @@ COMPONENT_OVERRIDES = {
         # function listings, etc). Drop -V to keep CI logs readable;
         # --output-on-failure still surfaces output for any failing test.
         "ctest_verbose": False,
+        # Depending on the load of the system when a test is ran, it could fail
+        # once and then pass on subsequent runs.
+        "ctest_repeat_until_pass": 3,
     },
     # rocwmma installs three independent CTestTestfile.cmake fragments:
     #   bin/rocwmma/             - per-target plain runs + regression_tests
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/test_executable_scripts/test_runner.py
```diff
@@ -528,6 +535,15 @@ def build_ctest_command(
     # Add common ctest parameters
     cmd.append("--output-on-failure")
 
+    # Retry flaky tests: re-run any failing test up to N times and only count
+    # it as a failure if it never passes. Off by default; opt in per component
+    # via COMPONENT_OVERRIDES[...]["ctest_repeat_until_pass"] = N.
+    repeat_until_pass = COMPONENT_OVERRIDES.get(test_component_job_name, {}).get(
```
geomin12, line 541: this is a great addition and will fix flaky tests, but we need to root cause and figure out why tests are flaky in the first place!

maybe, can we add this flag to specific components that request it?

geomin12 (commented):
just one minor discussion topic

build_tools/github_actions/test_executable_scripts/test_runner.py
kcossett-amd: > maybe, can we add this flag to specific components that request it?

The only component that requests it, as of right now, is `rocprofiler-systems`

> this is a great addition and will fix flaky tests, but we need to root cause and figure out why tests are flaky in the first place!

You are correct that this may hide actual flaky tests, but I'd argue that at least `--repeat until-pass:2` would be ideal, because the test could be flaky for reasons out of our control/unlucky events.

For example: 
 - any `binary-rewrite`/`runtime-instrument` could fail if the OpenMP parallelized region (in dyninst) discovers the same basic blocks at the same time. In practice, this is exceedingly rare, and I've only seen this happen once in TheRock CI.
 - a test could timeout due to the system being under load in that particular moment (this I have observed a few times). 
 - Whenever MPI is supported, also have a small chance that it fails once before passing.
 
 If a test were to fail twice in a row, then that would mean there is a serious problem. 
 Although, in the end, I suppose you could argue both for and against doing this.
 
 Edit: I can switch it to `2`.

PR #6735, opened 2026-07-21
build(pytorch): enable gfx1250 wheels for release/2.12 and nightly

ethanwee1:
JIRA ID : AIPYTORCH-804

## Summary

Enable `gfx125X-dcgpu` (gfx1250) in TheRock's PyTorch release matrix for **`release/2.12`** and **`nightly`**, matching the existing `release/2.11` behavior.

Previously, `configure_pytorch_release_matrix.py` filtered gfx1250 out of those refs even though PyTorch now supports the arch:

- `release/2.12`: ROCm/pytorch#3421
- `nightly`: upstream pytorch/pytorch#188597 (via TheRock's upstream nightly checkout)

## Changes

- Remove `gfx125X-dcgpu` from the unsupported-family filter for `release/2.12` and `nightly` in `configure_pytorch_release_matrix.py`.

## Validation

Both the single-arch and multi-arch gfx1250 PyTorch wheel builds are green with this matrix change (Python 3.12, ROCm `7.15.0a20260712`, index `rocm.nightlies.amd.com/whl-multi-arch`, `test_amdgpu_families=none`):

| Scenario | `amdgpu_families` | Run | Result |
|----------|-------------------|-----|--------|
| Single-arch (gfx1250 only) | `gfx125X-dcgpu` | https://github.com/ROCm/TheRock/actions/runs/29936052631 | ✅ success |
| Multi-arch (gfx942 + gfx1250) | `gfx94X-dcgpu;gfx125X-dcgpu` | https://github.com/ROCm/TheRock/actions/runs/29938470834 | ✅ success |

Notes:

- The **multi-arch** build (TheRock's production release/nightly path) works on stock `release/2.12` and `main` because `ck_gemm` is built as a separate library with gfx1250 filtered out of `HIP_ARCHITECTURES`; gfx1250 falls back to hipBLASLt at runtime, and CK GEMM stays enabled for gfx942/gfx950.
- The **single-arch** gfx1250-only case additionally requires the CK-GEMM CMake fix in ROCm/pytorch#3458 (`release/2.12`) and pytorch/pytorch#190683 (`main`/nightly). These are not required for the multi-arch production builds.


jithunnair-amd (approved):
Not sure if the unit tests are needed; will leave that to theRock team's judgement

.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ScottTodd, line 1: The changes to this file could be moved to a separate PR. I'd also prefer to keep the Linux and Windows workflows in sync and not add inputs to only one or the other: https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml

build_tools/github_actions/tests/configure_pytorch_release_matrix_test.py
```diff
@@ -105,6 +105,46 @@ def test_filters_exact_unsupported_family(self):
         matrix_families = ";".join(row["amdgpu_families"] for row in matrix)
         self.assertNotIn("gfx125X", matrix_families)
```
ScottTodd, line 106: I think the existing tests were sufficient here. We don't need 40 lines of test code for a 2 line change just checking that config file has some values.

ScottTodd (approved):
I'd prefer if this PR was limited to only the changes to `build_tools/github_actions/configure_pytorch_release_matrix.py`.

* One PR updating `.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml` _and_ `.github/workflows/multi_arch_build_windows_pytorch_wheels.yml`
* No PR for the extra test (it's not a useful test case when existing test cases already exercise the logic there)

ethanwee1:
@ScottTodd addressed, thanks for the review:

- Scoped this PR down to only `configure_pytorch_release_matrix.py`.
- Reverted the added `configure_pytorch_release_matrix_test.py` cases (existing tests already cover the filtering).
- Moved the `pytorch_gitrepo_origin` workflow input to a separate PR (#6782), applied to both the Linux and Windows workflows to keep them in sync.

ScottTodd (approved):

PR #6737, opened 2026-07-21
[rocm-sdk] Preload ASan runtime in wheel tests  (#6331)

### LINKED ISSUE #6331: [Issue] Linux ASAN CI gfx94X-dcgpu: rocm-sdk test fails — test_initialize_process_env_preload_1 aborts with `ASan runtime does not come first in library list`
## Context

- **Workflow:** Multi-Arch CI ASAN (`.github/workflows/multi_arch_ci_asan.yml`)
- **Run:** [Run 28635426241 (2026-07-03)](https://github.com/ROCm/TheRock/actions/runs/28635426241)
- **Affected arch:** `gfx94X-dcgpu` (gfx942 / MI300X)
- **Test:** `rocm_sdk.tests.base_test.ROCmBaseTest.test_initialize_process_env_preload_1`

## Summary:

`rocm-sdk test` exits with code 1 on every ASAN nightly run. The failing test sets `ROCM_SDK_PRELOAD_LIBRARIES=amdhip64` and calls `rocm_sdk.initialize_process()`, which invokes `ctypes.CDLL(libamdhip64.so, RTLD_GLOBAL)`. Because `libamdhip64.so` was compiled with AddressSanitizer, the ASan runtime demands it be the first `.so` loaded into the process. Since the Python interpreter itself was not started under `LD_PRELOAD=libclang_rt.asan.so`, ASan aborts immediately:

## Error Snippet:
```text
==217==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
##[error]Process completed with exit code 1.
```

The same failure occurs twice per run — once for the `rocm[libraries]` install step and once for `rocm[libraries,devel]`. All Python versions (3.10, 3.11, 3.12) and both container images (ubi10, ubuntu24.04) are affected.

## Failing Jobs:
- https://github.com/ROCm/TheRock/actions/runs/28635426241/job/84978878931
- https://github.com/ROCm/TheRock/actions/runs/28635426241/job/84978878971
- https://github.com/ROCm/TheRock/actions/runs/28635426241/job/84978878976
- https://github.com/ROCm/TheRock/actions/runs/28635426241/job/84978878950

## Root Cause

`test_initialize_process_env_preload_1` (and the other `preload_libraries` tests) unconditionally load an ASAN-instrumented ROCm shared library via `ctypes.CDLL`. ASan's runtime requires being the first library loaded — a condition only satisfiable via `LD_PRELOAD` at process startup, which is not set when `rocm-sdk test` is invoked. The test lacks an ASAN-guard equivalent to the `@pytest.mark.skipif(is_asan(), ...)` pattern used in other test scripts.

bgopesh:
## Summary

 Fixes #6331. On the *Multi-Arch CI ASAN* nightly (`gfx94X-dcgpu`), `rocm-sdk test` exits with code 1. The failing test `test_initialize_process_env_preload_1` sets                    
  `ROCM_SDK_PRELOAD_LIBRARIES=amdhip64` and calls `rocm_sdk.initialize_process()`, which `ctypes.CDLL`s an **ASan-instrumented** `libamdhip64.so` into an interpreter that was **not**   
  started with `LD_PRELOAD=libclang_rt.asan…so`. ASan requires its runtime to be first in the process' *initial* library list, so it aborts the whole process (`ASan runtime does not    
  come first in initial library list`) — an uncatchable `SIGABRT` that kills the entire run at the first in-process load.    

## Changes

- `.github/workflows/test_rocm_wheels.yml`: new `build_variant` input (both `workflow_dispatch` and `workflow_call`, default `release`). Before each `rocm-sdk test` step              
  (`rocm[libraries]` and `rocm[libraries,devel]`), when the variant matches `*asan*`, `LD_PRELOAD` the ASan runtime located via `amdclang++ -print-file-name=libclang_rt.asan-$(uname    
  -m).so` (placed first, preserving any existing value) and set `ASAN_OPTIONS=detect_leaks=0`. No-op on `release`.                                                                       
  - `.github/workflows/multi_arch_ci_linux.yml`: thread `build_variant` into the `test_rocm_wheels.yml` call from `fromJSON(inputs.build_config).build_variant_label`, so the preload    
  actually triggers on ASan builds. (`BUILD_VARIANT` was not otherwise available at wheel-test time.)  

## Test plan

  - [ ] Verify green on the Multi-Arch CI ASAN workflow (`gfx94X-dcgpu`) once ASan CI can be opted-in for this PR.                                                                       
  - [ ] Confirm non-ASAN wheels are unaffected (preload block is a no-op on `release`).  


build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
```diff
@@ -82,6 +82,10 @@ def testInstallationLayout(self):
             msg="Paths are not siblings",
         )
 
+    @unittest.skipIf(
+        utils.should_skip_asan_inprocess_load(),
+        "ASan-instrumented libraries require LD_PRELOAD at startup; see TheRock#6331",
+    )
     def testSharedLibrariesLoad(self):
```
ScottTodd: I think test failures are appropriate here and we shouldn't just skip tests, though we could improve the error message when using ASan and necessary libraries are not also preloaded. We _especially_ shouldn't just skip tests on our CI, as the whole point of CI is to run our own tests and catch issues. As needed we can have our test workflows preload libraries and do other test environment setup, matching the usage instructions that we expect users to follow.

Maybe this code could do the preloading itself though? It's already loading _some_ libraries, could it also load the asan libraries?

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
```diff
@@ -82,6 +82,10 @@ def testInstallationLayout(self):
             msg="Paths are not siblings",
         )
 
+    @unittest.skipIf(
+        utils.should_skip_asan_inprocess_load(),
+        "ASan-instrumented libraries require LD_PRELOAD at startup; see TheRock#6331",
```
ScottTodd: nit: prefer full URLs in logs and comments so humans (and agents etc.) can follow them directly, without needing to know what this expands to

`https://github.com/ROCm/TheRock/issues/6331`

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
bgopesh: yes makes sense. Let me try to reload asan lib

bgopesh:
@ScottTodd Hi Scott, Do you know how I can enable Linux ASAN jobs? I am not familiar with TheRock CI. Can you run `multi_arch_ci_asan.yml` via `workflow_dispatch` against my PR? Seems like my PR didnt trigger any asan jobs due to no code change.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
bgopesh: sure. will resolve it before final commit.

bgopesh:
This is a 2 part solution:
                                                                                                                                                                                                                                                                                                                                             
  - Part 1 (workflow LD_PRELOAD) is what makes the ASAN CI green — the tests run under a correctly-preloaded interpreter.                                                                
  - Part 2 (pre-check) is a safety net / UX improvement — if someone forgets to preload, they get a clear, catchable RuntimeError with the fix, instead of an uncatchable SIGABRT. 
  
  We can do part 2 in a separate PR as its an improvement. Let me know and I will remove it.

ScottTodd:
> @ScottTodd Hi Scott, Do you know how I can enable Linux ASAN jobs? I am not familiar with TheRock CI. Can you run `multi_arch_ci_asan.yml` via `workflow_dispatch` against my PR? Seems like my PR didnt trigger any asan jobs due to no code change.

I don't think we have support for this yet (other than creating branches in the shared repository and using workflow_dispatch). We should add a `ci:run-asan` label or something so PRs that don't touch submodules can easily opt-in (cc @geomin12).

We could rename the `build_variant:asan` label which was used by a prior iteration of CI workflows?
image
Docs for how to opt-in to jobs like that will go in https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

bgopesh:
> > @ScottTodd Hi Scott, Do you know how I can enable Linux ASAN jobs? I am not familiar with TheRock CI. Can you run `multi_arch_ci_asan.yml` via `workflow_dispatch` against my PR? Seems like my PR didnt trigger any asan jobs due to no code change.
> 
> I don't think we have support for this yet (other than creating branches in the shared repository and using workflow_dispatch). We should add a `ci:run-asan` label or something so PRs that don't touch submodules can easily opt-in (cc @geomin12).
> 
> We could rename the `build_variant:asan` label which was used by a prior iteration of CI workflows? image Docs for how to opt-in to jobs like that will go in https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request

so how do I test, if my PR fixes the ASAN preloading issue?

ScottTodd:
> so how do I test, if my PR fixes the ASAN preloading issue?

We should make the small change to let PRs opt-in to running ASan as I mentioned above. Did you run the tests locally? CI should mainly be used for verifying changes, not as the primary check. Also note that we aren't currently committed to publishing ASan Python packages at the same support level as ASan native linux packages and tarballs, see
* https://github.com/ROCm/TheRock/pull/6559
* https://github.com/ROCm/TheRock/blob/02ac6dd96ed67540b7dddb962e3001ee1331e7ea/build_tools/github_actions/publish_rocm_to_release_buckets.py#L335-L346

build_tools/packaging/python/templates/rocm/src/rocm_sdk/tests/core_test.py
bgopesh: done

ScottTodd:
Support for running ASan CI jobs to PRs was added in https://github.com/ROCm/TheRock/pull/6881. I've synced the branch for this PR and added the opt-in label.

ScottTodd:
Well tests didn't run, maybe due to https://github.com/ROCm/TheRock/blob/fe040db029d8a319ee445b24ece0ad3d040a22ee/build_tools/github_actions/configure_multi_arch_ci.py#L1010-L1036

see https://github.com/ROCm/TheRock/actions/runs/30298125629/job/90132266272?pr=6737 

:/

ScottTodd:
I guess we'll need to push a branch and trigger https://github.com/ROCm/TheRock/actions/workflows/test_rocm_wheels.yml manually.

.github/workflows/test_rocm_wheels.yml
```diff
@@ -167,6 +179,16 @@ jobs:
...
+          # first entry in the process' initial library list, so it must be
+          # LD_PRELOAD'd before the interpreter starts.
+          # See https://github.com/ROCm/TheRock/issues/6331.
+          if [[ "${{ inputs.build_variant }}" == *asan* ]]; then
+            asan_rt="$(amdclang++ -print-file-name=libclang_rt.asan-$(uname -m).so)"
+            echo "ASAN build: preloading ASan runtime ${asan_rt}"
+            export LD_PRELOAD="${asan_rt}${LD_PRELOAD:+:${LD_PRELOAD}}"
+            export ASAN_OPTIONS="detect_leaks=0${ASAN_OPTIONS:+:${ASAN_OPTIONS}}"
+          fi
           rocm-sdk test
```
ScottTodd, line 192: Sorry for the slow review here. Looking at a few other PRs related to ASan support, I'd like to see a general "setup ASan" utility that we can use for multiple test workflows that sets environment variables, see https://github.com/ROCm/TheRock/pull/7311#discussion_r3833612012

> we could have a "setup ASan" step that sets `LD_PRELOAD` and `ASAN_OPTIONS` environment variables for the whole job, then we wouldn't need to update individual test scripts with ASan support like this. Tests should generally follow our user instructions, which in this case would be https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md#using-asan-instrumented-libraries. Hopefully we can update our tests to
> 
> 1. Install ASan artifacts/packages
> 2. Set ASan-related environment variables
> 3. Run tests in the same way they are run without ASan

dileepr1:
@bgopesh The issue #6331 got assigned to me laterally on the assumption that my another PR #8123 already fixed this issue. When I found that the issue is similar but #8123 doesn't actually fix it I went ahead and created #8341 without going through the full history.
I see that this PR came first and fixes the problem, but since #8123 already handles Scott's ask of a shared ASan utility I think we can go ahead with #8341 which updates on top of that.
cc: @ScottTodd

PR #6739, opened 2026-07-21
Add rocSHMEM Dependency

EvanPhoukong:
## Motivation

Add rocSHMEM as a build dependency in TheRock so that rocprofiler-sdk can detect
and enable its rocSHMEM API-trace feature.

## JIRA Ticket

JIRA ID: AIPROFSDK-963

## Technical Details

rocprofiler-sdk does an optional `find_package(rocshmem)`, so rocSHMEM must be
configured before `profiler`. rccl, in the same directory, depends on the profiler
and must come after. Since `add_subdirectory()` is per-directory and one-shot, the
two can't share a file.

- Moved rocSHMEM into `comm-libs/rocshmem/`, added before `profiler` from the top
  level (same hoisting `media-libs` and `storage-libs` already use). `comm-libs`
  (rccl) is still added after.
- `artifact-rocshmem.toml` moved with it, since `DESCRIPTOR` resolves relative to
  the current source dir. Contents unchanged, and `BINARY_DIR` was adjusted to keep
  the stage path at `comm-libs/rocshmem`, so the artifact layout is identical.
- Declared `rocshmem` as an optional dep of both `rocprofiler-sdk` and `rccl` so it
  resolves from the dist rather than a stray build tree — for rccl this avoids an
  ExternalProject checkout of a pinned rocSHMEM commit. rocSHMEM's config
  references `MPI::MPI_CXX`, so `therock-openmpi` is declared too when MPI is
  enabled; otherwise rocprofiler-sdk falls back to its bundled trace headers.
- Added `rocprofiler-register` to rocSHMEM's runtime deps
  (`USE_ROCPROFILER_REGISTER`, on by default).
- `BUILD_TOPOLOGY.toml`: moved rocshmem into a new `comm-libs-shmem` group so it can
  depend on `base` (for rocprofiler-register) without changing rccl's deps, and so it
  builds ahead of rccl in the comm-libs job.

## Test Plan

Test in CI.

## Submission Checklist


jbonnell-amd (approved):
Looks fine, but I am curious about if it makes sense to add a `rccl` subdirectory as well within `comm-libs` since you added one for `rocshmem`. We should have a second review from someone in DevOps to confirm if the directory changes here are valid.

comm-libs/rocshmem/artifact-rocshmem.toml
geomin12, line 1: personally, this can stay in comm-libs/ dir

comm-libs/rocshmem/CMakeLists.txt
geomin12, line 1: we could combine this with the CMakeLists.txt in comm-libs dir

any reason to separate? ideally, if rocshmem is a part of this stage, it should only build comm-libs/CMakeLists.txt

comm-libs/CMakeLists.txt
```diff
@@ -117,84 +134,3 @@ if(THEROCK_ENABLE_RCCL)
       ${_rccl_subproject_names}
   )
 endif()
-
-if(THEROCK_ENABLE_ROCSHMEM)
```
geomin12, line 121: this can actually stay i believe

geomin12 (commented):
ah merge conflcits :(

comm-libs/rocshmem/artifact-rocshmem.toml
EvanPhoukong: rocprofiler-sdk does an optional find_package(rocshmem) so rocSHMEM has to be declared before profiler/, while rccl has to come after it.

comm-libs/rocshmem/CMakeLists.txt
EvanPhoukong: They're separate only because add_subdirectory() runs once per directory and these two need to land on opposite sides of profiler/. The comm-libs build job still builds rocshmem exactly as before.

comm-libs/CMakeLists.txt
EvanPhoukong: It can't stay in this file, since this file is added after profiler/ (rccl depends on rocprofiler-sdk), so with the block here rocprofiler-sdk would be declaring a dep on a rocshmem target that doesn't exist yet and the configure dies with "non-existent target rocshmem".

geomin12 (approved):
thanks for explanations

there was a GH outage, so pls make sure the builds pass before landing

PR #6743, opened 2026-07-21
[deprecate-hipcc] Replace hipcc with amdclang++ for libhipcxx

david-salinas:
## Motivation

JIRA ID : LCOMPILER-876

Removing uses of the HIPCC tool in favour of calling amdclang++ directly.  The Clang driver has enough changes that we no longer need an intermediatory tool, and users can just call amdclang++.

## Technical Details

- build_tools/github_actions/fetch_test_configurations.py: rename libhipcxx_hipcc CI job to libhipcxx_amdclang
- test_libhipcxx_hipcc.py -> test_libhipcxx_amdclang.py: update test script to invoke amdclang++ directly with GPU arch detection via offload-arch / rocminfo / --offload-arch=native fallback chain
- test_libhipcxx_hiprtc.py: replace HIP_HIPCC_EXECUTABLE with CMAKE_CXX_COMPILER=amdclang++
- libhipcxx_utils.py: improve GPU arch detection (capture offload-arch stdout on non-zero exit, rocminfo fallback)
- math-libs/CMakeLists.txt: remove hipcc from libhipcxx COMPILER_TOOLCHAIN and RUNTIME_DEPS
- math-libs/libhipcxx: bump submodule to 8e3369197 which merges all hipcc deprecation patches (amdclang++ as default compiler, AMDCLANG compiler ID, amdhip64 link fix, -O3 optimization, --offload-arch=native for Windows MxGPU)
-
## Test Plan

Full CQE/CI Staging PSDB

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_libhipcxx_amdclang.py
+++ b/build_tools/github_actions/test_executable_scripts/test_libhipcxx_amdclang.py
@@ -25,25 +25,6 @@ logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 gpu_arch = get_gpu_architecture_portable(OUTPUT_ARTIFACTS_DIR)
 logging.info(f"++ Detected GPU architecture: {gpu_arch}")
 
-# AMDGPU_TARGETS is set authoritatively by CI. MxGPU virtual GPUs can cause
-# offload-arch to return an incorrect default (e.g. gfx906) even when the
-# actual GPU is different (e.g. gfx1101). If the detected arch is not in
-# AMDGPU_TARGETS, fall back to the first entry in AMDGPU_TARGETS.
-amdgpu_targets = os.getenv("AMDGPU_TARGETS")
-if amdgpu_targets and gpu_arch is not None:
-    # Only apply the AMDGPU_TARGETS check when offload-arch returned something.
-    # If gpu_arch is None (offload-arch failed entirely), leave it as None so
-    # cmake's enable_language(HIP) auto-detects via the HIP runtime, which
-    # works correctly on MxGPU virtual GPUs even when offload-arch doesn't.
-    valid_targets = amdgpu_targets.split(",")
-    if gpu_arch not in valid_targets:
-        # offload-arch returned a wrong arch (e.g. gfx906); compile for all
-        # AMDGPU_TARGETS so the binary runs on whichever GPU is present.
-        gpu_arch = ";".join(valid_targets)
-        logging.info(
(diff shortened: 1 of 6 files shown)


lamb-j:
Should we generalize the title if it's applying to all components?

build_tools/github_actions/test_executable_scripts/test_libhipcxx_amdclang.py
```diff
@@ -25,6 +25,25 @@
...
+    # works correctly on MxGPU virtual GPUs even when offload-arch doesn't.
+    valid_targets = amdgpu_targets.split(",")
+    if gpu_arch not in valid_targets:
+        # offload-arch returned a wrong arch (e.g. gfx906); compile for all
+        # AMDGPU_TARGETS so the binary runs on whichever GPU is present.
+        gpu_arch = ";".join(valid_targets)
+        logging.info(
+            f"++ GPU arch not in AMDGPU_TARGETS, compiling for all targets: {gpu_arch}"
+        )
+
```
obersteiner: We should put all this into get_gpu_architecture_portable so that also the hiprtc tests get this enhancement. This will be important once we also support Windows there.

math-libs/CMakeLists.txt
```diff
@@ -534,13 +534,11 @@ if(THEROCK_ENABLE_LIBHIPCXX)
       therock_explicit_finders.cmake
     COMPILER_TOOLCHAIN
       amd-hip
-      hipcc
     BUILD_DEPS
       rocm-cmake
     RUNTIME_DEPS
       amd-llvm
       hip-clr
-      hipcc
```
obersteiner, line 543: Do we need to add amdclang here? I am not sure if amdclang is a default runtime dependency.

david-salinas:
> Should we generalize the title if it's applying to all components?

This change is really only affecting libhipcxx.  We will be removing the use of hipcc from other components, but we need the libhipcxx change first because other components depend on libhipcxx.

math-libs/CMakeLists.txt
david-salinas: No, because amdclang is part of the amd-llvm package.

build_tools/github_actions/test_executable_scripts/test_libhipcxx_amdclang.py
david-salinas: OK, good idea.

obersteiner (approved):
LGTM thanks!

lamb-j (approved):

Issue #6720, opened 2026-07-21
[Issue] Windows::release `Validate Artifact Structure: test_artifacts_covered_by_packages fails` — core-hipinfo [run] missing from package.json


TriveniTadapaneni:
## Context

- **Check:** `Windows::release / Validate Artifact Structure / Validate artifact structure`
- **Run:** https://github.com/ROCm/TheRock/actions/runs/29772617751/job/88535134703
- **PR:**
-  https://github.com/ROCm/TheRock/pull/6702
- https://github.com/ROCm/TheRock/pull/6704
- **Platform:** Windows
- **TheRock ref:** `f3bd4ad54af2d97fea1fb1f2babe9b2029c5260c`

## Failure Signature

```
ERROR    test_artifact_structure:test_artifact_structure.py:437 MISSING package coverage: core-hipinfo [run] (core-hipinfo_run_generic.tar.zst)
Failed: Found 1 artifact component(s) not covered by package.json:
  core-hipinfo [run] (core-hipinfo_run_generic.tar.zst)
```
## Summary

`test_artifacts_covered_by_packages` fails on the Windows artifact structure validation job because `core-hipinfo [run]` (`core-hipinfo_run_generic.tar.zst`) is present in the Windows build artifacts but has no entry in `build_tools/packaging/linux/package.json`.

## Suspected Root Cause

`core-hipinfo` is a Windows-only artifact (`platform = "windows"` in `BUILD_TOPOLOGY.toml`, added in [PR #1216](https://github.com/ROCm/TheRock/pull/1216)). The new `test_artifacts_covered_by_packages` test added by commit [`d7cdfa0a64ae`](https://github.com/ROCm/TheRock/commit/d7cdfa0a64ae3501936b5e07664c9424a09c7cf4) (PR #6401, 2026-07-20) checks all fetched artifact components against the Linux-only `package.json`. `("core-hipinfo", "run")` was never added to `KNOWN_UNCOVERED_COMPONENTS`.

## Ai Suggested Fix

Add `("core-hipinfo", "run")` to `KNOWN_UNCOVERED_COMPONENTS` in `tests/test_artifact_structure.py`:

```python
    ("core-hipinfo", "run"),  # Windows-only artifact; no Linux package.json entry.
```

TriveniTadapaneni:
@jonatluu Could you check this issue?

TriveniTadapaneni:
[...]

PR #6747, opened 2026-07-21
Add artifacts check to github workflow. Skip check for windows platform

jonatluu:
## Motivation

Add check to ensure all artifact components added in toml files are in package.json. This ensures that the changes to the toml files will be built into packages. Resolve failure caused by running test_artifacts_covered_by_packages on  windows platform https://github.com/ROCm/TheRock/issues/6720

## Technical Details

New test_artifacts_covered_by_packages added to existing TestArtifactStructure class in tests/test_artifact_structure.py. Reuse existing THEROCK_ARTIFACTS_DIR pointing to unextracted artifacts as well as resolves package.json relative to the source file

## Test Plan
Run tests locally and run test in CI 

## Test Result
https://github.com/ROCm/TheRock/actions/runs/29862120656

validate artifact structure now skipped on windows release: https://github.com/ROCm/TheRock/actions/runs/29862120656/job/88957817801?pr=6747

## Submission Checklist


tests/test_artifact_structure.py
```diff
@@ -81,6 +81,11 @@
 }
 
 
+def is_windows_platform() -> bool:
+    """Check if the artifacts being validated are Windows artifacts and return bool"""
+    return os.getenv("PLATFORM", "linux").lower() == "windows"
```
ScottTodd, line 86: This is an unreliable check and we should limit the number of environment variables needed to run tests (rule of thumb: write tests for local use _first_, then consider CI). See the code that I linked: https://github.com/ROCm/TheRock/blob/30ba1d83bd728dd3675e32df19d4102fa1707500/tests/test_rocm_sanity.py#L28-L29

It's trickier here though, as I should be able to test Linux and Windows artifacts alike on both operating systems.

Given that the artifacts this file is testing are not controlled by the test itself (see `ARTIFACTS_DIR = os.getenv("THEROCK_ARTIFACTS_DIR", "")`), I think it would be enough to default to checking `platform.system().lower()` and still allow overriding via the env var, expecting Windows developers to be testing Windows artifacts and vice-versa. Anyone who wants to test artifacts for a different operating system can then set the environment variable.

tests/test_artifact_structure.py
ScottTodd, line 50: Given my other comment,  let's put `PLATFORM` up here so make it clear where all environment variables are loaded

```python
ARTIFACTS_DIR = os.getenv("THEROCK_ARTIFACTS_DIR", "")
PLATFORM = os.getenv("PLATFORM", platform.system()).lower()
```

ScottTodd, line 26: Probably also worth mentioning the platform detail up here too.

Maybe this?

```
Typical usage:
    THEROCK_ARTIFACTS_DIR=/path/to/archives \\
        python -m pytest tests/test_artifact_structure.py -v --log-cli-level=info

Set PLATFORM=linux or PLATFORM=windows if you want to test artifacts
built for a different operating system than your current system.
```

ScottTodd (approved):
A few small suggestions, otherwise lgtm

tests/test_artifact_structure.py
jonatluu: Will add this in a future update. Wanted to merge the check first to prevent more failures from being missed.

Issue #3734, opened 2026-03-03
[Feature]: Standardize logging for TheRock


jonatluu:
### Suggestion Description

Logging is currently inconsistent in TheRock and standardized logging with structured data, consistent timestamps, and performance time will allow for easier readability and parsing of logs. Currently a logging function is copy pasted in many python files within build_tools, a minimal logging api will be introduced to replace these copy pasted logging functions. Once the minimal logging api is in use for build_tools further design discussions and more advanced features will be added to address concerns such as:
- logs not being consistently uploaded to persistent storage in an organized way
- ROCm build workflows logs will upload but ROCm test workflows and PyTorch/JAX/etc. build/release workflows do not 
- considering existing logging tools such as https://github.com/ROCm/TheRock/blob/main/build_tools/teatime.py and how best to use them with the addition of this logging api
- poor searchability due to no standardization of how logs are written, uploaded, stored, processed
- difficultly debugging CI/CD failures due to inconsistent log formatting

### Plan for new logging feature 

1. A minimal logging api will be added to replace to copy and pasted logging functions used in many python files in build_tools
2. Once the logging api is in use further discussions can take place for more complex features to address various concerns
3. The first feature proposed will be to standardize uploading logs to persistent storage for ROCm build workflows, ROCm test workflows, PyTorch/JAX/etc. build/release workflows. Further discussion will take place for this feature but the addition of a standardized logging api will help standardize uploading logs and solve the issue of ROCm build workflows having logs uploaded while ROCm test workflows, and PyTorch/JAX/etc. build/release workflows do not have uploaded logs. 

Minimal logger proposed changes: https://github.com/ROCm/TheRock/pull/4206
Old standardized logging pr for reference: https://github.com/ROCm/TheRock/pull/2620

raramakr:
Problem Statement                                                                                                                                               
                  
  - TheRock has 6 fragmented logging implementations (1,143+ print statements, 47 basicConfig() calls, custom wrappers) with no interoperability                  
  - teatime.py only captures subprocess output, NOT Python script's own print()/logging calls, causing loss of all Python-generated logs                          
  - No per-component log separation - all package build output mixed together, cannot isolate individual failures
  - No centralized configuration - 47 files independently call basicConfig() causing import-order conflicts
  - No automatic S3 upload for logs - packaging and test logs lost after 90-day GitHub Actions retention period

  Solution Overview

[...]

PR #6748, opened 2026-07-21
Add logging support for Linux packaging

raramakr:
Introduce unified logging infrastructure for the packaging system:

 Add log_utils.py providing:
  - configure_logging() for one-time setup with verbosity control
  - TheRockLogger class as the primary public API with familiar
    logger.info()/warning()/error() interface
  - capture_console() for capturing stdout/stderr to files with
    race-condition-safe shutdown (threads joined before closing FDs)
  - GitHub Actions integration (collapsible groups, annotations via
    github=True parameter with optional file/line support)
  - Thread-safe stream handling for subprocess output

Add per-package log files at {dest-dir}/logs/{pkgtype}-{pkg_name}.log
  Each package build generates its own log file for easier debugging
  and isolation of build issues.
  
  Environment variable controls:
  - THEROCK_LOG_LEVEL: Set log level (DEBUG, INFO, WARNING, ERROR)
  - THEROCK_LOG_ENABLED: Enable/disable logging (default: enabled)

  Add native packaging logs upload to S3.

  All Linux packaging modules (build_package.py, deb_package.py,
  rpm_package.py, packaging_utils.py, packaging_summary.py) use the
  unified TheRockLogger API.

  Add comprehensive test coverage with platform-specific decorators
  (@skipIf/@skipUnless) for POSIX and Windows capture paths.

JIRA ID: https://github.com/ROCm/TheRock/issues/3734

### DIFF
--- a/build_tools/_therock_utils/log_utils.py
+++ b/build_tools/_therock_utils/log_utils.py
@@ -1,0 +1,545 @@
+"""Unified logging utilities for TheRock build tools.
+
+Provides a consistent logging API across all Python scripts with:
+- Console output capture to files (Python print + subprocess output)
+- CI-friendly formatting (GitHub Actions groups, annotations)
+- Verbosity control via environment variables or API
+- Thread-safe stream handling
+
+Usage::
+
+    from _therock_utils.log_utils import configure_logging, get_logger, capture_console
+
+    # One-time setup in main()
+    configure_logging(verbose=args.verbose)
+
+    # Get logger for module
+    logger = get_logger(__name__)
+    logger.info("Processing artifacts")
+
+    # Capture all output to file
+    with capture_console("build.log"):
+        logger.info("Building...")
+        subprocess.run(["cmake", ".."])  # Also captured
+"""
+
+import io
+import logging
+import os
+import sys
+import threading
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Iterator, TextIO
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+DEFAULT_FORMAT = "[%(asctime)s][%(levelname)-8s]: %(message)s"
+DEFAULT_DATE_FORMAT = "%y-%m-%d %H:%M:%S"
+CI_FORMAT = "[%(levelname)-8s] %(name)s: %(message)s"
+
+# Environment variables
+ENV_LOG_ENABLED = "THEROCK_LOG_ENABLED"
+ENV_LOG_LEVEL = "THEROCK_LOG_LEVEL"
+ENV_GITHUB_ACTIONS = "GITHUB_ACTIONS"
+
+# Module state
+_configured = False
+_verbosity = 0
+_capture_depth = threading.local()  # Per-thread capture depth counter
+_MAX_CAPTURE_DEPTH = 3
+
+
+def _get_capture_depth() -> int:
+    """Get capture depth for current thread."""
+    return getattr(_capture_depth, "value", 0)
+
+
+def _increment_capture_depth() -> int:
+    """Increment and return capture depth for current thread."""
+    depth = getattr(_capture_depth, "value", 0) + 1
+    _capture_depth.value = depth
+    return depth
+
+
+def _decrement_capture_depth() -> None:
+    """Decrement capture depth for current thread."""
+    _capture_depth.value = max(0, getattr(_capture_depth, "value", 0) - 1)
+
+
+# ---------------------------------------------------------------------------
+# CI Detection
+# ---------------------------------------------------------------------------
+
+
+def is_ci() -> bool:
+    """Check if running in CI (GitHub Actions)."""
+    return os.environ.get(ENV_GITHUB_ACTIONS, "").lower() == "true"
+
+
+def is_logging_enabled() -> bool:
+    """Check if logging is globally enabled."""
+    return os.environ.get(ENV_LOG_ENABLED, "1") != "0"
+
+
+# ---------------------------------------------------------------------------
+# Flushing Handler
+# ---------------------------------------------------------------------------
+
+
+class FlushingStreamHandler(logging.StreamHandler):
+    """StreamHandler that flushes after every emit for CI compatibility."""
+
+    def emit(self, record: logging.LogRecord) -> None:
+        super().emit(record)
+        self.flush()
+
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+
+def configure_logging(
+    *,
+    enabled: bool | None = None,
+    level: int | str = logging.INFO,
+    verbose: bool = False,
+    ci_mode: bool | None = None,
+    stream: TextIO | None = None,
+) -> None:
+    """Configure root logger for script execution.
+
+    Call once at the start of main(). Replaces logging.basicConfig().
+
+    Args:
+        enabled: Enable/disable logging. None = check THEROCK_LOG_ENABLED env.
+        level: Base logging level (default INFO).
+        verbose: If True, sets level to DEBUG.
+        ci_mode: CI-specific formatting. None = auto-detect from GITHUB_ACTIONS.
+        stream: Output stream (default sys.stderr).
+    """
+    global _configured
+
+    # Check if globally disabled
+    if enabled is None:
+        enabled = is_logging_enabled()
+
+    if not enabled:
+        logging.disable(logging.CRITICAL)
+        _configured = True
+        return
+
+    # Determine level
+    env_level = os.environ.get(ENV_LOG_LEVEL, "").upper()
+    if env_level:
+        level = getattr(logging, env_level, logging.INFO)
+    elif verbose:
+        level = logging.DEBUG
+
+    # Determine CI mode
+    if ci_mode is None:
+        ci_mode = is_ci()
+
+    # Select format
+    fmt = CI_FORMAT if ci_mode else DEFAULT_FORMAT
+    datefmt = None if ci_mode else DEFAULT_DATE_FORMAT
+
+    # Configure root logger
+    root = logging.getLogger()
+    root.setLevel(level)
+
+    # Remove existing handlers to avoid duplicates
+    for handler in root.handlers[:]:
+        root.removeHandler(handler)
+
+    # Add new handler
+    handler = FlushingStreamHandler(stream or sys.stderr)
+    handler.setFormatter(logging.Formatter(fmt, datefmt))
+    root.addHandler(handler)
+
+    _configured = True
+
+
+def get_logger(name: str) -> logging.Logger:
+    """Get a configured logger instance.
+
+    Args:
+        name: Logger name (typically __name__).
+
+    Returns:
+        Configured Logger instance.
+    """
+    if not _configured:
+        configure_logging()
+    return logging.getLogger(name)
+
+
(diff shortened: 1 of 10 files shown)


build_tools/packaging/linux/upload_package_repo.py
```diff
@@ -777,6 +821,26 @@ def main():
     print(f"Package repository URL: {install_url}")
     _emit_github_output("package_repository_url", install_url)
 
+    # Upload packaging logs and write step summary (only available with new-style --run-id)
+    log_index_url = None
+    if args.run_id:
```
ScottTodd: What is "new-style --run-id"? The only use of this script sets `--run-id`: https://github.com/ROCm/TheRock/blob/0888ca35dad1ae9a5f9b46bf56269520fde79cdb/.github/workflows/multi_arch_build_native_linux_packages.yml#L134-L140

Let's just make `--run-id` required and remove the code branches?

https://github.com/ROCm/TheRock/blob/0888ca35dad1ae9a5f9b46bf56269520fde79cdb/build_tools/packaging/linux/upload_package_repo.py#L726-L732

build_tools/packaging/linux/upload_package_repo.py
```diff
@@ -777,6 +821,26 @@ def main():
...
+        )
+        if log_index_url:
+            _emit_github_output("packaging_logs_url", log_index_url)
+
+    # Write GitHub Actions step summary
+    pkg_type_upper = args.pkg_type.upper()
+    gha_append_step_summary(f"### {pkg_type_upper} Package Build Results")
+    gha_append_step_summary(f"[Package Repository]({install_url})")
+    if log_index_url:
+        gha_append_step_summary(f"[Packaging Logs]({log_index_url})")
```
ScottTodd: I'd like to see a clean CI run showing what this is doing. I retriggered a failed job on https://github.com/ROCm/TheRock/actions/runs/30571225711?pr=6748 to see if this will run.

A step summary formatted like what we have for rocm python packages is what I would expect:

Image

build_tools/_therock_utils/log_utils.py
```diff
@@ -0,0 +1,545 @@
...
+        logging.disable(logging.CRITICAL)
+        _configured = True
+        return
+
+    # Determine level
+    env_level = os.environ.get(ENV_LOG_LEVEL, "").upper()
+    if env_level:
+        level = getattr(logging, env_level, logging.INFO)
+    elif verbose:
+        level = logging.DEBUG
```
ScottTodd, line 154: May want to prefix this `ENV_LOG_LEVEL` with `THEROCK_`

build_tools/_therock_utils/log_utils.py
ScottTodd, line 1: We could write some unit tests for this module as a way to demonstrate the logging and check some of the finer details:

* Show how to configure verbosity, including checking that verbose logs are skipped when the logging level excludes them
* Show how to configure file output
* Show that capture depth includes output from subcommands

Importantly I'd like to see that working on both Linux and Windows so we can adopt this project-wide with more confidence.

ScottTodd (approved):

build_tools/_therock_utils/log_utils.py
raramakr: Added unit test covering the mentioned scenario. Also added  subprocess console capturing for windows

build_tools/_therock_utils/log_utils.py
raramakr: already prefixed as THEROCK_LOG_LEVEL (line 45).

build_tools/packaging/linux/upload_package_repo.py
raramakr: Made run-id a required parameter and removed the code branches. Also removed obsolete(single arch) arguments

build_tools/_therock_utils/log_utils.py
```diff
@@ -0,0 +1,692 @@
...
+    if is_ci():
+        print(f"::group::{title}", flush=True)
+    try:
+        yield
+    finally:
+        if is_ci():
+            print("::endgroup::", flush=True)
+
+
+def github_warning(
```
nunnikri: For github annotation do we need a different fn? Is it better to have a single log fn For both github and python?

```
# Default — Python logging style
log("warning", f"Failed to build {pkg_name}")
# → [WARNING ]: Failed to build amdrocm-fft

# With github=True — GitHub Actions annotation
log("warning", f"Failed to build {pkg_name}", github=True)
# → ::warning::Failed to build amdrocm-fft
```

build_tools/packaging/linux/build_package.py
```diff
@@ -627,26 +629,31 @@ def run(args: argparse.Namespace):
...
-            print(f"Creating {config.pkg_type} package: {pkg_name}")
+            logger.info(f"Creating {config.pkg_type} package: {pkg_name}")
 
             # Build all package variants for this package
-            output_list = build_package_variants(pkg_name, config)
+            # Capture per-package logs
+            logs_dir = Path(config.dest_dir) / "logs"
+            logs_dir.mkdir(parents=True, exist_ok=True)
+            pkg_log_file = logs_dir / f"{config.pkg_type}-{pkg_name}.log"
+            with capture_console(pkg_log_file):
```
nunnikri, line 639: We already have a build_tools/teatime.py which captures the console and logs. 
Did we checked already that cant be used? and need a new fn for this?

build_tools/_therock_utils/log_utils.py
```diff
@@ -0,0 +1,692 @@
...
+    orig_stderr_handle = kernel32.GetStdHandle(STD_ERROR_HANDLE)
+
+    # Validate handles
+    if orig_stdout_handle == INVALID_HANDLE_VALUE:
+        raise OSError("GetStdHandle failed for stdout")
+    if orig_stderr_handle == INVALID_HANDLE_VALUE:
+        raise OSError("GetStdHandle failed for stderr")
+    orig_stdout = sys.stdout
+    orig_stderr = sys.stderr
+    orig_stdout_fd = os.dup(1)
```
nunnikri, line 361: build_tools/_therock_utils/log_utils.py — Race condition in POSIX capture_console shutdown sequence

In the POSIX path of capture_console, orig_stdout_fd / orig_stderr_fd are closed before the tee threads are joined. The tee threads hold os.fdopen wrappers that reference those same fds and may still be writing when the close happens.

The race:
```
# finally block — current order (wrong)
os.dup2(orig_stdout_fd, 1)
os.dup2(orig_stderr_fd, 2)
os.close(orig_stdout_fd)     # ← fd closed here
os.close(orig_stderr_fd)     # ← fd closed here

for t in tee_threads:
    t.stop()
    t.join(timeout=5.0)      # ← tee thread may still write to the now-closed fd!

```
If _TeeWriter calls out.write(text) on the os.fdopen(orig_stdout_fd, ...) wrapper after os.close(orig_stdout_fd) runs on the main thread, it writes to an invalid (or recycled) file descriptor — undefined behaviour.

Fix — join threads before closing the fds:
```
# finally block — correct order
# 1. Join tee threads first so no thread is writing anymore
for t in tee_threads:
    t.stop()
    t.join(timeout=5.0)

# 2. Now safe to restore and close the original fds
os.dup2(orig_stdout_fd, 1)
os.dup2(orig_stderr_fd, 2)
os.close(orig_stdout_fd)
os.close(orig_stderr_fd)

```
The same ordering issue applies to the Windows implementation in _capture_console_windows — verify the shutdown sequence there as well.

build_tools/_therock_utils/log_utils.py
```diff
@@ -0,0 +1,692 @@
...
+        yield log_path
+        return
+
+    _increment_capture_depth()
+
+    # Ensure parent directory exists
+    log_path.parent.mkdir(parents=True, exist_ok=True)
+
+    # Use Windows-specific implementation for subprocess capture support
+    if sys.platform == "win32":
```
nunnikri, line 524: Usual practice is to code like
```

    @unittest.skipUnless(sys.platform == "win32", "Windows-only path")
    def test_windows_capture(self):
        ...

    @unittest.skipIf(sys.platform == "win32", "POSIX-only path")
    def test_posix_capture(self):
        ...

    # Platform-agnostic tests run on both:
    def test_basic_capture_writes_to_file(self):
        # Tests the public API — works regardless of which internal path runs
        with capture_console(self.log_file):
            print("hello")
        self.assertIn("hello", self.log_file.read_text())
```

build_tools/packaging/linux/build_package.py
ScottTodd: This is a good point. https://github.com/ROCm/TheRock/blob/main/build_tools/teatime.py can be used as a wrapper around arbitrary commands to capture their output to a file. There is definitely still space for our own scripts to use more structured logging to both the console and files alongside that.

We have a few particularly heavy workflow steps that would benefit from log capture and upload:
* https://github.com/ROCm/TheRock/blob/2dce0a073c49233d48c775e14f9c539b54979e4d/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml#L331 (see "Upload pytorch build logs to AWS" in https://github.com/ROCm/TheRock/issues/4987)
* https://github.com/ROCm/TheRock/blob/2dce0a073c49233d48c775e14f9c539b54979e4d/.github/workflows/build_portable_linux_python_packages.yml#L153
* https://github.com/ROCm/TheRock/blob/2dce0a073c49233d48c775e14f9c539b54979e4d/.github/workflows/multi_arch_build_native_linux_packages.yml#L114
* https://github.com/ROCm/TheRock/blob/2dce0a073c49233d48c775e14f9c539b54979e4d/.github/workflows/multi_arch_build_linux_jax_wheels.yml#L215

Some of these we have reasonable control over, some we don't.

ScottTodd:
I still think this is generally in "land and iterate" territory, but there have been a few changes since I last reviewed, so I re-requested review and can take another look at the changes next week.

build_tools/packaging/linux/build_package.py
raramakr: teatime.py cannot capture Python print/logging statements from the calling code - it only wraps a subprocess. 
capture_console() uses fd-level redirection to capture everything within the context block, including mixed Python and subprocess output.

build_tools/_therock_utils/log_utils.py
raramakr: The ::warning:: syntax is a special GitHub Actions workflow command
  (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message) that triggers GitHub's annotation system -
 logger.warning() can't achieve this since GitHub doesn't parse standard log output.

The github_warning function is not used anywhere currently
Either we can remove it or keep if for future use

build_tools/_therock_utils/log_utils.py
raramakr: Added the same

build_tools/_therock_utils/log_utils.py
raramakr: Done

build_tools/_therock_utils/log_utils.py
raramakr: Addressed the same
TheRockLogger class as the primary public API with familiar
logger.info()/warning()/error() interface . No more github_warning function. If we need githb annotation , arguments can be passed with public API

nunnikri (approved):

raramakr:
Merging the PR.
If any further review comments, will address in another PR

PR #6753, opened 2026-07-21
ci(jax): run the JAX suite at three sizes

gulsumgudukbay:
Stacked on #6853, which splits the suite into two runner-sized jobs. Review the top commit; the base retargets to `main` when that one merges.

## Motivation

- Since #6558 the JAX suite is all-or-nothing: ~29k tests, 35-70 minutes per job. That is right for a nightly and too much for a PR, so PR CI today builds the JAX wheels and never tests them.
- Multi-GPU runners are scarce, and the multi-accelerator tests are a small, slow-moving part of the suite. Taking an 8-GPU slot for them every night costs more than it finds.
- This adds a `test_size` that says how much of the suite a run is worth, and connects the JAX tests to the PR path.

| Size | Single-GPU job | Multi-GPU job | End-to-end workloads | Trigger | Step budget |
| --- | --- | --- | --- | --- | --- |
| `small` | the subset in `small_tests.txt` | no | none | PR CI | 45 min |
| `medium` | the whole single-accelerator suite | one day a week | two, on that day | nightly | 120 min, 90 for multi |
| `large` | the whole single-accelerator suite | yes | all of them | prerelease | 120 min, 90 for multi |

Replaces the earlier draft of this PR, which predated #6558 and had placeholder steps for the full suite as well.

## Technical Details

- **A size selects over the one suite.** The `ci/run_pytest_rocm*.sh` scripts stay the only way tests run. `run_jax_tests.py --test-list` takes a file of test paths and ignores everything else under `tests/`, so a suite script keeps its own pytest invocation, markers and deselects.
- **`--ignore`, not `-k`.** `-k` leaves the filtered-out modules imported, and importing them is most of what a 45-minute run has time for.
- **The list is matched against the checkout, not trusted.** A test a version has since dropped is warned about and skipped; a list matching nothing fails the run instead of quietly testing everything.
- **A size also decides runners.** The weekly cadence itself comes from #6853; this restates it in size terms, so `configure_jax_test_matrix.py` emits the multi-GPU job for `large` always and for `medium` one day a week. The weekday stays a constant next to the rule, and the rule stays unit-tested against fixed dates rather than the clock.
- **End-to-end workloads ride on the multi-GPU job.** They need those GPUs and they run on the same cadence, so they are a step of that job rather than a job of their own, and wiring them costs no second slot in that pool. `medium` asks for two workloads and `large` for the whole set.
- **They are a placeholder for now.** The workloads live in `ROCm/jax-lab`, which is private; the step warns and runs nothing until that repository is public.
- **PR path.** `multi_arch_build_linux_jax_wheels_ci.yml` gains the test matrix and the test job, wired from `multi_arch_ci_linux.yml` on one representative family, so a PR pays for the JAX tests once rather than once per family it builds. CI publishes wheels to a run-scoped page rather than a package index, so `install_jax_wheels.py` learns `--find-links`.
- **Release path.** `test_size` follows `release_type`: `prerelease` gives `large`, everything else `medium`.

## Test Plan

- `pre-commit run --all-files`, including actionlint on the five workflows.
- Unit tests for the size rule, the selection and the wheel source: `configure_jax_test_matrix_test.py`, `run_jax_tests_test.py`, `jax_install_scripts_test.py`.
- Run the matrix script for each size, on and off the weekly day.
- Dry-run both jobs against a real `ROCm/jax` checkout, one with the small list and one on the multi subset.

## Test Result

- pre-commit clean. 751 unit tests pass; the four failures (`stage_reuse_decision_test` x2, `configure_multi_arch_ci_test`, `post_stage_upload_test`) fail identically on `main`.
- The matrix holds: `small` gives one job on the 1-GPU runner; `medium` gives one on a Tuesday and adds `linux-gfx942-8gpu-ossci-rocm` on a Sunday; `large` gives both whatever day it is.
- The small job resolves to 42 tests and 177 `--ignore` arguments against that checkout, nothing skipped as missing, and execs `ci/run_pytest_rocm.sh`.
- The multi job execs `ci/run_pytest_rocm_multi.sh` with no selection applied, and is the only one that reaches the end-to-end step.
- GPU test execution on a real CI runner: TODO before merge.

## Submission Checklist


ScottTodd (commented):
I don't expect I'll have time to serve as primary reviewer on these PRs. Will defer to Erman who can tag me in as needed.

gulsumgudukbay:
[Build run 31568063824](https://github.com/ROCm/TheRock/actions/runs/31568063824):  test_size=large, which is an input that this PR brings. It allows the size mapping and the end-to-end placeholder step firing with E2E_SCOPE=full.
[Release run 31568067537](https://github.com/ROCm/TheRock/actions/runs/31568067537): release_type=dev derives test_size=medium, and today is wednesday, so this is the run that shows a nightly dispatch taking zero 8-GPU runners. That's the capacity argument cc: @amd-shiraz (cc'ing you because we had the capacity discussion in https://github.com/ROCm/TheRock/pull/6853 before and these two PRs are intertwined.)

external-builds/jax/install_jax_wheels.py
```diff
@@ -35,10 +38,24 @@
...
+
+    A find-links page adds to PyPI while an index replaces it, so a caller that
+    gave both meant the page: an index that does not carry the run's wheels
+    cannot install them.
+    """
+    if args.find_links:
+        return ["--find-links", args.find_links]
+    if args.index_url:
+        return ["--index-url", args.index_url]
+    return []
```
erman-gurses, line 52: Should we handle the case where neither find_links nor index_url is provided? In that case wheel_source() returns an empty list, so neither --find-links nor --index-url is passed to pip.

external-builds/jax/run_jax_tests.py
```diff
@@ -180,14 +204,78 @@ def jax_version_from_ref(jax_ref: str) -> str:
...
+    if missing:
+        log(
+            f"::warning::{len(missing)} selected test(s) are not in this JAX"
+            f" checkout and will not run: {', '.join(p.name for p in missing)}"
+        )
+    if len(missing) == len(wanted):
+        raise TestListError(
+            f"none of the {len(wanted)} selected tests exist under"
+            f" {jax_dir / RELATIVE_TESTS_DIR}"
+        )
```
erman-gurses, line 261: This is a bit interesting - what happens if a significant portion of the selected tests are missing? In the extreme case, if only one selected test still exists and it passes, the job would still be considered successful. Maybe we can consider failing or warning in that kind of case if majority of the tests are missed.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -117,6 +113,14 @@ on:
...
         type: string
         required: true
+      test_size:
+        description: "How much to run; medium and large also take a multi-GPU runner."
+        type: choice
+        options:
+          - small
+          - medium
+          - large
+        default: medium
```
erman-gurses, line 127: Maybe we can the workflow_dispatch description match the workflow_call behavior more closely?

.github/workflows/test_multi_arch_linux_jax_wheels.yml
```diff
@@ -296,16 +322,33 @@ jobs:
...
-        # job reached 100% of the tests and was still killed by the timeout.
-        timeout-minutes: 120
+        # The single-accelerator suite runs 35-70 min across versions, so 60
+        # left no headroom: one job reached 100% of the tests and was still
+        # killed. The multi-accelerator subset is a fraction of that.
+        timeout-minutes: ${{ inputs.test_subset == 'multi' && 90 || (inputs.test_size == 'small' && 45 || 120) }}
+        env:
+          # A size is a selection over the one suite; only small narrows it, and
+          # only for the subset it selects from.
+          JAX_TEST_LIST: ${{ (inputs.test_size == 'small' && inputs.test_subset == 'all') && 'external-builds/jax/test_selection/small_tests.txt' || '' }}
```
erman-gurses: I think there is an issue which comes from `test_size` and `test_subset` are exposed as two independent inputs. Please see agent's comment:


_What should happen for the small + multi combination? workflow_dispatch allows these inputs to be selected independently, but in that case JAX_TEST_LIST is empty and we run the full multi subset. Should you reject this combination, or clarify that small is only valid with test_subset=all?_

external-builds/jax/run_jax_tests.py
```diff
@@ -180,14 +204,78 @@ def jax_version_from_ref(jax_ref: str) -> str:
...
+    has nothing there to ignore.
+
+    Raises:
+        TestListError: if the selection names a path outside tests/, or if the
+            checkout has none of the tests it names. Silently running the whole
+            suite, or none of it, would both pass for a subset.
+    """
+    wanted = set()
+    for path in selected:
+        if path.is_absolute() or RELATIVE_TESTS_DIR not in path.parents:
```
erman-gurses: From agent:

_Does a path containing .. bypass this check? For example, tests/../benchmarks/a_test.py has tests in path.parents, but resolves outside the tests/ tree. Should we validate the resolved path against the resolved tests directory instead?_

.github/workflows/test_multi_arch_linux_jax_wheels.yml
erman-gurses, line 82: Since we have only all and multi -  we can make it choice.

erman-gurses (commented):
Please see my comments

external-builds/jax/install_jax_wheels.py
gulsumgudukbay: Yes you are right, I fixed it in `32af71f`: passing neither is now rejected. `jax-rocm7-plugin` is a published name on PyPI (our latest version is 0.11.0.post1 today), so a run with no wheel source could quietly install a released plugin in place of the one it had just built, which is the mismatch you were pointing at. `main()` now errors with "missing required argument: --find-links or --index-url", suggesting `https://pypi.org/simple` for the case where published wheels really are wanted, and a test covers it

external-builds/jax/run_jax_tests.py
gulsumgudukbay: Each missing test is already named in a `::warning::` annotation, and a selection where nothing exists at all is fatal, so the "one test passed, job green" case cannot happen. A majority threshold would be arbitrary across versions and the same list may lose files as upstream JAX renames or drops tests. So I stopped at the per test warning and the all-missing error.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
gulsumgudukbay: Done: I made `workflow_dispatch` carry the same naming as `workflow_call` (what each size runs on the 1-GPU runner and when it also takes a multi-GPU one). Same for `test_scope` on the base PR, where the dispatch description had drifted from the call one.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Here is the full picture, since the two inputs interact in one place only. `small` is defined by `small_tests.txt`(42 single-accelerator files chosen to cover the LLM-weighted module graph) and that is a selection over `ci/run_pytest_rocm.sh`. 
The multi job runs a different script, `ci/run_pytest_rocm_multi.sh`, and there is no PR-sized cut of it. that is, there is no reduced selection for the multi subset and no reason to build one: the only thing that would use it is a PR check, and small never takes a multi-GPU runner, so CI cannot produce this pair anyways, only a manual dispatch can. That is why `JAX_TEST_LIST` is set only when the size is `small` **and** the subset is `all`.

| size + subset | what runs |
| --- | --- |
| `small` + `all` | the 42 files in `small_tests.txt` |
| `medium` / `large` + `all` | the whole single-accelerator suite |
| `small` + `multi` | all multi-accelerator tests, workloads `partial` |
| `medium` / `large` + `multi` | same tests; `large` makes the workloads `full` |

So `small` + `multi` is defined rather than broken: nothing is skipped and nothing reports a false pass, the run just is not smaller, and size still applies on that job, to the end-to-end scope instead of to test selection. I left it legal because it is the natural way to dispatch the multi subset by hand, and rejecting it would fail that dispatch purely over the size box; both `test_size` descriptions now say the PR-sized selection exists for the `all` subset only.

external-builds/jax/run_jax_tests.py
gulsumgudukbay: you are right, it did bypass it. Each selected path is now resolved and checked with `is_relative_to()` against the resolved `tests/` directory, so `tests/../benchmarks/a_test.py` is rejected, and two tests cover it

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Done, `workflow_dispatch` offers it as a `choice` of `all` / `multi`. It had to stay `type: string` on the `workflow_call` side, since `choice` is not allowed there, and the change itself sits in the base PR #6853.

gulsumgudukbay:
@erman-gurses  the runs are below:
[Run 31770267839](https://github.com/ROCm/TheRock/actions/runs/31770267839), test_size=large. The size equivalent of the above: both jobs, plus the end-to-end step at full scope (still the placeholder warning).
[Run 31770273845](https://github.com/ROCm/TheRock/actions/runs/31770273845), test_size=small. The PR-sized path: one 1-GPU job running the 42-file selection, which covers the resolved-path validation in ignore_arguments().

external-builds/jax/run_jax_tests.py
erman-gurses: I agree that the missing tests won't be silent because of the warnings, but I think the one-test-passes / job-green case can still happen. For example, with 42 selected tests and 41 missing, len(missing) == len(wanted) is false, so the remaining test runs and the job can still pass. I'm fine if we intentionally accept that behavior, but wanted to make sure we're aligned on it.

erman-gurses (approved):
LGTM -  @ScottTodd, FYI, I do not see any blocker from my side.

gulsumgudukbay:
newest CI workflows: https://github.com/ROCm/TheRock/actions/runs/31985907382 https://github.com/ROCm/TheRock/actions/runs/31986007641

gulsumgudukbay:
Rebased onto the rebased #6853 (itself now on main with #7387 in it) and validated on the nightly packages the HUD runs against (10.1.0a20260818): https://github.com/ROCm/TheRock/actions/runs/32202425094

`release_type: dev` puts the run at `test_size: medium`, which is what a nightly takes, so the size this exercises is the one the nightly will.

amd-shiraz:
had a detailed convo about gpu utilization: this will not add a lot of new load on gfx942 1 gpu and multi gpu runs once a week on sunday. good to land from my side. @ScottTodd  pls check if any open items from your end based on conversations i see here. if we can land this before tonight's nightly will be good. thnx

amd-shiraz (approved):

.github/workflows/test_multi_arch_linux_jax_wheels.yml
```diff
@@ -300,17 +327,33 @@ jobs:
...
-        # multi-accelerator subset is a fraction of that.
-        timeout-minutes: ${{ inputs.test_subset == 'multi' && 90 || 120 }}
+        # The single-accelerator suite runs 35-70 min across versions, so 60
+        # left no headroom: one job reached 100% of the tests and was still
+        # killed. The multi-accelerator subset is a fraction of that.
+        timeout-minutes: ${{ inputs.test_subset == 'multi' && 90 || (inputs.test_size == 'small' && 45 || 120) }}
+        env:
+          # A size is a selection over the one suite; only small narrows it, and
+          # only for the subset it selects from.
+          JAX_TEST_LIST: ${{ (inputs.test_size == 'small' && inputs.test_subset == 'all') && 'external-builds/jax/test_selection/small_tests.txt' || '' }}
```
ScottTodd: Seems like we could move a bunch of these `${{ inputs.test_subset == 'multi' && 90 || (inputs.test_size == 'small' && 45 || 120) }}` branches into `configure_jax_test_matrix.py` as outputs that get set on the matrix for the workflow to just read instead of needing to decide on its own.

ScottTodd (approved):
No concerns on my end, thanks to others for reviewing and your patience here.

.github/workflows/test_multi_arch_linux_jax_wheels.yml
gulsumgudukbay: Done in 2e4ae69. `configure_jax_test_matrix.py` now emits `test_timeout_minutes` and `test_list` with each matrix entry, and the workflow reads them:

```yaml
timeout-minutes: ${{ inputs.test_timeout_minutes }}
env:
  JAX_TEST_LIST: ${{ inputs.test_list }}
```

so the size rule lives in one place next to the multi-GPU one, with the numbers as named constants and unit tests over the emitted jobs. The two new inputs default to 120 minutes and the whole subset, which is what a direct `workflow_dispatch` of the test workflow gets unless it names a list.

gulsumgudukbay:
Rebased onto the updated #6853 and validated on the nightly packages the HUD runs against (10.1.0a20260819): https://github.com/ROCm/TheRock/actions/runs/32307758886

`release_type: dev` puts the run at `test_size: medium`, so it exercises the step budget and the selection now coming from the matrix rather than from expressions in the test workflow, and the notice a nightly prints off the weekly day. The PR-sized path is covered by the presubmit `Multi-Arch CI` run on this head.

PR #6754, opened 2026-07-21
ci(manifest-diff): Post manifest-diff report link as a bump-PR comment

amd-hsivasun:
## Motivation

The manifest-diff HTML report is generated and uploaded to S3 on every multi-arch CI run and already surfaces in the **setup** job's step summary for general discoverability (see #6147). But for submodule-bump PRs specifically (`bump-{submodule}-{sha}`, created by `bump_automation.py`), the step summary link is buried inside a job that reviewers of a one-line version-bump PR rarely open. This PR posts that same report link directly as a PR comment, but **only on bump PRs**.

### `generate_manifest_diff_report.py` gains a `post_comment` subcommand

The new behavior is added as a second argparse subcommand (`post_comment`), alongside the existing `generate` subcommand — each with its own scoped arguments. Existing local/manual `generate` invocations are unaffected.

- `generate` additionally factors the "Commit Range" line out of `generate_step_summary()` into a new `build_commit_range_summary(diff)` function, emitted as a step output (`commit_range_summary`) via `gha_set_output()`.
- `post_comment` (new) computes the S3 report URL the same way `upload_test_report_script.py` did when uploading the report, then calls `gha_update_pr_comment(...)` to post/update a sticky comment. The comment body reuses the exact `commit_range_summary` text from the `generate` step, passed through an environment variable rather than direct `${{ }}` interpolation (backticks in the summary text would otherwise be re-parsed as shell command substitution).

### Workflow changes

- `manifest-diff.yml`: added `pull-requests: write` to the `generate-report` job's permissions — required because `gha_update_pr_comment()` calls the GitHub API to create/update a PR comment, which the default `GITHUB_TOKEN` can't do without this scope. The "Generate manifest diff report" step now explicitly passes the `generate` subcommand. A new step after "Upload Report to S3" calls the same script's `post_comment` subcommand, gated to bump-PRs only and only after a successful report generation:
  ```yaml
  if: >-
    ${{
      steps.gen-report.conclusion == 'success' &&
      github.event_name == 'pull_request' &&
      startsWith(github.event.pull_request.head.ref, 'bump-')
    }}
  continue-on-error: true  # a flaky comment-post shouldn't fail the job
  env:
    COMMIT_RANGE_SUMMARY: ${{ steps.gen-report.outputs.commit_range_summary }}
  ```
- `multi_arch_ci.yml`: added the same `pull-requests: write` permission to the `manifest_diff` job, since job-level permissions on the caller cap what the called reusable workflow can request — without it here too, the token `manifest-diff.yml` receives would still lack the scope regardless of what it requests itself.

## Test Plan

- **Unit tests** cover the new `post_comment` subcommand's logic — S3 URL computation, comment body assembly, and subcommand dispatch:
  ```bash
  $ python3 -m pytest build_tools/tests/generate_manifest_diff_report_test.py -q
  23 passed in 0.08s
  ```
- **Live validation:** ran the `post_comment` step against real CI on this PR's branch and confirmed the sticky comment posts correctly with a working report link and correctly-formatted commit range. https://github.com/ROCm/TheRock/pull/6754#issuecomment-5247676679

## Test Result

The live validation is directly visible on this PR: the sticky comment posted by CI confirms the report link and commit-range text render correctly in production.

## Submission Checklist


docs/development/manifest_diff.md
```diff
@@ -17,14 +17,21 @@ flowchart TD
...
+    C -- yes --> D["generate_manifest_diff_report.py<br/>--action post_comment"]
+    D --> E[Sticky comment on the bump PR]
 ```
 
 The sibling job is wired only for those two events because they're the ones whose payloads carry the implicit refs. A manual `workflow_dispatch` of `multi_arch_ci.yml` skips it; for ad-hoc runs, dispatch [`manifest-diff.yml`](../../.github/workflows/manifest-diff.yml) directly — it exposes the full ref input surface.
 
+### Bump-PR comment
+
+Submodule-bump PRs (`bump-{submodule}-{sha}`, created by `bump_automation.py`) get the report link posted directly as a PR comment, in addition to the step-summary link every run gets — bump-PR reviewers rarely open the CI job to see the step summary. This is the last step in `manifest-diff.yml`'s `generate-report` job: `generate_manifest_diff_report.py --action post_comment` recomputes the report's S3 URL (same formula `upload_test_report_script.py` used) and posts/updates a sticky comment (marker `<!-- therock-report-manifest-diff -->`) via `gha_update_pr_comment()`. It runs with `continue-on-error: true`, so a transient GitHub API failure here never fails the job.
```
ScottTodd: > ## Test Plan
> 
> ```shell
> python3 -m pytest \
>   build_tools/tests/generate_manifest_diff_report_test.py \
>   build_tools/github_actions/tests/github_actions_api_test.py
> ```
> 
> * New test classes in `generate_manifest_diff_report_test.py` (mirroring `bump_automation_test.py`'s structure): `BuildCommitRangeSummaryTest`, `GenerateStepSummaryReusesCommitRangeSummaryTest`, `BuildPrCommentBodyTest`, `HandlePostCommentTest`, `MainDispatchTest` (covers `--action post_comment` routing/validation and the `generate` default).
> 
> * 99 tests pass in the touched files (+ the existing 47 in `github_actions_api_test.py`).
> 
> ## Test Result
> 
> All targeted tests pass locally.

This test plan is insufficient. Demonstrate how this actually operates on a PR (that can be either here in this repository or in your fork).

ScottTodd: > * Two pre-existing, unrelated issues were investigated and ruled out as untouched by this PR: a hang in `configure_multi_arch_ci_test.py`, and a missing `pyzstd` dependency failing one case in `post_stage_upload_test.py`.

This is not worth calling out in the pull request description. Write the PR description as you would a commit message, because it will be the commit message when merged.

In general please pay more attention to PR descriptions, they should be focused on details that actually matter to reviewers. I would strongly recommend writing the description yourself and studying other good PR descriptions. See https://llvm.org/docs/AIToolPolicy.html#details for example:
> To ensure sufficient self review and understanding of the work, it is strongly recommended that contributors write PR descriptions themselves (if needed, using tools for translation or copy-editing). The description should explain the motivation, implementation approach, expected impact, and any open questions or uncertainties to the same extent as a contribution made without tool assistance.

build_tools/generate_manifest_diff_report.py
```diff
@@ -1482,6 +1607,23 @@ def generate_step_summary(diff: ManifestDiff) -> None:
...
+                "$GITHUB_RUN_ID)",
+                file=sys.stderr,
+            )
+            return 1
+        if not args.pr_number:
+            print(
+                "--pr-number is required for --action post_comment",
+                file=sys.stderr,
+            )
+            return 1
```
ScottTodd: Use `parser.error()`, not `print()` and `return 1`.

* https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#use-argparse-for-cli-flags
* https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error

build_tools/generate_manifest_diff_report.py
```diff
@@ -233,6 +263,41 @@ def parse_args(argv: list[str] | None) -> argparse.Namespace:
...
+
+    # --- --action post_comment only ---
+    parser.add_argument(
+        "--run-id",
+        default=os.environ.get("GITHUB_RUN_ID"),
+        help=(
+            "[post_comment] GitHub Actions run ID whose report was uploaded "
+            "(default: $GITHUB_RUN_ID)."
+        ),
+    )
```
ScottTodd: Use either https://docs.python.org/3/library/argparse.html#subcommands or https://docs.python.org/3/library/argparse.html#argument-groups, not a comment like "post_comment only" or "[post_comment]" in the help text.

build_tools/generate_manifest_diff_report.py
```diff
@@ -233,6 +263,41 @@ def parse_args(argv: list[str] | None) -> argparse.Namespace:
...
+    parser.add_argument(
+        "--github-repository",
+        default="ROCm/TheRock",
+        help="[post_comment] Repository in 'owner/repo' format.",
+    )
+    parser.add_argument(
+        "--platform",
+        default=platform.system().lower(),
+        help="[post_comment] Platform for workflow output paths.",
+    )
```
ScottTodd: Why does this need `--platform` at all?

build_tools/generate_manifest_diff_report.py
```diff
@@ -3,7 +3,17 @@
...
-Arguments:
+Actions (--action):
+  generate               (default) Build the report + step summary. Uses the
+                         arguments below.
+  post_comment           Post/update the report link as a comment on a
+                         submodule-bump PR. Runs as a separate step, after
+                         --action generate's report has been uploaded to S3
+                         (see manifest-diff.yml) -- uses --run-id/--pr-number/
+                         --commit-range-summary/--github-repository/--platform
+                         instead of the generate-only arguments below.
```
ScottTodd: Per my other comment, should probably use a subcommand instead of `--action`: https://docs.python.org/3/library/argparse.html#subcommands

build_tools/tests/generate_manifest_diff_report_test.py
```diff
@@ -367,5 +384,173 @@ def test_non_superrepo_html_includes_component_row_and_data_component(self):
...
+
+    def test_post_comment_action_requires_pr_number(self):
+        result = main(["--action", "post_comment", "--run-id", "123"])
+
+        self.assertEqual(result, 1)
+
+    def test_default_action_is_generate(self):
+        args = parse_args(["--end", "abc123"])
+
+        self.assertEqual(args.action, "generate")
```
ScottTodd: These tests are useless - they only exercise argparse and not the meaningful logic in the script.

Please write tests by hand or review generated test cases more carefully. This is adding lots of noise to reviews.

build_tools/tests/generate_manifest_diff_report_test.py
```diff
@@ -367,5 +384,173 @@ def test_non_superrepo_html_includes_component_row_and_data_component(self):
...
+        self.assertIn("2 submodules changed", body)
+        # Marker must appear so gha_update_pr_comment can find/update this comment.
+        self.assertTrue(body.startswith(PR_COMMENT_MARKER))
+
+    def test_omits_summary_line_when_blank(self):
+        body = build_pr_comment_body("https://example.com/index.html", "")
+
+        self.assertIn(PR_COMMENT_MARKER, body)
+        self.assertIn("https://example.com/index.html", body)
+        self.assertNotIn("Commit Range", body)
```
ScottTodd: 30 lines of tests for 7 lines of logic? You could move more of these implementation details to private/internal functions and only test the higher level API contract on the script entry points. It's really not worth this level of exhaustive testing for individual functions that do basic string manipulation as then any changes to the implementation also need to update all these tests too.

docs/development/manifest_diff.md
amd-hsivasun: Fixed test plan, added a test run on this branch with if: always(), then reverting back to original conditions.

build_tools/generate_manifest_diff_report.py
amd-hsivasun: Removed, and replaced with a constant as this job only runs on linux.

build_tools/generate_manifest_diff_report.py
amd-hsivasun: Switched to use subcommands instead

build_tools/tests/generate_manifest_diff_report_test.py
amd-hsivasun: Removed the tests that are not providing any meaningful checks. Will be more thorough in the future.

build_tools/generate_manifest_diff_report.py
```diff
@@ -1474,6 +1530,50 @@ def generate_step_summary(diff: ManifestDiff) -> None:
...
+
+    Runs as its own step, after `generate`'s report is uploaded to S3, so it
+    recomputes the report URL from --run-id/PLATFORM (same formula
+    upload_test_report_script.py used to upload it) rather than reusing a
+    ManifestDiff from that earlier step.
+    """
+    output_root = WorkflowOutputRoot.from_workflow_run(
+        run_id=args.run_id, platform=PLATFORM
+    )
+    report_url = output_root.log_file("manifest-diff", "index.html").https_url
```
dysu-amd: Any reason to not link to `.../TheRockReport.html` directly? Just saves one click every time.

dysu-amd (approved):
Lgtm, though it would be nice to see what the PR comment looks like when there are submodule changes.

build_tools/generate_manifest_diff_report.py
```diff
@@ -1397,12 +1441,24 @@ def generate_html_report(diff: ManifestDiff, output_dir: Path | None = None) ->
     return HtmlReportGenerator(diff, output_dir).generate()
 
 
+def build_commit_range_summary(diff: ManifestDiff) -> str:
+    """One-line commit-range + changed-submodule-count summary.
+
+    Shared by generate_step_summary() and handle_post_comment() so both
+    surfaces show identical text.
+    """
+    changed_count = len(diff.changed)
```
erman-gurses: I checked the `class ManifestDiff:` and certain functions only consider sub-modules not the super-repos  like `def changed`- This can ignore the changes in the super-repos and miscalculate `changed_count`
https://github.com/ROCm/TheRock/blob/2932c7aef6d93ba60eb5ab6b0f99e490adc2877d/build_tools/generate_manifest_diff_report.py#L134-L156

build_tools/tests/generate_manifest_diff_report_test.py
```diff
@@ -315,56 +333,120 @@ def test_pr_base_ref_takes_precedence_over_find_last_run(self):
...
-        """Report tables have header row with class report-table-header-row."""
-        html = create_table(["Component", "Commits"], [])
-        self.assertIn("report-table-header-row", html)
+    def test_posts_comment_with_computed_report_url(self):
+        args = argparse.Namespace(
+            run_id="99999",
+            pr_number=1234,
+            commit_range_summary="**Commit Range:** `aaa` -> `bbb` (1 submodule changed)",
+            github_repository="ROCm/TheRock",
+        )
```
erman-gurses, line 472: Can you also assert exact commit-range summary passed into handle_post_comment()?

erman-gurses (commented):
Please see my comments:

build_tools/generate_manifest_diff_report.py
amd-hsivasun: Thanks for pointing this out, just pushed a fix

build_tools/generate_manifest_diff_report.py
amd-hsivasun: Changed this. Also added some submodule changes.

build_tools/tests/generate_manifest_diff_report_test.py
amd-hsivasun: Done

erman-gurses (approved):
LGTM

.github/workflows/multi_arch_ci.yml
```diff
@@ -164,6 +164,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      pull-requests: write
```
ScottTodd, line 167: Let's be careful about expanding permissions here. We could limit the `pull-requests: write` permission to a more narrowly scoped workflow. Also note that PRs from forks will get a downgraded GITHUB_TOKEN with only read access per https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories, so the `post_comment` job will likely fail then `continue-on-error` for such runs.

cc @marbre @amd-shiraz @haribabug who may have opinions on this too

.github/workflows/manifest-diff.yml
```diff
@@ -123,3 +125,21 @@ jobs:
             --subfolder manifest-diff \
             --report-path reports \
             --index-file-name index.html
+
+      - name: Post manifest-diff report link to bump PR
```
ScottTodd, line 129: The example https://therock-ci-artifacts.s3.amazonaws.com/31648304381-linux/logs/manifest-diff/TheRockReport.html report does looks useful on PRs, though we could instead teach the bump automation itself to embed the relevant parts of the diff in bump PR descriptions themselves rather than post comments? That would be even more visible.

The most relevant section of the report is the per-component list of commits

Image

That could go in a PR description as e.g.

Component | Commits
-- | --
amdsmi | <ul><li>https://github.com/ROCm/rocm-systems/commit/586fe6d016e0ac4a699191a0a7aca6b2d91cb6f0 "test(amdsmi): packaging install-path and multi-Python-version coverage"</li><li>https://github.com/ROCm/rocm-systems/commit/b827f62b13ced53cdaabcce8b0d627c8fd8c657b "doc(amdsmi): [AILITOOLS-109] Python unit test readme cleanup"</ul>
aqlprofile | _Component has no commits in this range_
clr | ...

This is generally reinventing things that dependabot has already solved, see PRs like https://github.com/ROCm/TheRock/pull/7208 for an example of what it does: 

Image

We could also check the source code for dependabot (https://github.com/dependabot/dependabot-core) and lift whatever is used to generate that.

ScottTodd (commented):
Thanks for your patience here.

Looking at this with fresh eyes, I'm not convinced that the security change and complexity is worth the benefit when we could instead make the PR descriptions include more information directly.

.github/workflows/manifest-diff.yml
ScottTodd: We could also have the bump automation generate a diff report itself and include that in the description instead of inlining the text, if the rendered HTML format is useful enough compared to what github supports directly.

amd-hsivasun:
This PR was not meant to be merged, for some reason Scott's comments did not show up for me until after I had merged the PR. I will create a revert PR.

.github/workflows/manifest-diff.yml
amd-hsivasun: These are great points, I will shift to using something like dependabot or the approach you mentioned. Thanks for the feedback.

Issue #6752, opened 2026-07-21
Several unit tests for CI-related scripts are too slow (10+ seconds)


ScottTodd:
The "unit tests" job is now taking upwards of 7 minutes, e.g. https://github.com/ROCm/TheRock/actions/runs/29864059313

history on https://github.com/ROCm/TheRock/actions/workflows/unit_tests.yml?query=branch%3Amain shows it taking 3 minutes prior to https://github.com/ROCm/TheRock/commit/ee7bf1050a75123bc9f4d70f0d05c23daddd756c

```
pytest build_tools/tests/  --durations=0
================================================== slowest durations ==================================================
12.72s call     tests/configure_stage_test.py::ManifestValidationTest::test_artifact_subprojects_matches_cmake
6.87s call     tests/artifact_descriptor_overlap_test.py::ArtifactDescriptorOverlapTest::test_no_duplicate_basedirs_across_descriptors
2.48s call     tests/artifact_backend_test.py::TestS3BackendCredentials::test_list_objects_unsigned
2.00s call     tests/setup_git_mirrors_test.py::CreateMirrorTest::test_all_retries_exhausted
1.17s call     tests/memory_monitor_test.py::test_monitor_start_stop
1.06s call     tests/os_util_test.py::RmtreeWithRetryTest::test_retries_on_locked_file
1.03s call     tests/memory_monitor_test.py::test_collect_stats
1.02s call     tests/find_latest_artifacts_test.py::FindLatestArtifactsTest::test_skips_commits_missing_artifacts
0.93s call     tests/find_latest_artifacts_test.py::FindLatestArtifactsMultiGroupTest::test_returns_none_when_no_commit_has_all_groups
0.89s call     tests/find_latest_artifacts_test.py::FindLatestArtifactsTest::test_returns_none_when_all_artifacts_missing
0.82s call     tests/find_latest_artifacts_test.py::FindLatestArtifactsMultiGroupTest::test_skips_commit_with_partial_groups
0.74s call     tests/memory_monitor_test.py::test_responsive_shutdown
0.55s call     tests/os_util_test.py::RmtreeWithRetryTest::test_gives_up_after_max_attempts
0.53s call     tests/find_artifacts_for_commit_test.py::FindArtifactsForCommitMultiGroupTest::test_multiple_groups_preserves_requested_order
0.51s call     tests/find_latest_artifacts_test.py::FindLatestArtifactsTest::test_returns_first_commit_with_artifacts
[...]

PR #6760, opened 2026-07-22
feat(ci) Add safe automatic stage reuse from exact diff bases

erman-gurses:
## Summary

Issue: #3399  
Progress on: #6752

This PR adds safe automatic CI stage reuse using an exact, verified diff-base baseline.

It is the first part of the split from original #6760 and contains only the behavior-changing stage-reuse safety and execution work. Report-only test impact, external source-provenance reporting, and workflow timing telemetry are intentionally excluded.

Automatic stage reuse now requires the selected baseline workflow run to:

- Have a `head_sha` that exactly matches the resolved diff-base commit.
- Be within the configured baseline recency window.
- Have healthy required build jobs.
- Contain all required artifacts for the selected platforms and GPU families.

If any requirement cannot be verified, CI rebuilds conservatively.

### Exact diff-base baseline validation

Stage impact is calculated over:

```text
diff_base_commit -> diff_head_commit
```

Automatic reuse only considers a baseline run when:

```text
candidate_run.head_sha == diff_base_commit
```

This ensures that reused artifacts were produced from the same commit used as the starting point for changed-file impact analysis.

The matching baseline must also:

- Be within the configured recency window.
- Have healthy build jobs.
- Contain all required artifacts for the selected platforms and GPU families.

If any requirement cannot be verified, CI rebuilds conservatively.

### Automatic stage-reuse decision flow

```text
+-----------+     +------------------+     +----------------------------+
| Diff base | --> | Calculate impact | --> | Exact, healthy baseline    |
+-----------+     +------------------+     | with required artifacts?   |
                                           +-------------+--------------+
                                                         |
                                              +----------+----------+
                                              |                     |
                                             YES                    NO
                                              |                     |
                                              v                     v
                                    +-------------------+   +------------------------+
                                    | Reuse unaffected  |   | Rebuild conservatively |
                                    | verified stages   |   +------------------------+
                                    +-------------------+
```

Older ancestors are no longer accepted because doing so could omit changes between the candidate commit and the actual diff base.

The previous branch-history and ancestry lookup has been removed, including:

- `stage_reuse_commit_history`
- `STAGE_REUSE_COMMIT_HISTORY`
- `STAGE_REUSE_CURRENT_SHA`
- Recent-commit history queries used for ancestry validation

### `workflow_dispatch` automatic reuse

A new `stage_reuse_diff_base` input allows `workflow_dispatch` runs to use automatic stage reuse safely.

The same input is used for both:

```text
changed_files = diff(stage_reuse_diff_base, HEAD)
expected_baseline_sha = resolve(stage_reuse_diff_base)
```

This prevents stage impact from being calculated against one commit while artifacts are reused from another.

Automatic dispatch reuse remains disabled when `stage_reuse_diff_base` is empty.

The checkout now fetches full history so an explicitly selected diff-base commit can be resolved even when it is older than the immediate parent.

### Platform- and family-aware artifact validation

Reusable-stage artifacts are validated independently for each selected platform.

- Target-neutral artifacts require only the `generic` archive.
- Target-specific artifacts require archives for every selected GPU artifact family.
- Linux and Windows family requirements are evaluated separately.
- Artifacts restricted to or disabled on a platform are not required for other platforms.
- Conditional platform disables remain conservatively required when resolved build flags are unavailable.

A stage is reused only when every required artifact is available for every selected platform.

### Manual and automatic reuse precedence

Explicit manual reuse inputs continue to take precedence over automatic reuse.

When `prebuilt_stages` is provided:

- Only the manually selected stages are marked prebuilt.
- The manually supplied `baseline_run_id` is preserved.
- Automatically selected reusable stages are reported but not applied.

### Cross-repository behavior

Automatic cross-repository baseline selection remains disabled.

External repositories must continue to provide explicit manual inputs such as:

- `baseline_repository`
- `baseline_run_id`
- `prebuilt_stages`

The explicit cross-repository configuration is preserved without combining it with automatic stage-reuse decisions from the triggering repository.

### Conservative fallback behavior

Automatic reuse is disabled and CI rebuilds when:

- No changed-file range is available.
- The exact diff-base SHA is unavailable.
- No baseline run exactly matches the diff base.
- The matching run is stale or unhealthy.
- Required artifacts are missing.
- Linux and Windows cannot use a consistent verified baseline.
- Automatic cross-repository reuse would otherwise be required.

## Stage-reuse summary

The stage-reuse step summary now reports:

- Whether the conservative full-CI fallback was triggered.
- The effective build scope.
- Stages that must be rebuilt.
- Unaffected reuse candidates.
- The baseline run checked.
- Artifact availability per platform.
- Stages actually marked for reuse.
- Reasons for conservative fallback or non-application.

## Tests

Regression coverage includes:

- Exact baseline SHA acceptance.
- Rejection of non-matching and missing SHAs.
- Stale and unhealthy baseline rejection.
- Missing-artifact fallback.
- Target-neutral and target-specific artifact requirements.
- Platform-specific and platform-disabled artifacts.
- Independent Linux and Windows family validation.
- Manual reuse precedence.
- Automatic reuse application.
- Automatic cross-repository reuse suppression.
- `workflow_dispatch` diff-base parsing.
- Dispatches without an explicit diff base.
- Propagation of the resolved diff-base SHA into baseline selection.

Unit tests are passed.

TODO PRs
PR 2 — CI impact reporting: Report-only stage-to-test impact, test inventory, CI summaries, and external source/ref reporting. 
PR 3 — Workflow timing telemetry: Actions API timing collection, runner/job classification, and Markdown/JSON timing artifacts.

### DIFF
--- a/.github/workflows/multi_arch_ci.yml
+++ b/.github/workflows/multi_arch_ci.yml
@@ -42,6 +42,13 @@ on:
         type: string
         default: ""
         description: "Workflow run ID to copy prebuilt stage artifacts from; required when prebuilt_stages is set"
+      stage_reuse_diff_base:
+        type: string
+        default: ""
+        description: >-
+          Git commit or ref used as both the changed-file diff base and the
+          exact automatic stage-reuse baseline. Required for automatic
+          reuse-stage on workflow_dispatch; empty disables automatic reuse.
       stage_reuse_mode:
         type: choice
         default: "dry-run"
@@ -57,10 +64,6 @@ on:
         type: string
         default: "72"
         description: "Automatic stage reuse: reject baseline runs older than this many hours"
-      stage_reuse_commit_history:
-        type: string
-        default: "50"
-        description: "Automatic stage reuse: number of recent branch commits to fetch for ancestry"
       changed_projects:
         type: string
         default: ""
@@ -105,9 +108,9 @@ jobs:
       windows_test_labels: ${{ inputs.windows_test_labels || '' }}
       prebuilt_stages: ${{ inputs.prebuilt_stages || '' }}
       baseline_run_id: ${{ inputs.baseline_run_id || '' }}
+      stage_reuse_diff_base: ${{ inputs.stage_reuse_diff_base || '' }}
       stage_reuse_mode: ${{ inputs.stage_reuse_mode || 'dry-run' }}
       stage_reuse_max_age_hours: ${{ inputs.stage_reuse_max_age_hours || '72' }}
-      stage_reuse_commit_history: ${{ inputs.stage_reuse_commit_history || '50' }}
       build_pytorch: ${{ github.event_name != 'workflow_dispatch' || inputs.build_pytorch }}
       build_jax: ${{ github.event_name != 'workflow_dispatch' || inputs.build_jax }}
 
--- a/.github/workflows/setup_multi_arch.yml
+++ b/.github/workflows/setup_multi_arch.yml
@@ -50,16 +50,18 @@ on:
           be reused; "reuse-stage" actually reuses them.
         type: string
         default: "dry-run"
+      stage_reuse_diff_base:
+        description: >-
+          Git commit or ref used as both the changed-file diff base and the
+          exact automatic stage-reuse baseline. Intended for same-repository
+          workflow_dispatch runs; empty disables automatic dispatch reuse.
+        type: string
+        default: ""
       stage_reuse_max_age_hours:
         description: >-
          Automatic stage reuse: reject baseline runs older than this many hours. default: "72"
         type: string
         default: "72"
-      stage_reuse_commit_history:
-        description: >-
-          Automatic stage reuse: number of recent branch commits to fetch for ancestry. default: "50"
-        type: string
-        default: "50"
       build_pytorch:
         description: "Build PyTorch wheels."
         type: boolean
@@ -147,8 +149,9 @@ jobs:
         with:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref }}
-          # We need the parent commit to do a diff
-          fetch-depth: 2
+          # Fetch full history so diff-base commits can be resolved for path filtering
+          # and exact automatic stage-reuse validation.
+          fetch-depth: 0
 
       - name: Checkout CI config
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -176,10 +179,9 @@ jobs:
           WINDOWS_TEST_LABELS: ${{ inputs.windows_test_labels }}
           PREBUILT_STAGES: ${{ inputs.prebuilt_stages }}
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
+          STAGE_REUSE_DIFF_BASE: ${{ inputs.stage_reuse_diff_base }}
           STAGE_REUSE_MODE: ${{ inputs.stage_reuse_mode }}
-          STAGE_REUSE_CURRENT_SHA: ${{ steps.checkout.outputs.commit }}
           STAGE_REUSE_MAX_AGE_HOURS: ${{ inputs.stage_reuse_max_age_hours }}
-          STAGE_REUSE_COMMIT_HISTORY: ${{ inputs.stage_reuse_commit_history }}
           # For cross-repo artifact reuse, use repository input (defaults to GITHUB_REPOSITORY).
           # External repos set repository to "ROCm/TheRock" to reuse TheRock's prebuilt artifacts.
           THEROCK_REPOSITORY: ${{ inputs.repository }}
--- a/build_tools/_therock_utils/build_topology.py
+++ b/build_tools/_therock_utils/build_topology.py
@@ -129,6 +129,25 @@ class Artifact:
     )  # Artifacts needed for testing this artifact (e.g., ["core-hiptests"])
 
 
+def artifact_may_apply_to_platform(
+    artifact: Artifact,
+    platform_name: str,
+) -> bool:
+    """Return whether an artifact may apply to a platform.
+
+    Explicit platform restrictions are honored. Conditional disables are not
+    evaluated because callers do not have the resolved build flags, so the
+    artifact remains applicable conservatively.
+    """
+    if artifact.platform is not None and artifact.platform != platform_name:
+        return False
+
+    if platform_name in artifact.disable_platforms:
+        return False
+
+    return True
+
+
 class BuildTopology:
     """
     Parses and provides operations on BUILD_TOPOLOGY.toml.
--- a/build_tools/github_actions/baseline_runs.py
+++ b/build_tools/github_actions/baseline_runs.py
@@ -80,20 +80,20 @@ class WorkflowJobHealth:
 
 @dataclass(frozen=True)
 class CommitCompatibility:
-    """Result of checking a candidate run's commit against the current commit.
+    """Result of comparing a candidate run with the exact expected baseline.
 
-    A baseline run is only safe to reuse when it was built from source that the
-    current commit is based on. Reusing artifacts from a divergent or newer-than-current
-    commit risks mixing incompatible binaries, so those are rejected.
+    A baseline run is safe to reuse only when its ``head_sha`` exactly matches
+    ``expected_head_sha``. Accepting an older or different commit could omit
+    changes covered by the current impact range and reuse stale artifacts.
     """
 
-    current_commit_sha: str
+    expected_head_sha: str
     candidate_head_sha: str
     relationship: str
 
     @property
     def is_valid(self) -> bool:
-        return self.relationship in ("same", "ancestor")
+        return self.relationship == "same"
 
 
 @dataclass(frozen=True)
@@ -587,61 +587,36 @@ def parse_run_timing(workflow_run: dict) -> RunTiming:
 def validate_commit_compatibility(
     *,
     candidate_head_sha: str,
-    current_commit_sha: str,
-    ordered_commit_shas: Sequence[str],
+    expected_head_sha: str,
 ) -> CommitCompatibility:
-    """Validate that a candidate run was built from a compatible commit.
+    """Validate that a candidate run matches the exact expected baseline.
 
-    ``ordered_commit_shas`` is the branch history newest-first (as returned by
-    ``gha_query_recent_branch_commits``). The current commit is expected to be
-    at or near the front of that list. A candidate is:
+    Stage impact is calculated over:
 
-    * ``same`` when its head_sha equals the current commit;
-    * ``ancestor`` when its head_sha appears *after* the current commit in the
-      newest-first history (i.e. the current commit is built on top of it);
-    * ``descendant_or_divergent`` when it appears *before* the current commit
-      (newer than current, so unsafe to reuse);
-    * ``unknown`` when it does not appear in the provided history window.
+        diff_base_commit -> diff_head_commit
 
-    Only ``same`` and ``ancestor`` are considered valid. ``unknown`` is treated
-    as unsafe rather than assumed-good: a candidate outside the history window
-    cannot be confirmed as an ancestor, so widen the window if legitimate
-    ancestors are being rejected.
+    Therefore, a candidate run is considered compatible only when its
+    ``head_sha`` exactly equals ``diff_base_commit``. Accepting an older ancestor
+    would omit changes between the candidate commit and ``diff_base_commit`` and
+    could reuse stale artifacts.
+
+    TODO: Investigate safe per-platform baseline selection and ancestor reuse.
+    Any ancestor-based approach must account for the complete change range from
+    the candidate commit to ``diff_head_commit``, rather than checking ancestry
+    alone.
     """
-    current = _normalize_sha(current_commit_sha)
+    expected = _normalize_sha(expected_head_sha)
     candidate = _normalize_sha(candidate_head_sha)
-    if not current:
-        raise ValueError("current_commit_sha must be non-empty")
+
+    if not expected:
+        raise ValueError("expected_head_sha must be non-empty")
     if not candidate:
         raise ValueError("candidate_head_sha must be non-empty")
 
-    if candidate == current:
-        relationship = "same"
-    else:
-        history = [_normalize_sha(sha) for sha in ordered_commit_shas]
-        try:
-            current_index = history.index(current)
-        except ValueError:
-            current_index = None
-        try:
-            candidate_index = history.index(candidate)
-        except ValueError:
-            candidate_index = None
-
-        if candidate_index is None:
-            relationship = "unknown"
-        elif current_index is None:
-            # Current commit is outside the window but the candidate is in it;
-            # we cannot establish ordering, so treat as unknown.
-            relationship = "unknown"
-        elif candidate_index > current_index:
-            # Newest-first: a larger index is older than the current commit.
-            relationship = "ancestor"
-        else:
-            relationship = "descendant_or_divergent"
+    relationship = "same" if candidate == expected else "different"
 
     return CommitCompatibility(
-        current_commit_sha=current,
+        expected_head_sha=expected,
         candidate_head_sha=candidate,
         relationship=relationship,
     )
@@ -685,8 +660,7 @@ def select_baseline_run(
     max_runs: int = 20,
     exclude_run_ids: Iterable[str] = (),
     required_successful_job_name_substrings: Iterable[str] = ("Build",),
-    current_commit_sha: str | None = None,
-    ordered_commit_shas: Sequence[str] | None = None,
+    expected_head_sha: str | None = None,
     max_age_hours: float | None = None,
     now: datetime | None = None,
     workflow_runs: Sequence[dict] | None = None,
@@ -709,13 +683,9 @@ def select_baseline_run(
             match at least one completed successful job in the candidate run.
             Defaults to ``("Build",)`` so unrelated test failures do not
             automatically disqualify otherwise reusable build artifacts.
-        current_commit_sha: When provided, enables the commit-compatibility
-            rule: a candidate run is only accepted when its head_sha equals or
-            is an ancestor of this commit. Requires ``ordered_commit_shas``.
-        ordered_commit_shas: Branch history newest-first (e.g. from
-            ``gha_query_recent_branch_commits``) used to establish ancestry for
-            the commit-compatibility rule. Ignored unless ``current_commit_sha``
-            is set.
+        expected_head_sha: When provided, a candidate run is accepted only
+            when its head_sha exactly equals this SHA. Automatic stage reuse
+            passes the resolved diff-base commit here.
         max_age_hours: When provided, enables the recency rule: candidate runs
             older than this many hours (by ``created_at``) are rejected.
         now: Reference time for the recency rule; defaults to the current UTC
@@ -741,16 +711,11 @@ def select_baseline_run(
     )
     excluded = {str(run_id) for run_id in exclude_run_ids}
 
-    check_commit = current_commit_sha is not None
-    if check_commit:
-        if not current_commit_sha.strip():
-            raise ValueError(
-                "current_commit_sha must be non-empty when commit checking is enabled"
-            )
-        if ordered_commit_shas is None:
-            raise ValueError(
-                "ordered_commit_shas is required when current_commit_sha is set"
-            )
+    check_commit = expected_head_sha is not None
+    if check_commit and not expected_head_sha.strip():
+        raise ValueError(
+            "expected_head_sha must be non-empty when commit checking is enabled"
+        )
     check_recency = max_age_hours is not None
 
     candidates = (
@@ -794,21 +759,24 @@ def select_baseline_run(
         if check_commit:
             candidate_head_sha = (workflow_run.get("head_sha", "") or "").strip()
             if not candidate_head_sha:
-                # A run without a head_sha cannot be confirmed compatible.
-                logger.info("Skipping run %s: no head_sha to compare", run_id)
+                logger.info(
+                    "Skipping run %s: no head_sha to compare with expected baseline",
+                    run_id,
+                )
                 continue
+
             commit_compatibility = validate_commit_compatibility(
                 candidate_head_sha=candidate_head_sha,
-                current_commit_sha=current_commit_sha,
-                ordered_commit_shas=ordered_commit_shas or (),
+                expected_head_sha=expected_head_sha,
             )
+
             if not commit_compatibility.is_valid:
                 logger.info(
-                    "Skipping run %s: commit %s is %s relative to current %s",
+                    "Skipping run %s: candidate commit %s does not exactly match "
+                    "expected baseline commit %s",
                     run_id,
                     candidate_head_sha,
-                    commit_compatibility.relationship,
-                    commit_compatibility.current_commit_sha,
+                    commit_compatibility.expected_head_sha,
                 )
                 continue
 
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -203,7 +203,7 @@ class CIInputs:
         python_version = os.environ.get("PYTHON_VERSION", "").strip()
 
         pr_labels: list[str] = []
-        base_ref: str | None = "HEAD^1"
+        base_ref: str | None = None
         if event_name == "pull_request":
             # Extract label name strings from the event payload's label objects:
             #   Sample input:  [{"name": "ci:skip", "color": "fff", ...}, ...]
@@ -225,6 +225,21 @@ class CIInputs:
                 # diff against. HEAD^1 would only diff the final pushed commit
                 # against its parent, missing earlier commits in the push.
                 base_ref = None
+        elif event_name == "workflow_dispatch":
+            # This one ref defines both:
+            #
+            #   changed_files = diff(base_ref, HEAD)
+            #   expected_baseline_sha = resolve(base_ref)
+            #
+            # Keeping those values tied together prevents artifacts from one
+            # commit being reused for an independently calculated impact range.
+            base_ref = (
+                os.environ.get(
+                    "STAGE_REUSE_DIFF_BASE",
+                    "",
+                ).strip()
+                or None
+            )
 
         # Test labels come from two sources:
         # 1. LINUX/WINDOWS_TEST_LABELS env vars (workflow_dispatch inputs)
@@ -302,10 +317,11 @@ class GitContext:
 
     @staticmethod
     def empty() -> "GitContext":
-        """Empty context with no git data.
+        """Return a context without a validated change range.
 
-        This should typically be used for schedule/workflow_dispatch events
-        where we don't want to diff against a prior commit.
+        Used for schedules, dispatches without stage_reuse_diff_base, external
+        repository calls, and events where no reliable diff base is available.
+        Automatic stage reuse must remain disabled for this context.
         """
         return GitContext()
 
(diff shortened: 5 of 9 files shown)


.github/workflows/setup_multi_arch.yml
```diff
@@ -211,6 +212,56 @@ jobs:
...
+          THEROCK_REPOSITORY: ${{ inputs.repository || github.repository }}
+          THEROCK_REQUESTED_REF: ${{ inputs.ref }}
+          THEROCK_RESOLVED_SHA: ${{ steps.checkout.outputs.commit }}
+          CALLING_WORKFLOW_REF: ${{ github.workflow_ref }}
+          REUSABLE_WORKFLOW_PATH: .github/workflows/setup_multi_arch.yml
+        run: python ./build_tools/github_actions/super_repo_signal.py
+
+      # Temporary smoke test for PR #6760. Remove after verification.
+      - name: Test super-repo source/ref summary
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.number == 6760 }}
```
ammallya: Please remove this before merge

build_tools/github_actions/stage_to_test_impact.py
```diff
@@ -0,0 +1,314 @@
...
+        if normalized.startswith("test:"):
+            normalized = normalized.removeprefix("test:")
+
+        if normalized:
+            components.append(normalized)
+
+    return tuple(dict.fromkeys(components))
+
+
+def _artifact_applies_to_platform(
```
ammallya: This function is used and defined exactly the same in test impact and reuse_decision maybe you could import it and define it just in one (But not a blocker)

build_tools/github_actions/workflow_summary.py
```diff
@@ -244,8 +247,44 @@ def main(argv: list[str]) -> int:
...
+
+            print(
+                f"Wrote {timing_json_path} with "
+                f"{len(collected_records)} timing record(s)."
+            )
+            print(
+                f"Wrote {timing_markdown_path} with "
+                f"{len(collected_records)} timing record(s)."
+            )
+        except Exception as exc:
```
ammallya: I think this exception would cause more problems and I think we want to fail the workflow  as per Scotts comments on my PR even if the summary step fails.

build_tools/github_actions/stage_to_test_impact.py
erman-gurses: Thanks -  I moved the shared conservative artifact/platform applicability check into _therock_utils.build_topology and updated both stage-reuse and test-impact analysis to use it. The existing platform and conditional-disable tests continue to pass.

.github/workflows/multi_arch_ci.yml
```diff
@@ -172,11 +167,24 @@ jobs:
...
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - name: Evaluate workflow results
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
           python build_tools/github_actions/workflow_summary.py \
-            --needs-json '${{ toJSON(needs) }}'
+            --needs-json '${{ toJSON(needs) }}' \
+            --collect-timing
+      - name: Upload workflow timing
```
geomin12: nit: new space between

geomin12: hm, i know we can pull this via HUD / external sources. should we collect via that?

.github/workflows/setup_multi_arch.yml
```diff
@@ -211,6 +205,36 @@ jobs:
...
+          EXTERNAL_REPO_CONFIG_JSON: ${{ steps.external_repo_config.outputs.config_json }}
+          SUPER_REPO_SOURCE_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+          THEROCK_REPOSITORY: ${{ inputs.repository || github.repository }}
+          THEROCK_REQUESTED_REF: ${{ inputs.ref }}
+          THEROCK_RESOLVED_SHA: ${{ steps.checkout.outputs.commit }}
+          CALLING_WORKFLOW_REF: ${{ github.workflow_ref }}
+          REUSABLE_WORKFLOW_PATH: .github/workflows/setup_multi_arch.yml
+        run: python ./build_tools/github_actions/super_repo_signal.py
+
+  ci_impact_report:
```
geomin12: we write to github summary. should we do the updates there? instead of in console?

build_tools/github_actions/super_repo_signal.py
```diff
@@ -0,0 +1,183 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Render the super-repo source and TheRock checkout used by CI."""
```
geomin12: maybe an example on top of this script pls?

build_tools/github_actions/stage_to_test_impact.py
geomin12, line 1: oh this is great reporting, but could be very useful for actual test selection and put into CI plumbing

build_tools/_therock_utils/build_topology.py
```diff
@@ -129,6 +129,25 @@ class Artifact:
     )  # Artifacts needed for testing this artifact (e.g., ["core-hiptests"])
 
 
+def artifact_may_apply_to_platform(
```
geomin12, line 132: isn't it pretty deterministic if the artifact applies to platform?

build_tools/github_actions/baseline_runs.py
```diff
@@ -80,20 +80,15 @@ def is_valid(self) -> bool:
...
-
-    current_commit_sha: str
+    expected_head_sha: str
     candidate_head_sha: str
     relationship: str
 
     @property
     def is_valid(self) -> bool:
-        return self.relationship in ("same", "ancestor")
+        return self.relationship == "same"
```
geomin12, line 96: oh... are we supposed to remove ancestor?

build_tools/github_actions/baseline_runs.py
```diff
@@ -586,61 +581,22 @@ def parse_run_timing(workflow_run: dict) -> RunTiming:
...
-    * ``descendant_or_divergent`` when it appears *before* the current commit
-      (newer than current, so unsafe to reuse);
-    * ``unknown`` when it does not appear in the provided history window.
-
-    Only ``same`` and ``ancestor`` are considered valid. ``unknown`` is treated
-    as unsafe rather than assumed-good: a candidate outside the history window
-    cannot be confirmed as an ancestor, so widen the window if legitimate
-    ancestors are being rejected.
-    """
-    current = _normalize_sha(current_commit_sha)
```
geomin12, line 611: oh this is good commenting. let's add back

build_tools/github_actions/baseline_runs.py
```diff
@@ -80,20 +80,15 @@ def is_valid(self) -> bool:
 
 @dataclass(frozen=True)
 class CommitCompatibility:
-    """Result of checking a candidate run's commit against the current commit.
+    """Result of comparing a candidate run with the exact expected baseline."""
 
-    A baseline run is only safe to reuse when it was built from source that the
```
geomin12, line 85: oh this is good commenting. let's add back

build_tools/github_actions/workflow_timing.py
geomin12, line 1: i feel this data pulling should be separate and polled. CI should only handle build + test, while we shouldn't do data collection (for submodule bumps) in here

geomin12 (commented):
first pass comments!

I feel for data collection, it should be done externally as an infra perspective. CI should handle build + test + relevant data for developer

build_tools/github_actions/baseline_runs.py
erman-gurses: Yes, I removed `ancestor` purposely. 

Stage impact is calculated over:

    `diff_base_commit -> diff_head_commit`

If we accept an older ancestor as the baseline, changes between that ancestor and `diff_base_commit` are outside the range analyzed by stage impact.

For example:
```
    A  older baseline
    |
    B  diff_base_commit
    |
    C  diff_head_commit
```
When impact is calculated over B -> C, a stage may appear unaffected, but its artifact from A may still be stale because that stage could have changed between A and B.

The safe invariant for the current implementation is therefore:

    `candidate_run.head_sha == GitContext.diff_base_commit`

for both pull requests and pushes. If no run exists at that exact SHA, automatic reuse is not applied and CI rebuilds normally.

Let me know if it makes sense.

build_tools/github_actions/baseline_runs.py
erman-gurses: If we agreed with the explanation above - I will update or add these comments based on that.

build_tools/_therock_utils/build_topology.py
erman-gurses: The function is deterministic, but the final applicability is not always known from `artifact` and `platform_name` alone.

We can deterministically honor `artifact.platform` and `artifact.disable_platforms`, but `disable_platforms_if_flags_not_set` depends on the resolved build flags. These stage-reuse and test-impact callers do not currently have those flags, so a conditionally enabled artifact is kept applicable conservatively.

That is why the name uses `may`: it means “not definitely excluded by the available platform metadata,” rather than “definitely built on this platform.” I can rename it to `artifact_is_potentially_applicable_to_platform` to make that distinction clearer.

build_tools/github_actions/workflow_timing.py
erman-gurses: Thanks -  I agree with you -  will address that.

.github/workflows/setup_multi_arch.yml
erman-gurses: Good point. The CI impact report is already written to `GITHUB_STEP_SUMMARY`, but currently it is written from a separate `ci_impact_report` job, so it appears as a separate job summary.

Since the impact report is computed during Setup, I will append it to the Setup job summary alongside the existing configuration and super-repo summaries, and remove the extra report job.

I will keep concise console logging for diagnostics, but the GitHub summary will be the primary human-facing report.

build_tools/github_actions/stage_to_test_impact.py
erman-gurses: Agreed - using this analysis for actual test selection is the intended next step.

I kept it report-only in this PR intentionally because the current mapping is based primarily on artifact ownership and producer stages. It does not yet model every runtime/test dependency; for example, some tests fetch multiple artifact groups such as `rocfft + rand or rocprofiler-compute + rocprofiler-sdk`. Applying the current recommendation directly could therefore skip a test when one of its secondary dependencies changed.

The workplan also calls for starting with dry-run output and enabling filtering only after the ownership and dependency mapping has been validated.

As a follow-up, I plan to add explicit per-test dependency metadata, emit platform-specific selected-component outputs, and plumb those into the existing PROJECTS_TO_TEST filtering behind an opt-in mode. Once the dry-run
recommendations are proven safe, that can drive the real test matrix.

.github/workflows/multi_arch_ci.yml
erman-gurses: Yes, HUD may be a better source for the raw job timing data. We still need queue and execution timing for the savings report, plus the rebuilt/copied/skipped/forced decision from this workflow.

build_tools/github_actions/baseline_runs.py
erman-gurses: So the mistake is in here:
```
Stage impact is calculated from B -> C,
while baseline selection may choose A,
and nothing checks changes from A -> B.
```
@ammallya - can you please verify that.

ammallya: Your Analysis is correct, Only thing is requiring head_sha == diff_base_commit will miss a lot of reuse in practice and any rebase/force-push moves the base SHA, so most PRs won't find an exact-match baseline and will just rebuild, but its okay because we'd rather be conservative and rebuild than ever reuse a stale artifact. There also would be another issue if commit A -> B has a good linux build but bad windows build both will be rebuilt, Since the new logic will deem no good ancestors, Might want to see how to get around that.

build_tools/github_actions/stage_to_test_impact.py
```diff
@@ -0,0 +1,295 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Report-only mapping from semantic stage impact to test components."""
+
+from __future__ import annotations
```
ammallya: Illegal import please change this please refer https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/super_repo_signal.py
```diff
@@ -0,0 +1,183 @@
...
+
+from __future__ import annotations
```
ammallya: Illegal import please change this please refer https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/workflow_timing.py
```diff
@@ -0,0 +1,407 @@
...
+def _duration_seconds(
+    start: datetime | None,
+    end: datetime | None,
+) -> float | None:
+    """Return a non-negative duration between two timestamps."""
+    if start is None or end is None:
+        return None
+
+    return max(
+        0.0,
```
ammallya: I think it might be better to make it uniform with baseline_runs._seconds_between so we can reuse the helper function, We can change the None in seconds_between to 0.0 if that helps this so it can be uniform

erman-gurses: Updated - now I use `seconds_between`

build_tools/github_actions/baseline_runs.py
erman-gurses: Agreed. Exact SHA matching will reduce reuse, especially after rebases or force-pushes, but correctness should take priority. We can handle per-platform baseline selection as a follow-up so a valid Linux build is not discarded because Windows failed.

build_tools/github_actions/baseline_runs.py
erman-gurses: Updated docstring

build_tools/github_actions/baseline_runs.py
erman-gurses: Added it back under CommitCompatibility and updated the wording to explain the new exact-baseline SHA requirement.

ammallya (approved):
Looks good to me

erman-gurses:
@ScottTodd , FYI, first pass reviews are done.

ScottTodd:
I'll need at least a few days to review a PR this large. Have you considered splitting into smaller, more reviewable chunks of work? Here's an automated review with some initial findings and suggestions for possible splits: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_6760.md.

Also please assign yourself to https://github.com/ROCm/TheRock/issues/3399 and post architecture plans, ideally prior to starting work, see https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#using-github-issues-for-feature-development
> When planning complex changes, please solicit feedback and announce your intent to work on a pull request early in development, as this gives other contributors time to offer advice and avoid duplicating effort.

erman-gurses:
> mated review with some initial findings and suggestions for possible splits:

I was actually hesitating to bring that large PR. I will split it - thanks for initial feedback.

erman-gurses:
> > mated review with some initial findings and suggestions for possible splits:
> 
> I was actually hesitating to bring that large PR. I will split it - thanks for initial feedback.

Split into three PRs based on the feedback:

This PR: 
PR 1 — Automatic stage reuse: Exact diff-base validation, platform/family artifact checks, conservative fallback, reuse execution, and manual/automatic precedence.

TODO PRs
PR 2 — CI impact reporting: Report-only stage-to-test impact, test inventory, CI summaries, and external source/ref reporting. PR 3 — Workflow timing telemetry: Actions API timing collection, runner/job classification, and Markdown/JSON timing artifacts.
CC: @ScottTodd

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -871,35 +1006,60 @@ def decide_jobs(
...
     # verify come straight from the resolved target selection.
 
+    all_families = get_all_families_for_trigger_types(
+        ["presubmit", "postsubmit", "nightly"]
+    )
+
+    linux_artifact_families = [
+        all_families[family_name]["linux"]["family"]
+        for family_name in targets.linux_families
+    ]
```
ScottTodd, line 1024: Why is this always using "all families"? I would expect that this should be looking for a workflow run that has _the same families as the current run_. For example, if CI jobs default to `gfx942,gfx1100` and a PR opts-in to building `gfx942,gfx1100,gfx1200`, it should _not_ reuse artifacts from prior jobs that used `gfx942,gfx1100` (or if it does, it should be particularly careful about which artifacts are reused as-is and which are rebuilt/patched)

Reusing artifacts from a run that used different targets/families can cause issues. It is *not* enough to just find artifacts that contain a superset of what the current run is planning on using:
* https://github.com/ROCm/TheRock/issues/4687

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -177,27 +232,43 @@ def _stage_artifacts_available(
...
+        found = False
+
+        for component in ARTIFACT_COMPONENTS:
+            for extension in ARTIFACT_EXTENSIONS:
+                filename = (
+                    f"{requirement.name}_"
+                    f"{component}_"
+                    f"{requirement.target_family}"
+                    f"{extension}"
+                )
```
ScottTodd, line 260: I don't think you trust this file name construction to include all possible artifact names. ASan builds produce artifacts with `:xnack+` suffixes for example:
* https://github.com/ROCm/TheRock/actions/runs/30920225093
* https://therock-ci-artifacts.s3.amazonaws.com/30920225093-linux/index.html
* [blas_lib_gfx942:xnack+.tar.zst](https://therock-ci-artifacts.s3.amazonaws.com/30920225093-linux/blas_lib_gfx942%3Axnack%2B.tar.zst)

I think the safer sequence of checks would be:
1. Did a workflow run use the same build environment (e.g. build container, compiler flags, enabled GPU targets/families) as the current run?
3. Did a workflow run use the same commit for artifacts that this run does not need to rebuild?
    * Example 1: this PR did not modify `compiler/amd-llvm` or `rocm-systems/`, so we can reuse the compiler-runtime stage
    * Example 2: this PR did not modify `rocm-systems/projects/rccl/`, so we can reuse the comm-libs stage
3. Did a workflow run succeed in building all artifacts

That saves us from checking individual requirements, components, artifact suffixes, etc.

As for how to determine "Did a workflow run succeed in building all artifacts", here are a few ideas:
* check the workflow status (relies on the github API, potentially ambiguous with test jobs included too)
* write a marker file to the artifact folder after the tests in https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_artifacts_structure.yml complete.
* write a a marker file to the artifact folder after all stages in https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_portable_linux.yml

edit: I see how `_required_artifacts_for_stages()` is listing artifacts using the topology file and `artifact_may_apply_to_platform` now but I'm not sure if we can determine which artifacts should exist from just the source code easily, see also https://github.com/ROCm/TheRock/blob/02bb54063b1e398ab07f2a1d8c4c3eb48023a2ff/build_tools/fileset_tool.py#L124-L138

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -225,6 +225,151 @@ def test_push_created_ref_disables_path_filtering(self):
...
+            },
+            extra_env={
+                "STAGE_REUSE_DIFF_BASE": "base-commit-sha",
+            },
+        )
+
+        self.assertEqual(
+            inputs.base_ref,
+            "base-commit-sha",
+        )
```
ScottTodd, line 242: nit: lots of whitespace / line breaks in these tests. If you drop the trailing `,` characters then black formatting should condense quite a bit

like this?

```suggestion
    def test_workflow_dispatch_reads_stage_reuse_diff_base(self):
        inputs = _run_from_environ(
            "workflow_dispatch", { "inputs": {} },
            extra_env={ "STAGE_REUSE_DIFF_BASE": "base-commit-sha" }
        )

        self.assertEqual(inputs.base_ref, "base-commit-sha")
```

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -225,6 +225,151 @@ def test_push_created_ref_disables_path_filtering(self):
...
+            clear=False,
+        ):
+            result = cm._load_git_context(
+                inputs,
+            )
+
+        self.assertIs(
+            result,
+            expected,
+        )
```
ScottTodd, line 298: nit: Same comment about extra whitespace/newlines due to trailing `,` characters here

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -528,6 +673,156 @@ def test_explicit_prebuilt_stages(self):
...
+
+        self.assertEqual(
+            baseline_run_id,
+            "manual-run-100",
+        )
+
+        self.assertEqual(
+            adjusted_auto_stage_reuse.applied_reuse_stages,
+            (),
+        )
```
ScottTodd, line 730: Same here - reformat these onto a single line

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -608,6 +903,262 @@ def test_asan_tests_only_run_on_nightly_triggers(self):
...
+            baseline_html_url=None,
+            available_stages=("compiler-runtime",),
+            unavailable_stages=(),
+            applied_reuse_stages=("compiler-runtime",),
+            reasons=(),
+            report_lines=(),
+            platform_available={
+                "linux": ("compiler-runtime",),
+            },
+        )
```
ScottTodd, line 1092: This is a large/complicated mock. Can a real version be used instead? Could the functions be refactored to make this style of testing easier without mocks?

build_tools/github_actions/tests/baseline_runs_test.py
```diff
@@ -519,68 +519,83 @@ def workflow_jobs_fetcher(workflow_run, github_repository):
...
-    def test_unknown_candidate_is_invalid(self):
-        result = baseline_runs.validate_commit_compatibility(
-            candidate_head_sha="sha-divergent",
-            current_commit_sha="sha-current",
-            ordered_commit_shas=self.HISTORY,
+        self.assertEqual(
+            result.relationship,
+            "same",
         )
-        self.assertEqual(result.relationship, "unknown")
```
ScottTodd, line 557: More test code using trailing `,` characters that result in lines spilling over after formatting is applied

build_tools/github_actions/tests/baseline_runs_test.py
```diff
@@ -519,68 +519,83 @@ def workflow_jobs_fetcher(workflow_run, github_repository):
...
+                )
+                self.assertEqual(
+                    result.candidate_head_sha,
+                    candidate_sha,
+                )
+                self.assertEqual(
+                    result.relationship,
+                    "different",
+                )
+                self.assertFalse(result.is_valid)
```
ScottTodd, line 567: trailing commas (please check all edits in this PR)

.github/workflows/setup_multi_arch.yml
```diff
@@ -147,8 +149,9 @@ jobs:
         with:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref }}
-          # We need the parent commit to do a diff
-          fetch-depth: 2
+          # Fetch full history so diff-base commits can be resolved for path filtering
+          # and exact automatic stage-reuse validation.
+          fetch-depth: 0
```
ScottTodd, line 154: `fetch-depth: 0` is going to be _very_ expensive in rocm-libraries/rocm-systems (upwards of 10 minutes to checkout)

I'd rather walk the git history and fetch commits as needed with code like https://github.com/ROCm/TheRock/blob/b070c6f38860cec6e28f37e1c4bb66c6002146e1/build_tools/github_actions/configure_ci_path_filters.py#L30-L63

That code is already in use with https://github.com/ROCm/TheRock/blob/b070c6f38860cec6e28f37e1c4bb66c6002146e1/build_tools/github_actions/configure_ci_path_filters.py#L83-L116 - is the "stage reuse" code using the output of that `get_git_modified_paths` function?

build_tools/github_actions/baseline_runs.py
ScottTodd, line 1: Is there a way to run this locally _for a source build_ or are you only testing on CI? Can you point me to some docs if so?

I created https://github.com/ROCm/TheRock/blob/main/build_tools/find_latest_artifacts.py and https://github.com/ROCm/TheRock/blob/main/build_tools/find_artifacts_for_commit.py to be the entry points for "get me prebuilt artifacts so I can build quickly" and I think we should lean more on local development support before going straight to "get multi-arch CI working across all repositories".

ScottTodd (commented):
I have ~60% confidence in some of the plumbing and decision making code in here. I'm concerned that directly integrating into the github actions workflows is going to be too big of a jump and we'll have trouble validating all the complex logic, edge cases, cross-repository work, asynchronous github actions runs / retries, etc.

Simpler steps:
* Extend https://github.com/ROCm/TheRock/blob/main/build_tools/find_latest_artifacts.py for more complete developer usage
* Prove out artifact reuse in https://github.com/ROCm/ROCgdb, https://github.com/ROCm/rocm-libraries, and other narrow slices _before_ tackling the fully general solution. For ROCgdb in particular that's used for the debug-tools stage (https://github.com/ROCm/TheRock/tree/main/debug-tools) and what really matters is using artifacts that correspond to `THEROCK_COMMIT_REF` here: https://github.com/ROCm/ROCgdb/blob/79fd0fd22c7dc6c542387024f27e264e42257731/.github/workflows/therock-ci-linux.yml#L16-L17

build_tools/github_actions/stage_reuse_decision.py
geomin12, line 1: i would like to be careful about "success". in a workflow, success = green at the end. however, due to flaky infra / test issues, the workflow is most likely never green! 

so we need to look at build only and determine "success" via that, so that we can re-use stages when applicable

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -225,6 +225,21 @@ def from_environ() -> "CIInputs":
...
+            #
+            # Keeping those values tied together prevents artifacts from one
+            # commit being reused for an independently calculated impact range.
+            base_ref = (
+                os.environ.get(
+                    "STAGE_REUSE_DIFF_BASE",
+                    "",
+                ).strip()
+                or None
+            )
```
geomin12, line 242: for workflow dispatch, we already have a "copy" option, where you can specify which stages to copy. still needed here?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -855,6 +871,125 @@ def _determine_test_type(
...
+    """Return the exact commit whose artifacts may be reused.
+
+    Automatic reuse is allowed only when changed-file analysis was computed
+    from the same resolved base commit:
+
+        diff_base_commit -> diff_head_commit
+    """
+
+    if not (
+        ci_inputs.is_pull_request or ci_inputs.is_push or ci_inputs.is_workflow_dispatch
```
geomin12, line 887: could we just use if ci_inputs.is_schedule?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -855,6 +871,125 @@ def _determine_test_type(
...
+        )
+
+    if not allow_automatic_reuse:
+        if auto_stage_reuse.applied_reuse_stages:
+            reason = (
+                "automatic stage reuse was not applied because "
+                "cross-repository reuse requires explicit inputs"
+            )
+
+            auto_stage_reuse = replace(
```
geomin12, line 961: woah! never seen replace before? any reason we use replace rather than resetting

aka:
`auto_stage_reuse = (..,...,..)`

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -871,35 +1006,60 @@ def decide_jobs(
...
+
+    linux_artifact_families = [
+        all_families[family_name]["linux"]["family"]
+        for family_name in targets.linux_families
+    ]
+
+    windows_artifact_families = [
+        all_families[family_name]["windows"]["family"]
+        for family_name in targets.windows_families
+    ]
```
geomin12, line 1029: might be good to add this to amdgpu_family_matrix.py

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -871,35 +1006,60 @@ def decide_jobs(
...
+
+    baseline_repository = ci_inputs.baseline_repository
+    current_repo = os.environ.get("GITHUB_REPOSITORY", "")
```
geomin12, line 1032: oh i notice in setup_multi_arch_ci.yml, we use THEROCK_REPOSITORY. maybe we can use that?

i realize in rocm-libs / rocm-sys, we don't have artifacts to copy, so we need to use TheRock's repo artifacts

geomin12 (commented):
first pass through!

PR #6766, opened 2026-07-22
fix: Skip empty packages in DEB and RPM packaging

raramakr:
Filter out non-existing directories from sourcedir_list before checking
  if a package has content. This ensures packages without artifacts are
  correctly skipped rather than being created as empty packages.

  Non-versioned packages and metapackages are unaffected and will still
  be created as expected.

  Changes:
  - DEB: Add path.is_dir() filter before empty package check
  - RPM: Return False from generate_spec_file() to skip empty packages
  - RPM: Check return value in caller to avoid rpmbuild failure
  - Now when a dependency lacks gfx artifacts but has host artifacts, we fall back
  to the host version
  
  JIRA ID: https://github.com/ROCm/TheRock/issues/6815

### DIFF
--- a/build_tools/packaging/linux/packaging_utils.py
+++ b/build_tools/packaging/linux/packaging_utils.py
@@ -602,6 +602,7 @@ def process_main_dependencies_kpack(
     """
     is_meta = is_meta_package(pkg_info)
     pkg_name = pkg_info.get("Package")
+    host_fallback_deps = []  # Track deps that need host fallback
 
     if is_meta:
         if config.gfx_arch == GFX_META:
@@ -612,8 +613,7 @@ def process_main_dependencies_kpack(
         else:
             # Arch-specific metapackage: depend on actual runtime packages
             dep_list = pkg_info.get(field_key, [])
-            # Filter deps without artifacts
-            dep_list = filter_dependencies_by_artifacts(
+            dep_list, host_fallback_deps = filter_dependencies_by_artifacts(
                 dep_list, config.artifacts_dir, config.gfx_arch
             )
     elif config.gfx_arch == GFX_META:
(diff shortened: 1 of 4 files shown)


raramakr:
The PR will be unit tested using  https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/linux/tests/build_package_test.py

nunnikri:
The ci build is failing in rpm package installs and deb sanity check. @raramakr  @jonatluu

raramakr:
> The ci build is failing in rpm package installs and deb sanity check. @raramakr @jonatluu

Will check this

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -653,15 +661,27 @@ def process_main_dependencies_kpack(
...
         )
         dep_list = [pkg_name] + gfxarch_deps
 
-    if not dep_list:
+    # Resolve dependencies at the end
+    if not dep_list and not host_fallback_deps:
         return ""
-    return resolve_versioned_dependencies(dep_list, config, is_meta)
+
+    deps = resolve_versioned_dependencies(dep_list, config, is_meta)
```
arvindcheru, line 673: Would request to add more documentation/doc update for Host fallback mechanism

build_tools/packaging/linux/packaging_utils.py
```diff
@@ -1068,6 +1088,10 @@ def has_artifact_for_arch(pkg_name, artifacts_dir, gfx_arch):
         else:
             artifact_suffix = gfx_arch
 
+        # GFX_HOST uses "generic" artifacts
```
arvindcheru, line 1091: one query to double confirm - If I understand correctly, artifact_suffix = "generic", combined with the new fallback resolution using replace(config, gfx_arch=GFX_HOST), this means host-fallback resolution depends on "generic" artifacts existing and being named consistently — worth confirming there's no case where a package has neither gfx-specific nor "generic"-suffixed artifacts but does have some other host-equivalent naming

PR #6769, opened 2026-07-22
[hipthreads] Add hipThreads build recipe and artifact to TheRock

marsavicc:
## Motivation

JIRA ID : ROCM-23943
  
Onboard hipThreads (GPU-side `std::thread`-style concurrency library, now living in rocm-libraries) into TheRock so it builds as part of ROCm. 

## Technical Details

- Add the `hipthreads` build recipe in `math-libs/CMakeLists.txt`, pulling source from `${THEROCK_ROCM_LIBRARIES_SOURCE_DIR}/projects/hipthreads`. 
- Register `[artifacts.hipthreads]` in `BUILD_TOPOLOGY.toml` (target-neutral: the static `libhipthreads.a` is identical across architectures) and add `math-libs/artifact-hipthreads.toml`. 
- Register libhipcxx as a CMake package provider so hipThreads can `find_package(libhipcxx)` (hipThreads is the first consumer of it this way).
- Build with `-DDISABLE_WERROR=ON` (HIP runtime headers trip hipThreads' strict warnings) and document the new `THEROCK_ENABLE_HIPTHREADS` flag.

## Test Plan

Build hipThreads via TheRock against rocm-libraries.

## Test Result

Builds succesfully and produces the hipthreads artifact. 

Note: CI is red until the TheRock rocm-libraries submodule is bumped to a SHA that includes `projects/hipthreads` (migration landed after the current pin).

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

[ROCM-23943]: https://amd-hub.atlassian.net/browse/ROCM-23943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6770, opened 2026-07-22
[hipthreads] Add hipThreads lit and example-app test jobs to TheRock CI

marsavicc:
## Motivation

JIRA ID : ROCM-23943

Run hipThreads' lit unit tests and example apps in TheRock CI on Linux and Windows. Depends on the hipThreads build PR #6769 .

## Technical Details

- Add `test_hipthreads.py` (runs the lit suite) and `test_hipthreads_examples.py` (builds + runs saxpy, InOneWeekend, and spmm against the packaged artifact via find_package).
- Register the `hipthreads` and `hipthreads_examples` jobs in `fetch_test_configurations.py` (linux + windows) and add the `--hipthreads` fetch flag to `install_rocm_from_artifacts.py`.
- Set `HIPTHREADS_VCORES_PER_WGP=1` at test runtime so the scheduler doesn't over-subscribe shared CI GPUs (runtime setting; the shipped library keeps its default).

## Test Plan

Run both scripts against a built artifact with `OUTPUT_ARTIFACTS_DIR` / `THEROCK_BIN_DIR` set, on Linux and Windows.

## Test Result

lit suite and all three example apps pass on Linux and Windows.

## Submission Checklist

[ROCM-23943]: https://amd-hub.atlassian.net/browse/ROCM-23943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


marbre:
Leaving the second round of review (after @ammallya reviewed and approved) as well as the final approval to @geomin12 who is more into testing.

marsavicc:
I rebased this branch on top of `main` to include new rocm-libraries bump where in which our library is included.

build_tools/github_actions/test_executable_scripts/test_hipthreads.py
```diff
@@ -0,0 +1,185 @@
...
+environ_vars = os.environ.copy()
+
+# Resolve absolute paths
+OUTPUT_ARTIFACTS_PATH = Path(OUTPUT_ARTIFACTS_DIR).resolve()
+THEROCK_BIN_PATH = Path(THEROCK_BIN_DIR).resolve()
+
+# Set up ROCm/HIP environment
+environ_vars["ROCM_PATH"] = str(OUTPUT_ARTIFACTS_PATH)
+environ_vars["HIP_DEVICE_LIB_PATH"] = str(
+    OUTPUT_ARTIFACTS_PATH / "lib/llvm/amdgcn/bitcode/"
```
geomin12: could we use proper Path (so this is compatible with windows)


```suggestion
    str(OUTPUT_ARTIFACTS_PATH / "lib" / "llvm" / "amdgcn" / "bitcode")
```

build_tools/github_actions/test_executable_scripts/test_hipthreads_examples.py
```diff
@@ -0,0 +1,263 @@
...
+    source_dir = EXAMPLES_ROOT / example["subdir"]
+    if not source_dir.exists():
+        raise FileNotFoundError(f"Example source dir not found: {source_dir}")
+
+    build_dir = source_dir / "build"
+    llvm_bin = OUTPUT_ARTIFACTS_PATH / "lib" / "llvm" / "bin"
+    clang = llvm_bin / f"clang{EXE_SUFFIX}"
+    clangxx = llvm_bin / f"clang++{EXE_SUFFIX}"
+
+    configure_cmd = [
```
geomin12, line 138: ah so can we include this in build?

the intention of these scripts was just a python wrapper to help the CI run these tests. we shouldn't do any builds here (as this is running on GPU runner). ideally out of the box (from artifacts), the tests should run without any additional build steps

geomin12 (commented):
obviously tests fail as build is not include yet. let's wait for build PR to land, then, we can make sure these tests are passing

marsavicc:
> obviously tests fail as build is not include yet. let's wait for build PR to land, then, we can make sure these tests are passing

Yes. Just a quick question, we have this PR open https://github.com/ROCm/rocm-libraries/pull/9755 that addresses a small issue that came up in one of the recent therock nighlies regarding lit tests, but it's stuck at this `waiting on math-ci` and we can't merge it. How do we skip/trigger/resolve this math ci thing?

build_tools/github_actions/test_executable_scripts/test_hipthreads.py
marsavicc: Yeah no problem

build_tools/github_actions/test_executable_scripts/test_hipthreads_examples.py
marsavicc: agree in principle that the GPU runner should just run, not build. The wrinkle is that our artifact is `TARGET_NEUTRAL` . `libhipthreads.a` is arch-independent, but an example executable embeds device code for a specific gfx arch. So pre-building the examples in the build stage would make the artifact target-specific (per-arch example binaries), which is a change to the artifact model rather than just relocating the build step. 

Given the release timeline I'd prefer (if possible) to keep the configure+build in the wrapper for now. 
Our goal is to make a switch and ship a `.so` library so with that we could move the example compilation into the build stage (as a target-specific test artifact) for 10.1/10.2.  

Just noting that the lit suite compiles on the runner too, by design, so this wrapper isn't uniquely doing builds.

geomin12: sounds good. in this case, can we add a comment in the code noting this?

geomin12:
> > obviously tests fail as build is not include yet. let's wait for build PR to land, then, we can make sure these tests are passing
> 
> Yes. Just a quick question, we have this PR open [ROCm/rocm-libraries#9755](https://github.com/ROCm/rocm-libraries/pull/9755) that addresses a small issue that came up in one of the recent therock nighlies regarding lit tests, but it's stuck at this `waiting on math-ci` and we can't merge it. How do we skip/trigger/resolve this math ci thing?

sure! i think since hipthreads seems new and not a part of math-ci, we can force merge it. let me know or you can ask in the Gardeners - rocm-libraries channel

for CI checks, since rocm-libraries uses TheRock build system (and the build PR hasn't landed), we can't add PR checks until after unfortunately

geomin12 (commented):
oh unit tests fail :( 

but will re-review once lands #6769, we can work together to getting CI checks in rocm-systems when both #6769 and this PR lands

build_tools/github_actions/test_executable_scripts/test_hipthreads_examples.py
marsavicc: Yeah no problem

math-libs/CMakeLists.txt
```diff
@@ -723,6 +723,7 @@ if(THEROCK_ENABLE_HIPTHREADS)
       rocm-cmake
       libhipcxx
     RUNTIME_DEPS
+      amd-llvm
```
marsavicc, line 726: @marbre I added `amd-llvm` runtime dep for lit tests here

geomin12 (commented):
waiting on tests, but let's fix unit tests pls

marsavicc:
> waiting on tests, but let's fix unit tests pls

this [commit](https://github.com/ROCm/rocm-libraries/commit/dc7fc5aa1da53f912e1318f80f61e64858d94ae1) was missing yesterday so let's just wait for the CI to finish

math-libs/CMakeLists.txt
marbre, line 726: Yeah, for lit test you might need this but in that case you most likely can drop hipcc.

math-libs/CMakeLists.txt
```diff
@@ -723,6 +723,7 @@ if(THEROCK_ENABLE_HIPTHREADS)
...
       hip-clr
       hipcc
```
marbre:
```suggestion
```

marsavicc:
@marbre For the current windows test failures, are we missing `argv.append("core-hip_run")` in `build_tools/install_rocm_from_artifacts.py)`? It looks to me that HIP Runtime DLL is not in `_lib` but in `_run` component.

math-libs/CMakeLists.txt
marsavicc: Done

marbre:
> @marbre For the current windows test failures, are we missing `argv.append("core-hip_run")` in `build_tools/install_rocm_from_artifacts.py)`? It looks to me that HIP Runtime DLL is not in `_lib` but in `_run` component.

If you need it, please model it not just via the install script and rather as a runtime dependency but please check the exact package you really want to depend on.

marbre:
Unit tests fail with
```
FAILED tests/install_rocm_from_artifacts_test.py::TestDebugToolsAmdLlvmDev::test_debug_tools_includes_amd_llvm_dev - AttributeError: 'Namespace' object has no attribute 'hipthreads'
```

marsavicc:
> > @marbre For the current windows test failures, are we missing `argv.append("core-hip_run")` in `build_tools/install_rocm_from_artifacts.py)`? It looks to me that HIP Runtime DLL is not in `_lib` but in `_run` component.
> 
> If you need it, please model it not just via the install script and rather as a runtime dependency but please check the exact package you really want to depend on.

Yeah, I added `argv.append("core-hip_run")`, the HIP runtime DLL (`amdhip64*.dll`) is in core-hip's `_run` component (`bin/**`), not `_lib`. On modeling it as a runtime dep: the recipe already declares `hip-clr` in `RUNTIME_DEPS`, which maps to the core-hip artifact, so the dependency edge is already there; the gap was just that the test fetch pulled `_lib`/`_dev` but not `_run`. 

As I see, adding `_run` there matches how other test jobs pull runtime components (miopen_run, rocprofiler-*_run, etc.).

build_tools/github_actions/fetch_test_configurations.py
```diff
@@ -676,6 +676,34 @@ def _build_container_options(job_config: dict, platform: str) -> dict:
...
+        "platform": ["linux", "windows"],
+        "total_shards_dict": {
+            "linux": 1,
+            "windows": 1,
+        },
+    },
+    # hipthreads example apps (build + run consumer samples against the artifact).
+    # Linux-only for now: on Windows the InOneWeekend example hits a clang HIP
+    # device-lambda parse bug (ROCM-25863). TODO: add "windows" back once a
+    # rocm-libraries submodule bump includes the fix (commit 2784675cb5).
```
marbre: LGTM but as another submodule bump landed recently you may be able to re-add Windows already. Can also be a follow up, up to you!

marbre (approved):

build_tools/github_actions/fetch_test_configurations.py
marsavicc: yep saw that, I'll re-add it here

dysu-amd:
Hi @marsavicc, the Windows `hipthreads_examples` test has been flaky since merging, depending on which gfx110x runner group is used, please see issue https://github.com/ROCm/TheRock/issues/7132 for more information.

PR #6776, opened 2026-07-22
fix(rocprofiler-sdk): add openmpi to BUILD_DEPS

araravik-psd:
## Problem

  When openmpi was added as a super-project component in TheRock, the dep provider (`therock_subproject_dep_provider.cmake:68`) began intercepting `find_package(MPI)` calls from
  subprojects. rocprofiler-sdk calls `find_package(MPI)` in `cmake/rocprofiler_config_interfaces.cmake:408` — this was added as part of rocshmem tracing support (AIPROFSDK-886) because
  rocshmem's installed CMake config references `MPI::MPI_CXX` in its link interface and must be located before `find_package(rocshmem)` can succeed.

  Since openmpi was not declared in rocprofiler-sdk's `therock_cmake_subproject_declare`, the dep provider hard-errors with:

  CMake Error at cmake/therock_subproject_dep_provider.cmake:68 (message):
    Project contains find_package(MPI) for a package available in the
    super-project but not declared: Add a BUILD_DEPS or RUNTIME_DEPS
    appropriately

Commit in therock causing this: https://github.com/ROCm/TheRock/commit/f1a803131ccf8d3e5799c059f21e2bde4d8782ec

**After this commit:** The package registry is disabled inside all subprojects. Now when rocprofiler-sdk calls `find_package(MPI)`, the dep provider enforces that MPI must be declared in `BUILD_DEPS` or `RUNTIME_DEPS`. Since the RCCL CI build passes `-DTHEROCK_ENABLE_MPI=ON` (via `THEROCK_ENABLE_COMM_LIBS=ON`), MPI is in `THEROCK_STRICT_PROVIDED_PACKAGES` — but rocprofiler-sdk's declaration in `TheRock/profiler/CMakeLists.txt` lists no MPI dependency → `FATAL_ERROR`.

  This caused rocprofiler-sdk configure to fail, which in turn caused the RCCL CI build to fail in rocm-systems after the TheRock hash bump to `22004a6` (2026-07-20).

  ## Fix

  Add `openmpi` to `BUILD_DEPS` in the `therock_cmake_subproject_declare(rocprofiler-sdk ...)` block in `profiler/CMakeLists.txt`. This tells the dep provider the `find_package(MPI)`
  call is intentional and allows configure to proceed normally.

  ## Testing

  - rocprofiler-sdk configure should complete without error
  - RCCL CI build in rocm-systems should pass after bumping to a TheRock hash that includes this fix

Test rocm-systems run with pinned TheRock commit: https://github.com/ROCm/rocm-systems/actions/runs/29940406058/job/88993031854?pr=9029

### DIFF
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -130,14 +130,13 @@ if(THEROCK_ENABLE_ROCPROFV3)
       rocprofiler-sdk-tests
     COMPILER_TOOLCHAIN
       amd-hip
-    BUILD_DEPS
-      ${_openmpi_optional_dep}
     RUNTIME_DEPS
       aqlprofile
       hip-clr
       rocprof-trace-decoder
       rocprofiler-register
       therock-yaml-cpp
+      ${_openmpi_optional_dep}
       ${_rocprofiler_sdk_optional_deps}
       ${THEROCK_BUNDLED_ELFUTILS}
       ${THEROCK_BUNDLED_LIBDRM}
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -130,6 +130,8 @@ if(THEROCK_ENABLE_ROCPROFV3)
       rocprofiler-sdk-tests
     COMPILER_TOOLCHAIN
       amd-hip
+    BUILD_DEPS
+      openmpi
     RUNTIME_DEPS
       aqlprofile
       hip-clr
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -131,7 +131,7 @@ if(THEROCK_ENABLE_ROCPROFV3)
     COMPILER_TOOLCHAIN
       amd-hip
     BUILD_DEPS
-      openmpi
+      ${_openmpi_optional_dep}
     RUNTIME_DEPS
       aqlprofile
       hip-clr


profiler/CMakeLists.txt
```diff
@@ -130,6 +130,8 @@ if(THEROCK_ENABLE_ROCPROFV3)
       rocprofiler-sdk-tests
     COMPILER_TOOLCHAIN
       amd-hip
+    BUILD_DEPS
+      ${_openmpi_optional_dep}
     RUNTIME_DEPS
       aqlprofile
```
ScottTodd: Should this `_openmpi_optional_dep` be a build dep or a runtime dep?

The other uses have it in `RUNTIME_DEPS`:
* https://github.com/ROCm/TheRock/blob/29faf3b6639bd85ea589a17a447485e24887d37f/profiler/CMakeLists.txt#L257-L268
* https://github.com/ROCm/TheRock/blob/29faf3b6639bd85ea589a17a447485e24887d37f/profiler/CMakeLists.txt#L348-L361

ScottTodd (approved):
You may also be interested in https://github.com/ROCm/TheRock/pull/6610 and https://github.com/ROCm/TheRock/pull/5627 (cc @minseobshin11). I'm not sure if we want to keep `THEROCK_ENABLE_MPI` given the packaging/distribution complexity there. If it's just a runtime dep used at test time then we can hopefully satisfy it via the test dockerfile, so long as the projects are at least able to build with dynamic linking MPI support somehow...

profiler/CMakeLists.txt
araravik-psd: I believe, It should stay in BUILD_DEPS. The find_package(MPI) in rocprofiler-sdk fires at configure time. the dep provider intercepts it before a single source file is compiled. I feel, the other subprojects that put _openmpi_optional_dep in RUNTIME_DEPS are most likely only using MPI for test execution, not as a compile-time linking dependency.

geomin12 (approved):

profiler/CMakeLists.txt
ScottTodd: All `RUNTIME_DEPS` are also `BUILD_DEPS`, not all `BUILD_DEPS` are `RUNTIME_DEPS`, see https://github.com/ROCm/TheRock/blob/29faf3b6639bd85ea589a17a447485e24887d37f/cmake/therock_subproject.cmake#L244-L248

If MPI is _only_ used at build time then `BUILD_DEPS` is appropriate. If it's also used at run time then it should go in `RUNTIME_DEPS`.

araravik-psd: Thanks @ScottTodd for linking #6610 and #5627, Keeping BUILD_DEPS, MPI is intentionally not distributed with ROCm artifacts (from #6610 which added the dedicated OpenMPI test image for exactly this reason, and #5627 which wires rocprofiler-sdk tests to use it). libmpi.so is satisfied at test runtime via the container
image, not shipped with release artifacts, so BUILD_DEPS would be accurate here?

ScottTodd: Maybe? I'd prefer local consistency with a small fix like this.

If we need MPI available at configure time but we don't want to distribute it (leaving it as an optionally dynamically linked runtime dependency), then I think `BUILD_DEPS` _could_ work, but we'd want to be careful about how we represent that in the build and packaging files. Pretty much all of our other dependencies expect to be vendored and distributed together with the libraries and tests.

araravik-psd: Good point on local consistency, I'll move it to RUNTIME_DEPS to match rocprofiler-compute and rocprofiler-systems. The proper long-term fix (guarding find_package(MPI) in rocprofiler-sdk's cmake so it's not reached in standard builds at all) will come as a follow-up PR in rocprofiler-sdk, which will make the dep declaration here moot when THEROCK_ENABLE_MPI=OFF. Moving now to unblock RCCL CI.

marbre (commented):
Post-review: Adding Open MPI as a runtime dep is indeed correct here.

PR #6778, opened 2026-07-22
fix(ci): Re-enable amd-smi static for gfx1151 in CI sanity check

### LINKED ISSUE #2789: [Issue] `amd-smi static` command is not enabled for Linux `gfx1151` see stuck issue on CI
Currently, `amdsmi metrics` support is not available for Linux gfx1151. The `amd-smi static` command seems to be stuck on a CI run for gfx1151, so it is presently disabled in CI with #2790. We are working on manually trying this on gfx1151 to check if the stuck issue is seen consistently. For now, this issue is a reminder to re-enable amd-smi static command.

darren-amd:
## Motivation

Closes #2789

The CI "Driver / GPU sanity check" skips `amd-smi static` for `gfx1151` because the command used to hang and time out on gfx1151 (tracked in #2789). `amd-smi static` no longer times out on gfx1151, so the skip can be removed to restore sanity-check coverage on that family.

JIRA ID : ROCM-9123

## Technical Details

`gfx1151` was the only remaining entry in `unsupported_amdsmi_families` in `build_tools/print_driver_gpu_info.py`. Rather than leave an empty list, this removes the list, the `AMDGPU_FAMILIES` env read, and the now-dead conditional (and its `TODO(#2789)`) entirely, so `amd-smi static` always runs on Linux.

## Test Plan

Verified on a gfx1151 machine by running `amd-smi static` and `AMDGPU_FAMILIES=gfx1151 python3 ./build_tools/print_driver_gpu_info.py` before and after the change.

## Test Result

`amd-smi static` completes in ~0.1s with no timeout. The sanity script now runs the `amd-smi static` section end-to-end for gfx1151 with no errors, and the default (unset `AMDGPU_FAMILIES`) path still works.

## Submission Checklist

### DIFF
--- a/build_tools/print_driver_gpu_info.py
+++ b/build_tools/print_driver_gpu_info.py
@@ -24,9 +24,6 @@ import subprocess
 import sys
 from typing import List, Optional
 
-AMDGPU_FAMILIES = os.getenv("AMDGPU_FAMILIES")
-unsupported_amdsmi_families = ["gfx1151"]
-
 
 def log(*args, **kwargs):
     print(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


marifamd:
unit tests not needed, this is enabling testing

geomin12 (approved):
lgtm. due to limited count of linux gfx1151, this only runs on nightlies

we can land and see how it does! if not good, we'll revert

PR #6764, opened 2026-07-22
fix(wsl-rocdxg): add missing env vars to Upload stage logs WSLENV

amd-chiranjeevi:
## Motivation

The "Upload stage logs" step in `multi_arch_build_wsl_rocdxg_artifacts.yml` fails with S3 AccessDenied on ci-typed runs (e.g. nightly release builds). The build itself succeeds; only log upload is broken.

PR https://github.com/ROCm/TheRock/pull/6658 refactored the wsl-rocdxg workflow to run builds in a WSL-native directory for performance. In doing so, it added a  new WSLENV to the "Upload stage logs" step — but omitted RELEASE_TYPE from it. The "Push stage artifacts" step immediately above has the correct full WSLENV including RELEASE_TYPE.

 Without RELEASE_TYPE propagated into WSL, post_stage_upload.py → s3_buckets.py resolves the wrong bucket role (therock-nightly) which lacks s3:PutObject on the ci-prefixed path, causing all 4 log files to fail upload after 3 retries. This only manifests on release_type: ci runs. Prior nightly runs (release_type: nightly) worked because the nightly role matched the path.

Failing run: https://github.com/ROCm/rockrel/actions/runs/29879687428/job/88800898315
Last passing run: https://github.com/ROCm/rockrel/actions/runs/29789834120/job/88562095214

GitHub issue: https://github.com/ROCm/TheRock/issues/6763

## Fix

  Fix

  Add the missing env vars to WSLENV in the "Upload stage logs" step to match "Push stage artifacts":
```diff
-   WSLENV: THEROCK_HOST_WORKSPACE/p:AWS_ACCESS_KEY_ID:AWS_SECRET_ACCESS_KEY:AWS_SESSION_TOKEN:AWS_DEFAULT_REGION:AWS_REGION
+  WSLENV: THEROCK_HOST_WORKSPACE/p:GITHUB_REPOSITORY:GITHUB_EVENT_NAME:GITHUB_EVENT_PATH/p:RELEASE_TYPE:AWS_ACCESS_KEY_ID:AWS_SECRET_ACCESS_KEY:AWS_SESSION_TOKEN:AWS_DEFAULT_REGION:AWS_REGION
```

## Test Result

Waiting for CI results.
mulit_arch_realese dev run: https://github.com/ROCm/rockrel/actions/runs/29901093180 --> passed

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

Issue #6763, opened 2026-07-22
[Issue] multi_arch_release nightly : `wsl-rocdxg stage log upload fails with S3 AccessDenied`


TriveniTadapaneni:
## Context

- **workflow:** `multi_arch_release.yml`
- **Run:** https://github.com/ROCm/rockrel/actions/runs/29879687428
- **Failing job:** [Stage - WSL ROCDXG · run 29879687428](https://github.com/ROCm/rockrel/actions/runs/29879687428/job/88800898315)

## Problem

The nightly `multi_arch_release` WSL ROCDXG job fails in its "Upload stage logs" step with an S3 `AccessDenied` error after [ROCm/TheRock#6658](https://github.com/ROCm/TheRock/pull/6658) migrated the build to a WSL-native directory.

## Error

```text
An error occurred (AccessDenied) when calling the PutObject operation:
User: arn:aws:sts::692859939525:assumed-role/therock-nightly/GitHubActions
is not authorized to perform: s3:PutObject on resource:
"arn:aws:s3:::therock-ci-artifacts/29879687428-linux/logs/wsl-rocdxg/therock-wsl-rocdxg_build.log"
because no identity-based policy allows the s3:PutObject action.

RuntimeError: Failed to upload 4/4 files.
```

## Suspected Root Cause

[TheRock commit 69aa122](https://github.com/ROCm/TheRock/commit/69aa122d1b8366445208254c716bed36e7d58a78) changed the "Upload stage logs" step to run inside WSL (`shell: wsl-bash`) but omitted `GITHUB_REPOSITORY` from that step's `WSLENV`. Inside WSL the variable is unset, so `post_stage_upload.py` resolves `github_repository` to `ROCm/TheRock` (the inner checkout) instead of `ROCm/rockrel`. This causes the bucket selector to choose `therock-ci-artifacts` (the CI bucket) instead of `therock-nightly-artifacts`. The `therock-nightly` IAM role is not authorized to write to `therock-ci-artifacts`.

The build itself succeeds — all 31 compile units linked and the three main artifacts uploaded to `therock-nightly-artifacts` without issue.

## AI Suggested Fix

[...]

PR #6781, opened 2026-07-22
ci(wsl-rocdxg): move most WSLENV into common job-level env

dysu-amd:
## Motivation

Move most definitions of `WSLENV` into a common job-level variable as suggested in https://github.com/ROCm/TheRock/pull/6764/changes#r3632422006

Might have prevented https://github.com/ROCm/TheRock/issues/6763 from happening.

## Technical Details

- Removed most step-level definitions of `WSLENV` and move their values to the job-level `WSLENV` definition
  - `THEROCK_WIN_SDK_SHARED` is able to be specified in job-level `WSLENV`, even though it is defined in a runtime step.
    - `WSLENV` is parsed by WSL at runtime, so it is able to pick up the SDK path without issues as long as `env: THEROCK_WIN_SDK_SHARED: ...` is set for each step that requires it.
    - Added comments to the three steps using the SDK var
- The AWS credential envs cannot be set for the whole job, so append them to `WSLENV` at the step level.
  - The AWS envs are set during the `Configure AWS credentials` step. Before this step is run, they are empty.
  - `Fetch inbound artifacts`, which runs before `Configure`, will read the empty envs and try to pull from an empty S3 URL, so they cannot be set before this step is run.
  - Therefore we don't set them in job-level `WSLENV` and instead append them in any steps that require them.
  - `env: WSLENV: ${{ env.WSLENV }}:...` allows appending to `WSLENV` rather than overwriting it.

## Test Plan

- [x] WSL CI workflow passes
- [x] WSL release workflow passes

## Test Result

Release workflow run: https://github.com/ROCm/TheRock/actions/runs/29964916475

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
+++ b/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
@@ -64,7 +64,7 @@ jobs:
       STAGE_NAME: ${{ inputs.stage_name }}
       RELEASE_TYPE: ${{ inputs.release_type }}
       THEROCK_HOST_WORKSPACE: ${{ github.workspace }}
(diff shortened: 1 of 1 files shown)


dysu-amd:
@ScottTodd @amd-chiranjeevi Is there a way I could test out these changes in a 'release' context?

ScottTodd:
> @ScottTodd @amd-chiranjeevi Is there a way I could test out these changes in a 'release' context?

You can trigger a dev release using workflow_dispatch here: https://github.com/ROCm/TheRock/actions/workflows/multi_arch_release.yml

ScottTodd (approved):

PR #6782, opened 2026-07-22
ci(pytorch): add optional PyTorch git origin override to wheel builds

ethanwee1:
ISSUE ID : https://github.com/ROCm/TheRock/issues/6735

## Summary

Add an optional `pytorch_gitrepo_origin` workflow input to the portable Linux and Windows PyTorch wheel build workflows so CI can checkout PyTorch from a fork URL (for pre-merge validation) while defaulting to `ROCm/pytorch` when unset.

## Changes

- `multi_arch_build_portable_linux_pytorch_wheels.yml`: new `pytorch_gitrepo_origin` input; pass through to `pytorch_torch_repo.py checkout --gitrepo-origin`.
- `multi_arch_build_windows_pytorch_wheels.yml`: same input and wiring for the Windows wheel pipeline.

## Testing

- Default behavior unchanged when `pytorch_gitrepo_origin` is empty (still uses `https://github.com/ROCm/pytorch.git`).
- Used to validate upstream PyTorch packaging fixes against TheRock Windows wheel builds before merge.

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
@@ -46,6 +46,12 @@ on:
         description: PyTorch ref to checkout. Typically "nightly" or "release/X.Y".
         type: string
         required: true
+      pytorch_gitrepo_origin:
+        description: >-
+          Optional PyTorch git origin override (for example, a fork URL for
+          pre-merge testing). Defaults to ROCm/pytorch for stable refs.
+        type: string
+        default: ""
       rocm_version:
         description: ROCm package version to install and build against (e.g. 7.10.0.dev0)
         type: string
@@ -104,6 +110,12 @@ on:
         description: PyTorch ref to checkout (for example, "release/2.12").
         type: string
         default: "release/2.12"
+      pytorch_gitrepo_origin:
+        description: >-
+          Optional PyTorch git origin override (for example, a fork URL for
+          pre-merge testing). Defaults to ROCm/pytorch for stable refs.
+        type: string
+        default: ""
       rocm_version:
         description: ROCm package version to install and build against (e.g. 7.10.0.dev0)
         type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
```diff
@@ -229,8 +241,12 @@ jobs:
       - name: Checkout PyTorch source repos (stable)
         if: ${{ inputs.pytorch_git_ref != 'nightly' }}
         run: |
+          pytorch_origin="${{ inputs.pytorch_gitrepo_origin }}"
+          if [ -z "${pytorch_origin}" ]; then
+            pytorch_origin="https://github.com/ROCm/pytorch.git"
+          fi
           ./external-builds/pytorch/pytorch_torch_repo.py checkout \
-            --gitrepo-origin https://github.com/ROCm/pytorch.git \
+            --gitrepo-origin "${pytorch_origin}" \
```
ScottTodd, line 249: I have a slight preference for this style of providing workflow inputs and default values where you load the input into an environment variable and provide a default if the input is empty: https://github.com/ROCm/TheRock/blob/9289636133ab11a7d83c76fa1245709a5e711353/.github/workflows/build_portable_linux_python_packages.yml#L111-L113

For the approach I linked:
* multiple steps throughout the workflow can use the value like `${{ env.ARTIFACT_RUN_ID }}`
* The default close to where the input is

For the approach currently on this PR:
* Multiple uses would need to duplicate the default logic (possibly diverging)
* Inline bash should be avoided, see https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash (we could clarify/extend that to also cover yaml patterns that are preferred over bash too)

.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
ScottTodd, line 148: We could also include the gitrepo_origin in the run-name if one is provided, but I think just the ref should be sufficient. What do you think? Would you want it included in the run name? We might be short on character count since repository URLs are pretty long...

If we did make that sort of change, we'd want a multiline string so optional values can be checked, like this: https://github.com/ROCm/TheRock/blob/9289636133ab11a7d83c76fa1245709a5e711353/.github/workflows/multi_arch_release.yml#L104-L107

ScottTodd (approved):

ethanwee1:
All therock-pr-bot checks green (approved). Expanded PR description with summary/changes/testing. Ready for maintainer merge.

Issue #6786, opened 2026-07-22
dev packages don't pull the runtime library; find_package() fails at the imported-target check on dev-only installs


lucbruni-amd:
### Description

Installing only a math-library `-dev` package leaves `find_package(<lib>)` unable to configure: the imported target references the runtime `.so`, which no installed package provides.

### Reproduction

Ubuntu 24.04, repo.amd.com multi-arch 7.14. Install `amdrocm-blas-dev7.14`, then `find_package(hipblas REQUIRED)`:

```
CMake Error at .../hipblas-targets.cmake:85 (message):
  The imported target "roc::hipblas" references the file
     "/opt/rocm/core-7.14/lib/libhipblas.so.3.5"
  but this file does not exist.
```

`libhipblas.so.3.5` and the `libhipblas.so` dev symlinks both ship in `amdrocm-blas-host7.14` (18 MB download / 280 MB installed, no device code).

### Root cause

A kpack-only regression, not a missing declaration. All 9 affected `-devel` packages already declare the dependency in `package.json` (`amdrocm-blas-devel` -> `amdrocm-blas`), and single-arch still honors it: `amdrocm-blas-dev7.13-gfx110x` depends on `amdrocm-blas7.13-gfx110x`, and 512 of 568 `-dev` packages in `repo.amd.com/rocm/packages/ubuntu2404` carry a ROCm dependency.

#5198 made kpack `-devel` packages keep their dependencies (existence check, no arch filter). #5099 replaced that with `filter_dependencies_by_artifacts(dep_list, artifacts_dir, gfx_arch)`. For `-devel`, `gfx_arch` is `""`, which matches no artifact directory, so every gfxarch dependency is dropped. The comment directly above the call still reads "include everything".

### Open decision

Which post-split variant `-dev` should point at: `-host` (arch-independent, 280 MB for blas) or the meta (host plus 25 device packages, 4.6 GB). `-host` is the only option that preserves the arch-independence that makes `-devel` non-gfxarch in the first place.

### Affected packages

blas, dnn, fft, hiptensor, rand, rccl, rocalution, solver, sparse.

lucbruni-amd:
The cmake-config dependency fixes referenced above are in #6787. This issue tracks the remaining runtime-library (`.so`) dependency gap that #6787 does not address.

lucbruni-amd:
[...]

PR #6787, opened 2026-07-22
Add missing cmake dependencies to library -devel packages

lucbruni-amd:
## Motivation

The `*-config.cmake` files in the math-library `-devel` packages call `find_dependency(hip)`, `find_dependency(hipblas-common)` and `find_dependency(rocblas)`, but the `-devel` packages declare no dependency on the packages that ship those cmake configs (`amdrocm-blas-dev7.14` ships `Depends: libc6` only).

Fixes ROCm/ROCm#6447, where `find_package(hipblas)` fails with `Could not find a package configuration file provided by "hipblas-common"`.

## Technical Details

`build_tools/packaging/linux/package.json`, adding to `DEBDepends` and `RPMRequires`:

| `-devel` package | added dependency | satisfies |
| --- | --- | --- |
| blas | `amdrocm-runtime-devel`, `amdrocm-hipblas-common-devel` | `hip`, `hipblas-common` |
| solver | `amdrocm-runtime-devel`, `amdrocm-blas-devel` | `hip`, `rocblas` |
| fft, rand, sparse, ccl, rccl, dnn, rocalution, hiptensor | `amdrocm-runtime-devel` | `hip` (dnn also `amd_comgr`, via runtime-devel) |

**Why the existing deps are dropped:** `is_gfxarch_package()` classifies every non-meta `-devel` package as non-gfxarch in kpack mode (`packaging_utils.py:257-260`), so `-devel` builds via `build_simple_package_variants()` (no host/device/meta split) and its deps take the `not is_gfxarch_package(pkg_info)` branch. `filter_dependencies_by_artifacts()` then drops the existing `amdrocm-blas` dep, because that one *is* gfxarch and `has_artifact_for_arch("amdrocm-blas", dir, gfx_arch="")` matches nothing.

**Why the added deps survive:** all are `-devel`, so the same classification makes them non-gfxarch, and `filter_dependencies_by_artifacts()` keeps non-gfxarch deps unconditionally (`packaging_utils.py:1171-1176`).

**On `amdrocm-runtime-devel` pulling `amdrocm-llvm-devel`** (~148 MB download / 2.0 GB installed): not avoidable from this file. `hip-config-amd.cmake` calls `find_dependency(AMDDeviceLibs)`, whose config ships in `amd-llvm_dev`; the `lib/cmake/AMDDeviceLibs` entry in `base_lib` is a bare `include()` redirect to it. `install_rocm_from_artifacts.py` pulls `amd-llvm_dev` for the same reason (kfdtest, rpp, libhipcxx, hipthreads). The dep is pre-existing on `amdrocm-runtime-devel`; narrowing it belongs there, not here.

## Known limitation

A `-dev`-only install still fails one line later at the imported-target check, because the runtime `.so` (shipped by `amdrocm-blas-host`) is not pulled in. Tracked in #6786.

## Test Plan

Generate the control file for each affected `-devel` package via the packaging harness and confirm the added dependencies are present; run the packaging unit tests.

## Test Result

Every affected `-devel` package now lists the expected dependencies (previously `Depends: libc6` only). `build_package_test.py` passes 21/21.

## Submission Checklist


lucbruni-amd:
@nunnikri maybe you can help with this. Also filed https://github.com/ROCm/TheRock/issues/6786 for follow-up.

build_tools/packaging/linux/package.json
```diff
@@ -567,10 +567,12 @@
     "BuildArch": "x86_64",
     "DEBDepends": [
       "libc6",
-      "amdrocm-fft"
+      "amdrocm-fft",
+      "amdrocm-runtime-devel"
```
nunnikri, line 544: The changes looks fine. But thie runtime-devel will install llvm-devel package which usually is huge. 

cc: @raramakr

nunnikri (approved):

build_tools/packaging/linux/package.json
marbre: Any way to avoid this?

lucbruni-amd: Bumping

nunnikri (approved):

build_tools/packaging/linux/package.json
nunnikri: For removing runtime dev dependency, the components needs change. They shouldnt include hip/rocr headers. Which I dont know its possible. 
Another option is to split amdrocm-runtime-devel package, which also i wont suggest

lucbruni-amd:
Merging as failing CI is unrelated. Can certainly revert if this causes issues. Side-effect is `amdrocm-llvm-devel` (for a total of about ~+2GB) is now pulled transitively in for affected `-dev` packages.

This will not fix the linked issue retroactively, so `apt install amdrocm-hipblas-common-dev7.14` as a workaround.

ROCm/TheRock#6786 for follow-up.

PR #6789, opened 2026-07-22
fix: Fix command injection vulnerability in upload_package_repo.py

raramakr:
Refactor run_command() to prevent shell injection attacks:
  - Accept list[str] instead of string commands with shell=True
  - Add stdout parameter for file redirects instead of shell redirection
  - Add type hints and docstring

Additional improvements:
  - Add encoding="utf-8" to all text file operations
  - Convert os.path usage to pathlib.Path for consistency

JIRA ID: https://amd-hub.atlassian.net/browse/ROCM-26880

### DIFF
--- a/build_tools/packaging/linux/upload_package_repo.py
+++ b/build_tools/packaging/linux/upload_package_repo.py
@@ -496,7 +496,7 @@ def run_command(cmd: list[str], cwd=None, stdout=None):
     """
     print(f"Running: {' '.join(cmd)}")
     if stdout is not None:
-        with open(stdout, "w", encoding="utf-8") as f:
+        with open(stdout, "wb") as f:
             subprocess.run(cmd, check=True, cwd=cwd, stdout=f)
     else:
         subprocess.run(cmd, check=True, cwd=cwd)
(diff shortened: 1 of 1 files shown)


nunnikri (commented):
Looks good to me. Lets complete the ci run

build_tools/packaging/linux/upload_package_repo.py
```diff
@@ -465,14 +483,28 @@ def regenerate_repo_metadata_from_s3(
...
+        cmd: Command and arguments as a list of strings
+        cwd: Working directory for the command
+        stdout: Optional path to redirect stdout to a file
+
+    Raises:
+        subprocess.CalledProcessError: If the command exits with non-zero status
+    """
+    print(f"Running: {' '.join(cmd)}")
+    if stdout is not None:
+        with open(stdout, "w", encoding="utf-8") as f:
```
nunnikri: Its safer to open it as binary with "wb". Else some of the files output like gzip can get corrupted. 

Previous command was a shell rewrite. 
gzip -9c Packages > Packages.gz
Now it opens as a file.
open(stdout, "w", encoding="utf-8"). but gzip output is binary. Opening it as text can corrupt the output.

nunnikri:
```suggestion
        with open(stdout, "wb", encoding="utf-8") as f:
```

nunnikri (approved):
Overall looks ok. Lets update the comments and merge only after complete ci pass

build_tools/packaging/linux/upload_package_repo.py
raramakr: Done

PR #6791, opened 2026-07-22
ci: clarify PyTorch build dispatch inputs and default the ROCm index

### LINKED ISSUE #6790: PyTorch build workflows: clarify dispatch inputs and default the ROCm index
### Summary

Manually dispatching the multi-arch PyTorch build workflows (`multi_arch_build_portable_linux_pytorch_wheels.yml`, `multi_arch_build_windows_pytorch_wheels.yml`) is error-prone for people outside the CI team. Two recurring mistakes:

1. **Wrong repo in `repository`/`ref`.** These inputs select which **TheRock** checkout the build scripts come from, but they're easily mistaken for the PyTorch source and set to `rocm/pytorch`. The outer checkout then grabs the wrong repo and every build step fails with `build_tools/github_actions/*.py: No such file or directory` (surfaced as the misleading "Executing the custom container implementation failed").

2. **No ROCm package source.** Leaving both `rocm_package_index_url` and `rocm_package_find_links_url` blank makes `pip install rocm[libraries,devel]==<nightly>` resolve against PyPI, where `rocm` only exists as a placeholder `0.1.0`, so the pinned nightly version can't be found.

### Proposed fix

- Clarify the `repository`/`ref` input descriptions (TheRock, not PyTorch; use `pytorch_git_ref` for the PyTorch branch).
- Clarify the ROCm index/find-links descriptions (use one OR the other).
- Default `rocm_package_index_url` on the Linux portable `workflow_dispatch` path to the nightly multi-arch index (`https://rocm.nightlies.amd.com/whl-multi-arch`) so a bare manual build works. Leave `workflow_call` defaults untouched so programmatic callers that pass find-links are unaffected.

ethanwee1:
Closes #6790

## What

Manual dispatches of the multi-arch PyTorch build workflows commonly fail in two avoidable ways. This PR makes the inputs self-explanatory and gives the Linux path a sane default.

1. **`repository`/`ref` mistaken for the PyTorch source.** They select which **TheRock** checkout the build scripts come from. When set to `rocm/pytorch`, the outer checkout grabs the wrong repo and every step fails with `build_tools/github_actions/*.py: No such file or directory` (shown as the misleading "Executing the custom container implementation failed").
2. **No ROCm package source.** Leaving both `rocm_package_index_url` and `rocm_package_find_links_url` blank makes `pip install rocm[libraries,devel]==<nightly>` resolve against PyPI, where `rocm` is only a placeholder `0.1.0`, so the pinned nightly can't be found.

## Changes

- Clarify `repository`/`ref` descriptions on both build workflows: they refer to **TheRock** (build scripts/workflow), not the PyTorch source; use `pytorch_git_ref` to pick the PyTorch branch.
- Clarify the ROCm index vs. find-links descriptions (use one **or** the other).
- Default `rocm_package_index_url` on the **Linux portable** `workflow_dispatch` path to the nightly multi-arch index (`https://rocm.nightlies.amd.com/whl-multi-arch`) so a bare manual build works.
- `workflow_call` defaults are intentionally left unchanged, so programmatic callers (release/CI wrappers, some of which pass `find-links`) are unaffected.

Windows keeps an empty ROCm-URL default (its packages are not served from the Linux multi-arch index), but its input descriptions get the same clarifications.

## Test

- `pre-commit` (incl. actionlint) passes on both workflow files.
- Docs/config-only change to `workflow_dispatch` metadata; `workflow_call` behavior is unchanged.

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
@@ -109,13 +109,20 @@ on:
         type: string
         required: true
       rocm_package_find_links_url:
-        description: URL for pip --find-links to install ROCm packages (flat page; use this OR rocm_package_index_url)
+        description: >-
+          URL for pip --find-links to install ROCm packages (flat page). Use
+          this OR rocm_package_index_url, not both. If you set this, clear
+          rocm_package_index_url (otherwise both get passed to pip).
         type: string
         default: ""
       rocm_package_index_url:
-        description: URL for pip --index-url to install ROCm packages (PEP 503 index, e.g. repo.amd.com; use this OR rocm_package_find_links_url)
+        description: >-
+          URL for pip --index-url to install ROCm packages (PEP 503 index).
+          Defaults to the nightly multi-arch index; override to build against a
+          different ROCm channel or a specific build's index. Use this OR
+          rocm_package_find_links_url, not both.
         type: string
-        default: ""
+        default: "https://rocm.nightlies.amd.com/whl-multi-arch"
       release_type:
         description: 'Release type; developer-triggered jobs should use "dev".'
         type: choice
@@ -123,11 +130,17 @@ on:
           - dev
         default: dev
       repository:
-        description: "Repository to checkout. Defaults to github.repository."
+        description: >-
+          TheRock repository to check out the build scripts/workflow from (NOT
+          the PyTorch source). Leave blank to use this repository. To build a
+          different PyTorch branch, set pytorch_git_ref instead.
         type: string
         default: ""
       ref:
-        description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
+        description: >-
+          Branch, tag, or SHA of TheRock (the repository above) to check out.
+          Leave blank to use the triggering ref. This does NOT select the
+          PyTorch source ref; use pytorch_git_ref for that.
         type: string
         default: ""
       cache_type:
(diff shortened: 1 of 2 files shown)


.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
```diff
@@ -109,25 +109,38 @@ on:
...
       rocm_package_index_url:
-        description: URL for pip --index-url to install ROCm packages (PEP 503 index, e.g. repo.amd.com; use this OR rocm_package_find_links_url)
+        description: >-
+          URL for pip --index-url to install ROCm packages (PEP 503 index).
+          Defaults to the nightly multi-arch index; override to build against a
+          different ROCm channel or a specific build's index. Use this OR
+          rocm_package_find_links_url, not both.
         type: string
-        default: ""
+        default: "https://rocm.nightlies.amd.com/whl-multi-arch"
```
ScottTodd: "If you set this, clear (some other variable)" doesn't work with inputs that have a default. If you set to empty string then github will revert back to the default.

See here for example: https://github.com/ROCm/TheRock/blob/29faf3b6639bd85ea589a17a447485e24887d37f/.github/workflows/multi_arch_release.yml#L74-L79

.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
ScottTodd: Keep this in sync with Linux please.

```yml
        env:
          # Use the index if given; else the find-links page; else default to
          # the nightly multi-arch index (both empty is the bare-dispatch case).
          ROCM_INDEX_URL: ${{ inputs.rocm_package_index_url || (inputs.rocm_package_find_links_url == '' && 'https://rocm.nightlies.amd.com/whl-multi-arch' || '') }}
          ROCM_FIND_LINKS: ${{ inputs.rocm_package_find_links_url }}
```

```yml
            --index-url="${ROCM_INDEX_URL}" \
            --find-links="${ROCM_FIND_LINKS}" \
```

ScottTodd (approved):

PR #6794, opened 2026-07-22
[math-libs] Fix benchmark build on Windows for primitives libs

amd-hsong:
## Motivation

Binaries of benchmarks for primitives libraries are missing from the Window build. This PR fixes the issue.

JIRA ID: ROCM-27542

## Technical Details

Benchmarks on Windows do not monitor GPUs (no amdsmi dependency). The condition of building them on Windows was coupled with `NOT TARGET amdsmi`. They are now decoupled.

## Test Plan

CI on Windows with `THEROCK_BUILD_TESTING=ON` should now build the binaries and include them in the tarball.

## Test Result

Check the CI and confirm the issue is fixed.

## Submission Checklist

### DIFF
--- a/math-libs/CMakeLists.txt
+++ b/math-libs/CMakeLists.txt
@@ -2,11 +2,17 @@
 # SPDX-License-Identifier: MIT
 
 if(THEROCK_ENABLE_RAND)
+  # On Windows the primlibs benchmarks build without GPU monitoring, so no
+  # amdsmi dependency is needed.
+  if(WIN32)
+    set(_primlibs_benchmark_enable ${THEROCK_BUILD_TESTING})
+    set(_primlibs_benchmark_deps)
+    set(_primlibs_benchmark_includes)
   # amdsmi is Linux-only and only declared when THEROCK_ENABLE_CORE_AMDSMI is ON
   # (see core/CMakeLists.txt). Benchmarks depend on primbench which requires
-  # amd_smi, so they must be disabled when amdsmi is unavailable (e.g., the
+  # amdsmi, so they must be disabled when amdsmi is unavailable (e.g., the
   # prim-isolated CI group THEROCK_ENABLE_ALL=OFF / THEROCK_ENABLE_PRIM=ON).
-  if(WIN32 OR NOT TARGET amdsmi)
+  elseif(NOT TARGET amdsmi)
     set(_primlibs_benchmark_enable "OFF")
     set(_primlibs_benchmark_deps)
     set(_primlibs_benchmark_includes)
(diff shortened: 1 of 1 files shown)


umfranzw (approved):
I'm not a codeowner on this repo, so my approval doesn't have much weight here, but I just wanted to confirm that these changes look correct from the primitives libraries side.

amd-hsong:
@geomin12 @marbre Could you please take a look at this PR - it fixes an urgent build issue? I checked the Multi-Arch CI run https://github.com/ROCm/TheRock/actions/runs/29965485248 and confirmed that the Windows build now contains the missing benchmark executables.

math-libs/CMakeLists.txt
```diff
@@ -2,11 +2,17 @@
...
+    set(_primlibs_benchmark_enable ${THEROCK_BUILD_TESTING})
+    set(_primlibs_benchmark_deps)
+    set(_primlibs_benchmark_includes)
   # amdsmi is Linux-only and only declared when THEROCK_ENABLE_CORE_AMDSMI is ON
   # (see core/CMakeLists.txt). Benchmarks depend on primbench which requires
-  # amd_smi, so they must be disabled when amdsmi is unavailable (e.g., the
+  # amdsmi, so they must be disabled when amdsmi is unavailable (e.g., the
   # prim-isolated CI group THEROCK_ENABLE_ALL=OFF / THEROCK_ENABLE_PRIM=ON).
-  if(WIN32 OR NOT TARGET amdsmi)
+  elseif(NOT TARGET amdsmi)
```
marbre: This seems to be the exact same logic also used later in `THEROCK_ENABLE_PRIM`. We should either use distinct variables here or, what I think might be preferred, is to move this before `THEROCK_ENABLE_RAND` and make the variables available to the entire `math-libs/CMakeLists.txt` and get rid of the code duplication.

Furthermore, it seems we can clean this up and always call
```
set(_primlibs_benchmark_enable ${THEROCK_BUILD_TESTING})
 set(_primlibs_benchmark_deps)
set(_primlibs_benchmark_includes)
```
and than alter those where needed (`elsif(NOT TARGET amdsmi)` and `else()`).

amd-hsong: Fixed in https://github.com/ROCm/TheRock/pull/6794/changes/c57d887d88c1368b5f4a3be03b6ecf2867645f7e.

amd-hsong:
@marbre I have implemented the suggestions in [c57d887](https://github.com/ROCm/TheRock/pull/6794/commits/c57d887d88c1368b5f4a3be03b6ecf2867645f7e). Thanks.

geomin12 (commented):
lgtm but going to let marius review this as he had concerns

amd-hsong:
@marbre Please see my previous comment.

math-libs/CMakeLists.txt
```diff
@@ -1,21 +1,23 @@
...
-  if(WIN32 OR NOT TARGET amdsmi)
+set(_primlibs_benchmark_enable ${THEROCK_BUILD_TESTING})
+set(_primlibs_benchmark_deps)
+set(_primlibs_benchmark_includes)
+# amdsmi is Linux-only and only declared when THEROCK_ENABLE_CORE_AMDSMI is ON
+# (see core/CMakeLists.txt). Benchmarks depend on primbench which requires
+# amdsmi, so they must be disabled when amdsmi is unavailable (e.g., the
+# prim-isolated CI group THEROCK_ENABLE_ALL=OFF / THEROCK_ENABLE_PRIM=ON).
+if(NOT WIN32)
+  if(NOT TARGET amdsmi)
```
marbre, line 12: This misses to disable benchmarking if building for Windows doesn't it?

amd-hsong: Yes, it does. And that is what this PR tries to do - we still want to build benchmark binaries for Windows. Previous logic accidentally disabled it for Windows and this PR fixes it.

amd-hsong:
@marbre ping

math-libs/CMakeLists.txt
marbre, line 12: You may want to flatten but would than also need to replace the inner `else()`.
```suggestion
if(NOT WIN32 AND NOT TARGET amdsmi)
```

This is more a stylistic decision and either works.

math-libs/CMakeLists.txt
```diff
@@ -1,21 +1,23 @@
...
     set(_primlibs_benchmark_enable "OFF")
-    set(_primlibs_benchmark_deps)
-    set(_primlibs_benchmark_includes)
   else()
```
marbre, line 14:
```suggestion
  elseif(NOT WIN32)
```

math-libs/CMakeLists.txt
```diff
@@ -1,21 +1,23 @@
...
-    set(_primlibs_benchmark_enable ${THEROCK_BUILD_TESTING})
     set(_primlibs_benchmark_deps amdsmi)
     set(_primlibs_benchmark_includes therock_primlibs_benchmark_deps.cmake)
   endif()
```
marbre, line 17:
```suggestion
```

math-libs/CMakeLists.txt
marbre: Got it.

marbre (approved):

PR #6810, opened 2026-07-23
[test][do-not-merge] rocm-systems @ HRR Windows teardown fix candidate

andraghetti:
## Summary
Test-only PR (do not merge) to run the Windows gfx110X hip-tests (PAL) against a candidate fix for the intermittent HRR roundtrip teardown hang tracked in ROCm/rocm-systems#8772 (shard 1/4 hangs ~2h; `Unit_HRR_EmbeddedPtrStruct_Direct` finishes but the process never exits, holding the device so later GPU tests time out at 600s).

This points the `rocm-systems` submodule at branch `users/andraghetti/hrr-win-teardown-fix` (`43161da0f6`), which is the same commit the failing bump used (`62bc371`) plus one change: uninstall the crash exception handler in `hip_capture_shutdown()` so it cannot fire during runtime/PAL teardown on Windows.

## Notes
- Not for merge; opened purely to exercise CI. The submodule is intentionally based on an older commit for a clean A/B against the known-hanging state.
- The hang is intermittent, so the PAL shard needs several runs to draw any conclusion; a single green run is not sufficient.

## Test plan
- [ ] Windows gfx110X hip-tests (PAL) shard 1/4 passes across multiple runs


andraghetti:
Closing this test-only draft because rocm-systems can dispatch TheRock Multi-Arch CI directly against the candidate branch, targeted to Windows gfx110X and hip-tests only. That direct run is now in progress.

PR #6812, opened 2026-07-23
fix(TheRock): ppc64le platform support for TheRock build

benrichard-amd:
## Motivation

Part of https://github.com/ROCm/TheRock/issues/5518 non-x86 portability support / upstream ppc64le

## Technical Details

- Do not build `rocprofiler-systems` and `rdc` on non-x86 platforms as they depend on x86-specific libraries (`Dyninst`, `esmi_lib`)
- Do not link `libquadmath` for `Flang` on `ppc64le` as `quadmath.h` is not available on POWER.

## Test Plan

- Verify TheRock build is successful on x86 and ppc64le
- Verify `rocprof-sys` and `rdc` are built on x86

## Test Result

- Builds successfully on x86 and ppc64le
- `rocprof-sys` and `rdc` are built on x86

## Submission Checklist

- [*] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.


cmake/therock_features.cmake
```diff
@@ -120,6 +120,27 @@ function(therock_finalize_features)
...
+
+  # rdc depends on esmi_lib, which is only supported on x86_64
+  # Disable it automatically on other archs.
+  if(THEROCK_ENABLE_RDC AND NOT CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|i.86")
+    message(WARNING
+        "rdc is not supported on ${CMAKE_SYSTEM_PROCESSOR}, "
+        "disabling THEROCK_ENABLE_RDC automatically.")
+    set(THEROCK_ENABLE_RDC OFF CACHE BOOL "" FORCE)
+    set(THEROCK_ENABLE_RDC OFF PARENT_SCOPE)
+  endif()
```
ScottTodd, line 142: I'd prefer for this `therock_finalize_features` function to only contain generic code. We shouldn't have project-specific options leak into these generic functions.

How about adding a `disable_processors` setting to https://github.com/ROCm/TheRock/blob/main/BUILD_TOPOLOGY.toml similar to `disable_platforms`  and using that for rdc and rocprofiler-systems?

That would involve quite a lot of additional boilerplate though.

What would it take to get these projects building on other processors?

benrichard-amd:
Closing. Instead will work to get rdc and rocprofiler-systems building and running on POWER.

PR #6813, opened 2026-07-23
fix(py_packaging): eliminate code injection via exec() on user-controlled input (SEC-00224)

araravik-psd:
## Problem

`py_packaging.py` constructed Python source code by interpolating `target_family` into an f-string. The value is parsed from artifact directory names using the permissive regex in `ArtifactName.from_path`.

This generated source was used in two places:

* The on-disk `_dist_info.py` file.
* An in-memory module created by calling `exec()` on the same interpolated source.

A crafted `target_family` value—such as one supplied through an unusual artifact directory name—could inject arbitrary Python code. That code could execute either on the packaging runner or later, during installation through `setup.py` (SEC-00224, SWSPLAT-28569).

`version` and `version_suffix` are also interpolated in this code. In practice, however, they come from the following CI argument, where the suffix is a run ID:

```bash
--version-suffix "${{ env.ARTIFACT_RUN_ID }}"
```

Therefore, `target_family` is the realistic attack vector—not workflow-dispatch input or a field in `artifact_manifest.json`.

## Fix

This change has two parts:

1. **In-memory `self.dist_info` module**

   `exec()` now runs only the static `_dist_info.py` template, which contains no interpolated values. The `version`, `version_suffix`, and `target_family` values are assigned directly as module attributes afterward, so they are never synthesized into executable source.

2. **On-disk `dist_info_contents`**

   All interpolated values are wrapped with `repr()`. This keeps the generated `_dist_info.py` safe when it is later executed or imported by each package’s `setup.py` during installation.

## Testing

Extended the existing `RestrictFamiliesTest` in `build_tools/tests/py_packaging_test.py`. This test already builds real `Parameters` and `PopulatedDistPackage` objects and executes the generated `_dist_info.py`.

The new tests exercise the actual production path instead of reimplementing the `repr()` and `exec()` behavior inline:

* `test_malicious_version_suffix_is_inert`

  Verifies that a `version_suffix` payload designed to escape a `repr()`-quoted string round-trips as inert data and is not executed.

* `test_malicious_artifact_target_family_is_inert`

  Verifies that a malicious `target_family` value reaching `Parameters` through a real artifact directory name round-trips as inert data and is not executed.

* `test_dist_info_object_matches_generated_file`

  Verifies that the in-memory `dist_info` module and the generated on-disk file agree on every user-controlled field, preventing the two initialization paths from silently drifting apart.

## Out of Scope

A separate, pre-existing shell-interpolation issue exists in the CI workflow YAML. Values such as `${{ inputs.package_version }}` are interpolated directly into `run:` blocks before Bash parses them.

That issue is independent of the `exec()` injection fixed by this PR and is not addressed here. It should be handled in a follow-up.

## References

* JIRA: SWSPLAT-28569
* SEC-00224 (High): `exec()` on f-string-built Python source in the wheel packager
* https://amd-hub.atlassian.net/browse/ROCM-26885


araravik-psd:
Test run locally:

(.venv) arravikum@aravind-azure-testVM:~/TheRock_sec00224/TheRock$ python3 -m pytest build_tools/tests/py_packaging_test.py::DistInfoInjectionTest -v
======================================================= test session starts =======================================================
platform linux -- Python 3.10.12, pytest-8.4.2, pluggy-1.6.0 -- /home/arravikum/TheRock_new/TheRock/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/arravikum/TheRock_sec00224/TheRock/build_tools
configfile: pyproject.toml
collected 4 items

build_tools/tests/py_packaging_test.py::DistInfoInjectionTest::test_default_target_family_injection_neutralized PASSED      [ 25%]
build_tools/tests/py_packaging_test.py::DistInfoInjectionTest::test_normal_version_suffix_unaffected PASSED                 [ 50%]
build_tools/tests/py_packaging_test.py::DistInfoInjectionTest::test_target_family_injection_neutralized PASSED              [ 75%]
build_tools/tests/py_packaging_test.py::DistInfoInjectionTest::test_version_suffix_injection_neutralized PASSED             [100%]

======================================================== 4 passed in 0.03s ========================================================

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,69 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+
+class DistInfoInjectionTest(unittest.TestCase):
+    """Verify that repr() prevents code injection in _dist_info.py generation.
+
+    SEC-00224: user-controlled values (version_suffix, target_family) were
+    previously interpolated directly into exec()-executed source code. The fix
+    wraps all such values in repr() for the on-disk file and assigns them
+    directly as attributes for the in-memory module, eliminating the attack
+    surface entirely.
+    """
```
ScottTodd, line 1681: Please be careful adding unit tests. Coding agents in particular bias heavily towards appending new tests rather than modifying existing tests. Push back _hard_ against that. This code should have existing test coverage already, which can be extended to include test cases for the scenarios this PR is addressing.

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,69 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+
+    def test_version_suffix_injection_neutralized(self):
+        """Malicious version_suffix must not execute when used in generated code."""
+        version = "7.0.0"
+        version_suffix = "'\nimport os; os.system('id')\n#"
+
+        line = f"__version__ = {repr(f'{version}{version_suffix}')}\n"
+        namespace = {}
+        exec(line, namespace)
```
ScottTodd: Avoid logic in tests: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html. The code changes here use `repr` as an **implementation detail** to harden against command injection. The unit tests should not need to be aware of that implementation detail and should not themselves also use `repr`. The tests can attempt command injection and demonstrate how it is thwarted.

build_tools/_therock_utils/py_packaging.py
```diff
@@ -126,12 +126,14 @@ def __init__(
...
+        # All user-controlled values are wrapped in repr() to prevent injection
+        # when the file is later imported by setup.py at install time.
         # Base: version and nonce only — no family lines. Used as the starting
         # point for restrict_families packages so they can write clean family
         # content without a .clear() dance.
         dist_info_base = DIST_INFO_PATH.read_text()
-        dist_info_base += f"__version__ = '{version}'\n"
-        dist_info_base += f"PY_PACKAGE_SUFFIX_NONCE = '{version_suffix}'\n"
+        dist_info_base += f"__version__ = {repr(f'{version}{version_suffix}')}\n"
+        dist_info_base += f"PY_PACKAGE_SUFFIX_NONCE = {repr(version_suffix)}\n"
```
ScottTodd, line 139: There is a behavior change here - previously `__version__` was just `{version}`, now it is `{version}{version_suffix}`. Why? https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_6813.md#1-blocking-preserve-version_suffix-as-a-module-name-nonce

build_tools/_therock_utils/py_packaging.py
```diff
@@ -126,12 +126,14 @@ def __init__(
         self.runtime_artifact_names: set[str] = set()
 
         # Load and interpolate the _dist_info.py template.
+        # All user-controlled values are wrapped in repr() to prevent injection
+        # when the file is later imported by setup.py at install time.
```
ScottTodd: Are these actually "user-controlled" values? I don't think so.

Even if they were, this mitigation doesn't seem sufficient, as some inputs are expanded by github actions in bash prior to even reaching the python script: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_6813.md#3-blocking-close-or-retract-the-claimed-workflow-input-attack-path

build_tools/_therock_utils/py_packaging.py
```diff
@@ -150,23 +152,42 @@ def __init__(
...
+                "rocm-sdk-libraries"
+            )
+        if self.default_target_family is not None:
+            self.dist_info.DEFAULT_TARGET_FAMILY = self.default_target_family
+        for target_family in self.available_target_families:
+            self.dist_info.AVAILABLE_TARGET_FAMILIES.append(target_family)
+        for target_family in self.linux_target_families:
+            self.dist_info.LINUX_TARGET_FAMILIES.append(target_family)
+        for target_family in self.windows_target_families:
+            self.dist_info.WINDOWS_TARGET_FAMILIES.append(target_family)
```
ScottTodd: This is duplicating all the code from a few lines up. Why? https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_6813.md#4-important-keep-one-source-of-truth-for-generated-and-in-memory-state

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,69 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+
+    def test_default_target_family_injection_neutralized(self):
+        """Malicious default_target_family must not execute as an assignment."""
+        default_target_family = "gfx942'\nimport os; os.system('id')\n#"
+
+        line = f"DEFAULT_TARGET_FAMILY = {repr(default_target_family)}\n"
+        namespace = {}
+        exec(line, namespace)
+
+        self.assertEqual(namespace["DEFAULT_TARGET_FAMILY"], default_target_family)
```
ScottTodd: See my other comment... as this is testing is the python built-in `repr` function. It's not actually testing any of the `py_packaging.py`. Unit tests should exercise the code they are meant to test, not the standard library.

ScottTodd (changes requested):
Review from my agent has detailed findings: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_6813.md. I'll comment inline with a few specific issues.

build_tools/tests/py_packaging_test.py
araravik-psd: Fair point, RestrictFamiliesTest already builds real Parameters/PopulatedDistPackage objects and execs the generated _dist_info.py to inspect the result, which is exactly the coverage these tests needed. Folded the 3 tests in there instead of keeping a separate class: 
extended its existing _make_params/_exec_dist_info helpers (optional version/version_suffix override, optional exec namespace) rather than duplicating them. DistInfoInjectionTest is removed entirely in dd8615244.

build_tools/tests/py_packaging_test.py
araravik-psd: I have taken care of this, The test this comment points to (test_version_suffix_injection_neutralized, which built the repr()-quoted line inline and duplicated the implementation under test) was replaced in 5171f07b8. The current test doesn't call repr() at all, it passes a malicious version_suffix through the real Parameters/PopulatedDistPackage constructors, lets production code generate the actual _dist_info.py, and execs that generated file. It also swaps os.system('id') for an in-process SENTINEL dict, so a real regression flips a flag instead of shelling out during a test run.

build_tools/_therock_utils/py_packaging.py
araravik-psd: Updated — the comment now names the specific values (version, version_suffix, target_family) instead of the blanket "user-controlled" framing, and calls out that the practical risk is target_family parsed from artifact directory names via ArtifactName's permissive regex, not workflow inputs. Same change applied to the exec() comment below it. The separate CI-workflow shell-interpolation issue you linked is real but out of scope for this PR, will raise it as a follow-up.

build_tools/_therock_utils/py_packaging.py
araravik-psd: Good catch, and you're right that it was a behavior change — but not an intentional one. It was a bug introduced partway through this PR's own iteration (never present in main), not a deliberate design decision. __version__ is now back to just {version}, with version_suffix isolated exclusively in PY_PACKAGE_SUFFIX_NONCE, fixed in 5171f07b8

build_tools/_therock_utils/py_packaging.py
araravik-psd: Fixed in 5171f07b8 the four .append() loops now .extend() directly from the already-computed self.available_target_families/etc., so there's one source of truth instead of a duplicated copy. test_dist_info_object_matches_generated_file guards against future drift.

build_tools/tests/py_packaging_test.py
araravik-psd: Agreed, same issue as the other comment this test was reimplementing repr()/exec() inline instead of exercising py_packaging.py. It's gone:
 DistInfoInjectionTest (including this test) was replaced in 5171f07b8 with tests that build a real Parameters/PopulatedDistPackage and exec the actual generated _dist_info.py, then folded into RestrictFamiliesTest in dd8615244. test_malicious_artifact_target_family_is_inert now covers this specific case (target_family injection) through the real code path.

araravik-psd (commented):
@ScottTodd thanks for the detailed review. I have addressed the PR comments and added replies inline for the them. Please take a look when you get a chance

ScottTodd (approved):
Thanks for the updates, particularly to the test cases. Code looks good to me now.

PR #6816, opened 2026-07-23
feat(ci): Support both count and weight for runner selection

geomin12:
Update select_weighted_label to support both selection methods:
- count: Integer representing number of available runners (preferred)
- weight: Float 0.0-1.0 probability (fallback for backwards compatibility)

Test runner labels use count-based selection (e.g., count: 5, 28, 5).
Build runner labels continue to use weight-based selection (weight: 1.0).

Changes:
- select_weighted_label: Check for count first, fallback to weight
- amdgpu_family_info_matrix: gfx94x test labels use count
- BUILD_RUNNER_LABELS: Keep weight-based (unchanged)
- Tests updated for both count and weight scenarios


build_tools/github_actions/amdgpu_family_matrix.py
```diff
@@ -79,7 +79,36 @@ def is_asan():
 
 
 def select_weighted_label(labels_config: list[dict], context_name: str) -> str:
-    """Select a runner label based on weighted random selection."""
+    """Select a runner label based on random selection.
```
HereThereBeDragons: have a look at  #3653 `build_tools/github_actions/new_amdgpu_family_matrix_runners.py:_select_weighted_label()`
you can simplify this to:

```
def _select_weighted_label(entries: list[_RunnerEntry]) -> str:
    """Weighted random pick across a non-empty pool. Solo entries are returned
    as-is regardless of weight; weights are relative and need not sum to 1.0."""
    if len(entries) == 1:
        return entries[0].label
    return random.choices(entries, weights=[e.weight for e in entries], k=1)[0].label
```

please also double check but this should also work with the percentage based ones. you just might want to add a check "needs to sum up to 1.0"

rocm-libraries
HereThereBeDragons, line 1: seems like this got accidentally added?

HereThereBeDragons (approved):
sorry didnt see that you had requested it already a couple of days ago

PR #6822, opened 2026-07-23
[compiler] Deprecate hipcc: use amdclang++ directly in amd-hip toolchain

david-salinas:
## Motivation

ROCM-8370 : [hipcc] Remove HIPCC use from all TheRock components
 https://amd-hub.atlassian.net/browse/ROCM-8370

Second phase of the plan to deprecate HIPCC, removes uses of hipcc in TheRock builds in remaining components.
Calls to hipcc will be replaced with calls to amdcalng++ with appropriate compiler options.

## Technical Details

- cmake/therock_subproject.cmake: set CMAKE_HIP_COMPILER to amdclang++ and inject --rtlib=compiler-rt --unwindlib=libgcc linker flags via CMAKE_EXE_LINKER_FLAGS_INIT and CMAKE_SHARED_LINKER_FLAGS_INIT in the generated amd-hip toolchain file. This replaces the equivalent flags that hipcc provided as a wrapper.

- tests/hipcc_check.cpp -> tests/hip_check.cpp: rename and update the ROCm sanity test to compile with amdclang++ directly (with explicit --hip-path, --hip-device-lib-path, and -x hip flags) instead of using the hipcc wrapper.

## Test Plan

Full CI/CQE Staging 

## Submission Checklist


cmake/therock_subproject.cmake
```diff
@@ -1742,8 +1742,8 @@ function(_therock_cmake_subproject_setup_toolchain
     # we commingle them, but they are different:
     #   "amd-llvm": Just the base LLVM compiler and device libraries. This
     #     doesn't know anything about hip (i.e. it doesn't have hipconfig, etc).
-    #   "amd-hip": Superset of "amd-llvm" which also includes hipcc, hip headers,
-    #     and hip version info. This has hipconfig in it.
+    #   "amd-hip": Superset of "amd-llvm" which also includes HIP headers and
+    #     HIP version info. Uses amdclang++ directly; does not require hipcc.
```
lamb-j:
```suggestion
    #     HIP version info.
```

cmake/therock_subproject.cmake
```diff
@@ -1813,6 +1813,9 @@ function(_therock_cmake_subproject_setup_toolchain
     list(APPEND _compiler_toolchain_addl_depends "${_hip_stamp_dir}/stage.stamp")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-path=@_hip_dist_dir@\")\n")
     string(APPEND _toolchain_contents "string(APPEND CMAKE_CXX_FLAGS_INIT \" --hip-device-lib-path=@_amd_llvm_device_lib_path@\")\n")
+    string(APPEND _toolchain_contents "set(CMAKE_HIP_COMPILER \"@AMD_LLVM_CXX_COMPILER@\")\n")
+    string(APPEND _toolchain_contents "string(APPEND CMAKE_EXE_LINKER_FLAGS_INIT \" --rtlib=compiler-rt --unwindlib=libgcc\")\n")
```
lamb-j: What's the libunwind add here needed for?

tests/test_rocm_sanity.py
```diff
@@ -129,28 +125,33 @@ def test_hip_printf(self):
             offload_arch is not None
         ), f"Expected offload-arch to return gfx####, got:\n{process.stdout}"
 
-        # Compiling .cpp file using hipcc
-        hipcc_check_executable_file = f"hipcc_check{platform_executable_suffix}"
+        # Compiling .cpp file using amdclang++ directly (hipcc is deprecated)
```
lamb-j:
```suggestion
        # Compiling .cpp file using amdclang++
```

tests/test_rocm_sanity.py
```diff
@@ -129,28 +125,33 @@ def test_hip_printf(self):
...
+                hip_check_executable_file,
             ],
             cwd=str(THEROCK_BIN_DIR),
         )
 
         # Running and checking the executable
         platform_executable_prefix = "./" if not is_windows() else ""
-        hipcc_check_executable = f"{platform_executable_prefix}hipcc_check"
-        process = run_command([hipcc_check_executable], cwd=str(THEROCK_BIN_DIR))
+        hip_check_executable = f"{platform_executable_prefix}hip_check"
```
lamb-j, line 146: should these be amdclang_check_*?

cmake/therock_subproject.cmake
david-salinas: the unwind was only needed because we had added --rtlib=compiler-rt.  But we don't need either.

cmake/therock_subproject.cmake
david-salinas: ok

tests/test_rocm_sanity.py
david-salinas: ok

tests/test_rocm_sanity.py
david-salinas: The test source does actually utilize basic HIP RT functions, so it's more than just testing if amdclang exists and can compile.

tests/test_rocm_sanity.py
```diff
@@ -82,9 +82,9 @@ def test_rocm_output(self, rocm_info_output, to_search):
             f"Failed to search for {to_search} in rocminfo output",
         )
 
-    # TODO(#3313): Re-enable once hipcc test is fixed for ASAN builds
+    # TODO(#3313): Re-enable once amdclang++ test is fixed for ASAN builds
     @pytest.mark.skipif(
-        is_asan(), reason="hipcc test fails with ASAN build, see TheRock#3313"
+        is_asan(), reason="amdclang++ test fails with ASAN build, see TheRock#3313"
```
lamb-j: If this test is updated to use amdclang instead of hipcc (which is should be), can we flip the test back on and see if it passes?

david-salinas: ok lets try it.  I'll push a commit to enable and see how the CI jobs do.

david-salinas:
@lamb-j ping.  I think this PR is ready to land.

tests/test_rocm_sanity.py
```diff
@@ -82,18 +82,15 @@ def test_rocm_output(self, rocm_info_output, to_search):
             f"Failed to search for {to_search} in rocminfo output",
         )
 
-    # TODO(#3313): Re-enable once hipcc test is fixed for ASAN builds
+    # TODO(#4755): Re-enable test for windows once offload-arch.exe is fixed
```
lamb-j, line 85: We fixed this last week it looks like. So can we turn this back on now?

lamb-j: (It can be done as a separate PR)

lamb-j (approved):

PR #6828, opened 2026-07-23
feat(ci): Enabling external repos for artifact re-use

geomin12:
The branch enables external repos (rocm-libraries, rocm-systems) to copy prebuilt artifacts from TheRock's S3 bucket via:
  1. stage_reuse_baseline_repository input to query TheRock for baselines
  2. THEROCK_SOURCE_REPOSITORY env var for artifact copy operations

Working here for [rocm-libraries](https://github.com/ROCm/rocm-libraries/actions/runs/30043559586?pr=9741) and [rocm-systems](https://github.com/ROCm/rocm-systems/actions/runs/30043587313?pr=9060)

Next: 
- Enable non-blocking pre-submit for multi-arch CI in superrepos
- Determine "good baseline run ID" from TheRock dynamically instead of hard-coded
- Dynamic determination of what stages to copy vs what to rebuild (currently hard-coded)
- Enable "granular" re-use of artifacts (ex: BLAS update, only build BLAS, copy the remaining artifacts such as FFT, PRIM, etc) as no need to rebuild

ISSUE ID: #3343

### DIFF
--- a/.github/workflows/multi_arch_ci.yml
+++ b/.github/workflows/multi_arch_ci.yml
@@ -61,6 +61,10 @@ on:
         type: string
         default: "50"
         description: "Automatic stage reuse: number of recent branch commits to fetch for ancestry"
+      stage_reuse_baseline_repository:
+        type: string
+        default: ""
+        description: "Repository to query for baseline runs (default: current repo). Set to 'ROCm/TheRock' for external repos."
       changed_projects:
         type: string
         default: ""
@@ -108,6 +112,7 @@ jobs:
       stage_reuse_mode: ${{ inputs.stage_reuse_mode || 'dry-run' }}
       stage_reuse_max_age_hours: ${{ inputs.stage_reuse_max_age_hours || '72' }}
       stage_reuse_commit_history: ${{ inputs.stage_reuse_commit_history || '50' }}
+      stage_reuse_baseline_repository: ${{ inputs.stage_reuse_baseline_repository || '' }}
       build_pytorch: ${{ github.event_name != 'workflow_dispatch' || inputs.build_pytorch }}
       build_jax: ${{ github.event_name != 'workflow_dispatch' || inputs.build_jax }}
 
--- a/.github/workflows/multi_arch_ci.yml
+++ b/.github/workflows/multi_arch_ci.yml
@@ -61,10 +61,6 @@ on:
         type: string
         default: "50"
         description: "Automatic stage reuse: number of recent branch commits to fetch for ancestry"
-      stage_reuse_baseline_repository:
-        type: string
-        default: ""
-        description: "Repository containing baseline_run_id; defaults to the current repository."
       changed_projects:
         type: string
         default: ""
@@ -112,7 +108,6 @@ jobs:
       stage_reuse_mode: ${{ inputs.stage_reuse_mode || 'dry-run' }}
       stage_reuse_max_age_hours: ${{ inputs.stage_reuse_max_age_hours || '72' }}
       stage_reuse_commit_history: ${{ inputs.stage_reuse_commit_history || '50' }}
-      stage_reuse_baseline_repository: ${{ inputs.stage_reuse_baseline_repository || '' }}
       build_pytorch: ${{ github.event_name != 'workflow_dispatch' || inputs.build_pytorch }}
       build_jax: ${{ github.event_name != 'workflow_dispatch' || inputs.build_jax }}
 
--- a/.github/workflows/multi_arch_ci.yml
+++ b/.github/workflows/multi_arch_ci.yml
@@ -64,7 +64,7 @@ on:
       stage_reuse_baseline_repository:
         type: string
         default: ""
-        description: "Repository to query for baseline runs (default: current repo). Set to 'ROCm/TheRock' for external repos."
+        description: "Repository containing baseline_run_id; defaults to the current repository."
       changed_projects:
         type: string
         default: ""
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -86,8 +86,16 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
           RELEASE_TYPE: ${{ inputs.release_type }}
         run: |
+          # Build the command with optional --source-repository for cross-repo copies
+          SOURCE_REPO="${{ fromJSON(inputs.build_config).baseline_repository }}"
+          SOURCE_REPO_ARG=""
+          if [ -n "$SOURCE_REPO" ]; then
+            SOURCE_REPO_ARG="--source-repository=$SOURCE_REPO"
+          fi
+
           python build_tools/artifact_manager.py copy \
             --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} \
+            $SOURCE_REPO_ARG \
             --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" \
             --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" \
             --expand-family-to-targets
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -85,17 +85,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
           RELEASE_TYPE: ${{ inputs.release_type }}
+          THEROCK_SOURCE_REPOSITORY: ${{ fromJSON(inputs.build_config).baseline_repository }}
         run: |
-          # Build the command with optional --source-repository for cross-repo copies
-          SOURCE_REPO="${{ fromJSON(inputs.build_config).baseline_repository }}"
-          SOURCE_REPO_ARG=""
-          if [ -n "$SOURCE_REPO" ]; then
-            SOURCE_REPO_ARG="--source-repository=$SOURCE_REPO"
-          fi
-
           python build_tools/artifact_manager.py copy \
             --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} \
-            $SOURCE_REPO_ARG \
             --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" \
             --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" \
             --expand-family-to-targets
--- a/.github/workflows/multi_arch_ci_linux.yml
+++ b/.github/workflows/multi_arch_ci_linux.yml
@@ -85,10 +85,10 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
           RELEASE_TYPE: ${{ inputs.release_type }}
-          THEROCK_SOURCE_REPOSITORY: ${{ fromJSON(inputs.build_config).baseline_repository }}
         run: |
           python build_tools/artifact_manager.py copy \
             --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} \
+            --source-repository=${{ fromJSON(inputs.build_config).baseline_repository }} \
             --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" \
             --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" \
             --expand-family-to-targets
--- a/.github/workflows/multi_arch_ci_windows.yml
+++ b/.github/workflows/multi_arch_ci_windows.yml
@@ -83,21 +83,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
           RELEASE_TYPE: ${{ inputs.release_type }}
+          THEROCK_SOURCE_REPOSITORY: ${{ fromJSON(inputs.build_config).baseline_repository }}
         run: |
-          # Build the command with optional --source-repository for cross-repo copies
-          $SOURCE_REPO = "${{ fromJSON(inputs.build_config).baseline_repository }}"
-          $SOURCE_REPO_ARG = ""
-          if ($SOURCE_REPO) {
-            $SOURCE_REPO_ARG = "--source-repository=$SOURCE_REPO"
-          }
-
-          python build_tools/artifact_manager.py copy `
-            --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} `
-            $SOURCE_REPO_ARG `
-            --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" `
-            --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" `
+          python build_tools/artifact_manager.py copy \
+            --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} \
+            --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" \
+            --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" \
             --expand-family-to-targets
-        shell: pwsh
 
   build_multi_arch_stages:
     name: Build Multi-Arch Stages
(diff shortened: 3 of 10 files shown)


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -883,10 +891,15 @@ def decide_jobs(
     if auto_stage_reuse.applied_reuse_stages and auto_stage_reuse.baseline_run_id:
         baseline_run_id = auto_stage_reuse.baseline_run_id
 
+    # For cross-repo artifact reuse, pass through the baseline_repository so
+    # the copy job knows which repository to fetch artifacts from.
+    baseline_repository = ci_inputs.baseline_repository
```
erman-gurses, line 887: Can we ensure that baseline_run_id and baseline_repository always refer to the same repository?

When automatic reuse is enabled, auto_stage_reuse.baseline_run_id can replace the manually supplied run ID, but _default_baseline_selector() still queries GITHUB_REPOSITORY. We then unconditionally pair that run ID with ci_inputs.baseline_repository.

For an external repo using stage_reuse_baseline_repository: "ROCm/TheRock" and stage_reuse_mode: "reuse-stage", this could select a run from the external repository and then try to look up that run ID in ROCm/TheRock.

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -504,5 +504,89 @@ def test_both_platforms(self):
         )
 
 
+class CrossRepoBaselineSelectorTest(unittest.TestCase):
+    """External repos (rocm-libraries, rocm-systems) can use TheRock's baselines.
+
+    When STAGE_REUSE_BASELINE_REPOSITORY is set to a different repo than
+    GITHUB_REPOSITORY, commit compatibility checking is disabled because
+    the two repos have unrelated commit histories.
+    """
```
erman-gurses: CAn we replace these tests with coverage of the behavior added by the final version of this PR?

This class describes cross-repo baseline selection, but neither test supplies a non-empty STAGE_REUSE_BASELINE_REPOSITORY, and the final diff no longer changes _default_baseline_selector(). These tests would pass even if the new repository value were never propagated to the artifact copy operation.

We might instead add regression tests verifying that:

- --source-repository takes precedence over THEROCK_SOURCE_REPOSITORY.
- The environment variable is used when the flag is empty.
- The resolved repository is passed to WorkflowOutputRoot.from_workflow_run().
- baseline_repository is preserved in the generated Linux and Windows build configs?

erman-gurses (commented):
I did my initial review for that -  will go another pass

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -879,14 +887,30 @@ def decide_jobs(
     # Only reuse-stage mode returns non-empty applied_reuse_stages.
     for stage in auto_stage_reuse.applied_reuse_stages:
         stage_decisions.setdefault(stage, JobAction.PREBUILT)
+
+    # For cross-repo artifact reuse, pass through the baseline_repository so
+    # the copy job knows which repository to fetch artifacts from.
+    baseline_repository = ci_inputs.baseline_repository
```
erman-gurses: Thanks, this prevents the auto-selected run ID from being paired with a different repository.

Could we also disable applying auto_stage_reuse.applied_reuse_stages for cross-repo reuse?

Those stages are currently merged into stage_decisions before the is_cross_repo check. If automatic analysis returns stages, the stage list could be verified against a run from GITHUB_REPOSITORY, while the copy operation uses the manually supplied run from baseline_repository.

Since automatic cross-repo selection is deferred to a follow-up, could we skip the entire automatic reuse result when is_cross_repo is true, rather than only preventing the run ID override?

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -504,5 +504,148 @@ def test_both_platforms(self):
         )
 
 
+class CrossRepoArtifactCopyTest(unittest.TestCase):
+    """Test cross-repo artifact copying plumbing.
+
+    External repos (rocm-libraries, rocm-systems) can copy artifacts from
+    TheRock's baseline runs. This tests that the source_repository parameter
+    flows correctly through the artifact copy operation.
+    """
```
erman-gurses: We might updating these tests to exercise the production code directly rather than repeating the implementation logic?

The current flag/env precedence tests mirror the expression in do_copy(), and the cross-repo tests mirror the condition in decide_jobs(). This means they could still pass even if the production behavior changed or was removed.

It may be more robust to call the real functions with mocks to verify repository precedence, forwarding to WorkflowOutputRoot, cross-repo behavior, and propagation into the generated build configs. The artifact-manager-specific coverage might also fit better in its own test module.

.github/workflows/multi_arch_ci.yml
```diff
@@ -61,6 +61,10 @@ on:
         type: string
         default: "50"
         description: "Automatic stage reuse: number of recent branch commits to fetch for ancestry"
+      stage_reuse_baseline_repository:
+        type: string
+        default: ""
+        description: "Repository to query for baseline runs (default: current repo). Set to 'ROCm/TheRock' for external repos."
```
erman-gurses: I think we should update this description to match the current behavior.

`stage_reuse_baseline_repository` identifies the repository containing the manually supplied `baseline_run_id`; automatic selection still uses `GITHUB_REPOSITORY`.

Maybe: “Repository containing `baseline_run_id`; defaults to the current repository.”

erman-gurses (commented):
Some additional comments

.github/workflows/setup_multi_arch.yml
```diff
@@ -60,6 +60,13 @@ on:
           Automatic stage reuse: number of recent branch commits to fetch for ancestry. default: "50"
         type: string
         default: "50"
+      stage_reuse_baseline_repository:
+        description: >-
+          Repository to query for baseline workflow runs (default: current repository).
+          Set to "ROCm/TheRock" for external repos (rocm-libraries, rocm-systems) to
+          reuse TheRock's prebuilt artifacts.
+        type: string
+        default: ""
```
erman-gurses: Could we update this description to match the one in multi_arch_ci.yml?

erman-gurses (approved):
LGTM - Please see my last comment.

ammallya (approved):
LGTM I think Erman addressed the concerns well

geomin12:
this is actually a fairly minor change and aligns with #3343, going to land and continue on #3343

ScottTodd:
> this is actually a fairly minor change and aligns with #3343, going to land and continue on #3343

Please wait for my review. I'm also going to add myself to CODEOWNERS for some of these files.

geomin12:
> > this is actually a fairly minor change and aligns with #3343, going to land and continue on #3343
> 
> Please wait for my review. I'm also going to add myself to CODEOWNERS for some of these files.

oh okay cool! thanks for adding yourself back

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -876,17 +884,31 @@ def decide_jobs(
...
+
+    # Determine if this is cross-repo reuse. When baseline_repository differs
+    # from GITHUB_REPOSITORY, automatic stage reuse is not applicable because
+    # it queries the current repo, not the baseline repo.
+    current_repo = os.environ.get("GITHUB_REPOSITORY", "")
+    is_cross_repo = baseline_repository and baseline_repository != current_repo
+
+    # Only apply automatic stage reuse for same-repo scenarios.
+    # Cross-repo reuse uses manually supplied baseline_run_id and prebuilt_stages.
```
ScottTodd: This seems like an odd condition to enforce? Can you expand on the comment here?

Artifact reuse is a form of caching. What we really care about is that the source code on a given branch produces (or uses prebuilt) the artifacts that correspond to that code. Is there something particularly unique about same-repo vs cross-repo that makes this harder?

build_tools/artifact_manager.py
```diff
@@ -892,7 +892,10 @@ def copy_single_artifact(request: CopyRequest) -> bool:
 
 
 def _create_source_backend(
-    source_run_id: str, platform: str, local_staging_dir: Optional[Path] = None
+    source_run_id: str,
+    platform: str,
+    local_staging_dir: Optional[Path] = None,
+    source_repository: Optional[str] = None,
 ) -> ArtifactBackend:
```
ScottTodd, line 899: nit: group related arguments. `source_run_id` should be next to `source_repository`. It's fine to make it required to not mix optional and required args.

build_tools/artifact_manager.py
```diff
@@ -948,10 +954,17 @@ def do_copy(args: argparse.Namespace):
     target_families = parse_target_families(args)
 
     # Create source and dest backends
+    # --source-repository flag takes precedence, then THEROCK_SOURCE_REPOSITORY env var
+    source_repository = (
+        getattr(args, "source_repository", None)
+        or os.environ.get("THEROCK_SOURCE_REPOSITORY")
+        or None
+    )
```
ScottTodd: Follow the style guide: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#access-argparse-attributes-directly (no `getattr` when working with argparse)

A better way to handle this would be to set `default=` on the new argument, so the flag gets the expected value at the source, rather than later in only this one function. Similar to https://github.com/ROCm/TheRock/blob/64a68c4fd35579a580f0404a460cce537ae91481/build_tools/artifact_manager.py#L1124-L1129

.github/workflows/setup_multi_arch.yml
```diff
@@ -60,6 +60,10 @@ on:
           Automatic stage reuse: number of recent branch commits to fetch for ancestry. default: "50"
         type: string
         default: "50"
+      stage_reuse_baseline_repository:
+        description: "Repository containing baseline_run_id; defaults to the current repository."
+        type: string
+        default: ""
```
ScottTodd: This list of `stage_reuse_*` arguments in `setup_multi_arch.yml` is growing pretty long.

* Are workflows actually going to be providing explicit input values here that couldn't be discovered implicitly from the environment?
* Why do these parameters matter? They look like they should be all implementation details of the CI/build system and not dynamic values using all this plumbing. If/when we get to the point where we have a central database managing artifact reuse (essentially a cache server), some of these parameters would be a part of that.
  * For example, let's say the fingerprint for the amd-llvm artifact on a given commit is `abcdef12346`. If I ask the cache server for artifacts matching that fingerprint, I don't care which run (from a github actions run in some repository or otherwise) produced those artifacts. I also wouldn't care how old those artifacts are.

Actually building such a cache server with centralized artifact fingerprints for reuse will take more planning and work, but it will scale much more efficiently than the pure "reuse artifacts from a workflow run in this repository" that was initially built into these workflows. I don't want to go too far down the path of plumbing workflow inputs around and setting up scripts with lookup behavior and other heuristics if that paints us into too much of a corner, and requiring all these extra workflow inputs takes some steps in that direction.

.github/workflows/multi_arch_ci_linux.yml
ScottTodd, line 94: This is running a script so it can directly pass the `source_repository` argument instead of using an environment variable:

```diff
        env:
          GITHUB_TOKEN: ${{ github.token }}
          RELEASE_TYPE: ${{ inputs.release_type }}
-         THEROCK_SOURCE_REPOSITORY: ${{ fromJSON(inputs.build_config).baseline_repository }}
        run: |
          python build_tools/artifact_manager.py copy \
            --source-run-id=${{ fromJSON(inputs.build_config).baseline_run_id }} \
+           --source-repository=${{ fromJSON(inputs.build_config).baseline_repository }} \
            --stage="${{ fromJSON(inputs.build_config).prebuilt_stages }}" \
            --amdgpu-families="${{ fromJSON(inputs.build_config).dist_amdgpu_families }}" \
            --expand-family-to-targets
```

Relating to some of my other comments, I'm also a bit concerned about having three potentially different repository variables: `GITHUB_REPOSITORY`, `inputs.repository`, and now also `baseline_repository` on the build config. That's somewhat unavoidable with the current build and CI architecture, since artifact reuse is done via this upload and then skipping entire build stages. I don't see an easy way around that with the current build system.

.github/workflows/multi_arch_ci_windows.yml
ScottTodd: Same comment here, use `--source-repository` and not an env var for the plumbing

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -504,5 +504,128 @@ def test_both_platforms(self):
...
+    """Test cross-repo artifact copying plumbing.
+
+    External repos (rocm-libraries, rocm-systems) can copy artifacts from
+    TheRock's baseline runs. These tests call production code to verify
+    the cross-repo behavior.
+    """
+
+    def test_baseline_repository_preserved_in_build_config(self):
+        """baseline_repository is preserved in BuildRocmDecision."""
+        import configure_multi_arch_ci as cma
```
ScottTodd: Follow the style guide: imports go at the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -504,5 +504,128 @@ def test_both_platforms(self):
...
+
+        build_rocm = cma.BuildRocmDecision(
+            action=cma.JobAction.RUN,
+            stage_decisions={},
+            baseline_run_id="12345",
+            baseline_repository="ROCm/TheRock",
+        )
+
+        self.assertEqual(build_rocm.baseline_repository, "ROCm/TheRock")
+        self.assertEqual(build_rocm.baseline_run_id, "12345")
```
ScottTodd: This is a lot of test code for a single check that a property is tracked on a dataclass. I'd rather test meaningful behavior.

Same comment for `test_ci_inputs_stores_baseline_repository`.

I would expect existing test cases that use `BuildRocmDecision` and `CIInputs` to be able to add a `self.assertEqual(ci_inputs.baseline_repository` check.

(In general please watch for AI generated test cases always appending new tests and not modifying any existing tests)

build_tools/github_actions/tests/stage_reuse_decision_test.py
```diff
@@ -504,5 +504,128 @@ def test_both_platforms(self):
...
+
+    def test_cross_repo_skips_auto_stage_reuse(self):
+        """Cross-repo reuse skips automatic stage decisions."""
+        import os
+        import configure_multi_arch_ci as cma
+
+        old_github_repo = os.environ.get("GITHUB_REPOSITORY")
+        os.environ["GITHUB_REPOSITORY"] = "ROCm/rocm-libraries"
+
+        try:
```
ScottTodd: nit: I'd prefer this style of env var overriding: https://github.com/ROCm/TheRock/blob/d383661dfdf6eaff76a60297c1757ec678780102/build_tools/github_actions/tests/configure_multi_arch_ci_test.py#L1828-L1836

that lets the tests stay focused without needing to more carefully manage this `try: finally` in the individual test cases

.github/workflows/setup_multi_arch.yml
geomin12: oh i updated this! i actually think the artifacts will come from repository input

build_tools/artifact_manager.py
```diff
@@ -948,10 +954,12 @@ def do_copy(args: argparse.Namespace):
     target_families = parse_target_families(args)
 
     # Create source and dest backends
+    source_repository = args.source_repository or None
     source_backend = _create_source_backend(
         source_run_id=args.source_run_id,
         platform=args.platform,
         local_staging_dir=args.local_staging_dir,
+        source_repository=source_repository,
     )
```
ScottTodd, line 962: ? Why use a local variable? Just pass it directly without the `or None`? Trust argparse and let the default propagate down to the called function

build_tools/artifact_manager.py
ScottTodd: This comment is still not addressed.

ScottTodd (changes requested):
Please actually read the feedback and respond.
* https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#ai-tool-use-policy
* https://tombedor.dev/human-attention-and-human-effort/

build_tools/artifact_manager.py
geomin12: thanks for patience, i was definitely running on fumes while doing this code but being stubborn and trying to finish it up

ScottTodd (approved):
Thanks for the updates. Code looks cleaner now.

PR #6847, opened 2026-07-24
Support extra pip index URL for ROCm deps in PyTorch wheel tests

ethanwee1:
## Problem

PyTorch wheel tests install torch from a single `--index-url`. When a wheel is built against a released ROCm, `rocm==<version>` is not on that torch index, so venv setup fails.

## Changes

- Add optional `rocm_package_index_url` (`pip --extra-index-url`) to `test_pytorch_wheels.yml` and `test_pytorch_wheels_full.yml`. Empty default: no behavior change.
- Example extra index is the deps-only URL `https://stable.repo.amd.com/rocm/core/whl-next/`.
- Thread the input through Linux portable and Windows `multi_arch_build_*_pytorch_wheels.yml` test dispatches.
- `build_tools/setup_venv.py` accepts `--extra-index-url`.

Does not change the weekly nightly full-test cadence gate.

## Validation

Rebased onto current `main` (AIPYTORCH-1210). No wheel test run on this rebase.

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
@@ -441,6 +441,9 @@ jobs:
       amdgpu_family: ${{ matrix.amdgpu_family }}
       test_runs_on: ${{ matrix.test_runs_on }}
       package_index_url: ${{ needs.build_pytorch_wheels.outputs.package_index_url }}
+      # Resolve ROCm runtime deps (e.g. a released ROCm not on the torch index)
+      # from the same index the wheels were built against.
+      rocm_package_index_url: ${{ inputs.rocm_package_index_url }}
       python_version: ${{ inputs.python_version }}
       torch_version: ${{ needs.build_pytorch_wheels.outputs.torch_version }}
       # TODO(#5110): pass manifest_url here and let test_pytorch_wheels.yml use
@@ -492,6 +495,7 @@ jobs:
               "test_runs_on": "linux-gfx942-1gpu-ossci-rocm",
               "test_runs_on_multi_gpu": "linux-gfx942-8gpu-ossci-rocm",
               "package_index_url": "${{ needs.build_pytorch_wheels.outputs.package_index_url }}",
+              "rocm_package_index_url": "${{ inputs.rocm_package_index_url }}",
               "python_version": "${{ inputs.python_version }}",
               "torch_version": "${{ needs.build_pytorch_wheels.outputs.torch_version }}",
               "pytorch_git_ref": "${{ inputs.pytorch_git_ref }}",
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -21,6 +21,15 @@ on:
         required: true
         type: string
         default: "https://rocm.nightlies.amd.com/whl-multi-arch/"
+      rocm_package_index_url:
+        description: >-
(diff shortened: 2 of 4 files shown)


.github/workflows/multi_arch_build_portable_linux_pytorch_wheels.yml
ScottTodd, line 1: Linux and Windows workflows should always be kept in sync.

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -21,6 +21,15 @@ on:
         required: true
         type: string
         default: "https://rocm.nightlies.amd.com/whl-multi-arch/"
+      rocm_package_index_url:
+        description: >-
+          Optional additional package index (pip --extra-index-url) used to
+          resolve ROCm runtime deps (e.g. rocm==7.14.0) that are not published
+          to package_index_url. Set to a released ROCm index (e.g.
+          https://repo.amd.com/rocm/manylinux/rocm-rel-7.14/) when testing
```
ScottTodd, line 29: This https://repo.amd.com/rocm/manylinux/rocm-rel-7.14/ URL does not exist. Did you mean https://repo.amd.com/rocm/whl-multi-arch/ or some other URL?

.github/workflows/test_pytorch_wheels.yml
```diff
@@ -21,6 +21,15 @@ on:
...
+          wheels built against a released ROCm.
```
ScottTodd, line 30: I'm not convinced that we should support this. We currently expect each release index to be self-contained and the tests prove that `pip install torch` is able to successfully pull dependent `rocm` packages from the same index.

Mixing release indices and types is unstable territory, so we provide https://github.com/ROCm/TheRock/blob/main/.github/workflows/copy_release.yml to pull nightly release package sfrom the nightly release index into the dev release index. See the docs at https://github.com/ROCm/TheRock/blob/main/docs/development/github_actions_debugging.md#testing-pytorch-release-workflows.

Adding `--extra-index-url` here would let you:
* Build a "dev" release against "nightly"/"prerelease"/"release" packages from their corresponding index
* Test that release with `--index-url=...dev... --extra-index-url=...nightly...`

but then the packages could _only_ be installed with that extra index url.

---

Here are some alternate approaches to consider:

1.Copy into the dev release index automatically
    * For the "dev" releases type, copy deps from the requested input index into the dev release index automatically (instead of requiring a separate run of copy_release.yml)
    * Then the build and test steps can operate directly on the dev release index without needing to pass `--extra-index-url` and users can also install packages directly
2. Create separate workflows and/or buckets specifically for testing different pytorch configurations, to limit the impact of different package/index combinations.
3. For test runs, copy packages into a workflow-scoped index (see https://github.com/ROCm/TheRock/blob/main/docs/development/workflow_outputs.md and the `python/` folder) where we can mix package types however we want.

@marbre thoughts?

marbre: @ethanwee1, @ScottTodd, catching up here finally after we refreshed repo.amd.com. With the new structure, dependencies are home at e.g. https://stable.repo.amd.com/rocm/core/whl-next/ (same for every stream). I have some work to make sure that all streams serve the same version of the dependencies. What we could further do is to move dependencies from `rocm/core/whl-next/` to `rocm/third-party/whl` for example but this needs to be aligned with @saadrahim.

With that we would offer an "extra index" that only serves dep without the risk to pull a ROCm or torch package. The aggregated indexes still serve everything for our entire set of packages.

Issue #6711, opened 2026-07-20
Introduce a TESTING.md or docs/development/testing.md file documenting our testing strategy


ScottTodd:
This will serve as then authoritative page for our testing strategy and philosophy. This should cover how changes to TheRock are validated, maintained, and evolved over time.

We'd like to add similar TESTING.md pages to https://github.com/ROCm/rocm-libraries and https://github.com/ROCm/rocm-systems subprojects, so what we add here in TheRock can serve as both an example to follow and an integration point where the code all comes together into releases.

Scope for testing is different across areas of TheRock, so I think we can treat these like "subprojects" or "components" to some degree:
* The CMake build system (.cmake, .txt, artifacts)
  * Submodule updates run our typical CI workflows with "all archs" tested
  * Other changes may want to opt-in to testing with certain gfx archs
* Packaging (python, native linux, native windows, tarballs) and package publishing / promotion
  * These have some unit tests and workflows that build and test them on CI, could note potential gaps with testing across _all_ supported environments (python versions, OS distros, etc.)
* GitHub Actions workflows
  * Mention workflows that run on `pull_request` and those that run on `schedule` as an area to be particularly careful (especially where workflows are used by https://github.com/ROCm/rockrel and other repositories too)
* Python scripts
  * These have associated unit tests that run on the CPU locally and on CI via https://github.com/ROCm/TheRock/blob/main/.github/workflows/unit_tests.yml (which is a required check on all PRs). 
* Framework integrations

See existing docs which this can compliment:
* https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md
* https://github.com/ROCm/TheRock/blob/main/skills/therock_pr_bot/FAQ.md
* https://github.com/ROCm/TheRock/blob/main/docs/development/adding_tests.md
* https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md

For inspiration:
* https://llvm.org/docs/TestingGuide.html
* https://github.com/pytorch/pytorch/blob/main/CONTRIBUTING.md#unit-testing

ScottTodd:
See also what hipdnn has: https://github.com/ROCm/rocm-libraries/tree/develop/projects/hipdnn/docs/testing

ScottTodd:
[...]

PR #6365, opened 2026-07-06
feat: Adding python API module installaton for rocprof-trace-decoder

ApoKalipse-V:
## Motivation

Rocprof-trace-decoder is developing a python wrapper to facilitate data analysis with thread trace data.
This PR adds the python artifacts to therock builds.

rocm-systems PR: https://github.com/ROCm/rocm-systems/pull/7931 

JIRA ID : AIPROFSDK-928

## Test Plan

Added a packaging test

## Submission Checklist

- [X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6850, opened 2026-07-24
fix(profiler): Bundle `libprofiler-hub.so` into the rocm-profiler wheel

adjordje-amd:
## Motivation

The `rocm-profiler` nightly wheel ships rocprof-sys binaries and `librocprof-sys.so` hard-linked against `libprofiler-hub.so.0`, but that library is never bundled in the wheel, breaking every rocprof-sys tool at load time. Tracked as AIPROFSYST-669.

## Technical Details

rocprofiler-systems' `ProfilerHub.cmake` already builds and installs `libprofiler-hub.so.0` into its own `stage/lib/`, and it is already captured into the `rocprofiler-systems` TheRock artifact's `lib` component via the generic `*.so`/`*.so.*` glob in `ComponentDefaults`. The only gap was that the `rocm-profiler` wheel's own file allowlist in `build_tools/build_python_packages.py` never listed `lib/libprofiler-hub.so*`, so the file was silently dropped when the wheel was assembled from an otherwise-correct artifact. This adds that one glob entry next to the existing `lib/librocprof-sys*` entry.

A second, smaller commit extends `ensure_profiler_library_symlinks()` to also recreate the unversioned `libprofiler-hub.so -> libprofiler-hub.so.0` symlink, mirroring the existing treatment of `librocprof-sys*.so` (`populate_runtime_files()` only materializes the SONAME-matching filename as a real file, dropping any other symlink names pointing at it).

## Issue Tracking

JIRA ID : AIPROFSYST-669.

## Test Plan

Built `rocprofiler-systems` from `main` (pinned to the exact rocm-systems commit named in the bug's bisection) and ran `build_python_packages.py` directly against the resulting artifacts, before and after this change. Additionally built real `.whl` files, `pip install`ed them into a fresh venv, and ran the exact repro command from the bug report.

## Test Result

Before: `_rocm_profiler/lib/` contained all `librocprof-sys*.so*` files but zero `libprofiler-hub.so*` files, and `rocprof-sys-sample --version` failed with `error while loading shared libraries: libprofiler-hub.so.0: cannot open shared object file: No such file or directory` - matching the bug report exactly. After: `libprofiler-hub.so.0` (and its unversioned symlink) are populated into `_rocm_profiler/lib/`, and `rocprof-sys-sample --version` runs cleanly from a real pip-installed wheel with no `LD_LIBRARY_PATH` override.

## Submission Checklist

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -258,22 +258,7 @@ def run(args: argparse.Namespace):
 
     profiler_artifacts = params.filter_artifacts(
         profiler_artifact_filter,
-        includes=[
-            # rocprofiler-systems
-            "bin/rocprof-sys-*",
-            "include/rocprofiler-systems/**",
-            "lib/librocprof-sys*",
-            "lib/libprofiler-hub.so*",
-            "lib/python/site-packages/rocprofsys/**",
-            "lib/rocprofiler-systems/**",
-            "libexec/rocprofiler-systems/**",
-            "share/**/rocprofiler-systems/**",
-            # rocprofiler-compute
-            "bin/rocprof-*",
-            "libexec/rocprofiler-compute/**",
-            "lib/rocprofiler-compute/**",
-            "share/**/rocprofiler-compute/**",
-        ],
+        includes=PROFILER_WHEEL_INCLUDES,
     )
 
     if profiler_artifacts.artifact_names:
@@ -571,6 +556,29 @@ def profiler_artifact_filter(an: ArtifactName) -> bool:
         "rocprofiler-compute",
         "rocprofiler-systems",
     ] and an.component in ["lib", "run"]
+
+
+# File-path allowlist for the rocm-profiler wheel, applied on top of
+# profiler_artifact_filter(). rocprofiler-systems' own ProfilerHub.cmake
+# vendors profiler-hub as a runtime .so dependency (NEEDED libprofiler-hub.so.0);
+# it stages into the same lib/ dir as librocprof-sys* but needs its own glob
+# entry here, or it gets silently dropped from the wheel.
+PROFILER_WHEEL_INCLUDES = [
+    # rocprofiler-systems
+    "bin/rocprof-sys-*",
+    "include/rocprofiler-systems/**",
+    "lib/librocprof-sys*",
+    "lib/libprofiler-hub.so*",
+    "lib/python/site-packages/rocprofsys/**",
+    "lib/rocprofiler-systems/**",
+    "libexec/rocprofiler-systems/**",
+    "share/**/rocprofiler-systems/**",
+    # rocprofiler-compute
+    "bin/rocprof-*",
+    "libexec/rocprofiler-compute/**",
+    "lib/rocprofiler-compute/**",
+    "share/**/rocprofiler-compute/**",
+]
 
 
 def device_artifact_filter(target: str, an: ArtifactName) -> bool:
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -67,15 +67,16 @@ def load_therock_manifest(artifact_dir: Path) -> dict:
 
 
 def ensure_profiler_library_symlinks(profiler: PopulatedDistPackage) -> None:
-    """Recreate unversioned profiler library symlinks expected by dlopen()."""
+    """Recreate unversioned library symlinks for profiler runtime dependencies."""
     profiler_lib_dir = profiler.platform_dir / "lib"
 
-    for target in profiler_lib_dir.glob("librocprof-sys*.so.*"):
-        if target.is_symlink():
-            continue
-        link = target.with_suffix("")
-        if not link.exists():
-            link.symlink_to(target.name)
+    for pattern in ("librocprof-sys*.so.*", "libprofiler-hub*.so.*"):
+        for target in profiler_lib_dir.glob(pattern):
+            if target.is_symlink():
+                continue
+            link = target.with_suffix("")
+            if not link.exists():
+                link.symlink_to(target.name)
 
 
 def _platform_targets(
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -262,6 +262,7 @@ def run(args: argparse.Namespace):
             "bin/rocprof-sys-*",
             "include/rocprofiler-systems/**",
             "lib/librocprof-sys*",
+            "lib/libprofiler-hub.so*",
             "lib/python/site-packages/rocprofsys/**",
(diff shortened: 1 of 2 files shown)


build_tools/build_python_packages.py
dgaliffiAMD: Just in case you need to create a symbolic link from `libprofiler-hub.so.0` to `libprofiler-hub.so`, this is where you would add it.

dgaliffiAMD (approved):

build_tools/build_python_packages.py
adjordje-amd: Done

ScottTodd:
> ## Test Plan
> 
> Built `rocprofiler-systems` from `main` (pinned to the exact rocm-systems commit named in the bug's bisection) and ran `build_python_packages.py` directly against the resulting artifacts, before and after this change. Additionally built real `.whl` files, `pip install`ed them into a fresh venv, and ran the exact repro command from the bug report.

This test plan puts too much weight on manual testing. We should add automated tests for the profiler python package(s). See also my comment here: https://github.com/ROCm/TheRock/pull/6365#discussion_r3640096531

ScottTodd:
(specifically we should have integration tests that install the packages and use them, not just unit tests for package construction)

adjordje-amd:
Added automated tests in the latest commit (aabe9830b): build_tools/tests/py_packaging_test.py now covers the actual regression (libprofiler-hub.so* being selected into the profiler wheel via PROFILER_WHEEL_INCLUDES, extracted into a named constant for testability) plus ensure_profiler_library_symlinks() behavior for both libprofiler-hub and librocprof-sys.

dgaliffiAMD:
> (specifically we should have integration tests that install the packages and use them, not just unit tests for package construction)

Yes, I agree that we need a proper integration test for the rocm[profiler] package. @erman-gurses, is this something that is already on your radar? If not, then I will create a new GitHub Issue to track it. 

For now, I'd like to get this fix in, so the `rocm[profiler]` package is in a usable state.

dgaliffiAMD:
I have created https://github.com/ROCm/TheRock/issues/6856 to track workflow changes.

dgaliffiAMD:
I have independently confirmed that the [CI-built Python Packages](https://therock-ci-artifacts.s3.amazonaws.com/30125023193-linux/python/index.html) contain the new libraries. See below image.

image

erman-gurses:
> `rocm[profiler]` pac

@dgaliffiAMD , sorry I missed that - let me know If we need further follow-up.

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,177 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+        )
+
+    def test_profiler_wheel_includes_libprofiler_hub(self):
+        """libprofiler-hub.so* staged inside the rocprofiler-systems artifact
+        must be selected into the profiler wheel, same as librocprof-sys*.
+        """
+        from build_python_packages import (
+            PROFILER_WHEEL_INCLUDES,
+            profiler_artifact_filter,
+        )
```
ScottTodd, line 1611: Imports go at the top of files: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,177 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+        self.assertTrue(profiler.files.has("lib/librocprof-sys.so.1"))
+
+    def test_profiler_wheel_excludes_unrelated_lib_files(self):
+        """PROFILER_WHEEL_INCLUDES is a targeted allowlist, not a bare lib/**
+        catch-all - an unrelated file must not sneak into the profiler wheel.
+        """
+        from build_python_packages import (
+            PROFILER_WHEEL_INCLUDES,
+            profiler_artifact_filter,
+        )
```
ScottTodd, line 1647: Imports to the top: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,177 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+        params = self._make_params(artifact_dir)
+        profiler_artifacts = params.filter_artifacts(
+            profiler_artifact_filter,
+            includes=PROFILER_WHEEL_INCLUDES,
+        )
+        profiler = PopulatedDistPackage(params, logical_name="profiler")
+        profiler.populate_runtime_files(profiler_artifacts)
+
+        self.assertTrue(profiler.files.has("lib/libprofiler-hub.so.0"))
+        self.assertFalse(profiler.files.has("lib/libunrelated-dependency.so.1"))
```
ScottTodd, line 1670: This is not really a useful test IMO. Tests shouldn't need to look at implementation details like this `PROFILER_WHEEL_INCLUDES` list or include their own logic. All this test case is really checking is that the `PROFILER_WHEEL_INCLUDES` list contains a pattern matching `lib/libprofiler-hub.so.0` but not one matching `lib/libunrelated-dependency.so.1`. We don't have similar tests for the other subproject packaging code.

build_tools/tests/py_packaging_test.py
```diff
@@ -1560,5 +1560,177 @@ def test_requires_dist_pins_version_and_uses_target_in_name(self):
...
+        """
+        from build_python_packages import ensure_profiler_library_symlinks
+
+        (self.lib_dir / "librocprof-sys.so.1").write_text("fake soname file")
+
+        ensure_profiler_library_symlinks(self._fake_profiler())
+
+        link = self.lib_dir / "librocprof-sys.so"
+        self.assertTrue(link.is_symlink())
+        self.assertEqual(os.readlink(link), "librocprof-sys.so.1")
```
ScottTodd, line 1716: These test cases are nearly identical. Why duplicate? A subtest covering `.so.0` and `.so.1` would be simpler.

ScottTodd (commented):
> I have created #6856 to track workflow changes.

Thanks. I'm also working on adding an authoritative "testing.md" document that will explain the testing standards that we expect for all code in this project as part of https://github.com/ROCm/TheRock/issues/6711. This code needs an installation test, not just unit tests.

---

The fix + unit tests are helpful, but please follow the style guide (https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md) and ensure that tests are both meaningful and follow local testing practices. We'll enforce reviews on these code paths via codeowners if more changes like this are merged without following those policies.

PR #6872, opened 2026-07-27
feat(rocjitsu): package HSA hotswap runtime

Groverkss:
Publish libhsa_hotswap_rocjitsu.so as a dedicated rocjitsu artifact component. ROCr's HSA_HOTSWAP_ENABLE=2 mode requires this library, so the HSA hotswap hook must be included in runtime distributions.

Keeping it as an independently selectable component lets runtime packages ship the HSA hotswap library without pulling in the general librocjitsu shared library or introducing a ROCr build or link dependency. The currently validated unstripped shared library is 5,996,136 bytes (approximately 5.72 MiB).

Include the component in the rocm-sdk-core wheel, the amdrocm-runtime native package, and base artifact installations. Validate the staged shared library and add focused coverage for wheel selection and base artifact retrieval.

ISSUE ID: https://github.com/ROCm/rocm-systems/issues/8939


Groverkss:
Depends on https://github.com/ROCm/rocm-systems/pull/9262

stellaraccident (approved):
Given the urgency, we can land like this, but generally we don't arbitrarily extend components like that. We should come back and set it up as its own artifact. Please add such a todo ti te toml file before landing.

Groverkss:
Can be landed after https://github.com/ROCm/TheRock/pull/6897 (and after ci passes ofcourse)

marbre:
> Can be landed after #6897 (and after ci passes ofcourse)

This PR was marked and closed as stale but I assume another bump landed that ships the base for this?

Groverkss:
> > Can be landed after #6897 (and after ci passes ofcourse)
> 
> This PR was marked and closed as stale but I assume another bump landed that ships the base for this?

Yes

Groverkss:
> Given the urgency, we can land like this, but generally we don't arbitrarily extend components like that. We should come back and set it up as its own artifact. Please add such a todo ti te toml file before landing.

Created a new artifact instead of leaving a TODO.

build_tools/install_rocm_from_artifacts.py
```diff
@@ -350,6 +350,7 @@ def retrieve_artifacts_by_run_id(args):
         "core-hipinfo_run",
         "core-runtime_run",
         "core-runtime_lib",
+        "rocjitsu-hotswap_hook",
```
marbre: I don't have enough context on rocjitsu and neither on hotswap but is this really a base requirement to run tests? I bet it is not (as of today).

emulation/artifact-rocjitsu.toml
```diff
@@ -4,6 +4,11 @@ default_patterns = false
 include = [
   "lib/librocjitsu.so",
 ]
+[components.hook."emulation/rocjitsu/stage"]
```
marbre: What is `components.hook`? I think this should either be in `components.lib` if it a runtime component or just go into a separate artifact descriptor which can then be made optional?

emulation/CMakeLists.txt
```diff
@@ -52,6 +58,17 @@ if(THEROCK_ENABLE_ROCJITSU)
...
+  if(THEROCK_ENABLE_ROCJITSU_HOTSWAP)
+    therock_provide_artifact(rocjitsu-hotswap
+      TARGET_NEUTRAL
+      DESCRIPTOR artifact-rocjitsu.toml
+      COMPONENTS
+        hook
+      SUBPROJECT_DEPS
+        rocjitsu
+    )
+  endif()
```
marbre: If you want to use a shared artifact descriptor I would rather extend the COMPONENTS list for `rocjitsu` above, but it seems cleaner to move it to a separate artifact descriptor as for mirage. I don't think it is immediately obvious what coupling is introduced here.

build_tools/install_rocm_from_artifacts.py
Groverkss: It's not needed for tests today yes. But eventually, we plan to choose which backend to make default and if we do make it default any tests on a gfx1250 A0 machine CI won't work, because the runtime will use this library. For context, anything that uses rocr-runtime (almost everything that runs on a gpu) for gfx1250 A0 needs hotswap. The comgr backend is default today, but we might switch to the rocjitsu backend. If any CI decides to use the rocjitsu comgr backend, it will not be able to run tests without this.

emulation/artifact-rocjitsu.toml
Groverkss: done

emulation/CMakeLists.txt
Groverkss: done

emulation/CMakeLists.txt
```diff
@@ -40,6 +40,12 @@ if(THEROCK_ENABLE_ROCJITSU)
     LIB_NAMES
       librocjitsu.so
   )
+  if(NOT WIN32)
+    therock_test_validate_shared_lib(PATH rocjitsu/dist/lib
+      LIB_NAMES
+        libhsa_hotswap_rocjitsu.so
+    )
+  endif()
```
marbre: Here you use `rocjitsu/dist/lib` while you use `rocjitsu/stage/lib` in L72 (dist vs stage). Both should work for consistency, we should pick one tree and use it in both places eventually.

emulation/CMakeLists.txt
```diff
@@ -40,6 +40,12 @@ if(THEROCK_ENABLE_ROCJITSU)
     LIB_NAMES
       librocjitsu.so
   )
+  if(NOT WIN32)
```
marbre: This is already guarded by `THEROCK_ENABLE_ROCJITSU` and this disabled on Windows (`disable_platforms = ["windows"]` in the build topology). So I think this guard can be dropped.

emulation/CMakeLists.txt
```diff
@@ -52,6 +58,33 @@ if(THEROCK_ENABLE_ROCJITSU)
     SUBPROJECT_DEPS
       rocjitsu
   )
+
+  if(THEROCK_ENABLE_ROCJITSU_HOTSWAP)
```
marbre: This is fine to be at top level IMHO as `rocjitsu-hotswap` depends on `rocjitsu`. So you should be able to move this out of `if(THEROCK_ENABLE_ROCJITSU)` (same as for mirage).

marbre (commented):
This looks mostly good, a few NITs.

emulation/CMakeLists.txt
Groverkss: fixed

emulation/CMakeLists.txt
Groverkss: done

emulation/CMakeLists.txt
Groverkss: fixed

Groverkss: fixed

marbre (approved):
Verified `_rocm_sdk_core/lib/libhsa_hotswap_rocjitsu.so` made it's way into rocm-sdk-core. CI failures are unrelated.

PR #6876, opened 2026-07-27
Harden docker helper downloads against supply-chain tampering

marbre:
## Motivation

Add integrity protection for the pinned third-party binaries and source archives fetched by the docker helper install scripts. These downloads were previously trusted based only on HTTPS and then executed or extracted immediately.

JIRA ID: ROCM-26750

Generated with Codex

## Test Plan

New image build via https://github.com/ROCm/TheRock/actions/runs/30248091598.

## Test Result

Building the new image passed, available at https://github.com/ROCm/TheRock/pkgs/container/therock_build_manylinux_x86_64/1069706460?tag=users-marbre-harden-downloads.

## Submission Checklist

Generated with Codex

### DIFF
--- a/dockerfiles/install_awscli.sh
+++ b/dockerfiles/install_awscli.sh
@@ -7,14 +7,31 @@
 
 set -euo pipefail
 
+AWSCLI_VERSION="${1:-2.36.7}"
 ARCH="$(uname -m)"
+case "${AWSCLI_VERSION}:${ARCH}" in
+    2.36.7:x86_64)
+        AWSCLI_SHA256="d641283d37f1a2168457a9f26a20d4e29167652e9ab1719b37114ef1ebe859f4"
+        ;;
+    2.36.7:aarch64)
+        AWSCLI_SHA256="85826b67912b44bb45d1e46c6e66f383c14405ee0b2f4686f73bdf949c93bd61"
+        ;;
+    *)
+        echo "Unsupported AWS CLI version/architecture: ${AWSCLI_VERSION}/${ARCH}" >&2
+        exit 1
+        ;;
+esac
+
+AWSCLI_ARCHIVE="awscli-exe-linux-${ARCH}-${AWSCLI_VERSION}.zip"
 
 # https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
 curl --silent --fail --show-error --location \
-    "https://awscli.amazonaws.com/awscli-exe-linux-${ARCH}.zip" \
-    --output "awscliv2.zip"
+    "https://awscli.amazonaws.com/${AWSCLI_ARCHIVE}" \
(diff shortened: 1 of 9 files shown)


HereThereBeDragons (approved):
general design question:
we have 1) some install scripts (aws) that decide on the version and have the hash inside the install script and then we have 2) other scripts that take any version but check only for a specific version the hash and fail otherwise.

i think it would be good to unify them. wondering if when you select the version on the outside you should also provide the hash - or as we support normally only one version (except python) anyway have it hardcoded inside the script?

marbre:
> general design question: we have 1) some install scripts (aws) that decide on the version and have the hash inside the install script and then we have 2) other scripts that take any version but check only for a specific version the hash and fail otherwise.
> 
> i think it would be good to unify them. wondering if when you select the version on the outside you should also provide the hash - or as we support normally only one version (except python) anyway have it hardcoded inside the script?

For now we only want to pin a single version as this is purely for our Docker containers.

dockerfiles/install_awscli.sh
ScottTodd, line 1: I still want to fully remove this `install_awscli.sh` script and use only `boto3` for interfacing with AWS.

Fine to keep supporting it for now...

dockerfiles/install_shared_pythons.sh
```diff
@@ -23,15 +23,29 @@ PYTHON_VERSIONS=(
   "3.13.2"
   "3.14.0"
 )
+declare -Ar PYTHON_SHA256S=(
+  ["3.10.16"]="f2e22ed965a93cfeb642378ed6e6cdbc127682664b24123679f3d013fafe9cd0"
+  ["3.11.11"]="883bddee3c92fcb91cf9c09c5343196953cbb9ced826213545849693970868ed"
+  ["3.12.9"]="45313e4c5f0e8acdec9580161d565cf5fea578e3eabf25df7cc6355bf4afa1ee"
+  ["3.13.2"]="b8d79530e3b7c96a5cb2d40d431ddb512af4a563e863728d8713039aa50203f9"
+  ["3.14.0"]="88d2da4eed42fa9a5f42ff58a8bc8988881bd6c547e297e46682c2687638a851"
+)
```
ScottTodd, line 32: When we next add a version here we'll also need to compute the sha256 value. Could add a comment that demonstrates how to do that. Maybe a note in https://github.com/ROCm/TheRock/blob/main/dockerfiles/README.md shared across all scripts?

ScottTodd (approved):

dockerfiles/install_shared_pythons.sh
marbre: Generated with Codex ;) Well yeah, let me figure out a way on my workstation and add this in a follow up. I assume we want to bump a few versions anyway soon.

PR #6884, opened 2026-07-27
Use filters to make python tar file extractions safer

HereThereBeDragons:
Restrict tar extraction to recommended behavior following PEP 706: https://peps.python.org/pep-0706/#filters

Using "tar" as filter for build_tools/install_rocm_from_artifacts.py, because the extracted release tarball contains .so linker symlinks that filter="data" can reject as unsafe; "tar" preserves them while still blocking ../ and absolute-path traversal, consistent with the sibling extractors artifact_manager.py and fetch_artifacts.py

Doc change is just to have at least the minimum of extraction safety, matching the script analyze_ccache_logs.py.

JIRA ID: ROCM-26879

### DIFF
--- a/build_tools/install_rocm_from_artifacts.py
+++ b/build_tools/install_rocm_from_artifacts.py
@@ -273,7 +273,7 @@ def _untar_files(output_dir: Path, destination: Path):
     """
     log(f"Extracting {destination.name} to {str(output_dir)}")
     with tarfile.open(destination) as extracted_tar_file:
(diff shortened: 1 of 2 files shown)


HereThereBeDragons:
@amd-shiraz here is another example why the unit test enforcement is not "ideal"

marbre (approved):

marbre:
Conditional approval if CI passes.

ScottTodd (approved):

PR #6888, opened 2026-07-27
Set top-level permissions in hip_tagging_automation.yml

marbre:
## Motivation

Adds top-level permissions to restrict to `contents: read` by default.

JIRA ID: ROCM-26883

## Submission Checklist

### DIFF
--- a/.github/workflows/hip_tagging_automation.yml
+++ b/.github/workflows/hip_tagging_automation.yml
@@ -10,6 +10,9 @@ on:
     paths:
         - rocm-systems
 
+permissions:
+  contents: read
+
 jobs:
   tag-rocm-systems:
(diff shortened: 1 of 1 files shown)


.github/workflows/hip_tagging_automation.yml
ScottTodd, line 38: I guess this will still work if the workflow itself just has
```yml
permissions:
  contents: read
```
?

As needed we could adjust and also set https://github.com/actions/create-github-app-token#create-a-token-with-specific-permissions

ScottTodd (approved):

.github/workflows/hip_tagging_automation.yml
marbre: I think it does, yes, otherwise we will follow up.

PR #6896, opened 2026-07-27
feat(ci): Add AMDGPU DKMS package version to driver/GPU sanity check

dayatsin-amd:
## Summary

Adds the installed AMDGPU DKMS package version to the Linux CI runner sanity check in `build_tools/print_driver_gpu_info.py`, alongside the existing `amd-smi static`, `rocminfo`, and kernel version (`uname -r`) output. This makes runner driver state easier to diagnose from CI logs.

Rather than shelling out to `dkms status` (which requires root on some systems and would break running the script as a non-root user), a new `log_dkms_status()` helper reads `/var/lib/dkms/<module>/` directly. It walks the on-disk `<version>/<kernel>/<arch>` layout and reports each entry as `installed`/`built`, producing output equivalent to `dkms status` without elevated privileges. The `kernel-*` symlinks and the `original_module` backup directory are skipped.

Note: the script assumes Python 3.10+ (it already uses the `str | Path` PEP 604 annotation).

## Test plan

Verified as a non-root user on runners with different distros:

- [x] `dayatsin-test-01` (Python 3.12): lists both installed kernels for `amdgpu/6.18.10-2299007.24.04`
- [x] `f05-1` (Python 3.12): lists both installed kernels for `amdgpu/6.19.0-2372803.24.04`
- [x] `original_module` backup directory is correctly skipped (verified against b06-2's `/var/lib/dkms/amdgpu` layout)
- [x] Output matches prior `dkms status` and requires no `sudo`
- [x] Graceful handling when the module has no DKMS entry (prints "no DKMS entry found under /var/lib/dkms")

### DIFF
--- a/build_tools/print_driver_gpu_info.py
+++ b/build_tools/print_driver_gpu_info.py
@@ -9,7 +9,7 @@ On Linux:
   - run "amd-smi static"
   - run "rocminfo"
   - run "uname -r"
-  - run "dkms status -m amdgpu"
+  - read "/var/lib/dkms" for the AMDGPU DKMS package version
 
 On Windows:
   - run "hipInfo.exe"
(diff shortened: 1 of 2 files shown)


dayatsin-amd:
At the moment, this function is not able to get the DKMS version on the CI systems, but works in the existing systems that I have access to.

=== amdgpu DKMS package version ===
amdgpu: no DKMS entry found under /var/lib/dkms

I need access to a system that is similar to the CI systems to be able to debug further and this system has already been requested.
Will get back to this one we have debugged on that syste,/

build_tools/print_driver_gpu_info.py
```diff
@@ -92,6 +94,57 @@ def run_command_with_search(
     log(f"{command}: command not found")
 
 
+DKMS_ROOT = Path("/var/lib/dkms")
```
geomin12, line 94: ah so we run this on a docker container, so it does not have access to the underlying files on machine

this is why it doesn't exist. might need to do a docker mount for those paths, but this may bring up security concerns

geomin12:
> At the moment, this function is not able to get the DKMS version on the CI systems, but works in the existing systems that I have access to.
> 
> === amdgpu DKMS package version === amdgpu: no DKMS entry found under /var/lib/dkms
> 
> I need access to a system that is similar to the CI systems to be able to debug further and this system has already been requested. Will get back to this one we have debugged on that syste,/

note: this is due to docker container, and this is why dkms is not accessible. added a comment in review too

build_tools/print_driver_gpu_info.py
dayatsin-amd: ok. Thanks. Makes sense.
I only need read-access to this path, so we can add the "ro" tag to the mounts, e.g:

-v /path/to/host/dir:/path/in/container:ro

dayatsin-amd:
Pushed 3ad2233b8: the sanity check runs inside the test Docker container, so I added a read-only bind mount of the host's `/var/lib/dkms` to `_BASE_CONTAINER_OPTIONS` in `build_tools/github_actions/fetch_test_configurations.py`:

```
-v /var/lib/dkms:/var/lib/dkms:ro
```

This exposes the host DKMS state to the container without write access, addressing the security concern @geomin12 raised. `print_driver_gpu_info.py` only reads that path, so read-only is sufficient. This should let the sanity check report the AMDGPU DKMS version in CI now.

dayatsin-amd:
This is still not working on the CI's, but work when I use the same docker container on my test system. Moving this PR to draft until I have cycles to look into this again.

Issue #5384, opened 2026-05-21
Add component tests that use Python and native packages


marbre:
Currently component tests fully rely on artifacts. Only install and sanity testing is performed for Python packages and native Linux packages. As a result, packaging related issues get caught to late, requiring fixes like #5376.

We need to add more component test coverage that uses the Python packages and native Linux packages.

PR #6909, opened 2026-07-27
[hipthreads] Include hipThreads in rocm-sdk Python packages

marsavicc:
## Summary
Adds `hipthreads` to the devel-wheel artifact allowlist in `build_tools/build_python_packages.py`, so the hipThreads headers, static
library, and CMake package config ship in `rocm-sdk-devel`.

hipThreads provides only a static library (no runtime `.so`), so it belongs in the devel wheel alongside the other static-only libs (rocshmem), not in the runtime `rocm-sdk-libraries` wheel.

JIRA ID : ROCM-23943


ammallya (approved):
LGTM

marbre (approved):
LGTM. However, let's please land #6769 first.

marbre (approved):
Approving as code owner, but let's verify CI produced packages before merging!

build_tools/build_python_packages.py
```diff
@@ -380,6 +380,8 @@ def _run_kpack_split(
             "nlohmann-json",
             # rocshmem only provides a static library.
             "rocshmem",
+            # hipthreads only provides a static library.
+            "hipthreads",
```
ScottTodd, line 384: Inspecting https://therock-ci-artifacts.s3.amazonaws.com/30364308720-linux/python/rocm_sdk_devel-7.15.0.dev0%2B66caf6137482a115dd76f339012c448544fd6817-py3-none-linux_x86_64.whl from the CI run on this PR (https://github.com/ROCm/TheRock/actions/runs/30364308720?pr=6909), I see these files:

```
rocm_sdk_devel-7.15.0.dev0+66caf6137482a115dd76f339012c448544fd6817-py3-none-linux_x86_64.whl\
  rocm_sdk_devel\_devel.tar\_rocm_sdk_devel\
    include\            # various source files, such as...
      hipthreads\hip\__mutex\lock_guard.h
   lib\
     hipthreads\libhipthreads.a
```

Looks like that's all you're expecting?

This is probably fine. I'd like to find a way to test changes like this with automation at some point - maybe install the python devel packages and have a set of tests for each component that should be included to make sure it's usable? Maybe we could run the scripts from https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions/test_executable_scripts ?

I only checked the Linux packages FYI. Looks like hipthreads is not excluded on Windows and the windows package build is still running at https://github.com/ROCm/TheRock/actions/runs/30364308720/job/90329776677?pr=6909 https://github.com/ROCm/TheRock/blob/52b4d2a916ccba7bade61af26265a7f588c9e1d2/BUILD_TOPOLOGY.toml#L643-L648

ScottTodd (approved):

build_tools/build_python_packages.py
marsavicc: Yes, that's exactly the expected payload, hipThreads only ships a dev component (headers under `include/hipthreads/hip/**` plus the static `lib/hipthreads/libhipthreads.a` and its CMake config), no runtime `.so`, so headers + `.a` is the complete set.
  
On Windows: the packaging jobs have since finished and are green (Build Python Packages, Test ROCm wheels, and Validate Artifact Structure all passed), so hipThreads is included cleanly on Windows too. 

I'm just looking at the test PR #https://github.com/ROCm/TheRock/pull/6770 and there is an issue with windows tests , we might need to change something about testing on Windows but that's pretty much it.

ScottTodd: SGTM, thanks. Note that (for all subprojects) we currently just run those tests on the _artifacts_, not the _native linux packages_ or the _python packages_. That's what I'd like to expand somehow, so we can verify that each distribution we publish has all expected subprojects working as expected. See
* https://github.com/ROCm/TheRock/issues/5384

marsavicc: I'll be happy to extend the hipThreads runners (`test_hipthreads.py` / `test_hipthreads_examples.py`) to run against the Python wheel and native packages once that approach lands, they're basically consumer-path tests already.

Issue #6927, opened 2026-07-28
Organize project unit tests so all are run continuously


ScottTodd:
The https://github.com/ROCm/TheRock/blob/main/.github/workflows/unit_tests.yml workflow runs our unit tests on all PRs/commits but it uses some limited path filters that do not cover all files:

* https://github.com/ROCm/TheRock/blob/390021ce8c36bb439325895411cf20c58a04a977/.github/workflows/unit_tests.yml#L40-L53
* https://github.com/ROCm/TheRock/blob/390021ce8c36bb439325895411cf20c58a04a977/build_tools/pyproject.toml#L1-L9

At least these tests are not running continuously:
* https://github.com/ROCm/TheRock/blob/main/skills/therock_pr_bot/test_policy_check_ut.py
* https://github.com/ROCm/TheRock/blob/main/test_tools/tests/determine_rocm_test_dependencies_test.py
* https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/linux/tests/native_linux_package_install_ut_test.py

As part of this I think it's time to move https://github.com/ROCm/TheRock/tree/main/build_tools/github_actions/test_executable_scripts out of `build_tools/github_actions/`. Maybe to a new subdirectory under https://github.com/ROCm/TheRock/tree/main/tests ? I may also want to add more subdirectories there for ROCm package tests that don't fit naturally under https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/python or https://github.com/ROCm/TheRock/tree/main/build_tools/packaging/linux.

PR #6941, opened 2026-07-28
feat(ci): derive test-subproject selection from the consumer graph 

endurthiabhilash:
## Motivation

TheRock CI selects which GPU test suites to run from the set of changed projects, via a hand-maintained `TEST_SUBPROJECTS` list on each `therock_cmake_subproject_declare()`. That list is only partially populated, so a change to most subprojects selects nothing beyond itself, producing passing CI runs that never exercised the affected downstream suite (false greens).

This PR makes test selection a *derived* property of the dependency edges the build already declares, rather than a parallel list kept in sync by hand. Selection walks the reverse-dependency (consumer) graph to a per-component **level** on a 1–5 gating ladder — default level 4 (the changed component + its direct consumers), with broader/narrower tiers available per component. See
[RFC0013](https://github.com/endurthiabhilash/TheRock/blob/main/docs/rfcs/RFC0013-Consumer-Based-Test-Selection.md) for the design and gating background.

Issue ID: #3343

## Technical Details
- **Consumer graph (CMake).** `therock_cmake_subproject_declare()` records a reverse edge per `BUILD_DEPS`/`RUNTIME_DEPS`; `therock_emit_consumer_graph()` serializes it at configure time. The graph is committed to `test_tools/therock_consumer_graph.json` so the per-PR change-detection job reads it directly — no fetch or configure on the hot path. Removes the `TEST_SUBPROJECTS` keyword and its per-component uses.
- **Level-based selection (Python).** `determine_rocm_test_dependencies.py` loads the committed graph and, for each changed subproject, walks consumer edges to the depth set by its policy level (5 = self only, 4 = direct consumers, 3 = transitive closure). Build stage is not consulted.
- **Policy overrides (`test_tools/test_policies.toml`).** A small, hand-reviewed file layers explicit intent on the graph: per-component `level`, plus `test_include` (additive) and `test_exclude` (subtractive, applied last) for couplings the graph can't express, test-only targets with no link edge (e.g. `rocgdb-cpu`, `hipinfo`) and reverse edges absent from forward deps. Broad-blast-radius runtimes (e.g. `rocr-runtime`) use `level = 3` to auto-derive their downstream set instead of listing it by hand.
- **Drift check (CI).** `consumer_graph_drift.yml` regenerates the graph from a full-tree configure and fails if the committed copy is stale, so the checked-in graph can never silently diverge from the declared dependency edges. 

## Test Plan
- `test_tools/tests/determine_rocm_test_dependencies_test.py` covers the level walk (each depth), `test_include`/`test_exclude`, and policy validation against the committed graph.
- `build_tools/github_actions/tests/check_consumer_graph_drift_test.py` covers the drift check's normalization and pass/fail behavior.

## Submission Checklist
 - [ ] Reviewed CONTRIBUTING.md
 - [ ] Unit tests pass
 - [ ] RFC0013 reflects the merged design


docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+DEFCON 4, "local **and intimately related** project builds/tests", is precisely
+the selection this RFC automates: for a changed subproject, run it plus the
+projects directly downstream of it. Doing that for *every* component requires a
+reliable, low-maintenance answer to "what is directly downstream of X?" across the
+whole tree. The hand-maintained `TEST_SUBPROJECTS` lists were that answer in
+principle but not in practice (below); the consumer graph makes the DEFCON 4
+baseline derivable rather than curated, and the override keys become the knobs for
+moving a component off that baseline.
+
+[defcon]: https://amd.atlassian.net/wiki/spaces/MLSE/pages/1795331309/DEFCON+Enablement+Gating+Levels
```
ScottTodd: No direct links to internal pages in open source code. You can indirect through a u.amd.com shortlink as needed, but if we want to refer to this from open source the design should be public (I also don't know if we'd want to use the term "defcon" - @astyrrian1 ?)

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+that fixes *how much* every PR must build and test, sized to the component's
+downstream blast radius and regression history.
+
+| Level        | What every PR must build & test                                     | Intended for                                             |
+| ------------ | ------------------------------------------------------------------- | -------------------------------------------------------- |
+| **DEFCON 1** | Full QA integration + nightly suites                                | Components actively destabilizing releases               |
+| **DEFCON 2** | Frameworks built and their tests run                                | High downstream blast radius (e.g. framework-facing libs)|
+| **DEFCON 3** | All rocm-libraries projects/tests + rest of rocm-systems (rccl etc.)| Broad-dependency components with regression history      |
+| **DEFCON 4** | Local + intimately related project builds/tests only                | Well-tested components with contained surface            |
+| **DEFCON 5** | Fast unit tests only                                                | Proven track record + strong local test coverage         |
```
ScottTodd: By this definition, isn't this between 3 and 4?

4 is just "intimately related projects", e.g.
* hip-clr edited -> test hip-tests
* rocRoller edited -> test hipBLASLt (which directly depends on rocRoller), do NOT test MIOpen (which directly depends on hipBLASLt and _indirectly_ depends on rocRoller)

3 would be "all projects/tests" which is for "I don't trust the build/test system to compute the impact of this change for me, so just build/test everything"

What using the build graph lets us do is compute direct and indirect users of each project and decide how far along those dependency graph edges we want to walk, e.g.:
* rocRoller edited - > test hipBLASLt and MIOpen
* hip-clr edited -> test all users of it (rocRAND, rocPRIM, rocRoller, hipBLASLt, MIOpen, etc.), do NOT test projects that don't have a direct or indirect dependency on hip-clr, like rocr-debug-agent or rdc

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+a change to those projects selects nothing beyond itself. The mechanism is sound;
+the coverage is incomplete, and completing it by hand has proven impractical:
+
+- **Manual duplication.** `TEST_SUBPROJECTS` restates dependency relationships
+  that are *already* declared via `BUILD_DEPS` / `RUNTIME_DEPS`. When a new
+  consumer of a library is added, the library's `TEST_SUBPROJECTS` must be edited
+  too, or its tests silently stop running on relevant changes.
+- **Drift and false greens.** A stale or missing entry produces a passing CI run
+  that never exercised the affected downstream suite: the most dangerous class of
+  gap because it is invisible.
```
ScottTodd: Here's a recent example: https://github.com/ROCm/rocm-systems/commit/920e5a8a70669d28b930c33fcef2b18c4b47d152

A change to https://github.com/ROCm/rocm-systems/tree/develop/projects/amdsmi broke https://github.com/ROCm/rocm-systems/tree/develop/projects/rdc and the CI didn't catch it because https://github.com/ROCm/rocm-systems/blob/develop/.github/scripts/therock_matrix.py was not modeling the build dependency relationship there.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+   `build/therock_consumer_graph.json`, a configure-time side effect analogous to
+   `compile_commands.json`. It is **dynamic and never committed**.
+3. **Derive test subprojects in Python.** `determine_rocm_test_dependencies.py`
+   loads the graph, computes each subproject's build stage from the committed
+   `artifact-*.toml` descriptors + `BUILD_TOPOLOGY.toml`, and for each changed
+   subproject selects its **same-stage** direct consumers.
+4. **Apply explicit overrides from `BUILD_TOPOLOGY.toml`.** Couplings the graph
+   cannot express (test-only projects with no link edge, cross-stage couplings,
+   and stage-less foundational deps) are declared with `test_include` /
+   `test_exclude` / `test_fanout_all` keys on the relevant artifact.
```
ScottTodd: I'm not sure I agree with the usage of `BUILD_TOPOLOGY.toml` stages here. "cross-stage couplings" are commonplace, e.g. many projects depend on `amd-hip` (the compiler), `hip-clr` (the runtime), and `rocprofiler-sdk` (the profiler).

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+opt-in and auditable rather than accidental.
+
+The same-stage cut is what keeps the default at **DEFCON 4** ("local + intimately
+related") rather than sliding into DEFCON 3 ("all rocm-libraries / rocm-systems").
+Selecting every transitive consumer of a foundational dep is effectively DEFCON 3
+behavior on every touch of that dep; the cut confines the automatic selection to
+the directly-related, same-stage neighborhood and keeps runs inside the 10–15 min
+SLA. Deliberately widening a component toward DEFCON 3/2 is then an explicit,
+reviewable act (`test_fanout_all` / `test_include`) rather than an accident of the
+graph's shape.
```
ScottTodd: Relating to my other comments... well yeah, if something depends on `hip-clr` then its builds/tests could be affected by changes there. I don't think "build stage" is the right boundary here. Viewing this through the lens of "direct" and "indirect" deps may be more useful.

The framing here as "an accident of the graph's shape" doesn't make sense to me - if a project takes a dependency on another project that's no accident.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+determines test dependencies:
+
+1. `pip install -r requirements.txt`
+2. `fetch_sources.py` — all subproject source trees must be present so every
+   `therock_cmake_subproject_declare()` runs and the emitted graph is complete.
+3. A configure-only `cmake -B build -DTHEROCK_ENABLE_ALL=ON ...` that emits
+   `build/therock_consumer_graph.json` as a side effect.
+
+`determine_rocm_test_dependencies.py` then reads the freshly emitted graph. The
+graph is never committed, so it cannot drift from the build definitions.
```
ScottTodd: This is pretty expensive to run: https://github.com/ROCm/TheRock/actions/runs/30411761253/job/90461735601?pr=6941. It takes around 6 minutes on a GitHub-hosted runner (4m40s on checkout, 1m40s on configure).

We'll also want more of this information to be available _prior to building_, so we can ensure
* all projects we plan on testing are also built 
* (if we want) _only_ projects we plan on testing are built

Some ideas:
* Generate the graph and check it in / cache it somehow, so we don't need to recompute with this dynamic code
* Have the CI configure step that runs at the start of the pipeline compute the graph (something similar is needed for build artifact reuse too, cc @erman-gurses @ammallya )
* Have each build stage emit a subset of the graph and then merge those subgraphs back together when running tests

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+Levels 3, 4, and (by radius) 5 are expressible with the current selection engine.
+`test_fanout_all` is precisely the level-3 transitive reach, so it collapses into
+the ladder and can be deprecated in favor of `defcon = 3`.
+
+### What the generalization adds
+
+1. **A `defcon` key** on `[artifacts.<name>]`, overridable in the
+   `test_overrides.<subproject>` sub-table and resolved by the existing
+   `get_test_overrides_for_subproject()` path. Absent means the default level 4,
+   which is exactly the RFC's "Phase 1 starts every component at DEFCON 4."
```
ScottTodd: I like the idea of having a clear place to say "changes to these code paths/project should use this level of testing". I'm not convinced that the topology or artifacts are the right place for that though. Need to think on this some more.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
```diff
@@ -0,0 +1,336 @@
...
+### Overview
+
+1. **Register consumers at declare time.** Every
+   `therock_cmake_subproject_declare()` appends the subproject to a global
+   registry and records a reverse edge (`consumer`) for each of its
+   `BUILD_DEPS` / `RUNTIME_DEPS`.
+2. **Emit a consumer graph.** At the end of the top-level configure,
+   `therock_emit_consumer_graph()` serializes the registry to
+   `build/therock_consumer_graph.json`, a configure-time side effect analogous to
+   `compile_commands.json`. It is **dynamic and never committed**.
```
ScottTodd: Can you post the current `therock_consumer_graph.json` somewhere we can look at it? This comment thread, a gist linked from the PR description, etc.

ScottTodd (commented):
I left some initial comments on the design.

In general I think we do need this dynamic graph analysis in some form to make CI build/test selection scale well.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: Ack. My earlier idea was to generate the graph, check it in and have a CI configure step. However, I didn't want to introduce a new json file to check in and went with dynamically generating each time.

I'll explore adding the consumer_graph.json and add a CI step to validate at the start of the pipeline.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: I'll remove the direct amd.atlassian.net link and can add a short public summary of the gating levels so the RFC is self-contained. 

On the "defcon" term itself — I'd like to wait for confirmation from others (@astyrrian1, @suchitravemuri1) before renaming. If we settle on a different public name for the levels I'll rename throughout.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: Agreed the dynamic-every-run configure is too expensive on the per-PR hot path - I've moved off it in the RFC. The approach I'm proposing:

1. Commit therock_consumer_graph.json to the repo. The per-PR change-detection job reads the checked-in file directly — no cmake, no source fetch on the hot path.
2. Add a CI validity check that regenerates the graph and fails if the committed copy is stale. It runs only when dependency-related files change (**/CMakeLists.txt, the emit/registration cmake, the committed graph itself), so the ~6-minute configure is confined to that low-frequency job instead of every PR. This turns silent drift into a CI failure similar to the contract BUILD_TOPOLOGY.toml already carries.

ScottTodd: That works for me. The main thing I want to avoid is having a complex graph that is only manually updated, as then we can easily overlook that graph going stale or accidentally missing dependency edges that it should have. A dynamically generated graph _or_ a static graph that is produced by a script (ideally with CI enforcing that it stays up to date) would both be okay.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: ACK

If topology or artifacts are not the right places, I'm leaning toward a dedicated file at the root or in testing_tools, e.g. test_policies.toml, keyed by component, carrying both the per-component level and any test_include / test_exclude. Keying by component removes the artifact-sharing issue and gives one clear place to answer "what testing does a change to this component trigger?". 
Rough sketch:

```
[component.hipblaslt]
level = 4
test_include = ["hipsparselt"]
```
Let me know if you have other suggestions. Otherwise, I can document this approach in RFC and implement.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: Agree with all the comments related to build stages. I'll come up with a plan on how to implement `direct`, `transitive` dependencies, update the RFC for another round of review.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
endurthiabhilash: I will check in therock_consumer_graph.json once I build it with `direct`, `indirect` consumers.

docs/rfcs/RFC0013-Consumer-Graph-Test-Selection.md
ScottTodd: I'm coming around to the idea of putting more code in https://github.com/ROCm/TheRock/tree/main/test_tools specifically for these sorts of details in the CI system.

Ideally I'd like to have what build systems like Bazel have, where all build/package/test/etc. targets are known to the build system and you can just run `bazel test` and have it understand what targets already have cached outputs/results and which need to get built/run. To that end, I've hoped we could get the CMake superproject here to support similar features as natively as possible via our CMake modules and supporting files like BUILD_TOPOLOGY.toml.

I think a reasonable middle ground between "the build system fully models all dependencies and knows how to cache" and "we manually specify every dependency relationship and testing level for each project" would be to have code in https://github.com/ROCm/TheRock/tree/main/test_tools or a similar location that is generated from knowledge about the build system and has a few escape hatches / customization points where we can say "changes to this project should by default also build/test projects 'foo' and 'bar'" or "changes to this project should by default build/test all projects that depend on it: [generated list]"

endurthiabhilash: That matches with where I've landed. Concretely I'd split it into two files:

therock_consumer_graph.json: generated, CI-monitored. The full reverse dependency graph, emitted from the build system, no hand editing. The validity check from the CI-cost thread is the "CI enforcing it stays up to date" part.
A hand-maintained overrides file (e.g. test_policies.toml): small and reviewed. Only the escape hatches, per-component level (how far to walk the generated graph) plus test_include / test_exclude for what the graph can't express.

The escape hatches you named map directly: "also build/test component a and b" -> test_include; 
"ignore testing component c" -> test_exclude;
"build/test all projects that depend on it" -> level = 3 (transitive walk over the generated graph, so the list stays generated rather than pasted in).

I'd keep the generated graph and the overrides as separate files because they have opposite lifecycles: the graph is machine-generated and shouldn't be hand-edited (edits get clobbered on regen, and the whole point is it can't drift), the overrides are human intent. Splitting them lets the validity check just regenerate-and-diff the graph, and the overrides get reviewed like code, with no file that's half-generated and half-authored.

Drawbacks of the two-file split, to be upfront:

Two places to look. Answering "what does a change to X test?" means reading the graph and the overrides together instead of one entry. I'd add a determine_rocm_test_dependencies.py helper that prints the resolved selection for a component so we rarely compose by hand.
Referential integrity across files. An override can name a component that no longer exists in the graph, and the override file alone won't catch it. Needs a small validation step that checks override keys against the current graph.
Two-step edit sometimes. A genuinely new dep edge regenerates the graph (automatic) and may need an override tweak (manual), where a single hand-file would be one edit (but then it isn't generated at all).
Precedence has to be defined. With level plus include/exclude on a generated base, the resolution order (walk to level -> apply include -> apply exclude) needs to be specified and tested so results are deterministic.
I think the split is worth it, it's the only way the graph half stays CI-owned and drift-proof while the intent half stays hand-reviewed, and the drawbacks are all addressable with a resolver helper + an override validation check.

.github/workflows/consumer_graph_drift.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          cmake -B build -S . -GNinja \
+            -DTHEROCK_ENABLE_ALL=ON \
+            -DTHEROCK_AMDGPU_FAMILIES=gfx110X-all
+
+      - name: Check committed consumer graph is up to date
+        # Compares the freshly emitted graph (build/) against the committed copy
+        # (test_tools/) after normalizing both the same way, and fails on any real
+        # edge change. Logic lives in a unit-tested script, not inline here.
+        run: python3 ./build_tools/github_actions/check_consumer_graph_drift.py --check
```
ScottTodd: No need for verbose comments like "Logic lives in a unit-tested script, not inline here.", the code already expresses that.

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+#   test_exclude = ["name", ...]      # Optional. Consumers to prune. SUBTRACTIVE,
+#                                     #   applied LAST (after walk + include).
+#
+# Resolution order: walk to the component's level -> union test_include ->
+# subtract test_exclude (last). Every component KEY must name a real
+# consumer-graph key (it is the change trigger; a cross-file validation, task #5,
+# fails the build otherwise). test_include / test_exclude VALUES need NOT be graph
+# keys: their whole purpose is to name test-only targets the graph cannot express
+# (e.g. ctest suites like rocgdb-cpu). There is no `test_fanout_all`: "test all
+# downstream consumers" is expressed as `level = 3`.
```
ScottTodd: What is `task #5`? (I've seen internal task planning state leak into code comments authored by AI agents, did that happen here?)

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+#
+# Components are DEFCON 4 by default (level 4 = self + direct consumers, so the
+# level key is omitted). A few broad-blast-radius runtimes set level = 3 (full
```
ScottTodd: nit: this sentence is redundant with the table above that says the default level is 4. Could trim these comments down a bit.

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+# transitive consumer closure) to auto-derive their large downstream set instead
+# of listing it by hand. The test_include lists carry the downstream "Dependents"
+# that the reverse-dependency graph cannot derive on its own, so that (graph walk)
+# UNION (test_include) covers the dependent set from the "DEFCON Enablement:
+# Gating Levels" tracking page.
```
ScottTodd: What is the "DEFCON Enablement: Gating Levels tracking page"? (careful about notes that make sense _to an agent with context_ spilling into open source code that needs to stand on its own)

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+# --- Foundational runtimes: broad downstream dependent sets --------------------
+# rocr-runtime / hip-clr / amd-llvm sit under most of the tree. Their direct
+# graph consumers miss many real downstream projects (framework-facing libs, test
+# harnesses), so the DEFCON 4 dependent list is carried explicitly here.
```
ScottTodd: The framing of this comment is confusing. The amd-llvm compiler is used to compile ~all projects and it is very close to the root of the build graph, so of course the "direct graph consumers" list is short for it. Changes to the compiler should build _all_ projects that use the compiler, direct dependency or otherwise.

Note that deps on the compiler are carried via `COMPILER_TOOLCHAIN` (`amd-hip` or `amd-llvm`) and _not_ `BUILD_DEPS` or `TEST_DEPS`. For example: https://github.com/ROCm/TheRock/blob/bfe0b550286db235bd4d3b1f9dde226783c9d6d3/math-libs/CMakeLists.txt#L60-L73

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.rocr-runtime]
+# Broad blast radius: level 3 walks the full transitive consumer closure, which
+# auto-derives the bulk of the DEFCON 4 dependent set. Only the couplings the
+# graph has no path for are pinned below (amd-dbgapi, ocl-icd, rocgdb).
+level = 3
+test_include = ["amd-dbgapi", "ocl-icd", "rocgdb"]
```
ScottTodd, line 19: This comment is full of jargon. What does it actually mean? This file is designed to be reviewed and authored by humans so I would rephrase like

```suggestion
[component.rocr-runtime]
# ROCr is used by many projects so test self + transitive users.
level = 3
```

Then I don't see why amd-dbgapi, ocl-icd, or rocgdb should be included here. They don't depend on or use rocr-runtime as far as I can tell? If there's no path in the build graph then they shouldn't be added.

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.amd-llvm]
+test_include = [
+    "hip-tests",
+    "rccl",
+    "rocgdb",
+    "rocprofiler-compute",
+    "rocprofiler-sdk",
+    "rocprofiler-systems",
+]
```
ScottTodd: The LLVM compiler should build/test _everything_, not a narrow list like this.

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+test_include = ["hipfft", "rocfft", "rocprofiler-systems-examples"]
+
+[component.rocprofiler-register]
+test_include = ["hipfile"]
+
+[component.aqlprofile]
+test_include = ["rocprofiler-systems-examples"]
+
+[component.roctracer]
+test_include = ["hip-clr", "rocprofiler-sdk"]
```
ScottTodd, line 50: I looked at each of these and I'm not seeing why they need explicit `test_includes`. If they have dependencies on these components the build system *must* know about them for source builds to work as expected.

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+test_include = ["hipsolver"]
+
+[component.hipcub]
+test_include = ["rocprim"]
+
+[component.hiprand]
+test_include = ["miopen", "rocprim"]
+
+[component.hipsparse]
+test_include = ["rocsparse"]
```
ScottTodd: hipsparse _depends on rocsparse_, so this is wrong
* changes to rocsparse can affect hipsparse and could run hipsparse tests
* changes to hipsparse _cannot_ affect rocsparse and should not run rocsparse tests

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+[component.amd-dbgapi]
+test_include = ["rocgdb-cpu", "rocgdb-gpu"]
+
+[component.rocgdb]
+test_include = ["rocgdb-cpu", "rocgdb-gpu"]
+
+# --- Math / ML library couplings ----------------------------------------------
+
+[component.hipblas]
+test_include = ["hipsolver"]
```
ScottTodd, line 64: Why should changes to hipblas run hipsolver tests? Looking at https://github.com/ROCm/TheRock/blob/main/math-libs/BLAS/CMakeLists.txt:

hipSOLVER has deps on
* rocBLAS
* rocSOLVER

hipBLAS has deps on
* rocBLAS
* rocSOLVER

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.hipcub]
+test_include = ["rocprim"]
```
ScottTodd: Why should changes to hipcub run rocprim tests? hipcub has a dependency on rocprim

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.hiprand]
+test_include = ["miopen", "rocprim"]
```
ScottTodd, line 67: Why should changes to hiprand run rocprim tests?

miopen I think could be okay, though I wonder why hiprand specifically is included while MIOpen's other deps are not. Note that since MIOpen is a leaf of the build graph, building it requires building ~all math library projects https://github.com/ROCm/TheRock/blob/bfe0b550286db235bd4d3b1f9dde226783c9d6d3/ml-libs/CMakeLists.txt#L57-L64

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.hipsparse]
+test_include = ["rocsparse"]
+
+[component.libhipcxx]
+test_include = ["rocthrust"]
```
ScottTodd, line 70: Why should changes to libhipcxx test rocthrust? I don't see a dependency there

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.miopen]
+test_include = ["hipdnn_integration_tests", "hipdnn_samples"]
+
+[component.hipkernelprovider]
+test_include = ["hipdnn_integration_tests", "hipdnn_samples"]
+
+[component.rocthrust]
+test_include = ["rocprim"]
```
ScottTodd: rocthrust has a dependency on rocprim, it should not run rocprim tests

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.rocroller]
+test_include = ["hipdnn", "miopen"]
```
ScottTodd, line 79: I don't see a dependency on rocroller in either hipdnn or miopen. Why should changes to rocroller run those tests?

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.rocrand]
+test_include = ["hipfft", "rocfft", "rocprim"]
```
ScottTodd, line 82: These includes look inverted too -
rocFFT depends on hipRAND which depends on rocRAND, should not run rocfft tests on changes to rocrand

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+[component.rocsparse]
+test_include = ["rocblas", "rocprim", "rocsolver"]
```
ScottTodd, line 85: * rocsparse depends on rocblas, should not run rocblas tests on changes to rocsparse
* rocsparse depends on rocprim, should not run rocprim tests on changes to rocsparse
* I'm not seeing the dep from rocsolver to rocsparse

cmake/therock_subproject.cmake
```diff
@@ -552,6 +550,14 @@ function(therock_cmake_subproject_declare target_name)
     THEROCK_FPRINT_SOURCE_HASH "${ARG_FPRINT_SOURCE_HASH}"
   )
 
+  # Register this subproject in the global consumer registry so that
+  # therock_emit_consumer_graph() can later derive "who to test when X changes"
+  # without requiring a hand-maintained TEST_SUBPROJECTS list.
+  set_property(GLOBAL APPEND PROPERTY THEROCK_ALL_SUBPROJECTS "${target_name}")
+  foreach(_dep IN LISTS ARG_BUILD_DEPS ARG_RUNTIME_DEPS)
+    set_property(GLOBAL APPEND PROPERTY "THEROCK_CONSUMERS_OF_${_dep}" "${target_name}")
+  endforeach()
```
ScottTodd: We could qualify this further as `THEROCK_DIRECT_CONSUMERS_OF_`, indicating that to get the full set of consumers you also need to walk the graph

.github/workflows/consumer_graph_drift.yml
```diff
@@ -0,0 +1,65 @@
...
+name: Consumer Graph Drift Check
+
+on:
+  pull_request:
+    paths:
+      - "**/CMakeLists.txt"
+      - "cmake/therock_emit_consumer_graph.cmake"
+      - "cmake/therock_subproject.cmake"
+      - "test_tools/therock_consumer_graph.json"
+      - ".github/workflows/consumer_graph_drift.yml"
```
ScottTodd: I'm not thrilled about introducing a new workflow with its own triggering for this check given how many workflows we already have in https://github.com/ROCm/TheRock/tree/main/.github/workflows. I'd prefer to extend an existing workflow instead, but there aren't many great candidates for a job that only sometimes needs to run, needs to fetch all sources, needs to configure CMake but not build, and wants to report a quick status.

Existing entry point options would be:
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml
  * Our standard CI workflow that runs on PRs as well as postsubmit. We could add a new job to that workflow that calls this workflow, similar to how https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_artifacts_structure.yml is run here: https://github.com/ROCm/TheRock/blob/820c35cb91ece10eb753127d601ed51bed679f19/.github/workflows/multi_arch_ci_linux.yml#L116-L126 However, we'd probably want to apply this style of path filtering to that job, and that's easier to do from a new workflow like this
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/unit_tests.yml
  * This should complete in <5 minutes, so adding a ~10 minute job (due to how long sources take to fetch) isn't great (however, if we optimize checkout somehow then it would be viable in there)
* https://github.com/ROCm/TheRock/blob/main/.github/workflows/pre-commit.yml
  * This should complete in <1 minute, and it relies on https://github.com/ROCm/TheRock/blob/main/.pre-commit-config.yaml which runs on _every commit_ locally too, ideally "you edited cmake files but didn't regenerate the .json" would be a super fast pre-commit compatible check, but as this relies on the full CMake configure to produce the file that's a bit out of reach right now

So... I guess this is fine, but let's keep an eye on it.

.github/workflows/consumer_graph_drift.yml
ScottTodd, line 1: We could at least name this file `test_consumer_graph_draft.yml` so it sorts alphabetically next to some other workflows that "test" different properties of the project in https://github.com/ROCm/TheRock/tree/main/.github/workflows

.github/workflows/consumer_graph_drift.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: python3 ./build_tools/fetch_sources.py --jobs 12 --depth 1
+
+      - name: Configure (emits the consumer graph)
+        # Configure only, no build. This emits build/therock_consumer_graph.json as a
+        # configure-time side effect. THEROCK_ENABLE_ALL=ON ensures every subproject
+        # is declared so the emitted graph matches the committed full-tree copy.
+        run: |
+          cmake -B build -S . -GNinja \
+            -DTHEROCK_ENABLE_ALL=ON \
+            -DTHEROCK_AMDGPU_FAMILIES=gfx110X-all
```
ScottTodd: If any projects are enabled/disabled based on the GPU targets/families then this configure will not model those dependencies. At least for gfx110X, there are exclusions for `hipSPARSELt` and `rocprofiler-compute`: https://github.com/ROCm/TheRock/blob/820c35cb91ece10eb753127d601ed51bed679f19/cmake/therock_amdgpu_targets.cmake#L193-L198

How about switching to `gfx94X-dcgpu`? https://github.com/ROCm/TheRock/blob/820c35cb91ece10eb753127d601ed51bed679f19/cmake/therock_amdgpu_targets.cmake#L92-L96

Unfortunately the only targets that include support for all subprojects are Linux-only, so Windows developers (like meeee) may find it hard to work with this so long as it requires actually running cmake configure to completion. Maybe we could add a simulation mode that does a faster walk through the dep graph without actually needing the projects sources to exist or to be buildable on the current platform.

.github/workflows/unit_tests.yml
```diff
@@ -37,6 +37,14 @@ jobs:
     - name: Install python deps
       run: pip install -r requirements-test.txt
 
+    # Referential-integrity check for test_tools/test_policies.toml against the
+    # committed consumer graph. This runs on EVERY PR (unlike the path-gated
+    # consumer_graph_drift.yml) because a stale policy key silently disables test
+    # selection for that component. It needs only the two committed files — no
+    # source fetch, no cmake configure. See RFC0013 "CI integration".
+    - name: Validate test policies against consumer graph
+      run: python test_tools/determine_rocm_test_dependencies.py --validate-policies
```
ScottTodd: I don't want special cases in this file. Running all project unit tests should be as simple as possible, with a common test runner.

Let's have https://github.com/ROCm/TheRock/blob/main/test_tools/tests/determine_rocm_test_dependencies_test.py check what this `--validate-policies` checks and then make sure that `test_tools/tests/` is running here as part of
* https://github.com/ROCm/TheRock/issues/6927

build_tools/github_actions/check_consumer_graph_drift.py
```diff
@@ -0,0 +1,124 @@
...
+
+import argparse
+import difflib
+import json
+import sys
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_COMMITTED_PATH = _REPO_ROOT / "test_tools" / "therock_consumer_graph.json"
+_EMITTED_PATH = _REPO_ROOT / "build" / "therock_consumer_graph.json"
```
ScottTodd: There is no hard requirement that local builds use this path or that there is only one build directory, so let's make this build directory configurable with a `--build-dir` argument and/or environment variable (argparse preferred).

build_tools/github_actions/check_consumer_graph_drift.py
```diff
@@ -0,0 +1,124 @@
...
+    print(
+        "A dependency edge changed but the committed consumer graph was not "
+        "regenerated."
+    )
+    print("Fix: run a full-tree configure, then regenerate and re-commit:")
+    print("    python3 ./build_tools/fetch_sources.py")
+    print(
+        "    cmake -B build -GNinja -DTHEROCK_ENABLE_ALL=ON "
+        "-DTHEROCK_AMDGPU_FAMILIES=gfx110X-all"
+    )
```
ScottTodd, line 57: Also update the recommended family to gfx94X-dcgpu here if addressing my related comment

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,435 @@
...
+    return result
 
-            if test_subprojects_match:
-                deps_str = (test_subprojects_match.group(1) or "").strip()
-                deps = [d.strip().lower() for d in deps_str.split() if d.strip()]
-                test_deps[subproject_name] = deps
 
-    return test_deps
+def get_subprojects_to_test(
+    changed_subprojects: list[str],
```
ScottTodd, line 199: What is this function expecting for "changed_subprojects"? My review agent caught this:

> The old stage-map mismatch was removed, but the end-to-end pipeline still uses
> at least three different identifiers without an authoritative mapping:
> 
> * [`configure_external_repo_ci.py`](https://github.com/ROCm/TheRock/blob/727e4423ba674347d331671ed1aef3c70643d1bc/build_tools/github_actions/configure_external_repo_ci.py#L290-L296)
>   emits external-repository subtree paths such as `projects/clr` and
>   `shared/rocroller`.
> * [`determine_rocm_test_dependencies.py`](https://github.com/ROCm/TheRock/blob/727e4423ba674347d331671ed1aef3c70643d1bc/test_tools/determine_rocm_test_dependencies.py#L466-L485)
>   strips only `projects/`, then looks the resulting string up in a graph keyed
>   by CMake target names such as `hip-clr`, `rocroller`, and
>   `hipdnn_integration_tests`.
> * [`fetch_test_configurations.py`](https://github.com/ROCm/TheRock/blob/727e4423ba674347d331671ed1aef3c70643d1bc/build_tools/github_actions/fetch_test_configurations.py#L835-L865)
>   schedules jobs only on an exact match with test-matrix keys such as
>   `hipdnn-integration-tests` and `hipdnn-samples`.
> 
> Concrete current-head results:
> 
> ```text
> projects/clr                     -> warning; ["clr"]
> shared/rocroller                 -> warning; ["shared/rocroller"]
> projects/hipdnn-integration-tests -> warning; ["hipdnn-integration-tests"]
> ```
> 
> The first two select no component job beyond sanity. The policy has the same
> problem: [`test_policies.toml`](https://github.com/ROCm/TheRock/blob/727e4423ba674347d331671ed1aef3c70643d1bc/test_tools/test_policies.toml#L121-L129)
> adds `hipdnn_integration_tests` and `hipdnn_samples` for MIOpen and
> hipkernelprovider, while the real matrix keys are
> [`hipdnn-integration-tests` and `hipdnn-samples`](https://github.com/ROCm/TheRock/blob/727e4423ba674347d331671ed1aef3c70643d1bc/build_tools/github_actions/fetch_test_configurations.py#L546-L560).
> An end-to-end local run for a MIOpen change scheduled `miopen` and
> `miopenprovider`, but silently omitted both intended hipDNN suites.
> 
> The unit tests stop at the selector's returned set, so they accept names that
> cannot create a real CI job.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,435 @@
...
+
+    A missing file yields empty policies (so every component uses the default
+    level and no include/exclude) rather than crashing — useful for clean-checkout
+    introspection, though the file is committed and normally present.
     """
-    test_deps = {}
-    cmake_files = list(Path(therock_dir).rglob("CMakeLists.txt"))
+    policies_path = _test_policies_path(therock_dir)
+    if not policies_path.exists():
+        return {}
```
ScottTodd: Seems like a missing file should raise an exception rather than default to empty policies: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#fail-fast-behavior

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,435 @@
...
+Level ladder (walk depth over consumer edges)
+---------------------------------------------
+* Level 5 -> depth 0: self only.
+* Level 4 -> depth 1: self + direct consumers (one hop). The default.
+* Level 3 -> unbounded: self + transitive consumers (full BFS closure).
+
+The per-component level, plus test_include/test_exclude, come from the
+hand-maintained `test_tools/test_policies.toml`. A component with no table there
+uses the default level 4 (depth 1) with no include/exclude. See the schema
+contract in docs/development/test_selection_schemas.md.
```
ScottTodd: `test_selection_schemas.md` does not exist. Redirect to the RFC in `docs/rfcs`?

docs/rfcs/RFC0013-Consumer-Based-Test-Selection.md
```diff
@@ -0,0 +1,458 @@
...
+machine-generated and must never be hand-edited — edits would be clobbered on
+regeneration, and the whole point is that it cannot silently drift from the build
+definitions. The policy file is human intent that CI cannot derive. Combining them
+would create a file that is half-generated and half-authored, which the drift check
+cannot cleanly "regenerate and diff": it would have to diff *around* the
+hand-authored regions. Keeping them apart lets the graph be fully CI-owned
+(regenerate-and-fail-on-diff) and the policy be fully human-owned (reviewed like
+code). This mirrors ScottTodd's "generated from build-system knowledge, with a few
+escape hatches" framing: the graph is the generated knowledge, the policy file is
+the escape hatches.
```
ScottTodd: nit: can drop the reference to my specific username and say instead "during review we decided..." or similar ;) the full discussion history is here on the PR

docs/rfcs/RFC0013-Consumer-Based-Test-Selection.md
```diff
@@ -0,0 +1,458 @@
...
+- **Cross-repo selection.** How should the graph interact with external-repo CI
+  (rocm-systems / rocm-libraries) where only a subset of source is present at
+  configure time?
+- **Test tiers for levels 1/2/5.** The depth ladder covers levels 3-5 by
+  selection radius, but true level 5 (unit-only) and level 1 (full QA / nightly)
+  need a per-project `test_tier` output the test-runner job honors. Should that
+  tier plumbing be part of this feature or a dedicated follow-up RFC?
+- **Level naming in open source.** The "DEFCON" label is internal. Should the
+  public RFC and code use a different name for the gating levels? (Pending input
+  from @ScottTodd / @astyrrian1.)
```
ScottTodd: lol, @astyrrian1 @suchitravemuri1  how you feel about us using "DEFCON" as the public name for these test tiers?

ScottTodd (commented):
Architecturally LGTM. A mix of comments on style, specific policies to apply to individual subprojects for correctness, and some details on how to integrate into our existing infrastructure then I'll be okay with approving.

docs/rfcs/RFC0013-Consumer-Based-Test-Selection.md
endurthiabhilash: My bad, I cleaned up all internal references, other internal comments but, clearly some escaped. Let me fix.

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: Ack

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: Ack

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: Earlier I was using an test override to mimic changed subprojects in `rocm-systems`, `rocm-libraries`. I removed that logic from here, created another test branch off of it for my e2e testing.

test_tools/test_policies.toml
endurthiabhilash: This makes sense, I will update to use COMPILER_TOOLCHAIN for building consumer graph for compilers.

test_tools/test_policies.toml
endurthiabhilash: Internal page, will remove.

test_tools/test_policies.toml
endurthiabhilash: My bad, I cleaned up all internal references, other internal comments but, clearly some escaped. Let me fix.

test_tools/test_policies.toml
endurthiabhilash: The override data for test_policies.toml is from an internal consumer document. We have got consumer information from component teams. Whatever components were not captured as part of the consumer graph, I added them as explicit include. I will remove override data for now (while still keeping the structure) and confirm with development teams.

test_tools/test_policies.toml
endurthiabhilash: The override data for test_policies.toml is from an internal consumer document. We have got consumer information from component teams. Whatever components were not captured as part of the consumer graph, I added them as explicit include. I will remove override data for now (while still keeping the structure) and confirm with development teams.

test_tools/test_policies.toml
endurthiabhilash: The override data for test_policies.toml is from an internal consumer document. We have got consumer information from component teams. Whatever components were not captured as part of the consumer graph, I added them as explicit include. I will remove override data for now (while still keeping the structure) and confirm with development teams.

test_tools/test_policies.toml
endurthiabhilash: I'll check the internal document, remove this override.

test_tools/test_policies.toml
endurthiabhilash: I'll check the internal document, remove this override.

test_tools/test_policies.toml
endurthiabhilash: This was given by a component owner. Let me confirm the dependency direction and fix.

test_tools/test_policies.toml
endurthiabhilash: I'll check the internal document, remove this override.

test_tools/test_policies.toml
endurthiabhilash: Let me check and fix

test_tools/test_policies.toml
endurthiabhilash: Ack, will fix

.github/workflows/consumer_graph_drift.yml
endurthiabhilash: Agree, I'll rename it.

.github/workflows/unit_tests.yml
endurthiabhilash: Sure, will remove it from unit_tests.yml, add it to determine_rocm_test_dependencies_test.py

build_tools/github_actions/check_consumer_graph_drift.py
endurthiabhilash: Ack

test_tools/test_policies.toml
endurthiabhilash: I'll check the internal document, remove this override.

test_tools/test_policies.toml
ScottTodd: Actually, maybe something got swapped / lost in translation for these. We probably want changes to rocsparse to run hipsparse builds and tests, just not the other way around. Similarly for some of the other entries here (in general "hip" projects depend on "roc" projects, then a few other projects are closely related and can be expected to run downstream tests, _not upstream tests_ regularly)

lumachad:
Just so I understand the impact of this change on ROCgdb CI and what we're trying to achieve here. We're throwing away TEST_SUBPROJECTS entirely and replacing it with another mechanism that does things automatically?

I went through the generated consumer graph json and there might be a gap here. Take for instance ROCgdb. ROCgdb does not depend directly on rocr-runtime, but when ROCgdb runs its tests that definitely depends on rocr-runtime. In fact that's been a frequent area of uncaught regressions from rocm-systems changes.

Could we please make sure we have that covered?

Another point is that we have a ROCgdb testing split to optimize capacity usage. rocgdb-cpu tests only have to be exercised due to changes in the amd-llvm compiler. rocgdb-gpu needs to be exercised whenever amd-llvm or rocr-runtime/clr change etc. Do we have any provisions to cater for that split so we don't run useless tests when not required?

test_tools/test_policies.toml
```diff
@@ -0,0 +1,140 @@
...
+
+# rocgdb-cpu / rocgdb-gpu are ctest test-tier targets of rocgdb, not declared
+# subprojects, so they are not graph keys -- but that is exactly what test_include
+# is for (test-only targets the graph cannot express). Kept as include values.
+
+[component.amd-dbgapi]
+test_include = ["rocgdb-cpu", "rocgdb-gpu"]
+
+[component.rocgdb]
+test_include = ["rocgdb-cpu", "rocgdb-gpu"]
```
ScottTodd, line 59: > Just so I understand the impact of this change on ROCgdb CI and what we're trying to achieve here. We're throwing away TEST_SUBPROJECTS entirely and replacing it with another mechanism that does things automatically?
> 
> I went through the generated consumer graph json and there might be a gap here. Take for instance ROCgdb. ROCgdb does not depend directly on rocr-runtime, but when ROCgdb runs its tests that definitely depends on rocr-runtime. In fact that's been a frequent area of uncaught regressions from rocm-systems changes.
> 
> Could we please make sure we have that covered?
> 
> Another point is that we have a ROCgdb testing split to optimize capacity usage. rocgdb-cpu tests only have to be exercised due to changes in the amd-llvm compiler. rocgdb-gpu needs to be exercised whenever amd-llvm or rocr-runtime/clr change etc. Do we have any provisions to cater for that split so we don't run useless tests when not required?

I would expect:

* This policy says "when rocgdb is modified, run rocgdb-cpu and rocgdb-gpu"
* A policy for rocr-runtime of
    ```toml
    [component.rocr-runtime]
    # ROCr is used by many projects so test self + transitive users
    level = 3
    ```
    says "when rocr-runtime is changed, test every project that depends on rocr-runtime directly _or indirectly_"
* A policy for rocr-runtime of
    ```toml
    [component.rocr-runtime]
    test_include = ["rocgdb"]
    ```
    (using the default level) would instead say "when rocr-runtime is changed, test rocr-runtime itself and the one additional project 'rocgdb'"

As for the cpu/gpu split, I don't see a clear spot for that in the current design. We could add some additional tiers of tests along those lines 🤔

.github/workflows/consumer_graph_drift.yml
```diff
@@ -0,0 +1,65 @@
...
+      - "cmake/therock_emit_consumer_graph.cmake"
+      - "cmake/therock_subproject.cmake"
+      - "test_tools/therock_consumer_graph.json"
+      - ".github/workflows/consumer_graph_drift.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  consumer_graph_drift:
```
geomin12: let's add test as if it is making sure it is not drifting

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,435 @@
...
 sys.path.insert(
     0, str(Path(__file__).resolve().parents[1] / "build_tools" / "github_actions")
 )
 from github_actions_api import gha_set_output
 
+# Level -> BFS walk depth over `consumers` edges. depth 0 = self only,
+# depth 1 = self + direct consumers, None = unbounded (transitive closure).
+# Levels 1 and 2 are the "stricter/broader" tiers on the DEFCON ladder; until a
+# separate test-tier output exists they walk transitively (same depth as 3), so
+# the accepted policy range [1, 5] and the engine never disagree.
```
geomin12: can we do a better comment?

Maybe a 
```
depth 0: 
depth 1: 
depth 2:
```

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -138,45 +445,44 @@ def main():
...
     args = parser.parse_args()
+    therock_dir = Path(args.therock_dir).resolve() if args.therock_dir else None
 
-    therock_dir = Path(args.therock_dir).resolve()
+    if args.validate_policies:
+        ok, messages = validate_policies(therock_dir)
+        stream = sys.stdout if ok else sys.stderr
+        for line in messages:
+            print(line, file=stream)
+        sys.exit(0 if ok else 1)
```
geomin12: maybe instead of sys.exit, could we raise an Exception to indicate policy is not ok?

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,435 @@
...
+        )
+    lines.append(f"  level:        {level} (walk depth {depth_label}: {depth_desc})")
+    lines.append("  graph walk:   " + ", ".join(sorted(walk)))
+    if include:
+        marked = [f"{v} (test-only)" if v not in known else v for v in sorted(include)]
+        lines.append("  test_include: " + ", ".join(marked))
+    else:
+        lines.append("  test_include: (none)")
+    if exclude:
+        lines.append("  test_exclude: " + ", ".join(sorted(exclude)))
```
geomin12: could we use \t instead of spaces?

geomin12 (commented):
first pass

test_tools/test_policies.toml
endurthiabhilash: For cpu, gpu split, can we add explicit excludes for amd-llvm, hip-clr changes like below? I need to test, but this should work. If we need to add more specific tiers of tests then we can do that in the next iteration once this is merged.

```
[component.amd-llvm]
test_include = ["rocgdb-cpu"]
test_exclude = ["rocgdb-gpu"]
```

lumachad: Thanks for the examples. As long as we can specify these dependencies clearly (or if they get resolved automatically, even better), this should work for us. The split is more of an optimization of capacity use, not a hard requirement.

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: done

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: done

test_tools/determine_rocm_test_dependencies.py
endurthiabhilash: done

.github/workflows/consumer_graph_drift.yml
endurthiabhilash: Edge drift is caught by test_check_fails_on_edge_change in the unit tests, and by this workflow regenerating from a full configure and diffing. Let me know if you are thinking of other cases that I missed.

test_tools/determine_rocm_test_dependencies.py
```diff
@@ -2,128 +2,431 @@
...
 from github_actions_api import gha_set_output
 
+# Level -> BFS walk depth over `consumers` edges:
+#   depth 0    = self only
+#   depth 1    = self + direct consumers
+#   unbounded  = self + transitive consumers (full closure)
+# Levels 1 and 2 walk transitively (same as 3) until a separate test-tier output
+# distinguishes them.
+_DEFAULT_LEVEL = 4
+_LEVEL_TO_DEPTH: dict[int, int | None] = {5: 0, 4: 1, 3: None, 2: None, 1: None}
```
dileepr1, line 57: As per this, levels 1 and 2 also will get same treatment as level 3 - a transitive walk. 
As per RFC0013:
Level 2 = self + a named frameworks group 
Level 1 = everything (projects_to_test = "*") + a full-QA / nightly workflow dispatch

endurthiabhilash: Yes, this PR is enabling only levels 3-5. So, temporarily levels 1, 2 are falling back to level 3. I'll add a TODO flag here to implement levels 1, 2 in the next iteration as improvement.

test_tools/test_policies.toml
ScottTodd: Want to adjust these includes as part of this PR or a follow-up?

ScottTodd (approved):
Looks complete enough to land and iterate.

You may want to sync the PR branch and let CI run again prior to merge. There was a github actions outage yesterday that caused jobs to not trigger. The code here *mostly* doesn't affect CI yet though as far as I can tell so it could be safe.

endurthiabhilash:
> Looks complete enough to land and iterate.
> 
> You may want to sync the PR branch and let CI run again prior to merge. There was a github actions outage yesterday that caused jobs to not trigger. The code here _mostly_ doesn't affect CI yet though as far as I can tell so it could be safe.

Synced the branch, also cleaned up some reverse edges in test_policies.toml. Any remaining ones will be cleaned up in next iteration. Do I also need approval from @ROCm/rocm-product-managers to merge this PR?

ScottTodd:
> Do I also need approval from @ROCm/rocm-product-managers to merge this PR?

Technically yes, but admins could also merge through that requirement. The revirew request there is due to CODEOWNERS for `docs/rfcs`, see https://github.com/ROCm/TheRock/pull/5424. @saadrahim do you want to review the RFC about test selection here? We could also move the file to https://github.com/ROCm/TheRock/tree/main/docs/design to get around that review requirement if this doesn't feel as much like a product RFC so much as an engineering design doc. Not sure where we want to draw the lines there.

endurthiabhilash:
> > Do I also need approval from @ROCm/rocm-product-managers to merge this PR?
> 
> Technically yes, but admins could also merge through that requirement. The revirew request there is due to CODEOWNERS for `docs/rfcs`, see #5424. @saadrahim do you want to review the RFC about test selection here? We could also move the file to https://github.com/ROCm/TheRock/tree/main/docs/design to get around that review requirement if this doesn't feel as much like a product RFC so much as an engineering design doc. Not sure where we want to draw the lines there.

I have re-triggered CI after syncing the branch. All the required checks pass. I see some non-blocking test failures, but the PR should be safe to merge. cc: @geomin12, @suchitravemuri1 

@ScottTodd, will you be able to override @ROCm/rocm-product-managers approval requirement and merge the PR?

ScottTodd:
Sure. If there are concerns about the RFC file, post-merge feedback is welcome.

test_tools/test_policies.toml
ScottTodd: Another option to consider for cheap/fast CPU unit tests is `therock_cmake_subproject_build_test` which runs _during the build itself_, on CPU runners. That runs in parallel to other build jobs and does not need to wait until all other subproject builds run, the GPU sanity check job is scheduled, etc.

ScottTodd: Yeah that may be a better fit for rocgdb-cpu tests instead of building so many branches into the existing GPU testing scripts and workflows. See also https://github.com/ROCm/TheRock/issues/4789

PR #6944, opened 2026-07-28
Enable RUNPATH to RPATH conversion by default in packaging

raramakr:
- Call convert_runpath_to_rpath on artifacts before packaging starts
  - RPATH conversion is now enabled by default; use --no-rpath-pkg to disable
  - Remove enable_rpath from PackageConfig (not needed in config)
  - Remove pyelftools dependency (script uses system readelf/patchelf)
  - Update native_packaging.md documentation to reflect new behavior
  
  JIRA ID: https://github.com/ROCm/TheRock/issues/6946

PR #6951, opened 2026-07-29
test(packaging): add runpath_to_rpath unit tests

arvindcheru:
## Motivation

Add unit test coverage for _get_rpath() and update_rpath() using real ELF fixtures built from hello_lib.c. Tests exercise readelf parsing and patchelf conversion paths, including the remove-then-set workflow for older patchelf versions.

## Technical Details

 Adds module-level unit tests for `build_tools/packaging/linux/runpath_to_rpath.py` under `build_tools/packaging/linux/tests/`.
- Builds real `DT_RUNPATH` and `DT_RPATH` fixture libraries in test setUp and validates:
  - `_get_rpath()` reads rpath values from both tag types via readelf
  - `update_rpath()` preserves rpath values across conversion
  - DT_RUNPATH → DT_RPATH conversion via patchelf remove-then-set (reference path for older patchelf)
- Adds `hello_lib.c` as minimal shared-library source for fixtures.

**Sequencing:** #6944 landed the patchelf/readelf rewrite of `runpath_to_rpath.py`.
This PR adds tests only — no script modifications.

Post-install verification on packaged ROCm libraries are tracked in #7035.

## Test Plan

Prerequisites (Ubuntu 24.04):
```bash
sudo apt install -y build-essential binutils patchelf python3
```
- [x] Local test run (Ubuntu 24.04 Docker, patchelf 0.18.0):
python3 build_tools/packaging/linux/tests/runpath_to_rpath_test.py -v

## Test Result

Result: 3 tests — 2 ok, 1 skipped (patchelf 0.10; expected), exit 0
- [x] Re-run after #6944 merges / when stacked on updated base

## Submission Checklist


build_tools/packaging/linux/tests/runpath_to_rpath_test.py
```diff
@@ -0,0 +1,144 @@
...
+    python3 build_tools/packaging/linux/tests/runpath_to_rpath_test.py -v
+
+  From build_tools/packaging/linux/tests::
+
+    python3 runpath_to_rpath_test.py -v
+
+Files in this directory
+-----------------------
+  runpath_to_rpath_test.py  - this test module
+  hello_lib.c               - minimal shared-library source built in setUp
```
ScottTodd: These comments aren't useful.

* The "how to run" is just "how does unittest/pytest work?"
* The "Files in this directory" for "this test module" is obvious from the file name itself

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
ScottTodd, line 1: Let's make sure that tests in this directory are running continuously on CI before adding any new tests. Tests that aren't run on CI have only marginal utility. See https://github.com/ROCm/TheRock/issues/6927.

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
```diff
@@ -0,0 +1,144 @@
...
+    runpath_so = out_dir / "lib_runpath.so"
+    rpath_so = out_dir / "lib_rpath.so"
+    common = ["gcc", "-shared", "-fPIC", "-Wl,-rpath," + RPATH_VALUE]
+    subprocess.check_call(
+        common + ["-Wl,--enable-new-dtags", "-o", str(runpath_so), str(FIXTURE_SRC)]
+    )
+    subprocess.check_call(
+        common + ["-Wl,--disable-new-dtags", "-o", str(rpath_so), str(FIXTURE_SRC)]
+    )
+    return runpath_so, rpath_so
```
ScottTodd, line 77: Is this test representative enough? This is a unit test for the `runpath_to_rpath` module that doesn't check that the libraries we build as part of ROCm itself (mostly compiled via clang) get handled correctly.

Is there also an integration test? That would be useful to mention in the docstring at the top of the file.

ScottTodd (approved):
Generally lgtm, but I want to make sure that any new tests we add are running continuously on CI.

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
arvindcheru: Thank you for the comments.

The gcc fixtures are intentional for now — they exercise the DT_RUNPATH → DT_RPATH readelf/patchelf path that #6944 fixed, using small controllable ELFs instead of full artifact trees in unit CI. The same logic applies whether the library was linked with gcc or clang. End-to-end checks on installed ROCm packages are tracked in [#7035](https://github.com/ROCm/TheRock/pull/7035); we will check this and update

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
arvindcheru: Updated docstring in latest version. Thanks

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
arvindcheru: Thank you for the comments. 
The new tests are part of the Unit Tests CI
On ubuntu-24.04 CI, all three tests pass under the #7004 directory testpaths. Windows skips (no Linux ELF toolchain), which matches other Linux-only packaging tests.

build_tools/packaging/linux/tests/runpath_to_rpath_test.py
arvindcheru: Renamed the framing to **unit tests** (module-level, gcc fixtures + readelf/patchelf)
so it is distinct from install-level integration. End-to-end checks on installed
ROCm packages remain in #7035; updated docstrings accordingly.

PR #6956, opened 2026-07-29
fix(jax): match jax_rocm plugin wheels by ROCm major version

marifamd:
## Summary

The JAX multi-arch build fails at the version-detection step with:

```
FileNotFoundError: Did not find jax_rocm7_plugin wheel
```

`build_tools/github_actions/write_jax_versions.py` hardcodes `jax_rocm7_plugin` and `jax_rocm7_pjrt`, but the JAX 0.10.0 build now emits `jax_rocm10_plugin` / `jax_rocm10_pjrt`. The ROCm major version embedded in the wheel name tracks the ROCm release, so a fixed `7` no longer matches.

## Fix

Detect the wheels with a `jax_rocm*_plugin` / `jax_rocm*_pjrt` glob so the step tracks whatever ROCm major version the build produces (rocm7, rocm10, and future), instead of a hardcoded major. Docstring and error messages updated to match.

## Scope

Limited to the failing build-time detection logic. The other `jax_rocm7_*` references live in the promotion/download allowlists (`download_python_packages.py`, `promote_packages.py`, `s3_management/manage.py`), which are separate jobs and intentionally explicit; they are left out of this change.

## Test

- `pre-commit run --files build_tools/github_actions/write_jax_versions.py` passes (black, etc.).
- Glob verified locally against real 0.10.0 wheel names: `jax_rocm*_plugin-*.whl` matches `jax_rocm10_plugin-0.10.0+devrocm10.0.0-...whl` while the old `jax_rocm7_plugin-*.whl` matches nothing (reproducing the failure).

## CI

This is a CI script change; the JAX build job exercises it.

### DIFF
--- a/build_tools/github_actions/write_jax_versions.py
+++ b/build_tools/github_actions/write_jax_versions.py
@@ -4,17 +4,20 @@
 
 """Writes jaxlib_version, jax_plugin_version, jax_pjrt_version, and jax_version to GITHUB_OUTPUT.
 
-Fails if required JAX wheels (jax_rocm7_plugin, jax_rocm7_pjrt) are not found.
+Fails if required JAX wheels (jax_rocm<major>_plugin, jax_rocm<major>_pjrt) are
+not found. The ROCm major version in the wheel name tracks the ROCm release
+(e.g. jax_rocm7_* for ROCm 7, jax_rocm10_* for ROCm 10), so the names are
+matched with a glob rather than a hardcoded major version.
 
 For JAX <= 0.9.0, jaxlib is built and expected in the wheelhouse.
 For JAX >= 0.9.1, jaxlib is not built - it is installed from upstream PyPI
-(e.g. `pip install jaxlib==0.9.1`). Only jax_rocm7_plugin and jax_rocm7_pjrt
-are built.
+(e.g. `pip install jaxlib==0.9.1`). Only jax_rocm<major>_plugin and
+jax_rocm<major>_pjrt are built.
 
 Expected wheels:
 * jaxlib (not built for JAX >= 0.9.1)
-* jax_rocm7_plugin
-* jax_rocm7_pjrt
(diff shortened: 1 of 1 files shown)


amd-chiranjeevi:
will this solution work for jax v0.9.1..
whl generated here has the prefix `jax_rocm1_plugin`
https://github.com/ROCm/rockrel/actions/runs/30419729909/job/90473961488#step:9:16375

marifamd:
Resolves fails in https://github.com/ROCm/TheRock/pull/6960

gulsumgudukbay (changes requested):
The jax_rocm1_* naming only happens on the 0.9.1 path. That branch builds through ROCm/rocm-jax's wrapper, which passes the truncated value straight through:

build/build.py build --wheels=jax-rocm-plugin --rocm_path=/opt/rocm-10.0.0a20260729 --rocm_version=1 ...
The 0.10.0 path never passes --rocm_version at all and it derives the namespace inside the JAX build from ROCM_PATH via rocm-sdk path --root, which is why it correctly produces jax_rocm10_*. Since #6708 leaves the release set at 0.10.0 through 0.11.0 and deletes the native flow, the only code path that can emit jax_rocm1_* goes away with it.

gulsumgudukbay (approved):

gulsumgudukbay:
This fix is necessary but not sufficient, test_linux_jax_wheels_partial.yml:240 and test_linux_jax_wheels.yml:253 still pip install "jax_rocm7_plugin==...", plus the allowlists in download_python_packages.py, s3_management/manage.py, and promote_packages.py, so the build will go green and the next stage will fail. Second, the small robustness point: get_wheel_version takes wheel_paths[0] in filesystem order when the glob matches more than one wheel, so sorting or validating the major against the ROCm version would make it deterministic.

PR #6967, opened 2026-07-29
Add RFC0014 for code coverage infrastructure in TheRock

reboss:
Proposes per-project coverage builds with isolation to prevent profraw contamination from upstream dependencies, maintaining separate coverage pipeline from pre-checkin testing.

https://github.com/ROCm/TheRock/issues/6572

### DIFF
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -103,10 +103,20 @@ TheRock CI splits build nodes and test nodes and does not retain the entire CMak
 ### Solution for Most Projects
 For most projects, the solution is straightforward:
 1. Instrument code during build
-2. Run tests using the Test Filter Standard (test_categories.yaml)
-3. Minor modifications to test_categories_coverage.yaml:
-   - Add `llvm_profile_file:` key to control profraw output location
-   - Modify test_runner.py to handle coverage mode
+2. Run tests using the Test Filter Standard (test_categories.yaml or test_categories_coverage.yaml)
+3. Schema additions to test_categories_coverage.yaml:
+   - **`llvm_profile_file`**: Path pattern for profraw file output (with `%m-%p` interpolation)
+   - **`ignore-filename-regex`**: Pattern to exclude test code from coverage reports
+   - **`test_names`**: List of test binary names for `-object` flags (header-only libraries only)
+
+**Declarative approach:** These fields can potentially be calculated at the top level rather than explicitly configured per-project:
+- **`test_names`**: Discovered by listing test executables matching `test_*` in a standardized test directory
+- **`llvm_profile_file`**: Hardcoded to standard path (e.g., `./coverage-report/%m-%p.profraw`)
+- **`ignore-filename-regex`**: Set to exclude anything outside the src directory (e.g., patterns matching test files)
+
+This reduces per-project configuration overhead and ensures consistency across components.
+
+**Error handling:** If profraw files are missing entirely, the coverage job fails. This indicates a fundamental problem with instrumentation or test execution.
 
 ### Problematic: Header-Only Static Libraries
 Some components are header-only static libraries (e.g., rocPRIM, hipCUB) and do not produce a simple `.so` file for `llvm-cov show -object <lib>.so`.
@@ -117,20 +127,21 @@ Some components are header-only static libraries (e.g., rocPRIM, hipCUB) and do 
 - Requires explicit reference to test binaries rather than library objects
 - **The `-object` flag does not accept wildcards** - cannot use `-object test_*`, must expand to `-object test_1 -object test_2 ... -object test_n`
 
-**Why test_categories_coverage.yaml may be needed:**
-Because `llvm-cov show -object` does not accept wildcard patterns, we must explicitly list each test binary. This requires either:
-1. **Separate test_categories_coverage.yaml** that explicitly lists all test binaries for coverage processing
-2. **Declarative approach with standard naming** - if tests follow `test_*` convention:
-   ```bash
-   # Generate -object flags for all tests
-   llvm-cov show $(for test in $(ls tests/); do echo -n "-object $test "; done) -instr-profile=coverage.profdata
-   ```
+**Handling mixed library types:**
+- **Shared libraries (.so)**: Use `-object <lib>.so`, only need `llvm_profile_file`
+- **Static libraries**: May require `-object` for each test binary plus `ignore-filename-regex` and `test_names`
+- **Header-only libraries**: Require all three fields (test binaries, ignore patterns, profraw path)
 
-**Proposed standardization:**
-- Standardize test naming: `test_<test_name>` across all components
-- Allows declarative generation of `-object` flags at the top level
-- With standard naming, may be able to reuse test_categories.yaml structure without a separate coverage variant
-- **This remains an open discussion topic** - the tradeoffs between explicit listing vs. declarative generation need team input
+The test runner can determine library type and adjust llvm-cov invocation accordingly. With declarative field calculation, the same test_categories_coverage.yaml schema works for all library types.
+
+**Why test_categories_coverage.yaml may be needed:**
+Because `llvm-cov show -object` does not accept wildcard patterns, we must explicitly list each test binary for header-only libraries. However, with standardized `test_*` naming:
+```bash
+# Declaratively generate -object flags for all tests
+llvm-cov show $(for test in $(ls tests/); do echo -n "-object $test "; done) -instr-profile=coverage.profdata
+```
+
+This allows the test runner to discover test binaries automatically rather than requiring explicit per-project configuration.
 
 ## Design Proposal
 
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -334,4 +334,4 @@ Specific node sizing, runtime benchmarks, and storage quotas remain open topics 
 ## Revision History
 
 - 2026-07-28: jorobbin: Initial version
-- 2026-07-30: jorobbin: Clarified downstream independence and llvm-cov wildcard limitations
+- 2026-07-30: jorobbin: Clarified downstream independence, llvm-cov wildcard limitations, added therock_configure_coverage.py design, test schema details, CI workflow integration, codecov.io integration, resource allocation, and failure handling
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -266,6 +266,18 @@ Modify existing test workflows:
 - Coverage data: `coverage.profdata`
 - Coverage report: `coverage.info` (lcov format)
 
+### Coverage Report Upload
+
+**Codecov.io integration:**
+- Already configured for Math CI at https://app.codecov.io/gh/ROCm/rocm-libraries
+- Upload token stored as GitHub repository secret
+- Service handles PR commenting and coverage aggregation automatically
+- Configuration via `codecov.yaml` at repository root
+- Upload step in CI workflow sends `coverage.info` to codecov.io
+
+**Provider flexibility:**
+The specific coverage reporting service (codecov.io vs alternatives) and its detailed configuration are outside the scope of this RFC. The design supports any service that accepts lcov format reports.
+
 ### Nightly Runs
 
 **Recommendation:** Achieve parity with Math CI:
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -32,11 +32,16 @@ TheRock CI needs to support code coverage while:
 
 ### Coverage Report Scope
 Coverage reports for a given project should only instrument that project's code, not upstream or downstream dependencies. For example:
-- **rocFFT coverage** should instrument only rocFFT, not hipFFT
+- **rocFFT coverage** should instrument only rocFFT, not hipFFT (downstream)
 - **rocBLAS coverage** should not instrument hipBLASLt or downstream consumers
 - **hipBLASLt coverage** should not instrument upstream dependencies like rocBLAS, rocRAND, or rocPRIM
 
-Downstream testing is already functionally covered by TheRock CI. Coverage for hipBLASLt does not affect coverage for rocBLAS.
+**Why exclude downstream components?** Coverage in one component does not affect coverage in downstream components - they are independent concerns. For example:
+- Generating a coverage report for rocBLAS has no effect on coverage for hipBLASLt
+- Generating a coverage report for rocFFT has no effect on coverage for hipFFT
+- Each component's coverage is measured independently against its own codebase
+
+Downstream functional testing is already covered by TheRock CI's existing test infrastructure, which validates that downstream components work correctly with their dependencies. Coverage testing focuses solely on exercising the code paths within the component being measured.
 
 ### Instrumentation Behavior and Profraw Contamination
 Instrumented binaries emit `.profraw` files during execution. If we instrument code upstream to the project being measured, those upstream libraries will emit profraw files whenever the downstream project calls their APIs.
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -168,13 +168,29 @@ Code coverage requires its own build/test pipeline separate from pre-checkin:
   - Negligible gain from multi-arch coverage
 
 #### Build Phase
-1. Determine coverage flag from PR changes:
-   ```bash
-   python build_tools/therock_configure_coverage.py
-   ```
-   Output: `-D<PROJECT_NAME>_ENABLE_COVERAGE=ON`
+1. Determine coverage-enabled projects from PR changes via `therock_configure_coverage.py`:
+   - Leverages existing TheRock CI infrastructure (`therock_matrix.py`) to detect changed projects
+   - Maintains a whitelist of coverage-enabled projects in `COVERAGE_PROJECT_METADATA`
+   - For each coverage-enabled project that changed:
+     - Outputs project-specific CMake options (e.g., `-DTHEROCK_ENABLE_RAND=ON -DTHEROCK_ENABLE_ALL=OFF`)
+     - Pins build to single project only (not merged mega-group)
+     - References project's `test_categories_coverage.yaml` file
+   - Creates independent coverage job per project (even if multiple projects changed)
+   - **Multi-component PRs**: Each coverage-enabled project gets its own isolated build → test → report pipeline
 
-2. Run coverage build:
+   Example metadata output for hipRAND:
+   ```json
+   {
+     "project_name": "HIPRAND",
+     "cmake_target": "hipRAND",
+     "build_dir": "TheRock/build-coverage/ml-libs/hipRAND/build",
+     "cmake_options": "-DTHEROCK_ENABLE_RAND=ON -DTHEROCK_ENABLE_ALL=OFF",
+     "coverage_config": "projects/hiprand/test_categories_coverage.yaml",
+     "projects_to_test": "hiprand"
+   }
+   ```
+
+2. Run coverage build for each project:
    ```bash
    cmake -B build -GNinja \
      -DTHEROCK_AMDGPU_FAMILIES=gfx942 \
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -115,11 +115,22 @@ Some components are header-only static libraries (e.g., rocPRIM, hipCUB) and do 
 - Must instrument the test binaries themselves (headers are compiled into the test binary)
 - Must use `--ignore-filename-regex` to exclude test code from the report
 - Requires explicit reference to test binaries rather than library objects
+- **The `-object` flag does not accept wildcards** - cannot use `-object test_*`, must expand to `-object test_1 -object test_2 ... -object test_n`
+
+**Why test_categories_coverage.yaml may be needed:**
+Because `llvm-cov show -object` does not accept wildcard patterns, we must explicitly list each test binary. This requires either:
+1. **Separate test_categories_coverage.yaml** that explicitly lists all test binaries for coverage processing
+2. **Declarative approach with standard naming** - if tests follow `test_*` convention:
+   ```bash
+   # Generate -object flags for all tests
+   llvm-cov show $(for test in $(ls tests/); do echo -n "-object $test "; done) -instr-profile=coverage.profdata
+   ```
 
 **Proposed standardization:**
-- Standardize test naming: `test_<test_name>`
-- Allows declarative control of `--ignore-filename-regex` at the top level
-- Reuses existing test_categories.yaml structure without modifications
+- Standardize test naming: `test_<test_name>` across all components
+- Allows declarative generation of `-object` flags at the top level
+- With standard naming, may be able to reuse test_categories.yaml structure without a separate coverage variant
+- **This remains an open discussion topic** - the tradeoffs between explicit listing vs. declarative generation need team input
 
 ## Design Proposal
 
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -297,6 +297,21 @@ The specific coverage reporting service (codecov.io vs alternatives) and its det
 
 Specific node sizing, runtime benchmarks, and storage quotas remain open topics for discussion and will be refined based on initial deployment experience.
 
+### Failure Handling
+
+**Coverage test failures:**
+- If a test fails during coverage runs, the pipeline fails
+- Coverage failures block PR merges (must be fixed before merge)
+- Treated the same as any other test failure in the CI pipeline
+
+**Missing profraw files:**
+- If profraw files are missing entirely, the coverage job fails
+- Indicates fundamental instrumentation or test execution problem
+
+**Flaky tests:**
+- Can be disabled at component owner's discretion
+- Same policy as regular pre-checkin test handling
+
 ### Nightly Runs
 
 **Recommendation:** Achieve parity with Math CI:
--- a/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
+++ b/docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
@@ -1,0 +1,283 @@
+---
+author: John Robbins (jorobbin)
+created: 2026-07-28
+modified: 2026-07-28
+status: draft
+discussion: TBD
+---
+
+# Code Coverage Infrastructure for TheRock
+
+This RFC proposes integrating code coverage builds and reporting into TheRock CI, maintaining per-project isolation while leveraging TheRock's super-build infrastructure.
+
+## Motivation
+
+Currently, Math CI builds projects individually for code coverage:
+```bash
+mkdir projects/rocfft/build && cd projects/rocfft/build
+cmake -DBUILD_CODE_COVERAGE ..
+make -j
+LLVM_PROFILE_FILE="./coverage-report/%m-%p-rocfft.profraw" make -j coverage
+```
+
+This workflow relies on prepackaged dependencies and component-specific coverage targets that handle test execution, profraw merging, and report generation.
+
+TheRock CI needs to support code coverage while:
+1. Maintaining per-project isolation for instrumentation
+2. Accommodating the build/test node split architecture
+3. Avoiding performance impacts on pre-checkin and performance testing builds
+4. Supporting nightly coverage runs for individual components
+
+## Background: Why Per-Project Isolation is Required
+
+### Coverage Report Scope
+Coverage reports for a given project should only instrument that project's code, not upstream or downstream dependencies. For example:
+- **rocFFT coverage** should instrument only rocFFT, not hipFFT
+- **rocBLAS coverage** should not instrument hipBLASLt or downstream consumers
+- **hipBLASLt coverage** should not instrument upstream dependencies like rocBLAS, rocRAND, or rocPRIM
+
+Downstream testing is already functionally covered by TheRock CI. Coverage for hipBLASLt does not affect coverage for rocBLAS.
+
+### Instrumentation Behavior and Profraw Contamination
+Instrumented binaries emit `.profraw` files during execution. If we instrument code upstream to the project being measured, those upstream libraries will emit profraw files whenever the downstream project calls their APIs.
+
+Example: If both hipBLASLt and rocBLAS are instrumented, but we only care about hipBLASLt coverage:
+- When hipBLASLt calls rocBLAS APIs, rocBLAS emits profraw files
+- These upstream profraw files contaminate the coverage report
+- We must either:
+  - Explicitly ignore them with `--ignore-filename-regex`, or
+  - Accept non-deterministic fluctuations in the coverage denominator that change with upstream modifications
+
+This contamination propagates up the entire dependency chain to rocRAND and rocPRIM, which are high in the dependency tree and consumed by many downstream projects.
+
+### Performance Impact
+Coverage builds have significant performance degradation:
+- **Example (rocPRIM)**: Normal pre-checkin tests take ~20 minutes; coverage runs take ~3 hours
+- Most projects have less extreme degradation, but the impact is still substantial
+- Coverage builds add debug instrumentation and disable compiler optimizations
+
+### Separation from Pre-Checkin Testing
+Coverage builds must be separate from pre-checkin builds:
+- **Pre-checkin goal**: Test binaries as close to production as possible
+- **Coverage builds**: Add debug lines, disable optimizations, enable instrumentation
+- Using coverage builds for pre-checkin creates an inconsistent testing environment
+- Performance testing (planned for TheRock CI) cannot use coverage builds
+
+## Current Math CI Workflow
+
+Projects in rocm-libraries already support coverage builds via `-DBUILD_CODE_COVERAGE`:
+
+```bash
+mkdir projects/rocfft/build && cd projects/rocfft/build
+cmake -DBUILD_CODE_COVERAGE ..
+make -j
+LLVM_PROFILE_FILE="./coverage-report/%m-%p-rocfft.profraw" make -j coverage
+```
+
+The `coverage` target:
+1. Runs the test suite
+2. Drops `*.profraw` files to `LLVM_PROFILE_FILE` location (with `%m-%p` interpolation for unique identifiers)
+3. Executes report generation commands:
+   ```bash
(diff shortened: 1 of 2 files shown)


docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+- **rocFFT coverage** should instrument only rocFFT, not hipFFT (downstream)
+- **rocBLAS coverage** should not instrument hipBLASLt or downstream consumers
+- **hipBLASLt coverage** should not instrument upstream dependencies like rocBLAS, rocRAND, or rocPRIM
+
+**Why exclude downstream components?** Coverage in one component does not affect coverage in downstream components - they are independent concerns. For example:
+- Generating a coverage report for rocBLAS has no effect on coverage for hipBLASLt
+- Generating a coverage report for rocFFT has no effect on coverage for hipFFT
+- Each component's coverage is measured independently against its own codebase
+
+Downstream functional testing is already covered by TheRock CI's existing test infrastructure, which validates that downstream components work correctly with their dependencies. Coverage testing focuses solely on exercising the code paths within the component being measured.
```
ScottTodd: I'm interested in considering alternatives here, particularly since rebuilding individual subprojects separately will add overhead and job graph complexity to CI workflows. We've had some offline discussions about this, but for the sake of argument and for others to have the chance to join the discussion:

* What if we instrumented all of rocm-libraries together as one unit? Or at least artifact groups or build stages like "blas" or "math-libs"?
    * We could generate a single code coverage number for all subprojects. Individual changes would have less of an impact on the overall number in that case though.
    * This would allow (for better or worse) code like https://github.com/ROCm/rocm-libraries/blob/develop/projects/rocsparse/library/src/primitives/rocsparse_find_max.cpp to contribute to coverage for rocprim/rocprim.hpp
* Can we _build_ all projects together with instrumentation but then only choose to _test_ a subset, or can coverage reports not be filtered that way?
    * I guess this is answered below
        > If we instrument code upstream to the project being measured, those upstream libraries will emit profraw files whenever the downstream project calls their APIs.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+- **Example (rocPRIM)**: Normal pre-checkin tests take ~20 minutes; coverage runs take ~3 hours
+- Most projects have less extreme degradation, but the impact is still substantial
+- Coverage builds add debug instrumentation and disable compiler optimizations
+
+### Separation from Pre-Checkin Testing
+Coverage builds must be separate from pre-checkin builds:
+- **Pre-checkin goal**: Test binaries as close to production as possible
+- **Coverage builds**: Add debug lines, disable optimizations, enable instrumentation
+- Using coverage builds for pre-checkin creates an inconsistent testing environment
+- Performance testing (planned for TheRock CI) cannot use coverage builds
```
ScottTodd: This is suggesting that we could treat "coverage" as a high level build variant, similar to "asan". That probably warrants
1. A CMake preset in https://github.com/ROCm/TheRock/blob/main/CMakePresets.json, see also https://github.com/ROCm/TheRock/issues/5453
2. A dedicated github _workflow_ that will use a unique directory so artifacts from other build variants don't overlap, like how https://github.com/ROCm/TheRock/actions/runs/30503084933?pr=6982 (release) and https://github.com/ROCm/TheRock/actions/runs/30503085021?pr=6982 (asan) on the same PR uploaded to https://therock-ci-artifacts.s3.amazonaws.com/30503084933-linux/index.html and https://therock-ci-artifacts.s3.amazonaws.com/30503085021-linux/index.html
    * Needing to build each subproject separately with instrumentation complicates this though.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+- `-DHIPBLASLT_ENABLE_COVERAGE`
+- etc.
+
+Add TheRock passthrough:
+```cmake
+# In TheRock super-build
+cmake -B build -GNinja \
+  -DTHEROCK_AMDGPU_FAMILIES=gfx942 \
+  -DROCFFT_ENABLE_COVERAGE=ON
+```
```
ScottTodd: To set coverage enabled/disabled per project, we discussed some other options on https://github.com/ROCm/TheRock/pull/3761, like the existing `{project}_CMAKE_ARGS` which can be used like `-DhipDNN_CMAKE_ARGS="-DHIPDNN_ENABLE_COVERAGE=ON"`.

A few other ideas:
* Similar to `-DTHEROCK_DEV_PROJECTS=...`, add  `-DTHEROCK_COVERAGE_PROJECTS=...`
* Match subproject casing with `-DrocFFT_ENABLE_COVERAGE`

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Required Compiler Flags (for device-side coverage)
+Recent amd-llvm changes enable device-side coverage, requiring:
+1. `-DCOMPILER_RT_BUILD_PROFILE_ROCM=ON`
+   - Ensures device profiler is built into instrumented device code
+2. `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF`
+   - Build only one artifact for testing (no multi-GPU architecture splits)
```
ScottTodd: `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF` is going to be tough... the flag is already default `ON` in most places and I'd really like to _remove the flag_, not add new features that depend on it being disabled.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+
+### TheRock CI Workflow
+
+#### Separate Coverage Pipeline
+Code coverage requires its own build/test pipeline separate from pre-checkin:
+- **Reason**: Custom instrumented build with different compiler flags
+- **Architecture**: Single architecture (typically gfx942)
+  - Minimal variation between architectures
+  - Negligible gain from multi-arch coverage
```
ScottTodd: Single architecture isn't that compelling a reason to build a new pipeline. We have plenty of other workflows that only build for a single architecture but still use the "multi-arch" pipelines that build in multiple stages and I'd like to use as much of the same workflows and scripts as we can for each build variant (release, asan, now coverage)

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Build Phase
+1. Determine coverage-enabled projects from PR changes via `therock_configure_coverage.py`:
+   - Leverages existing TheRock CI infrastructure (`therock_matrix.py`) to detect changed projects
+   - Maintains a whitelist of coverage-enabled projects in `COVERAGE_PROJECT_METADATA`
+   - For each coverage-enabled project that changed:
+     - Outputs project-specific CMake options (e.g., `-DTHEROCK_ENABLE_RAND=ON -DTHEROCK_ENABLE_ALL=OFF`)
+     - Pins build to single project only (not merged mega-group)
```
ScottTodd: I believe the filtering details like `therock_matrix.py` and project-specific CMake options for enabling/disabling components are changing with the migration to multi-arch CI. Maybe @geomin12 could confirm?

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+     -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF \
+     -D<PROJECT_NAME>_ENABLE_COVERAGE=ON
+   ninja -C build
+   ```
+
+3. Upload artifacts:
+   ```bash
+   python build_tools/artifact_manager.py
+   python build_tools/post_build_upload.py
+   ```
```
ScottTodd: `post_build_upload.py` has been replaced by https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/post_stage_upload.py for multi-arch CI. The only uses of the old script are in repositories other than TheRock

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+   - Merges all profraw files from all shards into single profdata
+   - Exports coverage in lcov format for codecov.io upload
+   - Uses component-specific library object file
+
+**Artifact naming conventions:**
+- Coverage builds use same artifact naming as regular builds
+- Artifacts downloaded via existing `fetch_artifact_args` mechanism
+- Profraw files: `${TEST_COMPONENT}-shard${SHARD_INDEX}-%p-%m.profraw`
+- Coverage data: `coverage.profdata`
+- Coverage report: `coverage.info` (lcov format)
```
ScottTodd: > Coverage builds use same artifact naming as regular builds

Artifacts built using different settings either need unique names (e.g. `-asan` suffix) or we need to be careful about where they are uploaded. We don't want developers, users, or CI systems downloading artifacts of a different type than they are expecting.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Coverage Report Upload
+
+**Codecov.io integration:**
+- Already configured for Math CI at https://app.codecov.io/gh/ROCm/rocm-libraries
+- Upload token stored as GitHub repository secret
+- Service handles PR commenting and coverage aggregation automatically
```
ScottTodd: If the upload token is stored in a github secret, it won't be accessible for PRs from forks. Is that okay? We can also load credentials onto our test runners (or some other self-hosted runner that bridges between the test runners and the codecov service)

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+
+**Architecture scope:**
+- Coverage runs on single architecture only (typically gfx942)
+- Multi-architecture testing provides negligible gain in coverage reporting
+- No downstream testing - only the changed project itself
+
+**Node allocation:**
+- One build node per coverage-enabled project that changed
+- One test node per coverage-enabled project that changed
+- Resource sizing should align with single-project testing requirements
```
ScottTodd: RE: the prior comments about multi-arch CI with multiple build stages: in general with a build system of this complexity, "one build node" is just not a reasonable starting point - everything is sharded across multiple machines (including tests).

If we use the multi-arch CI architecture with build stages separated across jobs, we can also use prebuilt artifacts for unmodified projects (we can also do that without separating stages across jobs, but the current CI workflows already support that)

ScottTodd (commented):
My main reservations here are about what the CI job graph and uploaded artifacts will look like if we need to rebuild each component project with different flags. Some metrics from experiments and graphs/diagrams of the proposed architecture could help in discussing that.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: Do you have a way of collating coverage from multiple devices?

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: `Negligible gain from multi-arch coverage` - if you are going to look at code coverage to inform developers and users whether code is tested, then architecture-specific code should be part of this coverage.  The assumption of "Minimal variation between architectures" is dangerous here.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: And multi-gpu testing is important. If code coverage is important, then knowing code-coverage for multi-gpu libraries is important.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+# In TheRock super-build
+cmake -B build -GNinja \
+  -DTHEROCK_AMDGPU_FAMILIES=gfx942 \
+  -DROCFFT_ENABLE_COVERAGE=ON
+```
+
+#### Required Compiler Flags (for device-side coverage)
+Recent amd-llvm changes enable device-side coverage, requiring:
+1. `-DCOMPILER_RT_BUILD_PROFILE_ROCM=ON`
+   - Ensures device profiler is built into instrumented device code
```
malcolmroberts: What is the plan for projects which generate device code?

malcolmroberts (changes requested):
There are some dangerous assumptions in this document.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: > `Negligible gain from multi-arch coverage` - if you are going to look at code coverage to inform developers and users whether code is tested, then architecture-specific code should be part of this coverage. The assumption of "Minimal variation between architectures" is dangerous here.

@malcolmroberts I agree that that statement is dangerous and misleading, so I can refine it.  However, there's a tradeoff here.  There is a way of collating coverage across multiple devices, but there's a problem with nightly runs if we do both multiple architectures AND separate runs for each component.

If we don't separate each component into their own run, then we end up doing an instrumented super build, which is likely to have way too much overhead to run feasibly.  A good example is rocPRIM's instrumented test run takes 3 hours to run, whereas the non-instrumented test run only takes 22 minutes.  rocPRIM, as I understand it, is upstream to everything, so it's slowness will compound down the dependency chain to the actual component we care about.

So it's implied that we will need to separate each component into its own nightly run, and then if we also have to do a per architecture per component, on every PR and in nightly, my assumption is that it can get out of hand rather quickly.

There's a few options, one is that we control which architecture it gets tested on automatically, based on the architecture that's relevant to the code changes and have a predominant architecture that it always runs on, or we accept that code coverage only applies to code that is shared across all architectures.

Seems as though the first option is probably best.  I can draft up that design and include it in the document for your review.  Unless you think there's another option I haven't considered?

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: > Can we build all projects together with instrumentation but then only choose to test a subset, or can coverage reports not be filtered that way?

Not only will the upstream components output profraw files, but more importantly, those upstream components will be instrumented themselves, and their _slowness_ will compound down the chain of dependencies.  Using rocPRIM as an example, which is upstream to pretty much everything, it's instrumented test runs take 3 hours to run, whereas its non-instrumented takes 22 minutes.

> What if we instrumented all of rocm-libraries together as one unit? Or at least artifact groups or build stages like "blas" or "math-libs"?

I read this in two ways: 
1. We instrument all of rocm-libraries and then test the component we're interested in consuming the most recent stable, instrumented builds from nightly runs for everything else.  Then we're running the test with all of it's upstream components instrumented, which doesn't work due to the aforementioned slowness issue.
2. We instrument all of rocm-libraries in a nightly job and then save each artifact as the most recent, stable, instrumented build.  These pre-existing binaries should never be consumed by a PR for a specific component, since we will need to build that component fresh and we don't want to include any other instrumented library.

What the latter case could be useful for is our nightly runs, where all of rocm-libraries is built as one, and then the test runners consume only the component they are testing from the instrumented super build, and everything else from the non-instrumented super build.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: If there's a need to do a full instrumented super build, then a preset is necessary.  Otherwise, I'm not sure if it would be valuable.  Can you give me an example of what you're envisioning the preset to look like for coverage?

So far, if I want to instrument a specific component, I just do:
```bash
cmake -D<COMPONENT_NAME>_ENABLE_COVERAGE=ON ..
#for rocPRIM
cmake -DROCPRIM_ENABLE_COVERAGE=ON ..
```

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: > Match subproject casing with -DrocFFT_ENABLE_COVERAGE

I think this is the cleanest approach, but will try a few of them before updating the doc

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: I can do some more testing with `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=ON` instead.

There is a way to collate coverage that was ran on different devices.  Codecov.io, or whatever coverage reporting tool we end up using, should be able to handle that simply by us uploading the report with a specific tag.  Otherwise, we can do it ourselves by merging the profraw files dropped by several runs in one place and generating the report that way.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: I'll update this to make use of post_stage_upload.py instead

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: The secrets are only for codecov.io, or other reporting mechanism, upload.  We don't want users publishing reports, we only want our CI to be doing that.

I do see us using secrets in our workflows already.  Do those secrets not get populated when a developer makes a PR?

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: There have been some recent changes to enable device side coverage using `llvm-cov`.  This line, and the `--offload-new-driver` added to the build targets, are supposed to enable device side coverage, if exists.  So seemingly no changes are needed to enable device coverage

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: Why is post_build_upload.py still present in the repo?  If it's still needed, these two scripts should be merged and controlled with conditional logic, rather than just create a brand new, disjoint script, that intuitively does the same thing.

ScottTodd: There are still several scripts and workflows that should be deleted or migrated as part of the switch from per-family CI to multi-arch CI:
* https://github.com/ROCm/TheRock/issues/3340

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+   ```
+4. Uploads to codecov.io, which:
+   - Comments on PRs showing coverage deltas
+   - Can fail CI if coverage drops below 80%
+
+### Nightly Runs
+Math CI performs nightly coverage runs on `develop` for each component individually. This approach avoids resource contention and provides per-component coverage trends.
+
+## Challenges in TheRock
+
```
malcolmroberts: therock has some very essential infrastructure in it; what is the plan for code coverage for therock?

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: alternatively, you could have a list of projects `-DENABLE_COVERAGE="rocFFT,hipBLASls"` which could also be `none` or `all`.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
ScottTodd: For the python scripts in TheRock, see:
* https://github.com/ROCm/TheRock/blob/main/TESTING.md#python-scripts-and-tools---limitations-and-known-gaps
* https://github.com/ROCm/TheRock/blob/35ed46c376f22431411a497dac08db894060a522/.github/workflows/unit_tests.yml#L59-L67

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+   - Ensures device profiler is built into instrumented device code
+2. `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF`
+   - Build only one artifact for testing (no multi-GPU architecture splits)
+
+### TheRock CI Workflow
+
+#### Separate Coverage Pipeline
+Code coverage requires its own build/test pipeline separate from pre-checkin:
+- **Reason**: Custom instrumented build with different compiler flags
+- **Architecture**: Single architecture (typically gfx942)
```
malcolmroberts: I don't see how we can stick to just one architecture.  We may have minimal code differences between architectures now, but that's likely to change.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: Ok, we discussed a possible solution for this which is to have a base code coverage report (initial or monthly) which is updated with per-architecture or single/multi-gpu code for PRs which exercise different code paths in these separate hardware configurations.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,337 @@
...
+- Storage: Coverage artifacts similar size to regular artifacts; profraw files are temporary and merged/deleted after report generation
+
+Specific node sizing, runtime benchmarks, and storage quotas remain open topics for discussion and will be refined based on initial deployment experience.
+
+### Failure Handling
+
+**Coverage test failures:**
+- If a test fails during coverage runs, the pipeline fails
+- Coverage failures block PR merges (must be fixed before merge)
+- Treated the same as any other test failure in the CI pipeline
```
malcolmroberts: This is at risk of Goodhart's law.

For example, one should handle hipMalloc failures, but the code to handle these cases are untestable unless we mock the hip runtime, allowing us to inject failures.  Developers are therefore incentivized to minimize or remove code to handle these failures in order to meet coverage % targets.

malcolmroberts (changes requested):
Some discussion left on this.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: That can work, but we still need to pass the <project_name__ENABLE_COVERAGE flag to the individual projects.  We need to move away from `-DBUILD_CODE_COVERAGE` flags that are used, so that we don't accidentally enable coverage for other projects, when they all use the same flag.

I'll explain this point more clearly in the docs and outline a few of the options above.

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
ScottTodd: Other infrastructure components (packages, github actions workflows, the cmake build system, etc.) are more difficult to get "coverage" reports/metrics for, but they're also discussed in that TESTING.md

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: Please see: https://github.com/ROCm/TheRock/pull/6967/changes/646d8163f7edddc0af3f70456070edbf5e173e1b for options and recommended approach.

Section _Component-Specific Coverage Flags_ and subsection _CI Workflow Integration Consideration:_

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: See https://github.com/ROCm/TheRock/pull/6967/changes/96fabcb182ba08f3bf5c6821565cafb1225121ec for multi-arch coverage strategy.

Sections _Resource Allocation_ and _Multi-Architecture Coverage Strategy_
As well as _Open Questions_

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: https://github.com/ROCm/TheRock/pull/6967/changes/e380ac501fd14c089a5c98abe15a50d0117ddfd6

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: https://github.com/ROCm/TheRock/pull/6967/changes/f344ebc69043d33528559f2df5e4816f0cec48b4

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: I don't see any problem with using multi-arch and have reflected that in the _Resource Allocation_ section of the RFC.

Your comment also shed light on a different gap in the design: test sharding.  Because tests are sharded across multiple nodes, we can't simply generate the coverage report on the test system, after they run, since we'll only get the coverage for that shard.  Instead, each test node must upload the profraw files- from their run and when all nodes finish testing, a separate node must run to pull the profraw files and merge them before publishing the coverage report.

All of this is captured in https://github.com/ROCm/TheRock/pull/6967/changes/0e0389c84356e295a455960241c9a719bc0e8bf2

See sections 
1. _Test Phase_ 
    - _4. Upload profraw files as artifacts (conditional on coverage_enabled):_ 
    - _5. Coverage aggregation job (separate node that runs after all test shards complete):_
2. _Resource Allocation:_
    - _Node allocation:_

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: I documented a requirement for amd-llvm to have code coverage smoke tests, to avoid future problems when bumping hashes, like what happened when moving from 7.13 to 7.14: https://github.com/ROCm/TheRock/pull/6967/changes/944a690e2012673b8cd272056232a8916f6d0349

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: https://github.com/ROCm/TheRock/pull/6967/changes/96fabcb182ba08f3bf5c6821565cafb1225121ec

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
reboss: I think this commit addresses your concern about mock and multi-gpu coverage: https://github.com/ROCm/TheRock/pull/6967/changes/009d7f307b21503f8e850ee37b4d31bc581ba547

docs/rfcs/RFC0013-Code-Coverage-Infrastructure.md
malcolmroberts: > Other infrastructure components (packages, github actions workflows, the cmake build system, etc.) are more difficult to get "coverage" reports/metrics for, but they're also discussed in that TESTING.md

Coverage is difficult, but this essential functionality which needs to be tested effectively.  I've made PRs to scripts that are still present in therock, but which have been superceded and are now stale - this would have been avoided by a code coverage report.

malcolmroberts (approved):
Thanks for the edits!

I think that this is workable now.  I do feel like we need phased implementation to demonstrate that we have the infrastructure available (eg multi-gpu, multi-arch, mocked upstream dependencies) before things are gating.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,598 @@
...
+  - Explicitly ignore them with `--ignore-filename-regex`, or
+  - Accept non-deterministic fluctuations in the coverage denominator that change with upstream modifications
+
+This contamination propagates up the entire dependency chain to rocRAND and rocPRIM, which are high in the dependency tree and consumed by many downstream projects.
+
+### Performance Impact
+Coverage builds have significant performance degradation:
+- **Example (rocPRIM)**: Normal pre-checkin tests take ~20 minutes; coverage runs take ~3 hours
+- Most projects have less extreme degradation, but the impact is still substantial
+- Coverage builds add debug instrumentation and disable compiler optimizations
```
ScottTodd, line 62: Metrics per component would be useful here. hipFFT for example has ~2 orders of magnitude more tests than is reasonable so we currently run a random selection of 2% of the tests with `"--test_prob" "0.02"` *even on releases*: https://github.com/ROCm/rockrel/actions/runs/32679226580/job/97293147174#step:15:52

At a certain point we need to optimize the component code / project structure, not just stress the limits of the CI system.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,598 @@
...
+     "projects_to_test": "hiprand"
+   }
+   ```
+
+2. Run coverage build for each project:
+   ```bash
+   cmake -B build -GNinja \
+     -DTHEROCK_AMDGPU_FAMILIES=gfx942 \
+     -DCOMPILER_RT_BUILD_PROFILE_ROCM=ON \
+     -DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF \
```
ScottTodd, line 313: As stated before, `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF` is going away. The flag has been default-enabled for many months now and will be retired.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,598 @@
...
+   - Shard 4 only covers tests 76-100 (25% coverage)
+   
+   Aggregation merges profraw from all shards to produce complete 100% coverage report.
+
+**Artifact naming conventions:**
+- Coverage builds require unique artifact names to avoid confusion with regular builds
+- Add `-coverage` suffix to artifact names (similar to `-asan` for ASAN builds)
+- Example: `hiprand-gfx942-coverage.tar.gz` vs `hiprand-gfx942.tar.gz`
+- Prevents developers/CI from accidentally downloading instrumented coverage artifacts when expecting regular builds
+- Artifacts downloaded via `fetch_artifact_args` mechanism with coverage-specific paths
```
ScottTodd: We might have done this at one point, but I don't currently see `-asan` suffixes on artifact names (example: https://github.com/ROCm/TheRock/actions/runs/30936559628 --> https://therock-ci-artifacts.s3.amazonaws.com/30936559628-linux/index.html).

As needed we can use `build_variant_suffix`: https://github.com/ROCm/TheRock/blob/e10e8e23d59cd3ed88fdf57eaa4e9650333c1c86/build_tools/github_actions/amdgpu_family_matrix.py#L156-L160

What we currently rely on is:
1. Each build variant is its own workflow, so it gets a unique directory name in the S3 buckets like `30936559628-linux` for run 30936559628
2. Package building and publishing choosing folders like `tarball-asan` or just `tarball`: https://github.com/ROCm/TheRock/blob/e10e8e23d59cd3ed88fdf57eaa4e9650333c1c86/build_tools/github_actions/publish_rocm_to_release_buckets.py#L106-L110

If these coverage builds will be just for CI and not released, then maybe (1) will be enough. Any scripts that download artifacts would look at workflow runs of "therock ci", not workflow runs of "therock ci coverage" or a similarly named workflow.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,598 @@
...
+2. `-DTHEROCK_FLAG_KPACK_SPLIT_ARTIFACTS=OFF`
+   - Build only one artifact for testing (no multi-GPU architecture splits)
+
+### TheRock CI Workflow
+
+#### Separate Coverage Pipeline
+Code coverage requires its own build/test pipeline separate from pre-checkin:
+- **Reason**: Custom instrumented build with different compiler flags
+- **Architecture strategy**: Phased approach to multi-architecture coverage
+  - **Phase 1 (initial)**: Single default architecture (gfx942 or gfx950) for all coverage runs - achieves parity with Math CI
```
ScottTodd, line 277: Lets start with an on-demand and scheduled workflow and then move towards coverage on pull requests

ScottTodd (approved):
Generally LGTM, but I'd like to see more focus in the implementation/rollout plan on the standard ordering from low frequency to high frequency workflow runs:
1. on-demand (workflow_dispatch)
2. nightly (schedule)
3. push
4. pull_request

Until we have concrete data on how long these jobs take to run it's hard to commit to running them on every pull request.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
reboss: The plan now is to have the regular nightly run kick off the code coverage run downstream, and each will have their own run-id to pull artifacts from.  So yes, there's no need for the -coverage suffix, since the distinction can be made with separate run-ids for instrumented versus non-instrumented artifacts.

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
reboss: Addressed with https://github.com/ROCm/TheRock/pull/6967/commits/5b18c0acef12a6df868399b156058d592393966b

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
```diff
@@ -0,0 +1,1066 @@
...
+- **`llvm_profile_file`**: Hardcoded to standard path (e.g., `./coverage-report/%m-%p.profraw`)
+- **`ignore-filename-regex`**: Set to exclude anything outside the src directory (e.g., patterns matching test files)
+
+This reduces per-project configuration overhead and ensures consistency across components.
+
+**Error handling:** If profraw files are missing entirely, the coverage job fails. This indicates a fundamental problem with instrumentation or test execution.
+
+### Problematic: Header-Only Static Libraries
+
+Some components are header-only static libraries (e.g., rocPRIM, hipCUB) and do not produce a simple `.so` file for `llvm-cov show -object <lib>.so`.
```
saadrahim, line 140:
```suggestion
Some components are header-only libraries (e.g., rocPRIM, hipCUB) and do not produce a simple `.so` file for `llvm-cov show -object <lib>.so`.
```
not sure if static is needed here

reboss: I think the use of static is appropriate.  It suggests that they are .a static headers, and not .so dynamically linked headers

reboss:
```suggestion
Some components are header-only static libraries (e.g., rocPRIM, hipCUB) that produce `.a` files, and do not produce a simple `.so` file for `llvm-cov show -object <lib>.so`.
```

Would that be more clear?

saadrahim (approved):

docs/rfcs/RFC0014-Code-Coverage-Infrastructure.md
saadrahim: I don't think there is even .a files? Maybe just ask @stanleytsang-amd

PR #6974, opened 2026-07-29
feat: Verify installed files are owned by root in native package install test

dileepr1:
## Summary
- Add a post-install ownership check to `run_basic_verification()` in `native_linux_package_install_test.py` so both `sanity` and `full` test modes confirm every file under the install prefix is owned by `root:root`.
- Uses `find` (C-level traversal) instead of a Python walk with per-file `lstat` to stay fast on large ROCm install trees; flags and prints offending paths (first 10) and fails basic verification if any are found.
- Degrades gracefully to a `[WARN]` (non-fatal) if `find` is unavailable or the check times out (`OWNERSHIP_TIMEOUT_SEC = 120`).

GitHub issue URL : https://github.com/ROCm/ROCm/issues/6496

## Test plan
- [x] Run against a real repo install (e.g. `--test-type sanity --os-profile ubuntu2404 ...`) inside a matching container and confirm the new `Verifying installed files are owned by root...` step reports `[PASS]`.
- [x] Confirm that a deliberately chowned file under the install prefix causes basic verification to `[FAIL]`.
- [x] Confirm `--test-type full` still runs steps 1-3 and the ownership check participates in step 2's pass/fail.
- [x] `pre-commit run --files build_tools/packaging/linux/native_linux_package_install_test.py` passes (verified locally).

## Notes
- Both uid **and** gid must be 0. If some ROCm files are intentionally shipped with a non-root group (device/render bits), an allowlist may be needed to avoid false failures.

### DIFF
--- a/build_tools/packaging/linux/native_linux_package_install_test.py
+++ b/build_tools/packaging/linux/native_linux_package_install_test.py
@@ -187,7 +187,6 @@ ZYPP_REFRESH_TIMEOUT_SEC = 120
 DNF_CLEAN_TIMEOUT_SEC = 60
 INSTALL_TIMEOUT_SEC = 1800  # 30 minutes
 ROCMINFO_TIMEOUT_SEC = 30
-FILE_SECURITY_TIMEOUT_SEC = 120
 # rdhc.py ``--all`` runs the full ROCm deployment health check suite; 30s was too
 # short in container CI (timeouts under load). Optional cluster checks are skipped
 # separately via ``--skip-optional-cluster-checks`` in ``test_rdhc``.
@@ -995,12 +994,11 @@ gpgcheck=0
 
         - Ownership: its owner uid or group gid is not 0 (not ``root:root``);
           a non-root-owned path can be tampered with by that owner.
-        - Writability: it is writable by group or other (mode bits ``0o022``),
-          which would let an unprivileged user drop or replace a binary in a
+        - Writability: it is writable by group or other (mode bits ``0o022``).
+          This would let an unprivileged user drop or replace a binary in a
           directory on ``PATH`` and have another user (or root) execute it.
-          Group/other-writable *sticky* directories (mode bit ``0o1000``,
-          e.g. ``/tmp``-style dirs) are allowed, since the sticky bit prevents
-          users from tampering with files they do not own.
+          ROCm ships no group/other-writable paths (including sticky
+          directories), so any such path is flagged.
         - setuid/setgid: it carries mode bits ``0o6000``, a direct privilege
           escalation surface.
 
@@ -1010,7 +1008,8 @@ gpgcheck=0
         would produce false positives. The install prefix itself is commonly a
         symlink (e.g. ``/opt/rocm/core`` -> ``/opt/rocm/core-X.Y``), so
         ``find -H`` follows that top-level link to scan the real tree while not
-        following links found *inside* the tree.
+        following links found *inside* the tree. ``-xdev`` keeps the scan on the
+        prefix's own filesystem so bind mounts inside the tree are not crossed.
 
         Uses ``find`` (C-level traversal) for speed on large install trees and
         falls back to a pure-Python ``os.walk`` scan if ``find`` is unavailable
@@ -1029,8 +1028,11 @@ gpgcheck=0
                     # follow the prefix if it is a symlink, but not links inside
                     "-H",
                     install_prefix,
+                    # do not cross into other filesystems mounted under prefix
+                    "-xdev",
                     "(",
                     # not owned by root:root
+                    "(",
                     "!",
                     "-uid",
                     "0",
@@ -1038,6 +1040,7 @@ gpgcheck=0
                     "!",
                     "-gid",
                     "0",
+                    ")",
                     "-o",
                     # insecure permissions, but skip symlinks whose own mode
                     # bits are meaningless (the target is checked on its own)
@@ -1046,14 +1049,9 @@ gpgcheck=0
                     "-type",
                     "l",
                     "(",
-                    # group/other-writable, excluding sticky-bit dirs
-                    "(",
+                    # group/other-writable
                     "-perm",
                     "/022",
-                    "!",
-                    "-perm",
-                    "-1000",
-                    ")",
                     "-o",
                     # setuid/setgid
                     "-perm",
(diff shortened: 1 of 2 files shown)


build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -186,6 +187,7 @@ def _env(key: str, default: str) -> str:
 DNF_CLEAN_TIMEOUT_SEC = 60
 INSTALL_TIMEOUT_SEC = 1800  # 30 minutes
 ROCMINFO_TIMEOUT_SEC = 30
+FILE_SECURITY_TIMEOUT_SEC = 120
```
nunnikri: Style Guide: timeout on subprocess violates "No timeouts on basic binutils"

The Python style guide explicitly discourages timeouts on subprocess calls. find on a large install tree could be slow but that should be addressed at the CI job level, not with an in-process timeout.

# Current — violates style guide
subprocess.run([...], timeout=FILE_SECURITY_TIMEOUT_SEC)

# Preferred — remove timeout; let CI job-level timeout handle runaway processes
subprocess.run([...])

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+        exempt from the permission rules since their mode bits are meaningless
+        on Linux. Slower than ``find`` on large trees but portable.
+
+        Returns:
+        True if no offending path is found, False if any offending path is found.
+        """
+        bad: list[str] = []
+        for root, dirs, files in os.walk(install_path):
+            for name in dirs + files:
+                entry = os.path.join(root, name)
```
nunnikri: As per coding guideline use pathlib instead of os.path
# Current — violates "Use pathlib for filesystem paths"
entry = os.path.join(root, name)

# Preferred
entry = Path(root) / name

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+        so the check is not silently skipped.
+
+        Returns:
+        True if no offending path is found (or the check could not be run),
+        False if any offending path is found.
+        """
+        print("\nVerifying installed files are owned by root with safe permissions...")
+        install_prefix = str(Path(self.install_prefix))
+        try:
+            result = subprocess.run(
```
nunnikri, line 1078: subprocess.run return code not checked

find can exit non-zero (e.g. permission denied on a subtree) while still printing partial output. The code checks only stdout content, not result.returncode, so a partial scan passes silently.
# Add after subprocess.run():
if result.returncode != 0:
    print(f" [WARN] find exited {result.returncode}: {result.stderr[:200]}")

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+                    "-o",
+                    # setuid/setgid
+                    "-perm",
+                    "/6000",
+                    ")",
+                    ")",
+                    ")",
+                    "-print",
+                ],
+                stdout=subprocess.PIPE,
```
nunnikri, line 1123: Improvement suggestion : not mandatory to merge PR.
stderr=subprocess.PIPE captures stderr but it is never read. find reports "Permission denied" entries on stderr for inaccessible directories. Log it when non-empty.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+        print("\nVerifying installed files are owned by root with safe permissions...")
+        install_prefix = str(Path(self.install_prefix))
+        try:
+            result = subprocess.run(
+                [
+                    "find",
+                    # follow the prefix if it is a symlink, but not links inside
+                    "-H",
+                    install_prefix,
+                    "(",
```
nunnikri, line 1086: Ownership check should use explicit grouping

! -uid 0 -o ! -gid 0 -o (...) is correct but relies on implicit POSIX precedence. Add explicit parentheses around the ownership clause for clarity:

( ! -uid 0 -o ! -gid 0 ) -o ( ! -type l ( ... ) )

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+        Returns:
+        True if no offending path is found (or the check could not be run),
+        False if any offending path is found.
+        """
+        print("\nVerifying installed files are owned by root with safe permissions...")
+        install_prefix = str(Path(self.install_prefix))
+        try:
+            result = subprocess.run(
+                [
+                    "find",
```
nunnikri, line 1080: Consider -xdev to avoid crossing filesystem boundaries

If the install tree has bind mounts or other filesystems mounted inside it, find will traverse them. Adding -xdev limits the scan to the same filesystem as the prefix.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -967,11 +974,170 @@ def run_basic_verification(self) -> bool:
...
+                    "0",
+                    "-o",
+                    # insecure permissions, but skip symlinks whose own mode
+                    # bits are meaningless (the target is checked on its own)
+                    "(",
+                    "!",
+                    "-type",
+                    "l",
+                    "(",
+                    # group/other-writable, excluding sticky-bit dirs
```
nunnikri: We dont expect any directory with sticky bit in ROCm. so this check may be unnecessary.We should flag if a sticky bit folder is present in /opt/rocm.
This might be a simpler check. 
```
find -H /opt/rocm/core \
  ( \
    ! -uid 0 \
    -o ! -gid 0 \
    -o ( ! -type l ( -perm /022 -o -perm /6000 ) ) \
  ) \
  -print

```Or the python version 

```
group_other_writable = bool(mode & 0o022)
setid = bool(mode & (stat.S_ISUID | stat.S_ISGID))
if not_root or group_other_writable or setid:
    bad.append(entry)

```

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Removed timeout= from subprocess.run, deleted the FILE_SECURITY_TIMEOUT_SEC constant and the TimeoutExpired handler

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Added if result.returncode != 0: print(WARN + stderr[:200]) so a partial scan (e.g. permission denied on a subtree) no longer passes silently

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Added -xdev right after the prefix

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Wrapped the ownership clause: ( ! -uid 0 -o ! -gid 0 ) -o ( ! -type l ( ... ) )

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Dropped ! -perm -1000 from find and the sticky exemption in Python (group_other_writable = bool(mode & 0o022)), so any group/other-writable path — sticky or not — is flagged

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: Updated.

build_tools/packaging/linux/native_linux_package_install_test.py
dileepr1: covered by earlier subprocess.run change, also logs non-empty stderr on a clean exit

nunnikri: Now overall looks good. But want to see the full ci run to confirm no other unexpected conditions we need to take care

nunnikri (approved):

PR #6604, opened 2026-07-16
Check return value in GPU driver info print / sanity check

ScottTodd:
## Motivation

Follow-up to https://github.com/ROCm/TheRock/pull/2272 which addressed https://github.com/ROCm/TheRock/issues/2209.

This script is used as a "Driver / GPU sanity check" but it wasn't actually checking if the test commands succeeded, causing future job steps to run on machines with faulty GPU state like on https://github.com/ROCm/TheRock/actions/runs/29448973188/job/87559403977?pr=6550:

* 
    ```
    Run python ./build_tools/print_driver_gpu_info.py
    === Sanity check: driver / GPU info ===
    
    === hipInfo.exe ===
    ++ Exec [B:\runner\_work\TheRock\TheRock]$ 'build\bin\hipInfo.exe'
    
    checkHipErrors() HIP API error = 0100 "no ROCm-capable device is detected" from file <C:\home\runner\_work\TheRock\TheRock\rocm-systems\projects\hip-tests\samples\1_Utils\hipInfo\hipInfo.cpp>, line 181.
    
    === End of sanity check ===
    ``` 
* 
    ```diff
     Run python build_tools/github_actions/test_executable_scripts/test_runner.py
     Test project B:/runner/_work/TheRock/TheRock/build/bin/hipblas
     test 1
         Start 1: hipblas-test_quick_suite
     
     1: Test command: B:\runner\_work\TheRock\TheRock\build\bin\hipblas-test.exe "--gtest_filter=*quick*-*known_bug*"
     1: Working Directory: B:/runner/_work/TheRock/TheRock/build/bin/hipblas
     1: hipBLAS version 3.6.0.fa9cdd18
     1: Query device error: cannot get device count 
     1: Error: No devices found
     1/1 Test #1: hipblas-test_quick_suite .........***Failed    0.08 sec
     hipBLAS version 3.6.0.fa9cdd18
    -Query device error: cannot get device count 
    -Error: No devices found
    
     0% tests passed, 1 tests failed out of 1
    ```

## Test Plan

Watch CI on this PR, see if any runners have GPU issues that this catches. We could also directly trigger https://github.com/ROCm/TheRock/actions/workflows/test_artifacts.yml.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #6976, opened 2026-07-29
Add "Driver / GPU sanity check" to pytorch test workflows

HereThereBeDragons:
Follow `.github/workflows/test_component.yml` and add the "Driver / GPU sanity check" also to pytorch test workflows to see if the GPU is healthy before running any tests on it.

https://github.com/ROCm/TheRock/issues/6977

### DIFF
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -192,6 +192,11 @@ jobs:
           python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
           pip freeze
 
+      - name: Driver / GPU sanity check
+        timeout-minutes: 3
+        run: |
+          python ./build_tools/print_driver_gpu_info.py
+
       - name: Run rocm-sdk sanity tests
         timeout-minutes: 5
         run: |
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -308,6 +308,11 @@ jobs:
         run: |
           ./build_tools/health_status.py
 
+      - name: Driver / GPU sanity check
+        timeout-minutes: 3
+        run: |
+          python ./build_tools/print_driver_gpu_info.py
+
       - name: Run rocm-sdk sanity tests
         timeout-minutes: 5
         run: |


.github/workflows/test_pytorch_wheels.yml
```diff
@@ -192,6 +192,11 @@ jobs:
           python -m pip install -r external-builds/pytorch/pytorch/.ci/docker/requirements-ci.txt
           pip freeze
 
+      - name: Driver / GPU sanity check
+        timeout-minutes: 3
+        run: |
+          python ./build_tools/print_driver_gpu_info.py
```
ScottTodd, line 198: I think we could move this up prior to the "checkout source repos" and "set up virtual environment" steps, so we can fail the job quickly on runners without a healthy GPU. Those steps can take several minutes.

We could make the same change to https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_component.yml

ScottTodd, line 198: Note that this script does not currently error if a check fails. I have a PR that will change that, but I need to revise test_component.yml to only run the script when using a GPU test runners (some recently added tests run on CPU runners instead):
* https://github.com/ROCm/TheRock/pull/6604

.github/workflows/test_pytorch_wheels.yml
ScottTodd, line 1: Shall we also update https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_linux_jax_wheels.yml and https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_linux_jax_wheels_partial.yml ? I think that could be in this PR, or a separate PR.

ScottTodd (approved):
Thanks. As part of https://github.com/ROCm/TheRock/issues/6711, I'm also planning on documenting how we expect to run GPU health/sanity checks like this
1. Before jobs even start, via prolense / the runner control plane somehow
2. Before component/framework test jobs (as this PR is doing)

.github/workflows/test_pytorch_wheels.yml
HereThereBeDragons: are you sure? we are needing amd-smi and rocminfo. are we not taking those from the venv?

please merge at your disgression if you want it in for this nightly, otherwise i will do it tomorrow.

ScottTodd: We probably are actually using them from the venv, yeah. We could at least run the check between "setup venv" and "checkout source repos", but that would be a larger workflow step reordering. Fine to have this as-is on this PR.

ScottTodd:
I ran some tests after merge (these tests could have been in the PR description, until we get pytorch tests running on PRs)
* Linux: https://github.com/ROCm/TheRock/actions/runs/30565987573/job/90950498017

    ```
    ++ Exec [/__w/TheRock/TheRock]$ /__w/TheRock/TheRock/.venv/bin/amd-smi static
    ```

* Windows: https://github.com/ROCm/TheRock/actions/runs/30567945417/job/90957062385

    ```
    ++ Exec [C:\actions-runner\_work\TheRock\TheRock]$ 'C:\actions-runner\_work\TheRock\TheRock\.venv\Scripts\hipInfo.exe'
    ```

So yes this is working, and the tools are coming from the venv.

Issue #6978, opened 2026-07-29
[Issue]: JAX plugin wheels are built with an empty git hash


WBobby:
## What happens

Every `jax_rocm7_plugin` wheel produced by the multi-arch Linux JAX workflows is
built with an empty `jaxlib_git_hash`, so a published wheel cannot be mapped back
to the JAX commit it was built from. The job succeeds, so this is invisible unless
you read the log.

## Evidence

Release run
[30326020986](https://github.com/ROCm/rockrel/actions/runs/30326020986/job/90191315298)
succeeded in 1h14m, and the "Build JAX Wheels in manylinux container" step logs:

    fatal: detected dubious ownership in repository at '/workspace/jax-source'

The wheel it published confirms the effect. In
`jax_rocm7_plugin-0.10.0+rocm7.15.0a20260728-cp314-...whl`, the file
`jax_rocm7_plugin/version.py` contains:

    _release_version: str = '0.10.0+rocm7.15.0a20260728'
    _git_hash: str | None = None

The version was substituted, the hash was not. No 40-hex string appears anywhere
in the wheel.

## Root cause

The build command is wrapped in `bash -lc '...'` with single quotes, so
`--bazel_options=--//jaxlib/tools:jaxlib_git_hash=$(git rev-parse HEAD)` is
expanded by the *container's* shell, not the host's. The `/workspace/...` path in
the error message is what identifies where it ran; the host path would have been
`.../rockrel/jax-source`.

The container runs `--user root` against `${GITHUB_WORKSPACE}` mounted from the
runner, and the `rocm-jax` manylinux image does not set `safe.directory`, so git
aborts. Note that TheRock's own `dockerfiles/build_manylinux_x86_64.Dockerfile`
does set `safe.directory '*'`; that image is not the one used here.

Nothing catches the failure. A failed command substitution in an *argument*
position does not trip `set -e` (only assignments do), and the inner `bash -lc`
does not enable it anyway. `build.py` therefore receives
`jaxlib_git_hash=` and builds normally.

## Affected

- `.github/workflows/multi_arch_build_linux_jax_wheels.yml`
- `.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml`

Both carried the identical line, so neither the release nor the CI path was
stamped.

## Proposed fix

Resolve the hash on the host, where the checkout is owned by the user running the
step; validate it as a 40-hex sha so a missing hash fails the job instead of
producing an unstamped wheel; forward it into the container with `--env`.
[...]

PR #6979, opened 2026-07-29
fix(jax): stamp JAX wheels with the real source commit hash

WBobby:
## Motivation

`jaxlib_git_hash` was resolved by a `git` call that ran *inside* the manylinux
container, where git rejects the runner-owned checkout as dubiously owned. The
command substitution failed silently, so every JAX plugin wheel from these two
workflows was published without a git hash while the job stayed green.

Issue: https://github.com/ROCm/TheRock/issues/6978

## Technical Details

`actions/checkout` already resolves the commit it materialised and exposes it as
`steps.<id>.outputs.commit` (internally `git log -1 --format=%H` in the checkout
directory). Taking the value from there removes the git call from the workflow
entirely, so the change is an `id:` on the checkout step plus one modified line.
* The runner substitutes `${{ ... }}` while rendering the `run:` block, so no
  shell ever sees a variable and nothing has to cross the container boundary.
  This is exactly why the old line failed: `$(...)` *is* shell syntax, so the
  single quotes deferred it to the container, where git rejects the runner-owned
  checkout. `${{ }}` is not shell syntax and the quotes do not apply to it.
* Nothing needs `--env` or `safe.directory`. The container makes no git call for
  this at all: `build/build.py` never shells out to git, and neither `.bazelrc`
  nor `build/rocm/rocm.bazelrc` sets `--stamp` or a `workspace_status_command`.
  `//jaxlib/tools:jaxlib_git_hash` is a plain `string_flag` passed through to
  `build_wheel.py` as a string.
* The resolved sha appears in the step's echoed script, so no extra `echo` is
  needed for observability.
Both the release and CI workflows carried the identical line and are changed
identically.

## Test Plan

Dispatched `multi_arch_build_linux_jax_wheels.yml` on this branch for a single
matrix cell (`build_mode=manylinux`, `python_version=3.14`,
`jax_ref=rocm-jaxlib-v0.10.2`, `rocm_version=7.15.0a20260728`,
`release_type=dev`) with the GPU test job enabled, then inspected the published
wheel.

## Test Result

Run [30513543402](https://github.com/ROCm/TheRock/actions/runs/30513543402), success in 1h17m14s. All 21 steps green; the native v0.9.1 step was skipped as expected.
* The expression is substituted before any shell runs, and Bazel receives it
  intact:
  `--bazel_options=--//jaxlib/tools:jaxlib_git_hash="0dad77d6b293b87573a9e8760b0ddf677a32ea43"`
  `'build' options: ... --//jaxlib/tools:jaxlib_git_hash=0dad77d6b293b87573a9e8760b0ddf677a32ea43`
* That sha is `refs/heads/rocm-jaxlib-v0.10.2` in ROCm/jax (`git ls-remote`) and
  matches what the `Checkout JAX` step reported.
* `dubious ownership`, `git rev-parse` and `GIT_CONFIG` each appear 0 times in
  the log: the container no longer invokes git, so dropping the `safe.directory`
  environment variables is confirmed harmless.
* Published `jax_rocm7_plugin` wheel: `_git_hash: str = '0dad77d6...'`, where the
  20260728 nightly had `None`. `_release_version` unchanged at
  `'0.10.2+rocm7.15.0a20260728'`.
* The GPU test job installed the wheels from this run and passed on
  gfx94X-dcgpu: 982, 480, 19 and 2 passed, 0 failed.

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_linux_jax_wheels.yml
+++ b/.github/workflows/multi_arch_build_linux_jax_wheels.yml
@@ -148,6 +148,7 @@ jobs:
           ref: ${{ inputs.jax_ref }}
 
       - name: Checkout JAX
+        id: checkout_jax
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           path: jax-source
@@ -219,7 +220,7 @@ jobs:
                 --bazel_options=--repo_env=ROCM_PATH=$(rocm-sdk path --root) \
                 --bazel_options=--repo_env=ML_WHEEL_TYPE=release \
                 --bazel_options=--repo_env=ML_WHEEL_VERSION_SUFFIX="${ML_WHEEL_VERSION_SUFFIX}" \
-                --bazel_options=--//jaxlib/tools:jaxlib_git_hash=$(git rev-parse HEAD) \
+                --bazel_options=--//jaxlib/tools:jaxlib_git_hash="${{ steps.checkout_jax.outputs.commit }}" \
                 --verbose \
                 --detailed_timestamped_log \
                 --output_path=$(pwd)/dist
--- a/.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
+++ b/.github/workflows/multi_arch_build_linux_jax_wheels_ci.yml
@@ -137,6 +137,7 @@ jobs:
           ref: ${{ inputs.jax_ref }}
 
       - name: Checkout JAX
+        id: checkout_jax
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           path: jax-source
@@ -208,7 +209,7 @@ jobs:
                 --bazel_options=--repo_env=ROCM_PATH=$(rocm-sdk path --root) \
                 --bazel_options=--repo_env=ML_WHEEL_TYPE=release \
                 --bazel_options=--repo_env=ML_WHEEL_VERSION_SUFFIX="${ML_WHEEL_VERSION_SUFFIX}" \
-                --bazel_options=--//jaxlib/tools:jaxlib_git_hash=$(git rev-parse HEAD) \
+                --bazel_options=--//jaxlib/tools:jaxlib_git_hash="${{ steps.checkout_jax.outputs.commit }}" \
                 --verbose \
                 --detailed_timestamped_log \
                 --output_path=$(pwd)/dist


.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -244,12 +244,25 @@ jobs:
...
           PYTHON_VERSION: ${{ inputs.python_version }}
         run: |
+          # Resolved on the host: the container runs as root against a
+          # runner-owned checkout, where git rejects the repo as dubiously owned.
+          JAXLIB_GIT_HASH="$(git rev-parse HEAD)"
+          if [[ ! "${JAXLIB_GIT_HASH}" =~ ^[0-9a-f]{40}$ ]]; then
+            echo "::error::refusing to build: jax-source HEAD resolved to '${JAXLIB_GIT_HASH}'"
+            exit 1
+          fi
+          echo "jaxlib_git_hash=${JAXLIB_GIT_HASH}"
```
ScottTodd: The `if` and `echo` here look unnecessary, remove?

* If `python build/build.py` wants to reject `JAXLIB_GIT_HASH` values that don't match some pattern, that should be up to it. Keep workflows as simple as possible: https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/github_actions_style_guide.md#prefer-python-scripts-over-inline-bash
* Commands are already echo'd by default in github actions, no need to also run `echo` manually like this

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -244,12 +244,25 @@ jobs:
           ROCM_VERSION: ${{ inputs.rocm_version }}
           PYTHON_VERSION: ${{ inputs.python_version }}
         run: |
+          # Resolved on the host: the container runs as root against a
+          # runner-owned checkout, where git rejects the repo as dubiously owned.
```
ScottTodd: Could the container itself apply ownership settings for the checkout to fix instead? Again, prefer to keep workflows as simple as possible. Every single line of bash in a workflow should be scrutinized. Treat workflows as recipes for **users** to run. You wouldn't ask a user to run any of this code... right? Users should just run scripts in sequence. Same for workflows.

WBobby:
Both removed — and the host-side resolution is gone entirely. The hash now comes
from `steps.checkout_jax.outputs.commit`, so the change is an `id:` on the
`Checkout JAX` step plus one modified line, with no bash added to either workflow.

ScottTodd (approved):
The code changes here LGTM. There are some merge conflicts to resolve.

WBobby:

WBobby:
Fixed the merge conflicts, and the JAX build passed.

PR #6981, opened 2026-07-29
feat(packaging): add aggregate DEB/RPM metadata scripts

nunnikri:
Add two Python scripts and a setup guide for aggregating DEB and RPM package metadata from multiple sources (e.g. ROCm core + RVS) into a single APT/DNF repository endpoint, backed by S3 and CloudFront.

JIRA ID : https://amd-hub.atlassian.net/browse/ROCM-25115

## Motivation
ROCm users currently need to configure separate repo entries for core ROCm packages and supplementary packages like RVS. This creates friction at install time. These scripts merge the metadata from multiple sources into a single repository index so users can install from a single baseurl, while the actual package files stay in their original buckets — only metadata is aggregated.

## Technical Details

**`aggregate_deb_metadata.py`**
- Fetches `Packages.gz` from multiple APT sources (dists-style or flat-style)
- Rewrites `Filename:` with owner-prefixed pool paths (`pool/core/`, `pool/rvs/`) so
  CloudFront can route each prefix to the correct origin bucket
- Merges all stanzas with NEVRA-based collision detection (errors on duplicate
  package+version+arch from different sources)
- Generates `Release` file with MD5/SHA1/SHA256 checksums (RFC 2822 date format
  required by APT)
- Optional GPG signing via `--sign-key` producing `Release.gpg` + `InRelease`

**`aggregate_rpm_metadata.py`**
- Fetches `primary.xml.gz` via `repomd.xml` from multiple RPM sources
- Rewrites `<location href>` with owner-prefixed paths (`core/`, `rvs/`)
- Merges all `<package>` entries with NEVRA-based collision detection
- Regenerates `repomd.xml` with fresh SHA256 checksums
- Optional GPG signing producing `repomd.xml.asc`

**Common design:**
- Sources passed as repeatable `--source` CLI arguments — no hardcoded limit, no
  hardcoded sources in the script
- `--output-dir` writes locally (no dependencies beyond stdlib); `--output-bucket`
  uploads to S3 (requires `boto3`); both can be used together
- 0 sources → error; 1 source → works (no-op merge, useful for format conversion)
- `filelists.xml` intentionally omitted — ROCm packages use name-based and
  soname-based dependencies only, not file-path dependencies

**`README-aggregate-repo.txt`** documents:
- Step-by-step usage with ROCm nightly + RVS as example sources
- Optional S3 upload with prerequisites
- CloudFront behavior configuration for production (path pattern → origin mapping
  for DEB and RPM)
- End-user repo setup commands for Ubuntu and RHEL
- Optional GPG signing client setup

### DIFF
--- a/build_tools/packaging/linux/aggregate_deb_metadata.py
+++ b/build_tools/packaging/linux/aggregate_deb_metadata.py
@@ -82,16 +82,16 @@ def parse_source(value: str) -> dict:
         raise argparse.ArgumentTypeError(
             f"--source must be 'name,base_url,style,pool_prefix', got: {value!r}"
         )
-    name, base_url, style, pool_prefix = parts
+    name, base_url, style, pool_prefix = [p.strip() for p in parts]
     if style not in ("dists", "flat"):
         raise argparse.ArgumentTypeError(
             f"style must be 'dists' or 'flat', got: {style!r}"
         )
     return {
-        "name": name.strip(),
-        "base_url": base_url.strip(),
-        "style": style.strip(),
-        "pool_prefix": pool_prefix.strip(),
+        "name": name,
+        "base_url": base_url,
+        "style": style,
+        "pool_prefix": pool_prefix,
     }
 
 
--- a/build_tools/packaging/linux/aggregate_deb_metadata.py
+++ b/build_tools/packaging/linux/aggregate_deb_metadata.py
@@ -9,6 +9,14 @@ owner-prefix so APT resolves package files through the correct CloudFront path,
 merges all stanzas, regenerates Release/InRelease, and writes output locally
 and/or uploads to S3.
 
+Limitations:
+  - Packages index compression: tries .xz, .bz2, .gz, and uncompressed in
+    APT preference order. Fails fast if none are available for a source.
+  - A single --architecture value is supported per run. To serve multiple
+    architectures, run the script once per architecture. Note that each run
+    overwrites dists/<suite>/Release, so multi-arch runs must use separate
+    --output-prefix values or the Release file must be manually merged.
+
 Usage:
     python aggregate_deb_metadata.py \\
         --source "core,https://repo.amd.com/rocm/packages-multi-arch/ubuntu2404,dists,pool/core" \\
@@ -44,12 +52,15 @@ At least one of --output-dir or --output-bucket must be provided.
 """
 
 import argparse
+import bz2
 import gzip
 import hashlib
+import lzma
 import os
 import subprocess
 import sys
 import tempfile
+import urllib.error
 import urllib.request
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


build_tools/packaging/linux/aggregate_deb_metadata.py
```diff
@@ -0,0 +1,488 @@
...
+        name = source["name"]
+        base_url = source["base_url"]
+        style = source["style"]
+        pool_prefix = source["pool_prefix"]
+
+        print(f"\n[{name}] Fetching Packages.gz from {base_url}")
+        print(f"         style={style}  pool_prefix={pool_prefix}")
+
+        try:
+            raw = fetch_packages_gz(base_url, style, suite, component, arch)
```
marbre: This fetches` Packages.gz` only and repos serving `.xz`-only or uncompressed indices will 404. Can you confirm that all ROCm sources guarantee `.gz`?

build_tools/packaging/linux/aggregate_deb_metadata.py
```diff
@@ -0,0 +1,488 @@
...
+            "pool_prefix is written into Filename: (e.g. 'pool/core')."
+        ),
+    )
+    parser.add_argument("--suite", default="stable", help="APT suite (default: stable)")
+    parser.add_argument(
+        "--component", default="main", help="APT component (default: main)"
+    )
+    parser.add_argument(
+        "--architecture", default="amd64", help="Package architecture (default: amd64)"
+    )
```
marbre, line 343: As we a single architecture is provided here we would need to run it a second time if we also want to produce packages for another architecture. However, the second run would overwrite `dists/<suite>/Release`. This limitation at least needs to be documented.

build_tools/packaging/linux/aggregate_rpm_metadata.py
```diff
@@ -0,0 +1,428 @@
...
+def rewrite_location_href(pkg: ET.Element, href_prefix: str) -> None:
+    """Rewrite <location href> to href_prefix/<basename> in-place.
+
+    DNF resolves this relative path against the repo baseurl.
+    CloudFront routes <prefix>/* requests to the correct origin bucket.
+    """
+    ns = {"c": NS_COMMON}
+    location = pkg.find("c:location", ns)
+    if location is None:
+        return
```
marbre: Shouldn't this fail fast instead?

marbre (approved):
I think this scripts have some issues and don't follow our style guides e.g. with not failing fast in several places. This needs to be improved _before_ we use those scripts in production. Furthermore, the scripts are lacking tests.

I am aware that this is a prototype but let's make sure we address these before bringing the scripts into use. Fine to go in for now to allow others to more easily contribute and iterate. Please file issues before merging.

build_tools/packaging/linux/aggregate_deb_metadata.py
nunnikri: Thanks @marbre for the catch. Now the script is improved to find all the possible options in the standard order.   .xz/.bz2/.gz and uncompressed Packages

build_tools/packaging/linux/aggregate_deb_metadata.py
nunnikri: yes currently we are supporting only 1 arch, added this as a limitation. Will revisit in a later point of time if we need to enhance on this

build_tools/packaging/linux/aggregate_rpm_metadata.py
nunnikri: Yes done.

nunnikri:
> I think this scripts have some issues and don't follow our style guides e.g. with not failing fast in several places. This needs to be improved _before_ we use those scripts in production. Furthermore, the scripts are lacking tests.
> 
> I am aware that this is a prototype but let's make sure we address these before bringing the scripts into use. Fine to go in for now to allow others to more easily contribute and iterate. Please file issues before merging.

Thanks @marbre for the feedback. Have addressed most of the comments and added unit testing as well. And that has helped to fix some issues.

PR #6987, opened 2026-07-30
parallelize native package builds using ProcessPoolExecutor

raramakr:
Use ProcessPoolExecutor to build packages concurrently. Each package
  builds in an isolated temp directory with its own log file, then
  artifacts are moved to the final destination. 
  
  - Add -j/--parallel flag (defaults to all CPU cores)
  - Isolate builds in temp directories to avoid conflicts
  - Per-package cleanup instead of global cleanup
  - Simplify error handling (executor handles failures)


raramakr:
Closing this and will be addressed in https://github.com/ROCm/TheRock/pull/7380

PR #6867, opened 2026-07-26
tests(rocm[profiler]): Add sanity tests after the rocm[profiler] pip install

dgaliffiAMD:
## Motivation

The `rocm[profiler]` package bundles `rocprof-sys` and its runtime dependencies.
`libprofiler-hub.so` is a newly introduced library that `librocprof-sys` links
against (NEEDED), but it was initially missed from the packaging manifest, so
the wheel shipped without it and `rocprof-sys` tools failed to load at runtime
(fixed in #6850). Nothing in CI caught this because we had no test that
exercises the *installed* profiler package.

This PR adds that missing coverage: an integration test that installs the
profiler wheel and verifies its runtime integrity, plus a CI job that runs it.

## Issue Tracking

For https://github.com/ROCm/TheRock/issues/6856.
JIRA ID - AIPROFSYST-699

## Technical Details

**New installation test module (`rocm_sdk/tests/profiler_test.py`)** — imports the 
physically installed rocm-profiler package (not the source tree) and runs four 
checks. The package import and shared-library probe happen in setUpClass, 
so the module is safe to import on any platform and skips cleanly (rather than 
erroring at collection) when the package is absent or on Windows. Helper sub-processes 
(`ldd`, console scripts) run under a bounded timeout.

Package-wide tests are unprefixed; tests specific to rocprofiler-systems are prefixed 
`rocprof_sys_` so other profiler components (e.g. rocprof-compute) can add distinguishable 
tests to the same module:

- `test_installation_layout` — `rocm_sdk` and the profiler module are real
  (non-namespace) packages and siblings in the same `site-packages`, the
  precondition for the cross-package `$ORIGIN` RPATHs to resolve.
- `test_all_required_libraries_resolve` — runs `ldd` on every profiler `.so` and
  asserts no NEEDED dependency is "not found". Generalizes the #6850 guard from
  one named library to *any* missing bundled / cross-package / system
  dependency, and stays GPU-free and deterministic (resolves the link map
  without executing anything). Also asserts ldd is on PATH and treats a non-zero ldd 
  exit as a failure, so a broken or missing ldd can't produce a false pass.
- `test_rocprof_sys_unversioned_library_aliases` — mirrors both source glob
  patterns (`librocprof-sys*.so.*`, `libprofiler-hub*.so.*`) from
  `ensure_profiler_library_symlinks()` and asserts the unversioned `.so` alias
  exists for each versioned library.
- `test_rocprof_sys_console_scripts` — execs
[...]

PR #6995, opened 2026-07-30
fix(profiler): expose profiler runtime libs from pip wrapper

anujshuk-amd:
## Motivation

The rocm[profiler] pip package includes the rocprof-sys tools and their shared libraries inside the Python environment. Direct profiling works, but once an executable is rewritten with rocprof-sys-instrument, the new binary needs librocprof-sys-rt.so at startup. In the pip layout that library lives under site-packages, which the dynamic loader does not search by default, so the run fails unless users manually set LD_LIBRARY_PATH.

This change makes the pip-installed profiler tools set up that library path automatically, so instrumented binaries can be profiled without extra manual environment setup.

## Technical Details

The pip entry points for the profiler tools are small Python wrappers. They resolve the installed _rocm_profiler package location and then execv into the native binaries, such as bin/rocprof-sys-run and bin/rocprof-sys-instrument.

The wrapper was already extending LD_LIBRARY_PATH for ROCm sysdeps from _rocm_sdk_core, but it was not adding the profiler package’s own lib directory. That directory contains librocprof-sys-rt.so, which is required by binaries rewritten with rocprof-sys-instrument.

This change updates the wrappers LD_LIBRARY_PATH setup to also prepend _rocm_profiler/lib when it exists. The path is discovered from the installed profiler module location, so it does not depend on a hardcoded Python version or virtualenv path. Existing LD_LIBRARY_PATH entries are preserved, and duplicate paths are avoided.

JIRA ID : ROCM-28454

## Test Plan

Tested on AAC6 with the nightly ROCm pip package. I was able to reproduce the original librocprof-sys-rt.so missing library error with an instrumented binary. After the fix, I reran the same rocprof-sys-instrument and rocprof-sys-run flow, and the instrumented executable ran successfully without manually setting LD_LIBRARY_PATH.

## Test Result

Described tests issue passes.

## Submission Checklist


build_tools/packaging/python/templates/rocm-profiler/src/rocm_profiler/_cli.py
ScottTodd, line 1: > ## Test Plan
> 
> Tested on AAC6 with the nightly ROCm pip package. I was able to reproduce the original librocprof-sys-rt.so missing library error with an instrumented binary. After the fix, I reran the same rocprof-sys-instrument and rocprof-sys-run flow, and the instrumented executable ran successfully without manually setting LD_LIBRARY_PATH.

Let's build on the new tests that https://github.com/ROCm/TheRock/pull/6867 is adding to also test for 

> Direct profiling works, but once an executable is rewritten with rocprof-sys-instrument, the new binary needs librocprof-sys-rt.so at startup.

That can be a follow-up to both PRs.

ScottTodd: To reduce the risk of `os.environ` changes leaking outside of this module and this `os.execv` call, I think this code could use `os.execve` instead: https://docs.python.org/3/library/os.html#os.execve

ScottTodd: Can you add some docstring comments to this `_extend_ld_library_path()` function and `_extend_pythonpath_for_compute()`?

In particular I think it's worth highlighting
* These CLI calls replace the currently running process and thus can't use the python library preloading documented at https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#performing-initialization
* Why rocprof binaries can't otherwise find/load their dependencies without both `LD_LIBRARY_PATH` and `PYTHONPATH` (we don't have similar setup for binaries in https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py, so what's special/different about these?)

ScottTodd (approved):

build_tools/packaging/python/templates/rocm-profiler/src/rocm_profiler/_cli.py
anujshuk-amd: - switched to os.execve with a local env copy
- added docstrings on the LD_LIBRARY_PATH / PYTHONPATH helpers
- will add automated coverage for instrumented binaries loading librocprof-sys-rt.so as a follow-up after #6867

anujshuk-amd: Added short docstrings on both helpers covering the exec/preload constraint, why profiler needs both env vars, and how this differs from rocm_sdk_core/_cli.py.

anujshuk-amd: Done. Copy os.environ, apply path changes on that local dict, and call os.execve so nothing leaks back into the wrapper process.

dgaliffiAMD: I agree we should add this test now that the scaffolding is there, let's create an issue ticket to track that.

dgaliffiAMD (approved):

PR #7002, opened 2026-07-30
fix(rocprofiler-compute): Gate rocprof-compute analyze deps behind an opt-in compute-analyze extra

vedithal-amd:
## Motivation

`rocprof-compute analyze` needs dash, pandas and numpy; `rocprof-compute profile` needs nothing. The `rocm-profiler` wheel declares no dependency metadata, so analyze fails at import.

The two modes want different environments:

| mode | install | analyze deps |
|---|---|---|
| profile | `pip install rocm rocm-profiler torch` | none, workload environment untouched |
| analyze | `pip install --extra-index-url <rocm_index> rocm-profiler[compute-analyze]` | full stack, no torch to conflict with |

Under that split the deps belong in an extra, and exact pins remain appropriate because the analyze venv is isolated.

## Related Issues

ISSUE ID: https://github.com/ROCm/TheRock/issues/7003

JIRA ID: AIPROFCOMP-788

## Technical Details

- Deps are read at build time from the staged payload at `libexec/rocprofiler-compute/requirements.txt`, which `PROFILER_WHEEL_INCLUDES` already carries into the wheel. rocprofiler-compute stays the single source of truth, so pin changes need no change here.
- `populate_runtime_files()` stages that payload before `build_packages()` invokes `setup.py`, so the file is on disk when read. `setup.py` already loads `_dist_info.py` off disk the same way.
- The extra is declared even when the payload lacks the file, so `pip install rocm-profiler[compute-analyze]` resolves against any build rather than warning about an unknown extra.
- The analyze venv installs with `--extra-index-url <rocm_index>`, so the extra's deps come from PyPI and `rocm-profiler` comes from the ROCm index.
- Named per tool, not per mode: this wheel also carries rocprofiler-systems, so a bare `analyze` gets ambiguous once anything else here needs Python packages. A second tool should get its own extra reading its own requirements file, not a merged list, which would force both tools' pins to relax.
- `rocm[profiler]` does not chain into the extra, by design. Analyze deps stay opt-in.

## Test Plan

Staged the profiler package with `PopulatedDistPackage`, built the wheel through TheRock's own `build_packages()`, then read METADATA and ran `pip install --dry-run` in a clean venv.

## Test Result

Every dep from the staged `requirements.txt` is gated behind the extra, with comment lines dropped. A plain install pulls no analyze deps; `[compute-analyze]` pulls the full set. A payload staged without `requirements.txt` builds cleanly with the extra declared and empty.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/python/templates/rocm-profiler/README.md
+++ b/build_tools/packaging/python/templates/rocm-profiler/README.md
@@ -32,8 +32,11 @@ invoking a console script will result in a `FileNotFoundError`.
 
 ## Packaging Notes
 
-- No `install_requires` are declared.
-- Dependencies are managed by the `rocm` meta package.
+- `rocprof-compute analyze`'s Python dependencies are declared under the
+  `compute-analyze` extra, read at build time from the staged rocprofiler-compute
+  payload at `libexec/rocprofiler-compute/requirements.txt`.
+- Analyze mode should run from its own virtualenv:
+  `pip install rocm-profiler[compute-analyze]`.
 - Versioning is centrally managed via `rocm_sdk._dist_info`.
 
 When integrated into the full ROCm packaging pipeline, this package will be
--- a/build_tools/packaging/python/templates/rocm-profiler/setup.py
+++ b/build_tools/packaging/python/templates/rocm-profiler/setup.py
@@ -32,11 +32,39 @@ def _load_local_dist_info():
     return module
 
 
+def _read_compute_requirements(platform_package_name: str) -> list[str]:
+    """Read rocprof-compute's analyze deps from the staged wheel payload.
+
+    Reading rocprofiler-compute's installed requirements.txt rather than
+    duplicating the list keeps it the single source of truth.
+    """
+    requirements_path = (
+        THIS_DIR
+        / "platform"
+        / platform_package_name
+        / "libexec"
+        / "rocprofiler-compute"
+        / "requirements.txt"
+    )
+    if not requirements_path.exists():
+        return []
+
+    requirements = []
+    for raw_line in requirements_path.read_text().splitlines():
+        line = raw_line.split("#", 1)[0].strip()
+        # pip options such as `-r other.txt` are not valid PEP 508 requirements.
+        if not line or line.startswith("-"):
+            continue
+        requirements.append(line)
+    return requirements
+
+
 dist_info = _load_local_dist_info()
 my_package = dist_info.ALL_PACKAGES["profiler"]
 packages = find_packages(where="./src")
 platform_package_name = my_package.get_py_package_name()
 packages.append(platform_package_name)
+extras_require = {"compute-analyze": _read_compute_requirements(platform_package_name)}
 
 version = os.environ.get("ROCM_SDK_VERSION")
 if version is None:
@@ -57,6 +85,7 @@ setup(
     },
     include_package_data=True,
     zip_safe=False,
+    extras_require=extras_require,
     options={
         "bdist_wheel": {
             "plat_name": os.getenv(


jbonnell-amd (commented):
Assuming we are good with adding all of `requirements.txt` dependencies into the `rocm-profiler` install, should we also do a follow-up PR to add in requirements for sdk and systems as well? We would also need to eliminate duplicates and ensure version requirements do not mismatch, but this would resolve any other potential dependency issues we'd have.

vedithal-amd:
> Assuming we are good with adding all of `requirements.txt` dependencies into the `rocm-profiler` install, should we also do a follow-up PR to add in requirements for sdk and systems as well? We would also need to eliminate duplicates and ensure version requirements do not mismatch, but this would resolve any other potential dependency issues we'd have.

@jbonnell-amd I dont think systems has any requirements.txt python deps and sdk is not part of rocm-profiler package

vedithal-amd:
@jbonnell-amd per my comment in ROCM-28653 we should not merge this PR since compute's python dependencies will clash with those of torch and the python application being profiled.

Folks would be better of creating a virtual environment for analysis mode based of our requirements.txt and a message in analysis mode instructing users to do so along with the path to requirements.txt should be enough

marbre:
> @jbonnell-amd per my comment in ROCM-28653 we should not merge this PR since compute's python dependencies will clash with those of torch and the python application being profiled.
> 
> Folks would be better of creating a virtual environment for analysis mode based of our requirements.txt and a message in analysis mode instructing users to do so along with the path to requirements.txt should be enough

If you don't want to merge this, maybe close or move to draft?

build_tools/packaging/python/templates/rocm-profiler/setup.py
```diff
@@ -32,11 +32,44 @@ def _load_local_dist_info():
...
+
+    Returns an empty list when the payload does not carry the file.
+    """
+    requirements_path = (
+        THIS_DIR
+        / "platform"
+        / platform_package_name
+        / "libexec"
+        / "rocprofiler-compute"
+        / "requirements.txt"
```
ScottTodd, line 47: Another comment on top of the other concerns: https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-compute/requirements.txt currently pins exact versions of many packages. If users install that while they have other packages installed too, pip's dependency resolver will likely warn about incompatible package version requirements.

For distribution for `install_requires` I would prefer to follow the best practices in https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/
* The list should be _minimal_
* Indicate any known lower or upper bounds
    > It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

vedithal-amd: Agreed, dropped `install_requires`. Analyze now runs from its own venv via `pip install rocm-profiler[compute-analyze]`, so the pins only ever apply there and never reach the environment being profiled.

vedithal-amd:
> > @jbonnell-amd per my comment in ROCM-28653 we should not merge this PR since compute's python dependencies will clash with those of torch and the python application being profiled.
> > Folks would be better of creating a virtual environment for analysis mode based of our requirements.txt and a message in analysis mode instructing users to do so along with the path to requirements.txt should be enough
> 
> If you don't want to merge this, maybe close or move to draft?

I think we can still merge this PR if we don't use `install_requires` but instead add a `rocm-profiler[compute-analyze]` extra which pull in compute's analyze deps. This way we can provide a way for users to setup a venv for analyze only.

I have implemented this change.

marbre:
> > > @jbonnell-amd per my comment in ROCM-28653 we should not merge this PR since compute's python dependencies will clash with those of torch and the python application being profiled.
> > > Folks would be better of creating a virtual environment for analysis mode based of our requirements.txt and a message in analysis mode instructing users to do so along with the path to requirements.txt should be enough
> > 
> > 
> > If you don't want to merge this, maybe close or move to draft?
> 
> I think we can still merge this PR if we don't use `install_requires` but instead add a `rocm-profiler[compute-analyze]` extra which pull in compute's analyze deps. This way we can provide a way for users to setup a venv for analyze only.
> 
> I have implemented this change.

@jbonnell-amd @vedithal-amd how do you want to proceed here?

vedithal-amd:
> > > > @jbonnell-amd per my comment in ROCM-28653 we should not merge this PR since compute's python dependencies will clash with those of torch and the python application being profiled.
> > > > Folks would be better of creating a virtual environment for analysis mode based of our requirements.txt and a message in analysis mode instructing users to do so along with the path to requirements.txt should be enough
> > > 
> > > 
> > > If you don't want to merge this, maybe close or move to draft?
> > 
> > 
> > I think we can still merge this PR if we don't use `install_requires` but instead add a `rocm-profiler[compute-analyze]` extra which pull in compute's analyze deps. This way we can provide a way for users to setup a venv for analyze only.
> > I have implemented this change.
> 
> @jbonnell-amd @vedithal-amd how do you want to proceed here?

Bit busy with other priority items, will circle back here.

But high level - we really need `compute-analyze` extra in pip install rocm-profiler - so this pr is important and should not be closed

Basically, folks need an easy way to create python venv with compute analyze deps pulled in instead of running pip install /path/to/requirements

vedithal-amd:
Hi @ScottTodd and @marbre could you please review this PR.

The idea is to support `pip install --extra-index-url <rocm_index> rocm-profiler[compute-analyze]` which forwards the transitive closure of 44 python dependencies (and growing) defined in https://github.com/ROCm/rocm-systems/blob/develop/projects/rocprofiler-compute/requirements.txt - rocprofiler-compute documentation and runtime output will instruct folks to create a separate venv and run this command so that they have everything they need for `rocprofiler-compute analyze` mode.

Note that `rocprofiler-compute profile` mode requires no python dependencies, so it plays well with workload to be profiled.

However, `rocprofiler-compute analyze` wants its own venv to get the plethora of dependencies without disturbing their ROCm environment.

Currently, folks need to run `pip install -r /path/to/rocprofiler-compute/requirements.txt` to get the python dependencies for analysis. However, with this PR that step is no longer needed. It's expected that these dependencies have to come from PyPI not ROCm index hence `--extra-index`

build_tools/packaging/python/templates/rocm-profiler/setup.py
```diff
@@ -32,11 +32,39 @@ def _load_local_dist_info():
...
+        requirements.append(line)
+    return requirements
+
+
 dist_info = _load_local_dist_info()
 my_package = dist_info.ALL_PACKAGES["profiler"]
 packages = find_packages(where="./src")
 platform_package_name = my_package.get_py_package_name()
 packages.append(platform_package_name)
+extras_require = {"compute-analyze": _read_compute_requirements(platform_package_name)}
```
ScottTodd, line 67: We may want a dedicated set of tests that install this package extra and use the profiler analyze functionality.

See also:
* https://github.com/ROCm/TheRock/blob/main/TESTING.md#therock-feature-area-packaging
* https://github.com/ROCm/TheRock/blob/8438943f723884b46da7036527a49efb1456a520/.github/workflows/test_rocm_wheels.yml#L201-L227

ScottTodd (approved):
Thanks, this looks good to me now.

I checked the generated whl from the CI build:
* https://github.com/ROCm/TheRock/actions/runs/35145382838?pr=7002
* https://therock-ci-artifacts.s3.amazonaws.com/35145382838-linux/python/index.html
* https://therock-ci-artifacts.s3.amazonaws.com/35145382838-linux/python/rocm_profiler-10.2.0.dev0%2B6ee43d2fadf65113c3c23212e823c93b6b953281-py3-none-linux_x86_64.whl

Here's the metadata:

```
Metadata-Version: 2.4
Name: rocm-profiler
Version: 10.2.0.dev0+6ee43d2fadf65113c3c23212e823c93b6b953281
Summary: ROCm profiler applications (rocprofiler-systems and rocprofiler-compute)
Requires-Python: >=3.8
Provides-Extra: compute-analyze
Requires-Dist: astunparse==1.6.2; extra == "compute-analyze"
Requires-Dist: dash-bootstrap-components==2.0.4; extra == "compute-analyze"
Requires-Dist: dash-svg==0.0.12; extra == "compute-analyze"
Requires-Dist: dash==3.4.0; extra == "compute-analyze"
Requires-Dist: numpy==1.26.4; python_version < "3.13" and extra == "compute-analyze"
Requires-Dist: numpy==2.5.2; python_version >= "3.13" and extra == "compute-analyze"
Requires-Dist: pandas==2.2.3; extra == "compute-analyze"
Requires-Dist: plotext==5.3.2; extra == "compute-analyze"
Requires-Dist: plotille==5.0.0; extra == "compute-analyze"
Requires-Dist: pyyaml==6.0.3; extra == "compute-analyze"
Requires-Dist: sqlalchemy==2.0.46; extra == "compute-analyze"
Requires-Dist: tabulate==0.9.0; extra == "compute-analyze"
Requires-Dist: textual==7.3.0; extra == "compute-analyze"
Requires-Dist: textual_plotext==1.0.1; extra == "compute-analyze"
Dynamic: provides-extra
Dynamic: requires-python
Dynamic: summary
```

build_tools/packaging/python/templates/rocm-profiler/README.md
```diff
@@ -32,8 +32,13 @@ invoking a console script will result in a `FileNotFoundError`.
 
 ## Packaging Notes
 
-- No `install_requires` are declared.
-- Dependencies are managed by the `rocm` meta package.
+- `rocprof-compute analyze`'s Python dependencies are declared under the
+  `compute-analyze` extra, read at build time from the staged rocprofiler-compute
+  payload at `libexec/rocprofiler-compute/requirements.txt`.
+- Analyze mode should run from its own virtualenv:
+  `pip install --extra-index-url <rocm_index> rocm-profiler[compute-analyze]`.
```
marbre, line 39: Should we consider mirroring the deps you need?

vedithal-amd: Folks have been doing pip install our requirements.txt (which uses PyPI by defualt) separately just to use rocprof-compute analyze - this PR improves that user experience by automatically pulling in deps. That being said, it should be fine to only pull rocm[profiler] from ROCm index and other analyze deps. (we use lot of third party deps and their transitive closure is huge and this will grow) can be pulled from PyPI.

But I see your point, in a future PR we should think about distributing them as part of ROCm index.

build_tools/packaging/python/templates/rocm-profiler/setup.py
vedithal-amd: Good point, will look into what `rocm-sdk test` does for rocm[profiler] wheels

vedithal-amd: Added CI coverage for the extra rather than a unit test:

- `test_rocm_wheels.yml` now installs `rocm-profiler[compute-analyze]` into its own venv and runs `rocprof-compute analyze --help`. The launcher imports the analyze stack before argparse runs, so a missing dependency fails the step.
- Kept it in a separate venv because the existing `rocm[profiler]` venv deliberately installs without the extra.
- Added `--extra-index-url` to `setup_venv.py` for this. The extra's dependencies come from PyPI, and `--index-url` on its own replaces PyPI rather than adding to it.

vedithal-amd:
Pushed the CI coverage for the extra. The new steps are in the `Test ROCm wheels` jobs on Linux, after `Test rocm[profiler]`:

- `Install rocm-profiler[compute-analyze]`
- `Show installed packages (compute-analyze)`
- `Test rocprof-compute analyze`

Worth confirming these are green before merge. The install step resolves the analyze pins from PyPI, so a failure there is dependency resolution rather than anything wrong with the wheel.

vedithal-amd:
Seems like the `compute-analyze` extra pip install wheel tests have passed - see https://github.com/ROCm/TheRock/actions/runs/35400766420/job/105797810618?pr=7002

Merging this PR

PR #5795, opened 2026-06-11
[ci] Update workflow to use internal ubuntu container instead of public

nunnikri:
Replace ubuntu:24.04 with ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest in the multi-arch build native packages workflow and the container image lookup used by the test native packages install workflow.

## Motivation

We're hitting Docker rate limits o.O [Link per-ISA device files into rocm-sdk-devel by marbre · Pull Request #5778 · ROCm/TheRock](https://github.com/ROCm/TheRock/pull/5778)
```
Starting job container
/usr/bin/docker pull ubuntu:24.04
Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. [Docker Hub usage and limits](https://www.docker.com/increase-rate-limit)
Warning: Docker pull failed with exit code 1, back off 7.153 seconds before retry.
/usr/bin/docker pull ubuntu:24.04
Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. [Docker Hub usage and limits](https://www.docker.com/increase-rate-limit)
Warning: Docker pull failed with exit code 1, back off 9.266 seconds before retry.
/usr/bin/docker pull ubuntu:24.04
Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. [Docker Hub usage and limits](https://www.docker.com/increase-rate-limit)
Error: Docker pull failed with exit code 1

```
## Test Plan

CI testing pass can be treated as pass

PR #4932, opened 2026-04-29
install test support functions - gpg url, repo url update

arvindcheru:
## Motivation

This pull request significantly enhances the flexibility and clarity of the `get_url_repo_params.py` tool for Linux packaging, especially around how repository URLs and GPG key URLs are derived and output. The changes introduce support for multiple repository layouts, improve argument handling, and make the code more robust and user-friendly for automation scenarios.

These changes make the tool more powerful, easier to use in CI/CD workflows, and better aligned with the evolving ROCm packaging infrastructure.

## Technical Details

Key improvements and new features include:

**Repository URL and Layout Handling:**
- Added support for both "per_family" and "multi_arch" repository layouts, with normalization logic and new helper functions to generate URLs according to the selected layout. This enables the tool to generate URLs compatible with both legacy and new multi-architecture packaging schemes. [[1]](diffhunk://#diff-103884a38a5e372dab0d2471e73262aad4da14d0fd85ff8541f93067fbe281a0R46-R96) [[2]](diffhunk://#diff-103884a38a5e372dab0d2471e73262aad4da14d0fd85ff8541f93067fbe281a0L139-R384)
- Improved logic for inferring the native package type (`deb` or `rpm`) from the OS profile if not explicitly provided, reducing manual input requirements.

**GPG Key URL Derivation:**
- Enhanced GPG key URL derivation to support both per-family and multi-arch layouts, with new mapping tables and logic to generate correct URLs based on release type and layout.
- Refined which release types require signed GPG key URLs, now including "stable" and "prereleases" in addition to "prerelease" and "release".

**Command-line Interface and Usability:**
- Updated CLI usage and help text to reflect new argument options, including mutually exclusive arguments for `get-base-url`, and new options for `get-repo-url` such as `--layout`, `--from-url`, and optional `--native-package-type`. [[1]](diffhunk://#diff-103884a38a5e372dab0d2471e73262aad4da14d0fd85ff8541f93067fbe281a0L12-R30) [[2]](diffhunk://#diff-103884a38a5e372dab0d2471e73262aad4da14d0fd85ff8541f93067fbe281a0L268-R494) [[3]](diffhunk://#diff-103884a38a5e372dab0d2471e73262aad4da14d0fd85ff8541f93067fbe281a0L316-R544)
- The `get-repo-url` command now outputs both `repo_url` and `gpg_key_url` in a single call, streamlining automation and reducing the number of required invocations.

## Test Result

https://github.com/arvindcheru/TheRock/actions/runs/25092190403

[...]

PR #7004, opened 2026-07-30
fix(ci): run get_url_repo_params tests and correct stale assertions

lsudarsh-amd:
## Motivation

The unit tests in `build_tools/packaging/linux/tests/` never run in CI, because the directory is not listed in the pytest `testpaths` in `build_tools/pyproject.toml`. That is one of the two places #6927 points at as the reason some tests are not running continuously.

This change adds that directory to `testpaths`. It was opened adding a single file, because the other two did not pass. #7017 fixed one of them and a missing test dependency accounts for the other, so all four now run.

Contributes to #6927 and #6986.

## Technical Details

The change touches five files.

`build_tools/pyproject.toml` gains `packaging/linux/tests` in `testpaths` and `packaging/linux/tests/*` in the coverage `omit` list, which keeps test files out of the coverage report as the four existing entries do. Adding the directory rather than naming files is deliberate: a file list is what let `aggregate_metadata_test.py`, added by #6981, go uncollected after this PR was opened.

`requirements-test.txt` gains `jinja2>=3.0.0`. `build_package_test.py` imports `build_package.py`, which reaches `jinja2` through `deb_package.py` and `rpm_package.py`. It is already declared in `requirements.txt` and was missing only from the file `unit_tests.yml` installs, so the test was never broken; its environment just did not match the declared runtime one. That file is also read by `test_artifacts_structure.yml` and the shared `setup_test_environment` action, and `jinja2` is a pure-Python wheel.

`get_url_repo_params_test.py` needed the container-image string assertions #7017 deferred to this PR in its own module docstring: "Not covered here (P2 / separate PRs): ... #7004 container-image string assertions". The suite it left compares `get_container_image` against itself, in both the mapping test and the CLI smoke, so neither pins a value. Since eb7319cb93 (#5795), `get_container_image()` has returned `ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest` for `ubuntu*` and `debian*` profiles, and `.github/workflows/test_native_linux_packages_install.yml` feeds that value straight into its job's `container: image:`. Changing the function would change which container CI runs in, so the tests are what pin it. They now assert the exact image per profile, `rhel8` included, which the trimmed suite had dropped.

`get_url_repo_params.py` had the same stale value in the `help=` text for the `get-container-image` subparser, which advertised `ubuntu2404 -> ubuntu:24.04` while the function's own docstring was already current. The two now agree.

`native_linux_package_install_test.py` built the install prefix with `pathlib.Path`, which renders `\opt\rocm\core` on Windows. The prefix is a path on the target Linux filesystem passed to `find(1)`, so it now uses `PurePosixPath`, which is identical on Linux. Nothing was broken in practice, since the harness only runs there, but collecting the directory turned the `windows-2022` leg red on the first push and this is what it caught.

### Why the whole directory now

When this PR was opened, neither of the other two files passed. `build_package_test.py` never reached collection, for the dependency reason above. `native_linux_package_install_ut_test.py` had seven failures, two of which shelled out to real `sudo` because a mock did not intercept the write in `setup_deb_repository`. On a runner with passwordless sudo those two would have passed, turning a required check green while writing to its `/etc/apt/sources.list.d/`.

#7017 fixed that file, so the reason for leaving it out is gone, and #6927 names it specifically. It now passes 118 tests, and run as a non-root user with passwordless sudo, mirroring a GitHub runner, it writes nothing to `/etc/apt/sources.list.d`, `/etc/yum.repos.d`, `/etc/zypp/repos.d`, `/usr/share/keyrings` or `/etc/apt/keyrings`.

## Test Plan

Everything ran in a `python:3.12` container with `pip install -r requirements-test.txt`, which matches the Python version and dependency set that `unit_tests.yml` installs on its `ubuntu-24.04` leg. Each check ran twice, against a clean checkout of `main` and again with the change applied, so the two states could be compared directly.

The checks were the directory alone, the exact command CI runs from `build_tools/`, the `scan_tools` suite CI runs separately, and `pre-commit`. Three negative controls tested whether the change does anything: dropping the `testpaths` entry, removing `jinja2`, and breaking the `ubuntu` mapping. The `windows-2022` failure above was reproduced and re-checked locally by forcing the Windows path flavour through the module, which is the one part of that leg reproducible off Windows; everything else there is left to CI on this pull request.

## Test Result

The full `build_tools` suite, run in that container against both states:

| | Passed | Skipped | Failed |
|---|---|---|---|
| Clean checkout of `main` | 1401 | 33 | 0 |
| With this change | 1603 | 33 | 0 |

The difference is exactly the 202 tests in the directory. Nothing was removed, and no test outside it changed status. The skip count depends on the environment, since there is no `GITHUB_TOKEN` and a few cases are Windows only.

The negative controls:

| Control | Result |
|---|---|
| Drop the `testpaths` entry | 0 collected from the directory, against 202 with it |
| Remove `jinja2` | `build_package_test.py` fails collection |
| Break the `ubuntu` mapping | three of the new assertions fail, where the suite as #7017 left it passes 21/21 |

`scan_tools` is unaffected at 31 passed on both states, since it has its own `pyproject.toml`. `pre-commit` passes every hook without reformatting anything, and `get_url_repo_params.py --help` renders the corrected mapping.

## Note on scope

This is wider than the version that was approved. That version added one file to `testpaths`; this adds the directory, one line to `requirements-test.txt`, and a one-line path fix in `native_linux_package_install_test.py` that the `windows-2022` leg required. #7017 cleared the reason the other files were left out.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/native_linux_package_install_test.py
+++ b/build_tools/packaging/linux/native_linux_package_install_test.py
@@ -139,7 +139,7 @@ import subprocess
 import sys
 import traceback
 from argparse import ArgumentParser, Namespace
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 _SCRIPT_DIR = Path(__file__).resolve().parent
 if str(_SCRIPT_DIR) not in sys.path:
@@ -1073,7 +1073,10 @@ gpgcheck=0
         False if any offending path is found.
         """
         print("\nVerifying installed files are owned by root with safe permissions...")
-        install_prefix = str(Path(self.install_prefix))
+        # PurePosixPath, not Path: this is a path on the target Linux filesystem
+        # handed to find(1). Path follows the local flavour, so on Windows it
+        # renders "\opt\rocm\core" and the scan silently targets the wrong path.
+        install_prefix = str(PurePosixPath(self.install_prefix))
         try:
             result = subprocess.run(
                 [
--- a/build_tools/pyproject.toml
+++ b/build_tools/pyproject.toml
@@ -2,6 +2,7 @@
 testpaths = [
     "tests",
     "github_actions/tests",
+    "packaging/linux/tests",
     "packaging/python/tests",
     "third_party/s3_management/tests",
 ]
@@ -14,6 +15,7 @@ source = ["."]
 omit = [
     "tests/*",
     "github_actions/tests/*",
+    "packaging/linux/tests/*",
     "packaging/python/tests/*",
     "third_party/s3_management/tests/*",
     "*/__pycache__/*",
--- a/build_tools/packaging/linux/get_url_repo_params.py
+++ b/build_tools/packaging/linux/get_url_repo_params.py
@@ -742,7 +742,10 @@ def main(argv: list[str] | None = None) -> int:
     # get-container-image: get container image for an OS profile
     p_img = subparsers.add_parser(
         "get-container-image",
-        help="Get container image for a given OS profile (e.g. ubuntu2404 -> ubuntu:24.04).",
+        help=(
+            "Get container image for a given OS profile "
+            "(e.g. ubuntu2404 -> ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest)."
+        ),
     )
     p_img.add_argument(
         "--os-profile",
--- a/build_tools/packaging/linux/tests/get_url_repo_params_test.py
+++ b/build_tools/packaging/linux/tests/get_url_repo_params_test.py
@@ -19,7 +19,7 @@ Coverage (trimmed suite, table-driven via ``subTest``):
   - One happy-path CLI smoke per remaining subcommand
 
 Not covered here (P2 / separate PRs): ``upload_package_repo._package_install_url``,
-#7004 container-image string assertions, full CLI error-matrix.
+full CLI error-matrix.
 
 Prerequisites:
 
@@ -331,20 +331,25 @@ class ExtractGfxArchTest(unittest.TestCase):
 
 
 class GetContainerImageTest(unittest.TestCase):
-    """Tests for ``get_container_image()`` (asserts against implementation output)."""
+    """Tests for ``get_container_image()``."""
+
+    # test_native_linux_packages_install.yml feeds this value straight into its
+    # container image, so each profile is pinned to the exact string rather than
+    # compared against another call to the function under test.
+    _EXPECTED_IMAGES = [
+        ("ubuntu2404", "ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest"),
+        ("debian12", "ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest"),
+        ("sles16", "registry.suse.com/bci/bci-base:16.0"),
+        ("rhel8", "registry.access.redhat.com/ubi8/ubi:8.10"),
+        ("rhel10", "registry.access.redhat.com/ubi10/ubi:10.1"),
+    ]
 
     def test_profile_mapping(self):
-        # Ubuntu/debian share one image; sles/rhel use distinct registry images.
-        ubuntu = get_url_repo_params.get_container_image("ubuntu2404")
-        self.assertEqual(get_url_repo_params.get_container_image("debian12"), ubuntu)
-        self.assertEqual(
-            get_url_repo_params.get_container_image("sles16"),
-            "registry.suse.com/bci/bci-base:16.0",
-        )
-        self.assertEqual(
-            get_url_repo_params.get_container_image("rhel10"),
-            "registry.access.redhat.com/ubi10/ubi:10.1",
-        )
+        for os_profile, expected in self._EXPECTED_IMAGES:
+            with self.subTest(os_profile=os_profile):
+                self.assertEqual(
+                    get_url_repo_params.get_container_image(os_profile), expected
+                )
 
 
 class MainSubcommandsTest(unittest.TestCase):
@@ -408,12 +413,13 @@ class MainSubcommandsTest(unittest.TestCase):
         self.assertIn("gfx_arch=gfx94x,gfx1100", output)
 
     def test_get_container_image_cli(self):
-        expected = get_url_repo_params.get_container_image("ubuntu2404")
         code, output = _run_main_with_output(
             ["get-container-image", "--os-profile", "ubuntu2404"]
         )
         self.assertEqual(code, 0)
-        self.assertIn(f"container_image={expected}", output)
+        self.assertIn(
+            "container_image=ghcr.io/rocm/no_rocm_image_ubuntu24_04:latest", output
+        )
 
 
 if __name__ == "__main__":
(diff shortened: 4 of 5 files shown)


build_tools/pyproject.toml
```diff
@@ -2,6 +2,7 @@
 testpaths = [
     "tests",
     "github_actions/tests",
+    "packaging/linux/tests/get_url_repo_params_test.py",
```
ScottTodd: This is okay for now but I'd rather switch to a directory pattern once possible. Ideally we should be able to run `pytest build_tools/`. The details in the PR description about the tests that currently fail at collection time are useful, thanks.

ScottTodd (approved):

arvindcheru:
Hello @lsudarsh-amd, Need a few changes to test code, current unit test there are some failures.  I am updating a PR for this, I would request hold merge this PR till the changes are in. (https://github.com/ROCm/TheRock/pull/4932 is one under rework now) Thank you.

build_tools/pyproject.toml
lsudarsh-amd: Thanks for the review. Agreed, I'd like this to get to the point where we are running these tests against the directory itself. For this directory specifically it's two dependencies in `requirements-test.txt` plus the remaining failures in `native_linux_package_install_ut_test.py`, and then this single-file entry can become the directory.

lsudarsh-amd:
> Hello @lsudarsh-amd, Need a few changes to test code, current unit test there are some failures. I am updating a PR for this, I would request hold merge this PR till the changes are in. (#4932 is one under rework now) Thank you.

@arvindcheru No problem, I can hold the merge. Thanks for letting me know. This PR only adds `get_url_repo_params_test.py` to `testpaths` rather than the whole directory, so the overlap with your rework should be small, but I'll rebase and adjust as needed once your changes land.

Could you please link the replacement PR when it's up, or give me a rough sense of timing? I have follow on work in `get_url_repo_params.py` and I'd rather build on your version than around it.

arvindcheru:
> > Hello @lsudarsh-amd, Need a few changes to test code, current unit test there are some failures. I am updating a PR for this, I would request hold merge this PR till the changes are in. (#4932 is one under rework now) Thank you.
> 
> @arvindcheru No problem, I can hold the merge. Thanks for letting me know. This PR only adds `get_url_repo_params_test.py` to `testpaths` rather than the whole directory, so the overlap with your rework should be small, but I'll rebase and adjust as needed once your changes land.
> 
> Could you please link the replacement PR when it's up, or give me a rough sense of timing? I have follow on work in `get_url_repo_params.py` and I'd rather build on your version than around it.

Hello @lsudarsh-amd, as discussed proposed the updated PR (https://github.com/ROCm/TheRock/pull/7017), and the changes I was proposed to main only

arvindcheru:
> > Hello @lsudarsh-amd, Need a few changes to test code, current unit test there are some failures. I am updating a PR for this, I would request hold merge this PR till the changes are in. (#4932 is one under rework now) Thank you.
> 
> @arvindcheru No problem, I can hold the merge. Thanks for letting me know. This PR only adds `get_url_repo_params_test.py` to `testpaths` rather than the whole directory, so the overlap with your rework should be small, but I'll rebase and adjust as needed once your changes land.
> 
> Could you please link the replacement PR when it's up, or give me a rough sense of timing? I have follow on work in `get_url_repo_params.py` and I'd rather build on your version than around it.

Sorry @lsudarsh-amd  I mistakenly update the PR, I was about to comment on rebasing before merging.

lsudarsh-amd:
Rebased onto `b070c6f388`, plus a follow-up commit that fixes the `windows-2022` leg.

@arvindcheru thanks for the heads up on timing. One conflict, in `get_url_repo_params_test.py`, resolved by keeping your rewritten suite and adding back the container-image string assertions your module docstring deferred to #7004. Collecting the directory also turned `windows-2022` red on one test from #6974: `verify_installed_file_security()` built the `find(1)` prefix with `pathlib.Path`, which renders `\opt\rocm\core` on Windows. It now uses `PurePosixPath`, which is identical on Linux.

@ScottTodd the scope grew past your approval, so this needs another look. The PR now adds the whole directory instead of one file, one line in `requirements-test.txt`, and that path fix. #7017 cleared the reason the other files were excluded. The test matrix and negative controls are in the description. Happy to split any of it into a follow-up if you would rather review it separately.

build_tools/packaging/linux/native_linux_package_install_test.py
ScottTodd, line 1090: We may want to just skip this test case (and maybe others) on Windows. The test is exercising a Unix file system and is using the `find` command (which has a Windows version in system32 that is different from unix, so it probably can't handle these arguments: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/find)

lsudarsh-amd: One thing worth clarifying here is, `find` never actually runs. The 18 tests in this class patch `native_linux_package_install_test.subprocess.run`, so the test only asserts the argv that would be passed. That is why it caught the bug rather than erroring on Windows' find.

With that said, I agree with the larger point you are making here. I'd recommend keeping the `PurePosixPath` change, since the prefix is a path on the target filesystem and should not depend on the local flavour. I can add a Windows skip as a follow up PR.

ScottTodd: This is very strange test case design. 40 lines of CLI flags to a command that isn't run? Then this below?
```python
        except OSError as e:
            # find not available on this host; fall back to a Python walk so the
            # check still runs rather than being silently skipped.
            print(f" [WARN] 'find' unavailable ({e}); falling back to Python scan")
            return self._verify_installed_file_security_python(Path(install_prefix))
```

Tests should be simple and free from logic - they should exercise the real code paths as much as possible and use mocks in narrow ways.

lsudarsh-amd: Yeah agreed, the current setup in this file is a bit awkward. For instance, there are two implementations of the same check, `find` plus a Python fallback. The test as a result pins argv for one of them. So it does make sense to invest some time in improving the design of the test cases in this file. 

Since this PR only changes a line in this file and is more about `testpaths`, I would keep the test design work separate. I will file an issue for it so that it doesn't get lost.

arvindcheru: Thank you @ScottTodd and @lsudarsh-amd for the findings/review comments — the Windows leg and unit-test design points were helpful.

@lsudarsh-amd’s clarification stands: The real install security check (Unix find against a Linux install tree) only runs on Linux install hosts/containers, which is where native package verification actually belongs, and on windows-2022 these tests mock subprocess.run

On the unit-test cleanup Scott raised (and the dual find/Python path Lalit noted): I can support follow-up work to simplify that coverage — I have exploratory test-side changes and can help in unblocking this.

lsudarsh-amd: @ScottTodd @arvindcheru FYI I've created the issue to track the test case rework here: #7128.

arvindcheru (approved):
LGTM

UT refactor exploratory changes are started on users/acheruva/test_unit_test_organize @ ba32cd9 (Adding helpers and simplified find-security unit tests for addressing the follow up comments) — not blocking this PR.

PR #7010, opened 2026-07-30
fix(packaging): Include unversioned `libamd_smi.so` in `rocm-sdk-core` package

nunnikri:
## Summary

Materialize the unversioned `libamd_smi.so` in `rocm-sdk-core` so that
`dlopen("libamd_smi.so")` works at runtime without requiring `rocm-sdk-devel`
or `rocm-sdk init`.

JIRA ID : https://amd-hub.atlassian.net/browse/ROCM-27833

## Motivation

PyTorch's `IntraNodeComm::getNvlMesh` calls `dlopen("libamd_smi.so")` (unversioned
name) during NVLink mesh detection for intra-node all-reduce. This caused two
distributed NCCL tests to hang and timeout after 300 seconds:

The unversioned `libamd_smi.so` symlink only existed in `_rocm_sdk_devel/lib/`
(after `rocm-sdk init`), but `LD_LIBRARY_PATH` covers `_rocm_sdk_core/lib/` —
not the devel tree. So `dlopen` failed at runtime even when `rocm-sdk-devel` was
installed and initialized.

## Technical Details

`_populate_runtime_symlink` in `py_packaging.py` handles symlinks from the ROCm
install tree when building wheels (which cannot contain symlinks). It materializes
a symlink as a real file only when the symlink name equals the SONAME. The
unversioned `.so` (e.g. `libamd_smi.so`) has a different name from the SONAME
(`libamd_smi.so.26`), so it was silently deferred to the devel expansion path.

Fix: add a `_RUNTIME_DLOPEN_UNVERSIONED_LIBS` allowlist. Libraries in this set
have their unversioned `.so` materialized as a real file copy in the runtime
package, placing them under `_rocm_sdk_core/lib/` and making them reachable via
`LD_LIBRARY_PATH` without devel.

The tradeoff is a duplicate copy of the library bytes in `rocm-sdk-core`
(`libamd_smi.so` and `libamd_smi.so.26` contain the same bytes). 

**File changed:** `build_tools/_therock_utils/py_packaging.py`

## Test Plan

1. Build `rocm-sdk-core` wheel with this fix
2. Install wheel without `rocm-sdk-devel`:
   ```bash
       "rocm[libraries,device-gfx942]"

## Test Results
Before fix — only versioned file in core, unversioned only after devel init:


$ find . -name "libamd_smi.so*"
./_rocm_sdk_devel/share/amd_smi/amdsmi/libamd_smi.so
./_rocm_sdk_devel/lib/libamd_smi.so.26.5.0
./_rocm_sdk_devel/lib/libamd_smi.so
./_rocm_sdk_devel/lib/libamd_smi.so.26
./_rocm_sdk_core/lib/libamd_smi.so.26
After fix — unversioned file present in core without devel:


$ find . -name "libamd_smi.so*"
./_rocm_sdk_core/lib/libamd_smi.so        ← NEW: dlopen("libamd_smi.so") now resolves
./_rocm_sdk_core/lib/libamd_smi.so.26

### DIFF
--- a/build_tools/_therock_utils/py_packaging.py
+++ b/build_tools/_therock_utils/py_packaging.py
@@ -38,6 +38,15 @@ assert DIST_INFO_PATH.exists()
 
 ENABLED_VLOG_LEVEL: int = 5
 
+# Unversioned .so symlinks that must be present in the runtime package (not
+# just in devel) because they are dlopen'd at runtime by their unversioned name.
+# Example: PyTorch's IntraNodeComm::getNvlMesh calls dlopen("libamd_smi.so").
+_RUNTIME_DLOPEN_UNVERSIONED_LIBS: frozenset[str] = frozenset(
+    [
+        "libamd_smi.so",
+    ]
+)
+
 
 def log(*args, vlog: int = 0, **kwargs):
     if vlog > ENABLED_VLOG_LEVEL:
@@ -450,6 +459,11 @@ class PopulatedDistPackage:
         # Case 2: Shared library.
         if file_type == "so" and (soname := get_soname(link_target)):
             if soname == src_entry.name:
+                self._populate_file(relpath, dest_path, src_entry, resolve_src=True)
+            elif src_entry.name in _RUNTIME_DLOPEN_UNVERSIONED_LIBS:
+                # Some libraries are dlopen'd at runtime using their unversioned
+                # name (e.g. dlopen("libamd_smi.so")). Materialize the unversioned
+                # symlink in the runtime package so devel is not required for this.
                 self._populate_file(relpath, dest_path, src_entry, resolve_src=True)
             else:
                 self.files.soname_aliases[relpath] = soname


build_tools/_therock_utils/py_packaging.py
```diff
@@ -38,6 +38,15 @@
...
 ENABLED_VLOG_LEVEL: int = 5
 
+# Unversioned .so symlinks that must be present in the runtime package (not
+# just in devel) because they are dlopen'd at runtime by their unversioned name.
+# Example: PyTorch's IntraNodeComm::getNvlMesh calls dlopen("libamd_smi.so").
+_RUNTIME_DLOPEN_UNVERSIONED_LIBS: frozenset[str] = frozenset(
+    [
+        "libamd_smi.so",
+    ]
+)
```
ScottTodd, line 48: I don't think we should distribute unversioned .so files outside of devel packages or encourage downstream users to depend on them.

Code like
```c++
    const char* rocm = std::getenv("ROCM_PATH");
    std::string path =
        rocm ? std::string(rocm) + "/lib/libamd_smi.so" : "libamd_smi.so";
```

in pytorch from https://github.com/pytorch/pytorch/commit/51e761b0e952b21b808f6faf9efbb76e51747d3e is brittle and depends on global environment variables. Changes like https://github.com/pytorch/pytorch/pull/188454 are also wrong, should follow https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#performing-initialization or _maybe_ use RPATH. A versioned dlopen like `handle = dlopen("libnvshmem_host.so.3", RTLD_LAZY);` (https://github.com/pytorch/pytorch/blob/ca7df6f0b43ebae7bad59cae48b820680f0a0152/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu#L57-L64) may also be okay, at least compared to the unversioned open.

ScottTodd (changes requested):

build_tools/_therock_utils/py_packaging.py
ScottTodd: This also conflicts with https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#general-design

> Runtime packages: Most packages are runtime packages. These are wheel files that are ready to be expanded directly into a site-lib directory. They contain only the minimal set of files needed to run. Critically, they do not contain symlinks (instead, **only SONAME libraries are included**, and any executable symlinks are emitted by dynamically compiling an executable that can execle a relative binary).

nunnikri: Thanks @ScottTodd for the feedback. I will close this PR and will push for the proper usage in pytorch side

raramakr: cc @marifamd , FYI

nunnikri:
Closing in favor of https://github.com/ROCm/TheRock/pull/7057 and follow up pytorch changes.

Issue #6965, opened 2026-07-29
[Issue] PyTorch nightly builds are missing `_rocm_init.py`. Windows fails to load `c10_hip.dll`, Linux fails to load `libMIOpen.so.1`


TriveniTadapaneni:
## Context

  - **Workflow:** Multi-Arch Release Windows PyTorch Wheels (`ROCm/rockrel`)
  - **Failing stage:** post-build `import torch` sanity check in `build_prod_wheels.py`, before the torchaudio build step even starts
  - **PyTorch version:** `nightly`
  - **Python versions affected:** 3.10, 3.11, 3.12, 3.13, 3.14 
  - **Example failing run:** https://github.com/ROCm/rockrel/actions/runs/30427147064
    - `Build | py 3.10 | torch nightly`: https://github.com/ROCm/rockrel/actions/runs/30427147064/job/90496057248
    - `Build | py 3.11 | torch nightly`: https://github.com/ROCm/rockrel/actions/runs/30427147064/job/90496057049
    - `Build | py 3.12 | torch nightly`: https://github.com/ROCm/rockrel/actions/runs/30427147064/job/90496057153
    - `Build | py 3.13 | torch nightly`: https://github.com/ROCm/rockrel/actions/runs/30427147064/job/90496057465
    - `Build | py 3.14 | torch nightly`: https://github.com/ROCm/rockrel/actions/runs/30427147064/job/90496057382

  ## Exact Error

  The main `torch` wheel builds and installs successfully; the failure happens at the sanity-check step right after install, before
  torchaudio is ever built:

  ```text
  +++ Installing built torch:
  ++ Exec [...] 'python.exe' -m pip install 'B:\src\pytorch\dist\torch-2.14.0a0+rocm10.0.0a20260729-cp311-cp311-win_amd64.whl'
  +++ Sanity checking installed torch (unavailable is okay on CPU machines):
  ++ Capture [...] 'python.exe' -c 'import torch; print(torch.cuda.is_available())'
  Error capturing output: Command [...] returned non-zero exit status 1.
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "C:\home\runner\_work\_tool\Python\3.11.9\x64\Lib\site-packages\torch\__init__.py", line 309, in <module>
      _load_dll_libraries()
    File "C:\home\runner\_work\_tool\Python\3.11.9\x64\Lib\site-packages\torch\__init__.py", line 305, in _load_dll_libraries
      raise err
[...]

PR #7027, opened 2026-07-31
[Windows] Fix torch_hip.dll import: preload rocblas/rocsolver + ship aotriton_v2.dll

ethanwee1:
ISSUE ID: https://github.com/ROCm/TheRock/issues/6965

## Problem

Windows ROCm PyTorch wheels fail the `import torch` sanity check with:

```
OSError: [WinError 126] The specified module could not be found.
Error loading ".../torch/lib/torch_hip.dll" or one of its dependencies.
```

A `dumpbin` dependency-tree walk of `torch_hip.dll` shows several Windows-side gaps in how its runtime dependencies reach the process at load time. On Linux these all resolve via RPATH; Windows has no equivalent, so each dependency must either be preloaded into the process namespace or shipped in `torch/lib`:

1. `torch_hip.dll` links `rocblas.dll` and `rocsolver.dll` **directly**. The generated `torch/_rocm_init.py` preloads every *other* ROCm dependency, but `rocblas`/`rocsolver` were absent from the Windows preload list and could not be added because they had **no registered `rocm_sdk` shortname** in `_dist_info.py`.
2. When Flash Attention is enabled, `torch_hip.dll` also links `aotriton_v2.dll` **directly**. CMake installs it into `torch/lib`, but scikit-build-core drops it from the wheel (same gitignore-driven class of bug as `_rocm_init.py`), so it is missing at import time.
3. `aotriton_v2.dll` in turn links `liblzma.dll` (used to decompress its GPU kernels). On Windows PyTorch builds this from its bundled xz into `build/xz-install/bin`, and it is never copied into `torch/lib`, so once `aotriton_v2.dll` is shipped it becomes the next `NOT FOUND` leaf.

## Changes

1. `build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py`: register `rocblas` and `rocsolver` as `rocm_sdk` `LibraryEntry` shortnames.
2. `external-builds/pytorch/build_prod_wheels.py`:
   - add `rocblas` and `rocsolver` to `WINDOWS_LIBRARY_PRELOADS`, ordered after `amdhip64` and with `rocblas` before `rocsolver` (which depends on it), so `rocm_sdk.initialize_process()` preloads them (`RTLD_GLOBAL`) before torch loads `torch_hip.dll`;
   - force-include `torch/lib/aotriton_v2.dll` into the Windows wheel when Flash Attention is enabled;
   - force-include `build/xz-install/bin/liblzma.dll` into `torch/lib` when Flash Attention is enabled, next to `aotriton_v2.dll`.

On Linux, `rocblas`/`rocsolver` resolve via RPATH and `aotriton`/`liblzma` are packaged already, so `LINUX_LIBRARY_PRELOADS` and Linux packaging are unchanged.

## Companion PR — both are required for Windows `import torch`

The preload half of this PR only takes effect if `torch/_rocm_init.py` is present in the wheel and executes. That inclusion is provided by the upstream PyTorch change:

- **pytorch/pytorch#191632** — ships `torch/_rocm_init.py` in the wheel (scikit-build-core drops it otherwise). Equivalent to the CMake `install(FILES ...)` alternative in pytorch/pytorch#191625.

Neither PR is sufficient alone on Windows: #191632 gets `_rocm_init.py` into the wheel; this PR makes the initializer preload `rocblas`/`rocsolver` and ships the missing `aotriton_v2.dll` + `liblzma.dll`. Both must be present for `import torch` to succeed.

## Validation

Validated iteratively on the TheRock multi-arch Windows PyTorch wheels workflow (gfx1100, py3.12, torch nightly), with the nightly checkout path building pytorch/pytorch#191632 so `_rocm_init.py` is in the wheel. A `dumpbin` dependency-tree diagnostic of `torch_hip.dll` was run each iteration to identify the next unresolved leaf:

- [TheRock Actions run 30944161464](https://github.com/ROCm/TheRock/actions/runs/30944161464) — after `rocblas`/`rocsolver` are preloaded, the diagnostic showed `aotriton_v2.dll` as the single remaining `NOT FOUND` dependency.
- [TheRock Actions run 30952311829](https://github.com/ROCm/TheRock/actions/runs/30952311829) — with the `aotriton_v2.dll` force-include added (plus a shim simulating a post-merge `rocm_sdk` nightly carrying the `_dist_info.py` registration). This confirmed `aotriton_v2.dll` is now shipped (`-> torch/lib`), but **import still failed** and the diagnostic revealed `liblzma.dll` as the next and only remaining `NOT FOUND` leaf (`aotriton_v2.dll` links it).
- Confirming run with the `liblzma.dll` force-include: **pending** (to verify the full dependency tree resolves and `import torch` succeeds).

Note: because the `rocblas`/`rocsolver` shortname registration lives in the published `rocm_sdk` nightly (installed via `--index-url`), a fully clean end-to-end validation of this PR without the shim requires a ROCm nightly built after this `_dist_info.py` change lands (an older `rocm_sdk` would raise `Unknown rocm library 'rocblas'`).

### DIFF
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -190,6 +190,12 @@ LINUX_LIBRARY_PRELOADS = [
 WINDOWS_LIBRARY_PRELOADS = [
     "amd_comgr",
     "amdhip64",
+    # rocblas and rocsolver are direct load-time deps of torch_hip.dll on
+    # Windows (verified via dumpbin). Preload them (after amdhip64, and rocblas
+    # before rocsolver which depends on it) so they are resolved in-namespace
+    # before torch loads torch_hip.dll. On Linux these resolve via RPATH.
+    "rocblas",
+    "rocsolver",
     "hiprtc",
     "hipblas",
     "hipfft",
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -1206,6 +1206,16 @@ def do_build_pytorch(
                 "-Cwheel.force-include.torch/lib/libomp140.x86_64.dll="
                 "torch/lib/libomp140.x86_64.dll"
             )
+            # aotriton_v2.dll is a direct load-time dependency of torch_hip.dll
+            # when Flash Attention is enabled. CMake installs it into
+            # torch/lib, but scikit-build-core drops it from the wheel (same
+            # gitignore-driven class of bug as _rocm_init.py), so torch_hip.dll
+            # fails to load with WinError 126. Force-include it when present.
+            if use_flash_attention == "1":
+                build_command.append(
+                    "-Cwheel.force-include.torch/lib/aotriton_v2.dll="
+                    "torch/lib/aotriton_v2.dll"
+                )
     if is_windows:
         # The PyPI `ninja` package is unusable on Windows: 1.11.1 loops without
         # making progress and 1.13.0 has an MSVC link.exe RSP-file regression
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
@@ -284,13 +284,13 @@ LibraryEntry("rccl", "libraries", "librccl.so*", "")
 LibraryEntry("miopen", "libraries", "libMIOpen.so*", "MIOpen*.dll")
 LibraryEntry("origami", "libraries", "liborigami.so*", "origami*.dll")
 LibraryEntry("hipdnn", "libraries", "libhipdnn_backend.so*", "hipdnn_backend*.dll")
+LibraryEntry("rocblas", "libraries", "librocblas.so*", "rocblas*.dll")
+LibraryEntry("rocsolver", "libraries", "librocsolver.so*", "rocsolver*.dll")
 
 # Others we may want:
 # hiprtc-builtins
-# rocblas
 # rocfft
 # rocrand
-# rocsolver
 # rocsparse
 
 # Overall ROCM package version.


external-builds/pytorch/build_prod_wheels.py
```diff
@@ -420,6 +369,37 @@ def initialize():
...
+                # Keep the handle alive for the lifetime of the process so the
+                # directory remains on the search path when torch later loads
+                # torch_hip.dll.
+                global _ROCM_SDK_DLL_DIRECTORY
+                _ROCM_SDK_DLL_DIRECTORY = _os.add_dll_directory(str(_rocm_bin_dir))
+            """
+        ),
+        "    ",
+    )
+    return init_block + windows_dll_dir_block
```
ScottTodd: `rocm_sdk.initialize_process` is designed to handle this: https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#performing-initialization . I don't think we should add this alternate init behavior to TheRock.

If our init code is no longer working in pytorch we should fix that. My impressions from https://github.com/pytorch/pytorch/pull/188454 and https://github.com/pytorch/pytorch/pull/190276 was that:
* We would continue using `_rocm_init.py` for both Linux and Windows for ROCm builds
* PyTorch upstream would use RPATH for Linux and solve for Windows later (maybe adopting `_rocm_init.py` in some form)

ScottTodd:
fyi I also have a local pytorch build running so I'll be able to test shortly

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -190,6 +190,12 @@
 WINDOWS_LIBRARY_PRELOADS = [
     "amd_comgr",
     "amdhip64",
+    # rocblas and rocsolver are direct load-time deps of torch_hip.dll on
+    # Windows (verified via dumpbin). Preload them (after amdhip64, and rocblas
+    # before rocsolver which depends on it) so they are resolved in-namespace
+    # before torch loads torch_hip.dll. On Linux these resolve via RPATH.
+    "rocblas",
+    "rocsolver",
```
ScottTodd, line 198: > Root cause, diagnosed via CI debug run [30574604058](https://github.com/ROCm/TheRock/actions/runs/30574604058): `dumpbin /dependents torch_hip.dll` shows it links `rocblas.dll` and `rocsolver.dll` **directly**. The generated `torch/_rocm_init.py` already preloads (via `rocm_sdk.initialize_process`) every _other_ ROCm dependency of `torch_hip.dll` (`hipblas`, `hipfft`, `hipsparse`, `hipsolver`, `libhipblaslt`, `MIOpen`, `amdhip64`), but `rocblas`/`rocsolver` were missing from the Windows preload list — and could not simply be added, because they had **no registered `rocm_sdk` shortname** (they were only commented-out "Others we may want" entries in `_dist_info.py`).

I'm not confident this fixes the loads. My local build without extra preloads _succeeded_, including `import torch` sanity checks. I think this is a red herring and libomp is missing:

Image

Let's debug around
* https://github.com/ROCm/TheRock/blob/36024af03ae158c332383a66dec3f307d33a36cd/external-builds/pytorch/build_prod_wheels.py#L1198-L1202 
* https://github.com/ROCm/TheRock/blob/36024af03ae158c332383a66dec3f307d33a36cd/external-builds/pytorch/build_prod_wheels.py#L1100-L1103

(hipblas imports rocblas and rocsolver, so these explicit preloads shouldn't be needed as far as I can tell)

ScottTodd: Or actually, `rocm-openblas.dll`?
image

ScottTodd: Hmmm, `rocm-openblas` is included in `_rocm_init.py` that I have:
```python
def initialize():
    import rocm_sdk
    rocm_sdk.initialize_process(
        preload_shortnames=['amd_comgr', 'amdhip64', 'hiprtc', 'hipblas', 'hipfft', 'hiprand', 'hipsparse', 'hipsparselt', 'hipsolver', 'hipblaslt', 'miopen', 'hipdnn', 'rocm-openblas'],
        check_version='10.1.0a20260731')
```

ScottTodd: I didn't get any further debugging today. My next step was going to be to edit build_prod_wheels.py and the build workflow to upload torch wheels even if the sanity check import fails, so we can download them and inspect what the CI is doing differently. Could also add more logging (e.g. list files in the `lib/` dir, cat the `_torch_init.py` file contents, etc.).

external-builds/pytorch/build_prod_wheels.py
```diff
@@ -1206,6 +1206,16 @@ def do_build_pytorch(
...
+            # aotriton_v2.dll is a direct load-time dependency of torch_hip.dll
+            # when Flash Attention is enabled. CMake installs it into
+            # torch/lib, but scikit-build-core drops it from the wheel (same
+            # gitignore-driven class of bug as _rocm_init.py), so torch_hip.dll
+            # fails to load with WinError 126. Force-include it when present.
+            if use_flash_attention == "1":
+                build_command.append(
+                    "-Cwheel.force-include.torch/lib/aotriton_v2.dll="
+                    "torch/lib/aotriton_v2.dll"
+                )
```
ScottTodd, line 1218: Oof, we should flip the default here: https://github.com/ROCm/TheRock/blob/02bb54063b1e398ab07f2a1d8c4c3eb48023a2ff/external-builds/pytorch/build_prod_wheels.py#L1519-L1530

my local build had flash attention off

ScottTodd: Specifically here: https://github.com/ROCm/TheRock/blob/02bb54063b1e398ab07f2a1d8c4c3eb48023a2ff/external-builds/pytorch/build_prod_wheels.py#L1105-L1115

should match https://github.com/ROCm/TheRock/blob/02bb54063b1e398ab07f2a1d8c4c3eb48023a2ff/external-builds/pytorch/build_prod_wheels.py#L1050-L1061

I think now.

ScottTodd: I'm testing that and the `aotriton_v2.dll` change myself now, fyi.

ethanwee1: https://github.com/ROCm/TheRock/actions/runs/30952311829
I see this also liblzma.dll missing also

ScottTodd: Okay, that file exists in `torch/lib/` alongside `libomp140.x86_64.dll` and `aotriton_v2.dll` in my local build so could just be missing another copy into the packages

ethanwee1: pushed the liblzma.dll force-include to the test branch and kicked off a confirming build with all three fixes + the dumpbin diagnostic: https://github.com/ROCm/TheRock/actions/runs/31024048213

ScottTodd: That works for me
image

```
++ Exec [C:\Users\NOD-SH~1\AppData\Local\Temp]$ 'D:\projects\TheRock\external-builds\pytorch\.venv\Scripts\python.exe' -m pip install 'D:\b\pytorch\dist\torch-2.14.0a0+rocm10.1.0a20260731-cp313-cp313-win_amd64.whl'
+++ Sanity checking installed torch (unavailable is okay on CPU machines):
++ Capture [C:\Users\NOD-SH~1\AppData\Local\Temp]$ 'D:\projects\TheRock\external-builds\pytorch\.venv\Scripts\python.exe' -c 'import torch; print(torch.cuda.is_available())'
Sanity check output:
True
```

I didn't need the `rocblas` or `rocsolver` loads but they seem harmless to include

ScottTodd:
I'm testing https://github.com/ROCm/TheRock/pull/7131 which enables aotriton by default on windows and includes the aotriton fix (among a few other cleanups)

ethanwee1:
Closing in favor of https://github.com/ROCm/TheRock/pull/7131

PR #7029, opened 2026-07-31
feat(jax): Avoid rebuilding LLVM in JAX manylinux wheel builds

erman-gurses:
Progress on: #6388

## Summary

Avoid building LLVM 18 from source as part of every JAX manylinux wheel build.

JAX releases use two different host compiler configurations today:

- JAX 0.10.0 and 0.10.1 use `clang_local` and expect
  `/usr/lib/llvm-18/bin/clang`.
- JAX 0.10.2 and 0.11.0 use `rocm_clang_hermetic` for
  `rocm_release_wheel`.

This change preserves those existing JAX/XLA toolchain configurations while
removing the expensive LLVM source build from the manylinux image.

For JAX 0.10.0 and 0.10.1, a pinned prebuilt LLVM 18.1.8 binary is installed
at `/usr/lib/llvm-18`.

For JAX 0.10.2 and 0.11.0, LLVM is not installed in the Docker image because
the release-wheel build already uses JAX/XLA's hermetic LLVM toolchain.

## Changes

- Add `prepare_jax_manylinux_dockerfile.py` to prepare the JAX manylinux
  Dockerfile based on the requested JAX release.
- For JAX 0.10.0 and 0.10.1:
  - Replace the LLVM 18 source build with the LLVM 18.1.8 prebuilt binary.
  - Install it at `/usr/lib/llvm-18`, matching the path expected by
    `clang_local`.
  - Preserve the existing GCC toolchain configuration using
    `gcc-toolset-14`.
- For JAX 0.10.2 and 0.11.0:
  - Remove the LLVM 18 source-build block entirely.
  - Keep the existing `rocm_clang_hermetic` configuration unchanged.
- Keep the normal JAX build flow:
  - `--config=rocm_release_wheel`
  - `ROCM_PATH` provided by TheRock.
- Remove TheRock-side JAX compiler configuration overrides.

## Why

The JAX manylinux Dockerfile currently builds LLVM 18.1.8 from source during
each image build.

In recent CI runs, this step alone took roughly 8-11 minutes per job.

For JAX 0.10.2 and 0.11.0, this LLVM installation is unnecessary because
`rocm_release_wheel` already selects `rocm_clang_hermetic`, which obtains the
host LLVM toolchain as a prebuilt hermetic binary.

JAX 0.10.0 and 0.10.1 still use `clang_local` and explicitly expect:

```text
/usr/lib/llvm-18/bin/clang
```
## Test: 
https://github.com/ROCm/TheRock/actions/runs/31745865495


build_tools/github_actions/configure_jax_rocm_llvm.py
ScottTodd, line 1: I would prefer for the JAX build itself to use LLVM from the ROCm install/packages and for us to not need this `build_tools/github_actions/configure_jax_rocm_llvm.py` script. PyTorch builds get the compiler from a mix of standard and custom environment variables in the build system itself, without needing scripts like this. Example: https://github.com/ROCm/TheRock/blob/ac0283a86d02298655a998f722bfc3ac1328e6fa/external-builds/pytorch/build_prod_wheels.py#L621-L640

ScottTodd (commented):
Please get a review from someone who works directly on JAX before me for this PR.

erman-gurses:
> Please get a review from someone who works directly on JAX before me for this PR.

Agreed - currently I am communicating with them - this PR will change based on the discussion with them or maybe discarded.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+
+LOCAL_LLVM_REFS = {
+    "rocm-jaxlib-v0.10.0",
+    "rocm-jaxlib-v0.10.1",
+}
+
+HERMETIC_LLVM_REFS = {
+    "rocm-jaxlib-v0.10.2",
+    "rocm-jaxlib-v0.11.0",
+}
```
gulsumgudukbay, line 41: This ref table duplicates configure_jax_release_matrix.py L26-49. Adding a branch there alone yields a job that dies at L70 mid-build. Add an llvm_mode key to JAX_REF_CONFIGS, emit it as a matrix row key, and pass --llvm-mode here.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--dockerfile", type=Path, required=True)
+    parser.add_argument("--jax-ref", required=True)
+    args = parser.parse_args()
+
+    dockerfile = args.dockerfile.read_text(encoding="utf-8")
+
+    start_marker = "# Install LLVM 18 from source"
+    end_marker = "# ld.lld (hermetic linker)"
```
gulsumgudukbay, line 53: Markers are prose comments in ROCm/rocm-jax, maintained on 4 branches; a reword breaks all JAX builds. Same failure just killed all 15 test jobs in the linked run (run_pytest_rocm.sh has no section between ...). 
a better way to do it is to patch the Dockerfile per rocm-jaxlib-v* branch in rocm-jax, the branch already encodes the release, so no table and no script.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+
+    start = dockerfile.find(start_marker)
```
gulsumgudukbay, line 55: find silently takes the first hit. it is safer to assert each marker occurs exactly once.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
gulsumgudukbay, line 53: End marker is the libstdc++/lld block. If rocm-jax moves it, the script deletes the wrong span without raising. please match to the end of the RUN instead.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+    end = dockerfile.find(end_marker, start)
+
+    if start == -1 or end == -1:
+        raise RuntimeError("Could not find LLVM source-build block")
```
gulsumgudukbay, line 59: it'd be nice if you could include the marker text and branch in the error for easy debugging

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+    tar -xJf /tmp/llvm.tar.xz \
+      -C /usr/lib/llvm-18 \
+      --strip-components=1 && \
+    rm /tmp/llvm.tar.xz && \
+    printf '%s\n' \
+      '--gcc-toolchain=/opt/rh/gcc-toolset-14/root/usr' \
+      > /usr/lib/llvm-18/bin/clang.cfg && \
+    cp /usr/lib/llvm-18/bin/clang.cfg \
+      /usr/lib/llvm-18/bin/clang++.cfg && \
+    /usr/lib/llvm-18/bin/clang --version
```
gulsumgudukbay, line 29: clang --version exits 0 even if clang.cfg is ignored, and the --gcc-toolchain pin depends on it. 
maybe grepping for Configuration file is a better option so this fails when it stops loading. (it does work today, please check build job 94600371011.)

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+    "rocm-jaxlib-v0.10.1",
+}
+
+HERMETIC_LLVM_REFS = {
+    "rocm-jaxlib-v0.10.2",
+    "rocm-jaxlib-v0.11.0",
+}
+
+
+def main():
```
gulsumgudukbay, line 44: this doesnt have a module docstring, no help=; both are the convention in this dir. please make it main(argv) -> int like configure_jax_release_matrix.py L149 so it's testable.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+import argparse
+from pathlib import Path
+
+LLVM18_PREBUILT = r"""
```
gulsumgudukbay, line 8: nit: leading newline plus appended "\n" gives a triple blank line in the output.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+
+    start = dockerfile.find(start_marker)
+    end = dockerfile.find(end_marker, start)
+
+    if start == -1 or end == -1:
+        raise RuntimeError("Could not find LLVM source-build block")
+
+    if args.jax_ref in LOCAL_LLVM_REFS:
+        print(f"{args.jax_ref}: using prebuilt LLVM 18.1.8")
+        replacement = LLVM18_PREBUILT + "\n"
```
gulsumgudukbay, line 63: nit: leading newline plus appended "\n" gives a triple blank line in the output.

build_tools/github_actions/prepare_jax_manylinux_dockerfile.py
```diff
@@ -0,0 +1,79 @@
...
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+import argparse
+from pathlib import Path
+
+LLVM18_PREBUILT = r"""
+# Install prebuilt LLVM 18.1.8 for JAX releases using clang_local.
+RUN --mount=type=cache,target=/var/cache/dnf \
+    dnf install -y wget xz ncurses-compat-libs && \
```
gulsumgudukbay, line 11: could you please double check this? Ubuntu 18.04 tarball on ncurses-compat-libs logs "libtinfo.so.5: no version information available" message on every build.

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -190,10 +189,17 @@ jobs:
           THEROCK_VERSION: ""
           GFX_ARCH: ${{ inputs.gfx_arch }}
           ROCM_PACKAGE_FIND_LINKS_URL: ${{ inputs.rocm_package_find_links_url }}
+          JAX_REF: ${{ inputs.jax_ref }}
         run: |
+          set -euxo pipefail
```
gulsumgudukbay, line 194: set -euxo pipefail on the image step only; wheel-build step keeps the default (L194 vs L222).

.github/workflows/multi_arch_build_linux_jax_wheels.yml
```diff
@@ -224,20 +230,23 @@ jobs:
             --workdir /workspace/jax-source \
             "${MANYLINUX_IMAGE_TAG}" \
             bash -lc '
+              set -euxo pipefail
+
+              ROCM_ROOT="$(rocm-sdk path --root)"
```
gulsumgudukbay, line 235: good catch, real quoting bugs. Land these regardless (same for L242 and L248 as well).

gulsumgudukbay (changes requested):
tests/prepare_jax_manylinux_dockerfile_test.py; 30 siblings follow that name. Cover both modes, unknown ref, missing markers, and that the # ld.lld block survives.

a comment for the PR body: cite the build-log evidence: Configuration file: /usr/lib/llvm-18/bin/clang.cfg, ~77 s vs 8-11 min.

erman-gurses:
Closing since it was implemented at upstream

PR #6115, opened 2026-06-24
Revert "fix: Add missing rocm compiler configs (#5628)"

estewart08:
This reverts commit 43820c5278df2d1f653b34dc2d52d6f5719c02b6.

GitHub Issue: 6113

Issue #6113, opened 2026-06-24
[Bug]: clang++.cfg in rocm-sdk-core-7.14.0a20260624 injects -Wl,--enable-new-dtags causing PyTorch inductor PCH compilation to fail


chinmaydk99:
## Problem Description

`rocm-sdk-core-7.14.0a20260624` introduced a new `clang++.cfg` → `rocm.cfg` configuration file that unconditionally injects `-Wl,--enable-new-dtags` into every `clang++` invocation. This breaks PyTorch inductor's precompiled header (PCH) compilation with:

```
clang++: error: cannot specify -o when generating multiple output files
```

When compiling a PCH (`-x c++-header ... -o file.pch`), clang++ treats `-Wl,--enable-new-dtags` as implying an additional linker output, making total outputs > 1, which conflicts with the explicit `-o flag`.

This causes all inductor shards to fail immediately at the CPU sanity check — before any GPU tests run — in PyTorch CI: https://github.com/ROCm/TheRock/actions/runs/28110484343

### Root cause bisection

Reproduced on `banff-s80-2` inside `ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26`.

**Jun12 SDK (passes):** `rocm-sdk-core-7.14.0a20260612` — no `clang++.cfg`, `clang++` commit `52226beb248f`

**Jun24 SDK (fails):** `rocm-sdk-core-7.14.0a20260624` — has `clang++.cfg` referencing `rocm.cfg`, `clang++` commit `46fcb339fb61`

Contents of new `clang++.cfg`:
```
@rocm.cfg
```

Contents of `rocm.cfg`:
```
-Wl,--enable-new-dtags
--rocm-path='<CFGDIR>/../../..'
-frtlib-add-rpath
```

Minimal repro confirming `-Wl,--enable-new-dtags` is the sole trigger:

```bash
# Fails (Jun24, cfg active):
clang++ -x c++-header test.h \
  -fno-trapping-math -funsafe-math-optimizations -ffinite-math-only \
  -fno-signed-zeros -fno-finite-math-only -fno-unsafe-math-optimizations \
  -fmath-errno -ffp-contract=off -march=native -fPIC -std=c++20 \
  -o test.h.pch
# clang++: error: cannot specify -o when generating multiple output files

# Passes (Jun24, cfg disabled):
[...]

PR #7030, opened 2026-07-31
fix(base): Add missing rocm compiler configs

estewart08:
This adds the main rocm.cfg and alias configs:
`clang-${LLVM_VERSION_MAJOR}.cfg, clang++.cfg, clang.cfg, clang-cl.cfg, clang-cpp.cfg, flang.cfg`

These seem to have been lost when switching to TheRock.

rocm.cfg contents:
Enable RUNPATH instead of RPATH via tail option:
`$-Wl,--enable-new-dtags`

Ensures proper path base rocm install directory to determine hip version and linking of bitcode libraries: --rocm-`path='<CFGDIR>/../../..'`

Pass the architecture specific clang resource directory. This is not passed by default:
`-frtlib-add-rpath`

JIRA ID: LCOMPILER-2285
Reland of (tail option change): #5628

### DIFF
--- a/base/aux-overlay/CMakeLists.txt
+++ b/base/aux-overlay/CMakeLists.txt
@@ -51,6 +51,28 @@ if(NOT WIN32)
     "${CMAKE_CURRENT_BINARY_DIR}/bin/offload-arch"
     DESTINATION bin
   )
+
+  # Begin creation of compiler configs. We will need LLVM_VERSION_MAJOR.
+  include("${THEROCK_SOURCE_DIR}/compiler/amd-llvm/cmake/Modules/LLVMVersion.cmake")
+  set(rocm_config_name "rocm.cfg")
+  set(rocm_config_dir "${CMAKE_CURRENT_BINARY_DIR}/lib/llvm/bin")
+  set(rocm_config_content "-Wl,--enable-new-dtags\n--rocm-path='<CFGDIR>/../../..'\n-frtlib-add-rpath")
+
+  # Generate rocm.cfg
+  file(MAKE_DIRECTORY "${rocm_config_dir}")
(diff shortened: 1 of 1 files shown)


base/aux-overlay/CMakeLists.txt
```diff
@@ -51,6 +51,28 @@ if(NOT WIN32)
     "${CMAKE_CURRENT_BINARY_DIR}/bin/offload-arch"
     DESTINATION bin
   )
+
+  # Begin creation of compiler configs. We will need LLVM_VERSION_MAJOR.
+  include("${THEROCK_SOURCE_DIR}/compiler/amd-llvm/cmake/Modules/LLVMVersion.cmake")
+  set(rocm_config_name "rocm.cfg")
+  set(rocm_config_dir "${CMAKE_CURRENT_BINARY_DIR}/lib/llvm/bin")
+  set(rocm_config_content "$-Wl,--enable-new-dtags\n--rocm-path='<CFGDIR>/../../..'\n-frtlib-add-rpath")
```
ScottTodd, line 59: Looks like the only difference between this change and https://github.com/ROCm/TheRock/pull/5628 is the leading `$` here. You could structure this PR as two commits to make that style of review easier:
1. a revert of the revert (https://github.com/ROCm/TheRock/pull/6115) 
2. a new commit that has the fix

estewart08: Done

estewart08:
Downloaded the artifacts locally and verified the PCH compilation with rocm.cfg does not result in an error. This was the reason the initial patch was reverted. See #6113.

ScottTodd (approved):

PR #6389, opened 2026-07-07
ci: libhipcxx switch to amd-smi

obersteiner:
## Motivation

As rocm-smi is getting deprecated we switch with this bump to amd-smi.

## Technical Details

JIRA ID : ROCM-966

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7034, opened 2026-07-31
fix(ci): Guard libhipcxx amdsmi runtime dep on the target existing

### LINKED ISSUE #7036: hipThreads Linux nightly fails at CMake configure: libhipcxx declares an amdsmi runtime dep that doesn't exist when CORE_AMDSMI is off
## Summary

The hipThreads Linux nightly builds have been failing at CMake configure since Jul 29. `libhipcxx` declares `amdsmi` as an unconditional Linux runtime dependency, but the `amdsmi` target only exists when `THEROCK_ENABLE_CORE_AMDSMI` is on. The hipThreads CI group configures narrow with `THEROCK_ENABLE_ALL=OFF`, so the target never exists and configure dies before anything compiles.

Fix is up in #7034.

## Symptom

Both Linux hipThreads build jobs fail in the `Configure Projects` step. About fifteen instances of this, from various call sites in `therock_subproject.cmake`:

```
CMake Error at cmake/therock_subproject.cmake:1388 (get_target_property):
  get_target_property() called with non-existent target "amdsmi".
Call Stack (most recent call first):
  cmake/therock_subproject.cmake:1510 (_therock_assert_is_cmake_subproject)
  math-libs/CMakeLists.txt:524 (therock_cmake_subproject_declare)
```

Affected jobs in the `ROCm/rocm-libraries` TheRock CI Nightly:

- `Linux (hipthreads | gfx94X-dcgpu) / Build (gfx94X-dcgpu)`
- `Linux (hipthreads | gfx950-dcgpu) / Build (gfx950-dcgpu)`

Latest run: https://github.com/ROCm/rocm-libraries/actions/runs/30614819132

Two things in those jobs are cascade noise rather than separate problems. The `Report` step failing with `du: cannot access 'TheRock/build/dist/rocm'` and `Post Build Upload` failing with `FileNotFoundError: therock_manifest.json not found` are both just consequences of nothing having been built. The `Executing the custom container implementation failed. Please contact your self hosted runner administrator` annotations are the container wrapper's generic message on any non-zero exit, which makes this look like a runner or infrastructure problem when it isn't.

## Root cause

`math-libs/CMakeLists.txt` currently adds the dep based only on platform:

```cmake
set(libhipcxx_runtime_deps)
if(NOT WIN32)
  # Required on Linux.
  list(APPEND libhipcxx_runtime_deps amdsmi)
endif()
```

The `amdsmi` target is only declared under `THEROCK_ENABLE_CORE_AMDSMI` (see `core/CMakeLists.txt`). In a full build that's always on, so the dep resolves fine and this has never been a problem. In the hipThreads CI group it is off. From the failing configure log:

```
-- Implicitly enabled features: CORE_RUNTIME HIP_RUNTIME COMPILER LIBHIPCXX KPACK SYSDEPS BASE
--   * LIBHIPCXX (-DTHEROCK_ENABLE_LIBHIPCXX=ON)
--   * HIPTHREADS (-DTHEROCK_ENABLE_HIPTHREADS=ON)
-- Disabled features:
--   * CORE_AMDSMI (-DTHEROCK_ENABLE_CORE_AMDSMI=OFF)
```

`HIPTHREADS` implicitly pulls in `LIBHIPCXX`, but nothing pulls in `CORE_AMDSMI`, so `libhipcxx` ends up depending on a target that was never declared.

## Reproduce

```
cmake -B build -GNinja . \
  -DTHEROCK_AMDGPU_FAMILIES=gfx950-dcgpu \
  -DTHEROCK_ENABLE_HIPTHREADS=ON \
  -DTHEROCK_ENABLE_ALL=OFF
```

## Regression window

The unconditional dep landed in #6389 (merged Jul 27), which was correct for every configuration that existed at the time. It became reachable when the hipThreads build recipe landed in #6769 (Jul 28) and the nightly matrix entry landed in ROCm/rocm-libraries#9705 (Jul 29). The Jul 28 nightly was the last green one. It has been red on Jul 29, 30, and 31.

Worth being clear that no individual change here was wrong on its own. The dep was fine until a narrow configuration existed, and the narrow configuration couldn't have known about the dep.

## Precedent

This exact problem was already solved for the prim libs, in the same file, and the comment on it describes this scenario almost word for word:

```cmake
# amdsmi is Linux-only and only declared when THEROCK_ENABLE_CORE_AMDSMI is ON
# (see core/CMakeLists.txt). Benchmarks depend on primbench which requires
# amd_smi, so they must be disabled when amdsmi is unavailable (e.g., the
# prim-isolated CI group THEROCK_ENABLE_ALL=OFF / THEROCK_ENABLE_PRIM=ON).
if(WIN32 OR NOT TARGET amdsmi)
```

So the guard idiom exists and is documented, it just isn't enforced anywhere. My worry is that the next narrow CI group added will hit this again, since nothing stops a new recipe from copying the older platform-only pattern.

## Follow-ups

1. `hipthreads` isn't registered in `repos-config.json`, so TheRock CI skips PRs touching `projects/hipthreads` (ROCm/rocm-libraries#10218). As it stands the fix won't be exercised on the PR and we'll confirm at the next nightly.
2. It might be worth a general rule that any recipe depending on an optionally-enabled target guards on `TARGET`, or that the feature graph makes the implication explicit so the dep can stay unconditional. I don't have a strong opinion on which, and I'd defer to whoever owns the subproject machinery.
3. A PR that adds a CI matrix entry can't validate itself today. ROCm/rocm-libraries#9705 only touched `therock_matrix.py`, so nothing in it ran the new configuration.

marsavicc:
## Summary

The hipThreads Linux nightly has been failing at CMake configure since Jul 29 with `get_target_property() called with non-existent target "amdsmi"`. `THEROCK_ENABLE_HIPTHREADS=ON` implicitly pulls in libhipcxx, which adds `amdsmi` to its runtime deps unconditionally on Linux, but the `amdsmi` target is only declared under `THEROCK_ENABLE_CORE_AMDSMI`. In the narrow nightly graph (`THEROCK_ENABLE_ALL=OFF`) that feature is off, so the dep dangles and configure fails before anything compiles.

Fix: guard the dep with `AND TARGET amdsmi`, matching the prim/rand recipes in the same file.

Closes #7036

## Risk Assessment

Risk 2. One-line, configure-time guard in `math-libs/CMakeLists.txt` with no device or product code touched. In any configuration where `CORE_AMDSMI` is enabled, `TARGET amdsmi` is true and the resulting dependency graph is byte-for-byte what it was before, so full builds are unaffected. The only behavior change is in narrow graphs that previously could not configure at all.

Known trade-off, stated deliberately: the guard makes the dependency silently absent when the target does not exist, rather than failing loudly. That is the same trade-off the prim-libs guard a few hundred lines up in this file already makes. If a narrow configuration ever genuinely needs libhipcxx to have amdsmi at runtime, this will produce a quietly under-linked libhipcxx instead of an error. I think that is acceptable here because libhipcxx in a hipThreads-only graph has no consumer that needs amd-smi, but it is worth a reviewer's eye.

## Related

- Work tracking: https://github.com/ROCm/TheRock/issues/7036
- Introduced by: #6389 (added the unconditional Linux dep; correct for every configuration that existed at the time)
- Exposed by: #6769 (hipThreads build recipe) and ROCm/rocm-libraries#9705 (nightly matrix entry)
- Failing runs: https://github.com/ROCm/rocm-libraries/actions/runs/30614819132 (also red Jul 29 and Jul 30)
- Related follow-up: ROCm/rocm-libraries#10218 (`hipthreads` is not registered in `repos-config.json`, so TheRock CI skips PRs touching `projects/hipthreads`)

Architecture-independent. This is a configure-time CMake guard; no kernel, dispatch, default-selection, or support-surface change, so no device sweep is warranted.

Two different things need verifying, and PR CI only covers the first:

1. **No regression to the full graph.** Covered by this PR's own CI, which builds the full dependency graph across the Linux and Windows math-libs stages.
2. **The narrow hipThreads graph now configures.** Not covered by PR CI. The hipThreads group only runs in the rocm-libraries nightly, so confirmation comes from the next nightly on `gfx94X-dcgpu` and `gfx950-dcgpu`.

## Testing Summary

- Full-graph build via PR CI, confirming the guard does not change behavior where `amdsmi` is present.
- Local configure of the failing narrow configuration, confirming the reported error is gone.
- Nightly confirmation on the two affected hipThreads lanes, which is the only place the regression is actually observable.

## Testing Checklist

- [x] pre-commit - `pre-commit run --all-files` - Status: Passed
- [x] Unit Tests - ubuntu-24.04 and windows-2022 - Status: Passed
- [x] Linux math-libs stages - gfx110X-all, gfx1151, gfx120X-all - Status: Passed
- [ ] Linux math-libs stage - gfx94X-dcgpu - Status: Pending
- [ ] Windows math-libs stages - gfx110X-all, gfx1151, gfx120X-all - Status: Pending
- [ ] Local narrow-config repro - `cmake -B build -GNinja . -DTHEROCK_AMDGPU_FAMILIES=gfx950-dcgpu -DTHEROCK_ENABLE_HIPTHREADS=ON -DTHEROCK_ENABLE_ALL=OFF` - Status: Pending
- [ ] rocm-libraries nightly, hipThreads Linux lanes - Devices: gfx94X-dcgpu, gfx950-dcgpu - Status: Pending

## Flags / Guardrails

None. No new option or default is introduced.

## Adjacent Tests Considered

The same guard pattern already exists in this file for the prim libs (`if(WIN32 OR NOT TARGET amdsmi)`), which was added for the equivalent `THEROCK_ENABLE_ALL=OFF / THEROCK_ENABLE_PRIM=ON` group. Those recipes are unaffected by this change.

After this change nothing in `math-libs/CMakeLists.txt` declares `amdsmi` without a guard, but other files still do. `math-libs/BLAS/CMakeLists.txt` (hipBLASLt and hipSPARSELt) and `profiler/CMakeLists.txt` carry the same unguarded pattern, and are safe today only because their artifacts declare `core-amdsmi` in `BUILD_TOPOLOGY.toml`. That is out of scope here and tracked separately in #7037.

## Risk Acceptance / Waivers

`W-BUILD` — build-configuration-only change with no product behavior. No new automated test is added because the regression is a configure-time failure that the existing hipThreads nightly lane already detects; a new test would duplicate a lane we already run. The reproduction evidence is the three consecutive red nightlies before the change and the narrow-config configure after it.

## Technical Changes

- `math-libs/CMakeLists.txt`: change the libhipcxx runtime-dep guard from `if(NOT WIN32)` to `if(NOT WIN32 AND TARGET amdsmi)` so the dep is only declared when `CORE_AMDSMI` actually provided the target.

### DIFF
--- a/math-libs/CMakeLists.txt
+++ b/math-libs/CMakeLists.txt
@@ -516,7 +516,7 @@ if(THEROCK_ENABLE_LIBHIPCXX)
   ##############################################################################
 
   set(libhipcxx_runtime_deps)
-  if(NOT WIN32)
+  if(NOT WIN32 AND TARGET amdsmi)
     # Required on Linux.
     list(APPEND libhipcxx_runtime_deps amdsmi)
   endif()
--- a/math-libs/CMakeLists.txt
+++ b/math-libs/CMakeLists.txt
@@ -516,8 +516,12 @@ if(THEROCK_ENABLE_LIBHIPCXX)
   ##############################################################################
 
   set(libhipcxx_runtime_deps)
+  # Used on Linux when amd-smi is available. amdsmi is only declared when
+  # THEROCK_ENABLE_CORE_AMDSMI is ON (see core/CMakeLists.txt), so guard on the
+  # target existing to keep narrow configs (e.g. THEROCK_ENABLE_LIBHIPCXX=ON /
+  # THEROCK_ENABLE_ALL=OFF, as in the hipThreads nightly) from referencing a
+  # non-existent target.
   if(NOT WIN32 AND TARGET amdsmi)
-    # Required on Linux.
     list(APPEND libhipcxx_runtime_deps amdsmi)
   endif()
 


tony-davis:
Thanks for turning this around so fast, Marko. The guard is the right shape and it matches what prim and rand already do in that file, so no argument from me on the approach. Two things before it goes in, one trivial and one that's really a question for Michael.

**The comment needs to change.** Line 520 still says `# Required on Linux.` but it now sits inside a guard that makes the dep optional. As written, a reader will stop and ask "if it's required, why is it conditional, and what happens on Linux when it's missing?" Either the dep is genuinely required, in which case guarding it quietly under-links libhipcxx instead of failing loudly, or it isn't, in which case the comment should say something like "used on Linux when amd-smi is available." One word either way, but I don't think it can stay as-is.

**Which fix do we actually want?** I went digging in `BUILD_TOPOLOGY.toml` and libhipcxx looks like the only artifact that consumes amdsmi without declaring `core-amdsmi`. `blas`, `rccl`, `rdc`, `rocprofiler-systems`, `rocrtst`, `dctools-core` and `profiler-apps` all list it in `artifact_deps`. So there's a competing one-line fix: add `core-amdsmi` to `[artifacts.libhipcxx].artifact_deps`, which makes the topology honest and lets the CMake dep stay unconditional with its comment intact. The trade-off is real, since that makes the hipThreads build pull in and compile amdsmi, so the narrow group gets slower and less narrow. I don't have a strong opinion. I'd rather defer to Michael here, since he added the dep and knows whether libhipcxx genuinely needs amd-smi at runtime. If it's genuinely optional then this PR is exactly right and we just fix the comment.

Tiny nit, take it or leave it: `core-amdsmi` has `disable_platforms = ["windows"]`, so `TARGET amdsmi` is already false on Windows and the `NOT WIN32` is redundant now. The prim guard keeps its `WIN32 OR` too, so there's a consistency argument for leaving it. Not worth a round trip.

Something I noticed while looking around, and it is not this PR's job: `math-libs/BLAS/CMakeLists.txt` has the same unguarded pattern for hipBLASLt and hipSPARSELt, and `profiler/CMakeLists.txt` has it as well. They're safe today only because their artifacts declare `core-amdsmi`. I've filed #7037 to track that separately.

Last thing on validation. PR CI here builds the full graph, so green tells us this doesn't regress anything, but nothing in TheRock PR CI runs the narrow hipThreads config. We won't actually know the nightly is fixed until it runs. I'll watch tomorrow's and report back either way.

tony-davis (approved):
It should unblock the CI. Looks good, with comments above.

math-libs/CMakeLists.txt
```diff
@@ -516,7 +516,7 @@ if(THEROCK_ENABLE_LIBHIPCXX)
   ##############################################################################
 
   set(libhipcxx_runtime_deps)
-  if(NOT WIN32)
+  if(NOT WIN32 AND TARGET amdsmi)
     # Required on Linux.
     list(APPEND libhipcxx_runtime_deps amdsmi)
   endif()
```
ScottTodd, line 526: If amdsmi is required by libhipcxx on Linux as this comment suggests, `core-amdsmi` should be added to `artifact_deps` in the build topology for `libhipcxx` as done here:
* https://github.com/ROCm/TheRock/blob/36024af03ae158c332383a66dec3f307d33a36cd/BUILD_TOPOLOGY.toml#L587-L591
* https://github.com/ROCm/TheRock/blob/36024af03ae158c332383a66dec3f307d33a36cd/math-libs/BLAS/CMakeLists.txt#L135-L144

marsavicc:
> Thanks for turning this around so fast, Marko. The guard is the right shape and it matches what prim and rand already do in that file, so no argument from me on the approach. Two things before it goes in, one trivial and one that's really a question for Michael.
> 
> **The comment needs to change.** Line 520 still says `# Required on Linux.` but it now sits inside a guard that makes the dep optional. As written, a reader will stop and ask "if it's required, why is it conditional, and what happens on Linux when it's missing?" Either the dep is genuinely required, in which case guarding it quietly under-links libhipcxx instead of failing loudly, or it isn't, in which case the comment should say something like "used on Linux when amd-smi is available." One word either way, but I don't think it can stay as-is.
> 
> **Which fix do we actually want?** I went digging in `BUILD_TOPOLOGY.toml` and libhipcxx looks like the only artifact that consumes amdsmi without declaring `core-amdsmi`. `blas`, `rccl`, `rdc`, `rocprofiler-systems`, `rocrtst`, `dctools-core` and `profiler-apps` all list it in `artifact_deps`. So there's a competing one-line fix: add `core-amdsmi` to `[artifacts.libhipcxx].artifact_deps`, which makes the topology honest and lets the CMake dep stay unconditional with its comment intact. The trade-off is real, since that makes the hipThreads build pull in and compile amdsmi, so the narrow group gets slower and less narrow. I don't have a strong opinion. I'd rather defer to Michael here, since he added the dep and knows whether libhipcxx genuinely needs amd-smi at runtime. If it's genuinely optional then this PR is exactly right and we just fix the comment.
> 
> Tiny nit, take it or leave it: `core-amdsmi` has `disable_platforms = ["windows"]`, so `TARGET amdsmi` is already false on Windows and the `NOT WIN32` is redundant now. The prim guard keeps its `WIN32 OR` too, so there's a consistency argument for leaving it. Not worth a round trip.
> 
> Something I noticed while looking around, and it is not this PR's job: `math-libs/BLAS/CMakeLists.txt` has the same unguarded pattern for hipBLASLt and hipSPARSELt, and `profiler/CMakeLists.txt` has it as well. They're safe today only because their artifacts declare `core-amdsmi`. I've filed #7037 to track that separately.
> 
> Last thing on validation. PR CI here builds the full graph, so green tells us this doesn't regress anything, but nothing in TheRock PR CI runs the narrow hipThreads config. We won't actually know the nightly is fixed until it runs. I'll watch tomorrow's and report back either way.

Thanks! Fixed the comment.

math-libs/CMakeLists.txt
marsavicc: Should that be a separate PR or can I add it while I'm here?

tony-davis: I'd very slightly lean toward landing this as-is, but writing up a follow-up PR right away to clean all that up and link it to this one. That way we can get some CI unblocked but really clean it up. However, if you want to put it in here, that's fine too. The urgency on the nightly rocm-libraries isn't super-high. We have a little time.

marsavicc: Ok, merging this then and I'll follow up with build topology changes in other PR

PR #7038, opened 2026-07-31
[ROCM-20223] CV Integration - RPP build (1/3)

kiritigowda:
## Summary

Adds RPP (ROCm Performance Primitives) as a new `cv-libs/` component. RPP is target-neutral: it uses `USE_DIST_AMDGPU_TARGETS` to compile a single binary covering all distribution targets, so there are no per-architecture variants.

- New `cv-libs/` subproject directory with the RPP CMake declaration and `artifact-rpp.toml`
- `cv-libs` build stage and target-neutral `rpp` artifact in `BUILD_TOPOLOGY.toml`
- `THEROCK_ENABLE_CV_LIBS` / `THEROCK_ENABLE_RPP` feature flags, on by default on Linux and opt-in on Windows
- Wires the stage into the portable Linux build and the multi-arch Linux workflow

This is part 1 of 3, split out of #5708 per review feedback so that build, packaging and CI test changes can each go to the right area owners:

1. **Build** (this PR)
2. **Packaging** — `rocm-sdk-libraries` wheel and native Linux packages
3. **CI tests** — `test_rpp.py` and test configuration (remains on #5708)

## Test plan

- [x] `pytest build_tools/tests build_tools/github_actions/tests` — 1303 passed. The 4 remaining failures are pre-existing Windows symlink-privilege errors in `fileset_tool_test` and `py_packaging_test`; neither module is touched by this PR.
- [x] cv-libs stage built in the manylinux CI container from prebuilt target-neutral artifacts. `librpp.so` links clean with no unresolved `DT_NEEDED`.
- [ ] CI cv-libs stage green

PR #7039, opened 2026-07-31
[ROCM-20223] CV Integration - RPP packaging (2/3)

kiritigowda:
## Summary

Packages RPP into the ROCm distributions. Part 2 of 3, split out of #5708.

- Adds `librpp.so` to the `rocm-sdk-libraries` wheel, placed at the head of the libraries block in `_dist_info.py` since RPP depends only on core
- Adds a `lib/llvm/lib` rpath dep to the libraries wheel so RPP can resolve `libomp`
- Adds an `--rpp` flag to `install_rocm_from_artifacts.py`, with docs
- Adds the RPP components to the native Linux packages

> [!IMPORTANT]
> Depends on #7038 (the cv-libs build change). The `rpp` artifacts referenced here do not exist until that lands, so please merge #7038 first.

### Note for reviewers

`build_python_packages.py` takes an rpath dep on `lib/llvm/lib` for the libraries wheel. RPP's only need from that directory is `libomp.so`, confirmed via `readelf -d` on a built `librpp.so`. The `core` and `profiler` wheels already take the same dep, and `rpath_dep()` is package-scoped with no per-library granularity, so narrowing it to RPP alone would need new machinery. Happy to take a different approach if preferred — see the discussion on #5708.

## Test plan

- [x] `pytest build_tools/tests/install_rocm_from_artifacts_test.py build_tools/tests/py_packaging_test.py` — 78 passed. The 3 failures are pre-existing Windows symlink-privilege errors in the profiler symlink tests, unrelated to RPP.
- [ ] Wheel build green once #7038 lands


rrawther (approved):

ScottTodd:
> Depends on #7038 (the cv-libs build change). The `rpp` artifacts referenced here do not exist until that lands, so please merge #7038 first.

You could try to use https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests for these PRs.

build_tools/packaging/linux/package.json
```diff
@@ -3299,6 +3299,116 @@
...
+        "Artifact_Subdir": [
+          {
+            "Name": "rpp",
+            "Components": [
+              "test"
+            ]
+          }
+        ]
+      }
+    ],
```
ScottTodd, line 3334: Per my comment on https://github.com/ROCm/TheRock/pull/7038#discussion_r3706801416, the first PR that just adds the build should _add_ an exclusion here: https://github.com/ROCm/TheRock/blob/0888ca35dad1ae9a5f9b46bf56269520fde79cdb/tests/test_artifact_structure.py#L67-L83

then this PR that adds the package.json change should _remove_ that exclusion

ScottTodd (commented):
Code changes LGTM. Using stacked PRs (either the built in feature or by setting the base branch for each PR) would let us see CI results on this PR. Without that, we'll have to either merge the PRs in sequence and wait for full CI cycles each time, or merge without CI signal.

kiritigowda:
Superseded by #7080, which targets the build layer directly rather than `main` — thanks for the stacked-PR pointer. The blocker was that both branches lived in a fork, and [cross-fork stacks are not supported](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs); re-pushing them into this repo as `users/kiriti/*` branches made it work. #7079 and #7080 are now stack #7083.

Keeping this PR open for the review history, but **please review #7080 instead.**

Your exclusion request is implemented across the stack: #7079 adds the four `rpp` entries to `KNOWN_UNCOVERED_COMPONENTS`, and #7080 removes them alongside the `package.json` coverage. Neither layer can pass CI in the wrong order, so the dependency is machine-checked now instead of being a note in the description.

kiritigowda:
replaced by #7080

PR #7043, opened 2026-07-31
Rocprofiler-systems: build with ROCPROFSYS_BUILD_PORTABLE=ON to avoid AVX-512 in artifacts

sputhala-amd:
## Motivation

The rocprofiler-systems binaries in the nightly tarball crash at startup with `SIGILL` (Illegal instruction, exit 132) on CPUs without AVX-512:
- `rocprof-sys-avail --version` -> `Illegal instruction (core dumped)`
- Faulting instruction is AVX-512: `vmovups -0x...(%rip),%zmm0`
- `objdump -d` shows `%zmm` / `vmovdqu64` in `rocprof-sys-avail` and `librocprof-sys.so`

Confirmed dependency on AVX-512 support in the *runtime* CPU:
| CPU | AVX-512 | Result |
|---|---|---|
| Intel Xeon 8480C (Sapphire Rapids) | yes | works |
| AMD EPYC 9354 (Zen4/Genoa) | yes | works |
| AMD EPYC 7763 / 7A53 (Zen3, incl. MI250X hosts / Frontier) | no | SIGILL |

JIRA ID: ROCM-27668
JIRA ID: ROCM-28107
JIRA ID: AIPROFSYST-702

**Root cause**: 
Rocprofiler-systems builds its vendored `timemory` with `TIMEMORY_BUILD_PORTABLE=OFF` (timemory's default). With that OFF, `timemory/cmake/Modules/ConfigCpuArch.cmake` **host-detects the build machine's CPU features** and adds `-m<feature>` flags for each (e.g. `-mavx512f`), and can fall back to `-march=native`. When the CI/build host is AVX-512-capable (Sapphire Rapids or Zen4/Genoa), `-O3`/`-ftree-vectorize` then emits 512-bit `%zmm` code into ordinary auto-vectorized loops. Those instructions are illegal on Zen3 -> SIGILL.

## Technical Details

Depends on:

- [ ] https://github.com/ROCm/rocm-systems/pull/9580

- Add `-DROCPROFSYS_BUILD_PORTABLE=ON` to the rocprofiler-systems build configuration. This option (added upstream in rocprofiler-systems) forwards to `TIMEMORY_BUILD_PORTABLE=ON`, which caps `ConfigCpuArch` at AVX2 and disables host-native / AVX-512 arch flags. The result is a builder-independent baseline that runs on all supported Instinct hosts.

- **Risk** - Low. rocprof-sys is a profiler - its hot paths are sampling, unwinding, buffer and hash management, not dense numeric kernels - so the AVX2-vs-AVX-512 difference is negligible in practice, and it does not affect the profiled application's performance. Correctness/portability across Instinct hosts far outweighs any marginal SIMD width loss (and on some AVX-512 parts AVX2 avoids downclocking).

## Submission Checklist


sputhala-amd:
Closing this draft PR.

Issue #2653, opened 2025-12-19
buildctl.py: expunging a disabled component leaves it disabled permanently


eble-amd:
### Steps to reproduce

The state of your build directory is that a full build has occurred and all components are enabled.  `find . -name stage.prebuilt` finds no files.

1. `./build_tools/buildctl.py enable --build-dir=/therock/output/build rocSOLVER`

The output of buildctl.py shows that it has disabled all components other than rocSOLVER by creating stage.prebuilt files.  Check it for hipCUB specifically:

```
$ ls -l build/math-libs/hipCUB/stage.prebuilt
-rw-r--r-- 1 eble redacted 0 Dec 18 13:57 build/math-libs/hipCUB/stage.prebuilt
```

2. Imagine that you are new to TheRock, weeks have passed since step 1, you have forgotten that building hipCUB is disabled, and now you are trying to figure out why hipCUB is not rebuilding when you touch a file, so you decide to try expunging it.
3. `ninja -v -C /therock/output/build hipCUB+expunge`
4. `cmake -B /therock/output/build -GNinja .`
5. `ninja -v -C /therock/output/build hipCUB`

hipCUB is still disabled. (Find stage.prebuilt with same command as above.)

5. Imagine that someone reminds you of buildctl.py, so you decide to try re-enabling everything.
6. `./build_tools/buildctl.py enable --build-dir=/therock/output/build`

The output of buildctl.py does _not_ show hipCUB.  hipCUB is still disabled.  (Find stage.prebuilt with same command as above.)

### Workaround

Remove hipCUB's stage.prebuilt file by hand.

PR #6984, opened 2026-07-30
feat(build): add typed ROCm build flag provider protocol

stellaraccident:
Provide one fail-closed build flag contract for integrated and standalone ROCm project builds. Values are resolved while building a library and do not add an installed package dependency.

Changes:
- Add versioned BOOL and INTEGER provider state and consumer resolution.
- Add unconditional aux-overlay C and C++ conformance compilation.
- Document adoption, precedence, validation, and distribution boundaries.

Tests:
- 12 focused Python and CMake fixture tests passed.
- Minimal TheRock configure and therock-aux-overlay+build passed.
- pre-commit passed for every changed TheRock file.

PR #7046, opened 2026-08-01
fix(build): remove stage.prebuilt marker on subproject expunge

Alexander-Popescu:
## Motivation

`buildctl.py disable` writes a `stage.prebuilt` marker next to a project's `stage/` dir. `+expunge` deletes the build, stage, stamp and dist dirs, but the marker sits next to `stage/` instead of inside it, so it survived. The next configure picked it up and the project stayed disabled. Only way out was deleting the file by hand.

ISSUE ID: #2653

## Technical Details

Adds `${_prebuilt_file}` to the paths `+expunge` deletes.

## Test Plan

`build_tools/tests/therock_subproject_expunge_test.py` builds a small project that includes the real `cmake/therock_subproject.cmake`, drops the marker, runs `+expunge` and checks it's gone. Needs cmake and ninja, which the `unit_tests.yml` runners already have.

## Test Result

Fails on main but passes with change.

## Submission Checklist

Written with AI but I reviewed.

### DIFF
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -1126,10 +1126,12 @@ function(therock_cmake_subproject_activate target_name)
   add_custom_target("${target_name}+dist")
 
   # expunge target
+  # The prebuilt marker sits next to the stage dir, not inside it, so it has to
+  # be deleted here too or the project stays disabled.
   add_custom_target(
     "${target_name}+expunge"
     COMMAND
-      ${CMAKE_COMMAND} -E rm -rf "${_binary_dir}" "${_stage_dir}" "${_stamp_dir}" "${_dist_dir}"
+      ${CMAKE_COMMAND} -E rm -rf "${_binary_dir}" "${_stage_dir}" "${_stamp_dir}" "${_dist_dir}" "${_prebuilt_file}"
   )
   add_dependencies(therock-expunge "${target_name}+expunge")
 


build_tools/tests/therock_subproject_expunge_test.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Regression test for the subproject '+expunge' target.
```
ScottTodd: A 90 line regression test file for a 20 _character_ change is overly specific. I think we could write unit tests for that part of the CMake build but they should be closer to the `build_tools/tests/rocm_build_flags_test.py` tests that https://github.com/ROCm/TheRock/pull/6984 is adding.

For a simulated test exercising just the `.cmake` files:
1. Write a `CMakeLists.txt` that uses `therock_subproject.cmake` and defines a simple graph of subprojects
2. Build one project
3. Build the expunge target
4. Check that rebuilds work as expected

For a test against the real CMake project (this may be too expensive for a unit test):

1. Configure CMake
2. Build `therock-aux-overlay`
3. Build the expunge target
4. Check that rebuilds work as expected

ScottTodd (commented):
I think the code change here is fine. I don't want to add a dedicated integration test for such a small change though. We could add general tests that cover this, but I wouldn't block the fix on having that test coverage.

build_tools/tests/therock_subproject_expunge_test.py
Alexander-Popescu: Yeah that's true. Removed. This is my first PR so I was erring on the side of more tests.

For the more general test you were describing, if you think it would be useful to add, I could give it a shot.

ScottTodd: Semi-useful, particularly after the PR I linked is merged and we have a good example in the project to model future test cases after.

I've been working on a "testing philosophy" page at https://github.com/ROCm/TheRock/issues/6711 that will explain how we test changes to different parts of the project. The hope there is that we'll have sufficiently detailed documentation for developers (and coding agents) to know what sort of testing is expected for each change and how those tests are run (locally, on CI, as part of releases, etc.).

ScottTodd (approved):

PR #7054, opened 2026-08-02
fix(pytorch): Add libamd_smi to the Linux library wheel preload list

nunnikri:
## Summary

Add `libamd_smi` to the Linux library preload list in the PyTorch wheel build
so that amdsmi symbols are available via `dlsym(RTLD_DEFAULT, ...)` in
`IntraNodeComm::getNvlMesh`.

JIRA ID : ROCM-27833

## Motivation

Two distributed NCCL tests hang and timeout after 300 seconds on ROCm Python
wheel installs (ROCM-27833):

Root cause: `IntraNodeComm::getNvlMesh` calls `dlopen("libamd_smi.so")` to
load amdsmi for NVLink/xGMI mesh detection. In ROCm Python wheel installs,
only the versioned `libamd_smi.so.26` exists in `_rocm_sdk_core/lib/` — there
is no unversioned `libamd_smi.so` in the runtime package. `dlopen` by
unversioned name fails, causing the intra-node communication path to break.

## Technical Details

`_rocm_init.py` (generated by `get_rocm_init_contents()` and injected into
the torch wheel at build time) calls `rocm_sdk.initialize_process()` with a
list of library shortnames on `import torch`. Each library is loaded via
`ctypes.CDLL(path, mode=RTLD_GLOBAL)`, which exports all its symbols into
the process-wide global symbol table.

Adding `"amd_smi"` to `LINUX_LIBRARY_PRELOADS` causes `libamd_smi.so.26` to
be loaded with `RTLD_GLOBAL` on torch import. The paired PyTorch change
(`intra_node_comm.cpp`) then uses `dlsym(RTLD_DEFAULT, "amdsmi_init")` as the
primary lookup — which finds the symbol in the global table regardless of
what filename the library was opened with — and falls back to `dlopen` for
system ROCm installs where the library is not preloaded.

## Test Plan

- Rebuild PyTorch wheel with this change
- Install wheel without `rocm-sdk-devel`:
  ```bash
      "torch[device-gfx942]"


nunnikri:
closing in favor of https://github.com/ROCm/TheRock/pull/7057

Issue #6695, opened 2026-07-20
[Issue]: TheRock Pytorch build wheels _rocm_init.py to not preload rocprofiler-sdk.so


venkat1361:
### Problem Description

PyTorch wheels built by TheRock currently preload rocprofiler-sdk during torch startup through the generated `torch/_rocm_init.py` flow. This behavior was introduced by commit `6a55012d931359f125b540ea6f609c44629bc60c` to fix nightly smoke test failures after PyTorch/Kineto started depending on `librocprofiler-sdk.so.1`.

That fixed the missing runtime dependency, but it introduced a regression for profiler-launched workflows. When Python is launched under an external rocprofv3(not shipped by rocm or from rocprofiler-sdk build directory), rocprofv3 will already preload and initialize its own rocprofiler-sdk provider. Later, importing PyTorch explicitly loads the wheel-bundled rocprofiler-sdk, which can create a duplicate SDK/provider path in the same process.

This can cause rocprofv3’s tool configuration path to be invoked after rocprofiler-sdk has already locked configuration, resulting in:
```
ROCPROFILER_STATUS_ERROR_CONFIGURATION_LOCKED
Configuration request occurred outside of valid rocprofiler configuration period
```

### Operating System

Ubuntu 22/24

### CPU

AMD EPYC 9654 96-Core Processor

### GPU

MI300X, All archs

### ROCm Version

rocm 7.14

### Steps to Reproduce

- Install a TheRock PyTorch wheel that includes the rocprofiler-sdk preload added by commit `6a55012d931359f125b540ea6f609c44629bc60c`.
- Build rocprofiler-sdk from rocm-systems

Create a minimal PyTorch smoke script:

```
# torch_smoke.py
import torch
print(torch.__version__)
print(torch.cuda.is_available())
Run the script under external rocprofv3:

```

- `../rocprofiler-sdk-build/bin/rocprofv3 --hip-trace -- python3 torch_smoke.py` Observe that rocprofv3 initializes its SDK/tool path before Python imports torch.

- During import torch, PyTorch executes generated ROCm startup code and explicitly preloads the wheel-bundled rocprofiler-sdk.

- The process can fail when rocprofv3 attempts API-table intercept registration after rocprofiler-sdk configuration is already locked.

- Attached .md file for clear instructions 

[therock-pytorch-rocm-install-notes.md](https://github.com/user-attachments/files/30196118/therock-pytorch-rocm-install-notes.md)

[...]

Issue #3177, opened 2026-01-30
[CI] Expand CI workflows to build and test packages (ROCm Python, PyTorch, JAX, Native Linux, etc.)


ScottTodd:
## Background

Our CI (per-commit) and CD (dev/nightly/stable release) workflows diverged a while ago. They are building and testing different sets of jobs.

As of 2026-02-09, coverage looks like (✨ indicates "new since the issue was filed"):

Job description | CI in rocm-libraries<br>and rocm-systems | CI in TheRock | Releases in TheRock and rockrel
-- | -- | -- | --
Build subset of ROCm artifacts | ✅ Yes | ❌No | ⚪ N/A
Test subset of ROCm artifacts | ✅ Yes | ✅ Yes ([see docs](https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md)) | ⚪ N/A
Build all ROCm artifacts | ❌No | ✅ Yes | ✅ Yes
Test all ROCm artifacts | ❌No | ✅ Yes | ❌No (https://github.com/ROCm/TheRock/issues/2281)
Upload ROCm tarballs | ❌No | ❌No | ✅ Yes 
Build ROCm Python packages | ❌No | ✅ Yes (https://github.com/ROCm/TheRock/issues/1559) | ✅ Yes 
Test ROCm Python packages | ❌No | ✅ Yes (https://github.com/ROCm/TheRock/issues/1559) ✨ | ✅ Yes 
Build PyTorch packages | ❌No | 🟡 Partial (https://github.com/ROCm/TheRock/issues/3291) ✨ | ✅ Yes 
Test PyTorch packages | ❌No | ❌No (https://github.com/ROCm/TheRock/issues/3291) | 🟡 Partial (see https://github.com/ROCm/TheRock/issues/2258 and https://github.com/ROCm/TheRock/issues/2507)
Build JAX packages | ❌No | ❌No | ✅ Yes 
Test JAX packages | ❌No | ❌No | ✅ Yes (partial?)
Build native Linux packages | ❌No | ❌No | ✅ Yes 
Test native Linux packages | ❌No | ❌No | ❌No (https://github.com/ROCm/TheRock/pull/3126)

Notes:
* When coverage is only included in release workflows, commits that pass presubmit can break releases
* When coverage is only included in CI workflows, cherry-picks onto release branches carry additional risk
[...]

PR #7057, opened 2026-08-02
fix(pytorch): Add libamd_smi to the Linux library wheel preload list

nunnikri:
## Summary

Add `libamd_smi` to the Linux library preload list in the PyTorch wheel build
so that amdsmi symbols are available via `dlsym(RTLD_DEFAULT, ...)` in
`IntraNodeComm::getNvlMesh`.

JIRA ID : ROCM-27833

## Motivation

Two distributed NCCL tests hang and timeout after 300 seconds on ROCm Python
wheel installs (ROCM-27833):

Root cause: `IntraNodeComm::getNvlMesh` calls `dlopen("libamd_smi.so")` to
load amdsmi for NVLink/xGMI mesh detection. In ROCm Python wheel installs,
only the versioned `libamd_smi.so.26` exists in `_rocm_sdk_core/lib/` — there
is no unversioned `libamd_smi.so` in the runtime package. `dlopen` by
unversioned name fails, causing the intra-node communication path to break.

## Technical Details

`_rocm_init.py` (generated by `get_rocm_init_contents()` and injected into
the torch wheel at build time) calls `rocm_sdk.initialize_process()` with a
list of library shortnames on `import torch`. Each library is loaded via
`ctypes.CDLL(path, mode=RTLD_GLOBAL)`, which exports all its symbols into
the process-wide global symbol table.

Adding `"amd_smi"` to `LINUX_LIBRARY_PRELOADS` causes `libamd_smi.so.26` to
be loaded with `RTLD_GLOBAL` on torch import. The paired PyTorch change
(`intra_node_comm.cpp`) then uses `dlsym(RTLD_DEFAULT, "amdsmi_init")` as the
primary lookup — which finds the symbol in the global table regardless of
what filename the library was opened with — and falls back to `dlopen` for
system ROCm installs where the library is not preloaded.

## Test Plan

- Rebuild PyTorch wheel with this change
- Install wheel without `rocm-sdk-devel`:
  ```bash
      "torch[device-gfx942]"

### DIFF
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -161,10 +161,14 @@ is_windows = platform.system() == "Windows"
 # LLVM download URL for triton-windows
 LLVM_BASE_URL = "https://oaitriton.blob.core.windows.net/public/llvm-builds"
 
-# List of library preloads for Linux to generate into _rocm_init.py
+# List of library preloads for Linux to generate into _rocm_init.py.
+# These are loaded with RTLD_GLOBAL on `import torch` via _rocm_init.py so
+# that their symbols are available via dlsym(RTLD_DEFAULT, ...) without
+# requiring a successful dlopen by unversioned name (which fails in wheel
+# installs where only the versioned .so exists in the runtime package).
 LINUX_LIBRARY_PRELOADS = [
     "amd_comgr",
-    "amd_smi",  # Preload with RTLD_GLOBAL so amdsmi symbols are resolvable via dlsym(RTLD_DEFAULT, ...)
+    "amd_smi",
     "amdhip64",
     "rocprofiler-sdk",  # Linux only: needed by torch since kineto uses rocprofiler-sdk.
     "rocprofiler-sdk-roctx",  # Linux only for the moment.
--- a/external-builds/pytorch/build_prod_wheels.py
+++ b/external-builds/pytorch/build_prod_wheels.py
@@ -164,6 +164,7 @@ LLVM_BASE_URL = "https://oaitriton.blob.core.windows.net/public/llvm-builds"
 # List of library preloads for Linux to generate into _rocm_init.py
 LINUX_LIBRARY_PRELOADS = [
     "amd_comgr",
+    "amd_smi",  # Preload with RTLD_GLOBAL so amdsmi symbols are resolvable via dlsym(RTLD_DEFAULT, ...)
     "amdhip64",
     "rocprofiler-sdk",  # Linux only: needed by torch since kineto uses rocprofiler-sdk.
     "rocprofiler-sdk-roctx",  # Linux only for the moment.
--- a/external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
+++ b/external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
@@ -1,10 +1,13 @@
 # Copyright Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
+import ctypes
 import logging
-import torch
-import pytest
+import platform
 import re
+
+import pytest
+import torch
 
 
 class TestROCmAvailability:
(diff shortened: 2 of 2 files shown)


external-builds/pytorch/build_prod_wheels.py
```diff
@@ -164,6 +164,7 @@
 # List of library preloads for Linux to generate into _rocm_init.py
 LINUX_LIBRARY_PRELOADS = [
     "amd_comgr",
+    "amd_smi",  # Preload with RTLD_GLOBAL so amdsmi symbols are resolvable via dlsym(RTLD_DEFAULT, ...)
```
ScottTodd: I don't think the comment here on _just_ `amd_smi` is useful. A similar comment could be added up a few lines since it applies to _all_ library preloads.

ScottTodd: > ## Motivation
> 
> Two distributed NCCL tests hang and timeout after 300 seconds on ROCm Python wheel installs (ROCM-27833):
> 
> * `CommTest::test_intra_node_comm_all_reduce_custom_group_name_False` 
> * `CommTest::test_intra_node_comm_all_reduce_custom_group_name_True`



Rather than rely on just unit tests in pytorch, we may want to add some more specific tests for ROCm libraries and compatibility between torch, amdsmi, rocprofiler, etc.

Here's one spot where we could add a similar test, or we could create some new test files: https://github.com/ROCm/TheRock/blob/4d7626337cac97ed87801f3f57224e12fe509860/external-builds/pytorch/smoke-tests/pytorch_smoke_test.py#L134-L142

See also:
* https://github.com/ROCm/TheRock/issues/6695

ScottTodd (approved):

external-builds/pytorch/build_prod_wheels.py
nunnikri: Thanks @ScottTodd for the comments. Added 2 tests in external-builds/pytorch/smoke-tests/pytorch_smoke_test.py.
1. Testing all the exported libraries in external-builds/pytorch/build_prod_wheels.py are available and findable.
2. Confirming the available libraries as LD_PRELOAD-ed by checking dlsym if few amd_smi fns.

external-builds/pytorch/smoke-tests/pytorch_smoke_test.py
```diff
@@ -158,3 +161,68 @@ def test_lapack_available_svd(self):
...
+        for shortname, entry in _dist_info.ALL_LIBRARIES.items():
+            if entry.optional:
+                continue  # skip optional libs (e.g. rocdxg — WSL only)
+            # Skip if the package that provides this library is not installed
+            py_package_name = entry.package.pure_py_package_name
+            if importlib.util.find_spec(py_package_name) is None:
+                continue
+            paths = rocm_sdk.find_libraries(shortname)
+            assert paths, f"rocm_sdk.find_libraries('{shortname}') returned no paths"
+            assert paths[0].exists(), f"Library '{shortname}' not found at {paths[0]}"
```
ScottTodd: This is a test of rocm, not torch. It should go in a different file.

ScottTodd: The torch specific behavior is `_rocm_init.py`, and that's independent of what this code is testing. What matters is that torch can initialize and any rocm libraries that it uses can also be loaded _with the code in torch itself_.

Most/all of this function I think is already tested by `rocm-sdk test`

nunnikri: Good catch @ScottTodd 
Need to keep remind myself to review closely AI generated code.

nunnikri: Updated the test, with focus on the libraries is PRELOADED and is able to use with torch import

external-builds/pytorch/build_prod_wheels.py
nunnikri: Updated the comments to a common area

ScottTodd (approved):

nunnikri:
The test results from pytorch smoke test run in the nightly. https://github.com/ROCm/rockrel/actions/runs/30872304988/job/91889067154

```
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_core_preloaded_symbol_resolvable_via_rtld_default[amd_smi-amdsmi_init] PASSED [ 75%]
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_core_preloaded_symbol_resolvable_via_rtld_default[amd_smi-amdsmi_get_socket_handles] PASSED [ 80%]
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_core_preloaded_symbol_resolvable_via_rtld_default[amd_smi-amdsmi_get_processor_handles] PASSED [ 85%]
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_core_preloaded_symbol_resolvable_via_rtld_default[amd_smi-amdsmi_is_P2P_accessible] PASSED [ 90%]
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_core_preloaded_symbol_resolvable_via_rtld_default[amdhip64-hipGetDeviceCount] PASSED [ 95%]
external-builds/pytorch/smoke-tests/pytorch_smoke_test.py::TestRocmSdkLibraries::test_libraries_preloaded_symbol_resolvable_via_rtld_default[hipblas-hipblasCreate] PASSED [100%]
```

ScottTodd:
Looks like this failed on Windows: https://github.com/ROCm/TheRock/issues/7103. Did you test there prior to merge?

nunnikri:
> Looks like this failed on Windows: #7103. Did you test there prior to merge?

Since these smoke test are not in ci, i verified locally. But unfortunately i tested only in Linux env. 
Should we enable this smoke test in ci as well?

ScottTodd:
> Should we enable this smoke test in ci as well?

Yes, we've been wanting to run pytorch tests on CI for a long time, but doing so is not simple.

See the pinned issue in the repository:
* https://github.com/ROCm/TheRock/issues/3177
* https://github.com/ROCm/TheRock/issues/3291

ScottTodd:
https://github.com/ROCm/TheRock/actions/workflows/test_pytorch_wheels.yml can also be triggered directly for both Linux and Windows.

nunnikri:
> https://github.com/ROCm/TheRock/actions/workflows/test_pytorch_wheels.yml can also be triggered directly for both Linux and Windows.

Thats what i was also thinking. But it will add another 30min to ci time. Shall i go ahead and add this?

ScottTodd:
What do you mean by "add this"? When making changes to the tests you can run the workflows directly with `workflow_dispatch` (not using the CI entrypoint on pull requests). In the issues I linked we do want to run pytorch tests on PRs. However, we can't do that without some changes to how package are uploaded, package indexes are constructed, and the workflows install from those indexes.

nunnikri:
> What do you mean by "add this"? When making changes to the tests you can run the workflows directly with `workflow_dispatch` (not using the CI entrypoint on pull requests). In the issues I linked we do want to run pytorch tests on PRs. However, we can't do that without some changes to how package are uploaded, package indexes are constructed, and the workflows install from those indexes.

Ah.. Now its clear. Thanks for the background and time @ScottTodd 
If you want support on https://github.com/ROCm/TheRock/issues/3291 please feel free to assign activities.

PR #7062, opened 2026-08-02
fix(profiler): link correct rocSHMEM in rocprofiler-systems examples

lamikr:
Rocprofiler systems examples tried to link with
the wrong version of rocSHEMEM from /opt/rocm folder

Move the rocprofiler-systems examples declaration
after comm-libs so the rocSHMEM subproject exists
before the examples dependency is declared.

Register the package using the lowercase rocshmem
name expected by find_package(rocshmem). Disable the example when rocSHMEM is not enabled.

This prevents fallback to /opt/rocm, which caused:

[rocprofiler-systems-examples] lld: error: undefined symbol: ROCSHMEM_TEAM_WORLD
[rocprofiler-systems-examples] >>> referenced by /tmp/rocshmem-gfx1250-223d86.out.lto.o:(__clang_gpu_used_external)
[rocprofiler-systems-examples] clang++: error: amdgcn-link command failed with exit code 1 (use -v to see invocation)
[rocprofiler-systems-examples] ninja: build stopped: subcommand failed.
[rocprofiler-systems-examples FAILED WITH CODE 1 in 15 seconds]

Tested:
- Configured TheRock successfully.
- Built rocprofiler-systems-examples successfully.

## Motivation

Fixes therock build error.

## Technical Details

rocprofiles-systems-examples were build before the comm-libs
providing the rocshmem and rocshmem was also not defined as
a runtime dependency for the rocm-systems examples. This
caused linking to try to fallback to link the wrong rocshmem version 
it that was available in /opt/rocm directory. 

Sometimes this could cause a build break, sometimes it could work by luck if the
library versions are similar enough.

Fixes issue: https://github.com/ROCm/TheRock/issues/7064

## Test Plan

Test that build sucecess

## Test Result

Build works now for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -515,6 +515,7 @@ add_subdirectory(emulation)
 # Data center tools (RDC, etc.)
 add_subdirectory(dctools)
 add_subdirectory(comm-libs)
+include(profiler/rocprofiler-systems-examples.cmake)
 add_subdirectory(math-libs)
 add_subdirectory(ml-libs)
 
--- a/comm-libs/CMakeLists.txt
+++ b/comm-libs/CMakeLists.txt
@@ -184,7 +184,7 @@ if(THEROCK_ENABLE_ROCSHMEM)
     SUBDIRS
       src
   )
-  therock_cmake_subproject_provide_package(rocshmem ROCSHMEM lib/cmake/rocshmem)
+  therock_cmake_subproject_provide_package(rocshmem rocshmem lib/cmake/rocshmem)
   therock_cmake_subproject_activate(rocshmem)
   list(APPEND _rocshmem_subproject_names rocshmem)
 
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -386,49 +386,4 @@ if(THEROCK_ENABLE_ROCPROFSYS)
       rocprofiler-sdk
       rocprofiler-systems
     )
-
-
-  # TODO: This should be built as target-specific, but that's not currently possible as
-  # the artifact splitter strips per-arch HIP offload bundles out of the example binaries,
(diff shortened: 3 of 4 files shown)


marbre (changes requested):
This fix isn't sufficient. While it tries to fix the order in CMake it does neither touch the topology defined in `BUILD_TOPOLOGY.toml` nor does it correctly introduces artifact deps in download script or CI stages (`Stage Profiler Apps` runs in parallel to `Stage Comm-Libs`).

There are a few projects now which depend on RCCL or have other deps they may want to use for testing. Splitting those into their own projects is a necessary first step, like already done for `rocprofiler-systems-examples` but there are additional efforts required to get this working with the sharded multi-arch builds we're running in CI. @ScottTodd  and I discussed this earlier and next step is for sure a tracker issue to get a better overview of these additional requirements to make sure those cases are covered in the design.

ajanicijamd:
> This fix isn't sufficient. While it tries to fix the order in CMake it does neither touch the topology defined in `BUILD_TOPOLOGY.toml` nor does it correctly introduces artifact deps in download script or CI stages (`Stage Profiler Apps` runs in parallel to `Stage Comm-Libs`).
> 
> There are a few projects now which depend on RCCL or have other deps they may want to use for testing. Splitting those into their own projects is a necessary first step, like already done for `rocprofiler-systems-examples` but there are additional efforts required to get this working with the sharded multi-arch builds we're running in CI. @ScottTodd and I discussed this earlier and next step is for sure a tracker issue to get a better overview of these additional requirements to make sure those cases are covered in the design.

Nothing much to add from my side - I agree with the assessment. Adding dependency in `BUILD_TOPOLOGY.toml` is necessary in order to resolve the parallelism issue.

PR #7072, opened 2026-08-03
Fix typo in CMake foreach excluded projects

idubinov:
## Motivation

The lowercase "in" is treated as a one member of iteration list instead of cycle syntaxis.

For example:
```
foreach(project_name in ${PROJECTS})
    message("${project_name}")
endforeach()
```

will print
```
in
projA
projB
```

The patch fixes that - use upper case "IN LISTS"

https://github.com/ROCm/TheRock/issues/7121

## Submission Checklist


ScottTodd:
> ISSUE ID #42 to unlock CI
> ## Technical Details
> ## Test Plan
> ## Test Result

https://github.com/ROCm/TheRock/pull/42 looks unrelated? Can you be more specific about what this change affects?

ScottTodd (approved):
Okay, so the code was working before, but it _also_ included the string "in", which was harmless. That could have been more clear from the description and issue.

idubinov:
@ScottTodd , I am sorry that wrote not clear enough description. 
I have found such problem during another experiments, and it took me some time to debug why we have "in" target in exclude list. I have spent some time and decided to fix it even if it is not fire anywhere _yet_.

I really did not want to waste so much of somebody's time for such small change to review.

Created new issue with issue description https://github.com/ROCm/TheRock/issues/7121

PR #7078, opened 2026-08-03
Fix rocm_sdk console-scripts failing when install path has spaces

mapatel-amd:
## Motivation

On Windows, `rocm-sdk` console-scripts and the compiler wrappers fail whenever the
install path contains a space (e.g. a user folder like `C:\Users\First Last`).
Running `amdclang++` — or `rocm-sdk test` — produces:

First: binary 'First' not prefixed by 'amd'.
'C:\Users\First' is not recognized as an internal or external command

GitHub issue: https://github.com/ROCm/TheRock/issues/7075

## Technical Details

The Windows branch of the `rocm_sdk_core` trampoline (`_cli.py::_exec`) used
`os.spawnv`, which does **not** quote its arguments. A space in the executable
path caused the child process's CRT to re-split the command line, so the child
received a truncated `argv[0]` (`C:\Users\First`). The `amdllvm` wrapper
(`clang/tools/amdllvm/amdllvm.cpp`) then took the basename (`First`) and failed
its `amd`-prefix check, yielding the error above. This broke standalone
`amdclang++`, every `amdclang*` console-script, and `hipconfig`.

Replaced `os.spawnv` with `subprocess.run([str(full_path)] + sys.argv[1:])`,
which quotes the exe path and args correctly via `subprocess.list2cmdline`,
still waits for the child (preserving the `P_WAIT` semantics the old comment
cited from bugs.python.org/issue19124), and propagates the exit code.

Scope: only the `core` trampoline is touched — the `devel`/`libraries`
trampolines declare empty `console_scripts` (dead code) and `profiler` refuses
to run on Windows.

## Test Plan

On a Windows install whose path contains a space (`C:\Users\First Last\...`):

pip install --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ "rocm[libraries,devel,device-gfx942]"
rocm-sdk init
rocm-sdk test
amdclang++ --version

## Test Result

`rocm-sdk test` (`testConsoleScripts` for `amdclang`, `amdclang++`,
`amdclang-cpp`, `amdclang-cl`, `amdflang`, `amdlld`) and standalone
`amdclang++` now resolve the space-containing path correctly instead of
truncating `argv[0]`.

## Submission Checklist

*Disclosure: this change and PR description were drafted with AI assistance (Claude Code). Root cause was verified against the CPython `os.spawnv`/CRT quoting behavior and the `amdllvm.cpp` argv[0] prefix check; I have reviewed the diff and take responsibility for its correctness. *


build_tools/packaging/python/tests/rocm_sdk_core_cli_test.py
ScottTodd, line 1: These new tests are not useful.

We already have tests that run as part of `rocm-sdk test`, but we currently only run them on paths without spaces in them. We could change that.

(Prefer to look for existing tests and analyze why they missed issues before jumping to adding new test files)

build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
```diff
@@ -115,9 +115,16 @@ def _exec(relpath: str, expand_devel=True):
...
+        # background (https://bugs.python.org/issue19124), and os.spawnv, while
+        # it waits, does NOT quote its arguments: a space in the path (e.g. a
+        # user folder like "C:\Users\First Last") gets split so the child sees a
+        # truncated argv[0] (see ROCm/TheRock#7075). subprocess.run() both waits
+        # for the child and quotes the exe path and args correctly via
+        # subprocess.list2cmdline.
+        import subprocess
+
+        sys.exit(subprocess.run([str(full_path)] + sys.argv[1:]).returncode)
     os.execv(full_path, [str(full_path)] + sys.argv[1:])
```
ScottTodd, line 124: Another possible fix is to keep `os.spawnv` but use `subprocess.list2cmdline` to handle quoting:
```python
        args = subprocess.list2cmdline([str(full_path)] + sys.argv[1:])
        os._exit(os.spawnv(os.P_WAIT, full_path, [args]))
```
or 
```python
        args = [str(full_path)] + sys.argv[1:]
        quoted_args = [subprocess.list2cmdline([arg]) for arg in args]
        os._exit(os.spawnv(os.P_WAIT, full_path, quoted_args))
```
The comments here would be justification to use `subprocess.list2cmdline`, _not_ to use an entirely different method.

However, I see that `os.spawnv` has been soft deprecated on Windows, and reading https://discuss.python.org/t/how-to-deal-with-unsafe-broken-os-spawn-arg-handling-behavior-on-windows/20829, switching to `subprocess` is being recommended more and more.

One reason we use `os.execv` is that we want these CLI entry points to behave as closely as possible to native programs, so if someone runs `hipcc`, `amdclang`, or some other command from the Python venv it should appear as if they ran the native tool directly. Using subprocess keeps the parent process as Python, which could interfere with using debuggers, profilers, and other tools.

I think given the history on Windows here and limitations with `os.spawnv` and `os.execv`, switching to `subprocess.run()` is probably fine.

How about we reword the comment like so?
```python
    if is_windows:
        # Windows has no real exec() and subprocess is recommended instead.
        # os.execv runs the child in the background (https://bugs.python.org/issue19124)
        # os.spawnv has brittle argument handling (https://discuss.python.org/t/how-to-deal-with-unsafe-broken-os-spawn-arg-handling-behavior-on-windows/20829)
        import subprocess

        sys.exit(subprocess.run([str(full_path)] + sys.argv[1:]).returncode)
```

If you want to keep the reference to `ROCm/TheRock#7075`, use a full URL like `https://github.com/ROCm/TheRock/issues/7075` (I'm seeing lots of AI agents using this pattern and it's hostile to humans, since that style of referencing a github organization, repository, and issue number is not well handled by text editors and code viewers)

build_tools/packaging/python/tests/rocm_sdk_core_cli_test.py
mapatel-amd: Yea, wasn't going to add the test originally as I felt it wasn't necessary, but the PR bot was warning that there wasn't any test files so I added it. Let me get rid of this test, and I guess other issues missed this because they never had a space in them so I guess I can look into that too.

ScottTodd: Here's where we run `rocm-sdk test` today: https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_rocm_wheels.yml. Linux and Windows both use `VENV_DIR: ${{ github.workspace }}/.venv`. I could see an argument for switching that directory to one with spaces on Windows. It will look a bit funny, but the test coverage for user installs at common paths (`C:/Program Files/...`) will be really useful.

mapatel-amd: Agreed

build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
mapatel-amd: Sure let me reword the comment better, I will also look into seeing if there are any other ways without `subprocess.run()` so we can have it be more like a native tool.

build_tools/packaging/python/tests/rocm_sdk_core_cli_test.py
mapatel-amd: I implemented the case for a space in the path for windows specifically so that's done.

build_tools/packaging/python/templates/rocm-sdk-core/src/rocm_sdk_core/_cli.py
mapatel-amd: Yea I don't think there is any other way as far as I could tell, so I updated the comment as asked.

ScottTodd (approved):
Windows tests passed: https://github.com/ROCm/TheRock/actions/runs/30917910251/job/92060253845?pr=7078 with `Creating venv at 'B:\runner\_work\TheRock\TheRock\.venv with spaces'`

Linux used `Creating venv at '/__w/TheRock/TheRock/.venv'` as expected at https://github.com/ROCm/TheRock/actions/runs/30917910251/job/92060259393?pr=7078 (installs have been timing out due to some network issue on those runners)

Issue #1658, opened 2025-10-01
[Feature]: Script to set TheRock-related environment variables automatically


schung-amd:
### Suggestion Description

Similar to Visual Studio's vcvarsall, it would be convenient to have a script to set environment variables for TheRock. This would improve the developer experience on both Linux and Windows, and would smooth integration with various ML frameworks given the variety of file locations users may arrive at with different sources of TheRock (tarballs, Python wheels, build from source).

We have some documentation regarding environment variables but ideally this feature will automate and centralize the process.

ScottTodd:
By "TheRock-related environment variables" I believe we're thinking of "variables needed to use installed ROCm packages produced by TheRock". That roughly maps to any environment variables already set by system-wide ROCm installers (which we may want to provide for TheRock-based builds too).

These docs could benefit from using scripts:
* https://github.com/ROCm/TheRock/blob/main/RELEASES.md#using-installed-tarballs
* https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#using-packages-from-frameworks

janantos:
absolutely, @ScottTodd  is this sufficient or I am missing something
```
export ROCM_PATH=$(rocm-sdk path --root)
export HIP_PLATFORM=amd
export HIP_PATH=$ROCM_PATH
export HIP_CLANG_PATH=$ROCM_PATH/llvm/bin
export HIP_INCLUDE_PATH=$ROCM_PATH/include
export HIP_LIB_PATH=$ROCM_PATH/lib
export HIP_DEVICE_LIB_PATH=$ROCM_PATH/lib/llvm/amdgcn/bitcode
export PATH="$ROCM_PATH/bin:$HIP_CLANG_PATH:$PATH"
export LD_LIBRARY_PATH="$HIP_LIB_PATH:$ROCM_PATH/lib:$ROCM_PATH/lib64:$ROCM_PATH/llvm/lib:${LD_LIBRARY_PATH:-}"
export LIBRARY_PATH="$HIP_LIB_PATH:$ROCM_PATH/lib:$ROCM_PATH/lib64:${LIBRARY_PATH:-}"
export CPATH="$HIP_INCLUDE_PATH:${CPATH:-}"
export PKG_CONFIG_PATH="$ROCM_PATH/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
```

schung-amd:
[...]

PR #7082, opened 2026-08-03
feat(compiler): Enable LLVM_ENABLE_PER_TARGET_RUNTIME_DIR

estewart08:
## Motivation
Add support to use LLVM_ENABLE_PER_TARGET_RUNTIME_DIR=ON. This option is currently disabled by default in ROCm. There is talk upstream of keeping this ON by default and removing the option entirely. This also is a prequel step to using multlib build variants for debug/asan libraries for openmp/offload.

The layout change has broad ripple effects:

- Host runtimes such as libomp.so move from lib/llvm/lib/ into lib/llvm/lib/\<host-triple\>/.
- Clang sanitizer runtimes may report a per-target directory that contains only flang_rt, while compiler-rt remains under lib/linux.
- CI scripts, test harnesses, and Python wheel RPATH logic that assumed the flat layout need updating.
- Downstream consumers (hipBLASLt, rocprofiler sidecars, RPP tests, etc.) must resolve libomp.so from the new location.

| Area | Change |
|------|--------|
| **Compiler build** | `LLVM_ENABLE_PER_TARGET_RUNTIME_DIR=ON`; amd-llvm provides openmp package |
| **Sanitizer / ASAN CI** | Runtime dir derived dynamically from clang; hard fail on unresolvable launcher |
| **Test harnesses** | rocprofiler-systems, rocprofiler-sdk, and RPP tests updated for host-triple runtime paths |
| **Build deps / graph** | mxDataGenerator wired to amd-llvm in BUILD_DEPS and consumer graph |
| **Python wheels** | Host-triple dir included in wheel RPATH |

## Test Plan

Inspect current CI tests with my PR's modifications, which should not show any issues loading libraries found in lib/llvm/lib/x86_64-unknown-linux-gnu or lib/clang/23/lib/x86_64-unknown-linux-gnu. This needs to include asan builds.

ROCmValidationSuite (RVS) needs updated as well as the rvs binary uses libomp.so:
https://github.com/ROCm/ROCmValidationSuite/pull/1301

Fixes:
ISSUE ID: [#6801](https://github.com/ROCm/TheRock/issues/6801)

Partially AI Assisted


estewart08:
The patch files here will be removed later once they trickle into TheRock. I just need to test everything together.

estewart08:
Failing checks seem to be unrelated (Tests for: rocgdb-corefile, rocgdb-gpu, rocprofiler-sdk, rocr-debug-agent). Will retrigger failing checks.

build_tools/github_actions/test_executable_scripts/test_rocprofiler_systems.py
ScottTodd, line 101: Not needed as part of this PR, but we could extract some of the common `LD_LIBRARY_PATH` environment setup to a shared script handled as part of initializing the test artifacts. I've been wanting something like that for ASan setup too (a script that sets `LD_LIBRARY_PATH` and `LD_PRELOAD` by following the instructions at https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md#using-asan-instrumented-libraries)

See also:
* https://github.com/ROCm/TheRock/issues/1658

(That's more user-facing for building/running applications after installing ROCm, while this would be for running ROCm tests)

math-libs/support/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ therock_cmake_subproject_declare(mxDataGenerator
   COMPILER_TOOLCHAIN
     amd-hip
   BUILD_DEPS
+    amd-llvm
     therock-googletest
```
ScottTodd, line 19: Is this new build dep on `amd-llvm` (and the corresponding change to `test_tools/therock_consumer_graph.json`) for `mxDataGenerator` really needed? Most projects just need to depend on the `COMPILER_TOOLCHAIN`.

ScottTodd (approved):
Generally LGTM. I'm a bit worried that downstream consumers of ROCm may be depending on the current directory so they may need similar extensions to `LD_LIBRARY_PATH` to pick up the host triple path though...

math-libs/support/CMakeLists.txt
estewart08: mxDataGenerator (interface library) has target_link_libraries(mxDataGenerator INTERFACE OpenMP::OpenMP_CXX). mxDataGenerator-config.cmake then uses the openmp target for INTERFACE_LINK_LIBRARIES:

set_target_properties(mxDataGenerator PROPERTIES INTERFACE_LINK_LIBRARIES "OpenMP::OpenMP_CXX" ...

The amd-hip toolchain is not enough to resolve this openmp target in the superproject provided packages. amd-llvm subproject is updated in this patch to provide openmp via therock_cmake_subproject_provide_package(amd-llvm OpenMP lib/llvm/lib/cmake/openmp).

estewart08:
> Generally LGTM. I'm a bit worried that downstream consumers of ROCm may be depending on the current directory so they may need similar extensions to `LD_LIBRARY_PATH` to pick up the host triple path though...

At least locally, the changes to `test_rocprofiler_systems.py` to detect the new path may be a duplication now that my patch landed in rocm-systems: https://github.com/ROCm/rocm-systems/pull/7334/changes#diff-b05a7bb20c2b572f98ad07f39f62c5d6a110a3636a83b2603939c89b49ec266aR81. It bakes the path detection into `config.py`.

The change to `rocprofiler-systems` is not the same solution needed for `rpp` or `rocprofiler-sdk`. That's a possible follow up to this patch.

estewart08:
Had to rebase and fix conflict in compiler/CMakeLists.txt.

Issue #7067, opened 2026-08-03
[Issue] Jax tests failed: No FFI handler registered for `hipsolver_getrf_ffi / hipsolver_potrf_ffi` on ROCM


amd-chiranjeevi:
## Description

  39 tests in `jax/tests/scipy_stats_test.py` fail on ROCm with:
```
  jax.errors.JaxRuntimeError: NOT_FOUND: No FFI handler registered for hipsolver_getrf_ffi on a platform ROCM (canonical rocm)
  jax.errors.JaxRuntimeError: NOT_FOUND: No FFI handler registered for hipsolver_potrf_ffi on a platform ROCM (canonical rocm)
```

  The FFI handlers for HIPSolver's LU factorization (`hipsolver_getrf_ffi`) and Cholesky factorization `hipsolver_potrf_ffi`) are not being registered by `jax_rocm10_plugin` at runtime, causing any call into `lax.linalg.lu` or `lax.linalg.cholesky` to fail at  compile time.

  ## Context
- **Run**: https://github.com/ROCm/rockrel/actions/runs/30779391447/job/91587159958
-  **Release_type**:" `nightly`
- **Effected Jax Version**: ' v0.10.0, v0.10.1, v0.10.2`
- **Effected Arch:** `gfx942`

## Failing Tests (39 total)

  **`hipsolver_getrf_ffi` (LU factorization) — 24 failures:**
  - `testKde{0-9}`
  - `testKdeIntegrateBox1d{0,1}`
  - `testKdeIntegrateGaussian{0,1,2}`
  - `testKdeIntegrateKde{0,1,2}`
  - `testKdePyTree`
  - `testKdeResample1d{0,1}`
  - `testKdeResampleShape{0,1,2}`

  **`hipsolver_potrf_ffi` (Cholesky factorization) — 15 failures:**
  - `testMultivariateNormalLogpdf{2,4,5}`
  - `testMultivariateNormalLogpdfBatch{0-5}`
  - `testMultivariateNormalLogpdfBroadcasted{0-5}`

## Root Cause Analysis

JAX recently migrated LAPACK ops from the legacy custom-call API to the newer XLA FFI API.  The CUDA plugin registers `cusolver_getrf_ffi` / `cusolver_potrf_ffi` as FFI handlers.  The ROCm plugin (`jax_rocm10_plugin`) appears to be missing the equivalent registrations  for `hipsolver_getrf_ffi` / `hipsolver_potrf_ffi`.

[...]

PR #825, opened 2025-06-12
[windows] Build nightly tarballs for Windows.

ScottTodd:
Fixes https://github.com/ROCm/TheRock/issues/542. More documentation (https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-from-tarballs), script changes, and index pages for uploaded files would make these releases easier to use, but this at least gets tarballs building and uploading.

Tested here: https://github.com/ROCm/TheRock/actions/runs/15597803320. That produced https://therock-dev-tarball.s3.amazonaws.com/therock-dist-windows-gfx110X-dgpu-6.5.0.dev0%2B8fa909d3ab28e3c728d79334f6339a6ab933f787.tar.gz (4GB compressed, 20.5GB uncompressed), which I was able to extract then build PyTorch wheels with using:
```bash
cd TheRock\external-builds\pytorch
.\.venv\Scripts\activate.bat
set ROCM_HOME=D:/scratch/therock/therock-dist-windows-gfx110X-dgpu-6.5.0.dev0+8fa909d3ab28e3c728d79334f6339a6ab933f787
set PYTORCH_ROCM_ARCH=gfx1100
bash ./build_pytorch_windows.sh
```

PR #822, opened 2025-06-12
Adding "fetch_job_status.py" python script to fetch job status to be used for further CI analysis

araravik-psd:
Added a Python script to obtain the Job status and its specific step details to be used in further CI analysis and historical testcase success and failure analysis. 

Script uses Github APIs and extract job status outputs using the below URI:

"https://api.github.com/repos/RoCm/TheRock/actions/runs/{run_id}/attempts/{attempt_id}/jobs:

Changes in Workflow file to be added:

`      - name: Determine job status for workflow run
        if: '!cancelled()'
        env:
          run_id: ${{ github.run_id }}
          attempt: ${{ github.attempt }}
        run: python build_tools/github_actions/fetch_job_status.py`

PR #7084, opened 2026-08-03
ci(jax): teach the installed JAX stack about new ROCm major versions

gulsumgudukbay:
## Motivation

Plugin wheels carry the ROCm major in their package name (`jax_rocm7_plugin`, `jax_rocm10_plugin`), and both `jaxlib` and the PJRT shim look that name up from a hardcoded list of majors. On ROCm 10 the lookup misses, so the GPU kernel modules
resolve to `None` and no FFI handlers register: `'NoneType' object has no attribute ...` and `No FFI handler registered for hipsolver_*`, which is what #7067 reports.

Both sides are fixed at the source, the shim in ROCm/jax #822-#825 and `jaxlib` upstream in jax-ml/jax#39634. The gap is that `jaxlib` comes from upstream PyPI and the pinned 0.10.0-0.11.0 predate that fix, so CI and end users still need the installed copy patched.

This is the first half of the #6558 split: the ROCm 10 fixes that stand on their own. The full-suite enablement, the flake deselects and the fresh-process retry stay in #6558, since they only exist once the job runs the whole suite.

## Technical Details

- `patch_installed_jax_rocm_plugin_names.py` rewrites both lookups in the installed files between wheel install and test. It is idempotent, a no-op on ROCm 7, and detects the upstream fix so it disappears on its own once a fixed wheel ships.
- Wire the step into `test_linux_jax_wheels.yml` and `test_linux_jax_wheels_partial.yml`.
- RELEASES.md documents the same workaround where the JAX install section verifies `jax.devices()`, which is where the failure surfaces for users.

## Test Plan

Validated on the branch that also carries #6558, where the same code paths run against ~28k tests per job:

- Before the patch step, run 30561401221: all 15 jobs failed, 1468 failures each, essentially all plugin-name lookups.
- With the patch step, run 30701879601: those failures are gone.
- Full matrix green on the combined tree, runs 30781739300 and 30831999820.

## Submission Checklist

### DIFF
--- a/.github/workflows/test_linux_jax_wheels_partial.yml
+++ b/.github/workflows/test_linux_jax_wheels_partial.yml
@@ -264,6 +264,13 @@ jobs:
               "jaxlib==${JAX_VERSION}"
           fi
 
+      # jaxlib comes from upstream PyPI and only knows the ROCm majors that
+      # existed when it was released, so it has to be taught about newer ones.
+      - name: Patch installed JAX for the ROCm major version
+        run: |
+          python external-builds/jax/patch_installed_jax_rocm_plugin_names.py \
+            --plugin-package "${JAX_PLUGIN_PACKAGE}"
+
       - name: Run JAX tests
         timeout-minutes: 60
         env:
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -455,6 +455,22 @@ print(jax.devices())
 # [RocmDevice(id=0), RocmDevice(id=1), ...]
 ```
 
+> [!NOTE]
+> On ROCm 10, a released `jaxlib` (0.10.0 through 0.11.0) does not know the
+> `jax_rocm10_plugin` name, so the GPU kernel modules resolve to `None` and no FFI
+> handlers are registered. Symptoms are `'NoneType' object has no attribute ...` and
+> `No FFI handler registered for hipsolver_*`. Those versions predate the upstream fix
+> ([jax-ml/jax#39634](https://github.com/jax-ml/jax/pull/39634)); until a `jaxlib`
+> carrying it ships, teach the installed copy about the new major:
+>
+> ```bash
+> python external-builds/jax/patch_installed_jax_rocm_plugin_names.py \
+>     --plugin-package "jax_rocm${ROCM_MAJOR}_plugin"
+> ```
+>
+> The script edits the installed files in place, is idempotent, and is a no-op on ROCm 7
+> or once a fixed `jaxlib` is installed.
+
 ### Installing multi-arch native packages
 
 Native packages are installable via operating system package managers.
--- a/build_tools/github_actions/tests/patch_installed_jax_rocm_plugin_names_test.py
+++ b/build_tools/github_actions/tests/patch_installed_jax_rocm_plugin_names_test.py
@@ -1,0 +1,316 @@
+"""
+Unit tests for external-builds/jax/patch_installed_jax_rocm_plugin_names.py
+"""
+
+import importlib
+import os
+import sys
+import tempfile
+import textwrap
+import unittest
+from pathlib import Path
+
+THIS_DIR = Path(__file__).resolve().parent
+JAX_DIR = THIS_DIR.parents[2] / "external-builds" / "jax"
+
+sys.path.insert(0, os.fspath(JAX_DIR))
+
+import patch_installed_jax_rocm_plugin_names as patcher
+
+# jaxlib/plugin_support.py as released in jaxlib 0.10.0 through 0.11.0, i.e.
+# before jax-ml/jax#39634.
+JAXLIB_RELEASED = textwrap.dedent(
+    '''\
+    from collections.abc import Sequence
+    import importlib
+    import re
+
+    _PLUGIN_MODULE_NAMES = {
+        "cuda": ["jax_cuda13_plugin", "jax_cuda12_plugin"],
+        "rocm": ["jax_rocm7_plugin", "jax_rocm60_plugin"],
+        "oneapi": ["jax_oneapi_plugin"],
+    }
+
+
+    def import_from_plugin(plugin_name: str, submodule_name: str):
+      """Docstring mentioning "rocm": ["..."] should not confuse the patch."""
+      return _PLUGIN_MODULE_NAMES[plugin_name]
+    '''
+)
+
+# The same file once a jaxlib carrying jax-ml/jax#39634 ships.
+JAXLIB_FIXED = textwrap.dedent(
+    '''\
+    _PLUGIN_MODULE_NAMES = {
+        "cuda": ["jax_cuda13_plugin", "jax_cuda12_plugin"],
+        "rocm": ["jax_rocm10_plugin", "jax_rocm7_plugin", "jax_rocm60_plugin"],
+        "oneapi": ["jax_oneapi_plugin"],
+    }
+
+
+    @functools.cache
+    def _discovered_rocm_plugin_module_names() -> tuple[str, ...]:
+      """Finds installed ROCm plugin packages not listed above."""
+      return ()
+    '''
+)
+
+# jax_plugins/rocm/__init__.py before the ROCm/jax fix, installed as
+# jax_plugins/xla_rocm<major>/__init__.py.
+SHIM_RELEASED = textwrap.dedent(
+    """\
+    import importlib
+    import jax._src.xla_bridge as xb
+
+    # rocm_plugin_extension locates inside jaxlib. `jaxlib` is for testing without
+    # preinstalled jax rocm plugin packages.
+    for pkg_name in ['jax_rocm7_plugin', 'jax_rocm60_plugin', 'jaxlib.rocm']:
+      try:
+        rocm_plugin_extension = importlib.import_module(
+            f'{pkg_name}.rocm_plugin_extension'
+        )
+      except ImportError:
+        rocm_plugin_extension = None
+      else:
+        break
+    """
+)
+
+# The same file with the ROCm/jax fix, which every rocm-jaxlib-v0.10.x and
+# v0.11.0 branch now carries.
+SHIM_FIXED = textwrap.dedent(
+    """\
+    import importlib
+
+    _pkg_names = [
+        'jax_rocm10_plugin', 'jax_rocm7_plugin', 'jax_rocm60_plugin', 'jaxlib.rocm'
+    ]
+
+    _rocm_major = (__package__ or '').rpartition('xla_rocm')[2]
+    if _rocm_major.isdigit() and f'jax_rocm{_rocm_major}_plugin' not in _pkg_names:
+      _pkg_names.insert(0, f'jax_rocm{_rocm_major}_plugin')
+
+    for pkg_name in _pkg_names:
+      try:
+        rocm_plugin_extension = importlib.import_module(
+            f'{pkg_name}.rocm_plugin_extension'
+        )
+      except ImportError:
+        rocm_plugin_extension = None
+      else:
+        break
+    """
+)
+
+
+class RewriteJaxlibTest(unittest.TestCase):
+    def test_plugin_is_inserted_first(self) -> None:
+        result = patcher.rewrite_jaxlib_plugin_names(
+            JAXLIB_RELEASED, "jax_rocm10_plugin"
+        )
+
+        self.assertIn(
+            '"rocm": ["jax_rocm10_plugin", "jax_rocm7_plugin", "jax_rocm60_plugin"]',
+            result.text,
+        )
+        self.assertIn("jax_rocm10_plugin", result.detail)
+
+    def test_other_platforms_are_untouched(self) -> None:
+        result = patcher.rewrite_jaxlib_plugin_names(
+            JAXLIB_RELEASED, "jax_rocm10_plugin"
+        )
+
+        self.assertIn('"cuda": ["jax_cuda13_plugin", "jax_cuda12_plugin"]', result.text)
+        self.assertIn('"oneapi": ["jax_oneapi_plugin"]', result.text)
+
+    def test_result_is_valid_python(self) -> None:
+        result = patcher.rewrite_jaxlib_plugin_names(
+            JAXLIB_RELEASED, "jax_rocm10_plugin"
+        )
+
+        compile(result.text, "plugin_support.py", "exec")
+
+    def test_wrapped_list_is_rewritten(self) -> None:
+        wrapped = textwrap.dedent(
+            """\
+            _PLUGIN_MODULE_NAMES = {
+                "rocm": [
+                    "jax_rocm7_plugin",
+                    "jax_rocm60_plugin",
+                ],
+            }
+            """
+        )
+
+        result = patcher.rewrite_jaxlib_plugin_names(wrapped, "jax_rocm10_plugin")
+
+        self.assertIn('"jax_rocm10_plugin", "jax_rocm7_plugin"', result.text)
+        compile(result.text, "plugin_support.py", "exec")
+
+    def test_rewrite_is_idempotent(self) -> None:
+        once = patcher.rewrite_jaxlib_plugin_names(JAXLIB_RELEASED, "jax_rocm10_plugin")
+        twice = patcher.rewrite_jaxlib_plugin_names(once.text, "jax_rocm10_plugin")
+
+        self.assertIsNone(twice.text)
+        self.assertIn("already lists jax_rocm10_plugin", twice.detail)
+
+    def test_fixed_jaxlib_is_left_alone(self) -> None:
+        result = patcher.rewrite_jaxlib_plugin_names(JAXLIB_FIXED, "jax_rocm99_plugin")
+
+        self.assertIsNone(result.text)
+        self.assertIn("discovers", result.detail)
+
+    def test_unexpected_layout_raises(self) -> None:
+        with self.assertRaisesRegex(ValueError, "plugin list"):
+            patcher.rewrite_jaxlib_plugin_names(
+                "_PLUGIN_MODULE_NAMES = load_plugin_names()\n", "jax_rocm10_plugin"
+            )
+
+
+class RewritePjrtShimTest(unittest.TestCase):
+    def test_plugin_is_inserted_first(self) -> None:
+        result = patcher.rewrite_pjrt_shim_plugin_names(
+            SHIM_RELEASED, "jax_rocm10_plugin"
+        )
+
+        self.assertIn(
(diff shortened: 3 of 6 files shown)


.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -281,6 +288,11 @@ jobs:
           DEBUG_HIP_DYNAMIC_QUEUES: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '0' }}
           HSA_NO_SCRATCH_RECLAIM: ${{ inputs.jax_ref == 'rocm-jaxlib-v0.10.2' && '1' }}
 
+          # On gfx950 MIOpen's Winograd assembly does not build under the
+          # ROCm 10 comgr, so every convolution fails to autotune. The other
+          # solvers work, so drop Winograd until MIOpen assembles again.
+          MIOPEN_DEBUG_CONV_WINOGRAD: "0"
```
erman-gurses: Can you scope this workaround to gfx950 on ROCm 10?

It is currently exported for every GPU family and ROCm version handled by this workflow, although the documented failure is specific to gfx950 with ROCm 10 comgr. I think this unnecessarily disabling Winograd coverage on configurations where it is healthy.

RELEASES.md
```diff
@@ -455,6 +455,22 @@ print(jax.devices())
...
+> carrying it ships, teach the installed copy about the new major:
+>
+> ```bash
+> python build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py \
+>     --plugin-package "jax_rocm${ROCM_MAJOR}_plugin"
+> ```
+>
+> The script edits the installed files in place, is idempotent, and is a no-op on ROCm 7
+> or once a fixed `jaxlib` is installed.
+
```
erman-gurses, line 473: I know the struggle about to maintain the compatibility but I am asking just to make sure about it. Could we avoid making this an additional required step for end users?

This command directly modifies the installed jaxlib after the published wheels have been installed. That means the documented ROCm 10 JAX installation is not functional from pip install alone, and every user must obtain this repository script and rerun it for each environment or reinstall.

Could this compatibility fix instead be applied to the wheel before publication, or otherwise incorporated into the supported installation path? Keeping the script as a temporary CI workaround seems reasonable, but requiring users to patch files under site-packages does not seem like a sustainable release workflow.

external-builds/jax/patch_installed_jax_rocm_plugin_names.py
```diff
@@ -0,0 +1,164 @@
...
+_PLUGIN_PACKAGE_RE = re.compile(r"^jax_rocm(\d+)_plugin$")
+
+# _PLUGIN_MODULE_NAMES = {..., "rocm": ["jax_rocm7_plugin", ...], ...}
+_JAXLIB_ROCM_LIST_RE = re.compile(r'("rocm"\s*:\s*\[)([^\]]*?)(\s*\])', re.DOTALL)
+
+# for pkg_name in ['jax_rocm7_plugin', 'jax_rocm60_plugin', 'jaxlib.rocm']:
+_SHIM_LIST_RE = re.compile(r"(for pkg_name in \[)([^\]]*?)(\]\s*:)", re.DOTALL)
+
+# Marker for the ROCm/jax fix that derives the major from the package name.
+_SHIM_FIXED_MARKER = "rpartition('xla_rocm')"
```
erman-gurses, line 55: Could we add focused unit tests for these matching and transformation assumptions?

This script relies on a package-name regex, two layout-sensitive source regexes, and a specific marker for detecting the upstream shim fix, but the PR does not add direct coverage for them.

At minimum, we could test valid and invalid package names, the old jaxlib and shim formats, running the script twice, already-fixed shims, and unexpected file formats?

build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py
```diff
@@ -0,0 +1,164 @@
...
+    if importlib.util.find_spec(args.plugin_package) is None:
+        raise FileNotFoundError(
+            f"'{args.plugin_package}' is not installed, so there is nothing to"
+            " point the installed JAX stack at"
+        )
+
+    print(f"Patching the installed JAX stack for {args.plugin_package}:")
+    patched = patch_jaxlib_plugin_names(args.plugin_package)
+    if not args.skip_pjrt_shim:
+        patched |= patch_pjrt_shim(args.plugin_package, rocm_major)
```
erman-gurses: Please avoid leaving the installed environment partially patched.

patch_jaxlib_plugin_names() updates jaxlib before the PJRT shim is checked. If patch_pjrt_shim() then fails because the shim is missing or has an unexpected format, the command exits after only the first file has been modified.

erman-gurses (dismissed):

gulsumgudukbay:
Release matrix all green: https://github.com/ROCm/TheRock/actions/runs/30909284834

external-builds/jax/patch_installed_jax_rocm_plugin_names.py
ScottTodd, line 1: Copying a comment from https://github.com/ROCm/TheRock/pull/6558#discussion_r3715041410

> I'd rather this file go in https://github.com/ROCm/TheRock/tree/main/external-builds/jax , similar to https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/windows_patch_fat_wheel.py. The `build_tools/github_actions/` folder is for code that we only expect to run on CI, but this file is referenced by `RELEASES.md` that users will see.
> 
> Then you could document it a bit in https://github.com/ROCm/TheRock/blob/main/external-builds/jax/README.md, similar to https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/README.md#bundling-pytorch-and-rocm-together-into-a-fat-wheel

RELEASES.md
ScottTodd: Copying a comment from https://github.com/ROCm/TheRock/pull/6558#discussion_r3715031403

> I'm not thrilled by this, but it seems like an acceptable workaround while we wait for upstream. In the longer term I'd prefer if we weren't dependent on upstream to make the packages we distribute functional. That may require building packages ourselves or at least mirroring them into our index with a patching/repackaging step.

I think the technical and user-centric concerns have been heard and adequately addressed, so this temporary solution is okay for now.

external-builds/jax/patch_installed_jax_rocm_plugin_names.py
ScottTodd: Yeah tests would be nice given the complexity here.

If the file moves to `external-builds/jax/`, you could keep the test file in `build_tools/github_actions/tests/` so we run it on CI as part of https://github.com/ROCm/TheRock/blob/b070c6f38860cec6e28f37e1c4bb66c6002146e1/.github/workflows/unit_tests.yml#L40-L49

That's currently awkward due to how test files are scattered throughout the project. Here's one example to reference:

* Code in `external-builds/`: https://github.com/ROCm/TheRock/blob/main/external-builds/pytorch/checkout_from_manifest.py
* Test in `build_tools/github_actions/`: https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/pytorch_checkout_from_manifest_test.py

ScottTodd (commented):
Generally LGTM ~ the file location and Erman's other comments.

.github/workflows/test_linux_jax_wheels_partial.yml
gulsumgudukbay: Scoped, and I verified the premise first: run 30646054646 had 99 deterministic conv failures on gfx950 with MIOpen's assembler rejecting .amdhsa_reserve_xnack_mask does not match target id, while the gfx942 runs of the same suite were clean. So the workaround now comes from a step gated on gfx950 plus ROCm major 10, and MIOpen keeps its defaults everywhere else. I set it through GITHUB_ENV rather than a ternary in env: so that on other configurations the variable is absent rather than empty, since an empty value is not obviously "enabled" to MIOpen.

RELEASES.md
gulsumgudukbay: > I know the struggle about to maintain the compatibility but I am asking just to make sure about it. Could we avoid making this an additional required step for end users?
> 
> This command directly modifies the installed jaxlib after the published wheels have been installed. That means the documented ROCm 10 JAX installation is not functional from pip install alone, and every user must obtain this repository script and rerun it for each environment or reinstall.
> 
> Could this compatibility fix instead be applied to the wheel before publication, or otherwise incorporated into the supported installation path? Keeping the script as a temporary CI workaround seems reasonable, but requiring users to patch files under site-packages does not seem like a sustainable release workflow.

The lookup that breaks is in upstream jaxlib, which comes from PyPI, we neither build nor publish it, so there is no wheel of ours to fix before publication. The half that is ours, the PJRT shim, is already fixed at the source: every rocm-jaxlib-v0.10.x and v0.11.0 branch derives the plugin name from the ROCm major, so the wheels we publish need no patching.

jaxlib is fixed upstream in [jax-ml/jax#39634](https://github.com/jax-ml/jax/pull/39634) (merged Jul 31), where it discovers installed plugin packages instead of matching a hardcoded list. It simply is not in a release yet; 0.11.0 is still the newest on PyPI. Until that release lands, a released jaxlib on ROCm 10 cannot find jax_rocm10_plugin, and I would rather the release notes say so than have users hit 'NoneType' object has no attribute ... with no explanation.

The step is temporary by construction: the script is idempotent, a no-op on ROCm 7, and detects the upstream fix, so it does nothing once a fixed jaxlib is installed and the note comes out with it. The way to avoid it entirely would be publishing our own jaxlib, which we do not want, the point of last year's work upstreaming ROCm CI into JAX was to ship against upstream jaxlib rather than a fork.

build_tools/github_actions/patch_installed_jax_rocm_plugin_names.py
gulsumgudukbay: Fixed. Both files are planned before either is written: the rewrite_* functions return the new contents, and main() writes only after every plan has succeeded, so a layout the patch does not recognize raises with the installation untouched. Two tests cover it from both directions: a broken shim leaves jaxlib unpatched, and a broken jaxlib leaves the shim unpatched.

external-builds/jax/patch_installed_jax_rocm_plugin_names.py
gulsumgudukbay: > Could we add focused unit tests for these matching and transformation assumptions?
> 
> This script relies on a package-name regex, two layout-sensitive source regexes, and a specific marker for detecting the upstream shim fix, but the PR does not add direct coverage for them.
> 
> At minimum, we could test valid and invalid package names, the old jaxlib and shim formats, running the script twice, already-fixed shims, and unexpected file formats?

Added build_tools/github_actions/tests/patch_installed_jax_rocm_plugin_names_test.py, 18 tests. It covers valid and invalid package names (invalid exits before anything is touched), the released jaxlib and pre-fix shim layouts, wrapped list literals, that the rewritten text still compiles as Python, running twice, both upstream-fix markers, unrecognized layouts, --skip-pjrt-shim, and a missing plugin package. The fixtures are the actual file contents from the jaxlib 0.11.0 wheel and from ROCm/jax before the shim fix, so the regexes are tested against real layouts rather than invented ones.

build_tools/github_actions/tests/patch_installed_jax_rocm_plugin_names_test.py
```diff
@@ -0,0 +1,315 @@
...
+        twice = patcher.rewrite_jaxlib_plugin_names(once.text, "jax_rocm10_plugin")
+
+        self.assertIsNone(twice.text)
+        self.assertIn("already lists jax_rocm10_plugin", twice.detail)
+
+    def test_fixed_jaxlib_is_left_alone(self) -> None:
+        result = patcher.rewrite_jaxlib_plugin_names(JAXLIB_FIXED, "jax_rocm11_plugin")
+
+        self.assertIsNone(result.text)
+        self.assertIn("discovers", result.detail)
```
ScottTodd, line 161: FYI `rocm11` may trigger some security scans that watch for mentions of unannounced versions. We might want to have the test use a more obviously "test" version like `99`, `123`, `12345`, `1000`, etc.

ScottTodd (commented):
The unit tests LGTM, thanks

external-builds/jax/patch_installed_jax_rocm_plugin_names.py
gulsumgudukbay: Moved to [external-builds/jax/patch_installed_jax_rocm_plugin_names.py](https://github.com/ROCm/TheRock/blob/main/external-builds/jax/patch_installed_jax_rocm_plugin_names.py) and documented it in the JAX README under the test instructions, with the workflows and RELEASES.md pointing at the new path. The tests stayed in build_tools/github_actions/tests/ so pytest build_tools still runs them, following the checkout_from_manifest.py precedent you linked.

build_tools/github_actions/tests/patch_installed_jax_rocm_plugin_names_test.py
gulsumgudukbay: thanks, I switched to jax_rocm99_plugin. It only appeared in the tests that assert the patch leaves an already-fixed jaxlib or shim alone and that a missing plugin package is an error, so a nonsense major reads better there anyway. ROCm 10 is now the only real version named in the PR.

.github/workflows/test_linux_jax_wheels_partial.yml
```diff
@@ -264,6 +265,23 @@ jobs:
...
+
+      # Only gfx950 on ROCm 10: MIOpen's Winograd assembly does not build under
+      # the ROCm 10 comgr for that target (".amdhsa_reserve_xnack_mask does not
+      # match target id"), so every convolution fails to autotune. The other
+      # solvers work, so drop Winograd here and leave MIOpen at its defaults
+      # everywhere else, where Winograd is healthy.
+      - name: Disable MIOpen Winograd
+        if: startsWith(inputs.test_amdgpu_family, 'gfx950') && env.ROCM_MAJOR == '10'
+        run: |
+          echo "MIOPEN_DEBUG_CONV_WINOGRAD=0" >> "$GITHUB_ENV"
```
ScottTodd: I'm not a fan of this method of filtering:
* The ROCm major version covers a large span of versions and will apply even after the root cause is fixed in source
* Are we confident that MIOpen's Winograd assembly works on both old ROCm releases and with the latest releases / source code for non-gfx950 targets?

I'd rather file an issue against MIOpen in https://github.com/ROCm/rocm-libraries explaining what is failing where, linking that issue here, and disabling individual test cases that are known failing like here: https://github.com/ROCm/TheRock/blob/b070c6f38860cec6e28f37e1c4bb66c6002146e1/external-builds/pytorch/skip_tests/generic.py#L235-L237

I'm not sure what the implications of testing with this setting on only gfx950 would be, please involve the MIOpen team (cc @JonathanLichtnerAMD / @BrianHarrisonAMD maybe one of you could help send us in the right direction?)

gulsumgudukbay: Both fair points, and this PR turns out to be the wrong place for the workaround: the JAX jobs here only run multi_device_test.py, core_test.py, util_test.py and scipy_stats_test.py, none of which contain a single convolution call site, so MIOPEN_DEBUG_CONV_WINOGRAD has no effect on anything this PR tests. It matters only in the full-suite job in #6558, where the conv tests run and where the 99 failures showed up. I'll drop the commit here and take the discussion to #6558.

On your second question, I don't have data to make that claim. The only positive evidence is gfx942 on this ROCm 10 build, where the same suite ran clean; I have nothing for other families or older ROCm. The gate did leave every other family and major at MIOpen's defaults, so nothing outside gfx950 on ROCm 10 was affected either way, but your first point stands regardless: it would keep applying across all of ROCm 10, including after a fix lands, and nothing in the workflow can tell when that happens.

I'll file the issue in ROCm/rocm-libraries with the assembler output (.amdhsa_reserve_xnack_mask does not match target id and unknown directive .amdhsa_ieee_mode out of amd_comgr_do_action, from run [30646054646](https://github.com/ROCm/TheRock/actions/runs/30646054646/)) and link it. For the mitigation in #6558, the trade-off is worth flagging for the MIOpen folks: deselecting the conv tests took out roughly 636 tests, whereas disabling the Winograd solvers kept 392 of them running through MIOpen's other solvers, with a single straggler that passes on rerun. So deselects test MIOpen's default configuration but drop the conv surface entirely on gfx950, while the env var keeps conv coverage at the cost of not exercising Winograd there. A third option that answers your "applies even after the fix" concern is a preflight conv probe that only disables Winograd when the failure signature is present, so it disappears by itself once MIOpen assembles again. Happy to go whichever way you and the MIOpen team prefer.

gulsumgudukbay:
Removed the WINOGRAD env var and restarted the workflow: https://github.com/ROCm/TheRock/actions/runs/30945474437
@ScottTodd  @erman-gurses

ScottTodd (approved):

erman-gurses (approved):
LGTM - Assuming Release Workflow tests are green

.github/workflows/test_linux_jax_wheels_partial.yml
BrianHarrisonAMD: I think there was issues with winograd in gfx950, but I thought those have been fixed.
If you are still seeing it in a recent version, then please provide replication details (via an issue, or ticket), and it'll need to be prioritized + fixed.

---

If you need to temporarily disable certain solvers to unblock your CI, or to move forward with some upgrade, then that's fine, but it should be associated with an issue tracking it, and a comment so we can revisit the workaround once the fix lands.

gulsumgudukbay: @BrianHarrisonAMD I retested on a current build and you're right (I previously tested on a Jul 31 TheRock build, MIOpen 3.6.0 build bb8b1d6c), it's fixed, so no ticket needed. 
In my experiment I had a different case in its own process with a fresh MIOpen user DB: on ROCm 10.0.0.dev0+4b50725 (rocm-libraries bb8b1d6c, MIOpen 3.6.0) fp32 3x3 stride1/stride2, 5x5, 1x1 and fp16 3x3 all fail with .amdhsa_reserve_xnack_mask does not match target id on Conv_Winograd_v30_3_1_*. 
However on ROCm 10.1.0.dev0+0888ca35 (rocm-libraries 9f3670b4) all of them pass. 

The only MIOpen Winograd/gfx950 change in the 52 commits between those is #9014 (Winograd Rage v4_6_1/v4_9_1, enable gfx950, merged Jul 31), so that looks like the fix, though I didn't bisect. I dropped the MIOPEN_DEBUG_CONV_WINOGRAD workaround.

PR #7087, opened 2026-08-04
ci(jax): group the JAX release jobs by release instead of calling them all builds

gulsumgudukbay:
## Motivation

Every job of a called workflow is displayed under the name of the job that called it, so a JAX release matrix cell said "Build" over its own tests, and the Python version and the jax ref were repeated at all three levels. Name the cell for the release it covers and let the jobs inside say what they are, matching how the PyTorch jobs read.

Cosmetic only; peeled out of #6558 so that PR stays about test behavior.

## Technical Details

- Matrix cell becomes `Release | py 3.12 | JAX 0.10.0`, covering both the build and the tests underneath it.
- The called jobs drop the version and ref they inherit from the cell heading and say what they are: `Test | gfx94X-dcgpu | <runner>` and `Test JAX | gfx94X-dcgpu`.
- `configure_jax_release_matrix.py` adds a `jax_label` field, the ref without its `rocm-jaxlib-v` prefix, because a workflow expression cannot strip it.

## Test Plan

- `configure_jax_release_matrix_test.py` covers the new field.
- Job naming as rendered in run 30781739300.

JIRA ID: AIJAX-313

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_build_linux_jax_wheels.yml
+++ b/.github/workflows/multi_arch_build_linux_jax_wheels.yml
@@ -276,7 +276,7 @@ jobs:
   test_jax_wheels:
     if: ${{ inputs.test_amdgpu_family != '' }}
     needs: [build_jax_wheels, generate_target_to_run]
-    name: Test JAX wheels | py ${{ inputs.python_version }} | jax ${{ inputs.jax_ref }}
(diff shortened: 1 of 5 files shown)


ScottTodd (approved):

erman-gurses (approved):

gulsumgudukbay:
https://github.com/ROCm/TheRock/actions/runs/30974266602 Passed

PR #7090, opened 2026-08-04
fix: support fetching xnack-suffixed artifacts

raramakr:
Add base-arch matching to find_available_artifacts() so requesting a
  base architecture (e.g., gfx942) also matches xnack variants
  (gfx942:xnack+, gfx942:xnack-).

  Changes:
  - Add _get_base_arch() to strip xnack suffixes from targets
  - Add _matches_target() for base-arch comparison
  - Refactor find_available_artifacts() to iterate available artifacts and filter using ArtifactName parsing (like fetch_artifacts.py)
  - Add _sort_key to preserve .tar.zst preference over .tar.xz

  This enables --expand-family-to-targets to correctly fetch artifacts
  for ASAN builds which use xnack-suffixed targets.

JIRA ID: https://amd-hub.atlassian.net/browse/ROCM-28314

### DIFF
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -77,7 +77,11 @@ ARTIFACT_COMPONENTS = ["lib", "run", "dev", "dbg", "doc", "test"]
 
 
 def _get_base_arch(target: str) -> str:
-    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'."""
+    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'.
+
+    Note: This strips everything after the first colon. Any suffix (not just
+    xnack) will be removed, e.g., 'gfx942:foo' -> 'gfx942'.
+    """
     if not target:
         return ""
     base = target.split(":")[0]
@@ -85,7 +89,17 @@ def _get_base_arch(target: str) -> str:
 
 
 def _matches_target(artifact_target: str, requested_targets: set[str]) -> bool:
-    """Match if the artifact's base arch equals any requested target's base arch."""
+    """Match if the artifact's base arch equals any requested target's base arch.
+
+    Uses base-arch matching: both the artifact target and requested targets are
+    stripped to their base arch before comparison. This means requesting 'gfx942'
+    will match 'gfx942', 'gfx942:xnack+', 'gfx942:xnack-', or any 'gfx942:*' variant.
+
+    Known limitation: All colon-suffixed variants of the same base arch will match.
+    For example, requesting 'gfx942' matches 'gfx942:xnack+', 'gfx942:xnack-',
+    'gfx942:abc', etc. This is intentional to ensure all relevant artifacts are
+    fetched for a given GPU architecture.
+    """
     if not artifact_target:
         return False
     requested_bases = {_get_base_arch(t) for t in requested_targets}
@@ -164,7 +178,8 @@ def find_available_artifacts(
     and components. Uses base-arch matching: requesting a base arch (e.g., "gfx942")
     will also match variants with suffixes (e.g., "gfx942:xnack+", "gfx942:xnack-").
 
-    Prefers .tar.zst over .tar.xz when both exist.
+    Prefers .tar.zst over .tar.xz when both exist, because zstd offers faster
+    decompression and better compression ratios for our artifact workloads.
     """
     excluded_components = excluded_components or set()
     targets_to_match = set(target_families)
@@ -173,7 +188,7 @@ def find_available_artifacts(
     matched = []
     seen = set()
 
-    # Sort with .tar.zst before .tar.xz so dedup prefers zstd
+    # Sort .tar.zst before .tar.xz so dedup prefers zstd (faster decompression)
     def _sort_key(f: str) -> tuple:
         base = f.rsplit(".tar", 1)[0]
         priority = 0 if f.endswith(".tar.zst") else 1
--- a/build_tools/artifact_manager.py
+++ b/build_tools/artifact_manager.py
@@ -76,6 +76,22 @@ from _therock_utils.workflow_outputs import WorkflowOutputRoot
 ARTIFACT_COMPONENTS = ["lib", "run", "dev", "dbg", "doc", "test"]
 
 
+def _get_base_arch(target: str) -> str:
+    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'."""
+    if not target:
+        return ""
+    base = target.split(":")[0]
+    return base if base else target
+
+
+def _matches_target(artifact_target: str, requested_targets: set[str]) -> bool:
+    """Match if the artifact's base arch equals any requested target's base arch."""
+    if not artifact_target:
+        return False
+    requested_bases = {_get_base_arch(t) for t in requested_targets}
+    return _get_base_arch(artifact_target) in requested_bases
+
+
 def log(msg: str):
     """Print message and flush."""
     print(msg, flush=True)
@@ -144,22 +160,43 @@ def find_available_artifacts(
 ) -> list[str]:
     """Find which artifacts exist in the available set.
 
-    Iterates artifact_names × target_families × components × extensions,
-    returning filenames that are present in `available`. Prefers .tar.zst
-    over .tar.xz when both exist.
+    Iterates available artifacts and filters by artifact_names, target_families,
+    and components. Uses base-arch matching: requesting a base arch (e.g., "gfx942")
+    will also match variants with suffixes (e.g., "gfx942:xnack+", "gfx942:xnack-").
+
+    Prefers .tar.zst over .tar.xz when both exist.
     """
     excluded_components = excluded_components or set()
+    targets_to_match = set(target_families)
+
+    # Use structured ArtifactName parsing for reliable matching (like fetch_artifacts.py)
     matched = []
-    for artifact_name in sorted(artifact_names):
-        for tf in target_families:
-            for comp in ARTIFACT_COMPONENTS:
-                if comp in excluded_components:
-                    continue
-                for ext in ARTIFACT_EXTENSIONS:
-                    filename = f"{artifact_name}_{comp}_{tf}{ext}"
-                    if filename in available:
-                        matched.append(filename)
-                        break  # Found this artifact, don't check other extensions
+    seen = set()
+
+    # Sort with .tar.zst before .tar.xz so dedup prefers zstd
+    def _sort_key(f: str) -> tuple:
+        base = f.rsplit(".tar", 1)[0]
+        priority = 0 if f.endswith(".tar.zst") else 1
+        return (base, priority)
+
+    for filename in sorted(available, key=_sort_key):
+        an = ArtifactName.from_filename(filename)
+        if not an:
+            continue
+        if an.name not in artifact_names:
+            continue
+        if an.component in excluded_components:
+            continue
+        if not _matches_target(an.target_family, targets_to_match):
+            continue
+
+        # Dedupe: prefer .tar.zst over .tar.xz for same artifact
+        key = (an.name, an.component, an.target_family)
+        if key in seen:
+            continue
+        seen.add(key)
+        matched.append(filename)
+
     return matched
 
 
--- a/build_tools/tests/artifact_manager_tool_test.py
+++ b/build_tools/tests/artifact_manager_tool_test.py
@@ -593,12 +593,13 @@ class TestFetchStageAll(ArtifactManagerTestBase):
             )
 
     def test_fetch_exclude_components_skips_test_artifacts(self) -> None:
-        """Test that --exclude-components filters artifact components."""
+        """Test that --exclude-components filters artifact components including xnack variants."""
         import artifact_manager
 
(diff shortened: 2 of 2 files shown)


build_tools/artifact_manager.py
```diff
@@ -76,6 +76,22 @@
 ARTIFACT_COMPONENTS = ["lib", "run", "dev", "dbg", "doc", "test"]
 
 
+def _get_base_arch(target: str) -> str:
```
nunnikri, line 79: Looks like these helper fns are already there in fetch_artifacts.py. But good to add unit testing here as well.

build_tools/artifact_manager.py
```diff
@@ -144,22 +160,43 @@ def find_available_artifacts(
...
     """Find which artifacts exist in the available set.
 
-    Iterates artifact_names × target_families × components × extensions,
-    returning filenames that are present in `available`. Prefers .tar.zst
-    over .tar.xz when both exist.
+    Iterates available artifacts and filters by artifact_names, target_families,
+    and components. Uses base-arch matching: requesting a base arch (e.g., "gfx942")
+    will also match variants with suffixes (e.g., "gfx942:xnack+", "gfx942:xnack-").
+
+    Prefers .tar.zst over .tar.xz when both exist.
```
nunnikri: Why zst is preferred better to explain in the comments

build_tools/artifact_manager.py
```diff
@@ -144,22 +160,43 @@ def find_available_artifacts(
...
-                if comp in excluded_components:
-                    continue
-                for ext in ARTIFACT_EXTENSIONS:
-                    filename = f"{artifact_name}_{comp}_{tf}{ext}"
-                    if filename in available:
-                        matched.append(filename)
-                        break  # Found this artifact, don't check other extensions
+    seen = set()
+
+    # Sort with .tar.zst before .tar.xz so dedup prefers zstd
```
nunnikri: The comments are not very helpful to explain the logic.

build_tools/artifact_manager.py
```diff
@@ -144,22 +160,43 @@ def find_available_artifacts(
...
+            continue
+        if not _matches_target(an.target_family, targets_to_match):
+            continue
+
+        # Dedupe: prefer .tar.zst over .tar.xz for same artifact
+        key = (an.name, an.component, an.target_family)
+        if key in seen:
+            continue
+        seen.add(key)
+        matched.append(filename)
```
nunnikri, line 213: Any entry matching the base name. gfx942:xnack+, gfx942:xnack-, gfx942:abc, gfx942:abc:xnack+ all will get matched here. 
Which might be ok here. But good to note as a limitation

nunnikri (approved):
Overall LGTM. minor docstring comments, please take care before merging

build_tools/artifact_manager.py
ScottTodd, line 1: Add tests in https://github.com/ROCm/TheRock/blob/main/build_tools/tests/artifact_manager_tool_test.py

ScottTodd (changes requested):

build_tools/artifact_manager.py
raramakr: Addressed the same

build_tools/artifact_manager.py
raramakr: Done

build_tools/artifact_manager.py
raramakr: Done

build_tools/artifact_manager.py
raramakr: Addressed the same

build_tools/artifact_manager.py
raramakr: Done

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -1490,5 +1490,99 @@ def test_explicit_targets_still_added(self):
...
+        self.assertEqual(self.am._get_base_arch("gfx942"), "gfx942")
+        self.assertEqual(self.am._get_base_arch("gfx942:xnack+"), "gfx942")
+        self.assertEqual(self.am._get_base_arch("gfx950:xnack-"), "gfx950")
+
+    def testMatchesTarget_HandlesEdgeCases(self):
+        """Test _matches_target with empty and various inputs."""
+        requested = {"generic", "gfx942"}
+        self.assertFalse(self.am._matches_target("", requested))
+        self.assertTrue(self.am._matches_target("gfx942", requested))
+        self.assertTrue(self.am._matches_target("gfx942:xnack+", requested))
```
ScottTodd: I would rather unit tests only exercise the public API and not private helper functions (`_` prefix).

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -1490,5 +1490,99 @@ def test_explicit_targets_still_added(self):
...
+            artifact_names={"blas", "rccl"},
+            target_families=["generic", "gfx942"],
+            available=available,
+        )
+        self.assertIn("blas_lib_generic.tar.zst", result)
+        self.assertIn("blas_lib_gfx942.tar.zst", result)
+        self.assertIn("blas_test_gfx942.tar.zst", result)
+        self.assertIn("rccl_test_gfx942:xnack+.tar.zst", result)
+        self.assertIn("rccl_lib_gfx942:xnack-.tar.zst", result)
+        self.assertNotIn("blas_lib_gfx950.tar.zst", result)
```
ScottTodd: I could see a case for merging this test case into other tests that push or pull artifacts with `artifact_manager.py push`, rather than test the `find_available_artifacts` function.

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -1490,5 +1490,99 @@ def test_explicit_targets_still_added(self):
...
+        available = {
+            "fft_lib_gfx942.tar.zst",
+            "fft_lib_gfx942.tar.xz",
+        }
+        result = self.am.find_available_artifacts(
+            artifact_names={"fft"},
+            target_families=["gfx942"],
+            available=available,
+        )
+        self.assertEqual(result, ["fft_lib_gfx942.tar.zst"])
```
ScottTodd: Why is this test case under `XnackArtifactMatchingTest`? There is no xnack code in here.

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -1490,5 +1490,99 @@ def test_explicit_targets_still_added(self):
...
+    """Unit tests for xnack-suffixed artifact matching in artifact_manager.
+
+    These tests mirror the tests in fetch_artifacts_test.py since the
+    _get_base_arch and _matches_target functions are duplicated.
+    """
+
+    def setUp(self):
+        import artifact_manager as am
+
+        self.am = am
```
ScottTodd: I think we should (separate from this PR) make a cleanup pass through this file a nd move these imports to the top. There are 27 calls to `import artifact_manager` in this file already.

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -1490,5 +1490,99 @@ def test_explicit_targets_still_added(self):
...
+            "fft_test_gfx942.tar.zst",
+            "fft_test_gfx942:xnack+.tar.zst",
+        }
+        result = self.am.find_available_artifacts(
+            artifact_names={"fft"},
+            target_families=["gfx942"],
+            available=available,
+            excluded_components={"test"},
+        )
+        self.assertEqual(result, ["fft_lib_gfx942.tar.zst"])
```
ScottTodd: This test case is similar to https://github.com/ROCm/TheRock/blob/58d2f9771438624c0e7b26e2dc09c66a84424352/build_tools/tests/artifact_manager_tool_test.py#L595-L596

could merge the two? Or at least put them next to each other in the file?

ScottTodd (commented):
Test case design needs more work here. Please resist the urge to just append new test cases - prefer to update existing tests where they exist instead (and/or review the output of AI agents before just pushing and relying on reviewers to do that for you).

raramakr:
> Test case design needs more work here. Please resist the urge to just append new test cases - prefer to update existing tests where they exist instead (and/or review the output of AI agents before just pushing and relying on reviewers to do that for you).

Thanks @ScottTodd  for the guide lines .Will make sure to  reuse test cases rather than adding new ones

build_tools/tests/artifact_manager_tool_test.py
raramakr: Merged the same

build_tools/tests/artifact_manager_tool_test.py
raramakr: Will address this in a  different PR

build_tools/tests/artifact_manager_tool_test.py
raramakr: Removed the same. It adds little value

build_tools/tests/artifact_manager_tool_test.py
raramakr: Remove the changes and added in the existing test case

build_tools/tests/artifact_manager_tool_test.py
raramakr: Removed the  test cases for private helper function

ScottTodd (approved):
Thanks. We may also want integration tests checking some combination of
* `artifact_manager.py fetch` can download _all_ artifacts produced by CI and none were omitted (this `_get_base_arch` code may have missed something else)
* https://github.com/ROCm/TheRock/blob/main/build_tools/build_tarballs.py (which uses `artifact_manager.py fetch`) can produce tarballs including _all_ files produced by CI 
* ASan packages (tarballs, native packages) include xnack artifacts/files (in addition to the subproject tests that use those files)

raramakr:
> Thanks. We may also want integration tests checking some combination of
> 
> * `artifact_manager.py fetch` can download _all_ artifacts produced by CI and none were omitted (this `_get_base_arch` code may have missed something else)
> * https://github.com/ROCm/TheRock/blob/main/build_tools/build_tarballs.py (which uses `artifact_manager.py fetch`) can produce tarballs including _all_ files produced by CI
> * ASan packages (tarballs, native packages) include xnack artifacts/files (in addition to the subproject tests that use those files)

Ack, Will address these things in a different PR along with the cleanup suggested .

marbre (commented):
Unit tests are failing on Windows:
```
FAILED tests/artifact_manager_tool_test.py::TestFetchStageAll::test_fetch_exclude_components_skips_test_artifacts - OSError: [WinError 87] The parameter is incorrect
FAILED tests/artifact_manager_tool_test.py::TestFetchAmdgpuTargets::test_fetch_with_amdgpu_targets_finds_individual_target_archives - OSError: [WinError 87] The parameter is incorrect
```

build_tools/tests/artifact_manager_tool_test.py
```diff
@@ -593,12 +593,13 @@ def mock_extract(request):
...
 
     def test_fetch_exclude_components_skips_test_artifacts(self) -> None:
-        """Test that --exclude-components filters artifact components."""
+        """Test that --exclude-components filters artifact components including xnack variants."""
         import artifact_manager
 
         self._create_staged_artifact("test-artifact", "lib", "generic")
         self._create_staged_artifact("test-artifact", "test", "generic")
         self._create_staged_artifact("test-artifact", "test", "gfx942")
+        self._create_staged_artifact("test-artifact", "test", "gfx942:xnack+")
```
ScottTodd: > Unit tests are failing on Windows:
> 
> ```
> FAILED tests/artifact_manager_tool_test.py::TestFetchStageAll::test_fetch_exclude_components_skips_test_artifacts - OSError: [WinError 87] The parameter is incorrect
> FAILED tests/artifact_manager_tool_test.py::TestFetchAmdgpuTargets::test_fetch_with_amdgpu_targets_finds_individual_target_archives - OSError: [WinError 87] The parameter is incorrect
> ```

https://github.com/ROCm/TheRock/actions/runs/30945905710/job/92115696083?pr=7090

https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions

Maybe since `:` is a reserved character on Windows and can't be used in file names

raramakr: Can we skip the  xnack tests for windows?

ScottTodd: Well we don't support ASan builds (host or device) on Windows currently, so .... yes. I would prefer if we didn't generate files on one operating system that are invalid on another operating system though.

build_tools/artifact_manager.py
```diff
@@ -76,6 +76,36 @@
...
 
+def _get_base_arch(target: str) -> str:
+    """Strip xnack/other suffixes: 'gfx942:xnack+' -> 'gfx942'.
+
+    Note: This strips everything after the first colon. Any suffix (not just
+    xnack) will be removed, e.g., 'gfx942:foo' -> 'gfx942'.
+    """
+    if not target:
+        return ""
+    base = target.split(":")[0]
```
geomin12, line 87: do we see other symbols being used? wonder if we can do other punctuation

this is fine for now and fixes though

geomin12 (approved):

Issue #7092, opened 2026-08-04
[Issue]: Add clinfo (OpenCL) coverage to the ROCm sanity tests


lamb-j:
### Summary

The ROCm sanity suite (`tests/test_rocm_sanity.py`) validates the ROCr/HSA enumeration path via `rocminfo`, but has no check for the OpenCL runtime. OpenCL (`OCL_ICD` / `OCL_RUNTIME`) is a CORE feature built by default, and `clinfo` + the ICD loader (`libOpenCL.so` -> `libamdocl64.so`) ship in the `core-ocl` artifact, so the OpenCL enumeration path currently goes unexercised by the always-on sanity job.

### Proposal

Add a `clinfo`-based sanity check that enumerates the GPU through the OpenCL ICD loader, mirroring the existing `rocminfo` search-based tests (AMD platform, `CL_DEVICE_TYPE_GPU`, `gfx` device name).

Note: the sanity job fetches artifacts with `--base-only`, which pulls `core-ocl_lib`/`_dev` but not `core-ocl_run` (where the `clinfo` binary and `etc/OpenCL/vendors/amdocl64.icd` live), so the base artifact set needs `core-ocl_run` added.

### Scope

- Linux only initially. Windows OpenCL uses the PAL runtime backend, which this flow does not build/validate.
- Skipped under ASAN (runtime GPU enumeration is flaky under ASAN, cf. #3312).

Complements the deeper OpenCL-CTS conformance work (PR #5310) with a fast, always-on smoke check.

PR #5310, opened 2026-05-18
Add OpenCL-CTS testing

litvaOo:
## Motivation
We want to run OpenCL-CTS tests as part of TheRock test suite. This PR provides this functionality as part of the test matrix. 

## Technical Details
This PR is using the [ROCm fork of OpenCL-CTS](https://github.com/rocm/opencl-cts) for testing. Tests are built against upstream [OpenCL-Headers](https://github.com/KhronosGroup/OpenCL-Headers/) and [SPIR-V-Headers](https://github.com/KhronosGroup/SPIRV-Headers) from Khronos Group.

## Test Plan
The test suite should pass completely on CI.

## Test Result
Waiting for CI results.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

Issue #3312, opened 2026-02-09
[issue] `rocminfo` test fails when compiled with ASAN sanitizer


geomin12:
After compiling ASAN with `rocminfo`, we are noticing errors:

```
INFO     test_rocm_sanity:test_rocm_sanity.py:33 ++ Run [None]$ /__w/TheRock/TheRock/build/bin/rocminfo
ERROR    test_rocm_sanity:test_rocm_sanity.py:42 Command failed!
ERROR    test_rocm_sanity:test_rocm_sanity.py:43 command stdout:
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 ROCk module version 6.14.14 is loaded
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 =====================    
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 HSA System Attributes    
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 =====================    
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 Runtime Version:         1.18
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 Runtime Ext Version:     1.16
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 System Timestamp Freq.:  1000.000000MHz
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 Sig. Max Wait Duration:  18446744073709551615 (0xFFFFFFFFFFFFFFFF) (timestamp count)
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 Machine Model:           LARGE                              
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 System Endianness:       LITTLE                             
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 Mwaitx:                  DISABLED
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 XNACK enabled:           NO
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 DMAbuf Support:          YES
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 VMM Support:             YES
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 ==========               
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 HSA Agents               
ERROR    test_rocm_sanity:test_rocm_sanity.py:45 ==========               
[...]

PR #7093, opened 2026-08-04
test(sanity): add clinfo OpenCL sanity check

lamb-j:
## Motivation

The ROCm sanity suite (`tests/test_rocm_sanity.py`) validates the ROCr/HSA enumeration path via `rocminfo`, but nothing exercises the OpenCL runtime. OpenCL (`OCL_ICD` / `OCL_RUNTIME`) is a CORE feature built by default, and `clinfo` + the `amdocl64` vendor runtime ship in the `core-ocl` artifact — so the OpenCL enumeration path currently goes unchecked by the always-on sanity job.

This is especially useful on **Windows**, where `rocminfo` is HSA-only and unsupported: `clinfo` (OpenCL, PAL backend) is the GPU-enumeration sanity path that actually works there.

## Technical Details

- Adds a `clinfo_output` fixture and a parametrized `test_clinfo_output` that enumerates the GPU through the **system** OpenCL ICD loader and searches for the AMD platform, `CL_DEVICE_TYPE_GPU`, and a `gfx` device name — mirroring the existing `rocminfo` search tests. Runs on **Linux and Windows**.
- The test points `OCL_ICD_FILENAMES` at this build's `amdocl64` vendor runtime, so it exercises TheRock's OpenCL through the system loader (`libOpenCL` on Linux, `OpenCL.dll` on Windows) without relying on any `/etc` or registry ICD registration.
- Sets the `sanity` component's `container_image` to `no_rocm_image_ubuntu24_04_ocl_rt`, which adds the Khronos loader (`ocl-icd-libopencl1`) on top of the base image (Linux only; ignored on Windows, which uses its system `OpenCL.dll`). Same image family used by the OpenCL-CTS work in #5310.
- Adds `core-ocl_run` to `base_artifact_patterns` so the `--base-only` sanity fetch includes the `clinfo` binary. The `amdocl64` vendor is already pulled via `core-ocl_lib`.

### Design note: system loader, not bundled

The ICD **loader** is intentionally sourced from the system/distro (container package on Linux, System32 on Windows) rather than bundled from `core-ocl-icd`. Base artifacts carry only the `clinfo` binary and the `amdocl64` vendor. This keeps the loader a system component.

### Scope / notes
- **Skipped under ASAN** (runtime GPU enumeration is flaky under ASAN, cf. #3312).
- Complements the deeper OpenCL-CTS conformance work (#5310) with a fast, always-on smoke check.

## Test Result

Validated locally against a `gfx1030` (RX 6950 XT) build using the **system** `libOpenCL.so.1` loader (same package the `_ocl_rt` container installs) + `OCL_ICD_FILENAMES` at TheRock's vendor:

```
test_clinfo_output[clinfo - AMD Platform Search] PASSED
test_clinfo_output[clinfo - GPU Device Type Search] PASSED
test_clinfo_output[clinfo - GFX Name Search] PASSED
```

Windows behavior (device-name format under PAL, headless OpenCL on the Windows GPU runner) will be confirmed by CI on this draft.

ISSUE ID: #7092


chinmaydd:
Seems fine to me overall. Curious why we are seeing failures

lamb-j:
Failing checks seem unrelated, so I think this one is good to go/ready for review

tests/test_rocm_sanity.py
```diff
@@ -55,7 +57,68 @@ def rocm_info_output():
...
+            raise FileNotFoundError(f"OpenCL vendor runtime not found: {vendor}")
+        env = os.environ.copy()
+        env["PATH"] = str(THEROCK_BIN_DIR) + os.pathsep + env.get("PATH", "")
+        # The system ICD loader can fall back to a registered driver. Load the
+        # built vendor directly via clinfo's application-local OpenCL.dll.
+        # amdocl64 exports the OpenCL entry points used by clinfo.
+        with tempfile.TemporaryDirectory(prefix="therock-clinfo-") as directory:
+            clinfo = Path(directory) / "clinfo.exe"
+            shutil.copy2(THEROCK_BIN_DIR / "clinfo.exe", clinfo)
+            shutil.copy2(vendor, Path(directory) / "OpenCL.dll")
```
chinmaydd: Any reason to not copy the OpenCL.dll directly into THEROCK_BIN_DIR? I'm not too aware of how the test env runs so maybe we are trying to avoid that artifact persisting on there?

tests/test_rocm_sanity.py
```diff
@@ -55,7 +57,68 @@ def rocm_info_output():
...
+    if env.get("LD_LIBRARY_PATH"):
+        library_path.append(env["LD_LIBRARY_PATH"])
+    env["LD_LIBRARY_PATH"] = os.pathsep.join(library_path)
+
+    with tempfile.TemporaryDirectory(prefix="therock-ocl-icd-") as icd_dir:
+        (Path(icd_dir) / "amdocl64.icd").write_text(f"{vendor}\n", encoding="utf-8")
+        env["OCL_ICD_VENDORS"] = icd_dir
+        return run_command([str(THEROCK_BIN_DIR / "clinfo")], env=env).stdout
+
+
```
chinmaydd: Unnecessary extra line

```suggestion
```

tests/test_rocm_sanity.py
```diff
@@ -55,7 +57,68 @@ def rocm_info_output():
...
+        "to_search",
+        [
+            r"Platform\s*Name:\s*AMD Accelerated Parallel Processing",
+            r"Device\s*Type:\s*CL_DEVICE_TYPE_GPU",
+            r"Name:\s*gfx",
+        ],
+        ids=[
+            "clinfo - AMD Platform Search",
+            "clinfo - GPU Device Type Search",
+            "clinfo - GFX Name Search",
```
chinmaydd: Does this check if the return value of clinfo is 0 ? For instance, there's a chance that the above strings are present in output but due to device-libs mismatch you never see the string:

```
Extensions: cl_khr_fp64 cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_int64_base_atomics cl_khr_int64_extended_atomics cl_khr_3d_image_writes cl_khr_byte_addressable_store cl_khr_fp16 cl_khr_gl_sharing cl_amd_device_attribute_query cl_amd_media_ops cl_amd_media_ops2 cl_khr_d3d10_sharing cl_khr_d3d11_sharing cl_khr_dx9_media_sharing cl_khr_image2d_from_buffer cl_khr_subgroups cl_khr_gl_event cl_khr_depth_images cl_khr_mipmap_image cl_khr_mipmap_image_writes cl_amd_copy_buffer_p2p cl_amd_planar_yuv
```

^^ This output is generated when the test kernel actually compiles

tests/test_rocm_sanity.py
lamb-j: Apparently the python formatter expects one there? Because it's between a top-level and class?

tests/test_rocm_sanity.py
lamb-j: Ya it's to avoid any persistent one

build_tools/install_rocm_from_artifacts.py
```diff
@@ -390,6 +390,7 @@ def retrieve_artifacts_by_run_id(args):
         "core-kpack_lib",
         "core-ocl_lib",
         "core-ocl_dev",
+        "core-ocl_run",  # clinfo for the OpenCL sanity test
```
ScottTodd: Similar to https://github.com/ROCm/TheRock/pull/8494#discussion_r4097901767 for `core-hip_run` in `hip-tests`, if this isn't required for _all tests_, we may want to move it out of `base_artifact_patterns` and to a separate group for just sanity tests (see the code below this). The `clinfo` binary is ~90KB so it's not _too_ expensive to always download, but it could add up.

tests/test_rocm_sanity.py
ScottTodd: Yes the line is required by the formatter, see https://peps.python.org/pep-0008/#blank-lines

> [Blank Lines](https://peps.python.org/pep-0008/#blank-lines)
>
> Surround top-level function and class definitions with two blank lines.
>
> Method definitions inside a class are surrounded by a single blank line.

ScottTodd (approved):

Issue #6627, opened 2026-07-16
[issue] (asan) - CI Validation Gates


geomin12:
Enable ASAN gates across TheRock, R-L, R-S.

## Current State

| Repo | Pre-submit | Post-submit | Nightly |
|------|------------|-------------|---------|
| **TheRock** | disabled  | disabled  | full-asan |
| **R-L** | disabled | disabled | disabled |
| **R-S** | disabled | disabled | disabled |

## Future State

| Repo | Pre-submit | Post-submit | Nightly |
|------|------------|-------------|---------|
| **TheRock** | host-asan | host-asan | full-asan |
| **R-L** | host-asan | host-asan | full-asan |
| **R-S** | host-asan | host-asan | full-asan |

## Tasks
order:
- [x] Multi-arch ASAN Nightly pipelines for both r-s and r-l -> Enable here if not already done
- [x] Post-submit on rocm-systems
- [x] CI label:asan which can be attached at pre-submit for r-s & r-l
- [x] Post submit on r-l
- [x] Pre-submit on r-s (disable post-submit)
- [ ] Pre-submit on r-l -> enable until optimizations are in place

## Rollout
1. Start with post-submit to monitor queue impact
2. Promote to pre-submit once capacity confirmed

sa-faizal:
Completed:
- Multi-arch ASAN nightly pipelines for R-S and R-L 
- Post-submit on rocm-systems https://github.com/ROCm/rocm-systems/pull/9216
- CI label asan for pre-submit on Rock
- Post-submit on rocm-libraries https://github.com/ROCm/rocm-libraries/pull/10066

Not done:
- Pre-submit on R-S, R-L
- CI label ASAN for pre-submit on R-S, R-L

geomin12:
all subtasks completed

sa-faizal:
Initial Plan - 
* Label based triggers in r-l
* Pre-submit trigger in r-s

PR #7110, opened 2026-08-04
feat(ci): use host-asan for presubmit ASAN builds

geomin12:
Remap asan -> host-asan for pull_request events, not just push. This makes presubmit ASAN builds faster and more focused on host-side sanitizer checks rather than full device ASAN.

ISSUE ID: https://github.com/ROCm/TheRock/issues/6627

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -1236,8 +1236,8 @@ def expand_build_configs(
     all_families = _apply_external_family_overrides(all_families)
     build_variant = ci_inputs.build_variant
     # for ASAN CI runs, workflow_dispatch and scheduled events are "asan".
-    # Otherwise, push events run "host-asan"
-    if build_variant == "asan" and ci_inputs.is_push:
+    # Otherwise, push and pull_request events run "host-asan"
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1236,8 +1236,8 @@ def expand_build_configs(
     all_families = _apply_external_family_overrides(all_families)
     build_variant = ci_inputs.build_variant
     # for ASAN CI runs, workflow_dispatch and scheduled events are "asan".
-    # Otherwise, push events run "host-asan"
-    if build_variant == "asan" and ci_inputs.is_push:
+    # Otherwise, push and pull_request events run "host-asan"
+    if build_variant == "asan" and (ci_inputs.is_push or ci_inputs.is_pull_request):
         build_variant = "host-asan"
```
ScottTodd, line 1241: I would rather we add explicit `ci:asan` and `ci:host-asan` labels that directly control the behavior instead of hiding the logic all the way down here on line 1240 of a script, but this is okay for now.

ScottTodd (approved):

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: I'll do that in a follow-up! I agree, this is very hidden and tribal knowledge-type

PR #7112, opened 2026-08-04
docs(rfc): align flag lifecycle RFC with the typed build flag provider protocol (#6984)

tony-davis:
ISSUE ID: #6118

Amends [#6118](https://github.com/ROCm/TheRock/pull/6118) to line the feature flag lifecycle RFC up with [#6984](https://github.com/ROCm/TheRock/pull/6984), @stellaraccident's typed build flag provider protocol. Separate PR against Brian's branch so the diff is reviewable on its own instead of landing as another force of churn on #6118.

@ScottTodd tagged @BrianHarrisonAMD on #6984 over the "where flags are declared" tension between the two documents. This is my attempt at resolving it, and it resolves in #6984's favor.

**The substantive change: origination.** The RFC said "all flags start in TheRock as the single source of truth." Stella's follow-up on #6984 says she wants to "originate a flag in either place so we aren't so strictly interlocked," and I think she's right. The old rule interlocks the repos at exactly the wrong moment; a team can't guard code until a declaration lands in TheRock *and* that version reaches their branch. The RFC now says TheRock owns the mechanism and the aggregated inventory, not the birthplace. Where a TheRock declaration is still involved, advancing the pinned ref is self-service (resync with `develop`, per [rocm-libraries #9602](https://github.com/ROCm/rocm-libraries/pull/9602)) rather than a wait on the 12h bump cron.

**Flags are typed.** The glossary's "named boolean toggle," the taxonomy table, and the `therock_declare_flag` example were all boolean-only. They now carry `TYPE` (`BOOL` default, `INTEGER` with `VALID_VALUES`), plus a note on what a multi-valued flag means for the promotion train and for both-state CI. Runtime flags stay `BOOL`-only for now, but nothing precludes typing them later.

**Distribution boundary.** #6984 forbids installing build-time inputs. This RFC ships an installed `feature_flags.json`, which reads as a contradiction unless it's addressed head-on. New subsection adopts the boundary unchanged for build-time flags and explains why runtime state sits on the other side of it by construction (a value you can't read after install isn't a runtime flag).

**Fail-closed on names.** The runtime reader still falls back silently on a *missing file*, since a standalone build has none. But a misspelled *flag name* now has to be detectable, which is what the `ROCM_BUILD_FLAG(name)` macro trick buys on the build-time side.

**Mechanism deference.** New non-goal, new P0 phase for #6984 itself, two resolved decisions, and a table in the Build-Time section that describes #6984's pieces instead of re-proposing them.

Two new open questions fall out: how the inventory aggregates flags declared outside TheRock, and whether runtime flags should be typed too.

One editorial note. The RFC body states the design directly, with no "an earlier revision said..." anywhere; the differential is here in the description and in the revision history entry at the bottom. I'd like to drop that revision history section entirely before #6118 merges, since we're headed toward a clean doc, but it's useful while the amendments are still in flight.

**Testing:** `pre-commit run --all-files` clean on the changed file (mdformat, trailing whitespace, line endings). Documentation-only change; no build or CI impact.


BrianHarrisonAMD (approved):
LGTM thanks for the update. I think these are good additions, and work with the initial RFC.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -337,6 +364,33 @@ that step, and it ships without additional configuration.
 > own per-component override file under its own tree (for example, `lib/<lib>/feature_flags.json`):
 > that library then adds one `include` line to its component toml. The global file needs none.
 
+### Distribution boundary (runtime state is a deliberate exception)
+
+The build-flag provider protocol (PR #6984) draws a hard line: `ROCMBuildFlags.cmake`, the provider
+state file, and any generated private flag header are **build-time inputs only**. They are never
+installed, never exported as a target, and never appear in an installed package config, so an
+application calling `find_package()` on an installed ROCm library acquires no build-system
+dependency. This RFC adopts that boundary unchanged for build-time flags.
```
BrianHarrisonAMD: I agree with this, but think we might still want to ship information tracking which build flags were enabled for a specific package / install. It wouldn't be part of the cmake, or intended to be consumed.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
```diff
@@ -1119,11 +1216,15 @@ identical. No per-repo reimplementation, no drift.
...
 1. **Promotion cadence period:** weekly, bi-weekly, or monthly (and the exact schedule on the multi-arch CI). Only "fixed and automated" is settled; the period and schedule are open.
+1. **How the inventory aggregates flags declared outside TheRock.** Because origination is allowed in either place, the hygiene audit and the manifest need to see super-repo-declared flags. Options: TheRock scans the checked-out consuming super-repositories at configure time; each super-repo emits its own declarations into the provider handshake; or the audit runs per-repo and reports into a shared place. Preference not settled; PR #6984's follow-on work will constrain the answer.
+1. **Whether runtime flags should also be typed.** Build-time flags are `BOOL` or `INTEGER` (PR #6984). The runtime contract here is BOOL-only, which keeps `ROCM_FEATURE_<NAME>=0|1` and the JSON simple. Symmetry argues for typing runtime flags too; simplicity and the fact that no runtime use case yet needs a non-BOOL value argue for waiting. The RFC deliberately does not preclude it: nothing in the JSON, the manifest block, or the precedence order depends on the value being boolean.
```
BrianHarrisonAMD: I would be fine with runtime flags supporting `BOOL` or `INTEGER` as well, but I also think we can start with just bool, and extend later as needed.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
tony-davis: Good catch, and I think we already have what you want; my wording just hid it. `therock_finalize_flags()` writes `flag_settings.json`, which gets merged into the shipped `share/therock/therock_manifest.json` under `"flags"`, so the as-built flag state ships today. This RFC's metadata extension enriches it further with owner, stage, and expiry.

The problem is my two-column table read as "no build-flag information ships at all," which isn't the case and isn't what #6984 means either. I was conflating the *input* (provider state, helper, private header, which genuinely must never ship) with the *record* (the manifest block, which should). Your "wouldn't be part of the cmake, or intended to be consumed" is exactly the right line.

Reworked it: the table is now keyed by artifact with the manifest as its own row, and the section says the rule is about direction rather than secrecy. Build-time flag state is fully observable on an installed system even though the provider state file isn't installed.

docs/rfcs/RFC00XX-Feature-Flag-Lifecycle-and-Canary-Promotion.md
tony-davis: Agreed, and that's what the RFC does. I moved it out of the open questions and into resolved decisions so it's a stated position rather than a pending one: runtime flags start `BOOL`-only and may gain `INTEGER` later. Called out explicitly that nothing in the JSON, the manifest `runtime_flags` block, or the precedence order depends on the value being boolean, so extending it later is additive.

Issue #3161, opened 2026-01-29
[Feature]:  Build rocm Python packages with ASan


radhaksri:
### Suggestion Description

As of 01/29/2016, ASAN builds of ROCm generated by [CI ASAN Workflow](https://github.com/ROCm/TheRock/actions/workflows/ci_asan.yml) can only be downloaded and flattened using install_rocm_from_artifacts.py. 

A pip installable ASAN build would enable seamless diagnosing and debugging of workflows, especially when deploying over cluster.

### Context

We are engaged with training issues at a customer site every day. For many weeks (greater than one month), they have been struggling with an illegal memory access issue which is not being caught correctly. Our team has done extensive attempts at log analysis, but they can't tell where these issues are bubbling up from. As a result, in January, we embarked on a journey to enable ASAN at all parts of the ROCm chain that's being used by the customer's training flow.
 
Address Sanitization sounds a lot to me like valgrind/Purify in that it implements/injects memory fencing, so that memory access errors are handled more gracefully than a segfault & core file. This is important because frontier labs are very careful about what files and data they will share - so we need this advanced logging / fencing behavior in order to identify which bit of code is causing the issue.
 
The goal is that any user who is experiencing illegal memory access issues will be able to deploy an alternate build of the ROCm stack (including the Frameworks) which has ASAN enabled, and run their workloads via that stack to identify what's causing the exception.

There will never (I'm pretty sure) be an expectation that ASAN-enabled builds would run in production. ASAN carries a significant performance impact. There's plenty of concern that when using ASAN, due to Heisenbug-principles, the issue may not appear. However, the team is fully blocked at moving forward, so we need to go this route.
 
### Operating System

All

### GPU

All

### ROCm Component

ROCk

PR #7113, opened 2026-08-04
fix(py_packaging): strip xnack suffix from wheel package names

raramakr:
Target families like 'gfx942:xnack+' contain characters (':' and '+')
  that are invalid in Python package names (PEP 508). This caused wheel
  builds to fail with:

    error: Invalid distribution name or version syntax:
    rocm-sdk-device-gfx942-xnack--10.1.0.dev0+...

  Fix by stripping the xnack suffix in get_dist_package_name() so that
  'gfx942:xnack+' produces 'rocm-sdk-device-gfx942' instead of the
  invalid 'rocm-sdk-device-gfx942:xnack+'.

  Fixes multi-arch ASAN builds for gfx94x targets.

JIRA ID : https://amd-hub.atlassian.net/browse/ROCM-28314

### DIFF
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
@@ -331,8 +331,13 @@ def get_target_family_platform_marker(target_family: str) -> str:
     """
     if not LINUX_TARGET_FAMILIES or not WINDOWS_TARGET_FAMILIES:
         return ""
-    in_linux = target_family in LINUX_TARGET_FAMILIES
-    in_windows = target_family in WINDOWS_TARGET_FAMILIES
+    # Compare base targets (strip xnack suffix) since platform lists may contain
+    # xnack-suffixed entries like 'gfx942:xnack+' while we receive base targets.
+    base_target = target_family.split(":")[0]
+    linux_base_targets = {t.split(":")[0] for t in LINUX_TARGET_FAMILIES}
+    windows_base_targets = {t.split(":")[0] for t in WINDOWS_TARGET_FAMILIES}
+    in_linux = base_target in linux_base_targets
+    in_windows = base_target in windows_base_targets
     if in_linux and not in_windows:
         return 'sys_platform == "linux"'
     if in_windows and not in_linux:
@@ -353,13 +358,16 @@ def build_per_target_extras() -> "dict[str, list[str]]":
     the generic extras already in setup.py's EXTRAS_REQUIRE suffice).
     """
     result: dict[str, list[str]] = {}
-    if len(AVAILABLE_TARGET_FAMILIES) <= 1:
+    # Deduplicate by base target (strip xnack suffix) to avoid redundant iterations
+    # when both 'gfx942' and 'gfx942:xnack+' exist in AVAILABLE_TARGET_FAMILIES.
+    base_targets = sorted(set(tf.split(":")[0] for tf in AVAILABLE_TARGET_FAMILIES))
+    if len(base_targets) <= 1:
         return result
     for pkg in ALL_PACKAGES.values():
         if not pkg.is_target_specific or pkg.required:
             continue
         all_requires: list[str] = []
-        for tf in sorted(AVAILABLE_TARGET_FAMILIES):
+        for tf in base_targets:
             extra_name = f"{pkg.logical_name}-{tf}"
             req = pkg.get_dist_package_require(target_family=tf)
             marker = get_target_family_platform_marker(tf)
--- a/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
+++ b/build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
@@ -96,7 +96,8 @@ class PackageEntry:
             )
         kwargs = {}
         if target_family is not None:
-            kwargs["target_family"] = target_family
+            # Strip xnack suffix (e.g., 'gfx942:xnack+' -> 'gfx942') for valid package names
+            kwargs["target_family"] = target_family.split(":")[0]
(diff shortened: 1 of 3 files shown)


build_tools/tests/py_packaging_test.py
```diff
@@ -536,6 +536,14 @@ def test_populate_device_files_copies_all_files(self):
         artifact_dir = self._setup_kpack_split_artifacts()
         params = self._make_params(artifact_dir, kpack_split=True)
 
+        # Verify xnack-suffixed targets produce valid package names by stripping
+        # the suffix (e.g., 'gfx942:xnack+' -> 'rocm-sdk-device-gfx942')
+        device_entry = params.dist_info.ALL_PACKAGES["device"]
+        self.assertEqual(
```
nunnikri, line 544: Good to add xnack- and just gfx942 for completeness.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
```diff
@@ -96,7 +96,8 @@ def get_dist_package_name(self, target_family: str | None = None) -> str:
             )
         kwargs = {}
         if target_family is not None:
-            kwargs["target_family"] = target_family
+            # Strip xnack suffix (e.g., 'gfx942:xnack+' -> 'gfx942') for valid package names
+            kwargs["target_family"] = target_family.split(":")[0]
```
nunnikri, line 100: A similar fix might be needed in build_per_target_extras() fn as well.
Else the extras keys will still have xnack.

build_tools/tests/py_packaging_test.py
```diff
@@ -536,6 +536,14 @@ def test_populate_device_files_copies_all_files(self):
...
+            device_entry.get_dist_package_name("gfx942:xnack+"),
+            "rocm-sdk-device-gfx942",
+        )
+
```
nunnikri, line 570: Some more test which can be added.
1. build_per_target_extras with xnack target  — verify both the key (device-gfx942) and the value (rocm-sdk-device-gfx942==...) are clean when AVAILABLE_TARGET_FAMILIES contains gfx942:xnack+
2. get_dist_package_require with xnack — verify the version requirement string is valid
3. get_py_package_name with xnack — verify the internal Python module name is clean

marbre (commented):
Do we have a run were we observed that this is failing? I am wondering where we will push those packages and if we need to alter the package name further.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
ScottTodd, line 1: > JIRA ID : https://amd-hub.atlassian.net/browse/ROCM-28314

Issues with open source packaging code need **GitHub** issues, see https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#communication-channels. Please redirect/close jira issues before people get further in the habit of using them. We can attach this to https://github.com/ROCm/TheRock/issues/3161 or file subissues under that issue.

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
ScottTodd, line 100: Is stripping the suffix rather than sanitizing it the right fix? Do device ASan builds produce _only_ `:xnack+` packages or do they produce multiple packages that need to coexist?

Looking at a recent asan nightly release (https://github.com/ROCm/rockrel/actions/runs/30963528156), the artifacts at https://therock-nightly-artifacts.s3.amazonaws.com/30963528156-linux/index.html include:
* `blas_lib_gfx950.tar.zst` containing various .co and .hsaco files under `math-libs/BLAS/hipBLASLt/stage/lib/hipblaslt/library/gfx950`
* `blas_lib_gfx950:xnack+.tar.zst` containing just `math-libs/BLAS/hipBLASLt/stage/.kpack/blas_lib_gfx950_xnack+.kpack`

Do we want to include files from both those artifacts into a common `gfx950` package in the asan release index, or would we want both `gfx950` and `gfx950_xnack+` (or some other sanitized/normalized name).

nunnikri: I think its safer to package both as gfx950. And the runtime can decide which files to load.

ScottTodd: Does this merge files, or does it produce two packages where one overwrites the other?

raramakr:
> Do we have a run were we observed that this is failing? I am wondering where we will push those packages and if we need to alter the package name further.

https://github.com/ROCm/TheRock/actions/runs/30958264761/

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
raramakr: Done

build_tools/tests/py_packaging_test.py
raramakr: Done

build_tools/packaging/python/templates/rocm/src/rocm_sdk/_dist_info.py
raramakr: Good catch @ScottTodd  .It was overriding. 
Fixed the same  and now it merge files

build_tools/tests/py_packaging_test.py
raramakr: Done

nunnikri (approved):
LGTM

raramakr:
Passed ASAN build with this PR and  https://github.com/ROCm/TheRock/pull/7090
https://github.com/ROCm/TheRock/actions/runs/31038584860

ScottTodd (approved):

PR #3943, opened 2026-03-12
[torch] Remove Windows test process exit hacks

ScottTodd:
## Motivation

Now that
* https://github.com/ROCm/TheRock/issues/999

is fixed, we shouldn't need this hack introduced in
* https://github.com/ROCm/TheRock/pull/2265

Progress on running more Windows PyTorch tests on CI/CD:
* https://github.com/ROCm/TheRock/issues/2258
* https://github.com/ROCm/TheRock/issues/3291

## Test Plan and Results

* Tested on CI for gfx1151 + torch 2.9: https://github.com/ROCm/TheRock/actions/runs/23020511361/
  * Test runner is working as expected
  * Attempt 1: failure due to https://github.com/ROCm/TheRock/issues/3173
  * Attempt 2: failure
    ```
    external-builds\pytorch\pytorch\test\test_nn.py::TestNNDeviceTypeCUDA::test_transformerencoderlayer_gelu_cuda_float32 PASSED [0.0955s] [  2%]
    :0:C:\home\runner\_work\TheRock\TheRock\rocm-systems\projects\clr\hipamd\src\hiprtc\hiprtcInternal.hpp:89  : 23043456529 us:  Unable to add internal header
    external-builds\pytorch\pytorch\test\test_nn.py::TestNNDeviceTypeCUDA::test_triplet_margin_with_distance_loss_cuda 
    ```
  * Attempt 3: passing
* Testing locally on gfx1100 + torch 2.9
  * Test runner is working as expected
    ```
    ==================== 15748 passed, 26182 skipped, 112 deselected, 45 xfailed in 550.34s (0:09:10) =====================
    Pytest finished with return code: 0
    ```

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #2265, opened 2025-11-21
[torch] Support running unit tests on Windows 

ScottTodd:
## Motivation

Progress on https://github.com/ROCm/TheRock/issues/2258 and https://github.com/ROCm/TheRock/issues/1073. This changes the `test_pytorch_wheels.yml` workflow from only being able to run our [PyTorch smoke tests](https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch/smoke-tests) to being able to run the full set in our `run_linux_pytorch_tests.py` script (renamed to `run_pytorch_tests.py`).

## Technical Details

Due to https://github.com/ROCm/TheRock/issues/999, I added a `force_exit_with_code()` hack to `run_pytorch_tests.py`. Since the test process does not terminate on its own, even after all test cases complete, I kill the process with `os.kill()`. I tried to use nicer methods like `sys.exit()` and `os._exit()` but these were not sufficient. A consequence of this is that the exit code of the process is now always 15 (SIGTERM) on Windows, so the script now writes `exit_code.txt` to the current directory for the `test_pytorch_wheels.yml` workflow to use.

Two test cases caused additional issues:
* `test_cublas_config_nondeterministic_alert_cuda` in `test_torch.py`
* `test_graph_error` in `test_cuda.py`

These test cases should be fixed or conditionally skipped in the upstream pytorch test files. Until then, I marked them as skipped using our new test filtering under a new "platform/windows" category.

> [!WARNING]
> Because we have test failures, running the tests will stop promotion from v2-staging to v2 on Windows for GPU families where we have test runners like gfx1151 and gfx110X, as is already done on Linux. See [Gating releases with Pytorch tests](https://github.com/ROCm/TheRock/tree/main/external-builds/pytorch#gating-releases-with-pytorch-tests).

## Test Plan

> [!IMPORTANT]
> **Given how unstable the tests are, I will first get this code in place and then enable the CI runs in follow-up PRs**

* Ran tests locally with `python D:/projects/TheRock/external-builds/pytorch/run_pytorch_tests.py --pytorch-dir D:/b/pytorch --amdgpu-family=gfx110X-dgpu > C:\Users\Nod-Shark16\.therock\logs\run_pytorch_tests_%date%_%time::=%.txt 2>&1`
* Test runs on CI machines:
  * gfx1151, torch 2.9, `test_torch.py` only: https://github.com/ROCm/TheRock/actions/runs/19585433265/job/56093159807
[...]

PR #7127, opened 2026-08-05
ci(Pytorch): Fixes Windows Pytorch test process_hangs after test completion

SolerHo:
## Motivation

Fixes issue : Windows PyTorch test process hangs after test completion due to
leaked daemon threads from torch's HIP runtime initialization.

`run_pytorch_tests_full.py` already applies this fix via `os._exit()`, but
`run_pytorch_tests.py` (used for Windows testing in the full test suite workflow)
uses `sys.exit()` which blocks indefinitely on Windows.

The existing workflow workaround (`continue-on-error: true` + reading from
a nonexistent `run_pytorch_tests_exit_code.txt`) silently discards all Windows
test results (always exits 1).

## Changes

1. **`run_pytorch_tests.py`**: Replace `return retcode` with
   `os._exit(retcode if retcode >= 0 else 1)`, matching the fix in
   `run_pytorch_tests_full.py`.

2. **`test_pytorch_wheels_full.yml`**: Remove the Windows exit-code hack
   (`continue-on-error: true` + the file-read workaround). Normal error
   propagation is restored.

## Test Plan

- [ ] Verify `run_pytorch_tests.py` exits cleanly on Windows (no hang)
- [ ] Confirm Windows test results are correctly reflected in CI (pass/fail)
- [ ] Verify Linux behavior is unaffected (no regression)

### DIFF
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -349,23 +349,11 @@ jobs:
       # See https://github.com/ROCm/TheRock/issues/999.
       - name: (Windows) Run PyTorch tests
         if: ${{ runner.os == 'Windows'}}
-        continue-on-error: true
         run: |
           python ./external-builds/pytorch/run_pytorch_tests.py -- \
             --continue-on-collection-errors \
             -v
 
-      - name: (Windows) Read and propagate exit code
-        if: ${{ runner.os == 'Windows'}}
-        run: |
-          if [ -f run_pytorch_tests_exit_code.txt ]; then
-            EXIT_CODE=$(cat run_pytorch_tests_exit_code.txt)
-            echo "Exit code from file: ${EXIT_CODE}"
-            exit ${EXIT_CODE}
-          else
-            echo "No run_pytorch_tests_exit_code.txt found"
-            exit 1
-          fi
 
   # Aggregate every shard's JUnit XML test reports into a single failure table
   # on the run's summary page, so all failures across configs/shards are visible
--- a/external-builds/pytorch/run_pytorch_tests.py
+++ b/external-builds/pytorch/run_pytorch_tests.py
@@ -311,7 +311,12 @@ def main() -> int:
 
         retcode = pytest.main(pytest_args)
         print(f"Pytest finished with return code: {retcode}")
-        return retcode
+        # Force-exit immediately. PyTorch's pytest is known to hang after
+        # all test files complete due to leaked daemon threads or orphan child
+        # processes (https://github.com/ROCm/TheRock/issues/999). os._exit()
+        # terminates without waiting for threads or running atexit handlers.
+        os._exit(retcode if retcode >= 0 else 1)
+
     except (ValueError, IndexError) as e:
         print(f"[ERROR] Exception in PyTorch unit-tests runner: {e}")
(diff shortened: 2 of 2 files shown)


.github/workflows/test_pytorch_wheels_full.yml
```diff
@@ -349,23 +349,11 @@ jobs:
...
-        if: ${{ runner.os == 'Windows'}}
-        run: |
-          if [ -f run_pytorch_tests_exit_code.txt ]; then
-            EXIT_CODE=$(cat run_pytorch_tests_exit_code.txt)
-            echo "Exit code from file: ${EXIT_CODE}"
-            exit ${EXIT_CODE}
-          else
-            echo "No run_pytorch_tests_exit_code.txt found"
-            exit 1
-          fi
```
ScottTodd, line 368: While I do think we should make this change, it probably won't be sufficient on its own to get this `.github/workflows/test_pytorch_wheels_full.yml` workflow running on Windows.

What we should do is match https://github.com/ROCm/TheRock/blob/bfe0b550286db235bd4d3b1f9dde226783c9d6d3/.github/workflows/test_pytorch_wheels.yml#L227-L256

* Make `- name: (Linux) Run PyTorch full test suite` run on both platforms and drop `if: ${{ runner.os == 'Linux' }}`
* Drop these steps entirely (including the "HACK: The test process does not terminate on its own gracefully" comment)

That's going to be hard to test from an external contribution from a fork PR as this workflow is only run on-demand or as part of nightly releases at the moment, and not via our CI workflows that run on pull requests.

external-builds/pytorch/run_pytorch_tests.py
```diff
@@ -311,7 +311,12 @@ def main() -> int:
 
         retcode = pytest.main(pytest_args)
         print(f"Pytest finished with return code: {retcode}")
-        return retcode
+        # Force-exit immediately. PyTorch's pytest is known to hang after
+        # all test files complete due to leaked daemon threads or orphan child
+        # processes (https://github.com/ROCm/TheRock/issues/999). os._exit()
+        # terminates without waiting for threads or running atexit handlers.
+        os._exit(retcode if retcode >= 0 else 1)
```
ScottTodd, line 318: Is that issue still occurring? We did recently fix some hangs that re-occurred. I would rather not add this workaround back in unless absolutely necessary.

If we _do_ add the workaround, is `os._exit` enough? See:
* https://github.com/ROCm/TheRock/pull/3943
* https://github.com/ROCm/TheRock/pull/2265

I found that `os._exit` was not enough to interrupt stalled threads and I needed to `os.kill(os.getpid(), signal.SIGTERM)`

```python
    # In order from "asking nicely" to "tear down immediately":
    #   1. `sys.exit(retcode)`
    #   2. `os._exit(retcode)`
    #   3. `os.kill(os.getpid(), signal.SIGTERM)`
    #   4. `subprocess.Popen(f'taskkill /F /PID {os.getpid()}', shell=True)`
    # As options (1) and (2) are not sufficient, we use option (3) here.
    os.kill(os.getpid(), signal.SIGTERM)
```

ScottTodd (commented):
Thanks for the PR, but I think this is going to be difficult to take an external contribution for given how the pytorch test workflows aren't directly integrated into our CI workflows

PR #7130, opened 2026-08-05
feat(ci): add dynamic stage determination for external repos

geomin12:
Enable external repos (rocm-libraries, rocm-systems) to dynamically determine which TheRock stages need rebuilding based on their changes.

Changes:
- setup_multi_arch.yml: Add EXTERNAL_REPO env var and GITHUB_TOKEN for baseline run API calls. Use THEROCK_REPOSITORY for baseline selection.
- configure_multi_arch_ci.py: Add GitContext.from_external_repo() to create context with repo name as changed file. Skip path filtering for external repos and apply stage reuse from TheRock baselines.
- stage_reuse_decision.py: Use THEROCK_REPOSITORY for baseline run selection so external repos can query TheRock baselines.
- configure_external_repo_ci.py: Update docstring and exclude workflow files from full test triggers.

How it works:
1. External repo passes EXTERNAL_REPO JSON with repository name
2. configure_multi_arch_ci.py extracts repo name (e.g., "rocm-libraries")
3. Stage impact analysis maps repo name to affected stages (e.g., math-libs)
4. Unaffected stages are candidates for reuse from TheRock baselines
5. Commit-compatible baseline runs provide prebuilt artifacts

Working here: https://github.com/ROCm/rocm-libraries/actions/runs/31032656859/job/92397575440?pr=10339#step:4:198 (however, no baseline as this is on a feature branch)
```
Writing job summary:
### Stage reuse analysis

- mode: `reuse-stage`
- full rebuild required: `False`
- baseline run checked: _none_
- unaffected candidates: `comm-libs`, `compiler-runtime`, `dctools-core`, `debug-tools`, `media-libs`, `profiler-apps`, `runtime-tests`, `storage-libs`, `wsl-rocdxg`
- available in baseline: _none_
- applied: _none_
- available per platform:
  - linux: _none_
  - windows: _none_
```

ISSUE ID: https://github.com/ROCm/TheRock/issues/3343

### DIFF
--- a/.github/workflows/setup_multi_arch.yml
+++ b/.github/workflows/setup_multi_arch.yml
@@ -177,14 +177,15 @@ jobs:
           PREBUILT_STAGES: ${{ inputs.prebuilt_stages }}
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
           STAGE_REUSE_MODE: ${{ inputs.stage_reuse_mode }}
+          # The checkout commit is used for commit-compatibility checking.
           STAGE_REUSE_CURRENT_SHA: ${{ steps.checkout.outputs.commit }}
           STAGE_REUSE_MAX_AGE_HOURS: ${{ inputs.stage_reuse_max_age_hours }}
           STAGE_REUSE_COMMIT_HISTORY: ${{ inputs.stage_reuse_commit_history }}
-          # For cross-repo artifact reuse, use repository input (defaults to GITHUB_REPOSITORY).
-          # External repos set repository to "ROCm/TheRock" to reuse TheRock's prebuilt artifacts.
-          THEROCK_REPOSITORY: ${{ inputs.repository }}
-          # Skip path filtering for external repos (git sha won't exist in TheRock)
-          SKIP_PATH_FILTERS: ${{ inputs.external_repo != '' && 'true' || '' }}
+          # For baseline run selection, use repository input (defaults to GITHUB_REPOSITORY).
+          # External repos set repository to "ROCm/TheRock" to fetch TheRock baseline runs.
+          THEROCK_REPOSITORY: ${{ inputs.repository || github.repository }}
+          # External repo JSON (e.g., {"repository":"ROCm/rocm-libraries","ref":"..."})
+          EXTERNAL_REPO: ${{ inputs.external_repo }}
           CI_CONFIG_PATH: ci-config
           # External repo family overrides allow callers to add/modify GPU family configs
           # (e.g., adding test runners for architectures not enabled in TheRock's default config)
--- a/.github/workflows/setup_multi_arch.yml
+++ b/.github/workflows/setup_multi_arch.yml
@@ -184,6 +184,8 @@ jobs:
           # For baseline run selection, use repository input (defaults to GITHUB_REPOSITORY).
           # External repos set repository to "ROCm/TheRock" to fetch TheRock baseline runs.
           THEROCK_REPOSITORY: ${{ inputs.repository || github.repository }}
+          # Token for GitHub API calls (branch history, baseline workflow runs)
+          GITHUB_TOKEN: ${{ github.token }}
           # External repo JSON (e.g., {"repository":"ROCm/rocm-libraries","ref":"..."})
           EXTERNAL_REPO: ${{ inputs.external_repo }}
           CI_CONFIG_PATH: ci-config
--- a/.github/workflows/setup_multi_arch.yml
+++ b/.github/workflows/setup_multi_arch.yml
@@ -124,6 +124,7 @@ on:
 
 permissions:
   contents: read
+  actions: read
 
 jobs:
   setup:
--- a/build_tools/github_actions/configure_external_repo_ci.py
+++ b/build_tools/github_actions/configure_external_repo_ci.py
@@ -5,7 +5,14 @@
 """Configure CI for external repos (rocm-systems, rocm-libraries).
 
 This script determines which projects changed and whether to run/skip tests.
-It consolidates logic previously duplicated across external repo workflows.
+
+Stage Reuse:
+    TheRock's stage_reuse_decision.py handles stage impact analysis and
+    baseline run selection using commit compatibility. The external repo
+    workflow should:
+    1. Resolve therock_ref using resolve_therock_ref.py (merge-base pinning)
+    2. Pass that ref to setup_multi_arch.yml
+    3. TheRock's stage_reuse_decision.py finds commit-compatible baseline runs
 
 Usage:
     python configure_external_repo_ci.py \
@@ -30,7 +37,6 @@ import subprocess
 import sys
 import time
 from dataclasses import dataclass, fields
-from pathlib import Path
 from typing import (
     Callable,
     Iterable,
@@ -58,8 +64,10 @@ SKIPPABLE_PATH_PATTERNS = [
 ]
 
 # Patterns that trigger a full test run when changed (CI infrastructure)
+# NOTE: .github/workflows/therock* is intentionally excluded since workflow
+# changes should still use stage reuse to determine which stages to rebuild.
+# The workflow itself doesn't affect TheRock build stages.
 FULL_TEST_TRIGGER_PATTERNS = [
-    ".github/workflows/therock*",
     ".github/scripts/therock*",
     ".github/scripts/get_changed_projects.py",
     ".github/scripts/ci_utils.py",
--- a/build_tools/github_actions/configure_external_repo_ci.py
+++ b/build_tools/github_actions/configure_external_repo_ci.py
@@ -64,10 +64,8 @@ SKIPPABLE_PATH_PATTERNS = [
 ]
 
 # Patterns that trigger a full test run when changed (CI infrastructure)
-# NOTE: .github/workflows/therock* is intentionally excluded since workflow
-# changes should still use stage reuse to determine which stages to rebuild.
-# The workflow itself doesn't affect TheRock build stages.
 FULL_TEST_TRIGGER_PATTERNS = [
+    ".github/workflows/therock*",
     ".github/scripts/therock*",
     ".github/scripts/get_changed_projects.py",
     ".github/scripts/ci_utils.py",
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -1429,17 +1429,18 @@ def main():
         # external_repo is JSON like {"repository":"ROCm/rocm-libraries","ref":"..."}
         try:
             external_repo = json.loads(ci_inputs.external_repo)
-            repo_full_name = external_repo.get("repository", "")
-            external_repo_name = (
-                repo_full_name.split("/")[-1] if "/" in repo_full_name else ""
-            )
-        except (json.JSONDecodeError, TypeError):
-            external_repo_name = ""
+        except (json.JSONDecodeError, TypeError) as exc:
+            raise ValueError(
+                f"EXTERNAL_REPO contains invalid JSON: {ci_inputs.external_repo!r}"
+            ) from exc
 
-        if external_repo_name:
-            git_context = GitContext.from_external_repo(external_repo_name)
-        else:
-            git_context = GitContext.empty()
+        repo_full_name = external_repo.get("repository", "")
+        if not repo_full_name:
+            raise ValueError(
+                f"EXTERNAL_REPO missing 'repository' field: {ci_inputs.external_repo!r}"
+            )
+        external_repo_name = repo_full_name.split("/")[-1]
+        git_context = GitContext.from_external_repo(external_repo_name)
     elif (ci_inputs.is_pull_request or ci_inputs.is_push) and ci_inputs.base_ref:
         # 'pull_request' and 'push' events can use the list of changed files
         # compared to the "prior commit" to affect job selections/options.
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -161,6 +161,10 @@ class CIInputs:
     # Repository to query for baseline runs (for cross-repo artifact reuse)
     baseline_repository: str = ""
 
+    # External repo JSON (e.g., '{"repository":"ROCm/rocm-libraries","ref":"..."}')
+    # Non-empty when an external repo calls TheRock workflows
+    external_repo: str = ""
+
     def log(self) -> None:
         """Log parsed inputs for CI diagnostics."""
         print("CIInputs:")
@@ -260,6 +264,7 @@ class CIInputs:
             prebuilt_stages=os.environ.get("PREBUILT_STAGES", ""),
             baseline_run_id=os.environ.get("BASELINE_RUN_ID", ""),
             baseline_repository=os.environ.get("THEROCK_REPOSITORY", ""),
+            external_repo=os.environ.get("EXTERNAL_REPO", ""),
         )
 
 
@@ -308,6 +313,17 @@ class GitContext:
         where we don't want to diff against a prior commit.
         """
         return GitContext()
+
+    @staticmethod
+    def from_external_repo(external_repo_name: str) -> "GitContext":
+        """Create context for external repo builds (e.g., rocm-libraries).
+
+        For external repos, we treat the repo name as the changed file so that
+        stage reuse analysis can determine which TheRock stages are affected.
+        The external repo name maps to a submodule in TheRock.
+        """
+        print(f"External repo detected: {external_repo_name}")
+        return GitContext(changed_files=[external_repo_name])
 
     @property
     def has_submodule_changes(self) -> bool | None:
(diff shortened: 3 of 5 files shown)


ammallya:
The PR in general looks good but please add test cases in unit tests

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -892,14 +917,17 @@ def decide_jobs(
...
+    # we reuse stages from TheRock baselines. For same-repo runs, baseline_repository
+    # is empty or matches the current repo.
+    # reuse-stage mode returns non-empty applied_reuse_stages.
+    for stage in auto_stage_reuse.applied_reuse_stages:
+        stage_decisions.setdefault(stage, JobAction.PREBUILT)
+    if auto_stage_reuse.applied_reuse_stages and auto_stage_reuse.baseline_run_id:
+        baseline_run_id = auto_stage_reuse.baseline_run_id
+        # For external repos, use their baseline_repository (ROCm/TheRock)
+        if ci_inputs.baseline_repository:
+            baseline_repository = ci_inputs.baseline_repository
```
ammallya: This is already done in line 917

ammallya (approved):
LGTM

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -309,6 +314,17 @@ def empty() -> "GitContext":
...
+    @staticmethod
+    def from_external_repo(external_repo_name: str) -> "GitContext":
+        """Create context for external repo builds (e.g., rocm-libraries).
+
+        For external repos, we treat the repo name as the changed file so that
+        stage reuse analysis can determine which TheRock stages are affected.
+        The external repo name maps to a submodule in TheRock.
+        """
+        print(f"External repo detected: {external_repo_name}")
+        return GitContext(changed_files=[external_repo_name])
```
erman-gurses: Should this also populate `submodule_paths`?

The docstring says the external repository name maps to a TheRock submodule, but returning only `changed_files` means `has_submodule_changes` evaluates to `None`.

That changes downstream test behavior: `_determine_test_type()` does not apply the submodule-change test policy, and families with `submodule_bump_tests_only` have their test runners disabled because they require `has_submodule_changes is True`.

An external `rocm-libraries` or `rocm-systems` change appears semantically equivalent to changing that submodule in TheRock, so should this return:

```python
GitContext(
    changed_files=[external_repo_name],
    submodule_paths=[external_repo_name],
)
```

Could we also add a test asserting that `GitContext.from_external_repo(...).has_submodule_changes` is `True`?

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -474,7 +474,11 @@ def _default_baseline_selector(*, platform: str) -> BaselineSelector:
     extra "passing build" check is needed here.
     """
 
-    github_repository = os.environ.get("GITHUB_REPOSITORY", "ROCm/TheRock")
+    # THEROCK_REPOSITORY is set by setup_multi_arch.yml to the repository input
+    # (ROCm/TheRock for external repos, or github.repository for normal runs).
+    github_repository = os.environ.get(
+        "THEROCK_REPOSITORY", os.environ.get("GITHUB_REPOSITORY", "ROCm/TheRock")
+    )
```
erman-gurses, line 481: This correctly redirects external-repo baseline queries to TheRock. However, with this new external-repo path, could we fail closed if the branch-history query below fails or returns no commits while STAGE_REUSE_CURRENT_SHA is set? 
The current fallback clears effective_commit_sha, allowing baseline selection without verifying compatibility with the pinned TheRock commit. It would be safer to select no baseline and rebuild in that case.

.github/workflows/setup_multi_arch.yml
```diff
@@ -177,14 +177,15 @@ jobs:
...
           STAGE_REUSE_MAX_AGE_HOURS: ${{ inputs.stage_reuse_max_age_hours }}
           STAGE_REUSE_COMMIT_HISTORY: ${{ inputs.stage_reuse_commit_history }}
-          # For cross-repo artifact reuse, use repository input (defaults to GITHUB_REPOSITORY).
-          # External repos set repository to "ROCm/TheRock" to reuse TheRock's prebuilt artifacts.
-          THEROCK_REPOSITORY: ${{ inputs.repository }}
-          # Skip path filtering for external repos (git sha won't exist in TheRock)
-          SKIP_PATH_FILTERS: ${{ inputs.external_repo != '' && 'true' || '' }}
+          # For baseline run selection, use repository input (defaults to GITHUB_REPOSITORY).
+          # External repos set repository to "ROCm/TheRock" to fetch TheRock baseline runs.
+          THEROCK_REPOSITORY: ${{ inputs.repository || github.repository }}
```
erman-gurses, line 187: The PR description mentions adding GITHUB_TOKEN for the baseline API calls, but I don't see it passed into this step's environment.

Should this include:

`GITHUB_TOKEN: ${{ github.token }}`

The stage-reuse path now queries branch history and baseline workflow runs, so explicitly passing the workflow token would ensure those requests are authenticated.

build_tools/github_actions/configure_external_repo_ci.py
```diff
@@ -58,8 +64,10 @@
 ]
 
 # Patterns that trigger a full test run when changed (CI infrastructure)
+# NOTE: .github/workflows/therock* is intentionally excluded since workflow
+# changes should still use stage reuse to determine which stages to rebuild.
+# The workflow itself doesn't affect TheRock build stages.
 FULL_TEST_TRIGGER_PATTERNS = [
-    ".github/workflows/therock*",
```
erman-gurses, line 62: I understand the intent to let workflow-only changes reuse unaffected build-stage artifacts.

Could you clarify how removing this pattern affects test coverage downstream? FULL_TEST_TRIGGER_PATTERNS controls run_all_tests, so a change only under .github/workflows/therock* now appears to fall through with `run_all_tests=False` and no matched projects.

Stage reuse and test selection seem independent: we could reuse unchanged build artifacts while still running the full test suite to validate the workflow being modified. Is that still guaranteed elsewhere?

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -330,6 +365,17 @@ def test_release_pr_without_submodule_change_runs(self):
...
+    @patch("configure_multi_arch_ci.is_ci_run_required")
+    def test_external_repo_skips_path_filter(self, mock_filter):
+        """External repo builds skip path filtering and always run CI."""
+        inputs = self._inputs(
+            external_repo='{"repository":"ROCm/rocm-libraries","ref":"abc123"}'
+        )
+        git = cm.GitContext(changed_files=["rocm-libraries"])
+        self.assertFalse(cm.should_skip_ci(inputs, git))
+        # Path filter should not be called for external repos
+        mock_filter.assert_not_called()
```
erman-gurses, line 384: can you add a test covering the full external-repo stage-reuse path?

The tests verify that EXTERNAL_REPO is parsed and that path filtering is skipped, but they do not check that an external repository name is passed into stage-impact analysis and that unaffected stages are applied as PREBUILT from a TheRock baseline.

erman-gurses (commented):
Please see my comments:

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -511,6 +529,12 @@ def _default_baseline_selector(*, platform: str) -> BaselineSelector:
             )
             ordered_commit_shas = None
         if not ordered_commit_shas:
+            if is_external_repo:
+                logger.warning(
+                    "%s empty branch history for external repo; "
+                    "failing closed - no baseline will be selected."
+                )
```
erman-gurses, line 537: Can you add LOG_PREFIX here, as in the warning above?

.github/workflows/setup_multi_arch.yml
```diff
@@ -177,14 +177,17 @@ jobs:
...
+          # Token for GitHub API calls (branch history, baseline workflow runs)
+          GITHUB_TOKEN: ${{ github.token }}
```
erman-gurses, line 189: Thanks for adding GITHUB_TOKEN. Do we also need actions: read in the workflow permissions?

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -608,6 +661,34 @@ def test_asan_tests_only_run_on_nightly_triggers(self):
...
+        result = cm.decide_jobs(
+            self._inputs(
+                external_repo='{"repository":"ROCm/rocm-libraries","ref":"abc123"}'
+            ),
+            git_context=git,
+            targets=cm.TargetSelection(),
+        )
+
+        # auto_stage_reuse should have processed the external repo context
+        self.assertIsNotNone(result.auto_stage_reuse)
```
erman-gurses: This test does not seem to exercise stage-impact analysis or baseline reuse because TargetSelection() has no Linux or Windows families. compute_auto_stage_reuse() therefore exits early with no build platforms selected.

TargetSelection() only constructs an object with empty family lists; actual target selection happens through select_targets(ci_inputs).

https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/build_tools/github_actions/configure_multi_arch_ci.py#L336-L341

Actual target selection happens here:
https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/build_tools/github_actions/configure_multi_arch_ci.py#L1361

As a result, assertIsNotNone(result.auto_stage_reuse) only verifies that a result object was created. It does not verify the behavior described by the test name and docstring.

Could we either:

- make this a plumbing test by mocking compute_auto_stage_reuse() and asserting that changed_files=["rocm-libraries"] is passed, an unaffected stage becomes PREBUILT, the baseline run ID is propagated, and baseline_repository remains ROCm/TheRock; or
- provide non-empty target families and a fake baseline so the actual stage-impact/reuse path is exercised?

build_tools/github_actions/stage_reuse_decision.py
```diff
@@ -511,6 +529,12 @@ def _default_baseline_selector(*, platform: str) -> BaselineSelector:
...
+                return lambda required_artifacts: None
```
erman-gurses, line 538: Non blocker: we could add unit tests for both external-repo fail-closed cases. 

- gha_query_recent_branch_commits() raises GitHubAPIError.
- It returns an empty list.

erman-gurses (commented):
Some additional comments:

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -309,6 +314,22 @@ def empty() -> "GitContext":
...
 
+    @staticmethod
+    def from_external_repo(external_repo_name: str) -> "GitContext":
+        """Create context for external repo builds (e.g., rocm-libraries).
+
+        For external repos, we treat the repo name as both a changed file and
+        a submodule path so that:
+        1. Stage reuse analysis can determine which TheRock stages are affected
+        2. has_submodule_changes returns True, enabling submodule_bump_tests_only
+           families to run their tests
```
ScottTodd, line 325: This doesn't need to be part of this PR, but let's consider removing or reworking `submodule_bump_tests_only` to simplify here. I want the CI configuration to be as explicit as possible so it is easier to understand and edit.

The behavior that `submodule_bump_tests_only` provides is:
* `push` events include "postsubmit" amdgpu families in addition to "presubmit" amdgpu familes
* `pull_request` events can opt-in to any amdgpu families they want via labels ([docs here](https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md#pull-request)).
  * _If a PR that does not modify submodules opts in via the `ci:run-all-archs` or a `gfx****` label, the build will be included but not the tests
  * Our automated submodule update PRs apply the `ci:run-all-archs` label

Here are a few proposals that I think would simplify:
1. Remove the "postsubmit" category entirely and rely on just "presubmit" (with opt-ins, notably all automated submodule update PRs) and "nightly". If someone adds an opt-in label to a PR that doesn't update submodules, run the tests as requested.
2. Turn `submodule_bump_tests_only` into `pull_request_tests_only` to skip running tests on `push` but still run them on `pull_request` (any opt-in, submodule update or otherwise)

We should also consider moving gfx950 from postsubmit/nightly to presubmit (at least in rocm-systems, later rocm-libraries and TheRock) given its increasing importance. If we don't have enough runner capacity for tests, we could also split out "build targets" from "test targets" to give a more directly configurable way to say "this PR shouldn't just build gfx950, it should also run gfx950 tests". I don't want to hide "run tests" behind requirements like "only if the PR touches these files".

Here's the code:

* https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/build_tools/github_actions/configure_multi_arch_ci.py#L1085-L1096
* https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/build_tools/github_actions/amdgpu_family_matrix.py#L293-L322

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -594,6 +618,12 @@ def should_skip_ci(
     if "ci:asan" in ci_inputs.pr_labels and ci_inputs.build_variant == "asan":
         print("  Running: 'ci:asan' PR label triggers ASAN CI")
 
+    # External repo builds skip path filtering - they always run CI and use
+    # stage reuse to determine which stages to rebuild.
+    if ci_inputs.external_repo:
+        print("  External repo build: skipping path filter checks, using stage reuse")
+        return False
```
ScottTodd, line 627: This doesn't seem flexible enough, at least on the surface. If a PR to rocm-libraries or rocm-systems only modifies non-code files like `*.md` documents or code files that are *known* to not affect CI (e.g. github actions workflow files unrelated to TheRock CI, .gitignore files, CMake files for experimental projects not yet integrated into the build system, etc.), the CI should still be able to short-circuit and skip. We shouldn't need to run expensive logic that decides "no builds affected, copy prebuilt files" and then "no tests affected, skip tests" IMO.

Can we reuse the skip path filters that currently exist in those repositories like https://github.com/ROCm/rocm-libraries/blob/7df08c3d5d7ca4087a21810122f19d65fc102975/.github/scripts/therock_configure_ci.py#L28-L70 ?

We could start with this as you have it, but we get frequent requests for ways to skip CI where it isn't applicable, so I don't want to regress there. For example,
* the docs team is accustomed to adding a `ci:docs-only` label to PRs and having that skip TheRock CI
* changes to https://github.com/ROCm/rocm-libraries/tree/develop/tools/libraries_pr_bot should skip CI
* changes to https://github.com/ROCm/rocm-systems/tree/develop/experimental can skip TheRock CI
* changes to https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/update-docs.yml can skip TheRock CI

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1393,15 +1423,23 @@ def configure(ci_inputs: CIInputs, git_context: GitContext) -> CIOutputs:
...
+        # External repo: use repo name for stage reuse analysis.
+        # external_repo is JSON like {"repository":"ROCm/rocm-libraries","ref":"..."}
+        try:
+            external_repo = json.loads(ci_inputs.external_repo)
+            repo_full_name = external_repo.get("repository", "")
+            external_repo_name = (
+                repo_full_name.split("/")[-1] if "/" in repo_full_name else ""
+            )
+        except (json.JSONDecodeError, TypeError):
+            external_repo_name = ""
```
ScottTodd: `JSONDecodeError` seems serious here. Might want to raise an exception rather than fallback to empty string, or at least log an error.

* If the CI system is misconfigured, we should get an obvious error that we can fix
* If github sometimes produces some output we don't expect, it is likely to also produce _other_ output we don't expect elsewhere, so just continuing with fallback behavior seems risky

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: ah good point! seems we just need to advance our labelling system and make it known. 

As I was making updates for gfx90a, it was pretty messy with all the "conditionals" we are doing for postsubmit due to limited capacity. 100% agree! I'll do this in a follow up PR as this will also impact gfx90a

plus as this becomes the norm, we can educate ROCm devs about this and utilize (if they know it impacts gfx950, enable it!)

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: yes! do not worry, I am aware of this. I'm trying to make incremental PRs all in parallel to handle scoped items so it isn't massive! this is at least letting CI know that rules that apply to TheRock do not apply to external repo (such as CMakeLists.txt edit, don't run everything!)

ScottTodd: Could add a TODO here or link to an issue with the plan so we don't lose track and if someone later finds the code they know what the ideal/planned state is.

ScottTodd (approved):

erman-gurses (approved):

dysu-amd (approved):
TBH I'm not familiar enough with the content to give a meaningful review, but LGTM

PR #7133, opened 2026-08-05
fix(ci): Fix ASAN builds to build native Linux packages, skip Python packages

geomin12:
ASAN builds should:
- Build native Linux packages (deb/rpm) - set build_native_linux=True
- Skip Python packages - add new build_python_packages field, set to False for ASAN

This fixes the issue where ASAN builds were incorrectly building Python packages instead of native Linux packages during submodule bumps.

### DIFF
--- a/build_tools/github_actions/configure_multi_arch_ci.py
+++ b/build_tools/github_actions/configure_multi_arch_ci.py
@@ -494,6 +494,7 @@ class BuildConfig:
     build_variant_suffix: str
     build_variant_cmake_preset: str
     build_native_linux: bool
+    build_python_packages: bool
     build_pytorch: bool
     build_jax: bool
     test_python_packages_matrix: list[dict[str, str]] = field(default_factory=list)
@@ -1154,6 +1155,9 @@ def _expand_build_config_for_platform(
     # Python package build is disabled. Then multi_arch_ci_* can also condition
     # build_python_packages on that decision.
 
+    # ASAN builds native Linux packages (deb/rpm) but not Python packages.
+    is_asan = suffix == "asan"
+
     return BuildConfig(
         per_family_info=per_family_info,
         dist_amdgpu_families=dist_amdgpu_families,
@@ -1161,7 +1165,8 @@ def _expand_build_config_for_platform(
         build_variant_label=variant_config["build_variant_label"],
         build_variant_suffix=suffix,
         build_variant_cmake_preset=variant_config["build_variant_cmake_preset"],
-        build_native_linux=(suffix != "asan"),
+        build_native_linux=True,
+        build_python_packages=(not is_asan),
         build_pytorch=build_pytorch,
         build_jax=build_jax,
         pytorch_build_matrix=pytorch_build_matrix,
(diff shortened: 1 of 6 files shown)


build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1658,9 +1660,15 @@ def test_windows_workflow_uses_all_ci_fields(self):
         workflow_path = WORKFLOWS_DIR / "multi_arch_ci_windows.yml"
         yaml_fields = self._extract_build_config_fields(workflow_path)
         python_fields = {f.name for f in fields(cm.BuildConfig)}
-        # build_native_linux is Linux-only. JAX builds are release-only and
-        # Linux-only for now, so Windows CI workflows do not consume them.
-        unused_fields = {"build_native_linux", "build_jax", "jax_build_matrix"}
+        # build_native_linux and build_python_packages are Linux-only. JAX builds
```
marbre: `build_python_packages ` is not Linux-only.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1658,9 +1660,15 @@ def test_windows_workflow_uses_all_ci_fields(self):
...
+        # are release-only and Linux-only for now, so Windows CI workflows do not
+        # consume them.
+        unused_fields = {
+            "build_native_linux",
+            "build_python_packages",
```
marbre: Why does Windows no longer uses `build_python_packages`? This seems wrong.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -1154,14 +1155,18 @@ def _expand_build_config_for_platform(
...
         dist_amdgpu_families=dist_amdgpu_families,
         artifact_group=f"multi-arch-{suffix or 'release'}",
         build_variant_label=variant_config["build_variant_label"],
         build_variant_suffix=suffix,
         build_variant_cmake_preset=variant_config["build_variant_cmake_preset"],
-        build_native_linux=(suffix != "asan"),
+        build_native_linux=True,
+        build_python_packages=(not is_asan),
         build_pytorch=build_pytorch,
         build_jax=build_jax,
```
ScottTodd, line 1174: We could also plumb this up to `setup_multi_arch.yml` as was done for pytorch and jax:
* https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/.github/workflows/multi_arch_ci_asan.yml#L51-L60
* https://github.com/ROCm/TheRock/blob/e94ddfdc1789d4ff2e80fa64bf52d6fdb2d3828b/.github/workflows/multi_arch_ci.yml#L98-L112

That will let us have `workflow_dispatch` triggers in TheRock and multi-arch CI workflow runs in rocm-libraries/rocm-systems/etc. choose if they want to build python packages (and also native packages), instead of just relying on this lower level config code.

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
ScottTodd: Glad I added those tests 😅 that would have been subtle / easy to miss otherwise

ScottTodd (approved):

ScottTodd (approved):

Issue #7139, opened 2026-08-06
hip-tests: Unit_hip_linker_spirv_input and Unit_hipExtModuleLaunchKernel_CheckCodeObjAttr fail with Compiler ww28 SMP 2.5


lajagapp:
## Summary

Two hip-tests UTs fail in TheRock Multi-Arch CI when built with the Compiler ww28 SMP 2.5 submodule promotion ([TheRock#7052](https://github.com/ROCm/TheRock/pull/7052), amd-llvm `7287e2062fa6`). These failures block compiler promotion CI until resolved.

## Failing tests

| Test | Source |
|------|--------|
| `Unit_hip_linker_spirv_input` | `rocm-systems/projects/hip-tests/catch/unit/module/hipLinkCreate.cc` |
| `Unit_hipExtModuleLaunchKernel_CheckCodeObjAttr` | `rocm-systems/projects/hip-tests/catch/unit/module/hipExtModuleLaunchKernel.cc` |

## CI impact

- **Component:** `hip-tests` (`build_tools/github_actions/test_executable_scripts/test_hiptests.py` → ctest over `share/hip/catch_tests`)
- **Platforms / families observed:**
  - Linux: `gfx94X-dcgpu`, `gfx950-dcgpu`
  - Windows: `gfx110X-all`
- **Blocking PR:** [TheRock#7052](https://github.com/ROCm/TheRock/pull/7052)

## Workaround

Temporarily exclude both tests via `TEST_TO_IGNORE` in `build_tools/github_actions/test_executable_scripts/test_hiptests.py` until a runtime/compiler fix lands. CI disable PR: _link here after opened_.

## Expected fix

Fix should land in **rocm-systems (hip-tests)** and/or **compiler (amd-llvm)** as appropriate. Re-enable TheRock CI exclusions once fixed on main.

## References

- Compiler promotion: [TheRock#7052](https://github.com/ROCm/TheRock/pull/7052)
- amd-llvm submodule: `7287e2062fa6`

PR #7052, opened 2026-08-01
chore(compiler): set SMP bump for Compiler ww28 SMP 2.5

ronlieb:
set submodule pointer for Compiler ww28 SMP 2.5

llvm   7287e2062fa6
spirv  28fa9b05
hipify a27e6519

JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-HOTSWAP
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2150
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2241
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2458
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2500
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2403
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2452
JIRA ID : https://amd-hub.atlassian.net/browse/LCOMPILER-2491

CP's
LCOMPILER-2491 64fa6b89b8d5 [MLIR][OpenMP] Restore debug-info fixup for declare target function. (#3654)
LCOMPILER-2382 ef833c9b6390 Revert "Revert "device-libs: Move trig edge case to input (#1700)""

CP's:
LCOMPILER-2452 4b2adbad0efb Pass `-O0` by default for HIP no-RDC compilations
LCOMPILRE-2403 7f4d33f95b31 [AMDGPU] Fix instruction size of LDS-DMA buffer loads (#211302)
LCOMPILER-2500 1d9f8c45183c Revert "[clang] Handle constructor closures with consteval default args (#203554)
LCOMPILER-2458 6c3996842e34 Revert "[AMDGPU] Select the high-half 16-bit packing idiom to v_lshl_or_b32 (#206058)
LCOMPILER-2382 21c4ca9d9a6e device-libs: Guard trig reduction quadrant index against NaN UB (#2752)
     ""        96c977988aa3 [DAGCombiner] Fold NaN-guard fptosi/fptoui select to saturating variant (#201435)
LCOMPILER-2241 795861910c68 Revert "[AMDGPU] Enable runtime loop unrolling (#194924)"
LCOMPILER-2150 973d77e112f4 Revert "device-libs: Move trig edge case to input (#1700)"

CP's:
[...]

PR #7144, opened 2026-08-06
hip-tests: Disable two hip-tests UTs blocked by Compiler ww28 SMP 2.5

lajagapp:
## Summary

Temporarily exclude `Unit_hip_linker_spirv_input` and `Unit_hipExtModuleLaunchKernel_CheckCodeObjAttr` from TheRock hip-tests CI so [TheRock#7052](https://github.com/ROCm/TheRock/pull/7052) (Compiler ww28 SMP 2.5) promotion is not blocked.

## Motivation

Multi-Arch CI hip-tests fail with amd-llvm `7287e2062fa6` from #7052. These issues will be fixed in rocm-systems, which is exposed by the compiler promotion PR. CI exclusion is a short-term unblock until runtime/compiler owners deliver a fix.

## Test plan

- Multi-Arch CI hip-tests pass on gfx94X / gfx950 (Linux) and gfx110X (Windows) with #7052 compiler artifacts
- Confirm excluded tests are skipped in CI logs (`ctest --exclude-regex` in `test_hiptests.py`)
- After #7139 is resolved, remove `TEST_TO_IGNORE` entries and re-run hip-tests on main

## Related

- Unblocks: [TheRock#7052](https://github.com/ROCm/TheRock/pull/7052)
- Tracking issue: #7139

### DIFF
--- a/build_tools/github_actions/test_executable_scripts/test_hiptests.py
+++ b/build_tools/github_actions/test_executable_scripts/test_hiptests.py
@@ -71,9 +71,6 @@ TEST_TO_IGNORE = {
             "Unit_NonHost_Printf_loop",
             "Unit_NonHost_Printf_multiple_Threads",
             "Unit_NonHost_Printf_BufferAvailability",
-            # TODO(#7139): Compiler ww28 SMP 2.5 (TheRock#7052) — re-enable after fix.
-            "Unit_hip_linker_spirv_input",
-            "Unit_hipExtModuleLaunchKernel_CheckCodeObjAttr",
         ]
     },
     "gfx94X-dcgpu": {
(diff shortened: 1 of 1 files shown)


ronlieb (approved):

ronlieb:
Is this landable now from DevOps perspective ?

ScottTodd (approved):

ScottTodd:
Yes, though the list of skipped targets may not be exhaustive in that file. We run nightly tests on all devices we have test runners for in https://github.com/ROCm/rockrel/actions/workflows/test_artifacts.yml

lajagapp:
> Yes, though the list of skipped targets may not be exhaustive in that file. We run nightly tests on all devices we have test runners for in https://github.com/ROCm/rockrel/actions/workflows/test_artifacts.yml

Thanks. Yes, it only covered the three family/platform combinations that ran in the #7052 presubmit, but nightly tests around nine families, so these two tests would still fail there.

Neither failure depends on the GPU architecture. So, I've pushed a follow-up commit that moves both into a `GENERIC_TEST_TO_IGNORE` list. That list applies to every family and platform. The real arch-specific skips stay where they are.

ronlieb (approved):

ScottTodd (approved):

PR #7160, opened 2026-08-06
ci(pytorch): enable gfx1250 builds for release/2.13

ethanwee1:
enable 1250 builds for release/2.13

### DIFF
--- a/build_tools/github_actions/tests/configure_pytorch_release_matrix_test.py
+++ b/build_tools/github_actions/tests/configure_pytorch_release_matrix_test.py
@@ -114,26 +114,6 @@ class ConfigurePytorchReleaseMatrixTest(unittest.TestCase):
                     ],
                 )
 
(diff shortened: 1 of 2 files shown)


build_tools/github_actions/tests/configure_pytorch_release_matrix_test.py
```diff
@@ -114,6 +114,26 @@ def test_filters_exact_unsupported_family(self):
...
+        self.assertEqual(
+            matrix,
+            [
+                {
+                    "python_version": "3.12",
+                    "pytorch_git_ref": "release/2.13",
+                    "amdgpu_families": "gfx125X-dcgpu",
+                }
+            ],
+        )
```
ScottTodd: I think the existing test cases should be sufficient here

ScottTodd (approved):

Issue #6207, opened 2026-06-29
[Issue] CI (Linux gfx942/gfx94X-dcgpu): MIOpen build failure — `HOST_ASAN` passes `-fsanitize=address` to device compile without `xnack+`


amd-chiranjeevi:
## Description

MIOpen fails to build under the `host-asan` CI workflow on gfx942 / gfx94X-dcgpu. The root cause is in TheRock's `cmake/therock_sanitizers.cmake`: when `THEROCK_SANITIZER=HOST_ASAN`, it appends `-fsanitize=address` to `CMAKE_CXX_FLAGS_INIT` (same as full ASAN) but does **not** rewrite `GPU_TARGETS` to `gfx942:xnack+` (unlike full ASAN). As a result, every MIOpen HIP TU is compiled with both `-fsanitize=address` and `--offload-arch=gfx942` (no xnack+), which clang rejects. Because MIOpen builds with `-Werror`, the `[-Woption-ignored]` diagnostic becomes a fatal error and kills the entire build.

This is a **regression introduced by TheRock PR #6180** (commit `14b3cb73`, merged 2026-06-26 18:50 UTC), which switched bump PRs and push events from full `asan` to `host-asan`. Before that commit, math-libs on gfx942 passed (e.g. run `28214286898` from 2026-06-26 02:59). Failures started at 18:50–18:52 UTC, within minutes of the merge.

## Context

- **Platform:** Linux
- **Arch:** gfx942 (gfx94X-dcgpu target)
- **Workflow:** Multi-Arch CI ASAN
- **Repo:** ROCm/TheRock
- **Regression commit:** https://github.com/ROCm/TheRock/commit/14b3cb73cacb462323699b6d839eda103d9c7d9e
- **Regression PR:** https://github.com/ROCm/TheRock/pull/6180
- **Example failing PR run:** https://github.com/ROCm/TheRock/actions/runs/28306898700
- **Failing job:** https://github.com/ROCm/TheRock/actions/runs/28306898700/job/83878700553
- **Last passing ASAN run (pre-regression):** https://github.com/ROCm/TheRock/actions/runs/28214286898

| Platform | GPU Arch      | Job ID      |
|----------|---------------|-------------|
| Linux    | gfx94X-dcgpu  | 83878700553 |

## Log Excerpt

```
[MIOpen] FAILED: src/CMakeFiles/MIOpen.dir/expanduser.cpp.o
[...]

PR #7164, opened 2026-08-06
fix(ci): Add -Xarch_host flags to HOST_ASAN builds

estewart08:
Recently HOST_ASAN is default for Multi-Arch CI to improve efficiency: (https://github.com/ROCm/TheRock/pull/7110)

There is an issue with HOST_ASAN Multi-Arch CI. MIOpen uses -Werror and during HOST_ASAN
this reports a warning that -fsanitize=address is not supported unless xnack+ is
appended to the gfx target. Regular ASAN builds override the gfx942 target with
gfx942:xnack+ and do not see this issue.

-fno-gpu-sanitize is used to explicitly disable sanitizer checks for GPU device target
compilations, but allows host sanitization to remain. This option currently removes
driver arguments, which I believe to be a bug. Instead -Xarch_host is used to separate
host only sanitizer flags. The clang driver bug will be fixed in the next SMP update for amd-llvm.

ISSUE ID: 6207

## Motivation
Unblock sanitizer build.

### DIFF
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -47,10 +47,8 @@ function(therock_sanitizer_configure
       # Avoids: (1) -Woption-ignored on gfx942 without :xnack+ (MIOpen failure due to -Werror)
       #         (2) handleSanitizeOption dropping all device -I/-D with -fno-gpu-sanitize (Bug in Driver)
       # TODO: If this is indeed a Driver bug, then this can be replaced with -fno-gpu-sanitize when that is fixed.
-      string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
-      string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
-      string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-Xarch_host -fsanitize=${_sanitizer_string}>\n")
-      string(APPEND _stanza "  $<$<AND:$<LINK_LANGUAGE:C,CXX>,$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>>:-Xarch_host -shared-libsan>)\n")
+      string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -Xarch_host -fsanitize=address -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
+      string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -Xarch_host -fsanitize=address -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
     else()
       # TODO: Support ASAN_STATIC/TSAN_STATIC to use static sanitizer linkage. Shared is almost always the right thing,
       # so make the sanitizer imply shared linkage.
--- a/cmake/therock_sanitizers.cmake
+++ b/cmake/therock_sanitizers.cmake
@@ -42,10 +42,21 @@ function(therock_sanitizer_configure
       set(_sanitizer_string "thread")
     endif()
 
-    # TODO: Support ASAN_STATIC/TSAN_STATIC to use static sanitizer linkage. Shared is almost always the right thing,
-    # so make the sanitizer imply shared linkage.
-    string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -fsanitize=${_sanitizer_string} -fno-omit-frame-pointer -g\")\n")
-    string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -fsanitize=${_sanitizer_string} -fno-omit-frame-pointer -g\")\n")
+    if(_sanitizer STREQUAL "HOST_ASAN")
+      # Scope to host arch only so HIP device -cc1 never sees -fsanitize=*.
+      # Avoids: (1) -Woption-ignored on gfx942 without :xnack+ (MIOpen failure due to -Werror)
+      #         (2) handleSanitizeOption dropping all device -I/-D with -fno-gpu-sanitize (Bug in Driver)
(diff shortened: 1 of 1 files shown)


geomin12 (approved):
lgtm

estewart08:
There is an issue with -fno-gpu-sanitize where the Driver is dropping include entries (-I) looking at another fix/workaround.

cmake/therock_sanitizers.cmake
```diff
@@ -42,10 +42,21 @@ function(therock_sanitizer_configure
...
-    string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -fsanitize=${_sanitizer_string} -fno-omit-frame-pointer -g\")\n")
+    if(_sanitizer STREQUAL "HOST_ASAN")
+      # Scope to host arch only so HIP device -cc1 never sees -fsanitize=*.
+      # Avoids: (1) -Woption-ignored on gfx942 without :xnack+ (MIOpen failure due to -Werror)
+      #         (2) handleSanitizeOption dropping all device -I/-D with -fno-gpu-sanitize (Bug in Driver)
+      # TODO: If this is indeed a Driver bug, then this can be replaced with -fno-gpu-sanitize when that is fixed.
+      string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
+      string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
+      string(APPEND _stanza "add_link_options($<$<LINK_LANGUAGE:C,CXX>:-Xarch_host -fsanitize=${_sanitizer_string}>\n")
+      string(APPEND _stanza "  $<$<AND:$<LINK_LANGUAGE:C,CXX>,$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>>:-Xarch_host -shared-libsan>)\n")
```
BrianHarrisonAMD: These two lines don't link. The genex payload is two tokens, so CMake splits on the space and the `>` lands on the wrong one:

```
clang++ "\$<1:-Xarch_host" "-fsanitize=address>" "-shared-libsan>" ... -o m
clang++: error: no such file or directory: '$<1:-Xarch_host'
clang++: error: unsupported argument 'address>' to option '-fsanitize='
```

The existing unquoted form at line 69 works because each genex wraps a single token. Reproduces on Ninja and Make, CMake 3.25 policy and 3.28.

Simplest fix is to drop both lines. A HIP link is a host link, so the unconditional `add_link_options` below already produces the same runtime set (`libclang_rt.asan-x86_64.so`, `asan-preinit`, `asan_static`); `-Xarch_host` adds nothing at link time and currently just duplicates `-fsanitize=address`. `-Xarch_host` still matters on the compile flags above.

If you'd rather keep them, quoting plus `SHELL:` works:

```cmake
string(APPEND _stanza "add_link_options(\"$<$<LINK_LANGUAGE:C,CXX>:SHELL:-Xarch_host -fsanitize=${_sanitizer_string}>\"\n")
```

cmake/therock_sanitizers.cmake
```diff
@@ -42,10 +42,21 @@ function(therock_sanitizer_configure
...
-    # so make the sanitizer imply shared linkage.
-    string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -fsanitize=${_sanitizer_string} -fno-omit-frame-pointer -g\")\n")
-    string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -fsanitize=${_sanitizer_string} -fno-omit-frame-pointer -g\")\n")
+    if(_sanitizer STREQUAL "HOST_ASAN")
+      # Scope to host arch only so HIP device -cc1 never sees -fsanitize=*.
+      # Avoids: (1) -Woption-ignored on gfx942 without :xnack+ (MIOpen failure due to -Werror)
+      #         (2) handleSanitizeOption dropping all device -I/-D with -fno-gpu-sanitize (Bug in Driver)
+      # TODO: If this is indeed a Driver bug, then this can be replaced with -fno-gpu-sanitize when that is fixed.
+      string(APPEND _stanza "string(APPEND CMAKE_CXX_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
+      string(APPEND _stanza "string(APPEND CMAKE_C_FLAGS_INIT \" -Xarch_host -fsanitize=${_sanitizer_string} -Xarch_host -fno-omit-frame-pointer -Xarch_host -g\")\n")
```
BrianHarrisonAMD: Minor: this branch is gated on `HOST_ASAN` but interpolates `${_sanitizer_string}`, which reads as if TSAN could reach it. It can't, since TSAN takes the `else()`. Hardcoding `address` here would be clearer.

BrianHarrisonAMD:
Thanks for this fix!
I believe this will resolve #6207 once it merges.

estewart08:
I was experimenting with different ways to unblock HOST_ASAN as the original idea was to use `-fno-gpu-sanitize`. This unfortunately uncovered a bug in the clang driver that will be fixed in the next SMP bump of llvm-project. Hopefully that SMP bump will land this week.

TLDR: I will go back to just using `-fno-gpu-sanitize`.

cmake/therock_sanitizers.cmake
estewart08: I was experimenting with SHELL locally, but was running into more issues later in the build. Going to abandon this in favor of `-fno-gpu-sanitize` once the llvm fix lands.

estewart08:
Actually, I will go ahead and make the suggested changes while we wait for the llvm fix.

cmake/therock_sanitizers.cmake
estewart08: done

cmake/therock_sanitizers.cmake
estewart08: Dropped the two lines.

BrianHarrisonAMD:
> I was experimenting with different ways to unblock HOST_ASAN as the original idea was to use `-fno-gpu-sanitize`. This unfortunately uncovered a bug in the clang driver that will be fixed in the next SMP bump of llvm-project. Hopefully that SMP bump will land this week.
> 
> TLDR: I will go back to just using `-fno-gpu-sanitize`.

Yea I was also investigating this, and I hit the same issue I believe.
Happy with either approach!

BrianHarrisonAMD (approved):
Thanks for the updates.

LGTM!

Issue #6156, opened 2026-06-25
[Issue]: Missing items in pip install vs package manager install


pbhandar-amd:
### Problem Description

There are some items missing when doing a pip install of ROCm (`rocm[libraries,devel,device-gfx1100,profiler]==7.14.0a20260619"`) that are available when doing a package manager install (`amdrocm-core-sdk7.14-gfx1100`).

These missing items are as follows:

1. rdc (ROCm Data Center Tools)
2. libhipcxx
3. elfio

### Operating System

Ubuntu 24.04

### CPU

N/A

### GPU

Navi31XTX

### ROCm Version

7.14 June 18 nightly

### ROCm Component

rdc, rocprofiler

### Steps to Reproduce

Pip install repro
```
sudo apt install libatomic1 libquadmath0

# Install Python 3.12 or 3.13.
sudo apt install python3.12 python3.12-venv

# Setup Python virtual environment.
python3.12 -m venv .venv
source .venv/bin/activate

# Install ROCm wheels packages.
python -m pip install --pre --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ "rocm[libraries,devel,device-gfx1100,profiler]==7.14.0a20260619"
rocm-sdk init
```

Missing items
```
(.venv) root@b718d8451a74:/mnt# find .venv -name rdc -o -name elfio -o -name libhipcxx
(.venv) root@b718d8451a74:/mnt#
```

Whereas, if you do a package manager install, you will find these 3 items installed

```
## Register ROCm packages ##
sudo tee /etc/apt/sources.list.d/rocm.list << EOF
deb [trusted=yes] https://rocm.nightlies.amd.com/packages-multi-arch/deb/20260619-27797822902 stable main
EOF

sudo apt update
sudo apt install amdrocm-core-sdk7.14-gfx1100
```

Items not missing

```
root@b1795aae83f8:/mnt# find /opt/rocm -name rdc -o -name elfio -o -name libhipcxx
/opt/rocm/core-7.14/share/rdc
/opt/rocm/core-7.14/share/doc/rdc
/opt/rocm/core-7.14/share/elfio
/opt/rocm/core-7.14/include/rdc
/opt/rocm/core-7.14/include/elfio
/opt/rocm/core-7.14/include/libhipcxx
/opt/rocm/core-7.14/libexec/rdc
/opt/rocm/core-7.14/lib/rdc
/opt/rocm/core-7.14/lib/cmake/rdc
[...]

PR #7191, opened 2026-08-07
fix(packaging): include elfio dev artifact in rocm-sdk-devel wheel

lucbruni-amd:
## Motivation

Addresses #6156.

`elfio` ships in the deb/rpm packages (via `amdrocm-sysdeps`) but is absent
from every Python wheel.

## Technical Details

`elfio` is header-only (`add_library(elfio INTERFACE)`), so its only
non-empty component is `dev`. The wheel selectors never claimed it:
`core_artifact_filter` gates on `lib`/`run` components, and it was missing
from the `devel` `addl_artifact_names` list. Adding it there routes
`include/elfio/**` and `lib/cmake/elfio/**` into `rocm-sdk-devel`.

## Test Plan

Install nightly `rocm[devel]` built from this branch and check for the
headers under `_rocm_sdk_devel/include/elfio`.

## Test Result

Drove the real `populate_devel_files` path (as `_run_kpack_split` does)
over an artifact catalog of the real `nlohmann-json`/`flatbuffers` dev
artifacts plus a structure-faithful `elfio_dev_generic` fixture:

- With `elfio` in `addl_artifact_names`: `_devel` tarball gains
  `_rocm_sdk_devel/include/elfio/*.hpp` and `lib/cmake/elfio/*.cmake`.
- Control (elfio removed): those entries are absent; nothing else changes.

Full wheel build via nightly CI still pending.

## Submission Checklist

### DIFF
--- a/build_tools/build_python_packages.py
+++ b/build_tools/build_python_packages.py
@@ -377,6 +377,8 @@ def _run_kpack_split(
             "prim",
             "rocwmma",
             "libhipcxx",
+            # Header-only ELF reader.
+            "elfio",
             # Third party dependencies needed by hipDNN consumers.
             "flatbuffers",
             "nlohmann-json",


marbre (changes requested):
`elfio` is not ROCm core and a third-party dependency. ROCm wont become a full OS-like operating system distribution :) So I don't think this should be included "solely for consistency" and we rather need confirmation if this is really a requirement. Blocking until things got clarified.

lucbruni-amd:
Thanks @marbre. I'm in favour of closing this PR once clarification is given, so the linked issue can be closed without loose ends.

ScottTodd:
Oh the other hand: https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md
> The `rocm-sdk-devel` package is the catch-all for everything.

Is there much harm in including this?

lucbruni-amd:
@saadrahim your input is appreciated here.

ScottTodd (commented):
I'll defer to Marius and Saad on this one, marking as 'reviewed' for me.

PR #7197, opened 2026-08-07
feat(ci); Adding therock_ref workflow/script in TheRock to propogate to superrepos

geomin12:
From rocm-libraries (https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci.yml#L112-L139) and https://github.com/ROCm/rocm-libraries/blob/develop/.github/scripts/resolve_therock_ref.py, we are copying and creating a reusable workflow for other external repos to use

the goal is to resolve the ROCm/TheRock commit a multi-arch CI run should build against so we don't need to dynamically retrieve for superrepos

This is only needed for superrepos


geomin12:
@ScottTodd  mostly @tony-davis 's work, but I remember having a discussion in 8/6's meeting about this. feel free to take a scan!

geomin12:
few things from brain:
- logging needs to be better, why this commit hash?
- instead of time checking, why not good source check? can be paired with multi-arch re-use artifacts and also update here: https://github.com/ROCm/TheRock/blob/main/docs/development/ci_overview.md
-

build_tools/github_actions/resolve_therock_ref.py
```diff
@@ -0,0 +1,449 @@
...
+# SPDX-License-Identifier: MIT
+"""Resolve the ROCm/TheRock commit an external repo CI run should build against.
+
+Policy (see the "Merge-base-time TheRock ref" design):
+
+- On a ``pull_request`` event, pin TheRock to the tip of ``main`` as it existed
+  at the time of the PR branch's merge-base with its base branch. The pin stays
+  frozen while the author pushes new commits and only advances when they merge
+  or rebase the base branch back in. This keeps the ROCm build underneath a PR
+  stable during authoring and debugging.
```
ScottTodd, line 12: This looks brittle in risky ways, see my review over at https://github.com/ROCm/rocm-libraries/pull/9602#pullrequestreview-4953478484

Issue #7189, opened 2026-08-07
Tarball indexer creates misleading page title


marbre:
The page title in `index_generation_s3_tar.py` is determined by the following code:

https://github.com/ROCm/TheRock/blob/3a435130da465033df051fd18d93e2305ddda0a9/build_tools/index_generation_s3_tar.py#L98-L107

This is correct for the current, soon legacy, publishing but wrong for with regards to RFC0012 where the indexer be used for per-product local indexes which are not `ROCm SDK` (I think SDK here corresponds to `core`) furthermore, the bucket names won't match and as a result the page title will always be `ROCm SDK tarballs`.

I would suggest to 1. find a better default and push tis (without closing the ticket) and 2. either show the path or the per-product name as page title (and then close the issue after deploying the updated indexer).

PR #7198, opened 2026-08-07
Use a fixed "Tarballs" page title instead of bucket-name pattern matching

araravik-psd:
## Summary
  - Replaces the bucket-name pattern matching in `generate_index_s3()` (`index_generation_s3_tar.py`) that guessed a page title from substrings
  like "dev", "nightly"/"nightlies", and "prerelease" in the bucket name with a fixed `"Tarballs"` title.
  - Addresses the immediate fix requested in #7189: per RFC0012, this indexer will be repurposed for per-product local indexes that aren't ROCm
  SDK-specific, so bucket-name-based title guessing no longer applies and was silently falling back to a misleading "ROCm SDK tarballs"
  default.

ISSUE ID: #7189 

  ## Why
  The old logic hard-coded assumptions about bucket naming conventions ("ROCm SDK dev/nightly/prerelease tarballs") that don't hold once this
  indexer is used for arbitrary per-product roots. Rather than keep producing an incorrect/misleading title, this uses a neutral,
  always-correct default.

  ## Follow-up
  Per #7189, a follow-up change will make the title reflect the actual file path or product name being indexed. This PR only covers the
  immediate fix so it can be deployed without waiting on that larger change. Leaving #7189 open for the follow-up.

  Fixes the immediate part of #7189.

### DIFF
--- a/build_tools/index_generation_s3_tar.py
+++ b/build_tools/index_generation_s3_tar.py
@@ -99,7 +99,7 @@ def generate_index_s3(
         raise FileNotFoundError(f"No .tar.gz files found in bucket {bucket_name}.")
 
     # Page title
-    page_title = "Tarballs"
+    page_title = "Tarball index"
 
     # Prepare filter options and files array for JS
     gpu_families = extract_gpu_details(files)
--- a/build_tools/index_generation_s3_tar.py
+++ b/build_tools/index_generation_s3_tar.py
@@ -100,14 +100,7 @@ def generate_index_s3(
 
     # Page title
     bucket_lower = bucket_name.lower()
-    if "dev" in bucket_lower:
-        page_title = "ROCm SDK dev tarballs"
-    elif "nightly" in bucket_lower or "nightlies" in bucket_lower:
-        page_title = "ROCm SDK nightly tarballs"
-    elif "prerelease" in bucket_lower:
-        page_title = "ROCm SDK prerelease tarballs"
-    else:
-        page_title = "ROCm SDK tarballs"
+    page_title = "Tarballs"
 
     # Prepare filter options and files array for JS
     gpu_families = extract_gpu_details(files)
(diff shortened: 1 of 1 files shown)


build_tools/index_generation_s3_tar.py
```diff
@@ -99,15 +99,7 @@ def generate_index_s3(
...
-    bucket_lower = bucket_name.lower()
-    if "dev" in bucket_lower:
-        page_title = "ROCm SDK dev tarballs"
-    elif "nightly" in bucket_lower or "nightlies" in bucket_lower:
-        page_title = "ROCm SDK nightly tarballs"
-    elif "prerelease" in bucket_lower:
-        page_title = "ROCm SDK prerelease tarballs"
-    else:
-        page_title = "ROCm SDK tarballs"
+    page_title = "Tarballs"
```
ScottTodd: The "tarballs" name is pretty non-specific / ambiguous. I think we should still differentiate between:
* release channel
* product (rocm sdk, others)
* build variant (release, asan)

We also have an active RFC at https://github.com/ROCm/TheRock/pull/4438 that keeps using tar for Linux archives but uses zip for Windows archives.

Brainstorming some alternatives to just "Tarballs"
* "Index of {bucket name}"
* "Tarball archives"
* "Tarball index"

marbre: Tarball index works for me. The script is used for several different products (e.g. rvs, ROCm core) and release streams.

marbre: @araravik-psd ping.

araravik-psd: I have the a follow up task in the PR, 

`make the title reflect the actual file path or product name being indexed` I can open an issue to track this for the next iteration or we can iterate over it on this PR itself

marbre: Why not following Scott's suggestion for now and name it `Tarball index`?

araravik-psd: The Tarball index is a good suggestion and is more generic till we iterate over this and display the product name in the display page title. I have updated the PR to reflect Tarball index now. Thanks

ScottTodd (approved):

marbre (approved):

Issue #7120, opened 2026-08-05
[Issue] Linux gfx94X-dcgpu nightly: rocshmem comprehensive tests fail — `hipSetDevice(rank) invalid device ordinal `


TriveniTadapaneni:
## Context
- **Workflow:** `ROCm/rockrel .github/workflows/test_artifacts.yml` → `ROCm/TheRock .github/workflows/test_artifacts.yml` → `test_component.yml`
- **Failing run:** [Test Artifacts (gfx94X-dcgpu, nightly, comprehensive, release, linux-gfx942-1gpu-ccs-csp-ossci-rocm)](https://github.com/ROCm/rockrel/actions/runs/30967121312) (2026-08-05)
- **Failing job:** [test / Test rocshmem / Test rocshmem (shard 1/1) (gfx94X-dcgpu)](https://github.com/ROCm/rockrel/actions/runs/30967121312/job/92206596014)
- **Affected arch:** `gfx94X-dcgpu`
- **platform:** Linux

## Summary

Every rocshmem test in the gfx94X-dcgpu nightly fails immediately with `hipSetDevice(rank): invalid device ordinal`. The root cause is a runner-routing precedence bug in `test_artifacts.yml`: because the nightly workflow dispatches with `workflow_dispatch` and passes a non-empty `test_runs_on` (the single-GPU runner `linux-gfx942-1gpu-ccs-csp-ossci-rocm`), the `workflow_dispatch` guard at the top of the `test_runs_on` expression wins for **all** matrix components — overriding the per-component `multi_gpu_runner: linux-gfx942-8gpu-ossci-rocm` that is correctly configured for rocshmem.

rocshmem tests require ≥2 GPUs: they fork two SLR processes, each calling `hipSetDevice(rank)`. On a 1-GPU container (`ROCR_VISIBLE_DEVICES` maps a single physical GPU), the rank-1 process fails immediately.

## Error

```text
test 1
        Start   1: put_n2_w1_z1_1048576B_uuid

[...]

PR #7200, opened 2026-08-07
feat(ci); `test_component.yml` runner priority update for workflow_dispatch triggers

geomin12:
Previously, we were running into errors where the test_component ruleset was not obeyed due to workflow dispatch triggers in rockrel

Now, we only support workflow_dispatch inputs (if user-triggered) - working here: https://github.com/ROCm/TheRock/actions/runs/31195115985/job/92922362899#step:1:2

Otherwise, we obey the ruleset (if a bot triggered like here: https://github.com/ROCm/rockrel/actions/runs/31157157060), working here: https://github.com/ROCm/rockrel/actions/runs/31197168830/job/92942923530 with triggered  https://github.com/ROCm/rockrel/actions/runs/31216565619/job/92993606398#logs

the correct runners are used
Screenshot 2026-08-07 135047
https://github.com/ROCm/rockrel/actions/runs/31216565619/job/92993606746#step:1:2

Adding onto ISSUE ID: #7120

### DIFF
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -216,19 +216,25 @@ jobs:
       artifact_group: ${{ inputs.artifact_group }}
       amdgpu_families: ${{ inputs.amdgpu_families }}
       amdgpu_targets: ${{ inputs.amdgpu_targets }}
-      # Route to the appropriate runner:
-      #   1. Multi-GPU components use the multi-GPU runner (required for correct execution)
-      #   2. For workflow_dispatch, honor the explicit test_runs_on input if provided
-      #   3. Benchmark components use the dedicated benchmark runner
-      #   4. CPU-only components use standard GitHub runners (azure-linux-scale-rocm)
-      #   5. Per-component runner (set by fetch_test_configurations.py)
+      # Runner selection priority rules:
+      #
+      # 1. USER-TRIGGERED runs (workflow_dispatch or workflow_call): Honor test_runs_on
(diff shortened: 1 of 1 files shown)


.github/workflows/test_artifacts.yml
```diff
@@ -217,18 +217,20 @@ jobs:
...
+      #   2. Multi-GPU components use the multi-GPU runner (required for correct execution)
       #   3. Benchmark components use the dedicated benchmark runner
       #   4. CPU-only components use standard GitHub runners (azure-linux-scale-rocm)
       #   5. Per-component runner (set by fetch_test_configurations.py)
       test_runs_on: >-
         ${{
-          (matrix.components.multi_gpu_runner != '' &&
-            matrix.components.multi_gpu_runner) ||
-          (github.event_name == 'workflow_dispatch' &&
+          (!contains(github.actor, 'github-actions') &&
```
dileepr1, line 229: since 'workflow_dispatch' condition is dropped, will a pre-submit get caught in the first condition and get a 1gpu?

geomin12: workflow_dispatch condition was not dropped, it works only if workflow_dispatch of an actual user was done.
otherwise for github-actions bots, it uses the rules

for presubmit: that is a pull_request trigger so it will not be impacted and will follow the rules here

dileepr1: Thanks for the clarification, looks good!

dileepr1 (approved):

PR #7201, opened 2026-08-07
fix(test): add platform-specific llvm / rocm_sysdeps to LD_LIBRARY_PATH in CI

archana-ramalingam:
## Motivation
  - Add platform-specific llvm / rocm_sysdeps to LD_LIBRARY_PATH in FFM CI 
  - Without these paths, test binaries fail to resolve shared libraries when running from installed artifacts.
  
## Technical Details
  - Detect platform-specific LLVM lib subdir via `glob("*-linux-*")` and add to `LD_LIBRARY_PATH`
  - Add `lib/rocm_sysdeps/lib/` when it exists
  - Both paths are conditionally added (only when dirs exist on disk) — no breakage on other platforms

## Test Plan
  - `python -m pytest build_tools/github_actions/tests/pytest_runner_test.py` — 31/31 pass
  - Verified: in TheRock artifacts, `libomp.so` is installed under the platform triple subdir (`lib/llvm/lib/x86_64-unknown-linux-gnu/`), not directly in `lib/llvm/lib/`

## Test Result
- [x] 31/31 unit tests passed.
  
## Submission Checklist

JIRA ID : AIHPBLAS-3988


build_tools/github_actions/test_executable_scripts/pytest_runner.py
```diff
@@ -195,13 +195,23 @@ def build_environment(rocm_path, component_name):
...
-    # (in lib/llvm/lib/) — both are needed or the client segfaults at load.
+    # (in lib/llvm/lib/<triple>/ due to GNUInstallDirs) and librocm_sysdeps_z.so
+    # (in lib/rocm_sysdeps/lib/).
     lib_path = rocm_path / "lib"
     llvm_lib_path = rocm_path / "lib" / "llvm" / "lib"
+    sysdeps_lib_path = rocm_path / "lib" / "rocm_sysdeps" / "lib"
+    ld_paths = [str(lib_path), str(llvm_lib_path)]
+    # GNUInstallDirs puts libomp.so under a platform-specific triple subdir
+    # (e.g. x86_64-unknown-linux-gnu/) instead of lib/llvm/lib/ directly.
+    llvm_platform_dirs = list(llvm_lib_path.glob("*-linux-*"))
```
dileepr1: Please add a unit test as well to make sure this works.
Won't this path selection break once a new triple is added to the location?

dileepr1 (changes requested):

geomin12 (commented):
will let dileep review this as he has more context

build_tools/github_actions/test_executable_scripts/pytest_runner.py
archana-ramalingam: Good catch — switched `glob(...)[0]` to appending all `*-linux-*` triple dirs, sorted: deterministic, and the correct-arch dir is always on the path (loader skips wrong-arch libs). 

Chose this over `amdclang --print-target-triple` — glob-all's failure mode is benign, whereas host-triple resolves to one dir and silently misses libomp if it's staged under a different triple (cross/sysdeps).

Can switch to host-triple if you'd prefer matching rocprofiler_sdk?

archana-ramalingam: Added unit tests.

dileepr1 (approved):
Looks good to me with the changes.

Issue #7091, opened 2026-08-04
Add a read-only HTTP artifact backend, and make bucket layout and public/CDN URLs data-driven


PeterCDMcLean:
## Problem

There is no way to fetch build artifacts without S3 credentials, even though the artifact buckets
are publicly readable over HTTPS.

`artifact_manager fetch` only has an `S3Backend`, which needs boto3 and a credential chain. That
forces credentials on cases that shouldn't need them:

- a developer who wants the artifacts from a public CI run and has no AWS account;
- a CI job in a repository that consumes TheRock artifacts but has no reason to hold write
  credentials for the bucket;
- anything running where boto3 isn't installed.

The practical result is that consumers write their own HTML-index scrapers against the artifact
index pages. Those reimplementations tend to skip the sha256 verification that `S3Backend` does,
and they hardcode a URL layout they have no way to look up.

## Why a plain `HTTPBackend` isn't enough on its own

The URL a backend should fetch from is not derivable from the code today.

`StorageLocation.https_url` hardcodes `https://{bucket}.s3.amazonaws.com/{path}`. The CloudFront
distributions that actually front these buckets are recorded only as a prose markdown table in
[`docs/development/s3_buckets.md`](https://github.com/ROCm/TheRock/blob/main/docs/development/s3_buckets.md),
and buckets may store objects under a key prefix that the S3 URL and the CDN URL express
differently.

Because that table isn't machine-readable, every caller that needs a public URL re-derives it.
There are currently five hand-rolled copies of the release-type → CDN map:

- `build_tools/setup_venv.py`
- `build_tools/github_actions/publish_jax_to_release_bucket.py`
- `build_tools/github_actions/publish_pytorch_to_release_bucket.py`
- `build_tools/github_actions/summarize_test_pytorch_workflow.py`
- `build_tools/install_rocm_from_artifacts.py`

plus a duplicated S3 key-prefix formula in `build_tools/find_artifacts_for_commit.py`.

An `HTTPBackend` built on `https_url` would bake raw S3 URLs into the one component that most needs
to know about CDNs. So the bucket metadata has to become data first.

## Proposal

**One machine-readable source of truth for bucket layout and public URLs, and an HTTP artifact
backend that gets CDN awareness for free from it.**

### 1. Bucket-config-driven URI resolution

Extend the existing `build_tools/_therock_utils/s3_buckets.py` registry rather than adding a
parallel resolution path:

[...]

PR #4199, opened 2026-03-27
STACKED PR B: Introduce StorageConfig dataclass for URL schema configuration

PeterCDMcLean:
## Summary

Consolidate `s3_url_schema`, `https_url_schema`, and `bucket_schema` parameters into a single `StorageConfig` frozen dataclass. This addresses PR review feedback about threading ~280 lines of boilerplate through multiple classes.

**Key changes:**
      - Add `StorageConfig` dataclass with placeholder validation and `from_json()` parsing
      - Replace individual schema parameters with single `storage_config` throughout codebase
      - Replace `--s3-url-schema`, `--https-url-schema`, `--bucket-schema` CLI args with single `--storage-config` JSON arg
      - Fix `_fetch_index()` error handling (bare `except Exception` → specific HTTP/URL errors)
      - Remove unused `_discover_gfx_families_from_master_index` stub

## Stacked on

This PR is stacked on #4088 (`users/pmclean/workflow_summary_artifact_backend`)

## Test plan

      - [x] All 69 tests pass in `workflow_outputs_test.py`
      - [x] Pre-commit checks pass
      - [ ] CI validation

PR #7238, opened 2026-08-10
Make bucket layout and public/CDN URLs data-driven, including the repo.amd.com product buckets

PeterCDMcLean:
ISSUE ID : https://github.com/ROCm/TheRock/issues/7091

## Motivation

`StorageLocation.https_url` hardcodes `https://{bucket}.s3.amazonaws.com/{path}`, but most of
these buckets are fronted by a CDN. Those mappings live only as a prose table in
`docs/development/s3_buckets.md`, so nothing can read them — and everyone who needs a public URL
re-derives one by hand.

The repo.amd.com rewire (#7547, #7392, #7553) made this sharper rather than obsolete. RFC0012's
layout is a **key-prefix rewrite**:

```
s3://therock-repo-amd-nightly-core/v5/rocm/core/tarball/X
                 ↓  v5/ stripped
    https://nightly.repo.amd.com/rocm/core/tarball/X
```

Nothing in the tree derives that address. `v5/rocm/<product>/` is a hardcoded f-string in
`python_package_paths.py:127` and repeated through `publish_rocm_to_release_buckets.py`, and
`get_product_release_bucket_config` builds `therock-repo-amd-{stream}-{product}` from a formula
that `python_package_paths.py:65` now implements a second time — under a comment reading
"Centralized here". #7351 added a third copy of the `v5/rocm` prefix in
`download_python_packages.py:249`. The duplication this PR removes has been growing.

Downstream is the sharpest case. `rocm-npi-dev` monkey-patches
`WorkflowOutputRoot.from_workflow_run` in two scripts, stores its `v3/artifacts/` key prefix in
the **`external_repo`** field (a fork-namespace field — semantic abuse it documents itself), and
keeps two copies of an S3→CDN rewrite table installed by regex-sniffing TheRock's module layout
at run time. Its three rewrites are the same shape as `v5/`: one CDN host, three different
`v3/…` prefixes to strip.

## Technical Details

Four commits, reviewable in order.

**1. `CdnRule` and `.public_url`.** A frozen `CdnRule(key_prefix, url_prefix)` and a `cdn_rules`
field on `S3BucketConfig`. `StorageLocation.public_url` resolves through the registry, longest
matching prefix first, and falls back to `.https_url`.

`.https_url` deliberately stays the raw S3 URL rather than changing in place. It is named after a
fact; returning CloudFront from it removes any way to get the origin back, and would flip
behaviour at every call site the day someone adds a rule — in production CI for TheRock,
rocm-libraries, rocm-systems, llvm-project and rockrel. `storage_location.py` is bundled into the
artifact-index Lambda and has no project imports, so the resolver lives in `s3_buckets.py` behind
a function-local import.

**2. `key_prefix` on the bucket config.** It is a property of the bucket, not of a workflow run —
it applies to objects `WorkflowOutputRoot` never produces. Folded into `relative_path` at
construction so `relative_path` *is* the S3 key, which makes CDN stripping fall out of the
existing rule matching. This lets `external_repo` keep meaning "fork namespace", and the literal
`== "therock-ci-artifacts-external"` compare becomes a `namespace_external_repos` flag.

**3. Registry file.** `--bucket-config-file` on `artifact_manager.py`, or
`THEROCK_S3_BUCKETS_FILE` for wrapper scripts that shell out to several entry points. The flag
wins and the choice is logged. Merges are additive; shadowing an in-tree name is **rejected**
unless the entry sets `"override": true`, in which case it fully replaces and logs. Unknown keys
at any level are an error — a typo'd `cdn_rule` would otherwise mean "this bucket has no CDN",
exactly the silently-wrong-URL outcome the registry exists to prevent.

**4. Product buckets and `anonymous_s3_read`.** See below.

### Registering the v5 product buckets

`get_product_release_bucket_config` synthesized a config that was not in the inventory, so
`lookup_bucket_config` could not see it and every object in those buckets fell back to a raw S3
URL. They are now registered like any other bucket and the getter looks them up; the config it
returns is identical field for field, pinned by a test.

Each carries `key_prefix="v5/"` and a rule mapping `v5/` to `https://<stream>.repo.amd.com/`,
built from **one formula** rather than a per-stream table — RFC0012 gives every stream the same
hierarchy under its own subdomain, so a stream still cutting over needs no edit here when it
finishes.

### `anonymous_s3_read`

The split between machine-consumed URLs (raw S3, to avoid CloudFront egress charges) and
human-facing links (CDN) was a global policy standing in for a per-bucket fact — and the fact
does not hold globally. Probed anonymously on 2026-08-25 by requesting a key that does not exist,
so `404 NoSuchKey` distinguishes readable from `403 AccessDenied`:

| Readable | Refused |
| --- | --- |
| every `*-artifacts` bucket | every `*-packages` bucket |
| `dev`/`nightly` python + tarball | `prerelease`/`release` python + tarball |
| | all `therock-repo-amd-*` product buckets |

So the prerelease buckets are not merely expensive to read over raw S3, they are **unreadable** —
which is the root cause of #2139, and the same reason `rocm-npi-dev` passes a CloudFront URL as
its pip index in nine workflows.

`StorageLocation.download_url` now picks per bucket, and raises rather than returning a URL that
answers 403. **Nothing changes in TheRock today**: `package_find_links_url`,
`package_repository_url` and `tarball_urls` all resolve to artifacts buckets, which are readable
and keep the default.

### CDN rule verification

The inventory holds 24 rules across 16 distinct CDN prefixes. Probed 2026-08-25: every legacy
prefix and the `dev`, `nightly` and `rc` stream roots serve. Not yet reachable are
`bkc.repo.amd.com` (no DNS), `rc/rocm/core/tarball/`, and the `stable` paths — streams and paths
still being brought up.

That is deliberate. The registry encodes RFC0012's target layout, derived from a single
`f"https://{stream}.repo.amd.com/"` formula, so each becomes correct as it comes online with no
code change here. Emitting only "what serves today" would mean another round of hand-edits across
five files per stream.

Where a mapping is not a prefix rewrite at all, no rule is emitted: the prerelease and release
*package* CDNs serve a distro-partitioned apt/dnf repo, and no CDN is published for the ASAN
prefixes. No rule means the raw S3 URL, i.e. today's behaviour exactly.

### Blast radius

**Zero today.** Every `.public_url` call site migrated here resolves to a `therock-*-artifacts`
bucket, and all of those carry `cdn_rules=()`. `TestArtifactsBucketsHaveNoCdn` pins that, so a
future rule addition fails loudly rather than silently rewriting production CI job-summary links.

### Designs considered and rejected

Prior art: #4088, #4167, #4199. Both earlier URI designs were rejected in review and are not
revived:

- **`THEROCK_HTTPS_URL_<bucket>` env var** (#4088): one variable per bucket, no schema, no
  validation, cannot express key prefixes, cannot refuse to clobber a production bucket.
- **`StorageConfig` format-string templates** (#4199): rejected as encoding a schema with
  component replacement rules.

A per-bucket `cdn_url` string — the shape #7517 proposes — is not enough either. It supplies a
URL to prepend but no prefix length to strip, which is exactly what `v5/` and NPI's `v3/` need.
`CdnRule` pairs the two; for a whole-bucket mapping it degenerates to `CdnRule("", url)`.

The registry file differs from the rejected env var on every point: one pointer to one document,
validated on load, key-prefix aware, and shadowing requires explicit opt-in. It is still
process-wide state and the docs say so; the honest distinction is *explicit and validated*, not
*non-ambient*.

## Test Plan

- `cd build_tools && python -m pytest`
- `pre-commit run --all-files`
- Every CDN prefix and derived index URL fetched anonymously; every bucket probed for anonymous
  read. Results above.

New coverage in `s3_buckets_test.py`, `workflow_outputs_test.py` and
`find_artifacts_for_commit_test.py`: `CdnRule` normalization and rejection of malformed prefixes;
per-bucket round trips; longest-prefix-wins; registry file valid / invalid / duplicate /
unknown-key / unknown-version / shadow-without-override; `key_prefix` composition with and
without `external_repo`; the v5 round trip asserted as literal URLs;
`TestForkPrConfigPreservesNewFields`, which pins the one field-by-field config rebuild that
silently dropped every new field on fork PRs until it became a `dataclasses.replace`; and a
downstream round trip in the shape `rocm-npi-dev` needs, so an upstream refactor cannot break
that repo as silently as its module-sniffing patches allow today.

## Test Result

`cd build_tools && python -m pytest` on this branch (Linux, Python 3.12):
**2132 passed, 8 skipped, 244 subtests passed**, 0 failed.

`pre-commit run` clean across all 18 changed files.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -4,9 +4,16 @@
 """Inventory of S3 buckets used by CI/CD systems and related functions.
 
 See docs/development/s3_buckets.md.
+
+Repositories outside TheRock that reuse these build tools against their own
+buckets can extend the in-tree inventory with a JSON registry file, pointed at
+by the ``THEROCK_S3_BUCKETS_FILE`` environment variable or by an explicit
+``--bucket-config-file`` argument. See ``load_bucket_registry_file`` for the
+schema and merge rules.
 """
 
 from dataclasses import dataclass, field, replace
+import json
 import os
 import sys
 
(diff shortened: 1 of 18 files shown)


marbre (commented):
The repo structure is going to change with https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0012-Repo-Structure.md being implemented right now. I fear this patch cannot be ported straight forward to the new structure where we will have one bucket _per-product + release-stream_, whereas we currently have one bucket _per-package-type + release-stream_. As we urgently need to land the RFC0012 implementation, it might be best to hold this changes back for now. At least I can't invest in a deeper review to help landing this before we need to switch to the new repo structure.

PR #7240, opened 2026-08-10
Derive release index URLs instead of hand-copying them

PeterCDMcLean:
ISSUE ID : https://github.com/ROCm/TheRock/issues/7091

## Motivation

The public "which URL do I install this channel from" mapping is still written out by hand, and
after the repo.amd.com rewire there are now **two** families of it instead of one:

| Copy | Where |
| --- | --- |
| `ROCM_INDEX_URLS_MAP` (new stream URLs) | `setup_venv.py:56` |
| `LEGACY_MULTI_ARCH_INDEX_URLS` | `publish_pytorch_to_release_bucket.py:45` |
| the same dict again, verbatim | `publish_jax_to_release_bucket.py:27` |
| the nightly index as an argparse default | `summarize_test_pytorch_workflow.py:110` |

The legacy map did not go away when the streams landed — it was renamed and extended with BKC
keys. `main` already has `get_release_package_index_url`, but only for the aggregate wheel index,
and nothing uses it from `setup_venv.py`.

Stacked on #7239 for a linear review order; it shares no files with that PR.

## Technical Details

Extends `main`'s `get_release_package_index_url` rather than adding a rival name, and adds two
siblings:

```python
get_release_package_index_url("nightly")             # aggregate pip index
get_release_package_index_url("nightly", "pytorch")  # product-local index
get_release_tarball_index_url("release")             # where users browse tarballs
get_legacy_release_index_url("nightly")              # pre-RFC0012 layout
```

`get_legacy_release_index_url` derives from the selected bucket's `cdn_rules` rather than from a
copy of the mapping. The `v4/whl/` and `v4/tarball/` prefixes are now stated once, in
`_LEGACY_INDEX_KEY_PREFIXES`, and the seeded `CdnRule`s are built from it — the prefix a rule
matches and the prefix an index URL is built from have to be the same string or the lookup
silently misses, so they should not be two literals kept in sync by hand.

All eight derived URLs are **byte-identical** to the literals they replace.

### The read set is wider than the write set

`get_release_stream` and `get_release_bucket_config` keep rejecting `"release"`: the stable
buckets have no automated upload credentials — they are promoted by hand — so nothing may publish
there. But it is the channel most users install from, and `setup_venv.py`'s `stable` index already
pointed at it. `get_index_release_stream` accepts it and the publish-side lookups do not, which is
what lets one widen without widening the other.

The registry file's `release_buckets` map is validated against the same wider set, so a downstream
repo can redirect its `release` channel; validating it against the publish set would leave
`get_legacy_release_index_url` reading a slot no file could ever fill.

`setup_venv.py` keeps `stable` as the user-facing `--index-name` spelling, aliased to the
`release` release type, so no command lines change.

### A note on `stable`

`https://stable.repo.amd.com/rocm/whl-next/` currently returns 404 — that stream has not finished
cutting over. This PR does not introduce or fix that: it is the exact value `main` already holds
in `setup_venv.py` after #7547, and the consolidation preserves it. Flagging it because it means
`--index-name stable` does not work today, which is worth someone's attention independently of
this change.

## Test Plan

- `cd build_tools && python -m pytest`
- `pre-commit run --all-files`
- Every derived index URL fetched anonymously. The `dev` and `nightly` streams and every legacy
  index serve; `rc/rocm/core/tarball/` and the `stable` paths are not up yet.

Every expected URL in `TestReleaseIndexUrls` is asserted as a literal rather than rebuilt from the
same streams and rules the implementation reads. These go to pip, so a test that derived them
would agree with any rule the code happens to hold, including a wrong one. That includes an
explicit check that the four `--index-name` values resolve to exactly what `setup_venv.py` held
before, and that the five legacy values match the deleted `LEGACY_MULTI_ARCH_INDEX_URLS` entries.

Also covered: `"release"` is accepted by the read path and rejected by both publish-side lookups;
`"packages"` has no single index and raises, because that CDN serves a distro-partitioned apt/dnf
repo rather than a prefix rewrite.

## Test Result

`cd build_tools && python -m pytest` on this branch (Linux, Python 3.12):
**2188 passed, 8 skipped, 259 subtests passed**, 0 failed.

`pre-commit run` clean across all changed files.

## Submission Checklist

### DIFF
--- a/build_tools/_therock_utils/s3_buckets.py
+++ b/build_tools/_therock_utils/s3_buckets.py
@@ -159,6 +159,15 @@ _RELEASE_STREAM_BY_TYPE = {
     "prerelease": "rc",
 }
 
+# Channels a user can install *from*, deliberately wider than
+# _RELEASE_STREAM_BY_TYPE, which is the set a workflow can publish *to*. The
+# stable channel has no automated upload credentials - it is promoted by hand -
+# so the publish path must keep rejecting it, but it is the channel most users
+# install from, and setup_venv.py's "stable" index already pointed at it.
+# Keeping the read set and the write set separate is what lets one widen
+# without widening the other.
+_INDEX_STREAM_BY_RELEASE_TYPE = {**_RELEASE_STREAM_BY_TYPE, "release": "stable"}
+
 # The key prefix every product bucket stores under, stripped from the public URL:
 # s3://therock-repo-amd-nightly-core/v5/rocm/core/tarball/X is served at
 # https://nightly.repo.amd.com/rocm/core/tarball/X.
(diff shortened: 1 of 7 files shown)


marbre (commented):
https://github.com/ROCm/TheRock/pull/7238#pullrequestreview-4904502723:

> The repo structure is going to change with https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0012-Repo-Structure.md being implemented right now. I fear this patch cannot be ported straight forward to the new structure where we will have one bucket _per-product + release-stream_, whereas we currently have one bucket _per-package-type + release-stream_. As we urgently need to land the RFC0012 implementation, it might be best to hold this changes back for now. At least I can't invest in a deeper review to help landing this before we need to switch to the new repo structure.

PR #7243, opened 2026-08-10
feat(ci): introducing granular artifact reuse into multi arch CI

geomin12:
Enable granular artifact-level reuse for external repos (rocm-libraries, rocm-systems). When only specific components change (e.g., rocBLAS), skip rebuilding unaffected stages.

 Changes
  - Add component-to-artifact mapping in BUILD_TOPOLOGY.toml
  - Analyze changed_projects to determine reusable artifacts
  - Wire artifact reuse through CI workflows
  - Rename --projects to --artifacts in configure_stage.py
  - Remove deprecated consumer graph drift check

Working [rocm-libraries test](https://github.com/ROCm/rocm-libraries/actions/runs/33424918677/job/99604095908?pr=10487) from [pr](https://github.com/ROCm/rocm-libraries/pull/10487)

ISSUE ID: https://github.com/ROCm/TheRock/issues/3343

Next steps: 
- determine baseline run ID compatible within artifact-level granularity and ensure this is correct
- handle build dependencies and rebuild what it depends on (currently only handles one project)


BUILD_TOPOLOGY.toml
```diff
@@ -525,6 +526,9 @@ feature_group = "ALL"
 
 # --- Runtime Artifacts ---
 
+# Components map monorepo project directories to artifacts for granular CI reuse.
+# Component names are directory basenames within projects/ subdirs of monorepos.
```
ScottTodd: All fields should be documented at the top of the file in this block: https://github.com/ROCm/TheRock/blob/ac0283a86d02298655a998f722bfc3ac1328e6fa/BUILD_TOPOLOGY.toml#L1-L72

BUILD_TOPOLOGY.toml
```diff
@@ -607,76 +615,95 @@ artifact_group = "runtime-tests"
...
 artifact_deps = ["core-hip"]
 feature_group = "CORE"  # Part of core, enabled by default
+components = ["hip-tests"]
 
 # --- Math Libraries (per-arch) ---
 
+# Components map monorepo project/shared directories to artifacts for granular
+# CI reuse analysis. When changes are isolated to specific components, only the
+# affected artifacts need rebuilding while others can be reused from baseline.
+# Component names are directory basenames within projects/ or shared/ subdirs.
```
ScottTodd: Same here - document at the top of the file, not way down here.

BUILD_TOPOLOGY.toml
```diff
@@ -685,26 +712,31 @@ artifact_group = "ml-libs"
...
 artifact_group = "ml-libs"
 type = "target-neutral"
 artifact_deps = ["core-runtime", "core-hip", "spdlog"]
+components = ["hipdnn"]
 
 [artifacts.hipdnn-integration-tests]
 artifact_group = "ml-libs"
 type = "target-neutral"
 artifact_deps = ["core-runtime", "core-hip", "hipdnn"]
+components = ["integration-tests"]
```
ScottTodd: This component name should be qualified further, just "integration-tests" could apply to any of the 50+ rocm subprojects.

BUILD_TOPOLOGY.toml
```diff
@@ -831,13 +875,15 @@ artifact_group = "debug-tools"
...
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
+components = ["rocdbgapi"]
 
 [artifacts.rocr-debug-agent]
 artifact_group = "debug-tools"
 type = "target-neutral"
 artifact_deps = ["amd-llvm", "core-runtime", "amd-dbgapi", "sysdeps"]
 feature_group = "DEBUG_TOOLS"  # Part of DEBUG_TOOLS, enabled by default via THEROCK_ENABLE_ALL
 disable_platforms = ["windows"]
+components = ["rocr-debug-agent"]
```
ScottTodd: The majority of these `components =` are just matching the `artifacts.` name. There are just a few exceptions like `artifacts.base`, `artifacts.blas`, `artifacts.solver`, and `artifacts.core-hip`.

Ideas:
* Could this just default to the artifact name and only override when needed?
* Is the build topology the right place for this information in the first place?
* Do we need this granular a component<->artifacts mapping at this point? Isn't the math-libs stage with artifact reuse all-or-nothing right now? Is there a scenario where "rand" artifacts are used from prebuilt while "blas" artifacts are built from source?

Actually looking at https://github.com/ROCm/rocm-libraries/actions/runs/31208693701/job/92966415872?pr=10487#step:6:22, why does that need to copy all of `Copying 19 artifacts: blas, composable-kernel, fft, hipblasltprovider, hipdnn, hipdnn-integration-tests, hipdnn-samples, hipkernelprovider, hiptensor, hipthreads, libhipcxx, miopen, miopenprovider, rand, rocalution, rocwmma, solver, sparse, support` if all that wanted to do was build/test rocprim?

build_tools/configure_stage.py
```diff
@@ -380,7 +381,7 @@ def main(argv: List[str] = None):
         include_comments=args.comments and not args.oneline,
         platform_name=args.platform,
         manylinux=args.manylinux,
-        project_names=args.projects,
+        project_names=artifact_list if artifact_list else None,
```
ScottTodd: nit: is there any value in passing `artifact_list if artifact_list else None` over just `artifact_list` here? What other falsey values might it have?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -161,6 +213,8 @@ class CIInputs:
     baseline_run_id: str = ""
     # Repository to query for baseline runs (for cross-repo artifact reuse)
     baseline_repository: str = ""
+    # Changed projects from external repos (e.g., "projects/rocprim,projects/hipcub")
+    changed_projects: list[str] = field(default_factory=list)
```
ScottTodd, line 294: How does this handle https://github.com/ROCm/rocm-libraries/tree/develop/shared and https://github.com/ROCm/rocm-systems/tree/develop/shared ? There are code paths that are used by multiple projects.

.github/workflows/multi_arch_ci_linux.yml
```diff
@@ -55,7 +55,7 @@ permissions:
 jobs:
   copy_prebuilt_stages:
     name: Copy Prebuilt Stages
-    if: ${{ fromJSON(inputs.build_config).prebuilt_stages != '' && fromJSON(inputs.build_config).baseline_run_id != '' }}
+    if: ${{ (fromJSON(inputs.build_config).prebuilt_stages != '' || fromJSON(inputs.build_config).reusable_artifacts != '') && fromJSON(inputs.build_config).baseline_run_id != '' }}
```
ScottTodd: Under what conditions will `prebuilt_stages` be empty but `reusable_artifacts` be non-empty? Seems like this should only need to check `prebuilt_stages` here.

ScottTodd (commented):
A few comments, haven't looked through this fully. Behavior on the test runs look pretty reasonable, though I think excess artifacts are being copied and I'm not sure about the distinction/overlap between `prebuilt_stages` and `reusable_artifacts`

BUILD_TOPOLOGY.toml
geomin12: yeah.... i'll talk to hipdnn team about that

BUILD_TOPOLOGY.toml
geomin12: this is a good point: "Could this just default to the artifact name and only override when needed?" - however, some naming conventions do not match correct. i can take a look and use default behavior if applicable but hard-code if not

"Is the build topology the right place for this information in the first place?" - i wanted to keep it centralized. didn't want to introduce another json tracking file

"Do we need this granular a component<->artifacts mapping at this point? Isn't the math-libs stage with artifact reuse all-or-nothing right now? Is there a scenario where "rand" artifacts are used from prebuilt while "blas" artifacts are built from source?" - it is indeed reuse all or nothing, but i am looking to change that in this PR! without this, it would be a ton of wasted build time. in superrepos, let's say the rocprim team makes a change, we really only need to rebuild rocprim (build took maybe 40s in the PR test) + build dependencies. if we wanted to rebuild math-libs entirely, might be 1h+ and rebuild already built items. prim doesn't need fft! why rebuild fft

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -122,6 +122,58 @@ def _parse_prebuilt_stages(raw: str) -> list[str]:
...
+        normalized = project.split("/")[-1].lower()
+        if normalized in alias_map:
+            rebuild_artifacts.add(alias_map[normalized])
+        else:
+            print(
+                f"  WARNING: unknown project '{project}' - not mapped to any artifact"
+            )
+
+    # Artifacts not in rebuild set are reusable
+    reusable_artifacts = all_stage_artifacts - rebuild_artifacts
```
erman-gurses, line 163: Do we consider which other artifacts were built using this artifact? I think we need to expand rebuild_artifacts through the reverse artifact_deps graph before calculating reusable_artifacts.
For instance:
```
[artifacts.prim]
artifact_group = "math-libs"
type = "target-specific"
artifact_deps = ["core-runtime", "core-hip", "rand"]

[artifacts.sparse]
artifact_group = "math-libs"
type = "target-specific"
artifact_deps = ["blas", "prim"]

[artifacts.solver]
artifact_group = "math-libs"
type = "target-specific"
artifact_deps = ["blas", "prim", "sparse", "host-suite-sparse"]

[artifacts.rocalution]
artifact_group = "math-libs"
type = "target-specific"
artifact_deps = ["core-runtime", "core-hip", "blas", "prim", "rand", "sparse"]
```
Sparse depends on prim, while solver and rocalution depend on prim/sparse. With the current `reusable_artifacts = all_stage_artifacts - rebuild_artifacts`  calculation, those downstream consumers could still be copied from the baseline even though they were built against the previous prim.

erman-gurses (commented):
Please see my initial comment, will do one more pass.

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: when we get `changed_projects` of `shared/rocroller`, we parse `rocroller` and use the BUILD_TOPOLOGY.toml to map to BLAS! then we rebuild only BLAS which includes rocroller

.github/workflows/multi_arch_ci_linux.yml
geomin12: oh good call, there should never be a scenario like this

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: good call. i was going to do this in a follow-up PR, but it's actually a good spot for here. i'll add it but we can split it out if needed

BUILD_TOPOLOGY.toml
ScottTodd: > let's say the rocprim team makes a change, we really only need to rebuild rocprim (build took maybe 40s in the PR test) + build dependencies. if we wanted to rebuild math-libs entirely, might be 1h+ and rebuild already built items. prim doesn't need fft! why rebuild fft

Well the fft libraries are header-only and their builds complete in seconds :P

The other artifacts are useful to pull when building packages though, so this sounds reasonable to me now.

geomin12: hah bad example, but better example is BLAS. I've noted in https://github.com/ROCm/TheRock/issues/3343 that we need a further step to be more correct with sources within in-stage build, but this will work for now

build_tools/_therock_utils/build_topology.py
```diff
@@ -1324,3 +1364,53 @@ def get_stages_for_projects(
...
+                        artifacts.add(artifact.name)
+        return frozenset(artifacts)
+
+    @staticmethod
+    def extract_component_from_path(path: str) -> Optional[str]:
+        """Extract component name from projects/NAME/... or shared/NAME/... path."""
+        import re
+
+        match = re.match(r"^(?:projects|shared|dnn-providers)/([^/]+)(?:/|$)", path)
+        return match.group(1) if match else None
```
ScottTodd, line 1386: nit: I think we can just move `import re` to the top per https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md#import-organization

build_tools/github_actions/tests/configure_multi_arch_ci_test.py
```diff
@@ -1752,7 +1752,13 @@ def test_windows_workflow_uses_all_ci_fields(self):
...
         # build_native_linux is Linux-only. JAX builds are release-only and
         # Linux-only for now, so Windows CI workflows do not consume them.
-        unused_fields = {"build_native_linux", "build_jax", "jax_build_matrix"}
+        # rebuild_artifacts is Linux-only (used by portable build workflow).
+        unused_fields = {
+            "build_native_linux",
+            "build_jax",
+            "jax_build_matrix",
+            "rebuild_artifacts",
+        }
```
ScottTodd: It was added to `.github/workflows/multi_arch_build_portable_linux.yml`, let's also add to https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_build_windows_artifacts.yml ?

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: i'm actually going to do this in a follow up PR, so this PR stays light

build_tools/configure_stage.py
```diff
@@ -154,44 +154,38 @@ def generate_cmake_args(
...
-
-    # Manylinux Python executables for per-Python-version builds
-    # Quote values since they contain semicolons (CMake list separator)
     if manylinux:
         args.append(f'-DTHEROCK_DIST_PYTHON_EXECUTABLES="{DIST_PYTHON_EXECUTABLES}"')
         args.append(
             f'-DTHEROCK_SHARED_PYTHON_EXECUTABLES="{SHARED_PYTHON_EXECUTABLES}"'
         )
 
-    # Disable all features by default, then enable only what we need
```
ScottTodd, line 187: These comments look useful still, why delete in this PR?

build_tools/configure_stage.py
ScottTodd, line 308: Looks like you're changing `--projects` to `--artifacts` but the script still references `--projects` throughout docstrings and arguments like `--list-projects`. Can you choose one mode and then fully switch to it instead of staying half migrated?

build_tools/configure_stage.py
```diff
@@ -354,20 +355,20 @@ def main(argv: List[str] = None):
         parser.error(f"Unknown stage '{args.stage}'. Available stages: {available}")
 
     # Normalize project names (handle paths like "projects/hip" -> "hip")
-    if args.projects:
-        args.projects = [normalize_project_name(p) for p in args.projects]
+    if artifact_list:
+        artifact_list = [normalize_project_name(p) for p in artifact_list]
```
ScottTodd: This is normalizing "artifact" names, not "project" names now, right? Stale comment and function name.

build_tools/configure_stage.py
```diff
@@ -307,18 +301,25 @@ def main(argv: List[str] = None):
...
-        and args.projects is None
+        and not artifact_list
     ):
         parser.error(
             "--stage or --projects is required unless --list-stages, --list-projects, or --skip-stages is specified"
         )
 
-    if args.skip_stages and not args.projects:
+    if args.skip_stages and not artifact_list:
         parser.error("--skip-stages requires --projects")
```
ScottTodd: These `parser.error()` lines reference `--projects` when the option has been renamed to `--artifacts`

build_tools/configure_stage.py
```diff
@@ -354,20 +355,20 @@ def main(argv: List[str] = None):
...
+        artifact_list = [normalize_project_name(p) for p in artifact_list]
 
     # Validate projects if provided (fast-fail on unknown projects)
-    if args.projects:
+    if artifact_list:
         alias_map = topology.get_alias_to_artifact_map(args.build_dir)
-        unknown = [p for p in args.projects if p.lower() not in alias_map]
+        unknown = [p for p in artifact_list if p.lower() not in alias_map]
         if unknown:
             parser.error(f"Unknown project(s): {', '.join(unknown)}")
```
ScottTodd: "Validate projects" -> "Validate _artifacts_"?

"Unknown project(s)" -> "Unknown _artifact(s)_"?

build_tools/configure_stage.py
```diff
@@ -354,20 +355,20 @@ def main(argv: List[str] = None):
...
 
     # Output skip-stages if requested
     if args.skip_stages:
         required_stages = topology.get_stages_for_projects(
-            args.projects, args.build_dir
+            artifact_list, args.build_dir
         )
```
ScottTodd, line 380: Passing _artifact_ list to `get_stages_for_projects`, use a different function? Looks like it's already doing a remapping there:
```python
        # Resolve projects to artifacts
        alias_map = self.get_alias_to_artifact_map(build_dir)
```

https://github.com/ROCm/TheRock/blob/c3621af90a101adb38758ecff0a27b9dddface6d/build_tools/_therock_utils/build_topology.py#L1294-L1301

build_tools/configure_stage.py
ScottTodd: another case here where "artifacts" and "projects" are used interchangeably, please pick one and use it consistently

BUILD_TOPOLOGY.toml
```diff
@@ -70,6 +70,12 @@
 #                                           #   Canonical names: x86_64, aarch64, ppc64le
 #   python_requires = ["-r path/to/req.txt", "pkg"]  # Optional: pip install args
 #   split_databases = ["rocblas", "hipblaslt"]  # Optional: database handlers for kpack split
+#   components = ["comp1", "comp2"]         # Optional: monorepo subdirectory names that map
+#       to this artifact. Used for granular CI artifact reuse: when only specific
+#       components change in an external repo (e.g., rocm-libraries), only affected
+#       artifacts rebuild while others reuse baseline. Component names are directory
+#       basenames within projects/ or shared/ subdirs. Defaults to [artifact_name]
+#       if omitted.
```
ScottTodd: I think we should choose a more specific name here. "components" could mean so many things, but this is specifically a directory name/path within a repository / source set.

Ideally I think we should be able to derive this information from the artifact toml and cmake files though. Consider https://github.com/ROCm/TheRock/blob/main/math-libs/CMakeLists.txt where we have:

```cmake
  therock_cmake_subproject_declare(composable_kernel
    EXTERNAL_SOURCE_DIR "${THEROCK_ROCM_LIBRARIES_SOURCE_DIR}/projects/composablekernel"
    BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/composable_kernel"

  therock_provide_artifact(composable-kernel
    DESCRIPTOR artifact-composable-kernel.toml
    SUBPROJECT_DEPS
      composable_kernel
```

That tells us that the `composable-kernel` artifact (descriptor https://github.com/ROCm/TheRock/blob/main/math-libs/artifact-composable-kernel.toml) uses `${THEROCK_ROCM_LIBRARIES_SOURCE_DIR}/projects/composablekernel` as its source CMake project and any files in that directory or a subdirectory under it are likely relevant to CI.

https://github.com/ROCm/TheRock/blob/main/math-libs/BLAS/CMakeLists.txt is a more complicated case since the `blas` artifact depends on multiple subprojects
```cmake
therock_provide_artifact(blas
  DESCRIPTOR artifact-blas.toml
  COMPONENTS
      dbg
      dev
      doc
      lib
      run
      test
  SUBPROJECT_DEPS ${_blas_subproject_names}
)
```

We may want something closer to https://github.com/ROCm/TheRock/blob/main/test_tools/determine_rocm_test_dependencies.py and https://github.com/ROCm/TheRock/blob/main/test_tools/therock_consumer_graph.json that generates the mapping from source paths to artifacts, derived from the actual CMake files rather than separately maintained in the topology like this.

For now how about we keep the surface area of this PR smaller and just choose a more specific name like `source_paths`?

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -122,6 +122,58 @@ def _parse_prebuilt_stages(raw: str) -> list[str]:
...
+) -> tuple[list[str], list[str]]:
+    """Compute rebuild/reusable artifacts from external repo changed_projects.
+
+    Maps project paths (e.g., "projects/rocprim") to artifact names (e.g., "prim")
+    using BUILD_TOPOLOGY.toml mappings.
+
+    Returns:
+        (rebuild_artifacts, reusable_artifacts) tuple
+    """
+    from _therock_utils.build_topology import get_topology
```
ScottTodd: Move import to the top please

.github/workflows/setup_multi_arch.yml
```diff
@@ -93,6 +93,10 @@ on:
         description: "JSON object with 'repository' (e.g., 'ROCm/rocm-libraries') and 'ref' fields for external repo builds."
         type: string
         default: ""
+      changed_projects:
+        description: "Comma-separated list of changed projects from external repos (e.g., 'projects/rocprim'). Used for granular artifact-level reuse."
+        type: string
+        default: ""
```
ScottTodd, line 120: Could we add this to https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci.yml also? That way we could use workflow_dispatch from this repository to test what the CI does for changes to specific projects without needing to actually create PRs with such changes. I would like for most heuristics-driven computations to have a way to override / specify manually for testing.

ScottTodd (commented):
Architecturally I think this looks good to proceed.

* Let's settle on `artifacts` or `projects` in a few more spots of the code so they aren't used interchangeably
* Can you sync the PR to get a new CI workflow run? Looks like this was affected by a recent github outage and doesn't have recent logs.

build_tools/github_actions/tests/stage_impact_test.py
```diff
@@ -506,5 +509,169 @@ def test_stage_impact_result_shape(self):
...
+        result = analyze_stage_impact(
+            ["rocm-libraries/shared/common.hpp"],
+            topology=topology,
+        )
+
+        self.assertFalse(result.full_rebuild_required)
+        # Artifact-level analysis should NOT be enabled due to conservative fallback
+        self.assertFalse(result.artifact_level_analysis)
+        self.assertEqual(result.impacted_artifacts, ())
+        self.assertEqual(result.reusable_artifacts, ())
```
ScottTodd, line 625: (future optimization) We should be able to improve this by identifying which `shared/` paths affect which projects. For example, https://github.com/ROCm/rocm-libraries/tree/develop/shared/origami is just used by a few subprojects.

build_tools/github_actions/tests/project_artifact_map_test.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""Unit tests for source_path-to-artifact mapping in BuildTopology."""
```
ScottTodd: This seems like it should be folded into https://github.com/ROCm/TheRock/blob/main/build_tools/tests/build_topology_test.py rather than be added as a new test in the `build_tools/github_actions/` subfolder

build_tools/github_actions/configure_external_repo_ci.py
```diff
@@ -64,9 +64,10 @@
 ]
 
 # Patterns that trigger a full test run when changed (CI infrastructure)
+# TODO: Re-enable workflow patterns after testing granular artifact reuse
 FULL_TEST_TRIGGER_PATTERNS = [
-    ".github/workflows/therock*",
-    ".github/scripts/therock*",
+    # ".github/workflows/therock*",
+    # ".github/scripts/therock*",
     ".github/scripts/get_changed_projects.py",
```
ScottTodd, line 70: Do you want to keep these lines commented out in this PR or were these changes just for testing? The comments aren't providing enough context for me to tell with full confidence.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -123,6 +123,56 @@ def _parse_prebuilt_stages(raw: str) -> list[str]:
...
+    rebuild_artifacts: set[str] = set()
+    alias_map = topology.get_alias_to_artifact_map()
+    for project in changed_projects:
+        normalized = project.split("/")[-1].lower()
+        if normalized in alias_map:
+            rebuild_artifacts.add(alias_map[normalized])
+        else:
+            print(
+                f"  WARNING: unknown project '{project}' - not mapped to any artifact"
+            )
```
ScottTodd, line 160: (ok for a follow-up) We could move some of this logic out of `configure_multi_arch_ci.py` and into one of these:
* https://github.com/ROCm/TheRock/blob/main/build_tools/_therock_utils/build_topology.py
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/stage_impact.py
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/stage_reuse_decision.py

I'd like for this configuration code to be as simple as possible, leaning on other modules to assist with decision making, and this is +80 LOC added near the top of the file.

build_tools/configure_stage.py
ScottTodd: Bump - want to bring the comments back?

ScottTodd (approved):
Mostly comments for follow-up work. Could you at least take a look at the comments that were deleted and see if they should be brought back?

build_tools/github_actions/configure_external_repo_ci.py
geomin12: ah this was for testing, i am reverting

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: fair! i can do this in a follow-up, same for https://github.com/ROCm/TheRock/pull/7243#discussion_r3907211935

geomin12:
from example without `changed_projects`, everything works and builds: https://github.com/ROCm/rocm-libraries/actions/runs/33558572933/job/100032836655?pr=10487

i plan to land this and then open PRs to test in rocm-libraries / systems to test functionality. I have additional work on top of this to enhance in parallel

Issue #7248, opened 2026-08-10
ccache: release-build accuracy risk from compiler_check fingerprint coverage gaps


subodh-dubey-amd:
## Problem

Release builds read from the shared ccache remote cache, so the correctness of a cached object depends on the `compiler_check` fingerprint capturing every input that affects output.

The POSIX fingerprint (`build_tools/posix_ccache_compiler_check.py`) hashes the compiler binary and its `ldd` shared libraries, but not:

- the device bitcode libraries (ROCm-Device-Libs `*.bc`), or
- the linker (`lld`).

A toolchain change confined to those inputs (with the clang binary unchanged) would not move the cache key, so a stale object could in principle be reused. This matters most for builds whose artifacts ship (e.g. prereleases promoted to stable).

## Evidence

The fingerprint does invalidate on a real compiler change (recent nightlies on the release cache):

| Stage (nightly) | Aug 6 | Aug 7 | Aug 10 |
| --- | --- | --- | --- |
| compiler-runtime (LLVM) hits | 95.6% | 93.5% | 92.3% |
| math-libs (gfx94X) hits | 31% | 27% | 18% |

LLVM caches well (stable host compiler); math-libs is mostly misses because each night's rebuilt clang changes the fingerprint. The residual concern is only the uncovered inputs above.

## Design discussion + prototype

This issue is the general problem statement and is not tied to a specific fix. Approaches (harden the fingerprint, disable/limit the cache for release builds, etc.) are being weighed here: https://github.com/ROCm/TheRock/discussions/7259

A prototype for one approach (local cache only for prereleases) is at #7249 - draft, not a merge request.

ScottTodd:
> A prerelease build shows ~100% misses in the "Report" step and still produces artifacts; nightly hit rates are unchanged.

There should be zero _remote_ cache hits for these builds, but _local_ cache hits are probably okay. Some of our builds (particularly Windows) require a local cache for performance _and_ stability (without at least a local cache we've seen runners run out of resources and fail).

ScottTodd:
BTW please file issues or post on existing issues _before_ posting PRs if the design is uncertain. See https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#using-github-issues-for-feature-development
> When planning complex changes, please solicit feedback and announce your intent to work on a pull request early in development, as this gives other contributors time to offer advice and avoid duplicating effort.

[...]

PR #7249, opened 2026-08-10
ci: prerelease release builds use local ccache only (no remote reads)

subodh-dubey-amd:
## Motivation

Stable releases are a repackage of the **prerelease** release-candidate build, so the prerelease compile is what actually ships. Those objects should never come from a **remote** cache hit to eliminate cache poisoning risks.

* Background: #7248
* Design discussion: #7259

## Technical Details

Prerelease builds use the **local cache only** (remote disabled): zero remote hits, but the local cache still works normally (reads + writes).

- `build_tools/setup_ccache.py`: now maps `nightly-bkc` and `prerelease` values of `--release-type` to `--config-preset=local`

### Why local-only (not recache or full-disable)

- **Keep the local cache**: local hits are fine, and some builds (notably Windows) need at least a local cache for performance and runner stability - without one we have seen runners exhaust resources and fail.
- `recache` was rejected: it forces *local* misses too, defeating the local cache Windows needs.
- Full-disable was rejected: it removes the local cache and changes build-config parity with nightly/CI.

## Test Plan

- `build_tools/tests/setup_ccache_test.py` (new): pins which release types read the shared remote cache and which are local-only, plus both failure modes above.
- Local: run the tool as each workflow does for `ci` / `dev` / `nightly` / `prerelease` and compare the emitted config.
- Owner dispatch: a `prerelease` run shows zero remote hits (local hits fine) in each stage's "Report" step and still builds artifacts; a `dev` / `nightly` run is unchanged.

## Test Result

CI on this branch, all 9 new tests passing on both platforms:

| Check | Result |
| --- | --- |
| [Unit Tests :: ubuntu-24.04](https://github.com/ROCm/TheRock/actions/runs/32395965820/job/96512731944) | pass - 9/9 `setup_ccache_test.py` |
| [Unit Tests :: windows-2022](https://github.com/ROCm/TheRock/actions/runs/32395965820/job/96512731634) | pass - 9/9 `setup_ccache_test.py` |
| [pre-commit](https://github.com/ROCm/TheRock/actions/runs/32395965889/job/96512732019) | pass (black, mdformat, actionlint) |

Windows coverage matters here, not just for parity: `IS_WINDOWS` selects a different `compiler_check` mode, and Windows is the platform whose local-cache requirement is the reason this is local-only rather than a full disable.

`Cache mode` output from running the tool exactly as each workflow invokes it - only `prerelease` loses the remote, and it keeps its local cache:

```text
[setup_ccache] Cache mode: release_type=ci         preset=github-oss-dev     remote=http://bazelremote-svc.bazelremote-ns...:8080|layout=bazel|connect-timeout=50      local=/tmp/ccache/ci/cache
[setup_ccache] Cache mode: release_type=dev        preset=github-oss-dev     remote=http://bazelremote-svc.bazelremote-ns...:8080|layout=bazel|connect-timeout=50      local=/tmp/ccache/dev/cache
[setup_ccache] Cache mode: release_type=nightly    preset=github-oss-release remote=http://bazelremote-svc-rel.bazelremote-ns...:8080|layout=bazel|connect-timeout=50  local=/tmp/ccache/nightly/cache
[setup_ccache] Cache mode: release_type=prerelease preset=github-oss-release remote=disabled (--no-remote-cache)                                                       local=/tmp/ccache/prerelease/cache
```

## Submission Checklist


build_tools/setup_ccache.py
ScottTodd, line 296: Rather than have both `--release-type` _and_ a new `--no-remote-cache` option, we can route "prerelease" release_type to the existing `local` preset here:

```python
    if args.release_type is not None:
        if args.release_type in ("ci", "dev", "dev-bkc"):
            args.config_preset = "github-oss-dev"
        elif args.release_type in ("prerelease"):
            args.config_preset = "local"
        else:
            args.config_preset = "github-oss-release"
```

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -137,11 +137,15 @@ jobs:
...
+        # Prerelease builds use the local cache only (remote disabled) so a shipped
+        # object never comes from a remote cache hit; the local cache is kept for
+        # performance and runner stability. See docs/development/ccache_troubleshooting.md.
         run: |
           ./build_tools/setup_ccache.py \
             --release-type "${{ inputs.release_type }}" \
             --dir "$(dirname $CCACHE_CONFIGPATH)" \
-            --local-path "$CACHE_DIR/ccache"
+            --local-path "$CACHE_DIR/ccache" \
+            ${{ inputs.release_type == 'prerelease' && '--no-remote-cache' || '' }}
```
ScottTodd: I think we can just have this rely fully on `--release-type` and behavior in the script, no need for `${{ inputs.release_type == 'prerelease' && '--no-remote-cache' || '' }}`

ScottTodd (commented):
Looks like we can substantially simplify this. That would also make testing easier.

ScottTodd:
I pushed my suggested changes to this branch directly, as we'd like to merge this ahead of our next release builds and I didn't want to fully fork the PR to my own. I'll also edit the PR description to reflect the changes.

Sorry for stepping on your toes here, we're short on time here.

marbre (approved):

PR #7260, opened 2026-08-11
ci: harden test container security posture

lumachad:
Addresses ROCM-26826 and ROCM-26887. Three commits, one per change, for bisectability.

**Drop \`seccomp=unconfined\`**

Running with seccomp disabled gives containers access to the full Linux syscall table (bpf, mount, kexec_load, etc.), which none of our GPU tests need. The Docker default profile already permits the syscalls required for normal GPU workloads. Tests that genuinely need ptrace (rocgdb, rocprofiler-sdk) already request \`SYS_PTRACE\` explicitly via \`--cap-add\`, which is the correctly scoped alternative.

**Drop \`--ipc host\`**

Sharing the host IPC namespace exposes the host's \`/dev/shm\` and all POSIX IPC objects to every test container. No job actually needed it: RCCL and rocSHMEM use GPU IPC (peer-to-peer HIP memory handles via KFD), not POSIX shared memory. The PyTorch full-test workflow already uses the correctly scoped \`--shm-size=10g\` for the shared memory it genuinely needs. Also removes unjustified \`SYS_PTRACE\` from both PyTorch wheel workflows — our test drivers don't invoke gdb or ptrace.

**Narrow tester sudo grant**

The base image previously granted \`tester\` blanket \`NOPASSWD:ALL\` sudo, leaving a trivial privilege escalation path via \`sudo bash\`. This commit replaces that with a scoped \`/etc/sudoers.d/tester\` drop-in that allows only the commands CI actually needs at runtime: \`apt\`, \`apt-get\`, \`tee\`, \`mkdir\`, \`chmod\`, and \`dmesg\`. The \`tester\` user is removed from the \`sudo\` group entirely. The git-lfs packagecloud installer script (the only \`sudo bash\` call in the image build) is replaced with a direct \`apt-get install git-lfs\`, and all remaining build-time installs are consolidated into a single root-level \`RUN\` block before the \`USER tester\` switch.

Note: dropping \`--user 0:0\` is not included in this PR. The GHA runner writes \`/_temp/_runner_file_commands/\` state files as root from the host; containers running as a non-root user cannot write back to them, breaking \`actions/checkout\` post-step cleanup. This is the documented reason GitHub recommends \`--user 0:0\` for containerized jobs.

JIRA ID : ROCM-26826
JIRA ID : ROCM-26827

### DIFF
--- a/.github/workflows/test_native_linux_packages_install.yml
+++ b/.github/workflows/test_native_linux_packages_install.yml
@@ -196,7 +196,6 @@ jobs:
     runs-on: ${{ github.repository_owner == 'ROCm' && (inputs.test_runs_on || 'linux-gfx942-1gpu-ccs-csp-ossci-rocm') || 'ubuntu-24.04' }}
     container:
       image: ${{ needs.prepare_install_context.outputs.container_image }}
-      options: --ipc host
     env:
       REPO_URL: ${{ inputs.package_install_url }}
       GPG_KEY_URL: ${{ inputs.gpg_key_url || '' }}
--- a/.github/workflows/test_pytorch_wheels.yml
+++ b/.github/workflows/test_pytorch_wheels.yml
@@ -115,10 +115,7 @@ jobs:
       summary: ${{ steps.summary.outputs.summary }}
     container:
       image: ${{ contains(inputs.test_runs_on, 'linux') && 'ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26' || null }}
-      # SYS_PTRACE is required by PyTorch's test infra, which uses gdb to
-      # collect stack traces on segfaults and hangs (torch/testing/_internal).
-      options: --ipc host
-        --cap-add=SYS_PTRACE
+      options:
         --group-add video
         --device /dev/kfd
         --device /dev/dri
--- a/.github/workflows/test_pytorch_wheels_full.yml
+++ b/.github/workflows/test_pytorch_wheels_full.yml
@@ -195,10 +195,7 @@ jobs:
       matrix: ${{ fromJSON(needs.prepare_matrix.outputs.matrix) }}
     container:
       image: ${{ contains(inputs.test_runs_on, 'linux') && 'ghcr.io/rocm/no_rocm_image_ubuntu24_04@sha256:405945a40deaff9db90b9839c0f41d4cba4a383c1a7459b28627047bf6302a26' || null }}
-      # SYS_PTRACE is required by PyTorch's test infra which uses gdb to
-      # collect stack traces on segfaults and hangs (torch/testing/_internal).
       options: --shm-size=10g
-        --cap-add=SYS_PTRACE
         --group-add video
         --device /dev/kfd
         --device /dev/dri
--- a/.github/workflows/test_rocm_wheels.yml
+++ b/.github/workflows/test_rocm_wheels.yml
@@ -108,8 +108,7 @@ jobs:
     timeout-minutes: 30
     container:
       image: ${{ contains(inputs.test_runs_on, 'linux') && inputs.container_image_url != '' && inputs.container_image_url || null }}
-      options: --ipc host
-        --group-add video
+      options: --group-add video
         --device /dev/kfd
         --device /dev/dri
         --group-add 992
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -62,13 +62,11 @@ def _get_script_path(script_name: str) -> str:
 # --user 0:0 - Running as root, by recommendation of GitHub: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support#user
 # --ulimit memlock=-1:-1 - Prevents memory allocation issues with ROCm inside container
 # --ulimit nofile=1048576:1048576 - Increase open file limit for RCCL
-# --security-opt seccomp=unconfined - enables memory mapping, and is recommended for containers running in HPC environments
 _BASE_CONTAINER_OPTIONS = [
     "--ipc host",
     "--user 0:0",
     "--ulimit memlock=-1:-1",
     "--ulimit nofile=1048576:1048576",
-    "--security-opt seccomp=unconfined",
 ]
 
 # GPU-specific container options (only applied when linux_cpu_runner != True)
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -58,12 +58,10 @@ def _get_script_path(script_name: str) -> str:
 
 
 # Base container options applied to all Linux containers
-# --ipc host - Allows shared memory between host and container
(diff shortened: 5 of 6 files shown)


lumachad:
Needs some validation first.

lumachad:
@marbre You might want to keep an eye out for this one if it ends up passing CI.

marbre:
> @marbre You might want to keep an eye out for this one if it ends up passing CI.

As you know CI is tricky at the moment... One remark:

> The base test image (no_rocm_image_ubuntu24_04) already declares USER tester, a non-root user with passwordless sudo and render/video group membership.

If `tester` has passwordless sudo, it doesn't really strengthen security?

lumachad:
> > @marbre You might want to keep an eye out for this one if it ends up passing CI.
> 
> As you know CI is tricky at the moment... One remark:
> 

Yeah. I'm waiting for some stability so we can validate this one.

> > The base test image (no_rocm_image_ubuntu24_04) already declares USER tester, a non-root user with passwordless sudo and render/video group membership.
> 
> If `tester` has passwordless sudo, it doesn't really strengthen security?

True. I can try to drop that and see what effect it has on tests. I expect none, as we're (debug-tools) not using sudo for our TheRock CI testing scripts.

lumachad:
> > > @marbre You might want to keep an eye out for this one if it ends up passing CI.
> > 
> > 
> > As you know CI is tricky at the moment... One remark:
> 
> Yeah. I'm waiting for some stability so we can validate this one.
> 
> > > The base test image (no_rocm_image_ubuntu24_04) already declares USER tester, a non-root user with passwordless sudo and render/video group membership.
> > 
> > 
> > If `tester` has passwordless sudo, it doesn't really strengthen security?
> 
> True. I can try to drop that and see what effect it has on tests. I expect none, as we're (debug-tools) not using sudo for our TheRock CI testing scripts.

@marbre Experimental additional commit. I couldn't drop all sudo privileges, but it seems we can narrow things down considerably. Let's see how CI runs as I'm not entirely sure how other components will react.

lumachad:
@marbre Trying things without --user 0:0 did not go well, but it seems we did manage to narrow things down a bit.

In order to drop --user 0:0, we'd need a bit more setup that is beyond my reach. It seems information passed through `/_temp/_runner_file_commands/` as root from the host can't be written to from the container as non-root. We might need to sidestep that in some way. Until then, I think we need to stick with --user 0:0.

marbre:
> @marbre Trying things without --user 0:0 did not go well, but it seems we did manage to narrow things down a bit.
> 
> In order to drop --user 0:0, we'd need a bit more setup that is beyond my reach. It seems information passed through `/_temp/_runner_file_commands/` as root from the host can't be written to from the container as non-root. We might need to sidestep that in some way. Until then, I think we need to stick with --user 0:0.

@lumachad so how would you like to proceed? I think we can drop some of the flags or rather only add them for the use cases where really needed. Not sure if we need `--ipc host` for some multi-GPU tests to use shared memory for communication between worker processes or GPUs. Maybe we should file separate issues and try to land it in smaller chunks?

lumachad:
> > @marbre Trying things without --user 0:0 did not go well, but it seems we did manage to narrow things down a bit.
> > In order to drop --user 0:0, we'd need a bit more setup that is beyond my reach. It seems information passed through `/_temp/_runner_file_commands/` as root from the host can't be written to from the container as non-root. We might need to sidestep that in some way. Until then, I think we need to stick with --user 0:0.
> 
> @lumachad so how would you like to proceed? I think we can drop some of the flags or rather only add them for the use cases where really needed. Not sure if we need `--ipc host` for some multi-GPU tests to use shared memory for communication between worker processes or GPUs. Maybe we should file separate issues and try to land it in smaller chunks?

Well, we can proceed in one of the following directions...

1 - Take the CI signal as confirmation these changes are sane.
2 - If we're unsure about the effect on components, component owners need to chime in as I do not have the required context for all of them.
3 - I could narrow the scope down to only rocgdb, which would be a bit of a shame since I've already put the time to investigate and test this change in a few versions.

@marbre Please let me know what you'd prefer.

lumachad:
> > > @marbre Trying things without --user 0:0 did not go well, but it seems we did manage to narrow things down a bit.
> > > In order to drop --user 0:0, we'd need a bit more setup that is beyond my reach. It seems information passed through `/_temp/_runner_file_commands/` as root from the host can't be written to from the container as non-root. We might need to sidestep that in some way. Until then, I think we need to stick with --user 0:0.
> > 
> > 
> > @lumachad so how would you like to proceed? I think we can drop some of the flags or rather only add them for the use cases where really needed. Not sure if we need `--ipc host` for some multi-GPU tests to use shared memory for communication between worker processes or GPUs. Maybe we should file separate issues and try to land it in smaller chunks?
> 
> Well, we can proceed in one of the following directions...
> 
> 1 - Take the CI signal as confirmation these changes are sane. 2 - If we're unsure about the effect on components, component owners need to chime in as I do not have the required context for all of them. 3 - I could narrow the scope down to only rocgdb, which would be a bit of a shame since I've already put the time to investigate and test this change in a few versions.
> 
> @marbre Please let me know what you'd prefer.

Alternatively, since this is touching infra, we could re-assign this to someone working on infra. I'm reaching beyond rocgdb here, after all.

lumachad:
> > > > @marbre Trying things without --user 0:0 did not go well, but it seems we did manage to narrow things down a bit.
> > > > In order to drop --user 0:0, we'd need a bit more setup that is beyond my reach. It seems information passed through `/_temp/_runner_file_commands/` as root from the host can't be written to from the container as non-root. We might need to sidestep that in some way. Until then, I think we need to stick with --user 0:0.
> > > 
> > > 
> > > @lumachad so how would you like to proceed? I think we can drop some of the flags or rather only add them for the use cases where really needed. Not sure if we need `--ipc host` for some multi-GPU tests to use shared memory for communication between worker processes or GPUs. Maybe we should file separate issues and try to land it in smaller chunks?
> > 
> > 
> > Well, we can proceed in one of the following directions...
> > 1 - Take the CI signal as confirmation these changes are sane. 2 - If we're unsure about the effect on components, component owners need to chime in as I do not have the required context for all of them. 3 - I could narrow the scope down to only rocgdb, which would be a bit of a shame since I've already put the time to investigate and test this change in a few versions.
> > @marbre Please let me know what you'd prefer.
> 
> Alternatively, since this is touching infra, we could re-assign this to someone working on infra. I'm reaching beyond rocgdb here, after all.

I went ahead and re-assigned the ticket to @geomin12. This involves too many components, and if I don't have feedback, I can't act on it.

lumachad:
Back to draft. @geomin12 please feel free to build on top of this one.

.github/workflows/test_native_linux_packages_install.yml
```diff
@@ -196,7 +196,6 @@ jobs:
     runs-on: ${{ github.repository_owner == 'ROCm' && (inputs.test_runs_on || 'linux-gfx942-1gpu-ccs-csp-ossci-rocm') || 'ubuntu-24.04' }}
     container:
       image: ${{ needs.prepare_install_context.outputs.container_image }}
-      options: --ipc host
```
arvindcheru, line 199: FYI, We have changes for post install test fixes in #7706, we are updating the  .github/workflows/test_native_linux_packages_install.yml

PR #7269, opened 2026-08-11
feat(packaging): add amdrocm-repo package for ROCm repository setup

lsudarsh-amd:
## Motivation

Installing ROCm today needs a manual repository setup. The user has to fetch and dearmor the AMD signing key, write a sources or repo file by hand, and look up the correct URL for their distribution and release line.

This change adds `amdrocm-repo`, a repo configuration package that replaces that multi-step manual setup. Once it is installed, ROCm can be installed and updated with the native package manager.

This is the first of four pull requests:

- PR 0 (#7004) added `packaging/linux/tests` to `testpaths`, so the tests the later ones add actually run in CI.
- PR 1 (this one) adds the builder, its templates, its unit tests and the user documentation.
- PR 2 (#7276) adds the publish helper and the workflow output type it writes to.

Contributes to #6986.

## Technical Details

`build_repo_package.py` renders the package from the Jinja templates under `template/repo/`. It builds one package per OS profile (`ubuntu2404`, `rhel8`, `rhel10`, `sles16`) and release line (`prerelease`, `release`, `nightly`).

**Repository URLs are derived per release line.** The lines do not share a layout. Prerelease and release publish one repository per distribution. Nightly publishes one per package type, under a dated sub-folder. `repo_baseurl()` selects the correct form. It branches on the release line rather than on whether the line is signed, because those two properties are independent even though they correlate today.

**The package version carries the release line.** Two lines can therefore never produce an identically named package. Without this, installing the release package over the prerelease one at the same ROCm version reports success and silently leaves the old repository configured.

**The signing key is fetched at build time.** For signed lines it is retrieved over https, size-bounded, checked against a pinned fingerprint, and embedded in the package. It is never stored in the source tree. The nightly repository is unsigned upstream, so nightly packages ship no key and disable signature checking.

**The package conflicts with `amdgpu-install`**, which configures a different ROCm repository and raises it above the default priority. Its repository files use different names from ours, so nothing else stops both being installed, and its repository then wins resolution silently. On apt the conflict is settled by removing `amdgpu-install`, which leaves its configuration files behind. The deb `postinst` detects that state and tells the user to run `apt purge`. `dnf` and `zypper` refuse the pair outright and remove their `%config` files on erase, so no equivalent is needed there.

**Nothing builds the repo packages yet.** The script writes finished packages to `--dest-dir` and is run once per profile and release line. The workflow that calls it is in the third pull request; until then it is invoked by hand, as the documentation describes. Each format needs its own tooling, `dpkg-buildpackage` for deb and `rpmbuild` for rpm, so the CI job added later builds each profile in a container that has it.

**`build_tools/.gitignore` is deleted.** Its only entry was an unanchored `repo` pattern, which would hide the entire new `template/repo/` directory from git. The tool it guarded was removed in #128.

### Notes for reviewers

- `repo_baseurl()` derives the same URLs as `get_repo_url_multi_arch()` in `get_url_repo_params.py`, from the same five inputs. Where the two differ it is only a trailing slash, which this one adds on the deb URLs and on nightly rpm. They are not merged here because their base-URL contracts differ: that function takes a scheme and host and appends the layout segments itself, while this one requires them already present. That file is also on the next pull request's path, which is where the two are reconciled.
- `list-profiles` has no caller until the CI pull request. Unit tests cover it in the meantime.
- The test file sits in `build_tools/tests/` rather than `packaging/linux/tests/`, where the other tests for that module's directory live. `build_tools/tests/` is already in `testpaths`, so these tests run in CI from the first push, including on the Windows leg. No module name collides on the path the file inserts.
- The root-ownership directives in `rules.j2` and the spec are convention alignment, not bug fixes. Building as a non-root user already produces `root:root`. They are stated so that behaviour cannot change silently, and to match the other ROCm packages.
- No apt pin or `priority=` is shipped; the package sits at each manager's default. `amdgpu-install` pins its own repository above the default, so an equivalent here would mean bidding above another AMD package rather than resolving anything. Its pin also covers every package from that origin, so matching it would elevate this whole CDN over the distribution's own repositories. The conflict declaration, and the purge the `postinst` points at, address the overlap directly instead. Happy to add a pin if you would prefer one.
- `RELEASE_TYPES` covers three of the five canonical release types. `ci` has no published repository to point at, and `dev` is deferred until there is a reason to ship one.

## Test Plan

All testing ran in containers.

- Built the package for all four OS profiles, signed and unsigned, and inspected each payload with `dpkg-deb -c` or `rpm -qp`.
- Installed the package on `ubuntu2404`, `rhel8`, `rhel10` and `sles16`, refreshed metadata against the live prerelease repository, and resolved the `amdrocm` package.
- Tested the `amdgpu-install` conflict in both directions on both formats, using the real `amdgpu-install` 31.40 packages, then repeated it on each format with the declaration removed as a control.
- Installed the deb against all four `amdgpu-install` states: not installed, installed, removed but not purged, and purged.
- Ran the documented install, verification and build commands.
- Ran the unit tests with networking disabled, and again with the module's path handling bound to the Windows flavour.
- Ran `pre-commit`, `shellcheck`, and the full unit test suite.

## Test Result

| Check | Result |
| --- | --- |
| New unit tests | 106 pass, 1 skip. The skip needs a locale no standard image provides. |
| Full suite | 1747 passing before, 1853 after. The delta is exactly the new test file. |
| Package payload | Correct on all four profiles. Every file lands at its documented path, owned by `root:root`. |
| Signing key | The pinned fingerprint matched on every build that fetches the key. |
| Live repository | Metadata refreshes on all four profiles and `amdrocm` resolves. `apt-get update` verifies the repository signature, and `rpm --checksig` reports `digests signatures OK`. |
| Documented commands | All ran as written. `zypper` needs `--allow-unsigned-rpm` for a locally built file, which the documentation states. |
| Conflict enforcement | Refused in both directions on both formats. |
| Conflict controls | With the declaration removed, both packages install together and two then own competing repository files. |
| `postinst` | Warns only in the removed-but-not-purged state, and never fails an install. With `amdgpu-install` present the conflict refuses the install before the scriptlet runs. |
| Networking disabled | 106 pass, so no unit test reaches the network. |
| Windows path handling | Binding the Windows flavour makes the keyring test fail, so a regression there would be caught. |
| `pre-commit`, `shellcheck` | Clean. |

Three tests fail on this branch. They also fail on `main` without this change, all three are in `build_tools/github_actions/`, which this pull request does not touch, and the failure set is identical either way.

## Submission Checklist


dbieleck:
Folks, let's move this along this week.  Thanks.

docs/packaging/rocm_repo_setup.md
```diff
@@ -0,0 +1,346 @@
...
+your distro first using the previous section. Then install the resulting file,
+refresh the package manager's metadata, and install ROCm packages.
+
+The examples below assume `--dest-dir ./repo-package-out`. The built filename
+carries the package version, so the glob below is what selects it.
+
+Ubuntu:
+
+```bash
+sudo apt install ./repo-package-out/amdrocm-repo_*_all.deb
```
nunnikri, line 224: Once the repo packages are merged might have to update these steps in the RELEASES.md file

build_tools/.gitignore
```diff
@@ -1 +0,0 @@
-repo
```
nunnikri, line 1: Is this a oversight change?

build_tools/packaging/linux/build_repo_package.py
```diff
@@ -0,0 +1,900 @@
...
+(apt/dnf/zypper) to install AMD ROCm from the public ROCm repositories.
+
+One ``amdrocm-repo`` package is generated per target OS profile and RFC0012
+stream. It ships a single repository definition and, for signed streams, the AMD
+signing key, so that after installing the package a user can install ROCm with
+their native package manager. The signing key is fetched at build time and
+embedded in the package; it is never stored in the source tree.
+
+```
+python build_repo_package.py \
```
nunnikri, line 16: With these parameters, looks like the package created will be `amdrocm-repo_10.0.0_all.deb` Which is not as per the packaging name standard. 
amdrocm-<module>_<version>-<build_no>_amd64.deb

Also the package name will be same for ubuntu2404, ubuntu2604, debian10 or any other debian os flavors.

One possibility is to name the packages something like this. 
```
10.0.0-1~noble1      # Ubuntu 24.04 (Noble)
10.0.0-1~jammy1       # Ubuntu 22.04 (Jammy)
10.0.0-1~bookworm1    # Debian 12
```
Another option would be to have one deb package which has all the config for all deb flavors and install only the correct one checking the OS version

build_tools/packaging/linux/build_repo_package.py
```diff
@@ -0,0 +1,900 @@
...
+    --os-profile ubuntu2404 \
+    --stream stable \
+    --repo-base-url https://stable.repo.amd.com/rocm/core/packages \
+    --rocm-version 10.0.0 \
+    --dest-dir ./output
```
nunnikri, line 21: Looks like the parameter handling is bit complicated. Can you give more samples and the expected output(package name and version) for better review?

build_tools/packaging/linux/template/repo/rpm/amdrocm-repo.spec.j2
```diff
@@ -0,0 +1,31 @@
...
+%files
+%defattr(-, root, root, -)
+%config(noreplace) {{ rpm_repo_dir }}/{{ repo_id }}.repo
+{% if signed %}
+{{ rpm_gpg_key_path }}
+{% endif %}
+
+%changelog
+* {{ changelog_date }} {{ maintainer }} - {{ version }}-{{ release }}
+- ROCm {{ rocm_version }} repository configuration.
```
nunnikri, line 31: rpm is not having a post install script like debian. Which is a difference in package behavior

nunnikri (changes requested):

build_tools/packaging/linux/build_repo_package.py
lsudarsh-amd: Thanks for catching this. Deb now builds as:

```
amdrocm-repo_10.0.0-1~ubuntu2404_all.deb
amdrocm-repo_20260716.12345-1~ubuntu2404_all.deb    (nightly)
```

For the distro token I used the OS-profile name rather than a codename, to keep things consistent with the directory names and to avoid maintaining a codename mapping.

I applied the same fix to rpm, where `%{?dist}` is empty on SUSE, so `sles15` and `sles16` would have collided once a second SUSE profile was added. RHEL package names are unchanged.

I avoided taking the second approach you suggested regarding a single deb choosing config at install time, since it needs OS detection in a maintainer script and the repository URL is already per distro.

build_tools/packaging/linux/build_repo_package.py
lsudarsh-amd: The module docstring and [docs/packaging/rocm_repo_setup.md](../docs/packaging/rocm_repo_setup.md) have been updated to both carry the naming rule and a worked examples table. All rows use `--rocm-version 10.0.0` and the default `--version-suffix`. The nightly rows add `--repo-sub-folder 20260716-12345`.

| `--os-profile` | `--stream` | Resulting file                                         |
| -------------- | ---------- | ------------------------------------------------------ |
| `ubuntu2404`   | `stable`   | `amdrocm-repo_10.0.0-1~ubuntu2404_all.deb`             |
| `ubuntu2404`   | `rc`       | `amdrocm-repo_10.0.0~pre-1~ubuntu2404_all.deb`         |
| `ubuntu2404`   | `nightly`  | `amdrocm-repo_20260716.12345-1~ubuntu2404_all.deb`     |
| `rhel8`        | `stable`   | `amdrocm-repo-10.0.0-1.stable.el8.noarch.rpm`          |
| `rhel8`        | `rc`       | `amdrocm-repo-10.0.0-1.rc.el8.noarch.rpm`              |
| `rhel8`        | `nightly`  | `amdrocm-repo-20260716-12345.1.nightly.el8.noarch.rpm` |
| `rhel10`       | `stable`   | `amdrocm-repo-10.0.0-1.stable.el10.noarch.rpm`         |
| `sles16`       | `stable`   | `amdrocm-repo-10.0.0-1.stable.sles16.noarch.rpm`       |

build_tools/packaging/linux/template/repo/rpm/amdrocm-repo.spec.j2
lsudarsh-amd: rpm now has a post install script. It prints the same message the deb `postinst` does, with the refresh command for each family: `sudo dnf makecache` or `sudo zypper refresh`. I confirmed the message reaches the user on both `dnf install` and `zypper install`.

build_tools/.gitignore
lsudarsh-amd: This is deliberate. Scott asked for the file to be removed rather than edited in [#6901](https://github.com/ROCm/TheRock/pull/6901#discussion_r3677846801), since the `repo` tool it guarded has been unused since #128. Its bare `repo` pattern would also have matched the new `template/repo/` directory and hidden those files.

docs/packaging/rocm_repo_setup.md
lsudarsh-amd: Yes, I agree. `amdrocm-repo` has no public download URL yet, so those steps would not be actionable until the publish side lands. I have left `RELEASES.md` alone here and will update it as a follow up once the package is reachable.

PR #7273, opened 2026-08-11
Run full nightly PyTorch tests daily

ethanwee1:
JIRA ID: AIPYTORCH-1000

## Problem

Full PyTorch tests for the `nightly` ref run only on Sunday. Release branches already run daily. Waiting a week for nightly regressions is too slow.

## Changes

Remove the Sunday-only cadence gate in `dispatch_pytorch_wheels_full_test` so a successful gfx94X Python 3.12 wheel build with `run_full_pytorch_tests` always dispatches `test_pytorch_wheels_full.yml`, including `nightly`.

The `workflow-dispatch` action pin stays on v1.3.3 from main.

## Validation

Rebased onto current `main` (AIPYTORCH-1210). No new test run on this rebase. Prior capacity notes:

- rockrel already triggers nightly releases daily.
- Weekly load goes from 22 to 28 full suites (~27%), about 48 extra single-GPU jobs and 18 extra 8-GPU jobs.
- Still needs capacity sign-off from amd-shiraz; marbre noted gfx942 queue times.


ScottTodd (commented):
Can you add more context to the PR description? What are the current job schedules and how does this change them? How much will this increase workflow load?

@amd-shiraz should review this for capacity planning (we are currently short on supply for many runner types)

ethanwee1:
Done @ScottTodd

ScottTodd (commented):
I'll leave this review to @amd-shiraz . The workflow changes look fine, if we have capacity to support this.

marbre (commented):
We are currently seeing long queue times on gfx942 (single- and multi-gpu) thus we may want to hold back here.

PR #7017, opened 2026-07-31
test(packaging): fix native install UT drift and canonicalize get_url_repo_params URLs/GPG

arvindcheru:
## Motivation

This PR restores green unit coverage for the install driver and establishes get_url_repo_params.py plus its tests as the canonical, merge-time-checked spec for install repo URLs, GPG paths, and multi-arch layout—so URL/GPG regressions are caught in CI without running the full install matrix.

## Technical Details

- Align native_linux_package_install_ut_test.py with current install-test behavior 
- Correct per-family and multi-arch install repo URL and GPG path derivation in get_url_repo_params.py to match latest pkging specs. 
- Replace the expanded get_url_repo_params_test.py suite with a trimmed table-driven set covering URL/GPG paths, layout normalization, release-type policy, and CI-wired subcommands.
- replaces https://github.com/ROCm/TheRock/pull/4932

Part of 
#7028 
https://github.com/ROCm/TheRock/issues/7028

## Test Plan

[] python3 -m unittest build_tools.packaging.linux.tests.native_linux_package_install_ut_test -v — expect Ran 100 tests 
[] python3 -m unittest build_tools.packaging.linux.tests.get_url_repo_params_test -v — expect Ran 21 tests 
[] CI: test_native_linux_packages_install.yml — confirm wired subcommands still pass

## Test Result
CI: test_native_linux_packages_install.yml results: Success
https://github.com/ROCm/TheRock/actions/runs/30609387941
https://github.com/ROCm/TheRock/actions/runs/30660268891
https://github.com/ROCm/TheRock/actions/runs/30660389291
https://github.com/ROCm/TheRock/actions/runs/30660714020

## Submission Checklist

- [ ] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

PR #7276, opened 2026-08-11
feat(packaging): publish amdrocm-repo via workflow outputs

lsudarsh-amd:
## Motivation

The `amdrocm-repo` bootstrap package needs a published location before anyone can fetch and install it. This pull request adds the script that publishes it, along with the URL helpers the packaging workflow calls. The workflow job that runs them comes in the next pull request.

It also resolves two review comments on #6901, which this split supersedes. The publish helper imported `boto3` inline and did not use `WorkflowOutputRoot`. It now resolves its destination through `WorkflowOutputRoot` and writes through `StorageBackend`, which takes `boto3` out of this script entirely rather than hoisting the import to the top of the file. The dependency still exists behind `S3StorageBackend`, but the helper no longer touches it.

This is the third of four pull requests:

- PR 0 (#7004) added `packaging/linux/tests` to `testpaths`, so the tests the later ones add actually run in CI.
- PR 1 (#7269) adds the builder, its templates, its unit tests and the user documentation.
- PR 2 (this one) adds the publish helper and the workflow output type it writes to.

Contributes to #6986.

### What this adds

**A new output type.** `WorkflowOutputRoot.native_linux_repo_package(pkg_type, os_profile)` returns `{prefix}/packages/{pkg_type}/repo/{os_profile}/amdrocm-repo.{pkg_type}`. It follows the process in `docs/development/workflow_outputs.md`, which is why this adds the method, its tests and the documentation together.

The key sits beside the content repository rather than inside it. Clients fetch the file directly by URL, so it does not belong in the package index. The published name is fixed at `amdrocm-repo.<ext>` to keep that URL stable, so the per-profile directory is what stops one profile's package overwriting another's. It also keeps a client from resolving the package built for a different distro, since every profile uses the same package name.

**The publish helper derives its own destination.** `publish_repo_package.py` previously took `--bucket`, `--prefix` and `--endpoint-url`. It now takes `--run-id` and resolves the bucket and prefix through `WorkflowOutputRoot`, which is the single source of truth for CI path layout. Two consequences are worth calling out.

1. The `release` line now raises from inside this script rather than from a separate `write_artifacts_bucket_info.py` step. There is no artifacts bucket for that line, since every `therock-release-*` bucket has `iam_role=None` and release upload happens externally. The publishing job is gated off `release`, so the raise is a backstop.
2. The script needs `GITHUB_REPOSITORY`, `RELEASE_TYPE` and the workflow event payload in its job environment. The `--bucket` version needed none of them.

**Nothing calls this script yet.** No workflow on `main` invokes `publish_repo_package.py`, so changing its arguments breaks no existing caller. The workflow that will call it lands in the next pull request and uses the new ones.

**URL helpers.** `get_public_repo_base_url()`, `nightly_sub_folder()` and `get_container_image_map()` are added with their CLI subcommands. The workflow reads all three.

**Install harness.** `native_linux_package_install_test.py` gains a mode that configures the repository by installing the built package instead of writing repository files directly.

### Fixes found while adding the install harness

**Keyring path.** `APT_KEYRING_FILE` defaulted to `/etc/apt/keyrings/rocm.gpg`, where the current documented amdgpu package manager steps set up the signing key. No package owns that path, but the harness writes there with `sudo tee` and would overwrite a key the host is already using, so the default is now derived from `REPO_NAME`. `setup_gpg_key()` built its write target separately, so it could drift from the `signed-by=` reference. Both now read `APT_KEYRING_FILE`, and a test asserts they agree. This path is not reachable in CI today, since no caller supplies `gpg_key_url`.

**Shell removal.** The GPG import was a single `shell=True` pipeline interpolating the key URL and the keyring path, both of which are configurable. It is now three list-form calls to `wget`, `gpg --dearmor` and `sudo tee`.

**Input validation.** `run_id` must now be numeric. It becomes the leading path segment of the object key, and it is written into `$GITHUB_OUTPUT`, where `gha_set_output` uses a heredoc whose delimiter it does not verify. Both inputs come from CI, so this is defence in depth.

**Retired tests.** `get_url_repo_params_test.py` now takes main's `MainSubcommandsTest` as rewritten in #7017 and drops this branch's older version. Three cases do not carry across, all asserting that the CLI turns an error into exit code 1 rather than raising: `test_get_base_url_invalid_returns_one`, `test_get_repo_url_error_returns_one` and `test_extract_gfx_arch_empty_returns_one`. The underlying errors are still covered at function level.

**One fixture.** `RunTestsTestTypeTest._base_args` gains `repo_package_dir` and `repo_config_only` with their argparse defaults, both of which `main()` reads. Without them those tests fail, since #7004 made that directory collectable.

## Test Plan

- **Unit tests.** The new output type across both package types and all four OS profiles, the publish helper end to end, and the input validation. The publish tests drive a real `LocalStorageBackend` and assert the bytes on disk instead of mocking an S3 client, so only bucket resolution is stubbed.
- **Full suite.** `pytest build_tools/` in the CI container, against a baseline measured at the same base commit, comparing failure names rather than counts.
- **GPG import for real.** A key generated, served over localhost, and imported through the genuine `setup_gpg_key()` path, to confirm which file the harness actually writes.
- **Negative controls.** For every new assertion, revert the change it covers, confirm the test fails, then restore.
- **Interface check.** The exact command the workflow will run, replayed against the new CLI.
- **Lint.** `pre-commit` over the changed files.

## Test Result

| Check | Result |
|---|---|
| Unit tests | Pass. 268 tests and 68 subtests across the four changed test modules. |
| Full suite | **No new failures.** 9 failed and 1828 passed, against 9 failed and 1735 passed on `main` at `a3e3f450db`. The failure names are identical on both sides. |
| GPG import | Pass. The key landed at the path `signed-by=` pins, dearmored, fingerprint intact, mode 0644, and `/etc/apt/keyrings/rocm.gpg` was not written. |
| Negative controls | Pass. Every one fired, so no new assertion passes against the code it is meant to catch. |
| Interface check | Pass. The command placed the file where the workflow expects, and the old flags are rejected. |
| Lint | Pass, including `black`, `mdformat` and `actionlint`. |

Two notes on the numbers above. The nine pre-existing failures are in `compute_rocm_package_version_test.py`, `jax_install_scripts_test.py` and `stage_reuse_decision_test.py`, none of which this pull request touches, and that count moves from day to day, so it is worth re-measuring rather than trusting it. Running the GPG check against the code before the fix failed on both points, which is what confirms the fix does real work rather than passing a test written around it.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/native_linux_package_install_test.py
+++ b/build_tools/packaging/linux/native_linux_package_install_test.py
@@ -137,6 +137,7 @@ import re
 import stat
 import subprocess
 import sys
+import time
 import traceback
 from argparse import ArgumentParser, Namespace
 from pathlib import Path, PurePosixPath
@@ -166,7 +167,9 @@ APT_SOURCES_LIST = _env(
 # documented amdgpu package manager steps set the signing key up there
 # ("wget .../rocm.gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg")
 # and point signed-by= at it. This harness writes its keyring the same way, so
-# sharing the path would overwrite the key a host is already using.
+# sharing the path would overwrite the key a host is already using. Distinct
+# from the amdrocm-repo package's own keyring (AMDROCM_DEB_KEYRING), which is
+# package-owned and lives in a different directory.
 APT_KEYRING_FILE = _env("ROCM_APT_KEYRING_FILE", f"/etc/apt/keyrings/{REPO_NAME}.gpg")
 ZYPP_REPOS_DIR = _env("ROCM_ZYPP_REPOS_DIR", "/etc/zypp/repos.d")
 YUM_REPOS_DIR = _env("ROCM_YUM_REPOS_DIR", "/etc/yum.repos.d")
@@ -180,6 +183,21 @@ VERIFY_KEY_COMPONENTS = [
 # Relative path from install prefix to rdhc binary (script); overridable via ROCM_RDHC_REL_PATH
 RDHC_REL_PATH = _env("ROCM_RDHC_REL_PATH", "libexec/rocm-core/rdhc.py")
 
+# amdrocm-repo package: the on-disk paths it installs (the package's public
+# contract). Used by the install-via-package mode to confirm the repo config
+# landed. These are independent of REPO_NAME (which names the manually-written
+# repo file used by the default setup path).
+AMDROCM_REPO_ID = "amdrocm"
+AMDROCM_DEB_SOURCES = "/etc/apt/sources.list.d/amdrocm.sources"
+# Named amdrocm.gpg, not rocm.gpg, so it cannot collide with the keyring the
+# amdgpu driver setup from repo.radeon.com installs. Must match
+# build_repo_package.DEB_KEYRING_PATH.
+AMDROCM_DEB_KEYRING = "/usr/share/keyrings/amdrocm.gpg"
+AMDROCM_RPM_REPO_BASENAME = "amdrocm.repo"
+# Named -amdrocm, not -rocm, for the same collision reason as the deb
+# keyring. Must match build_repo_package.RPM_GPG_KEY_PATH.
+AMDROCM_RPM_KEY = "/etc/pki/rpm-gpg/RPM-GPG-KEY-amdrocm"
+
 # Pytest/CI only: becomes ``--rocm-version``.
 ENV_NATIVE_LINUX_INSTALL_ROCM_VERSION = "NATIVE_LINUX_INSTALL_ROCM_VERSION"
 
@@ -190,6 +208,12 @@ APT_UPDATE_TIMEOUT_SEC = 120
 ZYPP_CLEAN_TIMEOUT_SEC = 60
 ZYPP_REFRESH_TIMEOUT_SEC = 120
 DNF_CLEAN_TIMEOUT_SEC = 60
+DNF_MAKECACHE_TIMEOUT_SEC = 120
+REPO_PACKAGE_INSTALL_TIMEOUT_SEC = 120
+# Retry the post-install metadata refresh so a transient repo/CDN blip does not
+# fail the (network-dependent) config-only gate.
+REPO_REFRESH_ATTEMPTS = 3
+REPO_REFRESH_BACKOFF_SEC = 3
 INSTALL_TIMEOUT_SEC = 1800  # 30 minutes
 ROCMINFO_TIMEOUT_SEC = 30
 # rdhc.py ``--all`` runs the full ROCm deployment health check suite; 30s was too
@@ -372,11 +396,14 @@ class NativeLinuxPackageInstallTest:
         rocm_version: str | None = None,
         gpg_key_url: str | None = None,
         build_variant: str = "",
+        repo_package_dir: str | None = None,
+        repo_config_only: bool = False,
     ):
         """Initialize the native Linux package install test runner.
 
         Args:
-        repo_url: Full repository URL (constructed in YAML)
+        repo_url: Full repository URL (constructed in YAML). May be empty/None when
+        repo_package_dir is set (the amdrocm-repo package supplies the repo URL).
         os_profile: OS profile (e.g., ubuntu2404, rhel8, debian12, sles15, sles16, almalinux9, centos7, azl3)
         release_type: Type of release ('nightly' or 'prerelease')
         install_prefix: Installation prefix (default: /opt/rocm/core)
@@ -393,11 +420,20 @@ class NativeLinuxPackageInstallTest:
         inserted before the version suffix in package names so that variant
         packages are tested (e.g. ``amdrocm-asan7.15-gfx942`` instead of
         ``amdrocm7.15-gfx942``).
+        repo_package_dir: Directory holding the built ``amdrocm-repo`` package. When
+        set, the repository is configured by installing that package instead of
+        writing repo files directly.
+        repo_config_only: With ``repo_package_dir``, install the package and verify
+        the repo config, then stop (do not install ROCm).
         """
         self.os_profile = os_profile.lower()
         self.package_type = self._derive_package_type(os_profile)
-        self.repo_url = repo_url.rstrip("/")
+        # repo_url may be empty/None in install-via-package mode (the package
+        # supplies it); tolerate both without raising.
+        self.repo_url = (repo_url or "").rstrip("/")
         self.release_type = release_type.lower()
+        self.repo_package_dir = repo_package_dir
+        self.repo_config_only = repo_config_only
         self.install_prefix = install_prefix
         self.gfx_arch_list = normalize_target_list(
             gfx_arch, lowercase=True, dedupe=True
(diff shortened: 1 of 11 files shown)


ScottTodd:
No concerns for the `build_tools/_therock_utils/workflow_outputs.py` and related changes in this PR from me. Will let others more directly working on native linux packages review.

nunnikri:
Its a big change. need some time to review it. 
on a quick look i think might need a change in build_tools/github_actions/publish_rocm_to_release_buckets.py to actually publish the package to release area

arvindcheru:
Hi @lsudarsh-amd, reviewers,  
I Propose pulling the setup_gpg_key() rewrite (shell=True → list-form subprocess calls) and the APT_KEYRING_FILE path fix (hardcoded /etc/apt/keyrings/rocm.gpg → /etc/apt/keyrings/{REPO_NAME}.gpg) out of #7276 into their own PR. It's a pure sec/collision fix to existing stable code with zero functional coupling to the new amdrocm-repo feature — nothing in #7269 or #7280 references those two pieces.

**Learnings:**

To me, This PR bundles two independent changes:
1. GPG/keyring fix (existing code) — setup_gpg_key() rewritten from a shell=True pipeline to list-form subprocess.run calls, plus APT_KEYRING_FILE's default path changed from /etc/apt/keyrings/rocm.gpg to /etc/apt/keyrings/{REPO_NAME}.gpg. This fixe looks to be helpful for analysis/fix different issue and a latent collision with the AMDGPU driver installer's keyring — both affect the harness's default, already-running code path.
2. New amdrocm-repo bootstrap package feature — new storage layout, publish_repo_package.py, install-via-package mode, and supporting subcommands, all gated behind new opt-in flags (--repo-package-dir, --repo-config-only) that no existing caller sets.
Why split:
	• No dependency between them — traced the call graph and confirmed the new feature's install_repo_package()/assert_repo_configured() never call setup_gpg_key() or reference APT_KEYRING_FILE; they use a separate, independently-named keyring constant (AMDROCM_DEB_KEYRING). Splitting introduces no functional risk.
	• Different risk/urgency profiles - The new feature is larger, and warrants its own focused review cycle.
	• Cleaner review and rollback — bundling them makes it harder to isolate a regression to one or the other if something breaks post-merge, and forces reviewers to context-switch between a security fix and a feature design.
Proposed split: PR A = GPG/keyring rewrite + its dedicated tests (ConfiguredPathsTest, GPG-import test cases). PR B = everything else (new storage layout, publish_repo_package.py, package-mode install/verify, new get_url_repo_params.py subcommands, associated tests/docs).

Thanks

arvindcheru (changes requested):
Request to Split the GPG Key Ring Changes (which modifies existing running workflow) from the new feature proposed. Refer the Comments in the PR for more details. Thanks.

lsudarsh-amd:
> Its a big change. need some time to review it. on a quick look i think might need a change in build_tools/github_actions/publish_rocm_to_release_buckets.py to actually publish the package to release area

Thanks @nunnikri for starting to look at these changes. After your initial comment, I double checked the implementation, and no change should be needed in `publish_rocm_to_release_buckets.py`. I checked by running the copy, and the bootstrap package lands at `v4/packages/deb/repo/ubuntu2404/amdrocm-repo.deb` alongside `pool/` and `dists/`.

It's carried because `{run_id}-linux/packages/{pkg_type}/repo/{os_profile}/` sits inside the prefix that `publish_native_linux_packages` already copies at line 246. That copy is recursive on both backends, since S3 paginates with no `Delimiter` and the local backend uses `rglob`, and it preserves relative paths (`storage_backend.py:126`, `:346`, `:494`).

It stays out of the package index too, which is why it sits in `repo/`. The deb metadata is built from the locally uploaded files rather than by enumerating the prefix (`upload_package_repo.py:355`), and the rpm side lists only `{prefix}/x86_64/repodata/`. That matters because one `amdrocm-repo` is built per OS profile and they all share a package name.

I've added three tests to ensure proper coverage here. One drives the real promotion into a local backend, and two cover the S3 listing so a `Delimiter` or a dropped nested key fails instead of silently losing the package from every release.

lsudarsh-amd:
> Request to Split the GPG Key Ring Changes (which modifies existing running workflow) from the new feature proposed. Refer the Comments in the PR for more details. Thanks.

Thanks for the review @arvindcheru. As requested, the keyring work is now #7332, which carries the `setup_gpg_key()` rewrite, the `APT_KEYRING_FILE` path change and their tests. This PR keeps the storage layout, the publish helper and the install-via-package mode. The keyring commits are out of it as of b4b4152457, so its diff against `main` no longer touches any of that code.

Your call-graph trace was right. `install_repo_package()` and `assert_repo_configured()` never call `setup_gpg_key()` or read `APT_KEYRING_FILE`, and they use a separate constant (`AMDROCM_DEB_KEYRING`, `native_linux_package_install_test.py:188`). One case can't move with the rest. `test_package_keyring_is_separate_from_the_harness_keyring` asserts on `AMDROCM_DEB_KEYRING`, which doesn't exist on `main`, so it stays here, in a class renamed to `PackageKeyringPathTest` so it can't collide with the `ConfiguredPathsTest` that #7332 adds to the same module.

Isolating that code also turned up a bug, fixed and covered in #7332. The documented `ROCM_APT_KEYRING_FILE` override never worked, because only the `signed-by=` reference used it while the write target hardcoded `rocm.gpg`, so setting it makes `apt update` fail with "The repository is not signed". The fix also leaves `APT_KEYRING_DIR` with no reader, so #7332 removes it along with its docstring entry.

build_tools/packaging/linux/native_linux_package_install_test.py
```diff
@@ -175,6 +176,21 @@ def _env(key: str, default: str) -> str:
...
+# repo file used by the default setup path).
+AMDROCM_REPO_ID = "amdrocm"
+AMDROCM_DEB_SOURCES = "/etc/apt/sources.list.d/amdrocm.sources"
+# Named amdrocm.gpg, not rocm.gpg, so it cannot collide with the keyring the
+# amdgpu driver setup from repo.radeon.com installs. Must match
+# build_repo_package.DEB_KEYRING_PATH.
+AMDROCM_DEB_KEYRING = "/usr/share/keyrings/amdrocm.gpg"
+AMDROCM_RPM_REPO_BASENAME = "amdrocm.repo"
+# Named -amdrocm, not -rocm, for the same collision reason as the deb
+# keyring. Must match build_repo_package.RPM_GPG_KEY_PATH.
```
arvindcheru, line 198: Thank you for the updating the comments, Will review in detail.

Observation: native_linux_package_install_test.py hardcodes the GPG key paths as strings instead of importing them from build_repo_package.py, relying on a comment to keep them in sync. If the paths change there, this file won't notice.
Is it possible to import: Import DEB_KEYRING_PATH/RPM_GPG_KEY_PATH from build_repo_package.py instead — inspect_repo_package.py already does this, so it's a consistent, low-risk fix.

lsudarsh-amd: I agree with you on the risk you outlined. Two spellings of one path kept in sync by a comment is the same bug #7332 fixes.

I can't make the fix in this PR though. `build_repo_package.py` is added by #7269, so the import would fail at collection on this branch. `inspect_repo_package.py` gets away with it because it lands in #7280, after #7269.

I'd rather make this change in #7280, which anyway depends on #7269 and already imports both constants.

dbieleck:
Please review.

arvindcheru (approved):
LGTM,

Please resolve the Merge conflicts, and wait for CI tests
Approve #7276: Based on GPG split is done (#7332), promotion is covered by tests, scope is publish helper + workflow outputs + URL helpers + opt-in install mode — LGTM to merge after #7269 (and before #7280.)

Since there are multiple PR dependency Kindly wait for other review input also before proceeding to merge.

PR #7278, opened 2026-08-11
feat(sysdeps): add optional bundled libfwupd for amd-smi UMA carveout

adam360x:
## Motivation

`amd-smi` reads and sets the UMA "carveout" (dedicated GPU memory) BIOS setting on APUs. On UEFI-HII platforms — notably HP Strix Halo systems (ZBook Ultra G1a, Z2 Mini G1a) — the amdgpu `/sys/class/drm/<card>/device/uma/carveout` sysfs node is absent, and the setting is only reachable through the **fwupd** daemon's BIOS-settings interface. To let amd-smi talk to fwupd via `libfwupd` (over D-Bus) rather than shelling out to `fwupdmgr`, we need `libfwupd` and its dependency stack bundled as a portable sysdep.

## What this adds

An **optional** Linux sysdep artifact (`sysdeps-fwupd`, gated on `THEROCK_ENABLE_SYSDEPS_FWUPD`, off by default) that bundles:

| library | version | notes |
| --- | --- | --- |
| libffi | 3.4.6 | GLib (gobject) dependency |
| pcre2 | 10.42 | GLib core dependency |
| glib | 2.74.7 | glib/gobject/gio/gmodule/gthread |
| libcurl | 8.5.0 | built without TLS — the BIOS path is local D-Bus only |
| libfwupd | 2.1.7 | `-Dbuild=library` |

`zlib` (used by GLib's gio) is already a core sysdep, so it is reused.

### Why fwupd 2.1.7
2.1.x is the last release series **before** the Rust/FFI work that lands in 2.2.x / `main`, and it no longer depends on `libjcat` or `json-glib` (both are internal now). This keeps the stack small and Rust-free.

## Details
- Each library installs under `lib/rocm_sysdeps`, is SONAME-prefixed `librocm_sysdeps_*`, and is symbol-versioned `AMDROCM_SYSDEPS_1.0`, following the existing `libmnl` (autotools) and `libdrm` (meson) sysdep patterns.
- GLib's five interdependent libraries are renamed post-build via `patchelf --set-soname` / `--replace-needed`.
- Build-host requirements: `meson` (already required for bundled sysdeps) plus the `jinja2` Python module (used by fwupd's code generation).

## Testing
- Each sysdep builds standalone and produces the expected renamed, symbol-versioned `.so`.
- The assembled bundle links with **zero unresolved dependencies** (only `libz` → the zlib core sysdep, plus libc).
- The resulting `libfwupd.so` was validated on a real **HP ZBook Ultra G1a** (Ubuntu 26.04): it connects to the system fwupd daemon and reads the UMA carveout BIOS setting through `libfwupd`.

## Follow-ups (not in this PR)
- The five source tarballs need mirroring to the `rocm-third-party-deps` S3 bucket (upstream URLs + SHA256s are recorded in each `CMakeLists.txt`).
- The amd-smi consumer change (linking `THEROCK_BUNDLED_FWUPD`) is separate.

## Issue Tracking

JIRA ID : AILITOOLS-306

### DIFF
--- a/third-party/sysdeps/linux/libfwupd/CMakeLists.txt
+++ b/third-party/sysdeps/linux/libfwupd/CMakeLists.txt
@@ -1,0 +1,127 @@
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    # When included in TheRock, we download sources and set up the sub-project.
+    set(_source_dir "${CMAKE_CURRENT_BINARY_DIR}/source")
+    set(_download_stamp "${_source_dir}/download.stamp")
+
+    therock_subproject_fetch(therock-libfwupd-sources
+      SOURCE_DIR "${_source_dir}"
+      # Originally from: https://github.com/fwupd/fwupd/archive/refs/tags/2.1.7.tar.gz
+      # 2.1.x is the last pre-Rust release series (Rust/FFI lands in 2.2.x/main)
+      # and no longer depends on libjcat or json-glib.
+      URL "https://rocm-third-party-deps.s3.us-east-2.amazonaws.com/fwupd-2.1.7.tar.gz"
+      URL_HASH "SHA256=338ea5b139f9a37447ff2215d0a8d2cf23851fb5a39fba7b1b0ded52a40240cc"
+      TOUCH "${_download_stamp}"
+      DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+    )
+
+    therock_cmake_subproject_declare(therock-libfwupd
+      USE_DIST_AMDGPU_TARGETS
+      EXTERNAL_SOURCE_DIR .
+      FPRINT_SOURCE_DIR "${_source_dir}"
+      FPRINT_FILE_GLOBS "${CMAKE_CURRENT_LIST_DIR}/*"
+      BINARY_DIR build
+      NO_MERGE_COMPILE_COMMANDS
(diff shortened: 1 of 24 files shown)


third-party/sysdeps/linux/libfwupd/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+        --reconfigure
+        --prefix "/"
+        -Dpkgconfig.relocatable=true
+        -Dlibdir=lib
+        # library-only build: no daemon, plugins, tools, or Rust.
+        -Dbuild=library
+        -Ddocs=disabled
+        -Dman=false
+        -Dtests=false
+        -Dintrospection=disabled
```
superm1, line 104: You might want introspection if you want to use libfwupd in python.

adam360x: Thanks! For amd-smi we only consume libfwupd from C++ — `fwupd_carveout.cc` uses the C API via pkg-config (`pkg_check_modules(... IMPORTED_TARGET fwupd)`). The Python CLI doesn't link libfwupd; it shells out to `fwupdmgr --json`. So we don't need GObject introspection for this use case, and we're intentionally keeping the bundle minimal (same reasoning behind pinning fwupd 2.1.7 and dropping libjcat/json-glib to stay Rust-free).

If we later add a PyGObject-native path (so the Python CLI could use the bundled `libfwupd.so` instead of requiring `fwupdmgr` at runtime), we can flip `-Dintrospection=true` then.

superm1 (approved):

oliveiradan (approved):

PR #7303, opened 2026-08-12
build_tools: per-submodule CI labels and test:rocgdb group alias

lumachad:
## Summary

- Replaces the single `CI_LABELS` global in `bump_automation.py` with a `COMMON_CI_LABELS` base that each submodule expands. rocgdb bump PRs now get `ci:run-all-archs` and `test:rocgdb` rather than `ci:asan`, which is only relevant for rocm-systems and rocm-libraries.
- Adds `TEST_LABEL_GROUPS` to `fetch_test_configurations.py` — a dict that maps a group alias to the individual test matrix keys it expands to. The `rocgdb` group selects `rocgdb-cpu`, `rocgdb-gpu`, and `rocgdb-corefile`. Unknown labels continue to pass through as literal keys, so all existing `test:*` usage is unaffected.
- Adds intent comments to each `SUBMODULE_CONFIG` entry explaining what CI matrix the submodule change triggers.
- Switches new `TEST_LABEL_GROUPS` tests to use `patch.dict(os.environ)` rather than mutating the global environment directly.

JIRA ID: N/A

### DIFF
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -38,6 +38,10 @@ SUBMODULE_CONFIG = {
         "files": ROCM_SYSTEMS_FILES,
         "updater": "ref",
         "token_key": "systems",
+        # Changes to rocm-systems should run the full matrix of CI jobs:
+        #   * Build for all gfx archs
+        #   * Build for all variants (asan)
+        #   * All builds and tests (including downstream rocm-libraries jobs)
         "labels": [*COMMON_CI_LABELS, "ci:asan"],
     },
     "rocm-libraries": {
@@ -45,6 +49,10 @@ SUBMODULE_CONFIG = {
         "files": ROCM_LIBRARIES_FILES,
         "updater": "ci-env",
         "token_key": "libraries",
+        # Changes to rocm-libraries should run the full matrix of CI jobs:
+        #   * Build for all gfx archs
+        #   * Build for all variants (asan)
+        #   * All rocm-libraries tests
         "labels": [*COMMON_CI_LABELS, "ci:asan"],
     },
     "debug-tools/rocgdb/source": {
@@ -54,6 +62,9 @@ SUBMODULE_CONFIG = {
         # We will reuse the rocm-systems token for now.
         "token_key": "systems",
         "branch": "amd-staging-rocgdb-16",
+        # Changes to rocgdb can run a limited matrix of CI jobs:
+        #   * Build for all gfx archs
+        #   * rocgdb tests only (no impact on other project builds/tests)
         "labels": [*COMMON_CI_LABELS, "test:rocgdb"],
     },
     "third-party/sysdeps/linux/amd-mesa/mesa-fork": {
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -16,7 +16,7 @@ THEROCK_MAIN_BRANCH = "main"
 BOT_NAME = "therockbot"
 BOT_EMAIL = "therockbot@amd.com"
 
-CI_LABELS = ["ci:run-all-archs", "ci:asan"]
+COMMON_CI_LABELS = ["ci:run-all-archs"]
 
 ROCM_SYSTEMS_FILES = [
     ".github/workflows/therock-ci-linux.yml",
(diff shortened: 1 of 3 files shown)


lumachad:
@ScottTodd Is this roughly what you had in mind?

ScottTodd:
Roughly, yeah. Maybe use just `rocgdb` or `rocgdb-all` instead of `all-rocgdb` for different alphabetic sorting.

lumachad:
> Roughly, yeah. Maybe use just `rocgdb` or `rocgdb-all` instead of `all-rocgdb` for different alphabetic sorting.

Went with test:rocgdb.

lumachad:
I'm thinking we'll want to enable ci:asan in the future. If we're only running rocgdb tests, we'll save a bit of time anyway. But we'll deal with that when asan is a bit more stable and clean test-wise.

build_tools/github_actions/bump_automation.py
ScottTodd, line 77: How about we add some comments here explaining the intent?

Something like this:

```python
SUBMODULE_CONFIG = {
    "rocm-systems": {
        "repo": "ROCm/rocm-systems",
        "files": ROCM_SYSTEMS_FILES,
        "updater": "ref",
        "token_key": "systems",
        # Changes to rocm-systems should run the full matrix of CI jobs:
        #   * Build for all gfx archs
        #   * Build for all variants (asan)
        #   * All builds and tests (including downstream rocm-libraries jobs)
        "labels": [*COMMON_CI_LABELS, "ci:asan"],
    },
    "rocm-libraries": {
        "repo": "ROCm/rocm-libraries",
        "files": ROCM_LIBRARIES_FILES,
        "updater": "ci-env",
        "token_key": "libraries",
        # Changes to rocm-libraries should run the full matrix of CI jobs:
        #   * Build for all gfx archs
        #   * Build for all variants (asan)
        #   * All rocm-libraries tests 
        "labels": [*COMMON_CI_LABELS, "ci:asan"],
    },
    "debug-tools/rocgdb/source": {
        "repo": "ROCm/rocgdb",
        "files": [],
        "updater": "submodule-only",
        # We will reuse the rocm-systems token for now.
        "token_key": "systems",
        "branch": "amd-staging-rocgdb-16",
        # Changes to rocgdb can run a limited matrix of CI jobs:
        #   * Build for all gfx archs
        #   * rocgdb tests only (no impact on other project builds/tests)
        "labels": [*COMMON_CI_LABELS, "test:rocgdb"],
    },
    "third-party/sysdeps/linux/amd-mesa/mesa-fork": {
        "repo": "ROCm/mesa-fork",
        "files": [],
        "updater": "submodule-only",
        # We will reuse the rocm-systems token for now.
        "token_key": "systems",
        # Changes to mesa-fork can run a limited matrix of CI jobs:
        #   * Build for all gfx archs
        #   * media-libs tests only (no impact on other project builds/tests)
    },
}
```

---

then seeing it this way, we could also consider moving this part of "determine test type" into labels by having the bump automation apply `test_filter:standard` instead of https://github.com/ROCm/TheRock/blob/df21e24c90357697e202a03261d7678c4d1e245d/build_tools/github_actions/configure_multi_arch_ci.py#L887-L892

I could go either way on that, implicit or explicit. We do get submodule update PRs outside of this automation that having the implicit behavior as it is now is useful.

build_tools/github_actions/tests/fetch_test_configurations_test.py
```diff
@@ -546,6 +546,69 @@ def fake_get_all_families(_):
         hipblas = next(j for j in components if j["job_name"] == "hipblas")
         self.assertEqual(hipblas["test_runner"], "linux-gfx942-default")
 
+    # -----------------------
+    # TEST_LABEL_GROUPS expansion
+    # -----------------------
+
+    def test_all_rocgdb_label_selects_cpu_and_gpu_jobs(self):
+        """test:rocgdb should expand to both rocgdb-cpu and rocgdb-gpu."""
+        os.environ["TEST_LABELS"] = json.dumps(["test:rocgdb"])
```
ScottTodd: Rather than mutate the persistent/global `os.environ` in each test case, how about either:
1. Patching/mocking environment variables using code like `with patch.dict(os.environ`
2. Passing data through parameters into `self._get_components()` (or a public function - careful about testing implementation details with `_` prefixes), rather than only passing them through environment variables

ScottTodd (approved):
Generally LGTM, just one comment about test case design and code comments/documentation.

lumachad:
Thanks for the feedback. Let me go and address those.

geomin12 (approved):

build_tools/github_actions/bump_automation.py
lumachad: I'm leaning towards implicit here until the components have time to adapt to it. But we could introduce the label strategy already. But it sounds like we should do this in a separate PR. Does that sound reasonable?

build_tools/github_actions/tests/fetch_test_configurations_test.py
lumachad: I think I've address this. Please let me know otherwise. A further simplification with a public function getting arguments is possible, but might be best done as a separate PR.

ScottTodd (approved):

Issue #1293, opened 2025-08-20
[Feature]: Enable Windows Control Flow Guard for the HIP stack


chinmaydd:
### Suggestion Description

Recently there was an ask from a customer to enable the Windows [Control Flow Guard](https://learn.microsoft.com/en-us/windows/win32/secbp/control-flow-guard) across the compiler and runtime components -- when building the compute stack on Windows.

We have verified that enabling CFG leads to minimal performance impact and have now enabled it by default in the internal builds. TheRock should either adopt the same or provide a command line option to enable Windows CFG when building the compiler and runtime artifacts (until the impact can be verified).

Possible `CMakeLists.txt` modifications :

```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /guard:cf")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /guard:cf")
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /guard:cf /DYNAMICBASE")
```

Refer: SWDEV-546434

### Operating System

Windows

### ROCm Component

Compiler and runtime

Hoernchen:
How minimal is the impact and which workload was measured? I am somewhat concerned, because past experience has shown that is absolutely kills performance unless hand picked declspecs to disable it for hot paths are scattered all over the code after profiling...

lamb-j:
Can we create a top-level TheRock arg to turn this on? (OFF by default)

PR #7309, opened 2026-08-12
feat(ci): Add top-level TheRock flag WINDOWS_DRIVER_ABI to align with downstream Drivers builds 

chinmaydd:
## Motivation

The AMD driver package build compiles every Windows component with Control Flow
Guard and loads comgr under a driver-specific DLL name. TheRock does neither.
Adds a flag for both, default `OFF`.

Not addressed: CRT linkage, SDK/toolset pinning.

ISSUE ID https://github.com/ROCm/TheRock/issues/1293

## Technical Details

`THEROCK_FLAG_WINDOWS_DRIVER_BUILD` (default `OFF`, no effect off Windows):

1. **CFG** — `/guard:cf` compile (`-mguard=cf` under `COMPILER_TOOLCHAIN`) and
   `LINKER:/guard:cf` link. `/machine` and `/DYNAMICBASE` omitted: CMake emits
   the former, the linker defaults to the latter.
2. **`COMGR_DLL_NAME=amd_comgr_drivers.dll`** via the flag system's
   `CMAKE_VARS`/`SUB_PROJECTS`, scoped to `amd-comgr`, `hip-clr`, `ocl-clr`.
   Both CLR projects build `rocclr`, which is where the name is consumed.

The compile half goes in the generated toolchain file (as
`THEROCK_MINIMAL_DEBUG_INFO` does). The link half must use `add_link_options()`
in `_init.cmake` instead: that file runs during `project()` before
`enable_language()` populates `CMAKE_<TYPE>_LINKER_FLAGS` from `*_INIT`, so the
existing `string(APPEND CMAKE_SHARED_LINKER_FLAGS " /LIBPATH:...")` shadows the
cache and drops anything the toolchain set. The first build of this PR produced
binaries with `/guard:cf` on every compile line and no CFG bit.

zlib/zstd sysdeps run a nested `cmake` that forwards only compilers, so flags
are forwarded there too — CFG instrumentation is per-object, so linking with
`/guard:cf` cannot retrofit checks into uninstrumented static libs.

Files: `FLAGS.cmake`, `cmake/therock_subproject.cmake`,
`third-party/sysdeps/common/{zlib,zstd}/CMakeLists.txt`.

## Test Plan

Full clean Windows build, flag `ON`: MSVC 19.44, Ninja, `gfx110X-all`,
`THEROCK_ENABLE_ALL=OFF` + `COMPILER`/`HIP_RUNTIME`/`OCL_RUNTIME`. Then
`dumpbin /headers` over `dist/rocm/bin`. Configure-only check of the `OFF` path.

## Test Result

`BUILD_EXIT_CODE=0`, no compile or link errors. CFG set on all 11 shipped
binaries (`amd_comgr_drivers.dll`, `amdhip64_7.dll`, `amdocl64.dll`,
`hiprtc*.dll`, `cltrace.dll`, `OpenCL.dll`, `rocm_kpack.dll`, `clinfo.exe`,
`hipInfo.exe`, `hipcc.exe`) plus clang/lld-link; none missing. comgr built and
loaded as `amd_comgr_drivers.dll`. Flag `OFF`: no guard references, no
`COMGR_DLL_NAME` injection, forwarding intact.

## Submission Checklist

### DIFF
--- a/FLAGS.cmake
+++ b/FLAGS.cmake
@@ -105,6 +105,19 @@ therock_declare_flag(
     amd-llvm
 )
 
+therock_declare_flag(
+  NAME WINDOWS_DRIVER_BUILD
+  DEFAULT_VALUE OFF
+  DESCRIPTION "Windows: build for the AMD driver package (Control Flow Guard, driver comgr DLL name)"
+  GLOBAL_PROPAGATE_FLAG
+  CMAKE_VARS
+    COMGR_DLL_NAME=amd_comgr_drivers.dll
+  SUB_PROJECTS
+    amd-comgr
+    hip-clr
+    ocl-clr
+)
+
 ###############################################################################
 # Branch-specific flag overrides.
 # BRANCH_FLAGS.cmake is .gitignored on main but can be committed on
--- a/cmake/therock_subproject.cmake
+++ b/cmake/therock_subproject.cmake
@@ -87,6 +87,15 @@ if(WIN32)
   list(APPEND THEROCK_AMD_LLVM_DEFAULT_CXX_FLAGS -Wno-duplicate-decl-specifier)
 endif()
 
+# Options added to every subproject when THEROCK_FLAG_WINDOWS_DRIVER_BUILD is set.
+# The compile flag is spelled per compiler; the link flag goes through CMake's
+# LINKER: prefix, which handles the driver difference.
+# /machine and /DYNAMICBASE are omitted: CMake emits the former and the linker
+# defaults to the latter.
+set(THEROCK_WINDOWS_DRIVER_BUILD_MSVC_COMPILE_FLAGS "/guard:cf")
+set(THEROCK_WINDOWS_DRIVER_BUILD_CLANG_COMPILE_FLAGS "-mguard=cf")
+set(THEROCK_WINDOWS_DRIVER_BUILD_LINK_FLAGS "/guard:cf")
+
 # Generates a command prefix that can be prepended to any custom command line
 # to perform log/console redirection and pretty printing.
 # LOG_FILE: If given, command output will also be sent to this log file. If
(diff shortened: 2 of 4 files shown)


lamb-j (approved):

astrelsky:
Is control flow guard really necessary for userland code though? The msvc implementation of control flow guard is an abomination I've never fully wrapped my head around. I don't see a reason to be concerned about this in the rocm libraries, but I guess maybe that depends on where and how they're being used.

Obviously the driver needs to have control flow guard enabled. It's not perfect, but it makes gaining execution control significantly harder. rop gadgets would still work because return addresses aren't subject to cfg.



Edit: oh shoot I think I missed that some of these components may be included in the driver package. In which case this is correct and I should just be ignored 😅.

chinmaydd:
Appreciate your comments @astrelsky.

> Is control flow guard really necessary for userland code though? ... I don't see a reason to be concerned about this in the rocm libraries, but I guess maybe that depends on where and how they're being used.

Right, that makes sense. I think for most rocm-libraries (artifacts) that are critical to performance and are used exclusively in the userland, we probably arent worried about enabling the MSVC-provided CFG mitigation yet. It is however, a step towards preventing some (if not most) control flow related attack vectors that is provided by the compiler. I am quite the fan of such "free" security wins that likely will continue to improve over time. We should adopt it unilaterally when the performance cost-benefit ratio works in our favor.

> The msvc implementation of control flow guard is an abomination I've never fully wrapped my head around

I see. In the few experiments that we ran I didnt really see a major impact. Probably because the code paths that we were dealing with didnt involve much CFG related instrumentation.

> Obviously the driver needs to have control flow guard enabled. It's not perfect, but it makes gaining execution control significantly harder. rop gadgets would still work because return addresses aren't subject to cfg.

Sure, but I think setting up a precise ROP chain also involves having a write primitive which I'd like to think we dont expose. Unless the attack is so well designed that it uses known control flow targets in weird ways.

The reason for including this is also motivated by customer needs and requirements. And I tend to agree that for components that sit inside the driver, this is something that we definitely should work towards enabling.

Issue #7202, opened 2026-08-07
[issue] [asan] Support runtime-only host-ASan presubmit builds and tests


sa-faizal:
Parent tracking issue: #6624

## Summary

Make a focused host-ASan workflow the default automated presubmit for every qualifying pull request in rocm-systems.

The automated rocm-systems PR run must:

- build only `compiler-runtime` and `runtime-tests`
- run only `hip-tests` and `rocrtst`
- skip `math-libs` and every unrelated build stage
- skip packaging, Python packages, framework wheels, and other full-distribution validation
- enable host-ASan test runners for the presubmit trigger

This is not an opt-in test mode. It must run automatically when a rocm-systems pull request is opened, synchronized, or reopened, subject only to the workflow's normal PR/path filters. It must not require a `ci:asan` label, a test-component label, or manual dispatch.

## Motivation

The current reusable multi-arch workflow expands a host-ASan presubmit into the full build graph. Expensive stages such as `math-libs` are unnecessary for runtime validation and can extend the run by multiple hours. Sanitizer component tests may also be disabled for non-nightly triggers, leaving the presubmit without the intended HIP/runtime test coverage.

A fixed runtime-only host-ASan presubmit provides useful sanitizer signal on every rocm-systems PR while keeping runner usage and turnaround time practical.

## Requested behavior

1. Automatically trigger the focused host-ASan workflow for every qualifying rocm-systems pull request. No ASan label or manual action is required.
2. The automated PR build scope is fixed to:
   - `compiler-runtime`
   - `runtime-tests`
3. The automated PR test scope is fixed to:
   - `hip-tests`
   - `rocrtst`
4. PR labels must not expand the host-ASan test matrix beyond those two components.
5. Keep the generic sanity test and omit all full-build-only post-processing from this PR path, including:
   - complete artifact-structure validation
   - native package builds and install tests
   - Python package builds/tests
   - PyTorch and JAX builds
6. TheRock's reusable workflows must support this fixed downstream configuration while preserving current behavior for unrelated callers.
7. Full-ASan nightly and manual-dispatch workflows remain unchanged.

## Acceptance criteria

- Opening, synchronizing, or reopening a qualifying rocm-systems PR automatically starts a host-ASan run.
- The automated run does not require `ci:asan`, `ci:host-asan`, `test:*`, or any other opt-in label.
- The build graph runs `compiler-runtime` and `runtime-tests` only.
- Every other build stage, including `math-libs`, is skipped.
[...]

Issue #6624, opened 2026-07-16
[issue] (asan) ROCm ASAN enablement


geomin12:
Enable AddressSanitizer (host + device) across ROCm with QA-validated artifacts.

## Current State

| Repo | Pre-submit | Post-submit | Submodule Bump | Nightly | Artifacts |
|------|------------|-------------|----------------|---------|-----------|
| **TheRock** | host-asan | host-asan | full-asan | full-asan | tarball (beta) |
| **R-L** | disabled | disabled | N/A | disabled | — |
| **R-S** | disabled | disabled | N/A | disabled | — |

- Full-asan runs on Rock nightly + R-L/R-S submodule bumps
- No dedicated ASAN runners (competes with regular CI, queue times up to 440 min)
- Math-libs is critical bottleneck: 183-443 min for gfx94x, 117-391 min for gfx950
- No QA validation or test execution yet

## Future State

| Repo | Pre-submit | Post-submit | Submodule Bump | Nightly | Artifacts |
|------|------------|-------------|----------------|---------|-----------|
| **TheRock** | host-asan | host-asan | full-asan | full-asan | tarball, deb, rpm |
| **R-L** | host-asan | host-asan | N/A  | full-asan | — |
| **R-S** | host-asan | host-asan | N/A  | full-asan | — |

- Dedicated AWS runners with auto-scaling + MI325X/MI350 GPU sandboxes
- QA-validated ASAN artifacts
- Test execution on MI325X (gfx94x)

## Milestones

| Release | Workstream | Goal |
|---------|------------|------|
| **ROCm 7.14** | Packaging | Beta ASAN tarball artifacts (not QA validated) |
| **ROCm 7.15** | QA | QA validation of nightly ASAN artifacts begins |
| | CI Infrastructure | Full-ASAN nightly in R-L and R-S |
| | CI Infrastructure | Host-ASAN pre-submit and submodule bump gates |
| | Test Execution | Begin test execution from MI325X (gfx94x) |
| | Packaging | Roll out deb and rpm packages |
| **ROCm X.XX** | CI Infrastructure | Stable testing, build/test optimization |

sa-faizal:
Including the packaging phase optimization (using **zstd compression**) which reduced RPM ASAN packaging time from **5hrs to less than 1hr**. https://github.com/ROCm/TheRock/pull/6455

sa-faizal:
[...]

PR #7311, opened 2026-08-12
feat(ci): support runtime-only host-ASAN presubmit builds and tests

dileepr1:
Towards #7202 (parent: #6624).

## Problem

A host-ASAN presubmit currently expands into the full build graph. `math-libs` alone dominates the run (#6624 cites 183-443 min for gfx94x, 117-391 min for gfx950), and on top of that ASAN tests are disabled for any non-nightly trigger, so the presubmit costs hours and produces no sanitizer signal.

## Approach

This is the TheRock half of #7202: it adds the reusable-workflow support so a downstream caller can pin a narrow ASAN presubmit. Every input defaults to today's behavior, so nightly/manual ASAN runs and all other callers are unaffected. The rocm-systems side (adopting this in `therock-multi-arch-ci-asan.yml`) is the follow-up the issue describes.

New `setup_multi_arch.yml` inputs:

| Input | Default | Effect |
| --- | --- | --- |
| `build_stages` | `""` | Allowlist of build stages; the complement is dropped from the graph entirely. |
| `asan_presubmit` | `false` | Runs ASAN with no `ci:asan`/`ci:host-asan` label and allocates sandbox test runners on `push`/`pull_request`. |
| `lock_test_labels` | `false` | PR `test:*` labels no longer extend the caller's test scope or escalate the test level. |
| `validate_artifact_structure` | `true` | Set `false` for partial builds, which have no complete artifact tree. |
| `build_native_linux` | `true` | Set `false` to skip deb/rpm builds and their install tests. |
| `sanity_blocks_components` | `true` | Set `false` so a failing sanity check no longer skips the component tests. |

`sanity_blocks_components` addresses a failure mode this presubmit runs into immediately. Today a failing sanity check skips every component test, so a single failure hides all of the results a run was meant to produce. The host-ASAN presubmit hits exactly that: sanity reports a LeakSanitizer finding inside `libamdhip64` and `hip-tests` and `rocrtst` never run, even though those are the results the presubmit exists to collect. With the input set to `false` they run regardless, while sanity still runs and still reports its own status — the failure stays visible rather than being suppressed. The gate now also checks the sanity and configure job results explicitly instead of using `!failure()`, which cascaded from any dependency.

`build_stages` is deliberately an allowlist rather than a skip-list, so adding a stage to `BUILD_TOPOLOGY.toml` later cannot silently widen an existing caller's scope. It is distinct from `prebuilt_stages`: a prebuilt stage still contributes artifacts copied from a baseline run, whereas a stage outside `build_stages` is never built and has no artifacts at all. The complement is computed in `configure_multi_arch_ci.py` and passed to the build workflows as a new `skip_stages` input; an unknown stage name raises instead of silently building everything.

Combined with the existing `build_python_packages`, `build_pytorch` and `build_jax` switches, a caller gets the configuration #7202 asks for:

```yaml
with:
  build_variant: "asan"
  build_stages: "compiler-runtime,runtime-tests"
  linux_test_labels: "test:hip-tests,test:rocrtst"
  asan_presubmit: true
  lock_test_labels: true
  validate_artifact_structure: false
  build_native_linux: false
  build_python_packages: false
  build_pytorch: false
  build_jax: false
```

## Acceptance criteria

| Criterion from #7202 | Status |
| --- | --- |
| PR starts a host-ASAN run automatically | `asan_presubmit` bypasses the label gate in `should_skip_ci` |
| No `ci:asan` / `ci:host-asan` / `test:*` opt-in needed | same, plus `lock_test_labels` for `test:*` |
| Build graph runs `compiler-runtime` and `runtime-tests` only | `build_stages` allowlist |
| Every other stage, including `math-libs`, is skipped | `skip_stages` complement guards each stage's `if:` |
| Test matrix runs `hip-tests` and `rocrtst` only | `linux_test_labels` + `lock_test_labels` |
| Host-ASAN tests get an eligible sandbox runner on PR | `asan_presubmit` in `_expand_build_config_for_platform` |
| Native packages, artifact validation, Python, PyTorch, JAX skipped | `build_native_linux`, `validate_artifact_structure`, existing switches |
| Unrelated callers keep current behavior | all inputs default to current values |
| Full-ASAN nightly / dispatch unchanged | `asan_presubmit` is the only path that alters trigger handling |

## Verification

Ran the configure script end-to-end with the environment `setup_multi_arch.yml` would produce for a rocm-systems PR carrying an unrelated `test:rocblas` label:

```
build_variant_label         = 'host-asan'
build_variant_cmake_preset  = 'linux-release-host-asan'
skip_stages                 = 'wsl-rocdxg,math-libs,comm-libs,cv-libs,storage-libs,debug-tools,dctools-core,profiler-apps,media-libs'
validate_artifact_structure = False
build_native_linux          = False
build_python_packages       = False
build_pytorch               = False
build_jax                   = False
test-runs-on per family     = [('gfx94X-dcgpu', 'linux-mi325-gpu-rocm-cpu-sandbox')]
enable_build_jobs           = true
linux_test_labels           = ['test:hip-tests', 'test:rocrtst']
```

The same event with only `BUILD_VARIANT=asan` still reports `Skipping: ASAN PR without enabling label`, and a default release PR still produces `skip_stages=''` with all post-processing enabled.

## Test plan

- [x] `pytest build_tools/github_actions/tests/` — 792 passed, up from 757 on `main` (+35 new tests). The 2 failures in `stage_reuse_decision_test.py` are pre-existing on `main` and unrelated.
- [x] `pre-commit run` on all touched files, including `actionlint` and `black`.
- [x] The existing `TestBuildConfigWorkflowContract` field-parity tests cover the new `skip_stages`, `validate_artifact_structure` and `sanity_blocks_components` fields in both the Linux and Windows CI workflows.
- [x] `test_component.yml` has a single job with no job-level `if:`, so `test_sanity_check` can only resolve to `success`, `failure` or `cancelled` — never `skipped`. Replacing `!failure()` with an explicit result check therefore does not change behavior for any existing caller.
- [x] Confirmed on a real downstream run. [ROCm/rocm-systems#10088](https://github.com/ROCm/rocm-systems/pull/10088) points its ASAN workflow at this branch and passes the focused configuration; its `setup` job produced exactly the intended decisions (details in the comment below).

## Known limitation, not addressed here: rocrtst's timeout under ASAN

The downstream validation shows `rocrtst`, one of the two components this presubmit pins, being killed at its 15-minute `timeout_minutes` after spending roughly 14 of them inside `Memory_Alignment_Test`. That budget is tuned for a release build; instrumented, it cannot finish regardless of whether anything is wrong.

`hip-tests` is unaffected because it already accounts for instrumentation both at the job level (120 minutes) and per test (`timeout = 1500 if is_asan_instrumented() else 600`). `rocrtst` has neither.

Deliberately left out of this PR: making `timeout_minutes` variant-aware changes shared test configuration for every component and caller, which is a wider blast radius than the opt-in inputs here and is better reviewed on its own.

## Note for reviewers

With `linux_test_labels` set, `test_type` resolves to `full` for the two selected components, which is existing behavior for any caller that specifies test labels. A caller wanting a shallower presubmit can add `test_filter:standard` (or `quick`) to `linux_test_labels`, since `_determine_test_type` scans that list too. Happy to change the default here if you'd prefer the focused presubmit to pin a level explicitly.

## Stacked on #7318

This branch carries one commit that belongs to [#7318](https://github.com/ROCm/TheRock/pull/7318), `fix(tests): preload the ASAN runtime for host-asan builds`, and it should be reviewed there rather than here.

It is present because the downstream validation in [ROCm/rocm-systems#10088](https://github.com/ROCm/rocm-systems/pull/10088) checks TheRock out at the ref it pins, so the sanity test it fixes only runs if the fix is on *this* branch. Without it, sanity fails and the `hip-tests` and `rocrtst` jobs this PR is meant to schedule never get created.

Once #7318 merges, rebasing this branch drops the commit and the diff here goes back to the CI-scoping change alone.


dileepr1:
## Validated against a real downstream caller

[ROCm/rocm-systems#10088](https://github.com/ROCm/rocm-systems/pull/10088) is a draft PR that points its `therock-multi-arch-ci-asan.yml` at this branch and passes the focused configuration from the issue. The `setup` job on that PR produced exactly the intended decisions ([run 31645503865](https://github.com/ROCm/rocm-systems/actions/runs/31645503865)):

```
lock_test_labels: True
External repo detected: rocm-systems
  Running: ASAN configured as an automatic presubmit gate
  External repo build: skipping path filter checks, using stage reuse
  build_stages allowlist: ['compiler-runtime', 'runtime-tests']
  Skipping stages outside the allowlist: ['wsl-rocdxg', 'math-libs', 'comm-libs',
    'cv-libs', 'storage-libs', 'debug-tools', 'dctools-core', 'profiler-apps', 'media-libs']
  Using host-asan variant (push/pull_request default)
  gfx94x: using host-asan sandbox runner: linux-mi325-gpu-rocm-cpu-sandbox
```

Mapping that back to the acceptance criteria in ROCm/TheRock#7202:

| Criterion | Evidence |
| --- | --- |
| Runs with no `ci:asan` / `ci:host-asan` label | the PR carries no ASAN label and the run started |
| Host-ASAN, not full ASAN | job renders as `Linux::host-asan`; preset `linux-release-host-asan` |
| Only runtime stages build | `skip_stages` covers all nine other stages; the only build job is `Stage - Compiler Runtime` |
| ASAN sandbox test runner | `test-runs-on: linux-mi325-gpu-rocm-cpu-sandbox` |
| No post-processing | `validate_artifact_structure`, `build_native_linux`, `build_python_packages`, `build_pytorch`, `build_jax` all `false` |
| Test scope pinned | `linux_test_labels=['test:hip-tests', 'test:rocrtst']` with `lock_test_labels: true` |

## One fix this surfaced

The first validation run exposed a gap that the unit tests could not see, because it lives in the workflow's job graph rather than in the Python.

`test_python_packages_per_family` needs `build_python_packages`, but it guards only on `test_python_packages_matrix` being non-empty — unlike the PyTorch and JAX jobs, it has no `build_python_packages == true` check. Its `if:` contains `!cancelled()`, which stops GitHub from propagating the skip from the job it needs. So with `build_python_packages: false` the build job was correctly skipped while `Test Python` would still have run, and would have tried to install wheels that were never built.

`da44239` generates the matrix only when Python packages are actually built, which matches how `pytorch_build_matrix` and `jax_build_matrix` are already handled and needs no workflow change. The re-run confirms `"test_python_packages_matrix": []`. This was reachable before this PR by any caller passing `build_python_packages: false`, so it is worth having regardless of the rest of this change.

The build itself takes several hours; I will follow up here once the downstream job list is confirmed end to end.

raramakr:
Corresponding rocm-system PR : https://github.com/ROCm/rocm-systems/pull/10558

.github/workflows/setup_multi_arch.yml
```diff
@@ -73,6 +81,39 @@ on:
         description: "Build JAX wheels."
         type: boolean
         default: false
+      build_native_linux:
+        description: "Build native Linux packages (deb/rpm) and run their install tests."
+        type: boolean
+        default: true
```
ScottTodd, line 87: Please split this off into its own smaller PR. Large PRs that bundle multiple changes together like this one are considerably slower to review (this currently 800 LOC PR has been in my queue for over a week for example, a a 100 LOC PR would be easier to handle).

I want to review all new inputs to `.github/workflows/setup_multi_arch.yml` carefully, and it looks like the main thing this "host-ASAN presubmit" PR should be adding is the `asan_presubmit` option, all the other options could be prerequisite PRs.

build_tools/print_driver_gpu_info.py
```diff
@@ -24,19 +24,47 @@
...
+# Import ASAN helpers from amdgpu_family_matrix
+_THIS_DIR = Path(__file__).resolve().parent
+_GITHUB_ACTIONS_DIR = _THIS_DIR / "github_actions"
+if _GITHUB_ACTIONS_DIR.exists():
+    sys.path.insert(0, str(_GITHUB_ACTIONS_DIR))
+    from amdgpu_family_matrix import get_asan_lib_path, is_asan_instrumented
+else:
+    # Fallback if running standalone
+    def is_asan_instrumented():
+        return os.getenv("BUILD_VARIANT", "") in ("asan", "host-asan")
```
ScottTodd, line 36: I would prefer for this `print_driver_gpu_info.py` file to remain a standalone script that can run on _any system_, CI or otherwise. For that, it should not read CI configuration information from environment variables.

Looking at https://github.com/ROCm/TheRock/blob/e0d203f0c072431d2baa79760626cd79d771d2f4/.github/workflows/test_component.yml#L161-L179

we could have a "setup ASan" step that sets `LD_PRELOAD` and `ASAN_OPTIONS` environment variables for the whole job, then we wouldn't need to update individual test scripts with ASan support like this. Tests should generally follow our user instructions, which in this case would be https://github.com/ROCm/TheRock/blob/main/docs/development/sanitizers.md#using-asan-instrumented-libraries. Hopefully we can update our tests to
1. Install ASan artifacts/packages
2. Set ASan-related environment variables
3. Run tests in the same way they are run without ASan

I bet some project tests will need special handling there, but I'd rather handle it case by case in those tests and not by teaching generic utilities like this "print driver info" about build variants.

build_tools/print_driver_gpu_info.py
```diff
@@ -24,19 +24,47 @@
 import sys
 from typing import List, Optional
 
+# Import ASAN helpers from amdgpu_family_matrix
+_THIS_DIR = Path(__file__).resolve().parent
+_GITHUB_ACTIONS_DIR = _THIS_DIR / "github_actions"
+if _GITHUB_ACTIONS_DIR.exists():
+    sys.path.insert(0, str(_GITHUB_ACTIONS_DIR))
+    from amdgpu_family_matrix import get_asan_lib_path, is_asan_instrumented
```
ScottTodd, line 32: Why are ASAN helpers in `amdgpu_family_matrix`? See also my comment below about keeping this script standalone and generic.

tests/test_rocm_sanity.py
```diff
@@ -20,19 +20,19 @@
 
 AMDGPU_FAMILIES = os.getenv("AMDGPU_FAMILIES")
 
-# Importing is_asan from amdgpu_family_matrix.py
+# Importing ASAN helpers from amdgpu_family_matrix.py
 sys.path.append(str(THIS_DIR.parent / "build_tools" / "github_actions"))
-from amdgpu_family_matrix import is_asan
+from amdgpu_family_matrix import get_asan_lib_path, is_asan, is_asan_instrumented
```
ScottTodd, line 25: ASan helpers do not belong in `amdgpu_family_matrix.py`

.github/workflows/setup_multi_arch.yml
```diff
@@ -73,6 +81,39 @@ on:
...
+      sanity_blocks_components:
+        description: >-
+          Whether a failing sanity check prevents the component tests from
+          running. Set to false to keep collecting component results while a
+          known sanity failure is being resolved; the sanity check still runs
+          and still reports its own status.
+        type: boolean
+        default: true
```
ScottTodd, line 102: I don't think we should add this input. Sanity check tests *need* to be actual sanity checks. If a sanity check fails, tests should not run, otherwise there is no point of them existing and we should remove that job or make it always run in parallel to the other component test jobs.

.github/workflows/setup_multi_arch.yml
```diff
@@ -73,6 +81,39 @@ on:
...
+      lock_test_labels:
+        description: >-
+          Treat the caller's test labels as the complete test scope: PR 'test:*'
+          labels are ignored instead of extending the matrix.
+        type: boolean
+        default: false
```
ScottTodd, line 108: I don't understand the use case for this label. Can you elaborate? (Also please split this PR into smaller PRs, there is too much intersecting to review)

.github/workflows/setup_multi_arch.yml
```diff
@@ -73,6 +81,39 @@ on:
...
+      asan_presubmit:
+        description: >-
+          Run ASAN as an automatic presubmit gate. No 'ci:asan'/'ci:host-asan'
+          label is required, and ASAN sandbox test runners are allocated on
+          push/pull_request instead of nightly triggers only. Intended for
+          callers that also narrow the build via build_stages.
+        type: boolean
+        default: false
```
ScottTodd, line 116: Please align with @geomin12 on how you want to have ASan jobs trigger in each repository. There is another PR making a similar change under review at
* https://github.com/ROCm/TheRock/pull/7510#discussion_r3832358274

ScottTodd (changes requested):
This PR should be split into 3-5 smaller PRs. There is too much to review here in a single PR.

raramakr:
We'll create multiple smaller PRs to make reviewing easier.

geomin12 (changes requested):
i dropped a similar comment here: https://github.com/ROCm/TheRock/pull/7600#pullrequestreview-5022087447 in regards to skip stage

should we just re-use known good artifacts?

raramakr:
> i dropped a similar comment here: [#7600 (review)](https://github.com/ROCm/TheRock/pull/7600#pullrequestreview-5022087447) in regards to skip stage
> 
> should we just re-use known good artifacts?

We can't reuse artifacts, since the option is mainly for asan-host pre-submit

PR #7332, opened 2026-08-13
fix(packaging): correct the install harness GPG key import

lsudarsh-amd:
## Motivation

`native_linux_package_install_test.py` is an existing install test that CI runs today. It downloads a repository signing key and saves it to disk so apt can check package signatures. Three things are wrong with how it does that, and all of them predate the `amdrocm-repo` work.

**The import could fail and still report success.** The key was fetched with a single `shell=True` pipeline. A pipeline reports the exit status of its last command, so `check=True` was only ever checking `tee`. If the download returned something that wasn't a key, `gpg --dearmor` failed, `tee` wrote an empty keyring anyway, and the harness printed `[PASS]` and returned `True`. The run then failed much later on a signature error that gave no hint where it came from. This isn't a corner case. `gpg --dearmor` exits 2 on an empty body or on an HTML error page served with a 200, which is a normal thing for a server to return.

**The documented `ROCM_APT_KEYRING_FILE` override didn't work.** Only the `signed-by=` line used it. The code that writes the key built its own path and hardcoded `rocm.gpg`, so setting the variable saved the key in one place and told apt to look in another. `apt update` then failed with "The repository is not signed".

**The default path clashes with AMD's install instructions.** Those instructions put the ROCm signing key at `/etc/apt/keyrings/rocm.gpg`, and the harness writes over it with `sudo tee`. This doesn't happen in CI, which runs in throwaway containers, but it does on the VMs and workstations the script is also meant to run on.

Split out of #7276 because it changes existing code and doesn't depend on the new `amdrocm-repo` feature.

Contributes to #6986.

## Technical Details

**The shell pipeline is gone.** The import now runs as three `subprocess.run` calls, one each for `wget`, `gpg --dearmor` and `sudo tee`. Each takes its arguments as a list and sets `check=True`, so a failure stops the import there instead of leaving an empty keyring behind. The URL and the keyring path are no longer interpolated into a shell string, so neither can be used for injection.

**The keyring path comes from one constant.** `setup_gpg_key()` takes both the file and its parent directory from `APT_KEYRING_FILE`. That fixes the override, and it lets the default move to `/etc/apt/keyrings/{REPO_NAME}.gpg`, the way `APT_SOURCES_LIST` already works. The rename only works because of that. On its own it would have written to `rocm.gpg` while telling apt to check `rocm-test.gpg`. `APT_KEYRING_DIR` goes too, along with its docstring entry, since nothing reads it any more and nothing in the repo sets it.

**Paths are `PurePosixPath`.** These are paths on the target Linux filesystem, handed to `mkdir`, `tee` and `chmod`. `Path` follows the local flavour and renders `\etc\apt\keyrings` on Windows, which fails the `windows-2022` job.

**A timeout now returns `False`.** `subprocess.TimeoutExpired` is a `SubprocessError`, so neither the `CalledProcessError` nor the `OSError` handler caught it and a timeout killed the run instead of reporting a failed import. That was already true, but splitting the pipeline takes the timed calls in that block from two to four. Each carries `GPG_KEY_TIMEOUT_SEC` of 60, so fetch, dearmor and write together can take up to 180 seconds.

**The file runs in CI, but these lines don't.** Every change sits behind `if self.gpg_key_url:` (`:543`, `:551`, `:763`), which covers both `setup_gpg_key()` call sites, and neither caller passes a key URL. `test_native_linux_packages_install.yml:227` wires `GPG_KEY_URL` from an input (`:17`, `:62`, `:151`) nobody sets, and `multi_arch_build_native_linux_packages.yml:138` runs `--test-type simulate`, which skips repository setup.

## Test Plan

- **Unit tests.** Five new cases in `ConfiguredPathsTest`: the default doesn't collide with `/etc/apt/keyrings/rocm.gpg`, it derives from `REPO_NAME`, `APT_KEYRING_DIR` is gone, the write target matches what `signed-by=` pins, and no call uses a shell. Two new cases in `SetupGpgKeyTest` for the bad body and the timeout, plus the existing sequence check tightened to compare two argv tokens instead of one.
- **End to end against a signed repo.** Build a real GPG-signed apt repository in a container, serve it over localhost, and drive the genuine `setup_deb_repository()`, which runs a real `apt update`. Run the same scenarios against `main` to confirm each one fails there.
- **Full suite.** `pytest build_tools/` in the CI container against a freshly measured `origin/main` baseline, comparing failure names rather than counts.
- **Negative controls.** For every new assertion, break the code it covers, confirm the test fails, then restore.
- **Lint.** `pre-commit` over the changed files.

## Test Result

| Check | Result |
|---|---|
| Unit tests | Pass. 125 in the module, including the 7 new cases. |
| End to end against a signed repo | Pass, 12/12. The same suite scores 6/12 on `main`. |
| Full suite | **No new failures.** 9 failed / 1761 passed against 9 failed / 1754 on `main`. Identical failure names. |
| Negative controls | Pass. Six, each failing only the test it targets. |
| Lint | Pass. |

The six checks `main` fails are exactly the ones this fixes. It overwrites `/etc/apt/keyrings/rocm.gpg`, ignores the `ROCM_APT_KEYRING_FILE` override, and reports success on a body that isn't a key while leaving an empty keyring behind. The override case is the sharpest. There, `apt update` exits 100 with "The repository is not signed", so `main` doesn't just misplace the key, it breaks apt in a plain container.

The nine pre-existing failures are in `compute_rocm_package_version_test.py`, `jax_install_scripts_test.py` and `stage_reuse_decision_test.py`, none of which this touches.

**Not covered.** `PurePosixPath` can't be exercised on Linux. `Unit Tests :: windows-2022` covers it and passed on #7276 with this same code. The timeout is covered by unit test rather than end to end, since `GPG_KEY_TIMEOUT_SEC` is 60 and isn't overridable.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/native_linux_package_install_test.py
+++ b/build_tools/packaging/linux/native_linux_package_install_test.py
@@ -28,7 +28,7 @@ Test modes (--test-type):
   actual install. Requires --packages-dir.
 
 Path and repo name are overridable via environment variables: ROCM_REPO_NAME (repo id used for
-APT list, Zypper/Yum repo file and section), ROCM_APT_KEYRING_DIR, ROCM_APT_SOURCES_LIST,
+APT list, Zypper/Yum repo file and section), ROCM_APT_SOURCES_LIST,
 ROCM_APT_KEYRING_FILE, ROCM_ZYPP_REPOS_DIR, ROCM_YUM_REPOS_DIR,
 ROCM_RDHC_REL_PATH (relative path from install prefix to rdhc binary).
 
@@ -158,11 +158,16 @@ def _env(key: str, default: str) -> str:
 # ROCM_REPO_NAME: logical repo name used for APT list, Zypper/Yum repo file and section id
 # ROCM_APT_*, ROCM_ZYPP_*, ROCM_YUM_*, ROCM_RDHC_REL_PATH
 REPO_NAME = _env("ROCM_REPO_NAME", "rocm-test")
-APT_KEYRING_DIR = _env("ROCM_APT_KEYRING_DIR", "/etc/apt/keyrings")
 APT_SOURCES_LIST = _env(
     "ROCM_APT_SOURCES_LIST", f"/etc/apt/sources.list.d/{REPO_NAME}.list"
 )
-APT_KEYRING_FILE = _env("ROCM_APT_KEYRING_FILE", "/etc/apt/keyrings/rocm.gpg")
+# Derived from REPO_NAME like APT_SOURCES_LIST above, and deliberately not
+# /etc/apt/keyrings/rocm.gpg. No package owns that path, but the current
+# documented amdgpu package manager steps set the signing key up there
+# ("wget .../rocm.gpg.key | gpg --dearmor | sudo tee /etc/apt/keyrings/rocm.gpg")
+# and point signed-by= at it. This harness writes its keyring the same way, so
+# sharing the path would overwrite the key a host is already using.
+APT_KEYRING_FILE = _env("ROCM_APT_KEYRING_FILE", f"/etc/apt/keyrings/{REPO_NAME}.gpg")
 ZYPP_REPOS_DIR = _env("ROCM_ZYPP_REPOS_DIR", "/etc/zypp/repos.d")
 YUM_REPOS_DIR = _env("ROCM_YUM_REPOS_DIR", "/etc/yum.repos.d")
 VERIFY_KEY_COMPONENTS = [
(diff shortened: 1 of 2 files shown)


arvindcheru (commented):
Looks Good To Me.
Please verify and confirm the CI/Test results

Sharing reports I could capture: 
https://github.com/ROCm/TheRock/actions/runs/31805846179
https://github.com/ROCm/TheRock/actions/runs/31807809257
https://github.com/ROCm/TheRock/actions/runs/31807709572
https://github.com/ROCm/TheRock/actions/runs/31805980502

lsudarsh-amd:
> Looks Good To Me. Please verify and confirm the CI/Test results
> 
> Sharing reports I could capture: https://github.com/ROCm/TheRock/actions/runs/31805846179 https://github.com/ROCm/TheRock/actions/runs/31807809257 https://github.com/ROCm/TheRock/actions/runs/31807709572 https://github.com/ROCm/TheRock/actions/runs/31805980502

Thanks for the review @arvindcheru. All four runs you linked are green on 500f6f8687, the current head. They're the install tests for ubuntu2404, rhel8, rhel10 and sles16, which is exactly the harness this PR changes. Unit tests on both platforms, pre-commit, gitleaks and the bot are green too. The 10 red jobs are GPU tests plus the CI Summary aggregate. None touch build_tools, and this PR only changes Python under build_tools/packaging/linux.

If there are no other concerns, can this PR be approved for merge?

arvindcheru (approved):
There are a few tests in progress, request wait for completion Thank you.
https://github.com/ROCm/TheRock/actions/runs/32219610485/job/96112830918
https://github.com/ROCm/TheRock/actions/runs/32219309894/job/96114162407
https://github.com/ROCm/TheRock/actions/runs/32219490193/job/96114193749

lsudarsh-amd:
> There are a few tests in progress, request wait for completion Thank you. https://github.com/ROCm/TheRock/actions/runs/32219610485/job/96112830918 https://github.com/ROCm/TheRock/actions/runs/32219309894/job/96114162407 https://github.com/ROCm/TheRock/actions/runs/32219490193/job/96114193749

Thanks @arvindcheru. All three have finished and passed on 500f6f8687, nothing else is in progress.

dbieleck:
Please review.

Issue #6930, opened 2026-07-28
Bring Python scripts across the project into alignment with the style guide


ScottTodd:
We have style rules written in https://github.com/ROCm/TheRock/blob/main/docs/development/style_guides/python_style_guide.md . Some rules are less important than others, some files predate the style guide, and some files and sections of code can be considered exempt. As we're getting more large code contributions to python scripts and haven't yet documented enough of our testing philosophy (https://github.com/ROCm/TheRock/issues/6711), I think it's worth taking the time to fix several of the easier/glaring issues before too many coding agents pick up on patterns we want to avoid.

I'm planning on running a few project-wide audits to identify the scale/scope of issues and then have some agents chew through the rules one by one, flagging any cases that need more direct attention as they go.

Initial prompt:

> › A number of scripts and unit tests in TheRock are out of compliance with our style guide in docs/development/
>   style_guides/python_style_guide.md. I'd like your help auditing and fixing such issues. First can you run an audit for a few easily fixable rules? In particular I think we can handle this, but I'm interested in others that you may
>   suggest too:
>   * Access argparse attributes directly
>   * Import organization (my primary focus right now)
>   * Use `__main__` guard
>   * Use named arguments for complicated function signatures
>   * Distinguish between different error conditions
>   * Don't make assumptions about the current working directory (I think we don't have issues like this today but would
>   be good to confirm)
> 
>   Once we audit and fix enough of the smaller details, I'll want to step up and look for more difficult to fix issues
>   like code organization, testing practices (e.g. overuse of mocks, testing the standard library instead of the
>   application logic, duplicating/overlapping test coverage instead of reusing existing tests).
> 
>   For right now, I want to understand the scope of the issue with some audits. The next step would then be to fix
>   classes of issues in batches of commits/PRs (like one commit/PR for import organization, one for `__main__` guard, etc.).

and initial findings:

```
• The audit shows a meaningful backlog, especially for imports, __main__ guards, and broad exception handling. I audited
  the clean unit-tests-style branch at 38b34b159.

  Scope: 368 tracked Python files, with 350 repository-owned files analyzed and 18 files under third-party directories
  excluded. All 350 parsed successfully.

   Rule                                        Audit result    Confidence
[...]

PR #7335, opened 2026-08-13
Fix/skip unit tests that were failing locally on Windows

ScottTodd:
## Motivation

Related issues:
* https://github.com/ROCm/TheRock/issues/6930
* https://github.com/ROCm/TheRock/issues/6927

## Technical Details

Tests should isolate themselves from the environment
* Environment variables set by GitHub Actions like `GITHUB_REPOSITORY`
* Local git configuration settings (such as requiring commit signing for `git commit` operations)
* If a test only runs when certain conditions are met (such as having submodules initialized), it should take care that it passes outside of CI on a variety of systems, since CI might not run those tests

Tests should also cover _meaningful_ logic, not just commands ourside of our control like `git rev-parse --short`.

## Test Plan

```
pytest build_tools/github_actions/tests/ build_tools/tests/ -vv --durations=20
```

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/tests/stage_reuse_decision_test.py
+++ b/build_tools/github_actions/tests/stage_reuse_decision_test.py
@@ -105,22 +105,18 @@ def _selector(baseline):
 
 class ModeParsingTest(unittest.TestCase):
     def test_default_is_dry_run(self):
-        import os
-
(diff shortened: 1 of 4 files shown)


geomin12 (approved):
thanks for all the cleanup!

PR #7343, opened 2026-08-13
feat(fetch_sources): authenticate GitHub clones via GITHUB_CLONE_TOKEN

### LINKED ISSUE #7170: [Issue] CI: `Fetch sources` exceeds its 30-minute timeout cloning `compiler/amd-llvm` on both Linux and Windows runners
## Summary

The `Fetch sources` step in rocm-libraries CI regularly exceeds its 30-minute timeout while cloning
the `compiler/amd-llvm` submodule (`ROCm/llvm-project`). This affects **both Linux and Windows**
runner pools and blocks unrelated PRs with red builds that never reach CMake configure.

It is intermittent, not a hard break: on a good run the same clone completes in about **2.5 minutes**.
That points at throughput variance rather than a repo or configuration problem.

This is filed separately from #7129 because #7129 is scoped to the
`linux-gfx942-1gpu-ccs-csp-ossci-rocm` pool, whereas the failures below are on
`aws-linux-scale-rocm-prod` and `aws-windows-scale-rocm-prod`.

## The failure has two very different-looking error messages

This is the main reason it has been hard to spot as one issue.

**Windows** — explicit and obvious:

```text
Cloning into 'C:/home/runner/_work/rocm-libraries/rocm-libraries/TheRock/compiler/amd-llvm'...
##[error]The action 'Fetch sources' has timed out after 30 minutes.
```

**Linux** — misleading, looks like a runner-administrator problem:

```text
Cloning into '/__w/rocm-libraries/rocm-libraries/TheRock/compiler/amd-llvm'...
##[error]Executing the custom container implementation failed. Please contact your self hosted runner administrator.
```

On Linux the clone timeout kills the container step, and the container harness reports its own generic
error instead of the timeout. In both cases the timestamps show exactly 30 minutes between
`Cloning into ...amd-llvm` and the error, and no build code runs.

A common follow-on symptom is a confusing cascade in the next step, e.g.
`du: cannot access 'TheRock/build/dist/rocm': No such file or directory` or
`ls: cannot access 'B:\build'`, because nothing was ever built.

## Occurrences

All from `ROCm/rocm-libraries`. "Elapsed" is `Cloning into ...amd-llvm` to the error.

| Date (UTC) | Lane | Platform / pool | Job | Elapsed |
| --- | --- | --- | --- | --- |
| 2026-08-06 15:49 | `Windows (…\| gfx1151) / Build (gfx1151)` | `aws-windows-scale-rocm-prod` | [92641848402](https://github.com/ROCm/rocm-libraries/actions/runs/31108996094/job/92641848402) | 30m |
| 2026-08-05 21:59 | `Windows (hipkernelprovider \| gfx1151) / Build` | `aws-windows-scale-rocm-prod` | [92443569378](https://github.com/ROCm/rocm-libraries/actions/runs/31046423180/job/92443569378) | 30m |
| 2026-08-05 22:00 | `Windows (rocthrust,hipcub,rocprim \| gfx1151) / Build` | `aws-windows-scale-rocm-prod` | [92443569428](https://github.com/ROCm/rocm-libraries/actions/runs/31046423180/job/92443569428) | 30m |
| 2026-08-04 19:07 | `Windows (…\| gfx1151) / Build (gfx1151)` | `aws-windows-scale-rocm-prod` | [92091848309](https://github.com/ROCm/rocm-libraries/actions/runs/30938782012/job/92091848309) | 30m |
| 2026-08-03 16:51 | `Windows (…\| gfx1151) / Build (gfx1151)` | `aws-windows-scale-rocm-prod` | [91745210729](https://github.com/ROCm/rocm-libraries/actions/runs/30831054264/job/91745210729) | 30m |
| 2026-08-03 16:48 | `Linux (…\| gfx950-dcgpu) / Build` | `aws-linux-scale-rocm-prod` | [91745210694](https://github.com/ROCm/rocm-libraries/actions/runs/30831054264/job/91745210694) | 30m |
| 2026-08-03 16:48 | `Linux (…\| gfx125X-dcgpu) / Build` | `aws-linux-scale-rocm-prod` | [91745210911](https://github.com/ROCm/rocm-libraries/actions/runs/30831054264/job/91745210911) | 30m |
| 2026-08-03 15:58 | `Build (hipBLASLt \| gfx90a \| HOST_ASAN)` | `aws-linux-scale-rocm-prod` | [91731959997](https://github.com/ROCm/rocm-libraries/actions/runs/30827210051/job/91731959997) | 30m |

These span at least four unrelated PRs and multiple unrelated component groups
(`hipkernelprovider`, `rocthrust,hipcub,rocprim`, `hipblaslt,rocblas,…`), which rules out any
individual PR's code as the cause.

### Counter-example showing it is intermittent

In run [30827216696](https://github.com/ROCm/rocm-libraries/actions/runs/30827216696), jobs
[91732002078](https://github.com/ROCm/rocm-libraries/actions/runs/30827216696/job/91732002078) and
[91732002356](https://github.com/ROCm/rocm-libraries/actions/runs/30827216696/job/91732002356)
cloned `amd-llvm` between 15:29:23 and 15:31:52 — roughly **2.5 minutes** — and moved on to
`rocm-systems` normally. Same submodule, same command, same day.

## Impact

- Unrelated PRs show multiple red build lanes and get triaged repeatedly. Two examples this week:
  [rocm-libraries#10347](https://github.com/ROCm/rocm-libraries/pull/10347) and
  [rocm-libraries#8509](https://github.com/ROCm/rocm-libraries/pull/8509).
- Because the builds die before CMake configure, the affected lanes yield **zero signal** — neither a
  pass nor a real failure. In review threads this is easily misread as "the build is broken", which
  costs reviewer and gardener time and can stall a PR that is actually fine.
- 30 minutes of runner time is burned per occurrence, and the Windows lanes are already
  capacity-constrained.

## Suggested directions

Listed roughly easiest first; the packaging/CI owners are better placed to choose.

1. **Make the failure legible.** Have the Linux path surface the real cause rather than
   `Executing the custom container implementation failed...`, and log per-submodule clone durations in
   `fetch_sources.py` so slow fetches are visible before the timeout.
2. **Reduce what is transferred.** `compiler/amd-llvm` is the dominant cost. Options include a
   shallower or partial clone (`--depth` / `--filter=blob:none`) where full history is not needed, or
   a treeless/sparse checkout — related: #4559 (sparse checkouts in `fetch_sources.py`) and #460
   (streamlining `Fetch sources`).
3. **Cache or mirror.** A warm submodule cache or a local/regional mirror on the AWS pools would
   remove the dependency on WAN throughput to `github.com` for the largest repo.
4. **Retry rather than fail.** A bounded retry on submodule clone would convert many of these into a
   slow-but-green run.
5. **Revisit the 30-minute budget** only as a stopgap — raising it hides the throughput problem and
   burns more runner time.

## Notes

- Related: #7129 (network degradation on `linux-gfx942-1gpu-ccs-csp-ossci-rocm`, internal
  OSSCI-2850). Same class of symptom, different runner pools — worth checking whether both trace back
  to the same network path.
- Related: #4559, #460, #664.
- Filed while gardening `rocm-libraries`. Happy to gather more log samples if that helps narrow down
  whether this correlates with time of day or a specific subnet/AZ.

kailash-khalasi:
## Summary

Addresses intermittent `Fetch sources` timeouts caused by GitHub server-side throttling of anonymous git clone traffic from many concurrent CI runners (#7170, ROCm/TheRock-Infra#602).

When hundreds of runners simultaneously clone large repos like `compiler/amd-llvm` (ROCm/llvm-project), GitHub rate-limits unauthenticated requests, causing clones that normally take 2.5 minutes to stall past the 30-minute timeout.

**Solution:** Use a GitHub App installation token to authenticate all `fetch_sources.py` git clone operations. Authenticated requests get significantly higher rate limits and dedicated throughput from GitHub.

### `build_tools/fetch_sources.py`
- Added `_configure_github_clone_token()`: reads `GITHUB_CLONE_TOKEN` env var and configures `git url.insteadOf` to inject the token for all `https://github.com/` clones
- Called at the top of `main()` — no-ops silently if env var is not set (safe for local dev and forks)

### Workflows (4 files)
Added a `Generate clone token` step immediately before `Fetch sources` in every workflow that calls `fetch_sources.py`:
- `multi_arch_build_portable_linux_artifacts.yml`
- `multi_arch_build_windows_artifacts.yml`
- `multi_arch_build_wsl_rocdxg_artifacts.yml`
- `test_consumer_graph_drift.yml`

Uses `actions/create-github-app-token` with two org secrets:
- `ROCM_CLONE_APP_ID`
- `ROCM_CLONE_APP_PRIVATE_KEY`

### Tests
- Added `ConfigureGithubCloneTokenTest` to `build_tools/tests/fetch_sources_mirror_test.py`
  - Token set → git config called with correct args
  - Token not set → no-op
  - Token empty string → no-op

## Prerequisites

A GitHub App must be created and installed on the ROCm org with:
- **Permission:** `Contents: Read` on all repositories
- **Org secrets set:** `ROCM_CLONE_APP_ID`, `ROCM_CLONE_APP_PRIVATE_KEY`

## Test plan
- [ ] GitHub App created with `Contents: Read` org-wide
- [ ] `ROCM_CLONE_APP_ID` and `ROCM_CLONE_APP_PRIVATE_KEY` set as ROCm org secrets
- [ ] CI run on this PR shows `Generate clone token` step succeeding
- [ ] `Fetch sources` completes consistently without throttling delays
- [ ] Local `fetch_sources.py` runs (without `GITHUB_CLONE_TOKEN` set) are unaffected

Closes #7170

### DIFF
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -153,8 +153,18 @@ jobs:
             --output-dir="${BUILD_DIR}" \
             --bootstrap
 
+      - name: Generate clone token
+        id: clone-token
+        uses: actions/create-github-app-token@df432ceeac20ab8f7f8cc7a0e70e0c83f3eb6ff3 # v2.0.6
+        with:
+          app-id: ${{ secrets.ROCM_CLONE_APP_ID }}
+          private-key: ${{ secrets.ROCM_CLONE_APP_PRIVATE_KEY }}
+          owner: ROCm
+
       - name: Fetch sources
         timeout-minutes: 30
+        env:
+          GITHUB_CLONE_TOKEN: ${{ steps.clone-token.outputs.token }}
         run: ./build_tools/fetch_sources.py --stage ${STAGE_NAME} --jobs 12 --depth 1 ${EXTERNAL_FETCH_SOURCES_ARGS}
 
       - name: Get stage configuration
--- a/.github/workflows/multi_arch_build_windows_artifacts.yml
+++ b/.github/workflows/multi_arch_build_windows_artifacts.yml
@@ -181,8 +181,18 @@ jobs:
             --output-dir="${BUILD_DIR}" \
             --bootstrap
 
+      - name: Generate clone token
+        id: clone-token
+        uses: actions/create-github-app-token@df432ceeac20ab8f7f8cc7a0e70e0c83f3eb6ff3 # v2.0.6
+        with:
+          app-id: ${{ secrets.ROCM_CLONE_APP_ID }}
+          private-key: ${{ secrets.ROCM_CLONE_APP_PRIVATE_KEY }}
+          owner: ROCm
+
       - name: Fetch sources
         timeout-minutes: 30
+        env:
+          GITHUB_CLONE_TOKEN: ${{ steps.clone-token.outputs.token }}
         run: python build_tools/fetch_sources.py --stage ${STAGE_NAME} --jobs 12 --depth 1 ${EXTERNAL_FETCH_SOURCES_ARGS}
 
       - name: Get stage configuration
--- a/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
+++ b/.github/workflows/multi_arch_build_wsl_rocdxg_artifacts.yml
@@ -151,8 +151,18 @@ jobs:
         with:
           version: "3.62.0"
 
+      - name: Generate clone token
+        id: clone-token
+        uses: actions/create-github-app-token@df432ceeac20ab8f7f8cc7a0e70e0c83f3eb6ff3 # v2.0.6
+        with:
+          app-id: ${{ secrets.ROCM_CLONE_APP_ID }}
+          private-key: ${{ secrets.ROCM_CLONE_APP_PRIVATE_KEY }}
+          owner: ROCm
+
       - name: Fetch sources
         timeout-minutes: 30
+        env:
+          GITHUB_CLONE_TOKEN: ${{ steps.clone-token.outputs.token }}
         run: python build_tools/fetch_sources.py --stage ${STAGE_NAME} --jobs 12 --depth 1 ${EXTERNAL_FETCH_SOURCES_ARGS}
 
       - name: Set up WSL with Ubuntu
--- a/.github/workflows/test_consumer_graph_drift.yml
+++ b/.github/workflows/test_consumer_graph_drift.yml
@@ -36,8 +36,18 @@ jobs:
       - name: Install Python deps
         run: pip install -r requirements.txt
 
+      - name: Generate clone token
+        id: clone-token
+        uses: actions/create-github-app-token@df432ceeac20ab8f7f8cc7a0e70e0c83f3eb6ff3 # v2.0.6
+        with:
+          app-id: ${{ secrets.ROCM_CLONE_APP_ID }}
+          private-key: ${{ secrets.ROCM_CLONE_APP_PRIVATE_KEY }}
+          owner: ROCm
+
       - name: Fetch sources
         # All subproject sources must be present for the graph to be complete.
+        env:
+          GITHUB_CLONE_TOKEN: ${{ steps.clone-token.outputs.token }}
         run: python3 ./build_tools/fetch_sources.py --jobs 12 --depth 1
 
       - name: Configure (emits the consumer graph)
--- a/build_tools/fetch_sources.py
+++ b/build_tools/fetch_sources.py
@@ -45,6 +45,30 @@ def is_windows() -> bool:
     return platform.system() == "Windows"
 
 
+def _configure_github_clone_token() -> None:
+    """Configure git to authenticate GitHub clones using GITHUB_CLONE_TOKEN.
+
+    When set, injects the token into all https://github.com/ URLs via
+    git's url.insteadOf mechanism. This allows fetch_sources.py to clone
+    private and rate-limited ROCm org repos without hitting anonymous
+    GitHub API throttling limits.
+    """
+    token = os.environ.get("GITHUB_CLONE_TOKEN")
+    if not token:
+        return
+    subprocess.run(
+        [
+            "git",
+            "config",
+            "--global",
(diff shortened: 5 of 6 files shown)


.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -153,8 +153,18 @@ jobs:
             --output-dir="${BUILD_DIR}" \
             --bootstrap
 
+      - name: Generate clone token
+        id: clone-token
+        uses: actions/create-github-app-token@df432ceeac20ab8f7f8cc7a0e70e0c83f3eb6ff3 # v2.0.6
+        with:
+          app-id: ${{ secrets.ROCM_CLONE_APP_ID }}
+          private-key: ${{ secrets.ROCM_CLONE_APP_PRIVATE_KEY }}
+          owner: ROCm
```
ScottTodd, line 162: > A GitHub App must be created and installed on the ROCm org with:
> * **Permission:** `Contents: Read` on all repositories

No. Apps must be limited to only the minimal permissions/scopes. We will not create an app that grants read access to all repositories.

ScottTodd, line 162: Why is GITHUB_TOKEN insufficient? What does a custom token grant instead?

ScottTodd (commented):
This doesn't look right.

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kailash-khalasi: can we have a static list of repos to allow the app to read?

ScottTodd: Let's first confirm that an app is needed at all. All repositories used here are public, so we don't need privileged credentials to read from them.

kailash-khalasi: i will circle back

kailash-khalasi: GITHUB_TOKEN should suffice. In my previous test, it failed but it might have been a different issue and not permissions related. let me get back to testing GITHUB_TOKEN and validating that.

PR #7347, opened 2026-08-13
Float Stable Torch ROCm Dependency Pins

### LINKED ISSUE #7345: Pin ROCm `X.Y.*` in stable torch packages
So far all torch wheels strictly pinned the corresponding ROCm dependency. While this is desirable for dev and nightly builds as well as for prereleases, it isn't strictly for stable releases. For stable releases only the major and minor version must be pinned while the patch version should be floating. This allows to upgrading to a ROCm patch release (e.g. from 7.14.0 to 7.14.1) while not needing to upgrade the torch package.

marbre:
## Motivation

Torch wheels currently strictly pin their corresponding ROCm dependencies. This is desirable for development and nightly builds as well as prereleases, but is unnecessarily restrictive for stable releases.

Stable Torch wheels should be able to use later ROCm patch releases on the same major and minor version. For example, a wheel built against ROCm 7.14.0 should allow upgrading to ROCm 7.14.1 without requiring a new Torch package, while a wheel built against ROCm 7.14.1 should not allow downgrading to ROCm 7.14.0.

Closes #7345

## Technical Details

Stable torch wheels keep their exact `+rocmX.Y.Z` artifact identity while their ROCm package dependencies accept later patch releases on the same minor line.

Existing exact stable ROCm `Requires-Dist` pins in Torch-family metadata are changed from `==X.Y.Z` to the PEP 440 compatible-release specifier `~=X.Y.Z`. This is equivalent to `>=X.Y.Z, ==X.Y.*`, allowing later patches without permitting a downgrade below the ROCm version against which the wheel was built.

ROCm package wheels, prerelease promotions, compound specifiers, `rocm-bootstrap`, and Torch-to-Torch dependencies remain exact.
 
The generated `_rocm_init.py` runtime check uses `X.Y.*`. The runtime API supports glob patterns rather than PEP 440 specifiers, so it verifies same-minor compatibility while the package metadata enforces the forward-only lower bound.

## Test Plan

- python build_tools/packaging/tests/promote_keep_list_test.py
- python -m py_compile build_tools/packaging/promote_packages.py \ build_tools/packaging/tests/promote_keep_list_test.py

## Test Result

- Passed: promote_keep_list_test.py (34 tests)
- Passed: py_compile

## Submission Checklist

Generated with Codex

### DIFF
--- a/build_tools/packaging/promote_packages.py
+++ b/build_tools/packaging/promote_packages.py
@@ -348,6 +348,31 @@ def rewrite_metadata_rocm_line(
     return requires_match.group("prefix") + requires_match.group("name") + tail
 
 
+_CHECK_VERSION_RE = re.compile(
+    r"(?P<prefix>\bcheck_version\s*=\s*['\"])(?P<version>[^'\"]+)(?P<suffix>['\"])"
+)
+
+
+def rewrite_rocm_init_check_version(
+    line: str, old_rocm_version: str, new_rocm_version: str
+) -> str:
+    """Rewrite torch _rocm_init.py stable checks to ROCm major/minor wildcard."""
+    rewritten = line.replace(old_rocm_version, new_rocm_version)
+    check_version_match = _CHECK_VERSION_RE.search(rewritten)
+    if check_version_match is None:
+        return rewritten
+    if check_version_match.group("version") != new_rocm_version:
+        return rewritten
+    minor_spec = stable_minor_spec(new_rocm_version)
+    if minor_spec is None:
+        return rewritten
+    return (
+        rewritten[: check_version_match.start("version")]
+        + minor_spec
+        + rewritten[check_version_match.end("version") :]
+    )
+
+
 def update_metadata_rocm_requires_dist(
     new_dir_path: pathlib.Path,
     package_name_no_version: str,
@@ -1009,6 +1034,19 @@ def wheel_change_extra_files(
     with fileinput.input(files=files_to_change, encoding="utf-8", inplace=True) as f:
         for line in f:
             print(line.replace(old_rocm_version, new_rocm_version), end="")
+
+    if float_rocm_dependency_patch and "torch" == package_name_no_version:
+        rocm_init_path = new_dir_path / package_name_no_version / "_rocm_init.py"
+        with fileinput.input(
+            files=[rocm_init_path], encoding="utf-8", inplace=True
+        ) as f:
+            for line in f:
+                print(
+                    rewrite_rocm_init_check_version(
+                        line, old_rocm_version, new_rocm_version
+                    ),
+                    end="",
+                )
 
     print("    ...done")
 
--- a/build_tools/packaging/promote_packages.py
+++ b/build_tools/packaging/promote_packages.py
@@ -347,7 +347,8 @@ def rewrite_metadata_rocm_line(
         tail,
         count=1,
     )
-    return requires_match.group("prefix") + requires_match.group("name") + tail
+    tail_start, tail_end = requires_match.span("tail")
+    return rewritten[:tail_start] + tail + rewritten[tail_end:]
 
 
 def rewrite_rocm_init_check_version(
@@ -369,7 +370,7 @@ def update_metadata_rocm_requires_dist(
     *,
     float_rocm_dependency_patch: bool = False,
 ) -> None:
-    """Update Requires-Dist lines in METADATA that reference rocm, leaving others unchanged."""
+    """Update ROCm references in METADATA Summary and Requires-Dist fields."""
     metadata_path = (
         new_dir_path / f"{package_name_no_version}-{old_version}.dist-info" / "METADATA"
     )
@@ -394,7 +395,7 @@ def update_metadata_rocm_requires_dist(
                         ),
                         end="",
                     )
-                elif line.startswith("Requires-Dist"):
+                elif line.startswith("Requires-Dist") and "rocm" in lower_line:
                     print(
                         rewrite_metadata_rocm_line(
                             line,
@@ -981,12 +982,13 @@ def wheel_change_extra_files(
     if JAX_ROCM_PACKAGE_PATTERN.match(package_name_no_version):
         return
 
+    files_to_float: list[pathlib.Path] = []
+
     # rocm packages needing extra handling
     if new_dir_path.name.startswith("rocm"):
         files_to_change = [
             new_dir_path / package_name_no_version / "_dist_info.py",
         ]
-        files_to_float = []
 
         if new_dir_path.name.startswith("rocm_sdk_core"):
             files_to_change.append(
@@ -1004,14 +1006,10 @@ def wheel_change_extra_files(
         files_to_float = [
             new_dir_path / package_name_no_version / "_rocm_init.py",
         ]
-        if not float_rocm_dependency_patch:
-            files_to_change.extend(files_to_float)
-            files_to_float = []
     elif "apex" in package_name_no_version:
         files_to_change = [
             new_dir_path / package_name_no_version / "git_version_info_installed.py",
         ]
-        files_to_float = []
     else:
         # we have multiple packages that have a version.py that needs updating
         need_change_version_py = ["torchaudio", "torchvision", "jaxlib"]
@@ -1019,7 +1017,6 @@ def wheel_change_extra_files(
             files_to_change = [
                 new_dir_path / package_name_no_version / "version.py",
             ]
-            files_to_float = []
         else:
             # no additional (rocm-specific) files needed to be changed that contain the version
             # currently applying to: triton
@@ -1028,12 +1025,16 @@ def wheel_change_extra_files(
     def rewrite_exact(line: str) -> str:
         return line.replace(old_rocm_version, new_rocm_version)
 
-    def rewrite_rocm_init(line: str) -> str:
+    def rewrite_float_version(line: str) -> str:
         return rewrite_rocm_init_check_version(line, old_rocm_version, new_rocm_version)
+
+    if not float_rocm_dependency_patch:
+        files_to_change.extend(files_to_float)
+        files_to_float = []
 
     rewrite_groups = [
         (files_to_change, rewrite_exact),
-        (files_to_float, rewrite_rocm_init),
+        (files_to_float, rewrite_float_version),
     ]
     for files, rewrite_line in rewrite_groups:
         if not files:
--- a/build_tools/packaging/promote_packages.py
+++ b/build_tools/packaging/promote_packages.py
@@ -24,11 +24,11 @@ Common workflows:
 Other combinations are mechanically possible but not typical.
 
 When promoting torch-family wheels to a stable release, ROCm package
-dependencies in METADATA are rewritten to a patch-floating stable minor
-specifier (for example `rocm-sdk-core==7.13.*`). The torch wheel identity itself
-remains exact (for example `torch-2.8.0+rocm7.13.0-...whl`) so the artifact
-records the ROCm version it was built against, while allowing ROCm patch
-updates that do not require rebuilding torch.
+dependencies in METADATA are rewritten to a forward patch-compatible specifier
+(for example `rocm-sdk-core~=7.13.0`). The torch wheel identity itself remains
+exact (for example `torch-2.8.0+rocm7.13.0-...whl`) so the artifact records the
+ROCm version it was built against, while allowing later ROCm patch updates that
+do not require rebuilding torch.
 
 The keep-list pass (--multi-arch-targets) is independent of version promotion: it
 runs whenever --multi-arch-targets is supplied, alongside any version rewrite. To
@@ -291,6 +291,18 @@ def stable_minor_spec(version_str: str) -> str | None:
     return f"{version.release[0]}.{version.release[1]}.*"
 
 
+def stable_patch_compatible_spec(version_str: str) -> str | None:
+    """Return a forward patch-compatible specifier like `~=7.13.1`, or None."""
+    version = Version(version_str)
+    if version.is_prerelease or version.is_devrelease or version.local is not None:
+        return None
+    if len(version.release) < 3:
+        raise ValueError(
+            f"Stable version {version_str!r} must include major, minor, and patch"
+        )
+    return f"~={version}"
+
+
 _REQUIRES_DIST_NAME_RE = re.compile(
     r"^(?P<prefix>Requires-Dist:\s*)(?P<name>[A-Za-z0-9_.-]+)(?P<tail>.*)$"
 )
@@ -307,7 +319,9 @@ def rewrite_metadata_rocm_line(
 
     Summary lines and non-floating requirements use exact replacement. When
     `float_rocm_dependency_patch` is set, only existing exact `==<stable version>`
-    pins on ROCm dependency names are converted to `==X.Y.*`.
+    pins on ROCm dependency names are converted to `~=X.Y.Z`. This allows later
+    patches on the same minor line without permitting a downgrade below the
+    version against which the wheel was built.
     """
     rewritten = line.replace(old_rocm_version, new_rocm_version)
     if not float_rocm_dependency_patch or not rewritten.startswith("Requires-Dist"):
@@ -334,16 +348,17 @@ def rewrite_metadata_rocm_line(
 
     exact_version = specifier.version
     try:
-        minor_spec = stable_minor_spec(exact_version)
+        compatible_spec = stable_patch_compatible_spec(exact_version)
     except InvalidVersion:
         return rewritten
-    if minor_spec is None:
+    if compatible_spec is None:
         return rewritten
 
+    compatible_version = compatible_spec.removeprefix("~=")
     tail = requires_match.group("tail")
     tail = re.sub(
-        rf"(?P<operator>==\s*){re.escape(exact_version)}(?P<boundary>\b)",
-        rf"\g<operator>{minor_spec}\g<boundary>",
+        rf"==(?P<space>\s*){re.escape(exact_version)}(?P<boundary>\b)",
+        rf"~=\g<space>{compatible_version}\g<boundary>",
         tail,
         count=1,
     )
@@ -982,6 +997,8 @@ def wheel_change_extra_files(
     if JAX_ROCM_PACKAGE_PATTERN.match(package_name_no_version):
         return
 
+    # If more files need specialized rewrites, replace the parallel file groups
+    # below with (path, rewrite function) pairs so the dispatch stays explicit.
     files_to_float: list[pathlib.Path] = []
 
     # rocm packages needing extra handling
--- a/build_tools/packaging/promote_packages.py
+++ b/build_tools/packaging/promote_packages.py
@@ -23,6 +23,13 @@ Common workflows:
   - a  -> rc                 e.g. 7.13.0a20260501 -> 7.13.0rc1
 Other combinations are mechanically possible but not typical.
 
+When promoting torch-family wheels to a stable release, ROCm package
+dependencies in METADATA are rewritten to a patch-floating stable minor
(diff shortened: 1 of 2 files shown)


build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,114 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+    new_rocm_version: str,
+    *,
+    float_rocm_dependency_patch: bool,
+) -> str:
+    """Rewrite one METADATA line for ROCm version promotion.
+
+    Summary lines and non-floating requirements use exact replacement. When
+    `float_rocm_dependency_patch` is set, only existing exact `==<stable version>`
+    pins on ROCm dependency names are converted to `==X.Y.*`.
+    """
```
ScottTodd, line 325: Do we also need to update `check_version=` here? https://github.com/ROCm/TheRock/blob/400a95ded0b68acfc141f6dd039b99ace0e6af10/external-builds/pytorch/build_prod_wheels.py#L414-L429

See https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md#performing-initialization

> Note that you must preload any libraries that your native extensions depend on
so that when the operating system attempts to resolve linkage, they are already
in the namespace. The above is an example that, at the time of writing, was
suitable for PyTorch.
>
> **Note that it version locks to a specific ROCm SDK version.
This is generally appropriate for development/nightly builds. For production
builds, you will want to lock to a major/minor version only and use a wildcard
(\*) to match the suffix.**

marbre: Ohhhh, good fetch! Taking a look.

marbre: Addressed.

build_tools/packaging/promote_packages.py
```diff
@@ -261,9 +397,25 @@ def update_metadata_rocm_requires_dist(
                 if line.startswith("Summary:") and (
                     "TheRock" in line or "rocm" in line
                 ):
-                    print(line.replace(old_rocm_version, new_rocm_version), end="")
-                elif line.startswith("Requires-Dist") and "rocm" in line:
-                    print(line.replace(old_rocm_version, new_rocm_version), end="")
+                    print(
```
HereThereBeDragons, line 404: I dont think we want to replace this here. this is only the summary section of metdata (== title). And this was/is only relevant for jax. and as you saw in the offline discussion we now need to look for "ROCm". 

maybe we can do a case insensitive matching?

build_tools/packaging/promote_packages.py
```diff
@@ -878,6 +1035,19 @@ def wheel_change_extra_files(
         for line in f:
             print(line.replace(old_rocm_version, new_rocm_version), end="")
 
+    if float_rocm_dependency_patch and "torch" == package_name_no_version:
```
HereThereBeDragons: can we rearrange this a bit?

this breaks the current design style and has a bug

style-wise we first create the list of files and then have the for loop just above (line 1032-36) with `print(line.replace(old_rocm_version, new_rocm_version), end="")`

_rocm_init is also only relevant for "torch", which we are already overwriting above (line 1013). so your code here should not work how you expect it to work.

can we either introduce a flag or a new list "files_to_float" that keeps the old style of first creating a list for relevant files and then changing them? (and if float_rocm_dependency_patch == false just merge the 2 lists)

build_tools/packaging/promote_packages.py
```diff
@@ -779,6 +931,7 @@ def wheel_change_extra_files(
     old_version: Version,
     new_version: Version,
     multi_arch_targets: list[str] | None = None,
+    dest_version: str = "release",
```
HereThereBeDragons: i would rather have no default here.

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
     return args
 
 
+def is_torch_package_name(package_name: str) -> bool:
```
HereThereBeDragons, line 251: why not simplify it?
if "torch" in package_name or "triton" .. or "apex"?

please also note that triton and apex have no required-dist. they use "dynamic".
see also my comment before calling update_metadata_rocm_requires_dist()

```
    # Every wheel: rewrite METADATA Summary / Requires-Dist rocm refs.
    # No-op when no rocm refs are present (e.g. triton, apex)
    update_metadata_rocm_requires_dist(
```

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+            "amd-torch-device-",
+            "amd-torchvision-device-",
+        )
+    )
+
+
+def is_floating_rocm_dependency_name(package_name: str) -> bool:
+    """Return True for ROCm package deps that may float in torch metadata.
+
+    `rocm-bootstrap` intentionally stays out of this set. It is not part of the
```
HereThereBeDragons, line 273: good to leave a comment but we are already skipping rocm-bootstrap completely from the promotion process in main()

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+_CHECK_VERSION_RE = re.compile(
+    r"(?P<prefix>\bcheck_version\s*=\s*['\"])(?P<version>[^'\"]+)(?P<suffix>['\"])"
+)
+
+
+def rewrite_rocm_init_check_version(
+    line: str, old_rocm_version: str, new_rocm_version: str
+) -> str:
+    """Rewrite torch _rocm_init.py stable checks to ROCm major/minor wildcard."""
+    rewritten = line.replace(old_rocm_version, new_rocm_version)
```
HereThereBeDragons: the replacement has already happened before hand. see my other comment.


maybe i am a bit insensitive but there is no other content in rocm_init other than the one rocm version in 

```
    rocm_sdk.initialize_process(
        preload_shortnames=['amd_comgr', 'amdhip64', 'rocprofiler-sdk', 'rocprofiler-sdk-roctx', 'roctracer64', 'roctx64', 'hiprtc', 'hipblas', 'hipfft', 'hiprand', 'hipsparse', 'hipsparselt', 'hipsolver', 'rccl', 'hipblaslt', 'miopen', 'hipdnn', 'rocm_sysdeps_liblzma', 'rocm-openblas', 'rocm_smi64'],
        check_version='10.0.0rc0')

```

so why all the extra checking and not just calculating the new floating version and do the line.replace(old_rocm_version, floating_rocm_version)?

you should be able to just do something a la:

```
minor_spec = stable_minor_spec(new_rocm_version)
...
for line in f:
    print(line.replace(old_rocm_version, minor_spec), end="")
```

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+def rewrite_metadata_rocm_line(
+    line: str,
+    old_rocm_version: str,
+    new_rocm_version: str,
+    *,
+    float_rocm_dependency_patch: bool,
+) -> str:
+    """Rewrite one METADATA line for ROCm version promotion.
+
+    Summary lines and non-floating requirements use exact replacement. When
```
HereThereBeDragons, line 320: might be a bit misunderstood here. summary is taken care of somewhere else

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+    `float_rocm_dependency_patch` is set, only existing exact `==<stable version>`
+    pins on ROCm dependency names are converted to `==X.Y.*`.
+    """
+    rewritten = line.replace(old_rocm_version, new_rocm_version)
+    if not float_rocm_dependency_patch or not rewritten.startswith("Requires-Dist"):
+        return rewritten
+
+    requires_match = _REQUIRES_DIST_NAME_RE.match(rewritten)
```
HereThereBeDragons, line 330: as we are opening a wheel, we must expect that requires-dist is written correctly. dont think this check is needed

build_tools/packaging/tests/promote_keep_list_test.py
```diff
@@ -210,6 +238,141 @@ def test_empty_intersection_raises(self):
...
+            .replace(
+                "Requires-Dist: rocm-bootstrap\n",
+                "Requires-Dist: rocm-bootstrap\n"
+                "Requires-Dist: rocm-sdk-core==7.13.0a20260505\n",
+                1,
+            )
+            .replace(
+                "Requires-Dist: rocm-sdk-core==7.13.0a20260505\n",
+                "Requires-Dist: rocm-sdk-core==7.13.0a20260505\n"
+                'Requires-Dist: ROCm-SDK-Core == 7.13.0a20260505; extra == "core"\n',
```
HereThereBeDragons, line 256: quotes? " vs '

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,139 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+
+_CHECK_VERSION_RE = re.compile(
+    r"(?P<prefix>\bcheck_version\s*=\s*['\"])(?P<version>[^'\"]+)(?P<suffix>['\"])"
+)
+
+
+def rewrite_rocm_init_check_version(
+    line: str, old_rocm_version: str, new_rocm_version: str
+) -> str:
+    """Rewrite torch _rocm_init.py stable checks to ROCm major/minor wildcard."""
```
HereThereBeDragons:
```suggestion
    """Rewrite torch _rocm_init.py stable check to a ROCm major/minor wildcard."""
```

build_tools/packaging/promote_packages.py
marbre: I think we should rather be explicit here. It might be the case that another package with torch in the name is added that needs to be handled differently, `if "torch" in package_name` would catch this.

If they don't have an required-dist this won't hurt but would you rather suggest to drop them?

build_tools/packaging/tests/promote_keep_list_test.py
marbre: `''` needed due to `"core"`.

build_tools/packaging/promote_packages.py
marbre: Not sure what offline discussion you're referring to. Furthermore, it is unclear to me why this should only be relevant for JAX.

HereThereBeDragons: because jax is the only packages that has the rocm version in the summary line
jax has like `Summary: JAX Plugin for AMD GPUs (ROCm:unknown)`

before they had `TheRock` in it

build_tools/packaging/promote_packages.py
HereThereBeDragons: ok, but then it would be nice to add a comment that apex and triton currently dont have any specifically named dependencies for rocm

JeniferC99 (approved):

build_tools/packaging/promote_packages.py
```diff
@@ -841,6 +986,7 @@ def wheel_change_extra_files(
         files_to_change = [
             new_dir_path / package_name_no_version / "_dist_info.py",
         ]
+        files_to_float = []
```
HereThereBeDragons: could consider moving this up above "# rocm packages needing extra handling" and then just overwriting it whereever it is needed.

build_tools/packaging/promote_packages.py
```diff
@@ -853,30 +999,50 @@ def wheel_change_extra_files(
     # only torch and NOT triton, torchaudio, torchvision
     elif "torch" == package_name_no_version:
         files_to_change = [
-            new_dir_path / package_name_no_version / "_rocm_init.py",
             new_dir_path / package_name_no_version / "version.py",
         ]
+        files_to_float = [
+            new_dir_path / package_name_no_version / "_rocm_init.py",
+        ]
+        if not float_rocm_dependency_patch:
```
HereThereBeDragons: this "if" we could also put down near "rewrite_groups" - as it is generically applied to files_to_float indepedent in which branch it is set

build_tools/packaging/promote_packages.py
```diff
@@ -853,30 +999,50 @@ def wheel_change_extra_files(
...
 
-    for f in files_to_change:
-        print(f"      {f}")
-    with fileinput.input(files=files_to_change, encoding="utf-8", inplace=True) as f:
-        for line in f:
-            print(line.replace(old_rocm_version, new_rocm_version), end="")
+    def rewrite_exact(line: str) -> str:
+        return line.replace(old_rocm_version, new_rocm_version)
+
+    def rewrite_rocm_init(line: str) -> str:
```
HereThereBeDragons: think would be better to rename it to something like "rewrite_float_version"

build_tools/packaging/promote_packages.py
```diff
@@ -853,30 +999,50 @@ def wheel_change_extra_files(
...
+        return rewrite_rocm_init_check_version(line, old_rocm_version, new_rocm_version)
+
+    rewrite_groups = [
+        (files_to_change, rewrite_exact),
+        (files_to_float, rewrite_rocm_init),
```
HereThereBeDragons: we can leave it for now to get it in quickly, but probably seeing this logic here it would be good:

files_to_float could become a tuple (filename, func_for_replacing) so that dependent on the file name we can call different replacment functions.

as right now it reads a bit inconsistent: files_to_float sounds generic but the function called is just for `_rocm_init.py`

---

overall i like the current implemntation approach. it is quite elegant

build_tools/packaging/promote_packages.py
```diff
@@ -239,12 +248,126 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
 def update_metadata_rocm_requires_dist(
     new_dir_path: pathlib.Path,
     package_name_no_version: str,
     old_version: str,
     old_rocm_version: str,
     new_rocm_version: str,
+    *,
+    float_rocm_dependency_patch: bool = False,
 ) -> None:
     """Update Requires-Dist lines in METADATA that reference rocm, leaving others unchanged."""
```
HereThereBeDragons: i guess i forgot to update that docstring.. doesnt talk about the summary

build_tools/packaging/promote_packages.py
```diff
@@ -258,12 +381,29 @@ def update_metadata_rocm_requires_dist(
...
+                    print(
+                        rewrite_metadata_rocm_line(
+                            line,
+                            old_rocm_version,
+                            new_rocm_version,
+                            float_rocm_dependency_patch=False,
+                        ),
+                        end="",
+                    )
+                elif line.startswith("Requires-Dist"):
```
HereThereBeDragons: i think we should put `and "rocm" in  line` back. 
we should only be interested in requires-dist that have somewhere "rocm" stated. and especially also because `rewrite_metadata_rocm_line` doesnt look for "rocm" in the line before doing the version replacement (safes work and also makes a bit more secure)

e.g. torch has

```
Requires-Dist: rocm[libraries]==10.0.0rc0
Requires-Dist: triton==3.8.0+git4cff872c.rocm10.0.0rc0
Requires-Dist: rocm-bootstrap
Provides-Extra: device-gfx1010
Requires-Dist: amd-torch-device-gfx1010 == 2.13.0+rocm10.0.0rc0; extra == "device-gfx1010"
Requires-Dist: amd-torch-device-gfx1010 == 2.13.0+rocm10.0.0rc0; extra == "device-all"
```

build_tools/packaging/promote_packages.py
marbre: I will defer the tuple abstraction as it adds structure for now as other replacement functions for more rewrite do not exist yet. Please file an issue if think it's worth to cover in a follow up.

build_tools/packaging/promote_packages.py
marbre: Okay, I think this is now resolved with `casefold()`.

build_tools/packaging/promote_packages.py
marbre: Addressed by `files_to_change/files_to_float`.

build_tools/packaging/promote_packages.py
marbre: Done.

build_tools/packaging/promote_packages.py
marbre: Addressed.

build_tools/packaging/promote_packages.py
marbre: Done.

build_tools/packaging/promote_packages.py
marbre: updated

build_tools/packaging/promote_packages.py
marbre: Done.

marbre:
@HereThereBeDragons I think I addressed most of your feedback (cleanup, updated the docstring, restored case-insensitive ROCm-only requirement filtering). I deferred the tuple-based file/rewrite abstraction and retained malformed-requirement validation to preserve the fail-fast behavior.

In addition to your review feedback, my agent spotted a unnoticed newline-loss bug that could corrupt METADATA which should now be fixed and for which a test was added for.

@HereThereBeDragons, @ScottTodd PTAL.

build_tools/packaging/promote_packages.py
```diff
@@ -239,14 +248,129 @@ def parse_arguments(argv: list[str]) -> argparse.Namespace:
...
+
+
+def stable_minor_spec(version_str: str) -> str | None:
+    """Return a stable patch-floating specifier like `7.13.*`, or None."""
+    version = Version(version_str)
+    if version.is_prerelease or version.is_devrelease or version.local is not None:
+        return None
+    if len(version.release) < 2:
+        raise ValueError(f"Stable version {version_str!r} must include major and minor")
+    return f"{version.release[0]}.{version.release[1]}.*"
```
ScottTodd, line 291: I think `7.14.*` is fine, but we could also consider `~=7.14.1` if we want to set a lower bound on the patch version instead of allowing any patch version including downgrades.

In my review agent's words...

> PEP 440 already has the desired forward-only form: `~=7.14.1` means
> `>=7.14.1,==7.14.*`. An explicit `>=7.14.1,<7.15.0` range is equivalent and
> more obvious to readers. For a wheel built against `.0`, this still permits all
> later patches; for a wheel built against `.1`, it correctly rejects `.0`.
> 
> The local check against `packaging.SpecifierSet` produced:
> 
> ```text
> ==7.14.* : 7.14.0=True, 7.14.0.post1=True, 7.14.1=True
> ~=7.14.1 : 7.14.0=False, 7.14.1=True, later 7.14 patches=True
> ```

ScottTodd (approved):
I scanned through the changes currently between 7.14.0 and 7.14.1 (https://github.com/ROCm/TheRock/compare/therock-7.14...release/therock-7.14) and this looks safe, no major breaking changes in there that would warrant restricting the compatible versions.

build_tools/packaging/promote_packages.py
HereThereBeDragons: can we add an inline comment in the file? might be easier to be found when someone touches the code than some issue

HereThereBeDragons (approved):
lgtm
other than the one comment i added to the conversation above

build_tools/packaging/promote_packages.py
marbre: Goot catch. I refactored to "pin" `~=X.Y.Z` in `Requires-Dist` but will keep `X.Y.*` in the `_rocm_init.py` check.

Want to take another look @ScottTodd?

build_tools/packaging/promote_packages.py
marbre: Done.

ScottTodd (approved):

PR #7350, opened 2026-08-13
fix(package): update rocpd & roctx python versions to include 3.14

yhuiYH:
## Motivation

In ROCm 7.14 whl and tarballs, rocpd and roctx need to support multiple python versions.  Looks like it's been limited to 3.10-3.13.  When users installed onto Ubuntu 26.04 where the default python version is 3.14, they get an error:

```rocpd --help
  File "/opt/venv/lib/python3.14/site-packages/_rocm_sdk_devel/bin/rocpd", line 74, in <module>
    main()
    ~~~~^^
  File "/opt/venv/lib/python3.14/site-packages/_rocm_sdk_devel/bin/rocpd", line 46, in main
    raise ImportError(
    ...<5 lines>...
    )
ImportError: rocpd not supported for Python version 3.14 (sys.executable='/opt/venv/bin/python3').
rocpd supported python versions: 3.10, 3.11, 3.12, 3.13
```

## Technical Details

Digging into TheRock build, it looks like the python versions used are defined in `DIST_PYTHON_EXECUTABLES`.  However, the manylinux image SHA TheRock is using is slightly old, and only has up to 3.13 in /opt/python.  

To add 3.14, we would need to:
1) Update the manylinux image SHA to something newer, and then update `DIST_PYTHON_EXECUTABLES` or
2) Just use the `/opt/python-shared/cp314-cp314/bin/python3` that rocgdb is using

Opting for 2, to minimize changes. 
We may need an additional rocprofiler-sdk change to make the python 3.14 discoverable since CMake is old in manylinux image.  https://github.com/ROCm/rocm-systems/pull/10198

Addresses ISSUE ID https://github.com/ROCm/TheRock/issues/7398

## Test Plan

Check the build artifacts to see if python3.14 is now usable with rocpd.  There was discussion to add rocpd tests to TheRock, but there are additional issues that need to be resolved, before those tests can be added.  

## Test Result

rocpd should be usable with python 3.14

## Submission Checklist

### DIFF
--- a/build_tools/github_actions/manylinux_config.py
+++ b/build_tools/github_actions/manylinux_config.py
@@ -17,7 +17,8 @@ DIST_PYTHON_EXECUTABLES = (
     "/opt/python/cp310-cp310/bin/python;"
     "/opt/python/cp311-cp311/bin/python;"
     "/opt/python/cp312-cp312/bin/python;"
-    "/opt/python/cp313-cp313/bin/python"
+    "/opt/python/cp313-cp313/bin/python;"
+    "/opt/python-shared/cp314-cp314/bin/python3"
 )
 
 # Python executables with shared libpython, for embedded Python builds (rocgdb)
--- a/build_tools/github_actions/manylinux_config.py
+++ b/build_tools/github_actions/manylinux_config.py
@@ -18,7 +18,7 @@ DIST_PYTHON_EXECUTABLES = (
     "/opt/python/cp311-cp311/bin/python;"
     "/opt/python/cp312-cp312/bin/python;"
     "/opt/python/cp313-cp313/bin/python;"
-    "/opt/python-shared/cp314-cp314/bin/python3"
+    "/opt/python/cp314-cp314/bin/python"
 )
 
 # Python executables with shared libpython, for embedded Python builds (rocgdb)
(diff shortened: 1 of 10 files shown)


build_tools/github_actions/manylinux_config.py
ScottTodd, line 26: > To add 3.14, we would need to:
> 
> 1. Update the manylinux image SHA to something newer, and then update `DIST_PYTHON_EXECUTABLES` or
> 2. Just use the `/opt/python-shared/cp314-cp314/bin/python3` that rocgdb is using
> 
> Opting for 2, to minimize changes.

I don't think we should mix `python` and `python-shared` like this.

We can update our manylinux container if the current sha doesn't include `cp314` but newer shas do:
* https://github.com/ROCm/TheRock/blob/main/dockerfiles/README.md#updating-images-used-by-github-actions-workflows
* https://github.com/ROCm/TheRock/blob/400a95ded0b68acfc141f6dd039b99ace0e6af10/dockerfiles/build_manylinux_x86_64.Dockerfile#L12

yhuiYH: So you'd prefer the manylinux SHA update?  Heard.  Will take some time to read up, and go thru the motions since it needs the .Dockerfile update, and it's a multi-step/PR process.

ScottTodd: Yeah I don't know the specific details here but `python` and `python-shared` are different for a reason and this may have impacts on packaging / portability if done as you have it now.

Issue #7349, opened 2026-08-13
Support structured artifact promotion


marbre:
Manual promotion now needs to round trip artifacts through the repo.amd.com product buckets. For the implementation of RFC0012, the scripts need to support structured package download and upload support. The local promotion directories can be kept flat. This lets the existing promotion tool remain focused on version rewriting and the product-local paths need to be only handled during down- and upload.

PR #7351, opened 2026-08-13
Support structured artifact promotion

marbre:
## Motivation

Manual promotion now needs to round trip artifacts through the repo.amd.com product buckets. To be able to promote packages to the structure coming along with RFC0012, the scripts need to support structured package download and upload support. The local promotion directories are kept flat. This lets the existing promotion tool remain focused on version rewriting and the product-local paths need to be only handled during down- and upload.

Issue #7349

## Technical Details

Core, PyTorch, and JAX artifacts are routed to product-owned buckets. ROCm Core tarballs use the new normal and ASAN prefixes through the existing tarball flow. Further adds tests for uploading packages, which didn't existed before.

## Test Plan

python -m pytest \
  build_tools/packaging/tests/download_python_packages_test.py \
  build_tools/packaging/tests/upload_release_packages_test.py

## Submission Checklist

Generated with Codex


marbre:
Haven't reviewed this myself too closely but hoped it would be a better starting point than just a GH issue. @araravik-psd, @JeniferC99 or @HereThereBeDragons feel free to take over and iterate if you think further improvements are need.

FYI, I will create a follow up issue to rename `download_python_packages_test.py` (it is handling tarballs since a while -.-) but I want to land this first.

build_tools/packaging/download_python_packages.py
```diff
@@ -79,6 +79,18 @@
   # List multi-arch packages
   python download_python_packages.py --version=7.13.0rc1 --multi-arch --list-multi-arch-packages --bucket-prefix=v4/whl/
 
+  # Download structured product-local whl-next packages.
```
HereThereBeDragons, line 90: what does product-local mean?

build_tools/packaging/download_python_packages.py
```diff
@@ -79,6 +79,18 @@
...
+  python download_python_packages.py --version=7.13.0rc1 --structured --output-dir=./downloads
+
+  # Download only structured PyTorch whl-next packages.
```
HereThereBeDragons, line 93: i think some explanation what "structured" means would be good before the commands here

build_tools/packaging/download_python_packages.py
```diff
@@ -201,12 +214,67 @@
     "setuptools",
 }
 
+STRUCTURED_PRODUCTS = ("core", "pytorch", "jax")
+STRUCTURED_INDEXES = ("whl", "whl-next")
+STRUCTURED_DEFAULT_INDEX = "whl-next"
+STRUCTURED_ROOT_PREFIX = "v5/rocm"
```
HereThereBeDragons:
```suggestion
STRUCTURED_S3_ROOT_PREFIX = "v5/rocm"
```
so people unfamiliar with the code know it is form an s3 bucket? same for the other vars

build_tools/packaging/download_python_packages.py
```diff
@@ -201,12 +214,67 @@
...
+    if not values:
+        return []
+
+    parsed = []
+    for value in values:
+        parsed.extend(item.strip() for item in value.split(",") if item.strip())
+    return parsed
+
+
+def structured_python_root(product: str, index: str) -> str:
```
HereThereBeDragons, line 279: some small comment that these functions are all s3 bucket urls?

build_tools/packaging/download_python_packages.py
```diff
@@ -201,12 +214,67 @@
...
+
+def is_structured_package_key(root: str, key: str) -> bool:
+    if not key.startswith(root):
+        return False
+    relative = key[len(root) :]
+    parts = relative.split("/")
+    return len(parts) == 2 and all(parts)
+
+
+def is_python_artifact(filename: str) -> bool:
```
HereThereBeDragons, line 291: this would also return true for tarballs.
maybe add a comment for context

build_tools/packaging/download_python_packages.py
```diff
@@ -1234,8 +1465,13 @@ def download_prerelease_packages(
     bucket_prefix: str = "v3/whl/",
     include_dependencies: bool = False,
     multi_arch: bool = False,
+    structured: bool = False,
```
HereThereBeDragons, line 1524: docstring off this function needs update explaining the new params

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
         help="Enable multi-arch package handling. Requires bucket prefix compatible with multi-arch layout (e.g. v4/whl/)",
     )
 
+    parser.add_argument(
+        "--structured",
+        action="store_true",
```
HereThereBeDragons: action=argparse.BooleanOptionalAction, 

is preferred by our codestyle

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
...
+        help=(
+            "Download Python packages from product-local structured roots "
```
HereThereBeDragons: "product-local structured roots"
dont understand a thing :)

e.g.
```
    "Use per-product repo layout v5/rocm/<product>/<index>/ (RFC0012)instead"
    "of the flat prerelease bucket. Implies --multi-arch."
```

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
...
+            "v5/rocm/<product>/<index>/<package>/. Implies --multi-arch."
+        ),
+    )
+
+    parser.add_argument(
+        "--repo-stream",
+        default="rc",
+        choices=REPO_STREAMS,
+        help=(
+            "repo.amd.com stream used with --structured "
```
HereThereBeDragons: this also need a bit better phrasing, maybe like 

claude suggestions:
long
```
            "repo.amd.com release stream to promote to (only used with --structured). "
            "dev = per-commit developer builds, nightly = promoted develop-branch builds, "
            "rc = release candidates. Selects the per-stream product buckets "
            "therock-repo-amd-<stream>-<product>. Default: rc"
```

short
```
            "Target repo.amd.com stream for --structured promotion "
            "(dev|nightly|rc; picks therock-repo-amd-<stream>-<product> buckets; default: rc)"
```

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
...
+            "(selects therock-repo-amd-<stream>-<product> buckets; default: rc)"
+        ),
+    )
+
+    parser.add_argument(
+        "--python-index",
+        default=STRUCTURED_DEFAULT_INDEX,
+        choices=STRUCTURED_INDEXES,
+        help=(
+            "Product-local Python index used with --structured "
```
HereThereBeDragons: e.g.
```
help=f"Wheel index with --structured: whl (plain pip, all archs) or whl-next (explicit device extra) (default: {STRUCTURED_DEFAULT_INDEX})",

```

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
...
+            f"(default: {STRUCTURED_DEFAULT_INDEX})"
+        ),
+    )
+
+    parser.add_argument(
+        "--product",
+        action="append",
+        default=[],
+        help=(
+            "Product to download with --structured: core, pytorch, or jax. "
```
HereThereBeDragons, line 1320: e.g.
```
help="Product(s) to download with --structured: core, pytorch, or jax. Repeatable or comma-separated (default: all)",
```

build_tools/packaging/download_python_packages.py
```diff
@@ -1067,6 +1226,56 @@ def parse_arguments(argv):
...
+            "Can be repeated or comma-separated. Defaults to all products."
+        ),
+    )
+
+    parser.add_argument(
+        "--tarball-variant",
+        default="release",
+        choices=sorted(CORE_TARBALL_PREFIXES),
+        help=(
+            "Tarball prefix variant used with --structured --include-tarballs: "
```
HereThereBeDragons, line 1330: e.g.
```
help="Core tarball selection when using --structured --include-tarballs: release or asan (default: release)",
```

build_tools/packaging/download_python_packages.py
```diff
@@ -1078,10 +1287,32 @@ def parse_arguments(argv):
...
+            and args.output_dir is not None
+        ):
+            args.tarball_output_dir = args.output_dir / core_tarball_dir_name(
+                args.tarball_variant
+            )
+        elif (
+            args.tarball_output_dir is None
+            and args.output_dir is not None
+            and not args.multi_arch
+        ):
```
HereThereBeDragons: either delete `# Validate arguments` or move it up

build_tools/packaging/download_python_packages.py
```diff
@@ -1078,10 +1287,32 @@ def parse_arguments(argv):
 
     if args.arch:
         args.arch = [arch.strip() for arch in args.arch.split(",")]
+    args.product = parse_csv_args(args.product) or list(STRUCTURED_PRODUCTS)
```
HereThereBeDragons: should be part of the `if args.structured` below?

build_tools/packaging/download_python_packages.py
```diff
@@ -1078,10 +1287,32 @@ def parse_arguments(argv):
...
+            and args.tarball_output_dir is None
+            and args.output_dir is not None
+        ):
+            args.tarball_output_dir = args.output_dir / core_tarball_dir_name(
+                args.tarball_variant
+            )
+        elif (
+            args.tarball_output_dir is None
+            and args.output_dir is not None
+            and not args.multi_arch
```
HereThereBeDragons, line 1370: not true. delete the line.
the old style both supports multi_arch and single arch. both will be downloaded into tarball/

build_tools/packaging/download_python_packages.py
```diff
@@ -201,12 +214,67 @@
...
 def filter_package_entries(
-    packages: List[Tuple[str, int]],
+    packages: list[tuple],
     include_package_globs: List[str] | None = None,
     exclude_package_globs: List[str] | None = None,
-) -> List[Tuple[str, int]]:
+) -> list[tuple]:
     """Filter package entries by filename glob patterns.
 
     Args:
```
HereThereBeDragons, line 303: docstring update needed

build_tools/packaging/download_python_packages.py
```diff
@@ -201,12 +214,67 @@
...
+    return len(parts) == 2 and all(parts)
+
+
+def is_python_artifact(filename: str) -> bool:
+    return filename.endswith(STRUCTURED_PACKAGE_EXTENSIONS)
+
 
 def filter_package_entries(
-    packages: List[Tuple[str, int]],
+    packages: list[tuple],
```
HereThereBeDragons: can we please not do that. where packages can be either size of 2 or 3. very confusing and nowhere written down
(there will be another longer comment explaning what i mean + suggestion)

build_tools/packaging/download_python_packages.py
```diff
@@ -644,26 +760,43 @@ def handle_multi_arch_downloads(
...
 
-    # Validate that packages exist
-    if not exists_version_multi_arch(
-        s3_client, bucket_name, bucket_prefix, None, version
-    ):
-        print(f"[ERROR]: No packages found for version {version}")
-        sys.exit(1)
+    if structured:
+        products = products or list(STRUCTURED_PRODUCTS)
+        packages_to_promote = list_packages_structured(
```
HereThereBeDragons, line 837: packages_to_promote is here len() == 3 with entry_bucket, key, size

below in list_packages_multi_arch is it only len() == 2 with key, size


can we please unify it? either by having a new dataclass or at least using the same tupel for structued and old style.

then we can also delete from some functions the "bucket" as func param.

considering having a couple of
```
            key = entry[-2]
            size = entry[-1]
```
having a dataclass will be nicer to read.

build_tools/packaging/download_python_packages.py
```diff
@@ -492,6 +561,53 @@ def list_packages_multi_arch(
     return packages_to_promote
 
 
+def list_packages_structured(
```
HereThereBeDragons, line 634: can we please deduplicate this with `list_packages_multi_arch()`?

<details><summary>claude suggestion</summary>


Refactor suggestion: unify list_packages_multi_arch and list_packages_structured behind a target-based scanner

The two listing functions duplicate the same paginate/filter/version-match loop; they differ only in which (bucket, prefix, key_filter) locations they scan. Extracting that into a small ScanTarget plus builder functions removes the duplication and keeps each signature to only the args it needs (no boolean mode flag, no conditionally-valid parameters):

```
from dataclasses import dataclass
from functools import partial
from typing import Callable


@dataclass
class ScanTarget:
    bucket: str
    prefix: str
    key_filter: Callable[[str, str], bool] | None = None


def _scan_packages(s3_client, bucket, prefix, version, architectures=None, key_filter=None):
    paginator = s3_client.get_paginator("list_objects_v2")
    for page in paginator.paginate(Bucket=bucket, Prefix=prefix):
        for obj in page.get("Contents", []):
            key = obj["Key"]
            filename = key.split("/")[-1]
            if not filename or filename == "index.html":
                continue
            if key_filter and not key_filter(key, filename):
                continue
            if (version in filename or f"rocm{version}" in filename) and is_allowed_multi_arch_package(
                filename, architectures
            ):
                yield PackageEntry(key=key, size=obj["Size"], bucket=bucket)


def multi_arch_targets(bucket, prefix):
    return [ScanTarget(bucket=bucket, prefix=prefix)]


def _structured_key_filter(root, key, filename):
    return is_python_artifact(filename) and is_structured_package_key(root, key)


def structured_targets(stream, products, index):
    targets = []
    for product in products:
        root = structured_python_root(product, index)
        targets.append(
            ScanTarget(
                bucket=repo_product_bucket(stream, product),
                prefix=root,
                key_filter=partial(_structured_key_filter, root),
            )
        )
    return targets


def list_packages(s3_client, targets, version, architectures=None):
    entries = []
    for target in targets:
        entries.extend(
            _scan_packages(
                s3_client, target.bucket, target.prefix, version, architectures, target.key_filter
            )
        )
    if not entries:
        print(f"[ERROR]: No packages found for version {version}")
        sys.exit(1)
    return entries
```


Call sites:

```
packages = list_packages(s3_client, structured_targets(stream, products, index), version, arch)
packages = list_packages(s3_client, multi_arch_targets(bucket, prefix), version, arch)
```

This pairs naturally with using a PackageEntry dataclass (key, size, bucket) instead of mixed-length tuples, so every entry carries its own bucket and downstream consumers no longer need to branch on tuple length. list_packages_for_arch is intentionally left as-is since it categorizes results differently. Optional cleanup, not a blocker.

</details>

build_tools/packaging/tests/download_python_packages_test.py
```diff
@@ -75,5 +98,185 @@ def test_unknown_package_is_not_allowed(self):
...
+        args = parse_arguments(
+            [
+                "--version=7.13.0rc1",
+                "--multi-arch",
+                "--include-tarballs",
+                "--output-dir=downloads",
+            ]
+        )
+
+        self.assertIsNone(args.tarball_output_dir)
```
HereThereBeDragons, line 235: should we check here the actual path like in the test below?

build_tools/packaging/upload_release_packages.py
```diff
@@ -109,12 +134,59 @@
...
+
+def repo_product_bucket(stream: str, product: str) -> str:
+    return f"therock-repo-amd-{stream}-{REPO_BUCKET_PRODUCT_NAMES[product]}"
+
+
+def core_tarball_dir_name(tarball_variant: str) -> str:
+    return "tarball-asan" if tarball_variant == "asan" else "tarball"
+
+
+def infer_structured_product(filename: str) -> str:
```
araravik-psd: Thanks for working on this. One thing worth fixing before merge: infer_structured_product in upload_release_packages.py has no allowlist for "core", it's just the fallback for anything that doesn't match the PyTorch/JAX prefixes. That means a typo'd/renamed/unrelated package in wheels/ would silently get uploaded to the core repo.amd.com bucket instead of erroring, which is hard to reverse for a release pipeline. Relatedly, an unparseable filename currently raises ValueError uncaught, which would crash the whole batch mid-upload.

I've got a fix ready (explicit core-package allowlist shared via _therock_utils/python_package_paths.py, plus a try/except so a bad file is skipped with an error instead of crashing or misrouting) with tests passing locally. Happy to push it as a commit here if that's useful, please let me know.

These are the only concerns I have before merging the PR

araravik-psd (dismissed):
Thanks Marius for working on this. I have only 2 concerns which I have raised below.

build_tools/packaging/download_python_packages.py
```diff
@@ -492,6 +561,53 @@ def list_packages_multi_arch(
...
+
+                if (
+                    not filename
+                    or filename == "index.html"
+                    or not is_python_artifact(filename)
+                    or not is_structured_package_key(root, key)
+                ):
+                    continue
+
+                matches_version = version in filename or f"rocm{version}" in filename
```
HereThereBeDragons: claude rightfully pointed out that this (also used in the old version) is fragile and a regex would be better to be used.

as it does matching via `in`, if we have packages for versions 10.0.0rc1 and 10.0.0rc10, if we just want to download 10.0.0rc1 we would also download 10.0.0rc10, rc11, rc12,... packages 

rather unlikely to have so many rc candidates or patch release but if it is being refactored anyway would be good to also take care of that

HereThereBeDragons (commented):
see comments. overall i think some more docstrings need adjustments to reflect the newly added codechanges. 

thanks also for adding the new unittests!

HereThereBeDragons:
oops just saw i forgot to review 2 files :( 
`build_tools/packaging/how_to_do_release.md`
and `build_tools/packaging/upload_release_packages.py`
will do it in a moment

build_tools/packaging/how_to_do_release.md
```diff
@@ -73,6 +73,66 @@ Output structure:
     *.whl
 ```
 
+### Structured repo.amd.com packages
+
+Use `--structured` when candidates live in the product-owned repo.amd.com
```
HereThereBeDragons:
```suggestion
Use `--structured` when candidates live in the product-owned (core, pytorch, jax, ...) repo.amd.com
```

side note:
in quartz we call those "pipelines": "rocm" "pytorch" "jax" "native packages". wonder if we should adjust it.

build_tools/packaging/upload_release_packages.py
```diff
@@ -109,12 +134,59 @@
     sys.exit(1)
 
 
+STRUCTURED_INDEXES = ("whl", "whl-next")
```
HereThereBeDragons: these are also declared in the download one. want to move all those to _therock_utils?

build_tools/packaging/upload_release_packages.py
```diff
@@ -137,7 +209,23 @@ def upload_python_files(
...
     upload_count = 0
 
-    if multi_arch:
+    if structured:
+        wheels_dir = input_dir / "wheels"
+
+        if not wheels_dir.exists():
+            print(f"[ERROR]: Structured wheels directory not found: {wheels_dir}")
+            return 0
+
```
HereThereBeDragons, line 191: strucuted and multiarch here have a lot of common. worth deduplicating.

build_tools/packaging/upload_release_packages.py
```diff
@@ -100,6 +117,14 @@
 from pathlib import Path
 from typing import Tuple
```
HereThereBeDragons, line 118: claude:
from typing import Tuple / -> Tuple[int, int]. Legacy hint. Use builtin tuple[int, int], drop the import.

marbre:
@araravik-psd, @HereThereBeDragons I am into other things right now as pointed out in my first comment. Anyone wants to take this over and incoorportate the feedback?

build_tools/packaging/upload_release_packages.py
```diff
@@ -366,18 +504,28 @@ def parse_arguments(argv):
...
+            args.tarball_bucket = repo_product_bucket(args.repo_stream, "core")
+            args.tarball_bucket_prefix = core_tarball_prefix(args.tarball_variant)
+        else:
+            args.bucket = "therock-release-python"
+        if args.multi_arch and not args.structured:
             args.bucket_prefix = "v4/rocm/whl/"
             args.tarball_bucket = "therock-release-tarball"
             args.tarball_bucket_prefix = "v4/rocm/tarball/"
-        else:
+        elif not args.structured:
```
HereThereBeDragons: claude:
--use-release-buckets help says v3/whl but code sets v3/rocm/whl/ (529) and v4/rocm/whl/ (525). Stale help. Docstring diagram (91) also stale.

build_tools/packaging/download_python_packages.py
marbre: Packages need to go into a subdirectory with package name, compared to the flat index we've been using so far. `--structured` is used across all our scripts that upload packages, this is the last one missing support for this.

build_tools/packaging/download_python_packages.py
marbre: See the argument helps:
```
            "Upload flat local wheels from <input-dir>/wheels into product-local "
            "structured roots v5/rocm/<product>/<index>/<package>/. Implies --multi-arch."
 ```

File: build_tools/packaging/upload_release_packages.py [hunk 28]
```diff
@@ -319,6 +417,35 @@ def parse_arguments(argv):
         help="Upload to multi-arch flat layout (no arch subdirectories)",
     )
 
+    parser.add_argument(
+        "--structured",
+        action="store_true",
```
HereThereBeDragons: action=argparse.BooleanOptionalAction,

build_tools/packaging/upload_release_packages.py
```diff
@@ -319,6 +417,35 @@ def parse_arguments(argv):
...
+        help=(
+            "Upload flat local wheels from <input-dir>/wheels into product-local "
```
HereThereBeDragons: again the descirption could be a bit better for people not familiar with rfc0012

build_tools/packaging/upload_release_packages.py
araravik-psd: Thanks @marbre , I will take care of pushing the changes for this

build_tools/packaging/upload_release_packages.py
```diff
@@ -366,18 +504,28 @@ def parse_arguments(argv):
...
     args = parser.parse_args(argv)
 
+    if args.structured:
+        args.multi_arch = True
+        if args.execute and not args.use_release_buckets:
+            parser.error("--structured --execute requires --use-release-buckets")
+
     # Adjust testing tarball prefix for multi-arch uploads
-    if args.multi_arch and not args.use_release_buckets:
+    if args.multi_arch and not args.structured and not args.use_release_buckets:
```
HereThereBeDragons, line 507: you also want a testing bucket for structured?

build_tools/packaging/upload_release_packages.py
```diff
@@ -366,18 +504,28 @@ def parse_arguments(argv):
 
     args = parser.parse_args(argv)
 
+    if args.structured:
+        args.multi_arch = True
```
HereThereBeDragons, line 495: maybe then you should write that if --structured is used it will ignore whatever --multi-arch is set to -- or do a check (need then the default to be None to detect if a user has used it or not and only after the check set it to the current default)

build_tools/packaging/upload_release_packages.py
```diff
@@ -366,18 +504,28 @@ def parse_arguments(argv):
...
+        if args.execute and not args.use_release_buckets:
```
HereThereBeDragons: no test bucket allowed for --structured?

build_tools/packaging/upload_release_packages.py
```diff
@@ -425,8 +577,17 @@ def upload_release_packages(
...
     print("=" * 80)
     print(f"Input directory: {input_dir.absolute()}")
-    print(f"Python bucket: {bucket_name}")
-    print(f"Python bucket prefix: {bucket_prefix}")
+    if structured:
+        print(f"Repo stream: {repo_stream}")
+        print("Python buckets:")
+        for product in ("core", "pytorch", "jax"):
+            print(
+                f"  {product}: s3://{repo_product_bucket(repo_stream, product)}/"
```
HereThereBeDragons: here you are printing the full s3 URL - should we then also change it for the rest or reduce to have it without the s3://?

HereThereBeDragons (commented):
so now i have reviewed everything. happy reading :)

some claude comments:

This is a second axis of duplication beyond the v5/stream constants I already flagged — product-classification logic, not just path constants. And the two encodings do not agree:

Jax rule differs. download matches jax via JAX_ROCM_PACKAGE_PATTERN (regex, jax_rocm<major>_...). upload matches via bare prefixes ("jax","jaxlib","jax-rocm"). So jaxlib is "jax product" to upload but is NOT matched by download's jax pattern (it rides PACKAGES_TO_PROMOTE instead). Same wheel, two different mental models.

Unknown-case philosophy is opposite. download's categorize_package returns "unknown" for anything unrecognized (fail-loud, caller can skip). upload's infer_structured_product returns "core" as a silent default (fail-silent). So a stray/unknown wheel that download would flag as unknown, upload routes straight into the core production bucket under --execute --use-release-buckets. That is the misrouting risk I flagged earlier, and it is directly caused by the two scripts not sharing one classifier.

build_tools/packaging/upload_release_packages.py
araravik-psd: I have pushed a fix for this with cb770c60cd18cf44e3e28426d292963e620e00f7

build_tools/packaging/download_python_packages.py
araravik-psd: Fixed in the latest commit 32473fedf5664a95dfd225a5b26af6b606819095, replaced the in-based substring check with a bounded regex match (_version_matches() in download_python_packages.py) so a requested version can't accidentally match a longer version that contains it as a prefix/infix (e.g. 10.0.0rc1 no longer matches 10.0.0rc10, and 3.0.0 no longer matches 13.0.0).

This replaces all 7 call sites that were doing plain substring matching: exists_version_single_arch, exists_version_multi_arch, list_packages_multi_arch, list_packages_structured, list_packages_for_arch, and the two checks in list_tarball_for_package.

Also added regression tests (test_version_matching_rejects_rc_suffix_collision, test_version_matching_rejects_numeric_prefix_collision) that fail against the old substring-match logic and pass with the fix, since the existing suite didn't actually cover this case.

araravik-psd (commented):
@HereThereBeDragons  I have taken care of 2 must fix issues in the latest 2 commits. Let us sync up to figure out what else is needed to merge this.

build_tools/packaging/download_python_packages.py
araravik-psd: Looked into this closer and I don't think this line should come out. handle_multi_arch_downloads (line ~848) already has its own default for the non-structured multi-arch case, it sets tarball-multi-arch/, not tarball/, and only when tarball_output_dir is still None coming in. There's an existing test (test_parse_multi_arch_tarballs_keep_handler_default_output_dir) that encodes exactly this contract. Removing 
and not args.multi_arch here would make parse_arguments pre-fill tarball_output_dir with tarball/, which short-circuits that handler default and puts multi-arch tarballs in the wrong directory, the test fails immediately if I make this change. Keeping the guard as-is; Please let me know if I'm missing something about the "old style" behavior you had in mind.

build_tools/packaging/upload_release_packages.py
araravik-psd: Fixed it, updated the --use-release-buckets help text and the module docstring's S3 bucket-structure diagram (both the v3 and v4 sections) to say v3/rocm/whl / v3/rocm/tarball / v4/rocm/whl / v4/rocm/tarball, matching what the code actually sets.

build_tools/packaging/download_python_packages.py
araravik-psd: Moved --product parsing/validation now lives inside if args.structured:. Confirmed it's safe: args.product is only ever read downstream inside the structured branch of handle_multi_arch_downloads, so there was no reason to default/validate it when --structured isn't passed.

araravik-psd (commented):
Pushed fixes for the 2 must-fix issues: infer_structured_product now has an explicit core allowlist and fails loud instead of silently misrouting unknown packages to core (cb770c60), and version matching now uses a bounded regex instead of substring checks that could match 10.0.0rc1 against 10.0.0rc10 (32473fed). Also cleaning up a few smaller items in my next commit --structured switched to BooleanOptionalAction for consistency, --product parsing scoped inside if args.structured:, and the stale v3/whl/v4/whl help text and docstring diagram corrected to v3/rocm/whl/v4/rocm/whl. I'm holding off on the suggested removal of the and not args.multi_arch tarball-output-dir check since handle_multi_arch_downloads already relies on it defaulting to tarball-multi-arch/, and an existing test locks that behavior in, please let me know if I'm missing context there. Remaining open items (testing bucket for --structured, whether --structured silently overriding --multi-arch should be documented/validated, and the jax-classification duplication between download and the shared infer_structured_product) feel like good candidates for a follow-up rather than blockers let me know if that works and what else is
needed to merge @HereThereBeDragons .

araravik-psd:
Pushed a follow-up addressing the remaining structural feedback:

* Added `PackageEntry` to consistently carry bucket, key, and size; removed redundant bucket parameters from download helpers.
* Deduplicated package listing logic into `_scan_bucket_prefix()` and merged structured/multi-arch upload branches.
* Centralized repo-stream, bucket-name, and tarball-path constants/helpers in `_therock_utils/python_package_paths.py`, eliminating duplicated logic that caused the previous PyTorch bucket-name typo.
* Fixed the structured upload summary `s3:///` typo.
* Replaced remaining `typing.List/Tuple/Dict/Union` usages with built-in generics.
* Cleaned up CLI help text, docstrings, and comments.
* Fixed `how_to_do_release.md` to reference the correct PyTorch `-pytorch` bucket.
* Updated tests; all three suites pass:

  * `download_python_packages_test.py`
  * `upload_release_packages_test.py`
  * `python_package_paths_test.py`

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -178,3 +212,51 @@ def plan_key_copies(
             )
         )
     return plans
+
+
+# Product classification for the flat local promotion layout, where core,
```
HereThereBeDragons: float local promotion layout? whats that?

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -178,3 +212,51 @@ def plan_key_copies(
...
+# PyTorch, and JAX artifacts are intermixed in a single directory before being
+# routed to their per-product structured buckets. Kept as an explicit
+# allowlist (rather than defaulting unmatched packages to "core") so an
+# unrecognized or renamed package fails loudly instead of silently landing in
+# the wrong product bucket.
+PYTORCH_PACKAGE_NAMES = frozenset({"apex", "triton"})
+PYTORCH_PACKAGE_PREFIXES = ("torch", "amd-torch")
+JAX_PACKAGE_PREFIXES = ("jax", "jaxlib", "jax-rocm")
+CORE_PACKAGE_PREFIXES = (
```
HereThereBeDragons, line 237: there is also rocm-bootstrap that is not changed by the promotion script. i dont thinkwe need to download it or anything. but it owuld be good to add a comment about it.

it might be also good to know which must be correct matches and which like the device are wildcards.

build_tools/packaging/upload_release_packages.py
```diff
@@ -88,22 +88,22 @@
...
       package1.whl
       package2.whl
       ...
-    v3/tarball/
+    v3/rocm/tarball/
       therock-dist-linux-<arch1>-<version>.tar.gz
       therock-dist-windows-<arch2>-<version>.tar.gz
       ...
 
   S3 bucket structure for mult-arch:
```
HereThereBeDragons: multi-arch:

"i" missing

build_tools/packaging/upload_release_packages.py
```diff
@@ -209,34 +177,24 @@ def upload_python_files(
...
-        print("\nMulti-arch wheels")
+        print(f"\n{label} wheels")
 
-        files_to_upload = []
-        files_to_upload.extend(wheels_dir.glob("*.whl"))
-        files_to_upload.extend(wheels_dir.glob("*.tar.gz"))
+        # Structured layout also accepts sdists/zips (e.g. ROCm Core tarballs);
+        # multi-arch's flat layout is wheels/tarballs only.
+        extensions = (
+            ("*.whl", "*.tar.gz", "*.zip") if structured else ("*.whl", "*.tar.gz")
```
HereThereBeDragons: lets clarify if we have "*.zip" files even or if AI made it up

build_tools/packaging/upload_release_packages.py
```diff
@@ -270,7 +228,11 @@ def upload_python_files(
 
         for file_path in sorted(files_to_upload):
             if structured:
-                product = infer_structured_product(file_path.name)
+                try:
+                    product = infer_structured_product(file_path.name)
+                except ValueError as e:
```
HereThereBeDragons, line 239: shouldnt this be a hard failure if we have a file in the directory that shouldnt be uploaded?

build_tools/packaging/upload_release_packages.py
HereThereBeDragons: can we also get this addressed? in either direction

build_tools/packaging/download_python_packages.py
```diff
@@ -18,6 +18,13 @@
 
 Previously downloaded packages are skipped.
 
+"Structured" (--structured) downloads from the per-product repo.amd.com layout
```
HereThereBeDragons, line 21: should we add a comment that this is from rocm 10 on?

build_tools/packaging/download_python_packages.py
```diff
@@ -215,23 +236,34 @@
 }
 
 STRUCTURED_PRODUCTS = ("core", "pytorch", "jax")
```
HereThereBeDragons: could be extract from REPO_BUCKET_PRODUCT_NAMES.keys()?

build_tools/packaging/download_python_packages.py
```diff
@@ -267,31 +287,33 @@ def is_structured_package_key(root: str, key: str) -> bool:
...
+    # Also matches ROCm Core tarball filenames (.tar.gz), not just wheels/sdists.
     return filename.endswith(STRUCTURED_PACKAGE_EXTENSIONS)
 
 
 def filter_package_entries(
-    packages: list[tuple],
-    include_package_globs: List[str] | None = None,
-    exclude_package_globs: List[str] | None = None,
-) -> list[tuple]:
+    packages: list[PackageEntry | tuple[str, int]],
```
HereThereBeDragons, line 297: why does single arch still extra treatement? it can also use package entry.

but leave as is. i will rip out single arch after we merged the big changes.

build_tools/packaging/download_python_packages.py
```diff
@@ -445,6 +467,23 @@ def list_architectures(
         sys.exit(1)
 
 
+def _version_matches(filename: str, version: str) -> bool:
+    """Check whether `filename` contains an exact match of `version`.
+
+    A plain `version in filename` substring check is unsound for dotted
```
HereThereBeDragons: historical comment not needed

build_tools/packaging/download_python_packages.py
```diff
@@ -445,6 +467,23 @@ def list_architectures(
...
+    version strings: "7.13.0rc1" is a substring of "7.13.0rc10", and "3.0.0"
+    is a substring of "13.0.0". This only matches when `version` (or, for
+    torch-style local versions, `rocm{version}`) is not directly abutted by
+    another alphanumeric character on either side, so it can't partially
+    match a longer/different version.
+    """
+    for candidate in (version, f"rocm{version}"):
+        pattern = r"(?<![A-Za-z0-9])" + re.escape(candidate) + r"(?![A-Za-z0-9])"
```
HereThereBeDragons: claude thinks the regex is not correct:

**`_version_matches`: `.post`/`.dev` sdists leak through the boundary check**

The right boundary `(?![A-Za-z0-9])` treats `.` as a word boundary, so a query
for release `7.13.0` also matches its post/dev sdists:

- `_version_matches("rocm-7.13.0.post1.tar.gz", "7.13.0")` -> `True`
- `_version_matches("rocm-7.13.0.dev0.tar.gz", "7.13.0")` -> `True`

We cannot just add `.` to the boundary, because a real sdist also puts a `.`
after the version (`rocm-7.13.0.tar.gz`) -- a boundary-only regex cannot tell
the `.tar.gz` extension from another version segment.

Fix: use a positive terminator -- a version is legally followed only by a wheel
`-`, the `.tar.gz`/`.zip` extension, or end of string:

```python
pattern = (
    r"(?<![A-Za-z0-9])"
    + re.escape(candidate)
    + r"(?=-|\.tar\.gz|\.zip|$)"
)

File: build_tools/packaging/download_python_packages.py [hunk 44]
```diff
@@ -495,39 +534,38 @@ def exists_version_multi_arch(s3_client, bucket, prefix, directory, version):
...
+    """Paginate a bucket/prefix, filter by version+arch, return PackageEntry list.
 
     Args:
         s3_client: boto3 S3 client
         bucket: S3 bucket name
-        prefix: S3 prefix containing multi-arch packages
+        prefix: S3 prefix to paginate
         version: Version pattern to filter packages
         architectures: Optional list of architectures used for filtering
+        extra_filter: Optional callable(key, filename) -> bool applied before
```
HereThereBeDragons: do you understand this description? because i dont

build_tools/packaging/download_python_packages.py
```diff
@@ -568,38 +632,24 @@ def list_packages_structured(
...
-                ):
-                    packages_to_promote.append((bucket, key, obj["Size"]))
+        packages_to_promote.extend(
+            _scan_bucket_prefix(
+                s3_client,
+                bucket,
+                root,
+                version,
+                architectures,
+                extra_filter=lambda key, filename: is_python_artifact(filename)
```
HereThereBeDragons: claude:

The lambda closes over root, which is reassigned each loop iteration. This is safe here only because .extend(_scan_bucket_prefix(...)) forces the scan to run eagerly within the same iteration, so the lambda executes while root still holds the current product's value. If _scan_bucket_prefix were ever refactored to defer evaluation (return a generator consumed later), every lambda would capture the last root — the classic late-binding-closure trap. Cheap insurance is to bind it explicitly:


extra_filter=lambda key, filename, root=root: is_python_artifact(filename)
                                              and is_structured_package_key(root, key)
Optional, but it removes the latent footgun.

build_tools/packaging/download_python_packages.py
```diff
@@ -1240,8 +1287,9 @@ def parse_arguments(argv):
         default="rc",
         choices=REPO_STREAMS,
         help=(
-            "repo.amd.com stream used with --structured "
-            "(selects therock-repo-amd-<stream>-<product> buckets; default: rc)"
+            "repo.amd.com release stream to read from with --structured: "
+            "dev (continuous), nightly, or rc (release-candidate). Selects the "
```
HereThereBeDragons: delete "continuous". dont think rc needs to be explained

HereThereBeDragons (commented):
thanks looks already way better. but still a couple of thinks that need improving

build_tools/_therock_utils/python_package_paths.py
```diff
@@ -39,6 +39,40 @@
 
 # Valid aggregate index names (the second path segment).
 INDEX_NAMES = ("whl", "whl-next")
+DEFAULT_INDEX = "whl-next"
```
HereThereBeDragons, line 42: looking at #7547 we probably can then reuse some of those things.

build_tools/_therock_utils/python_package_paths.py
araravik-psd: meant "flat": before --structured routing, core/PyTorch/JAX artifacts all land side-by-side in one local directory (e.g. <output-dir>/wheels/) instead of being split by product. Worded the comment to spell that out explicitly.

build_tools/_therock_utils/python_package_paths.py
araravik-psd: Added a note that rocm-bootstrap isn't classified here because promote_packages.py already special-cases and skips it before this code path runs. Also clarified that PYTORCH_PACKAGE_NAMES is an exact-match set 
(apex/triton don't share a prefix with torch), while the *_PREFIXES tuples are str.startswith() wildcard-style 
matches, e.g. rocm_sdk_device_gfx1100 matching rocm-sdk-device.

build_tools/packaging/upload_release_packages.py
araravik-psd: Fixed the typo ("mult-arch" → "multi-arch").

build_tools/packaging/upload_release_packages.py
araravik-psd: This predates this PR. .zip sdist support goes back to #6622 (27eeb1a4f), and build_tools/tests/python_package_paths_test.py has had a passing case for it since (example test llnl-hatchet-2024.1.zip). 
This PR just centralizes the extension tuple into ACCEPTED_FILE_EXTENSIONS, it doesn't add new format support.

build_tools/packaging/upload_release_packages.py
araravik-psd: Agreed that this has to be a hard failure, changed to sys.exit(1) on an unrecognized package instead of skipping it with a printed error. Updated the test (test_structured_upload_hard_fails_on_unknown_package) to assert the SystemExit(1) instead of a soft-skip count.

build_tools/packaging/download_python_packages.py
araravik-psd: Added: "The structured layout is only used from ROCm 10 onward; earlier releases only exist in the flat prerelease bucket."

build_tools/packaging/download_python_packages.py
araravik-psd: Done, STRUCTURED_PRODUCTS = tuple(REPO_BUCKET_PRODUCT_NAMES.keys()). The upload side still needs the same treatment

build_tools/packaging/download_python_packages.py
araravik-psd: Understood, leaving as-is per your note, no change made here, will revisit once single-arch is ripped out.

build_tools/packaging/download_python_packages.py
araravik-psd: Trimmed the docstring down to just what the function does, dropped the "here's why a naive substring check is not good enough" history :).

build_tools/packaging/download_python_packages.py
araravik-psd: Confirmed and fixed: the old (?![A-Za-z0-9]) right boundary treated . as a valid stop, so 7.13.0 matched inside 
7.13.0.post1.tar.gz/7.13.0.dev0.tar.gz. Changed to a positive terminator — -, .tar.gz, .zip, or end-of-string and added test_version_matching_rejects_post_and_dev_suffix_collision to cover it.

build_tools/packaging/download_python_packages.py
araravik-psd: Reworded it to make it more clear, now says extra_filter is checked before version/arch matching, gives list_packages_structured's actual use (rejecting keys that aren't a direct child of the product's package directory) as the concrete example.

build_tools/packaging/download_python_packages.py
araravik-psd: Took care of this, trimmed to "dev, nightly, or rc. Selects the therock-repo-amd-<stream>-<product> .  buckets (default: rc)" applied the same trim in upload_release_packages.py's matching --repo-stream help text.

build_tools/packaging/download_python_packages.py
araravik-psd: Thanks for catching this, fixed by binding root as a default argument on the lambda (root=root), so it captures the value at loop time instead of relying on the outer variable later. Not an active bug today since the filter is used immediately each iteration, but it prevents a classic Python closure bug if the code changes later. Also added a docstring note explaining extra_filter.

build_tools/packaging/upload_release_packages.py
araravik-psd: Fixed the inconsistency, dropped the s3:// scheme from this print so it matches the plain Python bucket {bucket_name} / Tarball bucket: {tarball_bucket_name} format used below it. 

While in here, also swapped the hardcoded ("core", "pytorch", "jax") 
tuple for STRUCTURED_PRODUCTS = tuple(REPO_BUCKET_PRODUCT_NAMES.keys()), mirroring the download side, so both files derive the product list from the same source instead of maintaining separate copies.

build_tools/_therock_utils/python_package_paths.py
araravik-psd: Agreed, INDEX_NAMES, DEFAULT_INDEX, REPO_STREAMS, REPO_BUCKET_PRODUCT_NAMES, and repo_product_bucket() are all generic enough that #7547 should import and reuse them rather than redefining the same product/stream/bucket-naming scheme independently. I'll leave that as a follow-up once both PRs land, since 
 #7547 is a separate workflow-rewiring change

build_tools/packaging/upload_release_packages.py
HereThereBeDragons: @marbre as far as i am aware currently we dont have an .zip files? (i know wheel under the hood is zip)
just wondering how far we need to add this support (promotion scripts etc) as long as we dont have a live package that ends on .zip

build_tools/packaging/tests/download_python_packages_test.py
```diff
@@ -330,6 +330,54 @@ def test_version_matching_rejects_rc_suffix_collision(self):
             ],
         )
 
+    def test_version_matching_rejects_post_and_dev_suffix_collision(self):
```
HereThereBeDragons, line 333: you might also want to have this check for .whl as they end with `-cp310-cp310-win_amd64.whl` or `-py3-none-manylinux_2_27_x86_64.whl`, etc.

build_tools/packaging/download_python_packages.py
```diff
@@ -1583,16 +1591,23 @@ def download_prerelease_packages(
...
     print("Download Prerelease Packages")
     print("=" * 80)
-    print(f"Bucket: {bucket_name}")
     print(f"Version: {version}")
     if structured:
-        print(f"Structured products: {products or list(STRUCTURED_PRODUCTS)}")
+        structured_products = products or list(STRUCTURED_PRODUCTS)
         print(f"Repo stream: {repo_stream}")
         print(f"Python index: {python_index}")
+        print("Python buckets:")
```
HereThereBeDragons: S3 buckets?

HereThereBeDragons (commented):
2 small things and then i am happy with it :)

build_tools/packaging/upload_release_packages.py
marbre: The Python package is a bit more generic and also tests for `.zip` as this is a valid format for source dists but we use `.tar.gz` instead. Thus it would be acceptable to be more strict here and to not accept `.zip` but no too strong feelings.

Where we will have zip files is as archives for Windows (accompanying tarballs) but they will go do a different path.

build_tools/packaging/download_python_packages.py
marbre: Yeah, there is no "Python bucket". There was a python specific bucket in the legacy structure but this is no longer existing and we now rather have per-product buckets.

marbre:
Please rebase on latest main or merge main into this branch. Rewiring changes landed.

build_tools/packaging/upload_release_packages.py
```diff
@@ -361,23 +473,33 @@ def parse_arguments(argv):
...
     # Adjust testing tarball prefix for multi-arch uploads
-    if args.multi_arch and not args.use_release_buckets:
+    if args.multi_arch and not args.structured and not args.use_release_buckets:
         args.tarball_bucket_prefix = "release-upload-testing/tarball-multi-arch/"
 
     if args.use_release_buckets:
-        args.bucket = "therock-release-python"
-
-        if args.multi_arch:
+        if args.structured:
```
HereThereBeDragons, line 511: having had another chat with @marbre we found the following:
the current implementation does not work for uploading. as we currently only have `REPO_STREAMS = ("dev", "nightly", "rc")` but no stable-staging oder stable.

as such it would be good to keep `--use_release_buckets` for legacy and introduce a new args `--use_stable_staging` which overwrites repo_stream to the stable staging buckets. 
in general i also think we should have a safer default for `--repo-stream`. right now it defaults to upload to the "rc" stream. can we either have a safe bucket? maybe "dev"? or at least enforce that the user actively sets it and not having a default?

HereThereBeDragons (changes requested):

build_tools/packaging/tests/download_python_packages_test.py
araravik-psd: Good catch here, added test_version_matching_rejects_post_and_dev_suffix_collision_for_wheels covering the wheel filename shape (-cp310-cp310-win_amd64.whl, -py3-none-manylinux_2_27_x86_64.whl) with the same .post1/.dev0 collision case as the tarball test above. Passes with the existing _version_matches fix, no code changes are required here.

build_tools/packaging/download_python_packages.py
araravik-psd: Agreed, changed both to "S3 buckets:" since there's no dedicated "Python bucket" anymore, just the per-product buckets.

build_tools/packaging/upload_release_packages.py
araravik-psd: Thanks for flagging this. Implemented as discussed:
 - Kept --use-release-buckets untouched for the legacy flat layout.
 - Added --use-stable-staging (requires --structured), which overwrites --repo-stream to a new "stable-staging" value and also satisfies the --structured --execute production-bucket gate (in place  of, or alongside, --use-release-buckets).
 - Removed the --repo-stream default of "rc", it is now None, and --structured without an explicit --repo-stream (or --use-stable-staging) is a hard parser.error instead of silently uploading to rc.

build_tools/packaging/upload_release_packages.py
araravik-psd: No strong feelings either way too, left ACCEPTED_FILE_EXTENSIONS as-is (still includes .zip) since you said it's acceptable either way and this predates the PR (#6622). Happy to drop .zip from the structured upload path specifically if you'd rather tighten it now.

HereThereBeDragons:
whyyyy a force push? ;_;

build_tools/packaging/upload_release_packages.py
```diff
@@ -476,12 +477,37 @@ def parse_arguments(argv):
...
+            "Overwrite --repo-stream to 'stable-staging' for structured "
+            "stable-staging release uploads. Requires --structured; independent "
+            "of --use-release-buckets, which only affects the legacy flat "
+            "bucket layout."
+        ),
+    )
+
     args = parser.parse_args(argv)
 
+    if args.use_stable_staging and not args.structured:
```
HereThereBeDragons: this should be inside the if args.structured branch

build_tools/packaging/upload_release_packages.py
```diff
@@ -476,12 +477,37 @@ def parse_arguments(argv):
...
-        if args.execute and not args.use_release_buckets:
-            parser.error("--structured --execute requires --use-release-buckets")
+        if args.use_stable_staging:
+            args.repo_stream = "stable-staging"
+        elif args.repo_stream is None:
+            parser.error(
+                "--structured requires --repo-stream to be set explicitly "
+                "(dev, nightly, rc, or stable-staging), or pass --use-stable-staging"
+            )
+        if args.execute and not (args.use_release_buckets or args.use_stable_staging):
```
HereThereBeDragons: this entire if can go away. we are already checking above if repo_stream is set. 

use_release_buckets we dont need and use_stable_staging we test above

build_tools/packaging/upload_release_packages.py
araravik-psd: Done, moved it to an elif args.use_stable_staging: parser.error(...) paired with the same if args.structured block, instead of a separate leading check.

build_tools/packaging/upload_release_packages.py
araravik-psd: Agreed, removed. Wheel uploads already always route through repo_product_bucket(repo_stream, product) regardless of --use-release-buckets, and repo_stream is now guaranteed non-None by the branch
above whenever --structured is used, so the extra --execute gate wasn't protecting anything.

araravik-psd (commented):
@HereThereBeDragons Resolved the last 2 changes. please take a look if you are still around today

build_tools/packaging/upload_release_packages.py
HereThereBeDragons, line 511: i think this entire if-tree still needs work. please double check.

HereThereBeDragons: same here. the elif doesnt make sense. these here is the single arch path as far as i am aware

HereThereBeDragons (approved):
i will approve now to unblock you but you need to rework the ifs in the parsing of the args. see the comments. and please look also at the others in case something more got forgotten

build_tools/packaging/upload_release_packages.py
araravik-psd: Fixed it, restructured this into a single flat if/elif/else (structured / legacy multi-arch / legacy single-arch) instead of the previous two-stage if/else followed by a separate if/elif that could overwrite what the first block just set. 

Each branch now sets its own bucket/prefix values in one place, so there's no cross-block overwrite to trace through. See
   the updated block starting at if args.use_release_buckets:.

araravik-psd: You're right, this is the legacy single-arch (v3) path. It's now a plain else at the end of the same three-way if/elif/else (structured -> if, legacy multi-arch -> elif args.multi_arch, legacy single-arch -> else), with a 
`# Legacy flat single-arch layout`. comment so it's unambiguous. No more standalone elif not args.structured floating separately from the block above it.

HereThereBeDragons, line 510: but we dont use `--use_release_buckets` anymore with `--structured` ?

PR #7360, opened 2026-08-14
ci: gate on KFD IOCTL version in driver/GPU sanity check

lumachad:
## Summary

Extends `build_tools/print_driver_gpu_info.py` to query the KFD IOCTL
version from `/dev/kfd` and log it alongside the existing amd-smi and
rocminfo output. The sanity job now exits non-zero when the version falls
below the minimum rocdbgapi requires (>= 1.13), when `/dev/kfd` is
missing, or when the ioctl call fails. A version above the tested ceiling
(>= 2.0) produces a warning in the log but does not block testing, since
it may still work fine in practice.

The `test_sanity_check` job in `test_artifacts.yml` is gated directly
on `fromJSON(sanity_component).test_runner`, removing the separate
`run_gpu_sanity_check` output variable that was introduced in an earlier
iteration of this PR.

## Motivation

CI runs on runners with old KFD drivers (e.g. version 1.6) were producing
a cryptic rocdbgapi failure:

```
rocm-dbgapi: warning: AMD GPU driver's version 1.6 not supported (version must be >= 1.13 and < 2.0)
rocm-debug-agent: error: amd_dbgapi_process_attach(...) failed (rc=-10)
```

Surfacing the KFD version in the existing sanity step makes mismatches
immediately visible in CI logs and blocks further testing on a runner that
would produce misleading results. CPU-only runners are unaffected because
the KFD section only runs on Linux with a GPU runner.

ISSUE ID: N/A

### DIFF
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -129,7 +129,6 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       sanity_component: ${{ steps.configure.outputs.sanity_component }}
-      run_gpu_sanity_check: ${{ steps.configure.outputs.run_gpu_sanity_check }}
       components: ${{ steps.configure.outputs.components }}
       platform: ${{ steps.configure.outputs.platform }}
       shard_arr: ${{ steps.configure.outputs.shard_arr }}
@@ -194,7 +193,7 @@ jobs:
   test_sanity_check:
     name: 'Test Sanity Check'
     needs: configure_test_matrix
-    if: ${{ needs.configure_test_matrix.outputs.run_gpu_sanity_check == 'true' }}
+    if: ${{ fromJSON(needs.configure_test_matrix.outputs.sanity_component).test_runner != '' }}
     uses: './.github/workflows/test_component.yml'
     secrets: inherit
     with:
--- a/.github/workflows/test_artifacts.yml
+++ b/.github/workflows/test_artifacts.yml
@@ -129,6 +129,7 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       sanity_component: ${{ steps.configure.outputs.sanity_component }}
+      run_gpu_sanity_check: ${{ steps.configure.outputs.run_gpu_sanity_check }}
       components: ${{ steps.configure.outputs.components }}
       platform: ${{ steps.configure.outputs.platform }}
       shard_arr: ${{ steps.configure.outputs.shard_arr }}
@@ -193,6 +194,7 @@ jobs:
   test_sanity_check:
     name: 'Test Sanity Check'
     needs: configure_test_matrix
+    if: ${{ needs.configure_test_matrix.outputs.run_gpu_sanity_check == 'true' }}
     uses: './.github/workflows/test_component.yml'
     secrets: inherit
     with:
--- a/build_tools/print_driver_gpu_info.py
+++ b/build_tools/print_driver_gpu_info.py
@@ -8,21 +8,43 @@ Sanity check script for CI runners.
 On Linux:
   - run "amd-smi static"
   - run "rocminfo"
+  - report the KFD IOCTL version and warn if outside the range required by rocdbgapi (>= 1.13 and < 2.0)
 
 On Windows:
   - run "hipInfo.exe"
 
-This script prints only raw command output.
+This script always exits 0; the KFD version check is informational only.
 """
 
+import fcntl
 import os
 from pathlib import Path
 import platform
 import shlex
 import shutil
+import struct
 import subprocess
 import sys
-from typing import List, Optional
+from typing import List, Optional, Tuple
+
+
+# AMDKFD_IOC_GET_VERSION = _IOR('K', 0x01, struct { u32 major; u32 minor; })
+# _IOR: direction=0x80, size=8, type='K'=0x4b, nr=0x01 -> 0x80084b01
+_AMDKFD_IOC_GET_VERSION = 0x80084B01
+_KFD_DEVICE = "/dev/kfd"
+_KFD_VERSION_MIN = (1, 13)
+_KFD_VERSION_MAX = (2, 0)  # exclusive
+
+
+def _get_kfd_version() -> Tuple[int, int]:
+    fd = os.open(_KFD_DEVICE, os.O_RDWR)
+    try:
+        buf = bytearray(8)
+        fcntl.ioctl(fd, _AMDKFD_IOC_GET_VERSION, buf)
+        major, minor = struct.unpack("II", buf)
+        return major, minor
+    finally:
+        os.close(fd)
 
 
 def log(*args, **kwargs):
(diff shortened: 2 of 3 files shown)


lumachad:
Quick test for now.

lumachad:
Looks like the GPU sanity checks are running even on CPU runners. That doesn't seem right.

build_tools/print_kfd_version.py
geomin12, line 1: honestly, it'll be good to add this to the `print_driver_gpu_info.py` ? it might be good to log this every time and check this. this way too, it already does the "GPU"-only machine

ScottTodd: +1, looks like this code can fit in the existing script

lumachad: I can do that.

lumachad: Should be done now. I've kept the code to skip GPU Sanity check entirely on CPU runners. In my opinion it does not make sense to have that job running on CPU runners and it pollutes things.

build_tools/print_driver_gpu_info.py
```diff
@@ -8,21 +8,43 @@
 On Linux:
   - run "amd-smi static"
   - run "rocminfo"
+  - report the KFD IOCTL version and warn if outside the range required by rocdbgapi (>= 1.13 and < 2.0)
 
 On Windows:
   - run "hipInfo.exe"
 
-This script prints only raw command output.
+This script always exits 0; the KFD version check is informational only.
```
ScottTodd: "This script always exits 0" is no longer true as of https://github.com/ROCm/TheRock/pull/6604, the `run_command()` function now passes `check=True` so if there is a driver issue on the runner then the test job fails immediately. Previously tests would continue even after errors like `checkHipErrors() HIP API error = 0100 "no ROCm-capable device is detected"`

.github/workflows/test_artifacts.yml
```diff
@@ -190,6 +191,7 @@ jobs:
   test_sanity_check:
     name: 'Test Sanity Check'
     needs: configure_test_matrix
+    if: ${{ needs.configure_test_matrix.outputs.run_gpu_sanity_check == 'true' }}
```
ScottTodd: Is this new `run_gpu_sanity_check` variable needed? Can't we instead check the existing `sanity_component` and `fromJSON(needs.configure_test_matrix.outputs.sanity_component).test_runner`?

ScottTodd (approved):
Code changes LGTM % the workflow and comment details.

build_tools/print_driver_gpu_info.py
lumachad: Hmm, good catch. Let me try this again.

.github/workflows/test_artifacts.yml
lumachad: Might work. I'll check it.

build_tools/print_driver_gpu_info.py
lumachad: Updated now. I've made the KFD check gating as well (initially it was informational). I think it is fair to make it punt given it is obviously a broken setup (too old/not functional).

.github/workflows/test_artifacts.yml
lumachad: Done this now.

build_tools/print_driver_gpu_info.py
```diff
@@ -125,13 +150,33 @@ def run_sanity(os_name: str) -> None:
...
+            in_range = _KFD_VERSION_MIN <= (major, minor) < _KFD_VERSION_MAX
+            status = "supported" if in_range else "NOT supported"
+            log(f"KFD IOCTL version: {major}.{minor} ({status})")
+            log(
+                f"Required range for rocdbgapi: "
+                f">= {_KFD_VERSION_MIN[0]}.{_KFD_VERSION_MIN[1]}"
+                f" and < {_KFD_VERSION_MAX[0]}.{_KFD_VERSION_MAX[1]}"
+            )
+            if not in_range:
+                return 1
```
ScottTodd: Are out of range versions okay for components other than rocdbgapi?

If we update the kernel on test runners beyond `_KFD_VERSION_MAX`, do we really want all tests to start failing here?

ScottTodd (approved):

build_tools/print_driver_gpu_info.py
lumachad: That's what gives me a bit of pause. Technically everyone using kfd would need to pass this check, but the required versions might be slightly different (I think profilers also use kfd).

At the same time, I'm not sure we want to completely block testing for non-kfd users here. BUT, from a kfd user's perspective, that runner is useless and needs to be fixed. So maybe failing everything is an incentive to get the runner fixed?

As for having a newer version, that might be difficult to happen as we'd be on top of this. But that's a valid point. We might want to relax that a bit, at least for now.

ScottTodd: I don't have a particularly strong opinion either way, I think you're more tuned in to the tradeoffs here. Up to you. This code is easy enough to change later.

geomin12 (approved):
thanks! this is great

build_tools/print_driver_gpu_info.py
lumachad: I'm going with gating on /dev/kfd absence and lack of a minimal version. If the version is too new, we'll just warn.

lumachad:
Looks like CI is a bit flakey.

build_tools/print_driver_gpu_info.py
```diff
@@ -8,21 +8,47 @@
...
 
-This script prints only raw command output.
+Driver commands (amd-smi, rocminfo, hipInfo) are run with check=True and will
+cause this script to exit non-zero if they fail. The KFD version check also
+causes a non-zero exit when the version is below the supported minimum or
+cannot be queried. A version above the supported maximum produces a warning
+but does not fail.
 """
 
+import fcntl
```
ScottTodd, line 23: This import fails on Windows, and sanity checks failed on Windows on this PR's presubmit run: https://github.com/ROCm/TheRock/actions/runs/33499470761/job/99867709885.

Fixing in https://github.com/ROCm/TheRock/pull/7820

cc @amd-justchen , see https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0002-MonoRepo-Gardener-Rotations.md#responsibilities-and-priorities 

> ### Responsibilities and priorities
> 
> 1. When regressions occur that break CI workflows
>    - Communicate that something is broken
>    - Identify the culprit commit(s)
>    - Notify the authors/reviewers of those commits
>    - Revert those changes
>    - If it's not obvious how to, facilitate getting changes in to get back to green (see 2)

The general sequence is "comment on the PR about the break" --> "revert" --> "let the author fix"

ScottTodd: We can also add a unit test for this script that at least checks that it can be run on CPU runners without failing at import time (CPU runners won't have the tools, so those calls could be mocked out)

lumachad: These shouldn't be running on CPU runners at all.

ScottTodd: We missed this because the "Windows::release / Test gfx110X-all / Test Sanity Check / Test sanity (shard 1/1) (gfx110X-all)" job that failed ran 2 hours after the CI started and was not blocking. The "unit tests" job runs in < 5 minutes and is blocking, but it only runs on CPU (github-hosted) runners. A unit test that checked that the script could import on windows and linux would have caught this and blocked the merge.

lumachad: Ok, yeah that makes sense. I had the context wrong at first. Let me craft a PR for that.

PR #7373, opened 2026-08-14
ci: authenticate git clones via GITHUB_TOKEN and enforce step timeouts

kailash-khalasi:
## Summary

Hardens CI reliability by ensuring git clone operations can't hang indefinitely.

- Injects `GITHUB_TOKEN` into all git clones via `url.insteadOf` in workflows that call `fetch_sources.py`, reducing the risk of GitHub throttling on parallel submodule fetches (validated in #7354)
- Adds `timeout-minutes` to `actions/checkout` and `fetch_sources.py` steps in the 4 affected workflows
- Adds a unit test that enforces `timeout-minutes` on every `actions/checkout` and `fetch_sources.py` step across all workflows — failing the PR if any new step is added without one

## Why a custom unit test?

No off-the-shelf tool covers this gap:
- **actionlint** — validates timeout syntax but does not enforce its presence (open feature request, unimplemented)
- **zizmor** — security-focused, no timeout rules
- **`komiya-atsushi/action-enforce-timeout-minutes`** — job-level only, not step-level
- **Native GitHub rulesets** — can mandate which workflows run, but cannot inspect workflow structure

The custom test in `build_tools/github_actions/tests/workflow_step_timeouts_test.py` fills this gap and runs automatically in the existing `unit_tests` CI job.

## Note

The unit test currently reports 54 violations in other workflows not touched by this PR — those are emitted as warnings so CI passes. All violations are tracked and must be addressed in #7388. Once resolved, the warning will be reverted to a hard failure to restore full enforcement.

### DIFF
--- a/.github/workflows/build_portable_linux_python_packages.yml
+++ b/.github/workflows/build_portable_linux_python_packages.yml
@@ -103,7 +103,7 @@ jobs:
     permissions:
       id-token: write
     container:
-      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:a382085df3ba2419b58aa9051350883a0d0b732a4bc0a4ef60458f8161bb08c6
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb
     outputs:
       package_find_links_url: ${{ steps.upload.outputs.package_find_links_url }}
       kpack_split: ${{ steps.upload.outputs.kpack_split }}
--- a/.github/workflows/build_portable_linux_python_packages.yml
+++ b/.github/workflows/build_portable_linux_python_packages.yml
@@ -103,7 +103,7 @@ jobs:
     permissions:
       id-token: write
     container:
-      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb # 2026-08-18
     outputs:
       package_find_links_url: ${{ steps.upload.outputs.package_find_links_url }}
       kpack_split: ${{ steps.upload.outputs.kpack_split }}
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -84,7 +84,7 @@ jobs:
       contents: read
       id-token: write
     container:
-      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb # 2026-08-18
     env:
       QUARTZ_REPORTING_WORKFLOW: multi_arch_build_portable_linux_artifacts.yml
       QUARTZ_WORKFLOW_POSTFIX: "- build ROCm artifacts"
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -110,6 +110,7 @@ jobs:
           gh_app_private_key: ${{ secrets.GH_APP_HAULY_PRIVATE_KEY }}
 
       - name: Checkout Repository
+        timeout-minutes: 10
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: ${{ inputs.repository || github.repository }}
@@ -117,6 +118,7 @@ jobs:
 
       - name: Checkout external repository
         if: ${{ inputs.external_repo_config != '' }}
+        timeout-minutes: 10
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: ${{ fromJSON(inputs.external_repo_config).repository }}
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -130,6 +130,9 @@ jobs:
         run: |
           git config --global --add safe.directory $PWD
           git config fetch.parallel 10
+          git config --global \
+            url."https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/".insteadOf \
+            "https://github.com/"
 
       - name: Setup ccache
         run: |
--- a/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
+++ b/.github/workflows/multi_arch_build_portable_linux_artifacts.yml
@@ -132,9 +132,6 @@ jobs:
         run: |
           git config --global --add safe.directory $PWD
           git config fetch.parallel 10
-          git config --global \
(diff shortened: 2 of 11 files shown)


ndhillon07 (approved):

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
```diff
@@ -130,6 +132,9 @@ jobs:
         run: |
           git config --global --add safe.directory $PWD
           git config fetch.parallel 10
+          git config --global \
+            url."https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/".insteadOf \
+            "https://github.com/"
```
ScottTodd: Let's find a different pattern for this.

This breaks the LLVM build and for good reason: https://github.com/ROCm/TheRock/actions/runs/31848599692/job/94920071516?pr=7373#step:19:239
```

[amd-llvm] FAILED: include/llvm/Support/VCSRevision.h B:/build/compiler/amd-llvm/build/include/llvm/Support/VCSRevision.h 
[amd-llvm] C:\Windows\system32\cmd.exe /C "cd /D B:\build\compiler\amd-llvm\build\include\llvm\Support && "C:\Program Files\CMake\bin\cmake.exe" -DNAMES=LLVM -DLLVM_SOURCE_DIR=C:/home/runner/_work/TheRock/TheRock/compiler/amd-llvm/llvm -DHEADER_FILE=B:/build/compiler/amd-llvm/build/include/llvm/Support/VCSRevision.h -DLLVM_FORCE_VC_REVISION= -DLLVM_FORCE_VC_REPOSITORY= -P C:/home/runner/_work/TheRock/TheRock/compiler/amd-llvm/llvm/cmake/modules/GenerateVersionFromVCS.cmake"
[amd-llvm] CMake Error at C:/home/runner/_work/TheRock/TheRock/compiler/amd-llvm/llvm/cmake/modules/VersionFromVCS.cmake:48 (message):
[amd-llvm]   The git remote repository URL has an embedded password.  Remove the
[amd-llvm]   password from the URL or use `-DLLVM_FORCE_VC_REPOSITORY=<URL without
[amd-llvm]   password>` in order to avoid leaking your password (see
[amd-llvm]   https://git-scm.com/docs/gitcredentials for alternatives).
[amd-llvm] Call Stack (most recent call first):
[amd-llvm]   C:/home/runner/_work/TheRock/TheRock/compiler/amd-llvm/llvm/cmake/modules/GenerateVersionFromVCS.cmake:51 (get_source_info)
```

Full review: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7373.md

My agent suggests commands like `gh auth setup-git --hostname github.com` instead

build_tools/github_actions/tests/workflow_step_timeouts_test.py
```diff
@@ -0,0 +1,92 @@
...
+            errors.extend(_find_violations(workflow, workflow_path.name))
+
+        if errors:
+            import warnings
+
+            warnings.warn(
+                "The following checkout/fetch_sources steps are missing timeout-minutes "
+                "(fix in follow-up PRs):\n" + "\n".join(f"  - {e}" for e in errors),
+                stacklevel=2,
+            )
```
ScottTodd, line 159: Tests that only log are not useful, they should actually fail.

https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7373.md#blocking-the-timeout-test-warns-about-every-violation-and-fails-on-none

.github/workflows/multi_arch_build_portable_linux_artifacts.yml
kailash-khalasi: Thanks for the thorough review, Scott — both blockers addressed in 5fc964c.

**Auth:** Replaced the `url.insteadOf` rewrite with `gh auth setup-git --hostname github.com` scoped to the Fetch sources step via a step-level `GH_TOKEN` env var. The remote URL stays credential-free so LLVM's `VersionFromVCS.cmake` check won't trigger.

**Timeout test:** Replaced the blanket `warnings.warn` with an explicit `_KNOWN_VIOLATIONS` allowlist of the 53 pre-existing violations. Those still warn, but any new checkout/fetch_sources step added without a timeout now fails the test immediately — restoring the enforcement guarantee the PR description claimed.

ScottTodd: Looks like the Linux build container does not have `gh` :/

https://github.com/ROCm/TheRock/actions/runs/31850715001/job/94926102947?pr=7373#step:11:21
```
/__w/_temp/35ae680c-906a-4a96-bd02-0d272b7c3190.sh: line 1: gh: command not found
```

ScottTodd: Updated review: https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7373.md#blocking-portable-linux-requires-an-executable-that-its-container-does-not-provide

that suggested a scoped
```yml
env:
  GIT_CONFIG_COUNT: 1
  GIT_CONFIG_KEY_0: url.https://x-access-token:${{ github.token }}@github.com/.insteadOf
  GIT_CONFIG_VALUE_0: https://github.com/
```
that would not apply during the build step, just our own targeted checkout steps

I'm not thrilled about that but it might be worth it.

We should probably check online for other suggestions.

kailash-khalasi: I think it's crucial we get gh installed in the manylinux image. I will work on that and update these workflows with the new image.

ScottTodd: Sure, that's reasonable too.

kailash-khalasi: PR for gh cli in manylinux image - https://github.com/ROCm/TheRock/pull/7394

build_tools/github_actions/tests/workflow_step_timeouts_test.py
kailash-khalasi: The current implementation does call self.fail() for any new violations — the warning path is only for the pre-existing debt captured in _KNOWN_VIOLATIONS. Those are tracked for a follow-up in #7388. Would you prefer we remove the warning log for those entirely and just silently skip them?

ScottTodd: I think the warning logs are fine, we can fix those uses shortly after merging this work and then consider dropping the warning code to just make it consistently error on any new steps without timeouts.

See https://github.com/ROCm/TheRock/actions/runs/32081785990/job/95546039870?pr=7373#step:5:1884 where the warnings appear in the unit tests workflow logs.

(I can do a more complete review tomorrow)

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -103,7 +103,7 @@ jobs:
     permissions:
       id-token: write
     container:
-      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:a382085df3ba2419b58aa9051350883a0d0b732a4bc0a4ef60458f8161bb08c6
+      image: ghcr.io/rocm/therock_build_manylinux_x86_64@sha256:cf4f6d9909056906e4a1c0bc4c245658a0af4c6a6174082a86adfa2d518b7aeb
```
kailash-khalasi: @ScottTodd this new image has the gh cli binary. I've updated all the workflows in this repo. Do we need to update in the submodule repos too like rocm-systems?

ScottTodd: We should, yes:
* https://github.com/ROCm/rocm-libraries/blob/2a35abf66cea99024defe125d4e91fc49d35f2c3/.github/actions/ci-env/action.yml#L8-L10
* https://github.com/ROCm/rocm-systems/blob/dfd3f592c67a53fe1affcac95e45eeee6da136bb/.github/workflows/therock-rccl-ci-linux.yml#L44-L45
* https://github.com/ROCm/rocm-systems/blob/dfd3f592c67a53fe1affcac95e45eeee6da136bb/.github/workflows/therock-ci-linux.yml#L35-L36
* etc.

ScottTodd (approved):
Build jobs on https://github.com/ROCm/TheRock/actions/runs/32178932561?pr=7373 are successful so far (a few unrelated infra / github flakes that we're seeing on other runs)

PR #7378, opened 2026-08-14
[profiler] Advance rocm-systems pin to profiler-hub packaging and ASAN fixes

avansick-amd:
## JIRA ID

JIRA ID : ROCPDSNA-73

## Motivation

The profiler-hub build and packaging work above this PR needs two fixes that live in
rocm-systems, not in TheRock. This PR moves the `rocm-systems` submodule pointer forward to pick
them up, and does nothing else, so that the pointer move is a deliberate reviewable act rather
than incidental drift folded into a functional change.

Stacked on #7377 for a linear review order; it shares no files with that PR.

This PR is not mergeable as it stands, and is deliberately a draft. See the pin status below.

## Technical Details

- Moves the `rocm-systems` gitlink from `af82d0abb401f1e0bccebdb69d5053371507b6b5` to
  `a181b145190f0b9a73ea8969f34924c1e6e64ab6`. That is the entire diff: one line, one file.
- The target commit carries two things the stack needs. First, the profiler-hub installed CMake
  package config uses a conditional `find_package(spdlog QUIET)` instead of a fatal
  `find_dependency`, so an external project can call
  `find_package(profiler-hub CONFIG REQUIRED)` against an installed tree without spdlog being
  independently discoverable. Second, the profiler-hub unit tests carry the ASAN build fixes for
  the amd-llvm toolchain. Both come from ROCm/rocm-systems#9367.

Pin status, and why this is a draft:

- `a181b145190` is not reachable from rocm-systems `develop`. It sits on
  `users/avansick-amd/6922-rocm-systems-integration`, which merges #9367. A submodule pin must
  point at a landed upstream SHA before it can merge; this one does not yet, because no such
  commit exists until #9367 lands.
- ROCm/rocm-systems#9367 is open and awaiting approval. Once it merges to `develop`, this pin
  must be moved forward to a develop-reachable commit that contains it, and only then can this
  PR come out of draft. Re-pinning is tracked as separate follow-up work.
- Reviewers can review the intent of the move now; please do not merge it in its current state.

## Test Plan

- Confirm the diff against the PR below it in the stack is exactly the `rocm-systems` gitlink and
  nothing else.
- Confirm the target commit is present on the rocm-systems remote, so the gitlink resolves for
  anything that checks out this branch.
- Confirm the reachability claim above rather than asserting it.

## Test Result

- `git diff <base>..HEAD --name-only` returns exactly `rocm-systems`; the numstat is 1 insertion,
  1 deletion.
- `git ls-remote https://github.com/ROCm/rocm-systems.git` shows `a181b145190` at
  `refs/heads/users/avansick-amd/6922-rocm-systems-integration`, so the gitlink resolves against
  the remote.
- The same commit is not reachable from `develop`, which is exactly the condition recorded above
  and the reason this PR is a draft.

CI results pending on this PR. Note that TheRock CI resolves submodules with
`build_tools/fetch_sources.py --depth 1`, which fetches the recorded gitlink rather than a branch
tip; whether a shallow fetch resolves a non-tip SHA on this remote is the first thing this PR's
CI will establish.

## Submission Checklist

- [x] Looked over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.
- [x] The submodule pointer move is isolated in its own PR and is the only change.
- [x] The pin is disclosed as not develop-reachable, with the condition for resolving it stated.
- [x] Opened as a draft because it cannot merge in its current state.
- [x] Shares no files with any other PR in the stack.

### DIFF
--- a/BUILD_TOPOLOGY.toml
+++ b/BUILD_TOPOLOGY.toml
@@ -794,6 +794,14 @@ artifact_deps = ["rocprofiler-sdk", "openmpi"]
 feature_group = "PROFILER"
 disable_platforms = ["windows"]
 
+[artifacts.profiler-hub]
+artifact_group = "profiler-core"
+type = "target-neutral"
+artifact_deps = ["core-runtime", "base", "rocprofiler-sdk", "spdlog", "nlohmann-json"]
+feature_name = "PROFILER_HUB"
+feature_group = "PROFILER"
+disable_platforms = ["windows"]
+
 [artifacts.rocprofiler-systems]
 artifact_group = "profiler-apps"
 type = "target-neutral"
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -621,63 +621,56 @@ def run(args: argparse.Namespace):
         print("Error: No packages found to build. Package list is empty.")
         sys.exit(1)
 
-    current_pkg_idx = 0
-    try:
-        built_pkglist = []
-        failed_pkglist = []
+    built_pkglist = []
+    failed_pkglist = []
 
-        for current_pkg_idx, pkg_name in enumerate(pkg_list):
-            print(f"Creating {config.pkg_type} package: {pkg_name}")
+    for current_pkg_idx, pkg_name in enumerate(pkg_list):
+        print(f"Creating {config.pkg_type} package: {pkg_name}")
 
+        try:
             # Build all package variants for this package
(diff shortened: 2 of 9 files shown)


avansick-amd:
The rocm-systems pin on main has advanced to ef3b147376c6, which already contains ROCm/rocm-systems#9367, so this manual bump is no longer needed. #7379 has been rebased onto #7377 and now carries that pin through the stack. Closing without merging; leaving the branch in place for provenance.

PR #7380, opened 2026-08-14
 fix(feat): Add parallel package building with ProcessPoolExecutor

raramakr:
- Introduce -j/--parallel flag to control concurrent builds (defaults to CPU count)
   - Use ProcessPoolExecutor to build packages in parallel processes
   - Isolate each package build in a temporary directory to avoid conflicts
   - Capture per-package logs separately for easier debugging
   - Handle SystemExit and exceptions gracefully in worker processes
   - Fix RPM log message to show package name instead of 'specfile'

   This significantly reduces build time for multi-package runs by utilizing
   all available CPU cores.
   
   JIRA ID: https://github.com/ROCm/TheRock/issues/7381

### DIFF
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -812,8 +812,8 @@ def main(argv: list[str]):
         "-j",
         "--parallel",
         type=int,
-        default=os.cpu_count() or 1,
-        help="Number of parallel package builds (default: all available cores)",
+        default=min(8, os.cpu_count() or 1),
+        help="Number of parallel package builds (default: min(8, cpu_count))",
     )
 
     args = p.parse_args(argv)
--- a/build_tools/packaging/linux/build_package.py
+++ b/build_tools/packaging/linux/build_package.py
@@ -44,6 +44,7 @@ import tempfile
 from concurrent.futures import ProcessPoolExecutor, as_completed
 from dataclasses import replace
 from pathlib import Path
+from typing import NamedTuple
 
 # Setup paths
 SCRIPT_DIR = Path(__file__).resolve().parent
@@ -68,6 +69,14 @@ logger = TheRockLogger(__name__)
 
 # Default install prefix
 DEFAULT_INSTALL_PREFIX = "/opt/rocm/core"
+
+
+class BuildTask(NamedTuple):
+    """Task for parallel package building via ProcessPoolExecutor."""
+
+    pkg_name: str
+    config: PackageConfig
+    logs_dir: Path
 
 
 def load_kpack_from_manifest(artifacts_dir: Path) -> bool:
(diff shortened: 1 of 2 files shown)


build_tools/packaging/linux/build_package.py
```diff
@@ -768,6 +777,14 @@ def main(argv: list[str]):
         help="Specify the packages to be created",
     )
 
+    p.add_argument(
+        "-j",
+        "--parallel",
+        type=int,
+        default=os.cpu_count() or 1,
+        help="Number of parallel package builds (default: all available cores)",
+    )
```
ScottTodd, line 827: Out of curiosity, how many package builds will be running in parallel? We might want to limit if using all cores (up to 100+ on some machines) oversubscribes on disk I/O, RAM, etc.

See https://github.com/ROCm/TheRock/pull/4448 for example:

> I did wrap compression in a `ProcessPoolExecutor` since parallel compression does make efficient use of CPU cores, sample benchmarks showing speedup (so not oversubscribed):
> 
> Workers |  Wall (s) |   Avg/job | Speedup | Efficiency
> -------- | -----------| ----------- | --------- | ---------------
> 1 |     244.2 |      24.4 |    1.0x |      103%
> 2 |     128.3 |      25.6 |    2.0x |       98%
> 4 |      79.4 |      26.6 |    3.2x |       79%
> 6 |      54.4 |      27.2 |    4.6x |       77%
> 8 |      54.0 |      27.6 |    4.7x |       58%
> 10 |      28.8 |      28.6 |    8.8x |       88%

ScottTodd (commented):
Nice!

build_tools/packaging/linux/build_package.py
raramakr: We have currently 83 packages. So a maximum of 83 parallel builds  if that much cores are available.
Any recommendation on the number of cpus to be used?  os.cpu_count() /2 ?

ScottTodd: It may depend on available RAM, disk I/O speeds, _and_ CPU core count. I would run some benchmarks.

raramakr:
[benchmark_summary.txt](https://github.com/user-attachments/files/31201340/benchmark_summary.txt)

build_tools/packaging/linux/build_package.py
raramakr: Attached https://github.com/user-attachments/files/31201340/benchmark_summary.txt
Based on the that , will be 

Default to `min(8, os.cpu_count() or 1)` — provides ~4x speedup with >50% efficiency while avoiding resource waste from idle workers.

nunnikri (approved):
Changes looks good. Will have to add unit test for these fns as a follow up change

PR #5942, opened 2026-06-17
Generate dynamic matrix for ROCm python package tests in CI

ScottTodd:
## Motivation

Follow-up to https://github.com/ROCm/TheRock/pull/5731#discussion_r3404708070

We had been generating matrix entries for every built GPU family and then filtering in the child workflow:
* https://github.com/ROCm/TheRock/blob/2d417fa49bc4e657bcbcf0ae549238784175f0fe/.github/workflows/multi_arch_ci_linux.yml#L260-L261 https://github.com/ROCm/TheRock/blob/2d417fa49bc4e657bcbcf0ae549238784175f0fe/.github/workflows/multi_arch_ci_linux.yml#L270-L272
* https://github.com/ROCm/TheRock/blob/2d417fa49bc4e657bcbcf0ae549238784175f0fe/.github/workflows/test_rocm_wheels.yml#L111-L114

This cluttered the UI: 

image
image

This moves towards expanding the full matrix for the `test_python_packages_per_family` job in `build_tools/github_actions/configure_multi_arch_ci.py` so we only add jobs to the graph that we expect to run.

> [!IMPORTANT]
> I'd like to do the same thing for all other jobs (e.g. `build_pytorch` for CI using [`build_tools/github_actions/configure_pytorch_release_matrix.py`](https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_pytorch_release_matrix.py)), as this will also let us make the matrix generation dynamic based on included GPU targets and workflow run type (PR, release build, opt-in labels applied, etc.).
>
> We can put enough details in the logs and https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci_summary.py to explain what the CI runs and why.

## Test Plan

* Linux just gfx942: https://github.com/ROCm/TheRock/actions/runs/27723233016
* Windows just gfx101x (no test runner): https://github.com/ROCm/TheRock/actions/runs/27723318428
* Windows gfx101x (no test runner) and gfx1151 (`nightly_check_only_for_family`): https://github.com/ROCm/TheRock/actions/runs/27723794134
* Watch jobs on this PR (should see the same running jobs as on the baseline, without the extra skipped jobs) --> https://github.com/ROCm/TheRock/actions/runs/27724949201/job/82034179186?pr=5942
    image

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #5731, opened 2026-06-09
Expand Linux Python wheel test matrix to include python 3.10 and 3.11

ScottTodd:
## Motivation

We recently had test failures that were only caught by the time we also ran PyTorch tests across multiple Python versions:
* https://github.com/ROCm/TheRock/issues/5516

This expands presubmit coverage to also include Python 3.10 and 3.11 in addition to Python 3.12.

See also:

* https://github.com/ROCm/TheRock/issues/1559

## Technical Details

> [!IMPORTANT]
> This is expected to add 5-10 minutes of test time **on each GPU runner type** for every workflow run.

We can choose to conditionally enable these additional jobs to limit how many workflow runs include them. References:
* https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/configure_multi_arch_ci.py
* https://github.com/ROCm/TheRock/blob/main/docs/development/ci_behavior_manipulation.md

I'm limiting to just Linux for now. We could also expand coverage on Windows.

## Test Plan

Watch CI on this PR

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #4297, opened 2026-04-01
[ci] Run rocm-sdk wheel tests sequentially

lucbruni-amd:
Prerequisite of https://github.com/ROCm/TheRock/pull/4248

## Motivation

Using one job for both package stages keeps the matrix dimension free for container images later, so extra base images do not multiply jobs across libraries vs. devel. See https://github.com/ROCm/TheRock/pull/3119#discussion_r2736945862 for more details.

## Technical Details

Replace the `matrix.packages` job split with one job that installs `rocm[libraries]`, runs `rocm-sdk test`, then upgrades the same venv to `rocm[libraries,devel]` and runs `rocm-sdk test` again via `setup_venv` reuse: https://github.com/ROCm/TheRock/blob/76ad42a633edc612ae74c771fde169c1c2686263/build_tools/setup_venv.py#L80

## Test Plan

Observe PR action passed with sequential installation/testing enabled: https://github.com/ROCm/TheRock/actions/runs/24152917536/job/70519071421

## Test Result

Now conditionally installs/tests `devel` package if `libraries` was successfully installed.

Example of a run which fails during testing of `rocm[libraries]` and continues to install/test `rocm[libraries,devel]` as desired: https://github.com/ROCm/TheRock/actions/runs/24411710011/job/71344455349

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #3261, opened 2026-02-05
Wire ROCm Python package tests through ci.yml workflow paths

ScottTodd:
## Motivation

Fixes https://github.com/ROCm/TheRock/issues/1559. Now our CI workflows that run on `pull_request` and `push` triggers will
1. Upload the Python packages that they build to the same `therock-ci-artifacts` bucket where they already upload artifacts, artifact build logs, and other files.
2. Test those Python packages by installing them on GPU test runners and running `rocm-sdk test`.

This is a part of the broader "shift left testing" work tracked on https://github.com/ROCm/TheRock/issues/3177.

> [!IMPORTANT]
> There have been at least 5 PRs/commits in the past two weeks that produced test failures which this workflow code would have caught earlier if it were in place.

## Technical Details

~~The workflow code here is modeled after how we build and test PyTorch in [`build_portable_linux_pytorch_wheels.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_portable_linux_pytorch_wheels.yml). Future workflow refactoring of the CI and release workflows may bring that back into closer alignment with how ci.yml starts the workflow pipeline with a run of [`setup.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/setup.yml).~~

Workflow code is structured to match the setup in [`ci.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/ci.yml). I would also like to migrate the release workflows like [`release_portable_linux_packages.yml`](https://github.com/ROCm/TheRock/blob/main/.github/workflows/release_portable_linux_packages.yml) to use the same pipelined approach to building and testing packages in the future.

## Test Plan

* I've tested the underlying workflows in various ways over the past two weeks on both test commits and commits that materially impact (and have broken) the python package tests:
  * https://github.com/ROCm/TheRock/actions/workflows/build_portable_linux_python_packages.yml
  * https://github.com/ROCm/TheRock/actions/workflows/test_rocm_wheels.yml
* CI jobs on this PR should run the new code
  * https://github.com/ROCm/TheRock/actions/runs/21693735684?pr=3261 (failures are due to https://github.com/ROCm/TheRock/issues/3268 and should go away after syncing this branch)

## Test Results

On this PR's https://github.com/ROCm/TheRock/actions/runs/21693735684?pr=3261:

1. Python packages uploads are next to other uploads: 
    group | file | link
    -- | -- | --
[...]

PR #7383, opened 2026-08-14
Test change for test_rocm_wheels enablement in rockrel 

jonatluu:
## Motivation

Corresponding change to Therock so that wrapper test_rocm_wheels.yml for rockrel can be called https://github.com/ROCm/rockrel/pull/84

Issue: https://github.com/ROCm/rockrel/issues/82

## Technical Details

modified .github/workflows/multi_arch_release_linux.yml to call 

## Test Plan

Test ROCm Wheels: https://github.com/ROCm/TheRock/actions/runs/35660047381

Multi-Arch Release: https://github.com/ROCm/TheRock/actions/runs/35660197014

## Submission Checklist

### DIFF
--- a/.github/workflows/multi_arch_release_linux.yml
+++ b/.github/workflows/multi_arch_release_linux.yml
@@ -250,6 +250,34 @@ jobs:
               "ref": "${{ inputs.ref }}"
             }
 
+  # Wheel tests are dispatched asynchronously (like native install tests) so
+  # publish is not blocked on GPU runner queues. See https://github.com/ROCm/TheRock/issues/5212
+  trigger_test_wheels_per_family:
+    needs: [build_python_packages]
+    name: Trigger Test Wheels (${{ matrix.amdgpu_family }}, py${{ matrix.python_version }})
+    if: ${{ toJSON(fromJSON(inputs.build_config).test_python_packages_matrix) != '[]' }}
+    runs-on: ubuntu-24.04
+    permissions:
+      actions: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJSON(inputs.build_config).test_python_packages_matrix }}
+    steps:
+      - name: Trigger test_rocm_wheels
+        uses: benc-uk/workflow-dispatch@31e2b3319479a63f0ab15bf800eff9e913504e26 # v1.3.2
+        with:
+          workflow: test_rocm_wheels.yml
+          inputs: |
+            { "amdgpu_family": "${{ matrix.amdgpu_family }}",
+              "test_runs_on": "${{ matrix.test_runs_on }}",
+              "package_find_links_url": "${{ needs.build_python_packages.outputs.kpack_split == 'true' && needs.build_python_packages.outputs.package_find_links_url || format('{0}/{1}/index.html', needs.build_python_packages.outputs.package_find_links_url, matrix.amdgpu_family) }}",
+              "python_version": "${{ matrix.python_version }}",
+              "rocm_version": "${{ inputs.rocm_package_version }}",
+              "container_image_name": "${{ matrix.container_image_name }}",
+              "container_image_url": "${{ matrix.container_image_url }}",
+              "kpack_split": "${{ needs.build_python_packages.outputs.kpack_split }}"
+            }
 
   # TODO: Consider switch pulling ROCm packages from the CDN instead of the artifact
   #   bucket if the job order is not altered.
@@ -300,7 +328,7 @@ jobs:
 
   notify_quartz_completed:
     if: ${{ always() && vars.NOTIFY_QUARTZ_ENABLED == 'true' }}
-    needs: [build_artifacts, build_tarballs, build_python_packages, build_native_deb_packages, build_native_rpm_packages, publish_to_release_buckets, trigger_test_artifacts_per_family, trigger_test_native_install, trigger_release_pytorch_wheels, trigger_release_jax_wheels]
+    needs: [build_artifacts, build_tarballs, build_python_packages, build_native_deb_packages, build_native_rpm_packages, publish_to_release_buckets, trigger_test_artifacts_per_family, trigger_test_native_install, trigger_test_wheels_per_family, trigger_release_pytorch_wheels, trigger_release_jax_wheels]
     name: "Quartz - completed - release Linux"
     runs-on: ubuntu-24.04
     timeout-minutes: 10


.github/workflows/multi_arch_release_linux.yml
ScottTodd, line 1: Always keep Linux and Windows workflows in sync: https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_release_windows.yml

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -250,6 +250,34 @@ jobs:
...
+          inputs: |
+            { "amdgpu_family": "${{ matrix.amdgpu_family }}",
+              "test_runs_on": "${{ matrix.test_runs_on }}",
+              "package_find_links_url": "${{ needs.build_python_packages.outputs.kpack_split == 'true' && needs.build_python_packages.outputs.package_find_links_url || format('{0}/{1}/index.html', needs.build_python_packages.outputs.package_find_links_url, matrix.amdgpu_family) }}",
+              "python_version": "${{ matrix.python_version }}",
+              "rocm_version": "${{ inputs.rocm_package_version }}",
+              "container_image_name": "${{ matrix.container_image_name }}",
+              "container_image_url": "${{ matrix.container_image_url }}",
+              "kpack_split": "${{ needs.build_python_packages.outputs.kpack_split }}"
+            }
```
ScottTodd, line 280: Ah this isn't great - the matrix is going to dispatch lots of jobs here.

I would rather have releases dispatch closer to what test_artifacts does: one workflow run per amdgpu_family, not per [amdgpu_family, container, python_version].

Need to look at the workflow architecture a bit to see how we might simplify there.

ScottTodd: I haven't had much time to take a more detailed look here yet.

One idea would be to add a new workflow that takes `fromJSON(inputs.build_config).test_python_packages_matrix` as an input and have that workflow expand the matrix itself.

CI would run that new workflow here: https://github.com/ROCm/TheRock/blob/4444ac2bf896097776d65ff6a43dbc78cb533894/.github/workflows/multi_arch_ci_linux.yml#L285-L313

We've gone through a few iterations on the workflow matrix shape, see also 
* https://github.com/ROCm/TheRock/pull/3261
* https://github.com/ROCm/TheRock/pull/4297
* https://github.com/ROCm/TheRock/pull/5731
* https://github.com/ROCm/TheRock/pull/5942

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -275,6 +275,52 @@ jobs:
...
+        toJSON(fromJSON(inputs.build_config).test_python_packages_matrix)
+        != '[]'
+      }}
+    strategy:
+      fail-fast: false
+      matrix:
+        family_info: ${{ fromJSON(inputs.build_config).per_family_info }}
+
+    uses: ./.github/workflows/test_python_packages_per_family.yml
+    secrets: inherit
```
ScottTodd, line 296: We're going to start enforcing security scanner findings soon, so please fix this

https://github.com/ROCm/TheRock/actions/runs/35771641728/job/106894719298?pr=7383#step:6:302

```
warning[secrets-inherit]: secrets unconditionally inherited by called workflow
   --> .scan-target/.github/workflows/multi_arch_release_linux.yml:303:11
    |
303 |     uses: ./.github/workflows/test_python_packages_per_family.yml
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this reusable workflow
304 |     secrets: inherit
    |     ---------------- inherits all parent secrets
    |
    = note: audit confidence → High
    = help: audit documentation → https://docs.zizmor.sh/audits/#secrets-inherit
```

Can follow the examples from
* https://github.com/ROCm/TheRock/pull/8405

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -275,6 +275,52 @@ jobs:
...
+          needs.build_python_packages.outputs.kpack_split == 'true'
+          &&
+          needs.build_python_packages.outputs.package_find_links_url
+          ||
+          format(
+            '{0}/{1}/index.html',
+            needs.build_python_packages.outputs.package_find_links_url,
+            matrix.family_info.amdgpu_family
+          )
+        }}
```
ScottTodd, line 319: Please follow the same formatting used elsewhere in the project for code like this: https://github.com/ROCm/TheRock/blob/e0e238c0323f76aeabdd348cfd3798a653820fdc/.github/workflows/multi_arch_ci_linux.yml#L323-L330

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -684,6 +684,7 @@ class BuildConfig:
     build_pytorch: bool
     build_jax: bool
     test_python_packages_matrix: list[dict[str, str]] = field(default_factory=list)
+    test_matrix_by_family: dict[str, list[dict[str, str]]] = field(default_factory=dict)
```
ScottTodd, line 687: If `test_matrix_by_family` is specific to python package testing, it should include "python" in the name. We have ROCm artifact tests, ROCm python package tests, ROCm native package tests, PyTorch tests, etc. so each variable should qualify what it is referring to.

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd, line 1506: `build_rocm_python_test_matrix()` should only be called once. This calls it twice and always computes `test_matrix_by_family` even if python packages are not being built.

(Separately we should remove the  `not is_asan` code here and rely on just `ci_inputs.build_python_packages` to control whether they are enabled)

.github/workflows/multi_arch_release_linux.yml
ScottTodd: Please respond to comments about architecture when making architectural changes and don't just resolve/push/re-request review. I can guess from what you changed but the review will be easier if you engage with the discussion here.

I would prefer to CI and release workflows to use the same code, perhaps by redesigning `.github/workflows/test_rocm_wheels.yml` along the lines suggested in this comment. The direction the PR currently takes adds a different workflow file and input that is unused by CI (`test_matrix_by_family`).

ScottTodd (commented):
> ## Test Plan
> Test ROCm Wheels: https://github.com/ROCm/TheRock/actions/runs/35772363614
> Multi-Arch Release: https://github.com/ROCm/TheRock/actions/runs/35660197014

These runs are not up to date with the latest code? I want to see how the generated matrix is working.

.github/workflows/multi_arch_release_linux.yml
```diff
@@ -275,6 +275,52 @@ jobs:
...
+        !cancelled() &&
+        toJSON(fromJSON(inputs.build_config).test_python_packages_matrix)
+        != '[]'
+      }}
+    strategy:
+      fail-fast: false
+      matrix:
+        family_info: ${{ fromJSON(inputs.build_config).per_family_info }}
+
+    uses: ./.github/workflows/test_python_packages_per_family.yml
```
ScottTodd, line 295: This is not "triggering", it is running directly. Triggering should use `workflow_dispatch`. Please change to trigger/dispatch as the other test jobs do (releases should not use GPU runners in the entry point workflow run, CI uses `workflow_call` from a single workflow run).

.github/workflows/multi_arch_release_linux.yml
ScottTodd: I think we could change https://github.com/ROCm/TheRock/blob/main/.github/workflows/multi_arch_ci_linux.yml to call the new `test_python_packages_per_family.yml` instead of having its own `strategy: matrix` and this would be pretty close. Might want to pick some different names  and consolidate on a single "matrix" variable instead of plumbing through both `test_python_packages_matrix` and `test_matrix_by_family` in that case?

(and do the same on Windows)

Some other names to consider:

Single configuration | Matrix wrapper
-- | --
test_python_packages_single.yml |   test_python_packages.yml
test_python_packages.yml        |   test_python_packages_matrix.yml
test_python_packages_single.yml |   test_python_packages_per_family.yml
test_rocm_wheels.yml            |   test_rocm_wheels_matrix.yml

I think we could start with `test_rocm_wheels_matrix.yml` to keep it simple.

Issue #7320, opened 2026-08-13
[Issue] JAX nightly test jobs fail across all Python/JAX versions —` run_pytest_rocm.sh` has no section between ENV_SECTION_START/END


TriveniTadapaneni:
## Context
  - **Workflow:** `ROCm/rockrel` `.github/workflows/multi_arch_release_linux_jax_wheels.yml` → `build_jax_wheels` → `Test JAX`
  - **Failing run:** https://github.com/ROCm/rockrel/actions/runs/31660008484/job/94334671803
    (same failure repeats identically across every job in the run)
  - **Affected JAX versions:** `0.10.0`, `0.10.1`, `0.10.2`, `0.11.0` 
  - **Affected Python versions:** `3.11`, `3.12`, `3.13`, `3.14`

  ## Summary
  Every nightly `Test JAX` job across all Python/JAX version combinations fails at the "Run JAX tests" step with the same error.
  `ci/run_pytest_rocm.sh` no longer has any content between the `# Set up the generic test environment variables` and `# Run tests` marker comments, so TheRock's `run_jax_tests.py` can't reconstruct the environment for the serial retry pass and hard-fails by design rather than guess (a wrong environment there would decide the run's result).

  ## Error:
```
  Run python external-builds/jax/run_jax_tests.py \
  === Resources
    KUBE_CPU_REQUEST=25
    KUBE_MEMORY_REQUEST=78643200000
    cpus visible to this process=192
    MemTotal:       2377684840 kB (the node's, whatever this container was given)
    pinned to 25 cpus: 0-24
  Error: ci/run_pytest_rocm.sh has no section between '# Set up the generic test environment variables' and '# Run tests'. The script
  most likely renamed them, so update ENV_SECTION_START and ENV_SECTION_END to match it.
  Error: Process completed with exit code 1.

  ```
  ## Suspected Root cause
[...]

PR #7387, opened 2026-08-14
ci(jax): read the suite environment from where it lives now

gulsumgudukbay:
## What breaks today

Every JAX nightly test job fails before it runs a single test:

Error: ci/run_pytest_rocm.sh has no section between '# Set up the generic test environment variables' and '# Run tests'.

`run_jax_tests.py` evaluates the suite's environment so that the serial retry pass runs under the same settings as the run it is checking. It read that environment out of a marked section of `ci/run_pytest_rocm.sh`. ROCm/jax has since moved it into `ci/utilities/rocm_test_env.sh`, and every release branch from `rocm-jaxlib-v0.10.0` through `rocm-jaxlib-v0.11.0` now ships the file with no section left behind, so the lookup fails on all of them.

## The change

Source `ci/utilities/rocm_test_env.sh` when the checkout has it, and fall back to the section when it does not, since an older `--jax-ref` is still valid. The file only exports, echoes and queries devices, so sourcing it runs nothing.

Split out of #6853 so the nightly is unblocked without waiting on that stack.

Solves https://github.com/ROCm/TheRock/issues/7320

## Test plan

- [x] `pytest build_tools/github_actions/tests/run_jax_tests_test.py` (77 passed)
- [x] New coverage for the split layout, and the error now names both places to fix
- [x] Nightly JAX test job gets past the environment step

### DIFF
--- a/external-builds/jax/run_jax_tests.py
+++ b/external-builds/jax/run_jax_tests.py
@@ -104,11 +104,15 @@ REPORT_GLOB = "logs/pytest_results*.json"
 PYTEST_EXIT_ALL_PASSED = 0
 PYTEST_EXIT_TESTS_FAILED = 1
 
-# The section of the suite script that only computes the environment: exports,
-# echoes and device queries, with the installs above it and the test run below it.
-# Evaluating it is how the per-version environment reaches the retry pass, which
-# has to match the run it is checking. Sourcing the script itself would run the
-# tests, and copying its values here would be a second source of truth.
+# What the suite script sources for its environment: exports, echoes and device
+# queries, and nothing that installs or tests. Evaluating it is how the
+# per-version environment reaches the retry pass, which has to match the run it
+# is checking. Sourcing the suite script itself would run the tests, and copying
+# its values here would be a second source of truth.
+RELATIVE_SUITE_ENV_SCRIPT = Path("ci") / "utilities" / "rocm_test_env.sh"
+
+# Where that environment lived before it moved into the file above, as a section
+# of the suite script, with the installs above it and the test run below it.
 ENV_SECTION_START = "# Set up the generic test environment variables"
 ENV_SECTION_END = "# Run tests"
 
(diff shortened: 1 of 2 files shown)


gulsumgudukbay:
https://github.com/ROCm/TheRock/actions/runs/31863692105 release workflow

gulsumgudukbay:
newest release workflow: https://github.com/ROCm/TheRock/actions/runs/31985833035

external-builds/jax/run_jax_tests.py
```diff
@@ -209,25 +213,45 @@ def _read_env_dump(path: Path) -> dict[str, str]:
...
 
     Raises:
-        SuiteEnvironmentError: if the section cannot be found or evaluated. This
-            fails the run by design, because the retry pass decides the result and
-            a wrong environment there would decide it wrongly.
+        SuiteEnvironmentError: if the checkout holds neither.
     """
-    section = env_section((jax_dir / RELATIVE_SUITE_SCRIPT).read_text())
+    if (jax_dir / RELATIVE_SUITE_ENV_SCRIPT).exists():
+        return f"source {shlex.quote(os.fspath(RELATIVE_SUITE_ENV_SCRIPT))}"
```
erman-gurses: Nit: Since this can now source ci/utilities/rocm_test_env.sh, can you make the diagnostics in suite_environment() source-agnostic? They still say “the section of ci/run_pytest_rocm.sh” if evaluation fails, which would point at the wrong file when this path is used.

external-builds/jax/run_jax_tests.py
```diff
@@ -209,25 +213,45 @@ def _read_env_dump(path: Path) -> dict[str, str]:
...
+    Evaluated in a shell that dumps its environment on either side, so the result
+    is what the suite itself changed rather than a guess at which names look
+    interesting.
+
+    Raises:
+        SuiteEnvironmentError: if it cannot be found or evaluated. This fails the
+            run by design, because the retry pass decides the result and a wrong
+            environment there would decide it wrongly.
+    """
+    section = suite_env_program(jax_dir)
```
erman-gurses: Nit: The module docstring still describes the environment as being read from the marked section of run_pytest_rocm.sh. I think we need to update it to mention `rocm_test_env.sh` as the preferred source and the marked section as the legacy fallback?

erman-gurses (approved):
LGTM -  I pointed some nits.

external-builds/jax/run_jax_tests.py
gulsumgudukbay: fxed in b968469: suite_env_program now returns the file it picked alongside the program, so all three messages name that file rather than the suite script, covered by a new test that fails a rocm_test_env.sh and asserts the error names it.

external-builds/jax/run_jax_tests.py
gulsumgudukbay: updated in b968469: the docstring now names ci/utilities/rocm_test_env.sh as where the environment is read from, with the marked section as what a checkout from before the move has, and says a checkout holding neither fails with an error naming both.

gulsumgudukbay:
release workflow on b968469: https://github.com/ROCm/TheRock/actions/runs/32069712859

gulsumgudukbay:
release workflow on the nightly packages the HUD runs against (10.1.0a20260818): https://github.com/ROCm/TheRock/actions/runs/32166904957

Issue #7183, opened 2026-08-07
[Issue]: pip --upgrade downgrades torch from rocm10.x to rocm7.x due to local-version string sort


0xDELUXA:
### Problem Description

When both a rocm7.x and a rocm10.x nightly series are present on https://rocm.nightlies.amd.com/whl-multi-arch/, running an unpinned `--upgrade` install actively downgrades an already-installed rocm10.x torch build back to rocm7.x.

PEP 440 local-version segments are split on `.` and compared component-by-component, as integers when purely numeric and as strings otherwise. Because the version strings are built as `rocm{major}.{minor}.{patch}a{date}` with no separator between `rocm` and the major version digits, the first local-version component is e.g. `rocm10` vs `rocm7` - neither is purely numeric, so pip compares them as strings:

    "rocm10" < "rocm7"   (lexicographic: '1' < '7')

So pip's resolver believes `+rocm7.15.0a20260728` is newer than `+rocm10.1.0a20260806`, even though 10 > 7 numerically and the rocm7 build is roughly 10 days older by date.

`--upgrade` should never select an objectively older/lower ROCm build over a newer one already installed.

### Operating System

Windows 11

### CPU

Intel Core i5

### GPU

RX 9060 XT

### ROCm Version

rocm-sdk-core/rocm-sdk-devel 10.1.0a20260806 (also affects 7.15.0a20260728, the version pip incorrectly prefers)

### Steps to Reproduce

Environment with torch==2.12.0+rocm10.1.0a20260806 (gfx1200, Windows, cp312) already installed.

1. Run:
```
   pip install torch[device-gfx1200] torchvision[device-gfx1200] torchaudio rocm-sdk-devel --index-url https://rocm.nightlies.amd.com/whl-multi-arch/ --upgrade
```

2. Observe pip output:
```
   Requirement already satisfied: torchaudio in ... (2.11.0+rocm10.1.0a20260806)
   Collecting torchaudio
     Downloading torchaudio-2.11.0%2Brocm7.15.0a20260728-cp312-cp312-win_amd64.whl (337 kB)
   Requirement already satisfied: torch[device-gfx1200] in ... (2.12.0+rocm10.1.0a20260806)
   Collecting torch[device-gfx1200]
     Downloading torch-2.12.0%2Brocm7.15.0a20260728-cp312-cp312-win_amd64.whl (114.3 MB)
```

[...]

PR #6618, opened 2026-07-16
fix(build_tools): forward override_git_sha in compute_rocm_package_version

tony-davis:
## Summary

`compute_rocm_package_version.py` `main()` never forwarded `args.override_git_sha` to `compute_version()`, so `--override-git-sha` was silently ignored and package versions embedded `GITHUB_SHA`/`HEAD` instead. This forwards the override and adds a `main()`-level regression test. Needed so the rocm-libraries nightlies (ROCm/rocm-libraries#9507) can stamp the resolved TheRock commit into nightly package metadata.

## Risk Assessment

Risk 2. One-line behavioral fix on the version-string path plus a test, well covered. It changes the embedded SHA only when a caller explicitly passes `--override-git-sha` (today: `setup_multi_arch.yml`), moving it from the wrong value to the intended one. No API/schema change.

## Related

- Fixes #6617
- Related PRs: ROCm/rocm-libraries#9517 (consumes this for nightly package-metadata traceability)
- Work tracking: ROCM-28114

## Device / Architecture Coverage

None. Host-only Python packaging helper; architecture-independent. PR CI is sufficient.

## Testing Summary

- Unit tests for the packaging helper, including a new `main()`-level override test.

## Testing Checklist

- [x] compute_rocm_package_version tests - `python -m pytest build_tools/tests/compute_rocm_package_version_test.py` - Status: Passed (24)
- [x] Reproduction: new test fails before the fix (embeds HEAD `710812f4...`, not the override), passes after - Status: Verified
- [ ] PR CI - GitHub PR checks - Status: Pending

## Flags / Guardrails

None.

## Technical Changes

- `build_tools/compute_rocm_package_version.py`: forward `args.override_git_sha` into `compute_version()` in `main()`.
- `build_tools/tests/compute_rocm_package_version_test.py`: add `test_main_forwards_override_git_sha_to_all_package_types` exercising the `main()` path.

PR #4929, opened 2026-04-29
Fix release dev version commit ref in cross-repo workflow_call

ScottTodd:
## Motivation

A dev release run in rockrel https://github.com/ROCm/rockrel/actions/runs/25100364100/job/73547700450 checked out commit `42e0a8bd7ddd664c7e600c854b4f6e8cd22ab220` from TheRock but computed version `7.13.0.dev0+b7df0b9cb7302534cb45e9614ba29962d4ec6c2c` using the commit ref from rockrel. The version should be the commit from TheRock, not rockrel.

## Test Plan

1. Attempted to test at https://github.com/ROCm/rockrel/actions/runs/25118876533/job/73613659899 but the workflow has `Uses: ROCm/TheRock/.github/workflows/setup_multi_arch.yml@refs/heads/main`
2. Tested from a branch with a different pin: https://github.com/ROCm/rockrel/actions/runs/25119107861/job/73614526170
    
    ```diff
    Run actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
    Syncing repository: ROCm/TheRock
    Checking out the ref
    /usr/bin/git log -1 --format=%H
    +bce56b2318b2cb6ca35826faee750814d190702b
    ```
    
    ```diff
    Run python ./build_tools/compute_rocm_package_version.py \
      python ./build_tools/compute_rocm_package_version.py \
        --release-type=dev \
        --prerelease-version= \
    +   --git-sha=bce56b2318b2cb6ca35826faee750814d190702b
      shell: /usr/bin/bash -e {0}
    Loading version from file '/home/runner/work/rockrel/rockrel/version.json'
    Base version  : '7.13.0'
    Package type  : 'wheel'
    +Version suffix: '.dev0+bce56b2318b2cb6ca35826faee750814d190702b'
    Full version  : '7.13.0.dev0+bce56b2318b2cb6ca35826faee750814d190702b'
    ```

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #7389, opened 2026-08-14
[packaging] Refine package version tests and documentation

ScottTodd:
## Motivation

* Cleanup tests and docs prior to changes planned as part of https://github.com/ROCm/rockrel/issues/88
* Document https://github.com/ROCm/TheRock/issues/7183
* Progress on https://github.com/ROCm/TheRock/issues/3340

## Technical Details

* Added tests for Python package version schema compliance and sorting. We can add similar tests for deb/rpm packages using Linux tooling if installed/available.
* Merged `GitShaOverrideTest` tests into other test classes (seeing lots of contributions that only append tests without considering what is already tested)
* Reworked `MainFunctionMultiplePackageTypesTest` to follow repository practices for mocking and removed confusing/useless `test_existing_workflows_still_work` test

## Test Plan

* Unit tests + ran a few commands like `python build_tools/compute_rocm_package_version.py --release-type=nightly --override-base-version=7.99.0` locally.
* CI and releases should continue using their existing versions.

## Submission Checklist

### DIFF
--- a/build_tools/tests/compute_rocm_package_version_test.py
+++ b/build_tools/tests/compute_rocm_package_version_test.py
@@ -122,7 +122,6 @@ class PythonPackageVersionTest(unittest.TestCase):
         # pip install --upgrade selects the greatest available version, so enforce:
         # release > prerelease > nightly > dev.
         versions = self._compute_versions_by_release_type()
-        print(f"versions: {versions}")
 
         self.assertGreater(
             Version(versions["nightly"]),
@@ -336,15 +335,56 @@ class RpmPackageVersionTest(unittest.TestCase):
         self.assertEqual(version, "8.0.0~custom1")
 
 
+# Test meaningful combinations of argparse options through the real computation path,
+# including main()'s side effect of writing versions to GitHub Actions outputs.
 class MainFunctionTest(unittest.TestCase):
-    def test_sets_package_version_outputs(self):
+    def test_sets_dev_outputs_with_version_overrides(self):
+        override_git_sha = "abcdef1234567890abcdef1234567890abcdef12"
+        with (
+            mock.patch.dict(os.environ, {"GITHUB_SHA": "f" * 40}),
+            mock.patch.object(
+                compute_rocm_package_version, "gha_set_output"
+            ) as gha_set_output,
+        ):
+            compute_rocm_package_version.main(
+                [
+                    "--release-type",
+                    "dev",
+                    "--override-base-version",
+                    "7.99.0",
+                    "--override-git-sha",
+                    override_git_sha,
+                ]
+            )
+
+        gha_set_output.assert_called_once()
+        outputs = gha_set_output.call_args.args[0]
+        self.assertEqual(
+            set(outputs),
+            {
+                "rocm_package_version",
+                "rocm_deb_package_version",
+                "rocm_rpm_package_version",
+            },
+        )
+        self.assertEqual(
+            outputs["rocm_package_version"],
+            f"7.99.0.dev0+{override_git_sha}",
+        )
+        self.assertTrue(outputs["rocm_deb_package_version"].startswith("7.99.0~dev"))
+        self.assertTrue(outputs["rocm_rpm_package_version"].startswith("7.99.0~"))
+        self.assertTrue(outputs["rocm_rpm_package_version"].endswith("gabcdef12"))
+
+    def test_sets_prerelease_outputs(self):
         with mock.patch.object(
             compute_rocm_package_version, "gha_set_output"
         ) as gha_set_output:
             compute_rocm_package_version.main(
                 [
                     "--release-type",
-                    "release",
+                    "prerelease",
+                    "--prerelease-version",
+                    "2",
                     "--override-base-version",
                     "7.99.0",
                 ]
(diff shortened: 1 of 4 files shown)


build_tools/tests/compute_rocm_package_version_test.py
```diff
@@ -99,6 +109,61 @@ def test_nightly_with_override_base_version(self):
...
+    def test_versions_sort_by_release_type(self):
+        # pip install --upgrade selects the greatest available version, so enforce:
+        # release > prerelease > nightly > dev.
+        versions = self._compute_versions_by_release_type()
+        print(f"versions: {versions}")
+
+        self.assertGreater(
+            Version(versions["nightly"]),
+            Version(versions["dev"]),
+        )
```
ScottTodd, line 129: > * Added tests for Python package version schema compliance and sorting. *We can add similar tests for deb/rpm packages using Linux tooling if installed/available.*

@nunnikri @raramakr fyi ^ I considered writing some tests that use e.g. `dpkg --validate-version` and `dpkg --compare-versions` but I can't easily run those tests on my Windows dev machine.

nunnikri: There are some python only modules available like python-debian, will try to explore those more and add the test.

build_tools/tests/compute_rocm_package_version_test.py
```diff
@@ -99,6 +109,61 @@ def test_nightly_with_override_base_version(self):
...
+
+        for release_type, version in versions.items():
+            with self.subTest(release_type=release_type):
+                self.assertEqual(str(Version(version)), version)
+
+    def test_versions_sort_by_release_type(self):
+        # pip install --upgrade selects the greatest available version, so enforce:
+        # release > prerelease > nightly > dev.
+        versions = self._compute_versions_by_release_type()
+        print(f"versions: {versions}")
```
nunnikri: This print might be enough in case of failures

ScottTodd: Whoops, that was for local debugging and pytest didn't even show it :P I can remove it.

build_tools/tests/compute_rocm_package_version_test.py
```diff
@@ -262,132 +336,27 @@ def test_custom_version_suffix(self):
...
-            package_type="rpm",
-            release_type="dev",
-            override_base_version="8.1.0",
-            override_git_sha="abcdef1234567890",
-        )
-        # Should truncate to 8 chars
-        self.assertRegex(version, r"^8\.1\.0~[0-9]{8}gabcdef12$")
-
-    def test_main_forwards_override_git_sha_to_all_package_types(self):
-        """main() must forward --override-git-sha into the computed versions.
```
nunnikri, line 287: This test is getting removed. are we planning to remove this parameter?

build_tools/tests/compute_rocm_package_version_test.py
ScottTodd: I was considering making the tests log all the different versions and "checking that x > y" for these tests, together with https://docs.pytest.org/en/stable/how-to/logging.html#live-logs.

nunnikri (approved):
Overall looks fine. Minor comments given, better to take care.

build_tools/tests/compute_rocm_package_version_test.py
ScottTodd: `--override-git-sha` is still supported, I would just rather test the actual behavioral differences of `main()` compared to `compute_rocm_package_version()`, i.e. the call to `gha_set_output()`. I don't think it's very useful to have tests checking that argparse forwards inputs to a function.

ScottTodd: Actually I think I'll add back test coverage for this and any other argparse options, given that argument forwarding was missing at one point despite other tests passing:
* Added in https://github.com/ROCm/TheRock/pull/4929
* Dropped in https://github.com/ROCm/TheRock/pull/4359
* Added back in https://github.com/ROCm/TheRock/pull/6618

docs/packaging/versioning.md
```diff
@@ -134,6 +128,21 @@ True
...
+> versions using this schema:
+>
+> ```python
+> >>> from packaging.version import Version
+> >>> Version("2.12.0+rocm10.0") > Version("2.12.0+rocm7.0")
+> False
+> ```
+>
+> We plan on publishing 10.0.0+ packages to a new index to avoid this issue for
+> future installs.
```
marbre, line 144: To further harden this, we could also alter the local version identifier. Let's discuss this offline.

build_tools/tests/compute_rocm_package_version_test.py
ScottTodd: I explored a few ways to test argparse -> call function -> set outputs, settled on this setup:
```python
# Test meaningful combinations of argparse options through the real computation path,
# including main()'s side effect of writing versions to GitHub Actions outputs.
class MainFunctionTest(unittest.TestCase):
    def test_sets_dev_outputs_with_version_overrides(self):
    def test_sets_prerelease_outputs(self):
    def test_sets_custom_suffix_outputs(self):
```

We could also add more generic tests (or static analysis) that check that all argparse options are actually used, across all of our scripts. Maybe there's a pylint or similar check for that.

build_tools/tests/compute_rocm_package_version_test.py
ScottTodd: Nice, `python-debian` works on Windows too:
```bash
(3.12.venv) λ pip install python-debian
Collecting python-debian
  Downloading python_debian-1.1.1-py3-none-any.whl.metadata (2.6 kB)
Downloading python_debian-1.1.1-py3-none-any.whl (138 kB)
Installing collected packages: python-debian
Successfully installed python-debian-1.1.1
```

```python
(3.12.venv) λ python
Python 3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from debian.debian_support import NativeVersion
>>> NativeVersion("10.1.0~20260811+bkc.20260814") < NativeVersion("10.1.0~20260811")
False
>>> NativeVersion("abcdef+%%%random_characters")
  File "<stdin>", line 1, in <module>
  File "D:\scratch\therock\3.12.venv\Lib\site-packages\debian\debian_support.py", line 149, in __init__
    self.full_version = version
    ^^^^^^^^^^^^^^^^^
  File "D:\scratch\therock\3.12.venv\Lib\site-packages\debian\debian_support.py", line 176, in __setattr__
    self._set_full_version(str(value))
  File "D:\scratch\therock\3.12.venv\Lib\site-packages\debian\debian_support.py", line 154, in _set_full_version
    raise ValueError("Invalid version string %r" % version)
ValueError: Invalid version string 'abcdef+%%%random_characters'
>>>
```

This PR stack is already rather complex but I could work that into the branches...

ScottTodd: I sent a follow-up PR adding tests (which actually revealed that currently sorting is `dev > nightly`): https://github.com/ROCm/TheRock/pull/7433

PR #7390, opened 2026-08-14
[packaging] Add versioning for BKC release type

ScottTodd:
## Motivation

* Progress on https://github.com/ROCm/rockrel/issues/88

We're creating a new release stream that will branch off from nightly releases while carrying cherry-picks. This PR implements versioning for these releases with a local version segment (https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-segments) that extends the base nightly version:

Version type | Example
-- | --
nightly | `10.1.0a20260811`
nightly-bkc| `10.1.0a20260811+bkc.20260813`

To support testing this also adds a `dev-bkc` release type that may redirect the output buckets while keeping the same dev version scheme used for regular dev releases.

## Technical Details

To produce versions like:
```
10.1.0a20260811+bkc.20260813
10.1.0a20260811+bkc.20260814
10.1.0a20260811+bkc.20260815
```

we'll store `base-date` (e.g.  `20260811`) as metadata in version.json on the release branch and then have the script that computes the version load the current data into the local version segment.

## Test Plan

* Unit tests and local commands like `python build_tools/compute_rocm_package_version.py --release-type=nightly-bkc --override-base-version=7.99.0`
* Once the full stack of changes lands we'll be ready to trigger dev and nightly bkc releases in https://github.com/ROCm/rockrel, where this will be tested for real

## Submission Checklist

### DIFF
--- a/build_tools/compute_rocm_package_version.py
+++ b/build_tools/compute_rocm_package_version.py
@@ -31,7 +31,7 @@ Sample usage:
 
   python compute_rocm_package_version.py --release-type=nightly-bkc
   # rocm_package_version=10.1.0a20260811+bkc.20260814
-  # rocm_deb_package_version=10.1.0~20260811+bkc.20260814
+  # rocm_deb_package_version=10.1.0~20260811.bkc.20260814
   # rocm_rpm_package_version=10.1.0~20260811.bkc.20260814
 
   python compute_rocm_package_version.py --custom-version-suffix=.dev0
@@ -240,10 +240,7 @@ def compute_version(
             # Format: <rocm-version>~<YYYYMMDD>
             version_suffix_str = f"~{current_date}"
         elif release_type == "nightly-bkc":
-            bkc_separator = "+" if package_type == "deb" else "."
-            version_suffix_str = (
-                f"~{release_metadata['base-date']}{bkc_separator}bkc.{current_date}"
-            )
+            version_suffix_str = f"~{release_metadata['base-date']}.bkc.{current_date}"
         elif release_type == "prerelease":
             # Construct a prerelease version
             # deb format: <rocm-version>~pre<N>
--- a/build_tools/compute_rocm_package_version.py
+++ b/build_tools/compute_rocm_package_version.py
@@ -29,6 +29,11 @@ Sample usage:
   # rocm_deb_package_version=7.10.0~20251021
   # rocm_rpm_package_version=7.10.0~20251021
 
+  python compute_rocm_package_version.py --release-type=nightly-bkc
+  # rocm_package_version=10.1.0a20260811+bkc.20260814
+  # rocm_deb_package_version=10.1.0~20260811+bkc.20260814
+  # rocm_rpm_package_version=10.1.0~20260811+bkc.20260814
+
   python compute_rocm_package_version.py --custom-version-suffix=.dev0
   # 7.10.0.dev0
 
@@ -37,6 +42,7 @@ Sample usage:
 """
 
 import argparse
+from collections.abc import Mapping
 from datetime import datetime
 from pathlib import Path
 import json
@@ -62,6 +68,18 @@ def load_rocm_version() -> str:
     with open(version_file, "rt") as f:
         loaded_file = json.load(f)
         return loaded_file["rocm-version"]
+
+
+def load_release_metadata() -> dict[str, object]:
+    """Loads generic release metadata from the repository's version.json file."""
+    version_file = THEROCK_DIR / "version.json"
+    _log(f"Loading release metadata from file '{version_file.resolve()}'")
+    with open(version_file, "rt") as f:
+        loaded_file = json.load(f)
+        release_metadata = loaded_file.get("release-metadata", {})
+        if not isinstance(release_metadata, dict):
+            raise ValueError("release-metadata in version.json must be an object")
+        return release_metadata
 
 
 def get_git_sha(short: bool = False, override_git_sha: str | None = None):
@@ -112,17 +130,21 @@ def compute_version(
     prerelease_version: str | None = None,
     override_base_version: str | None = None,
     override_git_sha: str | None = None,
+    release_metadata: Mapping[str, object] | None = None,
 ) -> str:
     """Compute package version based on package type and release type.
 
     Args:
         package_type: Type of package ("wheel", "deb", or "rpm")
-        release_type: Release type ("ci", "dev", "nightly", "prerelease", or "release")
+        release_type: Release type ("ci", "dev", "dev-bkc", "nightly",
+            "nightly-bkc", "prerelease", or "release")
         custom_version_suffix: Custom suffix to override automatic suffix
         prerelease_version: Prerelease version number
         override_base_version: Override the base version from version.json
         override_git_sha: Explicit git SHA override, forwarded to get_git_sha().
             See get_git_sha() for details on when this is needed.
+        release_metadata: Generic metadata from version.json. Loaded automatically
+            for BKC release types when not provided.
 
     Returns:
         Computed version string appropriate for the package type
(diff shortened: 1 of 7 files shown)


docs/packaging/versioning.md
```diff
@@ -272,21 +322,25 @@ Native package versions are handled by scripts:
...
+| Release type | Version format                               | Version example                                                        |
+| ------------ | -------------------------------------------- | ---------------------------------------------------------------------- |
+| stable       | `X.Y.Z`                                      | `7.10.0`                                                               |
+| prerelease   | `X.Y.Z~preN`                                 | `7.10.0~pre0`<br>(The first release candidate for that stable release) |
+| nightly      | `X.Y.Z~YYYYMMDD`                             | `7.10.0~20251124`<br>(The nightly release on 2025-11-24)               |
+| nightly-bkc  | `X.Y.Z~BASEDATE+bkc.YYYYMMDD`                | `10.1.0~20260811+bkc.20260814`                                         |
+| dev          | `X.Y.Z~devYYYYMMDD`                          | `7.10.0~dev20251124`<br>(For dev build on 2025-11-24)                  |
+| dev-bkc      | `X.Y.Z~devYYYYMMDD`<br>(same as regular dev) | `10.1.0~dev20260814`                                                   |
```
ScottTodd, line 343: @nunnikri can you review the deb/rpm version changes for nightly-bkc here? I'm mostly trusting my agent on them, as I'm more familiar with python package versioning and typical conventions.

build_tools/compute_rocm_package_version.py
```diff
@@ -29,6 +29,11 @@
   # rocm_deb_package_version=7.10.0~20251021
   # rocm_rpm_package_version=7.10.0~20251021
 
+  python compute_rocm_package_version.py --release-type=nightly-bkc
+  # rocm_package_version=10.1.0a20260811+bkc.20260814
+  # rocm_deb_package_version=10.1.0~20260811+bkc.20260814
```
nunnikri: having + in a rpm version can cause issues. 
I would suggest to use 
`10.1.0~20260811bkc20260814 OR 10.1.0~20260811.bkc.20260814 `

ScottTodd: Do you want the same change for deb, or just rpm? I would prefer `10.1.0~20260811.bkc.20260814` (with `.` separators) over `10.1.0~20260811bkc20260814`

ScottTodd: I changed for just rpm (see the latest commit, the rest got rebased as part of updating and merging the base branch)

nunnikri: Yes this looks more readable. `10.1.0~20260811.bkc.20260814`
rpm package version doesnt allow `+`. The same issue is not there in debian package it allows `+` but its not a usual conventions. So if we are choosing `10.1.0~20260811.bkc.20260814` as the versioning scheme for rpm, i would prefer to keep same for debian as well.

ScottTodd: Ok, updated deb also. I couldn't find a source for "rpm package version doesn't allow +", but I might have just missed something, and the docs I found were harder to parse than those for python package versions.

Here's what my agent found, fyi:

<details><summary>Expand for agent findings</summary>
<p>


```
• The feedback is directionally reasonable, but RPM’s specification does not say + is invalid or unsafe.

  RPM explicitly allows ., _, and + as separators. Those separators are ignored during comparison, so these three
  versions are equivalent to RPM:

  10.1.0~20260811+bkc.20260814
  10.1.0~20260811.bkc.20260814
  10.1.0~20260811bkc20260814

  Numeric-to-alphabetic transitions create implicit segments, which is why the version without separators also compares
  identically. RPM version specification (https://rpm.org/docs/6.0.x/man/rpm-version.7), RPM spec-file documentation
  (https://rpm.org/docs/4.20.x/manual/spec.html)

  10.1.0~20260811
    < 10.1.0~20260811.bkc.20260814
    < 10.1.0~20260811.bkc.20260815
    < 10.1.0~20260812
    < 10.1.0

  The initial ~ makes all these prerelease versions older than 10.1.0; the base date positions BKC builds between regular
  nightlies; the final date orders BKC builds.

  My recommendation is:

  - Python: retain 10.1.0a20260811+bkc.20260814. Here + has defined PEP 440 local-version meaning.
  - DEB: retain 10.1.0~20260811+bkc.20260814. Debian explicitly allows +, and it is conventionally used for downstream
    extensions. Debian Policy (https://www.debian.org/doc/debian-policy/ch-controlfields.html)

  - RPM: use 10.1.0~20260811.bkc.20260814.

  The RPM change has essentially no downside:

  - It follows the reviewer’s practical guidance.
  - Dots are visually clearer than 10.1.0~20260811bkc20260814.
  - It avoids any surrounding tools that might give + non-RPM meaning.
  - RPM ordering is unchanged.

  I would not change both native formats merely to keep them identical. We already generate package-type-specific
  versions, and + is more idiomatic for Debian than it is useful for RPM.

  Other options:

  - Keep +: technically defensible, but it buys us nothing in RPM and leaves the review concern unresolved.
  - Remove the separator: valid and equivalent, but less readable.
  - Put BKC information in RPM’s Release: field: semantically plausible because RPM defines that field for packaging
    revisions. However, TheRock already uses config.version_suffix as Release:, so this would expand into packaging
    interfaces, dependency EVRs, and workflow changes. It is disproportionate for this PR. CPack RPM documentation
    (https://cmake.org/cmake/help/latest/cpack_gen/rpm.html)

  One edge case: if any +bkc RPMs have already been published, RPM considers a later dotted spelling with the same dates
  equal—not newer. We should change the format before publishing rather than rely on it as an upgrade.
```

</p>
</details>

nunnikri: What agent mentioned is correct `+` is not technically restricted in rpm spec. But some of the older OS doesnt use it. https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/. So its safer not to use `+` in rpm. 

For debian, the `+` is having a slightly different meaning. `~` means its pre release and `+` means its post release. https://www.debian.org/doc/manuals/debmake-doc/ch06.en.html#name-version
So 10.1.0~20260811+bkc.20260814 means its a 10.1 pre release and a 2026811 post release build. Which might be correct. 
I dont have a strong opinion of deb version here. Though my preference was to have same version.  But it looks like using `+` is more correct as per the spec.

nunnikri (approved):
Overall looks good to me. Deb version can update if needed.

ScottTodd:
Ok let's start with deb and rpm matching and we can refine later as needed. Given that...
1. this is a new release stream that is intended for system use and not too much direct consumer installation
2. we're planning on shifting from the current dev/nightly indexes to a new index with repo.amd.com

we have some leeway where we could delete/yank/repackage releases as needed after changing the version scheme.

build_tools/github_actions/tests/determine_version_test.py
```diff
@@ -27,6 +27,11 @@ def test_nightly_version_suffix(self):
         suffix = determine_version.derive_version_suffix(rocm_version)
         self.assertEqual(suffix, "+rocm7.0.0rc20250707")
 
+    def test_bkc_version_suffix(self):
+        rocm_version = "10.1.0a20260811+bkc.20260813"
+        suffix = determine_version.derive_version_suffix(rocm_version)
+        self.assertEqual(suffix, "+rocm10.1.0a20260811-bkc.20260813")
```
ScottTodd, line 33: note: there's a slight difference here between https://github.com/ROCm/rockrel/issues/88 and the code: this has `-bkc` instead of `.bkc`, but the package should actually be normalized from `-` to `.` by python packaging for us (see how https://rocm.devreleases.amd.com/whl-multi-arch/torch/ has e.g. `torch-2.10.0+devrocm7.14.0.dev0.271529c2b1f241d8b3c849f3e1d1ba879a0f23a8-cp310-cp310-linux_x86_64.whl` with `.` instead of `-`)

`test_dev_version_suffix` above has the same issue

I can send a follow-up PR updating docs and https://github.com/ROCm/TheRock/blob/f4d4233e73e1e8d73f8e1eab03bde0f958cdf43e/build_tools/github_actions/determine_version.py#L39-L62

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode
> Escaping and Unicode
>
> As the components of the filename are separated by a dash (-, HYPHEN-MINUS), this character cannot appear within any component.
>
> Version numbers should be normalised according to the [Version specifier specification](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers). Normalised version numbers cannot contain -.

PR #7395, opened 2026-08-15
fix(packaging): Suppress vendored TBB provides on amdrocm-profiler RPM

arvindcheru:
## Motivation

This pull request updates the RPM packaging logic to properly handle vendored TBB libraries in the `amdrocm-profiler` package. The main goal is to prevent the package manager from treating these bundled libraries as system-wide dependencies, which could otherwise interfere with package removal or introduce unwanted dependencies. The changes also add tests to ensure this behavior is correct and isolated to the profiler package.

## Technical Details

**Packaging logic updates:**

* Added logic in `rpm_package.py` to set an `exclude_vendored_tbb_metadata` flag for the `amdrocm-profiler` package, which is then passed into the RPM spec file template. This ensures that only the profiler package receives the special handling for vendored TBB libraries. [[1]](diffhunk://#diff-54b422a21ef4c708683938f44a09261371a8a12a9de894db3cef44aff000fd40R112-R116) [[2]](diffhunk://#diff-54b422a21ef4c708683938f44a09261371a8a12a9de894db3cef44aff000fd40R192)

**Spec file template changes:**

* Updated the `rpm_specfile.j2` template to conditionally add `%global __requires_exclude` and `%global __provides_exclude` directives for TBB libraries when the new flag is set, suppressing both auto-generated "provides" and "requires" metadata for TBB.

JIRA ID : ROCM-28385

## Test Plan

[] python3.12 -m unittest build_tools/packaging/linux/tests/build_package_test.py
[] Reproduce the issue with ROCM-28385 with the fix (OL 8.10 repro: install kernel-uek-devel + amdrocm-developer-tools, remove metapackage and zero amdrocm RPMs remain)
 rpm -q --provides amdrocm-profiler shows no libtbb* sonames
[] Check the amdrocm-profiler provides

## Test Result
Fix Verified with ROCM-28385
Verified: CI Build - https://therock-ci-artifacts.s3.amazonaws.com/34865153570-linux/packages/rpm/x86_64/
[root@CTR-SVDT-L004 /]# rpm -q --provides amdrocm-profiler10.2 | grep tbb

## Submission Checklist


arvindcheru:
Attaching Verification Logs with Fix:
[Verifying_28385_withFIX.txt](https://github.com/user-attachments/files/32271507/Verifying_28385_withFIX.txt)

nunnikri (approved):
LGTM. May be as a future change we need to clear the difference in macro definition. Some are using %define and some are using %global

PR #7425, opened 2026-08-17
fix(sysdeps): make rocm_sysdeps zlib.pc reliably relocatable

### LINKED ISSUE #7424: rocm_sysdeps zlib.pc is non-deterministically non-relocatable
## Description

`lib/rocm_sysdeps/lib/pkgconfig/zlib.pc` in the shipped ROCm SDK sometimes contains
an absolute build-machine prefix instead of a relocatable one. Which version ships
varies between nightlies with no source change in between.

Broken (Aug 13 and Aug 17 nightlies):

```
prefix=/__w/rockrel/rockrel/build/third-party/sysdeps/linux/zlib/build/stage/lib/rocm_sysdeps
exec_prefix=/__w/rockrel/rockrel/build/third-party/sysdeps/linux/zlib/build/stage/lib/rocm_sysdeps
libdir=${exec_prefix}/lib
sharedlibdir=${exec_prefix}/lib
includedir=${exec_prefix}/include

Name: zlib
Description: zlib compression library
Version: 1.3.2
License: Zlib

Requires:
Libs: -L${libdir} -L${sharedlibdir} -lz
Cflags: -I${includedir}
```

Correct (Aug 14 nightly), which is what `third-party/sysdeps/common/zlib/zlib.pc.in`
produces:

```
prefix=${pcfiledir}/../..
libdir=${prefix}/lib
includedir=${prefix}/include

Name: zlib
Description: zlib compression library
Version: 1.3.2

Requires:
Libs: -L${libdir} -lz
Cflags: -I${includedir}
```

The broken variant is upstream zlib's own generated file, distinguishable by the
extra `exec_prefix`, `sharedlibdir`, and `License:` fields.

## Root cause

`third-party/sysdeps/common/zlib/CMakeLists.txt` passes `-DSKIP_INSTALL_FILES=ON`
to the zlib sub-build with the comment "Disable writing pc files: we will do so
ourselves."

**zlib 1.3.2 removed that option.** In the mirrored tarball referenced in that same
file (SHA256 `bb329a0a2cd0274d05519d61c667c062e06990d72e125ee2dfa8de64f0119d16`), the
string `SKIP_INSTALL` does not appear anywhere. The only knob is `ZLIB_INSTALL`
(line 25), and `zlib.pc` is installed unconditionally under it (lines 114-115 and
297-299).

So the flag is a silent no-op and two different `zlib.pc` files are written to the
same destination on every build:

- upstream's, during `cmake --install b` inside the `build ALL` target
- TheRock's relocatable one, via `install(FILES ...)` at the bottom of the same file

Whichever lands last wins. The `build ALL` target also opens with
`rm -rf ${CMAKE_INSTALL_PREFIX}`, so a rebuild not followed by the super-project
install leaves only upstream's file. That is the likely source of the variation
between nightlies, though I have not confirmed which ordering each specific build hit.

Every other sysdep relativizes explicitly in its `patch_install.sh` (elfutils,
numactl, libcap, zstd). zlib is the only one relying on a build option that no
longer exists.

## Impact

Consumers that resolve zlib through pkg-config against
`lib/rocm_sysdeps/lib/pkgconfig` get a `-I` pointing at a non-existent directory.
CMake refuses to generate:

```
CMake Error in .../CMakeLists.txt:
  Imported target "PkgConfig::zlib" includes non-existent path
    "/__w/rockrel/rockrel/build/third-party/sysdeps/linux/zlib/build/stage/lib/rocm_sysdeps/include"
```

This broke rocprofiler-compute's native tool build on the Aug 17 nightly. The CMake
package config at `lib/rocm_sysdeps/lib/cmake/ZLIB/zlib-config.cmake` is relocatable
and correct in all three nightlies, so consumers using `find_package(ZLIB CONFIG)`
are unaffected, which is why this went unnoticed.

## Reproduction

```bash
head -1 $ROCM_PATH/lib/rocm_sysdeps/lib/pkgconfig/zlib.pc
# broken:  prefix=/__w/rockrel/...
# correct: prefix=${pcfiledir}/../..

PKG_CONFIG_PATH=$ROCM_PATH/lib/rocm_sysdeps/lib/pkgconfig \
  pkg-config --variable=includedir zlib
# broken if the printed path does not exist
```

Checked against `rocm-sdk-devel` 10.1.0a20260813, 10.1.0a20260814, and
10.1.0a20260817 from https://rocm.nightlies.amd.com/whl-multi-arch/.

vedithal-amd:
## Motivation

`lib/rocm_sysdeps/lib/pkgconfig/zlib.pc` non-deterministically ships upstream
zlib's non-relocatable file instead of TheRock's, breaking any consumer that
resolves zlib through pkg-config.

Fixes https://github.com/ROCm/TheRock/issues/7424

ISSUE ID: 7424
JIRA ID: AIPROFCOMP-816

## Technical Details

- `-DSKIP_INSTALL_FILES=ON` is a no-op: zlib 1.3.2 removed that option, so
  upstream's `zlib.pc` is always installed and races with the relocatable one
  installed from `zlib.pc.in`. Replace the flag with a comment saying so.
- Delete upstream's `zlib.pc` in `patch_install.py`, mirroring the existing
  removal of upstream's cmake config directory, so only TheRock's file can survive.

## Test Plan

- Confirm `SKIP_INSTALL_FILES` is absent from the mirrored zlib 1.3.2 sources
- Build the zlib sysdep and check `lib/rocm_sysdeps/lib/pkgconfig/zlib.pc` starts
  with `prefix=${pcfiledir}/../..`
- Confirm `pkg-config --variable=includedir zlib` against the staged pkgconfig dir
  resolves to a path that exists

## Test Result

- Verified the option is absent from the mirrored tarball
  (`bb329a0a2cd0274d05519d61c667c062e06990d72e125ee2dfa8de64f0119d16`), and confirmed
  the broken and good files alternate across the Aug 13, 14, and 17 nightly wheels
- Not built locally: a full TheRock superbuild was out of reach in my environment,
  so the staged-artifact checks above need to come from CI

## Submission Checklist

### DIFF
--- a/third-party/sysdeps/common/zlib/CMakeLists.txt
+++ b/third-party/sysdeps/common/zlib/CMakeLists.txt
@@ -98,8 +98,8 @@ add_custom_target(
       # On Windows we only use the static library; skip the shared build
       # to avoid having to track the DLL alongside the static artifact.
       "$<$<BOOL:${WIN32}>:-DZLIB_BUILD_SHARED=OFF>"
-      # Disable writing pc files: we will do so ourselves.
-      -DSKIP_INSTALL_FILES=ON
+      # zlib 1.3.2 dropped SKIP_INSTALL_FILES, so it always installs its own
+      # zlib.pc. patch_install.py removes it; we install our own below.
 
   COMMAND
     "${CMAKE_COMMAND}" --build "${CMAKE_CURRENT_BINARY_DIR}/b"
--- a/third-party/sysdeps/common/zlib/patch_install.py
+++ b/third-party/sysdeps/common/zlib/patch_install.py
@@ -26,3 +26,10 @@ if platform.system() == "Linux":
 cmake_dir = Path(PREFIX) / "lib" / "cmake" / "zlib"
 if cmake_dir.is_dir():
     shutil.rmtree(cmake_dir)
+
+# Remove zlib's auto-generated pkg-config file, which hardcodes the build-time
+# prefix. TheRock installs its own relocatable zlib.pc to the same path, so
+# leaving this one in place makes the result depend on install ordering.
+pc_file = Path(PREFIX) / "lib" / "pkgconfig" / "zlib.pc"
+if pc_file.exists():
+    pc_file.unlink()


jbonnell-amd (approved):
Looks good!

I assume this was a regression that was picked up when upgrading to the newer version that got missed. Should we add a test to ensure this doesn't occur again? Mostly just want to make sure we don't experience something like this again in the future.

vedithal-amd:
Multi arch linux tests are passing - merging the pr

PR #7431, opened 2026-08-17
feat(ci): populate baseline-run-id from merged PR's CI run

geomin12:
When bumping rocm-libraries therock-ref, also update the baseline-run-id in ci-env/action.yml with the completed Multi-Arch CI run ID from the merged PR. This allows rocm-libraries multi-arch CI to reuse prebuilt artifacts from TheRock.

Paired with https://github.com/ROCm/rocm-libraries/pull/10894

Changes:
- Add get_merged_pr_for_commit() to find the PR that produced a merge commit
- Add get_workflow_run_id_for_pr() to get successful CI run ID from PR head
- Add get_baseline_run_id_from_merged_pr() to orchestrate the above
- Modify update_ci_env_file() to also update baseline-run-id
- Modify handle_push() to fetch and pass baseline_run_id for ci-env updates

### DIFF
--- a/build_tools/github_actions/bump_automation.py
+++ b/build_tools/github_actions/bump_automation.py
@@ -107,6 +107,40 @@ def gh_api(
     return response.json()
 
 
+def get_baseline_run_id_from_merged_pr(
+    repo: str, token: str, merge_commit_sha: str, workflow_name: str = "Multi-Arch CI"
+) -> str | None:
+    """Get the baseline run ID from the PR that was just merged."""
+    # Find the PR that produced this merge commit
+    prs = gh_api(token, f"repos/{repo}/commits/{merge_commit_sha}/pulls")
+    pr = next(
+        (
+            p
+            for p in prs
+            if p.get("merged_at") and p.get("merge_commit_sha") == merge_commit_sha
+        ),
+        None,
+    )
+    if not pr:
+        print(f"[WARN] No merged PR found for commit {merge_commit_sha[:7]}")
+        return None
+
+    # Get the completed workflow run for this PR's head commit
+    pr_head_sha = pr["head"]["sha"]
+    runs = gh_api(
+        token, f"repos/{repo}/actions/runs?head_sha={pr_head_sha}&status=completed"
+    )
+    for run in runs.get("workflow_runs", []):
+        if run["name"] == workflow_name:
+            print(
+                f"[INFO] Found {workflow_name} run {run['id']} for PR #{pr['number']}"
+            )
+            return str(run["id"])
+
+    print(f"[WARN] No {workflow_name} run found for PR #{pr['number']}")
+    return None
+
+
 def latest_commit(repo: str, token: str, branch: str | None = None) -> str:
     """Return the SHA of the latest commit on the given branch, or the default branch."""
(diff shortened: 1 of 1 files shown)


build_tools/github_actions/bump_automation.py
```diff
@@ -107,6 +108,40 @@ def gh_api(
...
+    )
+    if not pr:
+        print(f"[WARN] No merged PR found for commit {merge_commit_sha[:7]}")
+        return None
+
+    # Get the completed workflow run for this PR's head commit
+    pr_head_sha = pr["head"]["sha"]
+    runs = gh_api(
+        token, f"repos/{repo}/actions/runs?head_sha={pr_head_sha}&status=completed"
+    )
```
erman-gurses, line 132: Should this require a successful workflow run rather than just completed?

build_tools/github_actions/bump_automation.py
```diff
@@ -397,6 +457,19 @@ def handle_push(before: str, after: str, tokens: dict[str, str]) -> None:
...
+    if config.get("updater") in ("ci-env", "ref-and-ci-env"):
+        baseline_run_id = get_baseline_run_id_from_merged_pr(
+            THEROCK_REPO, token, after, workflow_name="Multi-Arch CI"
+        )
+        if baseline_run_id:
+            print(f"[INFO] Using baseline run ID {baseline_run_id} from merged PR")
+        else:
+            print(
+                f"[WARN] Could not get baseline run ID from merged PR, proceeding without it"
+            )
```
erman-gurses: Some clarification would be good for proceeding without baseline. Please see agent's comment:

_Concerned about proceeding with the bump when we fail to resolve the baseline run ID._

_update_ci_env_file() will write baseline-run-id: "", but the paired rocm-libraries change still supplies a non-empty default prebuilt_stages. In TheRock, copy_prebuilt_stages only runs when both prebuilt_stages and baseline_run_id are non-empty. So this looks like it can skip the artifact copy while still telling the build that those stages are prebuilt._

_Can this fail the bump instead of proceeding? Alternatively, the downstream workflow would need to clear prebuilt_stages whenever there is no baseline ID so it falls back to building everything._

erman-gurses (commented):
Please see my comments:

build_tools/github_actions/bump_automation.py
endurthiabhilash: +1 on requiring a successful workflow. Also, another concern - is it possible that we don't get the multi-arch run returned within the 30 results we get from the API call? One way to avoid that would be to fetch multi arch workflow runs, other would be to increase the per_page number of results.

geomin12: due to the flakiness of CI, "successful" will never appear. Success = green builds and tests.

We really only care about passing builds, in which the PR checks are watched everyday by tiger team and are known green builds

build_tools/github_actions/bump_automation.py
geomin12: fixed in here: https://github.com/ROCm/rocm-libraries/pull/10894/changes

erman-gurses (approved):
LGTM

endurthiabhilash (approved):

PR #7434, opened 2026-08-17
Drop unused dependencies

### LINKED ISSUE #7432: `dbus_python` and `flatbuffers` are not index
Those dependencies are uploaded to our mirror but are not appearing in the index. They can be therefore dropped as this isn't causing a regression.

marbre:
## Motivation

`dbus_python` and `flatbuffers` are uploaded but have never been indexed, as the packages have not been whitelisted in `manage.py`. Hence, the dependencies have never been consumed from our mirror and can be dropped. This most likely works because `jaxlib` is consumed from PyPI.

Closes #7432.

## Submission Checklist

### DIFF
--- a/build_tools/third_party/s3_management/update_dependencies.py
+++ b/build_tools/third_party/s3_management/update_dependencies.py
@@ -66,8 +66,6 @@ _ALLOWED_CPYTHON_TAGS: frozenset[str] = frozenset(
 )
 
 PACKAGES_PER_PROJECT = {
-    "dbus_python": {"versions": ["latest"], "project": "jax"},
-    "flatbuffers": {"versions": ["latest"], "project": "jax"},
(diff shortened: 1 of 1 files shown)


ScottTodd (approved):
Ok if not currently used, but I would like for `jaxlib` to come from our own index at some point too, so users can install jax + rocm from a single pip command without mixing multiple indexes.

marbre:
> Ok if not currently used, but I would like for `jaxlib` to come from our own index at some point too, so users can install jax + rocm from a single pip command without mixing multiple indexes.

Agreed, but this required a few more changes due to the current repo.amd.com refactoring. As of now I'd like to avoid stale packages in our index.

erman-gurses:
From JAX point of view, I verified this with JAX team back in time so jax version > 0.8.2, jaxlib coming from PyPI

erman-gurses (approved):

PR #7421, opened 2026-08-17
fix(rocprofiler-compute): wire test-compression for TheRock install-mode CI

jamessiddeley-amd:
## Motivation

Added a `test-compression` gtest in [#9711 of rocprofiler-compute,](https://github.com/ROCm/rocm-systems/pull/9711), but TheRock CI can't run it in install mode yet:

- The post-hook sets install RPATH for the other gtests under `libexec/rocprofiler-compute/tests/` but not `test-compression`
- The subproject pulls bundled elfutils/libdrm/sqlite3 but not zlib, so `pkg_check_modules(zlib)` can link system `libz` instead of bundled sysdeps

The follow-up in rocm-systems repo is to add `test-compression` to `test_categories.yaml` after this merges.

## Technical Details

- Add `test-compression` to the RPATH post-hook foreach in `profiler/post_hook_rocprofiler-compute.cmake`.
- Add `${THEROCK_BUNDLED_ZLIB}` to `rocprofiler-compute` `RUNTIME_DEPS` in `profiler/CMakeLists.txt`.

## Test Plan

- Build with rocprofiler-compute + tests enabled.
- Confirm `test-compression` installs under `libexec/rocprofiler-compute/tests/` with correct RPATH.
- Run `ctest -R test-compression`.

## Test Result

- Pending CI.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

PR #7376, opened 2026-08-14
[profiler] Add profiler-hub as a TheRock subproject and artifact

avansick-amd:
Co-authored-by: Claude Opus 5 <noreply@anthropic.com>
Signed-off-by: Adam Van Sickle <adam.vansickle@amd.com>

## JIRA ID
ROCPDSNA-73

## Motivation

profiler-hub is a general-purpose, schema-versioned data storage library for profiling and tracing tools. It defines a common on-disk schema and a writer/reader API for storing trace, PMC and profiling data, so multiple independent tools can share a single storage format instead of each implementing their own.

This PR makes profiler-hub a first-class artifact end to end:

- built as its own subproject;
- published as its own artifact;
- shipped in the DEB/RPM packages and the rocm-profiler wheel;
- provided to rocprofiler-systems as a real CMake package.

The build, packaging and testing phases were originally split across three PRs. Build and packaging are combined here to avoid sequencing errors. #7379, the packaged-consumption tests, still sits above this PR.

## Technical Details

11 files, 294 added lines, 6 removed.

Build:

- `profiler/CMakeLists.txt` declares `profiler-hub` as its own subproject behind a new
  `THEROCK_ENABLE_PROFILER_HUB` flag. It is a pure host C++ library, so `DISABLE_AMDGPU_TARGETS`;
  it builds with in-tree clang via `COMPILER_TOOLCHAIN amd-llvm` so its tests are ASan-linked.
   Build deps are spdlog and fmt, plus googletest under `THEROCK_BUILD_TESTING`; the only
  runtime dep is the bundled sqlite3.
- The subproject registers `therock_cmake_subproject_build_test`, which runs profiler-hub's gtest
  binaries under ctest as part of the subproject's test step, and `therock_test_validate_shared_lib`
  to validate `libprofiler-hub.so`.
- `profiler/artifact-profiler-hub.toml` describes the components and `BUILD_TOPOLOGY.toml` registers
  the artifact (`artifact_group = profiler-core`, target-neutral, Linux only). The descriptor
  deliberately omits a bare `run` section: a bare component is a catch-all and would claim `dev`'s
  files.
- `profiler/profiler-hub-install-test/` is a small consumer project declared as its own subproject so it
  compiles during the build phase.
- `test_tools/therock_consumer_graph.json` is regenerated from a full `-DTHEROCK_ENABLE_ALL=ON`
  configure rather than hand-edited.
[...]

Issue #7298, opened 2026-08-12
Consumer graph drift check fails repo-wide: kfdtest edges missing from committed graph


BrianHarrisonAMD:
## Summary

`test_tools/therock_consumer_graph.json` is missing the kfdtest edges. The drift check compares the whole file, so this affects every PR that touches a graph-affecting path.

Possible fix: #7297.

## Details

kfdtest was added as a subproject in #6334. The drift check landed three days before it merged (#6941), so the check was new and its reviewers were not looking for it. It ran on #6334 and failed with these edges, but it is not part of the `CI Summary` gate, so it did not block the merge.

Since the comparison is whole-file, a PR cannot regenerate only its own edges. It either fails the check or commits the 10 kfdtest lines alongside unrelated work. Recent examples: `users/PeterCDMcLean/meson-version-script-per-target`, `users/marifamd/rocm-smi-disable`, `users/lakshmi/windows-mesa-updates`, `users/adam360x/sysdeps-libfwupd`. Open PRs carrying the lines: #7166, #7288, #6922.

`users/stella/hotswap-pre-cache` shows a second set accumulating for `rccl-translate`.

## Proposed actions

1. Land #7297 so the edges are attributed to kfdtest and reviewed by its owners.
2. Decide whether the drift check should gate merge, or document it as advisory.

## Scope

Regenerating does not change behavior. Selection can emit `kfdtest`, but there is no `kfdtest` key in `fetch_test_configurations.py`, so no job is scheduled. Adding one belongs with the kfdtest and rocr-runtime owners.

cc @andmar-amd @nunnikri @ScottTodd @endurthiabhilash

BrianHarrisonAMD:
Splitting this out of #7166 so the edges are reviewed by the kfdtest owners rather than riding along in an ml-libs change.

- #7297 regenerates the graph, kfdtest edges only.
- #7166 is now stacked on #7297 and carries a single graph line for its own rocrand edge.

#7297 needs to land first, otherwise #7166 and the other graph-touching PRs keep failing the drift check.

BrianHarrisonAMD:
Closing for now, regeneration PR has merged.

PR #7297, opened 2026-08-12
fix(ci): regenerate consumer graph for kfdtest edges

BrianHarrisonAMD:
## Motivation

kfdtest was added as a subproject in #6334, but the committed consumer graph was never regenerated. The drift check compares the whole file, so any PR touching a graph-affecting path either fails it or has to carry these lines in an unrelated diff.

GitHub issue: https://github.com/ROCm/TheRock/issues/7298

## Technical Details

Regenerated `test_tools/therock_consumer_graph.json` with `check_consumer_graph_drift.py --write`. No other files change.

The diff is the kfdtest node plus its seven reverse edges from `core/CMakeLists.txt`: `amd-llvm`, `rocr-runtime`, `therock-libdrm`, `therock-numactl`, `therock-yaml-cpp`, `therock-zlib`, `therock-zstd`.

No behavior change. Selection can now emit `kfdtest`, but there is no `kfdtest` key in `fetch_test_configurations.py`, so no job is scheduled.

## Test Plan

Drift check, policy validation, selection unit tests, pre-commit.

## Test Result

- `--validate-policies`: OK, 18 component keys against 116 graph nodes
- `determine_rocm_test_dependencies_test.py`: 49 passed
- pre-commit: passed

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md.

PR #7440, opened 2026-08-18
fix(ci): regenerate stale committed consumer graph

avansick-amd:
Co-authored-by: Claude Opus 5 <noreply@anthropic.com>
Signed-off-by: Adam Van Sickle <adam.vansickle@amd.com>

## Motivation

`rocprofiler-compute` gained a `THEROCK_BUNDLED_ZLIB` runtime dep in #7421, but the
committed consumer graph was never regenerated. The drift check compares the whole
file, so any PR touching a graph-affecting path either fails it or has to carry this
line in an unrelated diff. Same class as #7298 / #7297.

## Technical Details

Regenerated `test_tools/therock_consumer_graph.json` with
`check_consumer_graph_drift.py --write` from a full-tree `THEROCK_ENABLE_ALL=ON`
`gfx94X-dcgpu` configure. No other files change.

The diff is one line: `rocprofiler-compute` added to `therock-zlib`'s consumers.

`rocprofiler-compute` has a key in `fetch_test_configurations.py`, 
so changes to `therock-zlib` can now schedule its tests. 
That follows from the dependency #7421 added.

## Test Plan

- Regenerated locally, then confirmed the result matches what CI independently
computed on #7376.

## Test Result

- Consumer graph drift check: pass.

## Submission Checklist

### DIFF
--- a/test_tools/therock_consumer_graph.json
+++ b/test_tools/therock_consumer_graph.json
@@ -834,6 +834,7 @@
       "kfdtest",
       "rdc",
       "rocgdb",
+      "rocprofiler-compute",
       "rocr-runtime",
       "therock-elfutils",
       "therock-grpc"


ScottTodd:
That's a very long AI generated PR description for what could be summarized in one sentence as:
> Commit <> did not update `test_tools/therock_consumer_graph.json`, this fixes it.

The one additional detail that is useful in the current description is the lack of a `push` trigger in the workflow for post-merge coverage that would make this easier to track through the history. We will not "regenerate the committed graph after a merge" automatically though, as that would require granting automation commit access.

Please read https://github.com/ROCm/TheRock/blob/main/CONTRIBUTING.md#ai-tool-use-policy for future PRs.

ScottTodd:
For reference, here's another PR updating the committed graph: https://github.com/ROCm/TheRock/pull/7297. Still verbose, but about half as much text.

avansick-amd:
> For reference, here's another PR updating the committed graph: #7297. Still verbose, but about half as much text.

Thanks!  I was just looking through that one actually.  I was about to update with a new description more or less lifted from it.

ScottTodd (approved):
Passing here: https://github.com/ROCm/TheRock/actions/runs/32096784259/job/95589599898, lgtm.

I think we should also add a `push` trigger to https://github.com/ROCm/TheRock/blob/main/.github/workflows/test_consumer_graph_drift.yml to make it easier to spot when it started failing (it takes ~10 minutes to run on github-hosted runners so running on only pushes affecting the relevant files is a nice optimization instead of running on push for all file changes)

PR #7443, opened 2026-08-18
[ci] Add quartz_tracking_id to all release-related workflows

### LINKED ISSUE #7322: [ci] Add optional input parameter to workflows for quartz tracking
For quartz, itis important to create the correct graphs which release workflow dispatched which workflows. Github does not offer this for free.

Right now we are extracting from various input parameters including URLs and artifact_run_id. However this is rather fragile and pytorch test workflows do not have any way of tracking it.

Best and safest would be to introduce a new optional input parameter for all workflows triggering quartz_notify to have a consistent end-to-end tracking which workflows were started by top level orchestrator workflow.

HereThereBeDragons:
## Summary

This PR adds an explicit `quartz_tracking_id` input to every workflow in the
release graph reachable from `multi_arch_release.yml`, and threads it down every
internal edge. The goal is to give Quartz one authoritative field that identifies
the top-level run a workflow belongs to, so release lineage can be reconstructed
directly instead of inferred.

Today Quartz recovers a run's owner with a fragile heuristic: it probes unrelated
inputs (`artifact_run_id`, artifact-bucket URLs) and finally the immediate GitHub
parent. For workflows dispatched asynchronously through `benc-uk/workflow-dispatch`
(the PyTorch and JAX test suites), a fresh `run_id` is assigned and no signal
survives, so their lineage cannot be tracked at all. Carrying the id explicitly
removes the guesswork.

This change is the TheRock-side plumbing only. The Quartz classifier change and
the rockrel caller change are deliberate follow-ups (see below).

Closes https://github.com/ROCm/TheRock/issues/7322

_This will also enable to explicitly track for bkc or asan releases if needed at a later point._

## What changed

- `multi_arch_release.yml` gains a boolean `enable_quartz_tracking` input
  (default `false`) in both its `workflow_call` and `workflow_dispatch` blocks. It
  does **not** gain a `quartz_tracking_id` input, since it is the lineage root. When
  enabled, it computes the value as `<github.run_id>;<release_type>` and passes it
  to its children. In a `workflow_call` chain `github.run_id` is the top-level run
  id, which is exactly the id Quartz needs.
- Every other reachable workflow (25 files) declares a plain string
  `quartz_tracking_id` input (default `""`).
- The value is threaded down all internal edges: 20 reusable `workflow_call` edges
  via their `with:` blocks, and 7 asynchronous `benc-uk/workflow-dispatch` edges via
  their JSON `inputs:` payloads, where the key is placed last.
- The verbose "leave empty, set automatically" description is kept only in blocks
  whose sibling inputs are also documented (the UI-facing `workflow_dispatch` blocks
  and a few majority-documented `workflow_call` blocks); elsewhere the declaration is
  bare, matching the surrounding style.

The separator is `;` rather than `-`, because some release types contain `-`.

## How it reaches Quartz (no action change needed)

Every `notify_quartz` step already sends `workflow_inputs: ${{ toJSON(inputs) }}`.
Once `quartz_tracking_id` is a declared input, it is included in that payload
automatically. No change to the `notify_quartz` composite action or its script is
required.

## Merge ordering (important)

The asynchronous `benc-uk/workflow-dispatch` steps set no `repo:`, so under a
rockrel-triggered run they dispatch to `github.repository`, which resolves to
**rockrel**, and land on rockrel's wrapper copies (`test_artifacts.yml`,
`test_native_linux_packages_install.yml`, `multi_arch_release_linux_pytorch_wheels.yml`,
`multi_arch_release_linux_jax_wheels.yml`, `multi_arch_release_windows_pytorch_wheels.yml`,
`test_pytorch_wheels_full.yml`). Because the dispatch payload now always includes a
`quartz_tracking_id` key, and the `workflow_dispatch` API rejects undefined inputs
with `422 Unexpected inputs provided`, those wrappers must declare and forward the
input before this PR reaches production. **This PR should land together with (or
after) the matching rockrel wrapper update.** Direct TheRock dev dispatches are
unaffected, since there `github.repository` is TheRock and the targets already
declare the input.

## Out of scope (follow-ups)

- rockrel: set `enable_quartz_tracking: true` on the `multi_arch_release.yml` call,
  and add `quartz_tracking_id` to the six async wrapper copies.
- Quartz `therock_classify.py`: prefer the new `quartz_tracking_id` over the URL and
  parent heuristics in `derive_effective_owner_run_id()`.
- The CI graph (`multi_arch_ci*.yml`), `multi_arch_release_asan.yml`, and the `_ci`
  wheel variants are not plumbed; where they share a reusable workflow they simply
  default the input to `""`.

## Test plan

- [x] `pre-commit run actionlint --all-files` passes.
- [x] Static audit: all 25 reachable workflows declare the input; all 20 reusable
      edges and all 7 async edges thread it; `enable_quartz_tracking` appears only in
      `multi_arch_release.yml`.
- [ ] End-to-end dev dispatch on TheRock with `enable_quartz_tracking` on, confirming
      the value appears in Quartz payloads across the graph.
- [ ] Coordinated rockrel wrapper PR merged so nightly async dispatch does not 422.


HereThereBeDragons:
https://github.com/ROCm/TheRock/actions/runs/32134565582

setup.yml: 
`inputs: quartz_tracking_id: 32134565582;dev`

.github/workflows/multi_arch_release.yml
```diff
@@ -59,6 +59,15 @@ on:
...
         default: ""
+      # Enable Quartz release-lineage tracking. When on, this run's id is
+      # propagated to every triggered workflow as quartz_tracking_id, formatted
+      # as "<github.run_id>;<release_type>". Tracking is currently only
+      # meaningful for rockrel-initiated runs; enabling it for a manual TheRock
+      # dispatch has no effect on Quartz tracking.
+      enable_quartz_tracking:
+        description: "Enable Quartz release-lineage tracking (off by default)."
+        type: boolean
+        default: false
```
ScottTodd, line 70: > * Static audit: all 25 reachable workflows declare the input; all 20 reusable
>       edges and all 7 async edges thread it; `enable_quartz_tracking` appears only in
>       `multi_arch_release.yml`.

We can also add a unit test for this.

See my agent's full suggestion at https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7443.md#suggestion-turn-the-static-propagation-audit-into-a-test . I'm actually pretty shocked it caught this and knew to recommend extending https://github.com/ROCm/TheRock/blob/main/build_tools/github_actions/tests/workflow_dispatch_inputs_test.py.

> Consider extending the existing
> `build_tools/github_actions/tests/workflow_dispatch_inputs_test.py` machinery or
> adding a focused graph test that starts at `multi_arch_release.yml` and asserts:
> 
> * every reachable non-root workflow declares `quartz_tracking_id`;
> * every local reusable-workflow edge forwards it; and
> * every `benc-uk/workflow-dispatch` payload includes it and its target accepts
>   it.

ScottTodd (approved):
There are some merge conflicts to resolve but otherwise LGTM. I considered if we could instead use `GITHUB_RUN_ID` for `workflow_call` and quartz's tracking for `workflow_dispatch` to save on needing to edit so many files (including any future workflows), but the consistency will make it harder to accidentally miss a file.

.github/workflows/multi_arch_release.yml
ScottTodd: FYI I poked the agent and it said
```
• After accepting the explicit-propagation design, I looked for existing workflow-graph validation rather than proposing
  a new parser. I searched TheRock for terms around the mechanism:

  workflow_call
  benc-uk/workflow-dispatch
  workflow graph
  actionlint

  That surfaced:

  build_tools/github_actions/tests/workflow_dispatch_inputs_test.py

  I then read it and found it already:

  - discovers every benc-uk/workflow-dispatch invocation;
  - parses its JSON inputs;
  - resolves the target workflow locally;
  - rejects undefined inputs;
  - verifies required inputs are supplied; and
  - validates literal choice values.
```

We have a new section in TESTING.md for this that provides more directly guidance now though: https://github.com/ROCm/TheRock/blob/main/TESTING.md#therock-feature-area-github-actions-workflows . I'll reference that from https://github.com/ROCm/TheRock/blob/main/CLAUDE.md (and maybe rename to AGENTS.md or whatever is more portable to other agents)

HereThereBeDragons:
> There are some merge conflicts to resolve but otherwise LGTM. 

oh :( i had resolved them just before adding you as reviewer. will do it again before merging.

> I considered if we could instead use `GITHUB_RUN_ID` for `workflow_call` and quartz's tracking for `workflow_dispatch` to save on needing to edit so many files (including any future workflows), but the consistency will make it harder to accidentally miss a file.

please lets not do it. the quartz_tracking_id will allow us to add more info than just the run_id. e.g. right now we have it as `<run_id>;<release_type>` e.g. `32134565582;dev`. 
With this it will be easier to get each workflow directly assigned to the correct status.json/database input, independent if it is a normal release, bkc or asan.

.github/workflows/multi_arch_release.yml
HereThereBeDragons: i can have a look - but will be a separate pr. right now we dont want to track it in every workflow - just the workflows that are reachable via a release run started from rockrel.

arvindcheru (approved):
LGTM

.github/workflows/build_portable_linux_python_packages.yml
```diff
@@ -48,6 +48,13 @@ on:
         description: "Branch, tag, or SHA to checkout. Defaults to the triggering ref."
         type: string
         default: ""
+      quartz_tracking_id:
```
nunnikri, line 51: If there is no need for user to set this field. do we need this field in workflow_dispatch

HereThereBeDragons: especially for workflow dispatch we need it as we loose all information about the triggering release workflow

extract from the pr description:

> The goal is to give Quartz one authoritative field that identifies
the top-level run a workflow belongs to, so release lineage can be reconstructed
directly instead of inferred.
>
> Today Quartz recovers a run's owner with a fragile heuristic: it probes unrelated
inputs (artifact_run_id, artifact-bucket URLs) and finally the immediate GitHub
parent. For workflows dispatched asynchronously through benc-uk/workflow-dispatch
(the PyTorch and JAX test suites), a fresh run_id is assigned and no signal
survives, so their lineage cannot be tracked at all. Carrying the id explicitly
removes the guesswork.

.github/workflows/multi_arch_release.yml
ScottTodd: We'd walk the workflow call graph starting from `multi_arch_release.yml`, similar to how we walk it from `multi_arch_ci.yml` here: https://github.com/ROCm/TheRock/blob/48d493e696d795a0aa49c8f39969345eef8f4257/build_tools/github_actions/tests/configure_ci_path_filters_test.py#L215-L241

PR #7449, opened 2026-08-18
ci: restrict rocgdb-corefile tests to gfx942

lumachad:
## Summary

- Add an `include_family` opt-in filter alongside the existing `exclude_family` in the test matrix, so a job can list the gfx architectures it runs on instead of enumerating every architecture to skip. A job runs only when it matches an include (if any are listed for the platform) and matches no exclude.
- Restrict the rocgdb-corefile tests to gfx942 via `include_family`, since only gfx942 runners have GPU core-dump support. This also avoids extending an exclude list as new architectures come online.
- Add unit tests covering the include-only, exclude-only, both, and neither selection paths, plus the corefile behavior.

JIRA ID : N/A

## Test plan

- [x] Verify rocgdb-corefile job appears in the test matrix for gfx942 builds
- [x] Verify rocgdb-corefile job is absent from the test matrix for gfx110X, gfx120X, gfx1151, gfx90a, gfx950, and gfx103X builds

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -402,6 +402,8 @@ test_matrix = {
     },
     # Corefile tests require specific hardware support (GPU core dump capable runners).
     # test_runner is pre-pinned so the family-based runner selection loop skips it.
+    # Only gfx942 has core-dump support, so include_family opts the job in to that
+    # family alone rather than enumerating every other architecture to exclude.
     "rocgdb-corefile": {
         **_rocgdb_common,
         "job_name": "rocgdb-corefile",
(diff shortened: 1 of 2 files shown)


lumachad:
I'm considering adding the reverse filter, where we add the only gfx's we support running things on.

ScottTodd:
> I'm considering adding the reverse filter, where we add the only gfx's we support running things on.

That sounds easier to maintain, yes. Include list + exclude list? Similar to the logic here? https://github.com/ROCm/TheRock/blob/f4d4233e73e1e8d73f8e1eab03bde0f958cdf43e/build_tools/fetch_artifacts.py#L138-L166

geomin12 (approved):

PR #7451, opened 2026-08-18
Relocate /opt/rocm compat symlinks from amdrocm-core to amdrocm-sysdeps

lucbruni-amd:
Partially addresses https://github.com/ROCm/ROCm/issues/6599 (symlink point only).

## Motivation

`/opt/rocm/{bin,lib,include,core,...}` compat symlinks were created only by `amdrocm-core`, which is arch-specific and pulls the full math stack. Metapackages that don't depend on it (`amdrocm-developer-tools`, `amdrocm-opencl`, `amdrocm-runtime-dev`) install without them.

## Technical Details

Move the directory `update-alternatives` from `amdrocm-core` to `amdrocm-sysdeps`, registered from a post-unpack hook so every target exists: a dpkg path trigger on deb, an rpm `%transfiletriggerin`. Registration skips any slot whose target is absent, so no directory is pre-created (an earlier `mkdir` clobbered llvm's `amdgcn`/`llvm` symlinks and broke the rpm unpack). `amdrocm-core` keeps its `/usr/bin` slaves.

TODO (follow-up): dedup `is_debian_like`/`altscore` into the shared partial for `amdrocm-core`/`-devel`, and decouple `debian/triggers` emission from the `Postinstall` key.

## Test Plan

ubuntu:24.04 (deb) and rockylinux:9 (rpm): install a lean meta with a patched `amdrocm-sysdeps`; check the `/opt/rocm` symlinks, upgrade, and uninstall. `bash -n` all rendered scripts.

## Test Result

Both formats: all symlinks created (incl. symlink-typed `amdgcn`/`llvm`), no cpio conflict, `amdrocm-core` not pulled, prerm/erase clean. The old `mkdir` approach reproduced the CI failure; the trigger approach installs clean.

## Submission Checklist

### DIFF
--- a/build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
+++ b/build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
@@ -1,8 +1,7 @@
 #!/bin/bash
 {% include "template/scripts/_alternatives_lib.j2" %}
 {% if target == "deb" %}
-# deb: run on initial configure and whenever the dpkg path trigger fires after
-# payload packages have unpacked files under the prefix (see amdrocm-sysdeps triggers).
+# deb: register on configure and on the dpkg trigger (fires after payload unpack).
 case "$1" in
     (configure|triggered) alt_install ;;
     (abort-upgrade|abort-remove|abort-deconfigure) : ;;
--- a/build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
+++ b/build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
@@ -1,71 +1,15 @@
 #!/bin/bash
-
-is_debian_like() {
-    local id_like="$1"
-    case " $id_like " in
-        (*" debian "*|*" ubuntu "*)
-            return 0
-            ;;
-        (*)
-            return 1
-            ;;
-    esac
-}
-
-do_update_alternatives(){
-    # skip update if program doesn't exist
-    command -v update-alternatives >/dev/null || return 0
-    local altscore now
-    now=$(date -u +%s)
-
-    # Highest semver wins (else newest install); score must fit in 32 bits.
-    altscore=$(({{ version_major }} - 3))
-    altscore=$((altscore * 14 + {{ version_minor }})) # Allow up to 14 minor
-    altscore=$((altscore * 14 + {{ version_patch }})) # Allow up to 14 patch
-    altscore=$((altscore*1000000+(now-1600000000)/60))
-
-    {% if target == "deb" %}
-       PKG_INSTALL_PREFIX={{ install_prefix }}
-    {% else %}
-       PKG_INSTALL_PREFIX="$RPM_INSTALL_PREFIX0"
-    {% endif %}
-
-    # Guard against an empty prefix so the mkdir below never lands at /.
-    [ -n "$PKG_INSTALL_PREFIX" ] || return 0
-
-    # Seed dirs first: update-alternatives errors on a missing target, and sysdeps configures before payload ships them.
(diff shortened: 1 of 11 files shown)


build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+is_debian_like() {
```
lucbruni-amd: Recycled from https://github.com/ROCm/TheRock/blob/c6fac2c1bda525614d39bd478766045d45cca596/build_tools/packaging/linux/template/scripts/amdrocm-core-devel-postinst.j2#L3-L13

build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
```diff
@@ -0,0 +1,71 @@
...
+}
+
+do_update_alternatives(){
+    # skip update if program doesn't exist
+    command -v update-alternatives >/dev/null || return 0
+    local altscore now
+    now=$(date -u +%s)
+
+    # Highest semver wins (else newest install); score must fit in 32 bits.
+    altscore=$(({{ version_major }} - 3))
```
lucbruni-amd: Recycled from https://github.com/ROCm/TheRock/blob/c6fac2c1bda525614d39bd478766045d45cca596/build_tools/packaging/linux/template/scripts/amdrocm-core-devel-postinst.j2#L18-L45

nunnikri:
Approach seems ok to me. But will let me give some thoughts before approving. @raramakr could you also review this?

nunnikri:
@lucbruni-amd seeing multiple failures in ci.

https://github.com/ROCm/TheRock/actions/runs/32160007110/job/95814568512?pr=7451
install failures and scriplet errors. Can you check?

build_tools/packaging/linux/template/scripts/amdrocm-sysdeps-postinst.j2
```diff
@@ -0,0 +1,71 @@
...
+       PKG_INSTALL_PREFIX={{ install_prefix }}
+    {% else %}
+       PKG_INSTALL_PREFIX="$RPM_INSTALL_PREFIX0"
+    {% endif %}
+
+    # Guard against an empty prefix so the mkdir below never lands at /.
+    [ -n "$PKG_INSTALL_PREFIX" ] || return 0
+
+    # Seed dirs first: update-alternatives errors on a missing target, and sysdeps configures before payload ships them.
+    mkdir -p "$PKG_INSTALL_PREFIX"/bin "$PKG_INSTALL_PREFIX"/lib "$PKG_INSTALL_PREFIX"/include "$PKG_INSTALL_PREFIX"/llvm "$PKG_INSTALL_PREFIX"/amdgcn "$PKG_INSTALL_PREFIX"/libexec "$PKG_INSTALL_PREFIX"/share
```
nunnikri: Better to check if mkdir is success or not

if ! mkdir -p "$PKG_INSTALL_PREFIX"/bin "$PKG_INSTALL_PREFIX"/lib ...; then
    echo "Error: Failed to create required directories under $PKG_INSTALL_PREFIX" >&2
    exit 1
fi

lucbruni-amd: I see. Thanks for reviewing. This is also related to the failing jobs as `amdgcn`/`llvm` are shipped as symlinks so the `rpm` aborted. I dropped this completely in the latest commits moving registration to `amdrocm-sysdeps` which everything deps on, and uses a hook that registers the slot only if target exists. I validated w/ `deb`/`rpm` locally but this will need re-review. Resolving this thread.

PR #7236, opened 2026-08-10
fix(ci): pass build_runs_on to PyTorch CI workflow for builds

geomin12:
The PyTorch CI workflow was hardcoded to use aws-linux-scale-rocm-prod, ignoring the ASAN runner selected by configure_multi_arch_ci.py

Now, all PyTorch CI workflows are dynamically retrieved

Changes:
- Add build_runs_on input to all pytorch CI workflows
- Pass build_runs_on from build_config in multi_arch_ci_linux.yml

PR #7452, opened 2026-08-18
fix(ci): add build_runs_on input to workflow_dispatch for PyTorch builds

geomin12:
PR #7236 added build_runs_on to workflow_call but missed adding it to workflow_dispatch. This caused manual dispatches to use the hardcoded default runner, leading to OOM failures on resource-constrained runners.

Add the missing build_runs_on input to workflow_dispatch for both Linux and Windows PyTorch wheel build workflows, allowing users to specify appropriate runners when dispatching manually.

### DIFF
--- a/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
+++ b/.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
@@ -83,7 +83,7 @@ on:
       build_runs_on:
         description: "Build runner label (selected by configure script with weighted distribution)"
         type: string
-        default: "azure-windows-scale-rocm"
+        default: "aws-windows-scale-rocm-prod-mix"
   workflow_dispatch:
     inputs:
       amdgpu_families:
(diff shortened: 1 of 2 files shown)


.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
```diff
@@ -162,6 +162,10 @@ on:
           - sccache
           - ccache
         default: sccache
+      build_runs_on:
+        description: "Build runner label (selected by configure script with weighted distribution)"
+        type: string
+        default: "azure-windows-scale-rocm"
```
ScottTodd: Want to use AWS instead of Azure here and for the `workflow_call` default?

I see that @amd-justchen is routing back from 50%/50% AWS/Azure to 100% AWS with https://github.com/ROCm/therock-ci-config/pull/30

Looks like the current label is `aws-windows-scale-rocm-prod-mix`

ScottTodd (approved):

.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
geomin12: ah! i thought it was experimental, but seems it is official. let me update

ScottTodd (approved):

.github/workflows/multi_arch_build_windows_pytorch_wheels.yml
amd-justchen: > Want to use AWS instead of Azure here and for the `workflow_call` default?
> 
> I see that @amd-justchen is routing back from 50%/50% AWS/Azure to 100% AWS with [ROCm/therock-ci-config#30](https://github.com/ROCm/therock-ci-config/pull/30)
> 
> Looks like the current label is `aws-windows-scale-rocm-prod-mix`

That's correct!

PR #156, opened 2025-03-05
Adding TheRock sanity checks

geomin12:
Adding sanity checks, as well as some workflow configuration so we don't have to build the packages everytime we want to run a test workflow

(just to get this on paper, next will be adding integration tests)

PR #7457, opened 2026-08-18
fix(tests): disable for gfx125X in hip_printf sanity test due to unsupported kernel code pattern

geomin12:
## Summary

Skip `test_hip_printf` for gfx125X-dcgpu (MI455) due to a binary translator limitation.

## Problem

The gfx1250 binary translator cannot handle indirect branch instructions (`s_set_pc_i64`) generated by this test's kernel code pattern. The test fails with:

```
[hsa-hotswap-rj] error: indirect branch or call target recovery is not implemented for relocated kernel text
```

See #7458 for full details and reproduction steps.

## Changes

- Add `@pytest.mark.skipif` for `AMDGPU_FAMILIES == "gfx125X-dcgpu"`

### DIFF
--- a/tests/test_rocm_sanity.py
+++ b/tests/test_rocm_sanity.py
@@ -87,6 +87,11 @@ class TestROCmSanity:
         is_windows(),
         reason="Windows offload-arch.exe is not retrieving correct data, ignoring test",
     )
+    # TODO(#7458): Re-enable once gfx1250 binary translator supports this kernel code pattern
+    @pytest.mark.skipif(
+        AMDGPU_FAMILIES == "gfx125X-dcgpu",
+        reason="gfx1250 binary translator does not yet support this kernel code pattern, see #7458",
+    )
     def test_hip_printf(self):
         platform_executable_suffix = ".exe" if is_windows() else ""
 
--- a/tests/test_rocm_sanity.py
+++ b/tests/test_rocm_sanity.py
@@ -89,7 +89,7 @@ class TestROCmSanity:
     )
     # TODO(#7458): Re-enable once gfx1250 binary translator supports this kernel code pattern
     @pytest.mark.skipif(
-        AMDGPU_FAMILIES == "gfx125X-dcgpu",
+        AMDGPU_FAMILIES and "gfx125X-dcgpu" in AMDGPU_FAMILIES,
         reason="gfx1250 binary translator does not yet support this kernel code pattern, see #7458",
     )
     def test_hip_printf(self):


rahulc-gh (approved):

tests/test_rocm_sanity.py
```diff
@@ -87,6 +87,11 @@ def test_rocm_output(self, rocm_info_output, to_search):
         is_windows(),
         reason="Windows offload-arch.exe is not retrieving correct data, ignoring test",
     )
+    # TODO(#7458): Re-enable once gfx1250 binary translator supports this kernel code pattern
+    @pytest.mark.skipif(
+        AMDGPU_FAMILIES == "gfx125X-dcgpu",
+        reason="gfx1250 binary translator does not yet support this kernel code pattern, see #7458",
+    )
     def test_hip_printf(self):
         platform_executable_suffix = ".exe" if is_windows() else ""
```
ScottTodd, line 96: We could restructure the test to not printf, or add a second test. Sanity checks are pretty important to keep running and passing, I think the other GPU compile + launch behavior of the test is more important than print itself.

https://github.com/ROCm/TheRock/blob/main/tests/hip_check.cpp

tests/test_rocm_sanity.py
```diff
@@ -87,6 +87,11 @@ def test_rocm_output(self, rocm_info_output, to_search):
         is_windows(),
         reason="Windows offload-arch.exe is not retrieving correct data, ignoring test",
     )
```
ScottTodd, line 89: Let's follow-up on this too, https://github.com/ROCm/TheRock/issues/4755 is fixed now and I thought we re-enabled this on Windows already 🤔

tests/test_rocm_sanity.py
```diff
@@ -87,6 +87,11 @@ def test_rocm_output(self, rocm_info_output, to_search):
...
+    # TODO(#7458): Re-enable once gfx1250 binary translator supports this kernel code pattern
+    @pytest.mark.skipif(
+        AMDGPU_FAMILIES == "gfx125X-dcgpu",
```
ScottTodd: Is this `==` enough, or do we want `contains()` instead? The environment variable name suggests that it might be a list and could include multiple families

ScottTodd (commented):
Ok to unblock testing as needed, but a few details to be careful about

tests/test_rocm_sanity.py
geomin12: ah yes let us follow-up on that

tests/test_rocm_sanity.py
geomin12: we can do contains, safer

tests/test_rocm_sanity.py
geomin12: actually, the issue is deeper than printf. I updated issue here: https://github.com/ROCm/TheRock/issues/7458
seems minimal kernels fail, so hipcc kernel execution is of no value.

ScottTodd: 🤔 interesting, okay. I assume developers working on the relevant parts of the system are aware? I don't want to just turn off test coverage in the CI system without that.

ScottTodd (approved):

dan-garvey (approved):

tests/test_rocm_sanity.py
geomin12: yeah @kuhar is aware. it's noted as a "won't fix" :( on the JIRA tix, as other higher priorities. msg from Jakub: "it's one of the features that the translator can't support without major heroics".

we'll keep on radar so we won't forget

ScottTodd: Well, our "minimal kernel" for sanity checks should be something that passes. If this printf and other behavior is not representative of what users programs do / is not high priority enough to fix then we should move it to an extended test and use something else for sanity checks.

kuhar: you can have it add two arrays or something similar -- printf is actually not entirely straightforward on gpu

ScottTodd: Cool, we can definitely make edits in that direction. This sanity check test was added along time ago (https://github.com/ROCm/TheRock/pull/156) 😅

Issue #6948, opened 2026-07-28
ROCm Python Indexes: Aggregate whl-next Index


marbre:
## Goal

Generate a pip-compatible aggregate Simple API index for:

```text
/rocm/whl-next/
```

by aggregating the existing product-local indexes for ROCm Core, PyTorch, and
JAX.

The aggregate index generator may live in **TheRock** and be reused by
**TheRock-Infra**. **TheRock-Infra** remains responsible for the authoritative
deployment workflow, privileged S3 access, CloudFront routing deployment, and
publication of the generated aggregate index.

The generator should be implemented as a standalone tool that does not depend
on TheRock-specific publishing workflows.

### In scope

- Support `/rocm/whl-next/` only.
- Establish a checked-in package ownership manifest.
- Validate that manifest against the published product-local indexes.
- Generate `/rocm/whl-next/index.html`.
- Generate the exact package routing table consumed by the CloudFront Function.
- Generate both artifacts from the same validated ownership manifest.
- Produce validation output suitable for local development and CI.

### Out of scope

- `/rocm/whl/`.
- Changes to product publication workflows.
- Changes to `manage_structured.py`.
- Extras packages.
- Tracking which repository publishes a package.
- Legacy bare-domain indexes.
- CloudFront deployment logic inside the aggregate generator.

## Package ownership manifest

Maintain a checked-in manifest containing exact PEP 503-normalized package names
and their product-local public owner paths.

Example:

```yaml
schema_version: 1

python_indexes:
  - public_base: /rocm/whl-next
    packages:
      rocm-sdk-core:
        owner_path: core/whl-next
      rocm-sdk-libraries:
        owner_path: core/whl-next
      torch:
        owner_path: pytorch/whl-next
      torchvision:
        owner_path: pytorch/whl-next
      jax:
        owner_path: jax/whl-next
      jaxlib:
        owner_path: jax/whl-next
```

The manifest is the authoritative package ownership policy. It may be authored
manually or with the help of discovery tooling, but ownership changes must be
reviewed and checked in.

The manifest must use public owner paths only. It must not contain S3 bucket
names, internal S3 prefixes, repository names, workflow names, or release-stream
names.

## Product-local publication contract

Each owner path referenced by the manifest must provide:

```text
/rocm/<owner-path>/index.html
/rocm/<owner-path>/<normalized-package>/index.html
```

For example:

```text
/rocm/pytorch/whl-next/index.html
/rocm/pytorch/whl-next/torch/index.html
[...]

PR #7463, opened 2026-08-18
Add whl-next aggregate index generator

marbre:
## Motivation

Add a schema-versioned ownership manifest and standalone generator for the /rocm/whl-next aggregate Python index tracked in #6948. The generator emits the aggregate Simple API root, exact route table, and validation report.

The generator implements the aggregate index using exact package ownership and per-stream package availability. Stream availability is part of the manifest because package sets differ between release streams; some architectures are not included in release candidates or stable releases but are supported in nightly builds.

This intentionally differs from #6948 by making product-local content validation optional instead of requiring manifest-derived deployment artifacts. The tool can generate aggregate roots and exact route tables directly from the manifest for deployment artifacts. This routing table is used to generate CloudFront functions for routing.

## Technical Details

Optional product-local content validation is available for preflight checks and debugging. This uses an authoritative local snapshot. The validation report records whether output was manifest-derived or content-validated so consumers do not mistake manifest-only generation for verified product-index validation.

## Test Plan

- python -m py_compile build_tools/packaging/python/aggregate_index.py build_tools/packaging/python/tests/aggregate_index_test.py
- python -m pytest build_tools/packaging/python/tests/aggregate_index_test.py
- python build_tools/packaging/python/aggregate_index.py validate-manifest --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml

## Test Result

- Passed: py_compile
- Passed: aggregate_index_test.py (82 tests)
- Passed: validate-manifest

## Submission Checklist

Generated with Codex

### DIFF
--- a/build_tools/packaging/python/aggregate_index.py
+++ b/build_tools/packaging/python/aggregate_index.py
@@ -1,0 +1,1339 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""ROCm aggregate Python Simple API index tooling.
+
+Validates and generates the routed aggregate Simple API index for:
+
+    /rocm/whl-next/
+
+The generator is storage-independent. It consumes a checked-in ownership
+manifest and can optionally validate a local directory tree mirroring the
+public product-local index layout:
+
+    <content-root>/rocm/<owner_path>/index.html
+    <content-root>/rocm/<owner_path>/<normalized-package>/index.html
+
+For example, a package owned by ``pytorch/whl-next`` is validated from:
+
+    <content-root>/rocm/pytorch/whl-next/index.html
+    <content-root>/rocm/pytorch/whl-next/torch/index.html
+
+The required routed baseline writes:
+
+    <output>/rocm/whl-next/index.html
+    <output>/rocm-whl-next-routes.json
+    <output>/validation.json
+
+The aggregate root links to canonical package subdirectories under the
+aggregate namespace, such as ``torch/``. The generated route table maps those
+aggregate package requests to exact product-local package pages. The generator
+does not copy or rewrite product-local package pages; they remain authoritative
+for artifact links and metadata in the routed design.
+
+CLI subcommands:
+
+    # Validate only the ownership manifest.
+    python aggregate_index.py validate-manifest \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml
+
+    # Validate a local public-tree snapshot without writing outputs.
+    python aggregate_index.py validate-content \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml \
+        --content-root /tmp/rocm-whl-next-content
+
+    # Generate routed aggregate outputs directly from the manifest.
+    python aggregate_index.py generate \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml \
+        --stream nightly \
+        --output-dir /tmp/rocm-whl-next-output
+
+    # Validate a local public-tree snapshot while generating outputs.
+    python aggregate_index.py generate \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml \
+        --stream nightly \
+        --content-root /tmp/rocm-whl-next-content \
+        --output-dir /tmp/rocm-whl-next-output
+
+When ``--content-root`` is provided, the deployment validation invocation should
+use ``generate --strict-completeness`` and must not use ``--allow-unpublished``.
+Both set-comparison directions are deployment-visible failures: a manifest-owned
+package that is not published would route to a missing product-local package
+page, while a published package absent from the manifest is unreachable through
+the aggregate index.
+
+``--allow-unpublished`` is only for explicit pre-publication inventory workflows
+that also provide ``--content-root``. In that mode, manifest-owned packages
+absent from product roots are recorded as unpublished in ``validation.json`` and
+excluded from the aggregate root and exact route table.
+"""
+
+import argparse
+import dataclasses
+import html
+import html.parser
+import json
+import re
+import sys
+from collections.abc import Mapping, Sequence
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+from urllib.parse import quote, urlsplit
+
+from packaging.utils import canonicalize_name, is_normalized_name
+import yaml
+from yaml.constructor import ConstructorError
+from yaml.nodes import MappingNode
+
+
+MANIFEST_SCHEMA_VERSION = 2
+ROUTES_SCHEMA_VERSION = 2
+VALIDATION_SCHEMA_VERSION = 2
+SUPPORTED_PUBLIC_BASE = "/rocm/whl-next"
+ROUTES_FILENAME = "rocm-whl-next-routes.json"
+VALIDATION_FILENAME = "validation.json"
+
+_PUBLIC_PATH_SEGMENT_RE = re.compile(r"^[a-z0-9][a-z0-9._-]*$")
+_STREAM_RE = re.compile(r"^[a-z0-9][a-z0-9_-]*$")
+
+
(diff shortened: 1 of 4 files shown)


build_tools/packaging/python/tests/aggregate_index_test.py
```diff
@@ -0,0 +1,1343 @@
...
+            if value in {"ownership.yaml", "content", "out"}
+            else value
+        )
+        for value in argv
+    ]
+    with pytest.raises(SystemExit) as exc_info:
+        main(argv)
+    captured = capsys.readouterr()
+    assert exc_info.value.code == 2
+    assert "unrecognized arguments" in captured.err
```
ScottTodd: This `test_main_rejects_unsupported_content_flag` test checking for `unrecognized arguments` error text is only exercising argparse itself, not any of our own code. We can remove it.

build_tools/packaging/python/aggregate_index.py
```diff
@@ -0,0 +1,1339 @@
...
+CLI subcommands:
+
+    # Validate only the ownership manifest.
+    python aggregate_index.py validate-manifest \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml
+
+    # Validate a local public-tree snapshot without writing outputs.
+    python aggregate_index.py validate-content \
+        --manifest build_tools/packaging/python/rocm_whl_next_ownership.yaml \
+        --content-root /tmp/rocm-whl-next-content
```
ScottTodd, line 45: nit: `--stream` is a required argument on the `validate-content` parser, so this example is not runnable as-is

build_tools/packaging/python/aggregate_index.py
```diff
@@ -0,0 +1,1339 @@
+#!/usr/bin/env python3
+# Copyright Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+"""ROCm aggregate Python Simple API index tooling.
+
+Validates and generates the routed aggregate Simple API index for:
+
+    /rocm/whl-next/
```
ScottTodd, line 9: Can you (in this PR or a follow-up) document how our python packages are designed and generated? A new section in https://github.com/ROCm/TheRock/blob/main/docs/packaging/python_packaging.md seems appropriate.

Agent review at https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7463.md#important-add-durable-user-and-maintainer-documentation suggested topics to cover:
> Cover the manifest ownership/availability contract, the source of stream exceptions, safe deployment invocation, diagnostic modes, artifact schemas, the TheRock-Infra boundary, and worked commands. Link both the historical [PEP 503](https://peps.python.org/pep-0503/) and the maintained [Simple Repository API specification](https://packaging.python.org/en/latest/specifications/simple-repository-api/) where the code relies on normalized project names and simple-index links.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 CppHeaderParser>=2.7.4
 build>=1.2.2
 meson>=1.7.0
+packaging==25.0
```
ScottTodd, line 4: This is fine, but we may want to merge https://github.com/ROCm/TheRock/blob/main/build_tools/packaging/requirements.txt into this file at some point if the overlap is significant.

build_tools/packaging/python/rocm_whl_next_ownership.yaml
```diff
@@ -0,0 +1,275 @@
...
+      amd-torch-device-gfx900:
+        owner_path: pytorch/whl-next
+        streams:
+          - dev
+          - nightly
+      amd-torch-device-gfx906:
+        owner_path: pytorch/whl-next
+        streams:
+          - dev
+          - nightly
```
ScottTodd, line 84: This currently limits `[gfx900, gfx906, gfx90c]` to just the `dev` and `nightly` release streams, matching what we've been triggering for prerelease builds like https://github.com/ROCm/rockrel/actions/runs/31840058334 but https://github.com/ROCm/TheRock/blob/main/SUPPORTED_GPUS.md is still stale and presents conflicting information (e.g. `gfx950` is _not_ marked 'release ready'). We should fix that file and maybe link to it from this file in a comment?

(This was my own finding, then my agent wrote about it in more detail at https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7463.md#important-ownership-and-release-availability-are-now-one-policy-surface)

cc @sa-faizal

build_tools/packaging/python/aggregate_index.py
```diff
@@ -0,0 +1,1339 @@
...
+        help="Path to the ownership manifest YAML file",
+    )
+    validate_parser.set_defaults(func=_validate_manifest_command)
+
+    generate_parser = subparsers.add_parser(
+        "generate",
+        help="generate aggregate routed outputs, optionally validating "
+        "product-local content",
+        allow_abbrev=False,
+    )
```
ScottTodd, line 1278: I'm not familiar enough with the cloud component of this, can you comment on this agent review finding? https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7463.md#important-define-and-test-the-actual-cloudfront-consumer-format-and-size-budget

> For `nightly`, the generated route artifact contains 111 routes and is 18,241 bytes as written. Minifying the current list-of-records shape is still 13,884 bytes. AWS documents a non-adjustable 10 KB maximum for an entire [CloudFront Function](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html#limits-functions), including the routing code. A compact `package -> owner_path` map for the same data is approximately 4,738 bytes before JavaScript is added, so the design is feasible, but only if the Infra translation deliberately discards the repeated fields and remains compact as package count grows.
> 
> The route JSON therefore cannot be treated as an embeddable artifact in its
current form. It is an intermediate representation whose consumer transform is
part of the correctness contract.
>
> **Recommendation:** define the consumer-facing schema with TheRock-Infra now
(the map shape from #6948 is a better starting point), add an end-to-end route
rewrite test, and fail generation/deployment when the final CloudFront source
exceeds the 10 KB quota. Consider CloudFront KeyValueStore if long-term package
growth makes inline data too tight.

build_tools/packaging/python/tests/aggregate_index_test.py
ScottTodd, line 1: I think this test file could use some comments:
* At the top explaining what core behaviors are tested in each section
* In individual test cases explaining what is unique about each test case

I think the tests are good from a _coverage_ perspective but I'm curious to read more about the test _design_ (what is important to test, what is not tested, etc.) and comments are a good place for that

ScottTodd (commented):
Overall I think this is okay structurally. I made a few review passes with an agent with the raw findings at https://github.com/ScottTodd/rocm-workspace/blob/main/reviews/pr_TheRock_7463.md if you want to go through them in a batch. I'll flag a few details in particular that caught my attention.

build_tools/packaging/python/aggregate_index.py
marbre: Maximum right now is 6,251 bytes  for `rocm-whl-next-router-nightly.js` which holds the full routing information for our nightly builds (already checked into TheRock-Infra as an intermediate solution before wiring this fully up). This concern is valid but was taken care of and the JSON isn't embedded into the CF func and only is the foundation.

build_tools/packaging/python/tests/aggregate_index_test.py
ScottTodd: (I'm also skeptical of test code that captures and checks stdout/stderr as this does - logs are not typically load bearing or part of the meaningful API that should be tested)

marbre: Thanks for catching this. I was fearing that the agent wrote some tests like this which I didn't caught, glad you did. Will drop it.

ScottTodd (approved):
We discussed the review together offline. A few of the findings would be nice to fix prior to merge, while others (e.g. documentation) can be follow-ups as more of this work is fully landed.

PR #7479, opened 2026-08-19
ci(rocgdb): Improve wall clock time for rocgdb-* tests

lumachad:
Pass `--parallel -f 0.25` to all three rocgdb test jobs so tests run in parallel using 25% of the available cores on the machine. Also pass `--toolchain llvm` to `rocgdb-gpu` and `rocgdb-corefile` to restrict runs to the LLVM toolchain, which is the compiler used for GPU code paths.

I expect the rocgdb-cpu jobs to get a reasonable speed up due to parallel compilations/runs. The wall clock time for rocgdb-gpu/rocgdb-corefile tests will be slashed in half at least, plus additional reduction from parallel compilations.

@skip-pr-bot

### DIFF
--- a/build_tools/github_actions/fetch_test_configurations.py
+++ b/build_tools/github_actions/fetch_test_configurations.py
@@ -357,13 +357,13 @@ test_matrix = {
     "rocgdb-cpu": {
         **_rocgdb_common,
         "job_name": "rocgdb-cpu",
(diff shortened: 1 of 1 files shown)


lumachad:
This one should reduce the wall clock time for rocgdb CI jobs, hopefully freeing up some capacity.

geomin12 (approved):

PR #7481, opened 2026-08-19
Make bundled zlib/liblzma pkg-config files relocatable via patch_install

idubinov:
## Motivation

SKIP_INSTALL_FILES cmake flag was removed from zlib and liblzma. The flag skips .pc file generation, and TheRock creates its own. The difference in original .pc file and self-written TheRock's files is a relative paths inside original instead of absolute paths in TheRock's. 

This patch suggests using relativize_pc_file helper to process original files instead of using self-written fields.

Bug found during experiments, does not reproduce in CI due to the absolute paths on different CI machines were accidentally the same.

https://github.com/ROCm/TheRock/issues/7482

## Submission Checklist


idubinov:
@aobolensk @jbonnell-amd @vedenev-amd

PR #7483, opened 2026-08-19
ci: raise setup test environment timeout from 15 to 25 minutes

lumachad:
## Summary

- Raises the `Run setup test environment workflow` step timeout from 15 to 25 minutes in `test_component.yml`
- Updates both `timeout-minutes` and the matching `STEP_TIMEOUT_MINUTES` env var (kept in sync per the existing comment)

The step was observed timing out at 15 minutes (e.g. https://github.com/ROCm/TheRock/actions/runs/32165005832/job/96055002747?pr=7449).

## Test plan

- [x] Confirm the setup test environment step completes within 25 minutes in CI

@skip-pr-bot

### DIFF
--- a/.github/workflows/test_component.yml
+++ b/.github/workflows/test_component.yml
@@ -130,10 +130,10 @@ jobs:
           ref: ${{ inputs.ref }}
 
       - name: Run setup test environment workflow
-        timeout-minutes: 15
+        timeout-minutes: 25
         env:
           RETRY_THIS_STEP: "true"
-          STEP_TIMEOUT_MINUTES: "15" # should match timeout-minutes specified above at this step
+          STEP_TIMEOUT_MINUTES: "25" # should match timeout-minutes specified above at this step
         uses: './.github/actions/setup_test_environment'
         with:
           ARTIFACT_RUN_ID: ${{ env.ARTIFACT_RUN_ID }}


lumachad:
I think we're approaching the timeout limit for the setup environment if we need to pull a LOT of artifacts, like the math-libs.

.github/workflows/test_component.yml
```diff
@@ -130,10 +130,10 @@ jobs:
           ref: ${{ inputs.ref }}
 
       - name: Run setup test environment workflow
-        timeout-minutes: 15
+        timeout-minutes: 25
         env:
           RETRY_THIS_STEP: "true"
-          STEP_TIMEOUT_MINUTES: "15" # should match timeout-minutes specified above at this step
+          STEP_TIMEOUT_MINUTES: "25" # should match timeout-minutes specified above at this step
```
ScottTodd, line 136: Please no :(, or at least not for an extended prior outside of ongoing infrastructure incidents. The execution time expectations per https://github.com/ROCm/TheRock/blob/main/docs/development/test_filtering.md are:
* < 5 minutes for 'quick' (sanity check tests)
* < 30 minutes for 'standard' (tests for key functionality)

If we spend this much of our budget on setup that leaves no time for testing.

We should work to optimize the setup steps:
* https://github.com/ROCm/TheRock/issues/3674
* Improve network speed on runners
* Reduce the amount of data needed
    * experiment with compression
    * trim the list of packages in https://github.com/ROCm/TheRock/blob/main/requirements-test.txt

If runners are unable to download artifacts within budgets *they are not healthy enough to be used on CI*

lumachad: That's fair. I can't tell whether these are infra issues we're seeing or whether we grew the artifacts/setup enough that we're taking too long now to run things on time.

ScottTodd: The https://github.com/ROCm/TheRock/actions/runs/32165005832/job/96055002747?pr=7449 run you linked spent 6+ minutes downloading the 64MB [hip-python](https://pypi.org/project/hip-python/) python package but did not actually use it. I commented about that over at https://github.com/ROCm/TheRock/pull/6730#discussion_r3814522438

lumachad: I'm dropping this for now. And the scope of the optimization required here unfortunately is beyond what I have cycles to work with.

ScottTodd: Thanks, and sorry if I sounded too dismissive. If we just increase the timeouts that can help with test failures, but it puts increasing pressure on runner queues that lead to negative feedback loops (longer job time -> longer queues).

PR #7486, opened 2026-08-19
fix: external repo pin SHA in TheRock manifest

erman-gurses:
## Summary
Issue: https://github.com/ROCm/TheRock/issues/7487
Fix `therock_manifest.json` so external/super-repo CI records the commit that was actually built for the externally supplied repository.

Previously, manifest generation always derived submodule `pin_sha` values from the TheRock tree using `git ls-tree <the_rock_commit>`. This is correct for normal TheRock builds, but incorrect for super-repo CI when `rocm-systems`, `rocm-libraries`, or another supported external repo is checked out at a newer SHA than the submodule SHA currently pinned in TheRock.

For example, super-repo CI could build:

- `rocm-systems`: `c1dba564...`

while TheRock still pins:

- `rocm-systems`: `ef3b147...`

The build used `c1dba564...`, but the generated manifest incorrectly recorded `ef3b147...`.

## Changes

- Add `submodule_path` to the serialized `external_repo_config`.
- Pass the existing `external_repo_config` to Linux and Windows artifact build jobs through `EXTERNAL_REPO_CONFIG`.
- Update `generate_therock_manifest.py` to:
  - keep using the TheRock gitlink SHA for normal builds;
  - use `external_repo.ref` for the matching externally supplied submodule;
  - match the external repository to the manifest entry using `submodule_path`;
  - reject invalid external repo configs or unmatched submodule paths.

No CMake changes are required because the job-level environment is inherited by the manifest generation custom command.

## Behavior

Normal TheRock build:

pin_sha = SHA pinned by TheRock submodule gitlink

External/super-repo build:

matching external submodule pin_sha = external_repo.ref
all other submodules                = TheRock gitlink SHA

## Validation

Validated locally for:

- normal TheRock manifest generation;
- `rocm-systems` external override;
- `rocm-libraries` external override;
- `ROCgdb` nested submodule path override;
- non-target submodules remaining unchanged;
- partial manifest generation without git metadata;
- invalid/unmatched submodule paths;
- missing required external repo fields;
- invalid external repo JSON types;
- Linux and Windows workflow environment wiring.

Relevant unit tests:

61 passed

across:

- `build_tools/github_actions/tests/detect_external_repo_config_test.py`
- `build_tools/github_actions/tests/dvc_external_repo_test.py`
- `build_tools/tests/configure_external_repo_ci_test.py`

Final validation will be through a super-repo CI run to confirm the packaged `therock_manifest.json` contains the external commit SHA that was actually built.

### DIFF
--- a/build_tools/generate_therock_manifest.py
+++ b/build_tools/generate_therock_manifest.py
@@ -179,6 +179,27 @@ def rocm_version_from_package_version(rocm_package_version: str) -> str | None:
     return match.group(1)
 
 
+def parse_external_repo_config(raw: str | None) -> dict | None:
+    """Parse external repo configuration used by super-repo CI."""
+    if not raw:
+        return None
+
+    config = json.loads(raw)
+
+    if not isinstance(config, dict):
+        raise ValueError("EXTERNAL_REPO_CONFIG must be a JSON object")
+
+    submodule_path = config.get("submodule_path")
+    ref = config.get("ref")
+
+    if not submodule_path or not ref:
+        raise ValueError(
+            "EXTERNAL_REPO_CONFIG must contain both 'submodule_path' and 'ref'"
+        )
+
+    return config
+
+
 def build_manifest_schema(
     repo_root: Path,
     the_rock_commit: str,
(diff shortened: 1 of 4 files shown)


erman-gurses (commented):
I did the local tests - currently running CI test.

build_tools/generate_therock_manifest.py
```diff
@@ -179,21 +179,59 @@ def rocm_version_from_package_version(rocm_package_version: str) -> str | None:
     return match.group(1)
 
 
+def parse_external_repo_config(raw: str | None) -> dict | None:
```
geomin12: can we call this "validate"? seems it insn't really parsing

geomin12 (approved):
lgtm, just one nit

build_tools/generate_therock_manifest.py
erman-gurses: Thanks! - Updated the function name

PR #7460, opened 2026-08-18
comm-libs: pretranslate RCCL gfx1250 hotswap cache

ScottTodd:
> [!NOTE]
> This is a fork of https://github.com/ROCm/TheRock/pull/7288, rebased onto latest `main` with https://github.com/ROCm/rocm-systems/pull/10037 already included in the rocm-systems submodule.

Original PR description:

* Adds a rccl-translate sub-project with local scripting that orchestrates the hotswap AOT translator against the librccl binary, producing a cache.
* Configures the rccl artifact to use the new hotswap kpack handler that properly slips the pre-generated cache into the existing rccl_lib_gfx1250 artifact, ensuring that it makes its way into all packaging modalities.
* The rocm-systems patch will land first and then the submodule bump will be removed from this.

PR #7505, opened 2026-08-20
fix(rccl): enable RocJitsu for staged pretranslation

ammarwa:
## Summary

- make the RCCL artifact explicitly depend on `rocjitsu-hotswap`
- package the RocJitsu pretranslation binary and driver in `rocjitsu_run`
- add regression coverage for both the staged feature set and runtime artifact payload

## Problem

The gfx1250 RCCL pretranslation integration from #7460 is conditional on both
`THEROCK_ENABLE_ROCJITSU_HOTSWAP` and `THEROCK_FLAG_KPACK_SPLIT_ARTIFACTS`.

In [RockRel run 32199453871](https://github.com/ROCm/rockrel/actions/runs/32199453871),
the `comm-libs` stage configured with split KPACKs enabled but both RocJitsu
features disabled. As a result, `rccl-translate` was excluded and the published
RCCL KPACK had no install-owned translation cache.

After enabling the RocJitsu projects, [RockRel run 32316930942](https://github.com/ROCm/rockrel/actions/runs/32316930942)
exposed the second staged-build boundary: `rocjitsu_run` contained only
`share/rocjitsu/**`, so the fetched artifact omitted `bin/rj_pretranslate` and
`bin/rocjitsu-pretranslate.py`. `rccl-translate` then failed at configure time
because `rj_pretranslate` was absent from the RocJitsu dist root.

## Fix

Declare `rocjitsu-hotswap` as an artifact dependency of RCCL. Its transitive
dependency on `rocjitsu` makes both artifacts inbound to staged `comm-libs`
builds, so `configure_stage.py` enables `ROCJITSU` and `ROCJITSU_HOTSWAP`.

Also include the installed pretranslation binary and Python driver in the
`rocjitsu` runtime artifact. The staged RocJitsu dist root now has the same two
entry points that the monolithic build exposes to `rccl-translate`.

## Tests

```text
uvx --system-certs pre-commit run --files \
  BUILD_TOPOLOGY.toml emulation/artifact-rocjitsu.toml \
  build_tools/tests/configure_stage_test.py build_tools/tests/artifacts_test.py
  PASS

python -m unittest \
  build_tools.tests.artifacts_test \
  build_tools.tests.configure_stage_test \
  build_tools.tests.build_topology_test
  Ran 66 tests: PASS (1 skipped)

python build_tools/configure_stage.py --stage comm-libs \
  --dist-amdgpu-families gfx125X-dcgpu --platform linux --oneline
  includes THEROCK_ENABLE_ROCJITSU=ON and THEROCK_ENABLE_ROCJITSU_HOTSWAP=ON
```


atgutier (approved):

ScottTodd (commented):
I can give a more detailed review in a bit. Please fix the test failures: https://github.com/ROCm/TheRock/actions/runs/32386927143/job/96483568182?pr=7505

build_tools/tests/emulation_cmake_test.py
ScottTodd, line 1: Can you remove this test? We don't currently test the CMake source files like this, see https://github.com/ROCm/TheRock/blob/main/TESTING.md#super-project-cmake-build---validation-methods

We are considering some tests for CMake functions but they would likely follow the style of `build_tools/tests/rocm_build_flags_test.py` in https://github.com/ROCm/TheRock/pull/6984, not mocking entire CMake functions.

build_tools/tests/artifacts_test.py
```diff
@@ -334,6 +334,28 @@ def testEmptyComponentMatchesEverything(self):
...
+            artifact_name="rocjitsu",
+        )
+        scanner = builder.ComponentScanner(self.temp_dir / "src", descriptor)
+
+        run_files = set()
+        for pattern_matcher in scanner.components["run"].basedir_contents.values():
+            run_files.update(relpath for relpath, _ in pattern_matcher.matches())
+
+        self.assertIn("bin/rj_pretranslate", run_files)
+        self.assertIn("bin/rocjitsu-pretranslate.py", run_files)
```
ScottTodd: This test fabricates the expected files so all it's testing is the toml include list, when we don't have similar change detector tests for other components, please remove.

This was also added in front of other generic tests (ordering should be generic --> specific).

build_tools/tests/configure_stage_test.py
```diff
@@ -76,6 +80,20 @@ def test_rocblas_resolves_to_blas(self):
...
+
+    def setUp(self):
+        self.topology = get_topology()
+
+    def test_comm_libs_enables_rocjitsu_hotswap_for_rccl_pretranslation(self):
+        """The staged RCCL build must include its pretranslation toolchain."""
+        features = get_stage_features(self.topology, "comm-libs", platform_name="linux")
+
+        self.assertIn("ROCJITSU", features)
+        self.assertIn("ROCJITSU_HOTSWAP", features)
```
ScottTodd: This is sort of a change detector test too. We already have coverage for the build topology itself.

We could follow the precedent here and add a test to `build_topology_test.py` instead:

https://github.com/ROCm/TheRock/blob/48d493e696d795a0aa49c8f39969345eef8f4257/build_tools/tests/build_topology_test.py#L1126-L1136

comm-libs/rccl-translate/CMakeLists.txt
```diff
@@ -36,6 +41,12 @@ find_program(ROCJITSU_PRETRANSLATE_DRIVER
   NO_DEFAULT_PATH
   REQUIRED
 )
```
ScottTodd, line 43: Not directly related to this PR I think but we should look at this code more closely. Review agent findings:

> [`comm-libs/CMakeLists.txt` lines 84-96](https://github.com/ROCm/TheRock/blob/393afa52658d0a02a8eca069648eadbcad8adae8/comm-libs/CMakeLists.txt#L84-L96)
> passes `-DROCJITSU_PRETRANSLATE_DRIVER=<rocm-systems source path>` and also
> fingerprints that source file. In the child project,
> [`find_program(ROCJITSU_PRETRANSLATE_DRIVER ...)`](https://github.com/ROCm/TheRock/blob/393afa52658d0a02a8eca069648eadbcad8adae8/comm-libs/rccl-translate/CMakeLists.txt#L38-L43)
> does not search `${ROCJITSU_ROOT}/bin` when the variable is already defined.
> That is [CMake's documented behavior](https://cmake.org/cmake/help/latest/command/find_program.html):
> a defined result variable whose value is not `NOTFOUND` suppresses the search.
> 
> The failing RockRel comm-libs log confirms the configure command passed the
> source path while only `rj_pretranslate` was searched in the fetched dist.
> Consequently, adding `bin/rocjitsu-pretranslate.py` to `rocjitsu_run` does not
> make `rccl-translate` consume the packaged driver. This matters when a stage is
> reused or otherwise sourced from a different commit: the tool/library can come
> from the inbound artifact while the driver comes from the current checkout.
> The resulting version skew defeats the staged artifact boundary this PR is
> intended to repair.
> 
> **Required action:** Remove the
> `-DROCJITSU_PRETRANSLATE_DRIVER=...` override and its matching
> `EXTRA_DEPENDS` entry so the existing `find_program(... PATHS
> "${ROCJITSU_ROOT}/bin" NO_DEFAULT_PATH REQUIRED)` resolves the driver from the
> inbound RocJitsu artifact. Then verify the configure log records the dist path.

>   One substantive caveat remains: comm-libs/CMakeLists.txt predefines ROCJITSU_PRETRANSLATE_DRIVER as a source-tree path.
>   Because the variable is already set, CMake’s find_program() does not search the newly packaged script in
>   ${ROCJITSU_ROOT}/bin. Thus:
> 
>   - A normal same-checkout build should work.
>   - The packaged Python driver is not actually used by rccl-translate.
>   - Reused artifacts could pair an older tool/library with the current source-tree driver.

ScottTodd (commented):
The test changes here don't follow project conventions (https://github.com/ROCm/TheRock/blob/main/TESTING.md). Code changes look correct enough to me so we could merge after revising the tests.

build_tools/tests/emulation_cmake_test.py
ammarwa: Removed

build_tools/tests/artifacts_test.py
ammarwa: Removed this change-detector test.

build_tools/tests/configure_stage_test.py
ammarwa: Removed this stage-feature change-detector test; existing topology coverage remains unchanged.

comm-libs/rccl-translate/CMakeLists.txt
ammarwa: I verified that RCCL cache freshness is preserved here: ${RCCL_LIBRARY} is a direct dependency of the translation custom command, and rccl-translate has rccl as a build dependency. A newly staged RCCL therefore reruns unbundling and pretranslation instead of reusing the prior translation.stamp.

The potential version skew between the source-tree Python driver and reused RocJitsu artifacts is a separate artifact-boundary concern. Removing the override would change production/stage-reuse behavior, so I would prefer handling that in a focused follow-up with explicit stage-reuse validation. This PR remains scoped to restoring the missing staged executable/translator dependencies and generating the cache from each newly built RCCL. Is that acceptable?

ScottTodd (approved):

PR #7510, opened 2026-08-20
feat(ci): enable ASAN presubmit for external repos without labels

geomin12:
rocm-systems and rocm-libraries were initially enabled on presubmit but this conditional caused host-asan to not run. However from discussions, 
- we decided to do [post-submit only for rocm-libraries](https://github.com/ROCm/rocm-libraries/pull/11262) to continue getting signal and able to revert if issue is found. 
    - once rocm-libraries optimizations on multi-arch come into play, we will enable presubmit
- rocm-systems will get presubmit coverage in order to quickly block bad work

Plan here: https://github.com/ROCm/TheRock/issues/6627

Workflow dispatch test run: https://github.com/ROCm/rocm-libraries/actions/runs/33098621491/job/98617005999
@main pin no longer there due to https://github.com/ROCm/rocm-libraries/commit/bdcc4149c853a77b366c58396838f541d72b7710

### DIFF
--- a/.github/workflows/setup_multi_arch.yml
+++ b/.github/workflows/setup_multi_arch.yml
@@ -44,6 +44,14 @@ on:
         description: "Workflow run ID to copy prebuilt stage artifacts from."
         type: string
         default: ""
+      baseline_repository:
+        description: >-
+          Repository to search for baseline runs during automatic stage reuse.
+          Defaults to the 'repository' input (THEROCK_REPOSITORY). External repos
+          can set this to their own repo (e.g., github.repository) for
+          workflow_dispatch events to reuse baselines from their own runs.
+        type: string
+        default: ""
       stage_reuse_mode:
         description: >-
           Artifact-reuse mode: "off" disables all reuse, ignores prebuilt_stages
@@ -211,6 +219,7 @@ jobs:
           WINDOWS_TEST_LABELS: ${{ inputs.windows_test_labels }}
           PREBUILT_STAGES: ${{ inputs.prebuilt_stages }}
           BASELINE_RUN_ID: ${{ inputs.baseline_run_id }}
+          BASELINE_REPOSITORY: ${{ inputs.baseline_repository }}
           STAGE_REUSE_MODE: ${{ inputs.stage_reuse_mode }}
           # The checkout commit is used for commit-compatibility checking.
           STAGE_REUSE_CURRENT_SHA: ${{ steps.checkout.outputs.commit }}
--- a/build_tools/github_actions/stage_reuse_decision.py
+++ b/build_tools/github_actions/stage_reuse_decision.py
@@ -548,10 +548,15 @@ def _default_baseline_selector(*, platform: str) -> BaselineSelector:
     extra "passing build" check is needed here.
     """
 
-    # THEROCK_REPOSITORY is set by setup_multi_arch.yml to the repository input
-    # (ROCm/TheRock for external repos, or github.repository for normal runs).
-    github_repository = os.environ.get(
-        "THEROCK_REPOSITORY", os.environ.get("GITHUB_REPOSITORY", "ROCm/TheRock")
+    # BASELINE_REPOSITORY overrides THEROCK_REPOSITORY for baseline lookups.
+    # This allows external repos to specify a different repository for baseline
+    # runs (e.g., their own repo for workflow_dispatch instead of TheRock).
+    #
+    # Priority: BASELINE_REPOSITORY > THEROCK_REPOSITORY > GITHUB_REPOSITORY
+    github_repository = (
+        os.environ.get("BASELINE_REPOSITORY")
+        or os.environ.get("THEROCK_REPOSITORY")
+        or os.environ.get("GITHUB_REPOSITORY", "ROCm/TheRock")
     )
     branch = os.environ.get("STAGE_REUSE_BASELINE_BRANCH", "main")
     workflow_name = os.environ.get("STAGE_REUSE_BASELINE_WORKFLOW", "multi_arch_ci.yml")


sa-faizal (approved):
Lgtm!

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -607,25 +607,32 @@ def should_skip_ci(
         return True
 
     # Skip ASAN on PRs unless an enabling label is present.
-    # This avoids running expensive ASAN builds on every PR.
+    # This avoids running expensive ASAN builds on every PR in TheRock.
+    # External repos (e.g., rocm-libraries) that create an ASAN workflow have
+    # explicitly opted in, so they don't require labels.
```
ScottTodd: Do we really want all external repos with the workflow to run ASan all the time on PRs? It's a pretty expensive workflow, especially for rocm-libraries.

How about controlling from some combination of:
* inputs to the `setup_multi_arch.yml` workflow: https://github.com/ROCm/rocm-libraries/blob/8d0205de4a31b788c4d80f76ebe28edc5950332d/.github/workflows/therock-multi-arch-ci-asan.yml#L55-L65 and https://github.com/ROCm/TheRock/blob/f438028b32ad093ad56709f3bd26ee02d958c477/.github/workflows/multi_arch_ci_asan.yml#L52-L63
    * For example, TheRock could pass `skip_on_pull_request_unless_label: "ci:asan"` there and rocm-libraries could leave that option off (with some support for `ci:asan` and `ci:host-asan` 🤔 )
    * Or `always_run_on_pull_request: true`
* a policy level for each subproject (https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0013-Consumer-Based-Test-Selection.md, https://github.com/ROCm/TheRock/blob/main/test_tools/test_policies.toml), with some projects running asan and others not

ScottTodd: I also see that https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci-asan.yml only has `pull_request` and `workflow_dispatch` labels. Do we also want `push`? Postsubmit signal is less expensive than presubmit signal and will be essential to spotting when commits break the build (without it you have to check the runs on PRs themselves, which may have used outdated base branches)

I wrote these docs for another project: https://iree.dev/developers/general/github-actions/#workflow-triggers, note that the decision there was to have `push` be a superset of `pull_request`.

build_tools/github_actions/configure_multi_arch_ci.py
```diff
@@ -607,25 +607,32 @@ def should_skip_ci(
...
     has_asan_label = (
         "ci:asan" in ci_inputs.pr_labels or "ci:host-asan" in ci_inputs.pr_labels
     )
+    is_asan_variant = ci_inputs.build_variant in ("asan", "host-asan")
     if (
         ci_inputs.is_pull_request
-        and ci_inputs.build_variant == "asan"
+        and is_asan_variant
         and not has_asan_label
+        and not ci_inputs.external_repo  # External repos opt-in by creating the workflow
```
ScottTodd: Is the ASan [presubmit] workflow _working_ in rocm-libraries? Can you add a test plan to this PR's description?

As currently implemented with `uses: ROCm/TheRock/.github/workflows/setup_multi_arch.yml@main` and `ref: main` in https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci-asan.yml, merging this PR would cause that workflow to _immediately_ start getting used in rocm-libraries, so we should first check that it works, what the cost will be, etc.

* I see only one `workflow_dispatch` event of https://github.com/ROCm/rocm-libraries/actions/workflows/therock-multi-arch-ci-asan.yml?query=event%3Aworkflow_dispatch that was cancelled
* There are only 9 PRs that added the `ci:asan` label: https://github.com/ROCm/rocm-libraries/pulls?q=is%3Apr+label%3Aci%3Aasan , and I'm not seeing the workflow actually running on any of them

ScottTodd (changes requested):
Please write down the test plan for this change.

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: true! i think in that case, we'll keep it label based until we get the optimization we want. 

i'll make sure asan label works and is sync-ed up with rocm-libraries / rocm-systems. i'll close this out

geomin12:
<s>labels work fine here https://github.com/ROCm/rocm-libraries/pull/11257

we'll do postsubmit for now, where postsubmit runs host-asan</s>

build_tools/github_actions/configure_multi_arch_ci.py
ScottTodd: Please link to at least one `workflow_dispatch` run of the asan workflow in relevant repositories. Even with https://github.com/ROCm/rocm-libraries/pull/11262, there are still _already open PRs_ that use the `on: pull_request:` code together with `@main` pins (https://github.com/ROCm/rocm-libraries/blob/ef3ce79db43c76161b28eca7fb4e886257e28f5a/.github/workflows/therock-multi-arch-ci-asan.yml#L54-L56) that will immediately pick this change up once merged.

I don't want this to be merged and immediately start running and possibly failing. Let's make sure:
1. The workflow has the behavior we expect
     * Then check the duration against our runner capacity / budgets
2. New PRs at least won't start running these tests (e.g. by removing the `@main` pin from https://github.com/ROCm/rocm-libraries/blob/develop/.github/workflows/therock-multi-arch-ci-asan.yml and merging https://github.com/ROCm/rocm-libraries/pull/11262 ahead of this)

ScottTodd (commented):
Marking as 'reviewed' after my latest comment

build_tools/github_actions/configure_multi_arch_ci.py
geomin12: i updated the description, but failure :( 

let me investigate

geomin12:
other engineers are resolving this. closing

PR #7517, opened 2026-08-20
Fetch multi-arch tarballs from CloudFront instead of S3

### LINKED ISSUE #2139: Refactor `install_rocm_from_artifacts.py` to retrieve tarballs from CloudFront
Tarball releases are now consistently available from

* https://rocm.devreleases.amd.com/tarball/
* https://rocm.nightlies.amd.com/tarball/
* https://rocm.prereleases.amd.com/tarball/

`build_tools/install_rocm_from_artifacts.py` should be refactored to fetch tarballs from CloudCront instead from the S3 buckets. This is required to enable fetching prerelease tarballs as the bucket does not provide public read access and files can only be accessed via CloudFront.

**It is not an option to grant public read access to the prerelease tarball bucket!**

IgorKasianenko:
Fixes #2139.

`install_rocm_from_artifacts.py` downloaded and listed multi-arch tarballs through anonymous S3 (`boto3` with unsigned requests against `therock-*-tarball`). That cannot reach the prerelease bucket, which grants no anonymous read access, so prerelease tarballs were unreachable from the script.

This change routes all tarball traffic through CloudFront:

- `S3BucketConfig` gains a `cdn_url` field, populated for the packages, python and tarball buckets of every release type. The CDN path segment is recorded per bucket because it does not match the S3 key prefix (`tarball-multi-arch/` on the CDN is `v4/tarball/` in the bucket).
- Downloads use `urllib` against the CDN URL. `boto3`, `botocore` and the unsigned S3 client are gone from this script's tarball path.
- Discovery reads the JSON array that `build_tools/index_generation_s3_tar.py` embeds in `index.html` for client-side filtering. CloudFront does not expose the S3 listing API, and `index.json` / `files.json` next to the tarball index return 403, so this is the only listing available over the CDN. Per [#2139 (comment)](https://github.com/ROCm/TheRock/issues/2139#issuecomment-5348023442), a machine-readable index is a future optimisation rather than a prerequisite.
- `--release` accepts prerelease versions of the form `X.Y.ZrcN`, and versions are classified into nightly / prerelease / dev to pick the right bucket. The nightly and prerelease `rc` forms are disjoint: a nightly `rc` is followed by an 8-digit date, a prerelease `rc` by a shorter counter.
- The `--release` help text that lists where to find a version derives its URLs from the bucket configs rather than repeating the CDN domains, so the domains live in one place.

No command-line arguments were added, removed or renamed, so existing callers keep working; `--release` accepts a wider set of values than before.

There is no opt-in S3 path. `docs/development/s3_buckets.md` notes that CI may read the buckets directly to avoid CloudFront transfer costs, but keeping that path would keep the credentialed code this issue asks to remove, and per [#2139 (comment)](https://github.com/ROCm/TheRock/issues/2139#issuecomment-5348023442) TheRock's own CI does not use this script for tarballs.

Tests: `build_tools/tests/install_rocm_from_artifacts_test.py` and `build_tools/tests/s3_buckets_test.py`, 68 passing locally. Network access is stubbed; the tests exercise index parsing, version classification, URL construction and the error paths. The index fixture is built by calling `index_generation_s3_tar.generate_index_s3` rather than by hand-writing HTML, so the parser stays tied to the page format that is actually published.

Not done here: splitting "download artifacts" and "download tarballs" into separate entry points, or delegating to `dockerfiles/install_rocm_tarball.sh`, as raised in [#2139 (comment)](https://github.com/ROCm/TheRock/issues/2139#issuecomment-5348023442). That reshapes the script's public surface and downstream users, which seems worth its own change. Would you like a follow-up issue for it?

---

Disclosure: this change and this text were written with an AI coding assistant. I have checked the code and verified every claim above myself, and I am accountable for both.


IgorKasianenko:
The checks on this PR are all sitting at `action_required` and have not started: Multi-Arch CI, Multi-Arch CI ASAN, Unit Tests, Pre commit security and CodeQL. `therock-pr-bot` polls for those required checks, so it timed out after 15 minutes and failed ([run](https://github.com/ROCm/TheRock/actions/runs/32361278593/job/96401174244)) with:

```
❌ Timed out waiting for required checks to complete.
```

Nothing in the change has been exercised yet as a result. @ScottTodd would you approve the workflow runs?

build_tools/_therock_utils/s3_buckets.py
```diff
@@ -33,6 +33,16 @@ class S3BucketConfig:
...
     """IAM role name that grants write access to this bucket (e.g. 'therock-ci'), if any"""
 
+    cdn_url: str | None = field(default=None)
+    """Base URL of the CloudFront distribution in front of this bucket, if any.
+
+    Downloading through the CDN needs no AWS credentials and is the only public
+    way to read buckets that do not grant anonymous S3 read access (such as the
+    prerelease buckets). Note that the CDN path segment does not always match
+    the S3 key prefix: 'tarball-multi-arch/' on the CDN maps to 'v4/tarball/'
+    in the bucket, for example. See docs/development/s3_buckets.md.
```
ScottTodd: We have a few open PRs touching similar code in this file and there is a big rework in progress for the actual URLs (https://github.com/ROCm/TheRock/blob/main/docs/rfcs/RFC0012-Repo-Structure.md). I'll defer to @marbre on the order he wants to review/merge changes:

* https://github.com/ROCm/TheRock/pull/7240
* https://github.com/ROCm/TheRock/pull/7547

IgorKasianenko: Thanks for the comment, feel free to cherry pick and or close and suggest future work after big changes land

ScottTodd: Looks like we'll actually need this together with https://github.com/ROCm/TheRock/pull/7547, as the new buckets do not allow direct access through S3. Let's see what Marius wants to do.

marbre: #7547 and doc updates have been merged now and @PeterCDMcLean was holding back with #7240 until that was landed. No strong feelings on the order of further work touching this.

build_tools/_therock_utils/s3_buckets.py
```diff
@@ -51,24 +61,72 @@ def write_access_iam_role(self) -> str | None:
...
+    S3BucketConfig(
+        "therock-dev-python",
+        iam_role="therock-dev",
+        cdn_url="https://rocm.devreleases.amd.com/whl-multi-arch",
+    ),
+    S3BucketConfig(
+        "therock-dev-tarball",
+        iam_role="therock-dev",
+        cdn_url="https://rocm.devreleases.amd.com/tarball-multi-arch",
+    ),
```
marbre: These are still existing but are now only be used for legacy. Nice if legacy support it but actually I wouldn't put too much work into it. Legacy at least partly allows public read (except for pre-releases and releases) but none of the new buckets allow read through S3 and here downloading via CF is the bigger win.

build_tools/install_rocm_from_artifacts.py
```diff
@@ -635,16 +707,16 @@ def retrieve_artifacts_by_release(args):
...
         log("Please retrieve the correct release version from:")
         log(
             "\t - https://rocm.nightlies.amd.com/tarball-multi-arch/ (nightly examples: 6.4.0rc20250416, 7.10.0a20251024)"
         )
+        log(
+            "\t - https://rocm.prereleases.amd.com/tarball-multi-arch/ (prerelease examples: 7.13.0rc2, 10.0.0rc0)"
+        )
         log(
             "\t - https://rocm.devreleases.amd.com/tarball-multi-arch/ (dev-tarball example: 6.4.0.dev0+8f6cdfc0d95845f4ca5a46de59d58894972a29a9)"
         )
```
marbre: Could also be used for stable if one wants to but this is not needed. Instead of hardcoding the domain name here, you could get it from s3_buckets.py instead.

marbre (commented):
Sorry for landing colliding work even though your PR was up earlier. The other work was a higher priority though but your work on this is apprechiated. Maybe @PeterCDMcLean  wants to take a look as well to align and avoid further doubled efforts.

build_tools/_therock_utils/s3_buckets.py
IgorKasianenko: Understood on legacy. That changes what this PR should be aiming at, and I could not find the answer in the code or the docs, so I would rather ask than guess.

What I see after #7547: final publication goes to `therock-repo-amd-<stream>-<product>` and is read through `https://<stream>.repo.amd.com/rocm/...`, while `install_rocm_from_artifacts.py` on `main` still resolves tarballs from `therock-<type>-tarball` under `tarball-multi-arch/`. So the script sits entirely on the legacy path today, and this PR as written only moves that legacy path from S3 to CloudFront. Retargeting it at the new product buckets is a different change, not a rename — the layout is `rocm/core/tarball`, and CloudFront there is the only public read rather than an optimisation.

Please let me know what is best product behaviour and what are recommended user flows: whether `install_rocm_from_artifacts.py` should keep serving legacy tarballs, and whether the credential-free download is wanted on the new buckets first.

I will merge `main` and resolve the conflicts in `s3_buckets.py` once the scope is settled, so the merge does not have to be redone.

build_tools/install_rocm_from_artifacts.py
IgorKasianenko: Agreed, that block should derive the domains instead of repeating them. #7240 derives release index URLs from `cdn_rules` and looks like the right home for it, so I will not duplicate the mapping here — if it lands first, I will build the message from whatever accessor it exposes.

PeterCDMcLean: I'm working on #7238 which adds cdn_rules. It's a big stack so it will take a while for me to refine and for it to review and land. It's also not a high priority so that will delay it was well. Just something to keep in mind

build_tools/_therock_utils/s3_buckets.py
marbre: I didn't touched `install_rocm_from_artifacts.py` yet as it would be duplicating your work and the mechanism to download via S3 no longer works. Downloading via CDN works for legacy as well as for the new structure, thus I think we can just drop downloading release tarballs via S3. Does that make sense to you?

IgorKasianenko:
Answering @marbre's question on `s3_buckets.py`, with a measurement that I think changes what this PR should target.

Yes, dropping the S3 download path makes sense, and this PR already does it: there is no `boto3` or `botocore` import left in `install_rocm_from_artifacts.py`, and there is no opt-in S3 fallback.

What I had wrong is *which* CDN paths to point at. I listed the published nightly tarball indexes today, 2026-08-27:

| Index | Versions present | Files |
| --- | --- | --- |
| `https://nightly.repo.amd.com/rocm/core/tarball/` | `10.1.0a20260823` … `10.1.0a20260827` | 310 |
| `https://rocm.nightlies.amd.com/tarball-multi-arch/` | up to `10.1.0a20260822` | 4540 |

The two sets do not overlap. I read that as publication having moved to the new location on 2026-08-23, though a retention window on the new index would look the same from outside; either way, neither location on its own covers every version a user might hold.

Two consequences:

1. This PR as written maps every bucket to its legacy `*-multi-arch` CDN path, so after merging it the script would still fail on any nightly from 2026-08-23 onwards.
2. `main` already has a version of this problem. `docs/development/installing_artifacts.md` sends users to `https://nightly.repo.amd.com/rocm/core/tarball/` to find a version string, while the code resolves `therock-nightly-tarball` under `v4/tarball/`. A version copied from the documented index cannot be downloaded by the script today.

Prerelease does not follow the same pattern. `https://rc.repo.amd.com/` returns a CloudFront `403 ERROR — Request blocked` at the root, so I could not find a working new-structure index for it, while `https://rocm.prereleases.amd.com/tarball-multi-arch/` serves 510 files up to `10.0.0rc4`. Since #2139 is about prerelease tarballs being unreachable without credentials, the legacy CDN path is the only one that fixes that today. Dev has both indexes serving (742 files on `dev.repo.amd.com`, 8026 on `rocm.devreleases.amd.com`), but dev versions are commit hashes, so I cannot tell from the listing alone whether one is behind.

My proposal is to widen this PR: resolve each release kind from `https://<stream>.repo.amd.com/rocm/core/tarball/` where that is serving, and keep the legacy `*-multi-arch` CDN as a fallback for versions that exist only there. Both are CloudFront, so no S3 either way. That keeps older version strings working and makes current nightlies work.

Please let me know what is best product behaviour and what are recommended user flows: whether that fallback is wanted, or whether the script should follow the cutover and stop resolving nightlies from before 2026-08-23, and whether prerelease should stay on `rocm.prereleases.amd.com` until `rc.repo.amd.com` serves.

On merging: `s3_buckets.py` now merges with `main` without conflicts. The only conflict left is in `docs/development/installing_artifacts.md`, which the doc updates alongside #7547 rewrote. I will resolve it as part of whichever scope you pick, so it does not have to be redone.

On CI: the required checks (Multi-Arch CI, Multi-Arch CI ASAN, Unit Tests, Pre commit security, CodeQL) are still sitting at `action_required` and have never run, so nothing in this change has been exercised yet. @ScottTodd would you approve the workflow runs?

build_tools/_therock_utils/s3_buckets.py
IgorKasianenko: Yes, that makes sense, and this PR already drops the S3 download path: no `boto3` or `botocore` left in the script, and no opt-in fallback.

The part I had wrong is which CDN paths to use. Measured today, the legacy nightly index stops at `10.1.0a20260822` and `https://nightly.repo.amd.com/rocm/core/tarball/` starts at `10.1.0a20260823`, so this PR as written would not resolve any current nightly. I have put the numbers and a scope proposal in a PR comment rather than bury them in this thread.

build_tools/install_rocm_from_artifacts.py
IgorKasianenko: Withdrawing what I said about waiting for `cdn_rules`. With #7238 being a long, low-priority stack, blocking this on it is the wrong call, and it turns out I do not need it: this PR already adds `cdn_url` to every release bucket config, and `_multiarch_tarball_index_base_url()` reads it. So the help text can loop over `RELEASE_KIND_TARBALL_BUCKETS` and print derived URLs, with no domain mapping duplicated in this file. I will make that change here. If #7238 later exposes a better accessor, swapping to it is a small edit.

IgorKasianenko:
Two things since Thursday.

Re-measuring the nightly indexes today: `https://rocm.nightlies.amd.com/tarball-multi-arch/` has published nothing since `10.1.0a20260822` — the same 4540 files as two days ago — while `https://nightly.repo.amd.com/rocm/core/tarball/` is up to `10.1.0a20260829` and growing daily. So the legacy path this PR currently targets is frozen, and the script cannot reach any nightly from the last week. That settles the direction for me: the new `{stream}.repo.amd.com/rocm/core/tarball/` layout has to be the primary, which is the same move #7576 makes for packages.

What I still cannot answer from the code or the docs is whether to keep the legacy CDN as a fallback for the 4540 older versions, and what to do about prerelease: `https://rc.repo.amd.com/` still returns a CloudFront `403 — Request blocked`, including `/rocm/core/packages/ubuntu2404`, so there is no new-structure prerelease index to point at yet. Please let me know what is best product behaviour and what are recommended user flows.

Separately, the `pre-commit` check that `therock-pr-bot` requires has never run on this PR — all six workflows sit at `action_required`, so the bot times out waiting for a check that was never created. Run locally over this branch, `pre-commit` passes: every hook passed or had no matching files. @marbre @ScottTodd would one of you approve the workflow runs?

build_tools/install_rocm_from_artifacts.py
IgorKasianenko: Done in 6399b931d, without waiting for `cdn_rules`. `release_version_source_lines()` builds each line from `_multiarch_tarball_index_base_url(bucket)` over `RELEASE_KIND_TARBALL_BUCKETS`, so the three domains are no longer written in this file. The example version strings stay in their own table, and a test asserts each one classifies as the release kind it is listed under. Rendered output is unchanged apart from "dev-tarball example" becoming "dev example".

build_tools/_therock_utils/s3_buckets.py
```diff
@@ -53,24 +63,72 @@ def write_access_iam_role(self) -> str | None:
     S3BucketConfig("therock-bkc-artifacts", iam_role="therock-bkc"),
     # Release type "dev"
     S3BucketConfig("therock-dev-artifacts", iam_role="therock-dev"),
-    S3BucketConfig("therock-dev-packages", iam_role="therock-dev"),
-    S3BucketConfig("therock-dev-python", iam_role="therock-dev"),
-    S3BucketConfig("therock-dev-tarball", iam_role="therock-dev"),
+    S3BucketConfig(
```
marbre: I don't think we should bind the CDN URL to the S3BucketConfig as this does not provide much value. We rather want to introduce a function like https://github.com/ROCm/TheRock/blob/82ad6d3ba3d941ae89ccc6462f69f17865e3ad2a/build_tools/_therock_utils/s3_buckets.py#L197-L222

which get's you the correct CDN and, depending on the version number, the correct path. Elaboration more in my review feedback.

marbre (commented):
> My proposal is to widen this PR: resolve each release kind from https://<stream>.repo.amd.com/rocm/core/tarball/ where that is serving, and keep the legacy *-multi-arch CDN as a fallback for versions that exist only there. Both are CloudFront, so no S3 either way. That keeps older version strings working and makes current nightlies work.
> 
> Please let me know what is best product behaviour and what are recommended user flows: whether that fallback is wanted, or whether the script should follow the cutover and stop resolving nightlies from before 2026-08-23, and whether prerelease should stay on rocm.prereleases.amd.com until rc.repo.amd.com serves.

I think we should route everything via the CDN and, as mentioned before, simplify by dropping downloading via S3 access. From a code perspective we can route depending on the versions that are requested but I don't think the current changes to `s3_bucket_configs` is what we want. This binds the resolution to a bucket whereas it isn't needed any longer to know the bucket. It is fine to always go via the CDN.

For nightlies it is okay to only serve the new ones from nightly.repo.amd.com, same for  dev releases (if @scotttodd does not disagree). We can just cut off legacy here. For prereleases we need to serve them depending on the version number. Everything < 10.1 will get served from https://rocm.prereleases.amd.com/, newer version will come from https://rc.repo.amd.com/. Similar situation for stable, where ROCm 7.13 and 7.14 are served from https://repo.amd.com/rocm/tarball-multi-arch/ and ROCm 10.0+ are in https://stable.repo.amd.com. I think we don't need to make the script too complicated and we can drop support for 7.9-7.13 non-multi-arch which are in https://repo.amd.com/rocm/tarball/ (@scotttodd?).

---

> Two things since Thursday.
> 
> Re-measuring the nightly indexes today: `https://rocm.nightlies.amd.com/tarball-multi-arch/` has published nothing since `10.1.0a20260822` — the same 4540 files as two days ago — while `https://nightly.repo.amd.com/rocm/core/tarball/` is up to `10.1.0a20260829` and growing daily. So the legacy path this PR currently targets is frozen, and the script cannot reach any nightly from the last week. That settles the direction for me: the new `{stream}.repo.amd.com/rocm/core/tarball/` layout has to be the primary, which is the same move #7576 makes for packages.

Legacy nightly and dev is indeed frozen. Legacy will only be used for pre-ROCm 10 patch-releases (release candidates and stable releases).

> What I still cannot answer from the code or the docs is whether to keep the legacy CDN as a fallback for the 4540 older versions, and what to do about prerelease: `https://rc.repo.amd.com/` still returns a CloudFront `403 — Request blocked`, including `/rocm/core/packages/ubuntu2404`, so there is no new-structure prerelease index to point at yet. Please let me know what is best product behaviour and what are recommended user flows.

Let me point you to https://github.com/ROCm/TheRock/blob/main/RELEASES.md#installing-multi-arch-releases:

> The repo.amd.com layout documented below starts with ROCm 10.1 nightly releases and ROCm 10.0 stable releases. Earlier releases remain in the [legacy multi-arch release locations](https://github.com/ROCm/TheRock/blob/main/docs/packaging/legacy_multi_arch_releases.md). In particular, ROCm 7.14 nightlies and the ROCm 10.0.0 release candidates are only available from the legacy indexes.

For ROCm 10 and below, all pre-releases are in the old location. For ROCm 10.1, you will find pre-releases at https://rc.repo.amd.com/ but there is no content yet and therefore we it isn't full live.

> Separately, the `pre-commit` check that `therock-pr-bot` requires has never run on this PR — all six workflows sit at `action_required`, so the bot times out waiting for a check that was never created. Run locally over this branch, `pre-commit` passes: every hook passed or had no matching files. @marbre @ScottTodd would one of you approve the workflow runs?

No worries about the PR bot, this isn't blocking.

IgorKasianenko:
Thanks — that settles the direction. I will drop `cdn_url` from `S3BucketConfig` and the twelve bucket entries, and add a resolver in `s3_buckets.py` shaped like `get_product_release_bucket_config`, taking the release kind and the version string and returning the index base URL. Routing as you set it out: nightly and dev always from the new layout, prerelease split at 10.1, stable split at 10.0.

Two things that make the rework smaller than it looks. I checked the indexes today:

| Index | State |
| --- | --- |
| `nightly.repo.amd.com/rocm/core/tarball/` | 620 files, up to `10.1.0a20260901` |
| `dev.repo.amd.com/rocm/core/tarball/` | 1096 files, `10.1.0.dev0+<hash>` |
| `stable.repo.amd.com/rocm/core/tarball/` | 50 files, `10.0.0` only |
| `rc.repo.amd.com/rocm/core/tarball/` | 403, CloudFront `Request blocked` |
| `rocm.prereleases.amd.com/tarball-multi-arch/` | 510 files: 7.13.0rc0–rc2, 7.14.0rc0–rc3, 7.14.1rc0, 10.0.0rc0/rc2/rc3/rc4 |
| `repo.amd.com/rocm/tarball-multi-arch/` | 120 files: 7.13.0, 7.14.0, 7.14.1 |

The legacy contents match your cutoffs exactly: every legacy prerelease is below 10.1, and legacy stable is 7.13 and 7.14 only. And the new indexes embed the same `const files = [{"name": ..., "mtime": ...}]` array as the legacy ones, with the same flat `therock-dist-<platform>-<family>-<version>.tar.gz` names, so the index parsing and sorting in this PR carry over unchanged — only the base URL moves. A `HEAD` on `therock-dist-linux-gfx110X-all-10.1.0a20260901.tar.gz` returns 200 for 2.6 GB without credentials.

Open decisions before I push the rework:

1. Stable in scope? `install_rocm_from_artifacts.py` today accepts only nightly and dev versions for `--release`. Your routing table covers stable and prerelease as well, which widens what the flag accepts. Should this PR add both, or add prerelease now and leave stable to a follow-up?
2. `rc.repo.amd.com` returns 403 on every path I tried, including the root and `/rocm/core/tarball/index.html` — the CloudFront `Request blocked` page rather than a 404, so the distribution is blocked rather than empty. That means the 10.1-and-newer prerelease branch cannot be checked against a live index by anyone here. I would write it and cover it with unit tests, and it stays unverified until the distribution opens. Is that the right call, or should the script refuse 10.1+ prereleases with a message pointing at the legacy index until then?
3. @PeterCDMcLean's #7238 adds `cdn_rules`, which is the same data-driven CDN mapping this resolver needs. Building it by hand here may collide again. Please let me know what is best product behaviour and what are recommended user flows: whether to add a narrow tarball-only resolver now and let #7238 absorb it, or wait for the `cdn_rules` accessor.

One consequence worth recording: `.github/workflows/test_linux_jax_wheels.yml` calls this script with `--release`, so after the cutover that workflow resolves ROCm 10.1 and newer nightlies only.

On CI — understood that the PR bot is not blocking. The six real workflows on `6399b93` are all at `action_required`, including `Unit Tests`, so the test changes here have no CI signal yet. Locally the 68 tests in `install_rocm_from_artifacts_test.py` and `s3_buckets_test.py` pass. Should I finish all the work on the branch before triggering workflow runs?

IgorKasianenko:
Pushed `d43885503`, which does the rework you asked for.

`cdn_url` is gone from `S3BucketConfig`, and `s3_buckets.py` is now identical to `main` apart from the additions. In its place there is `get_release_tarball_index_url(release_type, version=None)`, next to `get_product_release_bucket_config`, which maps a release type and version to the index that publishes its tarballs. The bucket no longer takes part in resolving a download. The download-side stream mapping extends the upload one with `release` -> `stable`, so nothing on the upload path changes. `install_rocm_from_artifacts.py` passes index URLs, not bucket configs.

The cutoffs match what the indexes hold today, not only the wording in RELEASES.md: the new nightly and dev indexes carry `10.1` versions only, and the new stable index `10.0` only. A nightly or dev version below 10.1 is now refused with a pointer to `docs/packaging/legacy_multi_arch_releases.md`, rather than resolved to a URL that cannot serve it.

One defect came out of the verification, and I think it belongs in this PR rather than a follow-up, since this PR introduces it. Dev tarball names carry a `+` before the commit hash, and the CDN returns 404 for the unencoded character:

```
HEAD https://dev.repo.amd.com/rocm/core/tarball/therock-dist-linux-gfx110X-all-10.1.0.dev0+0d8328481ea99949822854dad92fbbd992c3a137.tar.gz    -> 404
HEAD https://dev.repo.amd.com/rocm/core/tarball/therock-dist-linux-gfx110X-all-10.1.0.dev0%2B0d8328481ea99949822854dad92fbbd992c3a137.tar.gz  -> 200, 2595637344 bytes
```

`https://rocm.devreleases.amd.com/tarball-multi-arch/` behaves the same way, so it is not specific to the new layout. Downloads through boto3 were unaffected because they addressed the S3 key rather than a URL, which is why this only appears once the download moves to the CDN. Every `--release <dev version>` would have failed without it. The asset name is now percent-encoded when the URL is built, with a regression test. It adds one line of behaviour and no new scope.

Two things are still gated on the questions above, so they are not in this push:

- accepting stable versions in `--release`. The resolver handles `release` already, including the 7.13/7.14 to 10.0 split, but the CLI still classifies only nightly, prerelease and dev. Wiring it in is a small separate commit if you want stable in this PR.
- what to do while `rc.repo.amd.com` is dark. A 10.1 or newer prerelease currently resolves to it and would 404. Refusing those with a message is a few lines, once you decide which behaviour you want.

Verification, for what it is worth without CI: 74 unit tests pass locally, `pre-commit` passes over the changed files, and I checked each resolved base with a `HEAD` for a real tarball — nightly, dev and stable in the product layout, and the legacy prerelease and stable indexes — all 200. `rc.repo.amd.com` returns 403 on every path, so that branch of the routing is covered by unit tests only.

